PR #796, opened 2015-09-23
V8 'stable' branch

jeroen:
Much software that depends on V8 today actually depends on version 3.14 ~ 3.15. This is also the 'stable' version in homebrew and most Linux distributions:
- Debian: https://packages.debian.org/sid/libv8-dev
- Fedora: https://apps.fedoraproject.org/packages/v8-devel
- Arch https://aur.archlinux.org/packages/v8-3.15/
- Homebrew: https://github.com/Homebrew/homebrew-versions/blob/master/v8-315.rb

The 3.15 branch is fully compatible with 3.14, but not with more recent versions.

### DIFF
--- a/mingw-w64-v8-315/PKGBUILD
+++ b/mingw-w64-v8-315/PKGBUILD
@@ -26,8 +26,9 @@ prepare() {
 
 build() {
   local BUILDTYPE=Release
+  local BUILDDIR="${srcdir}/build-${CARCH}"
   export PYTHON=/usr/bin/python2
-  mkdir -p "${srcdir}/build-${CARCH}"
+  mkdir -p "${BUILDDIR}"
 
   case ${CARCH} in
     i686)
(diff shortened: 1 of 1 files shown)


Alexpux:
I think you can have it locally yourself. Don't want waste my time for building multiple versions of different software

jeroenooms:
Yes I understand. This is a bit of a strange case though, because the latest version of v8 is kind of useless (the API changes very rapidly with no backward compatibility). All software uses the 3.14 / 3.15 stable branch. 

I don't think this recipe will need any further maintenance, it is pretty final. Could you make an exception?

mingwandroid:
@jeroenooms : do you still want us to merge this? should it be in preference to packaging any other versions of v8?

@Alexpux : maybe we can consider the PKGBUILD variable of not_packaged_rationale="some-reason" thing we mentioned in some other pull request already, I've not got the link at present, it's late and I'm tired ;-) maybe @elieux can remember 

Good night guys ..

jeroenooms:
Yes I would still really like that. Most applications that depend on libv8 require the 3.15 version. It would be really nice if this version can easily be found in msys2.

jeroenooms:
Not sure if it is relevant, but homebrew has a special [homebrew-versions](https://github.com/homebrew/homebrew-versions) repository where they maintain commonly needed stable branches of software.

PR #797, opened 2015-09-23
Added mingw-w64-usbview.

DavidEGrayson:
This pull request adds a package for usbview, a sample GUI application from Microsoft that shows interesting information about connected USB controllers and devices.

Three caveats:

* It uses an older ZIP file from Microsoft's website because the [latest one on github](https://github.com/Microsoft/Windows-driver-samples/tree/master/usb/usbview) cannot yet be compiled with mingw-w64 because of some missing structs/typedefs.  I will probably submit patches to mingw-w64 so we can eventually compile it.
* It compiles with warnings because of macro redefinitions and arrays that are assumed to have one element.
* xmlhelper.cpp uses .NET, so this package replaces it with a dummy file that just shows an error message.  This build will not be able to save data as an XML file.

If either of these reasons make you not want to merge this in, I would understand.

### DIFF
--- a/mingw-w64-usbview/PKGBUILD
+++ b/mingw-w64-usbview/PKGBUILD
@@ -1,0 +1,79 @@
+# Maintainer: David Grayson <davidegrayson@gmail.com>
+
+# Note: We currently build an older version of usbview from a ZIP file
+# that was on a page that Microsoft has since unpublished.  We are not
+# yet able to build the version of usbview from github because
+# mingw-w64 is missing some new structs/typedefs.
+#
+# https://github.com/Microsoft/Windows-driver-samples/tree/master/usb/usbview
+
+_realname=usbview
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.1.20140402
+pkgrel=1
+pkgdesc='GUI for browsing all USB controllers and connected USB devices on your computer (mingw-w64)'
+arch=('any')
+url='https://web.archive.org/web/20140824072648/http://code.msdn.microsoft.com/windowshardware/USBView-sample-application-e3241039'
+license=('custom')
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+  'patch'
+)
+depends=()
+options=('strip')
+source=(
+  "https://code.msdn.microsoft.com/windowshardware/USBView-sample-application-e3241039/file/51333/38/USBView%20sample%20application.zip"
+  'my_xmlhelper.c'
+  '01-strsafe-after-tchar.patch'
+  '02-sal.patch'
+  '03-resource-fix.patch'
+)
+
+sha256sums=('fd07e8b20f9cff7efd8c01bf2b9cc9ac39ba85c76e56287b80d8abf38ec92287'
+            'b3d551683be0398bca58e67bf7c91b22d88283e10292c9cda5ab367921511ebf'
+            'dd0dade894b612cc0d30b0cbe1c47647de3697c9feeff65b866c32b01430154e'
+            'b6959253bde6fd85af82cdd74c0993d0e9c82871f95cb507811b1fd8528cd301'
+            '2be3167e95f8d5d8abf91e43a49fac4593b8264df5c20478f476bdbe1bae89a3')
+
+prepare() {
+  cd "${srcdir}/C++"
+
+  # dos2unix -q *.c *.h *.rc
+
+  # strsafe.h must be after tchar.h in mingw-w64
+  patch -p1 -i "../01-strsafe-after-tchar.patch"
+
+  # Include sal.h in some places it is needed.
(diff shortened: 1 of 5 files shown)


mingw-w64-usbview/PKGBUILD
```diff
@@ -0,0 +1,80 @@
+# Maintainer: David Grayson <davidegrayson@gmail.com>
+
+# Note: We currently build an older version of usbview from a ZIP file
+# that was on a page that Microsoft has since unpublished.  We are not
+# yet able to build the version of usbview from github because
+# mingw-w64 is missing some new structs/typedefs.
+#
+# https://github.com/Microsoft/Windows-driver-samples/tree/master/usb/usbview
+
+_realname=usbview
```
Alexpux: Add line:
````pkgbase=mingw-w64-${_realname}````

File: mingw-w64-usbview/PKGBUILD [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+  "https://code.msdn.microsoft.com/windowshardware/USBView-sample-application-e3241039/file/51333/38/USBView%20sample%20application.zip"
+  'my_xmlhelper.c'
+  '01-strsafe-after-tchar.patch'
+  '02-sal.patch'
+  '03-resource-fix.patch'
+)
+
+sha256sums=(
+  'fd07e8b20f9cff7efd8c01bf2b9cc9ac39ba85c76e56287b80d8abf38ec92287'
+  'SKIP'
```
Alexpux: Run ````updpkgsums```` to generate checksums

mingw-w64-usbview/PKGBUILD
DavidEGrayson: I'm not totally sure what you mean, but I think you want a blank line after line 10.  I can do that.

Alexpux: Hmm my message was wrong.
Add line

DavidEGrayson:
OK, I added pkgbase and patch checksums, and amended the commit in the pull request.  Thanks!

DavidEGrayson:
@Alexpux, you said you couldn't build usbview?  If you tell me the error message you are getting I could try to help.  I was able to build it fine on two different machines, but maybe I accidentally installed a better mingw-w64 header on both of those machines or something.

Alexpux:
https://pastebin.com/6kSVTN8x

DavidEGrayson:
Line 39 of `uvcview.h` is `#include <usb.h>`.  On my system, I think that include resolves to `/mingw64/x86_64-w64-mingw32/include/usb.h`, which contains the typedef for `USBD_STATUS`.  Line 40 of `uvcview.h` includes `usbuser.h`, which uses the typedef.

My guess is that @Alexpux has a different `usb.h` that is taking precedence.  Do you have a `usb.h` in `/mingw64/include/usb.h`?

Alexpux:
Yeah, I have from libusb-compat

Alexpux:
So seems we have conflict

DavidEGrayson:
Have you had conflicts like this before, and how were they resolved?  I guess I could patch usbview to include the full path, or I could try using a `-I` option so it searches for the mingw-w64 headers first.

Alexpux:
need to think

DavidEGrayson:
The `-I` option does not work because of a quirk of GCC's [search path](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html):

> You can add to this list with the `-Idir` command-line option. All the directories named by `-I` are searched, in left-to-right order, before the default directories. **The only exception is when dir is already searched by default.** In this case, the option is ignored and the search order for system directories remains unchanged.

DavidEGrayson:
In the PKGBUILD, I think I can copy `usb.h` "${srcdir}/include" and then use the `-I` option to make it take precedence.

Alexpux:
ok

DavidEGrayson:
OK, I amended the pull request with this change:

~~~~{.patch}
--- a/mingw-w64-usbview/PKGBUILD
+++ b/mingw-w64-usbview/PKGBUILD
@@ -57,12 +57,16 @@ prepare() {

 build() {
   cd "${srcdir}"
-  mkdir -p "build-${MINGW_CHOST}"
+  mkdir -p "build-${MINGW_CHOST}/include"
   cd "build-${MINGW_CHOST}"

+  # Make sure mingw-w64's usb.h takes precedence over libusb-compat's.
+  cp ${MINGW_PREFIX}/${MINGW_CHOST}/include/usb.h include
+
   windres ../C++/uvcview.rc rc.so

   gcc -mwindows --std=c99 ${CFLAGS} ${LDFLAGS} \
+    -Iinclude \
     -DNTDDI_VERSION=0x06020000 -D_WIN32_WINNT=0x0602 \
     -DSTRSAFE_NO_DEPRECATE \
     ../C++/*.c rc.so \
~~~~

DavidEGrayson:
Oh, you already merged.  I'll make a new pull request.

PR #800, opened 2015-09-24
Adds mingw-w64-devcon-git

DavidEGrayson:
I think it's pretty cool that we can compile these Microsoft examples using mingw-w64.  (I just had to submit about 4 patches to make it work.)


DavidEGrayson:
Should I add `provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")` to this package?  The `-git` seems like an annoying detail and I don't want people to need to type that when they are installing the package.

DavidEGrayson:
OK, the pull request is good now.

mingw-w64-devcon-git/PKGBUILD
```diff
@@ -0,0 +1,70 @@
+# Maintainer: David Grayson <davidegrayson@gmail.com>
+#
+# You might need to run your shell as administrator for devcon to work.
+
+_realname=devcon
+pkgbase=mingw-w64-${_realname}-git
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+provides="${MINGW_PACKAGE_PREFIX}-${_realname}"
```
Alexpux: You also need ````conflicts```` entry here

DavidEGrayson:
OK, I amended the commit again, added a `conflicts` line.

PR #807, opened 2015-10-01
Fix antiword by making it relocatable

dscho:
Unfortunately, I had tested antiword only lightly before it was submitted. A very important part of `antiword` is to handle encodings, and those mapping files are in the `share/` directory. Sadly, antiword had no code to make itself relocatable, therefore I taught it this new trick.


dscho:
Thanks!

dscho:
Any word when a patched package will be available?

Alexpux:
today later.

dscho:
Thank you!

PR #809, opened 2015-10-06
libass: Update to 0.13.0

rossy:
This adds DirectWrite font selection, so fontconfig is now only needed for Windows XP. Also, the enca dependency was dropped and the PKGBUILD was changed to use the release tarball (Arch did this as well.)

### DIFF
--- a/mingw-w64-libass/PKGBUILD
+++ b/mingw-w64-libass/PKGBUILD
@@ -1,8 +1,9 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
 _realname=libass
+pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.12.3
+pkgver=0.13.0
 pkgrel=1
 pkgdesc="A portable library for SSA/ASS subtitles rendering (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mingw-w64-libass/PKGBUILD
```diff
@@ -38,7 +32,6 @@ build() {
     --enable-static \
     --enable-harfbuzz \
     --enable-fontconfig \
-    --enable-enca \
```
Alexpux: Maybe need directly disable it? Pass: ````--disable-enca````

mingw-w64-libass/PKGBUILD
```diff
@@ -2,7 +2,7 @@
 
 _realname=libass
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
```
Alexpux: Add line:
````pkgbase=mingw-w64-${_realname}````
before this line

mingw-w64-libass/PKGBUILD
rossy: Sorry, I meant to say that libass dropped enca support. See libass/libass@3d09c60

Alexpux: ok

mingw-w64-libass/PKGBUILD
rossy: Will do

PR #812, opened 2015-10-06
gitg: downgrade to version 0.2.7.

KrullKorg:
It builds but it doesn't work very well.

### DIFF
--- a/mingw-w64-gitg/001-win.patch
+++ b/mingw-w64-gitg/001-win.patch
@@ -1,0 +1,148 @@
+--- ./configure.ac.orig	2013-04-05 13:23:46.000000000 +0200
++++ ./configure.ac	2015-10-06 11:50:58.567044400 +0200
+@@ -65,7 +65,7 @@
+ 	gobject-2.0 >= $GLIB_REQUIRED_VERSION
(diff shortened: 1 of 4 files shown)


DavidEGrayson:
I couldn't even build the existing gitg PKGBUILD because I don't know where `mingw-w64-x86_64-webkit2gtk3` comes from.

Alexpux:
There are no working webkit2 package for GTK. We have ````mingw-w64-{i686,x86_64}-webkitgtk3```` packages but it contain only webkit1 as webkit2 not buildable for Windows

mingwandroid:
webkit2 would be a great task for someone to step up and hack on :-)

I've not got enough time unfortunately.

KrullKorg:
I downgrade gitg because version 0.2.7 doesn't depend on webkit2gtk3; i'll try ASAP to make it more usable under win

KrullKorg:
could you merge it?

mingwandroid:
How well is it working now?

KrullKorg:
not very well (for example sometime it shows the history and sometimes not) but it builds; 0.3.3 doesn't and it cannot be build until webkit2gtk3 is buildable

Issue #816, opened 2015-10-09
Strange Path Prefix `C:/building/`


yfwz100:
When I used `cmake` to compile PCL, I found that there're several errors in some config files. The path prefix `C:/building/` seems to be hard coded in many config files, like `$MSYS_ROOT/mingw64/bin/freetype-config`.

DavidEGrayson:
I can confirm that /mingw64/bin/freetype-config has invalid absolute Windows paths in it.  This is a general problem with several MSYS2 packages, and I wrote about it here: https://github.com/Alexpux/MSYS2-pacman/issues/16

You can help by fixing the [PKGBUILD script for freetype](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-freetype/PKGBUILD) and submitting a pull request here.  If you can't fix the problem at its root, you could at least use `sed` in the `build` function to fix the invalid paths.

yfwz100:
Can I used 
```
sed -e "s|$(cygpath -m /)|/|g" `grep -rl $pkgdir`
```
to replace all absolute path?

DavidEGrayson:
Yes, that should work.  It's a bit odd to use grep to find files, but I guess it works.  I might have used `find -type f $pkgdir` instead.

yfwz100:
Ok, I just send a PR. The fix is a little strange though. I hope a better solution could be applied. I have rebuilt several packages manually because of this problem.

DavidEGrayson:
Thanks for the PR.  Sorry, I just realized it's probably not a good idea to run sed on the binary files in a package.

yfwz100:
Maybe we should restrict the replacement to a single file `$MSYS_ROOT/mingw64/bin/freetype-config`. It seems that all configuration is related to this file. After I modified this file, everything goes well.

DavidEGrayson:
Yeah, if only one file needs to be edited, that would be fine.

PR #817, opened 2015-10-10
freetype: Fix hard coded path prefix.

yfwz100:
Seems to fix issue #816.


Alexpux:
For future, don't replace paths in **ALL** files because replacing strings in libraries and executables by **SED** break them

yfwz100:
Got it. :)

PR #824, opened 2015-10-15
New package: confuse

DavidEGrayson:
Thanks to some recent pull requests of mine, this library can be compiled and tested totally fine on MSYS2.  I'm interested in adding an MSYS2 package for it because it is an optional dependency of libftdi, which is an optional dependency of avrdude.

I'm not sure why, but the arch package name is [confuse](https://www.archlinux.org/packages/community/x86_64/confuse/) instead of "libconfuse", so I copied their naming.

### DIFF
--- a/mingw-w64-confuse/PKGBUILD
+++ b/mingw-w64-confuse/PKGBUILD
@@ -1,0 +1,49 @@
+# Maintainer: David Grayson <davidegrayson@gmail.com>
+
+_realname=confuse
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=2.8
+pkgrel=1
+pkgdesc='Library for parsing configuration files (mingw-w64)'
+arch=('any')
+url="https://github.com/martinh/libconfuse"
+license=('custom')
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
+depends=("${MINGW_PACKAGE_PREFIX}-gettext")
+options=('staticlibs' 'strip')
+source=(
+  "https://github.com/martinh/libconfuse/releases/download/v${pkgver}/confuse-${pkgver}.tar.xz"
+)
+
+sha256sums=('2a8102bfa3ccc846c14d94a81b0abfb4f5e855809f89ff3722aca1a9f314ea0d')
+
+build() {
+  rm -rf "${srcdir}/build-${MINGW_CHOST}"
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}"
+  cd "${srcdir}/build-${MINGW_CHOST}"
+  "../${_realname}-${pkgver}/configure" \
+    --prefix="${MINGW_PREFIX}" \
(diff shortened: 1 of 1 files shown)


mingw-w64-confuse/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+build() {
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}"
+  cd "${srcdir}/build-${MINGW_CHOST}"
+  "../${_realname}-${pkgver}/configure" \
+    --prefix="${MINGW_PREFIX}" \
+    --build="${MINGW_CHOST}" \
+    --host="${MINGW_CHOST}" \
+    --enable-static \
+    --enable-shared \
+    --disable-examples \
```
Alexpux: Remove slash as there are no more configure options after it.

mingw-w64-confuse/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+)
+options=('staticlibs' 'strip')
+source=(
+  "https://github.com/martinh/libconfuse/releases/download/v${pkgver}/confuse-${pkgver}.tar.xz"
+)
+
+sha256sums=('2a8102bfa3ccc846c14d94a81b0abfb4f5e855809f89ff3722aca1a9f314ea0d')
+
+build() {
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}"
```
Alexpux: Add before this line:
[[ -d build-${MINGW_CHOST} ]] && rm -rf build-${MINGW_CHOST}

mingw-w64-confuse/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+_realname=confuse
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=2.8
+pkgrel=1
+pkgdesc='Library for parsing configuration files (mingw-w64)'
+arch=('any')
+url="https://github.com/martinh/libconfuse"
+license=('custom')
+makedepends=(
```
Alexpux: If you have only one dependency - write all in one line
makedepends=(...)

mingw-w64-confuse/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+)
+depends=(
```
Alexpux: the same as previous

mingw-w64-confuse/PKGBUILD
DavidEGrayson: If you don't mind, I'll just do the `rm -rf ...` command.  The `-f` option already makes it OK to run that command when the directory does not exist.

Alexpux: ok

DavidEGrayson:
Thanks for the quick response, @AlexPux!  I have made those 4 changes to the PKGBUILD and force-updated the pull request.

PR #830, opened 2015-10-17
gdk-pixbuf2: introspection patch

bpisoj:
This fix utf8 support functions in typelib, without that pixbuf2 can only work with paths which have ascii characters


mingwandroid:
Thanks!

PR #838, opened 2015-10-20
python-setuptools: Update to 18.4.

DavidEGrayson:
Disclaimer:  I don't fully understand how python-setuptools works and I am having some weird issues where pip thinks that version 18.3.2 is installed, even though I installed my 18.4 package.  However, this pull request only changes a version number and a checksum, and I think it will help out users who try to install pips using the `-U` option, e.g. `pip install -U cryptography`.  I was discussing this problem here:

https://github.com/ARMmbed/yotta_windows_installer/issues/13#issuecomment-149430588


Alexpux:
As I know you need rebuild PIP when you upgrade SETUPTOOLS

DavidEGrayson:
Thanks, @AlexPux!  I wonder if there is something we could do short of rebuilding the pip package, like running a pip command to update its list of packages or something.

Alexpux:
I'm not python user at all, so don't know

PR #867, opened 2015-11-06
SDL*_mixer: help ./configure find shared libraries for dynamic loading

fabiangreffrath:
This avoids hard dependencies against the libraries providing the
music backends (with the exception of libmad). Instead, if the
libraries are found in the search path, they are loaded as plug-ins at
run-time.

Fixes https://github.com/Alexpux/MINGW-packages/issues/865

First, the ./configure script did check only for $host being
*-mingw32*, so it skipped the relevant parts for $host=*-mingw64*.

Then, the ./configure script did check for the wrong library names for
both libfluidsynth*.dll and ibvorbisidec*.dll.

Finally, the find_lib() function in the ./configure script searched
for the shared libraries in the wrong directories. Fixed this by
adding $PATH to the search paths.

### DIFF
--- a/mingw-w64-SDL2_mixer/PKGBUILD
+++ b/mingw-w64-SDL2_mixer/PKGBUILD
@@ -18,13 +18,17 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-fluidsynth")
 options=('staticlibs' 'strip')
 noextract=(${_realname}-${pkgver}.tar.gz)
-source=("$url/release/${_realname}-${pkgver}.tar.gz")
-md5sums=('65f6d80df073a1fb3bb537fbda031b50')
+source=("$url/release/${_realname}-${pkgver}.tar.gz"
+        SDL2_mixer-find_lib.mingw.patch)
+md5sums=('65f6d80df073a1fb3bb537fbda031b50'
+         'c891fc5d73238c85ff42cfe31213da7b')
 
 prepare() {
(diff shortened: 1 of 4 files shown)


mingwandroid:
Please add proper relocation support for finding the correct path. MSYS2 has a pathtools sample that can be reused. Out programs need to work outside of the msys2 shell.

mingwandroid:
Please add proper relocation support for finding the correct path. MSYS2 has a pathtools sample that can be reused. Our programs need to work outside of the msys2 shell.

fabiangreffrath:
@mingwandroid Sorry, I was just going to reply to your request when I saw that my PR has already been pulled, albeit with slight modifications. So, this discussion might come a little late. :(

However, I am not sure I understood your request properly.
- What so you mean by "proper relocation" and which part of the patch does this apply to? I guess you mean the part in which certain path strings are split up into single chunks, right?
- What am I supposed to do with the "pathtools sample"? There isn't a "pathtools" package available anywhere in mingw-w64 and what am I supposed to do with uncompiled C-code in a configure script?
- What do you mean with "work outside of the msys2 shell"? We are talking about a configure script that is supposed to be run by the shell, i.e. `/bin/sh`, right? What is the problem if this is started from anything than the msys2 shell? Again, I guess you mean the parsing of the path strings in this part of the configure script that my patch applies to, right? How am I supposed to get this more portable if one OS returns this string and the other OS returns the other?

While we are at discussing this: What I actually wanted to add to the search paths was the *linker path*, i.e. the path in which the linker searches for shared libraries that are referenced in executable objects. In other words, I was searching for the pendant to `/etc/ld.so.conf*` on Linux, but couldn't find it anywhere. That's why I got the idea to add `$PATH` in the first place.

PR #889, opened 2015-11-18
Add packages for pycparser and cffi Python modules.

fred-wang:
In my attempt to run buildbot with MSYS2, I found that there is a plan to replace pywin32 with cffi:

https://twistedmatrix.com/trac/ticket/7477

As a consequence, I've prepared some PKGBUILD for cffi and (its dependency) pycparser ; even if I'll continue to explore the pywin32 way. I have not really tested these Python packages, but at least the cffi demo for Windows seems to work (copying some text into the clipboard).

cffi needed a small patch to use the libffi, but the ifdef / sys.platform definitions on http://sourceforge.net/p/msys2/wiki/Contributing%20to%20MSYS2/ do not seem to help distinguishing with the case where MSVC.

### DIFF
--- a/mingw-w64-python-cffi/001-compile-and-setup-fixes.patch
+++ b/mingw-w64-python-cffi/001-compile-and-setup-fixes.patch
@@ -1,0 +1,23 @@
+diff -r 0c0093754b62 c/misc_win32.h
+--- a/c/misc_win32.h	Tue Nov 17 10:36:39 2015 +0100
++++ b/c/misc_win32.h	Wed Nov 18 10:34:34 2015 +0100
+@@ -238,4 +238,3 @@
+ /************************************************************/
+ /* obscure */
+ 
+-#define ffi_prep_closure(a,b,c,d)  ffi_prep_closure_loc(a,b,c,d,a)
+diff -r 0c0093754b62 setup.py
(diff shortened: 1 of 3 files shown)


DavidEGrayson:
This looks similar to the work I was doing here:

https://github.com/DavidEGrayson/MINGW-packages/tree/py_cffi/mingw-w64-python-cffi

However, I didn't get the cffi tests to pass.  Do they pass for you?  At the moment, I forget the details of how to run the tests unfortunately.

fred-wang:
I think I didn't figure out how to run the test and just verified the windows clipboard demo. I'll do a separate pull request for pycparser and will come back later for cffi.

fred-wang:
Apparently, you wrote in your patch that pytest is required to run the tests, so I can prepare a patch for it too.

fred-wang:
@DavidEGrayson I opened #897 and #896 for the missing dependencies you mentioned in the comments of your PKBUILD.

For py.test-2.7, I get the same result as you: one failure and the execution stops after a segmentation fault.

For py.test-3.5, the execution ends with 805 failures, 166 passed, 116 skipped and 163 errors :-( 

I'm not sure how much the tests rely on the assumption of being ran in a non-MSY2 environment...

PR #891, opened 2015-11-18
Add png2ico as dep to exm, update ecm

TheOneRing:
We might need to discuss the disabled patch for ecm

### DIFF
--- a/mingw-w64-extra-cmake-modules/PKGBUILD
+++ b/mingw-w64-extra-cmake-modules/PKGBUILD
@@ -4,17 +4,18 @@
 
 _realname=extra-cmake-modules
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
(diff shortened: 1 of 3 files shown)


mingwandroid:
@TheOneRing : I've fixed the AUTOSTATICPLUGIN problem now, but it requires rebuilds of the mingw-w64-qt5 packages (the non-static ones actually). You can manually edit the qt5 packages removing the STATIC_PLUGINS line from the _populate_*_plugin_properties macro in the .cmake files to save time:

    macro(_populate_*_plugin_properties Plugin Configuration PLUGIN_LOCATION)
    ..
        set_property(TARGET Qt5::* APPEND PROPERTY STATIC_PLUGINS ${Plugin})
    ..

mingwandroid:
Thanks a lot.

PR #897, opened 2015-11-20
Add package for pytest and py Python modules.

fred-wang:


mingwandroid:
Thanks! Again, maintainer please in future or when you update.

Issue #893, opened 2015-11-19
Add buildbot packages


fred-wang:
This is a meta bug for adding buildbot packages:

http://buildbot.net/
https://pypi.python.org/pypi/buildbot-slave
https://pypi.python.org/pypi/buildbot

One of the issue is that pywin32 does not build yet (cf #751 ). I tried to continue the work done there, but as @renatosilva said, MVSC's __try ... _except syntax is not supported by mingw64 (cf http://sourceforge.net/p/msys2/mailman/message/34378748/ ).

Anyway, it seems that @opalmer has started some work to replace pywin32 with cffi, so that seems to be the way to go:
https://twistedmatrix.com/trac/ticket/7477
https://github.com/opalmer/pywincffi/
https://pywincffi.readthedocs.org/en/latest/

I already opened #892 and #889 and will submit more pull requests as I make progress on this.

opalmer:
pywincffi is a work in progress and it will be a while before it replaces pywin32 in Twisted.  I don't want to hold this up though so maybe in the meantime you could reach out to @glyph who built [pypiwin32](https://pypi.python.org/pypi/pypiwin32/219) into a binary wheel to figure out how he was able to get it to build.

DavidEGrayson:
I was working on making a mingw-w64-python-cffi package but the tests were failing and then I stopped.  You can see my work here and use any of it in any way if it is useful to you:

https://github.com/DavidEGrayson/MINGW-packages/tree/py_cffi/mingw-w64-python-cffi

glyph:
I installed VS2K8 and ran `setup.py bdist_wheel`, and wrote a tiny patch to add to the `.pth` file to set the `PATH` environment variable so that the `.dll` files included in the package were loadable.  That's really it.  I still have a to-do somewhere in my infinitely-receding to-do list to contribute the patch upstream, since I think I have commit access to pywin32 now :).  But it's really not much work just to do the build.  You can look at the `.pth` file in the wheel or the uploaded sdist to see the line I'm talking about.

fred-wang:
Thank you everybody for your replies. While I was trying to update @opalmer 's patch for twisted, I actually did the official python.org setup and in particular @glyph 's  pypiwin32 package. It turns out that this now allows me to run buildslave in MINGW64 / MINGW32 consoles too and so this resolves my problem.

[...]

PR #898, opened 2015-11-20
Add packages for zope.event and zope.interface Python modules #893

fred-wang:
zope.interface is required for python.twisted (and I used zope.event to run the tests)


mingwandroid:
Thanks! Though you should set yourself as maintainer really.

PR #900, opened 2015-11-21
ada gtk solution for mingw

RickLeaf:
1. fix gtkada build issue
2. build mingw ada tools first version with gcc (not gprbuild)


mingwandroid:
Thanks!

PR #905, opened 2015-11-22
clang: Removed -DLLVM_ENABLE_THREADS=0 as it is causing segmentation fault w…

eidheim:
…hen performing multithreaded libclang parsing, and fixed a sha256sum.


mingwandroid:
Sorry, I followed your first link (I think the aspell one is unrelated).. ok, you have clang built as single threaded calling into something built as multithreaded? Merged as a trivial fix then. Thanks.

mingwandroid:
.. and apologies it took so long.

PR #914, opened 2015-11-25
We use gtk2 for windows, so we need a buildin ime for windows

RickLeaf:
1. Native input method interface is necessary for other language.
And If you use Windows 8 or later without this module, gtk2 will crash when system input method was set Chinese, Japanese or others.
2 . I add gdl library for gtk+ 2.0

### DIFF
--- a/mingw-w64-gdl2/PKGBUILD
+++ b/mingw-w64-gdl2/PKGBUILD
@@ -1,0 +1,53 @@
+# $Id$
+# Maintainer: Ricky Wu <rickleaf.wu@gmail.com>
+
+_realname=gdl2
+pkgbase=mingw-w64-${_realname}
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+pkgver=2.31.2
+pkgrel=1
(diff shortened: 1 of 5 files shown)


mingw-w64-gdl2/PKGBUILD
```diff
@@ -0,0 +1,52 @@
+# $Id$
+# Maintainer: Ricky Wu <rickleaf.wu@gmail.com>
+
+_realname=gdl2
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
```
Alexpux: Add "pkgbase" entry before "pkgname"

mingw-w64-gdlmm2/PKGBUILD
```diff
@@ -0,0 +1,52 @@
+# $Id$
+# Maintainer: Ricky Wu <rickleaf.wu@gmail.com>
+
+_realname=gdlmm2
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
```
Alexpux: As previous

RickLeaf:
I recommit these changes with only one commit

PR #917, opened 2015-11-29
emacs: Apply IME patch for Japanese language input

jo7ueb:
Emacs has a problem about Japanese language input with Widows GUI.
A patch is provided from following URL to fix this problem, and I applied it.
https://gist.github.com/rzl24ozi/a282260d2dd136daab14


Alexpux:
Don't see patch in pull request

jo7ueb:
mingw-w64-emacs/emacs-24.5-w32-ime.diff is added.
Is it OK?

mingwandroid:
Thanks!

Issue #923, opened 2015-12-06
CMake autogenTargetName different from official


Optiligence:
Is there a reason `autogenTargetName` is renamed from `_automoc` to `_autogen`? 
This i.e. makes it necessary to patch [cotire](https://github.com/sakra/cotire).

mingwandroid:
This bug report is quite context free! Can you specify the package, and maybe isolate it down to the commit in which the change was made? I know it's Qt5-related, but in the spirit of providing good bug reports I'd like to see more effort expended :-)

Optiligence:
https://github.com/Alexpux/MINGW-packages/commit/aa8f86ab267446002bcbb59e1455fbdf0d9f1cde#diff-4c9f45a9f09cd9ff5e35063610a2d954R29

mingwandroid:
Well, since autostaticplugin came along we put more than automoc created files in that folder so that name is less appropriate than autogen. For cotire to care about that folder name seems a bit like it's reaching behind the scenes at an implementation detail somewhat?

Anyway, if you want to supply a patch that modifies qt5-static-plugin-support.patch to change it back then that's fine with me, I'll merge it.

mingwandroid:
I'd like to see that patch necessary for cotire too if possible. Perhaps they can be convinced not to rely on implementation details?

Optiligence:
I wouldn’t say that this is just an implementation detail: [_automoc target](https://cmake.org/cmake/help/v3.4/prop_tgt/AUTOMOC.html?highlight=_automoc), [_automoc file](https://cmake.org/cmake/help/v3.4/prop_tgt/AUTOGEN_TARGET_DEPENDS.html?highlight=_automoc). 

I agree that `autogen` fits the purpose more, idiomatically, not functionally (except it’s intended that users notice that this CMake works differently regarding this property). 
CMake already uses this for more than just the `moc` (`uic`, `rcc`), but i think they are keeping the name for backwards compatibility, especially regarding the links above. 

cotire usage: [file](https://github.com/sakra/cotire/blob/c55aeec38077a1ce099faa3628b5fab785ba5fb4/CMake/cotire.cmake#L2858:L2860), [target](https://github.com/sakra/cotire/blob/c55aeec38077a1ce099faa3628b5fab785ba5fb4/CMake/cotire.cmake#L2876:L2877)
I’m not sure if it’s possible to query the file and target in question differently.

mingwandroid:
[...]

PR #927, opened 2015-12-08
Issue #923, CMake 3.4.1, cotire

Optiligence:
I tested `automoc`, `autorcc` and `autostaticplugins`, i have nothing to test `autouic` with though, but it should work nonetheless. 

CMake 3.4 split off `cmQtAutoGeneratorInitializer.cxx` from `cmQtAutoGenerators.cxx` and `GetLinkImplementationClosure` moved from `cmTarget` to `cmGeneratorTarget`. 
The hunk with most of the implementation actually applied without error, but at the wrong position in the source. 

The path in cotire’s packaging has to be changed with every major CMake release. (CMake only searches in the module directory of the current version)


mingwandroid:
Thank you very much for handling this and updating to cmake 3.4.1. I'll merge the changes as soon as you address the minor issues with cotire.

Optiligence:
done

mingwandroid:
Thanks.

PR #930, opened 2015-12-10
catch: Update to version 1.3.1.

DavidEGrayson:
For the packages I contribute, I try to monitor the upstream projects for new releases.  A new version of Catch was released just yesterday, so this pull request updates our MSYS2 package.

### DIFF
--- a/mingw-w64-catch/PKGBUILD
+++ b/mingw-w64-catch/PKGBUILD
@@ -1,15 +1,16 @@
 # Maintainer: David Grayson <davidegrayson@gmail.com>
 
 _realname=catch
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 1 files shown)


mingwandroid:
Thanks!

mingw-w64-catch/PKGBUILD
```diff
@@ -2,14 +2,14 @@
 
 _realname=catch
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
```
Alexpux: Add please "pkgbase" entry before "pkgname"

DavidEGrayson:
OK, I did that.  I also reset the pkgrel to 1, which I should have done originally I think.

DavidEGrayson:
Thanks, @Alexpux!  :christmas_tree:

PR #938, opened 2015-12-21
gitg: update to version 3.19.1

nacho:
Now that gitg does not depend on webkit we should be able
to build it without problems


mingwandroid:
Thanks!

mingwandroid:
Hi @nacho:

No package 'gio-unix-2.0' found
No package 'gee-0.8' found
No package 'libsecret-1' found

nacho:
Yes, working on it.

PR #939, opened 2015-12-21
gitg: add missing deps

nacho:
See that it will still not build due to the gio-unix dep.
This will be fixed by patch later on.

### DIFF
--- a/mingw-w64-gitg/PKGBUILD
+++ b/mingw-w64-gitg/PKGBUILD
@@ -2,7 +2,7 @@
 
 _realname=gitg
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.19.1
+pkgver=3.19.2
 pkgrel=1
 arch=('any')
 pkgdesc="git repository viewer for GTK+/GNOME (mingw-w64)"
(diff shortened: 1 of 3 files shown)


mingwandroid:
Yup, Just `package 'gio-unix-2.0' found`. I'd rather merge when all working if that's ok now?

nacho:
with this last patch it should start to work. As in builds and starts the app. But there are still bugs.

mingwandroid:
It's getting there I think :-)

nacho:
Finally done.

mingwandroid:
Thanks @nacho.

mingwandroid:
Oops.

mingwandroid:
gitg is really nice, thank you for this @nacho.

nacho:
Thanks to you too. I am pretty sure there are still bugs, so anything you find please report it and we will fix it asap.

mingwandroid:
The commit option doesn't seem to pick up the details giving instead:

    Author name <email@example.com>
    Committed by Committer name <email@example.com>

.. but it looks work-in-progress generally so it may not be an MSYS2 problem.

nacho:
Please file this in the gnome bugzilla.

PR #946, opened 2015-12-24
libusb-compat: Move usb.h to libusb-compat directory

DavidEGrayson:
This pull request has two commits: the first commit just updates the libusb-compat package so that it can build cleanly, but doesn't make any majors changes.

The second commit moves usb.h to the libusb-compat subdirectory of the include folder (e.g. `/mingw64/include/libusb-compat`).  Having usb.h in the top-level include folder caused a lot of problems for any native Windows program trying to use USB-related headers in mingw-w64.  The mingw-w64 headers have a header file named usb.h, and a few other header files like winusbio.h include usb.h using `#include <usb.h>`.  Those includes were pulling in the version of usb.h from libusb-compat, which caused lots of problems.  You can see an example of how we worked around this in the [usbview PKGBUILD](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-usbview-git/PKGBUILD#L68-L69).

I also brought this up on the mingw-w64 mailing list, but they didn't want to fix the problem on their end and said "it is that software that is supposed to adapt to the platform":  https://sourceforge.net/p/mingw-w64/mailman/message/34590803/


Alexpux:
This will break building software that depends on libusb-compat. So need to fix it in another way

DavidEGrayson:
Alexey: I thought it would be easy enough for people to fix their software so that it finds libusb-compat using `pkg-config`, or they can just hardcode the proper `-I` option in their build scripts.  The only packages in this repository that depend on libusb-compat are gnupg and openocd.  If I tested/fixed both of those packages, would you then accept the pull request?

Alternatively, would you accept a pull request where I add the following patch to the mingw-w64-headers package?  This is what I send to the mingw-w64 mailing list but it was rejected.  With this patch, at least the WinUSB headers will work.  But any Windows software that wants to use the mingw-w64 version of usb.h must include something like usbdi.h instead of including usb.h directly, so this patch is less friendly to the native Windows software.

~~~~{.patch}
--- a/mingw-w64-headers/include/usbdi.h
+++ b/mingw-w64-headers/include/usbdi.h
@@ -20,7 +20,7 @@
 
 #if WINAPI_FAMILY_PARTITION (WINAPI_PARTITION_DESKTOP)
 
-#include <usb.h>
+#include "usb.h"
 #include <usbioctl.h>
 
 #define USBD_STATUS_CANCELLING ((USBD_STATUS) 0x00020000)
diff --git a/mingw-w64-headers/include/winusbio.h b/mingw-w64-headers/include/winusbio.h
index eb4ac8f..6bde19a 100644
--- a/mingw-w64-headers/include/winusbio.h
+++ b/mingw-w64-headers/include/winusbio.h
@@ -50,7 +50,7 @@
 #if WINAPI_FAMILY_PARTITION (WINAPI_PARTITION_DESKTOP)
 
 #include <windows.h>
-#include <usb.h>	/* for USBD_PIPE_TYPE */
+#include "usb.h"       /* for USBD_PIPE_TYPE */
 
 #ifdef __cplusplus
 extern "C" {
~~~~

mingwandroid:
If it is for the long term good then I think it could be acceptable. If we can encourage upstream to accept it then for sure it is acceptable? David, would you please try that approach?

DavidEGrayson:
@mingwandroid I already tried to submit that patch to the mingw-w64 and they didn't like it.  Instead, I think we should use a new approach that I explained in pull request #951.

mingwandroid:
I didn't mean mingw-w64 as the upstream I meant libusb-compat. mingw-w64 must follow Microsoft as much as possible. libusb-compat are the ones with wriggle room. @Alexpux please reconsider. Not that much software depends on libusb-compat and if they need and ugly #ifdef then I think it is tough.

Alexpux:
I prefer to not change mingw-w64 headers. Need to fix it in libusb-compat and packages that use it.
This is implies way I think at the end.

DavidEGrayson:
How about this: I'll send to the libusb-devel mailing list (where libusb-compat is discussed) saying that we are thinking about using the `--includedir=/mingw64/include/libusb-compat` option and asking them if they have any opinions on that.  They might suggest a different name for the folder or something.  

I don't think I need to submit any patches to them because their software already supports the `--includedir` option.

DavidEGrayson:
Actually, does anyone actually want me to contact the libusb developers?  @mingwandroid said something about sending a patch to the upstream developers but I don't think any patch is necessary.

As far as I'm concerned, this pull request can be merged in now.  Please let me know if there's something else I should do.

mingwandroid:
I wouldn't mind getting their opinion on making this the default. Having all third party code uglified for this unfortunate naming clash is well, unfortunate.

mingwandroid:
@Alexpux, just to make sure, you are ok with this now?

DavidEGrayson:
For the record, @AlexPux said on IRC that he was OK with moving usb.h to a subdirectory.

Also, I posted my question to the libusb-devel mailing list on 2015-12-30: https://sourceforge.net/p/libusb/mailman/message/34729537/

Also, I just simplified this pull request.  It's one commit now and I took out the dos2unix stuff and added pkgbase.

mingwandroid:
Do you want to take on co-maintainership of this also David? I assume Andrew Leech wouldn't mind, but I've never heard from him directly actually.

DavidEGrayson:
Hmm, the git history shows no history of this Andrew Leech committing either; it looks like @mingwandroid made all the commits.  Well, I'll add myself to the list of maintainers.

mingwandroid:
One of the openocd developers asked me to help out since he wanted to use MSYS2 as the build env. for their Windows releases, and pointed me to Andrew's PKGBUILDs as a starting point. Andrew wasn't about (or at least, he never made pull requests or got in touch) so I just imported them in and got on with it.

DavidEGrayson:
Hey @AlexPux, the name clash between the two different `usb.h` files is still annoying; I have to manually fix `winusbio.h` to use quotes instead of angle brackets when including `usb.h`.  Could you rebuild libusb-compat some time?  The last time it was built was in 2015.  Do you want me to test it out to make sure it still builds these days?

PR #952, opened 2015-12-26
avrdude fixes

### LINKED ISSUE #947: avrdude depends on libusb-0.1-4.dll
Thanks, @AlexPux, for building AVRDUDE the other day!  However, there is something I overlooked, and this issue is mostly a note to myself about fixing it.

I noticed that the avrdude.exe you built depends on libusb-0.1-4.dll, which is not what I was expecting (and is not reflected in the package dependencies).  I thought that AVRDUDE would simply use libusb-1.0 instead of using libusb-compat.  I will have to look into that more and figure out why it is using libusb-compat and whether we can somehow tell avrdude to ignore libusb-compat and use libusb-1.0 instead.

For now, anyone wanting to use avrdude should manually install libusb-compat-git if they have not done so already.

DavidEGrayson:
- Added validpgpkeys
- Added a dependency on libusb-compat because there is code in AVRDUDE that uses it and cannot use libusb-1.0 (fixes #947).

### DIFF
--- a/mingw-w64-avrdude/PKGBUILD
+++ b/mingw-w64-avrdude/PKGBUILD
@@ -4,7 +4,7 @@ _realname=avrdude
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=6.2
-pkgrel=1
+pkgrel=2
 pkgdesc='Software for programming Atmel AVR Microcontrollers (mingw-w64)'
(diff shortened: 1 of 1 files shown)


DavidEGrayson:
I think this pull request is good.  I checked recently that avrdude can still find libusb-compat even though the header file has moved or is moving to a subdirectory; it's probably using pkg-config.

mingwandroid:
Thanks @DavidEGrayson

Issue #971, opened 2016-01-03
gdal needs a rebuild


parched:
because of updated dependencies, poppler and webp.

parched:
PR was merged

PR #974, opened 2016-01-03
mingw-w64-gdal: rebuild for poppler and webp

parched:
Needs a rebuild because of updated dependencies, poppler and webp. #971


parched:
I have built and tested this.

Issue #976, opened 2016-01-04
SDL2_image needs rebuild


TinoM:
It still depends on an older webp version.
Needed: libwebp-5.dll
Provided by current libwebp package: libwebp6.dll

mingwandroid:
Pleas submit a pull request with a bumped pkgrel so it is not forgotten. In the meantime, copy `libwebp-6.dll` as `libwebp-5.dll`

PR #978, opened 2016-01-04
Update SDL2 libaries

TinoM:


TinoM:
Should also fix #976

mingwandroid:
Thank you very much!

PR #981, opened 2016-01-05
Create PKGBUILD for enet 1.3.13 (http://enet.bespin.org/)

qme5400:
Please review new PKGBUILD: Nothing to do in "prepare()". No need to run "./configure".


qme5400:
Thank you!

PR #986, opened 2016-01-09
cURL: Get relocatable base from .dll instead of .exe

dscho:
Currently, cURL uses the path to the current executable to figure out
where the certificates are. This works as long as all executables are in
/mingw64/bin/ (or /mingw32/bin/).

However, the entire point of having a DLL with a linkable .dll.a is to
be able to use it in executables that other people build. For example,
if an .exe lives in %HOME%\bin, the path used by our libcurl to look up
the certificates is all wrong.

Same problem with Git for Windows: its executables are hidden away from
the user, in the /mingw64/libexec/git-core/ directory. So the path to
the certificates is all wrong again.

So let's just use the path to the *cURL library* instead of the path to
the current .exe. This requires Win32 API available in Windows XP & 2003
and later, well within the Windows versions supported by Cygwin (and
therefore MSys2).

Reported by David Ebbo.

Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>


mingwandroid:
Thanks!

PR #992, opened 2016-01-11
mingw-w64-libspectre: make package build against newer versions of ghostscript

darealshinji:


mingw-w64-libspectre/PKGBUILD
```diff
@@ -18,12 +18,16 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
 options=('strip' 'staticlibs')
```
Alexpux: Increase pkgrel please

darealshinji: Done.

PR #996, opened 2016-01-12
qt5: Some fixes for the no-ICU case.

DavidEGrayson:
The patch 0123-webkit-commit-151422.patch has already been applied upstream.  Also, I had to add another patch to fix a compiler error in qtwebkit.

After making these fixes, I was able to successfully compile a version of qt5 with use_icu=no and opengl=desktop.  I have not tested it to see if it works though.


mingwandroid:
Thanks @DavidEGrayson

PR #1006, opened 2016-01-16
Update snorenotify

TheOneRing:


mingw-w64-snorenotify/PKGBUILD
```diff
@@ -1,21 +1,21 @@
-# Maintainer: Patrick von Reth <vonreth@kde.org>
+# Maintainer: Hannah von Reth <vonreth@kde.org>
 
-_realname=Snorenotify
+_realname=snorenotify
```
mingwandroid: I'm pretty sure this then needs to have: `replaces=("${MINGW_PACKAGE_PREFIX}-Snorenotify")`
@Alexpux, can you confirm?

TheOneRing: Probably I'll change that.

PR #1007, opened 2016-01-16
Update URL for oniguruma

JanSchulz:
The URL seems to have been switched to https://github.com/kkos/oniguruma

### DIFF
--- a/mingw-w64-oniguruma/PKGBUILD
+++ b/mingw-w64-oniguruma/PKGBUILD
@@ -8,9 +8,9 @@ pkgrel=1
 pkgdesc="A regular expressions library (mingw-w64)"
 arch=('any')
 license=('BSD')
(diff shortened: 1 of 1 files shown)


JanSchulz:
Not that I currently use the following line on appveyor to change the URL when using your repo:

```
  - C:\msys64\usr\bin\bash -lc "cd /c/projects/jq/MINGW-packages/mingw-w64-oniguruma && sed -i s#http://www.geocities.jp/kosako3/oniguruma/archive/onig-5.9.6.tar.gz#https://github.com/kkos/oniguruma/releases/download/v5.9.6/onig-5.9.6.tar.gz# PKGBUILD"
```

JanSchulz:
I used my branch to build the package on appveyor and it least downloads the file and the hash is ok...

PR #1013, opened 2016-01-19
Switch dcadec to the 0.2.0 release and update FFmpeg to 2.8.5

rossy:
I submitted these together because the FFmpeg package is changed to depend on the dcadec release version. dcadec has releases now, so this changes it from using the git version to using a release tarball, though I'm not sure if I've changed the package metadata to correctly reflect this. All I did is add the ``replaces=()`` line.

The FFmpeg update is somewhat important because it fixes CVE-2016-1897 and CVE-2016-1898.


rossy:
So, was the ``mingw-w64-dcadec-git => mingw-w64-dcadec`` change done correctly? Will it be seamless for users of the existing ffmpeg and dcadec packages?

Alexpux:
not correctly al all but I will fix it myself

rossy:
Fair enough. For future reference, what did I do wrong?

Alexpux:
Need also add "replaces=(${MINGW_PACKAGE_PREFIX}-${_realname}-git)" entry

rossy:
Ah, thanks.

mingwandroid:
Sorry @Alexpux, I didn't check it very carefully.

PR #1015, opened 2016-01-21
libCello: new package.

sbwhitecap:


DavidEGrayson:
I can't speak for @Alexpux, but I think he will probably ask you to:

* Add a `pkgbase` line
* Rename the folder to match the package name (it's missing "-git").

Also, I think capital letters in package names are annoying.  I see that Arch Linux users just calls this [libcello](https://aur.archlinux.org/packages/libcello/) so I would suggest lowercasing the C unless you can find some reason to leave it uppercase.  I don't see "libcello" or "libCello" on the [project website](http://libcello.org/).

PR #1026, opened 2016-01-25
mingw-w64-qtcreator: Update to 3.6.0

nazar554:
It looks like it was recently released but no links on `download.qt.io` site page. Only this https://download.qt.io/official_releases/qtcreator/3.6/3.6.0/

Binaries are here: https://drive.google.com/folderview?id=0B_CR4N27LxsubF9hSS1EdWh6WU0&usp=sharing


mingwandroid:
There's no point in renaming patches like that, the patches are named according to the first version where they were necessary, please undo this part.

mingwandroid:
.. and squash the commits too please ;-)

mingwandroid:
Thank you for your patience!

PR #1034, opened 2016-01-30
mingw-w64-gcc: Add cc.exe

elieux:
This has been reported as a problem several times, that `cc` in mingw64 shell points to `/usr/bin/cc` or nothing (depending on whether the user installed the `gcc` package).

### DIFF
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -12,7 +12,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-ada"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-objc")
 pkgver=5.3.0
-pkgrel=1
+pkgrel=2
 pkgdesc="GCC for the MinGW-w64"
 arch=('any')
 url="http://gcc.gnu.org"
@@ -126,7 +126,7 @@ prepare() {
   patch -p1 -i ${srcdir}/230-build-more-gnattools.mingw.patch
 
   patch -p1 -i ${srcdir}/240-prettify-linking-no-undefined.mingw.patch
-  
+
   patch -p1 -i ${srcdir}/310-gcc-make-xmmintrin-header-cplusplus-compatible.patch
 
   if [ "${CARCH}" = "i686" ]; then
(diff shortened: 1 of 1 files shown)


Alexpux:
Maybe create shell scripts as we do for msys gcc? What about others like c++?

elieux:
Sorry for the delay, I didn't notice the email notification.

> Maybe create shell scripts as we do for msys gcc?

I think mingw-w64 `cc` should be executable from outside MSYS2. (Another option I could go for is to replace `/usr/bin/cc.exe` with a `/usr/bin/cc` script.)

> What about others like c++?

`c++` is already in both packages, see below. Only differences I see are `c89` and `c99`. I can add these two shell scripts to this pull request (even though I doubt their usefulness). Should I do that?

```
$ pacman -Qlq mingw-w64-x86_64-gcc | grep /bin/
/mingw64/bin/
/mingw64/bin/c++.exe
/mingw64/bin/cpp.exe
/mingw64/bin/g++.exe
/mingw64/bin/gcc-ar.exe
/mingw64/bin/gcc-nm.exe
/mingw64/bin/gcc-ranlib.exe
/mingw64/bin/gcc.exe
/mingw64/bin/gcov-tool.exe
/mingw64/bin/gcov.exe
/mingw64/bin/x86_64-w64-mingw32-c++.exe
/mingw64/bin/x86_64-w64-mingw32-g++.exe
/mingw64/bin/x86_64-w64-mingw32-gcc-5.3.0.exe
/mingw64/bin/x86_64-w64-mingw32-gcc-ar.exe
/mingw64/bin/x86_64-w64-mingw32-gcc-nm.exe
/mingw64/bin/x86_64-w64-mingw32-gcc-ranlib.exe
/mingw64/bin/x86_64-w64-mingw32-gcc.exe
```

```
$ pacman -Qlq gcc | grep /bin/
/usr/bin/
/usr/bin/c++.exe
/usr/bin/c89
/usr/bin/c99
/usr/bin/cc.exe
/usr/bin/cpp.exe
/usr/bin/g++.exe
/usr/bin/gcc-ar.exe
/usr/bin/gcc-nm.exe
/usr/bin/gcc-ranlib.exe
/usr/bin/gcc.exe
/usr/bin/gcov.exe
/usr/bin/x86_64-pc-msys-c++.exe
/usr/bin/x86_64-pc-msys-g++.exe
/usr/bin/x86_64-pc-msys-gcc-4.9.2.exe
/usr/bin/x86_64-pc-msys-gcc-ar.exe
/usr/bin/x86_64-pc-msys-gcc-nm.exe
/usr/bin/x86_64-pc-msys-gcc-ranlib.exe
/usr/bin/x86_64-pc-msys-gcc.exe
```

ian-abbott:
I'm not sure how useful the `c89` and `c99` scripts would be, but if someone really wants them, they could install the msys2/gcc package. The scripts will run whichever `gcc` is first in the `PATH`, which in a MinGW-w64 Win32 or Win64 environment would be the one from `/mingw32` or `/mingw64`.

ian-abbott:
A minor technical point about c66d4441ac61cf9cd2f2772ce534874c5b801782 - it may be better to use `ln -s` instead of `cp` to create `cc.exe`, as that is what the msys2/gcc package does (although MSYS2 will substitute the symbolic links with copies by default).

mingwandroid:
It's not better to use `ln -s` in MSYS2 PKGBUILDs since some people may run with `nativestrict` symlinks and thus generate packages with real symlinks in them.

PR #1038, opened 2016-02-02
Add drone build script.

fracting:
Similar to https://github.com/Alexpux/MSYS2-packages/pull/431, this patch add build script for MINGW packages, using mingw32 compilers.

mingw64 will be added once some strange Wine64 bugs are fixed (like https://bugs.wine-staging.com/show_bug.cgi?id=623)

Currently the server has only one CPU core, running on AWS. In the coming days I'll start to upgrade the server with more cores.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,14 @@
+build:
+  image: fracting/msys32
+  shell: msys32
+  commands:
+    - drone/build.sh
+
+notify:
+  irc:
+    prefix: build
+    nick: MINGW-packages
+    channel: msys2
+    server:
+      host: irc.oftc.net
+      port: 6667
--- a/drone/build.sh
+++ b/drone/build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# fetch first changed file, assume at most one package touched per commit
+TOUCHED=`git show --pretty="format:" --name-only | grep . | head -1`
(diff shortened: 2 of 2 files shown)


.drone.yml
```diff
@@ -0,0 +1,14 @@
+build:
+  image: fracting/msys32
+  shell: mingw32
```
Alexpux: I think that building packages need be builded under msys shell not mingw

fracting: @Alexpux thanks for pointing out that, does that mean I need to set MSYSTEM=MSYS even for building mingw packages?

Alexpux: makepkg-mingw switch environment variables itself

Alexpux:
You build under 32-bit MSYS2?

fracting:
Thanks, i update the patch to msys32 shell.

Test here http://wine-ci.org/fracting/MINGW-packages/5

(I found there is still random Wine bug which crashing the build, sorry for that, will try to fix that once they happen again)

fracting:
Yes, this time it is under 32-bit MSYS2.

64-bit MSYS2 is not supported by Wine yet.

Alexpux:
ok

fracting:
Thanks for testing! I would expect there are many things to test, to tweak and to improve, let's see how it works.

PR #1040, opened 2016-02-04
libbotan: Updated to 1.11.28

nazar554:
Is this patch okay? Making `_WIN32_WINNT` equal to `0x0600` breaks complitability with < Windows Vista.


Alexpux:
No we will not break compatibility with other windows versions.

nazar554:
Fixed by removing the call.

PR #1046, opened 2016-02-05
Tcllib fix md5

fracting:
https://github.com/Alexpux/MINGW-packages/pull/1045 doesn't build on CI server, see:

http://wine-ci.org/Alexpux/MINGW-packages/12

I manually verified on Windows in order to confirm it is not a CI server bug.

However, there is a CI server bug which reports fake success for matrix build at irc (and only at irc), I think this fake report is very misleading, so I temporary disable matrix build, will re-enable after fix irc notify.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,4 +17,5 @@ notify:
 matrix:
   runtest:
     - false
-    - true # allow failures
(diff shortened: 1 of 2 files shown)


fracting:
Successful buld here: http://wine-ci.org/fracting/MINGW-packages/13

PR #1048, opened 2016-02-05
Added Libswift to msys2 repo

genusistimelord:
Added LibSwift to msys2 for the first time"


fracting:
Thanks for the contribution!

Sorry for the noise here, I just need a place to put a cross-reference.
Interesting thing, when compiling this package on Wine, we found a Wine bug which cause objdump hangs, I've filed the Wine bug as: https://bugs.wine-staging.com/show_bug.cgi?id=643

Full build log is here http://wine-ci.org/Alexpux/MINGW-packages/18 (might be changed if anyone trigger restart in the future, but the same log has been uploaded to the Wine Staging bugzilla)

PR #1051, opened 2016-02-06
Fix relocatable gettext-runtime on 64-bit systems

dscho:
When compiling code on 64-bit systems, the default return value of previously undeclared functions is `int`, as in *32-bit* integer value. This is a problem when the return value is actually a pointer. It becomes a real problem when the upper 32 bits of the register used to return that pointer are still set from previous usage (as is the case e.g. when the [`FLG_LDR_TOP_DOWN` global flag](https://technet.microsoft.com/en-us/library/cc779664%28v=ws.10%29.aspx) is set, as pretty much all return addresses have pretty much all of the upper 32 bits set to 1).

This is the problem that let anything using `bindtextdomain()` of the `gettext-runtime` package cause segmentation faults in this developer's tests.

Declaring the functions before using them miraculously fixes the segmentation faults.

Note that there is more work to do, as e.g. `gdb.exe` *still* fails (it does not link `libintl` at all) and most likely has a similar issue in another code location.

Please also note that in my tests, the patched code was not even compiled when building `mingw-w64-gcc`. But then, `gcc.exe` did not crash with a segmentation fault to begin with.

And lastly, I did not manage to get `mingw-w64-gcc-git` to build at all, so I did not fix that package's ["Relocate libintl" patch](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-gcc-git/0001-Relocate-libintl.patch).

### DIFF
--- a/mingw-w64-gcc/000-relocatex-libintl-0.18.3.1.mingw.patch
+++ b/mingw-w64-gcc/000-relocatex-libintl-0.18.3.1.mingw.patch
@@ -1,25 +1,37 @@
-From 10d1c9d898b512792dc562b3fb9fff19ffd7e7be Mon Sep 17 00:00:00 2001
+From 4f52371759455e1091f4cac44298dbba92b79340 Mon Sep 17 00:00:00 2001
 From: Johannes Schindelin <johannes.schindelin@gmx.de>
 Date: Fri, 5 Feb 2016 13:38:09 +0100
 Subject: [PATCH] relocatex-libintl
 
+[jes: fix for 64-bit]
+
 Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
 ---
- intl/Makefile.in    |   7 +-
+ intl/Makefile.in    |   8 +-
  intl/bindtextdom.c  |  22 ++++
- intl/canonicalize.c | 329 ++++++++++++++++++++++++++++++++++++++++++++++++++++
- intl/relocatex.c    | 295 ++++++++++++++++++++++++++++++++++++++++++++++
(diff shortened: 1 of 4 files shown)


fracting:
Sadly the build timeout because multiple builds are running on the same single core server. I'll upgrade wine-ci.org next week.

dscho:
Wow, that is a pretty nasty Javascript on that page, too. It took my entire web  browser down several times. @fracting could that be fixed, please?

fracting:
@dscho Thanks for point out that, it annoyed me as well, sorry :(
I'm not familiar with Javascript, I'm just doing some simple hack based on the open source Drone project: https://github.com/drone/drone

If you can create an issue to drone, that would be appreciated, otherwise I'll add it to my todo list, but still need somebody else to work on it. At the meantime I have to focus on more urgent issues like several Wine bug affects MSYS2. Sorry for the trouble!

fracting:
@dscho BTW, is it possible that your browser has a bug which crashes when some web content is closed to 2MB?

This comment says "We've observed that > 2MB of logs will crash the browser. ": https://github.com/drone/drone/issues/1429#issuecomment-174211783 for

dscho:
> is it possible that your browser has a bug which crashes when some web content is closed to 2MB?

No. It is not a bug in the browser but a memory restriction: I run my browser in a VM and there were only 1.2GB (!) free. That was apparently not enough for that Javascript, so it simply over-allocated and since I do not give my VM any swap, the OS killed the browser.

It is rather drastic that 1.2GB are not enough to handle 2MB of logs, don't you think?

> If you can create an issue to drone

Sorry, I share your pain, I am out of time.

PR #1027, opened 2016-01-25
Update fossil

BMBurstein:

PR #1054, opened 2016-02-06
Update+fix sqlite3, sqlite-analyzer and fossil

elieux:
Reverts #1027.


mingwandroid:
@Alexpux can you merge this since you merged the fossil commit in the first place. Or @elieux, why not just go ahead and do it yourself? You're the expert on fossil after all!

elieux:
True. Originally, I was waiting to see how #1027 turns out.

PR #1070, opened 2016-02-13
ruby: Update to 2.3.0

kumaryu:


fracting:
Hi @kumaryu, thanks for the contribution!

Your patch doesn't pass the CI build, please also update md5sum of those three patches, thanks!

```
==>Validating source files with md5sums...(B
    ruby-2.3.0.tar.bz2 ... Passed
    0001-mingw-w64-time-functions.patch ... FAILED
    0002-use-gnu-printf.patch ... FAILED
    0003-fix-check-types.patch ... FAILED
==> ERROR One or more files did not pass the validity check!
```

fracting:
@kumaryu Congratulations, CI build passed this time! :)

Thanks for contribution, let's wait @mingwandroid or somebody else to merge or comment on the patch.

fracting:
@kumaryu btw, next time please squash this kind of two patches into one patch and force update the pull request branch, thanks!

mingw-w64-ruby/PKGBUILD
```diff
@@ -60,7 +60,7 @@ build() {
 package() {
   cd "${srcdir}/build-${MINGW_CHOST}"
   make -f GNUMakefile DESTDIR="${pkgdir}" install-nodoc
-  for script in {erb,gem,irb,rake,rdoc,ri}; do
```
DavidEGrayson: Why are you removing rake here?  Ruby is usually supposed to come with a copy of rake.  When I tried to upgrade this package to 2.3.0, rake was causing problems for me but we should try to make sure it gets installed.

DavidEGrayson:
@AlexPux:  I just confirmed that ruby-2.3.0 does ship with rake (a useful Ruby build tool like Make), and it successfully installs rake when you compile it on Linux.  I know that rake was causing an error in the build of ruby 2.3.0 on MSYS2, but I think it's better to fix that error or to stay on version 2.2.3 than to make a breaking change like removing rake.

DavidEGrayson:
Well, I don't really know what's going on, but I'm looking in to it.

In Ruby commit https://github.com/ruby/ruby/commit/c4ee0df8ba2cf7cc6aaa785c8df20a91b1719021, it looks like Rake was changed from a "default gem" to a "bundled gem", which involved removing the source code of Rake from `lib` and removing Ruby's `bin/rake` file.  Rake no longer lives in the Ruby repository, but it ships in the `gems` directory of the Ruby tarball.

When I install ruby 2.3.0 in Arch Linux (with prefix `/opt/rubies/ruby-2.3.0` for compatibility with chruby), I see a file at `/opt/rubies/ruby-2.3.0/bin/rake` which starts with these lines:

```
#!/opt/rubies/ruby-2.3.0/bin/ruby
#
# This file was generated by RubyGems.
```

So it seems like there is some part of the Ruby build/install process where Rubygems (Ruby's package manager) generates binaries for each of the bundled gems and puts them in the bin directory.  This works on Arch Linux but doesn't work on MSYS2 for whatever reason, and it would be good to know why.

mingwandroid:
Thanks for looking into this David.

DavidEGrayson:
Some details about debugging this:  There is this cryptic error message at the end of the `package` function's execution which seems to say something is wrong with the Rake gem:

```
installing default gems:      /mingw64/lib/ruby/gems/2.3.0 (build_info, cache, doc, extensions, gems, specifications)
                              bigdecimal 1.2.8
                              io-console 0.4.5
                              json 1.8.3
                              psych 2.0.17
                              rdoc 4.2.1
installing bundle gems:       /mingw64/lib/ruby/gems/2.3.0 (build_info, cache, doc, extensions, gems, specifications)
                              did_you_mean 1.0.0
                              minitest 5.8.3
                              net-telnet 0.1.1
                              power_assert 0.2.6
                              rake 10.4.2
D:/builds/mingw-packages/mingw-w64-ruby/pkg/mingw-w64-x86_64-ruby/mingw64/lib/ruby/gems/2.3.0/gems/rake-10.4.2/bin/rake maybe `gem pristine rake` will fix it?
                              test-unit 3.1.5
```

If we could figure out why that error message is produced, that would probably lead to a solution.

DavidEGrayson:
The error message about rake comes from the file "pkg/mingw-w64-x86_64-ruby/mingw64/lib/ruby/gems/2.3.0/gems/rake-10.4.2/bin/rake" not being present.  It looks like that file is not present because actually all of the bundled gems are getting installed into this weird directory named `pkg/mingw-w64-x86_64-rubyingw64` inside the build directory.  Somewhere, somehow, "\m" or "/m" is getting replaced with the empty string.

DavidEGrayson:
Ruby's script for installing ruby can be be found in `tool/rbinstall.rb`.  It contains this function:

```
def without_destdir(dir)
  return dir if !$destdir or $destdir.empty? or !dir.start_with?($destdir)
  dir = dir.sub(/\A\w:/, '') if File::PATH_SEPARATOR == ';'
  dir[$destdir.size..-1]
end
```

The second line of that method just seems totally wrong to me.  It removes "D:" or "C:" from `dir` even though we already know that `dir` starts with `$destdir`.  Then it returns `dir` with $destdir.size characters removed from the beginning of the string.  Since the drive letter was removed, it's guaranteed to remove two more characters than desired.

My adding `puts` statements, I can see that this function transforms directories like "D:/builds/mingw-packages/mingw-w64-ruby/pkg/mingw-w64-x86_64-ruby/mingw64/lib/ruby/gems/2.3.0/gems/rake-10.4.2" into "ingw64/lib/ruby/gems/2.3.0/gems/rake-10.4.2".

DavidEGrayson:
The Ruby people fixed this very quickly today after I made a few GitHub comments:

https://github.com/ruby/ruby/commit/9cedfbcb918176627a9ed0a462480e52c9991c6c

I will be submitting a pull request to MINGW-packages later today.

PR #1074, opened 2016-02-16
ruby: Fix the installation of the bundled gems.

DavidEGrayson:
See #1070 for more discussion.

By the way, the ruby package installs some `.bat` and `.cmd` files to /mingw64/bin.  I'm not sure if we like that or not.

I can add myself as a maintainer for this package if you guys would like.  Ruby is my go-to language for a lot of things.


DavidEGrayson:
I can add myself as a maintainer for this package if you guys would like.

PR #1075, opened 2016-02-16
CI: Update core packages before full system upgrade.

fracting:
** Please review before merge **

Dear folks,

This pull request has two changes:

Firstly, it adds `C:\msys64\usr\bin\pacman --sync --refresh --refresh --needed --noconfirm msys2-runtime msys2-runtime-devel bash pacman pacman-mirrors` to appveyor.yml, this is a workaround, once we support `update-core --noconfirm`, the long version of command line could be replaced.

Secondly, it moves `Prepare` after `Detect`, so when no package is changed, we don't need to waste time and waste bandwidth on system upgrade.

@renatosilva how do you think about this solution, which is slightly different to what we discussed last time? Please share your thoughts, thanks!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,2 +1,4 @@
 build_script:
+    # TODO: implement update-core --noconfirm and replace the below command line
+    - C:\msys64\usr\bin\pacman --sync --refresh --refresh --needed --noconfirm msys2-runtime msys2-runtime-devel bash pacman pacman-mirrors
     - C:\msys64\usr\bin\bash --login -c "$(cygpath ${APPVEYOR_BUILD_FOLDER})/ci-build.sh"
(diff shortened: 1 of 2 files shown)


renatosilva:
This will immediately fix all builds currently failing on AppVeyor due to outdated core packages. Once [non-interactive update-core](https://github.com/Alexpux/MSYS2-pacman/pull/23) is available we can move the whole update process to the configuration files.

fracting:
Thanks for comment!

Please consider merge the commit, let's start to have fun with Appveyor :)

fracting:
Hi @mingwandroid , could you merge this pull request? Thanks!

mingwandroid:
Sorry guys. Been a bit distracted. Only saw this because I couldn't sleep.

PR #1077, opened 2016-02-21
use secure urls

vszakats:
* and follow some redirects
* prefer http over ftp


vszakats:
In the hope this is taken well. Similar effort has been started in Homebrew in early 2015 and it is now adopted as a standard.

There is a lot more similar updates possible, also in `MSYS2-packages`.

Another important update would be to move to SHA256 checksums (which is already being used in a few places) from SHA1 and especially from MD5. The latter two are not considered secure.

fracting:
@vszakats Wow, thanks for the patch.

Indeed, this is a good stress testing for our Continuous Integration server :)

Please forgive the Wine based Drone CI server if there is false positive when building so many packages, as far as I know there are still many todos, for examples, https://bugs.wine-staging.com/show_bug.cgi?id=649 and https://bugs.wine-staging.com/show_bug.cgi?id=646

Hopefully these bugs could be fixed and our CI server would be better and better in the coming future.

vszakats:
@fracting Well, it appears it will be a nice stress test indeed :) (sorry for the big patch BTW)

Haven't heard of http://wine-ci.org so far. It looks interesting.

A workaround may be to use a curl build with a non-schannel SSL backend (f.e. OpenSSL), to avoid potential OS-emulation traps. But probably it's even better to fix the latter ones.

vszakats:
@fracting Thank you very much for those Wine fixes!

fracting:
The stress test failed, you won :) 
We hit https://bugs.wine-staging.com/show_bug.cgi?id=404 again, and the CI build hangs.

The CI server can't speak English, but it asks me to say sorry to you :D 

Regarding wine-ci.org: Congrats, you find the hidden magic door! It's not publicly announced yet, mostly because some Wine bugs are still not fixed yet. MSYS2 people are very nice to allow me to test wine-ci.org using their repo as a playground, once bug 404 is fixed we will publicly announce wine-ci.org and provide it to open source projects as a free CI service.

Regarding those Wine fixes, it's not my own work, kudos to Wine Staging team.

I like your stress test. However, wine-ci.org has a hard limitation of 900 mins for one build, I don't think we can build 88 packages in 15 hours, AppVeyor CI has a hard limitation of 50 mins by default, not sure how they set for MSYS2, but I guess it won't be longer than 15 hours as well. We are also going to hit https://github.com/drone/drone/issues/1429#issuecomment-176020154 sooner or later by building 88 packages in one CI build.

However, the patch itself should be valid regardless how our CI server works, it is just a pity that our CI server doesn't help when we need it the most :(

@mingwandroid or @Alexpux , comment or merge on the pull request?

vszakats:
If there is a way to make reduced test that just downloads packages and verifies the checksums, it'd be enough as far as this commit is concerned. If checksums match, the commit is correct.

[ These are some tried and tested URL updates, but mistakes can be made of course. ]

Not exactly sure about this syntax though:
https://github.com/Alexpux/MINGW-packages/pull/1077/files#diff-2df7f22e3b646a0ebc5ff2468afa496dL26

mingwandroid:
Thanks! I'll test the cmake-git one, the -git packages go out of date quickly (by their very nature of course, my hope is to attract a developer linked to the project to MSYS2 with these from source-control packages, it's not really working, yet ..).

vszakats:
Thank you @mingwandroid. In some spare time I may create further similar PRs, also for MSYS2.

fracting:
@vszakats I like you idea, I hacked ci-build.sh to temporary disable building, and push your commit to a testing branch in order to trigger a new CI build, now it is downloading and testing:

http://wine-ci.org/fracting/MINGW-packages/17

vszakats:
@fracting That's perfect, thanks!

fracting:
ok, Appveyor CI runs out of disk space... https://ci.appveyor.com/project/fracting/mingw-packages/build/1.0.63

```
 
:: Proceed with installation? [Y/n] error: Partition / too full: 169687 blocks needed, 154580 blocks free
error: failed to commit transaction (not enough free disk space)
 
```

So our stress testing is really hard :)

vszakats:
Ops! (VMs have ~20GB)

fracting:
Update:

Here is some error detected by CI:

```
curl: (51) SSL: no alternative certificate subject name matches target host name 'dmake.apache-extras.org.codespot.com'\r
==> ERROR: Failure while downloading https://dmake.apache-extras.org.codespot.com/files/dmake-4.12.2.tar.bz2
```


```
==> ERROR: Failure while downloading https://gcc.gnu.org/pub/gcc/releases/gcc-5.3.0/gcc-5.3.0.tar.bz2\r
```

Reference: http://wine-ci.org/fracting/MINGW-packages/19

vszakats:
Thank you very much for your tests. I've made a new PR with the two fixes:
https://github.com/Alexpux/MINGW-packages/pull/1079

fracting:
Hi @vszakats , while you are here, as openssl contributor and expert, could I ask you if you have any suggestions or idea how to debug https://bugs.wine-staging.com/show_bug.cgi?id=649#c6 ? Any hints would be great appreciated, move to Wine Staging bugzilla for further discussion would be super appreciated, thanks!

vszakats:
ncurses: Only the homepage was updated in this patch. The failure is instead caused by linking to an unversioned "latest" version, which changes content under the same URL. It's doesn't appear to be the latest version either, nor version 6.0.

Version 6.0 is hosted here:
https://ftp.gnu.org/gnu/ncurses/ncurses-6.0.tar.gz

vszakats:
@fracting As for `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)` errors:

One cause may be that SSL peer and certificate verification have been [enabled by default](https://www.python.org/dev/peps/pep-0476/) starting Python 2.7.9. Of these two, _certificate verification_ will only work if the local SSL stack (OpenSSL in this case) has been configured with a valid and up-to-date list of accepted *CA certificates*. OpenSSL _doesn't_ currently support reading the Windows Certificate Store, so this list must be provided as a `.pem` file, like this one: https://curl.haxx.se/ca/cacert.pem

Errors like this may be typically caused by failing to configure/find/read such PEM file.

It _appears_ this job isn't handled by Python centrally, but each component offering support for SSL communication would need to load the PEM file by their own means.

On a standalone Python installation these PEM files can be found:
   `X:\Python27\Lib\site-packages\botocore\vendored\requests\cacert.pem`
   `X:\Python27\Lib\site-packages\pip\_vendor\requests\cacert.pem`

On an MSYS2 config, these ones:
   `X:\msys64\etc\pki\ca-trust\extracted\pem\tls-ca-bundle.pem`
   `X:\msys64\usr\ssl\cert.pem`
   `X:\msys64\usr\ssl\certs\ca-bundle.crt`

So first of all it should be verified if a fresh and valid PEM file can be found on the failing system, then one should find the Python component responsible for the HTTPS download and verify if that component is configured with an existing PEM file, and that the PEM file is indeed found and loaded correctly.

fracting:
@vszakats thanks a lot, you save me a lot of time!

PR #1079, opened 2016-02-21
two fixes to the previous, url update commit

vszakats:
Ref: https://github.com/Alexpux/MINGW-packages/pull/1077#issuecomment-186880520

### DIFF
--- a/mingw-w64-dmake/PKGBUILD
+++ b/mingw-w64-dmake/PKGBUILD
@@ -11,7 +11,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
 options=('staticlibs' 'strip')
 license=('GPL')
 url="https://code.google.com/a/apache-extras.org/p/dmake/"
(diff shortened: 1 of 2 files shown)


vszakats:
Not sure why the dmake failure in AppVeyor. The URL is now the same as it was and it's a valid one with matching checksum and internal folder.

fracting:
The dmake failure in AppVeyor is introduced by a bug of `file` when stdin is closed, as a result wrong file type is detected and wrong tool is used to decompress:

Appveyor CI:
```
==> Extracting sources...
  -> Extracting dmake-4.12.2.tar.bz2 with bzip2
==> Sources are ready.
 ```

Drone CI:

```
==> Extracting sources...
  -> Extracting dmake-4.12.2.tar.bz2 with bsdtar
==> Sources are ready.
```

The `file` bug is fixed in upstream development branch, but not in stable release yet, so `file` from AppVeyor is still buggy. Our Drone CI environment was customized and works in a slightly different way than AppVeyor where stdin is not closed, so we worked around the `file` bug in Drone CI.

vszakats:
That explains it, thanks for the insight!

PR #1085, opened 2016-02-22
ncurses 6.0

vszakats:
* fix to point to a versioned source URL
  (to have a stable content/checksum)
* fix source to be the latest 6.0 (was 5.9?)
* use sha256 checksums

Ref: https://github.com/Alexpux/MINGW-packages/pull/1077

### DIFF
--- a/mingw-w64-ncurses/PKGBUILD
+++ b/mingw-w64-ncurses/PKGBUILD
@@ -2,24 +2,22 @@
 
 _realname=ncurses
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-_base_ver=6.0
-_date_rev=20150627
-pkgver=${_base_ver}.${_date_rev}
-pkgrel=2
-pkgdesc="System V Release 4.0 curses emulation library"
+pkgver=6.0
+pkgrel=3
+pkgdesc="System V Release 4.0 curses emulation library (mingw-w64)"
 arch=('any')
 depends=("${MINGW_PACKAGE_PREFIX}-libsystre")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
 url="https://www.gnu.org/software/ncurses/"
 license=('MIT')
 options=('staticlibs' 'strip')
-source=("${_realname}-${pkgver}.tar.gz"::http://invisible-island.net/datafiles/current/${_realname}.tar.gz
+source=("https://ftp.gnu.org/gnu/ncurses/${_realname}-${pkgver}.tar.gz"
         001-use-libsystre.patch)
-md5sums=('a2bed858184157081aa433671d9a02c4'
-         'b669861903d0699b6535b7c6e028880d')
+sha256sums=('f551c24b30ce8bfb6e96d9f59b42fbea30fa3a6123384172f9e7284bcf647260'
+            'eb28949297a4c1eda67da49cfbed1d60f181b83101a0b3715552a4564bd90ff3')
 
 prepare() {
-  cd $_realname-${_base_ver}-${_date_rev}
+  cd $_realname-${pkgver}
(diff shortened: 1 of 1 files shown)


mingw-w64-ncurses/PKGBUILD
```diff
@@ -2,10 +2,8 @@
 
 _realname=ncurses
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-_base_ver=6.0
-_date_rev=20150627
-pkgver=${_base_ver}.${_date_rev}
-pkgrel=2
+pkgver=6.0
+pkgrel=3
 pkgdesc="System V Release 4.0 curses emulation library"
```
mingwandroid: this should also end in (mingw-w64)

mingw-w64-ncurses/PKGBUILD
```diff
@@ -13,13 +11,13 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
 url="https://www.gnu.org/software/ncurses/"
 license=('MIT')
 options=('staticlibs' 'strip')
-source=("${_realname}-${pkgver}.tar.gz"::http://invisible-island.net/datafiles/current/${_realname}.tar.gz
+source=("${_realname}-${pkgver}.tar.gz"::https://ftp.gnu.org/gnu/ncurses/${_realname}-${pkgver}.tar.gz
         001-use-libsystre.patch)
-md5sums=('a2bed858184157081aa433671d9a02c4'
-         'b669861903d0699b6535b7c6e028880d')
+sha256sums=('f551c24b30ce8bfb6e96d9f59b42fbea30fa3a6123384172f9e7284bcf647260'
```
mingwandroid: I wonder if we shouldn't standardise:

```
agm2 --nofilename -o "sha[1-9]*sum" | sort | uniq -c

118 sha1sum
181 sha256sum
  3 sha512sum
  7 shasum
```

I was going to suggest sha1sum, but it seems sha256sum is prevalent.

vszakats: Indeed, sha1sum is not considered secure since quite some time, and since last year there is wholesale migration _away_ from it, to sha256. (future candidates are sha3 and maybe blake)

mingw-w64-ncurses/PKGBUILD
vszakats: Fix pushed for that (and squashed).

vszakats:
On the second run, AppVeyor run onto an error when preparing the i686 build:
```
[...]
==> Finished making: mingw-w64-x86_64-ncurses 6.0-3 (Mon, Feb 22, 2016  2:37:10 AM)
==> Making package: mingw-w64-i686-ncurses 6.0-3 (Mon, Feb 22, 2016  2:37:12 AM)
==> WARNING: Using existing $srcdir/ tree
==> Starting prepare()...
patching file configure
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file configure.rej
    Aborting...
```

Alexpux:
Do we really need to downgrade ncurses?

vszakats:
No. But the old URL has a volatile content. F.e. now it has different content than at the time of creating this PR.

vszakats:
Sorry, I compared to the wrong URL - I used the one published on the homepage, which has the same filename, but it's 5.9. Nevertheless, the content changes, hence the failing checksum that implied my fix attempt. Short of stable URLs this is not fixable. Closing.

vszakats:
There _are_ stable URL, the latest is:
http://invisible-island.net/ncurses/current/ncurses-6.0-20160220.tgz

PR #1087, opened 2016-02-22
openssl: Add ca-certificates as optdepends.

fracting:
For https://github.com/Alexpux/MINGW-packages/issues/1086

### DIFF
--- a/mingw-w64-openssl/PKGBUILD
+++ b/mingw-w64-openssl/PKGBUILD
@@ -3,33 +3,35 @@
 _realname=openssl
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-_ver=1.0.2e
+_ver=1.0.2f
 # use a pacman compatible version scheme
 pkgver=${_ver/[a-z]/.${_ver//[0-9.]/}}
 pkgrel=1
 arch=('any')
 pkgdesc="The Open Source toolkit for Secure Sockets Layer and Transport Layer Security (mingw-w64)"
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs" "${MINGW_PACKAGE_PREFIX}-zlib")
+optdepends=("${MINGW_PACKAGE_PREFIX}-ca-certificates")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
 options=('strip' '!buildflags' 'staticlibs')
 license=('BSD')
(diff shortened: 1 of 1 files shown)


Alexpux:
I think the proper will be adding ca-certificates dependency to openssl instead python

fracting:
msys/openssl set ca-certificates as `optdepends`, should we set ca-certificates as `depends` or `optdepends` for mingw/openssl?

Alexpux:
As optdepends

fracting:
I updated the pull request, now ca-certificates is an optdepends of openssl.

Drone CI build passed.

```
New optional dependencies for mingw-w64-i686-openssl
    mingw-w64-i686-ca-certificates
```

However, if it is only optdepends, do we have to manually install mingw-w64-i686-ca-certificates when installing mingw-w64-i686-python2? I'm concerning about our `makepkg --syncdeps` script, will `--syncdeps` automatically install optdepends?

fracting:
mingw-w64-i686-ca-certificates is also needed when calling fossil to fetch https repo. fossil is needed when building mingw-w64-i686-sqlite-analyzer.

Any ideas?

DavidEGrayson:
I'd say that the CI server should just always install the CA certificates packages.   I think of them as an implicit dependency to a lot of things, similar to gcc and make.

fracting:
> I'd say that the CI server should just always install the CA certificates packages. I think of them as an implicit dependency to a lot of things, similar to gcc and make.

This is not a problem for me. I can install that in one command line.

However, for a normal new MSYS2 user, it would be a great pain if s/he run python / fossil / etc to download from https connection and fails everywhere. There is no document saying the user should manually install mingw-w64-i686-ca-certificates. I took a lot of time debugging it, if I didn't luckyly get help from openssl contributors like this time, I guess I'll spend much more time, not to say average users.

At very least, I guess we should add mingw-w64-i686-ca-certificates as a member of mingw-w64-i686-toolchain group. Does that sound reasonable?

vszakats:
I think this package should be part of the base MSYS2 installation, so nobody can miss installing it (but can update it later with the rest of MSYS2). It's a basic requirement for several secure protocols.

vszakats:
If it's part of the base system, it will be automatically installed on any CI that offers MSYS2. No need to open Issues with each.

Alexpux:
Well for downloading sources with pacman used MSYS version of curl with MSYS version of ca-certificates. Do not mix it. Mingw packages will never be in ````base```` group.

PR #1092, opened 2016-02-23
atom-editor, atom-shell: fix invalid checksum-types

vszakats:
* fix two `shasum` checksum types (with `SKIP` content) to `sha256sum`
* add a missing dependency

### DIFF
--- a/mingw-w64-atom-editor/PKGBUILD
+++ b/mingw-w64-atom-editor/PKGBUILD
@@ -10,7 +10,7 @@ arch=('any')
 url='https://atom.io'
 license=('MIT')
 options=()
-shasums=('SKIP')
+sha256sums=('SKIP')
 makedepends=("${MINGW_PACKAGE_PREFIX}-ninja"
              "${MINGW_PACKAGE_PREFIX}-python2"
              "git")
--- a/mingw-w64-atom-editor/PKGBUILD
+++ b/mingw-w64-atom-editor/PKGBUILD
@@ -11,6 +11,7 @@ url='https://atom.io'
 license=('MIT')
 options=()
 sha256sums=('SKIP')
+depends=("${MINGW_PACKAGE_PREFIX}-nodejs")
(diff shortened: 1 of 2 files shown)


fracting:
Hi @vszakats 

According to build log, mingw-w64-nodejs is missing as makedepends, could you update your patch please? Thanks!

BTW, due to a known bug, nodejs does not work on Wine yet, so the Drone CI is going to fail anyway: https://bugs.wine-staging.com/show_bug.cgi?id=626

I have no idea about the error in Appveyor CI, hopefully someone else understand nodejs could have a look.

vszakats:
Thanks @fracting, update committed. Fingers crossed.

fracting:
I pushed an workaround for winpty to Drone CI: https://github.com/fracting/msys32-docker/commit/e071a65af40bf54206183eb56d32fc046033812b

Now Drone CI fails in the same way with Appveyor CI:

```
==> Starting package()...
    Currently, attempting to make a log fails!
     so for now, please call without -L option
input is not a tty
    Aborting...
```

fracting:
Cross Reference:
Another related error when building on my local machine
https://bugs.wine-staging.com/show_bug.cgi?id=655

vszakats:
May or may not be the same issue, but resembles a bit to the error `./<script>: line <n>: 0: Bad file descriptor` I was getting when running certain (or may I say _any_) `Configure` scripts.

The solution was to open a "dummy file descriptor":
```sh
   exec 0</dev/null && ./configure
```

vszakats:
Reworked this to include the fix only. `sha521sums` isn't a problem at all.

vszakats:
https://github.com/Alexpux/MINGW-packages/pull/1092#issuecomment-187852152 was identified and subsequently fixed as an AppVeyor bug: https://github.com/appveyor/ci/issues/652

fracting:
> #1092 (comment) was identified and subsequently fixed as an AppVeyor bug: appveyor/ci#652

Great!

PR #1098, opened 2016-02-24
Improvements for continuous integration

renatosilva:
Some more improvements ([MSYS2-packages#460](https://github.com/Alexpux/MSYS2-packages/pull/460) already updated with this):

* Detected commits are now listed even when no package is built.
* Some code refactoring.


mingwandroid:
The single line coding style is not very easy to read IMHO here but you and fracting maintain it, so it's your call.

PR #1108, opened 2016-02-27
fixed missing symbols in mingw-w64-angleproject

Martchus:
* The \*.def files are ignored when not specified within the project file in the way this patch does.
  That's the reason why the \*.dll files in your package
  [mingw-w64-x86_64-angleproject-git-2.1.r4611-1-any.pkg.tar.xz](http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-angleproject-git-2.1.r4611-1-any.pkg.tar.xz)
  don't provide the symbols defined in the *.def file and hence Qt 5
  applications using ANGLE provided by your package will crash.

* Qt 5 WebKit works for me without the angleproject-export-shader-symbols.patch
  However, applying this patch results in errors. Hence I removed it.

* The \*.def files for win32 don't differ from the win64 version so it should
  be safe to remove them. I guess the ANGLE devs would provide a 32-bit version if
  it would be required.

* I didn't test whether this PKGBUILD compiles now. When developing
  [mingw-w64-angleproject](https://github.com/Martchus/PKGBUILDs/tree/master/angleproject/mingw-w64)
  I noticed that there are a few more adjustmens
  and additional header files required so using this PKGBUILD to build
  the current version will likely fail, too.

* The manual selection of object files is not valid for the current version.
  Static libs might be built automatically using the build system. (See
  [mingw-w64-angleproject](https://github.com/Martchus/PKGBUILDs/tree/master/angleproject/mingw-w64).)
  Hence I removed the comments.


mingwandroid:
The .def files came from the Qt Project, but as you say, they are only appropriate for 32-bit as they're all stdcall decorated! This'll be because the Qt Project never build mingw-w64 64-bit variants. I think they've added these files to make sure that the final DLLs export only exactly that which should be exported by those libraries, but yeah, it's completely busted for 64-bit!

Thanks a lot for the work. I'm merging because it built ok on wine-ci/drone and it can't have built before.

Martchus:
> The *.def files for win32 don't differ from the win64 version so it should
> be safe to remove them. I guess the ANGLE devs would provide a 32-bit version if
> it would be required.
This assumption was not correct. Hence this commit doesn't define the correct symbols for the 32-bit version. (Before my commits, there were no symbols at all for both architectures, so at least I didn't make it worse.) However, the previous version of the 32-bit *.def files doesn't work with the latest ANGLE. Hence I need to update them.

Additionally, there is a little adjustment required in the Qt 5 PKGBUILD (if ANGLE is used) because otherwise the required declarations of ANGLE are not available. This has nothing to do with my commit, its just because the ANGLE devs added a condition in their headers.

I'll push the required changed if I'm done.

PR #1113, opened 2016-02-28
move to sha256 checksums, fix some packages

vszakats:


Alexpux:
Well you break many packages with changes for secure urls. Are you test your changes?
Please revert back ALL packages that use git sources.

vszakats:
I maybe not understanding correctly what you mean by "git sources". In above PR, there is only one package that used github.com as a source (`lua-lpeg`). The previous URL was broken, so I looked around the internet to find the original package that was referred to. It was nowhere to be found, but found one which checked out with the previous MD5, changed in commit 63387d4ab6c330463c26dca6cf6c20f371387fc9.

My point is, the URL that was used last time doesn't exist, so there is no way it would have passed a test. Reverting to the non-GitHub archive seemed safe to do based on above referenced commit, which only updated the source and MD5. This patch reverts that (because the old source works again/now) and rechecksummed it with SHA256.

But maybe you have something else in mind.

Alexpux:
I'm about your previously merged pull requests. For example I revert url back for x264:
https://github.com/Alexpux/MINGW-packages/commit/34e37be572d844c0aa2bb023677687b4d3d63a3a#diff-5bf6d77a186990a2f478a75e98661a7cL16

vszakats:
Okay, thanks, I'll have a look at these. These might indeed be problematic at times, but all of these were tested (separately). One drawback of https/http Git access is missing support for shallow clones. Is this something used by `pacman` possibly?

Alexpux:
pacman check the URL, if it started with http* it use curl/wget to download sources

vszakats:
Of couse, I might have missed that :( So wouldn't the solution be to prefix the URL by `git+` to force a git clone rather?

Alexpux:
Well test it yourself and decide.

PR #1120, opened 2016-03-01
go: update to 1.6

ys-nuem:


Alexpux:
Is it building ok?

fracting:
Sorry for off-topic:

@Alexpux 

Due to a known bug Wine does not support go lang yet: Multiple games and applications need realtime updates to KSYSTEM_TIME members in KUSER_SHARED_DATA (Star Wars: The Old Republic game client, GO 1.4+ runtime)
 https://bugs.winehq.org/show_bug.cgi?id=29168

I have some idea, one is patch Wine to implement an ugly support for golang, one is to patch msys2 go to use QueryPerformanceCounter/QueryPerformanceFrequency/NtQueryPerformanceCounter instead of directly memory access.

Do you think it is considerable to patch msys2 go runtime in order to support our Wine CI until we find an not ugly solution in the Wine side? If that's not acceptable to MSYS2, I'll try to patch go runtime in an unofficial repo and use that go version in Wine CI. Otherwise, I'll provide the patch to you for review and once it has been tested for some time I'll try to ask upstream go community. If you prefer patch directly to upstream first I can also try as well, but I fear that upstream will reject the patch just because it is Wine only.

Alexpux:
Well I'm building packages in real Windows so if it buildable I will merge it. Problems with Wine will be solved some later. But if you have solution for Wine we will review it too.

fracting:
I see, thank you!
BTW, thanks to our Appveyor CI, it now works great, go package builds fine on Appveyor CI:
https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.123

(Appveyor is real Windows, not Wine, heh)

fracting:
Hmm, something weird happens on Appveyor CI, the x86_64 version of go builds fine, but the i686 version of go hangs at line 710, I don't understand why...

https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.123#L710
```
==> Starting prepare()...
patching file src/cmd/dist/buildtool.go
Hunk #1 succeeded at 111 (offset 3 lines).
patching file src/make.bat
```

PR #1128, opened 2016-03-04
Update Tesseract Pkgbuild to use 3.04.01 release

Shreeshrii:
https://github.com/tesseract-ocr/tesseract/releases

https://github.com/tesseract-ocr/tesseract/archive/3.04.01.tar.gz


fracting:
Hi, CI build failed, complaining checksum failed, could you double check? Thanks!

```
    tesseract-ocr-3.04.01.tar.gz ... FAILED
    build-fixes.patch ... Passed
```

Alexpux:
It fixed in my local tree. Will apply on next week as we now have 4 weekends

PR #1145, opened 2016-03-08
potrace: Add new package

ys-nuem:


ys-nuem:
Thank you for pointing it out.

PR #1154, opened 2016-03-09
tesseract Data for all langs

Shreeshrii:


Shreeshrii:
These are language data files for tesseract ocr and do not change based on the architecture or host.
Current package build goes thru the whole process twice, once for mingw32 and then for mingw64.

If there is a way to change it to do it just once and create both set of packages that will be great.
Thanks!

PR #1155, opened 2016-03-09
fix source files

Shreeshrii:
add ?raw=true for github blob files so that complete file is downloaded.
correct the sha256sums based on newly downloaded files.


Alexpux:
Does this included in tesseract-data already?

Alexpux:
I think this package need to be removed as it already in tesseract-data for all languages

Shreeshrii:
Yes, we can do that if tesseract-data is available.

I am not able to access tesseract-data.

Thanks for all your help.

ShreeDevi
____________________________________________________________
भजन - कीर्तन - आरती @ http://bhajans.ramparivar.com

PR #1162, opened 2016-03-09
PSEH Library from mingw-w64.

JPeterMugaas:
PSEH library for 32-bit versions of Windows that is part of the mingw-w64 source-code.  More info at:
http://reactos-blog.blogspot.com/2009/08/inside-mind-of-reactos-developer.html


Alexpux:
Well we not add it earlier because it builds only for 32-bit

JPeterMugaas:
I decided to close this for now but I suspect that the PSEH library might be more important than we think.  The library itself seems to provide some type of Structured Exception Handling for GCC in Win32.
https://mingwpy.github.io/background-mingw.html .   The author of PSEH did make a blog entry about the library  http://reactos-blog.blogspot.com/2009/08/inside-mind-of-reactos-developer.html and there is an Wiki entry at http://www.reactos.org/wiki/PSEH about the library.

The reason it is 32-bit only is that GCC since 4.6 has supported the 64-bit SEH .  32-bit SEH was covered by a Borland patent that expired   But SEH was the default in the 32-bit version of Windows.

For everyone's info:

https://www.microsoft.com/msj/0197/exception/exception.aspx

I'm not completely sure I understand the situation myself but since I saw it in the mingw-w64 source-code, I felt it should not be disregarded.

mingwandroid:
Does this not need to get into GCC then?

JPeterMugaas:
Perhaps, it might but I don’t know.  And it may be possible that people might want to use it outside of GCC for their own projects or so that it could be shared by something.  It was a library for a reason, I think.  SEH is not native to GCC 32-bits which is why the library might be necessary (I don’t know for sure).   But I don’t think the library is needed for 64-bit software as SEH is different.

PR #1165, opened 2016-03-10
libbsdf: new package (DO NOT MERGE yet!!!)

mkae:
"makepkg-mingw -sLf" is locally crashing Windows Vista with a BAD_POOL_DATA blue screen here on my end.
         Want to see for now whether the CI systems behave similarly.


mkae:
This build failure on wine-ci.org I find disturbing:
```
-- Build files have been written to: C:/msys32/drone/src/github.com/Alexpux/MINGW-packages/mingw-w64-libbsdf/src/libbsdf-0.9.1/build-i686-w64-mingw32

wine: Bad EXE format for Z:\bin\sh..
Makefile:1076: recipe for target 'cmake_check_build_system' failed

process_begin: CreateProcess(NULL, /bin/sh -c "/C/msys32/mingw32/bin/cmake.exe -H/C/msys32/drone/src/github.com/Alexpux/MINGW-packages/mingw-w64-libbsdf/src/libbsdf-0.9.1 -B/C/msys32/drone/src/github.com/Alexpux/MINGW-packages/mingw-w64-libbsdf/src/libbsdf-0.9.1/build-i686-w64-mingw32 --check-build-system CMakeFiles/Makefile.cmake 0", ...) failed.

make (e=193): Error 193

mingw32-make: *** [cmake_check_build_system] Error 193

    Aborting...
==> Removing installed dependencies...
checking dependencies...
```

mkae:
@fracting, if all checks have passed, why does Appveyor claim that there is an error due to a "duplicate target" at the end:
```
Packages (1) mingw-w64-i686-eigen3-3.2.8-1
 
Total Removed Size:  3.46 MiB
 
:: Do you want to remove these packages? [Y/n] 
removing mingw-w64-i686-eigen3...
 
[MSYS2 CI] mingw-w64-libbsdf: Installing
 
error: 'libbsdf-0.9.1-1-x86_64.pkg.tar.xz': duplicate target
loading packages...
 
[MSYS2 CI] FAILURE: mingw-w64-libbsdf: Installing failed.
 
 
[MSYS2 CI] SUCCESS: All packages built and installed successfully.
 
Discovering tests...OK
Build success
```

fracting:
```
Bad EXE format for Z:\bin\sh.
```

This looks like a known bug of Wine, I'll take a look deeper later when I back from business trip ;)

mkae:
OK, thanks for checking on it. :)

mkae:
@fracting, I don't understand why wine-ci fails because of the `toolchain`...

mingw-w64-libbsdf/PKGBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer: Marko Käning <mk-lists@mailbox.org>
+
+pkgname=("libbsdf")
+pkgver=0.9.1
+pkgrel=1
+epoch=
+pkgdesc="libbsdf library (libraries)"
+arch=("i686" "x86_64")
```
Alexpux: Here need to be "any". See other PKGBUILDs

mingw-w64-libbsdf/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+url="https://github.com/KimuraRyo/libbsdf"
+license=('Mozilla Public License, version 2.0')
+groups=('libraries')
```
Alexpux: Remove this

mingw-w64-libbsdf/PKGBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer: Marko Käning <mk-lists@mailbox.org>
+
+pkgname=("libbsdf")
```
Alexpux: Wrong!!!!
You write MINGW package not MSYS. See header from any PKGBUILD

mingw-w64-libbsdf/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+source=("${pkgname}"::"https://github.com/KimuraRyo/${pkgname}/archive/v${pkgver}.tar.gz")
+sha256sums=('8f92a1d679adffc1fafc8c53bc397a5bb49ce07678e52630bbc1ca16817d350d')
+
+build() {
+  cd "${srcdir}/${pkgname}-${pkgver}"
+  [[ -d build-${MINGW_CHOST} ]] && rm -rf build-${MINGW_CHOST}
+  mkdir build-${MINGW_CHOST}
+  cd build-${MINGW_CHOST}
+  cmake -G"MSYS Makefiles" -DEIGEN3_INCLUDE_DIR=${MINGW_PREFIX}/include/eigen3 -DCMAKE_CXX_FLAGS=-std=c++11 ..
+  mingw32-make
```
Alexpux: Use regular make instead mingw32-make.

mingw-w64-libbsdf/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+makedepends=("${MINGW_PACKAGE_PREFIX}-ninja" "${MINGW_PACKAGE_PREFIX}-cmake" "${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-eigen3")
+source=("${pkgname}"::"https://github.com/KimuraRyo/${pkgname}/archive/v${pkgver}.tar.gz")
+sha256sums=('8f92a1d679adffc1fafc8c53bc397a5bb49ce07678e52630bbc1ca16817d350d')
+
+build() {
+  cd "${srcdir}/${pkgname}-${pkgver}"
+  [[ -d build-${MINGW_CHOST} ]] && rm -rf build-${MINGW_CHOST}
+  mkdir build-${MINGW_CHOST}
+  cd build-${MINGW_CHOST}
+  cmake -G"MSYS Makefiles" -DEIGEN3_INCLUDE_DIR=${MINGW_PREFIX}/include/eigen3 -DCMAKE_CXX_FLAGS=-std=c++11 ..
```
Alexpux: See any package that building with cmake and do the same. Split arguments one per line, you need set MSYS2_ARG_CONV_EXCL before cmake.

Alexpux:
This is really BAD PKGBUILD. See for example https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-eigen3/PKGBUILD and do the same.

mingwandroid:
There are also templates that people can use as a starting point. mingw-w64-PKGBUILD-templates.

Cheers!

fracting:
```
wine: Bad EXE format for Z:\bin\sh..
```

This looks similar to one known Wine bug, I'll do some research and come back later.

mkae:
Alexey, sorry for all the silly beginner's mistakes. These are my first PKGBUILD definitions ever. They weren't meant to get merged anyway until the issues are ironed out.
Building the package `libbsdf` was giving me blue screens reproducibly and the abuse of the CI system seemed the only way forward.
I agree it is really bad and I will crosscheck against the examples suggested in this thread. Thanks again for all the help!

mingwandroid:
Don't worry, you'll get there. I till make some stupid mistakes

mingw-w64-libbsdf/PKGBUILD
mkae: So, group "library" is only for MSYS2-specific builds? Could I use perhaps "science" as value?

mingw-w64-libbsdf/PKGBUILD
mkae: Sorry, yes, will fix that.

mingw-w64-libbsdf/PKGBUILD
mkae: I've found the advice to use the above on your wiki at SF. I must have misunderstood something.

Will correct that too.

mingw-w64-libbsdf/PKGBUILD
Alexpux: We not use groups in general for mingw packages.

mingw-w64-libbsdf/PKGBUILD
mkae: Will do so.

mkae:
So, now that I changed away from `mingw32-make` to normal `make` my machine doesn't crash with a blue screen anymore. :-) Do these crashes deserve attention for some investigation? I know now that I should be using the correct `make` but yet any other program shouldn't let my Vista crash with a blue screen...

Hoping that the CI systems can verify that once they've caught up...

mkae:
I guess I don't need the runtime dependency defined in 226f5c4, as that is only a library the application links against at build time.

Is this assumption correct?

@fracting, yet wine-ci is still failing to build `BSDFProcessor`, as it misses `mingw-w64-x86_64-toolchain`, which but wasn't needed here on my machine to get it to build successfully. Is that perhaps caused by a different behaviour of 32 and 64 bit versions?

AppVeyor still fails to build `BSDFProcessor` for the old reason of not finding package `libbsdf`. What is going on?

mkae:
@fracting, and finally wine-ci is gone, thus appveyor is probably failing on me soon again.

Giving up for now.

mkae:
TODO: Package name "*BSDFProcessor" needs to be converted into lower case, as all package names have to be in lower case on MSYS2.

mkae:
Done.

Wondering why the generated Makefiles do not contain an install block, which is why I was forced to come up with my own copying of stuff into the right places. Only now that I committed I see that I am missing `CMAKE_INSTALL_PREFIX` in the CMake call... I guess this is the reason for it. No more tonight. Well, I hope that's the last missing step to finalise this.

fracting:
I can't reproduce the ```wine: Bad EXE format for Z:\bin\sh..``` bug either on my local Ubuntu or on the CI server, while it looks similar to https://bugs.winehq.org/show_bug.cgi?id=37640, but I'm still not sure yet.

Let's skip it for now, if it happens again I'll investigate deeper.

http://wine-ci.org/fracting/MINGW-packages/54

mkae:
OK, thanks @fracting for looking into it though! :-)

PR #1167, opened 2016-03-10
Upgrade to 0.16

JPeterMugaas:


Alexpux:
Latest is 0.16.1. Fix it

PR #1169, opened 2016-03-11
Fix validpgpkeys.

JPeterMugaas:
You have to remove dirs because mv will not overwrite dirs and files.

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -130,23 +130,15 @@ prepare() {
   # See http://llvm.org/bugs/show_bug.cgi?id=4840
 
   cd "${srcdir}"/llvm-${pkgver}.src
-  # mv will not overwrite something that already exists so do "rm -rf"
-  rm tools/clang | true
-  mv "${srcdir}/cfe-${pkgver}.src"               tools/clang
-  rm -rf tools/clang/tools/extra | true
-  mv "${srcdir}/clang-tools-extra-${pkgver}.src" tools/clang/tools/extra
-  rm -rf tools/lld | true
-  mv "${srcdir}/lld-${pkgver}.src"               tools/lld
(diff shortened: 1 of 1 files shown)


mingw-w64-clang/PKGBUILD
```diff
@@ -132,13 +130,22 @@ prepare() {
   # See http://llvm.org/bugs/show_bug.cgi?id=4840
 
   cd "${srcdir}"/llvm-${pkgver}.src
+  # mv will not overwrite something that already exists so do "rm -rf"
+  rm tools/clang | true
```
Alexpux: There are nothing can be there. Will not accept "rm"

JPeterMugaas:
Let's see if this works better.

mingw-w64-clang/PKGBUILD
```diff
@@ -132,14 +130,15 @@ prepare() {
...
-  #mv "${srcdir}/testsuite-${pkgver}.src"         projects/test-suite
+# note you need to ignore failure if there is stuff that already exists
+  mv "${srcdir}/cfe-${pkgver}.src"               tools/clang | true
+  mv "${srcdir}/clang-tools-extra-${pkgver}.src" tools/clang/tools/extra | true
+  mv "${srcdir}/lld-${pkgver}.src"               tools/lld | true
+  #mv "${srcdir}/lldb-${pkgver}.src"              tools/lldb | true
+  mv "${srcdir}/compiler-rt-${pkgver}.src"       projects/compiler-rt | true
+  mv "${srcdir}/libcxxabi-${pkgver}.src"         projects/libcxxabi | true
+  mv "${srcdir}/libcxx-${pkgver}.src"            projects/libcxx | true
+  #mv "${srcdir}/testsuite-${pkgver}.src"         projects/test-suite | true
```
Alexpux: in case of fail move will not be done and you get wrong source tree

PR #1171, opened 2016-03-11
add libunwind and fix LLVM build issues.

JPeterMugaas:
libunwind 3.8.0 from LLVM distribution.
attempt to fix issues with validkey and ignore mv command failures in prepare if the target dir exists.


Alexpux:
Can lib unwind be builded together with clang/llvm as libc++ and others?

JPeterMugaas:
I'm going to try but in the meantime, you may want to consider adding two major bit I tried to get in:

1) It turns out that the source directory is not emptied after a build.  The code was using the "mv" command which would fail in that case.
2) When I tried your clang packages, I found that the signatures would not validate.  There were two validpgpkey lines.  The proper line is below and it should be placed right under the checksum lines:

 validpgpkeys=('B6C8F98282B944E3B0D5C2530FC3042E345AD05D') # Hans Wennborg, Google.

PR #1207, opened 2016-03-18
SDL_mixer: move openal to makedepends

fabiangreffrath:
It turns out that openal is not a run-time dependency of sdl_mixer, not even of libmikmod. It is merely needed to build the test compile triggered by sdl_mixer's configure script, because the string returned by `libmikmod-config --libs` contains `-lopenal`.

### DIFF
--- a/mingw-w64-SDL_mixer/PKGBUILD
+++ b/mingw-w64-SDL_mixer/PKGBUILD
@@ -4,7 +4,7 @@ _realname=SDL_mixer
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
 pkgver=1.2.12
-pkgrel=4
+pkgrel=5
 pkgdesc="A simple multi-channel audio mixer (mingw-w64)"
 arch=('any')
 url="https://libsdl.org/projects/SDL_mixer/"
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
I stand corrected! Something must be wrong with MSYS2's ldd:
```
$ ldd /mingw32/bin/libmikmod-3.dll
        ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x77860000)
        KERNEL32.DLL => /c/windows/SYSTEM32/KERNEL32.DLL (0x75870000)
        KERNELBASE.dll => /c/windows/SYSTEM32/KERNELBASE.dll (0x74b10000)
```
whereas
```
$ objdump -x /mingw32/bin/libmikmod-3.dll | grep 'DLL Name'
        DLL Name: dsound.dll
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
        DLL Name: USER32.dll
        DLL Name: WINMM.DLL
        DLL Name: libopenal-1.dll
        DLL Name: libgcc_s_dw2-1.dll
```
So, as it turns out, openal *is* a run-time dependency of libmikmod, but this is nowhere reflected in the package's dependencies.

fabiangreffrath:
> but this is nowhere reflected in the package's dependencies.

That is, openal is merely among libmikmod's makedepends but not among its depends.

maoueh:
@fabiangreffrath Check out with `ntldd` for software in `/mingw64`, I recall once having problem with `ldd` while `ntldd` gave "better" results. It's a shot in the dark, but worth to try :)

PR #1209, opened 2016-03-20
Add optimisations for non-MSVC compilers...

mkae:
... for lib and app.

Also fix for upper case still used for app's exe, since that leads to trouble in `makepkg-mingw`!


mkae:
Don't understand why drone doesn't find the patch file:
```
==> ERROR: optimisations-CMakeLists.txt.patch was not found in the build directory and is not a URL.
```
Could it be that this is caused by the upper case in the package's directory name?

fracting:
Hi @mkae the missing patch was putting into a separate directory: https://github.com/mkae/MINGW-packages/tree/master/mingw-w64-bsdfprocessor

It is interesting that this work on Windows, which means there is some bug in Wine + MSYS2, thanks for helping me find Wine bugs ;) However, the patch itself should also be fixed, could you double check?

Thanks!

mkae:
Hi, @fracting, I don't think it's a real bug in wine. Guess it just means that wine is running on a case-sensitive file system, while AppVeyor operated on a case-insensitive one. No?

mkae:
But yes, the repo itself has files for this package in two separate folders, with upper and lower case, as I see only now. So, my local clone on my Vista machine is thus different from my own repo at github!!

mkae:
Well, and so AppVeyor must be running on a case-insensitive file system...

Okay, will try to fix this with a checkout of my repo on Linux, in order to work on a case-sensitive file system.

PR #1226, opened 2016-03-25
wxwidgets 3.1.0

ncihnegn:
3.0.2 has this bug:
assert "m_menuDepth > 0" failed in wxTopLevelWindowMSW::DoSendMenuOpenCloseEvent(): No open menus?"

### DIFF
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -63,6 +63,7 @@ prepare() {
 
   # Add a manual page for bjam, based on the on-line documentation:
   # http://www.boost.org/boost-build2/doc/html/bbv2/overview.html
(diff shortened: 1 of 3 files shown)


Alexpux:
Is wxWidgets are compatible with all our wxWidget packages?

ncihnegn:
From wxwidgets.org:

"Please notice that while 3.1.0 is officially a "development" version because it is not fully compatible with 3.0.x, the list of backwards incompatible changes is very short, so you shouldn't have any problems updating to this version from 3.0.x in practice, and you're encouraged to try this version out and use it in production in spite of its odd version number."

Alexpux:
I think we will skip this version now

DoctorWho11:
Best to wait for the next release of wx 3.1 once 3.0 backwards compatibility is fixed

PR #1230, opened 2016-03-25
glbinding: Updated to 2.0.0 (shared library only)

nazar554:
The source tarball contains `CMakeLists.txt` with CRLF line endings, so I made a CRLF patch but git wants to make it LF. How to go around this?
Also upstream started using CMake export files, so side by side static and shared installation is not possible without separate prefix. Is it possible to make a `glbinding-static` package that conflicts with shared one?


nazar554:
I figured out a workaround, don't know if it's okay

Alexpux:
You can safely convert patch to have LF line endings

nazar554:
But I still need to make it CRLF in `prepare()`? Because otherwise patch fails due to different line endings.

Alexpux:
No it apply patch with different line endings

nazar554:
Looks like it worked. Want me to rebase this into a single commit?

PR #1243, opened 2016-03-26
Fix depend and bump version

mati865:
No problems locally.

### DIFF
--- a/mingw-w64-luajit-git/PKGBUILD
+++ b/mingw-w64-luajit-git/PKGBUILD
@@ -3,13 +3,13 @@
 _realname=luajit
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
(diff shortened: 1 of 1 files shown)


mingwandroid:
Please include the package you are changing in the commit message summary, such as:

`luajit-git: Fix depend and bump version`

mati865:
Fixed, sorry for inconvenience. I'm new to git.

mingwandroid:
No problem. Git is exceptionally good, if somewhat confusing. What you want to do is to squash all these commits into one using what's known as an interactive rebase then a force push (you might want to make sure your EDITOR variable is set to an editor that you can use first, as `vim` isn't the most user friendly. `nano` is more friendly).

```
git rebase -i HEAD~3
git push origin master --force
```

You will sometimes read that force pushes are frowned upon, but they are a core part of the github fork / pull request workflow. They are frowned upon in a different context; it is the upstream that cannot be force-pushed to, not the fork-for-PR clones.

PR #1254, opened 2016-03-28
[WIP] qt5-git: updated to actually clone and build a Qt5 GIT branch

mcallegari:
Submitting as WIP to start the review process.
At the moment, the prepare phase is OK and it also starts building a few libraries.

Unfortunately it fails when trying to build Qt5OpenGl due to an avalanche of errors that I hope you can help to figure out.
The opengl mode passed to configure is indeed "desktop" but apparently the build process tries to make also opengl ES stuff, and find unresolved macros.
This is an example: http://pastebin.com/KvfFJb5Q


aquiles2k:
Sry if i ask something obvious - how exactly do i use this, like what command do i need to enter on the msys shell to start a qt5-git build?

mati865:
@aquiles2k Keep in mind that compilation can take several hours per architecture depending on your machine.
You can use these commands and let it work for few hours:
```
pacman --needed -Sy base-devel
cd <build/directory>
git clone https://github.com/Alexpux/MINGW-packages.git
cd MINGW-packages/mingw-w64-qt5-git
makepkg-mingw -sfi --noconfirm
```

PR #1258, opened 2016-03-30
Remove optimization from libkml

dsogari:
I got an obscure segmentation fault when using libkml with gdal. I believe it was caused by a recent update of some library dependency, as it did not happen before (a week maybe), but I can't know for sure. This is the safest fix.

Also, the unit tests are not working anymore in 32-bit and I have no idea why. So I disabled the check() procedure from PKGBUILD.

Other than that, everything is fine.

### DIFF
--- a/mingw-w64-libkml/PKGBUILD
+++ b/mingw-w64-libkml/PKGBUILD
@@ -47,13 +47,13 @@ build() {
     --prefix=${MINGW_PREFIX} \
     --disable-java
 
-  make
+  make CXXFLAGS="-O0"
 }
 
-check() {
-  cd "${srcdir}/build-${MINGW_CHOST}"
-  make check
-}
+#check() {
+#  cd "${srcdir}/build-${MINGW_CHOST}"
+#  make check
+#}
 
 package() {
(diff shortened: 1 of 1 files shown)


dsogari:
Actually, the writing of some kml content using the libkml driver in gdal is giving errors. I'll have to look further into it.

dsogari:
It had to do with the LIBKML_STRICT_COMPLIANCE config option of gdal. When it is enabled, the IsValid method (which relies on the geos library) is called on geometries and this could cause seg fault. With LIBKML_STRICT_COMPLIANCE set to "NO" the seg fault can be avoided, to the expense of losing geometry validity checks.

Unfortunately, this is a run-time workaround that users of gdal should be aware of. I could not find another way to solve the issue. However, it is related to gdal and does not affect the correctness of libkml alone.

dsogari:
I just discovered that compiling geos without optimizations solves the issue with kml writing. This way, the LIBKML_STRICT_COMPLIANCE config option of gdal does not need to be set. I guess I'll close this pull request and keep using my own compiled version of geos and libkml.

PR #1272, opened 2016-04-01
Add wkhtmltopdf

dsogari:
Add wkhtmltopdf library (http://wkhtmltopdf.org/).

Note: I could not use the release tarball from version 0.12, as it depends on Qt 4.8. I had to take the development version from git branch labeled "0.13", which is already using Qt 5.x

### DIFF
--- a/mingw-w64-wkhtmltopdf/PKGBUILD
+++ b/mingw-w64-wkhtmltopdf/PKGBUILD
@@ -1,0 +1,72 @@
+# Maintainer: Diego Sogari <diego.sogari@gmail.com>
+
+_realname=wkhtmltopdf
+pkgbase=mingw-w64-${_realname}-git
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+_basever=0.13
+pkgver=0.13.r925.e03f1d1
+pkgrel=1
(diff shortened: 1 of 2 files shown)


mingw-w64-wkhtmltopdf/PKGBUILD
```diff
@@ -0,0 +1,73 @@
...
+build() {
+  cd "${srcdir}/${_realname}"
+
+  ln -fs `which windres` ./${MINGW_CHOST}-windres
+  export PREFIX=${MINGW_PREFIX}
+  PATH=`pwd`:$PATH
+
+  # detect architecture
+  if [[ ${MINGW_CHOST} == *"i686"* ]]; then
+    target=mingw-w64-cross-win32
```
Alexpux: Why cross?

dsogari: I could not find a way to build without the "scripts/build.py", and this script only specifies two options for MinGW: "mingw-w64-cross-win32" and "mingw-w64-cross-win64".

The cross in the name means that the project should be built on Linux hosts, targeting Windows. I believe this is the choice for us, even if in our case the host is Windows.

What do you think?

PR #1273, opened 2016-04-02
jucipp-git: Update PKGBUILD

eidheim:
Feel free to make this package installable through pacman -S. juCi++ has been very stable since v1.0.0 release, and I'll update the pkgver whenever we have major changes.


Alexpux:
We prefer release packages instead git. So git PKGBUILD's n repo only for developers. Commit PKGBUILD for release.

PR #1276, opened 2016-04-02
Fix crash in libidn on Win64

Karlson2k:
As title says.
Patch 0003 is most important.
Upstream was informed but no reaction for a few days.
Discovered while debugging crash in curl's simple command `curl -v http://яндекс.рф`.
More info:
https://github.com/curl/curl/issues/731
http://lists.gnu.org/archive/html/help-libidn/2016-03/msg00002.html


Alexpux:
Increase ````pkgrel````

Karlson2k:
@Alexpux Done

PR #1279, opened 2016-04-03
python-numpy: add patch to disable 64-bit warning

dthpham:
Every time a user, or another python program, imports numpy on a 64-bit Windows machine, it prints a warning telling people to report bugs upstream.

```
>>> import numpy
C:\msys64\mingw64\lib\python2.7\site-packages/numpy/core/__init__.py:14: Warning: Numpy built with MINGW-W64 on Windows 64 bits is experimental, and only available for
testing. You are advised not to use it for production.

CRASHES ARE TO BE EXPECTED - PLEASE REPORT THEM TO NUMPY DEVELOPERS
  from . import multiarray
```

Now, I use numpy and programs that depend on it a lot, and seeing this thing pop up all the time in my terminal is extremely annoying. I'm positive that other developers that use the 64-bit numpy package would appreciate that the warning be disabled by default, so here's a patch that takes it out.


mingwandroid:
Out of interest, how well does numpy work for you on MSYS2? How much do you stress it out and how does it compare with CPython's numpy?

PR #1285, opened 2016-04-05
python-cx_Freeze: add patch to fix manifest arch

dthpham:
This patch fixes `ImportError: DLL load failed: %1 is not a valid Win32 application.` when running modules packaged by cx_freeze on 64bit machines. This has to do with mixing Python 64 with 32bit modules. Switching the application manifest's `processorArchitecture="x86"` to `processorArchitecture="*"` fixes this problem. The star indicates that all processor architectures are supported.

This was [fixed upstream](https://bitbucket.org/anthony_tuininga/cx_freeze/pull-requests/71/changed-x86-in-windows-manifest-to/diff) in cx_Freeze 4.3.4. This patch wouldn't be necessary if the package itself was updated, but I was never able to get the latest version to successfully build.

### DIFF
--- a/mingw-w64-python-cx_Freeze/0001-fix-manifest-processor-arch.patch
+++ b/mingw-w64-python-cx_Freeze/0001-fix-manifest-processor-arch.patch
@@ -1,0 +1,21 @@
+diff -urN cx_Freeze-4.3.3.orig/source/bases/manifest.txt cx_Freeze-4.3.3/source/bases/manifest.txt
+--- cx_Freeze-4.3.3.orig/source/bases/manifest.txt	2012-02-27 07:52:03.000000000 -0700
++++ cx_Freeze-4.3.3/source/bases/manifest.txt	2016-03-30 08:16:18.423882000 -0700
+@@ -8,7 +8,7 @@
+   </trustInfo>
+   <dependency>
+     <dependentAssembly>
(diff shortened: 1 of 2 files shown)


mingw-w64-python-cx_Freeze/0001-fix-manifest-processor-arch.patch
```diff
@@ -0,0 +1,21 @@
...
++      <assemblyIdentity type="win32" name="Microsoft.VC90.CRT" version="9.0.21022.8" processorArchitecture="*" publicKeyToken="1fc8b3b9a1e18e3b"></assemblyIdentity>
+     </dependentAssembly>
+   </dependency>
+   <dependency>
+@@ -17,7 +17,7 @@
+               type="win32"
+               name="Microsoft.Windows.Common-Controls"
+               version="6.0.0.0"
+-              processorArchitecture="X86"
++              processorArchitecture="*"
```
Alexpux: You think it can used like this?

dthpham:
I've ran into this `ImportError` problem myself, and this patch fixed it. This same patch was merged upstream and there's a comment on a [MS documentation page](https://msdn.microsoft.com/en-us/library/aa374191%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) on application manifests that says `*` is a valid option. I'm pretty sure it won't break anything.

Thanks for merging it. :+1:

PR #1289, opened 2016-04-06
mingw-w64-python-path: use secure url

vszakats:


vszakats:
CI failure unrelated to this patch.

PR #1294, opened 2016-04-07
qt5: updated sha256 for qtwebkit-opensource-src-5.6.0.tar.xz

emmenlau:
For update from Wed, 06 Apr 2016 08:55:14 GMT (Unix time: 1459932914)

### DIFF
--- a/mingw-w64-libjpeg-turbo/0002-md5sum-fix.patch
+++ b/mingw-w64-libjpeg-turbo/0002-md5sum-fix.patch
@@ -1,0 +1,73 @@
+diff -Naur libjpeg-turbo-1.4.2/md5/md5cmp.c libjpeg-turbo-1.4.2.new/md5/md5cmp.c
+--- libjpeg-turbo-1.4.2/md5/md5cmp.c	2015-09-21 20:48:32.000000000 +0200
++++ libjpeg-turbo-1.4.2.new/md5/md5cmp.c	2016-04-07 21:37:46.396809300 +0200
+@@ -1,5 +1,5 @@
+ /*
+- * Copyright (C)2013 D. R. Commander.  All Rights Reserved.
++ * Copyright (C)2013, 2016 D. R. Commander.  All Rights Reserved.
+  *
+  * Redistribution and use in source and binary forms, with or without
+  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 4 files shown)


mati865:
Have you compiled it ourself?
@adrpo had error while compiling it https://github.com/Alexpux/MINGW-packages/issues/1283#issuecomment-206502282

adrpo:
@mati865: that might have been my error as I removed the Qt5 debug dlls from my msys2 installation and strangely they are needed to compile Qt5. I'm trying to compile again now.

emmenlau:
I think this patch is required, but it does not address the failing build. It only fixes the checksum, see https://tea-ci.org/Alexpux/MINGW-packages/566
```
...
    qt-everywhere-opensource-src-5.6.0.tar.xz ... Passed
    qtwebkit-opensource-src-5.6.0.tar.xz ... Passed
...
```

mati865:
Bots are buggy. And probably it would be better to see fail in 30 seconds than after 4 hours of compiling.

emmenlau:
@mati865 ok I see your point. But at least minimal builds seem to work, whatever that's worth.

DavidEGrayson:
I just built Qt5 myself a few days ago and the checksum (d5c4daac8eef1b24290cda59db432ccc390761af17661f64f3028482647521a1) for qtwebkit-opensource-src-5.6.0.tar.xz was correct.  It looks like it has indeed changed to 9ca72373841f3a868a7bcc696956cdb0ad7f5e678c693659f6f0b919fdd16dfe since then, as this pull request is implying.

That is kind of annoying; does anyone know why the Qt people would change their released tarball without updating its version number or something?

DavidEGrayson:
Well, I still have the old version of qtwebkit-opensource-src-5.6.0.tar.xz on my computer so I was able to compare them.  The new version adds the `qtwebkit-opensource-src-5.6.0/include` directory, which contains a bunch of header files for QtWebKit and QtWebKitWidgets.  Other than that, `diff -ur` reports no differences between the two.  I wonder if the addition of that directory has any implications for our PKGBUILD script.

Alexpux:
They update QtWebkit tar all because they forgot to synchronize with at headers. See my workaround in PKGBUILD where I run syncqt.

PR #1297, opened 2016-04-07
libjpeg-turbo: fix md5 code with backport from trunk

emmenlau:
Fixes a check() error that md5cmp hangs on all inputs


Alexpux:
increase pkgrel also

PR #1304, opened 2016-04-10
icu: Update to 57.1

skystrife:


Alexpux:
Well I will accept it but not build before New qt5 build as need rebuild many packages with icu update

skystrife:
Is that just finding every `PKGBUILD` that `depends` on icu and bumping `pkgrel`? Or is it more involved than that? (Wondering if there's anything I can do to help.)

mingwandroid:
Here is a list of the API changes between ICU 56 and ICU 57.

http://source.icu-project.org/repos/icu/icu/tags/release-57-1/APIChangeReport.html

Some projects may break because of these changes.

So it might require looking for patches on the source control repositories of those projects and adding them to MSYS2 but the main time concern for Alexey is that he must rebuild and retest everything.

PR #1307, opened 2016-04-11
mingw-w64-ghostscript: refresh patches

darealshinji:
Patches refreshed. Also removed `configure' from ghostscript-sys-zlib.patch since
the PKGBUILD will run autoconf anyway.


Alexpux:
Update SHA checksums and pkgrel

darealshinji:
Done.

PR #1314, opened 2016-04-12
Master

JPeterMugaas:
More cleaned up submit package without cmake work.


Alexpux:
clang,harfbuzz - increase pkgrel

PR #1316, opened 2016-04-13
Adjustment in graphite2 recipe

renatosilva:
This should fix the error *could not remove database entry* when updating the system.

### DIFF
--- a/mingw-w64-graphite2/PKGBUILD
+++ b/mingw-w64-graphite2/PKGBUILD
@@ -3,10 +3,8 @@
 _realname=graphite2
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.3.8
-pkgrel=1
+pkgrel=2
 pkgdesc="Font rendering capabilities for complex non-Roman writing systems (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


Alexpux:
Yeah. This is the issue exactly. I need read all 3rdparrty PKGBUILD more carefully

JPeterMugaas:
I wonder if some of the templates provided are adequate.  I was using the one for CMake on a git repository version but I was not doing that at all.  That's how the provides and replaces got in.

I admit that I made a mistake but at the moment I care more about how it was made.

mingwandroid:
Well I guess I wanted to show that those fields existed in the template, but it is wrong of me to to do so without making it clear with a comment that you should remove them in most cases!

Sorry about that guys.

JPeterMugaas:
Actually, it would make sense to use those fields when doing a -git version of something.  The provides would allow something dependent upon the release package to still work.  Replaces would be used to indicate a replacement for the original release package and indicate that the -git version is mutually exclusive with a release version.  So it's good practice in that instance.

The problem in the graphite2 case was this provides and replaces stuff was not applicable to what I was doing which was something taken from a release tarball where I was not replacing anything and there was no mutual exclusivity with something else.  And yes, for a non-repository version, there would be some other things that don't apply such as "pkgver" and cd "${_realname}".

The additional template I provided in Merge request #1317 is not meant to replace the GitHub one which you probably need   I wonder if you need something for autotools with a GitHub repository (-git) version.

mingwandroid:
Indeed, I was trying to demonstrate a variety of PKGBUILD features in the templates. I did consider a comment explaining this but left one out as I feared people would leave them in there.

PR #1326, opened 2016-04-16
Removed 0009-Work-around-un-too-late-resolved-__EH_FRAME_BEGIN__.patch

lhmouse:
The patch is no longer needed because the macro in question has been conditionally #undef'd for MinGW targets.

The issue has been fixed on GCC branch 'gcc-5-branch' by the following commit:
  https://github.com/gcc-mirror/gcc/commit/1f4cfb7ffe95dcf2e3568d98889ee6db914c0b40
... and on all other branches by the following commit:
  https://github.com/gcc-mirror/gcc/commit/1c934c582841982ea80a5aa3cb7afc82b9ef5e3d


mati865:
I think you should bump `pkgrel`.

mingwandroid:
`pkgrel` does not need to be bumped for -git pkgs, also this is a WIP commit because the package doesn't build yet.

mati865:
You are right, I've overlooked it.

mingwandroid:
Successful builds change the `pkgver`, there might be a way of getting it updated while the a build is not possible but I don't have the time to figure that out right now. If anyone knows how please share!

PR #1328, opened 2016-04-17
hunspell -upgrade to 1.3.4

JPeterMugaas:
Hunspell - upgrade to 1.3.4 and remove files created by patches.


renatosilva:
Were all these patches upstreamed? Please do not remove them without a reason. The removal of relocation for example may break projects that need this feature. Please try making them work with the new version instead.

mati865:
It is a bit misleading. In first commit https://github.com/Alexpux/MINGW-packages/pull/1328/commits/8b09978c63187a27cc715c2a8bd3ec8c6b04a32b he removed old patches for 1.3.3 and in second commit https://github.com/Alexpux/MINGW-packages/pull/1328/commits/5ee008e51ef1f1af68f142c2f440a67377a3a6f7 he recreated them for 1.3.4.

@JPeterMugaas could you rebase commits to merge it into 1 commit with a little better wording? I mean something like "hunspell: upgrade to 1.3.4" but as one commit.
vim is default editor but you can execute `export EDITOR=nano` if you don't have experience with it, then do following for your repo:
```
git rebase -i HEAD~2
git push origin master --force
```

renatosilva:
I see you did not remove patches, commit is clearer now :+1:

JPeterMugaas:
Okay.  Sorry for my bad wording.  What had happened was that I had to recreate many of the patches for 1.3.4 because the 1.3.3 patches did not work for the new version.  The new versions patch versions were named 1.3.4.  After the initial upgrade commit, I realized that I had forgotten to delete the old 1.3.3 files.

In addition, the new PKGBUILD had to remove some files that were created by the relocate patch.  Those files tend to create problems here.

mingwandroid:
Thanks for the continued contributions @JPeterMugaas.

PR #1333, opened 2016-04-18
ccache: add shortcuts for gcc

mati865:
Add shortcuts for gcc.
makepkg-mingw doesn't use it after changing `ccache` flag in `makepkg_mingw32.conf` and `makepkg_mingw64.conf`. I suppose it is makepkg-mingw bug and I'll further investigate it.
Manual compilation after exporting `export PATH="/mingw32/lib/ccache/bin/:$PATH"` for 32-bits or `export PATH="/mingw64/lib/ccache/bin/:$PATH"` for 64-bits works fine.


Alexpux:
Increase "pkgrel"

mati865:
Sorry, missed it as I focused on makepkg-mingw.

PR #1334, opened 2016-04-18
iso-codes

JPeterMugaas:
Update to 3.67.


mati865:
Good job but in next commits include name of modified package please. Ex. `iso-codes: update to 3.67`.

PR #1344, opened 2016-04-19
Move CUnit to checkdepends and add new template

JPeterMugaas:


mingw-w64-PKGBUILD-templates/PKGBUILD.setup-py-python-pkg
```diff
@@ -0,0 +1,85 @@
...
+  ${MINGW_PREFIX}/bin/python2 setup.py nosetests
+}
+
+# Note that build() is sometimes skipped because it's done in 
+# the packages setup.py install for simplicity if you can do so.
+package_python3-somepackage() {
+  depends=("${MINGW_PACKAGE_PREFIX}-python3")
+
+  cd "${srcdir}/python3-build"
+  MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
```
Alexpux: Well as I remember here can be more arguments. Check python modules packages

JPeterMugaas:
Which package in particular should I look at?

PR #1346, opened 2016-04-19
jsoncpp: Fix ${MINGW_PREFIX} getting hard-coded into the CMake package files

rpavlik:
Upstream CMake deficiencies are to blame, I think, in this case, since I'm pretty sure I worked around this in my branch of the repo, but this is just a simple sed post-install that can fix up consuming jsoncpp by CMake-using packages.

First time contributing to msys2 - wasn't sure if I needed to bump any of the versions, etc. or do any other special things with the commit, so be kind and let me know if I messed this one up, thanks!


Alexpux:
You need increase ````pkgrel````

rpavlik:
Thanks - I pushed an amended commit.

PR #1350, opened 2016-04-22
cmake: update to 3.5.2

mati865:
I had to make some changes to qt5 static patch however my machine is too slow to test it.
Do not merge until some report qt5-static build succeed.

### DIFF
--- a/mingw-w64-cmake/PKGBUILD
+++ b/mingw-w64-cmake/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=cmake
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.4.1
+pkgver=3.5.2
 pkgrel=1
 pkgdesc="A cross-platform open-source make system (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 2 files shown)


mati865:
After 7+ hours x86_64 build of Qt5-static finished. Compiling packages against it gives no errors.
I suppose it is safe to merge it if @mingwandroid doesn't have objections.

mingwandroid:
So finally you need to test a qt project that uses CMake as the build system and uses qt plugins. But to be honest you've done enough that I'm happy to say:

Looks good to me and also thanks a lot for the work.

mati865:
https://forum.qt.io/topic/30369/testing-qt-plugins/5 succeed (declarative, webkit, webkitwidgets, phonon were excluded).
If you know where I can find more code to test let me know and I'll give it a try.

mingwandroid:
It seems that there's a bug with qt5-static where it's being built with icu enabled. This needs to be fixed unfortunately.

mati865:
I'm seeing this error when building i686 qt5-static with Tea CI. However local build of x86_64 was fine. I suppose it's bug in icu but to be sure I've started qt5-static build with older cmake.

offtopic:
```
CMake Internal Error (please report a bug):
  Missing cmGeneratorTarget instance!
```
from https://tea-ci.org/Alexpux/MINGW-packages/667 should be fixed with new cmake but more tests are needed.

mingwandroid:
From my perspective, `_use_icu` should not be set for qt5-static as it is too big. I need to ask @Alexpux about this change. Besides, it causes build failures for some projects that use qt5-static since ICU symbols are not found in Qt5Core.

I am trying to build qt5-static without ICU but it will take some hours for the build to complete and for me to test it.

fracting:
@mati865 I just realized that your CI build is killed because of time out, I set the time limitation of your account to 900 mins, hopefully that helps :)

mati865:
@fracting Thank you, here is link for this build https://tea-ci.org/mati865/MINGW-packages/38

PR #1353, opened 2016-04-22
Adjustment in winpthreads-git

renatosilva:
The patches are now applied with patch instead of git, so:

* Mismatching package versions between architectures are avoided.
* The same source package is generated for both architectures as intended by Alexey.
* The real commit hash is present in package version.

### DIFF
--- a/mingw-w64-winpthreads-git/PKGBUILD
+++ b/mingw-w64-winpthreads-git/PKGBUILD
@@ -5,7 +5,7 @@
 _realname=winpthreads
 pkgbase=mingw-w64-${_realname}-git
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-git" "${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
-pkgver=5.0.0.4634.1a9aeb9
+pkgver=5.0.0.4651.7666854
 pkgrel=1
 pkgdesc="MinGW-w64 winpthreads library"
 url="http://mingw-w64.sourceforge.net"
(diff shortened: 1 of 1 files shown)


mingwandroid:
Would `git reset HEAD ^` achieve the same result? Using git apply has distinct advantages in maintainance of patches with attribution and comments.

renatosilva:
That reset command would need to be repeated for every patch. You can continue creating patches with git format-patch, but use patch instead of git am for applying them.

mingwandroid:
Yeah i get that detail. The issue here is that refreshing them with git format-patch is a *very* handy trick to prevent them from getting stale, losing attribution etc.

So in balance I'd much rather go for git reset.

DavidEGrayson:
I wish there was a patch-refreshing technique that didn't require using git.  It seems like there should be one, but I have not found it.

mingw-w64-winpthreads-git/PKGBUILD
```diff
@@ -36,7 +36,7 @@ pkgver() {
 
 prepare() {
   cd "${srcdir}/mingw-w64"
-  git am "${srcdir}"/0001-Define-__-de-register_frame_info-in-fake-libgcc_s.patch
+  patch -p1 < "${srcdir}"/0001-Define-__-de-register_frame_info-in-fake-libgcc_s.patch
```
Alexpux: Use ````-i```` switch please

renatosilva:
See 1ff8c30.

PR #1354, opened 2016-04-23
Update to lfcbase-1.8.2 and cego-2.30.3

bjlemke:


bjlemke:
Since there is a package dependency, lfcbase-1.8.2 has to be build first, before cego-2.30.3 can be build

PR #1355, opened 2016-04-23
Update lfcbase-1.8.2

bjlemke:


bjlemke:
There still seems to fail the dependency to lfcxml. Don't know, how to handle this :(

mati865:
@bjlemke It is Tea-CI issue and you shouldn't worry about it.
You can try to do it like that: 1st commit `lfcbase: update to 1.8.2`, 2nd commit `cego: update to 2.30.3`.

PR #1358, opened 2016-04-24
Catch: Update to 1.5.0.

DavidEGrayson:
I tested this and it builds fine on my machine.

By the way, I don't have an urgent desire to use this latest version of Catch.  The old one worked fine for my purposes.  But I just feel like I am doing my part to keep MSYS2 up-to-date which is generally good for all users.


DavidEGrayson:
Thanks!

PR #1362, opened 2016-04-25
qt5: deactivate patch 0037

filnet:
patch 0037 breaks cmake integration
patch file: 0037-qt-5.4.0-Improve-cmake-plugin-detection-as-not-all-are-suffixed-Plugin.patch

with the patch, a `find_package(Qt5Quick REQUIRED)` will fail with :

```
CMake Error at C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5Qml_QLocalClientConnectionFactory.cmake:2 (add_library):
  add_library cannot create imported target
  "Qt5::QLocalClientConnectionFactory" because another target with the same
  name already exists.
  C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5QmlConfigExtras.cmake:4 (include)
  C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5QmlConfig.cmake:176 (include)
  C:/msys64/mingw64/lib/cmake/Qt5Quick/Qt5QuickConfig.cmake:96 (find_package)
  CMakeLists.txt:84 (find_package)

CMake Error at C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5Qml_QQmlDebuggerServiceFactory.cmake:2 (add_library):
  add_library cannot create imported target "Qt5::QQmlDebuggerServiceFactory"
  because another target with the same name already exists.
  C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5QmlConfigExtras.cmake:4 (include)
  C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5QmlConfig.cmake:176 (include)
  C:/msys64/mingw64/lib/cmake/Qt5Quick/Qt5QuickConfig.cmake:96 (find_package)
  CMakeLists.txt:84 (find_package)

  ...
```

I have traced the source of the error to the patch.
Undoing its effect in my local environment addresses the issue.

I have not rebuilt qt5 but it should build fine.

### DIFF
--- a/mingw-w64-qt5/PKGBUILD
+++ b/mingw-w64-qt5/PKGBUILD
@@ -100,7 +100,7 @@ else
 fi
 
 pkgver=${_ver_base//-/}
-pkgrel=2
+pkgrel=3
 arch=('any')
 pkgdesc="A cross-platform application and UI framework (mingw-w64${_namesuff})"
 url='https://www.qt.io/'
@@ -340,7 +340,9 @@ prepare() {
 
   # Patches so that qt5-static can be used with cmake.
   patch -p1 -i ${srcdir}/0036-qt-5.3.2-win32-qt5-static-cmake-link-ws2_32-and--static.patch
-  patch -p1 -i ${srcdir}/0037-qt-5.4.0-Improve-cmake-plugin-detection-as-not-all-are-suffixed-Plugin.patch
+  # patch 0037 breaks 5.6.0 cmake integration
+  # the patch causes Qml extra factories to be added twice (correctly as factories and wrongly as plugins)
+  ###patch -p1 -i ${srcdir}/0037-qt-5.4.0-Improve-cmake-plugin-detection-as-not-all-are-suffixed-Plugin.patch
 
   pushd qtbase > /dev/null
     patch -p1 -i ${srcdir}/0038-qt-5.5.0-cmake-Rearrange-STATIC-vs-INTERFACE-targets.patch


mati865:
Other patches must be updated to reflect this change (grep patches modyfing `qtbase/mkspecs/features/data/cmake/Qt5BasicConfig.cmake.in` and update them).

mingwandroid:
Regardless, this disabling this patch is not the correct approach. It breaks qt5 static. I guess a change between qt5.5 and qt5.6 caused this problem too come to the surface?

filnet:
What would be a correct approach then ?

qt5 is broken w/r to using QtQuick in a cmake project.

mati865:
Maybe add case for `-static`.

filnet:
What's the problem with qt5 static ?

Shouldn't the issue be fixed by and in qt5 static ?

filnet:
Is it possible to reopen this review so I can see the CI errors ?

mati865:
If you look into PKGBUILD:
```
# Maintainer: Alexey Pavlov <alexpux@gmail.com>
# Maintainer: Ray Donnelly <mingw.android@gmail.com>

# Please keep this file exactly the same (except _variant=)
# as {mingw-w64-qt5/PKGBUILD,mingw-w64-qt5-static/PKGBUILD}
#_variant=-static
_variant=-shared
# If you want a quick turnaround during testing then
# use this. *Not* the same as _fastbuild; that's set
# always for static and never for non-static.
#_minimal="yes"
```

filnet:
Ok will take a look. Will also look at making a -static case (there are examples of doing that in other patches.

I also understand why the build fails. A later patch (0042) moves stuff around including changes done by 0037.

mati865:
@filnet exactly.

filnet:
Another approach would be to fix the patch so it ignores *Factory.cmake files

mingwandroid:
If the patches diverged we couldn't possible hope for them to be upstreamed, since one set breaks the other use case and vice versa.

PR #1365, opened 2016-04-27
Alexpux/master

JPeterMugaas:
Various updates, adjustments, and new template.

### DIFF
--- a/mingw-w64-python-py/PKGBUILD
+++ b/mingw-w64-python-py/PKGBUILD
@@ -1,27 +1,49 @@
 # Contributor: Frederic Wang <fred.wang@free.fr>
 
+_pyname=py
 _realname=py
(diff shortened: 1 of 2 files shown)


mingw-w64-python-py/PKGBUILD
```diff
@@ -1,27 +1,49 @@
 # Contributor: Frederic Wang <fred.wang@free.fr>
 
+_pyname=py
```
Alexpux: Why it here?

JPeterMugaas: It is part of the template I was using for this.  Some python packages have a capitalitalized name such as "Pygments" and the dir names use that capitalized name while most package names here have lower-case names such as "pygments".

PR #1376, opened 2016-05-01
Add unibilium

equalsraf:
Recipe for the unibilium library. The patch disables asan CFLAGS and removes the default terminfo paths.

### DIFF
--- a/mingw-w64-unibilium/PKGBUILD
+++ b/mingw-w64-unibilium/PKGBUILD
@@ -1,0 +1,32 @@
+_realname=unibilium
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.2.0
+pkgrel=1
+pkgdesc='A terminfo parsing library (mingw-w64)'
+arch=('any')
+url='https://github.com/mauke/unibilium/'
+license=(LGPL3)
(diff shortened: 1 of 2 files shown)


Alexpux:
Why we need this package?

equalsraf:
Unibilium is a portable terminfo library so it should be useful for native windows apps that need to deal with a terminal. Personally I'd like to be able to build Neovim using MSYS2 and this is one of its dependencies.

mingw-w64-unibilium/PKGBUILD
```diff
@@ -0,0 +1,32 @@
+_realname=unibilium
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.2.0
+pkgrel=1
+pkgdesc='A terminfo parsing library'
```
Alexpux: Add mingw-w64 at the end of description.

mingw-w64-unibilium/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+arch=(x86_64 i686)
```
Alexpux: For mingw packages arch ***MUST*** be always ````any````

equalsraf:
Ok, updated pkgdesc and arch

PR #1397, opened 2016-05-08
go: add env `$GOROOT` at loading /etc/profile

ys-nuem:


Alexpux:
Scripts from /mingw*/etc/profile not loading automatically by msys2.

ys-nuem:
The Go binary can be used from shell only if`$MSYSTEM` is set to `MINGW32` or `MINGW64`.
In such the situation, scripts in `/mingw**/etc/profile.d` are loaded automatically.

I think it is more desirable than installing into `/etc/profile.d` because `mingw-w64-go` package is not in MSYS2.

PR #1405, opened 2016-05-13
devcon-git: Disable wildcard globbing.

DavidEGrayson:
Our mingw-w64 crt globs wildcards it sees on `argv` by default because it was configured with `--enable-wildcard`.  That prevents us from running commands like `devcon hwids *`, which is the correct way to list the hardware IDs of all the devices on the system.  This pull request disables wildcard globbing for devcon.

Note that to run that command in Bash, you would actually have to type:

    devcon hwids \*

### DIFF
--- a/mingw-w64-devcon-git/PKGBUILD
+++ b/mingw-w64-devcon-git/PKGBUILD
@@ -5,9 +5,9 @@
 _realname=devcon
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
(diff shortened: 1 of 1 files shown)


mingwandroid:
I worry that some things won't build if they depend on the old behaviour. Should we think about fixing escaped astericks handling instead?

I'm completely on the fence here!

DavidEGrayson:
This patch should only affect deacon.  I was not proposing to change mingw-w64-crt-git.

elieux:
Sounds okay to me.

mingwandroid:
Ah oops, note to self, don't check commits on a mobile phone.

DavidEGrayson:
Thanks!

PR #1406, opened 2016-05-13
mingw-w64-imagemack: 7.0.1.3 update to latest version

JPeterMugaas:
mingw-w64-imagemack:  Update to altest version


mingw-w64-imagemagick/PKGBUILD
```diff
@@ -102,6 +108,7 @@ build() {
     --without-ltdl \
     --without-perl \
     --without-x \
+    --without-magick-plus-plus \
```
Alexpux: why it here?

mingw-w64-imagemagick/PKGBUILD
```diff
@@ -48,39 +48,44 @@ optdepends=("${MINGW_PACKAGE_PREFIX}-ghostscript: for Ghostscript support"
             )
 options=('staticlibs' 'strip' '!debug' 'libtool')
 source=(https://www.imagemagick.org/download/ImageMagick-${pkgver%.*}-${pkgver##*.}.tar.xz{,.asc}
-        ImageMagick-6.8.8.1-mingw.patch
-        001-relocate.patch
-        002-build-fixes.patch
+        ImageMagick-7.0.1.3-mingw.patch
+        001-7.0.1.3-relocate.patch
```
Alexpux: New patches added but seems old not removed from repo

svnpenn:
@JPeterMugaas @Alexpux where is the package for this? I check here:

http://repo.msys2.org/mingw/x86_64

but I only see the old versions

Alexpux:
@svnpenn no package, currently c++ part not building so I skip it. 
Also I'm very busy and will build/upload packages once in a week or two.

JPeterMugaas:
I found a newer 7.0.1.6 release.  I don't think they keep old releases from the 7.0.1 series (they probably should).  But anyway, it seems that bug still is not fixed.

PR #1416, opened 2016-05-20
libmypaint-git: depend on GI at build time

achadwick:
Per https://github.com/Alexpux/MINGW-packages/commit/414003c6b5077cda0b168eff3f6133076b7821be#commitcomment-17550373


achadwick:
Not needed, I think: G-I is not (yet) part of the standard libmypaint build.

PR #1420, opened 2016-05-21
qt5-git: removed obsolete patches

mcallegari:
and fixed installation path


Alexpux:
Are you sure that this patches not needed?

mcallegari:
I built Qt before opening this.
I think those patches were related to some issue found when compiling QtWebkit, which is no more there.

Also, the comment in the PKGBUILD file makes me think it was definitely something related to QtWebkit:
# Ossi says we should fix this in qwebkit with a $$(shell_path), but the ' quotes get in the way		
# due to qtPrepareTool adding them with $$shell_quote

Alexpux:
0030 at least need I think. It not related to build Qt itself but building some programs that include this header PTR is in conflict with mingw-w64 define

mcallegari:
Updated. Fixed also a couple of patches that got broken very recently.

PR #1422, opened 2016-05-21
libssh: repackage missing pkg-config files

wiiaboo:
Previous update stopped including pkg-config files which are needed for ex. FFmpeg.


wiiaboo:
I forgot to change the pkgrel. That must be changed or the package won't be rebuilt, right?

mati865:
Yes.

PR #1428, opened 2016-05-23
libzip: updated to version 1.1.2 and fixed bug

### LINKED ISSUE #1418: libzip failure due to rename failure
I debugged a problem encoutered while using libzip to create ZIP files programmatically. My program create a ZIP file, then adds new files one-by-one to this existing archive. It is not working on MSYS2 whereas it is working nicely on Linux and Mac OS X.

It turned out that at some point libzip tries to rename a temporary file using the system `rename` function (http://man7.org/linux/man-pages/man2/rename.2.html) and this function behaves wrongly in MSYS2. If the `newpath` already exists, the function should siliently atomically overwrite it (according to the man page). However, on MSYS2, the function simply fails in that case, thus breaking libzip in `zip_source_filep.c` line 245 (libzip version 1.0.1).

Since `rename()` is a widely used system function, I would not be surprized that this could be the source of many bugs in MSYS2.

omichel:
Fixes #1418.

This fixes a bug found in the current libzip package preventing the user from adding a new file to an existing archive.

### DIFF
--- a/mingw-w64-libzip/PKGBUILD
+++ b/mingw-w64-libzip/PKGBUILD
@@ -1,23 +1,29 @@
 # Maintainer: Jianfeng Liu <liujianfeng1994@gmail.com>
+# Contributor: Olivier Michel <Olivier.Michel@cyberbotics.com>
 
 _realname=libzip
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.1
-pkgrel=2
+pkgver=1.1.2
+pkgrel=1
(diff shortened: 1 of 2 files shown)


omichel:
:+1:

mingwandroid:
Thanks.

omichel:
This PR was merged two days ago and the corresponding package is still not available on the pacman repositories. Do you know about how long it will take?

mati865:
@Alexpux is currently busy. You can build package for yourself.

omichel:
All right, no problem, I was just wondering. It should be a matter of a couple of days I guess.

PR #1438, opened 2016-05-27
Updates for filesystem changes

renatosilva:
Updates for Alexpux/MSYS2-packages#612.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ MINGW-packages
 
 Package scripts for MinGW-w64 targets to build under MSYS2.
 
-To build these, run msys2_shell.bat then from the bash prompt.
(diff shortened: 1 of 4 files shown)


Alexpux:
Well I think octopi need more work. As I see you not try build recent version but just fix patch

mati865:
@Alexpux 
Yes, I've been trying to make Octopi 0.8.1 working in MSYS2.
My work is here: https://github.com/mati865/MINGW-packages/commits/octopi
Issue: https://github.com/mati865/MINGW-packages/issues/4

Current state: Package groups column is bugged, there is bug (could be fixed by now, need to test) with 32-bit Qt5-static crashing in Release mode.

PR #1437, opened 2016-05-27
Averted the relocation bug.

lhmouse:
According to <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70936>, GCC's '--with-gxx-include-dir' configure option is causing a relocation bug, We'd better remove it.
The newly added standard-conforming <cstdlib> and <cmath> C++ headers since GCC 6 suffer from this bug. A relocated GCC will complain about inability to find <stdlib.h> and <math.h>, respectively.
In addition, removal of this option changes the paths where GCC searches for libstdc++ headers. We make GCC search for them in the default directories by removing '--enable-version-specific-runtime-libs' as well.

---
Here is an abridged IRC log in #gcc on OFTC today:

> [16:12:21] \<lh_mouse> The #include_next bug. it hasn't been fixed.
> [16:12:25] \<jwakely> lh_mouse: your problems were on mingw though, right?
> [16:12:54] \<drako> anyways, my target version is 4.9.3, so it's good :-)
> [16:12:56] \<lh_mouse> jwakely, another person reported the same bug on Linux.
> [16:13:01] \<lh_mouse> jwakely, https://gcc.gnu.org/ml/gcc/2016-05/msg00092.html
> [16:13:02] \<jwakely> lh_mouse: link?
> [16:13:10] \<jwakely> that's not a bug report, is it in bugzilla?
> ...
> [16:16:06] \<lh_mouse> jwakely, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70936
> ...
> [16:47:30] \<lh_mouse> jwakely, 70936 updated with 'gcc -v' output and a minimal testcase.
> [16:47:47] \<jwakely> thanks
> [16:48:32] \<lh_mouse> Well the testcase consists of only a line of command:  echo '#include \<cstdlib>' | g++ -v -x c++ -
> [16:49:02] \<jwakely> it's still useful
> ...
> [17:00:11] \<jwakely> I suspect --with-gxx-include-dir is the problem
> [17:01:13] \<lh_mouse> So what should I do? remove it?
> [17:01:42] \<jwakely> be patient
> [17:01:59] \<lh_mouse> >_<
> [17:01:59] \<jwakely> but yes, you could try removing it from the build and see if it makes any difference
> [17:04:05] * davidlt (~davidlt@pb-d-128-141-241-251.cern.ch) has joined
> [17:04:45] \<lh_mouse> Removed. I am going to rebuild GCC which would take more than an hour with bootstrap disabled.
> ...
[...]

Issue #1374, opened 2016-04-30
GCC 6.1.0


BrisingrAerowing:
GCC 6.1.0 has been released, and I was wondering if work is being done to update to it.

Just curious. Some of the new features look nice.

mati865:
@Alexpux is off for 10 days.
There is some work done for gcc 6 in `gcc-git` package https://github.com/Alexpux/MINGW-packages/commits/master/mingw-w64-gcc-git

mingwandroid:
And also some problems reported regarding relocation by @lhmouse and some bugs reported by @tkelman, it might be a little while yet, but your experiences with a self-built mingw-w64-gcc-git package would be appreciated.

tkelman:
The new error and warning printing is nicer, but it did get more picky about some constructs so some code accepted by gcc 5 won't be by gcc 6.

lhmouse:
Known issue: ncurses 5.9 doesn't compile with GCC 6.

Alexpux:
Right now curses is 6.0. Do you try it?

lhmouse:
> Right now curses is 6.0. Do you try it?
Yes I have tried yesterday. 6.0 is fine

mati865:
@tkelman It's because of `The default mode for C++ is now -std=gnu++14 instead of -std=gnu++98`. There was analogical situation when GCC switched from ` -std=gnu89` to  `-std=gnu11`. Code which cannot be easily ported to new standard should be compiled with old flags.

tkelman:
@mati865 not entirely. I'm on a phone atm so don't have a link handy, but we found examples of code that gcc 5 compiled fine but gcc 6 could not be made to build with any set of flags, std or otherwise.

lhmouse:
@mati865 I used to always build ncurses-5.9 with -std=c99. Yes it fails on GCC 6.

BrisingrAerowing:
I'll get it building and see what happens. Might be a while, though.

BrisingrAerowing:
I get various errors trying to build, mostly file not found issues, with a few compiler errors thrown in for good measure. I'm not sure if it is my environment or the package itself, but I am giving up on trying to build this package. I am going to just build GCC 5.3.0 instead.

Leaving this open for now.

lhmouse:
@BrisingrAerowing
It sounds strange. Which files were not found?

mati865:
I cannot build GCC 6 git version. For 3 days it fails with:
```
[...]

PR #1444, opened 2016-05-28
Fixed packaging stuff.

lhmouse:
This a general patch for gcc-5-branch and gcc-6-branch that fixes issues mentioned in #1374 and #1437.

* In #1374:

The relocation bug mentioned in #1374 is a [GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70936) and has been averted.
libvtv*.dll is now excluded from packaging, though the reason why it isn't built is still unknown.

* In #1437:

Modification of the library path introduces problems when packaging DLLs. Libraries are known to be subject to this bug are `libatomic*`, `libgomp*`, `libquadmath*`, `libssp*`, `libobjc*` and `libgfortan*`.

The `build()` function now removes the destination directory before it invokes `make install` so missing libraries are now reported as errors. Otherwise, if a successful package has been created already, `makepkg` might be happy with existent files.

Two bootstraps with all languages enabled from scratch that had all indeterminate and installed files removed and targeted x86_64 and i686 were made to verify the script. 

@BrisingrAerowing @mati865 @tkelman 
Please test whether this script works.


mati865:
Compiled ffmpeg and some atomics examples using GCC 6 from this PR. Fine for me.

PR #1446, opened 2016-05-29
Update libmypaint-git

achadwick:
An assortment of little fixes that make sense in light of our recent release.

### DIFF
--- a/mingw-w64-libmypaint-git/PKGBUILD
+++ b/mingw-w64-libmypaint-git/PKGBUILD
@@ -40,15 +40,7 @@ sha256sums=('SKIP')
 
 prepare() {
   cd "${srcdir}"/${_realname}
-  #./autogen.sh --help
-  for f in README ChangeLog NEWS AUTHORS; do
-    if ! test -f "$f"; then
-      echo "Creating $f"
-      echo "missing" > "$f"
-    fi
-  done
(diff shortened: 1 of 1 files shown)


fracting:
Hi folks,

I sent a pull request to the codelite project, enable using Tea CI + PKGBUILD to build codelite per commit, if the build fails, we are able to get noticed and fix Msys2 PKGBUILDs:

https://github.com/eranif/codelite/pull/1299
```
    - svn co https://github.com/Alexpux/MINGW-packages/trunk/mingw-w64-codelite-git
    - cd mingw-w64-codelite-git
    - makepkg-mingw -s --noconfirm --skippgpcheck --noprogressbar
```

How do you like this idea? Anyone would like to work on a similar pull request to mypaint project?

BTW, is there anyway to limit the clone depth during `makepkg` ?

Thanks!

achadwick:
@fracting 
We already build libmypaint per commit on appveyor and travis-ci, and I want to do the same thing for mypaint master once MSYS2 has a binary package of the last libmypaint to build against. We could use the MSYS2 PKGBUILDs directly too, I guess. :rainbow: _needs more build matrix!_

(One of the things I want to do is get rid of SCons from MyPaint too. `setup.py` is a much better fit for MyPaint after the libmypaint split.)

PR #1449, opened 2016-05-30
Libzip - 1.1.3

JPeterMugaas:
Update to latest version.


omichel:
That seems to be a duplicate of #1448.

JPeterMugaas:
It is.  Let's just say great minds think alike :-)

PR #1458, opened 2016-06-05
Make x265 support winxp

timothytylee:
Instead of using _WIN32_WINNT to determine whether Windows XP support code in "winxp.cpp" and "winxp.h" is needed, enable it for 32-bit builds.

Also, pass -DWINXP_SUPPORT:BOOL=TRUE to cmake under 32-bit builds to enable Windows XP support.

### DIFF
--- a/mingw-w64-x265/PKGBUILD
+++ b/mingw-w64-x265/PKGBUILD
@@ -45,8 +45,8 @@ build() {
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
   ${MINGW_PREFIX}/bin/cmake \
     -G "MSYS Makefiles" \
-    -DWINXP_SUPPORT:BOOL=${_WINXP_SUPPORT} \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DWINXP_SUPPORT=${_WINXP_SUPPORT} \
     ../../../source
   make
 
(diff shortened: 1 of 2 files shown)


mingw-w64-x265/PKGBUILD
```diff
@@ -28,6 +45,7 @@ build() {
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
   ${MINGW_PREFIX}/bin/cmake \
     -G "MSYS Makefiles" \
+    -DWINXP_SUPPORT:BOOL=${_WINXP_SUPPORT} \
```
Alexpux: Remove ````:BOOL```` part, it not needed

mingw-w64-x265/PKGBUILD
```diff
@@ -39,16 +57,11 @@ build() {
...
-    _ENABLE_ASM=OFF
-  else
-    _ENABLE_ASM=ON
-  fi
-
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
   ${MINGW_PREFIX}/bin/cmake \
     -G "MSYS Makefiles" \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DWINXP_SUPPORT:BOOL=${_WINXP_SUPPORT} \
```
Alexpux: The same as previous

timothytylee:
Made the requested changes and tested build on Windows XP

PR #1469, opened 2016-06-10
Mingw w64 glfw: 3.2 Update to latest version

JPeterMugaas:


mingw-w64-glfw/PKGBUILD
```diff
@@ -15,12 +15,11 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
 options=('staticlibs' 'strip')
 source=(https://downloads.sourceforge.net/project/glfw/glfw/${pkgver}/${_realname}-${pkgver}.tar.gz
         001-cmake-config-location.patch)
```
mati865: You didn't remove patch from source and sha256sums.

mingw-w64-glfw/PKGBUILD
```diff
@@ -61,7 +60,8 @@ package() {
 
   cd "${srcdir}/shared-${MINGW_CHOST}"
   make DESTDIR=${pkgdir} install
-  mv ${pkgdir}${MINGW_PREFIX}/lib/glfw3dll.a ${pkgdir}${MINGW_PREFIX}/lib/libglfw3.dll.a
+
+#  mv ${pkgdir}${MINGW_PREFIX}/lib/glfw3dll.a ${pkgdir}${MINGW_PREFIX}/lib/libglfw3.dll.a
```
DavidEGrayson: When I tried to build glfw 3.2, I found that it was installing files with "blahdll.a", when it should be like "blah.dll.a".  You should run `find pkg -name *.a` after building the package to see if there are files like that and fix it.

Issue #1168, opened 2016-03-10
Add Vulkan header files and library to packages


FslNopper:
Please add the Vulkan header files and library as a package:
https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers

Alexpux:
Please provide PKGBUILD for it :)

FslNopper:
Never done a PKGBUILD script - so this may take a while.

DoctorWho11:
It's simple and I already have some tarballs provided from the official vulkan sdk (import libs generated from the defs of the vulkan dll though)

FslNopper:
It's easy when you have done it before :-) The honor will go to you, as soon it goes live. Thank you very much. Please close, as soon as the libs are available.

PR #1471, opened 2016-06-10
Vulkan

JPeterMugaas:


mingw-w64-vulkan
```diff
@@ -0,0 +1,56 @@
+# Maintainer: Some One <some.one@some.email.com>
```
mati865: I think you missed it.

mingw-w64-vulkan
```diff
@@ -0,0 +1,56 @@
...
+_realname=vulkan
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-headers" "${MINGW_PACKAGE_PREFIX}-${_realname}-man-pages")
+_pkgname=Vulkan-Docs
+_pkgver=1.0-core+wsi-20160527
+epoch=1
+pkgrel=1
+arch=('any')
+url="https://www.khronos.org/vulkan/"
+makedepends=(asciidoc)
```
mati865: Depends on `${MINGW_PACKAGE_PREFIX}-libxslt` also.

maoueh:
This will probably close #1168.

JPeterMugaas:
Perhaps but I still think you might need the library.  There is something at:
https://vulkan.lunarg.com/signin

But I have not yet done a package for it.  I think you just need the .DLL and that is part of the driver or the development SDK.  I do know that when I built glfew, cmake did detect the vulkan .DLL and compiled a test for it.

mati865:
@maoueh 
They are just headers, we would need Vulkan validation layers also.
But their code depend on MSVC for Windows so it would require a lot patches to compile it inside MSYS.

PR #1480, opened 2016-06-11
Openal

JPeterMugaas:


Alexpux:
fluidsync patch not needed anymore?

JPeterMugaas:
I don't think so.  At least, not for that.  I tried searching the current code and it did not mention anything about FluidSynth and the alc/midi code disappeared after the 1.16.0 release.  But that package might be needed by some other things.  You can't always know.  So I would recommend keeping that package.  I did drop the fluidynth because of what I described.

PR #1486, opened 2016-06-13
pexports: new package

ys-nuem:
... to help creating import libraries.


Alexpux:
This is not needed as for mingw-w64 there are ````mingw-w64-{i686,x86}-tools-git```` package that contain ````gendef```` program - mingw-w64 replacement for pexports

ys-nuem:
I've understand. removing the PR.

PR #1488, opened 2016-06-15
gedit: update to version 3.20.2

nacho:


fracting:
Hi @nacho, while you are here, could you also fix the build dependency?

See https://github.com/fracting/MINGW-packages/commit/77bf8f10e842aaec1e2af8d3f087b5c14badbe46

and https://tea-ci.org/fracting/MINGW-packages/87

(Sorry the CI server didn't work for some hours)

Thanks!

nacho:
here we go. Thanks!

fracting:
Tea CI hangs at `VAPIGEN gedit/gedit.vapi`, should be a Wine bug I guess because Appveyor CI works fine.

Anyone has idea what could be wrong?

PR #1499, opened 2016-06-22
bsdfprocessor: fix runtime deps

mkae:
When I committed this PKGBUILD previously I hadn't noticed that the runtime deps were empty. This should fix it.


mkae:
@fracting, do you have any clue what's going on with tea here?

mati865:
@mkae Looks like Tea crashed, you don't need to worry about it.

mkae:
mati865, thanks for the heads-up.

PR #1500, opened 2016-06-22
Qt4 static

vividsydney:
Fixed compilation issues in static variant and reworked PKGBUILD file similar to the one related to qt5.
Open question is what to do with _.prl and pkgconfig/_.pc files related to statically built libraries.


Alexpux:
If you look into qt5-static package you can see that we instal it in separate prefix not in the same as regular packages to not conflict with qt5-shared

mingwandroid:
Do you not want to proceed with this? A separate prefix is a requirement.

vividsydney:
Ray, there is some interest in proceeding but no time. I will try getting back to this work soon.

PR #1501, opened 2016-06-24
Updates for some packages

renatosilva:
* Fix regression with hunspell and enchant.
* Update pidgin and libmongoose.
* Install license for graphite2.


fracting:
```
error: target not found: mingw-w64-x86_64-hunspell<1.4
==> ERROR: 'pacman' failed to install missing dependencies.
```
Why this error?

Alexpux:
because hunspell more than 1.4?

fracting:
hunspell 1.3.4-10 is in repo:

http://mirrors.tea-ci.org/msys2/mingw/i686/mingw-w64-i686-hunspell-1.3.4-10-any.pkg.tar.xz

Is it possible a syntax error?

mingw-w64-enchant/PKGBUILD
```diff
@@ -14,7 +14,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-pkg-config")
 depends=(${MINGW_PACKAGE_PREFIX}-gcc-libs
          ${MINGW_PACKAGE_PREFIX}-glib2
          #${MINGW_PACKAGE_PREFIX}-aspell
-         ${MINGW_PACKAGE_PREFIX}-hunspell)
+         ${MINGW_PACKAGE_PREFIX}-hunspell{'>=1.3','<1.4'})
```
Alexpux: Wrong syntax. Also soon hunspell will be updated so what next then?

renatosilva:
That error is expected because hunspell needs to be downgraded. Alternatively, someone can make enchant work, bring that commit back, and have enchant updated instead. The version specification is normal bash syntax.

mati865:
I think we should revert hunspell till enchant is fixed.

PR #1506, opened 2016-06-25
openscenegraph: upgrade to 3.5.3

filnet:
* openscenegraph: upgrade to 3.5.3
* osgearth: rebuild with osg 3.5.3 and gcc 5.4.0


mingwandroid:
Why were all those dependencies listed as makedepends before?

filnet:
No idea...

Please don't merge yet. I think I broker freetype 2 support...

filnet:
freetype2 support should be fine now

PR #1509, opened 2016-06-26
mpg123 release 1.23.5

qme5400:


qme5400:
tea-ci.org fails at arch=32, please also verify removed patches

mati865:
Tea fail is related to https://github.com/Alexpux/MINGW-packages/commit/40222700c0788e0a4bc03e5537cc77f1a7eb3792
@renatosilva could you take a look?

PR #1512, opened 2016-06-26
Fixes for continuous integration

renatosilva:
Drone uses a 32-bit installation of MSYS2 for building both MINGW and MSYS packages for this architecture. Installing latest saneman from repository ci.msys fails because it has been generated by the AppVeyor installation, which is 64-bit.

One possible solution is using only a 64-bit installation in Drone, or generating one additional 32-bit MSYS repository as suggested by @fracting. For now, recipe check has been disabled for Drone 32-bit, thus avoiding installation of saneman.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -3,6 +3,8 @@ build:
   pull: true
   shell: msys$$arch
   commands:
+    # TODO: remove this option when not anymore needed
+    - if [ $$arch = 32 ]; then export DISABLE_QUALITY_CHECK=true; fi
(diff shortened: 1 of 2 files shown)


fracting:
Should be safe to merge like https://github.com/Alexpux/MSYS2-packages/pull/644, maybe @mingwandroid or @alexpux just too busy or forgot this one.

I'll try harder in the future to work with @renatosilva to take care about CI issues. Thank @renatosilva for the great work!

PR #1514, opened 2016-06-27
ffmpeg: update to 3.1

mati865:
libdcadec removed upstream https://ffmpeg.org/pipermail/ffmpeg-cvslog/2016-March/098996.html

There are references to `$(cygpath -m /)` inside /lib/pkgconfig/*.pc files and someone more experienced should take a look at it.


DavidEGrayson:
I think we usually use `sed` to fix the .pc files when that happens.

PR #1523, opened 2016-06-30
Update PKGBUILD for mingw-w64-nlopt

aphirst:
Issues with the build system resulted in the generation of libraries with absent/non-functioning Fortran API. The build system does weird things to generate the C++ library, and then goes back and redoes everything to generate the normal C library. After some discussion and investigation with people in the Freenode `##fortran` community, we were able to build working packages if a `make distclean` statement is placed inbetween those two build stages. I still have some suspicions about whether some Fortran related packages need to be in the build dependencies, but that depends on what is considered "standard" to be installed on the machine generating the packages for the repository.


elieux:
I think this change is fine. If you/we find anything else missing, it can be added in another change.

PR #1527, opened 2016-06-30
scons for mingw32/mingw64

manuelnaranjo:
packing SCons to be installed into /mingw{32,64} this is necessary in some cases, for example when trying to build a python extension from SCons and needed to call a library installed into the system Python.


DavidEGrayson:
Please don't merge this:  binary files do not belong in the MINGW-packages repo.

manuelnaranjo:
Yes sorry, I will close the request

PR #1544, opened 2016-07-11
mpv: update to 0.18.1

mati865:


mati865:
```
==> Starting prepare()...
err:seh:setup_exception stack overflow 7872 bytes in thread 0193 eip 000000007bc99c63 esp 00000000ffff8740 stack 0xffe00000-0xffffa000-0x100000000
```
@fracting 
Looks like mpv building can expose many Wine bugs.

fracting:
> Looks like mpv building can expose many Wine bugs.

@mati865 that's actually very nice... One more bug on my todo list :) Thanks for report!

PR #1548, opened 2016-07-12
unvanquished: new package, 0.50.0 version

mati865:
Depends on https://github.com/Alexpux/MINGW-packages/pull/1547.
Based on https://aur.archlinux.org/packages/unvanquished/

Not meant to be added to pacman yet as 32-bit version crashes with no console output (debug doesn't help either, probably NaCl have to be recompiled).
64-bit version is fine.

I'm going to holidays so I'll put it here if someone wants to help. After return I'll prepare git package.

### DIFF
--- a/mingw-w64-unvanquished-data/PKGBUILD
+++ b/mingw-w64-unvanquished-data/PKGBUILD
@@ -1,0 +1,65 @@
+# ArchLinux (AUR):
+# Maintainer: Viech <viech unvanquished net>
+# MSYS2:
+# Maintainer: Mateusz Mikuła <mati865@gmail.com>
+
+_realname=unvanquished-data
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=0.50.0
+pkgrel=1
+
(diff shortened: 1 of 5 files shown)


Alexpux:
Don't understand why we need this package

mati865:
Like chromium or vlc we don't need it. Just providing easy way to build it.
Of course I can move it to my personal repo.

Alexpux:
Chromium not building bu having web-browser in repo will be good.
VLC is good media player and I use it.

mati865:
Don't get me wrong VLC is great player and I do often use it but vlc git package is outdated and no longer building. I've tried to fix it but couldn't. It was failing at:
build-x86_64-w64-mingw32\src\Makefile:3260
```
fourcc_gen$(BUILDEXEEXT): misc/fourcc_gen.c misc/fourcc_list.h ../include/vlc_fourcc.h
	$(AM_V_at)rm -f -- $@
	$(AM_V_CC)$(BUILDCC) -I$(srcdir) -o $@ $< # here
```

Alexpux:
Git packages in most cases here for developers who want to build it. I'm not supporting git packages myself

PR #1549, opened 2016-07-12
cotire: update for cmake 3.6

Optiligence:


Optiligence:
This isn’t available in the repositories yet.

PR #1562, opened 2016-07-18
template-glib: new package.

KrullKorg:


Alexpux:
I think we need wait release instead building git version

KrullKorg:
i don't know when (and if) it will be released; in the meantime could you still insert it into the repository? without build it

Alexpux:
well then rename package name to ````*-git```` and folder too.

PR #1568, opened 2016-07-19
Update pyqt to 5.6

patstew:


Alexpux:
Seems it not buildable.

fracting:
Builds fine: https://tea-ci.org/Alexpux/MINGW-packages/1110

(The failure in https://tea-ci.org/Alexpux/MINGW-packages/1109 is caused by deleted branch, not a build error)

Alexpux:
For me not. I get infinite of qmake processes

Alexpux:
The last hunk in patch about "make" is cause troubles for me so I'm remove it. We need use mingw32-make instead make when building pyqt5

PR #1575, opened 2016-07-20
Updated lfcxml to version 1.2.4, dependency correction for cego package

bjlemke:


bjlemke:
Since there are dependencies from cego to lfcxml, please ensure the right install order for the packages.
Otherwise the build fails

Issue #1578, opened 2016-07-20
Rebuild wxWidgets with gcc 6.1


jessehager:
I just installed gcc 6.1 and programs built with wxWidgets stopped working:

![wxwidgetsfatalerrorabi1010](https://cloud.githubusercontent.com/assets/1455423/16992504/9c2dbf82-4e6d-11e6-8577-b3e39642b097.png)

The ABI version has been bumped to 1010 and wxWidgets needs to be recompiled with the new gcc version.

This is similar to issue #712 from a year ago.

mati865:
Just create PR with bumped `pkgrel`.

lazka:
wxWidgets has been rebuild multiple times in the meantime.

PR #1584, opened 2016-07-22
Update wxWidgets build files for GCC 6

jessehager:
Fix for issue #1578 for the changed ABI in GCC 6.1.

Bumped pkgrel and added patch from upstream for wxWidgets issue 17147 to
fix compile errors regarding overloaded abs() function in the Scintilla editor component.

Referenced wxWidgets bug http://trac.wxwidgets.org/ticket/17147


jhasse:
When will we see this in MSYS2's updates?

jhasse:
Any news on this? The binary wxWidgets package is still not working for me :(

PR #1604, opened 2016-07-30
update flatbuffers to version 1.3.0

dsogari:


mati865:
@fracting I'm not sure if it's wine related but 64 bit Tea failed: 
```
/usr/bin/colormake: line 34: /usr/bin/make.exe: Permission denied
make[1]: *** [CMakeFiles/Makefile2:253: CMakeFiles/flattests.dir/all] Error 126
```

@dsogari You don't need to worry about it.

PR #1611, opened 2016-08-02
Fixes for DllMain in glib2

jeroenooms:
This fixes part of the problem in https://github.com/Alexpux/MINGW-packages/issues/988.

The [current patch](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-glib2/0004-glib-prefer-constructors-over-DllMain.patch) to avoid `DllMain` is incomplete. If we statically link librsvg or cairo we get **multiple definition of DllMain** linking erros.

The current PR works around `DllMain` for static builds of `gobject.a` and `gdk-pixbuf.a` and only enables `DllMain` when building the shared library.


jeroenooms:
@manisandro has pointed out that similar patches are available here:

 - https://smani.fedorapeople.org/glib-prefer-constructors-over-DllMain.patch
 - https://smani.fedorapeople.org/gdk-pixbuf_static.patch

PR #1618, opened 2016-08-04
Add OpenCASCADE Community Edition (OCE)

nickoe:
Add OpenCASCADE Community Edition (OCE) library


mingw-w64-oce/PKGBUILD
```diff
@@ -0,0 +1,63 @@
+# Maintainer: Nick Østergaard <oe.nick at gmail dot com>
+
+_realname=oce
```
Alexpux: I think the better to have name opencascade instead oce

nickoe: It would be wrong to call it opencascade because it is not just opencascade. It is called oce by everyone else:
OpenSUSE: http://software.opensuse.org/package/OCE
Upstream: https://github.com/tpaviot/oce
Debian: https://packages.debian.org/source/wheezy/oce (in debian they are seperated to multiple packages and called liboce->>module<<)
fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/OCE/
AUR: https://aur.archlinux.org/packages/oce/

nickoe:
There is only a single issue with it as is. The _IMPORT_PREFIX in OCE-libraries.cmake and
OCE_INCLUDE_DIRS from OCEConfig.cmake to be windows paths (I mean with
C:/msys64/mingw64 instead of just /mingw64) when trying to build against it with another cmake project (here tested with kicad).

But I am not sure "fixing" the path in the package() step would be appropiate. And I have not tested this with the latest cmake in MSYS2. I have tested this with cmake 3.3.2, I don't know if those paths work fine in later versions.

nickoe:
Ok, I see, when using the mingw cmake, it is essentially a windows cmake, and it doesn't understand msys paths.

So I guess we should do something similar to https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-clang/llvm-x86_64.install

nickoe:
I have updated the pull request, I think this is ready to merge.

PR #1621, opened 2016-08-05
strigi: new package, version 0.7.8

mati865:


mati865:
Forgot to fix it. Will reopen when it's done.

PR #1634, opened 2016-08-14
qt5: add new functions better fit MSYS2 needs

mati865:
Probably final version of patch.
new functions: `QStandardPaths::standardLocationsMsys()`, `QStandardPaths::writableLocationMsys()`
They don't collide with anything and can be used in everything that calls 'QStandardPaths::standardLocations()` or `QStandardPaths::writableLocation()` but should search MSYS2 paths (e.x. `/mingw{32,64}/etc`, `/mingw{32,64}/share`) instead of Windows (e.x. `%APPDATA%`, `%PROGRAMDATA%`).

KDE porting would greatly benefit from this change.
Support for XDG_* is also included https://wiki.archlinux.org/index.php/XDG_Base_Directory_support

https://doc.qt.io/qt-5/qstandardpaths.html#standardLocations

Example dirs:
```
$ test.exe
GenericDataLocation: D:/msys64/home/mati865/.local/share,D:/msys64/mingw64/local/share,D:/msys64/mingw64/share
TempLocation: D:/msys64/tmp
ApplicationsLocation: D:/msys64/home/mati865/.local/share/applications,D:/msys64/mingw64/local/share/applications,D:/msys64/mingw64/share/applications
AppDataLocation: D:/msys64/home/mati865/.local/share/test,D:/msys64/mingw64/local/share/test,D:/msys64/mingw64/share/test
GenericConfigLocation: D:/msys64/home/mati865/.config,D:/msys64/mingw64/etc/xdg
RuntimeLocation: D:/msys64/tmp/runtime-mati865
HomeLocation: D:/msys64/home/mati865
MusicLocation: C:/Users/mati865/Music
```

### DIFF
--- a/mingw-w64-qt5/0200-create-writableLocationMsys-and-standardLocationsMsy.patch
+++ b/mingw-w64-qt5/0200-create-writableLocationMsys-and-standardLocationsMsy.patch
@@ -1,0 +1,401 @@
+From eb5aa8e9fbb9ae59f7167f5359facabb6e1b186d Mon Sep 17 00:00:00 2001
+From: mati865 <mati865@gmail.com>
+Date: Wed, 10 Aug 2016 16:03:23 +0200
+Subject: [PATCH] create writableLocationMsys and standardLocationsMsys
+
+Based on qstandardpaths_unix.cpp
+We need it to make porting of the KDE easier and improve it's quality.
+---
+ qtbase/src/corelib/io/io.pri                  |   1 +
+ qtbase/src/corelib/io/qstandardpaths.h        |   3 +
+ qtbase/src/corelib/io/qstandardpaths_msys.cpp | 351 ++++++++++++++++++++++++++
+ 3 files changed, 355 insertions(+)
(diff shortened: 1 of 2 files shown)


mati865:
I missed `locate()` and `locateAll()`.
Updated patch soon.

mkae:
This should be submitted to Qt's patch reviewing system once it is working and stable for MSYS2's needs.

mati865:
@mkae I don't think that they will accept patch in it's current state.

mkae:
Well, after all a solution should be developed, which would have a chance to get incorporated into official Qt, no!?

mati865:
I totally agree with you and we should discuss about the best way to fix it. But first let's get KDE working. I'll upload updated patch tomorrow when I get sober.

P. S.
qconfig.cpp patched for MSYS2/Mac detection could provide everything we need.

mati865:
Had to move to the new PR https://github.com/Alexpux/MINGW-packages/pull/1640 because of rebase.

PR #1640, opened 2016-08-17
qt5: add new functions better fit MSYS2 needs

mati865:
New functions: `QStandardPaths::standardLocationsMsys()`, `QStandardPaths::writableLocationMsys()`, `QStandardPaths::locateMsys()`, `QStandardPaths::locateAllMsys()`
They don't collide with anything and can be used in everything that calls `QStandardPaths::standardLocations()`, `QStandardPaths::writableLocation()`, `QStandardPaths::locateMsys()` or `QStandardPaths::locateAllMsys()` but should search MSYS2 paths (e.x. `/mingw{32,64}/etc`, `/mingw{32,64}/share`) instead of Windows (e.x. `%APPDATA%`, `%PROGRAMDATA%`).

KDE porting would greatly benefit from this change.
Support for `XDG_*` is also included https://wiki.archlinux.org/index.php/XDG_Base_Directory_support

https://doc.qt.io/qt-5/qstandardpaths.html#standardLocations

Example dirs:

```
$ test.exe
GenericDataLocation: D:/msys64/home/mati865/.local/share,D:/msys64/mingw64/local/share,D:/msys64/mingw64/share
TempLocation: D:/msys64/tmp
ApplicationsLocation: D:/msys64/home/mati865/.local/share/applications,D:/msys64/mingw64/local/share/applications,D:/msys64/mingw64/share/applications
AppDataLocation: D:/msys64/home/mati865/.local/share/test,D:/msys64/mingw64/local/share/test,D:/msys64/mingw64/share/test
GenericConfigLocation: D:/msys64/home/mati865/.config,D:/msys64/mingw64/etc/xdg
RuntimeLocation: D:/msys64/tmp/runtime-mati865
HomeLocation: D:/msys64/home/mati865
MusicLocation: C:/Users/mati865/Music

Locate() "kf5" in GenericDataLocation:
LocateMsys() "kf5" in GenericDataLocation: D:/msys64/mingw64/share/kf5
```


mati865:
`homeDir` needs improvement.
@mingwandroid Do you think `QT_CONFIGURE_PREFIX_PATH + QLatin1String("/../home") + qgetenv("USER")` would be OK?

mingwandroid:
We probably also need to be mindful of people who run stuff outside of our shells so `qgetenv("$HOME")` if set, with fallback to the `QT_CONFIGURE_PREFIX_PATH ...` maybe?

Also, if it's `qt5-static`, then it's installed one folder more deep, so you may want in that case:
`QT_CONFIGURE_PREFIX_PATH + QLatin1String("/../../home") + qgetenv("USER")`

Cheers!

mati865:
Should I create 2 patches or is there define for static build?

mati865:
@mingwandroid 
@RJVB and @mkae are trying to upstream fix for QStandardPaths https://trac.macports.org/ticket/49925
Here is the patch https://github.com/RJVB/macstrop/blob/master/aqua/qt5-kde/files/fix-qstandardpaths4.patch
We should add MSYS2 support to this patch and send it to Qt codereview, could you help?

RJVB:
On Thursday August 18 2016 14:33:14 Mateusz Mikuła wrote:
> @mingwandroid 
> @RJVB and @mkae are trying to upstream fix for QStandardPaths https://trac.macports.org/ticket/49925
> Here is the patch https://github.com/RJVB/macstrop/blob/master/aqua/qt5-kde/files/fix-qstandardpaths4.patch
> We should add MSYS2 support to this patch and send it to Qt codereview, could you help?

We should be very sure about exactly what we want, why and how before submitting a patch like this for review. An earlier version was already rejected because we failed to defend the "why" aspect. There's also the question which Qt version to target: QSP was refactored in 5.7 and the patch can thus not be the same between 5.6.x LTS and 5.7.x .

Also note that the patch linked above (probably) still contains code implementing a previous mechanism to select the proper mode, at runtime. This should be removed; unless you see an interest in a mechanism where an additional shared library implements the mode selector? 

R.

RJVB:
On Thursday August 18 2016 14:33:14 Mateusz Mikuła wrote:

Hi,


>@mingwandroid 
>@RJVB and @mkae are trying to upstream fix for QStandardPaths https://trac.macports.org/ticket/49925
>Here is the patch https://github.com/RJVB/macstrop/blob/master/aqua/qt5-kde/files/fix-qstandardpaths4.patch
>We should add MSYS2 support to this patch and send it to Qt codereview, could you help?
>
>

CC'ing David Faure who's been involved in this topic too, being the original QSP author.

I've made a first draft of a simplified patch that drops the whole activator link module concept. I haven't yet tested if it builds to a functional Qt but it should:

https://github.com/RJVB/macstrop/blob/master/aqua/qt5-kde-devel/files/fix-qstandardpaths5.patch

As mentioned elsewhere, this patch introduces an optional alternative mode (called XDG but one could just as well call it ALT; could make it more acceptable to the Qt people too). It does that by introducing protected class methods that take a mode switch, as well as a class mode state variable with getter and setter methods. Unmodified code that calls the regular QSP methods will receive paths that depend on the mode state variable (set to "native" by default; xdg/alt mode == OFF).

There is also a QExtStandardPaths class (defined purely in the QSP header file) which exposes the class methods with the mode switch, but gives that switch a default value which can be set at compile time; the default default (sic) value is still "native".

There are thus several options to select the alternative QSP locations:

1- call QSP::setXDGLocations() as early as possible (preferably in global instance of a special class with a ctor that calls this method): in this case every call to a QSP location method should return the XDG/ALT paths, even when made from a plugin
2- use QExtStandardPaths instead of QStandardPaths, and determine the desired mode while compiling. The patch provides 2 preprocessor tokens to handle this completely via the preprocessor.
3- use QExtStandardPaths and an explicit mode specification directly

In most cases 2) will be the preferred option: no code has to be patched, and the decision which QSP mode is appropriate for the library/application/plugin/whatever can be made at build time. Thus, an application can use the native paths but safely use plugins built to use the alternative paths (or vice versa). I cannot say to what extent this is a feasible scenario for MSYS2 but it is for MacPorts.

Note that there is a parallel effort aimed at MS Windows that tries to configure the QSP locations via qt.conf . It might be useful merge the approaches; the exact definition of the alternative mode/locations could be made via qt.conf . I haven't had time to investigate this; questions I raised about where qt.conf is expected on OS X haven't been answered (and I haven't gotten around to finding the answer myself). If this configuration file is expected inside individual application bundles on OS X then the whole idea becomes less appealing (qt.conf could still override a builtin default alternative location mode).

R.

RJVB:
On vendredi 19 ao�t 2016 11:48:29 CEST Ren� J.V. Bertin wrote:
> CC'ing David Faure who's been involved in this topic too, being the original
> QSP author.

The discussions at the last Randa meeting, with the people involved in porting 
KDE applications (e.g. kate/kdevelop/etc.) to OSX and Windows, convinced me 
that a XDG mode for QSP on OSX and Windows is unnecessary.

In any case, QExtStandardPaths is a hack, a proper Qt patch would add paths to 
QSP, rather than creating an "Ext" class.

-- 
David Faure, faure@kde.org, http://www.davidfaure.fr
Working on KDE Frameworks 5

RJVB:
On Friday August 19 2016 12:10:54 David Faure wrote:

> The discussions at the last Randa meeting, with the people involved in porting 
> KDE applications (e.g. kate/kdevelop/etc.) to OSX and Windows, convinced me 
> that a XDG mode for QSP on OSX and Windows is unnecessary.

I am pretty much convinced that none of those people were involved with porting KDE applications to projects like MacPorts, Fink etc. Or MSYS/Cygwin. On the contrary, I have every reason to believe that most of those people have a problem with MacPorts and the like.
Those efforts do need an alternative mode, there's just no way around that if KDE apps are to be installed and function as much as possible as they are on their native/traditional platform. And frankly, I don't believe it'll be feasible to port complex applications like KDE PIM to OS X using the all-encompassingly standalone app bundle approach.

I knew it was a monumental error not to be at Randa, but it just wasn't possible for me. How many of the porters present are actual long-term OS X users who really know the platform beyond what Apple like it to be?

> In any case, QExtStandardPaths is a hack, a proper Qt patch would add paths to 
> QSP, rather than creating an "Ext" class.

I resent calling the Ext class a hack. It is only there to make the patch as transparent as possible, and to limit the maintenance required for existing code. The current approach is also designed to avoid issues where host app and plugin interfere with their respective expectations, a possibility you evoked yourself (and which in practice happens only in very specific conditions).
How is one going to add paths to QSP in such a way that it doesn't apply globally - provide them as an argument to each and every QSP call? If so, why still bother having a QSP class?
I agree that it would be better if the alternative writable and additional readable paths can be specified at runtime rather than being hardcoded, but that still doesn't remove the need for a mode switch if a single Qt install is supposed to cater applications that require native QSP and ones that don't.

R.

RJVB:
On Friday August 19 2016 11:48:29 Ren� J.V. Bertin wrote:
> https://github.com/RJVB/macstrop/blob/master/aqua/qt5-kde-devel/files/fix-qstandardpaths5.patch

I've refactored the patch somewhat, mostly to avoid using the concept of XDG-compliance in the code (except for the actual functions in the Mac-specific file). I've only left in 2 functions with XDG in their names so that I didn't have to rebuild all dependent KF5 code (but it appears they're not even necessary).
The new version also reduces the amount of platform-specific code: more of the new/modified class methods are now available everywhere, but the mode argument only works on platforms that actually have multiple modes. This is determined in the header file, through a dedicated preprocessor token.

R.

RJVB:
On vendredi 19 ao�t 2016 13:18:23 CEST Ren� J.V. Bertin wrote:
> On Friday August 19 2016 12:10:54 David Faure wrote:
> > The discussions at the last Randa meeting, with the people involved in
> > porting KDE applications (e.g. kate/kdevelop/etc.) to OSX and Windows,
> > convinced me that a XDG mode for QSP on OSX and Windows is unnecessary.
> 
> I am pretty much convinced that none of those people were involved with
> porting KDE applications to projects like MacPorts, Fink etc. Or
> MSYS/Cygwin. On the contrary, I have every reason to believe that most of
> those people have a problem with MacPorts and the like. Those efforts do
> need an alternative mode, there's just no way around that if KDE apps are
> to be installed and function as much as possible as they are on their
> native/traditional platform. And frankly, I don't believe it'll be feasible
> to port complex applications like KDE PIM to OS X using the
> all-encompassingly standalone app bundle approach.

OK, you're probably right about OSX, but on Windows the shared data dir allows 
to do this, without the need for a XDG mode on Windows.
All that's needed is to install all of KDE PIM into the same prefix.
So I would suggest to leave Windows out of the discussion: separate issues, 
separate solutions.
Although I'm curious about what's in the commit I'm replying to, where is it?

> I knew it was a monumental error not to be at Randa, but it just wasn't
> possible for me. How many of the porters present are actual long-term OS X
> users who really know the platform beyond what Apple like it to be?
> > In any case, QExtStandardPaths is a hack, a proper Qt patch would add
> > paths to QSP, rather than creating an "Ext" class.
> 
> I resent calling the Ext class a hack. 

I'm just telling you, the Qt developers will never accept a class called 
QExtStandardPaths. The name itself reeks of bad API.

> It is only there to make the patch as
> transparent as possible, and to limit the maintenance required for existing
> code. The current approach is also designed to avoid issues where host app
> and plugin interfere with their respective expectations, a possibility you
> evoked yourself (and which in practice happens only in very specific
> conditions). How is one going to add paths to QSP in such a way that it
> doesn't apply globally - provide them as an argument to each and every QSP
> call? If so, why still bother having a QSP class? 

Well, the mode arg in each call is still a possible option, I don't see why it 
leads to a "why bother" conclusion. More work adapting KDE, but a cleaner API, 
possibly. I have trouble thinking about all this 10 minutes at a time every 
two weeks, it would be more efficient to sit together somewhere and dig into the 
issue, possibly with other OSX people. Will you be at QtCon/Akademy in two 
weeks?

> I agree that it would be
> better if the alternative writable and additional readable paths can be
> specified at runtime rather than being hardcoded, but that still doesn't
> remove the need for a mode switch if a single Qt install is supposed to
> cater applications that require native QSP and ones that don't.

Well, and I keep wondering if things wouldn't be much simpler without this 
requirement. I.e. Qt from MacPorts would be needed by MacPorts apps,
and that Qt can be told at compile time look into MacPorts-like paths, easy 
patch, end of story.

-- 
David Faure, faure@kde.org, http://www.davidfaure.fr
Working on KDE Frameworks 5

mingwandroid:
On MSYS2, I really want our Qts to be as flexible as possible, so that people who aren't bought in to the MSYS2 software distro can still use our Qt libraries to build their applications.

RJVB:
On Saturday August 20 2016 10:16:41 David Faure wrote:

> OK, you're probably right about OSX, but on Windows the shared data dir allows 
> to do this, without the need for a XDG mode on Windows.

Possibly, I must admit that I haven't looked at exactly what MSYS2's needs are here. It does seem likely that they don't need a configuration in which a single Qt build can cater to both needs. It wasn't my idea, initially, to combine our efforts, but since there is an evident overlap it does seem a good idea to pursue a single fits-all solution.

I've been simplifying my patch, moving away from the XDG terminology towards the concept of an ALTernative mode, which is something that I hope is more acceptable and requires less platform-specific code.

> So I would suggest to leave Windows out of the discussion: separate issues, 
> separate solutions.

Probably, though I see some practical advantages to come up with a patch that addresses the as yet unanswered needs, at least before we start to think of launching the review process(es).

> Although I'm curious about what's in the commit I'm replying to, where is it?

Sorry, it didn't survive my pruning:
https://github.com/Alexpux/MINGW-packages/pull/1623#issuecomment-241074130

> > I resent calling the Ext class a hack. 
> 
> I'm just telling you, the Qt developers will never accept a class called 
> QExtStandardPaths. The name itself reeks of bad API.

The name isn't set in stone, I just couldn't come up with a better name more in line with whatever guidelines Qt might have for this. I suppose you agree that a descendant class that overloads the QSP methods is the easiest way to introduce an extra argument without having to patch each and every QSP invocation and if you don't want to add the mode switch to the QSP API itself.

> Well, the mode arg in each call is still a possible option, I don't see why it 
> leads to a "why bother" conclusion.

Not the mode argument, but the idea of specifying a set of paths in each call. 

> More work adapting KDE, but a cleaner API, possibly.

Not just KDE, if adaptation is required for instance because of an obligatory new argument to class methods. Probably not impossible to propose, but it seems it would defeat the purpose of a number of C++ features to do this for an argument with a clearly defined default value.

I agree the API would be cleaner if the mode argument would simply always be there, everywhere, and there would only be the question of how you set the default value for that argument. The reason I've introduced the overloading QExt class to avoid changing the "official" API is that there are apparently concerns with Apple's App Store admission rules. I have no idea exactly how Apple go about checking whether QSP returns acceptable paths. If they just look at compiled application behaviour there shouldn't be any problem with introducing a mode argument with a default value in the current QSP API. Not as long as it doesn't change the default behaviour.

> I have trouble thinking about all this 10 minutes at a time every 
> two weeks, it would be more efficient to sit together somewhere and dig into the 

I agree, for me too.

> issue, possibly with other OSX people. Will you be at QtCon/Akademy in two 
> weeks?

Sadly, no, I don't think so. Is that the event in Berlin (IIRC)?
Maybe it would be possible to find a convenient moment to get together over some kind of video-conferencing during the Akademy?

> Well, and I keep wondering if things wouldn't be much simpler without this 
> requirement. I.e. Qt from MacPorts would be needed by MacPorts apps,
> and that Qt can be told at compile time look into MacPorts-like paths, easy 
> patch, end of story.

Yep. Of course. And MacPorts could also have 2 different Qt installs (and hope you can load a plugin using the one into a host application using the other Qt as long as their versions match). :)
On a more serious note: just how likely would it be to get such an easier patch accepted? It would have to be something that is under control of Qt's configure mechanism, and I have the strong impression that there's a lot of reluctance to make that mechanism any more complex than it already is.
On the other hand, if we're talking about an easy patch that won't be upstreamed then I don't see why I would give up on the idea of a mode switch...

For those who don't know (or forget ;)): the reason why I'm pursuing a solution in which a single Qt build can provide both "native" and alternative (XDG-compliant) QSP locations is that MacPorts aims to provide applications that behave as much as possible as they're expected to. This applies especially to applications that have been designed or specifically adapted to OS X (as opposed to simply being built on OS X and bundled into an app bundle). Those apps are to behave like they behave when you get them from another source, notably the official build. Example: VirtualBox is in MacPorts too; it's supposed to behave like the version provided by Oracle, and that also means both versions have to get their configuration files from the same locations (which means that a future Qt5-based version will need to use QStandardPaths in native mode).
KF5 applications are in a different reality: they'd be expected to behave (and thus interact which each other) like they do on other Unix platforms, and that also includes the interaction with other applications designed around Freedesktop/XDG guidelines that are not based on Qt5.

R

mati865:
`qt5-static` patch haven't been tested.

@RJVB 
Could you forward @mingwandroid message and my patches?

dfaure:
Why not just add these msys paths to the search paths in the existing methods?
What's in these msys paths? Is there any need to *write* to them, or only to look up stuff that happens to be there?

In the current idea for this patch, can you give examples of actual calls (e.g. in KF5 or apps) that would have to be adjusted ? Having to modify many calls to fooMsys() seems like a very intrusive approach (and certainly wrong for upstream).

The recent approach used by kde-windows was the opposite: install stuff where QStandardPaths is looking for it (e.g. BINDIR\data) rather than changing QSP to look where the stuff happens to be installed (XDG-like paths). But I can see how this would be incompatible with msys if the idea of msys is indeed "a unix-looking hierarchy".

Would it be an option to just have a compile-time choice in Qt itself to make it use xdg paths for writing ? (for reading it can just look in both types of locations)

RJVB:
On Saturday August 20 2016 04:02:16 Mateusz Mikuła wrote:

> @RJVB 
> Could you forward @mingwandroid message and my patches?

I have no idea who that is?!

R

mati865:
@dfaure

> Why not just add these msys paths to the search paths in the existing methods?
What's in these msys paths? Is there any need to write to them, or only to look up stuff that happens to be there?

MSYS2 paths must be listed before Windows native paths because if someone want to use another MSYS2 installation for testing configs would overlap. So need to write configs go to `/home/$USER` dir.
Also MSYS2 store things like Linux does but they go to `/mingw32` or `/mingw64` location so Qt have to search `/mingw{32,64}/bin/, `/mingw{32,64}/lib`, `/mingw{32,64}/shared`, etc.

> In the current idea for this patch, can you give examples of actual calls (e.g. in KF5 or apps) that would have to be adjusted ? Having to modify many calls to fooMsys() seems like a very intrusive approach (and certainly wrong for upstream).

It wasn't mean for upstream yet, things went faster than I expected. Creating fooMsys() was temporary solution to make porting of KDE possible. When I did grep it was almost 600 calls of `QStandardPaths::standardLocations()`, `QStandardPaths::writableLocation()`, `QStandardPaths::locate()` or `QStandardPaths::locateAll()` and each returned wrong paths so it wasn't even possible to build KDE Framework.
Example: https://github.com/KDE/kdoctools/blob/05bb1634ee9ae8d5df7fb6b8b25ccdeb4e4eb639/src/xslt.cpp#L412
`LocateAll()` searches `C:/Users/<USER>/AppData/Local`, `C:/ProgramData`, `<APPDIR>`, `<APPDIR>/data` but `kf5/kdoctools/` is actually located in `/mingw{32,64}/share`.

> The recent approach used by kde-windows was the opposite: install stuff where QStandardPaths is looking for it (e.g. BINDIR\data) rather than changing QSP to look where the stuff happens to be installed (XDG-like paths). But I can see how this would be incompatible with msys if the idea of msys is indeed "a unix-looking hierarchy".

At the beginning of porting KDE to MSYS2 stuff was indeed located in `BINDIR\data` and it allowed part KF5 to compile and even work. But it was ugly solution and for an example hicolor icon theme couldn't be used because it was located in `/mingw{32,64}/share`.

> Would it be an option to just have a compile-time choice in Qt itself to make it use xdg paths for writing ? (for reading it can just look in both types of locations)

Like @mingwandroid said:
> On MSYS2, I really want our Qts to be as flexible as possible, so that people who aren't bought in to the MSYS2 software distro can still use our Qt libraries to build their applications.

It would break apps compiled inside MSYS2 but meant to be used outside it.

dfaure:
I think in your case, my previous idea of adding support for environment variables QT_DATA_DIRS, QT_DATA_HOME, QT_CONFIG_DIRS and QT_CONFIG_HOME into QStandardPaths would be the best solution. Rather simple patch (especially if you do that in qstandardpaths_win.cpp for now), and would allow you to add the msys dirs in these env vars, without having to patch 600 QSP calls !

mati865:
I think it could be sufficient. @mingwandroid what do you think about it?

RJVB:
On Saturday August 20 2016 05:12:57 Mateusz Mikuła wrote:

> > Would it be an option to just have a compile-time choice in Qt itself to make it use xdg paths for writing ? (for reading it can just look in both types of locations)

FWIW, that would probably clash with Apple's App Store rules as applications would be able to pick up files from locations which are off-limits according to those rules.

> Like @mingwandroid said:
> > On MSYS2, I really want our Qts to be as flexible as possible, so that people who aren't bought in to the MSYS2 software distro can still use our Qt libraries to build their applications.
> 
> It would break apps compiled inside MSYS2 but meant to be used outside it.

So apparently MSYS2 also has a justifiable use for a solution which provides a QSP mode that can be selected at compile-time (compile time of the dependent code, not of Qt).
But to what extent is MSYS2 a system for the distribution of packages that provides a Unix-like environment to MS Windows? I know it mostly as a shell for the Ming compilers. How does this "used outside it" work, do users have to install the MSYS2 Qt package? Or do you bundle the required Qt libraries in an installer - and if so, would it be so much harder to just use the official Qt packages?

MacPorts is different in this aspect: it aims to provide as much of an all-inclusive, standalone environment as is required in order to guarantee reproducible builds. It can of course be used to provide certain libraries in software builds that are not part of MacPorts (Qt themselves use[d] MacPorts as the source for certain dependencies of their official builds on OS X) but it's not really supposed to be used like that.

R.

mkae:
@dfaure's patch is what I was using for a long time in the OSX/MacPorts-based KDI-CI-system for KF5 for more than a year and it makes things work fine.

I figure you've yourself understood, @mati865, that patching those 600 occurrences of QSP calls isn't the way to go.

mati865:
> So apparently MSYS2 also has a justifiable use for a solution which provides a QSP mode that can be selected at compile-time (compile time of the dependent code, not of Qt).

 dfaure commented an hour ago:
> I think in your case, my previous idea of adding support for environment variables QT_DATA_DIRS, QT_DATA_HOME, QT_CONFIG_DIRS and QT_CONFIG_HOME into QStandardPaths would be the best solution.

Probably @dfaure's solution would be sufficient in our case. We could add defines [here](https://github.com/Alexpux/MSYS2-packages/blob/master/filesystem/msystem)

> But to what extent is MSYS2 a system for the distribution of packages that provides a Unix-like environment to MS Windows? I know it mostly as a shell for the Ming compilers. How does this "used outside it" work, do users have to install the MSYS2 Qt package? Or do you bundle the required Qt libraries in an installer - and if so, would it be so much harder to just use the official Qt packages?

How I see it:
1. Environment to build software (which can be used without having MSYS2 on target machine, you just have to copy correct libraries alongside your app).
2. Standalone environment with easily installed apps.

> I figure you've yourself understood, @mati865, that patching those 600 occurrences of QSP calls isn't the way to go.

Like I said before, it was temporary solution to be able to port KDE. When people return from their holidays we would discuss about better solution.

mkae:
@mati865, so, will it be possible to let MSYS2 build all dependencies for a given application and then simply run it on any machine if you put all dependencies alongside?

I think it would be nice to have a dependency-walker-like script which can create a list of all dependent binaries for a given executable, so that it is easier to create standalone non-MSYS2'ish bundles...

I came across this in a discussion with @elieux on IRC, IIRC... Outcome was that there are indeed a few packages in MSYS2 which already try to get something like that implemented. I don't 100% remember now which ones that were, but I think git was one of them.

mkae:
AppVeyor fails because of a timeout, but what about tea, @mati865?

mati865:
Tea fail is probably some kind of the Wine bug.

Most of the packages is available as pre build binaries already. I usually manually check depends and copy them when using MSYS2 to build packages.

mkae:
@mati865, did you tell @fracting about the tea hang?

mati865:
I think it's reported already but cannot find it since I'm on my mobile right now.

mkae:
OK, good to know. My browser doesn't even render tea-ci's build log...

mkae:
BTW, at the top I noticed that the music location was `C:/Users/mati865/Music` which is probably a mistake or deliberate, as all the other ones were set to be in the MSYS2-prefix `D:/msys64/home/mati865/`...

mati865:
@mkae, I decided to keep it since we don't provide those locations and to let applications integrate with system.

mkae:
@mati865, I've triggered @fracting on gitter regarding tea-ci. :)

mati865:
@dfaure 
Could you also help with `QIcon::themeSearchPaths()` [doc](https://doc.qt.io/qt-5/qicon.html#themeSearchPaths)?
It leads to code which I don't really understand but it also could use `QT_DATA_DIRS.

RJVB:
On Saturday August 20 2016 05:16:57 dfaure wrote:

>I think in your case, my previous idea of adding support for environment variables QT_DATA_DIRS, QT_DATA_HOME, QT_CONFIG_DIRS and QT_CONFIG_HOME into QStandardPaths would be the best solution. 

As Marko reported, that is an excellent solution for testing, but it's not exactly appropriate when you cannot leave those variables set to a fixed value.

I'm not even sure if MS Windows allows you to do things like

%> env QT_DATA_DIRS=bla foo

from a command line, let alone when starting an application from the Explorer (via a .bat file?), and changes in an environment value are not visible to already running applications (double negation intended).

From what I recall, this kind of thing should be possible on MS Windows via qt.conf, as far as it is expected alongside the application executable.

Either solution can be used to define the alternative paths. Or, in a slightly different scenario, the alt mode could be defined as a mode that activates support for fetching information from a configurable source like the environment.

R.

mati865:
@RJVB is right. I missed the case when user would launch application without shell.

mingwandroid:
`qt.conf` sounds like a good idea for the place to switch this stuff on and off, along with the ability for the software to select which QSP mode to use manually. I see 3 values, `QSP_QT_NORMAL`, `QSP_DISTRO_SPECIFIC` (MacPorts and MSYS2) and `QSP_FROM_QT_CONF`. This would be the default and in that mode a value in `qt.conf` controls things.

mati865:
>  mingwandroid commented 2 hours ago • edited

> Re-reading the thread, my opinion is this:
> 
> qt.conf sounds like a good idea for the place to switch this stuff on and off, along with the ability for the software to select which QSP mode to use manually. I see 3 values, QSP_QT_NORMAL, QSP_DISTRO_SPECIFIC (MacPorts and MSYS2) and QSP_FROM_QT_CONF. This would be the default and in that mode a value in qt.conf controls things.
> 
> On a tangential note, at present I am working on enabling updating QtWebKit to the new fork (https://github.com/annulen/webkit) and adding it to MSYS2's qt5-static package. Fingers crossed.
> 

Quoting in case someone misses it in a mail.

Optimally if paths options were extened to include [QDir::homePath()](https://doc.qt.io/qt-5/qdir.html#homePath), [QIcon::themeSearchPaths()](https://doc.qt.io/qt-5/qicon.html#setThemeSearchPaths) and maybe something more.

I'm thinking about QIcon::themeName()](https://doc.qt.io/qt-5/qicon.html#fromTheme). I haven't looked into it's code yet but it probably shouldn't be hard to fix it for MSYS2.

RJVB:
On Sunday August 21 2016 16:13:35 Mateusz Mikuła wrote:

> > qt.conf sounds like a good idea for the place to switch this stuff on and off, along with the ability for the software to select which QSP mode to use manually. I see 3 values, QSP_QT_NORMAL, QSP_DISTRO_SPECIFIC (MacPorts and MSYS2) and QSP_FROM_QT_CONF. This would be the default and in that mode a value in qt.conf controls things.

Two things:

1- there will need to be a default for the QSP_FROM_QT_CONF values in case qt.conf is missing or lacks definitions
2- For the previously mentioned App Store rules the default should really be QSP_QT_NORMAL (I'd call it QSP_QT_NATIVE). I wouldn't be surprised if Microsoft have similar rules in place.

2) could be implemented (and should be, partly) by letting the paths from qt.conf override/complement the native paths. The problem with that is only that it still allows users to change the native paths for something that's not conform the App Store's sandboxing rules, and that's probably just as inacceptable.

Another problem that I've now looked into myself is this:

> QLibraryInfo will load qt.conf from one of the following locations:
> a- :/qt/etc/qt.conf using the resource system
> b- on OS X, in the Resource directory inside the application bundle, for
>   example assistant.app/Contents/Resources/qt.conf
> c- in the directory containing the application executable, i.e.
>   QCoreApplication::applicationDirPath() + QDir::separator() + "qt.conf"

In other words, we'd be looking at patching the build system so that it installs a qt.conf file in the appropriate location for each and every application, *unless* option a) can somehow take the form of a single qt.conf file that applies to all applications using a given Qt install (and that can thus be installed along with Qt).

Patching build systems so that an additional file is installed is not really an option as far as I'm concerned.

> I'm thinking about QIcon::themeName()](https://doc.qt.io/qt-5/qicon.html#fromTheme). I haven't looked into it's code yet but it probably shouldn't be hard to fix it for MSYS2.

That's really a separate issue, but if we start discussing this functionality: on platforms like OS X and MSWin Qt could really do with something like

QApplication::setWindowIconFromTheme(QString &name, QIcon &fallback=QIcon())

as a shortcut for

QApplication::setWindowIcon( QIcon::fromTheme(QString &name, QApplication::windowIcon() )

This is relevant for cross-platform software that has to specify the application dynamically on some platforms (X11/XDG) while on others that icon can already be provided via some resource bundle (and QIcon::fromTheme() is likely to return an empty icon).

That's probably an issue to raise via a Qt request ticket.

R.

fracting:
@mati865 @mkae which hang do you mean? Are you talking about https://bugs.wine-staging.com/show_bug.cgi?id=643 ?

Need someone give a hand to track down into a simple Win32 test case for that bug, or I'll do that few months later :)

mati865:
@fracting, When executing `configure.bat` Tea hangs for few hours/days.
```
==> Starting build()...
+ cd qtbase            # it hanged here for few days
```
https://tea-ci.org/Alexpux/MINGW-packages/1254

fracting:
@mati865 that hang is https://bugs.winehq.org/show_bug.cgi?id=32679

Some more tests is needed to add to Wine test suite and run on https://testbot.winehq.org/ in order to confirm the right fix.

mati865:
@fracting everybody understand it. Nobody is going to blame you for it.
And I must admit Wine is very slow with issues  not related to popular games.

fracting:
@mati865 thank you. I'm just CFH: crying for help :)

mati865:
Due to lack of time I'm closing this PR.
I'll keep this branch in my repo but it won't be updated unless I have time to do it.

PR #1644, opened 2016-08-18
gtk3: update to version 3.20.9

swilmet-ucl:
Here is an update of GTK+ 3. It compiles and works fine for me.


swilmet-ucl:
The above problem is:

```
Build execution time has reached the maximum allowed time for your plan (60 minutes).
```

Not a problem from my side. The patch is trivial, it's just an update to the latest micro version.

PR #1647, opened 2016-08-19
add PKGBUILD for git-lfs

aroig:
This adds a PKGBUILD for git-lfs: https://git-lfs.github.com


aroig:
The CI probably fails due to an old go version (1.4). This builds fine on my windows VM with go 1.6.2.

PR #1659, opened 2016-08-24
Add ANTLR4

dsogari:
ANTLR4 C++ runtime
https://github.com/DanMcLaughlin/antlr4/tree/master/runtime/Cpp

### DIFF
--- a/mingw-w64-antlr4-git/PKGBUILD
+++ b/mingw-w64-antlr4-git/PKGBUILD
@@ -9,7 +9,7 @@ pkgdesc="ANother Tool for Language Recognition v4 (mingw-w64)"
 arch=('any')
 url="http://www.antlr.org/"
 license=('BSD')
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-cmake")
 source=("${_realname}::git+https://github.com/DanMcLaughlin/antlr4.git"
         "${_realname}-${pkgver}.patch")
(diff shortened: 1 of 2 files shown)


Alexpux:
Is it not conflict with v3?

dsogari:
I think not, because the library name for this one is ```libantlr4-runtime.dll``` while ANTLR3 is ```libantlr3c.dll```. BTW, I need to adjust the structure of the installation directories, to make the dll go to the bin folder. I'll update the patch.

Alexpux:
What about headers and import libraries conflict?

dsogari:
Yes, perhaps the include directory needs to be adjusted too. I'll see to it.

All the library files are named ```libantlr4-runtime.*```. Do you think maybe that could cause problems?

mingw-w64-antlr4-git/PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+pkgver=4.5.3.r4904.6f62821
+pkgrel=1
+pkgdesc="ANother Tool for Language Recognition v4 (mingw-w64)"
+arch=('any')
+url="http://www.antlr.org/"
+license=('BSD')
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-cmake")
+source=("${_realname}::git+https://github.com/DanMcLaughlin/antlr4.git"
+        "${_realname}-${pkgver}.patch")
```
mati865: The patch cannot use `${_realname}-${pkgver}` format because it will break with update.

PR #1667, opened 2016-08-25
curl: add missing dependancy of libcurl

wiiaboo:
Fixes cmake not working if lz4 isn't installed.


mati865:
Are you sure that curl depends on lz4?
Cmake didn't work because libarchive depends on lz4 and there is already PR for it.

wiiaboo:
Oh ok. That's probably it then. Didn't notice libarchive was a cmake dependancy as well.

mati865:
In my case it couldn't find libgcc* so it can be confusing sometimes.

PR #1686, opened 2016-08-30
pkg-config: workaround glib-gettext.m4 error and don't override empty prefix setting

filnet:
PR provides two fixes:

1/ workaround glib-getext.m4 error
see https://github.com/Alexpux/MINGW-packages/issues/1351

2/ don't override empty prefix setting
see https://bugs.freedesktop.org/show_bug.cgi?id=97453

Second fix is a prerequisite to upgrading gstreamer from 1.6.3 to 1.8.x


filnet:
bump :)

PR #1694, opened 2016-09-02
jq: make it depend on oniguruma

vszakats:
Oniguruma is a required dependency since version 1.5


vszakats:
Currently if `jq` is installed by itself, invoking `jq` will result in an error due to a missing `.dll`. This patch ensures that the missing `.dll` &mdash; which is oniguruma &mdash; does exist when installing `jq`.

PR #1697, opened 2016-09-05
mingw-w64-oce: use secure checksum (sha256)

vszakats:


vszakats:
Is there any way to globally reject any submissions that are using MD5 or SHA1 checksums? These were proven insecure many years ago.

mati865:
There are warnings like `warning: mingw-w64-oce: deprecated MD5 checksums` when CI builds package but probably not much more can be done.

vszakats:
Can it be modified to be a fatal error instead of a warning?

mati865:
Of course by adding `--strict` to https://github.com/Alexpux/MINGW-packages/blob/master/ci-library.sh#L200.
Quality is checked by @renatosilva script https://github.com/renatosilva/pactoys/tree/master/source/saneman

vszakats:
Thank you for the pointers, @mati865.

vszakats:
Created a PR, in the hope that `msys2.sane.sh` is honored:
https://github.com/renatosilva/pactoys/pull/1/files

This seems to cause less side effect than going for `--strict`, though the latter may be a better option, if feasible.

vszakats:
Created a PR, in the hope that `msys2.sane.sh` is honored:
https://github.com/renatosilva/pactoys/pull/1

This seems to cause less side effect than going for `--strict`, though the latter may be a better option, if feasible.

vszakats:
Failed due to timeout on CI limit.

PR #1702, opened 2016-09-08
nodejs: update to 6.5.0

mati865:
Current progress:
- [x] node.js
- [x] npm
- [ ] node-gyp (undefined references to V8...)
- [ ] XP compatibility (need someone to test it)

I skipped `0008-Add-output-dir-option-to-configure-node_gyp-py.patch` and `0105-Add-gyp-debug-option-to-configure.patch` because I don't understand their point.
I'm open for suggestions.

`node-gyp` build fails during linkage, I've been struggling with it for few days but I couldn't find the reason. Somebody more advanced should look into it.


mati865:
@fracting, Déjà vu?
```
sem_init: Invalid argument
```

fracting:
@mati865 sorry, no time to work on that yet, no enough knowledge on exception handling and no time to learn in the near future.

@all any one could give a hand on https://bugs.wine-staging.com/show_bug.cgi?id=691 ?

Sebastian / Piotr / Alexandre / etc know Windows exception handling very well, but, unfortunately not me...
One important step is to write a simple test case using Win32 API + C + Assembly code to reproduce the bug on Wine, if anyone can make that happen, I think Piotr Caban or Sebastian Lackner can fix the bug on the Wine side very quickly

mati865:
It would be great if somebody with Win XP could test it. Hopefully it will work but without IPv6, or crash...

PR #1746, opened 2016-09-21
emacs: add missing make dependencies

ksjogo:


Alexpux:
This is common dependencies to make most packages. In wiki we write that developers need to install "base-devel" group of packages

ksjogo:
Ok, maybe `makepkg-mingw` should warn about these not being installed, would be more convenient to warn when using it, instead of letting people look it up in another step.

PR #1755, opened 2016-09-24
qt5[-static]: add upstream patch for QTBUG-55482

filnet:
workaround crashes in QtQml code related to dead-store elimination
see https://bugreports.qt.io/browse/QTBUG-55482


mati865:
Does it resolve https://github.com/Alexpux/MINGW-packages/issues/1713?

filnet:
Seems not :(

$ qtcreator
Segmentation fault

PS: I use official QtCreator not msys2 one.

filnet:
But it does address nasty crashes on application startup if Qt5Qml library is compiled with GCC 6.1.1/6.2.0

filnet:
Hi,

Any ETA for publishing the new qt5 ?

Cheers,
Philippe.

Alexpux:
I'm not plan to rebuild Qt5 in near time. Soon will be 5.6.2 released and then we will think about new packages.

filnet:
No problems. Thanks for answering.

Issue #1678, opened 2016-08-27
Issues when building with clang (mixed cstdio linkage)


vit9696:
Hello,

I am currently testing clang compilation and ran into several issues. Due to #1669 being unfixed at this moment I mostly relied on [3.8.0 win32](http://llvm.org/releases/3.8.0/LLVM-3.8.0-win32.exe) and [3.9.0 rc3 win32](http://llvm.org/pre-releases/3.9.0/rc3/LLVM-3.9.0-rc3-win32.exe), however I tested mingw-w64-i686-clang (3.8) as well. Due to reports getting large I decided to split them.

**3. Mixing gcc-compiled libraries with clang objects creates cstdio symbol duplication**

Imagine two files:
_example3.cpp_:

```
#include <cstdio>

void printStr(const char *str);

int main(int argc, char *argv[]) {
    char buf[1024];
    snprintf(buf, 1024, "<<%s>>", argv[0]);
    printStr(buf);
}
```

_example3_gcc.app_:

```
#include <cstdio>

void printStr(const char *str) {
    char buf[256];
    snprintf(buf, 256, "Some temp string %s", str);
    puts(buf);
}
```

Run the following commands on them:

```
clang++ -target i686-w64-mingw32 -std=c++11 -O1 -c example3.cpp
g++ -std=c++11 -O1 -c example3_gcc.cpp
ar rvs example3.a example3_gcc.o
clang++ -target i686-w64-mingw32 -static -std=c++11 -pthread -static example3.o example3.a -o example3
```

The last call (any clang) will produce a linkage error:

```
$ clang++ -target i686-w64-mingw32 -static -std=c++11 -pthread -static example3.o example3.a -o example3
example3.a(example3_gcc.o):example3_gcc.cpp:(.text$_Z8snprintfPcjPKcz[__Z8snprintfPcjPKcz]+0x0): multiple definition of `snprintf(char*, unsigned int, char const*, ...)'
example3.o:(.text[__Z8snprintfPcjPKcz]+0x0): first defined here
clang++.exe: error: linker command failed with exit code 1 (use -v to see invocation)
```

From what I can tell the issue happens due to a strange workaround for gcc present in _msys64\mingw32\i686-w64-mingw32\include\stdio.h_:

```
/* There seems to be a bug about builtins and static overrides of them
   in g++.  So we need to do here some trickery.  */
#ifdef __cplusplus
extern "C++" {
#endif
```

[...]

Issue #1677, opened 2016-08-27
Issues when building with clang (duplicate libstdc++.a symbols)


vit9696:
Hello,

I am currently testing clang compilation and ran into several issues. Due to #1669 being unfixed at this moment I mostly relied on [3.8.0 win32](http://llvm.org/releases/3.8.0/LLVM-3.8.0-win32.exe) and [3.9.0 rc3 win32](http://llvm.org/pre-releases/3.9.0/rc3/LLVM-3.9.0-rc3-win32.exe), however I tested mingw-w64-i686-clang (3.8) as well. Due to reports getting large I decided to split them.

**2. A number of symbols present in libstdc++ headers are exported by libstdc++.a**
The issue is pretty straight-forward. When performing optimised static linking to libstdc++.a one is likely to face multiple symbol definition issue.

Minimal sample (issue confirmed on any clang):
_clang++ -target i686-w64-mingw32 -static -std=c++11 -pthread -O1 example2.cpp_

```
#include <string>

int main () {
    auto s = std::string("1234");
    auto ws = std::u16string(u"1234");
    auto s1 = "#" + s;
}
```

This will produce the following error message:

```
C:\msys64\mingw32\lib\gcc\i686-w64-mingw32\6.1.0\libstdc++.a(string-inst.o):(.text$_ZStplIcSt11char_traitsIcESaIcEENSt7__cxx1112basic_stringIT_T0_T1_EEPKS5_RKS8_[__ZStplIcSt11char_traitsIcESaIcEENSt7__cxx1112basic_stringIT_T0_T1_EEPKS5_RKS8_]+0x0): multiple definition of `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > std::operator+<char, std::char_traits<char>, std::allocator<char> >(char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
C:\msys64\tmp\example2-b21e71.o:(.text[__ZStplIcSt11char_traitsIcESaIcEENSt7__cxx1112basic_stringIT_T0_T1_EEPKS5_RKS8_]+0x0): first defined here
[...]

PR #1768, opened 2016-09-29
[WIP] clang: update to 3.9.0

mati865:
Looks like a huge mess but works better than old Clang (except `_flat128` regression).
Rebased on new @martell patches and applied old and own patches where needed.

#### Compile with GCC, Clang doesn't like mingw-w64 headers.
Example:
```
In file included from D:/projekty/msys2/MATI865-packages/mingw-w64-clang/src/llvm-3.9.0.src/projects/libcxx/src/thread.cpp:35:
In file included from D:\msys64\mingw64\x86_64-w64-mingw32\include\windows.h:114:
D:\msys64\mingw64\x86_64-w64-mingw32\include\stralign.h:121:37: error: cannot initialize return object of type 'PUWSTR_C' (aka 'wchar_t *') with an rvalue of type 'const wchar_t *'
    if(WSTR_ALIGNED(String)) return wcschr((PCWSTR)String,Character);
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
D:\msys64\mingw64\x86_64-w64-mingw32\include\stralign.h:125:37: error: cannot initialize return object of type 'PUWSTR_C' (aka 'wchar_t *') with an rvalue of type 'const wchar_t *'
    if(WSTR_ALIGNED(String)) return wcsrchr((PCWSTR)String,Character);
```

### TODO:
- [ ] lldb (almost compiles)
- [ ] libunwind (untested)
- [ ] testsuite (untested)

### Issues:
- [x] https://github.com/Alexpux/MINGW-packages/issues/886
- [x] https://github.com/Alexpux/MINGW-packages/issues/1541 (fix sent to mingw-w64)
- [x] https://github.com/Alexpux/MINGW-packages/issues/1669
- [ ] https://github.com/Alexpux/MINGW-packages/issues/1676 (@martell we need your help)
- [ ] https://github.com/Alexpux/MINGW-packages/issues/1677
- [ ] https://github.com/Alexpux/MINGW-packages/issues/1678
- [x] https://github.com/Alexpux/MINGW-packages/issues/1679 (0018-experimental-workaround-for-limits-bug.patch)
- [ ] when using lld instead of ld apps segfault (we probably need lldb to solve it)

@martell there is problem with `0002-COFF-Fix-short-import-lib-import-name-type-bitshift.patch`: `test/tools/llvm-readobj/Inputs/imports.mangle.coff-x86_64` doesn't exist

EDIT: Tea fails because of outdated cmake
EDIT2: Added lld issue.


martell:
Adding a note.
We need to back port Reid's fix for libc++abi
https://reviews.llvm.org/rL279786

martell:
Thanks for all the work on this @mati865 btw :)
I assume 0002 is https://reviews.llvm.org/D23540
I'll push that upstream now after running the testbed.

mati865:
@martell, most of patches are your work.
I just had to comment out https://github.com/Alexpux/MINGW-packages/pull/1768/files#diff-fb941abd396fd02f8a6dabb60456016dR94
and add some patches to make it compile with GCC on Windows.
`0041-mingw-w64-hack-and-slash-fixes-for-libc` doesn't work nicely with GCC so I had to patch it with 0042 😄

Libunwind seems important to make libc++abi work but I had no time to even test if it compiles
lldb almost compiles with GCC but it's not as important as getting working Clang

martell:
This is what will replace the 0002 and be pushed to llvm master.
https://gist.github.com/martell/99e00b2f7c1e0bd954455f15a2cfab4f
This won't apply on top of 3.9.0 we can strip it to just the COFF.h change

My main focus for doing this was to get a clang standalone toolchain.
I don't have a high priority on supporting libstdc++ with clang.
So #1677 and #1678 will have to go on the back log for now.

martell:
https://reviews.llvm.org/rL283055
Send Upstream :)

vit9696:
> My main focus for doing this was to get a clang standalone toolchain.
> I don't have a high priority on supporting libstdc++ with clang.

Thanks for your work, but isn't libstdc++ currently the only way to support clang on Windows? I did send you an email questioning the current state of libc++, and it could have gotten lost. However, I do not expect anything to have changed. libc++ would be nice, but getting it work with exceptions is likely going to be years, just like with Linux.

mati865:
Added lld to issues:
> when using lld instead of ld apps segfault (we probably need lldb to solve it)

martell:
lld will not work with a mingw-w64-crt libs build with binutils dlltool.
It complies with the PE COFF spec fully which binutils does not.
That was the reason for the creation of genlib and finally llvm-genlib

mati865:
Ok, I'll try to get lldb building then.
@martell, you are our LLVM expert, feel free to coordinate this work and set priorities.

mati865:
Apparently GCC doesn't support `#pragma comment(lib...)` so we need to add `dbghelp` somewhere in `CMakeLists.txt`, any suggestions?

Alexpux:
@mati865, yes you need add extra link libraries in CMakeLists.txt

mati865:
More fixes with @Alexpux  help.
Currently it fails on linking: http://pastebin.com/xhi2WPag
When it's resolved it will fail here: https://github.com/llvm-mirror/lldb/blob/release_39/tools/lldb-mi/Platform.h#L89

mati865:
Disable lldb to compile it.
The only critical issue is `_float128` and will have to be disabled in Clang somehow.

PR #1771, opened 2016-09-30
SDL2_Mixer: build fixes

fabiangreffrath:
* Bump pkgrel to 2.
* Add smpeg2 to makedepends, some of its M4 macros are needed by autoreconf.
* Add another patch to compile version.rc into a libtool object (which suddenyl became
  necessary after running autoreconf).
* Unpack the zip archive with bsdtar. @mati865: I believe we prefer the zip file over
  tar.gz, because the latter contains symlinks that we have trouble to unpack.


Alexpux:
Well if you use zip file, then you need remove unpacking it in prepare and from noextract array

fabiangreffrath:
Alright, I removed the unpacking step from the prepare stage.

PR #1798, opened 2016-10-10
Openjpeg2

JPeterMugaas:


mingw-w64-openjpeg2/PKGBUILD
```diff
@@ -18,11 +18,12 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-pkg-config"
              "${MINGW_PACKAGE_PREFIX}-cmake")
 options=('staticlibs' 'strip')
-source=(${_realname}-${pkgver}.tar.gz::https://github.com/uclouvain/openjpeg/archive/v${pkgver}.tar.gz
+source=(https://github.com/uclouvain/openjpeg/archive/v$pkgver.tar.gz
```
Alexpux: Back it.

Alexpux:
Don't think this is good. Need to spend some time to fix it properly

JPeterMugaas:
Probably.  One thing I've noticed so far with that JPIP is there doesn't seem to be any clear headers where calling conventions are specified.  At least, I can't figure it out.

PR #1803, opened 2016-10-11
Update glib to 2.50.1

juml:
Does fix https://github.com/Alexpux/MINGW-packages/issues/1796 for me


Alexpux:
Already in my local tree.

PR #1805, opened 2016-10-11
rebuild some gstreamer related packages

filnet:
were using obsolete versions of avcodec, avutil and libx265


mati865:
gst-plugins-bad is not building

Alexpux:
I'm rebuild it yesterday successfully.

PR #1815, opened 2016-10-13
Fix glib2 regression introduced in 2.50.1

juml:
There is an regression introduced in glib 2.50.1
ref: https://bugzilla.gnome.org/show_bug.cgi?id=772054#c7

Fix from master


mati865:
Don't worry about CI failure, this PR is fine.

PR #1818, opened 2016-10-14
vtk: fix generated cmake files to not reference C:/building

filnet:
see https://github.com/Alexpux/MINGW-packages/issues/1670

### DIFF
--- a/mingw-w64-vtk/PKGBUILD
+++ b/mingw-w64-vtk/PKGBUILD
@@ -5,7 +5,7 @@ _realname=vtk
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=7.0.0
-pkgrel=1
+pkgrel=2
 pkgdesc="A software system for 3D computer graphics, image processing and visualization (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mingw-w64-gst-plugins-bad/PKGBUILD
```diff
@@ -48,7 +48,8 @@ depends=("${MINGW_PACKAGE_PREFIX}-celt"
          "${MINGW_PACKAGE_PREFIX}-orc"
          "${MINGW_PACKAGE_PREFIX}-schroedinger"
          "${MINGW_PACKAGE_PREFIX}-soundtouch"
-         "${MINGW_PACKAGE_PREFIX}-openh264")
+         "${MINGW_PACKAGE_PREFIX}-openh264"
+		 )
```
Alexpux: Always use SPACES instead TABS

filnet: that commit is spurious and was removed

Alexpux:
Well I don't remember why this is commented in the past but don't think that this solve problem

filnet:
opencv was failing because of the hard wired C:/building in vtk cmake files.
with this PR there are no more hard wired paths in vtk cmake files and opencv compiles fines.

Issue #1540, opened 2016-07-06
Cargo for rust package


marcthe12:
Cargo is used in rust, it manage rust dependencies and provides a build system. To some extent it is similar to pip. It is no present a all either in rust package or another package. It makes life easy for rust. It is a part of rust lang itself. Add it as a separate package or add it to rust.

It was last mentioned in the issue #1016

marcthe12:
#1016

Gheoan:
Fixed by https://github.com/Alexpux/MINGW-packages/issues/1825.

PR #1094, opened 2016-02-23
[DO NOT MERGE] Update to Rust 1.6

K900:
This should be everything that's needed, but I'm running another build on a clean install to make sure I didn't miss anything

Issue #1016, opened 2016-01-22
Rust package broken


jhasse:
I've installed the Rust package in a fresh MinGW install. Cargo, Rust's package manager, is missing and I can't run `rustc`:

```
jhasse MINGW64 ~
$ rustc --version
C:/msys64/mingw64/bin/rustc.exe: error while loading shared libraries: rustc_driver-e9599659.dll: cannot open shared object file: No such file or directory
```

I restarted the shell, which resulted in no output:

```
jhasse MINGW64 ~
$ rustc --version

jhasse MINGW64 ~
$
```

henryborchers:
rustc works for me.

univerio:
For me, rustc was missing zlib and libffi. I also had a `?` in place of the dll name, but I discovered that if you run rustc in `cmd.exe` a GUI dialog pops up with the name.

Gheoan:
This is fixed by https://github.com/Alexpux/MINGW-packages/issues/1825.

PR #1825, opened 2016-10-15
Include cargo with Rust

hcorion:
This updates the Rust PKGBUILD to build and install cargo as well.
I did some testing and it works well. (also, rust 1.12.0 actually installs now, before it was hitting errors with install Licenses)
This fixes bug #1540 and bug #1016 should be fixed as well. will need to be tested. PR #1094 could be closed as well now.


hcorion:
All the checks will fail because tea is using a old unsupported version of cmake and appveyor will fail because it takes more than 1 hour to build it.

hcorion:
I tested the compiler and cargo with various programs but some fail either because:
- Outdated package
- Developed on devel version of rust.
- The dependencies (or the program itself) use linux tools and haven't been adapted for windows.
- It requires the MSVC version of rust (namely [imgui-rs)](https://github.com/Gekkio/imgui-rs)
Some packages that did build fine were:
[rust64](https://github.com/kondrak/rust64)
[sokoban-rs](https://github.com/swatteau/sokoban-rs)
[zoc](https://github.com/ozkriff/zoc)

PR #1831, opened 2016-10-17
fix devhelp 3.8.2 prefs dialog bug, add devhelp.ico to PKG and hide control window

kuletco:
1. fix prefs dialog bug(notebook not expand).
2. generat devhelp.ico to $prefix/share/devhelp when packaging.
3. hide control window when start devhelp.exe directly in Windows.


mati865:
Increase `pkger` to 2 https://github.com/Alexpux/MINGW-packages/pull/1831/files#diff-382421269b36d29094e03bb4d6dca65cR7

PR #1841, opened 2016-10-19
Generate missing pixbuf loaders

peterbud:
The previous version has not generated the pixbuf loader dlls. After this change make is generating the appropriate loader dlls. I used the [arch version ](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/gdk-pixbuf2) of the PKGBUILD


mati865:
Bump `pkgrel` and it's fine.

PR #1852, opened 2016-10-22
Updated the go package to 1.7.1

hcorion:
Just a small update to the go package to 1.7.1. This is also a reminder to update the official pacman repo. The current version is 1.4.2 which was released at the start of 2015!


Alexpux:
I'm not be able to build GO on my machine. It always hang on building. So I think it will not be in repo soon.

PR #1853, opened 2016-10-22
Update to version 2.31.0

bjlemke:


mati865:
Remove unneeded newlines before `source`, `depends` and `sha256sums`.
Also rename to something like `cego: update to version 2.31.0`

bjlemke:
Removed newlines and changed commit message

PR #1860, opened 2016-10-24
Fix error of depends

akmll:


Alexpux:
This is wrong dependency.

PR #1880, opened 2016-10-30
PortAudio updated to stable release 190600_20161030

qme:
A fresh new stable release since January 30, 2014!


qme:
Consider also about the removing of the following patch files because they are not used in PKGBUILD:
- mingw-w64-cmake-allow-cxx.patch
- mingw-w64-fix.patch

mati865:
Remove them if they aren't needed anymore.

PR #1881, opened 2016-10-31
g-i-parser: fix race condition

achadwick:
For some reason I get the race condition with libmypaint almost always. I've investigated it a bit, and it seems the main culprit is dump.xml not being visible to Python's XML parser at [giscanner/gdumpparser.py:177](https://github.com/GNOME/gobject-introspection/blob/b638438e874fa437adff3bb0480710fe5cbe120c/giscanner/gdumpparser.py#L177).

The launched tmp helper binary sometimes can't see its input .txt after Python creates it, and it sometimes doesn't finish writing its output .xml to the disk by the time Python tries to parse() it. This causes an exception, which was then hidden by some earlier debugging code that looks as if it was trying to solve the same race condition, but only the rmtree() failure.

I'm using WIn7 64-bit, and an NTFS filesystem, with current MSYS2. Building `mingw-w64-libmypaint-git` with `makepkg-mingw -sfL`.

I've cleaned up the exception handling a bit, and just made rmtree() more tolerant than stock. I've also added some wait loops with tests that I'm not proud of but which get the job done. It logs anything that maintainers will need to know about, including all forced waits, and fails only after a 20s timeout.

By live-patching giscanner with some additional utility subprocess.check_call() functions echoing things and doing directory listings for debugging was enough to make the problem go away ~50% of the time. I can only imagine that it added enough delay into the system to allow the files from one process to hit the disk and be believed by the other process.


achadwick:
It seems that semantics are so different on Windows that you can't rely on a process seeing the same instantaneous state of the filesystem as another at all. Even between parent and child, files can be stat()able yet subprocesses can't open() them.

> ```sh
> Test User@win7test MSYS /1
> $ grep --color=auto -i 'functions.txt\|dump.xml' /2/*.log*
> /2/mingw-w64-libmypaint-git-1.3.0beta.1.10.g9491163-1-i686-build.log:<unknown>:: Warning: MyPaint: > dump.xml missing, waiting 2s (1/10)
> grep: /2/mingw-w64-libmypaint-git-1.3.0beta.1.10.g9491163-1-x86_64-build.log: No such file or directory
> ```
>
> So here, grepping my logs as my stress test loop for these changes ran, the shell could stat the new log file, but grep wasn't able to open it. I'm guessing that the native build happening over in /2 set the right share modes and everything in WinAPI's CreateFile() or whatever it's really using to create the log file, but oh my.

So (tl;dr) given the race condition with subprocess.check_call() - which supposedly does wait() for the launched process, but hey this is Windows - we have to do shit like this. In my testing with libmypaint builds though, these patches seem to fix the issue ~100% of the time.

PR #1899, opened 2016-11-05
Python: Fixed patch for distutils.util.change_root to account for cases when neither path has a drive letter

rhelmot:
Found this bug because on msys2, `python setup.py bdist_wheel` produces totally empty wheels, save for metadata! Turns out that `bdist_wheel` calls `install` with a chroot into the build directory set. Without this patch, applying the chroot caused the internal install to get installed to the root of the drive, so nothing ever got copied to the wheel.

This is the result of a solid 6 hours of debugging hahaha


mati865:
@renatosilva would you take a look into these CI false positives?

Issue #1885, opened 2016-11-01
mingw-w64-imagemagick: missing ENV variables


peterbud:
After you install mingw-w64-imagemagick with pacman, apparently it cannot locate its coder modules and configure path. Symptoms are the following errors::
```
$ convert -resize "800x800" test.jpg test.png
convert.exe: no decode delegate for this image format `JPG' @ error/constitute.c/ReadImage/508.
convert.exe: no images defined `test.png' @ error/convert.c/ConvertImageCommand/3253.
```
Adding the following to your bash_profile solves the problem:
```
export MAGICK_HOME="/mingw64/bin"
export MAGICK_CONFIGURE_PATH="/mingw64/etc/ImageMagick-7"
export MAGICK_CODER_MODULE_PATH="/mingw64/lib/ImageMagick-7.0.3/modules-Q16HDRI/coders"
```
` convert -resize 800X800 test.jpg test.png` runs without error

Should have been these variables defined at compile time? Or should they be defined with a profile.d script?

schmittl:
Just came across the same problem.

I am just guessing here, but the `001-7.0.1.3-relocate.patch` for the imagemagick package seems suspicious to me:

```
+CODER.XML      IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\coder.xml"
+COLORS.XML     IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\colors.xml"
+CONFIGURE.XML     IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\configure.xml"
+DELEGATES.XML  IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\delegates.xml"
+ENGLISH.XML    IMAGEMAGICK DISCARDABLE "..\\share\\ImageMagick-6\\english.xml"
+LOCALE.XML     IMAGEMAGICK DISCARDABLE "..\\share\\ImageMagick-6\\locale.xml"
+LOG.XML        IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\log.xml"
+MAGIC.XML      IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\magic.xml"
+THRESHOLDS.XML  IMAGEMAGICK DISCARDABLE "..\\etc\\ImageMagick-6\\thresholds.xml"
[...]

PR #1905, opened 2016-11-06
ImageMagick: Update to 7.0.3.5

peterbud:
Additionally:
Fixed the issue of not finding the coder and filter modules.  Fixes: #1885 
Updated constants to Imagemagick-7
Fixed compile warnings due to redefinition of MAGICKCORE_HAVE_ERF

### DIFF
--- a/mingw-w64-imagemagick/PKGBUILD
+++ b/mingw-w64-imagemagick/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=imagemagick
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=7.0.3.1
+pkgver=7.0.3.5
 pkgrel=1
 pkgdesc="An image viewing/manipulation program (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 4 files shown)


mingw-w64-imagemagick/PKGBUILD
```diff
@@ -48,17 +48,18 @@ optdepends=("${MINGW_PACKAGE_PREFIX}-ghostscript: for Ghostscript support"
             #"${MINGW_PACKAGE_PREFIX}-libwebp: for WEBP support"
             )
 options=('staticlibs' 'strip' '!debug' 'libtool')
-source=(https://www.imagemagick.org/download/ImageMagick-${pkgver%.*}-${pkgver##*.}.tar.xz{,.asc}
+source=(https://www.imagemagick.org/download/ImageMagick-${pkgver%.*}-${pkgver##*.}.tar.xz
```
Chocobo1: you can `makepkg  --skippgpcheck` instead of removing this.

JPeterMugaas: And I think the following line would actually help instead of removing the PGP signature.

validpgpkeys=('D8272EF51DA223E4D05B466989AB63D48277377A')

You could use something like:

gpg --recv-key D8272EF51DA223E4D05B466989AB63D48277377A

To download the public so you can verify with the key.  validpgpkeys should be used more often in these packages to verify the downloads.

Sorry for not replying sooner but I did not get as far as you did.

peterbud: I will get it back it was just my limited knowledge with handling PGP keys, What bothers me is that the Appveyor build has failed at autoreconf:
```

/usr/bin/m4:configure.ac:33: Warning: excess arguments to builtin `m4_define' ignored
autom4te: /usr/bin/m4 failed with exit status: 1
aclocal-1.15: error: echo failed with exit status: 1
autoreconf: aclocal failed with exit status: 1
```
Anybody know how can I fix it?

JPeterMugaas: I'm not sure but I did not see that on my computer.  I have gone ahead and made an adjustment here though.  See: https://github.com/Alexpux/MINGW-packages/pull/1908

The way gpg verification works is that not only do you download the source but also the associated .asc signature.  "makepkg" should automatically verify this with the key listed in "validpgpkeys".  The idea is to verify that you are downloading the proper source-code instead of something from a rogue.  BTW, I did succeed in resizing a JPEG into a PNG with your patches and my minor adjustment.

peterbud: Thanks @JPeterMugaas  I made PGP working also on my machine - next time I will do this way!
WRT the error above: its not appearing n my machine either, its just appearing when the Appveyor CI is building... so its only I want this little green check mark appear at the end of the PR :)

jho101:
Hi,
unfortunately with the latest ImageMagick (I use magick++ in my qt c++ app) the missing decoder problem is still there. Ticket #1885 
For every file format I tried I get the error message "NoDecodeDelegateForThisImageFormat".

peterbud:
There were a number of changes on ImageMagick since this PR in November. I'll look into this if time permits this week.

kirbyfan64:
@jho101 Same issue here.

PR #1921, opened 2016-11-09
clang: fix 32 bit LLDB build

mati865:
32-bit version fails to build


mati865:
This hack: [1.patch.txt](https://github.com/Alexpux/MINGW-packages/files/582322/1.patch.txt) allowed me to build lldb but it's untested.
If someone knows lldb and want's to test this patch the feedback would be nice.

mati865:
@martell what do you think about hack from previous comment?

martell:
According to https://msdn.microsoft.com/en-us/library/14h5k7ff.aspx it seem fine.
It depends on what the `struct stat` actually is.
https://github.com/llvm-mirror/lldb/blob/master/source/Host/windows/FileSystem.cpp#L256
If there is no run time error because of that then this looks good to me for a hack to get it working.

mati865:
Since Vista time is a 64 bit value so it can crash on WinXp. Yet it's probably better to have lldb for Vista and newer than not having it at all.
I'll update this PR.

mati865:
Neither 32 bit or 64 bit lldb crashes.
LLDB is probably fine to deploy it.

Issue #1906, opened 2016-11-06
adwaita-icon-theme: no windows cursors


bpisoj:
Try running gtk3-demo->Cursors and click on any buttons.
As a result cursor either disappear or is not anything like one pictured on button.
Adding line "gtk-cursor-theme-name=Adwaita" to etc/gtk-3.0/settings.ini does not changing anything except printing "Gdk-WARNING **: Failed to load cursor theme Adwaita"

Looks like adwaita-icon-theme ships XCursor and Gtk can not work with them.

If i manually convert first one called "alias" to windows cursor format "alias.cur" there is no more warnings and if you click on first button in "Drag and Drop" in before mention program you see that pointer start look as one pictured on button.

Don't know why PKGBUILD don't produce windows version of cursors but i see on https://github.com/GNOME/adwaita-icon-theme/tree/master/src/cursors two python files whose function is like to be that.

[alias.zip](https://github.com/Alexpux/MINGW-packages/files/574216/alias.zip)

juml:
In official tarball (https://download.gnome.org/sources/adwaita-icon-theme/3.22/adwaita-icon-theme-3.22.0.tar.xz) there are no python scripts (and src/cursors/pngs/make-w32.sh too).

lazka:
I've created an upstream patch for this: https://bugzilla.gnome.org/show_bug.cgi?id=774671

lazka:
It's now in upstream master: https://git.gnome.org/browse/adwaita-icon-theme/commit/?id=fa1a4b9181111c3e4d0111ee87f8b4288a1f1ecb

PR #1922, opened 2016-11-10
New package : adwaita-icon-theme-git

juml:
New package to build adwaita-icon-theme from git.

Created as released tarballs doesn't contain files to generate windows cursors (ani/cur).

Should address https://github.com/Alexpux/MINGW-packages/issues/1906


mingw-w64-adwaita-icon-theme-git/PKGBUILD
```diff
@@ -0,0 +1,81 @@
+# Maintainer: Alexey Pavlov <alexpux@gmail.com>
```
mati865: Place yourself here

mingw-w64-adwaita-icon-theme-git/PKGBUILD
```diff
@@ -0,0 +1,81 @@
...
+
+_realname=adwaita-icon-theme
+pkgbase=mingw-w64-${_realname}-git
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+pkgver=3.22.0
```
mati865: `pkgver()` is missing since it is a git package
Example https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-github-git#L25

juml:
Btw I noticed license is not set correctly (applies to adwaita-icon-theme too I did reuse)

https://git.gnome.org/browse/adwaita-icon-theme/tree/COPYING has:
This work is licenced under the terms of either the GNU LGPL v3 or
Creative Commons Attribution-Share Alike 3.0 United States License.

lazka:
Couldn't the missing files be patched into the tarball based package? That way everyone can benefit.

juml:
Too much files are missing in tarball.
Alternative is rework adwaita-icon-theme to build from git.
What's general approach here ?

Alexpux:
Well I think that issue with missing files will be in next stable release. Just will wait it. Adding new "git" package not solve your issue as I will not build it and place in repo.
"git" packages in repo in most cases for developers only - for your own build

lazka:
I've created an upstream patch for this: https://bugzilla.gnome.org/show_bug.cgi?id=774671

juml:
At this point package not being an solution for #1906. 
For now closing PR then. I will check once better win32 cursor generation available upstream.

PR #1926, opened 2016-11-13
Add graphene

achadwick:
Seemingly now a build dependency of git GTK.

Introspection isn't working yet.

### DIFF
--- a/mingw-w64-graphene/PKGBUILD
+++ b/mingw-w64-graphene/PKGBUILD
@@ -7,7 +7,7 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=${_realver}
 pkgrel=1
 arch=('any')
-pkgdesc="A thin layer of types for graphic libraries"
+pkgdesc="A thin layer of types for graphic libraries (mingw-w64)"
 depends=()
 makedepends=(
     "make"
--- a/mingw-w64-graphene/PKGBUILD
+++ b/mingw-w64-graphene/PKGBUILD
@@ -8,7 +8,10 @@ pkgver=${_realver}
 pkgrel=1
 arch=('any')
 pkgdesc="A thin layer of types for graphic libraries (mingw-w64)"
-depends=()
+depends=(
+    "${MINGW_PACKAGE_PREFIX}-glib2"
+    "${MINGW_PACKAGE_PREFIX}-gettext"
+)
 makedepends=(
     "make"
     "automake"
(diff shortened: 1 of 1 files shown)


mingw-w64-graphene/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=${_realver}
+pkgrel=1
+arch=('any')
+pkgdesc="A thin layer of types for graphic libraries (mingw-w64)"
+depends=(
+    "${MINGW_PACKAGE_PREFIX}-glib2"
+    "${MINGW_PACKAGE_PREFIX}-gettext"
+)
+makedepends=(
```
Alexpux: Remove all this from makedepends. This is common dependencies for all packages building. Here we write only specific for this package dependencies.

mingw-w64-graphene/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+    "libtool"
+    "pkg-config"
+    "gtk-doc"
+    # "${MINGW_PACKAGE_PREFIX}-gobject-introspection"
+)
+#options=('!strip' 'debug' 'staticlibs')
+options=('strip' '!debug')
+license=("MIT")
+url="http://ebassi.github.com/graphene"
+source=("https://github.com/ebassi/graphene/archive/${_realver}.tar.gz")
```
Alexpux: Change source entry to download tarsal into named  file. See other packages.

mingw-w64-graphene/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+
+prepare() {
+    cd "${srcdir}"/${_realname}-${_realver}
+}
+
+build() {
+    [[ -d build-${MINGW_CHOST} ]] && rm -rf build-${MINGW_CHOST}
+    cp -a "${srcdir}/${_realname}-${_realver}" build-${MINGW_CHOST}
+    cd build-${MINGW_CHOST}
+    ./autogen.sh \
```
Alexpux: I think proper is to run autogen in prepare and then run configure in build.

mingw-w64-graphene/PKGBUILD
achadwick: Okay. I dropped gtk-doc too, since it's not used - though omitting it has broken GTK in the past.

PR #1931, opened 2016-11-14
Cmake and deps

JPeterMugaas:


Alexpux:
Building libarchive with cmake break many stuff because of other library naming. I will revert thios changes. Libarchive build of with configure so no any reason to switch to cmake

JPeterMugaas:
It was about the only way I could figure to enable “Cryptography-Next Generation” which requires bcrypt.  With autoconf, it found becrypt but comnplained that it wouldn’t compile so bcrypt was not enabled.  We might have to the equivilant of what I was doing with cmake but I had not figured out how to do that.

PR #1938, opened 2016-11-15
cotire: update for cmake 3.7

Optiligence:
You can also remove the `_cmakegooduntil` if it is too annoying on rebuilds. 
(the previous version was never updated in the repos)
But I like to keep it as a reminder.


mingw-w64-cotire/PKGBUILD
```diff
@@ -4,8 +4,8 @@ _realname=cotire
 pkgbase=mingw-w64-${_realname}
 pkgname="$MINGW_PACKAGE_PREFIX-$_realname"
 _cotirever=1.7.8
-_cmakever=3.6
-_cmakegooduntil=3.7
+_cmakever=3.7
+_cmakegooduntil=3.8
 pkgver=${_cotirever}_${_cmakever}
 pkgrel=1
```
peterbud: I believe @mati865 is right. The current version of the package is the following:
```
$ pacman -Si mingw-w64-x86_64-cotire
Repository      : mingw64
Name            : mingw-w64-x86_64-cotire
Version         : 1.7.8-1
Description     : CMake module to speed up builds (automated PCH, unity builds).
                  (mingw-w64)
Architecture    : any
URL             : https://github.com/sakra/cotire
Licenses        : custom
Groups          : None
Provides        : None
Depends On      : None
Optional Deps   : None
Conflicts With  : mingw-w64-x86_64-cmake<3.4  mingw-w64-x86_64-cmake>=3.5
Replaces        : None
Download Size   : 35.75 KiB
Installed Size  : 161.00 KiB
Packager        : Alexey Pavlov <alexpux@gmail.com>
Build Date      : Tue, Apr 5, 2016 7:46:38 PM
```
A few things to observe: 
* As you see the Version shows only 1.7.8. 
* The refresh date is 5th April 2016.
That means that your [previous updates ](https://github.com/Optiligence/MINGW-packages/commit/c32d96a47a074c52413f5814f2ed6e72a1048555#diff-3ddd311cecf18ab3179b41e0091827bf) from July is still not in the repository. (BTW: April's `pkgver=1.7.8`, July update `pkgver=1.7.8_3.6`)

My advice would be to keep thing simple:
* use the original author's release numbers, which is still 1.7.8, as I see. 
* You need to increase `pkgrel`, otherwise pacman won't know that this is a newer version, and the new version won't be refreshed at the people who have this pakage already installed.

mati865: > Why? pkgver is incremented because it contains the CMake version. 

Let's ask @Alexpux why it wasn't updated.

Alexpux: Well, here is two possibilities. As cmake version changed we already have different version in name so its ok to not bump pkgrel, bt U think we need remove cmake version from versioning and leave only cotire version itself and then only change pkgrel when rebuilding with other cmake

Optiligence: @Alexpux The question is: Was there any reason for the previous package version to not be present in the repos?

Optiligence: @Alexpux I’m not sure that I understand. 
The package needs to be updated for every minor CMake release. 
Otherwise CMake won’t find it anymore, because the directory it is searching changes: 
`modules_dir="$pkgdir$MINGW_PREFIX/share/cmake-$_cmakever/Modules/"`

Alexpux: I mean that every rebuild you will just increase pkgrel. Also maybe we need to place files into unversioned folder:
````modules_dir="$pkgdir$MINGW_PREFIX/share/cmake/Modules/"````

In this case no need to rebuild package every cmake update

Optiligence: I don’t think CMake searches the unversioned directory, i think i tried that.

Alexpux: At least we have some packages that install files in to unversioned cmake folder in share. Also as option can be ````/mingw*/lib/cmake````

Optiligence: I just tried: `share/cmake/`, `share/cmake/Modules/`, `lib/cmake/` and what I don’t think exists anyway: `lib/cmake/Modules/`. None of them are searched by default.

PR #1939, opened 2016-11-15
latexila: updated to 3.22.0.

KrullKorg:


mati865:
Maybe it should be noted somewhere that `BUILDDIR=/tmp makepkg...`  have to be used?
@Alexpux what do you think?

aphirst:
I'm looking forward to this hitting the repo for download, I tried building this myself and, although the build seemed successful, the resulting binaries "run" but with no output.

Issue #1884, opened 2016-11-01
mingw-w64-gnome-doc-utils not in the repository?


peterbud:
I see that there should be a [mingw-w64-gnome-doc-utils  ](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-gdk-pixbuf2)in the MINGW repository, though you cannot find it by `pacman -Ss doc-utils` and you cannot install by `pacman -S mingw-w64-gnome-doc-utils`

Anybody knows why?

lazka:
`pacman -S gnome-doc-utils` works here

https://github.com/Alexpux/MSYS2-packages/tree/master/gnome-doc-utils

peterbud:
MSYS2 package is not the same as MINGW package.
This is the source of [MSYS2 package](https://github.com/Alexpux/MSYS2-packages/tree/master/gnome-doc-utils) and this is the source of [MINGW package](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-gnome-doc-utils)
I can access the MYS2 version, but I'd need the MINGW version.

Alexpux:
I'm not create mingw package for gnome-doc-utils because it need docbook-xml working.

peterbud:
I got it. Let me see whether I can help. Do you know what is **not** working today with the docbook-xml?

Alexpux:
Well it hard to tell. Mostly it not work because of wrong paths or wrong conversion

peterbud:
@Alexpux : I have spent some time with the mingw-w64-docbook-xml. What I have found is the following: the created XML catalog files and the docbook-xml file is OK, and all the DTD files seem to be in place. Nevertheless I was unsuccessful to validate a docbook file (`sample.xml`) with the following command:
`xmllint  --valid --nonet sample.xml`

Turning on the catalog debug reveals the following:
`XML_DEBUG_CATALOG=1 xmllint  --valid --nonet sample.xml`
```
Resolve: pubID -//OASIS//DTD DocBook XML V4.5//EN sysID http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd
Failed to parse catalog file:///etc/xml/catalog
```

So it seems that the default catalog location is **wrong**. With further research I have found that actually you can validate the docbook file with the installed DTDs if you use the following command:
`XML_DEBUG_CATALOG=1 XML_CATALOG_FILES="/mingw64/etc/xml/catalog" xmllint  --valid --nonet sample.xml`

[...]

Issue #1670, opened 2016-08-27
Many packages reference their build path ('C:/building/msys64'), not their install path 


emmenlau:
I have problems building an application that links vtk. Standard cmake tries to get libraries from `'C:/building/msys64'` that don't not exist on my machine. I've checked now a bit more broadly, and indeed there are hundreds of references to this build path. I've checked with:

```
grep -Irn 'C:/building/' /
```

and found literally hundreds of references. Here a few examples:

```
/mingw64/bin/curl-config:159:          echo ${exec_prefix}/lib/libcurl.a -LC:/building/msys64/mingw64/lib -pipe -LC:/building/msys64/mingw64/lib -LC:/building/msys64/mingw64/lib -lnghttp2 -lidn -lrtmp -lssh2 -lssl -lcrypto -lssl -lcrypto -lgdi32 -lwldap32 -lz -lws2_32
/mingw64/bin/cythonize2-script.py:1:#!C:/building/msys64/mingw64/bin/python2.exe
/mingw64/bin/freetype-config:144:  staticlibs="-lfreetype -L/mingw64/lib -lz -L/mingw64/lib -lbz2 -L/mingw64/lib -lpng16 -lz -L/mingw64/lib -LC:/building/msys64/mingw64/lib -L/mingw64/lib -lharfbuzz -lglib-2.0 -lintl -pthread -lws2_32 -lole32 -lwinmm -lshlwapi -lpcre -lintl -lpcre -lgraphite2"
/mingw64/bin/python-config.sh:43:prefix_build="C:/building/msys64/mingw64"
/mingw64/bin/python2-pyuic5.bat:1:@C:/building/msys64/mingw64/bin/python2.exe -m PyQt5.uic.pyuic %1 %2 %3 %4 %5 %6 %7 %8 %9
/mingw64/bin/python3-config:39:prefix_build="C:/building/msys64/mingw64"
/mingw64/bin/python3.5-config:39:prefix_build="C:/building/msys64/mingw64"
/mingw64/bin/python3.5m-config:39:prefix_build="C:/building/msys64/mingw64"
/mingw64/bin/xml2-config:89:            echo -lxml2 -LC:/building/msys64/mingw64/lib -lz -LC:/building/msys64/mingw64/lib -llzma   -liconv
[...]

PR #1948, opened 2016-11-16
libxml2: Eliminating hard-coded paths in pkg-config and xml-config files

### LINKED ISSUE #1864: Question: xml2-config contains hard-coded library path
I have noticed that the xml2-config file contains hard-coded library path, like "C;/building"

```
    --libs)
        if [ "`uname`" = "Linux" ]
    then
        if [ "-L${libdir}" = "-L/usr/lib" -o "-L${libdir}" = "-L/usr/lib64" ]
        then
        echo -lxml2 -LC:/building/msys64/mingw64/lib -lz -LC:/building/msys64/mingw64/lib -llzma   -liconv    
        else
        echo -L${libdir} -lxml2 -LC:/building/msys64/mingw64/lib -lz -LC:/building/msys64/mingw64/lib -llzma   -liconv    
        fi
    else
        echo -L${libdir} -lxml2 -LC:/building/msys64/mingw64/lib -lz -LC:/building/msys64/mingw64/lib -llzma   -liconv     -lws2_32
    fi
        ;;
```

Is this something which we should worry about, or it is OK?

peterbud:
Also eliminating one compile time warning due to LIBXML-STATIC redefinition
Fixes #1864 , partially #1670

### DIFF
--- a/mingw-w64-libxml2/0026-mingw-relocate.patch
+++ b/mingw-w64-libxml2/0026-mingw-relocate.patch
@@ -1,0 +1,696 @@
+diff -Naur '--exclude=.vscode' '--exclude=autom4te.cache' '--exclude=*.in' libxml2-2.9.4.orig/catalog.c libxml2-2.9.4/catalog.c
+--- libxml2-2.9.4.orig/catalog.c	2016-05-23 09:25:25.000000000 +0200
++++ libxml2-2.9.4/catalog.c	2016-11-20 10:19:34.468533500 +0100
+@@ -42,6 +42,7 @@
+ #include <libxml/globals.h>
+ 
+ #include "buf.h"
++#include "pathtools.h"
+ 
+ #define MAX_DELEGATE	50
+ #define MAX_CATAL_DEPTH	50
+@@ -69,16 +70,12 @@
+ 
+ #define XML_URN_PUBID "urn:publicid:"
+ #define XML_CATAL_BREAK ((xmlChar *) -1)
+-#ifndef XML_XML_DEFAULT_CATALOG
+-#define XML_XML_DEFAULT_CATALOG "file:///etc/xml/catalog"
+-#endif
++
+ #ifndef XML_SGML_DEFAULT_CATALOG
+ #define XML_SGML_DEFAULT_CATALOG "file:///etc/sgml/catalog"
+ #endif
+ 
+ #if defined(_WIN32) && defined(_MSC_VER)
+-#undef XML_XML_DEFAULT_CATALOG
+-static char XML_XML_DEFAULT_CATALOG[256] = "file:///etc/xml/catalog";
+ #if defined(_WIN32_WCE)
+ /* Windows CE don't have a A variant */
+ #define GetModuleHandleA GetModuleHandle
+@@ -3146,8 +3156,10 @@
+ 		}
+ 		catalogs = XML_XML_DEFAULT_CATALOG;
+     }
++#elif defined(__MINGW64_VERSION_MAJOR)
++	    catalogs = single_path_relocation(BINDIR, XML_XML_DEFAULT_CATALOG);
+ #else
+-	    catalogs = XML_XML_DEFAULT_CATALOG;
++		catalogs = XML_XML_DEFAULT_CATALOG;
+ #endif
+ 
+ 	catal = xmlCreateNewCatalog(XML_XML_CATALOG_TYPE,
+diff -Naur '--exclude=.vscode' '--exclude=autom4te.cache' '--exclude=*.in' libxml2-2.9.4.orig/include/libxml/catalog.h libxml2-2.9.4/include/libxml/catalog.h
+--- libxml2-2.9.4.orig/include/libxml/catalog.h	2016-02-09 11:17:32.000000000 +0100
++++ libxml2-2.9.4/include/libxml/catalog.h	2016-11-20 08:06:28.669019800 +0100
+@@ -175,6 +175,19 @@
+ XMLPUBFUN const xmlChar * XMLCALL
+ 		xmlCatalogGetPublic	(const xmlChar *pubID);
+ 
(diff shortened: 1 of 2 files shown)


peterbud:
Also fixes #1884

mingw-w64-libxml2/0026-mingw-relocate.patch
```diff
@@ -0,0 +1,709 @@
...
+-#ifndef XML_XML_DEFAULT_CATALOG
+-#define XML_XML_DEFAULT_CATALOG "file:///etc/xml/catalog"
+-#endif
++/*
++#if !defined(XML_XML_DEFAULT_CATALOG)
++	#if !defined(_WIN32)
++		#define XML_XML_DEFAULT_CATALOG "file:///etc/xml/catalog"
++	#elif defined(_WIN32) && defined(_MSC_VER)
++		static char XML_XML_DEFAULT_CATALOG[256] = "file:///etc/xml/catalog";
++	#elif defined(_WIN32) && defined(__MINGW32__)
```
mati865: It's not that simple because both 32 and 64 bit mingw defines `__MINGW32__ `.

mingw-w64-libxml2/0026-mingw-relocate.patch
```diff
@@ -0,0 +1,709 @@
...
++		#define XML_XML_DEFAULT_CATALOG "file:///mingw32/etc/xml/catalog"
++	#elif defined(_WIN32) && defined(__MINGW__)
```
mati865: You could check for `defined(_WIN64) && defined(__MINGW__)` first and for `_WIN32` later.

mati865: Or even better check for `__MINGW64__` only since `_WIN32` is covered by https://github.com/Alexpux/MINGW-packages/pull/1948/files/321a5d8fbf16b460eee0af990506608434280b53..9a558a0dcaf502173f093a45f2d94202dc1db3b7#diff-9e204cd6d4a4f7cc09292a308b5b430bR21 already

mingw-w64-libxml2/0026-mingw-relocate.patch
```diff
@@ -0,0 +1,709 @@
...
+ #define XML_URN_PUBID "urn:publicid:"
+ #define XML_CATAL_BREAK ((xmlChar *) -1)
+-#ifndef XML_XML_DEFAULT_CATALOG
+-#define XML_XML_DEFAULT_CATALOG "file:///etc/xml/catalog"
+-#endif
++/*
++#if !defined(XML_XML_DEFAULT_CATALOG)
++	#if !defined(_WIN32)
++		#define XML_XML_DEFAULT_CATALOG "file:///etc/xml/catalog"
++	#elif defined(_WIN32) && defined(_MSC_VER)
```
peterbud: True, just I did not want to change the existing codebase from the upstream, altough IMHO its rather sure it has never worked on Windows in that form

mingw-w64-libxml2/0026-mingw-relocate.patch
peterbud: Aaahm thats actually a typo... I meant __MINGW64__ thanks for spotting

mingwandroid:
Please squash these commits via:

`git rebase -i HEAD~4`

.. change the last three uncommented lines to either `s` or `squash`, edit the message then force push.

It keeps the history a lot cleaner.

mati865:
@mingwandroid It's not such a big deal since it will be merged into 1 commit like https://github.com/Alexpux/MINGW-packages/commit/6f39a40fe20483d881c4b2fc4b15b8cd87487982

mingwandroid:
Fair enough, I personally prefer to do it that way regardless.

mati865:
@mingwandroid could you suggest how to fix mess like this https://github.com/Alexpux/MINGW-packages/pull/1955 ?

mingwandroid:
Yeah, @JPeterMugaas should be using`git pull --rebase` instead of any `git merge` to avoid getting into that situation on the first place.

Once things get like this though, the safest option  is to rename the local repo folder to MINGW-packages.bak, reclone, use meld3 (or beyond compare) to copy changes from .bak to the new clone, recreate the commits then `git push --force` finally.

JPeterMugaas:
I am using GitHub Desktop.  I would frequently make branches off the main repository.  In addition, I would sometimes I use TortiseGit.  I think those might using Merge but I'm not sure.  I can't use the rebase properly and the command line is difficult to use so I prefer GUI's.

mati865:
Sometimes when I use GUI it's either GitKraken or Sourcetree. Had no problems so far.

mingwandroid:
I do like the GUI tools for some operations but I feel that learning the real/command line git gives an understanding of things as they actually work. Try to configure/drive them to via rebase if you can.

Nonetheless the actual work is the most important thing so thanks for both of your valuable and continued contributions.

peterbud:
@mingwandroid : Thanks for the advice. I now read a bit more about this, and I have a question: can I do rebase even though I have pushed these changes to this repository? Wouldn't that cause a problem here? I read something like this [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

mati865:
@peterbud You can do interactive rebase and force push.

peterbud:
I guess that works now. But that vi editor... I need to do a lookup every time even for the most basic commands. I need to configure something more convenient for me :)

mati865:
@peterbud you can set/export `EDITOR=nano` variable so you will be welcomed by more beginner friendly tool 😃

mingwandroid:
Force pushing all day long to your PR branches is totally fine and the best way to get your commits in order.

Force pushing to branches others have cloned and based their own commits on top of is really bad.. the others can get into a messed up situation when they `git pull --rebase` but even when that's happened the reflog can often be used to retrieve the work.

PR #1950, opened 2016-11-16
gdk-pixbuf2: Make the loaders cache relocatable

lazka:
And instead of creating it at install time include it in the package.


lazka:
Hm, on a second thought it might be better if the query executable would write relative paths in case the special env var isn't set and it was build with "relocatable"

The approach in this PR breaks easily when someone calls gdk-pixbuf-query-loaders.exe --update-cache.

I'll open a new PR when I have something.

PR #1955, opened 2016-11-18
Nghttp2

JPeterMugaas:


peterbud:
@JPeterMugaas : just a question - Why you have som many commits in this PR?

mati865:
@peterbud It happens when you use `git-merge` instead of `git-rebase`.

peterbud:
Hmmm I'm always using merge in my forked master branch, and then I create a new branch from there, then initiate a PR to the upstream, and I haven't had such symptoms...

mati865:
I'm not a git expert but I think it tries to do fast forward merge first, and if it fails then merge commits are created. GitHub sees that repositories have different history and tries to update history of https://github.com/Alexpux/MINGW-packages to look like repository from PR.

JPeterMugaas:
I have gone ahead and reset what I have.  I will be retrying this request.

PR #1973, opened 2016-11-22
New package: zopfli 1.0.1

Chocobo1:


mingw-w64-zopfli/PKGBUILD
```diff
@@ -0,0 +1,28 @@
+_realname=zopfli
```
Alexpux: Add header to PKGBUILD. See other packages.

Chocobo1: Done, however I don't want to disclose my email.

PR #1975, opened 2016-11-23
GNUPGP and some things it depends upon

JPeterMugaas:
mingw-w64-gnupg - 2.1.16 - Update to latest version -disable doc generation as it was segfaulting with 32-bit version
mingw-w64-gnutils - 3.5.6 - Update to latest version.
mingw-w64-libgpg-error - 1.25 - Update to latest version
mingw-w64-npth - 1.3 - Update to latest version
mingw-w64-readline - 7.0 Updated to first patch


Alexpux:
Why adding adns-git package? We have already adds from release.
I think for git packages that we not build for repo need create separate repo with scripts as git packages increase repo a lot.

JPeterMugaas:
I had added that adns package because the test build scripts where indicating that adns was missing for some odd reason.

Alexpux:
Well I think need to figure why current ADNS not funded instead adding new package

PR #1977, opened 2016-11-24
globjects: Add new package

nazar554:
[globjects ](https://github.com/cginternals/globjects)is an MIT licensed, cross-platform C++ wrapper for OpenGL API objects.

### DIFF
--- a/mingw-w64-globjects/mingw-unix-layout.patch
+++ b/mingw-w64-globjects/mingw-unix-layout.patch
@@ -1,0 +1,54 @@
+--- globjects-1.0.0/CMakeLists.txt.orig	2016-11-24 19:18:49.597354100 +0200
++++ globjects-1.0.0/CMakeLists.txt	2016-11-24 19:18:18.933902100 +0200
+@@ -103,9 +103,9 @@
+ set(project ${META_PROJECT_NAME})
+ 
+ # Check for system dir install
+-set(SYSTEM_DIR_INSTALL FALSE)
++option(SYSTEM_DIR_INSTALL OFF)
+ if("${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr" OR "${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr/local")
+-    set(SYSTEM_DIR_INSTALL TRUE)
++    set(SYSTEM_DIR_INSTALL ON CACHE BOOL "Check for system dir install")
+ endif()
+ 
+ # Installation paths
+@@ -123,6 +123,15 @@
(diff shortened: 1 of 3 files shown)


nazar554:
Any thoughts on this? tea-ci build currently fail due to old gcc version (5.4.0 when appveyor is 6.x)

mingw-w64-globjects/mingw-unix-layout.patch
```diff
@@ -0,0 +1,54 @@
...
+@@ -123,6 +123,15 @@
+     set(INSTALL_SHORTCUTS "share/applications")     # /usr/[local]/share/applications
+     set(INSTALL_ICONS     "share/pixmaps")          # /usr/[local]/share/pixmaps
+     set(INSTALL_INIT      "/etc/init")              # /etc/init (upstart init scripts)
++elseif(MINGW AND SYSTEM_DIR_INSTALL)
++    set(INSTALL_ROOT      "lib/cmake")
++    set(INSTALL_CMAKE     "lib/cmake/${project}")
++    set(INSTALL_DATA      "share/${project}")
++    set(INSTALL_BIN       "bin")
++    set(INSTALL_SHARED    "lib")
```
Alexpux: shared need to be installed in "bin" I think

nazar554: It is installed like this: 
```
# Library
install(TARGETS ${target}
    EXPORT  "${target}-export"            COMPONENT dev
    RUNTIME DESTINATION ${INSTALL_BIN}    COMPONENT runtime
    LIBRARY DESTINATION ${INSTALL_SHARED} COMPONENT runtime
    ARCHIVE DESTINATION ${INSTALL_LIB}    COMPONENT dev
)
```
According to CMake [doc](https://cmake.org/cmake/help/v3.0/command/install.html):

>  For DLL platforms the DLL part of a shared library is treated as a RUNTIME target and the corresponding import library is treated as an ARCHIVE target. All Windows-based systems including Cygwin are DLL platforms.

So `LIBRARY` is not even used.

Alexpux: I don't understand what you mean here.

nazar554: I mean its ok as is, `.dll` goes to `RUNTIME DESTINATION ${INSTALL_BIN}` and `.dll.a` to `ARCHIVE DESTINATION ${INSTALL_LIB}    COMPONENT dev`. If I was building a static library it would go to `LIBRARY DESTINATION ${INSTALL_SHARED}`.

PR #1979, opened 2016-11-26
Updated GTK2 to 2.24.31

rimas-kudelis:
Also added a few Windows-specific changes which didn't make it into this
release, but were committed later. These make windows dead keys work as
expected in GTK2 applications.


mingw-w64-gtk2/PKGBUILD
```diff
@@ -17,7 +17,12 @@ license=(LGPL2.1+)
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-gobject-introspection"
              "${MINGW_PACKAGE_PREFIX}-pkg-config"
-             "gtk-doc")
+             "gtk-doc"
+             "autoconf"
```
Alexpux: This is common dependencies for all packages. Remove the changes

rimas-kudelis: I can remove them, but just so you know – they aren't really obvious. At least as a Debian user, I would expect the packaging system to fetch everything required for building, when I ask it to, as opposed to fetching everything minus some defaults I should somehow magically be aware of.

Alexpux: Yeah remove them. In our wiki we wrote that for develop packages you need install ````base-devel```` group of packages.

rimas-kudelis: Done. Although I'd still prefer all packages (not just this one) to list their dependencies explicitly instead of expecting users to read your wiki. Why would that be a bad thing to do?

mati865: I think `makepkg-mingw` warns when `base-devel` and `mingw-w64-toolchain` are not installed.
This is Arch Linux approach and MSYS2 uses their packaging system.

mati865:
Unfortunately it broke Win XP support as `wcscpy_s` isn't available there.

PR #1980, opened 2016-11-26
Add styrene-git

achadwick:
Styrene is a developer tool for bundling up combinations of local native builds and binary MSYS packages into convenient app bundles shaped for regular Windows users. "All" developers need to do is maintain their normal .desktop files and write a config file for their app. The Styrene tool takes care of all the repackaging and installer-making details.

https://github.com/achadwick/styrene

Would it be OK to incorporate this relatively early draft of the tool into MINGW-packages in -git form? Stable PKGBUILD to follow when this thing is actually stable.

As discussed in Alexpux/MSYS-packages#759, the code belongs in MINGW-packages.

### DIFF
--- a/styrene-git/PKGBUILD
+++ b/styrene-git/PKGBUILD
@@ -1,0 +1,55 @@
+# Maintainer: Andrew Chadwick <a.t.chadwick@gmail.com>
+
+_realname=styrene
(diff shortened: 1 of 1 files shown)


mingwandroid:
Sure thing, thanks Andrew.

achadwick:
D'oh: #1981.

PR #1981, opened 2016-11-26
styrene-git: rename, use the normal convention

achadwick:
That's what I get for starting this PKGBUILD off in the other repo :flushed: :sob:


mingwandroid:
No problem!

achadwick:
Thanks for that! Sorry about the mistake. I think I managed to fix all the other obvious porting issues except for most obvious one :neutral_face:

mingwandroid:
I bl sometimes don't check -git as I assume took defer to upstream/the interested developer! Thanks.

PR #1992, opened 2016-11-30
mingw-w64-libarchive - 3.2.2 - Note it requires libxml-2.  Reported a…

JPeterMugaas:
…s https://github.com/Alexpux/MINGW-packages/issues/1991 .


Alexpux:
I think instead need to partially revert your commit https://github.com/Alexpux/MINGW-packages/commit/6f39a40fe20483d881c4b2fc4b15b8cd87487982 and add to configure ````--without-libxml2```` to use expat

PR #1994, opened 2016-11-30
python3-cairo-git: package(): fix argument order

achadwick:
Addresses the build failures in Alexpux/MINGW-packages#195.

### DIFF
--- a/mingw-w64-python3-cairo-git/PKGBUILD
+++ b/mingw-w64-python3-cairo-git/PKGBUILD
@@ -5,8 +5,8 @@ pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-python3-cairo-git"
 provides=("${MINGW_PACKAGE_PREFIX}-python3-cairo")
 conflicts=("${MINGW_PACKAGE_PREFIX}-python3-cairo")
-pkgver=1.10.1.36.c0fc2b9
-pkgrel=1
+pkgver=1.10.1.36.8910a35
+pkgrel=2
 pkgdesc="Python 3 bindings for the cairo graphics library (mingw-w64)"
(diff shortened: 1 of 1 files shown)


Alexpux:
increase pkgrel as no more commits in git since last build

achadwick:
Done. Although the version number itself is randomized nicely already by those `git am` applications :smile:

Alexpux:
Next random version can be lower than current and we get package with downgrade

achadwick:
Yeah. That's not a good thing. I'll determine the hwm+2 on master and come up with a proper fix.

We should be using `git apply` in future so that applying them doesn't mess with revision counts or HEAD hashes.

mingwandroid:
The best approach is to use:

https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-gcc-git/PKGBUILD#L96-L98

.. but the version will be somewhat random if you were to remove one of our local patches and at the same time upstream were to add a single commit.

Then again, these are for developers *anyway* so I expect they are only ever installed with `pacman -U` so that shouldn't matter much?

achadwick:
@mingwandroid 
I'm not sure about them always being installed with `pacman -U`. Sometimes we recommend the git packages as fixes to regular users like in the linked bug above, and sometimes they're the the best available version to build for the repository.

My [styrene](https://github.com/achadwick/styrene) toy assumes sensible versioning behaviour even for local builds. It assumes developers who are serious about their builds will make collections of the local things they want to use.

Anyway, #1995.

mingwandroid:
Well, the standards are all a bit rubbish, I mean, what can you do? Python has PEP440 where the version/tag is used + the number of commits since, but that would suffer the same issue we've been talking about. You could force a custom integer into `pkgver()` before the revision count and hash I guess?

PR #1995, opened 2016-11-30
Fix use of "git am": reinstate sensible versioning

achadwick:
This package was using plain "git am" to apply its patches, meaning that revision counts were 2 revisions ahead of master, and that pkgver was different every time because the hash was different.

This commit adds a temporary versioning scheme fix which will allow this git package to get back on track without resorting to epochs. It can be reverted later when it is not needed.

### Addendum

We've been ahead of upstream's true revision count since the cc1e846e5a6b78a4a3afe0b5ba5fb53cefac56a5 rebuild, and my c28b47a19303214fba2eb17916929f1ef554df17 only compounded the error in #1994 by making revisions go backwards. Sorry.

I do not think we should be using "git am" to apply *any* local patches to git packages. It messes with the revision counts too much. I believe we should only be presenting upstream's hashes too, not our own local commits. Even if we apply with `git am --committer-date-is-author-date`, we will increment the revision count. If upstream has not changed, then ideally `pkgver` should not change!

If we add or remove or change a patch while upstream remains the same, we should be bumping `pkgrel` instead, and not changing `pkgver`.

```
$ vercmp 1.10.1.37.really.34.5a45b0b 1.10.1.36.8910a35
1
```

The "`.37.really`" hack is disgusting but quite readable. And Debian do it sometimes, so it must be OK :wink:

### DIFF
--- a/mingw-w64-python3-cairo-git/PKGBUILD
+++ b/mingw-w64-python3-cairo-git/PKGBUILD
@@ -5,8 +5,8 @@ pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-python3-cairo-git"
 provides=("${MINGW_PACKAGE_PREFIX}-python3-cairo")
 conflicts=("${MINGW_PACKAGE_PREFIX}-python3-cairo")
-pkgver=1.10.1.36.8910a35
-pkgrel=2
+pkgver=1.10.1.37.really.34.5a45b0b
+pkgrel=1
 pkgdesc="Python 3 bindings for the cairo graphics library (mingw-w64)"
 url="http://www.cairographics.org/pycairo"
 arch=('any')
@@ -26,13 +26,21 @@ pkgver() {
   ver=$(grep "$PREFIX" setup.py | sed "s/$PREFIX//" | sed "s/'//")
   revision=$(git rev-list --count HEAD)
   hash=$(git log --pretty=format:'%h' -n 1)
-  echo ${ver}.${revision}.${hash}
+  echo ${ver}.37.really.${revision}.${hash}
+  # FIXME: Remove the ".37.really" hack when it's not needed.
+  #  We accidentally pushed a 1.10.1.36 which reflected 2 local "git am" revs
+  #  when the real upstream $revision would have been 34.
+  #  Can't use epochs to correct this mistake: the -git package has to match
+  #  pash and future non-git packages.
 }
 
 prepare() {
   cd ${_realname}
-  git am ${srcdir}/0001-msys2-port-to-subprocess.check_output.patch
-  git am ${srcdir}/0003-fix-PyVarObject_HEAD_INIT-problem.patch
+  # NOTE: avoid "git am". It will create a new pkgver() each time!
+  #  The ${hash} above should correspond to the ${revision} number on
+  #  the pycairo master.
+  git apply -v ${srcdir}/0001-msys2-port-to-subprocess.check_output.patch
+  git apply -v ${srcdir}/0003-fix-PyVarObject_HEAD_INIT-problem.patch
 }
 
 build() {


mingwandroid:
It's been discussed before and I disagree strongly that our commits should not appear in the commit count nor that our commits should not appear in the hash. Those things are sacred to me and not incrementing/changing them makes it appear that the source is exactly the same as upstreams when it clearly isn't and that's a much worse wrong.

For most things, upstream will hopefully be a lot more active than we are, and my opinion is the hacks are the lesser or two evils.

achadwick:
I remain in disagreement. What you're describing is *exactly* what the `pkgrel` field is for despite it getting reset.

Arch's wiki explains the ethos for [pkgver](https://wiki.archlinux.org/index.php/PKGBUILD#pkgver) lucidly: 

> ### pkgver
> The version of the package. This should be the same as the version released by the author of the package. [...]

So for the non-git case, it's not something we can make up. Not a thing with a special count of our special local patches. It _should be the same as the author's version_. The only exceptions are for comparison semantics and syntactic restrictions.

For the git case. Yes, that will involve a carefully thought out but mechanistically applied addendum. I do not think that our work should be reflected there.

As for pkgrel:

> As fixes and additional features are added to the PKGBUILD that influence the resulting package, the pkgrel should be incremented by 1. When a new version of the software is released, this value must be reset to 1. [...]

I believe that presenting upstream's versions, counts, and hashes in a `pkgver` plus local mechanistic amendments (as strictly required for correct versioning) is the right way to express the author's [moral rights](https://en.wikipedia.org/wiki/Moral_rights). Being too inventive with them or blurring the lines is akin to passing off our amendments as theirs.

(Feel free to chastise me for going from an should to a must, but to me the line that you're blurring is that little `-`. It is a clear demarcation between their bit and our bit, and it... should... be preserved.)

As for our own work, even a suffixed `-1` from `pkgrel` suffices for all patches, because we publish source and maintain histories. It is unwholesome for maintainers to be precious about our local fixes. They are not sacred. They would not exist without upstream. Find a way to contribute them back.

That resetting ought to be considered a kick in the general direction of avoiding fragmentation.

Alexpux:
I'm not accept this.

mingwandroid:
Primarily `-git` packages exist not to be packaged and distributed except in rare cases where no good upstream release or versioning strategy is in place. This should be kept in mind when talking about moral rights and passing off (for outside of binary packaging the concepts don't exist).

I still take issue with the claimed violation of moral rights: The sha1 sum *suffix* is not a creative work, it is a machine generated checksum. As a pkgver prefix we take the tag when sensible tags exist and should otherwise rummage around the source for a good value for that prefix. If you know of PKGBUILDs that don't do enough in that regard then we need to improve them.

I stand by my connection that in the rate cases where we do distribute `-git` packages distributing a binary with a hash indicating an upstream revision but different contents is passing-off and dishonest. The end user isn't necessarily going to know the details of our local patches and development procedures nor should they need to. If they were to look at one of thet binary filenames and see the sha1 they might reasonably expect to checkout that sha1 and reproduce the build and/or audit the source code.

While I can see your POV, I'd like to speak to the reasons for using `git am`. My reason for wanting `-git` recipes is to use them is as the development ground for patches needed for MSYS2 packages. The idea is you can iterate your work in the `src` folder using  `git add`, `git rebase` etc and then regenerate the recipe's patches (and more importantly the non-`-git`, packages version's patches) using `git format-patch`.

This in a much more efficient workflow than cloning in GitHub, making a local version of the PKGBUILD, modifying, pushing, rebasing on top of upstream etc, etc and therefore much more likely to work. Crucially, using `git am` / `git format-patch` ensures that the atomicity, commit messages and authorship of the local patches is retained. Until we went this route I lost these things for quite a few patches and that was terrible in a real functional sense. It made me less efficient and my work less certain and it made it impossible to contribute the patches upstream since the reasoning behind them was lost. Upstreaming or differences is the end goal for any patch (though not often enough achieved by me), the end goal overall is to benefit open source.

You are free to maintain your own repository of `-git` packages and even make releases of them using whatever pkgver/pkgrel/patching strategy you want but here I have outlined in detail the reason we do it this way and why I want to keep it this way.

If you do find packages we are making releases from where upstream have a good packaging and versioning story in place then please make PRs to add non`-git` versions to make the binary releases from, that way people who agree with you regarding moral rights and/or disagree with me regarding passing-off should both find some satisfaction.

mingwandroid:
Is there any good reason not to build `python3-cairo` from non`-git`? It may require using `git am` to get the 36 odd patches out, but if upstream do not seem inclined to make a new release I think it's the best course of action.

PR #2015, opened 2016-12-07
mingw-w64-xapian-core - 1.4.1 - add epoch and include libtool.la file…

JPeterMugaas:
… for a config script.


Alexpux:
Instead of revert commits you need rebase

mati865:
Hmm I thing interactive rebase should be fine here:
```
#if you want easier editor export or add EDITOR=nano before command
git rebase -i @~13
```
Then replace `pick` with `d` for commits that you want to remove.

PR #2020, opened 2016-12-09
ducible: new package

oscarfv:


peterbud:
If it's a git build you should use the name of a directory `mingw-w64-ducible-git`. But as I see there is a [release version available](https://github.com/jasonwhite/ducible/releases/tag/v1.1.1). Why not to use the released version?

oscarfv:
AFAIK the `-git` suffix is used for git-based packages when there is a non-git-based package available. When you use git just for downloading sources from a stable project, it does not make sense to reflect that on the package's name.

`ducible` is a small utility. Those projects tend to slow down development until it stops, because the project's goal was achieved. On that context, it does not make much sense to differentiate releases from the git sources. OTOH, if you need a quick fix (for a new g++ version, for example) you either have to wait for a new release or just do what I did: submit a patch upstream and rebuild our package afterwards.

The released version was not used because it does not build with MinGW.

Alexpux:
@oscarfv, this is really "git" package so modify script and folder for it

oscarfv:
Done.

Nice to see you back, @Alexpux.

PR #2050, opened 2016-12-28
Update OpenOCD pkgbuild

RoccoMarco:
I have commented provides and conflicts which produce error during makepkg and I have added a dependency: libftdi

### DIFF
--- a/mingw-w64-openocd-git/PKGBUILD
+++ b/mingw-w64-openocd-git/PKGBUILD
@@ -12,8 +12,8 @@ pkgrel=1
 pkgdesc="OpenOCD - Open On-Chip Debugger (mingw-w64)"
 arch=('any')
 url="http://openocd.sourceforge.net"
+license=("GPLv2")
 options=('!ccache')
-# What about ftdi?
 depends=("${MINGW_PACKAGE_PREFIX}-hidapi"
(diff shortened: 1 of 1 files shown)


mati865:
> I have commented provides and conflicts which produce error during makepkg

`provides` and `conflicts` must be an array like this:
```
provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
```

You should also add license variable (check other packages for reference).

RoccoMarco:
Done, waiting for checks

paulfertser:
mati865, these provides/conflicts look like boilerplate, is it really needed at all? I've checked the recently added libftdi PKGBUILD and it doesn't have those fields defined at all.
Other than that, from the OpenOCD POV this patch looks good to me, and I'd appreciate msys2 providing 0.10.0-rc1 binaries once this patch is accepted. Is it done automatically nowadays or some procedure need to be performed to have the build triggered?

mati865:
@paulfertser All `-git` packages have those fields because installing non git and git version at same time would fail (`example.exe already exists in filesystem`). And if someone decided to use `--force` to install it and then decides to remove one of those packages another package would be broken because they use files with the same names.

paulfertser:
Mateusz, thank you for the clarifications and for your work! TBH, I can't understand why it's not every windows user running MSYS2!

PR #2070, opened 2017-01-08
clang: update to 3.9.1

mati865:
Hello,
I've been very busy with assignments lately but it's getting better.

Back to the Clang topic:
3.9.1 brings some fixes, libclang.a could build without patching CMake so it should work (somebody please test).
I've included proper fix to limits.h bug which I had around but forgot or had no time to upload.
Also there are 2 patches related to dynamic linking which should be threaten as proof-of-concept (I originally made them to get shared 3.9.0 but it was very broken). This week I'll try to use upstream patches which should fix lld linking (replacing my hackish 0302).

Cheers


mati865:
Shared build should be fine by now but I haven't test it yet so for now there are shared libs build but executables  are linked statically (`-DLLVM_LINK_LLVM_DYLIB=OFF`).
In the following days I'll try to test and eventually enable it.

Issue #1952, opened 2016-11-18
Readline 7 buggy with GDB 7.12-2, echoes extra characters.


dimztimz:
With my pull request #1869 to add TUI in GDB, I modified the PKGBUILD script to build GDB with the system Readline because that was necessary for the build to work. TUI + GDB's readline was completely broken. So using the system readline was the only option. Back then, the system readline was at version 6.3 and everything was working fine.

After the update to readline 7 #1872 , things are slightly broken when gdb is running inside Windows Console (or with winpty).

I need to write `command + Enter + Random key` so the command gets regsitered. After the first enter the command is echoed again.
Also, Ctrl+key combinations are echoed, and they take effect after I press Enter twice.

## To reproduce

1. Install `pacman -S mingw-w64-x86_64-gdb`
2. Open mingw64 shell
3. Run `start gdb`. Will open gdb in windows console.
4. Write anything and press enter. Try Ctrl+C A and press Enter.

## Picture
![gdbbug](https://cloud.githubusercontent.com/assets/6236568/20415052/21d98702-ad36-11e6-974b-46ea91daa142.PNG)

mati865:
Picture is wrong, it should echo `No symbol table loaded.  Use the "file" command.` once.

dimztimz:
The issue is not about the way I should use gdb. I know that. It's about the library readline which handles the interactive prompt. The interactive prompt is messed up with readline 7.
For example, pressing Ctrl+X, A should not echo anything but it does.

dthpham:
I've encountered this myself after upgrading gdb from 7.12-1 to 7.12-2. This bug is actually what prompted me to try out lldb, but that doesn't work either #1941.

mati865:
@dthpham About lldb: it's upstream bug, and I cannot fix it. You have to use unofficial server for lldb.

dimztimz:
The patch https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commit;h=7f3c5ec870943f7f32c946ff9459dfd04fcb8e07 + reverting to GDB's readline seems to fix this bug. TUI works too. I will send a pull request.

genusistimelord:
Your fixed also fixed some issues a few of us had with IDE debugging in the 32bit GDB.

segevfiner:
@dimztimz I'm encountering this issue with mingw-w64-x86_64-gdb 8.1.1-1.

liudonghua123:
I have the same problem.

[...]

PR #2105, opened 2017-01-19
Fixes buggy readline 7 in GDB

dimztimz:
Fixes issue #1952

We revert to using Readline from GDB
We apply a patch from GDB master related to Readline and Mingw
TUI works too.


Alexpux:
Thanks!

PR #2108, opened 2017-01-21
clang: fix llvm-config shared lib detection

mati865:
Fixes https://github.com/Alexpux/MINGW-packages/issues/2107


Alexpux:
What about other dll's - LTO, clang?

mati865:
>// If LLVM_LINK_DYLIB is ON, the single shared library will be returned
  // for "--libs", etc, if they exist. This behaviour can be overridden with
  // --link-static or --link-shared.

Current output is:
```
$ llvm-config --libs --link-shared
-lLLVM.dll
```

The only minor problem is it returns static libs by default so one have to pass `--link-shared`. I'm investigating it.

mati865:
Ok, I know why why it's static by default: LLVM is still linked statically https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-clang/PKGBUILD#L287

PR #2119, opened 2017-01-23
Change ghc to use LLVM 3.9

ncihnegn:


mati865:
Fix patch name in https://github.com/Alexpux/MINGW-packages/pull/2119/files#diff-804ebf74ce827c7336cf4defd5417cdfL62

mati865:
```
patching file configure.ac
Hunk #1 FAILED at 521.
1 out of 1 hunk FAILED -- saving rejects to file configure.ac.rej
```

PR #2132, opened 2017-01-30
emmenlau grpc 1.1.0-pre1 update

emmenlau:
Updated grpc to 1.1.0-pre1


emmenlau:
Dear @mkae , the underscore version `_pkgver` uses the version number with a dash `1.1.0-pre1` which is not allowed in `pkgver`. Its not super nice but I think many packages have this. For 1.1.0 (which is due soon) I will get rid of this again.

mkae:
Ah, okay, that explains it. Thanks for responding so quickly. :-)

PR #2139, opened 2017-01-31
mingw-w64-gcc/PKGBUILD: removed copy of several missing source files

emmenlau:
Fixes the build of gcc 6.3.0


Alexpux:
Why you comment this stuff?

emmenlau:
For me the default build breaks, with error: no such file or directory share/info/cpp.info*. And when I check in more detail, then is are more missing stuff in share/info so I commented all missing things.
Does the build work for you? Then there must be something different with my environment. Maybe I'm missing some optional dependency?

emmenlau:
Closing outdated PR.

PR #2150, opened 2017-02-03
mingw-w64-mesa - 3.0.4 - Update to latest version

JPeterMugaas:
mingw-w64-readline - 7.0 - patchlevel 3
mingw-w64-libcerf - 1.5 - new package
mingw-w64-libsigsegv - git version - new package

### DIFF
--- a/mingw-w64-libcerf/PKGBUILD
+++ b/mingw-w64-libcerf/PKGBUILD
@@ -1,0 +1,45 @@
+# Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
+
+_realname=libcerf
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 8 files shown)


Alexpux:
First, import realine patche into repository.
Second, why not just add patch for Win64 for libsigsegv? In any case I will not build gitr version.

JPeterMugaas:
I have added the patches for GNU readline.

The problem with http://git.savannah.gnu.org/cgit/libsigsegv.git/commit/?id=96812beb65b75fdc374185670fbe1b0280c31b8b is that I can't get it to patch code from the download because something changed too much.  I tried hand rolling patch for the original version 2.10 but I think I botched it and that version is over 7 years old.

PR #2161, opened 2017-02-12
go: remove unneeded workaround [DO NOT MERGE]

aroig:
I seem to be able to build go just fine without this workaround that opens a new windows terminal window for building go. This interferes with automated builds, etc. Let's remove it.

@mingwandroid Can you please check if it works for you?

### DIFF
--- a/mingw-w64-go/PKGBUILD
+++ b/mingw-w64-go/PKGBUILD
@@ -40,13 +40,7 @@ build() {
   export GO_CFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
   export CFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
   export GO_BUILD_VERBOSE=1
-  # I have no idea why we need this to be run from a completely seperate
-  # MSDOS shell, but it does. Without this, the build exits with no info
-  # about what went wrong.
-  # ./make.bat
-  echo "cd $(cygpath -w $PWD)"                    > /tmp/bootstrap_go.bat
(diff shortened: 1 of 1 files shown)


hcorion:
You could also update the go version from 1.7.4 to 1.7.5, which is the current release.

hcorion:
Go 1.8 was released. I doubt it fixes anything but you could try.

Alexpux:
I'm not be able to build GO without hack still.

aroig:
@Alexpux @mingwandroid Ok, that's odd. Since I can build it fine without the hack, maybe we can get to the bottom of this by looking at what is different in our environments. I can't do it alone since I do not know how to break it.

I'm on a windows 10 VM after the anniversary update. [Here](https://gist.github.com/aroig/c355d6c50e1717aa8b9477879e05cc5e#file-msys2-environment-txt) are the environment variables at the time of build, seen from within the build() function in the PKGBUILD. Do you see any relevant difference?

@hcorion Thanks, I'll test-build and submit a new PR with go 1.8 later.

Alexpux:
I'm already built go 1.8. Will be in repo today.
About hack: I have installed nearly all packages that present in MSYS2. So trouble can be with one of the programs. It is hard to understand what

mati865:
Possibly mingw gettext issue?

Alexpux:
I think something with terminal output.

mingwandroid:
When I implemented the hack, I was trying my build in a brand new msys2 installation. I think this is some low level corruption of the msys2 process space.

PR #2162, opened 2017-02-12
git-lfs: bump version + some cleaning

aroig:
- Fetch go dependencies in prepare() instead of build.

- Use actual git sources for git-lfs instead of fetching them with go get.


aroig:
By the way, CI fails due to ancient version of go in the actual msys2 binary distribution.

mfld-fr:
Hello, I did manually what the script does, and it succeed to build on current MSYS2.

Go version : 
$ pacman -Q mingw-w64-x86_64-go
mingw-w64-x86_64-go 1.7.4-1

What are the missing steps to make this valuable package available for all ?

mfld-fr:
Old topic, a newer pull request about this package is here: https://github.com/Alexpux/MINGW-packages/pull/2406

PR #2171, opened 2017-02-15
librsvg: update to 2.41.0

qme:
This is a test PR.


qme:
Since version 2.41.0 the librsvg package requires rust. It seams that cargo is not installed properly even if rust is listed as makedepends:

configure: error: cargo is required.  Please install the Rust toolchain from https://www.rust-lang.org/

I've tried to avoid this problem with different commits without success.

mati865:
Upload configure logs. Cargo seems to be fine (I have just installed racer package using it).

PR #2181, opened 2017-02-18
Mitigate gdb bug 21057.

lhmouse:
See https://sourceware.org/bugzilla/show_bug.cgi?id=21057.


lhmouse:
Confirmed that it has fixed the undefined reference error.

PR #2187, opened 2017-02-19
mesa: update to version 17.0.0

crvv:
Besides, the header files are necessary for using osmesa. 
I wonder why they were deleted in a070cca548397fa11feb7dd4485aafdafaddfe40.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=13.0.4
+pkgver=17.0.0
 pkgrel=1
 pkgdesc="Open-source implementation of the OpenGL specification (mingw-w64)"
 arch=('any')
@@ -17,7 +17,7 @@ license=('MIT')
 options=('staticlibs' 'strip')
 source=(https://mesa.freedesktop.org/archive/${pkgver}/${_realname}-${pkgver}.tar.xz{,.sig}
         001-extra-libs.patch)
-sha256sums=('a95d7ce8f7bd5f88585e4be3144a341236d8c0fc91f6feaec59bb8ba3120e726'
+sha256sums=('39db3d59700159add7f977307d12a7dfe016363e760ad82280ac4168ea668481'
             'SKIP'
             'bc9bb5013ac80ded47ad164ae1ef58cc9a39784eb4bf61e8c7d654bb273b05a9')
 validpgpkeys=('8703B6700E7EE06D7A39B8D6EDAE37B02CEB490D')
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -16,10 +16,12 @@ url="http://www.mesa3d.org"
 license=('MIT')
 options=('staticlibs' 'strip')
 source=(https://mesa.freedesktop.org/archive/${pkgver}/${_realname}-${pkgver}.tar.xz{,.sig}
-        001-extra-libs.patch)
+        001-extra-libs.patch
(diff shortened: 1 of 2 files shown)


mingw-w64-mesa/PKGBUILD
```diff
@@ -58,6 +58,6 @@ package() {
   cp -f ${srcdir}/${_realname}-${pkgver}/build/windows-${_mach}/gallium/targets/osmesa/osmesa.dll ${pkgdir}${MINGW_PREFIX}/bin/
   cp -f ${srcdir}/${_realname}-${pkgver}/build/windows-${_mach}/gallium/targets/osmesa/libosmesa.a ${pkgdir}${MINGW_PREFIX}/lib/
 
-#  cp -f cd ${srcdir}/${_realname}-${pkgver}/include/GL/gl.h ${pkgdir}${MINGW_PREFIX}/include/GL/
-#  cp -f cd ${srcdir}/${_realname}-${pkgver}/include/GL/osmesa.h ${pkgdir}${MINGW_PREFIX}/include/GL/
+  cp -f ${srcdir}/${_realname}-${pkgver}/include/GL/gl.h ${pkgdir}${MINGW_PREFIX}/include/GL/
```
Alexpux: It is not copied because it conflict with system gl.h

crvv: On my computer, system gl.h is at C:\Program Files (x86)\Windows Kits\8.1\Include\um\gl\gl.h.
Other gl.h from MinGW is at /mingw64/x86_64-w64-mingw32/include/GL/gl.h. It belongs to package mingw-w64-x86_64-headers-git.
This gl.h is at /mingw64/include/GL/gl.h. They are at different position, and shouldn't conflict.

Alexpux: They conflict because they both in GL subfolder and both /mingw64/include and /mingw64/x86_64-w64-mingw32 are in default search path. So we we have troubles if /mingw64/include will be in first place when we build something

crvv: Mesa is an implementation of OpenGL, and it depends on its own header file gl.h.
With the gl.h from headers-git, I have troubles when I build something which use mesa.
(Just `echo "#include <GL/osmesa.h>" > test.c && gcc test.c`)
So without its own header files, this package doesn't work at all.

I also tried using gl.h from mesa to build something which doesn't use mesa, and it succeeded.
In Arch Linux, the /usr/include/GL/gl.h is also owned by mesa.
So I think the header files should be included.

Alexpux: As compromise you can install header to /mingw64/include/mesa/GL for example

mati865: `<GL/gl.h>` is provided by mingw-w64 headers:
```
$ pkgfile /mingw64/x86_64-w64-mingw32/include/GL/gl.h
mingw64/mingw-w64-x86_64-headers-git
```
GCC will look into `/mingw64/include` before `/mingw64/x86_64-w64-mingw32/include/`
but Clang will look into `/mingw64/x86_64-w64-mingw32/include/` before `/mingw64/include`.
We cannot keep them both at same time because of compatibility issues.

> In Arch Linux, the /usr/include/GL/gl.h is also owned by mesa.
glibc doesn't provide `GL/gl.h` so there is no conflict, mingw-w64 does.

`GL/osmesa.h` on the second hand could be provided by mesa package if it can work with mingw-w64 `GL/gl.h`.
If not mingw-w64 devs should be contacted.

crvv: OK, I have put them in /mingw64/include/mesa/GL.

mingw-w64-mesa/osmesa.pc
```diff
@@ -0,0 +1,11 @@
+prefix=/mingw64
```
mati865: What about 32 bit version?

PR #2194, opened 2017-02-21
Add "mingw-w64-libvisio" and update dependencies and license for "mingw-w64-libcdr"

Ede123:
First package, so let me know if I should change something...


mingw-w64-libvisio/PKGBUILD
```diff
@@ -0,0 +1,52 @@
+_realname=libvisio
```
Alexpux: Add yourself as maintainer at the top of PKGBUILD

Ede123: done

PR #2197, opened 2017-02-21
mingw-w64-pkg-config/PKGBUILD: add dependency on glib2

emmenlau:
Fixes build problem if glib2 is not installed.


Alexpux:
pkg-cionfig building with its own bundled GLIB. See configure options.

PR #2199, opened 2017-02-21
Python requests v2.13.0

alethea:
Add a build for `mingw-w64-python-requests` since Sphinx depends on it.


mingw-w64-python-requests/PKGBUILD
```diff
@@ -0,0 +1,70 @@
+# Contributor: Alethea Rose <alethea@alethearose.com>
+
+_pyname=requests
+_realname=requests
```
Alexpux: Remove _pyname and use _realname instead

alethea: Done! Thanks for the quick review.

PR #2202, opened 2017-02-22
binutils: backport patch for Alexpux/MSYS2-packages#454

mati865:
Possibly better solution for this issue.
Original commit https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commitdiff;h=ec8f76882145c71bef81a9cadf0bf51ff9fa5b35

Also fixes https://github.com/Alexpux/MINGW-packages/issues/2190


mati865:
@Alexpux binutils 2.28 were released. Should we stay with 2.27 or go forward?

PR #2204, opened 2017-02-24
Update GHC to 8.0.2, pull in upstream Arch scripts

alethea:
This should get it building again, and the upstream scripts should make it possible to handle Haskell libraries for distributing the rest of the Haskell platform.


mingw-w64-ghc/PKGBUILD
```diff
@@ -17,89 +19,174 @@ depends=("${MINGW_PACKAGE_PREFIX}-binutils"
...
+  cd "$srcdir/$CARCH/ghc-$pkgver"
+  patch -Np1 -i "$srcdir/0001-SysTools-Revert-linker-flags-change.patch"
+  patch -Np1 -i "$srcdir/0002-Switch-to-LLVM-version-3.9.patch"
+  patch -Np1 -i "$srcdir/0003-Use-system-build-tools.patch"
+  patch -Np1 -i "$srcdir/0004-Handle-CRLF-properly-for-C-dependencies.patch"
+  patch -Np1 -i "$srcdir/0005-Add-MinGW-_lock_file-symbol.patch"
+  patch -Np1 -i "$srcdir/0006-Add-MinGW-_unlock_file-symbol.patch"
+
+  ./boot
+  ./configure \
```
mati865: Move `./configure` to build().

mingwandroid:
@Mistuke, can I request a review of this please? Thanks in advance.

Mistuke:
@mingwandroid sure, I'll take a look tonight.

mingw-w64-ghc/PKGBUILD
```diff
@@ -17,89 +19,174 @@ depends=("${MINGW_PACKAGE_PREFIX}-binutils"
...
+fi
+_hooks=('ghc-register.hook' 'ghc-unregister.hook' 'ghc-rebuild-doc-index.hook')
+
 source=(
-  "https://downloads.haskell.org/~ghc/$pkgver/ghc-$pkgver-src.tar.xz"
-  "https://downloads.haskell.org/~ghc/$pkgver/ghc-$pkgver-testsuite.tar.xz"
+  "https://downloads.haskell.org/~ghc/$pkgver/ghc-$pkgver-src.tar.xz"{,.sig}
+  "https://downloads.haskell.org/~ghc/$pkgver/ghc-$pkgver-testsuite.tar.xz"{,.sig}
+  "https://downloads.haskell.org/~ghc/$pkgver/ghc-$pkgver-$_hsarch-unknown-mingw32.tar.xz"{,.sig}
   'build.mk'
```
Mistuke: what's in the build.mk?

mingw-w64-ghc/PKGBUILD
```diff
@@ -17,89 +19,174 @@ depends=("${MINGW_PACKAGE_PREFIX}-binutils"
...
-  if [[ $SYST == "i686" ]]; then
-     SYST="i386"
-  fi
-  mkdir -p host
-  curl -LO "https://www.haskell.org/cabal/release/cabal-install-1.24.0.0/cabal-install-1.24.0.0-$SYST-unknown-mingw32.zip"
-  curl -L "https://downloads.haskell.org/~${_realname}/$pkgver/${_realname}-$pkgver-$SYST-unknown-mingw32.tar.xz" | tar -Jx --strip-components=1 --directory=host
-  unzip -o cabal-install-1.24.0.0-$SYST-unknown-mingw32.zip -d host/bin
-  export PATH=$srcdir/host/bin:$PATH
-  cabal update
-  cabal install alex happy --prefix=$srcdir/host
```
Mistuke: `happy` and `alex` are required. I don't see them being provided?

mingw-w64-ghc/PKGBUILD
alethea: It's configuration settings for a release build of the GHC

mingw-w64-ghc/PKGBUILD
alethea: That's a good question. It builds without them, so I assumed they're in the source tree and the host compiler builds them, but I'll check, and add the binaries if necessary

mingw-w64-ghc/PKGBUILD
```diff
@@ -17,89 +19,174 @@ depends=("${MINGW_PACKAGE_PREFIX}-binutils"
...
+          "${MINGW_PACKAGE_PREFIX}-haskell-template-haskell"
+          "${MINGW_PACKAGE_PREFIX}-haskell-terminfo"
+          "${MINGW_PACKAGE_PREFIX}-haskell-time"
+          "${MINGW_PACKAGE_PREFIX}-haskell-transformers"
+          "${MINGW_PACKAGE_PREFIX}-haskell-win32"
+          "${MINGW_PACKAGE_PREFIX}-haskell-xhtml"
+          "${MINGW_PACKAGE_PREFIX}-haskell-cabal")
+
+
+makedepends=("${MINGW_PACKAGE_PREFIX}-python3-sphinx"
```
Mistuke: Ah, perfect!

mingw-w64-ghc/PKGBUILD
mati865: @Mistuke: https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-ghc/build.mk

mingw-w64-ghc/PKGBUILD
Mistuke: No, the host compiler doesn't build them as we need them to be resolved at configure time. Likely they're being picked up from a different location or a previous install. It's also a long existing limitation of `Cabal` that the solver can't currently resolve `build-tools` only dependencies see https://github.com/haskell/cabal/issues/4217 and https://github.com/haskell/cabal/pull/4241

mingw-w64-ghc/PKGBUILD
Mistuke: Ah, right, it's re-using the one from the previous patch. Got it.

mingw-w64-ghc/PKGBUILD
alethea: How do we feel about using `cabal-install` to install them? I don't like that it adds a download step that doesn't run through the SHA sums checks, but I think that's the cleanest thing to do until I can get them packaged.

Mistuke: @alethea Ah no need. I see that the files are pre-processed in the source distributions. Which is why configure allows them to be missing. Fair enough, from the `GHC` point of view this looks good.

alethea:
Not sure why CI is failing, looks like something to do with an older GCC? It build okay on my machine.

mati865:
> Not sure why CI is failing

It's Wine bug, Tea is Linux running MSYS2 under Wine.

PR #2206, opened 2017-02-24
mingw-w64-curl: add name suffix to alternative packages

Radrik5:
*UPDATE: Only mingw-w64-curl/PKGBUILD was modified (see discussion below).*

New package provides cURL compiled with WinSSL. It will be used with git as an alternative option to the ca-cert file: https://github.com/git-for-windows/git/issues/301#issuecomment-281644692.

I copied all the patches and the PKGBUILD file from `mingw-w64-curl` to `mingw-w64-curl-winssl` and modified the PKGBUILD files so that they're still the same except the value of the `_variant` variable.

The same technique is used in `mingw-w64-qt5` and `mingw-w64-qt5-static` packages. If there is a better way to share sources and patches among different packages, I'm ready to do some rework of the changes.

`mingw-w64-curl-winssl` provides the same binary files as `mingw-w64-curl`, so they're marked as conflict. You cannot have both installed at the same time. The `provides` variable is set in addition to the `conflicts` variable just in case there will be `mingw-w64-curl-gnutls` that should conflict with both `mingw-w64-curl` and `mingw-w64-curl-winssl`. Details: https://wiki.archlinux.org/index.php/PKGBUILD#conflicts

### DIFF
--- a/mingw-w64-curl/PKGBUILD
+++ b/mingw-w64-curl/PKGBUILD
@@ -3,9 +3,15 @@ _variant=-openssl
 #_variant=-winssl
 #_variant=-gnutls
 
+if [ "${_variant}" = "-openssl" ]; then
+  _namesuff=
+else
+  _namesuff="${_variant}"
+fi
+
 _realname=curl
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}${_namesuff}"
 pkgver=7.53.1
 pkgrel=1
 pkgdesc="An URL retrival utility and library. (mingw-w64)"
@@ -28,6 +34,10 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
             "${MINGW_PACKAGE_PREFIX}-ca-certificates" \
             "${MINGW_PACKAGE_PREFIX}-gnutls")
          )
+if [ -n "${_namesuff}" ]; then
+  provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+  conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+fi
 options=('staticlibs')
 source=("${url}/download/${_realname}-${pkgver}.tar.bz2"{,.asc}
         "0001-Make-cURL-relocatable.patch"


Radrik5:
@dscho suggested to use a script that patches the PKGBUILD file to build mingw-w64-curl-winssl package in the mingw-w64-curl folder: https://github.com/git-for-windows/git/issues/301#issuecomment-282321452. This eliminates the need to support two identical cURL packages but the ci-build.sh will build mingw-w64-curl package only because it simply enumerates folders with the PKGBUILD file.

@Alexpux which option do you prefer:
1. Leave the two packages in the separate folders;
2. Add make-winssl-pkg.sh to mingw-w64-curl as @dscho suggested;
3. Add PKGBUILD and PKGBUILD.winssl that only set the _variant variable and import PKGBUILD.base with the common code. For me it's a more straightforward way than a script that patches the PKGBUILD file.

Alexpux:
You can add to current PKGBUILD switches for different variants of building like you already have done. When building package you will uncomment proper variant that you need.

Radrik5:
As we need both versions in the Git installer, manual switching of the _variant variable is tedious and error-prone: when new cURL version is released you need to (1) build version with openssl, (2) wait until build is finished, (3) modify the PKGBUILD file and build version with winssl, (4) wait until build is finished, (5) upload new packages to the pacman repository. If you forget to build one of the versions, the installer will be shipped with a version that may contain bug or security hole. So it's important to have some automation here.

Alexpux:
@Radrik5, my repository is not vendor specific so I will not accept changes like you want. Specific changes for git-for-windows you can add to your own repository.

Radrik5:
@Alexpux, will you accept changes to the PKGBUILD file that add package name suffix and the conflicts array depending on the _variant variable (like in this PR)? Then I'll need a script that modifies the _variant variable only.

Alexpux:
@Radrik5 why not then during your build replace one text with another with sed for example?

Radrik5:
A sed script that modifies the _variant variable only (s/^_variant=.*/_variant=-winssl/) is simpler and less likely to break in the future than a sed script that modifies several variables and adds new ones after some line. For example, adding `conflicts` variable after a line with `makedepends` will be broken when `makedepends` changes to multi-line array like `depends`.

Alexpux:
@Radrik5 with sed you need modify only first occurence in PKGBUILD. Read something like 
http://stackoverflow.com/questions/148451/how-to-use-sed-to-replace-only-the-first-occurrence-in-a-file

Radrik5:
@Alexpux, yes, it will be more reliable. Will you accept changes to the PKGBUILD file so that I modified only _variant with sed? If yes, should I create separate pull request or force push to this one?

Alexpux:
You can force push here

dscho:
> why not then during your build replace one text with another with sed for example?

@Alexpux you misunderstood: that is what @Radrik5 wants to do, but *only* with the commented-out `_variant` line.

However, this is currently not enough. If you only do that, you will end up with a different, incompatible `mingw-w64-x86_64-curl-<version>.pkg.tar.xz` file. Nothing in that file suggests that it is the `-winssl` variant. That is simply unacceptable.

Hence @Radrik5's question whether you would accept changes in the `PKGBUILD` script that would make it possible to simply remove the comment in front of `#_variant=-winssl` and have the resulting package be named `mingw-w64-x86_64-curl-winssl-<version>.pkg.tar.xz` (note the `-winssl` in there) and have that package declare a conflict with `mingw-w64-x86_64-curl` so that you cannot install both at the same time.

I thought that @Radrik5's idea was rather sensible, and I think your terse reply may either reflect that you did not understand him at all, or that you are ignoring that splendid idea.

And maybe it would be better to close this Pull Request, then, because the new Pull request would not *Add mingw-w64-xul-winssl as alternative...*. It would do something pretty different. So a force-push would not be appropriate.

Alexpux:
@dscho too many words, seems we don't understand each other. What problem with last variant is pushed in pull?

Radrik5:
I've tested the PKGBUILD file with -openssl and -winssl variants, `makepkg-mingw` creates the following files:
```
mingw-w64-x86_64-curl-7.53.1-1-any.pkg.tar.xz
mingw-w64-x86_64-curl-winssl-7.53.1-1-any.pkg.tar.xz
```

`pacman` refuses to install mingw-w64-curl-winssl along with mingw-w64-curl:
```
$ pacman -U mingw-w64-x86_64-curl-winssl-7.53.1-1-any.pkg.tar.xz
loading packages...
resolving dependencies...
:: mingw-w64-x86_64-curl-winssl and mingw-w64-x86_64-curl are in conflict. Remove mingw-w64-x86_64-curl? [y/N] y

Packages (2) mingw-w64-x86_64-curl-7.53.1-1 [removal]
             mingw-w64-x86_64-curl-winssl-7.53.1-1

Total Installed Size:   2.11 MiB
Net Upgrade Size:      -0.07 MiB

:: Proceed with installation? [Y/n] Y
(1/1) checking keys in keyring                     [#####################] 100%
(1/1) checking package integrity                   [#####################] 100%
(1/1) loading package files                        [#####################] 100%
(1/1) checking for file conflicts                  [#####################] 100%
(2/2) checking available disk space                [#####################] 100%
(1/1) removing mingw-w64-x86_64-curl               [#####################] 100%
(1/1) installing mingw-w64-x86_64-curl-winssl      [#####################] 100%
```

I will rename the pull request to better illustrate what was done.

PR #2208, opened 2017-02-24
minor layout and coding style fixes for several PKGCONFIG files

emmenlau:
During implementing a continuous integration I found several minor inconsistencies, for example sometimes case differences between "depends" and "provides". This PR fixes the inconsistencies and should have no negative side effects.


mingw-w64-python-pywin32/PKGBUILD
```diff
@@ -8,7 +8,11 @@ license=('PSF')
 arch=('any')
 pkgver=219
 pkgrel=1
-makedepends=(${MINGW_PACKAGE_PREFIX}-python2 ${MINGW_PACKAGE_PREFIX}-python3 rsync)
```
emmenlau: Ok thats good, should I remove it or just remove it from the PR?

Alexpux: Remove from PR

mingw-w64-wxwidgets/PKGBUILD
```diff
@@ -5,6 +5,7 @@
 
 _realname=wxWidgets
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgbase=mingw-w64-${_realname}
```
emmenlau: Its not required, I can remove it if you find it not good. My reason was that the package name is wxWidgets (camel case) and the directory name is wxwidgets (lowecase), so I wanted to make it explicit. But I can also remove it.

Alexpux: If you want to do it you must replace yours _realname with wxwidgets. In other case you provide itself

emmenlau:
You are right, I removed the "provides" from wxwidgets.

PR #2209, opened 2017-02-25
Inkscape: update dependencies

Ede123:
Inkscape 0.92 has some new dependencies.

We now have libvisio (#2194) and already had potrace for some time.

### DIFF
--- a/mingw-w64-inkscape/PKGBUILD
+++ b/mingw-w64-inkscape/PKGBUILD
@@ -12,7 +12,7 @@ _realname=inkscape
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.92.1
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


Alexpux:
Increase pkgrel

Ede123:
> Increase pkgrel
done.

Should I always increase `pkgrel`?
I didn't do it before because I assume Inkscape was already built with those dependencies (they were just not yet listed) and therefore a rebuilt would not be necessary.

Alexpux:
If dependencies was not listed then after install you will not get all dependencies and users will report about it. So you need to increase pkgrel in cases when package need to be rebuild and reset pkgrel to 1 when new version is out.

Alexpux:
Don't yet see increased pkgrel

Ede123:
> Don't yet see increased pkgrel

Sorry, forgot to stage the change 🤦

mati865:
> Should I always increase pkgrel?

If you make changes that are important (like dependency changes, fixes, rebuilds, etc.) you have to bump pkgrel.
If your changes don't really affect compiled package (like coding style) you don't bump it.

PR #2212, opened 2017-02-26
Add SOCI package from git repo (most up to date)

dsogari:


Alexpux:
Are we really need this "git" package? As I'm tell earlier I will not build git packages, they only will be as scripts here

dsogari:
That's ok to discard this package. I added it because I need the latest functionality implemented by the library authors. I also made it link against `mariadbclient` to have a LPGL license for use with MySQL.

Since I have a [personal repo](https://github.com/dsogari/msys2-repos) for packages that are not present in the original MINGW-Packages, you can close this PR without merging. :)

PR #2219, opened 2017-02-26
gcc-git: generate _realpkgver from given branch

mati865:
`tar` is required by `libtool` which is part of `base-devel` group.


mati865:
Maybe moving this part to the top would be cleaner?

PR #2220, opened 2017-02-27
Bump IPython & related modules versions

oroppas:


Alexpux:
You must update checksums too, not only bump versions.

oroppas:
My bad. Should be fixed.

PR #2222, opened 2017-02-27
minor fixes for dependencies

emmenlau:
- libtorrent-rasterbar does not have a git suffix, fixed
 - webkitgtk2 referenced explicit version 0.10 of gstreamer and gst-plugins-base, but this version is not provided by any package? Replaced with head version as dependency.


Alexpux:
Webkitgtk2 need to be removed

emmenlau:
Ok crap I remove it, sorry

emmenlau:
Removed.

PR #2224, opened 2017-02-27
Update glog to latest git commit (ab65454)

dsogari:
Important!!
The recompilation also fixes a run-time error that occurs after the update of gflags.


Alexpux:
Rename package to "-git" and write pkgver function
Also add conflicts, provides and replaces

dsogari:
Great idea! I'll update it..

Alexpux:
I will not accept moving to unstable sources from git.

dsogari:
That's ok. But still, you must recompile this package (even at its current version) as it was giving run-time errors because of the update of gflags.

Alexpux:
Ok.

dsogari:
I'm sorry to bother you @Alexpux. I appreciate that you followed my request on this package, but it's still giving run-time error for the 64-bit version.
When I recompile it in my local machine, it runs ok... Perhaps when you recompiled it, it used the previous version of gflags?
Anyway, there's no need to take this as a priority. We'll workaround it for now. :)

Regards.

PR #2225, opened 2017-02-27
Update libkml to latest git revision

dsogari:
- rename package to "-git"
- add conflicts, provides and replaces
- update patch files

*note: the library authors merged my PRs for mingw into master, but they still did not make a new release; that's why I'm updating the package to use git


Alexpux:
@dsogari we move from git packages to stable versions so I will not accept your last pull requests.
You must fix issue in stable version with patch rather then move to git versions.

Alexpux:
@dsogari you move in wrong direction. We are reducing numbers of packages builded from git but you try increase it.

dsogari:
Ok @Alexpux , I understand now

PR #2226, opened 2017-02-28
npm: minor correction to PKGBUILD

mati865:
After mailing AppVeyor support they said CI is fixed.
This PR was created to verify it.

Also there was typo in npm PKGBUILD


mati865:
@mingwandroid is `input is not a tty` normal when running msys2 "headless"?

mati865:
@Alexpux somebody have to cancel all queued builds. I don't want to ask you for that because you are busy enough but probably only you have privilege to do it.

mingwandroid:
`input is not a tty` sounds ok here. It should instruct the software not to ask for user input and not to enter a REPL loop.

PR #2230, opened 2017-02-28
fixed tre libs build

emmenlau:
build of `mingw-w64-libsystre` was broken with a trivial problem missing round braces around provides, conflicts and replaces. build of `mingw-w64-libtre-git` did not specify a git commit, and the upstream head had changes.

### DIFF
--- a/mingw-w64-libsystre/PKGBUILD
+++ b/mingw-w64-libsystre/PKGBUILD
@@ -3,11 +3,11 @@
 _realname=libsystre
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-provides="${MINGW_PACKAGE_PREFIX}-libgnurx"
-conflicts="${MINGW_PACKAGE_PREFIX}-libgnurx"
-replaces="${MINGW_PACKAGE_PREFIX}-libgnurx"
+provides=("${MINGW_PACKAGE_PREFIX}-libgnurx")
+conflicts=("${MINGW_PACKAGE_PREFIX}-libgnurx")
+replaces=("${MINGW_PACKAGE_PREFIX}-libgnurx")
 pkgver=1.0.1
-pkgrel=2
+pkgrel=3
 pkgdesc="Wrapper library around TRE that provides POSIX API (mingw-w64)"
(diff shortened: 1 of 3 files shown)


mingw-w64-libtre-git/PKGBUILD
```diff
@@ -15,11 +16,11 @@ license=('BSD')
 options=(strip staticlibs)
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "git")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs" "${MINGW_PACKAGE_PREFIX}-gettext")
-source=("${_realname}"::"git+https://github.com/laurikari/tre.git"
+source=("${_realname}"::"git+https://github.com/laurikari/tre.git#commit=6fb7206"
```
Alexpux: Don't add commit hash here. We just build latest and ````pkgver````` function create ne version automatically

mingw-w64-libsystre/PKGBUILD
```diff
@@ -3,9 +3,9 @@
...
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-provides="${MINGW_PACKAGE_PREFIX}-libgnurx"
-conflicts="${MINGW_PACKAGE_PREFIX}-libgnurx"
-replaces="${MINGW_PACKAGE_PREFIX}-libgnurx"
+provides=("${MINGW_PACKAGE_PREFIX}-libgnurx")
+conflicts=("${MINGW_PACKAGE_PREFIX}-libgnurx")
+replaces=("${MINGW_PACKAGE_PREFIX}-libgnurx")
 pkgver=1.0.1
 pkgrel=2
```
Alexpux: Increase pkgrel to rebuild

mingw-w64-libtre-git/PKGBUILD
emmenlau: Please allow me to disagree. I think it would be good to always fix the upstream version, tag or commit. We can update upstream version together with `pkgrel` in regular intervals. But we should always update `pkgrel` (or `pkgver`) when updating upstream.

PR #2238, opened 2017-03-02
Update angleproject to latest git

matham:
This updates the package to build with latest angle git.

To be able to run this you need the latest mingw-w64-runtime. Specifically this commit: https://github.com/mirror/mingw-w64/commit/c628bd49a2c8cdee2250430c8d7b7cd95f1fe90a.

The main changes are as follows:

* `c++11` resulted in compilation errors similar to https://github.com/jb-alvarado/media-autobuild_suite/issues/433. Switching to `c++14` fixed it.
* The translator_lib has been merged into translator, so that's the name I used in the fixed patch.
* I had to disable the building of vulkan, which has been added to angle. Otherwise the build failed due to missing deps (at `spirv-headers`). Perhaps someone who knows how to build it may add that in later.
* The biggest thing, which I'm uncertain of, I removed `entry_points_shader.cpp` and its references. If I did not remove it I got errors saying that the stuff included in that file were redefined. I think these symbols for which this file was added (to support qt) may automatically be exported now. In fact they seem to have removed the defines that was used in that file to enable the exporting. However, someone with better knowledge should take a look.

While I haven't yet used the generated dlls, the build using `makepkg` completed successfully and I can see the dlls were properly generated.


Alexpux:
@matham see error building on CI

matham:
I don't understand. On appveyor it built it twice, the first [succeeded](https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.2420#L605) but the second [failed](https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.2420#L1019). What is the difference between the two builds?

Are they x64/x86 builds? I only tested locally with mingw64, not 32.

matham:
Btw, they reason why https://tea-ci.org/Alexpux/MINGW-packages/2454 failed is because it doesn't have the latest mingw-w64 runtime. You need this commit: https://github.com/mirror/mingw-w64/commit/c628bd49a2c8cdee2250430c8d7b7cd95f1fe90a as I mentioned above.

Alexpux:
@matham, we ll failed only 32 bit but not 64

matham:
Appveyor is now fixed for 64 and 32 bit. The tea CI fails due to the out of date runtime headers.

PR #2243, opened 2017-03-02
mingw-w64-ca-certificates: fixing download URL

emmenlau:
Until a real update of the certificates will happen, use another mirror of ca-certificates file. Build untested, but should work.


emmenlau:
Closing outdated PR, fixed upstream

PR #2256, opened 2017-03-07
gdb: Fix crash on mingw32

orgads:
Fixes https://github.com/Alexpux/MSYS2-packages/issues/716.

Future versions of gdb will have a flag for that[1].

[1] https://sourceware.org/bugzilla/show_bug.cgi?id=21187


Alexpux:
Increase pkgrel

PR #2264, opened 2017-03-08
mingw-w64-boost: updated to latest 1.63.0 checksum from boost.org

emmenlau:


emmenlau:
The CI builds failed, but not because of the checksum. In AppVeyor and TeaCI the new checksum is `Passed`, so I think this PR is correct. Should I increase the pkgrel (but nothing changed except checksum)?

Alexpux:
I\m build boost with old checksum, wonder why it changed. Need first verify myself

PR #2268, opened 2017-03-09
Fix detection of opencv lib path

matham:
When trying to fix the ITK builds I ran into a problem with the opencv cmake files. As you can see in the patch, opencv looks for the lib path at `${OpenCV_CONFIG_PATH}/${OpenCV_ARCH}/${OpenCV_RUNTIME}/lib`. However, mingw stores the lib files under `${OpenCV_CONFIG_PATH}/lib`, where `OpenCV_CONFIG_PATH` is e.g. `/msys64/mingw64` and doesn't separate them into arch etc like the original makefile does.


Alexpux:
Increase pkgrel

matham:
Fixed now. 

The tests are failing and I'm not sure why. I managed to build it successfully locally. The error is also completely uninformative.

It just says 
```
[ 15%] Built target opencv_imgproc
make: *** [Makefile:161: all] Error 2
```

matham:
So, the tests are failing because since https://github.com/Alexpux/MINGW-packages/pull/1818 was merged a new vtk package has not been published so it's suffering from that original issue. I hadn't realized opencv had a vtk dep.

PR #2270, opened 2017-03-09
switch to sourceforge.io homepages

vszakats:
after: https://sourceforge.net/blog/introducing-https-for-project-websites/


vszakats:
@Alexpux Thank you very much for the quick merges!

PR #2280, opened 2017-03-10
mingw-w64-qtbinpatcher: moved sources to public mirror

emmenlau:
And minor white space change: quote all paths with ${srcdir}


Alexpux:
I think better to import source archive here in the repository

emmenlau:
Ok I will do that

PR #2289, opened 2017-03-13
mingw-w64-extra-cmake-modules: updated from 5.20.0 to 5.32.0

emmenlau:


emmenlau:
This fixes also the problem that the sources are not available anymore.

PR #2290, opened 2017-03-13
mingw-w64-libcaca: explicitly disable docs build

emmenlau:
The build fails if doxygen is not available. Therefore I disabled the build of docs.
There is also a minor code cleanup.


emmenlau:
Should I add doxygen as a build-prerequisite, or is this PR good? Currently `mingw-w64-libcaca` will not build out of the box.

lazka:
A similar change made it in through d218ed35e82b1f78ee5c70887aff8db38cb

Issue #2143, opened 2017-02-01
mingw-w64-x86_64-rust 1.14.0-2  cargo does not work


kasehiroyuki:
With mingw-w64-x86_64-rust package version 1.14.0-2 , cargo command does not work with error.

```
$ cargo
C:/msys64/mingw64/bin/cargo.exe: error while loading shared libraries: zlib1.dll: cannot open shared object file: No such file or directory
```

But /mingw64/bin/zlib1.dll is already installed.
```
$ pacman -Qo /mingw64/bin/zlib1.dll
/mingw64/bin/zlib1.dll is owned by mingw-w64-x86_64-zlib 1.2.11-1
```

OS is Windows 10 Home 64bit.

Alexpux:
try under CMD. It will show you real missing library

kasehiroyuki:
Thank you for quick reply.
In cmd.exe, error dialog said that libssh2-1.dll is required.

Then I installed  mingw-w64-x86_64-libssh2 and zlib1.dll error was not shown.

ScreamingInDigital:
There is a bug with "ldd" on Windows 10.  It does not show missing dependencies.

Install "ntldd" and execute 'ntldd <library or executable name> | grep "not found" ' to see the list of missing dependencies.

Alexpux:
@ScreamingInDigital , no one tell here about ldd

ScreamingInDigital:
@Alexpux ,
Assume that "libssh2" was missing and "rust" was installed, the following could be executed to determine the missing library or libraries with "ntldd" from the msys2/mingw64 prompt.
ntldd cargo.exe | grep "not found"

Alexpux:
I know about ntldd, it is not bug free too. I prefer to use Dependency Walker in such cases. It also show if some imports are missing in libraries

ScreamingInDigital:
@Alexpux 
I also use dependency walker.  "ntldd" is currently not as "buggy" as "ldd", and he could have executed that immediately after trying to use "cargo".
In the case of the "cargo" command, it would have worked fine.

hcorion:
@kasehiroyuki Does the issue still persist with the latest version of rust?

kasehiroyuki:
@hcorion This issue is missing dependency to  mingw-w64-x86_64-libssh2 
and I checked 1.15.1-1 and seems to have the same problem.

Issue #2247, opened 2017-03-04
clang: boost/thread + boost/dll: Multiple definitions of GetModuleHandleA


askondro:
I have issues when using these headers in the same compilation unit and compile it with clang++.

```
#include <boost/thread.hpp>
#include <boost/dll.hpp>
```

g++ works OK.

with clang++ I get this error:

clang version 3.9.1 (tags/RELEASE_391/final)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:\msys64\mingw64\bin
[...]

Issue #1834, opened 2016-10-18
multiple definition with TLS when build clang5.0.0


yumetodo:
Here is small sample to reproduce.

https://bitbucket.org/yumetodo/tls_test

```
$cmake -G "MSYS Makefiles" -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_CXX_COMPILER=clang++
-- The CXX compiler identification is Clang 3.8.0
-- Check for working CXX compiler: D:/msys64/mingw64/bin/clang++.exe
-- Check for working CXX compiler: D:/msys64/mingw64/bin/clang++.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The C compiler identification is GNU 6.2.0
-- Check for working C compiler: D:/msys64/mingw64/bin/gcc.exe
-- Check for working C compiler: D:/msys64/mingw64/bin/gcc.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: D:/msys64/home/yumetodo/tls_test/TLS_test
$make
Scanning dependencies of target TLS_test
[ 33%] Building CXX object CMakeFiles/TLS_test.dir/a.cpp.obj
[ 66%] Building CXX object CMakeFiles/TLS_test.dir/main.cpp.obj
[100%] Linking CXX executable TLS_test.exe
CMakeFiles/TLS_test.dir/objects.a(main.cpp.obj):(.text+0xc0): `TLS wrapper function for Test::mt' が重複して定義されています
CMakeFiles/TLS_test.dir/objects.a(a.cpp.obj):(.text+0x140): ここで最初に定義されています
clang++.exe: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [CMakeFiles/TLS_test.dir/build.make:122: TLS_test.exe] エラー 1
make[1]: *** [CMakeFiles/Makefile2:68: CMakeFiles/TLS_test.dir/all] エラー 2
make: *** [Makefile:84: all] エラー 2
```

On msys2 GCC6.2.0 or Ubuntu 16.04 clang3.8.0 report no error.

[...]

PR #2294, opened 2017-03-13
clang: update to 4.0.0

mati865:
Do not merge yet, cc @martell .

`libc++`, `libc++abi`, `libunwind`, `lld` are disabled because they are broken too much.
Other packages should be working (Clang bugs reported here aren't fixed).
Additionally LLVM is now linked dynamically because otherwise `llvm-config --link-shared` breaks.


hcorion:
Can you test if rust still works? I'm pretty sure it depends on libunwind. I think it may build it's own version, but I'm not 100% sure.

mati865:
As far as I know LLVM libunwind was used only for libc++ and rust provides own libunwind using `gcc_eh` for mingw build.

BTW I'm preparing rust 1.16 release with rustbuild system.

hcorion:
Awesome! It would also be cool if you fixed bug #2143 in your 1.16 PR by adding libssh2 as a dependency.

mati865:
My plan is to move cargo to own package as well (like all Linux distributions do).
Of course fix will be included.

mati865:
Force pushed to add new patch: `0007-add-pthread-as-system-lib-for-mingw.patch` and restore old `0502-hack-to-use-64-bit-time-for-mingw.patch`.

mati865:
@Alexpux I think it's ok to merge, both libc++ and lld never worked in MSYS2 and now they don't build.

Alexpux:
I dont want to merge without libc++ and lld because we already have this packages in repository. If I'm upload new packages without those parts then we get issues with updating llvm packages on user systems who install libc++ and lld. And also I will need remove not builder parts from pacman databases

martell:
Hey @mati865,
Good start here.

Does `HasFloat128 = true;` fix the issue with the large revert we had to do previously?
Also could you share your errors when building lld libunwind 4.0 etc?
I can have a look.

Best,
Martell

mati865:
@martell, Hey.
`HasFloat128 = true;` tricks libstc++ to think Clang supports _float128 but doesn't expose `__FLOAT128__` define. Trying to use _float128 fill cause compilation error which is way better than segfault from running app.
Simple and brilliant, isn't it?
Downside is [failing test](https://gist.github.com/mati865/3a350709f853e4a8cb1fc40980a4f5b0) `Sema/128bitfloat.cpp` but I think we can live without it 😄.
Also I wanted to post my findings on the [LLVM bugzilla](https://bugs.llvm.org//show_bug.cgi?id=30685) but registration is "temporary" disabled...

LLD patches didn't apply and there was some code refactoring so I just disabled it as it is even less useful than before. Although it build fine so I'll update PR soon.

libc++ has seen big changes but after spending more hours than I wanted it is building without libunwind which is WIP. I think I'll get it ready tomorrow.

mati865:
Good news, everything builds now.

Some optimistic notes:

- LLD was already useless before and now without patches it's even more useless.
- libc++ was totally broken and now is even more broken (including <iostream> will cause error `D:\msys64\mingw32\include\c++\v1\__bsd_locale_fallbacks.h:121:17: error: use of undeclared identifier 'vasprintf'; did you mean 'vsprintf'?`)

In summary 2 broken features are even more broken.

martell:
LLD works just not with anything built with binutils :)

libc++ HEAD now almost passes fully for the MSVC target as standalone.
Adding MINGW support on top of that work is next on my list of tasks.

`vasprintf` is an actual function, you need to build with `_GNU_SOURCE` defined to 1 for it to work.

Alexpux:
Great @martell !!!

mati865:
@martell so it is broken for us 😶

martell:
If the mingw-w64-crt is built with llvm-dlltool then it can be used for us.
LLD should work for every other target though which is useful also.
I think the next major llvm release will have everything we need.

@mati865 How well is lldb working btw?

martell:
@mati865 This should be useful
http://clang-developers.42468.n3.nabble.com/Re-Newlib-v2-5-0-and-LibC-locale-support-BROKEN-td4054882.html

mati865:
@martell Now when I'm back home from college I cannot believe I missed that `_GNU_SOURCE`.
I guess we could define it in `${MINGW_PREFIX}/include/c++/v1__config`?

I tested LLDB with WSL (also called Bash on Ubuntu on Windows) when I updated LLVM/Clang to 3.9.0. It connected and worked fine as remote.
For pure Windows usage it is very limited as `lldb-server` uses POSIX API, I enabled it in CMake back at 3.9.0 and when it failed due to missing POSIX functions I disabled it.
So we have client only which can possibly work with https://github.com/facebook/ds2 but I haven't tested it.

Using libc++ now fails with impressive 1347 broken [1] references: https://gist.github.com/mati865/cb6daee33faafe06b877b562e705c3bf
Probably the best way (not easiest) to solve it would be to fix `-nodefaultlibs` and specify required libs. Also https://github.com/Alexpux/MINGW-packages/pull/2294/files#diff-ac7dd18141827df6a35a303e84901615L181 could need to be readded as `-lc++abi` is missing in linker args.

Currently at MINGW-packages we have 2 rather serious (#1677, #1834) and another 2 important (#1678, #2247) issues open for Clang and it would be great to have them fixed but I had to gave up.

[1] There is a word that describes it more accurately but let's avoid it in public.

revelator:
Small whoops, seems libunwind from clang breaks building glog with a multiple definition error. This might also cause other sources to fail because it includes the same symbols as libgcc which is allways linked in.
So if some source says hey i want to use libunwind with gcc it will break. You can offcourse get around that by adding a -Wl,--allow-multiple-definition to flags but it will propagate down to every source linking to that library afterwards.

PR #2297, opened 2017-03-14
libsodium: update to 1.0.12

woobs:


mingw-w64-libsodium/PKGBUILD
```diff
@@ -12,13 +12,15 @@ license=('custom:ISC')
...
-sha256sums=('a14549db3c49f6ae2170cbbf4664bd48ace50681045e8dbea7c8d9fb96f9c765'
+sha256sums=('b8648f1bb3a54b0251cf4ffa4f0d76ded13977d4fa7517d988f4c902dd8e2f95'
             'SKIP')
 
 build() {
   rm -rf "${srcdir}/build-${MINGW_CHOST}"
   mkdir "${srcdir}/build-${MINGW_CHOST}"
   cd "${srcdir}/build-${MINGW_CHOST}"
+  
+  export CFLAGS="-O3 -mtune=haswell"
```
mati865: I think it's a bad idea.

woobs: My understanding is that *mtune* will enable the AVX2 pathway at runtime for CPUs that can handle it, but will still allow code to run on older CPUs.  As opposed to *march* which would restrict the code to running on that specific architecture.  If this understanding is incorrect, then we can certainly go back to the usual build script.

Libsodium 1.0.12 with AVX2 selects a code path that results in 5-10x encryption speed improvements on Windows.

mati865: From [GCC docs](https://gcc.gnu.org/onlinedocs/gcc-6.3.0/gcc/x86-Options.html#x86-Options):
```
-mtune=cpu-type
    Tune to cpu-type everything applicable about the generated code, except for the ABI and the set of available instructions.
```
Doesn't look like it will allow using AVX2 and will possibly reduce performance for non Haswell CPUs.

woobs: Fair enough, I'll re-submit it.  We custom compile the library at our institution anyway.

PR #2298, opened 2017-03-14
Add package MathGL (version 2.3.5.1)

dsogari:
MathGL from http://mathgl.sourceforge.net/doc_en/Main.html

There are many optional dependencies that I made requirements (for the sake of simplicity, since I do not know exactly which of them are truly optional and which are needed at run-time).

### DIFF
--- a/mingw-w64-mathgl/PKGBUILD
+++ b/mingw-w64-mathgl/PKGBUILD
@@ -9,8 +9,7 @@ pkgdesc="Library for high-quality scientific graphics (mingw-w64)"
 arch=('any')
 url="http://mathgl.sourceforge.net/"
 license=('LGPL')
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
-             "${MINGW_PACKAGE_PREFIX}-cmake")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake")
 depends=("${MINGW_PACKAGE_PREFIX}-hdf5"
          "${MINGW_PACKAGE_PREFIX}-fltk"
          "${MINGW_PACKAGE_PREFIX}-libharu"
--- a/mingw-w64-mathgl/PKGBUILD
+++ b/mingw-w64-mathgl/PKGBUILD
@@ -1,0 +1,51 @@
+# Maintainer: Diego Sogari <diego.sogari@gmail.com>
+
+_realname=mathgl
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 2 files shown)


mingw-w64-mathgl/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+_realname=mathgl
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=2.3.5.1
+pkgrel=1
+pkgdesc="Library for high-quality scientific graphics (mingw-w64)"
+arch=('any')
+url="http://mathgl.sourceforge.net/"
+license=('LGPL')
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
```
mati865: `"${MINGW_PACKAGE_PREFIX}-gcc"` is part of `${MINGW_PACKAGE_PREFIX}-toolchain` therefore it's redundant.

mingw-w64-mathgl/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+         "${MINGW_PACKAGE_PREFIX}-fltk"
+         "${MINGW_PACKAGE_PREFIX}-libharu"
+         "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo"
+         "${MINGW_PACKAGE_PREFIX}-libpng"
+         "${MINGW_PACKAGE_PREFIX}-giflib"
+         "${MINGW_PACKAGE_PREFIX}-qt5"
+         "${MINGW_PACKAGE_PREFIX}-freeglut"
+         "${MINGW_PACKAGE_PREFIX}-wxWidgets")
+source=("${_realname}-${pkgver}.tar.gz::http://downloads.sourceforge.net/${_realname}/${_realname}-${pkgver}.tar.gz"
+        "${_realname}-${pkgver}.patch")
```
mati865: Use static name for patch.

mingw-w64-mathgl/PKGBUILD
dsogari: Ok, thanks! Will update it

dsogari:
@mati865 
Do you think I should create different packages for GPL and LGPL versions of the library?

I need the LPGL version, but there may be users who want the GPL version, which has support for GSL (https://www.gnu.org/software/gsl/).

mati865:
@dsogari I haven't heard about MathGL before.
Pinging @Alexpux and @mingwandroid

PR #2308, opened 2017-03-16
sqlite3: repackage

woobs:
It seems the 3.17.0 change 2 weeks ago was merged into this Github repository, but never submitted to the MSYS2/Pacman repository


Alexpux:
Well it doesn\t need to increase pkgrel in this case. Just need to ask

galeksandrp:
@Alexpux It seems also that MSYS2-packages/cmake 3.6.2-2 is not available in pacman.

Issue #2149, opened 2017-02-03
miniupnp 2.0


fr500:
Would it be possible to get miniupnp updated to 2.0?
there were quite a bit of changes to the API from what I understand and trying to get it to build has been frustrating...

galeksandrp:
#2309

PR #2309, opened 2017-03-16
miniupnpc

galeksandrp:
2.0 is successfully build with 001-link-libraries.patch, but now without it
2.0.20161216 does not build


mati865:
If you include message like "closes #2149" in pull request comment GitHub will close that issue when PR is merged.

galeksandrp:
@mati865 Included, thanks.

PR #2311, opened 2017-03-16
downloads.sourceforge.io -> downloads.sourceforge.net

galeksandrp:
sourceforge.io is for user's subdomains, not for download servers
seem that [someone](https://github.com/Alexpux/MINGW-packages/commit/92975171013db4c0c4b6298c94795ed2808bdb28) overplayed with sed :laughing:


vszakats:
It was me. Thank you for the fix!

PR #2319, opened 2017-03-17
Add package MinHook

int19h:


mingw-w64-MinHook/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+
+build() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+  make -f build/MinGW/Makefile clean
+  make -f build/MinGW/Makefile
+}
+
+package() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+  install -Dm644 include/MinHook.h "${pkgdir}/${MINGW_PREFIX}/include/MinHook.h"
```
Alexpux: ````MINGW_PREFIX```` already starts with ````/```` so no need to add it between ````pkgdir```` and ````MINGW_PREFIX````

int19h: Will fix. FWIW, this was copied from another PKGBUILD. I can't remember which now, but might be worth grepping for.

int19h:
Removed redundant slashes.

PR #2324, opened 2017-03-19
sqlcipher: new package

mati865:
Closes https://github.com/Alexpux/MINGW-packages/issues/2323

### DIFF
--- a/mingw-w64-sqlcipher/01-fix_dep.diff
+++ b/mingw-w64-sqlcipher/01-fix_dep.diff
@@ -1,0 +1,11 @@
+--- a/Makefile.in
++++ b/Makefile.in
+@@ -1074,7 +1074,7 @@
+    $(TOP)/ext/fts5/fts5_varint.c \
+    $(TOP)/ext/fts5/fts5_vocab.c  \
+ 
+-fts5parse.c:	$(TOP)/ext/fts5/fts5parse.y lemon 
++fts5parse.c:	$(TOP)/ext/fts5/fts5parse.y lemon$(BEXE)
+ 	cp $(TOP)/ext/fts5/fts5parse.y .
+ 	rm -f fts5parse.h
+ 	./lemon$(BEXE) $(OPTS) fts5parse.y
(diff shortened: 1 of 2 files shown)


sjemens:
The sqlite3 is not needed anywhere in compiling. Openssl is also needed in runtime.
I suggest changing to
depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs" "${MINGW_PACKAGE_PREFIX}-readline" "${MINGW_PACKAGE_PREFIX}-openssl")
makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "tcl")
Also if the wineditline package is installed sqlcipher will try to link to -ledit and the shell.c compilation will fail. So I suggest configure with the following parameters
  ${srcdir}/${_realname}-${pkgver}/configure \
    --prefix=${MINGW_PREFIX} \
    --build=${MINGW_CHOST} \
    --host=${MINGW_CHOST} \
    --enable-tempstore=yes \
    --disable-tcl \
    --disable-editline \
    --enable-readline \
    --with-readline-lib="-L${MINGW_PREFIX}/lib -lreadline" \
    --with-readline-inc="-I${MINGW_PREFIX}/include" \
    CFLAGS="$CFLAGS -DSQLITE_HAS_CODEC" LDFLAGS="-lcrypto"

mati865:
makedepends were taken from Arch, I'll fix it.

> Also if the wineditline package is installed sqlcipher will try to link to -ledit and the shell.c compilation will fail.

I have it installed and it doesn't fail if `-lhistory`  is added to LDFALGS. Can you explain why editline should be disabled?

sjemens:
You are right. With -lhistory (which belongs to package readline) explicity added to LDFLAGS the build does not fail, but i noticed that the sqlcipher.exe does not have the history command capabilities, which after all is why we linked in the first place to the library. If we did not want history capabilities in console, we could disable both editline and readline and skip the dependencies, but as i said, I do not think this is the desired approach.

With -lreadline added to LDFLAGS, everything is working as expected.  With just -ledit added the builing fails (it errors with undefined reference to 'stifle_history'). No need to depend on both since they try to offer the same (or at least I think is the same) functionality.

In conclusion, I think that explicity enabling readline and disabling editline states more clearly what is working and what is not.

mati865:
Updated

DX37:
Thanks, guys! :)

PR #2337, opened 2017-03-23
python3: segfault fix in GetTickCount64

PolarNick239:
On Windows XP SP3 there is no GetTickCount64, current workaround for this problem has a bug: ```GetProcAddress(hKernel32, "GetTickCount64")``` returns ```NULL```, but next if compares result with ```-1``` and so it tries to call function by ```NULL``` address. This leads to segfault.


Alexpux:
You need  increase pkgrel and update checksums

Alexpux:
@PolarNick239 ping

PolarNick239:
Sorry, but I even don't exactly know what about this repo, so I am not be able to easily update pkgrel and checksums. Also I don't have time for this :(

I just was googling about fixing python3.5 for windows, and found these patches - then I used them, but encountered segfault, fixed it and proposed this fix for you.

So you can reject pull request, or finish it on your own.

P.S. why checksums are not calculated at some packaging/building stage? Why they should be updated on each commit?

mati865:
Checksums are updated with `updpkgsums` tool available in MSYS2.

> P.S. why checksums are not calculated at some packaging/building stage? 

You need to know if files aren't malformed/damaged before you start using them. It's the only reasonable way of using checksums.

> Why they should be updated on each commit?

They only have to be updated when patches or sources are changed.

PR #2343, opened 2017-03-26
rust: update to 1.16.0

mati865:
Cargo is getting integrated closer into Rust and cannot be build on system without it (it requires itself), therefore I decided not to split it.

Closes https://github.com/Alexpux/MINGW-packages/issues/2143


mingw-w64-rust/PKGBUILD
```diff
@@ -5,120 +5,107 @@
...
-             "${MINGW_PACKAGE_PREFIX}-python2"
-             "${MINGW_PACKAGE_PREFIX}-llvm"
              "${MINGW_PACKAGE_PREFIX}-jemalloc"
-             "${MINGW_PACKAGE_PREFIX}-rust>=1.13.0"
-             "${MINGW_PACKAGE_PREFIX}-cmake")
+             "${MINGW_PACKAGE_PREFIX}-libffi"
+             "${MINGW_PACKAGE_PREFIX}-python2"
+             "${MINGW_PACKAGE_PREFIX}-cmake>=3.4.3"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-rust")
```
hcorion: libssh2 needs to be added as a dependency to fix https://github.com/Alexpux/MINGW-packages/issues/2143 . 
Also, why did you get rid of the rust minimum version requirement? That was to make sure the user (or tea-ci) didn't try to build it with an ancient version of rust that didn't include cargo.
Otherwise LGTM.

hcorion: libssh2 needs to be added as a dependency to fix #2143 .
Also, why did you get rid of the rust minimum version requirement? That was to make sure the user (or tea-ci) didn't try to build it with an ancient version of rust that didn't include cargo.
Otherwise LGTM.

mati865:
Ah, now I know know why I had feeling about forgetting of something.

Looks like I had older version checked but Tea fails on CMake as well.

hcorion:
Yeah, it's too bad wine doesn't support dwarf4, then tea-ci would be useful to the rust package.

PR #2362, opened 2017-03-31
Fix #2254: libzip is missing zipconf.h header

### LINKED ISSUE #2254: libzip is missing zipconf.h header
After installing mingw-w64-libzip, I cannot compile any code that does `#include "zip.h"` (which comes from that library), because that header in turn does `#include <zipconf.h>` - which is not installed in /mingw64/include.

Looking at the PKGBUILD for the package, it appears that it had a line that should address this:
https://github.com/Alexpux/MINGW-packages/blob/7d616ce3e4a490088d4e7e3c1925c3d5b9bd960d/mingw-w64-libzip/PKGBUILD#L45

However, it is commented out. After doing the same exact operation manually, I can compile successfully.

So it would seem that the PKGBUILD just needs to have that line uncommented. I'm not sure why it was commented in the first place - looking at history, it appears that it was already commented out in the initial commit for the package.

int19h:
Copy zipconf.h to where it can be found by the main header.

Remove redundant slashes after `${pkgdir}`.

### DIFF
--- a/mingw-w64-libzip/PKGBUILD
+++ b/mingw-w64-libzip/PKGBUILD
@@ -41,7 +41,6 @@ package() {
 
   make DESTDIR="${pkgdir}" install
 
-  # preserve old header path for compatibility
-  # ln -s ${MINGW_PREFIX}/lib/libzip/include/zipconf.h "${pkgdir}${MINGW_PREFIX}/include/zipconf.h"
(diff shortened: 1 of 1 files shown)


Alexpux:
Proper fix is to fix CMake install command to install header in right folder

PR #2372, opened 2017-04-05
plplot: Updated to version 5.12.0 and fixed building of package.

stahta01:
Changed include of "Windows.h" to "wx/msw/wrapwin.h" via patch file.
Removed "Fix CMake FindSWIG" code from package.
And, added defines:
  DEFAULT_NO_BINDINGS=ON
  ENABLE_ada=ON
  ENABLE_cxx=ON
  ENABLE_f95=ON
  ENABLE_tcl=ON
  ENABLE_tk=ON
  ENABLE_wxwidgets=ON


stahta01:
I just figured out that my path was being encoded in the .pc files.
Tim S.

PR #2375, opened 2017-04-07
crt-git: Commented out duplicate DEFINE_GUID in uuid.c;

stahta01:
that was also in extras-uuid.c.

When self building wxWidgets Monolithic DLL I get link errors from some of the duplicate  DEFINE_GUID that are in both uuid.c and extras-uuid.c.

Tim S.


Alexpux:
You must report it upstream to mingw-w64

PR #2376, opened 2017-04-07
gdk-pixbuf2: re-enable gdiplus loaders

Ede123:
(they were disabled in 5b53585ec2b119eaf37517d84ab2a1940430146b)

GDI+ loaders neet to be built into gdk-pixbuf to prevent the bug described in https://bugzilla.redhat.com/show_bug.cgi?id=795152#c1 (see also https://bugs.launchpad.net/inkscape/+bug/1467103)

### DIFF
--- a/mingw-w64-gdk-pixbuf2/PKGBUILD
+++ b/mingw-w64-gdk-pixbuf2/PKGBUILD
@@ -6,7 +6,7 @@ _realname=gdk-pixbuf2
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.36.6
-pkgrel=1
+pkgrel=2
 pkgdesc="An image loading library (mingw-w64)"
 arch=('any')
 url="http://www.gtk.org"
@@ -60,8 +60,7 @@ build() {
     --enable-shared \
     --enable-introspection \
     --enable-relocations \
-    --with-included-loaders=png \
-    --without-gdiplus
+    --with-included-loaders=png,gdip-bmp,gdip-emf,gdip-gif,gdip-ico,gdip-jpeg,gdip-tiff,gdip-wmf
   make
 }
 


Ede123:
@peterbud: Any specific reason for disabling gdiplus loaders? They work just fine, Inkscape makes use of them (that's how I noticed they were missing from MSYS2 packages).

One question out of curiosity: Why is png built into gdk-pixbuf, too? I noticed arch itself does it, but I don't see why...

peterbud:
@Ede123 : I had a very specific problem which I wanted to fix in October (which I have also reported upstream which they have [ignored](https://bugzilla.gnome.org/show_bug.cgi?id=773279) BTW). 

When I have tested gdk-pixbuf with GDI+ loaders (I have tried), it has crashed. I guess I have bumped into the same problem you have mentioned at the link. For me your changes looks OK, I'll test them later to see if they will work on 64-bit Win 10

Ede123:
I'm on Windows 10 x64 myself, so you should be fine...

Have you seen the comment at your upstream bug? From what I understand Bastien want's to say that `priv->animation` also returns something useful for a GdkPixbuf that is not an animation, i.e. the else is really only executed in case of a failure. In that case your patch would return the GdkPixbuf even in cases were it validly should not... Maybe you really hit an edge case?

peterbud:
Hi @Ede123 . That particular function is called [here](https://github.com/darktable-org/darktable/blob/f6d261c5d86a4de7e4bc4b1b54d39b7482dd7fad/src/common/camera_control.c#L358), in the darktable project.
IMHO this function is definitely wrong. It returns a `NULL` pointer, if the referenced image is NOT an animation (ie `priv->animation` is NULL, then it returns a `NULL`, instead if the image pointer), see [here](https://github.com/GNOME/gdk-pixbuf/blob/master/gdk-pixbuf/gdk-pixbuf-loader.c#L738). 
I have tested this with darktable (see source code above), and it always returned a NULL pointer. With my fix it is working nicely. 
Maybe it has something to do with different platforms, as the Linux version of `gdk-pixbuf` code is working OK (with Linux version of darktable), but not working on Windows without my patch.

Ede123:
I see. Maybe you should add that information upstream... Might well be that this in fact is a Windows-only issue!

Ede123:
@peterbud: I think I know how you can solve your issue!
You have to call [`gdk_pixbuf_loader_close()`](https://developer.gnome.org/gdk-pixbuf/stable/GdkPixbufLoader.html#gdk-pixbuf-loader-close) before calling [`gdk_pixbuf_loader_get_pixbuf()`](https://developer.gnome.org/gdk-pixbuf/stable/GdkPixbufLoader.html#gdk-pixbuf-loader-get-pixbuf).

From the documentation:
> Informs a pixbuf loader that no further writes with gdk_pixbuf_loader_write() will occur, so that it can free its internal loading structures. Also, tries to parse any data that hasn't yet been parsed; if the remaining data is partial or corrupt, an error will be returned.

That's for example [how Chromium does it](https://chromium.googlesource.com/chromium/chromium/+/master/ui/base/resource/resource_bundle_gtk.cc#34) and this issue is also discussed [here](http://purebasic.fr/english/viewtopic.php?p=472206).

peterbud:
Thanks for that, I'll test it this week and let you know what is the result!

Ede123:
@Alexpux: Ping! Anything missing or are you still waiting for feedback by @peterbud?

Alexpux:
Waiting feedback

peterbud:
Tomorrow I'll be back from vacation and over the weekend will look into this

peterbud:
@Ede123  Ok, I have tested - you are right, the suggested change is working. `gdk_pixbuf_loader_close()` must be called before `gdk_pixbuf_loader_get_pixbuf()` - thank you.
I'm wondering why that code is working though on Linux without calling `gdk_pixbuf_loader_close()`, and results this problem only on the Windows platform.
@Alexpux - I belive you can merge this PR

PR #2400, opened 2017-04-19
mingw-w64-thrift: use newer than WinXP as default build target

emmenlau:
Use newer than WinXP as default build target

### DIFF
--- a/mingw-w64-thrift/0001-thrift-build-option-defines.patch
+++ b/mingw-w64-thrift/0001-thrift-build-option-defines.patch
@@ -1,0 +1,16 @@
+diff --git a/lib/cpp/src/thrift/windows/config.h b/lib/cpp/src/thrift/windows/config.h
+index bc4aa42..174e8c1 100644
+--- a/lib/cpp/src/thrift/windows/config.h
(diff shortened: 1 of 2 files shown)


Ede123:
Why?

emmenlau:
I found a thread that lead me to believe that the WinXP code is inferior to the code path
for newer targets. Since the default target for MSYS2 is anyways newer than XP, I think it
makes sense to default to the newer code path.

See i.e.:
  * http://www.mail-archive.com/dev@thrift.apache.org/msg32675.html
  * https://github.com/apache/thrift/blob/master/lib/cpp/README.md (section "Windows Compatibility")

Ede123:
@emmenlau: Thanks!

It always helps *a lot* if commit messages specify *why* some changes are done, otherwise it's often impossible to understand the reasoning behind such a change in future (I stumbled across some of these lately).

emmenlau:
@Ede123 yes what you say makes sense, I will try to be more explicit in the future. The package is
pretty new, so I assumed to be the only user, and admittedly I was a bit sloppy :-)

PR #2406, opened 2017-04-22
Update git-lfs to v2.0.2

srjek:
Some problems:
 * ~~bsdtar chokes on the zip. Switching to the tar.gz release doesn't help. I can unzip it fine with the bsdtar on my Arch box, so it might be a problem with MSYS2 or a bug that was fixed in newer versions. As a workaround you can simply run makepkg-mingw twice, for some reason it only chokes on the zip when operating in a clean directory. I might try upgrading the libarchive PKGBUILD later to see if that helps.~~ (Resolved by Alexpux/MSYS2-packages#880)
 * GOROOT is set via a /mingwXX/etc/profile.d script, but makepkg-mingw doesn't pull them in. To get that environment variable I have to manually source the script, but the right fix might be to modify either makepkg-mingw or change how GOROOT is set.
 * We can now build man pages, I just had to install the ronn gem to the build system. I'm not sure if it's ok for a PKGBUILD to have side effects like that, so some input would be appreciated.

### DIFF
--- a/mingw-w64-git-lfs/0001-Translate-git-paths-from-msys-to-windows.patch
+++ b/mingw-w64-git-lfs/0001-Translate-git-paths-from-msys-to-windows.patch
@@ -1,79 +1,15 @@
-From 46ade92e68e3041f315867afd4b157258a1981b5 Mon Sep 17 00:00:00 2001
-Message-Id: <46ade92e68e3041f315867afd4b157258a1981b5.1483358798.git-series.abdo.roig@gmail.com>
-From: Abdo Roig-Maranges <abdo.roig@gmail.com>
-Date: Fri, 19 Aug 2016 13:46:27 +0200
-Subject: [PATCH] Translate git paths from msys to windows
-
----
- git/git.go | 26 ++++++++++++++++++++++----
- 1 file changed, 22 insertions(+), 4 deletions(-)
-
-diff --git a/git/git.go b/git/git.go
-index 72ab9c7..c66048c 100644
---- a/git/git.go
-+++ b/git/git.go
-@@ -574,6 +574,20 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
- 	}
- }
- 
-+func CygpathWindowsAbsolute(path string) (string, error) {
-+	cmd := subprocess.ExecCommand("cygpath", "-w", strings.TrimSpace(path))
-+	buf := &bytes.Buffer{}
-+	cmd.Stderr = buf
-+
-+	out, err := cmd.Output()
-+	output := strings.TrimSpace(string(out))
-+	if err != nil {
-+		return "", fmt.Errorf("Failed to call cygpath -w: %q", buf.String())
-+	}
-+
-+	return filepath.Abs(output)
-+}
-+
- func GitAndRootDirs() (string, string, error) {
- 	cmd := subprocess.ExecCommand("git", "rev-parse", "--git-dir", "--show-toplevel")
- 	buf := &bytes.Buffer{}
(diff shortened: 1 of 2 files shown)


srjek:
I was able to resolve the bsdtar issue by upgrading to 3.3.1 (Alexpux/MSYS2-packages#880)

mfld-fr:
@srjek : having git-lfs package available for end user would be so valuable that I would delay the man pages generation to a later step.

aroig:
> We can now build man pages, I just had to install the ronn gem to the build system. I'm not sure if it's ok for a PKGBUILD to have side effects like that, so some input would be appreciated.

Cool! It would be much cleaner to make a package for ronn, and add it on the `makedepends` list.

mfld-fr:
@srjek : even after upgrading to the lastest versions of libarchive (3.3.1-1) and bsdtar (3.3.1-1), the build of this package still fails on the extraction step:

```
$ makepkg-mingw -sCLf
==> Création du paquet mingw-w64-git-lfs 2.0.2-1 (jeu. 15 juin 2017 17:32:40)
==> Vérification des dépendances pour l’exécution...
==> Vérification des dépendances pour la compilation...
==> Récupération des sources...
  -> git-lfs-2.0.2.zip trouvé
  -> 0001-Translate-git-paths-from-msys-to-windows.patch trouvé
==> Validation des fichiers source avec sha256sums...
    git-lfs-2.0.2.zip ... Réussite
    0001-Translate-git-paths-from-msys-to-windows.patch ... Réussite
==> Suppression du répertoire $srcdir/ existant...
==> Extraction des sources...
  -> Extraction de git-lfs-2.0.2.zip avec bsdtar
git-lfs-2.0.2/docs/api/schemas/http-batch-request-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-batch-request-schema.json'
git-lfs-2.0.2/docs/api/schemas/http-batch-response-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-batch-response-schema.json'
git-lfs-2.0.2/docs/api/schemas/http-lock-create-request-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-lock-create-request-schema.json'
git-lfs-2.0.2/docs/api/schemas/http-lock-create-response-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-lock-create-response-schema.json'
git-lfs-2.0.2/docs/api/schemas/http-lock-delete-request-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-lock-delete-request-schema.json'
git-lfs-2.0.2/docs/api/schemas/http-lock-list-response-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-lock-list-response-schema.json'
git-lfs-2.0.2/docs/api/schemas/http-lock-verify-response-schema.json: Can't create 'git-lfs-2.0.2/docs/api/schemas/http-lock-verify-response-schema.json'
bsdtar: Error exit delayed from previous errors.
==> ERREUR : L’extraction de git-lfs-2.0.2.zip a échoué
    Abandon...
```

srjek:
@mfld-fr I could have swore it was working, but now I'm reproducing the issue on my machine as well, with the original binaries I compiled when I made the libarchive patch. As a work around for the short term, unzipping a second time should work.

Alexpux:
We have such issues with many packages. For workaround look into qtcreator PKGBUILD for example

mfld-fr:
Thanks @Alexpux for the suggestion, that I applied.

As this pull request has already been merged, I propose to continue this discussion in the issue https://github.com/Alexpux/MINGW-packages/issues/2595.

PR #2419, opened 2017-04-28
SDL2: keep the -mwindows flag in the pkg-config --libs output

### LINKED ISSUE #2413: MSYS2 removes -mwindows flag from --libs output in pkg-config file
Why on earth is the `-mwindows` flag removed from the `--libs` output in SDL2's pkg-config file?

https://github.com/Alexpux/MINGW-packages/commit/4d4676e9ffc6773b5dc38138a5cf8bc7227cc16e
(line 31)

This breaks e.g. compiling chocolate-doom's sdl2-branch:
https://github.com/chocolate-doom/chocolate-doom/issues/899

fabiangreffrath:
Removing this flag is an unexpected deviation from upstream which may
break building applications that rely on this flag being properly set
by SDL2, e.g.

https://github.com/chocolate-doom/chocolate-doom/issues/899

There may be applications out there using SDL2 which do not require
this flag, but they may still filter it out from the output of
'pkg-config --libs' themselves. We should not turn the exception into
the rule.

Fixes #2413


mati865:
@Alexpux do you remember why it was disabled?

Alexpux:
@mati865 as I remember some issues with ffmpeg

wiiaboo:
ffmpeg adds `-mconsole` to SDL2's LDFLAGS since http://git.videolan.org/?p=ffmpeg.git;a=commit;h=c5308eea29216c218b61599dc83911d2c1afe716, so that's probably not needed anymore.

holgern:
[gcc6.patch.txt](https://github.com/Alexpux/MINGW-packages/files/1218776/gcc6.patch.txt)
Please apply this patch as follows:
source=("https://libsdl.org/release/SDL2-${pkgver}.tar.gz"
        "gcc6.patch")
sha256sums=('442038cf55965969f2ff06d976031813de643af9c9edc9e331bd761c242e8785'
            'c99fff01674631aeda4543c412e839e5ba2fc94dea38406efa79a4575cd99884')

prepare() {
  cd "${srcdir}"/${_realname}-${pkgver}
  patch -p1 -i ${srcdir}/gcc6.patch
}

wiiaboo:
This is an unrelated issue. Please open another separate one or a pull request for it.

holgern:
Ok, i will

fabiangreffrath:
So, could we please remove this obscurity from the SDL2 package?

PR #2434, opened 2017-05-03
qemu: Update to 2.9.0 , add missing dependency, and enable libusb feature again

### LINKED ISSUE #1407: qemu doesn't find dlls
After installing `mingw-w64-x86_64-qemu` and trying to run `qemu-system-x86_64` I receive the following error message:

```
C:/msys64/mingw64/bin/qemu-system-x86_64.exe: error while loading shared libraries: zlib1.dll: cannot open shared object file: No such file or directory
```

despite having a `zlib1.dll` in `C:\msys64\mingw64\bin`. I tried to circumvent the issue by copying the dlls (`zlib1.dll`, `libssh2-1.dll`, ...) to `C:\msys64\usr\bin`. But eventually I reached:

```
C:/msys64/mingw64/bin/qemu-system-x86_64.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

Bobo1239:
Fixes #1407.

Tbh I'm not sure if enabling libusb is correct. The [commit disabling](https://github.com/Alexpux/MINGW-packages/commit/1be509d7aaef5deede76a3632e926ea26ed97799) it only said it was disabled to fix build issues. At least on my setup I can build with it enabled (and also with the other disabled features enabled).

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=qemu
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.8.0
+pkgver=2.9.0
 pkgrel=1
 pkgdesc="A generic and open source processor emulator (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


Bobo1239:
Maybe we also want to re-enable the other features. According to [the wiki](http://wiki.qemu.org/Hosts/W32#Native_builds_with_MSYS2) it should be fine but I don't really understand the features (or rather if they make sense in msys2) so somebody else would have to chip in.

Bobo1239:
Timeout on appveyor during installation, build succeeded.

PR #2444, opened 2017-05-07
Fix to build angleproject using newest git checkout by linking to -lsetupapi

irakhlin:
Fix to build angleproject using newest git checkout by linking to -lsetupapi
Add small patch to build outside of the source directory
Build static libraries after building shared libraries


irakhlin:
I am not sure what exactly i wrong with tea but it seems gyp-git package is broken there but it seems to build fine on appveyor

PR #2452, opened 2017-05-10
mingw-w64-gcc-git: Fix patch 0017 and PKGBUILD.

lhmouse:
1. Fix patch for std::experimental::filesystem for gcc-7-branch.
  2. `git archive` seems not supporting HTTPS. Revert to the GIT protocol.


lhmouse:
@niXman Sorry for the trouble but your patch can't be applied using `git am` so I have to apply it using `curl some-url | patch -p0`, add a commit message saying the author is you, then generate an `am`'able patch using `git format-patch`.

PR #2464, opened 2017-05-14
mingw-w64-gcc: Add a patch to fix the __in/__out name collision.

DavidEGrayson:
In C/C++ code, the names `__in` and `__out` have a special meaning in Microsoft-style code, and a recent commit to mingw-w64 (b7f44b677f43c3bb5f771ff5a430206de040c48c) made those definitions be more visible, causing compilation errors in libstdc++ headers which try to use `__in` and `__out` as variable names.  For what it's worth, this pull request adds a patch to mingw-w64-gcc (but not mingw-w64-gcc-git) which should fix the issue, but the mingw-w64-gcc package is not currently compiling so I was unable to really test this pull request.  I usually don't make a pull request if I can't build the package in question, and I would personally prefer to use `sed` instead of a 3000-line patch, but Alexey asked for this.

### DIFF
--- a/mingw-w64-gcc/0130-libstdc++-in-out.patch
+++ b/mingw-w64-gcc/0130-libstdc++-in-out.patch
@@ -1,0 +1,2982 @@
+Microsoft-style code uses __in and __out to annotate function
+parameters, which causes compilation in libstdc++ headers where they
+try to use __in and __out as variable names.
+
+This patch fixes the issue by adding one more underscore to the
+problematic variable names in libstdc++.
+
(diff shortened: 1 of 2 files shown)


Alexpux:
Add patch 0017 from gcc-git and increase pkgrel

DavidEGrayson:
The mingw-w64 people have decided to comment out `__in` and `__out`, so once they do that, we should remember to remove this huge patch.

bit2shift:
I'm going to try this patch on `mingw-w64-gcc-git`.

PR #2466, opened 2017-05-15
mingw-w64-dlfcn - 1.1.1 - Update to latest version

JPeterMugaas:


mingw-w64-dlfcn/PKGBUILD
```diff
@@ -14,12 +14,12 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
...
 source=(${_realname}-${pkgver}.tar.gz::https://github.com/dlfcn-win32/dlfcn-win32/archive/v${pkgver}.tar.gz
         unify-makefile.patch)
-sha256sums=('36f2e7ef1f1ba04f6ce682a71937eaddd3d6994f09e29df2c7578ec524e47450'
+sha256sums=('31434f96abc6978b80a437f072eebc6255d4f4c66fbd18aa073f3866a02b38a9'
             'e1668cf99f3e41b51be6ff75f48350d761f19d501f7c70a5322b72b7e746bfea')
 
 prepare() {
   cd "${srcdir}/dlfcn-win32-${pkgver}"
-  patch -p1 -i "${srcdir}"/unify-makefile.patch
+#  patch -p1 -i "${srcdir}"/unify-makefile.patch
```
peterbud: If we don't need a patch anymore can we remove it, not just comment out? Also removing the patch file? That would keep the source clean.

PR #2451, opened 2017-05-10
improve vcs protocols

vszakats:
Use encrypted Git protocol for gcc.gnu.org and sourceware.org VCS servers.

PR #2474, opened 2017-05-19
Revert "improve vcs protocols (#2451)"

lhmouse:
Cloning/updating repos through the HTTPS protocol results in Gigabytes of data transferred.

This reverts commit 113e9130aebbc000c6c840fdb41dcdd0e20fa188.


vszakats:
To lower the traffic and make cloning faster, the solution is to use shallow clones (f.e. using the `git clone` option `--depth 8`). Sadly, if you switch to an unprotected protocol for live repos, which have no way to verify them using a checksum/hash, a MITM can alter the downloaded source code (and thus the resulting binaries) as they wish, so it's highly insecure.

In particular the `gcc.gnu.org/git/gcc.git` repo should not have the same HTTPS bugs as `sourceware.org` seems to be having, so it'd be nice to keep the secure protocol and introduce shallow clones instead.

elieux:
> Sadly, if you switch to an unprotected protocol for live repos, which have no way to verify them using a checksum/hash

What do you mean? There's the SHA for the commit.

vszakats:
Unless I'm missing something the `*-git` packages would simple request the tip of a specific branch, and there is no reference to any specific commit or commit hash.

PR #2487, opened 2017-05-20
cppunit: Update to 1.14.0

lazka:
Explicitly disable Werror to prevent the build from failing

In this release cppunit-config was removed in favor of pkg-config.

PR #2509, opened 2017-05-21
cppunit: Updated to version "1.14.0".

stahta01:
And, added make check/check method.

I also had to add a patch to fix build errors.

Tim S.


lazka:
Also see #2487

PR #2512, opened 2017-05-22
portaudio: include static library. Fixes #2317

### LINKED ISSUE #2317: Portaudio: using --enable-static has no effect
If I grep the configure part of the log for static, it says:

$ cat mingw-w64-portaudio-190600_20161030-1-i686-build.log | grep static
checking if i686-w64-mingw32-gcc static flag -static works... yes
checking whether to build static libraries... yes
checking if i686-w64-mingw32-g++ static flag -static works... yes

But in the end no libportaudio.a file, just libportaudio.dll.a and .dll

lazka:
For some reason it either builds a static or a shared one.
Just build it twice and install both.

### DIFF
--- a/mingw-w64-portaudio/PKGBUILD
+++ b/mingw-w64-portaudio/PKGBUILD
@@ -4,28 +4,50 @@ _realname=portaudio
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=190600_20161030
-pkgrel=1
+pkgrel=2
 pkgdesc="A free, cross-platform, open source, audio I/O library (mingw-w64)"
 arch=('any')
 url="http://www.portaudio.com"
 license=("custom")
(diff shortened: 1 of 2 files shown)


lazka:
I don't know why the build fails, it works here. Maybe someone has an idea.

bagong:
I get exactly the same error as appveyor and tea at home... sorry...

lazka:
Good catch, I've removed the skeleton line from the makefile now. Maybe you can try again.

bagong:
That did it for me! Cool!
The tea failure is because of the CI-system, not because of the build.
Thank you so much!!

If you want to add an extra ;) Currently this build is a bit castrated, because the WASAPI backend isn't enabled. There is a fix for that in this PR:

https://app.assembla.com/spaces/portaudio/git/merge_requests/4858113?section=files

If you'd apply that, likely *all* portaudio backends could be enabled and would build (for ASIO this has to be done manually though, because the user has to download a SDK from Steinberg, and Steinberg does not allow that to be done automatically/for redistribution).

lazka:
Cool. I think it's best if you create another PR for this after this one is merged since you already know what needs to be done.

Issue #118, opened 2014-07-31
Error building mingw-w64-lame package


saghul:
https://gist.github.com/saghul/d21fa3788af9f3681caf

Issue #61, opened 2014-07-11
ImageMagick search paths assume C:\ drive


CounterPillow:
ImageMagick appears to be unable to find configuration files and modules if msys2 was installed on a drive different from C:\

Example snippet of output demonstrating the problem when running `convert -debug configure in.png out.jpeg`:

```
2014-07-11T15:48:36+02:00 0:00.002 0.000u 6.8.9 Configure convert.exe[6672]: ../magick/configure.c/GetConfigureOptions/679                                                      /Configure
  Searching for configure file: "C:\msys64\mingw64\share\ImageMagick-6\coder.xml"
2014-07-11T15:48:36+02:00 0:00.002 0.000u 6.8.9 Configure convert.exe[6672]: ../magick/configure.c/GetConfigureOptions/679                                                      /Configure
  Searching for configure file: "C:\msys64\mingw64\etc\ImageMagick-6\coder.xml"
2014-07-11T15:48:36+02:00 0:00.002 0.000u 6.8.9 Configure convert.exe[6672]: ../magick/configure.c/GetConfigureOptions/679                                                      /Configure
  Searching for configure file: "C:\msys64\mingw64\share\doc\ImageMagick-6\coder.xml"
2014-07-11T15:48:36+02:00 0:00.002 0.000u 6.8.9 Configure convert.exe[6672]: ../magick/configure.c/GetConfigureOptions/679                                                      /Configure
```

PR #2514, opened 2017-05-23
CMake tests

Optiligence:
not sure if this is wanted (therefore separately from #2513)
also not sure if this works to merge one after the other, but i’d be curious to find out

### DIFF
--- a/mingw-w64-cmake/PKGBUILD
+++ b/mingw-w64-cmake/PKGBUILD
@@ -5,7 +5,7 @@ _realname=cmake
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.17.3
-pkgrel=1
+pkgrel=2
 pkgdesc="A cross-platform open-source make system (mingw-w64)"
 arch=('any')
 url="https://www.cmake.org/"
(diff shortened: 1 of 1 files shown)


Optiligence:
I’m currently still running these, and apparently some fail, so far: 
#61: ExportImport
#118: CPackWiXGenerator (and some more related to the WiX toolset that i have installed, so removed that for a second run)

so it’s probably not so wise to enforce the tests yet

Alexpux:
Rebase changes

Optiligence:
test output https://pastebin.com/spN9F1xk
i couldn’t fix any of the failing tests yet
some are related to path handling, others to import libraries

peterbud:
I'm happy to see such PRs. I have found also packages which were failing some of the test cases, and those errors were undetected for a longer time simply due to lack of `make check`. I think it's OK to have some test cases fail, but at least they are known and people can work on those.

Optiligence:
I thought failing tests would prevent packages from successfully building. If that isn’t the case then this can be merged right away. Although i’m not sure where the discussion about failing cases should continue then.

mati865:
@Optiligence you can always use `make check | true`

lazka:
Needs a rebase again. Failing tests are fine imo, better then before at least..

PR #2515, opened 2017-05-23
gst-plugins-good: integrate upstream commits to fix 100% CPU issue

virtuald:
See https://bugzilla.gnome.org/show_bug.cgi?id=773681


lazka:
works here 👍

PR #2523, opened 2017-05-29
libtiff: Update to 4.0.8

lazka:
Drop patch included in the new release.
For changes see:
	http://www.simplesystems.org/libtiff/v4.0.8.html


Alexpux:
Rebase on current changes

PR #2531, opened 2017-05-30
Add libwebp as runtime dependency to opencv.

larskanis:
The current binary package "mingw-w64-x86_64-opencv-3.2.0-3-any.pkg.tar.xz"
links to the outdated DLL "libwebp-6.dll" but doesn't specify libwebp as a dependency.

Since libwebp is updated in the meanwhile to "libwebp-7.dll", it isn't sufficient to install the package.

Rebuilding opencv with webp dependency fixes this issue.


larskanis:
I wonder that the integration tests fail on two different issues. On my local Windows-10 machine the build succeeded at the first go and fixed the DLL loading issue.

lazka:
Also see #2529

Alexpux:
Fixed in other pull

larskanis:
@lazka @Alexpux Thank you!

PR #2533, opened 2017-06-01
mingw-w64 headers, tools, crt, libmangle, winstorecompat, winpthreads…

JPeterMugaas:
… - Update to latest git versions and add genstubsdll using a patch.


JPeterMugaas:
I accidentally deleted my branch.  I will make a new request.

PR #2547, opened 2017-06-03
mingw-w64-qt5: bump to Qt 5.9.0 LTS

mcallegari:
Qt 5.9.0 LTS has been released 3 days ago.
I need it for my project so I've updated the package.
I've been able to successfully build a release package (basically what most people need), but I encountered some openGL reference issues in building mapboxgl plugin from qtlocation module.
For now I disabled the whole module, even though it builds fine except for one plugin. If we can't find a solution for it, we'd just need to create a new patch just to disable mapboxgl plugin.

I also adopted the "yes/no flags" style I used in qt5-git so a custom Qt flavour can be built up to single features. Looks cleaner to me rather than "fastbuild" and "minimal"


Alexpux:
I'm already push some changes, so you need rebase your changes on top of my

mcallegari:
Damnit, perfect timing. One afternoon trashed. :cry:

stahta01:
FYI: I am working on fixing the qt5-git package to build using branches 5.6, 5.8, or 5.9.
My way of fixing was to create new patch files. Will likely be requesting a pull in a day or two.
Been working on it for over a week.
Tim S.

Issue #2534, opened 2017-06-02
full rebuild fails for mingw-w64-icu for x86_64


emmenlau:
I'm trying to make a full rebuild of a subset of MINGW-packages. The build fails repeatedly for latest `mingw-w64-icu` for `x86_64` with an `Error 127`:
```
make[2]: Entering directory '/d/Temp/mingw-w64-icu/src/icu/build-x86_64-debug/data'
rm -rf icupkg.inc
make[2]: Leaving directory '/d/Temp/mingw-w64-icu/src/icu/build-x86_64-debug/data'
Unpacking ./in/icudt58l.dat and generating out/tmp/icudata.lst (list of data files)
PATH=../lib:../stubdata:../tools/ctestfw:$PATH  ../bin/icupkg -d ./out/build/icudt58l --list -x \* ./in/icudt58l.dat -o out/tmp/icudata.lst
make[1]: *** [Makefile:493: out/tmp/icudata.lst] Error 127
make[1]: Leaving directory '/d/Temp/mingw-w64-icu/src/icu/build-x86_64-debug/data'
make: *** [Makefile:143: all-recursive] Error 2
==> ERROR: A failure occurred in build().
    Aborting...
```
Please find the build log here:
http://data.biodataanalysis.de/tmp/buildlogs/1496373224-mingw-w64-icu-buildlog.txt

JPeterMugaas:
I'm not seeing that here.  But then again, I am using some stuff the mingw-w64 from an updated package that I made and put in a PR.

lazka:
Same problem here with i686. All the binaries in bin like icupkg fail at the first function call to one of the icu DLLS and the process exists with 127.

Reverting the last update makes it work again.

lazka:
In case someone wants to revert until this is solved, I've opened #2552

JPeterMugaas:
Алексей suggested I add an export indicating the proper system.

https://github.com/Alexpux/MINGW-packages/pull/2553

Well, whoever has the better mousetrap wins.

Alexpux:
@JPeterMugaas I mean that functions in ICU 59 are available only since Win7. This is not about 58 build issue

mati865:
I just build mingw-w64_x86_64-icu 58.2 from fresh pull:
```
[All tests passed successfully...]
Elapsed Time: 00:00:13.943
[...]

PR #2552, opened 2017-06-04
Revert last icu update; breaks build on some machines. See #2534

lazka:
This reverts parts of https://github.com/Alexpux/MINGW-packages/pull/2424


Alexpux:
Well instead reverting need to see what is wrong.

lazka:
Sure, your call. Feel free to close.

Alexpux:
Problem with ICU building is that some functions call missing in msvcrt:
````
_create_locale
_free_locale
````

PR #2554, opened 2017-06-05
Two things (WIP)

ZachBacon:
Added builds for libfilezilla and filezilla. Also provides a small fix for wxwidgets and allowing filezilla to build against it.
Also attempts to fix the static linking issue with wxwidgets wanting to link against shared libs. (Still testing that one out locally)


Alexpux:
I will not accept changes to wxWidgets. We will not link its libraries statically.

ZachBacon:
Like I said it's a wip until I can get the system libraries situated. Because right now the story is this, it's currently trying to link to shared image libraries like libpng in the static build of wxwidgets.
Be patient and I'll have it worked out

PR #2556, opened 2017-06-05
Portaudio: Allow wasapi and wdmks backends to be built, tweak dsound

bagong:
This adds a patch to allow building WASAPI and WDMKS backends, and adds a tweak for DSOUND. Mainly from:

https://github.com/mxe/mxe/commit/f4bbc45682f021948bdaefd9fd476e2a04c4740f

For some reason not all MS headers can be removed, I kept the two that seem to be required in the current MSYS2 environment.

Verified to build 32/64-bit and also when including the ASIOSDK.


bagong:
Could somebody explain me what to do about the Tea error?

mati865:
Tea wasn't updated for a year, ignore it.

Alexpux:
Increase pkgrel

PR #2561, opened 2017-06-07
jsonrpc-glib: new package.

KrullKorg:
For now i disabled introspection due to an error that i didn't solve.


KrullKorg:
Don't merge because i saw an error in package directories.

PR #2573, opened 2017-06-09
openal: Update to 1.18.0

lazka:
* Refresh patches
* Disable building tool which depends on qt5


Alexpux:
I think we don't need to rename libraries as it broke many packages depended on openal

lazka:
I haven't changed any naming with this update.

Alexpux:
Ok, I misunderstood

lazka:
Hm, but the LIBNAME change seems indeed weird, I'll have another look just to be sure.

lazka:
Looks good. It used LIBNAME as the library name in older versions and hardcodes it to OpenAL now. The output name gets explicitly forced to LIBNAME so there is no change in the result.

PR #2574, opened 2017-06-09
pygobject: Update to 3.25.1

lazka:
While this is an unstable release, I've pushed various Windows related
fixes upstream. No need to wait until September for 3.26...


Alexpux:
We build only stable releases of gnome libraries. As I understand they have even number versions.

lazka:
OK, fair enough.

PR #2576, opened 2017-06-09
mingw-w64 packages, crt, headers, libmangle, tools, winpthreads, and …

JPeterMugaas:
…winstorecompat.  Update to latest versions.  Added genstubsdll through patch.


JPeterMugaas:
Now this is going to be a little more difficult since I can't see the "config.log" files so I'm working blind playing on hunches.

martell:
Hey @JPeterMugaas thank you for the contribution.

I wanted to ask what genstubsdll is used for and if this patch can be upstreamed to mingw-w64 instead?

Just one note also if that can not be done.
- can you update the patch to just contain the patch you want for genstubsdll because the maintainer alexey typically uses the pkgver versions of mingw-w64 packages to track the versions he has actually released. I.E it is changed after not before.

JPeterMugaas:
Generate stub-library acting like an import-library
                 using .def file syntax.

And yes, it should be upstreamed and might be.  I sent in a request for this and but kept on having problems sending the diff files.

JPeterMugaas:
It should be upstreamed.  I tried sending a diff to them but that may not worked out as well as I had hoped.  Some issues with blacklisted attachments.  I tried several times.

PR #2599, opened 2017-06-18
python3: Update to 3.6.2rc1

lazka:
This fixes the remaining problems I had with Python 3.6.1 and the faulthandler module.

3.6.2 will be released in two weeks, so if there are no plans to update the repo in the mean time feel free to ignore/close this PR and I will create a new one once the final 3.6.2 is out.


lazka:
Thanks for all the updates! :+1:

PR #2617, opened 2017-06-22
Move to new versions on Bintray

stephengroat:
Just a suggestion, you could use [dynamic downloads](https://bintray.com/docs/api/#_dynamic_download). By having a link pointing to `$latest` for a specific asset and always get the latest version. Current version `latest` isn't deploying because it's static

You could also only deploy on `master` with something like this in `appveyor.yml`:
```
provider:
    on:
        branch: master
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ deploy:
       subject: $(BINTRAY_ACCOUNT)
       repo: $(BINTRAY_REPOSITORY)
       package: $(PACMAN_REPOSITORY_NAME)
-      version: latest
+      version: $(APPVEYOR_BUILD_VERSION)
       publish: true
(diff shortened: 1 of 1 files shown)


jtanx:
Bintray premium only™

stephengroat:
Sorry about that.

[Get Package](https://bintray.com/docs/api/#_get_package) also returns a `latest_version` in it's result

stephengroat:
I realized that since the build scripts currently allow multiple packages to be modified in the same script, the bintray `version` tag would have to be based off of the `APPVEYOR_BUILD_ID` since i'm not 100% sure how you could assign different versions (something like `pkgver-pkgrel` out of `PKGBUILD`) to different artifacts.

PR #2619, opened 2017-06-22
python3: Make multiprocessing.Queue work; Add some tests. Fixes #837

### LINKED ISSUE #837: python3: multiprocessing is not fully supported.
I happen to use the multiprocessing lib in python 3 with this code:

```
from multiprocessing import Pool

def inc(a):
    return a + 1

with Pool(10) as p:
    print(p.map(inc, range(100)))
```

But it fail with this message:

```
ImportError: This platform lacks a functioning sem_open implementation, therefore, the required synchronization primitives needed will not function, see issue 3770.
```

How to solve it? The same program runs well on MSYS python 3.

lazka:
With non-mingw builds POSIX_SEMAPHORES_NOT_ENABLED is not defined and
sem_unlink is a no-op. Change it so that the no-op sem_unlink is also exposed under mingw.

This adds a minimal test suite which makes it easier to catch regressions
in the future for common MSYS2 related bugs.

### DIFF
--- a/mingw-w64-python3/1650-expose-sem_unlink.patch
+++ b/mingw-w64-python3/1650-expose-sem_unlink.patch
@@ -1,0 +1,11 @@
+--- Python-3.6.2rc1/Modules/_multiprocessing/multiprocessing.c.orig	2017-06-22 13:35:38.307651400 +0200
++++ Python-3.6.2rc1/Modules/_multiprocessing/multiprocessing.c	2017-06-22 13:40:12.238319300 +0200
+@@ -128,7 +128,7 @@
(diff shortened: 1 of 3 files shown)


cr1901:
@lazka Does this _only_ make `multiprocessing.Queue` work? I have a need for `multiprocessing.Lock` and while it makes sense that `multiprocessing.Lock` should work, I can't find the commit (if any) that says it works.

lazka:
No idea. I'm not aware of any other problems. Note that most of the msys2 patches afaik just fix assumptions of the code and build system regarding the build environment, so in most cases things just work, or not at all (like in this case where a Python function was just missing)

Issue #2249, opened 2017-03-05
mingw-w64-x86_64-npm package not present


TBBle:
In https://github.com/Alexpux/MINGW-packages/commit/2a7b363e05507b86d8ba520ed6b5a43c7987f732 npm was broken out of nodejs as a separate package.

Today I updated to `mingw-w64-x86_64-nodejs-6.9.5-1`, and it now correctly has an optional dependency on `mingw-w64-x86_64-npm`.

However, there is no `mingw-w64-x86_64-npm` package, even though the source clearly exists: https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-npm

In case the expectation is that either tea-ci or Appveyor will auto-upload new packages to make them available, both failed:
* https://tea-ci.org/Alexpux/MINGW-packages/2447 
````
==> Starting build()...
input is not a tty
==> ERROR: A failure occurred in build().
    Aborting...
````

* https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.2413
````
Deploying using BinTray provider
Uploading "ci.mingw.builds" to BinTray as alexpux/msys2/ci.mingw/latest/ci.mingw.builds...Error uploading artifact to BinTray (403 status): Version 'latest' has been published more than 180 days ago and can no longer be deployed to.
````

mati865:
It doesn't build on real machine for unknown reason.

TBBle:
If that's the apparent hang on the line `node cli.js install ../npm-marked-man-${_marked_man_ver}.tgz` I can confirm that it happens after npm attempts to return success, using `MINGW_INSTALLS=mingw64 makepkg-mingw -sLfC`

mati865:
> If that's the apparent hang on the line node cli.js install ../npm-marked-man-${_marked_man_ver}.tgz 

Workaround for it is to add `sleep 5` before because previous node command doesn't exit fast enough.

More serious problem is here: https://github.com/Alexpux/MINGW-packages/pull/2211#issuecomment-284254626

TBBle:
`sleep 5` didn't work around the problem for me. I didn't play around much harder than that. `sleep 30` also didn't work. It sounds like the problem is more general though.

leachim6:
I'm also having the same issue when building this package

___
 $ MINGW_INSTALLS=mingw64 makepkg-mingw -sLf
==> Making package: mingw-w64-npm 4.3.0-1 (Thu, Mar 23, 2017  4:25:40 AM)
[...]

PR #2622, opened 2017-06-23
nodejs: update to 6.11.0 and include npm

mati865:
It compiled locally but let's see how it will work on CI.


peterbud:
Is this addressing also #2249 ?

mati865:
I hope so but I forgot to commit 1 patch.
I'll test it again and force push.

mati865:
It worked fine with very small project but it would be nice if someone could test it after it is available in repo.

PR #2623, opened 2017-06-23
jasper: Added patch file to fix static library link error.

stahta01:
To use the static library you need to define JAS_DLL to zero.
-D JAS_DLL=0

I am trying to update qt5-git and I need this or another fix to enable linking to an static jasper library.

Tim S.


stahta01:
This is needed to fix the errors that was posted in https://github.com/Alexpux/MINGW-packages/issues/2321

Tim S.

PR #1772, opened 2016-09-30
Openjpeg2

JPeterMugaas:

PR #1372, opened 2016-04-30
Some various things I wanted to add before I forget

JPeterMugaas:

PR #1312, opened 2016-04-12
Master

JPeterMugaas:
Some new additions, and updates.

PR #2633, opened 2017-06-26
cmake and some updated dependencies

JPeterMugaas:
mingw-w64-cmake - 3.8.2 - Update to latest release version and remove 0002-bootstrap-Set-DEFAULT_CODEPAGE-in-cmake_cxx_flags_En.patch .  That patch is not needed for bootstrap
mingw-w64-expat - 2.2.1 - Update to latest version, add testing as part of build process.  Add 002-fix-tests.patch so those can work in MSYS environment.  remove expat-2.2.0-CVE-2016-0718-regression.patch as that is in the source-code.
mingw-w64-lncruses - 6.0 - 20170624 - Update to latest version


mati865:
Expat is already updated, rebase this PR.

JPeterMugaas:
I did not realize that until it was too late.  But what I'm trying to do with that new version is add some testing.

mati865:
Merge commit created a little mess here.

peterbud:
Don't take it personal, I'm just wondering when you can figure out your workflow:
#1312
#1372
#1772
#1955

JPeterMugaas:
I don't take this your criticism personally.  To be honest, I'm not sure I have a good workflow down.  I have a fork, and clone on my computer, then pull from Alexpuk's branch using "Rebase" like I'm supposed to using SourceFree.  Most of the time, this works.  But a merge conflict frequently causes me problems because of rebase and it seems like whatever fix I do, doesn't seem to help sometimes.  I can't git directly most of the time.  I just want what needs to be done, done.

With MSYS-packages, I can't even work with that anymore because SourceFree creates modifications that are unstaged and get in my way.

https://github.com/Alexpux/MSYS2-packages/issues/929

peterbud:
If that helps, here is my workflow which was working fine so far always:

1. Prerequisite: have a fork
2. Set up remote upstream pointing to the original repo (https://github.com/Alexpux/MINGW-package), as described [here](https://help.github.com/articles/configuring-a-remote-for-a-fork/)
3. When you have something to do/change **always** create a separate branch from your local master branch ( I see you do this as well)
4. From time to time pull remote upstream, and do a _local merge_ on your master branch. As you local master branch has no local changes this will be always a fast forward, and it won't add any additional commits and there won't be any merge conflict. There is no need to do any voodoo magic with `git rebase`

As front-end I'm using [Git for Windows](https://git-for-windows.github.io/), and it just works fine, both the GUI and the Git Bash as well.

PR #2637, opened 2017-06-27
Fix some missing -git suffixes in some git PKGBUILDS

lazka:
Instead of building the same name as as the non-git variant
build a -git variant and provide/conflict the normal one.
As is the case with other git packages.

### DIFF
--- a/mingw-w64-gst-rtsp-server-git/PKGBUILD
+++ b/mingw-w64-gst-rtsp-server-git/PKGBUILD
@@ -1,10 +1,10 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
 _realname=gst-rtsp-server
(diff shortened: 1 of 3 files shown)


mingw-w64-gst-rtsp-server-git/PKGBUILD
```diff
@@ -24,6 +24,8 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-gst-plugins-good-git"
          "${MINGW_PACKAGE_PREFIX}-gst-plugins-ugly-git"
          "${MINGW_PACKAGE_PREFIX}-gst-plugins-bad-git")
```
mati865: Remove `-git` from gst packages.

PR #2646, opened 2017-06-30
qt5: Fixed static library build issues.

stahta01:
By adding configure_opts for static build.
Static build uses "-no-icu", "-dbus-linked", and "-qt-freetype".
Added defines for static build of:
  JAS_DLL=0
  DBUS_STATIC_BUILD
Added library configuration info of:
  DBUS_LIBS=-ldbus-1 -lws2_32 -liphlpapi
And, added shared guard around export of "PKG_CONFIG".

Only had enough hard disk space to test building the 64 bit static release without docs or examples.
But, it built and installed without issues.
It took me many hours to figure out the export "PKG_CONFIG" was causing the static build issue.
Tim S.


Alexpux:
This  pull request not fix issue with jasper linking

PR #2647, opened 2017-06-30
clang: update to 4.0.1

mati865:
I have disabled libunwind as per https://github.com/Alexpux/MINGW-packages/pull/2294#issuecomment-310940009
CC @martell

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -21,7 +21,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-llvm"
          "${MINGW_PACKAGE_PREFIX}-lldb"
          "${MINGW_PACKAGE_PREFIX}-polly")
-pkgver=4.0.0
+pkgver=4.0.1
 pkgrel=1
 pkgdesc="C language family frontend for LLVM (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


Alexpux:
Dont disable it. You can just not install it but do not disable

mati865:
Okay, libc++abi is broken anyway so we can let it build with libunwind.

wirx6:
Clang/llvm can already perform self-build so I think we should set `_compiler` to `clang`.
It will make future patches more relevant because libcxx/cxxabi/unwind are used with clang anyway.

mati865:
Clang cannot build 64 bit libc++ until mingw-w64 gets fixed.

paniq:
I got an error trying to build this patch because it appears the public keys have changed
```
    llvm-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    cfe-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    compiler-rt-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    test-suite-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    libcxx-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    libcxxabi-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    clang-tools-extra-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    lld-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    lldb-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    libunwind-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
    polly-4.0.1.src.tar.xz ... FAILED (unknown public key 8F0871F202119294)
```

mati865:
@paniq you have to either import this key or compile with `--skippgpcheck`.

There is no compiler error in your log, maybe out of memory issue.
CI reached 77% before reaching it's 1.5h qouta https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.3224

mati865:
I'm working on 5.0.0 version.

PR #2650, opened 2017-07-01
some vulkan stuff

wirx6:
GCC 7.1.0 is required for compiling vcc.


mingw-w64-glslang/001-install-missing-dll.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/StandAlone/CMakeLists.txt b/StandAlone/CMakeLists.txt
+index 2a9a3c47..e3d46595 100644
```
Suppe1213: 1

Alexpux:
vpp-git not building for 32 bit

PR #2658, opened 2017-07-03
gcc: actually apply the patches

wiiaboo:
I'm getting the _ITM_RU1 relocation errors again from https://github.com/Alexpux/MINGW-packages/issues/1580
Wasn't the weak references in 64-bit patch merged upstream already?

Edit: (^ It was added to master the same day of the 7.1 release, so no.)


wiiaboo:
The patches probably need rebasing, now that makepkg is actually applying them.

PR #2664, opened 2017-07-04
mingw-w64-x264-git: adding nasm>=2.13 to makedepends

emmenlau:
`mingw-w64-x264-git` fails to build with error:
```
==> Starting build()...
Found no assembler
Minimum version is nasm-2.13
If you really want to compile without asm, configure with --disable-asm.
    Aborting...
```
Added nasm>=2.13 to makedepends. I'm not sure if it still requires `yasm` so I left yasm in makedepends.


Alexpux:
Use mingw version of nasm

Gramner:
yasm is no longer a required dependency since it's replaced by nasm.

PR #2665, opened 2017-07-04
waf: Add clang_compilation_database tool

jhasse:
This is quite useful when using clang for linting purposes.


Alexpux:
Increase pkgrel

PR #2671, opened 2017-07-05
Add freetds build script

larskanis:
This adds a build script for [FreeTDS](http://www.freetds.org/). I'll open a pull request for the required patch.

I tested the package by using `tsql.exe` (which is part of the resulting package) and by building and running [tiny_tds](https://github.com/rails-sqlserver/tiny_tds) against the installed libraries.

### DIFF
--- a/mingw-w64-freetds/0001-Don-t-use-MSYS2-file-libws2_32.patch
+++ b/mingw-w64-freetds/0001-Don-t-use-MSYS2-file-libws2_32.patch
@@ -1,0 +1,41 @@
+From 56e8972f66c3e948e2ad6885595c58fd23dcdb37 Mon Sep 17 00:00:00 2001
+From: Lars Kanis <kanis@comcard.de>
+Date: Thu, 6 Jul 2017 17:09:40 +0200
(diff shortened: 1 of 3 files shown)


larskanis:
I opened a PR for the patch here: https://github.com/FreeTDS/freetds/issues/132

Unfortunately the build fails on i686. I'll investigate this.

larskanis:
Now that the build succeeds, this could be merged from by point of view, now. Some additional thoughts, however:

1. I'm unsure, whether the patch to `configure` should be reported to autoconf. It mixes MSYS2 files into MINGW. But I'm not an autotools expert by no means.
2. I've opened https://github.com/FreeTDS/freetds/pull/133 for the `inet_pton()` patch.
3. Currently both static and shared libraries are built and packaged. Is there a policy what packages should ship?

Alexpux:
Looking to your patch  I see you define ````_WIN32_WINNT 0x601````, so it will work only on Win7 or higher

isapir:
How can I install `freetds`?  Running `pacman -S freetds` returns `error: target not found: freetds`.

`pacman -Syu` says that everything is up to date and there is nothing to do.

PR #2675, opened 2017-07-07
python3: Use is_cygpty() to detect a terminal; disable readline with …

lazka:
…a real Windows console

CPython uses isatty() to detect a terminal and change some settings
like line buffering and interactive mode. Use is_cygpty() to make
this also work under mintty.
See https://github.com/Alexpux/MINGW-packages/issues/2645

This also removes the bash script which forced the interactive mode
when python3 was started without arguments. This is no longer needed as
Python now detects the terminal output and does this automatically.

Also use is_cygpty() to detect when not under mintty and disable the readline
module there, as using it breaks input of certain characters and
leads to errors on shutdown when it tries to save the readline history.
(The readline module is not available in the official Python build)
See https://github.com/Alexpux/MINGW-packages/issues/2656


lazka:
rebased on master

Issue #2681, opened 2017-07-09
mingw-w64-libmariadbclient is missing mysql_config script


jeroen:
I would like to build `gdal` package `--with-mysql`. However this requires a `mysql_config` or `mariadb_config` executable. Trying to point it to the `libmariadbclient` directory, the gdal configure script errors with:

```
configure: error: --with-mysql argument is a directory.  It should be the path to the mysql_config 
  script, often somewhere like /mingw64/bin/mysql_config.
```

Is there some way to make `mingw-w64-libmariadbclient` produce the `mysql_config` script? It seems this should because it is also included in e.g. the [ubuntu libmariadbclient](https://packages.ubuntu.com/zesty/amd64/libmariadbclient-dev/filelist) package.

stahta01:
Can you apply a patch to the libmariadbclient package and build and test it?
If yes, I am willing to try to create a patch.

Tim S.

stahta01:
[patch](https://github.com/stahta01/stahta01.github.io/blob/master/msys2/patches/libmariadbclient-Add-install-of-mariadb_config.exe.patch)

Tim S.

jeroen:
Thanks! Perhaps make this a pull request?

stahta01:
Did it test OK? 
Because, I have no idea how to test the patch to confirm it works.
All I know is it builds, installs, and the exe gives output. No idea if the output is valid.

Tim S.

stahta01:
#2687 Pull request.

Tim S.

PR #2687, opened 2017-07-13
libmariadbclient: Add install of "mariadb_config.exe".

stahta01:
Possible fix for #2681 
I do NOT know enough about this package to say if my fix works or is the correct fix.
I have asked @jeroen to verify the fix works.

Tim S.

### DIFF
--- a/mingw-w64-libmariadbclient/PKGBUILD
+++ b/mingw-w64-libmariadbclient/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libmariadbclient
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.3.3
-pkgrel=1
+pkgrel=2
 pkgdesc="MariaDB client libraries (mingw-w64)"
 arch=('any')
 url="http://mariadb.org"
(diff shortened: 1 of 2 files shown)


jeroen:
Yes this works! It builds and installs and gives working cflags and libs! Thanks!

There is one minor annoyance that in addition to the correct paths it also returns a non-existing `-L/lib` and `-I/include`. Because these do not exist they simply get skipped by the preprocessor/linker so it's harmless. 

But perhaps we can tweak the `PKGBUILD` to properly set `${MINGW_PREFIX}` in `CFLAGS` and `LDFLAGS`?

```sh
Jeroen@MINGW64 ~/mingw-packages/mingw-w64-libmariadbclient
$ mariadb_config --libs
-L/lib/mariadb -lmariadb

Jeroen@MINGW64 ~/mingw-packages/mingw-w64-libmariadbclient
$ mariadb_config --cflags
-I/include/mariadb -I/include/mariadb/mysql -march=x86-64 -mtune=generic -O2 -pipe -Wunused -Wno-uninitialized

```

stahta01:
I do NOT have the time or energy to look for the -L/lib and -I/include issue.

Tim S.

jeroen:
OK I think this can be merged as is!

jeroen:
OK I figured it out! All we need to do is add this line to the `PKGBUILD`:

```
-DPREFIX_INSTALL_DIR=${MINGW_PREFIX} \
```

This macro is used in side `mariadb_config.exe`. @stahta01  do you want to add this to your PR or should I make a new PR?

stahta01:
Why don't you make a new pull. Post in this one after you do and I will close it then.

Tim S.

jeroen:
@stahta01 OK I have created a [new PR](https://github.com/Alexpux/MINGW-packages/pull/2702). Thanks again for your help!

PR #2690, opened 2017-07-14
xapian-core: Update to 1.4.4

lazka:


ojwb:
`0001-exp10-build-fix.patch` should no longer be required with 1.4.4.

PR #2697, opened 2017-07-14
gstreamer: Update to 1.12.2

filnet:


bagong:
In the current msys2 environment gstreamer, gst-editing-services and libav build fine (32 and 64 bit). In plugins-bad I get the following errors (building in the msys2 shell):

https://gist.github.com/bagong/e81ceb3e1af99be693413d986e9cc2c6

bagong:
plugins-good and plugins-ugly build fine.

PR #2702, opened 2017-07-18
Build mariadb_config.exe in libmariadbclient

jeroen:
This adds a working `mariadb_config.exe` which is needed by some other packages to get the proper cflags and libs to use this driver.


jeroen:
This is ready to merge :rocket:

PR #2705, opened 2017-07-19
pango: Update to 1.40.9

lazka:


Ede123:
@lazka - I'm currently investigating an issue with font substitution in Inkscape and have two questions:
* Does `pango-view.exe` work for you?
* Do you have an idea why the font "Sans" could use some strange serif font (instead of Arial or similar?)

lazka:
1) I don't know what pango-view should do, but `pango-view.exe mytext.txt -o foo.png` gives me a png with the rendered text.

2) I've never used anything besides the "standard" font in gtk+, no idea. I'd be happy to test things.

Ede123:
`pango-view` is supposed to show the rendered text in a Window as far as I understand, but it fails to show a Window for me. Rendering the output to .png is fine for my (testing) purposes though.

Using `pango-view.exe --text="The quick brown fox jumps over the lazy dog. 0123456789" --font="$fontname 50" -o out.png` I was able to determine that the bug is likely on Inkscape's side.<br/>
Using fontame `Sans`, `sans-serif` and even `sans` all result in the same sans-serif font ("Tahoma" on my system" being used).<br/>
In Inkscape I get a strange serif font (i.e. the font substitution for the generic family name is not working; it seems a bit as if "Sans" would be a different font in Inksacape as all fonts that can't be loaded fall back to the same serif font, too.

Ede123:
Just for completeness: I figured out the issue with Inkscape. The difference in Inkscape (compared to pango-view) is that it uses the fontconfig backend (not the native win32 backend) and due to a stupid error in the install scripts the directory `etc/fonts/conf.d` was not installed which made the font substitution fail.

PR #2706, opened 2017-07-19
gdk-pixbuf2: Update to 2.36.8

lazka:
Also drop a bogus patch.
The commit said "The function always returned NULL - now fixed", but
(1) that's not the case
(2) returning a random unrelated pointer wont help..


Ede123:
The patch `gdk_pixbuf_loader_get_pixbuf.patch` in question was discussed as (off-topic) part of #2376 (doesn't have anything to do with the gdiplus loaders).

lazka:
@Ede123 Thanks. I guess removing the patch is fine with you then?

peterbud:
It's OK to remove that patch, there is no need for that, it was my mistake.

juml:
Appears there is an regression in this particular version (ICO loader), ref: https://bugzilla.gnome.org/show_bug.cgi?id=785447, fixed in trunk.

Not however sure if affects gdip-ico loader from mingw package too.

lazka:
Thanks. Let's wait for the next release then.

lazka:
Updated to 2.36.8 which should include a fix: https://git.gnome.org/browse/gdk-pixbuf/tree/NEWS

PR #2711, opened 2017-07-20
vala: Update to 0.36.4

lazka:


lazka:
Not sure why this fails, ideas welcome.

mati865:
On AV it failed to generate docs but it worked on Tea with very outdated packages, I think the issue is in doc generating tools (missing rebuild?).

You can try disabling docs as workaround.

lazka:
I don't see any option to disable it (could patch I guess..)

I get the same error as tea-ci at the point where appveyor fails: "warning: failed to load external entity "../../../vala-0.36.4/doc/manual/version.xml"

As far as I see it calls xslt-proc.exe there, which is part of libxslt and installed in both cases.

I'll reopen if I figure something out.

PR #2727, opened 2017-07-28
hdf5 updated to 1.8.19

holgern:
Download link fixed
version set to 1.8.19
recipe quality improved


holgern:
32 bit works now :)
I will investigate why 64 bit fails...

holgern:
fixed :)

PR #2731, opened 2017-07-30
openblas: update to 2.20

holgern:
NUM_THREADS set to number of cores using nproc


mingw-w64-openblas/PKGBUILD
```diff
@@ -34,7 +35,7 @@ build() {
   make LIBNAME=libopenblasp.a NO_LAPACK=0 DYNAMIC_ARCH=1          \
        MAJOR_VERSION=3 NO_AFFINITY=1 CC=gcc FC=gfortran           \
        OPENBLAS_INCLUDE_DIR=${MINGW_PREFIX}/include/${_realname}  \
-       USE_THREAD=1 NUM_THREADS=64
+       USE_THREAD=1 NUM_THREADS=$(nproc)
```
mingwandroid: This is incorrect. `NUM_THREADS` should be a large constant number because it is the number of threads that can maximally be launched and limiting that to the number of processors on the build machine is wrong. Think about running on a very high-end server.

PR #2743, opened 2017-08-04
gnupg: Update to 2.1.22 Fixes #1778

### LINKED ISSUE #1778: GnuPG patches disable user profile based home directories
I'm trying to use MINGW packages for GnuPG deployments to multi-user machines.
I'd like GnuPG to keep its per user home directory files (keyrings, trust databases, etc) somewhere under users' profiles like GnuPG's default home directory (`%APPDATA%\Roaming\GnuPG` as explained in [FAQ](https://gnupg.org/faq/gnupg-faq.html#location_gpg_conf_file)).
Unfortunately, the MINGW GnuPG sets the home directory to `/home/` which all users would have to share.

```
$ gpg.exe -k -vv
gpg: NOTE: THIS IS A DEVELOPMENT VERSION!
gpg: It is only intended for test purposes and should NOT be
gpg: used in a production environment or with production keys!
gpg: keybox 'C:/Open/git/mingw64/home/pubring.kbx' created
gpg: C:/Open/git/mingw64/home/trustdb.gpg: trustdb created
gpg: using pgp trust model
```

The comments for the `--homedir` section of [the manual](https://gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html) explain this is the _portable_ home directory location, which can be enabled by creating 

> an empty file name gpgconf.ctl in the same directory as the tool gpgconf.exe

However, no such file exists, so portable mode enabled this way.
Instead, the comments for the [MINGW GnuPG patches](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-gnupg)

> Mingw always generate portable apps.

suggest patches were applied to _lock_ portable mode, leaving the user no way to undo it without recompilation.
Rather than lock the application into portability mode with patches, could you consider creating a blank gpgconf.ctl as the `--homedir` section of [the manual](https://gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html) recommends?

peterbud:
* HOME directory is now located at ~/.gnupg (`/home/<username>`)
* All basic tests are now running smoothly (21+4+1+2). Before this only the first 19 test cases were successful.
* There are tests which are using `TinySCHEME `scripts. I don't know this system, and some of the existing test scripts are failing, but I was not able to debug and fix them. Any help would be appreciated here.


Alexpux:
Why you remove all patches?

peterbud:
Apologies not answering earlier. Most of those old patches are not needed anymore with the new version of `gnupg`. I have now one new patch only which contains all necessary modifications on the top of the released `gnupg`.
Let me also call out once more: the current/original version of `gnupg` completes only the first 19 test cases, and then stops. With the changes which I have included here 28 test cases are successfully completing.

Alexpux:
@peterbud yesterday I'm upload gnupg 2.1.23. All old patches are applied. If you think that you have better patch then create new pull request for review

PR #2744, opened 2017-08-04
libidn: patch for gcc7 compatibilty added

holgern:


holgern:
I have no idea why the sha256sum did not pass for 0005-gcc7-compatibility.patch.
It passed on my pc.

jtanx:
I'd say calculating the sha256 after the fact defeats the purpose...

holgern:
Ok, after deleting and git revert, i am getting also 
0a826c91ff7b847574e1bc986289770caf1e7525ad41f9f7dfe6ef64d526f305

PR #2746, opened 2017-08-05
Firebird2-git: fix build

holgern:
fix Package


holgern:
works only on older gcc

PR #2750, opened 2017-08-07
Add pcre2

vszakats:


vszakats:
Git 2.14.0 now supports and recommends pcre2:
https://github.com/git/git/blob/master/Documentation/RelNotes/2.14.0.txt#L25
https://github.com/Homebrew/homebrew-core/commit/080c62c6196a2fa23994c04c8bd51bc29232d99d
It's also faster:
https://lists.exim.org/lurker/message/20170419.172322.833ee099.en.html

Issue #271, opened 2014-10-19
Provide mingw versions of msys2 packages where possible


elieux:
I know mingw-git is in the works, but I'd like to see subversion and other versioning systems, coreutils, patch/diff and others. I'd like to have these accessible (and working) in cmd without putting `msys/usr/bin` in my PATH.

gnuwin32 could help if it's not too outdated.

elieux:
Referencing: wget #569, diffutils #571, grep #553, sed #552.

elieux:
Quoting @mingwandroid from #552:

> perhaps we need the PKGBUILD format to include "makeconflicts", an array of packages that must not be installed for makepkg* to succeed?
> Qt can't be built if mingw-w64-perl is installed for example.

elieux:
How about using the shell script wrappers, as we do for tty-impaired mingw apps? For example, for sed, we create `/mingw??/bin/sed` with this contents:
```
#!/usr/bin/bash
/usr/bin/sed "$@"
```
That should prevent any interference from these new mingw programs. But I'm in doubt as to whether it's sustainable.

PR #2752, opened 2017-08-08
Update mingw-w64-file package

larskanis:
This updates mingw-w64-file util and library to 5.31.

The resulting package is currently missing in the repository, so my hope is, that this PR will trigger a rebuild and upload.


lazka:
My guess why it's not in the repo is potential conflicts with the msys2 version, see #271, but not sure...

larskanis:
My main reason to use the MINGW package is libmagic - not the "file" executable. Maybe adding a mingw-w64-libmagic package which excludes the file "file" ?

PR #2762, opened 2017-08-11
qtbinpatcher: Fix wrong install dir

holgern:
The install dir was mingw-w64-qtbinpatcher and not mingw-w64-i686-qtbinpatcher or mingw-w64-x86_64-qtbinpatcher
This patch fixes this. The variable ${pkgdir} seems to be overwritten and set to ${pkgbase}

I have no idea why this happens with this packages whereas using ${pkgdir} works fine on other packages...


Alexpux:
Fixed by https://github.com/Alexpux/MINGW-packages/commit/e87a973df0abc09f9e385bcaedc06c7009ca9a36

PR #2773, opened 2017-08-12
cego + lfcbase

bjlemke:
version update
lfcbase : Update to version 1.10.0
cego : Update to version 2.34.0


bjlemke:
Compile fails because of package dependencies lfcbase -> cego

PR #2774, opened 2017-08-13
mingw-w64-perl - 5.26.0 - Update to latest version.  Move build to pa…

JPeterMugaas:
…ckage since pkgdir is not the same value in build() as it is in package().  add and use helper macros.

mingw-w64-lcov - 1.13.\ - Update to latest version and use helper macros.

### DIFF
--- a/mingw-w64-cmake/0004-Implement-Qt5-static-plugin-support.patch
+++ b/mingw-w64-cmake/0004-Implement-Qt5-static-plugin-support.patch
@@ -35,23 +35,22 @@ index f11825083e..e8a07361c0 100644
             !(*ti)->GetPropertyAsBool("AUTORCC")) ||
            (*ti)->IsImported()) {
          continue;
-diff --git a/Source/cmQtAutoGeneratorInitializer.cxx b/Source/cmQtAutoGeneratorInitializer.cxx
-index 825eba0e1d..61652b80b1 100644
---- a/Source/cmQtAutoGeneratorInitializer.cxx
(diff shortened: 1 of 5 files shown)


peterbud:
My latest information was that the `mingw64` perl is broken, and we should use the msys2 version. Is this version working better? How have you tested?

JPeterMugaas:
Unfortunately, it still is broken although I'm still not sure why.  I would also point out that this package does interfere with some gtk build stuff and it has to do with autoconf and intltool.  My thinking is this, you probably will need two separate sets of perl libraries (or three if have both i686 and x86_64).  Then you could have a mingw-perl set of libraries along with the msys2 version.  Unfortunately, this seems have a bad habit of accessing the msys perl libraries and that's not desirable.

What I did was attempt to help matters by updating to the latest version of perl.  That distribution was based on 5.22 which is outdated.  Right now, I'm working on a 5.26 version of perl for msys2 and so far so good.

mingwandroid:
Perl 5.26 changed the handling of @INC (see http://blogs.perl.org/users/sawyer_x/2017/05/perl-5260-is-now-available.html)

As you can probably guess this breaks a *lot* of things. We need to update to automake 1.15.1 for example. I ran into a few more but the details elude me (it *probably* broke git).

JPeterMugaas:
I have gone ahead and made a MSYS2 version of Perl 5.26 at https://github.com/Alexpux/MSYS2-packages/pull/971 .

PR #2782, opened 2017-08-15
vala: Update to 0.36.4 [broken]

lazka:
Already had this open as #2711 which doesn't allow me to reopen. Still not sure what's the problem, maybe it solved itself in the meantime.


mati865:
Okay, I'm sure it's broken https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-libxslt/PKGBUILD#L70

lazka:
Uh, weird, why doesn't it fail on my machine then?

mati865:
Maybe because you have 32 and 64 bit packages installed at same time.
CI removes 64 bit packages after 64 bit build is done and downloads 32 bit packages for 32 bit build.

lazka:
Ah, makes sense. Thanks for helping!

Alexpux:
Now remove debug info

PR #2789, opened 2017-08-16
pango: Update to 1.40.10

lazka:
The new tarballs are created with meson so we have to run autogen.sh
and the tarballs no longer contain the built documentation.

### DIFF
--- a/mingw-w64-pango/PKGBUILD
+++ b/mingw-w64-pango/PKGBUILD
@@ -4,7 +4,7 @@
 _realname=pango
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.40.9
+pkgver=1.40.10
 pkgrel=1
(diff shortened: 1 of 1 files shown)


jtanx:
urgh meson sounds like crap

lazka:
What makes you say that?

lazka:
Just from my limited experience meson is a lot faster then autotools and easier to understand. And it will let us get rid of all cygwin build dependencies which seems to be in line with the goals of MSYS2.

Regarding the two issues I've mentioned in the commit:

* Upstream gnome is currently dropping documentation from the tarballs. This is related to the meson switch in that they didn't bother to reimplement doc generation there. I didn't try to create the docs here as I'm waiting for the new gtk-doc release which is ported from bash/perl to Python and will allow us to use a mingw version. Once that's out I'll try to enable doc generation/port things.
* For the configure file missing from the tarball. Again related to meson, but we were recreating the autotools files anyway, so it doesn't really matter.

I hope that helps.

PR #2790, opened 2017-08-16
python-pytest: Update to 3.2.1

lazka:


lazka:
Builds fails because the dependency added in #2742 is not in the repo.

PR #2795, opened 2017-08-17
Upgrade Git LFS to version 2.2.1

mfld-fr:
Also updated the patch for MSYS / MINGW environment.
Built and tested on MSYS2 / Win7.

Note : the documentation is still not generated, as it depends on `ronn` that is installed with a `gem` command line and not through a clean PKGBUILD.


mfld-fr:
Build in Tea is failed because it is still using Go version 1.4.
I is successful in AppVeyor with Go version 1.8.
Known problem...

PR #2747, opened 2017-08-06
firebird2-git: still broken

holgern:

Issue #2536, opened 2017-06-02
full rebuild fails for mingw-w64-firebird2-git for x86_64


emmenlau:
I'm trying to make a full rebuild of a subset of MINGW-packages. The build fails repeatedly for latest `mingw-w64-firebird2-git` for `x86_64` with error:
```
../src/common/classes/alloc.cpp: In function 'void* operator new(std::size_t)':
../src/common/classes/alloc.cpp:2127:7: error: declaration of 'void* operator new(std::size_t) throw (std::bad_alloc)' has a different exception specifier
 void* operator new(std::size_t s) THROW_BAD_ALLOC
       ^~~~~~~~
In file included from ../src/common/classes/alloc.cpp:32:0:
../src/include/../common/classes/alloc.h:502:7: note: from previous declaration 'void* operator new(std::size_t)'
 void* operator new(std::size_t s) THROW_BAD_ALLOC;
       ^~~~~~~~
../src/common/classes/alloc.cpp: In function 'void* operator new [](std::size_t)':
../src/common/classes/alloc.cpp:2131:7: error: declaration of 'void* operator new [](std::size_t) throw (std::bad_alloc)' has a different exception specifier
 void* operator new[](std::size_t s) THROW_BAD_ALLOC
       ^~~~~~~~
In file included from ../src/common/classes/alloc.cpp:32:0:
../src/include/../common/classes/alloc.h:503:7: note: from previous declaration 'void* operator new [](std::size_t)'
 void* operator new[](std::size_t s) THROW_BAD_ALLOC;
       ^~~~~~~~
make[3]: *** [../gen/make.rules:112: ../temp/boot/common/classes/alloc.o] Error 1
make[3]: Leaving directory '/d/Temp/mingw-w64-firebird2-git/src/build-x86_64-w64-mingw32/gen'
make[2]: *** [Makefile:233: gpre_boot] Error 2
make[2]: Leaving directory '/d/Temp/mingw-w64-firebird2-git/src/build-x86_64-w64-mingw32/gen'
make[1]: *** [Makefile:227: ../gen/firebird/bin/gpre_static.exe] Error 2
make[1]: Leaving directory '/d/Temp/mingw-w64-firebird2-git/src/build-x86_64-w64-mingw32/gen'
make: *** [Makefile:6: firebird] Error 2
==> ERROR: A failure occurred in build().
    Aborting...
```
Please find the build log here:
[...]

PR #2797, opened 2017-08-18
Firebird2-git

stahta01:
I am mainly an C Programmer working to better my C++ knowledge.
I have not tested the package; because I have no idea how to test it.

Used compiler option "-std=gnu++11" because I could not figure out how to fix the build error below.
Changed package to options=('debug' 'strip') because '!debug' is causing error in running create_db.exe empty.fdb during build.

See also #2747 and #2536

Error # 1 fixed by 0030-Use-nullptr.patch
    ../src/jrd/gds.cpp:278:7: error: conflicting declaration 'const UCHAR zero []'
      zero[]  = { op_line, 0 },
           ^
    In file included from ../src/jrd/../jrd/../jrd/val.h:32:0,
                     from ../src/jrd/../jrd/jrd.h:39,
                     from ../src/jrd/gds.cpp:151:
    ../src/jrd/../jrd/../jrd/../include/fb_blk.h:115:15: note: previous declaration as 'void* zero'
     static void * zero = (void *) 0;
                   ^~~~

Error # 2 work around is 0031-Fix-compiler-flags.patch "-std=gnu++11".
    ../gen/firebird/bin/gpre_current.exe -n -z -gds_cxx -raw -ids ../src/jrd/dpm.epp ../temp/boot/jrd/dpm.cpp
    gpre version NP-V2.5.8.27070 Firebird 2.5
    make[3]: *** [../gen/make.rules:80: ../temp/boot/jrd/dpm.cpp] Error 127

Error # 3 work around is doing debug option.
    rm -f empty.fdb
    ../gen/firebird/bin/create_db.exe empty.fdb
    make[2]: *** [../gen/Makefile.refDatabases:68: empty.fdb] Segmentation fault
    make[2]: *** Deleting file 'empty.fdb'

### DIFF
--- a/mingw-w64-firebird2-git/0030-Use-nullptr.patch
+++ b/mingw-w64-firebird2-git/0030-Use-nullptr.patch
@@ -1,0 +1,52 @@
+From 2adb6fa680b81e7f95429a6041f399d91d2c160a Mon Sep 17 00:00:00 2001
+From: Tim S <stahta01@users.sourceforge.net>
+Date: Wed, 16 Aug 2017 22:38:24 -0400
+Subject: [PATCH] Use nullptr instead of zero.
+
+---
+ src/include/fb_blk.h | 12 +++++-------
+ 1 file changed, 5 insertions(+), 7 deletions(-)
+
+diff --git a/src/include/fb_blk.h b/src/include/fb_blk.h
+index 4929c525c4..b2b8a1b6cc 100644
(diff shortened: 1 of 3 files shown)


mati865:
If non debug builds error on runtime that means there is most likely bug caused by optimizations.
If you want you can try with `-O1`, `-O0` and `-Os`.

PR #2798, opened 2017-08-18
cmake: remove obsolete cm_codecvt.patch

Optiligence:
was fixed for 3.9.0 and disabled since then anyway
https://gitlab.kitware.com/cmake/cmake/issues/16893


stahta01:
See #2803

PR #2806, opened 2017-08-19
[DO NOT MERGE] python2: Use is_cygpty() to detect a terminal; disable readline with …

lazka:
…a real Windows console

This is the same patch as https://github.com/Alexpux/MINGW-packages/pull/2675
adjusted to work with Python 2.7

### DIFF
--- a/mingw-w64-python2/2700-cygpty-isatty-disable-readline.patch
+++ b/mingw-w64-python2/2700-cygpty-isatty-disable-readline.patch
@@ -1,0 +1,429 @@
+diff -Nur Python-2.7.13.orig/Include/iscygpty.h Python-2.7.13/Include/iscygpty.h
+--- Python-2.7.13.orig/Include/iscygpty.h	1970-01-01 01:00:00.000000000 +0100
++++ Python-2.7.13/Include/iscygpty.h	2017-08-19 15:50:22.584217400 +0200
+@@ -0,0 +1,41 @@
++/*
++ * iscygpty.h -- part of ptycheck
++ * https://github.com/k-takata/ptycheck
++ *
++ * Copyright (c) 2015-2017 K.Takata
++ *
++ * You can redistribute it and/or modify it under the terms of either
++ * the MIT license (as described below) or the Vim license.
++ *
++ * Permission is hereby granted, free of charge, to any person obtaining
++ * a copy of this software and associated documentation files (the
++ * "Software"), to deal in the Software without restriction, including
++ * without limitation the rights to use, copy, modify, merge, publish,
++ * distribute, sublicense, and/or sell copies of the Software, and to
++ * permit persons to whom the Software is furnished to do so, subject to
++ * the following conditions:
++ *
++ * The above copyright notice and this permission notice shall be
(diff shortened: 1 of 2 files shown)


lazka:
Noticed that line buffering isn't enabled, needs more work.

lazka:
hm, ok..

lazka:
From what I see Python 2, unlike Python 3, doesn't change the buffering unless in unbuffered or interactive mode, so the missing line buffering is because mintty doesn't do it compared to cmd.exe. It would be possible to force unbuffered mode when mintty is detected (in Modules/main.c).... but Python 2 is on its way out and I don't care that much.

manisandro:
Hi
I've built a Fedora mingw-python2 package [1] which is basically this work here packaged as a RPM. I'm investigating an issue where stdout (and stderr) simply aren't printed to the console by python2.exe, but I see the outputs if I redirect them to a file. And python2.exe does not run in interactive mode. Is this issue here related to this? Also with `2700-cygpty-isatty-disable-readline.patch` I see no output on the console.

Thanks and apologies for potentially hijacking.

[1] https://src.fedoraproject.org/rpms/mingw-python2/tree/master

lazka:
Which type of console?

manisandro:
Standard Windows cmd.

lazka:
This patch should only change things with cygwin terminals. No idea what could cause your problem, sorry :/

The msys2 Python2 seems to work fine here in cmd.exe

lazka:
The only thing I can think of is that in case the readline module gets build in your build, try deleting it.

manisandro:
Okay thanks, I'll dig further then.

lazka:
In case your build includes the readline module, try deleting it.

mati865:
@manisandro I suppose Python has proper `fflush()` usage but you can try disabling stdout and stderr buffering anyway.

manisandro:
Hm I now build without readline, and indeed the `lib-dynload/readline.pyd` is not built anymore, but I still don't see anything, even if I start as `python2.exe -u`.

Ede123:
@lazka I just noticed this patch broke Windows XP support. Is it supposed to work on Windows as-is? Or do we need to set some compiler flags? It seems you put quite some amount of effort into supporting older versions of Windows without `GetFileInformationByHandleEx` so I hope it will be an easy fix...

lazka:
@Ede123 The code is a straight copy of https://github.com/k-takata/ptycheck According to the readme it should be possible to run on winxp with some defines. I'll have a look.

Ede123:
Thanks! I guess `USE_DYNFILEID` is what we probably want then?

lazka:
@Ede123 see #3187

lazka:
@Ede123 can you test the package: https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.4350/job/djsvrib29g5dhsd2/artifacts ?

Ede123:
Functionality restored on Windows XP, thank you @lazka!

lazka:
Thanks for testing, and sorry for the trouble

PR #2833, opened 2017-08-29
mbedtls: Updated to 2.6.0

nazar554:
https://tls.mbed.org/tech-updates/releases/mbedtls-2.6.0-2.1.9-and-1.3.21-released


jay:
:+1: As noted in that link there is an important vulnerability that was fixed, see https://tls.mbed.org/tech-updates/security-advisories/mbedtls-security-advisory-2017-02

PR #2839, opened 2017-09-02
updated mingw-w64-pkgconf added as git package

irakhlin:
Updated package, does not interfere with pkgconfig as its building right now. Creates a fully static little binary and seems to run pretty well.


stahta01:
The URL does not work for me.
url='https://github.com/nenolod/pkgconf'

Tim S.

irakhlin:
Good catch, I though I updated that earlier but must've not saved it before the commit

Alexpux:
I dont want this package as git

irakhlin:
The most recent release of pkgconf needs an additional patch that I have also made but have not created into a pull request. Would you consider an update of the current mingw-w64-pkgconf to the newest working version in a state that does not replace pkg-config at first to see if it can eventually replace pkg-config?
From my testing so far it seems to be much quicker and less error prone in addition to providing backward compatibility and added features like "--static --pure" and CFLAGS.private?

Link for reference:
http://pkgconf.org/features.html

Alexpux:
Updating current pkgcong is better than create new one

PR #2845, opened 2017-09-05
Fix the SFML build to work with the latest cmake

smarthall:
Currently SFML installs `FindSFML.cmake` in `/mingw32/share/cmake-3.7/Modules/FindSFML.cmake` and cmake can't find it. This moves it to `/mingw32/share/cmake-3.9/Modules/FindSFML.cmake`.

My build directory includes spaces so I've added quotes to support that, feel free to cherry pick that commit out, or let me know if you'd like it in a separate PR.


Alexpux:
Increase pkgrel. Build direcrtories with spaces can provide more issues, so try to avoid this in any case

smarthall:
I'm currently reinstalling things so that I don't have paths with spaces. I thought that better quoting couldn't hurt though.

I submitted the change to bump `pkgrel`, thanks for the tip!

PR #2847, opened 2017-09-05
clang: update to 4.0.1 und fix building

holgern:
fix compilation

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -73,7 +73,7 @@ source=(https://llvm.org/releases/${pkgver}/llvm-${pkgver}.src.tar.xz{,.sig}
         "0401-mingw-w64-hack-and-slash-fixes-for-libc.patch"
         "0501-minor-fixes-to-get-lldb-building.patch"
         "0502-hack-to-use-64-bit-time-for-mingw.patch"
-		"0503-fix-compile-error-TaskPool.patch"
+        "0503-fix-compile-error-TaskPool.patch"
         "0601-libunwind-add-support-for-mingw-w64.patch"
         "0701-fix-visibility-for-libc-abi-on-win32.patch")
 # Some patch notes :)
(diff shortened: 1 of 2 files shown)


mingw-w64-clang/PKGBUILD
```diff
@@ -73,6 +73,7 @@ source=(https://llvm.org/releases/${pkgver}/llvm-${pkgver}.src.tar.xz{,.sig}
         "0401-mingw-w64-hack-and-slash-fixes-for-libc.patch"
         "0501-minor-fixes-to-get-lldb-building.patch"
         "0502-hack-to-use-64-bit-time-for-mingw.patch"
+		"0503-fix-compile-error-TaskPool.patch"
```
Alexpux: Use spaces instead tabs

mingw-w64-clang/PKGBUILD
```diff
@@ -247,8 +250,9 @@ build() {
   # https://github.com/martine/ninja/issues/932
 
   if [ "${_compiler}" == "gcc" ]; then
-    export CC='gcc'
-    export CXX='g++'
+    msg "using gcc as compiler"
+    #export CC='gcc'
```
Alexpux: Why you disable this exports? I think without this by default will be clang used if it installed already

mingw-w64-clang/PKGBUILD
```diff
@@ -521,4 +525,4 @@ package_mingw-w64-x86_64-llvm(){
 
 package_mingw-w64-x86_64-polly(){
   package_polly
-}
+}
```
Alexpux: Add new line at the end

mingw-w64-clang/PKGBUILD
mati865: It was just a fail-safe.

mati865:
Add this https://github.com/Alexpux/MINGW-packages/pull/2647/files#diff-f8431e29bf404d2b6e837a3e6a7e6b00R339

mingw-w64-clang/PKGBUILD
holgern: The package does not build using makepkg-mingw in the msys2 terminal.
But it builds completely when doing make in the mingw32 terminal.
So i think, that the wrong gcc from msys2 is used.
removing both export lines fixes this.

mati865: @holgern this is wrong, it **has** to be build with `makepkg-mingw` in msys2 shell.

holgern:
export CC='gcc'
    export CXX='g++'
assure that gcc is used for compiling.
Then unset both variables prevent that a wrong gcc is used in the make process.

PR #2851, opened 2017-09-06
qbs: update to 1.9.0

holgern:


Alexpux:
Wrong update. Result import libraries have bad names

holgern:
I am working on a fix.
The library names are correct, but there is a problem with getting the current qbs version

Alexpux:
Ok

Issue #2784, opened 2017-08-15
Parallel Build for Boost


GyrosGeier:
Boost uses its own build system, which doesn't respect MAKEFLAGS, so it is difficult to pass a -j`nproc` to it, while here it would make a lot of sense.

It would be nice to have a way for parallel builds to work, e.g. by isolating a -j option from MAKEFLAGS and passing it on to b2.

mingwandroid:
This sounds like something you should be able to do with bash to me?

holgern:
I implemted it here: #2866

MehdiChinoune:
Fixed by #2866

GyrosGeier:
Definitely an improvement, but now you can't turn it off. :/

PR #2866, opened 2017-09-07
boost: parallel build implemented

holgern:


Alexpux:
Remove coreutils from dependencies.

holgern:
This solves #2784

PR #2869, opened 2017-09-09
nim: Updated to version 0.15.0

stahta01:
I have never used nim; and, I have no idea whether nimble the nim package manager is needed to use nim.
I tried and failed to get the nimble package to build; it needs someone who knows how to tell it where to find the system.nim file. Tim S.


stahta01:
Looks like my changes must have a problem; since the AppVeyor build failed. I will test on another computer and see if it works there.
Tim S.

Alexpux:
I'm update nim to 0.17.2

Issue #904, opened 2015-11-22
aspell segmentation fault on building dictionary


ncihnegn:
Shall I downgrade winpthreads?

==> Starting build()...
Finding Dictionary file location ... /mingw64/lib/aspell-0.60
Finding Data file location ... /mingw64/lib/aspell-0.60
/mingw64/bin/prezip-bin -d < en-common.cwl | /mingw64/bin/aspell  --lang=en create master ./en-common.rws
/mingw64/bin/prezip-bin -d < en-variant_0.cwl | /mingw64/bin/aspell  --lang=en create master ./en-variant_0.rws
/mingw64/bin/prezip-bin -d < en-variant_1.cwl | /mingw64/bin/aspell  --lang=en create master ./en-variant_1.rws
/mingw64/bin/prezip-bin -d < en-variant_2.cwl | /mingw64/bin/aspell  --lang=en create master ./en-variant_2.rws
/mingw64/bin/prezip-bin -d < en-w_accents-only.cwl | /mingw64/bin/aspell  --lang=en create master ./en-w_accents-only.rws
/mingw64/bin/prezip-bin -d < en-wo_accents-only.cwl | /mingw64/bin/aspell  --lang=en create master ./en-wo_accents-only.rws
/mingw64/bin/prezip-bin -d < en_CA-variant_0.cwl | /mingw64/bin/aspell  --lang=en create master ./en_CA-variant_0.rws
/mingw64/bin/prezip-bin -d < en_CA-variant_1.cwl | /mingw64/bin/aspell  --lang=en create master ./en_CA-variant_1.rws
/mingw64/bin/prezip-bin -d < en_CA-w_accents-only.cwl | /mingw64/bin/aspell  --lang=en create master ./en_CA-w_accents-only.rws
/mingw64/bin/prezip-bin -d < en_CA-wo_accents-only.cwl | /mingw64/bin/aspell  --lang=en create master ./en_CA-wo_accents-only.rws
/bin/sh: line 1:  1660 Done                    /mingw64/bin/prezip-bin -d < en-wo_accents-only.cwl
      7388 Segmentation fault      | /mingw64/bin/aspell --lang=en create master ./en-wo_accents-only.rws
Makefile:108: recipe for target 'en-wo_accents-only.rws' failed
make: **\* [en-wo_accents-only.rws] Error 139
make: **\* Waiting for unfinished jobs....
[...]

PR #2872, opened 2017-09-09
Aspell: update to 0.60.7-rc1 and fix segmentation fault

Ede123:
This allows to build with gcc 7 and fixes a segmentation fault.

(see previous description and patch by @jtanx for details)

<s>Unfortunately Aspell is still incompatible with recent versions of winpthreads making it SEGFAULT frequently, see #904 (and links in that bug).

Downgrading winpthreads temporarily is a workaround in some cases but is not really possible in others (e.g. I just realized that Inkscape, which is linked against the shared Aspell library, crashes *every* time on exit due to this issue).

Any help with this would be highly appreciated!

(P.S. In principle this could be merged, but as long as no fix is found there's probably not much reason to package a release candidate at this point; just wanted to make it available in case somebody want to try to build it)</s>

### DIFF
--- a/mingw-w64-aspell/0008-cache-Attempt-to-prevent-premature-destruction-of-gl.patch
+++ b/mingw-w64-aspell/0008-cache-Attempt-to-prevent-premature-destruction-of-gl.patch
@@ -1,0 +1,67 @@
+From 208e7e93466d02b583012029a60ad958b1eb234b Mon Sep 17 00:00:00 2001
+From: Jeremy Tan <jtanx@outlook.com>
+Date: Sun, 10 Sep 2017 13:55:41 +1000
+Subject: [PATCH] cache: Attempt to prevent premature destruction of global
+ cache lock
+
+https://stackoverflow.com/questions/335369/finding-c-static-initialization-order-problems#335746
+---
+ common/cache-t.hpp | 1 +
+ common/cache.cpp   | 8 ++++----
+ 2 files changed, 5 insertions(+), 4 deletions(-)
+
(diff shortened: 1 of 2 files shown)


jtanx:
I've taken a bit of a look, and it's because of this 'global cache lock':

https://github.com/GNUAspell/aspell/blob/b2df46185757361947303877c88ceba4392b0d63/common/cache.cpp#L8

If you set a breakpoint here:

https://github.com/GNUAspell/aspell/blob/b2df46185757361947303877c88ceba4392b0d63/common/lock.hpp#L38

You'll find that `global_cache_lock` is destroyed _before_ `GlobalCacheBase::~GlobalCacheBase` is called, which attempts to lock/unlock on `global_cache_lock`, probably resulting in some corrupted state in winpthreads.

Some seemingly relevant info:
https://stackoverflow.com/questions/75701/what-happens-to-global-variables-declared-in-a-dll

jtanx:
Works for me: https://github.com/jtanx/MINGW-packages/commit/cc617ef09e93a812546f1f8f7c49862c5523916d

Ede123:
Great!

I just confirmed the SEGFAULTs I was experiencing in Inkscape are gone. Also commands that could made Aspell segfault before seem to work now (`aspell dump config data-dir`) and building  mingw-w64-aspell-en works flawless, too.

I added your patch to this PR (hope that's OK). Do you think it's fine to propose this upstream, too?

jtanx:
Glad it works! I've made a PR upsteam as GNUAspell/aspell#532.

Ede123:
Great, thanks!

PR #2878, opened 2017-09-12
Make pkgbase of all packages match the name of the PKGBUILD directory

lazka:
This either changes the pkgbase to match the directory name
or the other way around.

The motivation for this is to make it possible to automatically generate
an URL to the PKGBUILD file from the package information alone.

This makes it possible to link to the package source from the web interface:
http://quodlibet.duckdns.org/msys2/base/mingw-w64-gtk3


lazka:
I haven't changed the pkgrel for the packages where pkgbase was changed as previous commits doing similar things didn't do it either. I can bump the pkgrel if wanted.

lazka:
Thanks!

filnet:
I have commented the commit. Looks like two of changes will break things or will not fulfill the stated goal.

PR #2885, opened 2017-09-13
Update kf5 libs to 5.38.0

zrax:
Update existing kf5 framework libraries to 5.38.0, including extra-cmake-modules (ECM).
Also adds some additional kf5 libs:
* KDocTools (Needed for building KJS)
* KSyntaxHighlighting
* ThreadWeaver

I may add more in the future if these end up working.


zrax:
FYI, the CI builds will always fail without internal/clever dependency resolution (ECM must be updated/installed before anything else, the Tier1 frameworks must be built/installed before T2, etc).  I did test all of these builds locally (shared qt5 only), but admittedly I did not test each with a clean sysroot.

PR #2895, opened 2017-09-15
json-glib: updated to 1.4.2.

KrullBorg:


lazka:
Note that building without introspection breaks building things depending on it, like clutter.

lazka:
ah, meson is broken with gi-scanner. I'll look into that

lazka:
See #2942

PR #2898, opened 2017-09-15
gdal: add crypto++ and netcdf dependencies

alex85k:
When testnig https://github.com/osmcode/libosmium with MSYS2, noticed that current gdal binaries were broken (asking missing dlls like libcryptopp.dll and libnetcdf.dll). Added the missing libraries to dependencies, but the package also needs rebuilding with new Poppler (currently it wants libpoppler-67 dll with libpoppler-70 present).
Postgresql client is also needed to run .exe-s , but currently postgres is a monolyt package (server+client) and I do not know if it should be a mandatory dependency.

### DIFF
--- a/mingw-w64-gdal/PKGBUILD
+++ b/mingw-w64-gdal/PKGBUILD
@@ -8,7 +8,7 @@ _realname=gdal
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.2.1
-pkgrel=1
+pkgrel=2
 pkgdesc="A translator library for raster geospatial data formats (mingw-w64)"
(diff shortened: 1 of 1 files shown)


Alexpux:
GDAL-2.1.1 builded with popper 0.59. Update packages

alex85k:
Thank you for super-fast reaction) I confirm that poppler dll problem is fixed.
I tried removing netcdf package `pacman -R mingw-w64-x86_64-netcdf`` and the dll error returned, so adding two dependencies is still needed.

alex85k:
Very strange: the error in tea CI is `mcpl_vsil_crypt.cpp:79:26: fatal error: cryptopp/dll.h: No such file or directory`,  but it had successfully installed and found libcrypto++. Appveyor build seems to pass.

PR #2900, opened 2017-09-15
glib2: updated to 2.54.0.

KrullBorg:
Before build you must update the gtk-doc package

### DIFF
--- a/mingw-w64-glib2/PKGBUILD
+++ b/mingw-w64-glib2/PKGBUILD
@@ -6,7 +6,7 @@
 _realname=glib2
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.52.3
+pkgver=2.54.0
 pkgrel=1
 url="https://www.gtk.org/"
 arch=('any')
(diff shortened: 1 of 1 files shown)


Alexpux:
i try to build GLIB2 and it have conflicts with files from other package seems. Did you try to build and install it yourself?

KrullBorg:
yes i built it and installed it without problems; what conflicts?

Alexpux:
glib-genmarshal

Alexpux:
problem seems that in previous version builded ````glib-genmarshal.exe```` and now without extension

Ede123:
You might want to hold of on that update because of
https://bugzilla.gnome.org/show_bug.cgi?id=787648

(glibmm 2.54.0 accidentally requires C++14)

mati865:
@Ede123 GCC is used to build packages (with gnu++14 standard as default). libc++ is broken so much it cannot be used with MinGW even for hello world programs.
We can ignore linked issue.

Ede123:
@mati865 It breaks the Inkscape build on ubuntu 17.10.1, see the [build log](https://launchpadlibrarian.net/337093458/buildlog_ubuntu-artful-amd64.inkscape-trunk_1%3A0.92.0+devel+16028+91~ubuntu17.10.1_BUILDING.txt.gz) which (if I'm not completely mistaken) uses gcc to build (Inkscape defines `-std=c++11`).

mati865:
@Ede123 you are right, looks like everything would have to switch to c++14.

lazka:
>problem seems that in previous version builded glib-genmarshal.exe and now without extension

While looking into fixing mingw gtk-doc I've used the following to create exe launchers for python scripts: https://bpaste.net/show/f2160a91e230

maybe that's also useful here

holgern:
Then we have to switch to an old version for gobject-introspection.
The actual version depends on glib 2.54.0.

Ede123:
glibmm 2.54.1 (which removes the C++14 dependency) has been released in the meantime (see https://bugzilla.gnome.org/show_bug.cgi?id=787648#c36), so we can update to that version.

Alexpux:
Ok

Ede123:
@Alexpux Do you want a PR for 2.54.1 (we now have 2.54.0) or will you update the pkgver yourself?

Alexpux:
you can create pr

lazka:
@Ede123 glib and glibmm are two different projects

Ede123:
Good catch... I think nobody noticed I was telling nonsense all this time (maybe Alex did). :-D

I somehow assumed this was about glibmm from the start (probably got tricked by the synced version numbers)  and saved that part in my brain... Thanks for playing along everybody (and sorry for the confusion). ;-)

PR #2911, opened 2017-09-16
gtk-doc: Make mingw build work

lazka:
With 1.26 gtk-doc was ported to Python2/3 so we can use it with mingw Python 3.

Regarding the patches:

* Some path replacements for paths leaking from the autotools build system
* Since some other tools like meson expect that the gtkdoc utils are
  executables the pyscript2exe script generates a launcher exe for all
  utils. This is similar to what setuptools does for console entry points.
* Various Windows fixes, don't depend on the locale encoding when opening
  files and don't depend on the shell for executing commands.
  I'll try to upstream these changes once this gets more testing.

I've only tested building gexiv2 with this, larger packages might expose
more bugs.

Regarding meson: Remove the hack to invoke gtkdoc through msys perl, it's no
longer needed when mingw gtk-doc is installed. As there are no packages using
meson+gtkdoc right now this shouldn't break anything.


lazka:
Any thoughts on this welcome.

PR #2912, opened 2017-09-16
gtk2: consistently use g_stat and GStatBuf

Ede123:
This resolves some serious breakages on Windows because
* field widths of "struct stat" are different for different variants of "stat" which can cause undefined behavior.
* Windows' "stat" expects filenames in proper encoding while gtk simply feeds it UTF-8 in many places which is bound to fail.

See [upstream report](https://bugzilla.gnome.org/show_bug.cgi?id=787772) for details.

Let me know if I overlooked anything.


peterbud:
@Ede123 : is this also a solution for [this](https://gnunet.org/sorting-out-stat-thing)? Is this a working solution for 32 and 64-bit as well?

Ede123:
@peterbud Yes, at least that's how it's advertised in [the documentation](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-stat).

If I understand correctly the idea is to [always use a struct with 32-bit time and size fields](https://github.com/GNOME/glib/blob/master/glib/gstdio.h#L30) and make sure a suitable (wide char) variant of the stat function is used.

If this still causes issues with respect to the article you posted it can probably be considered as an upstream bug and should be fixed.
With my work on the Inkscape project I can confirm however that the glib wrappers seem to have worked properly for 32-bit MinGW (TDM-GCC), stock mingw-w64 (64-bit), and also continue to work for both 32-bit and 64-bit flavors of MSYS2's mingw-w64 toolchains (at the very least I've never seen the cited compiler warning and didn't find any unexpected program behavior).

peterbud:
Thanks @Ede123, super news!

PR #2913, opened 2017-09-17
New package: gnuplot

peterbud:
This is a first version of the package using the mingw make file included in the package but this makefile has some limitations. Will look whether the default makefile can produce a better result

### DIFF
--- a/mingw-w64-gnuplot/PKGBUILD
+++ b/mingw-w64-gnuplot/PKGBUILD
@@ -29,7 +29,7 @@ options=('strip' 'staticlibs')
 source=("http://downloads.sourceforge.net/sourceforge/$_realname/$_realname-$pkgver.tar.gz"
         "01-gnuplot.patch" )
 sha256sums=('7dfe6425a1a6b9349b1fb42dae46b2e52833b13e807a78a613024d6a99541e43'
-            'b33a5c8be87a1c16045794865cb118c3d90e5d51b8284c7437be07d0c3b11199')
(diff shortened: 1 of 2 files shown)


peterbud:
Hmm, it was working in the CI [here](https://ci.appveyor.com/project/peterbud/mingw-packages/build/1.0.85)?

mingw-w64-gnuplot/PKGBUILD
```diff
@@ -0,0 +1,86 @@
...
+
+prepare() {
+  cd "$srcdir"/${_realname}-${pkgver}
+  patch -p1 -i ${srcdir}/01-gnuplot.patch
+}
+
+build() {
+  cd "$srcdir"/${_realname}-${pkgver}
+  [[ -d "${srcdir}"/build-${CARCH} ]] && rm -rf "${srcdir}"/build-${CARCH}
+  #mkdir -p "${srcdir}"/build-${CARCH} && cd "${srcdir}"/build-${CARCH}
```
mati865: Isn't is necessary?

mingw-w64-gnuplot/PKGBUILD
```diff
@@ -0,0 +1,86 @@
+# Maintainer: Some One <some.one@some.email.com>
```
mati865: Fix header.

mingw-w64-gnuplot/PKGBUILD
peterbud: I could not use mingw makefile in the package to build in a separate directory. Im working on the default makefile.am to make this work properly

PR #2925, opened 2017-09-19
plplot: Added to makedepends and changed defines passed to cmake.

stahta01:
Added "patch" and "bsdtar>=3.3" to make depends.
Added this define "ENABLE_qt=OFF" and
changed the "ENABLE_f95=ON" define to "ENABLE_fortran=ON".
Also, added paths and libname to help find the qhull lib.

I have not really found anything except the make depends that should stop the 64 bit build.
Note: It fails to find libgd or an good FORTRAN compiler on my computers.

Tim S.


stahta01:
Changing to Python3 resulted in the FORTRAN compiler being found as good. I hope the 64 bit build problem on the CI Travis is also fixed. Tim S.

mati865:
> I hope the 64 bit build problem on the CI Travis is also fixed.

We don't have Travis, if you mean Tea you can completely ignore it.

stahta01:
I meant AppVeyor instead of Tea fails to build 64 bit.
I have found out Python 2 work on 32 bit and Python 3 works on 64 bit build.
I am going to close this PR and try to find the correct Python option for this package.

Tim S.

PR #2927, opened 2017-09-19
Switch some packages to build with mingw-w64-gtk-doc

lazka:


lazka:
Might be that it needs to the new gtk-doc version that's not yet in the repo. I'll wait for that first.

lazka:
I've removed the one failing package for now. Should be good to merge.

PR #2945, opened 2017-09-23
 wxWidgets: Added provides="mingw-w64-wxmsw3.0".

stahta01:
wxWidgets: Added and Applying "003-wxWidgets-3.0.2-fix-access-sample.patch".
  Added "patch" and "make" to makedepends.
  Added to check() build samples.

wxWidgets: Deleted no longer used patch files. 

Tim S.


Alexpux:
There are no need to add ````patch```` and other msys2 packages to makedepends. Wiki says that every developer that want to make packages need install ````base-devel```` group of packages.

stahta01:
I removed the "patch" and "make" from makedepends. And, pushed the commits, again.

Tim S.

PR #2948, opened 2017-09-23
Wget 1.19.1 update

irakhlin:
1. Updated wget to 1.19.1
2. Added patch for support of windows cert store
3. Added patch for taskbar progress using cmd
4. Minor changes to PKGBUILD to fix gettext issue after running autoconf


irakhlin:
One of the depends should really be:
"${MINGW_PACKAGE_PREFIX}-libidn2" instead of "${MINGW_PACKAGE_PREFIX}-libidn" however for some reason appveyor was not finding the dependency "${MINGW_PACKAGE_PREFIX}-libidn2" as an available package..  I have just temporarily changed it to get the build running and make sure it works worth wise. I had committed mingw-w64-libidn2 a bit over a month ago and it seems to be available though the repo via pacman; does appveyor/tea pull their dependencies from somewhere else?

lazka:
tea is broken, ignore it. libidn2 should work on appveyor.

irakhlin:
This package was build with openssl before,  I am still building it with openssl so it will try to use the openssl certificate store file but if it cant be found it will try to use the windows certificate store. I thought this would be useful as its specifically the mingw-* package so I would imagine people may use them outside the msys2 environment and not have the same files setup to have openssl certificate store  in place.

PR #2950, opened 2017-09-23
libgd: Update to 2.2.5

peterbud:
Also fixing the pkg-config and gdlib-config, after the lib and headers were moved back from their subfolder to the /lib and /include folder, the .PC and the gdlib-config scripts were still pointing to the subfolders - this is now also fixed.

### DIFF
--- a/mingw-w64-libgd/libgd-export-pkg-config-build-in-subdirs.patch
+++ b/mingw-w64-libgd/libgd-export-pkg-config-build-in-subdirs.patch
@@ -352,3 +352,18 @@ diff -aur libgd-2.2.4/config/gdlib-config.in.orig libgd-2.2.4/config/gdlib-confi
  	echo "libdir:     $libdir"
  	echo "features:   @FEATURES@"
  	;;
+--- libgd-2.2.4/src/gd.h.orig	2017-08-30 04:05:54.000000000 -0700
++++ libgd-2.2.4/src/gd.h	2017-09-23 11:49:10.071934900 -0700
+@@ -63,7 +63,11 @@
+ #   define BGD_EXPORT_DATA_PROT __declspec(dllimport)
+ #  endif
+ # endif
+-# define BGD_STDCALL __stdcall
++# if !defined(__MINGW32__)
++#  define BGD_STDCALL __stdcall
++# else
++#  define BGD_STDCALL
++# endif
+ # define BGD_EXPORT_DATA_IMPL
+ #else
+ # if defined(__GNUC__) || defined(__clang__)
--- a/mingw-w64-libgd/libgd-export-pkg-config-build-in-subdirs.patch
+++ b/mingw-w64-libgd/libgd-export-pkg-config-build-in-subdirs.patch
@@ -99,7 +99,7 @@ diff -ur libgd-2.2.3/config/gdlib.pc.in libgd-2.2.3-new/config/gdlib.pc.in
 +prefix=@CMAKE_INSTALL_PREFIX@
 +exec_prefix=${prefix}
 +libdir=${prefix}/lib
-+includedir=${prefix}/include/libgd-@GD_VERSION_STRING@
++includedir=${prefix}/include/
  
  Name: gd
  Description: GD graphics library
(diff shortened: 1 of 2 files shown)


peterbud:
Also I have realized that the **32-bit** version of the `libgd` library is using C-style mangled exports. Has to fix that as well, patch comes soon.

mingw-w64-libgd/libgd-export-pkg-config-build-in-subdirs.patch
```diff
@@ -346,9 +346,24 @@ diff -aur libgd-2.2.4/config/gdlib-config.in.orig libgd-2.2.4/config/gdlib-confi
...
  	echo "features:   @FEATURES@"
  	;;
+--- libgd-2.2.4/src/gd.h.orig	2017-08-30 04:05:54.000000000 -0700
++++ libgd-2.2.4/src/gd.h	2017-09-23 11:49:10.071934900 -0700
+@@ -63,7 +63,11 @@
+ #   define BGD_EXPORT_DATA_PROT __declspec(dllimport)
+ #  endif
+ # endif
+-# define BGD_STDCALL __stdcall
++# if !defined(__MINGW32__)
```
Alexpux: I dont think that this is right patch. For 32-bit dlls mingw always need use __stdcall

peterbud: I'm no experienced with this to be honest. What would be the best way to avoid mangling?
The symptom is the following:
```
$ nm -C --defined-only -g /mingw32/lib/libgd.dll.a | grep gdImageCreateTrueColor
00000000 I _imp__gdImageCreateTrueColor@8
00000000 T gdImageCreateTrueColor@8
```
while 
```
$ nm -C --defined-only -g /mingw64/lib/libgd.dll.a | grep gdImageCreateTrueColor
00000000 I _imp__gdImageCreateTrueColor
00000000 T gdImageCreateTrueColor
```
That means for example that software using libgd can nicely link with a 64-bit version, but fail to link to the 32-bit version. Like here when I have tried with `gnuplot`:
64-bit version: [works fine](https://ci.appveyor.com/project/peterbud/mingw-packages/build/1.0.90#L810)
32-bit version [fails](https://ci.appveyor.com/project/peterbud/mingw-packages/build/1.0.90#L2095)

Alexpux: Output of ````nm```` is ok in both cases. You can verify on any other libraries.Problem maybe with passing right defines to link with libgd

peterbud: @Alexpux , are you sure on this? I have checked a few others:
ibcurl - no mangling:
```
$ nm -C --defined-only -g /mingw32/lib/libcurl.dll.a | grep curl_free
00000000 I _imp__curl_free
00000000 T curl_free
```
libz - no mangling:
```
$ nm -C --defined-only -g /mingw32/lib/libz.dll.a | grep deflateCopy
00000000 I _imp__deflateCopy
00000000 T deflateCopy
```
libpng - no mangling:
```
$ nm -C --defined-only -g /mingw32/lib/libpng.dll.a | grep png_image_free
00000000 I _imp__png_image_free
00000000 T png_image_free
```
Or maybe I' doing something wrong here?

peterbud: Also doing [further research](http://www.willus.com/mingw/yongweiwu_stdcall.html) confirms that `__stdcall` **will** decorate/mangle the function names.
What is your suggestion to use:
pass `-Wl, --kill-at` ? 
pass `-Wl,--add-stdcall-alias` ?
Or something else?

Alexpux: I will look some later.

peterbud: @Alexpux thanks for merging. Just to close out the previous dilemma (and to know what is the best approach for the future): how we should handle similar cases in the future?

PR #2953, opened 2017-09-23
protobuf: update to 3.4.1

holgern:
add missing dep
disable tests as building does not work


ArekPiekarz:
I'm just curious, how can you know everything works correctly in protobuf now if you disabled the tests? Wouldn't it be more beneficial to fix them instead? I thought that was one of the requirement of accepting a pull request here - that someone uploaded a tested, working version of software.

PR #2979, opened 2017-09-26
Upgrade podofo to 0.9.5

Dalan94:
It's only an upgrade of the podofo library


Dalan94:
Great sourceforge is down so the builds failed

Issue #2164, opened 2017-02-15
gtk3 native windows theme


Ede123:
Is gtk3 supposed to offer a native Windows theme?

This is what the "gtk3-demo-application.exe" about screen looks out-of-the-box for me on Windows 10 (i.e. it clearly uses the Adwaita theme):
![gtk3_windows10](https://cloud.githubusercontent.com/assets/904570/22955495/b595b048-f31c-11e6-98e0-b768b3422396.png)

Setting `gtk-theme-name=win32` in `etc\gtk-3.0\settings.ini` doesn't improve upon that all (remember I'm on Windows 10, so the theme is completely wrong):
![gtk3_windows10_win32-theme](https://cloud.githubusercontent.com/assets/904570/22955639/6ed2052a-f31d-11e6-8b62-be7a6a129a54.png)

Did the Gnome-people break native Windows theming or is there a packaging issue?

lazka:
Two things:

1) gtk3 never had a proper windows theme, some work went into improving it for 3.20/22. I'm using Adwaita (the default) on Windows and no one ever complained about it.
2) gtk3 for some reason defaults to client side window decorations, which means it draws the window border manually. You can disable them by setting the GTK_CSD env var to 0. They still get used if the program forces them by putting widgets in them etc.. Someone should probably file a bug report upstream to change the default behavior.

Ede123:
Wow, thank you so much! 2. is solid gold!
The non-matching window borders and title bar were the main reason for me to look into this. I can live if some controls looking different, but obviously the windows itself shouldn't deviate too much from the system default.

Also since everything else can be changed in CSS if necessary this basically would allow to make the rest at least look native, but transparent borders and stuff is probably nothing that can be recreated easily.

Ede123:
With `GTK_CSD=0` it already looks OK
![zwischenablage01](https://cloud.githubusercontent.com/assets/904570/23035817/bb4f1514-f480-11e6-80e1-ee1307e5cbdb.png)

With `GTK_CSD=0` and `gtk-theme-name=win32` it could pass as native
![zwischenablage02](https://cloud.githubusercontent.com/assets/904570/23035839/d4e11540-f480-11e6-9847-6a42316b6891.png)

Ede123:
Well, I filed [a bug](https://bugzilla.gnome.org/show_bug.cgi?id=778791) in the GNOME bugtracker, let's see if something comes from it.

[...]

PR #2981, opened 2017-09-28
gtk3: Don't force enable client side decoration

lazka:
So that for example gtk3-demo-application.exe uses the default
Windows window decorations.

### DIFF
--- a/mingw-w64-gtk3/0001-gtkwindow-Don-t-force-enable-CSD-under-Windows.patch
+++ b/mingw-w64-gtk3/0001-gtkwindow-Don-t-force-enable-CSD-under-Windows.patch
@@ -1,0 +1,35 @@
+From 07bbd9b206f9a8712e1d77a6f86e260023738247 Mon Sep 17 00:00:00 2001
+From: Christoph Reiter <reiter.christoph@gmail.com>
+Date: Thu, 28 Sep 2017 12:02:03 +0200
+Subject: [PATCH] gtkwindow: Don't force enable CSD under Windows
+
+CSD look a bit out of place under Windows, especially
+when the application doesn't use the headerbar and the CSD
+doesn't add any value.
+
+Instead let the application decide, like under X11.
+It can still be force enabled through GTK_CSD=1
+---
+ gtk/gtkwindow.c | 6 ------
+ 1 file changed, 6 deletions(-)
+
+diff --git a/gtk/gtkwindow.c b/gtk/gtkwindow.c
+index 9f1e52ab04..7c6d874afb 100644
+--- a/gtk/gtkwindow.c
++++ b/gtk/gtkwindow.c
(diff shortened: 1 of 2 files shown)


Ede123:
I appreciate this change very much. I actually thought something was broken initially (see #2164) and filed [an upstream bug](https://bugzilla.gnome.org/show_bug.cgi?id=778791) subsequently to request this behavior as (apart from the obvious visual impairment) CSD also seriously limits functionality (like Aero Snap and similar).

That being said people should be aware, though, that disabling CSD and letting the Window manager draw window decorations currently imposes a significant hit in performance, see [upstream bug](https://bugzilla.gnome.org/show_bug.cgi?id=781153). For all typical applications it's probably still fine but if your application depends on redraw performance (like Inkscape in my case) you might want to keep that in mind.<br/>Unfortunately I'm not familiar enough with gtk to even hope to achieve anything myself, @LRN seems busy with other things and no other gtk dev seems to even look into win32 issues, so - as unfortunate as it is - a fix might not be expected any time soon.

lazka:
I haven't noticed any performance regression personally, but as you said it could be because my app doesn't draw much.

Coming up with a small program showing the performance problem would be a good start imo.

@Ede123 regarding this PR, are you OK with it being merged without looking into the perf problem or do you think we should wait first?

Ede123:
> Coming up with a small program showing the performance problem would be a good start imo.

Have you checked out the modified fishbowl demo I posted in https://bugzilla.gnome.org/show_bug.cgi?id=781153 ?

I'm fine with merging this. After all it restores "sane" default behavior, can still be changed by devs and users alike, and might even increase visibility of the performance issue which I hope can be sorted out eventually...

lazka:
> Have you checked out the modified fishbowl demo I posted in https://bugzilla.gnome.org/show_bug.cgi?id=781153 ?

I have not. I just mean something which is small, easily switchable between csd/non-csd and where performance result doesn't take over a minute to settle would be nice.

lazka:
Using your modified fishbowl .ui file (and setting a default-width/height of 300x300) I get:

NO-CSD: 2385
CSD: 2290

Ede123:
Interesting... Could you test at larger screen sizes?

For Inkscape performance goes down highly nonlinear when increasing the window size. Also in my fisbowl tests I typically used 1280x1440 (half my screen).

If it's got something to do with timings as I speculated in the other bug it might just be that at a certain screen size some redraw takes to much time...

lazka:
ok, with ~ 1500x900 I get:

NO-CSD: 700
CSD: 1100

lazka:
@Alexpux can this get merged?

lazka:
Thanks!

juml:
Personally not sure about this change.

As gtk3 app are adwaita themed doens't integrate in Windows environment as native applications.
Previously at least application itself was consistent (compare gedit before and after change).

Now it is kind of hybrid and doesn't fit anywhere.

PR #2994, opened 2017-09-29
docbook-xsl: fix uri rewrites

lazka:
The last release changed the URI rewrites from sourceforge to
docbook.org. Things like gtk-doc docs reference the sourceforge ones,
so add them back.


lazka:
This fixes errors such as 

```
"I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/html/chunk.xsl"
```

Issue #2995, opened 2017-09-30
imagemagick: relocation support broken/incomplete


Ede123:
Imagemagick is currently not fully relocatable as it assumes its binaries to be in a `/bin` subfolder and searches for configuration/coders/modules in `../etc` and `../lib` (relative to the binaries).

This breaks applications where binaries are packaged to be located directly in the installation root, e.g. example from Inkscape:
```
terminate called after throwing an instance of 'Magick::ErrorMissingDelegate'
  what(): Magick: NoDecodeDelegateForThisImageFormat `PNG' @ error/blob.
```

A quick workaround is to use `--without-modules` when configuring the build . While this does obviously does not resolve the fundamental relocation issues it allows imagemagick at least to find its coders and do something useful in most scenarios.

A proper fix seems to require some additional work (I attempted to produce a fix yesterday but wasn't successful yet; as there are other things I want to work on first I'll likely not look at it again any time soon).

Previous discussion on the topic: #1905, #1885.

Alexpux:
@Ede123 our mingw programs using unix filesystem layout as much as possible. We don't want support installation all in one folder.

Ede123:
I understand, yet most well maintained packages *do* support a more Windows-oriented directory structure and I assume it's in the interest of the project not to enforce any usage scenarios.

This is obviously not a high-importance bug (especially in the scope of MSYS2) but was meant as a hint / / warning / feature request for people who might like to use ImageMagick outside of the MSYS2 ecosystem.

As it's trivial in many cases to add this support (especially with toolkits already [supporting this kind of relocations](https://developer.gnome.org/glib/stable/glib-Windows-Compatibility-Functions.html#g-win32-get-package-installation-directory-of-module) natively) I personally think it's usually well worth to look into these kind of things.

Alexpux:
Well if you have right fix to support both unix-like layout and "all in one folder" then create pull request and we will review

jho101:
Thanks @Ede123 : Now I can use imagemagick in my application (win 7 64 bit). I just created etc and lib folder outside my bin folder. Now it works for me. I can live with it like it is. 

By the way I am using the MSYS2 ecosystem but somehow ImageMagick was only running inside the MSYS2 folder that is not very convienient if you want to distribute your application.

Biswa96:
[...]

PR #2996, opened 2017-09-30
WIP: Updated PKGBUILD for imagemagick6 branch

Ede123:
As a byproduct of looking into #2995 I produced a working PKGBUILD for the imagemagick6 branch (latest release is 6.9.9-15; latest release in the MSYS2 repo is  6.9.3.7)

If anybody needs it feel free to use. As it's not compatible with the package mingw-w64-imagemagick (they would be conflicting) it should not be merged, though. If there's interest in having up-to-date imagemagick6 builds in the MSYS2 repo we'd need to look into making them co-exist somehow.


lazka:
@Ede123 Do you think this is still useful, or can we close it?

Ede123:
FWIW I still create official Inkscape packages with ImageMagick 6 (Inkscape is not compatible with ImageMagick 7).

However it should be possible to build with graphicsmagick as an alternative, and as I haven't updated this MR in a while, it's not useful as-is.

PR #2997, opened 2017-09-30
harfbuzz: Make cairo an optional depends.

stahta01:
While trying to fix fontconfig; I was getting these messages that I thought was wrong.

Tim S.

updating font cache... C:/Apps64/MSys2/mingw64/bin/fc-cache.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
done.

warning: dependency cycle detected:
warning: mingw-w64-i686-harfbuzz will be installed before its mingw-w64-i686-freetype dependency
warning: dependency cycle detected:
warning: mingw-w64-i686-cairo will be installed before its mingw-w64-i686-freetype dependency
warning: dependency cycle detected:
warning: mingw-w64-i686-fontconfig will be installed before its mingw-w64-i686-freetype dependency

### DIFF
--- a/mingw-w64-harfbuzz/PKGBUILD
+++ b/mingw-w64-harfbuzz/PKGBUILD
@@ -5,7 +5,7 @@ _realname=harfbuzz
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.5.1
-pkgrel=1
+pkgrel=2
 pkgdesc="OpenType text shaping engine (mingw-w64)"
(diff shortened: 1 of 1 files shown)


mati865:
If I remember correctly Arch also have cycle there and there is nothing wrong about it.

stahta01:
The fix I used was copied from ArchLinux.
Tim S.

mati865:
Ah, Arch have cycle between `harfbuzz` and `freetype2`. MSYS2 have cycle between `harfbuzz`, `freetype2` and `cairo`.

PR is fine then.

Issue #2954, opened 2017-09-24
fontconfig: default cache dir is "C:\building\msys64\var\cache\fontconfig"


Ede123:
If fontconfig is used with a fonts.conf that does not define any cache directories, fontconfig falls back to `C:\building\msys64\var\cache\fontconfig`.

It would be great if it would not fallback to a hardcoded path but use a sane default instead.

stahta01:
I am looking into the package and trying to fix it. I will need someone to test it if I find a fix.

Tim S.

Ede123:
Great! I can assist in testing, just ping me.

stahta01:
After more than a half dozen attempts; I believe that I found out how to fix it.
It should now default to /var/cache/fontconfig not sure that this is the most correct place.
But, it should be easy to change in the future to a better default.
I still need to finish the changes and verify it still builds and installs without error.

Tim S.

Alexpux:
Mingw programs doesnt understand UNIX paths, need fix sources to search cahce dir relative to executable

stahta01:
OK, I will stop trying to fix configure; and, instead edit "fontconfig/src/fcinit.c".
It should be easier.

Tim S.

Alexpux:
@stahta01 view any of the packages where we do relocation of directories. For example p11-kit

jtanx:
From my experience, fontconfig already falls back to `../share/fonts` where `..` is relative to the fontconfig DLL. This is what fontforge used to rely on (it's no longer the case because we use pango with the win32 backend by default now)

Although I could be wrong; I'd need to double-check if that behaviour is from pango or fontconfig

stahta01:
It is going to take at least a few days to understand the code I need to borrow from p11-kit; before I see the way to proceed. Tim S.

stahta01:
@Ede123 I have started to try to do the fix; and, I think I will need to patch the exe which has the problem.
What exe has the problem the worst? fc-cache.exe?

Tim S.

Ede123:
While I discovered the issue while using fc-cache the problem should not be specific to any particular executable but to the way the default configuration is created. I tracked this a bit through the code and there's a central place were all fontconfig configuration code is handled (including fallbacks if not set explicitly in the configuration file).

stahta01:
[...]

PR #2998, opened 2017-09-30
fontconfig: set with-cache-dir=yes  [HOLD Till Ede123 says it works]

stahta01:


Ede123:
I'm sorry, but this does not properly address the problem described in #2954:
* The default cache location (in fonts.conf) is properly changed to `LOCAL_APPDATA_FONTCONFIG_CACHE` but this results in the cache being created in `%localappdata%\fontconfig\cache`, even for a default MSYS2 environment. In this predictable case I assume we don't want a cache location outside of the MSYS2 directory tree.
* If no cachedir is specified in fonts.conf (i.e. the behavior that was supposed to be fixed) the cache does now end up in `$PWD/LOCAL_APPDATA_FONTCONFIG_CACHE` (i.e. the special folder name is used verbatim as a real folder name)

Issue #2989, opened 2017-09-29
gpg uses wrong 'home'


Karlson2k:
gpg use 'home' subdir inside 'mingw64' dir instead of 'HOME' env variable. 
````
$ LANG=c gpg --version
gpg (GnuPG) 2.2.0-unknown
libgcrypt 1.8.1
NOTE: THIS IS A DEVELOPMENT VERSION!
It is only intended for test purposes and should NOT be
used in a production environment or with production keys!
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: C:/msys64/mingw64/home
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2

$ echo $HOME
/home/User-Name
````

peterbud:
#2743 was supposed to fix this, but was not merged

Karlson2k:
@peterbud Looks like @Alexpux asked to updated patches against latest version.
To my experience, patches for current version merged pretty quickly in this repo.
Could you update your patches?

peterbud:
See #2999

Alexpux:
Won't fix

Karlson2k:
@Alexpux do you mean that you do not want to fix or do you mean that you are waiting for right patches?
Or you treat it as 'by design'?

Alexpux:
Mingw gnupg can bee installed outside msys2 tree. In this case home folder will be broken. 
All mingw packages are not related to msys2 filesystem so I can't do shared gnupg folder for both mingw and msys gnupg

PR #2999, opened 2017-09-30
Revising gnupg patches - Fixes #1778 #2989

### LINKED ISSUE #1778: GnuPG patches disable user profile based home directories

peterbud:
* HOME directory is now located at ~/.gnupg (/home/<username>)
* All basic tests are now running smoothly (21+4+1+2). Before this only the first 19 test cases were successful.
* Removing unnecessary patches


Alexpux:
What you mean by ````~/.gnupg (/home/)````?
You mean that it locate files inside msys2 home directory? If yes this is completely wrong for mingw package.
Also you MUST leave 03 patch. If you dont know why it there you must ask at leqast but not dellete what you dont add.

peterbud:
Hi @Alexpux  I did not know about the linking issues with openldap. OK then leave it is as is. Then please also close #1778 and #2989 as won't fix.

Alexpux:
Nothing to fix

PR #3007, opened 2017-10-03
glib: Update to 2.54.1

lazka:
* Fix wrong python exe wrapper for glib-gettextize
* Move to mingw gtk-doc


lazka:
Hm, ok, so I guess there is a gtk-doc bug when the package isn't installed. I'll investigate.

lazka:
https://bugzilla.gnome.org/show_bug.cgi?id=787768

PR #3014, opened 2017-10-05
opencv and suitesparse compile error fix.

bygreencn:
opencv 3.3.0 and suitesparse 4.5.5 compiling error fixed.


bygreencn:
tiny-dnn is not changed with the base branch.

PR #3034, opened 2017-10-10
qt5: Update to 5.9.2 [WIP]

filnet:
Had to disable an ICU patch. Not sure if it needs to be adapted or not.


Alexpux:
Already updated by me

filnet:
Yep. And in a better way.

gbschenkel:
The package appears to be broken. #3036

PR #3066, opened 2017-10-19
gst-libav: Build with system ffmpeg instead of internal one

lazka:
No need to include ffmpeg since it's likely getting pulled in by something
else anyway. This is also what Arch Linux does.

### DIFF
--- a/mingw-w64-gst-libav/PKGBUILD
+++ b/mingw-w64-gst-libav/PKGBUILD
@@ -4,17 +4,16 @@ _realname=gst-libav
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.12.3
-pkgrel=1
+pkgrel=2
 pkgdesc="GStreamer libav (mingw-w64)"
 arch=('any')
 url="https://gstreamer.freedesktop.org/"
 license=("LGPL")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-pkg-config"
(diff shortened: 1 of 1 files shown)


lazka:
@filnet (since you have updated gstreamer multiple times) Ok with this?

mingw-w64-gst-libav/PKGBUILD
```diff
@@ -4,17 +4,16 @@ _realname=gst-libav
...
 arch=('any')
 url="https://gstreamer.freedesktop.org/"
 license=("LGPL")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-pkg-config"
-             "${MINGW_PACKAGE_PREFIX}-python2"
-             "yasm")
-depends=("${MINGW_PACKAGE_PREFIX}-bzip2"
-         "${MINGW_PACKAGE_PREFIX}-gst-plugins-base")
+             "${MINGW_PACKAGE_PREFIX}-python2")
```
filnet: minor indent issue

filnet:
@lazka i more familiar with gstreamer and gst-plugins-* packages. That said, your change looks fine.

For information, gst-libav uses ffmpeg n3.3.4 internally. Msys2 has the same version.

lazka:
Cool, thanks.

filnet:
Let's just hope that gstreamer does not have strict requirements on *how* ffmpeg is compile.

lazka:
From what I see it just disables codecs for which it has support in other plugins: https://github.com/GStreamer/gst-libav/blob/a771795005fe5ebc06d788a4520e80890685aadc/configure.ac#L349

PR #3090, opened 2017-10-28
release package for GnuCOBOL (currently at version 2.2)

GitMensch:
Instead of the svn one (which I still would like to update to allow people to use "latest") this is the package for the most current release version.

A note concerning the patches:

* 001 will *only* work with MSYS2 (building), not with old mingw and not with any other environment - if I remember correctly this will just remove the format warnings during compilation (with the additional `-D__USE_MINGW_ANSI_STDIO=1`
* 002 is (slightly changed) in svn version already and will therefore be removed with the next release and not integrated in the svn package

A question: Is it possible to install the package with adjusting the user's environment? The generated binaries get the install-prefix' path hard-wired into them (always did) and won't work anywhere as the binaries need to use Windows paths (I could add a translation to Windows paths but this won't work for a user that has a different MSYS2 installation path than the machine creating the binaries).
The solution is to override the builtin default paths by environment variables, if we can't add a setting of these into the PKGBUILD the user would have to do so on his own before calling the binaries. Any idea to solve this problem?

### DIFF
--- a/mingw-w64-gnucobol/PKGBUILD
+++ b/mingw-w64-gnucobol/PKGBUILD
@@ -5,18 +5,15 @@
 _realname=gnucobol
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
-_base_ver=2.2
 pkgver=2.2
 pkgrel=1
 pkgdesc="GnuCOBOL, a free and modern COBOL compiler (mingw-w64)"
 _dirname=${_realname}-${pkgver}
 arch=('any')
-provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 url="https://www.gnu.org/software/gnucobol/"
 options=('strip')
 license=('GPL3' 'LGPL3')
-# --> maybe place autoconf/make here, at least for the svn version makedepends=("")
+# note: the final binaries will call GCC, therefore it is a dependency
 depends=("${MINGW_PACKAGE_PREFIX}-gcc"
          "${MINGW_PACKAGE_PREFIX}-gmp"
          "${MINGW_PACKAGE_PREFIX}-db")
--- a/mingw-w64-gnucobol/PKGBUILD
+++ b/mingw-w64-gnucobol/PKGBUILD
@@ -16,16 +16,22 @@ license=('GPL3' 'LGPL3')
 # note: the final binaries will call GCC, therefore it is a dependency
 depends=("${MINGW_PACKAGE_PREFIX}-gcc"
          "${MINGW_PACKAGE_PREFIX}-gmp"
+         "${MINGW_PACKAGE_PREFIX}-gettext"
+         "${MINGW_PACKAGE_PREFIX}-ncurses"
          "${MINGW_PACKAGE_PREFIX}-db")
-optdepends=("${MINGW_PACKAGE_PREFIX}-gettext: support for internationalization"
-            "${MINGW_PACKAGE_PREFIX}-ncurses: support for extended screenio"
-            "${MINGW_PACKAGE_PREFIX}-pdcurses: support for screenio")
+# the optional depencies cannot be used for creating the binary packages, moved to depends
+#optdepends=("${MINGW_PACKAGE_PREFIX}-gettext: support for internationalization"
+#            "${MINGW_PACKAGE_PREFIX}-ncurses: support for extended screenio"
+#            "${MINGW_PACKAGE_PREFIX}-pdcurses: support for screenio")
 source=("${_realname}::https://ftp.gnu.org/gnu/${_realname}/${_dirname}.tar.xz"
         "001-mingw-gnucobol-2.2-fixformatwarnings.patch"
-        "002-mingw-gnucobol-2.2-skipcursestests.patch")
+        "002-mingw-gnucobol-2.2-skipcursestests.patch"
+        "cobenv.sh" "cobenv.cmd")
 sha256sums=('DC18FC45C269DEBFE86A4BBE20A7250983CBA6238EA1917E135DF5926CD024A0'
             '3FBFDAF61EB6EC125258DE62B404C97BF6DCCF39BD3B9152510F794B5383CA33'
-            '92F4D25AA33E4058BC48E071334697091F1DD763A7FD39D93468D52081103549')
+            '92F4D25AA33E4058BC48E071334697091F1DD763A7FD39D93468D52081103549'
+            '7B26508D34F82514CB3940C78CE181DE5B187F5ECEE6EF7DD496017A779AD273'
+            'D7619C83EF967644A96C9F6FBECB3F4585FFCB8340AF3B69F32D065673B32AE2')
 
 prepare() {
   cd "${_dirname}"
@@ -46,9 +52,11 @@ build() {
     --prefix=${MINGW_PREFIX} \
     --infodir=${MINGW_PREFIX}/share/info \
     --mandir=${MINGW_PREFIX}/share/man \
+    --disable-rpath \
     --enable-shared \
-    --enable-static \
-    --disable-rpath
+    --disable-static
+    #  --> as using the static libcob would be the builtin default,
+    #      all generated COBOL modules would be GPLed (and big)...
 
   make -j1
 }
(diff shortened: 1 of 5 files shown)


GitMensch:
@Alexpux - the CI builds complain about:

> failure: mingw-w64-gnucobol: circular reference mingw-w64-gnucobol in provided packages
> failure: mingw-w64-gnucobol: circular reference mingw-w64-gnucobol in conflicted packages

but there is no self-dependency in the PKGBUILD - or do I fail to see it?

mingw-w64-gnucobol/PKGBUILD
```diff
@@ -0,0 +1,71 @@
+# Maintainer: Alexey Pavlov <alexpux@gmail.com>
+# Contributor: Jürgen Pfeifer <juergen@familiepfeifer.de>
+# Contributor: Simon Sobisch <simonsobisch@gnu.org>
+
+_realname=gnucobol
+pkgbase=mingw-w64-${_realname}
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+_base_ver=2.2
+pkgver=2.2
```
mati865: Why create 2 variables with the same value?

mingw-w64-gnucobol/PKGBUILD
```diff
@@ -0,0 +1,71 @@
...
+pkgrel=1
+pkgdesc="GnuCOBOL, a free and modern COBOL compiler (mingw-w64)"
+_dirname=${_realname}-${pkgver}
+arch=('any')
+provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
```
mati865: Remove `provides` and `conflicts`

mingw-w64-gnucobol/PKGBUILD
```diff
@@ -0,0 +1,71 @@
...
+url="https://www.gnu.org/software/gnucobol/"
+options=('strip')
+license=('GPL3' 'LGPL3')
+# --> maybe place autoconf/make here, at least for the svn version makedepends=("")
+depends=("${MINGW_PACKAGE_PREFIX}-gcc"
```
mati865: `${MINGW_PACKAGE_PREFIX}-gcc` is part of `${MINGW_PACKAGE_PREFIX}-toolchain`, remove it

mingw-w64-gnucobol/PKGBUILD
```diff
@@ -0,0 +1,71 @@
...
+pkgrel=1
+pkgdesc="GnuCOBOL, a free and modern COBOL compiler (mingw-w64)"
+_dirname=${_realname}-${pkgver}
+arch=('any')
+provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+url="https://www.gnu.org/software/gnucobol/"
+options=('strip')
+license=('GPL3' 'LGPL3')
+# --> maybe place autoconf/make here, at least for the svn version makedepends=("")
```
mati865: No need for it

mingw-w64-gnucobol/PKGBUILD
GitMensch: OK, but please explain it.
The project actually depends on the gcc package (it must be installed along this package as the binaries installed on the user side will *call* gcc). Is the `${MINGW_PACKAGE_PREFIX}-toolchain` always installed? If yes it may be a good idea to add this to "how to create a new package".

mingw-w64-gnucobol/PKGBUILD
GitMensch: There's no need for it, I guess I've just copied it over from the existing obsolete gnu-cobol-svn package.

mingw-w64-gnucobol/PKGBUILD
mati865: OK, my bad then.

mingw-w64-gnucobol/PKGBUILD
GitMensch: I've just left them in from the obsolete gnu-cobol-svn package - but after rechecking https://www.archlinux.org/pacman/PKGBUILD.5.html#_options_and_directives its obvious that these should not be integrated. I'll remove them.

mingw-w64-gnucobol/PKGBUILD
GitMensch: No need as these are always guaranteed to be available?

GitMensch:
Thank you @mati865 for the review, it is clear why these failed the CI now. I'm adjusting the PKGBUILD and will commit it.

Do you have any idea for the "user-side windows-path" issue asked about in the PR?

mati865:
@GitMensch this issue is a plague for MINGW packages on Windows.
Simplest solution is to use relative path, example: we have `app.exe` with some libs inside path `x/y/z/bin` and other libs in `x/y/z/lib`, app can access other libs dir using path `../lib`.
I hope it's clear enough.

GitMensch:
@mati865 

> Simplest solution is to use relative path...

I'd patch the build-specific `defaults.h` with relative paths, but as the binary "app.exe" will be called from different working directories relative paths wouldn't work.

Therefore the only option would be to use relative paths from within the applications path, and allow the generated `defaults.h` to have something like `EXEDIR\\..\\share\\gnucobol\\config` in (which could be `patch`ed in there before running `make`). But as this wouldn't change anything to the last release version and all CIs are working fine now I'd say this PR would actually be finished :-)

mati865:
There is another solution using `post_install()` like https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-fontconfig/fontconfig-x86_64.install#L1 but it reduces possibility of moving MSYS2 directory (this package will be broken).

After package is installed `pacman` calls `post_install()` which can modify headers/config files or run commands.

GitMensch:
As the paths are currently hard wired we can't change anything in there. The only option would be to add the following to the global profile and run them in the current environment, but it looks too brute force:

```
export COB_CONFIG_DIR=`cygpath.exe -pw $MINGW_PREFIX/share/gnucobol/config`
export COB_COPY_DIR=`cygpath.exe -pw $MINGW_PREFIX/share/gnucobol/copy`
export COB_LIBRARY_PATH=`cygpath.exe -pw $MINGW_PREFIX/lib/gnucobol`
```

Another option would be to include this in the already installed `cob-config` script as a new option `--source-msys2-mingw-env` or similar and let the user source it once if he wants to use the package.

Any thoughts on this?

GitMensch:
Note: As there is an additional shell script (for use withing mingw terminals) and an additional cmd file (for use in cmd.exe / Windows build scripts) which are tested and found to be working already I see this PR be finished and only needed to be merged by @Alexpux

Ede123:
Does somebody know what's going on with the two files `cobenv.cmd` and `cobenv.sh` which were added in this commit?

Whatever I do git tells me those two files would have been changed locally, but deleting/reverting them does not help. I suspect it has something to do with line endings as git diff tells me the whole file has changed (but does not show any actual changes).
However I already tried all three settings for `core.autocrlf` without success; I also tried actually changing line endings to `CRLF / LF / CR` respectively but with whatever setting I choose git still tells me the files would have been changed locally...

GitMensch:
@Ede123  Not sure but this may be because they are executable files. Just rechecked: these were committed with CRLF. In general it seems as your git client settings (no idea about details here as I mainly use svn) are set to change the files somehow...

Ede123:
OK, I tracked down the source of the issue and the output of `git ls-files --stage --eol` clears it up a bit:
```
$ git ls-files --stage --eol
100644 1861d74904921b367d6f34f990c5805d901dd0f6 0       i/lf    w/lf    attr/text eol=lf        001-mingw-gnucobol-2.2-fixformatwarnings.patch
100644 16b20eb038edbf16808362311a64a93d19d80a55 0       i/lf    w/lf    attr/text eol=lf        002-mingw-gnucobol-2.2-skipcursestests.patch
100644 cfabfa273cf4151d875a2a752ec2c1d25b8df45b 0       i/lf    w/lf    attr/text eol=lf        PKGBUILD
100644 8877af0acb2ee9ccdda027503e2bb91162abc35a 0       i/crlf  w/crlf  attr/text eol=lf        cobenv.cmd
100644 0dfd002bf3a5082a9a96c585a52c81e6c39d4d77 0       i/crlf  w/crlf  attr/text eol=lf        cobenv.sh
```

It's unrelated to the files being executable:
* in the index they are not, i.e. `644`
* default settings on Windows seem to be `core.filemode = false`, so it doesn't matter anyway

It definitely is a problem with line endings
* the files are CRLF in index
* the files are CRLF in my working tree
* the attributes of the file however say it *should* be lf (or so I think) which is [defined in `.gitattributes`](https://github.com/Alexpux/MINGW-packages/blob/master/.gitattributes#L2)

If I remove the linked line from `.gitattributes` the conflicts disappear...
If I understand the job of .gitattributes properly you therefore should have never been able to commit a file with CRLF line endings (i.e. you probably hit a bug in your git client). I'm unsure wether it's proper behavior that git now complains about local changes for me, which I seem to only be able to resolve if I create a commit which changes line endings...

GitMensch:
The client is not to be harmed :-) I've used the svn bridge provided by GitHub and svn clients can only use the svn properties (in this case svn:eol) which are not part of the commits.
Therefore the bugs seems to be a missing server-side pre-commit check.

If I understood you correctly you can do your commits (with the locally changed but not committed .gitattributes file), correct?

In any case someone should create an issue for GitHub about the missing server-side check. Can you do this?

And while it is fine to have the .sh file with lf it obviously is not fine to have .cmd or .bat files any other than crlf (as these must be executable from cmd.exe). Can you create a PR with the .gitattributes file changed to force these extensions crlf allowing @Alexpux to change this fast?

I try to do the necessary changes for the .sh file to lf, adjusting the SHA256 in the PKGBUILD and retest creating the package tomorrow.

Ede123:
Sorry, I didn't even know an svn bridge existed, so I certainly won't start opening bug reports experienced by someone else with a software I don't use.

Regarding .gitattributes:
I'm not familiar with the format either and I'm not even sure a change would be desirable.
What would be wrong with having LF in a .bat file? It seems to execute just fine (just tested it locally).

KrullBorg:
same here; i cannot stage anymore

GitMensch:
I've fixed the line endings, see https://github.com/Alexpux/MINGW-packages/pull/3193

PR #3094, opened 2017-10-30
Add NVPTX target to LLVM.

TravisWhitaker:
This simply adds NVPTX support to the LLVM built by the `mingw-w64-*-clang` package. It's very handy to have this in the MinGW-w64 package, since it's quite a hassle to build a usable LLVM from source on Windows (principally due to `llvm-config.exe` bugs, which I'd like to send PR's for upstream once I have the time). This will marginally increase the package build time and the size of `LLVM.dll`.

This is especially handy for [accelerate-llvm-ptx](https://hackage.haskell.org/package/accelerate-llvm-ptx) users on Windows, since the Stack Haskell build tool provides a sandboxed MinGW anyway.

I managed to build this on my machine without issue and the resulting package seems to work. Apologies if I've forgotten something or I'm doing something wrong; this is the first MinGW development I've done of any sort.

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -21,7 +21,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-lldb"
          "${MINGW_PACKAGE_PREFIX}-polly")
 pkgver=5.0.0
-pkgrel=3
+pkgrel=4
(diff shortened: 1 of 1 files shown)


TravisWhitaker:
@Alexpux, is there anything else I need to do to get this update into the package index? It still doesn't seem to be at http://repo.msys2.org/mingw/x86_64/.

TravisWhitaker:
@Alexpux, does this package fail to build? I'm happy to help/do whatever needs to be done to get this into the index.

PR #3103, opened 2017-11-01
icu: update to 60rc

Adsun701:
This updates icu to the latest version, 60rc.

### DIFF
--- a/mingw-w64-icu/PKGBUILD
+++ b/mingw-w64-icu/PKGBUILD
@@ -1,16 +1,12 @@
 # Maintainer: Alexey Pavlov <Alexpux@gmail.com>
 # Contributor: Ray Donnelly <mingw.android@gmail.com>
 
-# Note: I would like to update to 59.1 but I found
-# putil.cpp: In function 'const char* uprv_getDefaultLocaleID_59d()':
-# putil.cpp:1733:86: error: 'GetUserDefaultLocaleName' was not declared in this sc           ope
-# int length = GetUserDefaultLocaleName(windowsLocale, UPRV_LENGTHOF(windowsL           ocale));
-
 _realname=icu
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}" "${MINGW_PACKAGE_PREFIX}-${_realname}-debug-libs")
-pkgver=58.2
-pkgrel=2
+pkgver=60rc
+_pkgver=60
+pkgrel=1
 arch=('any')
 pkgdesc="International Components for Unicode library (mingw-w64)"
 arch=('any')
@@ -34,7 +30,7 @@ source=("http://download.icu-project.org/files/icu4c/${pkgver}/icu4c-${pkgver//.
         0016-icu-pkgconfig.patch
         0017-icu-config-versioning.patch
         0020-workaround-missing-locale.patch)
-sha256sums=('2b0a4410153a9b20de0e20c7d8b66049a72aef244b53683d0d7521371683da0c'
+sha256sums=('ee061fba196fc714106b1248001c84830ad004edb62dd379bbe122aa022998ae'
             '314765fc244e52b8a077f6470f2bbb21c6954d649481af6d85b14af13717f7a8'
             'ae73db58e714857f544f9e40de8fc6847f03c41346d5a8f42dd313ec99fc1ba9'
             '2038881890542797fe1fe9a384cf2bb6c5504b18fb26eab1368f5cb0e6f7f7af'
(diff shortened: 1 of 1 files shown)


mati865:
@Ede123, it breaks XP/Vista support. What do you think?

Ede123:
For one we certainly should wait for the final release instead of relying on a release candidate.

Regarding XP support it's a bit more complicated...
* In an ideal world it would be great to say goodbye to XP for good. Objectively people *should* stop using it and I'd be totally fine with dropping XP support.
* In the *real* world however it seems a lot of people are still using Windows XP these days.<br/>As the maintainer of the Windows builds of Inkscape  (why depends on ICU) this change would make my life a lot harder. I'd either have to tell those people that Inkscape 0.92.2 is the last version they can use (probably very unpopular) or I'd have to start to maintain my own builds of ICU and the dependent libraries [!] which is obviously possible but means a lot of work and additional infrastructure as well as more complicated build instructions for new developers which I had hoped to eliminate by using an MSYS2 build environment.

I assume other projects might be affected by such a change as well?

One thing I also can not judge is the drop of Vista compatibility (Windows 7 still seems supported) and how many people it might affect.

peterbud:
The support for XP has ended more than 3 years ago in 2014. I believe it does not make too much sense to make significant efforts to keep supporting an OS which is 15+ years old, and not even the manufacturer supports anymore.

lazka:
> I'd either have to tell those people that Inkscape 0.92.2 is the last version they can use (probably very unpopular)

What makes you think that? They are already using lots of unsupported and insecure software without any way to update.

Ede123:
As I said: You don't have to convince me but the users of the software we build...

Personally I won't shed tears over XP, some users certainly will (if it's a "relevant" amount - who knows - you always hear those few who make the most noise...)

If the MSYS2 project as a whole decides it's time to move on so be it - I'd just prefer if people took a moment to think before dropping XP support for ideological reasons. (E.g. @peterbud you're creating darktable - have you asked your users if there are some still using XP?)

> > I'd either have to tell those people that Inkscape 0.92.2 is the last version they can use (probably very unpopular)
> 
> What makes you think that?

The [mailing list thread on inkscape-devel](http://inkscape.13.x6.nabble.com/Roadmap-for-Inkscape-0-93-on-Windows-GTK3-devlibs-MSYS2-and-support-for-Windows-XP-tt4979046.html) I created some time ago when I started to switch the current development version (i.e. future 0.93) of Inkscape to use MSYS2 and mentioned we won't be able to support XP because GTK+3 does not support it (even then people started to complain and argue why XP still had its right to exist when I specifically explained that we *can't* support it).<br/>
Back then I got people of my back explaining it would only affect the development version but not 0.92.x stable branch.

lazka:
Hm, thanks for the info. I don't think that people on a mailing list are a good representation of a software user base, but I can understand that you want to keep active people involved in the development/community happy.

btw glib has dropped XP support now, so the release in 5 months will be win7+

peterbud:
@Ede123 I have not asked darktable users whether they use WinXP or Win95. 

Just one analogy which I believe is interesting: darktable is not being updated for Ubuntu 14.04 anymore, so if you are on that version (which BTW has been released as recently as 2014!), you stuck with version 2.0.7. See more details [here](https://launchpad.net/~pmjdebruijn/+archive/ubuntu/darktable-release)

So in other words if you keep an old version of OS is fine, but then you will only get an old version of the software. If you want tough you can try to build it for yourself, but there are good reasons why old OS versions cannot build new versions of software.

lazka:
Suggestion: Leave this until 2018-03-14, when glib without XP support is released and reevaluate then.

Ede123:
@lazka Your suggestion would certainly have my support (I kind of assumed this would be obvious, sorry if it wasn't).

Whatever happens I want to release at least an Inkscape 0.92.3 with XP support as 0.92.2 was the first release built with MSYS2 and some issues related to this switch were discovered and solved in the meantime.
Unfortunately I'm not even able to build for XP right now due to some mingw-w64 issue (which linked `_time32` which seems not to be available on XP; not sure if this was publicly discussed anywhere).

Related question: Does anybody know how long old packages are kept in the repos? Otherwise I'd just save the package databases once the build is successful again to make sure I have a backup plan available...

mati865:
You can browse repository, there still are packages uploaded years ago.

22.11.2017 21:24 "Eduard Braun" <notifications@github.com> napisał(a):

lazka:
glib requiring Win7+ is out: https://ftp.acc.umu.se/pub/GNOME/sources/glib/2.56/

lazka:
new glib is in the repos

lazka:
3cd54ad32894fe5bb7

PR #3125, opened 2017-11-09
SDL2_image: link in PNG support library

fabiangreffrath:
Link in the PNG library instead of relying on dlopen().

For proper read/write support for the most popular image format
the library should be linked in, we depend on the corresponding
package anyway.


fabiangreffrath:
Thanks for the prompt merge, should I have bumped `pkgrel`?

PR #3142, opened 2017-11-15
Drop gnome-icon-theme and gnome-icon-theme-symbolic

lazka:
* They are not included in the pacman repository
* They are deprecated upstream for some time now and replaced by adwaita-icon-theme


lazka:
(unless someone uses them of course, but I doubt it)

PR #3158, opened 2017-11-21
Add: qbittorrent

Chocobo1:


Chocobo1:
tea-ci fails because the packages there are too old. libtorrent-rasterbar was updated recently and the CI didn't pick up the update.

PR #3164, opened 2017-11-23
libxml2: update to version 2.9.7

Ede123:
Many patches (especially fixes for warnings) could be dropped due to upstream fixes (many of them submitted by @JPeterMugaas , thanks!)


Ede123:
libxml2-2.9.7/include/wsockcompat.h is included in both nanoftp.c and nanohttp.c (so ws2tcpip.h is included via this header; windows.h is not included in either of them, so we should be fine regarding inclusion order).

Revisiting this I guess we could go even further and drop 0004 altogether (as XML_SOCKLEN_T is now globally defined in wsockcompat.h) and we don't need to define HAVE_WS2TCPIP_H anymore as  0011 became obsolete due to wsockcompat.h now being included in nanoftp.c as explained above.

JPeterMugaas:
Perhaps, we could drop although it might be nice if libxml was to keep it for backwards compatibility with old systems that required Winsock 1.1 but that would in that library itself (yes, I know, Winsock 1.1 has not been popular for over a decade).  In the back of my mind, I would like to see IPv6 supported in nanoftp and might be doable (ideally in libxml itself) and I feel that this could be doable.

Ede123:
I just added a commit removing the two patches as they were not useful anymore with the current code.

I doubt we need to maintain backwards compatibility with even older systems? Winsock 2 was available since Windows 95 AFAIK, I assume that should cover most bases... Do you really build for (or know of) older systems? I would not even be sure if it's possible to run mingw-w64 compiled software at all there (and I certainly have no possibility to test).

JPeterMugaas:
No, I don't.  But such issues would probably be done upstream anyway.  At least, it is nice to put a few things on the chopping block.

jeroen:
FYI libxml2 2.9.8 is available as well.

JPeterMugaas:
https://github.com/Alexpux/MINGW-packages/pull/3616 - Enough has been said.

PR #3168, opened 2017-11-26
add mono package (version 5.4.1.7)

Adsun701:
This adds the mono package (version 5.4.1.7), a C# and .NET Framework implementation.

Unfortunately it requires an already existing installation of mingw-w64-mono for bootstrapping; however, I've uploaded existing mingw-w64-mono packages here for installation and bootstrapping:
https://github.com/Adsun701/msys2-mingw-binary-packages

Thanks.


mingwandroid:
Hod did you build these bootstrap binaries? Where is the recipe?

Adsun701:
I originally built the bootstrap mono package by using the monolite directory that came with the source. The recipe can be found here: [https://github.com/Adsun701/msys2-mingw-binary-packages/tree/master/mono-bootstrap](url)

mingwandroid:
I fixed your links. You aren't doing them right. I think you should include the bootrap recipe in a subdir here.

Adsun701:
Added bootstrap recipe. Sorry for not responding so long.

PR #3174, opened 2017-11-28
libmicrohttpd: updated to version 0.9.57, improved

Karlson2k:
* updated to version 0.9.57
* removed dependency on libgcrypt,
* enforced configure https flag to ensure that HTTPS support will be built.


Karlson2k:
While builds are succeed on tea, I noticed some strange messages in log, like 
````
  -> Generating .BUILDINFO file...
error: invalid option '--noprogressbar'
````
And
````
=> WARNING: You don't have installed mingw-w64 toolchain for architecture i686.
````

Not related to libmicrohttpd.

PR #3186, opened 2017-12-02
wxPython classic version 3.0.3.0 git package added

stahta01:
wxPython is doing a complete build change for version 4 also called Phoenix.

I thought leaving wxPython for the version 4 would be in keeping with MSys2 package naming.
The current wxPython is broken, I did *not* find out till I tried to use it in learning python.
Sorry about that, since I pull requested wxWidgets 3.0.3 that broke it.

I am going back to trying to fix plplot once more. Then, after that trying to update wxPython to version 4.

Tim S.


stahta01:
Removed package that uses wxWidgets 3.0.2 because of auto build issues with wxWidgets installed by wxpython3-git. I have decided I need to edit wxWidgets package before trying my wxpython3 package once more.

Tim S.

PR #3205, opened 2017-12-12
cairo: Update to 1.15.10

lazka:


lazka:
@Ede123 This release/snapshot switches filename encoding from ANSI to utf-8 (for example the first argument to cairo_pdf_surface_create()). This shouldn't make any difference if you use ASCII paths, but maybe this affects inkscape (I requires some changes in pycairo: https://github.com/pygobject/pycairo/issues/86)

Ede123:
Thanks for the notification Christoph, I'll have a look. I think Inkscape does not use cairo to write/read directly to/from disk in most cases but certainly doesn't hurt to make sure.

PR #3223, opened 2017-12-18
add unixodbc package (version 2.3.4)

Adsun701:
This adds unixodbc.

### DIFF
--- a/mingw-w64-unixodbc/0001-unixodbc-2.3.4-int.patch
+++ b/mingw-w64-unixodbc/0001-unixodbc-2.3.4-int.patch
@@ -1,0 +1,122 @@
+--- unixODBC-2.3.4/include/sqltypes.h.orig	2013-12-06 05:51:46.000000000 -0500
++++ unixODBC-2.3.4/include/sqltypes.h	2017-12-20 12:08:21.025954400 -0500
+@@ -174,9 +174,15 @@ typedef unsigned long   SQLSETPOSIROW;
+ #else
+ typedef long            SQLINTEGER;
+ typedef unsigned long   SQLUINTEGER;
++#ifdef _WIN64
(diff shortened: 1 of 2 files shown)


JPeterMugaas:
One problem that I've noticed is the source code gives numerous warnings such as:
warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]

The problem is that in _WIN64, the int and long do not grow to 64-bits like the code expects and this can have some consequences down the line.  I am look at this as I feel that this does need nipped in the bud.

JPeterMugaas:
[unixODBC-2.3.4-int.patch.txt](https://github.com/Alexpux/MINGW-packages/files/1583854/unixODBC-2.3.4-int.patch.txt)
Alright, I did look at this and I think I have patch for the issue I discussed.  There were two things.  In Win64, the types SQLLEN, SQLULEN, SQLSETPOSIROW are 64-bits wide.  This was pointed out by https://docs.microsoft.com/en-us/sql/odbc/reference/odbc-64-bit-information and is in the mingw-w64 sql headers.

Along with this but related, the code itself was typecasting pointers to something like SQLINTEGER which does not grow to 64-bits in Win64.  Unlike Unix-like operating systems, the long and int types do not grow in Windows.

BTW, you need to add the line:

conflicts=("${MINGW_PACKAGE_PREFIX}-firebird2-git")

That's because there is a /mingw64/bin/isql.exe and /mingw32/bin/isql.exe in that series of packages.  I may look at the firebird packages to see if there is some type of relationship or workaround for the issue.

Adsun701:
Fixed. In the Arch Linux firebird package (firebird-superserver, can be found in the Arch User Repository), the isql executable has -fb added to avoid conflicts with unixodbc.

PR #3229, opened 2017-12-22
lfcbase : Update to version 1.11.5

bjlemke:


bjlemke:
Checks succeed, if lfcbase only is commited. Seems that for cego build no current version of lfcbase is installed

PR #3243, opened 2017-12-30
make: gettext is a runtime dependency

TBBle:
Closes Alexpux/MINGW-packages#2233, Alexpux/MSYS2-packages#828,
Alexpux/MSYS2-packages#842.


TBBle:
Change is based on 47c0acdb1659609ad47a8ca9388946e475979365 for Alexpux/MINGW-packages#2048

PR #3251, opened 2018-01-02
libplist: Don't include DllMain in static lib

encounter:
This fixes an issue I was getting trying to include this static library. 

(I also noticed that openssl's libcrypto.a includes a DllMain, is this intentional?)


encounter:
@Alexpux It looks like some patches are missing from the current repo. Can't find `002-dont-include-winmain-in-static.patch` or `003-fix-cython-decode.patch`.

Alexpux:
@encounter I'm push missing patches

Alexpux:
@encounter  ping?

lazka:
No response, so closing. Feel free to resubmit if this is still a problem.

PR #3252, opened 2018-01-02
add a minor patch to gtkglext to fix building natively with makepkg-mingw

cbrt64:
Someone on IRC ran into this one while building packages on msys2 in Windows.
Basically `glib-mkenums --fprod "/* foo */"` was emitting `C:/mingw64/* foo */`, which of course confused the compiler. A newline before the comment, borrowed from the next Makefile rule, gets makepkg-mingw to pass for me.

This problem already has a fix in gtkglext git master, see [GNOME bug 608034](https://bugzilla.gnome.org/show_bug.cgi?id=608034), but this particular package looks unmaintained for a while.

I leave it to the maintainers to decide if this stop-gap solution is the right thing.


lazka:
Something like this should also work: https://github.com/Alexpux/MINGW-packages/pull/3136/files#diff-0e4cff845e0da5affe81f4e4b5167cd0R50

cbrt64:
Yes, using MSYS2_ARG_CONV_EXCL looks like a much better idea. Tested and updated.

PR #3260, opened 2018-01-03
Rebuild glibmm/fontconfig (without _time32 bug in mingw-w64)

Ede123:
Last build was with buggy version of mingw-w64 which was linking
with _time32 which is not available on Windows XP

(see https://sourceforge.net/p/mingw-w64/support-requests/133/)


Ede123:
poppler needs a rebuild, too (I saw it is WIP already)

More binaries containing the broken _time32 call (determined with grep on my local installation):
[_time32.txt](https://github.com/Alexpux/MINGW-packages/files/1602029/_time32.txt)

mingwandroid:
I don't think XP should be supported anymore. Providing software that runs on that OS does people a disservice since it encourages them to stay on an operating system that does not receive security patches.

Ede123:
I agree. However in this case there was an issue in the toolchain that is now fixed and shipping binaries that are known to be broken on purpose does not seem to be the proper way to end support for an OS...

mingwandroid:
It's a fine way to stop support IMHO, it's what everyone does to effect such a change.

Ede123:
Yes, this is now outdated.

Issue #200, opened 2014-09-15
wkhtmltopdf 


Shreeshrii:
Is there a wkhtmltopdf  package?

PR #3280, opened 2018-01-10
Exiv2: backport a memory allocation patch

peterbud:
We have discovered a serious [performance problem](https://github.com/Exiv2/exiv2/issues/200), this simple patch is addressing the current release until the new release will fix this on a permanent basis.

### DIFF
--- a/mingw-w64-exiv2/0002-memoryallocation.patch
+++ b/mingw-w64-exiv2/0002-memoryallocation.patch
@@ -1,24 +1,30 @@
 --- exiv2-0.26.orig/src/basicio.cpp.orig	2017-04-26 21:16:21.000000000 +0200
-+++ exiv2-0.26.orig/src/basicio.cpp	2018-01-10 22:50:16.497801700 +0100
-@@ -1093,10 +1093,11 @@
++++ exiv2-0.26.orig/src/basicio.cpp	2018-01-12 17:11:54.815825700 +0100
+@@ -1093,10 +1093,12 @@
      void MemIo::Impl::reserve(long wcount)
      {
          long need = wcount + idx_;
-+        long blockSize = 2*1024*1024;   // 32768
++        long    blockSize =     32*1024;   // 32768           `
++        long maxBlockSize = 4*1024*1024;
  
          if (!isMalloced_) {
-             // Minimum size for 1st block is 32kB
+-            // Minimum size for 1st block is 32kB
 -            long size  = EXV_MAX(32768 * (1 + need / 32768), size_);
++            // Minimum size for 1st block
 +            long size  = EXV_MAX(blockSize * (1 + need / blockSize), size_);
              byte* data = (byte*)std::malloc(size);
              std::memcpy(data, data_, size_);
              data_ = data;
-@@ -1107,7 +1108,7 @@
+@@ -1106,8 +1108,10 @@
+ 
          if (need > size_) {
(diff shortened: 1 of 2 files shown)


mingw-w64-exiv2/0002-memoryallocation.patch
```diff
@@ -0,0 +1,24 @@
+--- exiv2-0.26.orig/src/basicio.cpp.orig	2017-04-26 21:16:21.000000000 +0200
++++ exiv2-0.26.orig/src/basicio.cpp	2018-01-10 22:50:16.497801700 +0100
+@@ -1093,10 +1093,11 @@
+     void MemIo::Impl::reserve(long wcount)
+     {
+         long need = wcount + idx_;
++        long blockSize = 2*1024*1024;   // 32768
```
mati865: It's `2097152` instead of `32768`.
Not like it changes anything but it looks weird 😄

clanmills: This change is not ready to be submitted as I have explained in #200.  I'm surprised to see this raised as a PR while it's work in progress.

clanmills:
I've updated this patch for reasons explained in #200.  I'd like to more testing and instrumentation on this change before it is submitted to the code base.

mingw-w64-exiv2/0002-memoryallocation.patch
clanmills: I agree it looks weird.  The line of code really means blockSize = 2mb ; // used to be 32768

This is 'work-in-progress' and not intended to be submitted to 'master' at this time.

peterbud:
Ok, let's wait until you merge a final solution to the exiv2 master, I'll add as a patch here then. Apologies for the rush, just we have an issue which impacts a few hundred/thousand users.

clanmills:
You've given me another reason to be cautious!  Better to take another day and be sure that the fix is solid!  I'll do the work I have in mind today and give you an update.  With good fortune, we'll get the patch into MinGW tomorrow.

clanmills:
I've tested the following code and happy for this to be added to MinGW in src/basicio.cpp (to replace the existing code around line 1093).  This patch will be added to Exiv2 'master' in the next few days.
```
    void MemIo::Impl::reserve(long wcount)
    {
        long need = wcount + idx_;
        long    blockSize =     32*1024;   // 32768           `
        long maxBlockSize = 4*1024*1024;

        if (!isMalloced_) {
            // Minimum size for 1st block
            long size  = EXV_MAX(blockSize * (1 + need / blockSize), size_);
            byte* data = (byte*)std::malloc(size);
            std::memcpy(data, data_, size_);
            data_ = data;
            sizeAlloced_ = size;
            isMalloced_ = true;
        }

        if (need > size_) {
            if (need > sizeAlloced_) {
                blockSize = 2*sizeAlloced_ ;
                if ( blockSize > maxBlockSize ) blockSize = maxBlockSize ;
                // Allocate in blocks
                long want      = blockSize * (1 + need / blockSize );
                data_ = (byte*)std::realloc(data_, want);
                sizeAlloced_ = want;
                isMalloced_ = true;
            }
            size_ = need;
        }
    }
```

peterbud:
@clanmills , I have added the revised patch here and tested, still works OK, and performance problem solved.
@Alexpux Can be pushed to master IMHO

PR #3285, opened 2018-01-13
snappy: update to 1.1.7

Adsun701:
This updates snappy to latest version, 1.1.7. Snappy has switched from autotools to CMake.


Alexpux:
@Adsun701 snappy not buildable. Fix it

PR #3332, opened 2018-01-30
gdb: Switch to Python 3

lazka:


mingw-w64-gdb/PKGBUILD
```diff
@@ -77,7 +82,7 @@ build() {
     --disable-rpath \
     --with-system-gdbinit=${MINGW_PREFIX}/etc/gdbinit \
     --with-system-readline \
-    --with-python=${MINGW_PREFIX}/bin/python-config-u.sh \
+    --with-python=${MINGW_PREFIX}/bin/python3 \
```
Alexpux: Why not use python3-config.sh here?

lazka: https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=gdb/configure.ac;h=125e5f30e65a7b046b5fab038bde3c5a2fe485c3;hb=HEAD#l748 suggested to pass a python executable. Is there any downside?

Alexpux: Well, I don't exacly remember why we pass shell script for Python instead executable, maybe because of some path mangling issue. By link that you give it can be not executable:
````
NOTE: This needn't be the real python executable.
 In a cross-compilation scenario (build != host), this could be a shell script that provides what python-config.py provides for  --ldflags, --includes, --exec-prefix.
````

Alexpux: I will try with your changes. If some issue then back to shell script

Issue #3335, opened 2018-01-31
GDI+ gdk-pixbuf-loader performance issues


hitchfred:
Hi. I've been having issues with the gdk-pixbuf package. My Gtk 3 program stalls loading larger jpegs, and the GdkPixbufLoader does not produce area-updated signals for JPEGs as it does for PNGs. I've tracked it down to the package's use of GDI+ pixbuf loaders. After rebuilding gdk-pixbuf to use libjpeg, etc, my program becomes as responsive as it is on Linux, and area-updated signals are correctly produced.

Would it be possible to change the gdk-pixbuf package to disable the GDI+ loaders in favor of libjpeg, libtiff, etc loaders?

I think this problem has come up before:
https://mail.gnome.org/archives/gtk-list/2010-February/msg00128.html

lazka:
@Ede123 looks like you enabled them, thoughts?

lazka:
Thanks for the explanation. Would there by any downside in just using the gdi loads for formats for which no other loader exists (tiff/jpeg/..)?

lazka:
I did a small test using random jpeg images between 90x90 and 600x600, and the decoding with the gdi loader is between 10x and 20x times slower.

Regarding dependencies, I don't mind (also the current package depends on libjpeg/libtiff while not even linking against it..)

lazka:
I've opened #3339 feedback welcome

lazka:
@Ede123 the image in the Ubuntu bug loads fine with the libjpeg loader. In 0.088sec vs 2.075sec with gdi+

Ede123:
Sounds promising! I'll do some tests (probably this evening) to see if I can find any regressions, otherwise I'm fine with merging.

lazka:
The new package is in the repos.

hitchfred:
Closing. With the new package, everything is fine. Thanks

PR #3339, opened 2018-02-01
gdk-pixbuf2: build all available loaders and only use gdi loaders where needed. See #3335

lazka:
gdk-pixbuf only has a mode for building gdi loaders and then skipping
all formats supported through gdi, and not building gdi and building
all formats.

Since the gdi loaders can be slow, but we still want to get the extra format
support not covered by other loaders build all loaders including gdi
and then delete the gdi loaders where the format is covered by other loaders.


lazka:
This is the result:

```
> ls -1 loaders/*.dll
loaders/libpixbufloader-ani.dll*
loaders/libpixbufloader-bmp.dll*
loaders/libpixbufloader-gdip-emf.dll*
loaders/libpixbufloader-gdip-wmf.dll*
loaders/libpixbufloader-gif.dll*
loaders/libpixbufloader-icns.dll*
loaders/libpixbufloader-ico.dll*
loaders/libpixbufloader-jasper.dll*
loaders/libpixbufloader-jpeg.dll*
loaders/libpixbufloader-png.dll*
loaders/libpixbufloader-pnm.dll*
loaders/libpixbufloader-qtif.dll*
loaders/libpixbufloader-tga.dll*
loaders/libpixbufloader-tiff.dll*
loaders/libpixbufloader-xbm.dll*
loaders/libpixbufloader-xpm.dll*
```

Ede123:
One question: In #2376 I specifically used `--with-included-loaders` to avoid https://bugzilla.redhat.com/show_bug.cgi?id=795152.

This PR seems to re-introduce individual modules, so I wonder it it doesn't re-introduce said bug, too?

lazka:
I'll have a look

lazka:
Yeah, that's still a problem. While I couldn't find a "emf" file, loading a wmf and gif with two separate loaders crashes:
```
<__gi__.GdkPixbufGdipAnim object at 0x02c9af30 (GdkPixbufGdipAnim at 0x02a8b310)>
a.py:16: Warning: cannot register existing type 'GdkPixbufGdipAnim'
  print(GdkPixbuf.PixbufAnimation.new_from_file("a.gif"))
a.py:16: Warning: g_once_init_leave: assertion 'result != 0' failed
  print(GdkPixbuf.PixbufAnimation.new_from_file("a.gif"))
a.py:16: Warning: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
  print(GdkPixbuf.PixbufAnimation.new_from_file("a.gif"))
Segmentation fault
````

lazka:
This now includes all gidplus loaders in the main library and just doesn't load the ones where we have external ones.

Also included: a script which tests each loader and checks if it's using gdiplus or not
I hope that will make it easier to update things in the future.

Ede123:
Looking good as far as I can see.

I'm having some issues with the following two files (.ico and .tif): [raster.zip](https://github.com/Alexpux/MINGW-packages/files/1687107/raster.zip)
They won't reliably load but I suspect it might have to do with differences in how metadata is handled.

Also I can *not* load https://bugs.launchpad.net/inkscape/+bug/1746054/comments/5 (the image from the bug linked in #3335) -  @lazka are you sure you tested the correct file?

So in conclusion my first hunch seems to be confirmed with a slightly larger set of test files: GDI+ loaders seem more robust in comparison to their free counterparts.
Given the performance is significantly higher (in my "real world" tests in Inkscape it's about a factor of 5) I guess it's fine to prefer the non-GDI+ loaders, though. Bugtrackers of downstream projects will show if users agree. ;-)

lazka:
Thanks for testing, I'll have another look at those cases.

Regarding bugs, while gdk-pixbuf is not very well maintained, it would be nice if bugs would get reported upstream, so you get the benefit on Linux as well.

lazka:
the bug report jpeg:
It works with new_from_file(), but fails when using the pixbuf loader. Should be reported upstream.

tif.tif: It works with the pixbuf loader, but fails with new_from_file(). LRN has commited something related upstream (https://git.gnome.org/browse/gdk-pixbuf/commit/?id=ce52cefbbc7c6910cd6fd99c7321292b91202009) so loading from files might be indeed broken there.

ico.ico: this doesn't even load here with mspaint, the default win7 image viewer and gimp.. so I'd say it's fair that it's not loading.

Based on that, and with my limited experience, I'd switch the tiff loader back to gdi+ for now, report the jpeg upstream and leave the rest as is.

(and I have to extend the test script to test more api, as they clearly have different code paths for the loading part)

Ede123:
> ico.ico: this doesn't even load here with mspaint, the default win7 image viewer and gimp.. so I'd say it's fair that it's not loading.

Works fine in Paint for me (Windows 10) and shows in Explorer. The file was created with IrfanView.

lazka:
Added the libtiff patch from upstream, the tif.tif file works fine now.

re ico: should I enable the gdi loader for ico?

peterbud:
It's a bit sidetrack, and most likely an upstream question, but nevertheless: when you are testing the modules could you also test files with non-ascii character names in the file names like `Приве́тनमस्ते שָׁלוֹ.tiff`. I recall that e.g. that plain `TIFFOpen()` cannot open such files, and you need `TIFFOpenW()` for that, and I don't know how `gdk-pixbuf2` is coping with such filenames with different file format modules.

lazka:
This is using TIFFFdOpen and not TIFFOpen. But, while looking at the code I see that it casts handles to int, which wont work on 64bit Windows, and there is a bug from 2008 http://bugzilla.maptools.org/show_bug.cgi?id=1941

Alexpux:
@lazka maybe need try the patch from this bug report?

lazka:
I just learned that HANDLE can be cast to int in most cases, so ignore all this :P  https://bugzilla.gnome.org/show_bug.cgi?id=788388#c17

lazka:
@peterbud I've added some unicode to the filename tests of the test script and everything seems to work fine.

lazka:
@Ede123 :+1: or :-1:

Ede123:
Yes, I'm fine with merging...
As I stated above the advantages probably outweigh the few remaining issues (which we can try to address upstream if time permits).

lazka:
@Alexpux please merge

lazka:
I've filed some bugs upstream:

https://bugzilla.gnome.org/show_bug.cgi?id=794015
https://bugzilla.gnome.org/show_bug.cgi?id=794019
https://bugzilla.gnome.org/show_bug.cgi?id=794021

PR #3346, opened 2018-02-04
appveyor: disable test discovery

lazka:
With a mingw gdb build the auto test discovery takes 1.5 minutes, since
we don't have any tests disable them.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,8 @@ build_script:
     - C:\msys64\usr\bin\pacman --noconfirm --sync --refresh --refresh --sysupgrade --sysupgrade
     - C:\msys64\usr\bin\bash --login -c "$(cygpath ${APPVEYOR_BUILD_FOLDER})/ci-build.sh"
 
+test: off
+
 artifacts:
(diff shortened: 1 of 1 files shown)


Ede123:
https://gitlab.com/inkscape/inkscape/blob/master/buildtools/appveyor.sh#L12-13 shaves off another 50 seconds in the example if you want to optimize further.

lazka:
@Ede123 good idea, I've reused your snipped on all my repos. -> #3347

peterbud:
That inkscape appveyor script is great stuff @Ede123 - I have reused many of those in our [appveyor script](https://github.com/darktable-org/darktable/blob/master/.ci/ci-script-windows.sh). What about the fontconfig file? That is also super time consuming. Or that is not applicable always?

Ede123:
@peterbud Thanks!

Regarding fontconfig I'd be cautious to enable it by default as we do not know if packages might rely on certain fonts being available (for example Inkscape used to crash if no fonts at all were found). Not sure how relevant this is while building, though (I assume certain tests could potentially be affected)

PR #3350, opened 2018-02-06
Improvements to asciidoctor build.

JPeterMugaas:
1) Fix hardcoded pathes causing bug report ( https://github.com/Alexpux/MINGW-packages/issues/3338 )
2) Along similar lines, added .install files that fix the batch files to use DOS path references that should be valid.  This has to be done on the uer's system.
3) Add testing from the ruby gem.


mingw-w64-asciidoctor/PKGBUILD
```diff
@@ -4,24 +4,42 @@ _realname=asciidoctor
...
 
+  local _ruby_exe=$(cygpath -m ${MINGW_PREFIX}/bin/ruby.exe)
+  local _inst_dir=$(cygpath -m ${pkgdir}/${MINGW_PREFIX}/bin/)
+  local _w_ruby_exe=$(cygpath -w ${MINGW_PREFIX}/bin/ruby.exe | sed 's/\\/\\\\/g')
+
+  for _f in "${pkgdir}${MINGW_PREFIX}"/bin/*; do
+    sed -e "s|${_ruby_exe}|${MINGW_PREFIX}/bin/ruby|g" \
+        -e "s|${_w_ruby_exe}|${MINGW_PREFIX}/bin/ruby|g" \
+        -e "s|${_inst_dir}|${MINGW_PREFIX}/bin/|g" -i ${_f}
+  done
```
Alexpux: Really wrong for me.
After build we have two types of paths to convert in bat files: 
1. pkgdir
2. MINGW_PREFIX

Don't need to write ````/bin/ruby.exe```` as it already in path. 
You need just sed  ````pkgdir```` converted to windows to empty line and MINGW_PREFIX converted to windows to unix MINGW_PREFIX

PR #3365, opened 2018-02-09
add hdf4 package (for use in gdal)

jeroen:
The hdf4 library has been superseded by hdf5, but many scientists need both hdf4 as well as hdf5 to do their job (i.e. work with legacy systems in gdal).

This is mostly to enable the gdal driver, i.e. we could build gdal with `--enable-hdf4 --enable-hdf5`. The build script is a bit of a hack because the upstream authors only test with MSVC, so we need to disable some `_WIN32` macros. The resulting libs has been tested to support `--enable-hdf4` in `mingw-w64-gdal`.

### DIFF
--- a/mingw-w64-hdf4/PKGBUILD
+++ b/mingw-w64-hdf4/PKGBUILD
@@ -1,0 +1,57 @@
+# Maintainer: Jeroen Ooms <jeroen@berkeley.edu>
+
+_realname=hdf
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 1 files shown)


mingw-w64-hdf4/PKGBUILD
```diff
@@ -0,0 +1,57 @@
+# Maintainer: Alexey Pavlov <alexpux@gmail.com>
```
Alexpux: I'm not a maintainer of this package. Write yourself here

jeroen:
Sorry for that, I have now updated the maintainer field.

PR #3376, opened 2018-02-13
mingw-w64-python-wheel - 0.0.0 - New package - A built-package format…

JPeterMugaas:
… for Python


mingw-w64-python-wheel/PKGBUILD
```diff
@@ -0,0 +1,113 @@
+# Maintainer: Some One <some.one@some.email.com>
```
Alexpux: Fix maintainer field

PR #3387, opened 2018-02-14
Make sure that pytest executables actually run so we use those with o…

JPeterMugaas:
…ur packages.

### DIFF
--- a/mingw-w64-python-pytest/PKGBUILD
+++ b/mingw-w64-python-pytest/PKGBUILD
@@ -1,10 +1,11 @@
 # Contributor: Frederic Wang <fred.wang@free.fr>
 
-_realname=pytest
+_pyname=pytest
+_realname=${_pyname}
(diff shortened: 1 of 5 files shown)


lazka:
What does this fix?

JPeterMugaas:
Alright, without this, trying to execute something like "py.test" from the command line would not work.  Without it, it would probably complain about something not being found in the /usr path.

lazka:
executing py.test works fine here... but ok..

JPeterMugaas:
I think what it could be is that I build mingw-w64 packages and do testing from the MSYS prompt using the MSYS menu item as opposed to the "mingw64" or "mingw32" menu item.    But such things can be telling because some scripts that were built for mingw-w64 could mistakenly be run in the MSYS python interpreter and vice-versa.

lazka:
Would changing the shebang to just "#!python3.exe" work? It seems to here.

PR #3388, opened 2018-02-14
global: Update to GNU Global 6.6.2

akmll:


filnet:
Your commits contain something from 2016 and a merge.
You should squash everything into a single commit.

PR #3391, opened 2018-02-15
osgearth: fix linking against rocksdb

filnet:


Alexpux:
First, you forgot to commit patches.
Second, I think patches will not help because rocksdb shared library dont export C++ symbols needed for osgEarth.

filnet:
Will add the missing patches.

cache_rocksdb (the osgearth plugin that uses rocksdb) statically links rocksdb.

Issue #3369, opened 2018-02-11
No OpenGL SOIL package


AntonioCS:
Hey guys,

Is there a possibility to include the lib [SOIL](http://www.lonesock.net/soil.html) into a package?

Thanks!

ukhf-jm:
Why not use something like [stb_image](https://github.com/nothings/stb/blob/master/stb_image.h)? It's a single header which you can just include in your project.  
I'd hazard a guess and say that you're following a tutorial like [open.gl](https://open.gl/)? If so then the updated version at [learnopengl.com](https://learnopengl.com) would be better to follow.

AntonioCS:
@ukhf-jm I am trying to compile this game: https://github.com/ryanfleury/global_game_jam_2018
They use SOIL so I though it might be easy to add to msys2.
The learnopengl.com site has comments from 4 years ago. Is that the most up to date site to learn opengl?

MenaceInc:
@AntonioCS yes, the learnopengl.com site is perfectly fine. It focuses on the basics of modern OpenGL (version 3.3 onwards using the core profile) but has newer resources than open.gl.

Regarding that Global Game Jam game, I'm not sure if it can be compiled using mingw-w64 as it uses EFX-Util.h. I'll have a quick look at it though.

AntonioCS:
@MenaceInc Thanks! I did post about it here: https://github.com/Alexpux/MINGW-packages/issues/3368
Thought it might have been a bug or something.

MenaceInc:
@AntonioCS I've managed to get it compiled and running but I'm not happy about how the project is structured right now. You can find my fork of the game here: https://github.com/MenaceInc/global_game_jam_2018

Due to the lack of the SOIL library being available in MINGW-packages, I've had to include the static lib from the release zip from here: http://www.lonesock.net/soil.html

This means that it only compiles and works with the 32bit mingw-w64 compiler. You'll need to run `pacman -Sy mingw-w64-i686-gcc mingw-w64-i686-make mingw-w64-i686-glew mingw-w64-i686-glfw mingw-w64-i686-openal mingw-w64-i686-cmake` from MSYS2 and then build it from the MinGW 32bit shell.

Clone the repo and navigate to the root from within a MinGW 32bit shell then run the following:
```
mkdir build && cd build
cmake .. -DCMAKE_MAKE_PROGRAM=mingw32-make -G "MSYS Makefiles"
mingw32-make
```

[...]

PR #3405, opened 2018-02-17
Add SOIL package

MenaceInc:
Tested using
```
MINGW_INSTALLS=mingw64 makepkg-mingw -sLf
pacman -U mingw-w64-x86_64-soil-1.16.0-1-any.pkg.tar.xz
```
as well as
```
MINGW_INSTALLS=mingw32 makepkg-mingw -sLf
pacman -U mingw-w64-i686-soil-1.16.0-1-any.pkg.tar.xz
```

Both worked just fine when building a project with a SOIL dependency.

Related to issue #3369

### DIFF
--- a/mingw-w64-soil/PKGBUILD
+++ b/mingw-w64-soil/PKGBUILD
@@ -1,0 +1,35 @@
+# Maintainer: John Murray <john@menaceinc.com>
+
+_realname=soil
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.16.0
+pkgrel=1
+pkgdesc="C library used for loading image files into OpenGL (mingw-w64)"
+arch=('any')
+url='http://lonesock.net/soil.html'
+license=('Public Domain')
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+source=("http://www.lonesock.net/files/${_realname}.zip")
+sha256sums=('a2305b8d64f6d636e36d669bbdb0ca5445d1345c754b3d61d3f037dad2e5f701')
+
+build() {
+  mkdir -p ${srcdir}/build-${MINGW_CHOST}/obj
+  cd "${srcdir}/Simple OpenGL Image Library/projects/makefile"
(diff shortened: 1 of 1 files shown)


mingw-w64-soil/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.16.0
+pkgrel=1
+pkgdesc="C library used for loading image files into OpenGL (mingw-w64)"
+arch=('any')
+url='http://lonesock.net/soil.html'
+license=('Public Domain')
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+source=("http://www.lonesock.net/files/${_realname}.zip")
```
Alexpux: download in named and versioned archive

MenaceInc: Unfortunately, the original developer has not made the library available in a format using versions. Would it be better to have a mirror on GitHub and use that? There hasn't been any updates to the library in almost 10 years.

Alexpux: Look at other PKGBUILDS how we do it. It is simple.

MenaceInc: I have looked at a number of other PKGBUILDS but they all used GitHub or SourceForge or their own hosted solutions that have the package version as part of the URL.

The original developer of SOIL didn't make the library available with a URL using version numbers. Only "http://www.lonesock.net/files/soil.zip".

Considering the situation, would you prefer that I mirrored the library on GitHub and used an archive URL with a version number?

Alexpux: You need just add
````${_realname}-${pkgver}.zip::```` at the start of source URL

MenaceInc: Oh, right. I wasn't aware that was doable. Sorry.

I suppose that makes sense as the SHA will change anyway if the original developer updates the library. I'll edit the PKGBUILD now, thank you.

MenaceInc:
Updated the PKGBUILD as per request and also changed where the .a file was being placed so that no build artifacts are in the source directory.

PR #3419, opened 2018-02-19
python-virtualenv: fix post-install script for python2-virtualenv

Ede123:


Ede123:
That was fast, thanks!

PR #3420, opened 2018-02-20
suitesparse: Install static libraries

drhagen:
The compilation currently generates the static libraries, but does not install them. This adds a patch which modifies the makefiles to copy the static libraries to the install directory.


Alexpux:
Increase pkgrel

PR #3423, opened 2018-02-21
global: Fix segmentation fault caused by NULL parameter

akmll:
It seems the preferred build environment is different with the pull request pre-check, because the sha256 checksum of downloaded source code archive is different, see [this commit](https://github.com/Alexpux/MINGW-packages/commit/3c8999f9cb9073c7b541633cbe535431fb2815ce).


Alexpux:
Increase pkgrel

akmll:
Sorry, missed that when merging manually. Is it OK to **SKIP** the sha256 checksum of the downloaded source archive, or any suggestion to make it valid across all environments?

PR #3425, opened 2018-02-21
add colorspacious python module

briend:
Simple to use module for doing color conversions in CIECAM, XYZ, sRGB, etc


briend:
Looks like it failed because of a temporary http problem. Should I push a commit to trigger a new build?

PR #3429, opened 2018-02-21
gdal: enable hdf4 driver

jeroen:


jeroen:
Can't get this to work. It was working for me locally. I think the gdal autoconf script assumes libhdf is a static library, but @alexpux actually added the shared libs in the last `mingw-w64-hdf4` update.

PR #3438, opened 2018-02-23
Remove make parallel opts and add pinentry dep to gnupg

yuui-tanabe:
Removed make parallel opts because makepkg setted MAKEOPTS and added pinentry dependency to gnupg


yuui-tanabe:
Because Tea CI can't install missing dependencies he build was failed

PR #3449, opened 2018-02-26
djvulibre - 3.5.27 add libtiff and libiconv as dependencies

JPeterMugaas:
graphicsmagic - 2.3.28 - add gcc-libs, libwinpthreads, and txt2html as dependencies.  Rebuilding using txt2html to generate HTML documentation.
pdf2djvu - 0.9.8 - add exiv2, gettext, libiconv, poppler to dependendencies.  Add nose2 to checkdepends for future-proofing
djview - 4.10.6 - New package.  Portable DjVu viewer using the QT5 library.

### DIFF
--- a/mingw-w64-djview/PKGBUILD
+++ b/mingw-w64-djview/PKGBUILD
@@ -1,0 +1,41 @@
+# Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
+
+_realname=djview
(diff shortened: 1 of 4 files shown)


lazka:
`error: target not found: txt2html`

lazka:
ah, you added that in msys2, never mind!

PR #3476, opened 2018-03-07
New package: scummvm-2.0.0

rootfather:
After learning how to make PKGBUILDS, I thought it might be funny if there's a mingw-package of ScummVM (https://www.scummvm.org).

So here we go.


filnet:
I want Doom now.

rootfather:
Let me see what I can do about it. :)

rootfather:
@Alexpux, can you tell me how long it takes on average to make new packages available in the repository?

rootfather:
Hm... still not available in the repo?

PR #3487, opened 2018-03-13
glib-openssl: New package

lazka:
Allows glib to use openssl for TLS instead of gnutls

### DIFF
--- a/mingw-w64-glib-openssl/fix-cert-path.patch
+++ b/mingw-w64-glib-openssl/fix-cert-path.patch
@@ -1,0 +1,11 @@
+--- glib-openssl-2.50.8/tls/openssl/gtlsbackend-openssl.c.orig	2018-02-26 11:03:32.000000000 +0100
++++ glib-openssl-2.50.8/tls/openssl/gtlsbackend-openssl.c	2018-03-13 07:58:22.822305900 +0100
+@@ -191,7 +191,7 @@
+   gchar *cert_path;
+ 
+   module_dir = g_win32_get_package_installation_directory_of_module (NULL);
+-  cert_path = g_build_filename (module_dir, "bin", "cert.pem", NULL);
++  cert_path = g_build_filename (module_dir, "ssl", "cert.pem", NULL);
+   g_free (module_dir);
+ 
+   if (g_file_test (cert_path, G_FILE_TEST_IS_REGULAR))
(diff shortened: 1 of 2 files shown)


lazka:
(@nacho fyi)

mingw-w64-glib-openssl/fix-cert-path.patch
```diff
@@ -0,0 +1,11 @@
+--- glib-openssl-2.50.8/tls/openssl/gtlsbackend-openssl.c.orig	2018-02-26 11:03:32.000000000 +0100
++++ glib-openssl-2.50.8/tls/openssl/gtlsbackend-openssl.c	2018-03-13 07:58:22.822305900 +0100
+@@ -191,7 +191,7 @@
+   gchar *cert_path;
+ 
+   module_dir = g_win32_get_package_installation_directory_of_module (NULL);
+-  cert_path = g_build_filename (module_dir, "bin", "cert.pem", NULL);
++  cert_path = g_build_filename (module_dir, "ssl", "cert.pem", NULL);
```
nacho: I'd be more than ok getting this upstream

lazka: Not sure what's the right thing to do here, gvsbuild obviously expects it in the other dir.
Maybe asking openssl for the directory might work, I'll look into that.

lazka: @nacho

Something like this?

```diff
diff --git a/tls/openssl/gtlsbackend-openssl.c b/tls/openssl/gtlsbackend-openssl.c
index f118c53..4b44e1d 100644
--- a/tls/openssl/gtlsbackend-openssl.c
+++ b/tls/openssl/gtlsbackend-openssl.c
@@ -187,12 +187,15 @@ g_tls_backend_openssl_real_create_database (GTlsBackendOpenssl  *self,
   GTlsDatabase *database;
 
 #ifdef G_OS_WIN32
-  gchar *module_dir;
   gchar *cert_path;
+  const char *cert_file;
 
-  module_dir = g_win32_get_package_installation_directory_of_module (NULL);
-  cert_path = g_build_filename (module_dir, "bin", "cert.pem", NULL);
-  g_free (module_dir);
+  cert_file = g_getenv (X509_get_default_cert_file_env ());
+
+  if (!cert_file)
+    cert_path = g_strdup (X509_get_default_cert_file ());
+  else
+    cert_path = g_strdup (cert_file);
 
   if (g_file_test (cert_path, G_FILE_TEST_IS_REGULAR))
     anchor_file = cert_path;
```

nacho:
this is interesting, I wonder if we should do it for linux as well

nacho:
@lazka can you file a ticket upstream with the patch so we can do a proper review?

lazka:
https://bugzilla.gnome.org/show_bug.cgi?id=795782

PR #3493, opened 2018-03-17
appveyor: Reduce PATH to the default Windows entries + git

lazka:
PATH on appveyor contains 80+ directories and we don't need any of it.


lazka:
@Alexpux any thoughts on this?

PR #3520, opened 2018-03-28
openssl: Update to 1.0.2o

larskanis:
Out of curiosity: For [RubyInstaller](https://rubyinstaller.org) we're using OpenSSL-1.1.0 already, with an adjusted path-relocation patch. Are there plans to switch MSYS2/MINGW to OpenSSL-1.1.0 as well?

See: https://github.com/oneclick/rubyinstaller2-packages/tree/master/mingw-w64-openssl


Alexpux:
@larskanis need to check that all packages compatible with it.

larskanis:
@Alexpux When will you update the package repository?

larskanis:
I noticed, the new packages are built and uploaded now. Thank you very much!

PR #3522, opened 2018-03-30
gstreamer: update to 1.14.0

filnet:
see https://github.com/Alexpux/MINGW-packages/issues/3504


lazka:
I just saw you posted to the gst mailing list about the missing directsoundsink. I've filed a patch for it recently: https://bugzilla.gnome.org/show_bug.cgi?id=795750

lazka:
Maybe we should include it in msys2?

lazka:
> I suggest we wait for your patch to be merged upstream.

ok, sounds good.

> In the meantime, there is a workaround::
> 
> Delete the registry file located here:
> C:\Users\<user>\AppData\Local\Microsoft\Windows\INetCache\gstreamer-1.0
> Note that INetCache is a hidden directory.

yeah, problem is can't tell every user to do that when I ship 1.14.

filnet:
Note that the d3dvideosink plugin was renamed to d3d. So same issue there.

lazka:
Yeah. Maybe you can link to the bugzilla bug in the ML thread to give it some attention?

filnet:
Will do.

It is a recurring issue : https://bugzilla.gnome.org/show_bug.cgi?id=785168

lazka:
Thanks!

PR #3542, opened 2018-04-04
wxWidgets: Update to 3.0.4

stahta01:
And, commented out the check function I added in the recent past.
I decided it makes no sense to run the check function just to build the samples.

Note: The sample and the tests do build now. Did not run either the samples or the tests.
In version 3.0.3, the non gui test seg faulted; and, that is likely still true.

Tim S.


BrendanSimon-SEPL:
Apparently this is merged, but I can only see wxWidgets 3.0.2 with pacman.
Where is 3.0.4?

Or more importantly, where is 3.1.2?  This also is supposed to be merged, but I can't see it with pacman.  See #4748

PR #3546, opened 2018-04-05
glib2: Build with win32 threads model

accel2k:
Win32 threads are much more efficient than winpthreads. Some cases show 10x increase of performance. This patch forces glib to use win32 threads.

### DIFF
--- a/mingw-w64-glib2/PKGBUILD
+++ b/mingw-w64-glib2/PKGBUILD
@@ -78,7 +78,7 @@ build() {
     --disable-static --enable-shared \
     --disable-libelf \
     --with-python=${MINGW_PREFIX}/bin/python3.exe \
-    --with-threads=posix \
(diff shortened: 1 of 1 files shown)


uwehermann:
Please don't make this the only option, provide both options. Some projects and libs need POSIX thread model and cannot work with win32 (e.g. when you need C++11 std::thread support and other stuff).

mingwandroid:
AFAIK software built with both thread models can be mixed in general.

@accel2k, you need to back your claim up with benchmarks too though.

mingwandroid:
And by that I mean a script that can be used to reproduce them.

lazka:
afaik glib doesn't support static builds with win32 threads, see https://bugzilla.gnome.org/show_bug.cgi?id=698118

not that I care about static builds.. :)

PR #3547, opened 2018-04-05
Update for PDCurses 3.6 wincon (formerly win32)

GitMensch:
* updated from PDCurses 3.4 to PDCurses 3.6
* adjustments for port renaming "win32" to "wincon"
* adjustments for renamed and changed Makefile


GitMensch:
Note: it would actually be nice to provide more ports. This package provided the port win32 (now renamed to wincon), but at least SDL2 (and maybe SDL1 and later win32a/wingui) would be a nice addition.
Should these ports be defined as separate packages with the port added to the name and just the dependencies/make flags adjusted?

PR #3571, opened 2018-04-08
meson: Update to 0.45.1 and add patch for gtk-doc fix

lazka:
Also adds https://github.com/mesonbuild/meson/issues/3307


lazka:
@KrullBorg @qme ignoring the doc build error for now, any objections?

PR #3592, opened 2018-04-12
Teach cURL to find the certificate bundle also when validating the certificates of proxies

dscho:
In https://github.com/git-for-windows/git/issues/1493 the problem was identified where ca-bundle.crt was found for regular HTTPS connections, but not for *proxy* connections.

This is the obvious fix for that problem.

While at it, the diff was updated to the current cURL version, and it was also made a proper Git diff (generated from a commit pushed to dscho/MINGW-packages).


dscho:
FWIW I suspect the "tea" CI to fail because it apparently fails to update the package database?

PR #3599, opened 2018-04-13
Add SpellChecker plugin for Qt Creator

dsogari:
SpellChecker plugin version 1.2.0 from https://github.com/CJCombrink/SpellChecker-Plugin


Alexpux:
Is it not enought to use installed qt-creator instead unpacking its sources?

Alexpux:
We dont have Qt Creator plugins packages

lazka:
No update for a long time, so closing.

PR #3607, opened 2018-04-16
mingw-w64-cmake - Update to version 3.11

JPeterMugaas:
Drop 0004 patch about qtstatic plugins sk since I can't figure out what to do about it - see: https://github.com/Alexpux/MINGW-packages/issues/3545 .


Alexpux:
Well I will not update cmake while not fix qt static support

PR #3621, opened 2018-04-20
gtkspell3: drop gtk2 build

lazka:
gtkspell3 is a dependency of geany and gitg, and both are gtk3 now.
No need to pull in gtk2.


lazka:
If you think having the gtk2 build isn't a problem feel free to close this.

I'm just trying to slowly get rid of old stuff :)

Issue #3545, opened 2018-04-04
CMake: update to 3.11


ArekPiekarz:
CMake 3.11 has been released on March 28. Among others it adds support for Boost 1.66, which is currently available in Msys2 repositories. Can you update it please? Thanks in advance.

https://blog.kitware.com/cmake-3-11-0-available-for-download/

TBBle:
This is not a "trivial" update, as MSYS2 carries a patch to support `AUTOSTATICPLUGINS` for Qt, and the auto-generation code for Qt in CMake has been refactored to support parallel codegen. The patch's code needs to be refactored to reflect the same thing.

I started playing with this (since I did the last rebasing, for 3.10.1 in #3262) but the changes are significant enough that I'd need a test case, and I haven't had a chance to try and find an existing use-case to test it against.

ArekPiekarz:
Thanks for the information. Have you tried upstreaming the patch to ease merging in the future?

mingwandroid:
I wrote the patch for `AUTOSTATICPLUGINS` and it depends on some quite extensive changes to Qt too so the upstreaming effort would be extensive (and probably something I'd have to do and I do not have time for that).

TBBle:
Looking at the patch, and trying to find a working use-case (the best I have is `mingw-w64-kcwsh-qt5-git`, perhaps @mingwandroid can point me at a better one since that's failing to link for me right now), it's possible I broke the behaviour in #3262 by accidentally making it depend on also having `CMAKE_AUTORCC` set.

JPeterMugaas:
I hate to say this but I think it might be best to drop the staticplugin capability until someone can work on it.  But ideally, it should actually be in cmake itself.  I have actually made a version of cmake but I dn't port the patch and I don't use that capability.

TBBle:
On examination, `mingw-w64-qt-installer-framework-git` and `mingw-w64-octopi-git` don't use CMake. So I'd have to get either the static build-reqs for `mingw-w64-krita-git` in-place, or work out why `mingw-w64-kcwsh-qt5-git` fails.

mingwandroid:
If no one can find time to work on it then I'm ok with not applying it, but please leave it in the repo so it doesn't get forgotten about.

JPeterMugaas:
I have the time but I don't have the where-with-all the fix the problem and I might have goofed when trying to adjust the patch in previous versions making things worse.  

[...]

PR #3628, opened 2018-04-23
cmake: Update to 3.11.0 (#3545)

TBBle:
The patch series was recreated by applying it to a CMake git checkout at
v3.10.1, and then rebasing it onto v3.11.0.

Patch 0004-Implement-Qt5-static-plugin-support.patch had to be
extensively redone, to support a refactoring in the Qt5 auto-gen support
in CMake 3.11.0. The rest of the patches are changed only due to
artifacts of this rebasing.

### DIFF
--- a/mingw-w64-cmake/0001-Windows-Add-missing-stringapiset.h-include.patch
+++ b/mingw-w64-cmake/0001-Windows-Add-missing-stringapiset.h-include.patch
@@ -1,4 +1,4 @@
-From 1a42d44ec1a582a52d56eb44d4a130d285f7fa18 Mon Sep 17 00:00:00 2001
+From 7618a242db1a0a8792f0fd6b79da26a03ba883ba Mon Sep 17 00:00:00 2001
 From: Ray Donnelly <mingw.android@gmail.com>
 Date: Mon, 23 Jan 2017 12:46:49 +0000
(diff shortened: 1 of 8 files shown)


TBBle:
Per discussion in #3545, there's one current bug here:
* If you use `CMAKE_AUTOSTATICPLUGINS` without `CMAKE_AUTORCC` or `CMAKE_AUTOMOC`, it will still run a script step with no action "cmd /c". I'm not sure what's triggering this to happen.

TBBle:
This also fixes a bug introduced in the CMake 3.9.0 rebasing of patch 0004, where `CMAKE_AUTOSTATICPLUGINS` did not function if `CMAKE_AUTORCC` was not also set. See https://github.com/Alexpux/MINGW-packages/issues/3545#issuecomment-381339386

TBBle:
Also as part of the refactoring, we gained support for the `AUTOSTATICPLUGINS_SOURCE_GROUP` global property, see [AUTORCC_SOURCE_GROUP](https://cmake.org/cmake/help/v3.11/prop_gbl/AUTORCC_SOURCE_GROUP.html) for the analogous use-case.

Alexpux:
@TBBle cmake 3.11.1 released

PR #3634, opened 2018-04-25
cmake: Don't generate .orig artefacts from patches

TBBle:
All files in the Modules tree end up packaged in the final build,
including these patch-generated backups.

A *failed* patch will still generate .rej and .orig artefacts, so this
should not introduce any workflow issues for any package developers.


JPeterMugaas:
I was using the b parameter because that sometimes the backup .orig files weren't being generated and I needed to be certain that they were generated as part of debugging the patches.

TBBle:
Another option would be to add script to the packaging step to delete the .orig files from pkg/, I guess. Seems messier to me.

lazka:
Needs a rebase

lazka:
Thanks!

PR #3636, opened 2018-04-25
gcc-git: Update to GCC8 and remove patches for GCC 5 and GCC 6.

lhmouse:


lhmouse:
@Alexpux Will remove 0012 and re-push.

That color patch was adopted into trunk and is now available naturally in GCC 8.

lhmouse:
Done.

jeroen:
Is there any reason not to copy this set of changes to `mingw-w64-gcc`?

lhmouse:
It is just because I don't maintain `mingw-w64-gcc`.

PR #3643, opened 2018-04-27
libsodium - 1.0.16 - Update to latest version and use minisgn tool si…

JPeterMugaas:
…nce it's now impossible to retreive the GPG key for libsodium

minisign - 1.0.8 - Used by libsodium package to verify libsodium distribution.  New package

### DIFF
--- a/mingw-w64-libsodium/PKGBUILD
+++ b/mingw-w64-libsodium/PKGBUILD
@@ -10,16 +10,8 @@ arch=("any")
 url="https://github.com/jedisct1/libsodium"
 license=('custom:ISC')
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
-makedepends=("${MINGW_PACKAGE_PREFIX}-minisign")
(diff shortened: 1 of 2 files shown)


Alexpux:
@JPeterMugaas I dont think we need minisign package as we verify with gnupg

JPeterMugaas:
The problem was that I couldn't download the public pgp-keys for libsodium using gpg --recv-key which is what I usually would do for signed downloads.  So I decided to do what Archlinux had done with their libsodium package and I need that minisign to verify.

I am keenly aware of that the choice to use minisign to verify does create a chicken and egg problem since it too depends upon libsodium.  I was actually tempted to make a libsodium and minisign for MSYS and that could be an option if you want libsodium to be in MSYS as well.

JPeterMugaas:
Alright, I'll remove minisign usage from the libsodium package but include it in the pull request.  Minisign might be useful for some developers and it does depend upon libsodium.

PR #3646, opened 2018-04-28
Upgrade node.js to v8.11.1

dscho:
When node.js broke for the umptienth time due to an upgrade of `mingw-w64-icu` to a newer version (and hence the respective `.dll` files cannot be found anymore due to the version number being encoded in the file name), Git for Windows needed a solution, fast, as it relies on node.js to be working for lots of its release automation.

The solution by Git for Windows was not to try to get an obsolete node.js version to compile with an up to date GCC toolchain, but instead upgrade node.js to the current version (v8.11.1 LTS).

This Pull Request offers the same improvement to MSYS2.

The PR is actually *not* based on current `master`, as MSYS2 chose to try to work around compile errors of the obsolete node.js version v6.11.0 with current GCC. Instead, this PR is rooted in the commit before that work-around, and eventually merges that commit without taking any of its changes (`git merge -s ours`).

For the benefit of the record, this PR also contains the intermediate upgrade to node.js v6.14.1, which I performed to make the jump to the v8.x release train easier.


mati865:
Patch failed to apply

dscho:
> Patch failed to apply

That's funny because I built these packages successfully. Let me investigate.

dscho:
Oh... You know what? I forgot the crucial upgrade to the latest node.js version...

PR #3651, opened 2018-04-29
json-c: update to 0.13.1_20180305

Adsun701:
This updates json-c to 0.13.1_20180305.


jeroen:
FYI the change in versioning pattern will prevent upgrades:

```
warning: mingw-w64-x86_64-json-c: local (0.13_20171207-2) is newer than mingw64 (0.13.1_20180305-1)
```

PR #3675, opened 2018-05-02
cairo: add some patches posted upstream for improving the drawing performance

lazka:
See https://lists.cairographics.org/archives/cairo/2018-April/028606.html

This should help gtk3 apps without CSD.

@Ede123


lazka:
> Very interesting! Do I understand correctly that this will speed up one of the slowest parts determined for non-CSD Windows in https://bugzilla.gnome.org/show_bug.cgi?id=781153, i.e. copying the window content between surfaces?

As far as I understand, yes. See LRN's follow up (layered is CSD): https://lists.cairographics.org/archives/cairo/2018-April/028609.html

Ede123:
Just rebuilt Inkscape master with the patched cairo and got great results! The speedup is huge and brings performance in the development version to an almost usable level!

(as a matter of fact GTK_CSD=0 now seems even slightly faster than GTK_CSD=1 - before the former was completely blocking the rendering pipeline in certain scenarios!)

Inkscape master (gtk3) is still *a lot* slower than Inkscape 0.92.x (gtk2), but this patch was a huge step, thanks @lazka for spotting!

PR #3723, opened 2018-05-13
x265: update to 2.7

Adsun701:
This updates x265 to 2.7 and enables all bit depths as well as dynamic HDR10.


avih:
I think this update requires ffmpeg rebuild, see #3734 .

PR #3733, opened 2018-05-15
cmake 3.11.3 - rebuild to fix problem with json-cpp dll name.

JPeterMugaas:


Alexpux:
@JPeterMugaas it already rebuilded

JPeterMugaas:
The problem was that I was getting this:

D:/msys/mingw64/bin/cmake.exe: error while loading shared libraries: libjsoncpp-20.dll: cannot open shared object file: No such file or directory

When running cmake using the version from the package repository.

Alexpux:
You seems have your own json-cpp builded or not update it

PR #3739, opened 2018-05-16
Add texinfo to makedepends

jakschu:
binutils needs makeinfo (texinfo) to build


jakschu:
Crap, i missed things up in a hurry.

PR #3757, opened 2018-05-17
add python-more-itertools (version 4.1.0)

Adsun701:
This adds python-more-itertools (version 4.1.0).


Alexpux:
Why you add so many python packages? I don’t want to build all of them just for fun

Adsun701:
A lot of existing python packages now depend on packages that haven't been added yet. For example, python-more-itertools is a dependency of python-pytest.

PR #3758, opened 2018-05-17
New package: python-entrypoints

peterbud:


peterbud:
@Alexpux , I'm working on making `jupyter-notebook` work. These are prerequisite of that package, and sorry if there are a quite few of those prerequisites.
I'm not sure what is the strategy for python packages in this repo: should we add the as package explicitly like in Arch, or use `pip` to install them? In a few cases `pip `is failing to add/compile a more complex package, so I'd lean to add packages.
About _pyname: I'll remove them, I was using the python [PKGBUILD template](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.setup-py-python-pkg) which contains that structure, and I did not realize that that is optional.

Alexpux:
@peterbud  ok I will fix template.

peterbud:
I have updated the new packages, and pushed the changes to the original commits. Now they should be OK, pls check.
There will be a few additional ones, apologies in advance.

PR #3792, opened 2018-05-21
initial package of vapoursynth

ZachBacon:
Initial package of vapoursynth. At the moment the python module is having some issues, but it'll be ready for use with mpv


ZachBacon:
Kinda turned into a little something extra, just minor things really. Vulkan enablement in SDL2, enabling vapoursynth in mpv etc

PR #3806, opened 2018-05-23
WIP: Initial inclusion for hexchat

ZachBacon:
a rough draft for the inclusion of hexchat, meson has some interesting nuances in a PKGBUILD something isn't working quite right, could be something I'm doing, but local build there's something else, seems to not find itstools, may have to talk with tingping about it at some point. But I'll finish this sometime this week for inclusion with msys2


lazka:
afair meson gets confused if you have mingw perl installed because it tries to interpret the shebang (maybe that has been fixed..?). I'd maybe try without mingw perl for starters.

ZachBacon:
One particular issue is actually unrelated to peel. Been working with @tingping on the issue. I'll be working something out once I get off of work.

ZachBacon:
So one thing I did notice is that mingw32 is missing wbemcore while mingw64 has it by default. Might be a bug in the mingw-w64 runtime itself.

ZachBacon:
for now I'll send a pr to the mingw-w64 team for wbemcore and make it a patch for the crt here so that i686 can build, afterwards I plan on submitting a more complete patch for the mingw-w64 team for missing defs from each lib32 and lib64 crt for a bit of completeness

PR #3846, opened 2018-05-27
Add two helper macros to the PKGBUILD templates.  apply_patch_with_ms…

JPeterMugaas:
…g applies a patch ensuring a backup is made and with an output message showing the patch being applied to help with debugging.  del_file_exists delets files and is meant for where a patch creates new files.  Those added files may not be deleted if the PKGBUILD is reran causing patch to fail saying it's already been applied.

Fix the setup-py template to use a _dtoken variable for part of the URL in pypi.  Those use a download token for each software package and is unique to EVERY version.  Also, use python${pver}-build-${CARCH} consistantly.  The CARCH allows us to have for binary versions so that they could be examined if a build problem were to occur and speed up the build.

### DIFF
--- a/mingw-w64-PKGBUILD-templates/PKGBUILD.setup-py-python-pkg
+++ b/mingw-w64-PKGBUILD-templates/PKGBUILD.setup-py-python-pkg
@@ -18,12 +18,14 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-python2"
              "${MINGW_PACKAGE_PREFIX}-python3-setuptools"
              "${MINGW_PACKAGE_PREFIX}-python2-setuptools")
 options=('staticlibs' 'strip' '!debug')
-#older Python packages software might use this instead
-#source=("https://pypi.python.org/packages/source/P/${_realname}/${_realname}-${pkgver}.tar.gz")
-@_dtoken would be the part of the URL between the /package and filename.
-# I did this to make it so you could just update the token a new version instead of the complete
-# download URL.
+#Ideally, you should download the sources from github or some other place besides
+#pypi, but that ideal is not always possible.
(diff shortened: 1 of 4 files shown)


mingw-w64-PKGBUILD-templates/PKGBUILD.setup-py-python-pkg
```diff
@@ -18,18 +18,43 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-python2"
              "${MINGW_PACKAGE_PREFIX}-python3-setuptools"
              "${MINGW_PACKAGE_PREFIX}-python2-setuptools")
 options=('staticlibs' 'strip' '!debug')
-source=("https://pypi.python.org/packages/source/P/${_realname}/${_realname}-${pkgver}.tar.gz")
+#older Python packages software might use this instead
+#source=("https://pypi.python.org/packages/source/P/${_realname}/${_realname}-${pkgver}.tar.gz")
+@_dtoken would be the part of the URL between the /package and filename.
+# I did this to make it so you could just update the token a new version instead of the complete
```
filnet: minor typo: should be "update the token _to_ a new version" ?

peterbud: Based on my experience `dtoken` is not needed, and I find it very difficult to maintain. Source packages can be still referenced in a form of:
`https://pypi.python.org/packages/source/[first letter of the packages]/${_realname}/${_realname}-${pkgver}.tar.gz`

But even better, source packages should directly reference the github (or similar) source and not pypi. Also I see extremely rarely similar `dtoken `solution in Arch repositories.

PR #3847, opened 2018-05-27
mingw-w64-sphinx - 1.7.4 - Update to latest version.  Add sqlalchemy …

JPeterMugaas:
…to dependency

mingw-w64-sqlalchemy - 1.2.7 - New Package - Python 2 SQL toolkit and Object Relational Mapper

### DIFF
--- a/mingw-w64-python-sphinx/PKGBUILD
+++ b/mingw-w64-python-sphinx/PKGBUILD
@@ -5,7 +5,7 @@ _pyname=sphinx
 _realname=sphinx
 pkgbase=mingw-w64-python-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-python2-${_realname}" "${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
-pkgver=1.7.1
+pkgver=1.7.4
 pkgrel=1
 pkgdesc="Python documentation generator (mingw-w64)"
 arch=('any')
@@ -20,6 +20,8 @@ makedepends=(
             "${MINGW_PACKAGE_PREFIX}-python3-requests"
             "${MINGW_PACKAGE_PREFIX}-python3-setuptools"
             "${MINGW_PACKAGE_PREFIX}-python3-six"
+            "${MINGW_PACKAGE_PREFIX}-python3-sqlalchemy"
+            "${MINGW_PACKAGE_PREFIX}-python3-whoosh"
             "${MINGW_PACKAGE_PREFIX}-python2-colorama"
             "${MINGW_PACKAGE_PREFIX}-python2-docutils"
             "${MINGW_PACKAGE_PREFIX}-python2-imagesize"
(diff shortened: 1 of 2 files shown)


k-takata:
Why this reverts the change of #3461?
Now sphinx doesn't work because of the missing dependency to python-packaging.

mingw-w64-python-sphinx/PKGBUILD
```diff
@@ -28,6 +30,8 @@ makedepends=(
             "${MINGW_PACKAGE_PREFIX}-python2-requests"
             "${MINGW_PACKAGE_PREFIX}-python2-setuptools"
             "${MINGW_PACKAGE_PREFIX}-python2-six"
+            "${MINGW_PACKAGE_PREFIX}-python2-sqlalchemy"
+            "${MINGW_PACKAGE_PREFIX}-python3-whoosh"
```
k-takata: Isn't this a mistake of `python2-whoosh`?

Alexpux: @k-takata does packaging is direct dependency or maybe some other package need this dependency?
Show error

k-takata: Here is the error message when I run `sphinx-build` on AppVeyor:
https://ci.appveyor.com/project/universalctags/ctags-win32/branch/master/job/ujeyis6trlu4ve2n#L965
> pkg_resources.DistributionNotFound: The 'packaging' distribution was not found and is required by Sphinx

And `C:\msys64\mingw32\lib\python3.7\site-packages\Sphinx-1.7.6-py3.7.egg-info\requires.txt` has the following lines:

```six>=1.5
Jinja2>=2.3
Pygments>=2.0
docutils>=0.11
snowballstemmer>=1.1
babel!=2.0,>=1.3
alabaster<0.8,>=0.7
imagesize
requests>=2.0.0
setuptools
packaging                   <=== Here
sphinxcontrib-websupport
```

So this is a direct dependency.

Alexpux: @k-takata you are wrong.
File ````C:/msys64/mingw64/lib/python3.7/site-packages\pkg_resources\__init__.py```` are from setuptools package, so seems we ned add packaging as dependency for setuptools

k-takata: `pkg_resources` from setuptools just provides a mechanism to load other packages, and sphinx uses it to load the depended packages including `packaging`.
setuptools itself doesn't require `packaging`. It requires only certifi and wincertstore. Adding `packaging` as dependency of setuptools is not correct.

Please also see the source code of sphinx:
https://github.com/sphinx-doc/sphinx/blob/a27ace341789975ead996d20ef8646a701db9052/setup.py#L18-L31
It explicitly states that it requires `packaging`.

Alexpux: Fixed in repo

k-takata: Thanks, but I think this line still has a mistake: https://github.com/Alexpux/MINGW-packages/blob/1c2f28f5c694f95ddd7a9f4fbf822699bdef2570/mingw-w64-python-sphinx/PKGBUILD#L34
I think it should be `python2` instead of `python3`.

PR #3871, opened 2018-05-30
Mypaint brushes nitpicking overhaul

cbrt64:
Rationale:
1. Remove `_realver`, which is unnecessary, and occurs only here and in [libmypaint/PKGBUILD](https://github.com/Alexpux/MINGW-packages/blob/49894dbb562a65518abb3364482ad85f45f9181e/mingw-w64-libmypaint/PKGBUILD), where I presume it was borrowed from.
2. On Windows, `tar` can fail extracting a symlink if the target doesn't already exist on disk. So don't extract `README` (which isn't needed in the final product) so that `makepkg-mingw` doesn't need to be run twice.
3. This is a data package, providing only brush files, so update the description and remove the patch, reverting to the correct pkgconfig output directory.
4. Setting `PKG_CONFIG_PATH` is unnecessary here, it should rather be used by other packages looking for this one with pkgconfig, currently just GIMP.

I don't know if these commits should be or will be squashed, I will update them if needed.

Questions for @Jehan, the package owner:
1. Should the `license` be "ISC", or something else?
2. Do we need README and README.md, if they're the same?
3. Have I understood the build tools correctly this time (remember [GNOME 795587](https://bugzilla.gnome.org/show_bug.cgi?id=795587)?) ;)


cbrt64:
Well, the mypaint-brushes part passes, anyway.
If this PR is accepted, maybe the `mypaint-brushes` and `gimp` repo packages should be rebuilt at the same time due to the `PKG_CONFIG_PATH` change?

Jehan:
> Should the license be "ISC", or something else?

The license of what? The license of these scripts can be anything, I guess.
The licenses of mypaint-brushes is a mix of various licenses. None of these is ISC.
See: https://github.com/Jehan/mypaint-brushes/blob/master/Licenses.dep5

> Do we need README and README.md, if they're the same?

They are not the same. One is a symlink to the other. I guess my idea when I did this was that github was only reading README.md whereas the common file name is just README. I think since then (I have actually started this repository years ago) github updated its code and may now parse README files too, doesn't it?

> Have I understood the build tools correctly this time (remember GNOME 795587?) ;)

Having a quick look at your commits, yeah it seems so.
Basically this is not an error that we install the `.pc` file under `share/pkgconfig/`. This is the standard. Library `pc` files are under `lib/pkgconfig` but data packages `pc` are in `share/pkgconfig`.
You may have a look in `/usr/share/pkgconfig/` on any common GNU/Linux distribution, you'll see a bunch of files (from very common and major data packages, like `iso-codes.pc` or `systemd.pc` or `udev.pc` and many more!). This is simply the standard and the only proper way to find a data package such as mypaint-brushes is to set PKG_CONFIG_PATH environment variable appropriately (as you do), not to override its installation paths to have the pkg-config files install in the wrong directory.

Hope this helps! :-)

mingw-w64-gimp/PKGBUILD
```diff
@@ -75,6 +75,7 @@ build() {
   cp -rf ${_realname}-${_basever}${_prever} build-${MINGW_CHOST}
   cd build-${MINGW_CHOST}
 
+  PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/${MINGW_PREFIX}/share/pkgconfig" \
```
Alexpux: I think we need modify it here instead:
https://github.com/Alexpux/MSYS2-packages/blob/master/pacman/makepkg_mingw32.conf#L72
https://github.com/Alexpux/MSYS2-packages/blob/master/pacman/makepkg_mingw64.conf#L72

cbrt64:
@Alexpux, much better idea, wish I'd thought of it. So I'll just remove the gimp/PKGBUILD commit here, and make a PR over on MSYS-packages. I caught my extra '/' typo and fixed that there as well.

Alexpux:
@cbrt64 dont create pull on MSYS-packages as I already fixed it. Will push commits later

PR #3875, opened 2018-05-31
gcc-git: Remove the deprecated patch for <xmmintrin.h>.

lhmouse:
This has been fixed in mingw-w64 by 132be552ef8b9e7a637d4c3bcb26c409d1ef1c62 .

I have built GCC 8.1 without this patch and confirmed that the related GCC PR [56038](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56038) is no longer reproducible without this patch.


Alexpux:
@lhmouse did you build 8.1 with Ada?

lhmouse:
No. I only enable C, C++and LTO.

Alexpux:
@lhmouse ok.

PR #3877, opened 2018-05-31
gcc: upgrade to 8.2.0

jeroen:
This synchronizes `mingw-w64-gcc` with `mingw-w64-gcc-git`.


mingw-w64-gcc/PKGBUILD
```diff
@@ -98,37 +90,24 @@ del_file_exists() {
...
-    0100-gcc-4.8-libstdc++export.patch \
-    0110-gcc-4.7-stdthreads.patch
-
-  apply_patch_with_msg 0130-libstdc++-in-out.patch
-
-  apply_patch_with_msg 0140-gcc-7-Enable-std-experimental-filesystem.patch
-
-  # do not expect ${prefix}/mingw symlink - this should be superceded by
+  GIT_AM="git am --committer-date-is-author-date"
+  ${GIT_AM} ${srcdir}/0002-Relocate-libintl.patch
```
Alexpux: First, didyou try to build it yourself?
Second, this is not git package so return appluing patches as it was

jeroen: OK will do, thank you. Can you explain why a different method is needed to apply patches for gcc and gcc-git?

Alexpux: @jeroen you dont answer about testing package.

git packages can apply patches with ````git am``` while release sources need to use ````patch````

mati865: `git am` works only in git repositories. Source code tarballs are contain only the code, git repo is **not** included.

Alexpux:
@jeroen you need return 0130-libstdc++-in-out.patch and fix it for gcc 8.1.0
Then you need test yourself building and packaging GCC. 
If you not testing full building and packaging then pull request will not be accepted

jeroen:
Ok sorry for the extra work. I was testing from the same source tree as `mingw-w64-gcc-git` because a full rebuild takes so much time, but I must have made some wrong assumptions. 

I have now updated `0130-libstdc++-in-out.patch` and started a build on from the upstream tarball on my windows server. I'll give a status update in a few hours, so you can ignore this PR until then.

Thanks for your patience.

jeroen:
It now passes libstdc++ but fails when linking ada with lto. @Alexpux was this the same error that you were seeing? Why are we not seeing this issue for `mingw-w64-gcc-git` ?

screen shot 2018-06-01 at 6 15 34 pm

Alexpux:
Yes I have the same error. gcc-git not fail because it probably building without Ada language

jeroen:
Actually gcc-git [does have both ada and lto](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-gcc-git/PKGBUILD#L137) as well. It did order them differently though, as `--enable-languages=c,lto,c++,objc,obj-c++,fortran,ada` but that should not matter, right?

Alexpux:
@jeroen then probably this issue is fixed somewhere in git

jeroen:
I have been digging through the [commits in the gcc-8-branch](https://github.com/gcc-mirror/gcc/compare/gcc-8_1_0-release...gcc-8-branch) after the 8.1.0 release but I can't find anything that looks like it may fix this problem.

lhmouse:
@jeroen I don't enable Ada support because I have no knowledge about it and will not use it.

lhmouse:
I have started a fresh build of gcc-git with everything enabled (the vanilla one in MINGW-packages). I can have a look into it after the error happens.

lhmouse:
I got this response from <gcc@gcc.gnu.org>:  <https://gcc.gnu.org/ml/gcc/2018-06/msg00017.html>

jeroen:
Perhaps this is caused by our patch [0007-Build-EXTRA_GNATTOOLS-for-Ada.patch](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-gcc/0007-Build-EXTRA_GNATTOOLS-for-Ada.patch)?

Alexpux:
@jeroen try build without it

lhmouse:
Restarted.

lhmouse:
No it still fails without the patch.

jeroen:
Arch linux now seems to be shipping mingw-w64 5.0.4 with gcc 8.1. Maybe have a look what they have done.

mati865:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85638

Mistuke:
Hi All,

I've opened a pull request against this PR https://github.com/jeroen/MINGW-packages/pull/1 to include a fix for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86030 in the GCC8 branch. Tested the backport by compiling without ada.

If it could be included that would help us out a lot. I've also tried looking into the ada failure, but so far don't understand why it happens, the dll should be in the same directory as `xg++`...

/cc @angerman

Alexpux:
I see @lhmouse backport patches for std::filesystem in gcc-git, so maybe we can get them in this pull request also?

bbolli:
Would it not make sense to split the ADA and Fortran compilers out of this package? Maybe even gobjc(++)? I think most people would be happy with C and C++ support.

lhmouse:
The Fortran and Ada runtime libraries (DLLs) are split into separated packages, but they have to be built integrally after all.

Alexpux:
I found GCC commit that broke the build - https://github.com/gcc-mirror/gcc/commit/899af040b0ef0ec694a30851d3eee2a6ef15fddf
Reverting it fix the build.
@lhmouse can you tell about it o GCC guys?
Finally I push chnges to repo-  https://github.com/Alexpux/MINGW-packages/commit/a6cf796dd7de20c8a0c8940f5fa15beb42895447

Mistuke:
Great work @Alexpux . I'm a bit confused as to how this is working for other targets.. perhaps `CXX` here requires `-B../..` just like the `CC` case here https://github.com/gcc-mirror/gcc/blob/cfd0a3db624f0dcc1461d256c78e532ee5e01b83/gcc/ada/gcc-interface/Make-lang.in#L169

Alexpux:
32-bit build still failed but for another reason:
````
xgcc.exe: internal compiler error: Aborted signal terminated program gnat1
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://sourceforge.net/projects/msys2> for instructions.

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
xgcc.exe: internal compiler error: Aborted signal terminated program gnat1
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://sourceforge.net/projects/msys2> for instructions.
make[6]: *** [../gcc-interface/Makefile:296: g-exptty.o] Ошибка 4
make[6]: *** Ожидание завершения заданий…
make[6]: *** [../gcc-interface/Makefile:296: g-locfil.o] Ошибка 4

````

lhmouse:
> I found GCC commit that broke the build - gcc-mirror/gcc@899af04
> Reverting it fix the build.
> @lhmouse can you tell about it o GCC guys?
> Finally I push changes to repo- a6cf796

See <https://gcc.gnu.org/ml/gcc/2018-07/msg00410.html>.

Mistuke:
That patch does have a PR and seems to have broken more stuff, perhaps a comment https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81878 would be prudent too.

PR #3878, opened 2018-05-31
boost: update to 1.67.0

Adsun701:
This updates boost to 1.67.0.

### DIFF
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -28,10 +28,8 @@ source=(https://downloads.sourceforge.net/sourceforge/boost/boost_${_boostver}.t
         boost-1.58.0-pool-test_linking.patch
         boost-1.57.0-python-abi_letters.patch
         boost-1.57.0-python-libpython_dep.patch
-        #boost-1.58.0-address-model.patch
         boost-1.60.0-mingw-context.patch
         boost-1.61.0-fix-undefined-referense-to-codecvt.patch
-        #boost-1.64-default-python.patch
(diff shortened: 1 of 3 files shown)


Alexpux:
@Adsun701 after building check that both python 2 and 3 libraries builded and they linked with proper python libraries

Adsun701:
OK, so only python2 libraries were built. Let me work on linking the python3 libraries.

One thing I've noticed is the naming change from `${MINGW_PREFIX}/lib/libboost_python-mt.a` to `${MINGW_PREFIX}/lib/libboost_python27-mt.a`.

Alexpux:
@Adsun701 I think this is not good fix for python. All previous releases are building both python modules in one time. I disagree with such changes

Adsun701:
The boost python3 libraries are being built using the python2 include headers and overwriting the previous boost python2 libraries. I really don't know how to fix this, so I posted a gist here:

[https://gist.githubusercontent.com/Adsun701/ca75e86609571d267d6125c60d2a191c/raw/e9ee7459a1d928f37aeb4cdd8de7b9830e8c81f8/mingw-w64-boost-1.67.0-1-x86_64-build.log](url)

Alexpux:
@Adsun701 in the past version patch boost-1.64-default-python.patch fix this problem but now you comment it

Adsun701:
Actually, the patch boost-1.64-default-python.patch is not needed anymore, as the code that patch reverts has been reverted for version 1.67.0. The real problem was specifying more than one python version in the command arguments, which I just fixed in a new commit. Now both boost python libraries are built and link against the appropriate python.

PR #3887, opened 2018-06-02
opencl: use system provided library

wirx6:
*Vulkan validation layers can be added later as separate [package](https://github.com/KhronosGroup/Vulkan-ValidationLayers)


Alexpux:
@ why you remove all my patches? I build everything from source

wirx6:
Building everything from source is generating additional dependencies and need more work to maintain, we can use system dll because it's provided anyway.

wirx6:
@Alexpux would you accept only opencl commit(for now there is no library accessible)?

mingwandroid:
> building everything from source is generating additional dependencies and need more work to maintain, we can use system dll because it's provided anyway

Please use chocolately or some other non-from-source distro if that's how you like your software.

wirx6:
@mingwandroid so can you tell me why mingw is using `msvcrt.dll` and other system libraries(opencl is shipped with system, vulkan is installed with any graphic driver)?

Alexpux:
@wirx6 if you want to use system installed libraries then create your own custom repo and use your own package. We prefere to use builded by our own if possible. If you dont know why we use msvcrt and other system libraries - read mingw site.

wirx6:
As you wish, I only wanted to help

PR #3888, opened 2018-06-02
Rebuild libpng

jeroen:
Something weird is going on with libpng, it no longer seems to build in master.


jeroen:
Right so this no longer builds with the current toolchain. It works fine with an older toolchain. There seems to be some bug in the latest gcc 7.3.0 that causes the generated `libpngconf.h` file to contain a bunch of incorrect whitespace and linebreaks.

Alexpux:
@jeroen seems it related to https://github.com/Alexpux/MSYS2-packages/pull/1180

jeroen:
Is this a regression in the mingw-w64 crt then? The package builds fine on msys2 if I set `CC=` to another compiler.

jeroen:
A screencast to illustrate my concern that this is an issue in the compiler and not just gawk. So if I set CC to my old compiler everything works (no changes in PATH or gawk): https://youtu.be/fXf5YcZJA7o

Alexpux:
@jeroen compare config.log between working and not working build

Alexpux:
@jeroen I'm rebuild gawk and problem fixed for me

jeroen:
OK let me trigger a rebuild, to see if the problem is gone.

jeroen:
OK all seems fine now. Thank you for fixing gawk, @Alexpux! I still don't understand why the problem did not appear for my old compilers.

Alexpux:
@jeroen maybe you have other gawk in your compiler folder?

PR #3914, opened 2018-06-08
gdb-git: Update patch due to upstream changes.

lhmouse:
This makes patch 0005 apply again.
This only fixes the patch in question. At this moment gdbserver doesn't compile.


lhmouse:
@Alexpux This can be merged now.

PR #3924, opened 2018-06-09
add aom package and enable av1 support in ffmpeg

wirx6:
*building libaom as dll is not yet implemented


mingw-w64-aom-git/PKGBUILD
```diff
@@ -0,0 +1,44 @@
+_realname=aom
```
Alexpux: Add header with maintainer info. See other PKGBUILDs

wirx6:
done

PR #3941, opened 2018-06-13
gmp header file contains macro which decide it is shared or not.

Readon:
fix the bugs that mpfr will not compile dll.


Alexpux:
What error you have ? We build static libraries as bcc links with them

Readon:
The current version would not build and install shared library.
I want to use mpfr as independent shared library, which can be used in vala as arbitrary precision algorithm library.

Alexpux:
As I tell we use static gmp/mpfr to simplify GCC support.

Readon:
I think there would be some cases that use mpfr not only for GCC where arbitrary precision is needed.
just like some calculator.

Alexpux:
I don’t understand what problem you have with linking with our static library.

Readon:
ok, I just suppose that there would be more people like me just want to link with mpfr in shared version.

Alexpux:
@Readon issue here is that if I add shared libraries then GCC will link with shared library and if gmp/mpc/mpfr will be apdated and DLL name will be changed then automatically we get not working GCC.
So right now I will not accept this change

Alexpux:
You can create your own repo and place there custom packages that you want and share it with your users.

lazka:
There is some new discussion going on re shared gmp in #7024. Feel free to chime in there.

This PR contains many unrelated changes, so closing.

PR #4002, opened 2018-06-24
Add MKL-DNN (version 0.14)

msakai:
This adds MKL-DNN (version 0.14).


Alexpux:
I’m not accepting packages that builds only for one arch

msakai:
Thank you for your reply.

MKL-DNN supports only x86_64 officially. In the issue for packaging MKL-DNN for Debian, a developer of MKL-DNN said as follows:

https://github.com/intel/mkl-dnn/issues/206#issuecomment-377837134
> MKL-DNN has only been tested on intel64 machines, but I know that there are some forks that do work on non-x86 CPUs. The library has not been designed with non-x86 machines in mind, and the way it checks for the CPU features is not portable. So I would only package MKL-DNN for amd64.

And due to its use of JIT compilation, I (as a non-expert of x86 assembly) do not know how much effort is necessary for adding i686 support.

If such architecture-specific package cannot be accepted, I'll withdraw this PR and #4003.

PR #4003, opened 2018-06-24
Add Menoh (version 1.0.1)

msakai:
This adds Menoh DNN inference library (version 1.0.1).
This PR is intended to be merged after #4002.

### DIFF
--- a/mingw-w64-menoh/0000-remove-dllexport.patch
+++ b/mingw-w64-menoh/0000-remove-dllexport.patch
@@ -1,0 +1,13 @@
+diff --git a/include/menoh/menoh.h b/include/menoh/menoh.h
+index f2b413a..78ef87f 100644
+--- a/include/menoh/menoh.h
++++ b/include/menoh/menoh.h
+@@ -5,7 +5,7 @@
+ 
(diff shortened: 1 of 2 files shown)


mingw-w64-menoh/0000-remove-dllexport.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/include/menoh/menoh.h b/include/menoh/menoh.h
+index f2b413a..78ef87f 100644
+--- a/include/menoh/menoh.h
++++ b/include/menoh/menoh.h
+@@ -5,7 +5,7 @@
+ 
+ #ifndef DOXYGEN_SHOULD_SKIP_THIS
+ 
+-#if defined(_WIN32) || defined(__WIN32__) || defined(__CYGWIN__)
```
Alexpux: I think you dont need to patch this.

msakai: Menoh's windows support is still buggy and some non-exported symbols are referenced from outside the DLL. This patch avoids the problem by removing all `__declspec(dllexport)` annotation so that `gcc` fallbacks to export all symbols from DLL.

PR #4005, opened 2018-06-24
Do Not merge

stahta01:
oce: Update version to see if it builds on server


lazka:
(It took too long hit the timeout)

stahta01:
Builds on my local Win 7 64 bit; only built the 32 bit build.

PR #4008, opened 2018-06-24
enchant: update to 2.2.3

Adsun701:
This updates enchant to 2.2.3.


Alexpux:
@Adsun701 why you remove all relocation patches?

Adsun701:
I looked at the code and could not find anything that needed relocation. Version 2.2.3 is a large jump from 1.6.1, so I don't think the patches are needed anymore. I'll check this and see if this is the case.

Adsun701:
Confirmed that built-in relocation support was added with version 2.2.0:

https://github.com/AbiWord/enchant/issues/172

Ede123:
@Adsun701 Does enchant actually work for you with these changes?

At least the Aspell back-end is broken by this update (should report the "misspelled word "asdf"):
```
$ echo asdf > temp.txt && enchant-2 -l temp.txt && rm temp.txt
Couldn't create a dictionary for de_DE.UTF-8
```

I noticed this while attempting to fix gtkspell/gtkspell3 packages which are both broken by this update as well due to the library name change (enchant -> enchant-2).

Ede123:
Seems not limited to the Aspell backend.

With mingw-w64-x86_64-enchant-1.6.1-1:
```
$ enchant-lsmod.exe
aspell (Aspell Provider)
ispell (Ispell Provider)
myspell (Myspell Provider)
```

With mingw-w64-x86_64-enchant-2.2.3-1:
```
$ enchant-lsmod-2.exe

```
(empty output i.e. no providers found)

Adsun701:
@Ede123 I think aspell may have to be rebuilt against enchant version 2.2.3, as the current aspell package was definitely built with enchant version 1.6.1.

Ede123:
Aspell does not depend on enchant, so a rebuild should not be required.

I suspect an issue with the relocation logic introduced / removed in enchant-2.

lazka:
bt, this broke all reverse deps of enchant due to the so/dll version bump.

lazka:
I've openend #4041

Ede123:
Great, thanks @lazka! Now we only need to fix enchant itself...

lazka:
yeah, just a start... :/

PR #4042, opened 2018-07-05
[WIP] latexila: Rename to gnome-latex, update, add new dependencies

lazka:
needs #4041 first


Alexpux:
@lazka is gtksourceview4 not conflict with 3 version?

lazka:
> @lazka is gtksourceview4 not conflict with 3 version?

There are not conflicts. There was a conflict with the glade catalog file which is why I've disabled building that for now.

PR #4063, opened 2018-07-11
mingw-w64-pegtl - 2.6.0 - New Package - Parsing Expression Grammar (P…

JPeterMugaas:
…EG) Template Library


Alexpux:
@JPeterMugaas it really dont have dependencies, even on gcc libraries?

JPeterMugaas:
I have gone ahead and added the gcc-libs to depends.

PR #4064, opened 2018-07-11
meson: fix python module in case the official Python is installed

lazka:
meson uses the py.exe to figure out the path of the default python2/3
but msys2 python doesn't install it. This leads to it finding the
system Python and getting the wrong path.

This fixes the gobject-introspection build with meson (not the default yet..)


lazka:
I'll try to get this upstream

PR #4070, opened 2018-07-12
[WIP!!] test

lazka:


Alexpux:
I don't see in log that autoreconf is runned during prepare.

lazka:
> I don't see in log that autoreconf is runned during prepare.

yeah, good catch

lazka:
In either case appveyor seems to build it fine. Can you maybe try again with this PR?

lazka:
Not sure, maybe "aclocal --install" in autogen.sh does that?

I've filed an issue upstream re missing autogen.sh in the tarball: https://gitlab.gnome.org/GNOME/gnome-latex/issues/62

PR #4087, opened 2018-07-17
Change Default Generator to Ninja

yuui-tanabe:


mati865:
You cannot use `make` command with ninja build system.

Alexpux:
@yuui-tanabe you need add logic for make tool choosing depending on generator

Issue #4114, opened 2018-07-27
cmake broken due to missing dependency on libpsl


Rondom:
Cmake (both 32 and 64bit) fails due to missing dependency on libpsl. Installing `mingw-w64-{x86_64,i686}-libpsl` manually fixes this.

I will submit a pull request as well.

Rondom:
Closing this issue as a duplicate of #4111

mati865:
I think there is no PR for `curl`, you can fix it instead if you want.

Rondom:
Last I have checked it is still @Alexpux who is the only one who can do uploads and he mentioned that he will do it, Thus, I am not sure if me submitting a PR will help in any way.

Maybe this has changed, though. I will go ahead and submit a PR.

PR #4115, opened 2018-07-27
cmake: Add missing dependency to libpsl

Rondom:
Fixes: #4114


Rondom:
Closing, since it is curl depending on psl (and cmake depending on curl)

PR #4132, opened 2018-07-31
Add universal pkgbuild for cross-binutils

wirx6:
*multiple packages are generated from _targets array
*built tools can be used with clang or gcc in cross-compilation
*ccache can be used to speed up building this


lazka:
Any reason this has x86_64 in the directory name?

Alexpux:
@lazka I move it

lazka:
@Alexpux ok, it needs a pkgbase matching the directory as well

Alexpux:
@lazka done

lazka:
Thanks!

PR #4155, opened 2018-08-06
kicad-git: Add depends "${MINGW_PACKAGE_PREFIX}-curl"

stahta01:


stahta01:
Before creating and uploading an binary the wxPython package needs to be rebuilt with the new GCC version for the 64 bit build at least. The 64 bit build of Kicad-git has a run-time wx ABI error because of different GCC used.

PR #4156, opened 2018-08-06
fix gtkada build issue when use new version gprbuild

RickLeaf:


RickLeaf:
the original package no "mingw-w64-x86_64-gprbuild-bootstrap-git" in makedepends

PR #4158, opened 2018-08-06
fix gtkada build issue when use new version gprbuild

RickLeaf:


mingw-w64-gtkada/PKGBUILD
```diff
@@ -12,14 +12,16 @@ provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 url="https://libre.adacore.com/tools/gtkada/"
 options=('!staticlibs' 'strip')
 license=('GPL3' 'custom')
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc-ada")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc-ada"
+             "${MINGW_PACKAGE_PREFIX}-gprbuild-bootstrap-git")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-ada"
-         "${MINGW_PACKAGE_PREFIX}-gtk3")
+         "${MINGW_PACKAGE_PREFIX}-gtk3"
+         "${MINGW_PACKAGE_PREFIX}-gprbuild-bootstrap-git")
```
Alexpux: I think this is wrong because after bootstrapping gpr-build-bootstrap will be removed and gtkada will prevent doing this

RickLeaf: Your means is remove "${MINGW_PACKAGE_PREFIX}-gprbuild-bootstrap-git" ?

PR #4182, opened 2018-08-12
vala: update to 0.40.8

Adsun701:
This updates vala to 0.40.8.

### DIFF
--- a/mingw-w64-vala/PKGBUILD
+++ b/mingw-w64-vala/PKGBUILD
@@ -16,7 +16,6 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "help2man")
 depends=("${MINGW_PACKAGE_PREFIX}-glib2"
          "${MINGW_PACKAGE_PREFIX}-graphviz")
-options=('!emptydirs')
 source=(https://download.gnome.org/sources/${_realname}/${pkgver%.*}/${_realname}-${pkgver}.tar.xz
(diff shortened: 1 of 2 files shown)


lazka:
@Adsun701 have you actually tested if it works? Last time I tried updating it failed because it installs some new library not in PATH.

For example, does "valac --help" work?

Adsun701:
@lazka OK, it does not work. But I noticed that libvalaccodegen.dll is installed to ${MINGW_PREFIX}/lib/bin and not ${MINGW_PREFIX}/bin, so I will check that out right now.

Adsun701:
Confirmed it was libvalaccodegen.dll being installed to the wrong place. Now fixed.

lazka:
thanks, looks good!

mingw-w64-vala/PKGBUILD
```diff
@@ -42,6 +46,9 @@ build() {
 package() {
   cd "${srcdir}/build-${MINGW_CHOST}"
   make -j1 DESTDIR="${pkgdir}" install
-
+  
+  # Move libvalaccodegen.dll to correct directory
+  mv "${pkgdir}${MINGW_PREFIX}"/lib/bin/libvalaccodegen.dll "${pkgdir}${MINGW_PREFIX}"/bin/libvalaccodegen.dll
```
Alexpux: More general will be move *.dll and remove lib/bin folder

PR #4185, opened 2018-08-13
lpsolve: add lpsolve

yumetodo:
lpsolve is Mixed Integer Linear Programming (MILP) solver

cmake script was forked from
- https://github.com/PADrend/ThirdParty/blob/master/lpsolve-CMakeLists.txt
- https://github.com/PADrend/ThirdParty/blob/master/lpsolve-lpsolveConfig.cmake.in

ref:
- https://github.com/PADrend/ThirdParty/pull/2
- https://github.com/PADrend/ThirdParty/pull/3

### DIFF
--- a/mingw-w64-lpsolve/PKGBUILD
+++ b/mingw-w64-lpsolve/PKGBUILD
@@ -10,7 +10,7 @@ arch=('any')
 url="https://sourceforge.net/projects/lpsolve/"
 license=('LGPL2.1')
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
-makedepends=("automake-wrapper" "libedit" "${MINGW_PACKAGE_PREFIX}-gcc")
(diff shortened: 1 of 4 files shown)


mingw-w64-lpsolve/PKGBUILD
```diff
@@ -0,0 +1,70 @@
...
+  # Shared Install
+  cd "${srcdir}/build-shared"
+  make DESTDIR="${pkgdir}" install
+
+  # Static Install
+  cd "${srcdir}/build-static"
+  make DESTDIR="${pkgdir}" install
+}
+
+warn_build_references() {
```
Alexpux: Why this needed?

yumetodo:
```
==> WARNING: Package contains reference to $srcdir
```

This message will be produced. However, lpsolve use `__FILE__` macro for error handling, that is the reason the message produced.

ref:
- https://bbs.archlinux.org/viewtopic.php?id=239092

PR #4187, opened 2018-08-13
libatomic_ops: update to 7.6.6

Adsun701:
This updates libatomic_ops to 7.6.6. Both shared libraries and static libraries are now built.


Alexpux:
Replace tabs witrh spaces

PR #4192, opened 2018-08-14
Output source URL using for loop + Rebuild

yuui-tanabe:


lazka:
why?

yuui-tanabe:
decrease the code size

PR #4226, opened 2018-08-16
capstone: update to 3.0.5

Adsun701:
This updates capstone to 3.0.5.


Adsun701:
Will be closing this because of #4243 also updating capstone to 3.0.5 in addition to building python bindings.

PR #4245, opened 2018-08-18
graphite2: update to 1.3.12

Adsun701:
This updates graphite2 to 1.3.12.


Adsun701:
There is a segmentation fault with this version; I will work on fixing it.

Adsun701:
Fixed now; had to disable the GRAPHITE2_ASAN if check. I am not sure if it is needed.

PR #4267, opened 2018-08-21
zopfli - 1.0.2 package - attempt to fix.

JPeterMugaas:
I had to include two patches to make this work.


Alexpux:
I dont see any reason to split package for parts.

JPeterMugaas:
Hm, this seems to be the way that Archlinux was doing it in their AUR and I can see why.  It turns out that contains its own library for compressing PNG files.  zopfli is just a compression format.  I could combine the two but I have to have a provides for zopflipng.

Alexpux:
We not Arch Linux and have our own rules. For mingw packages we prefer to not splitting - only some cases.

PR #4288, opened 2018-08-22
libtiff - 4.0.9 - add CVE patches listed by Archlinux in their PKGBUILD

JPeterMugaas:


Alexpux:
You modify Libraw PKGBUILD instead tiff.

PR #4291, opened 2018-08-23
openjpeg2 - 2.3.0 - Fixes for a build issue on my computer with Java

JPeterMugaas:
I identified an issue on my computer with the build and Java.  It turns out that the JPIP includes two clients that are written in Java that will be installed if you have Java installed.  Unofrtunately, the Java version 9.x does not work with the JDK 1.5 source-code.  In any case, MINGW-W64 does support Java and any setups (well, just say all bets are OFF).


lazka:
Needs a rebase

lazka:
repo gone, so closing.

PR #4315, opened 2018-08-26
python-pyflake8 - 3.5.0 and packages that it depends upon

JPeterMugaas:
python-pycodestyle - 2.4.0 - new package
python-mccabe 0.6.1 - new package

### DIFF
--- a/mingw-w64-python-flake8/4fcbcccf381ce0987faa297173e4008b0490918f.patch
+++ b/mingw-w64-python-flake8/4fcbcccf381ce0987faa297173e4008b0490918f.patch
@@ -1,0 +1,37 @@
+From 4fcbcccf381ce0987faa297173e4008b0490918f Mon Sep 17 00:00:00 2001
+From: Jack Mustacato <jmustacato@knights.ucf.edu>
+Date: Fri, 27 Oct 2017 15:05:23 -0400
+Subject: [PATCH] Update minumum pycodestyle version to 2.3.0
+
+---
+ setup.cfg | 2 +-
+ setup.py  | 2 +-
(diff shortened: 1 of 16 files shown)


Alexpux:
@JPeterMugaas for what reasons you add so many python packages?

JPeterMugaas:
Alright, there's a big reason.  It turns out that a lot of validation with python uses various packages and I try to validate the validator using their tests.  This is what Archlinux is trying to do and I am trying to do this as well.  I use some of the Archlinux dependencies from their distribution as a model.  Some times, the test frameworks depend upon each other and sometimes, other things.  So in one sense, I have a "what comes first, chicken or the egg" problem so in one sense, I'm trying to bootstrap the process.   And it turns out that some things such as harfuzz use a python packages for their validation such as python-fontuils.  Fontutils (if you look at their GitHub page (https://github.com/fonttools/fonttools) use a number of packages for their capabilities and validation such as pyzopfli and sympy.  You could say this is like building a wall where you need numerous bricks that cement together.

JPeterMugaas:
Okay. Flake8 is a testing framework and some packages we have use it such as setuptools for their testing.

From /setuptools-40.2.0tests/requirements.txt
===
mock
pytest-flake8; python_version!="3.4"
pytest-flake8<=1.0.0; python_version=="3.4"
virtualenv>=13.0.0
pytest-virtualenv>=1.2.7
pytest>=3.0.2
wheel
coverage>=4.5.1
pytest-cov>=2.5.1
paver; python_version>="3.6"
===
Some other python packages we have might be using it.

Alexpux:
@JPeterMugaas python2-configparser is missing

PR #4332, opened 2018-08-28
python-pyzmq: update to 17.1.2

Adsun701:
This updates python-pyzmq to 17.1.2.

### DIFF
--- a/mingw-w64-python-pyzmq/001-mingw-python.patch
+++ b/mingw-w64-python-pyzmq/001-mingw-python.patch
@@ -1,6 +1,7 @@
---- py2zmq-16.0.2/setup.py	2016-07-19 16:31:03.100214900 +0100
-+++ pyzmq-16.0.2/setup.py	2016-07-19 16:38:12.475410600 +0100
-@@ -120,7 +120,7 @@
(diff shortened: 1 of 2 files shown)


mingw-w64-python-pyzmq/001-mingw-python.patch
```diff
@@ -36,14 +37,12 @@
...
-                 fatal("ZMQ directory must be specified on Windows via setup.cfg or 'python setup.py configure --zmq=/path/to/zeromq2'")
-             try:
-@@ -565,7 +565,7 @@
+@@ -558,7 +558,7 @@
              # this may not be sufficiently precise
              libzmq.define_macros.append(('ZMQ_USE_POLL', 1))
-         
+ 
 -        if sys.platform.startswith('win'):
 +        if sys.platform.startswith('win') and 'MSYSTEM' not in os.environ:
```
Alexpux: Probably this is not good condition. You need use _POSIX_BUILD. See for example patches from pyqt5

Adsun701: Ok, fixed now.

PR #4343, opened 2018-08-30
glib2: Update to 2.58.0 and port to meson

lazka:


lazka:
> is there any reason for g_fopen do not understend "{w,r}+" mode?
> and this patch do not care for alias for "wb+", original source does this (and msvc):

I didn't want to introduce any regressions and keep the code as it was with 2.56. Ideally this shoul dbe upstreamed as the other patches..

> and maybe i'm wrong but in g_open:
> when _fmode _O_TEXT and flags _O_BINARY _setmode will be called with _O_TEXT, instead of _O_BINARY

good catch, I've added a comment to the issue for this patch: https://github.com/Alexpux/MINGW-packages/issues/2414#issuecomment-417785476

PR #4378, opened 2018-09-06
PIP - fix to work on command line and use packages instead of statica…

JPeterMugaas:
…lly linking stuff in (just like Archlinux does)

python-pip - 18.0 - Build without third-party libraries.  We will package thos differently
python-distlib - 0.2.7 - new package required by pip
python-pytoml - 0.1.18 - new package required by pip
python-retrying - 1.3.3 - new package required by pip


Alexpux:
@JPeterMugaas you rewrite PKGBUILDs for other packages very often. Check what you doing. You wrongly change distorm PKGBUILD

JPeterMugaas:
This has gone badly.  I'll restart this branch and make another request.

Issue #4396, opened 2018-09-12
Use Azure Pipelines for CI


lazka:
https://azure.microsoft.com/en-us/services/devops/pipelines/

tl;dr; 10 parallel builds, 6 hours max build time per job, Server 2012/2016

I've played around a bit with it here: ~https://github.com/lazka/msys2-azure-pipelines~

This just installs MSYS2 and installs the toolchain in two parallel jobs: ~https://dev.azure.com/lazka/test/_build/results?buildId=17&view=logs~

This is how a PR looks like: ~https://github.com/lazka/msys2-azure-pipelines/pull/1~

Ede123:
Sounds interesting (although bloated), especially as they claim to offer builds for Mac, Linux and Windows at the same time.

Is there a possibility to publicly display the build logs without owning a Microsoft account? Every link I found produces a login page, which is not particularly helpful for open source projects...

lazka:
> Sounds interesting (although bloated), especially as they claim to offer builds for Mac, Linux and Windows at the same time.

appveyor also does windows+linux (but yeah, it's enterprisey..)

> Is there a possibility to publicly display the build logs without owning a Microsoft account? Every link I found produces a login page, which is not particularly helpful for open source projects...

Good point. In the end they just redirect to a public page, not sure if that is intentional...

lazka:
> > Is there a possibility to publicly display the build logs without owning a Microsoft account? Every link I found produces a login page, which is not particularly helpful for open source projects...
> 
> Good point. In the end they just redirect to a public page, not sure if that is intentional...

Does this work for you https://github.com/lazka/MINGW-packages/pull/1 ?

Ede123:
Yes, I can access the logs for the individual build steps without signing in:
![grafik](https://user-images.githubusercontent.com/904570/45483922-a8ae5a00-b752-11e8-8ce3-f78264f22d60.png)

lazka:
Cool, seems the login is only a thing when you add it through the github marketplace.

lazka:
ok, I need rights to add web hooks: "Unable to configure a service on the selected GitHub repository. This is likely caused by not having the necessary permission to manage hooks for the selected repository." :(

lazka:
@Alexpux I've send you an email invite. Please go to https://dev.azure.com/msys2/mingw/_build -> "GitHub" -> "New pipeline" -> "Authorize with OAuth" -> Select this git repo -> ...

peterbud:
[...]

PR #4398, opened 2018-09-13
Add support for Azure Pipelines. See #4396

lazka:
This creates a mingw32 and a ming64 job running on Server 2012r2

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,44 @@
+jobs:
+  - job: MINGW
+    timeoutInMinutes: 360
+    pool:
+      vmImage: vs2015-win2012r2
+    strategy:
+      maxParallel: 5
+      matrix:
+        i686:
+          MINGW_UPPER: MINGW32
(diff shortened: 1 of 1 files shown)


lazka:
Once that file is in master I can enable it on the azure side of things.

mati865:
While at CI topic Tea could be disabled as it doesn't help anymore and only confuses people.

lazka:
One thing I just noticed is that it doesn't support any UI stuff (interactive mode) compared to appveyor. I don't think we need it.. or do we?

lazka:
> One thing I just noticed is that it doesn't support any UI stuff (interactive mode) compared to appveyor. I don't think we need it.. or do we?

According to https://github.com/mesonbuild/meson/issues/4154#issuecomment-421220046 the server2016 one supports it, so if it turns out to be a problem we can switch there.

lazka:
I'm taking the liberty to just merge this so we can see how it works. With tea-ci we are used to CI not working anyway so things wont get worse anyway.

PR #4412, opened 2018-09-16
Update to OpenSSL 1.1.1 -

JPeterMugaas:
mingw-w64-openssl 1.1.1 - upgrade to latest version - This will break stuff compiled with stuff compiled with older versions of OpenSSL
mingw-w64-nghttp2 - 1l33l0 - rebuild for new OpenSSL version
mingw-w64-curl - 7.61.1 - rebuild with new openssl version


Alexpux:
This will broke repository for a long time. There are a lot of packages dependent on openssl

lazka:
>  The thing is that support for the 1.02 series will be dropped at the end of the year

https://www.openssl.org/policies/releasestrat.html

Version 1.0.2 will be supported until 2019-12-31 (LTS).

JPeterMugaas:
Alright, so far it seems to build out of the tree but I now get an error saying something about .a.new already existing.  Sorry for not responding earlier, but there's still some other things I'm doing.  It turns out that it is using an "int" for socket functions which doesn't always work in Windows since that can be a 64-bit value.  I will submit a patch to the OpenSSL developers about that one.  I can drop the compat package proposal but there still are a lot of things that have to be rebuilt for the new version.  An alternative might be to make a copy of the .DLL's under the old name.

MSP-Greg:
@Alexpux

Previously you've mentioned the issues with 1.1.0, and the same holds true for 1.1.1.

I work with Ruby, and we've been using 1.1.0 for quite a while and I've been using 1.1.1 for several weeks.

Working PKGBUILD and patch files are [here](https://github.com/MSP-Greg/ruby-makepkg-mingw).  I've only build 64 bit, and I also disable docs.  Packages are [here](https://ci.appveyor.com/project/MSP-Greg/ruby-makepkg-mingw/build/7/artifacts), with SHA512's in the messages section.  Not signed.

Checks/tests pass on both 1.1.1 and 1.1.0i.

Thank you for all your work here.

JPeterMugaas:
Alright, just to compare notes, I'm trying to fix some compiler warnings with the build.  A key problem I have is that the socket function calls in OpenSSL 1.1.1 are using the "int" type.  In Windows, that will ALWAYS be 32-bits wide and signed.  Microsoft seems have to defined the SOCKET type as 64-bits unsigned in Win64 and this would be used for socket functions such as "bind", "connect", "socket", "accept", etc.  As I say, I am intending to report this as a bug to the OpenSSL developers.  At the moment, I've got issues with the "z" in format strings even with "D__USE_MINGW_ANSI_STDIO=1" and I'm not completely sure what I am supposed to do beyond that.

MSP-Greg:
> I'm trying to fix some compiler warnings with the build

OpenSSL does CI on Appveyor & Travis, and I believe they cross-compile on Travis.  I'd suggest looking at the logs there and checking the warnings.  At some point when 1.1.1 was in beta, I posted an issue with warnings, and a commit was done that removed some of them.

I'm not a strong c type, and warnings are not uncommon.  Hence, I'll leave it up to the OpenSSL people to determine what is and is not a real issue.

FYI, the packages pass all tests, and also pass all tests when used in three different Ruby apps that compile with OpenSSL.

JPeterMugaas:
I am looking at your work and I have made a refinement.  I still have tests to do.  I am aware that some warnings are harmless.  But I suspect that the ones involving socket functions might not be harmless.  See this post:https://stackoverflow.com/questions/1953639/is-it-safe-to-cast-socket-to-int-under-win64 .  While we have been lucky that the current implementation might work with 32-bit values, we can't rely upon that to always be the case.  The thing is that OpenSSL has defined things as an "int" type which is ALWAYS 32-bits (4 bytes wide) which is not like a typical Unix-like system where "int" may 8-bytes wide.  SOCKET was defined as "UINT_PTR" in one version of the winsock2.h header and as u_int for 32-bit versions of Windows.

Now just a note about my compile problem.  It turns out that ranlib does not play nicely with McAfee Virus scan.  That was what what was causing the error "can not rename .a.new".

And another thing is that I think it's possible to do an out of source-build with 1.1.1 because they use a new build system since 1.1.0x.  We might also be able to enable rc5 since the patent on RC5 expired in 2015.

mati865:
Error from CI:
```
checking for file conflicts...
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/BN_print.3ssl.gz exists in filesystem
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/HMAC.3ssl.gz exists in filesystem
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/LHASH.3ssl.gz exists in filesystem
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/MD5.3ssl.gz exists in filesystem
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/MDC2.3ssl.gz exists in filesystem
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/RC4.3ssl.gz exists in filesystem
mingw-w64-x86_64-openssl: /mingw64/share/man/man3/UI.3ssl.gz exists in filesystem
```

JPeterMugaas:
Okay.  I have a made PR request that should fix your problem https://github.com/Alexpux/MINGW-packages/pull/4429 .

And I also made a bug report and included the patch file in the package at https://github.com/openssl/openssl/issues/7282 in reference to our earlier discussion about SOCKET type.

Alexpux:
@peterbud install with --force option
This issue because pacman look for case-sensitive filenames

peterbud:
Thanks, that's working for me! However as I see Appveyor scripts is failing due to this. Shall we add the --force option to Appveyor script?

PR #4428, opened 2018-09-20
update qpdf to version 8.2.1

oltolm:


jeroen:
Could we patch this to also build a static lib (rather than just the import lib)?

PR #4430, opened 2018-09-21
Update PKGBUILD

yuui-tanabe:


stahta01:
@yuui-tanabe why did you make this useless change?

Tim S.

PR #4435, opened 2018-09-21
Add python2-markdown package

eduardosm:
The PKGBUILD is based on archlinux's.


Alexpux:
Why only python2? I see it support python 3.4 and later too

eduardosm:
It seems python-markdown was already in the repository. When I first created this package (long before I opened the pull request), it wasn't.

PR #4441, opened 2018-09-22
openssl: build with deprecated interfaces

lazka:
Passing api=1.0.0 removes all API deprecated with 1.0.0. That's not the default
config and breaks code trying to support multiple openssl versions, so remove it.


lazka:
@JPeterMugaas any objections?

JPeterMugaas:
No.  Not really.  I wasn't sure what that line was about anyway.

PR #4460, opened 2018-09-25
opusfile: remove http support (doesn't build with openssl 1.1)

lazka:
https://github.com/xiph/opusfile/issues/13


Alexpux:
well we can wait while it will be fixed upstream

lazka:
ok. I also noticed that this code is there to hook openssl up to the windows cert store. Which means we can just disable it and let it use the msys2 certs like everything else. I'll open another PR for that.

lazka:
I've opened #4474 for a fix without disabling http support.

PR #4463, opened 2018-09-25
libplacebo: update to 0.5.0

Adsun701:
This updates libplacebo to 0.5.0.


Alexpux:
@Adsun701 it is not buildable for 32 bit

PR #4469, opened 2018-09-27
Add colour-science.org python library (git)

briend:
The excellent colour python library.


Alexpux:
i dont want to add git package

briend:
Ok, I will submit a released version. Thank you!

PR #4472, opened 2018-09-27
mingw-w64-xmlsec: Update after new openssl

peterbud:


lazka:
This might not be needed with #4475

PR #4475, opened 2018-09-27
openssl: revert recent patches

lazka:
They break ABI of public interfaces everywhere and the winsock2.h include
conflicts with openssl macros (leading to build errors with gstreamer, python3 etc)


MSP-Greg:
@lazka

Re PKGBUILD

1. The following items:
```
enable-camellia
enable-idea
enable-mdc2
enable-rfc3779
```
are included https://github.com/Alexpux/MINGW-packages/blob/4cef906da8b62a33009e810d62e079e04b06c419/mingw-w64-openssl/PKGBUILD#L78-L91

Looking at one of the Appveyor builds in OpenSSL, they are enabled by default in OpenSSL.  Your call, but if they're on by default, I would defer to the default OpenSSL config.  Note that rc5 is not enabled, but as mentioned elsewhere, I think the patent has indeed expired on it.


2. After the multi-line Configure statement shown above, adding the following will print out config info.  It is used in the OpenSSL CI:
```
perl configdata.pm --dump
```

MSP-Greg:
@lazka

Thanks for your work on this, I got the 64 bit package from Azure, ran a local Ruby build of trunk/master/head, and all is good with 1.1.1-4.  Greg

lazka:
> Looking at one of the Appveyor builds in OpenSSL, they are enabled by default in OpenSSL. Your call, but if they're on by default, I would defer to the default OpenSSL config. Note that rc5 is not enabled, but as mentioned elsewhere, I think the patent has indeed expired on it.

Thanks, good catch.

PR #4478, opened 2018-09-28
gtk4: Add new preview package (v3.94.0)

lazka:
No need to build this, but this makes it easy to test new releases.


lazka:
@Alexpux do you want to include it in git? Otherwise I'll close this.

PR #4490, opened 2018-09-30
Add libuhd package

eduardosm:


lazka:
Sorry for the long delay.

We no longer support Python 2, so the package needs to be adjusted to use Python 3 instead. Since the original fork for this PR seems to be gone I'm closing this. Feel free to create a new PR for libuhd and we'll try to review it in a more timely manner :)

PR #4504, opened 2018-10-03
Update Tier 2 KF5 modules to 5.50

zrax:
Includes 2 new frameworks: knotifications and syndication


zrax:
Oops, this is still missing a piece (phonon-qt5), and that piece still has some unresolved issues.  I'll reopen this when I fix that.

mingwandroid:
Will you end up adding and KDE applications? That'd be cool to see.

zrax:
Yeah, that's the eventual plan, but it'll depend on time and how many issues I run into.

PR #4524, opened 2018-10-10
add python websocket-client

thesmo:


thesmo:
I tried this only as 64 build myself so didn't see this error.
Any idea how I can fix this one?

Alexpux:
Youy need resolve name clashing in packages. Both packages have files with the same name and the same location

thesmo:
After checking some of the other python modules and looking at the example template I have fixed the issue.

PR #4525, opened 2018-10-10
add python comtypes

thesmo:


Alexpux:
the same name clashing error.
see our other python packages how to solve this

PR #4527, opened 2018-10-10
Add kate and kwrite

zrax:
Depends on #4526

### DIFF
--- a/mingw-w64-kate/PKGBUILD
+++ b/mingw-w64-kate/PKGBUILD
@@ -1,0 +1,89 @@
+# Maintainer (MSYS2): Michael Hansen <zrax0111@gmail.com>
+
+_variant=-${KF5_VARIANT:-shared}
+source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
+_realname=kate
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-kate"
+         "${MINGW_PACKAGE_PREFIX}-kwrite")
+pkgver=18.08.1
+pkgrel=1
+arch=('any')
+license=('GPL' 'LGPL' 'FDL')
+makedepends=("${MINGW_PACKAGE_PREFIX}-extra-cmake-modules"
+             "${MINGW_PACKAGE_PREFIX}-plasma-framework-qt5${_namesuff}"
+             "${MINGW_PACKAGE_PREFIX}-knewstuff-qt5${_namesuff}"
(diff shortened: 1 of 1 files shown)


Alexpux:
@zrax currently kate/kwrite fail to install because, first them install data folder in bin - it must not do it, second - only one package seems need to have this folder. I think you dont need to split package to kate/kwrite, instead add ````provides```` entry in PKGBUILD and package all in one

mati865:
Data in bin directory is configured in `extra-cmake-modules`, see https://github.com/mati865/MINGW-packages/blob/kdef5/mingw-w64-extra-cmake-modules/use-UNIX-paths-for-MSYS.patch for incomplete fix.

Alexpux:
@mati865 we use MINGW cmake for mingw packages, so ````NOT MSYS```` is not for MINGW target

mati865:
@Alexpux `MSYS` was defined somehow so it worked. There were more places requiring changes like that.

zrax:
The use of the 'data' folder in bin is due to Qt's unfortunate `QStandardPaths::GenericDataLocation` paths on Windows...  After digging into it for a while, I determined that leaving things that don't need to be accessed by other packages in bin/data was easier than fixing every use of `QStandardPaths` in KF5 (there are many) :(.  However, in my tests, these unfortunate install locations didn't cause any install conflicts, so I'm curious what you ran into...

Alexpux:
Both kwrite and Kate packages container the same files in bin data, so it conflicts. But in any cases I will not accept packages with bin/data folders

zrax:
Hmm, then will you accept a patch that changes the behavior of a fundamental Qt5 API?  That's the only other reasonable solution I can think of at the moment...

zrax:
Wow, that looks a whole lot more complicated than the solution I had in mind...   I'll look into it more later, but I was thinking to just add `QDir::cleanpath(QCoreApplication::applicationDirPath() + "../share")` to the end of the QStandardPaths::GenericDataLocation search paths.  It means it would (potentially) change the behavior of Qt apps, but it also means you don't have to patch every usage to make it work.

mati865:
Yes it's complicated but it (hopefully) doesn't break anything.
I'd avoid changing official APIs, many people use MSYS2 provided Qt build to distribute with their programs. Having API that doesn't mach official Qt API would be probably big turn-off for them.

zrax:
Hmm, good point.

gbschenkel:
I got error when trying to run it.
C:/dev/msys64/mingw64/bin/kate.exe: error while loading shared libraries: Qt5Xml.dll: cannot open shared object file: No such file or directory

https://github.com/Alexpux/MINGW-packages/issues/4534

PR #4575, opened 2018-10-21
python3: fix ctypes module build. Fixes #4572

### LINKED ISSUE #4572: python3: ctypes missing
```
$ python3 -c "import ctypes"
  File "<string>", line 1, in <module>
  File "C:/msys64/mingw32/lib/python3.7\ctypes\__init__.py", line 7, in <module>
    from _ctypes import Union, Structure, Array
ModuleNotFoundError: No module named '_ctypes'
```

lazka:
https://github.com/python/cpython/pull/9258 changed callproc to
use functions from the internal libffi which we don't use.
Just copy the functions into callproc directly instead.


Alexpux:
@lazka guard added functions with ````__MINGW32___````

PR #4589, opened 2018-10-24
gitg: move libraries to bin

albfan:
Fixes https://github.com/Alexpux/MINGW-packages/issues/4586

### DIFF
--- a/mingw-w64-gitg/PKGBUILD
+++ b/mingw-w64-gitg/PKGBUILD
@@ -28,7 +28,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-meson"
              "intltool"
              "itstool"
              "gettext-devel")
-options=('strip' 'staticlibs')
+options=('strip')
 license=("GPL2+")
 url="https://wiki.gnome.org/Apps/Gitg"
 install=${_realname}-${CARCH}.install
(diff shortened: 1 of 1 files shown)


mingw-w64-gitg/PKGBUILD
```diff
@@ -55,6 +55,7 @@ build() {
 package() {
   cd "$srcdir/build-${MINGW_CHOST}"
   DESTDIR=${pkgdir}${MINGW_PREFIX} ninja install
+  mv "${pkgdir}${MINGW_PREFIX}"/lib/*.dll* "${pkgdir}${MINGW_PREFIX}"/bin
```
mati865: Wouldn't this catch `.dll.a` files which have to be inside `/lib`?

albfan: Let me change it. I'm checking libgit2-glib and it has .dll in bin and .dll.a on lib. Why gitg don't do the same without this command

albfan:
gitg.exe doesn't show gitg icon. Am I doing something wrong?

mati865:
I have no experience with Meson so I can't tell for sure but `gitg` build system is not configured correctly for Windows.

lazka:
You need a Windows resource file which points to an .ico file and use "windows.compile_resources" in meson to include it with the .exe (glib uses windows.compile_resources for example to store some metadata in the .exe)

albfan:
@lazka Seems this is the problem. Let's try recovering that config, hopefully I will close this issue

albfan:
@lazka That works https://gitlab.gnome.org/GNOME/gitg/merge_requests/26. Any idea about why libgitg-1.0-0.dll and libgit-ext-1.0-0.dll do not deploy on bin directly? I think it has something to do with shared_library.

About compile_resources, any linux example? I think gitg lacks of it on linux, is something like gnome.compile_resources()? I suppose gitg.rc will not work on linux

albfan:
Seems removing install_dir from shared libraries, meson install correctly, so I will fix gitg master and launch a new release to fix this and other minor bugs, yay!

PR #4594, opened 2018-10-25
gitg: Update to version 3.30.1

albfan:


albfan:
I don't understand this error: https://ci.appveyor.com/project/Alexpux/mingw-packages/builds/19800785/job/w1udp4g3v1a5b6e1#L555
libgit2 is compiled with threadsafe: https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-libgit2/PKGBUILD#L37
The check uses liggit2-glib as wrapper to libgit2 to check that

Anyway Azure CI should be enough

PR #4598, opened 2018-10-28
gdk-pixbuf2: Update to 2.38.0; port to meson

lazka:


Ede123:
@lazka I just noticed there's a single `io-*.dll` in "mingw64\lib\gdk-pixbuf-2.0\2.10.0\loaders" now:  
"io-wmf.dll"

Is this correct?

PR #4608, opened 2018-10-31
Add cvode package

giraldeau:
CVODE is a solver for systems of ordinary differential equations.

More info: https://computation.llnl.gov/projects/sundials/sundials-software

This package is straight forward. The only issue was the DLLs are installed in the lib directory. I move them in the package step from lib to bin. 

Signed-off-by: Francis Giraldeau <francis.giraldeau@nrc-cnrc.gc.ca>


giraldeau:
Awesome, thanks!

PR #4631, opened 2018-11-12
Additional KF5 packages

NebuHiiEjamu:
Unfortunately, Okteta and KDesignerPlugin can't be built because KDocTools errors with this message, but I was able to build the new KDEWebKit and KCMUtils dependencies

```
kf5.kdoctools: Error: Could not find kdoctools catalogs
```


Alexpux:
@NebuHiiEjamu replace all TABs with spaces and add header to PKGBUILDs

PR #4646, opened 2018-11-14
python-setuptools - 40.6.2 -

JPeterMugaas:
Update to new version
fix easy-install script
fix some dependencies that are listed in requirements.txt including certifi and winse


lazka:
Note that setuptools will either use openssl or certifi or wincert certs. So this doesn't add much.

lazka:
extras_require are optional dependencies for extra features

PR #4650, opened 2018-11-14
cmocka - 1.1.3 - re-add MinGW DLL Naming Workaround

stephengroat:
patch prevents programs from finding the DLL at runtime. expecting `cmocka.dll`, not `libcmocka.dll`


Alexpux:
@stephengroat better to chnge program tosearcu libcmocka as mingw builded libraries prefer to use lib prefix in dll names

stephengroat:
@alexpux i did some digging the in the `dll` and `dll.a` and it seems like the `libcmocka.dll` name is not being passed into the `dll` or `dll.a` file, resulting in the issue.

is there any way you would consider reverting the MinGW DLL Naming Workaround? i'm not that familiar with the internals of cmake on mingw/msys to get an upstream patch

stephengroat:
@alexplux sorry my other comment didn't make sense, let me try again.

i went looking at how to fix my `cmocka` issue and why the compliation wasn't working. the `libcmocka.dll.a` file that is part of the msys package doesn't reference `libcmocka.dll` (as it logically should), it references `cmocka.dll` as the field `cmocka_dll_iname`. since the gnu components use these `dll.a` files i'm pretty sure that's what causing the program to fail at runtime (makes sense, since the `dll` is really needed at compile and the `dll.a` is stubbing out the function needed for compilation and linking).

with this stuff in mind, i'm hoping you'll reconsider this pr. i think there's some other stuff broken by the workaround that should probably be addressed upstream

Alexpux:
@stephengroat fixed by https://github.com/Alexpux/MINGW-packages/commit/b3197bc8703dbd2ee4208e65e25da50553b550f7

stephengroat:
@Alexpux thanks so much! works perfectly! thanks for all of the great work, really appreciated!

PR #4653, opened 2018-11-15
python-flexmock - 0.10.2 - new package required to verify python-ioca…

JPeterMugaas:
…pture


Alexpux:
@JPeterMugaas I will accept current pull requests for python packages but I dont want to increase number of python packages in such manner that you do anymore. 
I think you can create separate repository for python packages and maintaint it

Issue #4593, opened 2018-10-25
[qt5] Qt 5.11.2 can't be built on clang 7.0.0


mcallegari:
Up to 5.11.1 all good.
Yesterday I tried to build Qt 5.11.2 with my own flags set and build fails on qttools/qdoc with the error below.
Reverted to llvm+clang 6.0.0: problem solved.
Right now the current tree packages are not consistent.

```
make[3]: entering "/e/MINGW-packages/mingw-w64-qt5/src/i686/qttools/src/qdoc"
[...]

PR #4666, opened 2018-11-16
qt5: fix 32 bit linking with clang 7.0 (#4593)

paolo-sz:
g++ reports libclangEdit as a malformed archive.
Maybe a wrongly coded error message or a bug within the linker but libclangEdit.a seems to have no issue at all by itself.
Linking against the specific file (\<full path\>/libclangEdit.a) solve the issue.
Easily, moving -lclangEdit down within the clang libraries list (this PR) solve the issue as well.


paolo-sz:
@Alexpux done what you asked but I didn't check by my own... waiting continuous integration to do that for me...

PR #4674, opened 2018-11-19
mingw-w64-libwebp - 2.0.2 - Update to latest version - drop expirieme…

JPeterMugaas:
…ntal configure flag as that is not honored.


Alexpux:
Already done

PR #4676, opened 2018-11-19
gettext: Build with --enable-relocatable. Fixes #4392

### LINKED ISSUE #4392: xgettext not using included .its rules
I get the following for an appdata.xml file:

```
xgettext: warning: file 'data/io.github.quodlibet.QuodLibet.appdata.xml.in' extension 'xml' is unknown; will try C
```

rebuilding gettext fixes it so my guess is that it's not relocatable and doesn't find the .its rule files.

lazka:
This makes xgettext find its data files (like .its rules)

Fixes #4392


lazka:
That was easier than expected :)

PR #4684, opened 2018-11-22
Ghostscript

JPeterMugaas:


mingw-w64-ghostscript/PKGBUILD
```diff
@@ -57,14 +67,15 @@ prepare() {
...
-  cd ${srcdir}/${_realname}-${pkgver}
+  cd ${srcdir}/${_tarname}-${pkgver}
   ./configure \
     --prefix=${MINGW_PREFIX} \
     --host=${MINGW_CHOST} \
     --target=${MINGW_CHOST} \
     --build=${MINGW_CHOST} \
     --disable-contrib \
     --enable-fontconfig \
+    --with-fontpath=${MINGW_PREFIX}/share/fonts/gsfonts \
```
Alexpux: I think it will not work as it hardcode Unix path but windows program not understand them/ You also need relocate this path in code. And dont split package - do all in one.

JPeterMugaas: I'll look further at the issue you mentioned.  But the moment, I'm also looking at combining the split packages.  That might take a bit of work as I'm not completely sure if Archlinux did things correctly.  The thing is that you have three programs (Ghostscript, GhostXPS, and GhostPCL) generated by one or more libraries that are in a GhostPDL library (see: https://www.ghostscript.com/download.html ).  I could argue that if we combine all three programs into one package, then it might be called ghostpdl and I would have a "replaces=${MINGW_PACKAGE_PREFIX}-ghostscript" for backwards compatability.  This is just my thoughts on this that.  But I also want to review the patches because we will be doing something with the entire superset and I'm not quite sure what the original makefiles were doing but I'm looking at that now.  I don't know when I'll have something done.

lillolollo:
@JPeterMugaas Any news on this?

JPeterMugaas:
No news as I'm busy with some other parts.

lillolollo:
@JPeterMugaas Thanks

PR #4688, opened 2018-11-23
z3: update to 4.8.3

Adsun701:
This updates z3 to 4.8.3.


mingw-w64-z3/001-mingw-fixes.patch
```diff
@@ -51,21 +53,21 @@
...
  
  def display_help(exit_code):
-@@ -1984,6 +1991,9 @@ class MLComponent(Component):
+@@ -2188,6 +2195,9 @@
                  # Some ocamlmklib's don't like -g; observed on cygwin, but may be others as well.
                  OCAMLMKLIB += ' -g'
  
-+            if is_msys2() and not(is_msys2_mingw()):
-+                LIBZ3 = 'libz3.dll'
++                if is_msys2() and not(is_msys2_mingw()):
```
Alexpux: Why you change level of code?

Adsun701: That was a mistake. I fixed it now.

PR #4692, opened 2018-11-24
Update PKGBUILD

yuui-tanabe:


lazka:
@yuui-tanabe what's your goal here?

PR #4696, opened 2018-11-24
qt5: fix 32 bit linking with clang 7.0 (direct reference to <full_path>/clangEdit.a) (#4593)

paolo-sz:
Checked with a clean MSYS2 install.
This seems to be the only way to workaround the wrongly reported malformed archive issue (#4593)


mingw-w64-qt5-static/0052-qt-5.11-mingw-fix-link-qdoc-with-clang.patch
```diff
@@ -166,17 +166,15 @@
...
                          -lclangDaemon \
                          -lclangDriver \
                          -lclangDynamicASTMatchers \
 -                        -lclangEdit \
++                        $${CLANG_LIBDIR}/libclangEdit.a \
                          -lclangFormat \
                          -lclangFrontend \
-+                        -lclangEdit \
                          -lclangFrontendTool \
-                         -lclangHandleCXX \
```
Alexpux: why remove this libs?

paolo-sz: Because we are looking at a diff of a patch file (which is a diff file too).
Everything removed from the patch will be not touched in the original file where the patch will be applied to.

PR #4698, opened 2018-11-25
Add mingw-w64-python-biopython package

smalltalkman:
Biopython is a set of freely available tools for biological computation written in Python by an international team of developers.
It is a distributed collaborative effort to develop Python libraries and applications which address the needs of current and future work in bioinformatics.


mingw-w64-python-biopython/PKGBUILD
```diff
@@ -0,0 +1,61 @@
+
```
Alexpux: Add header

smalltalkman: Header added

PR #4702, opened 2018-11-27
qemu: fixed segmentation fault, disabled debug flags, added accelerating with HAXM

GentleFly:


mingw-w64-qemu/PKGBUILD
```diff
@@ -28,63 +28,30 @@ depends=("${MINGW_PACKAGE_PREFIX}-capstone"
...
-    --enable-snappy \
-    --disable-docs \
-    --enable-gtk \
-    --with-gtkabi=3.0 \
-    --disable-xen \
-    --disable-vnc-sasl \
-    --disable-kvm \
-    --disable-bluez \
-    --disable-spice \
-    --disable-libnfs \
```
Alexpux: You need to return this disable flags.

PR #4750, opened 2018-12-06
Add mingw-w64-python-xlrd

smalltalkman:
Library for developers to extract data from Microsoft Excel (tm) spreadsheet files (mingw-w64)

### DIFF
--- a/mingw-w64-python-xlrd/PKGBUILD
+++ b/mingw-w64-python-xlrd/PKGBUILD
@@ -1,0 +1,73 @@
+# Maintainer: gym603 <gui_yuan_miao@163.com>
+
+_realname=xlrd
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python2-${_realname}" "${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+pkgver=1.1.0
+pkgrel=1
+pkgdesc="Library for developers to extract data from Microsoft Excel (tm) spreadsheet files (mingw-w64)"
+arch=('any')
+license=('custom')
+url="https://github.com/python-excel/xlrd"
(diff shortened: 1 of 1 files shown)


smalltalkman:
@JPeterMugaas Thanks, the lost depends("${MINGW_PACKAGE_PREFIX}-python2" and "${MINGW_PACKAGE_PREFIX}-python3") have been added.

JPeterMugaas:
I think another thing you missed is that the script in the bin directory might not run correctly if it's invoked with the WRONG python interpreter.  In the bash shell, unlike Windows, you can't assume it is the one in the path which is why I typically hard-code the particular interpreter path in the shebang.  But I do that is by replacing python with something like "${MINGW_PREFIX}/bin/python3" for a python3 package.  Then, in the *.install, files I usually check in, I do a replacement with a "${MINGW_PREFIX}" with a value calculated by the script.  Look at the files in the "MINGW-packages/mingw-w64-PKGBUILD-templates/python-exe-installs/" .  I admit that those are oriented towards the .EXE launcher that the setuptools creates in Windows but that can be adapted for the script itself.  Actually, the .EXE just simply would invoke the interpreter with the name of the script.  eg.  "mypythonprog.exe" would run "python3 mypythonprog-script.py" so in one sense of the word, this is just the same EXACT logic with a different variation.  Let's just say that there are some reasons why I wrote the formula templates in PKGBUILD.setup-py the way I did.  The ideas are:

1) Make sure that any python-architecture compiled binaries are separated from eachother and placed in the appropriate package.  We don't want these being mixed during the build and install processes.
2) reduce the opportunity for mistakes to occur.
3) Handle some optional things such as the .exe or modify for a simple script in the bin dir.  On many that are just libraries, you would remove the "install=" line and the stuff referring to .EXE.
4) Facilitate testing by reducing the number of compiles down to one.

PR #4755, opened 2018-12-07
tcl/tk: update to 8.6.9

oscarfv:
Recent changes on MSYS2 makes pkg to fail for mingw-w64-tcl. The changes on this PR were tested on MSYS2 20180531.


Alexpux:
@oscarfv it is already in my local tree

PR #4765, opened 2018-12-10
gtk3: Backport some fixes

lazka:
* Fix the default scrolling sensitivity
* Make gtk use the default font on Windows 8/10
* Adds defaults for gtk fontconfig so we can drop the settings.ini

All commited upstream.


Alexpux:
@lazka we will wait for release

PR #4776, opened 2018-12-15
python-ifaddr - add dependency on nose and remove some comments from…

JPeterMugaas:
… templates along with unused code.


Alexpux:
no need to rebuild, so dont increase pkgrel

PR #4779, opened 2018-12-17
Libhandy git: added package

tallero:


Alexpux:
@tallero I will not accept git packages

tallero:
@Alexpux  I created a `-git` package because the fix to compile libhandy on mingw had just come upstream.
Anyway that is not necessary anymore since 0.0.6 just came out.
I will now remove the `-git`

PR #4778, opened 2018-12-17
PyWikiBot 3.0.20181203: added package

tallero:
Added PyWikiBot package.

PR #4780, opened 2018-12-17
Daty git: added package (depends on #4778 and #4779)

tallero:
Daty depends on libhandy (#4779) and pywikibot (#4778).


Alexpux:
@tallero i will not accept git packages

tallero:
I will try again when I have got a stable version.

PR #4829, opened 2019-01-04
kicad: Add packages to makedepends

stahta01:


ustsv:
"The packages in the depends array are implicitly required to build the package, they should not be duplicated here. " (С) Arch Wiki. Maybe not right.

stahta01:
> "The packages in the depends array are implicitly required to build the package, they should not be duplicated here. " (С) Arch Wiki. Maybe not right.

They are needed to build the package under MSys2. Tim S.

PR #4831, opened 2019-01-05
gstreamermm: New package

lb90:
GStreamermm is the official C++ binding for GStreamer.

https://gstreamer.freedesktop.org/bindings/cplusplus.html
https://gitlab.gnome.org/GNOME/gstreamermm

Note: while GStreamer GitLab infrastructure is on [FreeDesktop GitLab](https://gitlab.freedesktop.org/gstreamer), GStreamermm is hosted on [GNOME GitLab](https://gitlab.gnome.org/GNOME/gstreamermm)

### DIFF
--- a/mingw-w64-gstreamermm/PKGBUILD
+++ b/mingw-w64-gstreamermm/PKGBUILD
@@ -1,0 +1,42 @@
+# Maintainer: Luca Bacci <luca.bacci982@gmail.com>
+
+_realname=gstreamermm
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.10.0
+pkgrel=1
+pkgdesc="C++ bindings for gstreamer 1.0 (mingw-w64)"
+arch=('any')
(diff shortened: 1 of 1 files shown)


lb90:
I have adapted the PKGBUILd from the gtkmm3 PKGBUILD. Some notes:

- **pangomm** depends on both glibmm and cairomm, so we can remove glibmm and cairomm from depends()
- The package also contains a tool, **generate_plugin_gmmproc_file.exe**, useful for developing elements in C++.
In the repo there are **three copying files**: one for the library code (COPYING), one for the tool code (COPYING.tools) and one for the examples (COPYING.examples). We don't build the examples, so the PKGBUILD copies COPYING and COPYING.tools. Library code is LGPL v2.1 while tool code is GPL.

lb90:
Ready for review

mingw-w64-gstreamermm/PKGBUILD
```diff
@@ -0,0 +1,40 @@
+_realname=gstreamermm
```
Alexpux: Add header

PR #4850, opened 2019-01-09
mingw-w64-glib-networking - 2.58.0 - enable librpoxy and gnome-proxy …

JPeterMugaas:
…support to add glib-pacrunner.exe.

### DIFF
--- a/mingw-w64-glib-networking/PKGBUILD
+++ b/mingw-w64-glib-networking/PKGBUILD
@@ -29,22 +29,24 @@ build() {
   mkdir -p "${srcdir}/build-${MINGW_CHOST}"
   cd build-${MINGW_CHOST}
 
-  meson \
+  ${MINGW_PREFIX}/bin/meson \
     --buildtype plain \
     -Dlibproxy_support=true \
     -Dgnome_proxy_support=true \
     ../${_realname}-${pkgver}
 
-  ninja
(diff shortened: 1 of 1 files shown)


Alexpux:
@JPeterMugaas why we need this? Systemd is not applicable for Windows

JPeterMugaas:
I'm not sure but the meson.build system had installed it in the package and so I felt it was prudent to change the system file to use the correct path in the package.  I did look at: https://build.opensuse.org/package/view_file/windows:mingw:win32/mingw32-glib-networking/mingw32-glib-networking.spec and that system file is not included in their RPM for the Linux cross-compile to Windows.  However, it does include the glib-pacrunner.exe and org.gtk.GLib.PACRunner.service .  I also noticed another thing in the rpm spec.  I think the meson build system is doing something that might be different than the autoconf system if look at the parameters passed to autoconf by the RPM spec file.  We use meson for this package and that might be doing something different.

BTW, I decided to clarify in the PKGBUILD file that we are using the mingw-w64 versions of meson and minja.  I usually build from the MSYS prompts NOT the mingw-w64-i686 or the mingw-w64-x64_86 prompt.  We have three versions of ninja and meson and I want to make a template because quite a few things use meson to build so I want to be sure that the package logic is correct given what I've said.

lazka:
See #4855

PR #4863, opened 2019-01-14
mingw-w64-cmake - Update to latest version

JPeterMugaas:
I also disabled "0004-Implement-Qt5-static-plugin-support.patch" because that failed to merge.  I was able to use mingw-w64-PKGBUILD-common/build-kf5 with this but there was another failure with kcrash having to do with failing to find "QT:WinExtras".


Alexpux:
@JPeterMugaas need to port commented patch in other case we will stay on current package version

PR #4864, opened 2019-01-14
grantlee - 5.1.0 - new package

JPeterMugaas:
A string template engine based on the Django template system and written in Qt5


Alexpux:
@JPeterMugaas first it require qt5, second I dont want increase packages number without good reason

JPeterMugaas:
Alright.  It turns out that KDevelop and Konnart (part of the KDE suite require it).  I've been dabbling with updating the current KDE stuff but I don't have anything ready yet.

PR #4875, opened 2019-01-17
gtk3: backport switch fix

lazka:
3.24.3 broke the switch widget symbols: https://gitlab.gnome.org/GNOME/gtk/merge_requests/508

we can also wait for the next bugfix release...


lazka:
3.24.4 is about to be released, so closing.

PR #4909, opened 2019-01-22
Try to create GHC 8.6.3 build script.

ofry:
Also fixed https://github.com/Alexpux/MINGW-packages/issues/4800

but build still fails later (on stage 2)

Signed-off-by: ofry <tim4job@bmail.ru>


lazka:
What's the status of this?

PR #4940, opened 2019-02-01
add python-nvidia-ml

totaam:
The pull requests supersedes this buggy one: https://github.com/Alexpux/MINGW-packages/pull/4595 which you can reject and close.
The main differences are:

- use 2to3 to make the code python3 compatible
- I've actually tested the code on a system with an nvidia card - worked fine with both python2 and python3
- removed unused build code functions
- renamed the package (no need for "XXX-py" suffix when the package is already called "python-XXX")

Note: there is another github project which has ported the code to make it python3 compatible: https://github.com/fbcotter/py3nvml , but they've also added other functions. I felt it safer to go with the "official" nvidia source even if it meant using 2to3 to build with python3.

### DIFF
--- a/mingw-w64-python-nvidia-ml/PKGBUILD
+++ b/mingw-w64-python-nvidia-ml/PKGBUILD
@@ -29,7 +29,7 @@ prepare() {
   cd "python3-build-${CARCH}"
   for pyscript in pynvml.py nvidia_smi.py; do
     chmod 666 ${pyscript}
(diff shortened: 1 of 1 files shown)


totaam:
The CI build failed with: _'The agent: Hosted Agent 2 lost communication with the server. Verify the machine is running and has a healthy network connection.'_
I am reasonably confident that this is a transient error and that all the packages are fine.

PR #4945, opened 2019-02-05
gtk3: Update to 3.24.5

lazka:


lazka:
this version has a new default theme btw:

capture

PR #4947, opened 2019-02-05
gstreamer: rebuild bad and ugly

lazka:
and disable fdkaac, 1.14 doesn't support fdk-aac 2.x yet


lazka:
To should fix x264/x265/opencv related elements

PR #4948, opened 2019-02-05
Libyuv

totaam:
there are two patches:

- remove linking with libjpeg which fails with some cryptic error and isn't actually used in the library itself (only the extra utilities)
- fix cmake install target locations for .DLL and .EXE


totaam:
@Alexpux is there anything preventing this package from being merged?

Alexpux:
@totaam you need use official sources. Make package as git sources

totaam:
@Alexpux hah, thanks for the explanation.
I believe that https://github.com/msys2/MINGW-packages/pull/4948/commits/75317b954d366377f48b0d007bfd6b7be63ccf79 addresses this.

PR #4954, opened 2019-02-07
bundle pyopengl-accelerate

totaam:
This adds the acceleration bindings to python-pyopengl.

I'll be the maintainer because smo doesn't seem to be very interested.
The accelerate module is a separate package on pypi, but it **must** be packaged together because of the very tight coupling between the two. Failure to do so results in misery (we've tried in the past!).

updates in this patch:
- we have to rebuild the .c files from the .pyx source to make it build with python3.7, see https://github.com/mcfletch/pyopengl/issues/11
- to do that, we have to re-add some missing source files, for details see https://github.com/mcfletch/pyopengl/issues/12, which means we have to add cython as a build dependency build


Alexpux:
@totaam increase pkgrel

totaam:
@Alexpux done

PR #4957, opened 2019-02-08
Pandoc

JPeterMugaas:


Alexpux:
Binary packages is not acceptible in any case

JPeterMugaas:
Understood.  It looks like a better approach might be to try bootstrapping Haskell itself and trying to make a package.  It's taking a bit of work and the thing is that some of it is actually written in Haskell so I have to use a binary as part of the build process even if it's not included in the package. Trying to bootstrap this stuff can be fun.

PR #4969, opened 2019-02-12
gcc-git: Update filesystem patch.

lhmouse:
The upstream code has been rewritten for Windows so this is going to be a huge change.

The patch, as its commit message says, is still generated with the following commands:

```
    git cherry-pick 861db1097d3f 2fd48392d0a4 70fea18e0bbb d4fd5c4964cf
    git checkout master -- libstdc++-v3/src/filesystem/  \
      libstdc++-v3/include/bits/fs_\* libstdc++-v3/include/std/filesystem
```

However, the `git cherry-pick` now results in conflicts. Because all filesystem files will be overwritten by the `git checkout` anyway, the working tree is accepted (despite conflict markers) blindly with `git add -A && git cherry-pick --con`.


lhmouse:
ping?

PR #5000, opened 2019-02-25
libssh: update to 0.8.7

Adsun701:
This updates libssh to 0.8.7.


Adsun701:
Closing this because of #5004, which updates libssh and cmake.

JPeterMugaas:
I should add that one problem I had encountered was that the build couldn't work with cmake 1..13.3 because the FindDoxygen module had a bug that was fixed for 1.13.4.

PR #5007, opened 2019-02-26
mingw-w64-mesa: Clean-up dependencies and add a new GPG key

pal1000:
Tests I made indicate that '${MINGW_PACKAGE_PREFIX}-clang' on its own encapsulates '${MINGW_PACKAGE_PREFIX}-gcc' and '${MINGW_PACKAGE_PREFIX}-zlib'.

Flex and bison are required. patch is used inside PKGBUILD so it has to be added too.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -9,10 +9,7 @@ pkgdesc="Open-source implementation of the OpenGL specification (mingw-w64)"
 arch=('any')
 makedepends=("${MINGW_PACKAGE_PREFIX}-clang"
              "scons"
-             "python2-mako"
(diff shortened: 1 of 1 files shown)


mingwandroid:
Correct me if I misunderstand! Just because a package is also provided transitively by another it does not mean that the package should not declare its direct dependencies, err, directly.

Helenerineium:
From https://github.com/msys2/MSYS2-packages/blob/master/README.md

> Packages from the msys2-devel and base-devel groups are implicit build time dependencies. Make sure both are installed before attempting to build any package

I don't think packages in base-devel (bison, flex, patch) should be listed as dependencies.
(They are **implicit** dependencies)

PR #5011, opened 2019-02-28
readline 8.0

jeroen:


Alexpux:
@jeroen I think there are can be a lot of rebuilds during readline update so need to find what packages to rebuild it they builds with new readline

PR #5037, opened 2019-03-08
add mingw-w64-python-xpra

totaam:
For more information, see [xpra.org](https://xpra.org) and in particular this ticket: [1883: packaging for mingw64 / MSYS2](https://xpra.org/trac/ticket/1883)

### DIFF
--- a/mingw-w64-python-xpra/PKGBUILD
+++ b/mingw-w64-python-xpra/PKGBUILD
@@ -36,14 +36,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-cython"
              "${MINGW_PACKAGE_PREFIX}-gtk3"
              "${MINGW_PACKAGE_PREFIX}-pygobject-devel"
-             "${MINGW_PACKAGE_PREFIX}-python3-cairo"
-             #codec build deps:
-             "${MINGW_PACKAGE_PREFIX}-ffmpeg"
-             "${MINGW_PACKAGE_PREFIX}-libyuv"
-             "${MINGW_PACKAGE_PREFIX}-libvpx"
-             "${MINGW_PACKAGE_PREFIX}-x264-git"
-             "${MINGW_PACKAGE_PREFIX}-libwebp"
-             "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo")
+             "${MINGW_PACKAGE_PREFIX}-python3-cairo")
 source=("https://xpra.org/src/xpra-${pkgver}.tar.xz")
 sha512sums=('1a06a79fdcee8f4310ce8e6d31c9bfc23b384348b0c103a04683114d8f42ce03432556a62c0f9699219e8b3c8adfc4a7e4e741a3a0d249202c3ff9a4664ae902')
 
--- a/mingw-w64-python-xpra/PKGBUILD
+++ b/mingw-w64-python-xpra/PKGBUILD
@@ -1,0 +1,167 @@
+# Maintainer: Antoine Martin <totaam@xpra.org>
+
+_realname=xpra
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-common"
+         "${MINGW_PACKAGE_PREFIX}-python2-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+pkgver=2.5rc2
+pkgrel=1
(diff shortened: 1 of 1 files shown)


mingw-w64-python-xpra/PKGBUILD
```diff
@@ -0,0 +1,167 @@
...
+         "${MINGW_PACKAGE_PREFIX}-python2-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+pkgver=2.5rc2
+pkgrel=1
+pkgdesc="Remote access client / server software"
+arch=('any')
+url='https://xpra.org'
+license=('GPL2')
+optdepends=('libnotify: notification support'
+            'gst-plugins-bad: extra audio codecs'
```
Alexpux: All mingw packages have "${MINGW_PACKAGE_PREFIX}-" prefixes

totaam: @Alexpux the idea was to have the common dependencies here, those that aren't python version specific.
Are you saying that I need to duplicate them into each package instead?
If so, can I not declare a common_optdepends variable and reference that? (feels cleaner than cut & paste)

Alexpux: You must remove optdepends from header completely because:
1. Packages described in it not present
2. optdepends entries in subpackages override it.

totaam: Thanks for the feedback.

1. I must have copied the optdepends from an archlinux example, fixed
2. the deps are common to both python2 and python3, so I've used `optdepends+=`

Alexpux:
@totaam  see errors on CI:
````
error running (['uglifyjs', '--version'],),{'stderr': -1, 'stdout': -1}: [Error 2] The system cannot find the file specified
Warning: uglifyjs failed and return -1
Warning: yuicompressor module not found, cannot minify
Warning: no x264 and no vpx support!
 you should enable at least one of these two video encodings
'nvcc --version' failed with return code 1
stderr: None
updated ./xpra/build_info.py with:
````

**Comment by totaam** [comment]
@Alexpux Thanks, I am looking into this, I can't seem to figure it out.

First the harmless warnings:
- `Warning: uglifyjs failed and return -1` and `Warning: yuicompressor module not found, cannot minify
` - that's OK, the builtin html5 client won't be minified if none of the minifiers are found at build time
- `'nvcc --version' failed with return code 1` - also not a problem, without CUDA you won't get the NVENC and NVFBC codecs, but xpra works fine without those
- `Warning: no x264 and no vpx support!` - I don't understand this one, vpx and x264 are installed and the pkgconfig file is there too... same problem as:
- `Exception: ERROR: cannot find a valid pkg-config entry for pygobject-2.0 using PKG_CONFIG_PATH=C:\msys64\mingw32\lib\pkgconfig;C:\msys64\mingw32\share\pkgconfig` - I don't understand why this is not found - it is detected on my build system VM that I use every day, just not on a fresh install or the msys2 CI system.

This should always return 0:
```
python -c "import subprocess;proc=subprocess.Popen(['pkg-config', '--exists', 'pygobject-2.0'], stdout=subprocess.PIPE, stderr=subprocess.PIPE);print(proc.communicate(), proc.wait())"
```
And if you run the pkg-config command by hand, it does:
```
# pkg-config --exists pygobject-2.0
# echo $?
0
```
Why is python subprocess.Popen failing on those systems?

totaam:
Never mind, I found the solution here https://github.com/msys2/MINGW-packages/pull/5043, updated pull request will follow.

Alexpux:
@totaam no need add twice dependencies to makedepends and depends. Does x265 need to be in deps too?

totaam:
@Alexpux x265 is optional and not recommended at present.
`makedepends` is defined only once in the file.
`depends` is modified in the python2 vs python3 subpackages to add the python version-specific dependencies, the common bits are at the top. What do you suggest I should change here?

Alexpux:
@totaam remove this from makedepends as it already in depends:
````
             #codec build deps:
             "${MINGW_PACKAGE_PREFIX}-ffmpeg"
             "${MINGW_PACKAGE_PREFIX}-libyuv"
             "${MINGW_PACKAGE_PREFIX}-libvpx"
             "${MINGW_PACKAGE_PREFIX}-x264-git"
             "${MINGW_PACKAGE_PREFIX}-libwebp"
             "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo")
````

totaam:
@Alexpux Ah, gotcha. Thanks.
I didn't know that you could skip `makedepends` if already in `depends`.

PR #5056, opened 2019-03-13
Update Mesa to 19.0.0

pal1000:
- cleanup dependencies (mingw-w64-clang and mingw-w64-z3 are not needed, only gcc, llvm and zlib mingw packages are needed);
- update build fix patch to avoid spurious crosscompilation and MSVC build break when applied openning the posibility for the patch to be upstreamed, see upstream merge request: https://gitlab.freedesktop.org/mesa/mesa/merge_requests/362;
- cleanup compiler flags (-std=c++11 is passed correctly now and undefined M_PI doesn't seam to happen anymore.


Alexpux:
@pal1000 update checksums

PR #5065, opened 2019-03-15
WIP: python3: fix syntaxerror in sysconfig for some builds

### LINKED ISSUE #5048: python3 debug build syntax errors
Hi i build the python3 mingw package with the debug option in the PKGBUILD file

this results in a python3 installation that has a syntax error in

`python3.7\lib-dynload\_sysconfigdata_dm_win32_.py`

around the string

`INCLDIRSTOMAKE`

which breaks for example meson

is this a python bug or a problem with how msys builds that?

lazka:
The build scripts does basic sed replacement to make things relocatable
and fails when strings are wrapped (depending on the path lengths).

Disable wrapping when writing out the file to prevent that.

Ideally we would not use sed and write out the correct values in the first place,
but that's for another time.

Fixes #5048


lazka:
(The only things this does is add "width=999999", the rest are diff changes..)

lazka:
better fix -> #5069

PR #5071, opened 2019-03-16
Update gnucobol-svn to actually build from current development

GitMensch:
... leading to nearly 3.1RC, with new dependency libxml2 and complete checks upon request


GitMensch:
Can anyone explain to me why the Azure pipeline starts the build twice - and removes the dependencies before ?!?
@Alexpux: I assume the PR is considered fine as it is. Correct?

Out of interest: Is it possible to add artifacts with PKGBUILD?

Alexpux:
@GitMensch we have release package for gnucobol so I will not accept svn package

GitMensch:
I see the point.

We currently actually have a (heavily outdated and unstable) gnu-cobol-svn package which this PR "just" renames and updates.

I suggest to either accept this PR or remove the gnu-cobol-svn package, along with the packages mingw-w64-nsis-svn (which we have mingw-w64-nsis as release) and mingw-w64-clang-svn (which we also have mingw-w64-clang as release).

Alexpux:
I will remove those packages from repo

GitMensch:
4866ea8f3961d04b7f3356120a633db4aaacba23 provided consistency which is good.

The work done for this PR is not lost, just kept locally to check if a new version should build and run before doing the actual release version...

Issue #4818, opened 2019-01-01
Upgrade to GCC8 for mingw32


orgads:
Hi,

I've noticed in addfba3f1e68a9d30398771069b92fb6e66e70d6 that gcc7 was added because:

> gcc 8 can't be built for 32 bit

I just tried building it for mingw32 and it worked. Is this still relevant? Can it be upgraded for mingw32 too?

JPeterMugaas:
I think it could be if it works.

stahta01:
It failed to build for me; failed in the Ada/Gnat section like it was said to fail.

Tim S.

orgads:
Here is [my log](https://github.com/Alexpux/MINGW-packages/files/2749666/makepkg.log.gz). Where does it fail for you?

stahta01:
`checking for i686-w64-mingw32-ranlib... (cached) /mingw32/i686-w64-mingw32/bin/ranlib
checking command to parse /c/PKGBUILD/mingw-w64-gcc/src/build-i686-w64-mingw32/./gcc/nm output from /c/PKGBUILD/mingw-w64-gcc/src/build-i686-w64-mingw32/./gcc/xgcc -B/c/PKGBUILD/mingw-w64-gcc/src/build-i686-w64-mingw32/./gcc/ -L/mingw32/i686-w64-mingw32/lib -L/mingw32/lib -isystem /mingw32/i686-w64-mingw32/include -isystem /mingw32/include -B/mingw32/i686-w64-mingw32/bin/ -B/mingw32/i686-w64-mingw32/lib/ -isystem /mingw32/i686-w64-mingw32/include -isystem /mingw32/i686-w64-mingw32/sys-include    object... /c/PKGBUILD/mingw-w64-gcc/src/build-i686-w64-mingw32/./gcc/xgcc -B/c/PKGBUILD/mingw-w64-gcc/src/build-i686-w64-mingw32/./gcc/ -L/mingw32/i686-w64-mingw32/lib -L/mingw32/lib -isystem /mingw32/i686-w64-mingw32/include -isystem /mingw32/include -B/mingw32/i686-w64-mingw32/bin/ -B/mingw32/i686-w64-mingw32/lib/ -isystem /mingw32/i686-w64-mingw32/include -isystem /mingw32/i686-w64-mingw32/sys-include    -c -g -O2    -W -Wall -gnatpg -nostdinc   g-exptty.adb -o g-exptty.o

[...]

PR #5094, opened 2019-03-19
oce: support unicode filenames

carrotIndustries:
For some reason, opencascade developers thought that MINGW doesn't
support the windows-specific unicode extensions to fopen and friends
and thus use the non-unicode ones, making dealing with non-ASCII
filenames impossible.

This patch enables unicode fopen on MINGW.

### DIFF
--- a/mingw-w64-oce/PKGBUILD
+++ b/mingw-w64-oce/PKGBUILD
@@ -21,7 +21,9 @@ sha256sums=('c553d6a7bf52f790abc3b6bb7a1e91a65947e92a426bb1a88a11960c31f0966c'
 
 prepare() {
   cd "${srcdir}/oce-OCE-${pkgver}"
-  patch -p3 -i "${srcdir}"/unicode.patch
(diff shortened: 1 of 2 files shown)


carrotIndustries:
The 32 bit build failed since libstdc++ only gained support for wfopen fairly recently: https://github.com/gcc-mirror/gcc/commit/d4fd5c4964cf1862c19c0caa876fa43a677bc04a#diff-44b5c2c646cafc1c88f92e4c13f77fbb

Gcc 7.4.0 from i686 doesn't support wfopen. Why do we have an older gcc on i686 than on x86_64 in the first place?

Alexpux:
@carrotIndustries see https://github.com/msys2/MINGW-packages/issues/4818

carrotIndustries:
Oh well, that's unfortunate. Would only applying the patch on x86_64 be an acceptable workaround for the time being?

carrotIndustries:
Appveyor CI failed due to timeout.

carrotIndustries:
As far as I can tell we now have gcc 9 for i686 as well, so we could technically merge this PR.

PR #5125, opened 2019-03-30
gtk3: Update to 3.24.7

lazka:


lazka:
Thanks! With a bit of luck we'll switch to meson soon, so the build times will improve a lot for future updates.

PR #5127, opened 2019-03-31
Remove hack from the kicad package

nickoe:
With the latest release of glm the cmake hack is not needed in the kicad package anymore.


nickoe:
@ustsv why does this fail in the prepare step?

I think I will split this PR in two.

nickoe:
I split this up with the glm bump in https://github.com/msys2/MINGW-packages/pull/5128

Alexpux:
@nickoe fix conflicts

nickoe:
@Alexpux ok, thank you. I rebased it, but I think the glm pkgbump needs to be pushed to the package repo before this will pass the CI.

PR #5134, opened 2019-04-02
libgeotiff: update to 1.5.1

Adsun701:
This updates libgeotiff to 1.5.1.


dg0yt:
Now there is `libgeotiff-5.dll` instead of `libgeotiff-2.dll`. This needs a rebuild of `gdal` at least. I wish such changes would be detected and handled in MSYS CI.

PR #5135, opened 2019-04-02
Add mingw-w64-lib3mf package

podsvirov:


Alexpux:
@podsvirov provided by another commit

PR #5158, opened 2019-04-08
Bump to 19.04 and bump rel

ZachBacon:
Bumps FAudio to 19.04 and bumps the rel.


mingw-w64-faudio/PKGBUILD
```diff
@@ -3,8 +3,8 @@
 _realname=FAudio
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver="19.01"
-pkgrel=1
+pkgver="19.04"
+pkgrel=2
```
Adsun701: pkgrel should be reset to 1 when pkgver is updated.

PR #5182, opened 2019-04-13
WIP: glib2: Update to 2.60.0

lazka:
gtk-doc disabled for now since "ninja gio-doc" fails for some reason


lazka:
meh, 32bit builds fails. I'll have a look.

PR #5214, opened 2019-04-16
libpng-1.6.37 - add tests   and include  png2pnm pnm2png programs fro…

JPeterMugaas:
…m the /contrib/pngminus dir..

This is for cconsistency with Archlinux and in case those 2 programs are used by other stuff.

### DIFF
--- a/mingw-w64-libpng/PKGBUILD
+++ b/mingw-w64-libpng/PKGBUILD
@@ -42,8 +42,12 @@ build() {
   msg2 "Build contributed programs"
   mkdir -p ".libs/contrib/pngminus" && cd ".libs/contrib/pngminus"
   cp -r "${srcdir}/${_realname}-${pkgver}/contrib/pngminus"/* .
-  sed -e "s|#EXEEXT = .exe|EXEEXT = .exe|g" -i Makefile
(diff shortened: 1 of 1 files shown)


mingw-w64-libpng/PKGBUILD
```diff
@@ -39,11 +39,32 @@ build() {
...
     as_ln_s="cp -pR"
   make
+  msg2 "Build contributed programs"
+  mkdir -p ".libs/contrib/pngminus" && cd ".libs/contrib/pngminus"
+  cp -r "${srcdir}/${_realname}-${pkgver}/contrib/pngminus"/* .
+  sed -e "s|#EXEEXT = .exe|EXEEXT = .exe|g" \
+    -e "s|libpng.a|libpng16.a|g" \
+    -e "s|-lpng|-lpng16|g" \
+    -i Makefile
+  make PNGINC="-I${srcdir}/${_realname}-${pkgver} -I${srcdir}/build-${MINGW_CHOST}" CC="${MINGW_PREFIX}/bin/gcc" CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS" \
```
Alexpux: Probably you need add build dir where builded libpng.dll.a located to LDFLAGS, in other cases executables will be linked with system installed libpng

JPeterMugaas: I had considered that but I did see that when building without libpng instealled.  I think part of this is some confusion about the "libpng" name when linking.  I made a checkin a minute ago that hopefully classifies what I'm doing.  I also decided to add more parameters to the make call instead of modifying the Makefile with sed.  This version also includes statically linked versions of the contrib/programs in case developers might want to redistribute those.

PR #5227, opened 2019-04-17
lz4 - 1.9.0 Update to latest version, drop unneeded patch, and fix .l…

JPeterMugaas:
…ib file (it should be .dll.a)


Alexpux:
@JPeterMugaas  seems your commit is wrong because CI generate empty package:
````
make: Nothing to be done for 'install'.
````

Alexpux:
@JPeterMugaas return patch and adopt it to new version

Alexpux:
@JPeterMugaas why you not first check yourself on your machine? Pakcage is broken. As I told you return patch and fix it for new version

JPeterMugaas:
It seemed to work on my machine.

Alexpux:
@JPeterMugaas both CI failing

JPeterMugaas:
I know.  I'm working on this now.  I'm wondering if it might be better if we used cmake with "contrib/cmake_unofficial/CMakeLists.txt" instead of using the Makefile build system.  I think that the Makefiles could be doing something strange.

Alexpux:
done by https://github.com/msys2/MINGW-packages/commit/430ca85108f43ebcc9b05a9c509cbafa66d3a1d9

PR #5230, opened 2019-04-17
cmake: update to 3.14.2

Adsun701:
This updates cmake to 3.14.2.


Alexpux:
CI fail because need rebuild Qt5 with new ICU

PR #5261, opened 2019-04-21
Fix internal compiler error in dwarfout.c during compilation with GCC 8

hunbernd:
Fix GCC 8 stopping with internal compiler error when -fno-omit-frame-pointer is used.

Backported patch from here:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86593
https://gcc.gnu.org/ml/gcc-patches/2018-08/msg00559.html


lazka:
We are now at gcc 9.1.0, is this still needed?

hunbernd:
It works now without the patch

PR #5263, opened 2019-04-22
lz4 - 1.9.0 - redid the patch  so Make should work on MSYS2, CYGWIN, …

JPeterMugaas:
…and MINGW.  The idea is to kill several birds with one stone.  I have submitted this in bug report

https://github.com/lz4/lz4/issues/688 to the authors.


JPeterMugaas:
I made a better mousetrap.

Issue #3538, opened 2018-04-03
libcurl static linking error


galaxycash:
libcurl give errors with static linking, mingw32

zedxxx:
I found that problem with linking nghttp2. It seems that static version libcurl.a linked with dynamic version nghttp2 (i.e. it compiled without flag `-DNGHTTP2_STATIC`).

My test CMakeLists.txt:

```
cmake_minimum_required(VERSION 2.8)

project(curl_static_test)

add_executable(${PROJECT_NAME} "main.cpp")

add_definitions(-DCURL_STATICLIB)

set(LIBS ${LIBS} curl)
set(LIBS ${LIBS} nghttp2)

set(LIBS ${LIBS} brotlidec-static)
set(LIBS ${LIBS} brotlicommon-static)

set(LIBS ${LIBS} idn2)
set(LIBS ${LIBS} intl)
set(LIBS ${LIBS} unistring)
set(LIBS ${LIBS} iconv)

set(LIBS ${LIBS} ssl)
set(LIBS ${LIBS} crypto)

set(LIBS ${LIBS} ssh2)
set(LIBS ${LIBS} crypt32 z ws2_32 wldap32)

target_link_libraries(${PROJECT_NAME} ${LIBS})

set(CMAKE_EXE_LINKER_FLAGS "-static -static-libgcc -static-libstdc++")
```
Got errors like this:
`G:\dev\msys32\mingw32\lib\libcurl.a(libcurl_la-http2.o):-1: error: undefined reference to _imp__nghttp2_session_get_stream_user_data`

Properly compiled static version libcurl must require `_nghttp2_session_get_stream_user_data`.

Alexpux:
This errors tell that program want to link shared version of libcurl

williamjcm:
The question is: why does it want *only* a dynamic build of `libcurl` ?

Using `curl-config --static-libs` doesn't help in any way.

Alexpux:
@williamjcm do you add ````-static```` to link command?

williamjcm:
Yes.

Alexpux:
show full linking command

Alexpux:
And keep in mind that our project prefer shared linking instead static and this issues is not in our priority in any case

zedxxx:
My output:

```
[...]

PR #5268, opened 2019-04-23
fix typo

marcelotduarte:
references: libcurl static linking error #3538 
libcurl use this to compile, see: https://github.com/curl/curl/blob/master/winbuild/MakefileBuild.vc#L174


Alexpux:
@marcelotduarte  update checksums and increase pkgrel

Alexpux:
Fixed in https://github.com/msys2/MINGW-packages/commit/0d6166344802d86f0af3437887ba28b5a557be81

PR #5286, opened 2019-04-30
cmake tweak FindSWIG command such that users can set a search path fo…

gjanssens:
…r swig

As mingw is not typically installed in a path that's hard-coded into cmake, the original find
command would not find swig. By allowing the user to set search paths (that is, by removing
the NO_CMAKE_FIND_ROOT_PATH flag), a user can tell cmake where to look for swig.


gjanssens:
Hmm the checks fail, but I highly doubt that's due to my PR. It complains about a Qt dependency, but I haven't touched anything related to Qt...
Hopefully someone with more knowledge of the cmake dependencies can explain this build failure.

lazka:
Sorry for the late response.

We have packages that use cmake and swig (https://packages.msys2.org/package/mingw-w64-x86_64-ogre3d?repo=mingw64 for example), so I'm wondering if this is still required.

gjanssens:
Thanks for the reference to ogre3d. It triggered me to search around a bit in older PRs against msys2 and I discovered we were using msys2's swig rather than mingw32's swig. Switching to the latter solved our problem so you are correct the patch is not needed (and actually incorrect).
I'm glad we got this solved, thanks again!

lazka:
Cool, thanks for following up on this!

PR #5305, opened 2019-05-04
unicorn: drop unnecessary msys python2 dependency

lazka:
It builds fine with the mingw one.


Alexpux:
Unicorn install python bindings this is why python in dependencies

lazka:
> Unicorn install python bindings this is why python in dependencies

I haven't removed the mingw python dependency.

PR #5307, opened 2019-05-04
brotli: drop python2-brotli

lazka:
Nothing depends on it and there is python3-brotli for Python 3.

@JPeterMugaas any objections?

### DIFF
--- a/mingw-w64-brotli/PKGBUILD
+++ b/mingw-w64-brotli/PKGBUILD
@@ -3,35 +3,32 @@
 _realname=brotli
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
-         "${MINGW_PACKAGE_PREFIX}-python2-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-python3-${_realname}"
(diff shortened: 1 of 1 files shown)


lazka:
@JPeterMugaas

JPeterMugaas:
Sorry for not responding earlier.  I see what you say with Python2 being dropped and many python coders dropping support in their packages.  However, I think we would do well to keep Python 2 support in these packages as best we can.  Dropping Python 2 would be very difficult particularly if there's stuff that will not be ported.

totaam:
We (xpra) need python2-brotli, until we drop python2 support.
Which will happen, but we're not ready for that yet.

lazka:
@JPeterMugaas ok

@totaam oh, I did checked your list first :) let's not do this then.

/me stops removing python 2 packages for now

JPeterMugaas:
It may have been a good idea.  TBH, I actually had to deliberately introduce a ruby-gem that was discontinued by the authors nearly a decade ago because I wanted a package (ruby-nroff) that depends upon it.  Unfortunately, the newest branch depended upon an old version of something that conflicted with a newer version.  Sometimes, hard calls have to be made given limitations.

PR #5291, opened 2019-05-01
Python updates

peterbud:
* python-certifi: Update to 2019.3.9
* python-debtcollector: Update to 1.21.0
* python-decorator: Update to 4.4.0
* python-defusedxml: Update to 0.6.0
* python-importlib-metadata: Update to 0.9
* python-jedi: Update to 0.13.3
* python-mako: Update to 1.0.9
* python-more-itertools: Update to 7.0.0
* python-terminado: Update to 0.8.2
* python-urllib3: Update to 1.25.2 (pls do NOT upload to the repo yet, as python-requests is still explicitly requires urllib3<1.25)
* python-numpy: Update to 1.16.3 
* python-pytz: Update to 2019.1
* python-pandas: Update to 0.24.2
* python-pyzmq: Update to 18.0.1
* python-scandir: Update to 1.10.0
* python-prompt-toolkit: Update to 2.0.9
* python-cryptography: Update to 2.6.1
* python-yaml: Update to 5.1 
* python-prometheus-client: Update to 0.6.0
* pytest: Update to 4.4.1
* python-pytest-cov: Update to 2.7.1
* python-pytest-xdist: Update to 1.28.0
* python-hypothesis: Update to 4.18.3
* python-pyrsistent: Update to 0.15.1
* python-networks: Update to 2.3

PR #5363, opened 2019-05-17
python-more-itertools: Downgrade Python 2 version and split PKGBUILD

lazka:
more-itertools dropped Python 2 support with 6.0. The recent upgrade (#5291)
to 7.0 results in the package no longer working under Python 2 and breaks
packages depending on it like pytest.

Split the PKGBUILD up into a Python 2 (5.0) and Python 3 (7.0) version.

Add an epoch to the Python 2 version so that package upgrades work.


lazka:
@peterbud

PR #5393, opened 2019-05-21
some prerequisits for python-redbaron - a requirement of openmdao .

JPeterMugaas:
python-rply - - 0.7.7 - Update to  latest version
python-baron - 0.9 - New package required for redbaron (prerequisit topenmdao

Not ethat I can NOT yet provide redbarron because it requires pypandoc filters and a setuptools extension that requires pandoc.  Pandoc itself was written with the ghc compiler.

### DIFF
--- a/mingw-w64-python-baron/PKGBUILD
+++ b/mingw-w64-python-baron/PKGBUILD
@@ -1,0 +1,87 @@
+# Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
+
+_realname=baron
+pkgbase=mingw-w64-python-${_realname}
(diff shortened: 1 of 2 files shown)


lazka:
Do we need the Python 2 packages?

JPeterMugaas:
Okay.  From what I can gather, rebaron is a requirement for openmdao docs.  It's a good package to have anyway.  We still have openmdoa for python2.

lazka:
ok, note that openmdoa seems to support Python 3 as well.

lazka:
Feel free to re-create this PR with Python3 only packages.

Since the repo is gone I can't fix it myself, so closing.

PR #5410, opened 2019-05-25
Gdb 8.3 disable source highlight

StarWolf3000:
This will remove the new default dependency of boost and source-highlight on GDB 8.3 release, by providing the configure switch `--enable-source-highlight=no` and removing source-highlight from the package dependencies.

Has been tested locally, GDB doesn't crash if boost dlls are missing.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,108 @@
+# Maintainer: Alexey Pavlov <alexpux@gmail.com>
+# Contributor: Ray Donnelly <mingw.android@gmail.com>
+
+_realname=gdb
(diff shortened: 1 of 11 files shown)


lazka:
Something is not right here, you're not chaning the gdb PKGBUILD but adding a new one.

StarWolf3000:
Git(hub) didn't want me to make a PR without rebasing (or I did it the wrong way), so the files are recreated it seems.

lazka:
They are in the wrong path as well

StarWolf3000:
I saw it now, I do it again...

PR #5411, opened 2019-05-25
GDB 8.3 Disable Source Highlight

StarWolf3000:
This will remove the new default dependency of boost and source-highlight on GDB 8.3 release, by providing the configure switch `--enable-source-highlight=no` and removing source-highlight from the package dependencies.

Has been tested locally, GDB doesn't crash if boost dlls are missing.

(This time the PR should be correct)


lazka:
increase pkgrel

PR #5430, opened 2019-05-28
pidgin++  - 15.1 - change ncusrses dpenency to pdcurses.  OOtherwise,…

JPeterMugaas:
… it will get in the way of pdcurses update


Alexpux:
paging was removed from pacman database as not maintained. Need more work to update it

JPeterMugaas:
I wonder if it would be worth actually keeping in VCS then.

PR #5452, opened 2019-06-01
emacs-git: update

oscarfv:
Removed old patches, they failed to apply and had unknown purpose.

Remove dependency on dbus, emacs does not use it on Windows.

Move most libraries to makedepends and optdepends, as they are not
required to run emacs.

Use Universal Ctags instead of the unmaintained ctags.

Add dependency on harfbuzz. Emacs uses it by default if present.


kenkangxgwe:
Hi @oscarfv, it would be great if you also update the [non-git emacs package](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-emacs) to 26.2 and apply your changes if applicable.

oscarfv:
@Alexpux : any problem with this PR?

Alexpux:
@oscarfv please create pull request here https://github.com/msys2/MINGW-packages-dev

I'm cleaning up this repository from git packages that not providing via pacman. As we provide release emacs package in repository I will delete git package from here

PR #5475, opened 2019-06-05
Clang: Fix lld with g++

orgads:
and use symlinks instead of copies for all llvm tools.

### DIFF
--- a/mingw-w64-clang/0015-cmake-symlinks.patch
+++ b/mingw-w64-clang/0015-cmake-symlinks.patch
@@ -1,0 +1,48 @@
+From 31cf4b1ff245e0456a1f9b47fd818ed560afb259 Mon Sep 17 00:00:00 2001
+From: Orgad Shaneh <orgads@gmail.com>
+Date: Wed, 5 Jun 2019 08:55:57 +0300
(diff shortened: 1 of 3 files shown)


orgads:
Actually, since this is a single PR for 2 changes, maybe I should only change pkgrel once?

Alexpux:
@orgads MSYS doesnt support symlinks by default, instead it do deep copy

orgads:
I know, but this can be controlled by the MSYS env variable. This is better than telling cmake to unconditionally copy.

I'm currently building to see how makepkg behaves.

Alexpux:
Sorry but I will not accept this

orgads:
Ok, I'll separate the lld fix which is more important, and we'll discuss the links later.

Notice that the lld package contains 4 or 5 identical files of ~50 MB each because links are not used.

orgads:
Ok, I see that cmake doesn't behave well when the user doesn't have symlink permissions.

So I abandoned also the upstream change https://reviews.llvm.org/D62891

Can we consider replacing the copies with hardlinks at least when packaging? tar can preserve hardlinks.

Alexpux:
@orgads every user install MSYS2 and packages in defferent places, so we will have troubles with hardlinks

PR #5482, opened 2019-06-05
fix pyscript2exe.py so that it provides the propper shabbang just lik…

JPeterMugaas:
…e setuptools and then change the sshabbang for the user's particular system

mingw-w64-glib2
mingw-w64-gobject-introspection
mingw-w64-gtk-doc
mingw-w64-itstool

Note that the shabbang should be a file name and not "/usr/bin/env python2.exe" or something like that.  The env command is for accessing something in the environment.  I think this could be effecting some Gnome-build scripts.

### DIFF
--- a/mingw-w64-glib2/PKGBUILD
+++ b/mingw-w64-glib2/PKGBUILD
@@ -7,7 +7,7 @@ _realname=glib2
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.60.3
-pkgrel=1
+pkgrel=3
 url="https://www.gtk.org/"
 arch=('any')
 pkgdesc="Common C routines used by GTK+ 2.4 and other libs (mingw-w64)"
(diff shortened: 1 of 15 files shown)


lazka:
If you want to ignore PATH you can just use "!#python3.exe" like setuptools does. It will default to the directory of the "-script.py" file.

JPeterMugaas:
It works that way for .bat files but the shabbang and setuptools launcher seem to require a path particularly if the script is involved from the bash prompt.  It's usually set in  setuptools install script.  It's usually my practice to use post install scripts to set things properly for the user's system.

JPeterMugaas:
One part of this is that sometimes, you might want to invoke either the Win32 or Win64 mingw wrapper executable from the MSYS2 shell.  That's what I sometimes will do.  In the MSYSTEM=MSYS, the path is not set to point to /mingw64 or /mingw32 and often, invocation scripts require stuff that is compiled or placed in the particular ptyhon version-specific library and bash unlike "cmd" does not invoke stuff in the same dir.  An .exe will load a dll from the same dir but that's NOT what we are talking about (I only mention it to point out a difference).

But no matter what, a shabbang should never be something like "#!/usr/bin/env python2.exe"  And as I say, we want to be sure that the script runs in the proper python interpreter (the same basic version that  built the script).

lazka:
The script is never executed from the bash prompt, only through the setuptools launcher, so that shouldn't matter.

JPeterMugaas:
I think the setuptools .exe launcher works the same way.  That's been my experience and yes, I think I was calling the .exe from the MSYS command.

lazka:
Uhm, no, the launcher has its own shebang parsing code and we even patch it to understand /usr/bin/env.

You can try with meson.exe, it works without setting PATH and without any hardcoded paths so it's relocatable.

JPeterMugaas:
Meson's shabbang is:

#!python3.exe

so you might be right.  One reason why I was putting paths to the .exe in was because I have six versions of python that are not interchangable and I want to be ABSOLUTELY certain about which one of those the script is meant to invoke.  Let me explain the versions:

Python for MSYS - /usr/bin/python2.exe -/usr/bin/python3.exe
Python for mingw-w64-i686 - /mingw32/bin/python2.exe - /mingw32/bin/python3.exe
Python for mingw-w64-x86_64 - /mingw64/bin/python2.exe - /mingw64/bin/python3.exe

PR #5503, opened 2019-06-10
libmicrohttpd: added more build options

Karlson2k:
* Added support for debug build
* Added `check()` step.


Karlson2k:
@Alexpux, what the minimal requirement for OS for mingw packages? WinXP or something less old?

Alexpux:
@Karlson2k  we switch some times ago to minimum target Win7

PR #5504, opened 2019-06-10
libmicrohttpd: build for Win7+

Karlson2k:
Use Vista+ features in MHD

### DIFF
--- a/mingw-w64-libmicrohttpd/PKGBUILD
+++ b/mingw-w64-libmicrohttpd/PKGBUILD
@@ -7,7 +7,7 @@ _realname=libmicrohttpd
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.9.64
-pkgrel=1
+pkgrel=2
 pkgdesc="GNU libmicrohttpd is a small C library that is supposed to make it easy to run an HTTP server as part of another application (mingw-w64)"
 arch=('any')
 url="https://www.gnu.org/software/libmicrohttpd"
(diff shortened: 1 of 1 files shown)


Alexpux:
@Karlson2k there are no need to add it to CPPFLAGS as it already defined in mingw-w64 headers

Karlson2k:
Headers do not override previously defined `_WIN32_WINNT`, so usually application / library is free to define required operation system version.
By default MHD use `0x0501` value for `_WIN32_WINNT` before including headers, so binaries are compatible with WinXP.
If any value of `_WIN32_WINNT` is provided by CPPFLAGS, that supplied value is used.

Check 'configure summary' results: without CPPFLAGS it do not detect `WSAPoll`, with this patch `WSAPoll` is detected. Recently deleted `001-fix-verify-_WIN32_WINNT.patch` did not change configure results.

Changing default definition in headers do not change automatically OS target version for applications and libraries as default value is not enforced and used only if application/lib do not define its own `_WIN32_WINNT`.

Alexpux:
@Karlson2k then we need remove WINNT definition from libmicrohttpd, so it will not overridee system default value

Karlson2k:
@Alexpux System headers are designed in way to allow application/lib to choose default `_WIN32_WINNT`, right?
And headers still allow usage of values lower than `0x0603`.
Libmicrohttpd *do not* override `_WIN32_WINNT`, it define default value.

Same for `FD_SETSIZE` for example. Headers define default value only unless app define value of `FD_SETSIZE`.
Defining value of `_WIN32_WINNT` or `FD_SETSIZE` *before* including system headers is correct and proper way of providing default values.

Karlson2k:
MS Visual Studio headers are designed in the same way.
They provide **default** value for `_WIN32_WINNT` only if other value is not provided in compiler flags or in app source.
For VS project value of `_WIN32_WINNT` is chosen automatically and incorporated in precompiler definition (actually - to compiler '/D' flags).

Karlson2k:
From MS headers:
```` C
// Including SDKDDKVer.h defines the highest available Windows platform.

// If you wish to build your application for a previous Windows platform, include WinSDKVer.h and
// set the _WIN32_WINNT macro to the platform you wish to support before including SDKDDKVer.h.

#include <SDKDDKVer.h>
````
From SDKDDKVer.h:
```` C
#if !defined(_WIN32_WINNT) && !defined(_CHICAGO_)
#define  _WIN32_WINNT   0x0A00
#endif
````

This is official way: define `_WIN32_WINNT` before including system headers. Headers will correctly use provided value instead of default one.
This is not called **override**.

Alexpux:
@Karlson2k ok

Karlson2k:
@Alexpux :+1: Thanks.

Karlson2k:
I see how much confuse this generate, so I think special 'configure' flag to target specific OS version should be implemented.

PR #5512, opened 2019-06-11
Update Mesa3D to 19.1.0

pal1000:


Alexpux:
@pal1000 source extraction failed. Seems in source archive symlinks was added

Alexpux:
@pal1000 this is really wrong what you are doing. In this case sources will not present in source archive package.
You can workaround issue with symlinks like in ````qt-creator```` package or add git url to ````source````

Alexpux:
@pal1000 fixed by https://github.com/msys2/MINGW-packages/commit/9214f303756f0db77fe80202e86479b509dbb1d5

PR #5514, opened 2019-06-12
gcc-git: Add Ada and object c build guards

stahta01:
Copied the _enable_ada=no and _enable_objc=no changes to gcc-git.

Tim S.


Alexpux:
@stahta01 ehy you remove ````--enable-plugin```` from configure options?

stahta01:
> @stahta01 ehy you remove `--enable-plugin` from configure options?

I moved it so only GCC 9 version uses it; GCC 7 does not build with it.

Tim S.

Alexpux:
@stahta01 not only gcc9 but later too

stahta01:
Closing to do more testing and editing

PR #5521, opened 2019-06-12
gnome-calculator: Update version to 3.33.2

stahta01:
The decimal point and thousand seperator no longer looks for system set values.


Alexpux:
@stahta01 better to fix using proper separators because different languages use different separators.

lazka:
Not sure how easy this is with vala..

PR #5533, opened 2019-06-16
Fix glib g_fopen/g_open/g_creat patch to work with more file open modes

### LINKED ISSUE #5150: glib: patch causes g_fopen to fail on some open flags
This is caused by https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-glib2/0001-Use-CreateFile-on-Win32-to-make-sure-g_unlink-always.patch

Specifically, try opening a file in `a+`, `w+` or `wb+` mode, and it will fail. [msdn docs](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019) indicate that these are valid. It seems extremely fragile for this patch to be matching certain flags with `strcmp`, I don't even think that the order of the flags is meant to matter.

Can anyone shed light on what this patch is even trying to accomplish?

jtanx:
Fixes #5150


lazka:
upstream: https://gitlab.gnome.org/GNOME/glib/issues/539#note_532266

PR #5535, opened 2019-06-16
jsonschema: move attr, pyrsistent and setuptools from makedepends to depends

peterbud:
Interestingly the sha256sums was also wrong?


lazka:
> Interestingly the sha256sums was also wrong?

github generated tarballs are not stable, afaik they are cached and regenerated on demand... so sometimes they change.

peterbud:
I think it was my fault... I had a previously downloaded tarball, and the original checksum was the good. Now it looks better

lazka:
(I think setuptools-scm in makedepends might help)

peterbud:
Thanks @lazka ! Now it build nicely in CI as well.

PR #5569, opened 2019-06-24
Add a new package - dcraw

joankaradimov:
`dcraw` is a console tool for manipulation of raw image formats from digital cameras.


Alexpux:
@joankaradimov ````dcraw.exe```` is part of gimp-ufraw package

PR #5575, opened 2019-06-30
gtk3: fix g_content_type_is_a() errors

lazka:


lazka:
https://gitlab.gnome.org/GNOME/gtk/merge_requests/969

This helps with meld and gedit. envince is a different problem.

PR #5592, opened 2019-07-07
mingw-w64-make(4.2.1): bug fix to unlock GNU make's dll load feature (autotools build)

jannick0:
Addition to #5574:

1.  BugFix: Unlock GNU Make's dll load feature in the executable.
2.  BugFix: User-defined dll's for the dll load feature are linked against     the created import library libgnumake-1.dll.a which itself refers   to its associated executable.  The executable name is hardcoded in the     import library.  In order not to break the dll load feature, this implies     for the autotools build branch that in the installation process the     executable must not be renamed and autotools has to create the main     executable with our preferred name  (here: mingw32-make.exe).  Thus     the autotools package is to be informed about the new executable name     if different from 'make'.
3.  Update the build branch using build_w32.bat of the GNU Make package.     The standard executable is 'gnumake' which currently cannot be changed     without breaking the dll load feature.

Documentation on GNU make's dll load feature: [here](https://www.gnu.org/software/make/manual/html_node/Loading-Objects.html#Loading-Objects).

Apologies for the lengthy comments containing the essentials about the commit. Please feel free to edit and shorten as you might consider it apporpriate.

### DIFF
--- a/mingw-w64-make/PKGBUILD
+++ b/mingw-w64-make/PKGBUILD
@@ -1,12 +1,18 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
 _realname=make
+# NOTE: The name of the target executable (_prog_name) MUST remain unchanged
+#       when installed, otherwise GNU make's dll load feature in connection
+#       with the target executable breaks.  The reason is that the executable
+#       name is part of the import library (libgnumake-1.dll.a) used to link
+#       customized dll's.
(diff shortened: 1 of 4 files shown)


jannick0:
@Alexpux Hi Alex, AppVeyor succeeds, while CI fails in a customized test at the end of `package()` which checks whether the dll load feature really works or not.  It is about Windows calling `GetProcAddress` for a symbol in the test dll. My local checks with MinGW64 do work.

Is the CI failure a showstopper, while Appveyor succeeds?  I am not totally sure if I can do anything about the CI error.

PR #5604, opened 2019-07-12
New package: md4c

AntumDeluge:
C Markdown parser implementation compliant to CommonMark specification.

Homepage: https://github.com/mity/md4c


Alexpux:
@AntumDeluge see our other PKGBUILDs with cmake builds and cnage configure and install in the same way

AntumDeluge:
Will look at more when I get an opportunity.

AntumDeluge:
Okay, I made some changes. I set the variables ***extra_config*** & ***MSYS2_ARG_CONV_EXCL*** in the ***build*** function as I saw had been done in some other PKGBUILD files that use cmake.

Let me know if these are the changes you were talking about & if there is something else I need to do.

**Note:** I noticed that some of the other PKGBUILD files, such as [poppler](https://github.com/msys2/MINGW-packages/blob/1e34b45c64baec19b28ee6201d255d61cedd6501/mingw-w64-poppler/PKGBUILD) & [libjpeg-turbo](https://github.com/msys2/MINGW-packages/blob/70266199ec19f47b5a34db203856c004cf5dbe60/mingw-w64-libjpeg-turbo/PKGBUILD), do not set the ***extra_config*** variable. Perhaps because their configurations do not support the ***CMAKE_BUILD_TYPE*** flag?

PR #5605, opened 2019-07-12
gtk3: re-enable gtk debug features

lazka:
gtk ties the debug options to the build type, but since we use "plain" so that
the the makepkg CFLAGS get used we lose that.

Instead pass the CFLAGS manually.

See https://gitlab.gnome.org/GNOME/gtk/issues/2020


lazka:
@Ede123

PR #5625, opened 2019-07-21
Fix Creation of Python3 Virtualenv

### LINKED ISSUE #5001: Creation of a Python Virtualenv Fails
This is somewhat related to #2745, but the original issue posted is different, so I thought it would be worthwhile having an issue posted for virtualenv failing.

I would like to have Python virtualenv's working in msys2. When trying to make a virtualenv:

```
$ virtualenv venv
Using base prefix 'C:/msys64/mingw64'
New python executable in C:/Users/DYEAW/Projects/test-venv/venv/Scripts/python3.exe
Also creating executable in C:/Users/DYEAW/Projects/test-venv/venv/Scripts/python.exe
Installing setuptools, pip, wheel...
  Complete output from command C:/Users/DYEAW/Proje.../Scripts/python3.exe - setuptools pip wheel:
  File "<stdin>", line 3, in <module>
ModuleNotFoundError: No module named 'pkgutil'
----------------------------------------
...Installing setuptools, pip, wheel...done.
  File "C:\msys64\mingw64\bin\virtualenv-script.py", line 11, in <module>
    load_entry_point('virtualenv==16.0.0', 'console_scripts', 'virtualenv')()
  File "C:/msys64/mingw64/lib/python3.7/site-packages\virtualenv.py", line 712, in main
    symlink=options.symlink)
  File "C:/msys64/mingw64/lib/python3.7/site-packages\virtualenv.py", line 947, in create_environment
    download=download,
  File "C:/msys64/mingw64/lib/python3.7/site-packages\virtualenv.py", line 904, in install_wheel
    call_subprocess(cmd, show_stdout=False, extra_env=env, stdin=SCRIPT)
  File "C:/msys64/mingw64/lib/python3.7/site-packages\virtualenv.py", line 796, in call_subprocess
    % (cmd_desc, proc.returncode))
OSError: Command C:/Users/DYEAW/Proje.../Scripts/python3.exe - setuptools pip wheel failed with error code 1
```
Or trying it the now recommended way:
```
$ python3 -m venv venv
Error: [WinError 2] The system cannot find the file specified
```
These are the installed Python 3 packages:
```
mingw-w64-x86_64-python3 3.7.2-1
mingw-w64-x86_64-python3-apipkg 1.5-2
mingw-w64-x86_64-python3-appdirs 1.4.3-3
mingw-w64-x86_64-python3-atomicwrites 1.3.0-1
mingw-w64-x86_64-python3-attrs 18.2.0-1
mingw-w64-x86_64-python3-beaker 1.10.0-2
mingw-w64-x86_64-python3-cachecontrol 0.12.5-1
mingw-w64-x86_64-python3-cairo 1.18.0-1
mingw-w64-x86_64-python3-certifi 2018.11.29-2
mingw-w64-x86_64-python3-chardet 3.0.4-2
mingw-w64-x86_64-python3-colorama 0.4.1-1
mingw-w64-x86_64-python3-distlib 0.2.8-1
mingw-w64-x86_64-python3-gobject 3.30.4-1
mingw-w64-x86_64-python3-html5lib 1.0.1-3
mingw-w64-x86_64-python3-idna 2.8-1
mingw-w64-x86_64-python3-iniconfig 1.0.0-2
mingw-w64-x86_64-python3-lockfile 0.12.2-1
mingw-w64-x86_64-python3-mako 1.0.7-3
mingw-w64-x86_64-python3-markupsafe 1.1.0-1
mingw-w64-x86_64-python3-more-itertools 5.0.0-1
mingw-w64-x86_64-python3-msgpack 0.6.1-1
mingw-w64-x86_64-python3-packaging 19.0-1
mingw-w64-x86_64-python3-pep517 0.5.0-1
mingw-w64-x86_64-python3-pip 19.0.1-1
mingw-w64-x86_64-python3-pluggy 0.8.1-1
mingw-w64-x86_64-python3-progress 1.4-3
mingw-w64-x86_64-python3-py 1.7.0-2
mingw-w64-x86_64-python3-pyparsing 2.3.1-1
mingw-w64-x86_64-python3-pytest 4.3.0-1
mingw-w64-x86_64-python3-pytoml 0.1.20-1
mingw-w64-x86_64-python3-requests 2.21.0-1
mingw-w64-x86_64-python3-retrying 1.3.3-1
mingw-w64-x86_64-python3-setuptools 40.8.0-1
mingw-w64-x86_64-python3-six 1.12.0-1
mingw-w64-x86_64-python3-urllib3 1.24.1-1
mingw-w64-x86_64-python3-virtualenv 16.0.0-1
mingw-w64-x86_64-python3-webencodings 0.5.1-3
```

danyeaw:
This pull request fixes #5001, by fixing a few different issues:

# Issues
1. In Python 3.7.2, Windows now uses redirector scripts for `python[w].exe` instead of copying the actual binaries. The code to use these redirector scripts was causing the initial error messages in the issue above.

2. When a virtualenv was being created, the exe's were being installed in a `Scripts` directory instead of `bin`. Although this is consistent with Windows behavior, it is not consistent with other POSIX environments.

3. The script to activate a virtualenv was adding a Windows format path of the virtualenv to the front of the PATH, so activating the virtualenv didn't allow the shell to find the exe's inside the virtualenv which caused it to use the system base Python.

4. When the virtualenv was being created, it ran `ensurepip` to bootstrap pip in to the environment. This was causing AttributeErrors which would cause pip not to be functional in the virtualenv.

5. The `python.exe` and `python3.exe` executables were missing the `exe` file extension when they were copied in to the virtualenv.

# Solutions
1. Patch 2050 adds `_POSIX_BUILD` to the if statements to not use Windows redirectors,

2. Patch 2050 also adds a `_POSIX_BUILD` check so that it doesn't use the `Scripts` directory.

3. Patch 2051 converts the Windows path `__VENV_DIR__` using `cygpath`.

4. Patch 2052 removes the MSYSTEM environmental variable from `os.environ`, and then passes that env in when bootstrapping pip in to the virtualenv.

5. Path 2052 also adds the exe extension to the python executables.

# Testing
I have built the mingw64 package and successfully created a fully working virtualenv with the mingw64 shell. I recommend users set `export PIP_REQUIRE_VIRTUALENV=true` in their `.bashrc` so that pip always installs packages in the virtualenv, and keeps the virtualenv more isolated from the system packages.

# Other Issues
The virtualenv script that comes with the mingw-w64-ARCH-python3-virtualenv is no longer a required package and I didn't test it. [Python 3.6 deprecated this](https://docs.python.org/3/library/venv.html), so virtualenv's are now created using `python3 -m venv`. I can submit another PR to remove the Python3 packages for this.


mingw-w64-python3/2050-undo-venv-redirector.patch
```diff
@@ -0,0 +1,43 @@
...
+             f.write('include-system-site-packages = %s\n' % incl)
+             f.write('version = %d.%d.%d\n' % sys.version_info[:3])
+ 
+-    if os.name != 'nt':
++    if os.name != 'nt' or _POSIX_BUILD:
+         def symlink_or_copy(self, src, dst, relative_symlinks_ok=False):
+             """
+             Try symlinking a file, and if that fails, fall back to copying.
+             """
+-            force_copy = not self.symlinks
```
lazka: Why is this needed? Shouldn't self.symlinks be false anyway on Windows?

lazka:
bump pkgrel

lazka:
Nice work!

mingw-w64-python3/2052-force-pip-install-creating-venv.patch
```diff
@@ -0,0 +1,19 @@
...
++        # after the third time, and then forcing a reinstall:
++        # https://github.com/pypa/pip/issues/5820
+         cmd = [context.env_exe, '-Im', 'ensurepip', '--upgrade',
+                                                     '--default-pip']
++        subprocess.run(cmd, stderr=subprocess.DEVNULL)
++        subprocess.run(cmd, stderr=subprocess.DEVNULL)
++        subprocess.run(cmd, stderr=subprocess.DEVNULL)
++        cmd = [context.env_exe, '-Im', 'pip', 'install', '-U',
++               '--force-reinstall', 'pip']
+         subprocess.check_output(cmd, stderr=subprocess.STDOUT)
```
lazka: This seems to work:

```
        env = os.environ.copy()
        env.pop("MSYSTEM", None)
        subprocess.check_call(cmd, stderr=subprocess.STDOUT, env=env)
```

lazka:
I noticed that there is a "python" and "python3" file under "somevenv/bin" and they are lacking the ".exe" extension. This makes `python.exe -c "import sys; print(sys.prefix)"` not point to the venv.

mingw-w64-python3/2050-undo-venv-redirector.patch
danyeaw: @lazka I thought I was getting `Unable to symlink` errors previously, so this is why I made this change. After reviewing the code more, you are right, it shouldn't be needed. Thanks for the backup 👍

mingw-w64-python3/2052-force-pip-install-creating-venv.patch
danyeaw: I agree this is much less hacky. Do you know why removing MSYSTEM from the environment fixes the pip bootstrap? I made this update, and it works great, but I would like to understand it more if possible.

danyeaw:
@lazka Thanks for your help polishing this PR, I really appreciate it.

That's interesting because for me, I was getting the expected output:
```
(.venv)
DYEAW@WIN MINGW64 ~/gaphor
$ python3 -c "import sys; print(sys.prefix)"
C:/tools/msys64/home/DYEAW/gaphor/.venv
```
However, we are using Poetry to manage our dependencies for Gaphor, and with your suggested change it is no longer giving errors about not finding an activated virtualenv. Nice one! 🎉

mingw-w64-python3/2051-set-venv-activate-path-unix.patch
```diff
@@ -0,0 +1,11 @@
+--- Python-3.7.4/Lib/venv/scripts/common/activate.orig	2019-07-20 19:02:54.072985400 -0400
++++ Python-3.7.4/Lib/venv/scripts/common/activate	2019-07-20 20:10:57.797211100 -0400
+@@ -37,7 +37,7 @@
+ # unset irrelevant variables
+ deactivate nondestructive
+ 
+-VIRTUAL_ENV="__VENV_DIR__"
++VIRTUAL_ENV=$(cygpath "__VENV_DIR__")
```
Alexpux: This will work only under MSYS2 shell, not from CMD

lazka:
> That's interesting because for me, I was getting the expected output:

It's only affecting "python.exe" and your example runs python3

mingw-w64-python3/2052-force-pip-install-creating-venv.patch
lazka: No idea. My guess would be that something compares paths by string to look something up and with MSYSTEM set the path separator is different.

mingw-w64-python3/2051-set-venv-activate-path-unix.patch
lazka: Would the bash script be usable from CMD?

Alexpux: @lazka ah sorry, I forgo that this is bash script

lazka:
Since venv is now working I've played around with the pyperformance bechmark suite and building Python with (minimal) PGO and it results in a nice 15-20% speedup.

danyeaw:
Very cool extra benefit, speed ups are always great :smile:

PR #5643, opened 2019-07-25
libwebp: define WEBP_EXTERN as __declspec(dllexport) in CPPFLAGS

sezero:
Otherwise, the dlls export everything.


Alexpux:
@sezero i think it must be fixed in sources instead

sezero:
I didn't want to patch it, but will do so now in a new PR.

Alexpux:
@sezero passing this flags will broke static library

sezero:
SDL uses the same method, but no one links static-SDL in a dll, so I see your point.
I can leave it as is if you want.

PR #5644, opened 2019-07-25
libwebp: define WEBP_EXTERN as __declspec(dllexport)

sezero:
Otherwise, the dlls export everything.


Alexpux:
@sezero need more work because it will break static library

sezero:
More work will further complicate things. Will leave it as is.  Closing.

Alexpux:
@sezero see how it done in most other libraries. It not very complicated

sezero:
> @sezero see how it done in most other libraries. It not very complicated.

I know how it's done (I maintain a couple of such libraries mysel),
but I'm not motivated to do it for this one :)  Maybe one day the
mainstream handles it properly. Until that day I'm withdrawing this PR.

PR #5663, opened 2019-07-31
Termcap: Update URL & add pkg-config Support

AntumDeluge:
1. Changes homepage URL:
    - old: https://www.gnu.org/gnu/termcap/
    - new: https://www.gnu.org/software/termutils/
2. Creates file `lib/pkgconfig/termcap.pc` for pkg-config support.


Alexpux:
@AntumDeluge increase pkgrel

AntumDeluge:
Sorry about that. Done.

PR #5666, opened 2019-07-31
Enable building libraw with zlib and jpeg support

kmilos:


Alexpux:
@kmilos fix dependencies: libjpeg-turbo instead jpeg

PR #5684, opened 2019-08-05
Fixing pkg-config for gsettings-desktop-schemas

albfan:
Seems we are doing some unneeded replacements for pkg-config. Maybe those are things related to autotools, not needed anymore for meson.


Alexpux:
wrong change really

PR #5679, opened 2019-08-05
julia: remove libgcc_s_dw2-1.dll when packaging

smalltalkman:
@Alexpux Sorry, my final pkgbuild file was not submitted to pr.

PR #5688, opened 2019-08-06
julia: set `USE_LLVM_SHLIB = 0` and fix install dir

smalltalkman:
Continue from https://github.com/msys2/MINGW-packages/pull/5679


smalltalkman:
@Alexpux Why did AppVeyor build failed? The downloaded log is incomplete.

mati865:
1,5 hour time limit.

smalltalkman:
@mati865 Thanks.

vchuravy:
`USE_LLVM_SHLIB=0` is not an encouraged build configuration since it will break packages depending on LLVM.jl (https://github.com/maleadt/LLVM.jl), which uses the C-API of LLVM and is used to add support for GPU computing. 

Just FYI.

PR #5690, opened 2019-08-07
dlib: add python-dlib packages

kanonet:
Hi,
currently there is no python binding in mingw-w64-dlib package.

I think there are 3 different ways to add the missing binding:
1. include them in all inside the existing mingw-w64-dlib packages (e.g. OpenCV does this)
2. create a separate package script to create the individual python binding packages (e.g. zeromq does this)
3. extend the the existing package script to generate individual packages

In this PR I choose the third option, the package script will now create 6 packages:
mingw-w64-i686-dlib
mingw-w64-i686-python2-dlib
mingw-w64-i686-python3-dlib
mingw-w64-x86_64-dlib
mingw-w64-x86_64-python2-dlib
mingw-w64-x86_64-python3-dlib

@peterbud @Alexpux please review and make any changes you want. :)

### DIFF
--- a/mingw-w64-dlib/PKGBUILD
+++ b/mingw-w64-dlib/PKGBUILD
@@ -1,10 +1,10 @@
 # Maintainer: Peter Budai <peterbudai@hotmail.com>
 
 _realname=dlib
-pkgbase=mingw-w64-${_realname}
-pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgbase=("mingw-w64-${_realname}" "mingw-w64-python-${_realname}")
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}" "${MINGW_PACKAGE_PREFIX}-python2-${_realname}" "${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 pkgver=19.17
-pkgrel=1
+pkgrel=2
 pkgdesc="A toolkit for making real world machine learning and data analysis applications in C++ (mingw-w64)"
 arch=('any')
 url='http://dlib.net/'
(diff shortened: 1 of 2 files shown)


mingw-w64-dlib/PKGBUILD
```diff
@@ -49,25 +63,73 @@ build() {
...
       -DBUILD_SHARED_LIBS=ON \
       ../${_realname}-${pkgver}
-
   cmake --build . --config Release
+
+  for pver in {2,3}; do  
+    msg "Python ${pver} build for ${CARCH}"  
+    cd "${srcdir}/python${pver}-build-${CARCH}"
+    ${MINGW_PREFIX}/bin/python${pver} setup.py build -G "MSYS Makefiles" \
+	  --set USE_AVX_INSTRUCTIONS=ON \
```
Alexpux: Replace all tabs with spaces

mingw-w64-dlib/PKGBUILD
```diff
@@ -49,25 +63,73 @@ build() {
...
   make install DESTDIR="${pkgdir}"
 
   # Remove hard coded library path from cmake files
   local PREFIX_DEPS=$(cygpath -am ${MINGW_PREFIX})
-
   pushd "${pkgdir}${MINGW_PREFIX}/lib/cmake/dlib" > /dev/null
-  sed -s "s|${PREFIX_DEPS}|${MINGW_PREFIX}|g" -i ./dlibConfig.cmake
-  sed -s "s|${PREFIX_DEPS}|${MINGW_PREFIX}|g" -i ./dlib.cmake
+  sed -s "s|${PREFIX_DEPS}/include||g" -i ./dlibConfig.cmake
+  sed -s -e "s|${PREFIX_DEPS}/lib/||g" -e "s|${PREFIX_DEPS}/include||g" -i ./dlib.cmake
```
Alexpux: It looks not ok for me

mingw-w64-dlib/PKGBUILD
```diff
@@ -49,25 +63,73 @@ build() {
...
-
   cmake --build . --config Release
+
+  for pver in {2,3}; do  
+    msg "Python ${pver} build for ${CARCH}"  
+    cd "${srcdir}/python${pver}-build-${CARCH}"
+    ${MINGW_PREFIX}/bin/python${pver} setup.py build -G "MSYS Makefiles" \
+	  --set USE_AVX_INSTRUCTIONS=ON \
+      --set DLIB_ISO_CPP_ONLY=OFF \
+      --set DLIB_IN_PROJECT_BUILD=ON \
```
Alexpux: Probably you must also set CMAKE_INSTALL_PREFIX=${MINGW_PREFIX}

kanonet: Does this make a difference? Done.

mingw-w64-dlib/PKGBUILD
kanonet: What old Code does: rename machine specific absolute paths to msys/relative paths:
C:/msys64/ming64/include -> /mingw64/include
C:/msys64/ming64/lib/libgif.dll.a -> /mingw64/libgif.dll.a
But those paths do not work for me. When building a project with dlib, I always get an error message:
cmake: 'Imported target "dlib::dlib" includes non-existent path "/mingw64/include"'
'c++.exe: error: /mingw64/lib/libgif.dll.a: No such file or directory'
Of cause the lib _is_ installed in that directory, the compiler can just not find it given this path. Since this is the standard directory that the compiler is checking anyway, I decided to change the script: now renameing happens this way:
C:/msys64/ming64/include -> include
C:/msys64/ming64/lib/libgif.dll.a -> libgif.dll.a
It simply works. I can not tell why the old approach does not work anymore, because it used to work in the past. Something has changed, maybe its the compiler? (Or just my setup?) This Code just works around the Problem, without knowing what caused it.

PR #5694, opened 2019-08-08
julia: Use libopenblas.dll instead of liblapack.dll; Open the test function; Install lib{c,llvm}calltest.dll to the bin directory.

smalltalkman:
Use libopenblas.dll instead of liblapack.dll; 
Open the test function; 
Install lib{c,llvm}calltest.dll to the bin directory.


smalltalkman:
@Alexpux 
Llvm has been statically linked with julia, and other third-party libraries still use dynamic linking. However, these third-party libraries are dynamically loaded, so we put these third-party libraries in `bin/julia-$version`. 
Some of the defects were adjusted and the test cases passed almost completely. So I think we have nothing to do now.

PR #5729, opened 2019-08-17
Build 'bpy' module

NebuHiiEjamu:
Enabled building the bpy.pyd Python module, which should be useful for debugging Blender addons externally, as Blender's script debugging via internal Python console is rather lacking. While this builds, I'm not sure if Blender itself is built with it. Should this be a separate package?


Alexpux:
@NebuHiiEjamu this commit break building Blender. 
First, blender executable skip buildig, second - builded module installing in wrong place.
So revert this, if you need this then create new pull request with proper changes

PR #5738, opened 2019-08-21
Revert "python3: Replace python3-virtualenv package"

lazka:
This reverts commit a934b5a5d1d50e71d4f27e3036c0711e16b397e8.

python3 doesn't provide a virtualenv command


Alexpux:
Our virtual event package not working so replacing it to not try to use is ok I think

PR #5742, opened 2019-08-23
python-theano: add `gcc' to `optdepends', and fix `blas_info' value.

smalltalkman:


smalltalkman:
When gcc is not installed, you need to change the `return tuple([_find_exe_version(cmd) for cmd in commands])` in the file `lib\python3.7\distutils\cygwinccompiler.py (line 438)` to `return tuple([_find_exe_version(cmd) or '' for cmd in commands])`.

PR #5751, opened 2019-08-27
wxWidgets3.1: Add install scripts

stahta01:
And, no longer apply 001-wxWidgets-3.0.2-relocate-prefix-in-bin-wx-config.patch

Tim S.


Alexpux:
@stahta01 this is bad for me and i will not accept it because MSYS2 always have /mingw{32,64} mount points

PR #5760, opened 2019-09-02
libogg: update to 1.3.4

Adsun701:
This updates libogg to 1.3.4. Build is switched to cmake.


lazka:
> Build is switched to cmake.

Why?

Adsun701:
I did this to mirror Arch Linux's ogg package upstream. I think Arch switched to cmake for ogg because of the `*.cmake` files that would be installed for use by packages that are built with cmake and require ogg as a dependency.

lazka:
Ah, I see, thanks.

PR #5773, opened 2019-09-08
Add packages from Git for Windows

bjartur:
Do these packages, including Git, some SSH agency stuff, and a package for building a bleeding edge Emacs straight from Savannah, belong in mingw64?


bjartur:
The unreleased version of Emacs, in the least, does not belong here.

PR #5778, opened 2019-09-09
Stylus

JPeterMugaas:


lazka:
What's with the ninja change? (and don't you think having one of our base build tools depend on emacs is a bit much?)

JPeterMugaas:
Alright, the highlighters are included in the ninja distribution, not the emacs distribution.  And emacs should only be a make dependency and an opt dependency.   The thing with the mingw-w64 packages is that the bias tends to be including everything in one package (it was something that I did like but I felt that I had to do get things accepted).   For the MSYS2 packages, I had one PKGBUILD that would have a split package.  One would be main package and the other would be thing slike support for emacs and vim.   So can simply download those packages without vim and emacs support if you don't want them.

Alexpux:
@JPeterMugaas i will not accept stylus as nodejs is broken for me and maybe I decide to remove all nodejs packages as no one want to maintain it

PR #5769, opened 2019-09-07
WIP: glib: Update to 2.62.0

lazka:

PR #5782, opened 2019-09-10
WIP: gobject-introspection: Update to 1.62.0

lazka:


lazka:
This should be fine, but I'd like to look into finishing #5769 first.

PR #5787, opened 2019-09-12
New package: bin2header

AntumDeluge:
**Binary to Header**

*bin2header* is a command line tool. It takes any file as an argument and converts its binary data into a source header file for use in C/C++ applications. The data is stored as a character array.

Current version: 0.2.0

Upstream: https://github.com/AntumDeluge/bin2header


AntumDeluge:
Thank you! 😃

PR #5813, opened 2019-09-26
Fix insecure hashes, prefer HTTPS (or HTTP even) over FTP

vszakats:


lazka:
Please try to be more specific with your commit messages.

vszakats:
It really is just an assortment of fixing sloppy practices: MD5, SHA1, missing checksum, FTP protocol. I'm not even sure how some of these are still allowed in new commits.

Do you have a suggestion for a better title?

lazka:
> Do you have a suggestion for a better title?

Anything that somewhat describes what you are trying to do, like:

> It really is just an assortment of fixing sloppy practices: MD5, SHA1, missing checksum, FTP protocol. I'm not even sure how some of these are still allowed in new commits.

lazka:
thanks!

vszakats:
You're welcome and thanks for the merge.

(The 'missing checksum' case was in an orphan (old) file, so it didn't make it into the patch.)

PR #5818, opened 2019-09-29
[NO-MERGE] Mesa3D CI test with Python 3 and Scons.

pal1000:
Mesa3D claims it can now be built with Python 3 and Scons since mesa3d/mesa@3f92d178 so I want to test it, so merging please.


pal1000:
I don't know what's wrong, but this isn't ready.

PR #5827, opened 2019-10-03
ruby 2.6.5 - run updpkgprovs fpr mew  version.  Added ttf-dejavu font…

JPeterMugaas:
…s aa makedepends.


Alexpux:
For what reason you add ttf fonts to make depends?

JPeterMugaas:
I think it is used for documentation or might be the default font for graphviz or doxygen.  I'm not sure but I do know that ArchLinux does list that in it's makedepends.  In any case, the versions of the bundled gems needed to be updated.

Issue #5821, opened 2019-10-01
Bullet physics has significantly worse performance than when built from source


20k:
Hi there! I've noticed that the bullet physics version in the package manager has significantly worse performance than one built from source. I'm not familiar with how packages work on here otherwise I'd have a look myself, though if I had to guess I'd guess it were this https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-bullet/workaround-mingw-gcc6-bug.patch because it seems to mess with optimisation settings

Bullet on my end was built with 

cmake -G "Unix Makefiles" -DBUILD_PYBULLET=OFF -DBUILD_PYBULLET_NUMPY=OFF -DUSE_DOUBLE_PRECISION=OFF -DBULLET2_MULTITHREADING=ON -DCMAKE_BUILD_TYPE=Release -DBUILD_BULLET2_DEMOS=OFF -DBUILD_CPU_DEMOS=OFF -DBUILD_OPENGL3_DEMOS=OFF -DBUILD_ENET=OFF -DUSE_GRAPHICAL_BENCHMARK=OFF -DBUILD_UNIT_TESTS=OFF ../

With ~1000 rigidbodies, bullet from pacman rapidly hits 180ms/frame of execution time. With bullet built from source, bullet takes < 8ms/frame of execution time (which is gpu execution time in my test app), which is a massive difference. These are identical starting conditions, the only difference is where bullet came from

This has been a problem for at least a year so it would be great to see it fixed!

Windows 10, i5 3570k, r9 390, everything up to date, using mingw64 with no modifications

If you need any more information or you need a test application for this, let me know

oscarfv:
The condition and the comment look contradictory on that patch.

Try building the package after removing that patch. See this page for instructions about modifying PKGBUILD and building packages:

https://github.com/msys2/msys2/wiki/Creating-Packages

20k:
Ok, I navigated the pkgbuild process and can confirm that that gcc6 patch is the one that introduces the performance issue

Just to check I didn't do anything insane: I checked out this repo, built the package once with default settings and tested the produced headers + libs + dlls locally (slow), then changed line 40 of PKGBUILD to read   #patch -p1 -i ${srcdir}/workaround-mingw-gcc6-bug.patch, rebuilt, rebuilt the project with the headers/etc, and then tested again (fast). I definitely used the release build for both, under ./pkg/mingw-w64-x86_64-bullet

[...]

PR #5828, opened 2019-10-03
remove old patch that doesn't seem to have any function anymore, causes performance issues for bullet physics https://github.com/msys2/MINGW-packages/issues/5821

20k:
From this issue over here https://github.com/msys2/MINGW-packages/issues/5821

There seems to be an old patch for gcc6 that disables optimisations for bullet physics, resulting in very poor performance. This simply removes the patch

This is my first time submitting this sort of thing so let me know if anything needs changing


lazka:
Increase `pkgrel`.

Btw, if you push your changes to a different branch on your fork it makes it easier to submit multiple pull requests in the future.

20k:
Thanks, I've bumped pkgrel to 2, and I'll bear that in mind for next time

PR #5841, opened 2019-10-07
qt5-static 5.13.1

Optiligence:
ready now, with workarounds 

ref https://gitlab.kitware.com/cmake/cmake/issues/19804


Optiligence:
@Alexpux just making sure you haven’t missed that it’s ready now

PR #5850, opened 2019-10-08
meson: Update to 0.52.0

lazka:


lazka:
This release should allow us to get rid of various .pc file hacks to remove static internal libs. But I haven't tested that yet.

PR #5857, opened 2019-10-10
clang: fix lldb CMake when epydoc is present

oscarfv:
See https://bugs.llvm.org/show_bug.cgi?id=43637

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -91,6 +91,7 @@ source=(${_url}/$pkgver/llvm-${pkgver}.src.tar.xz{,.sig}
         "0507-declare-exports-for-all-windows-platforms.patch"
         "0508-add-share-header.patch"
         "0509-include-crtdbg.patch"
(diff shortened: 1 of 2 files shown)


mingw-w64-clang/PKGBUILD
```diff
@@ -91,6 +91,7 @@ source=(${_url}/$pkgver/llvm-${pkgver}.src.tar.xz{,.sig}
         "0507-declare-exports-for-all-windows-platforms.patch"
         "0508-add-share-header.patch"
         "0509-include-crtdbg.patch"
+	"0510-fix-CMake-when-LLDB_DISABLE_PYTHON-if-ON-and-ep.patch"
```
mati865: Don't use tabs
```suggestion
        "0510-fix-CMake-when-LLDB_DISABLE_PYTHON-if-ON-and-ep.patch"
```

oscarfv: Thanks. Fixed.

PR #5890, opened 2019-10-24
ca-certificates: revert the Python 3 port 

lazka:
something is wrong with it but I can't reproduce it locally, so revert.


Alexpux:
@lazka do you try to rebuild reverted package and see if it still not broke curl?

lazka:
All packages I build locally work with curl (Python 2 or Python 3). Something on your machine is different somehow...

PR #5899, opened 2019-10-26
Fix broken patches for GCC10 master

euloanty:


Alexpux:
@euloanty git packages moved to https://github.com/msys2/MINGW-packages-dev
Please open issues for such packages there

PR #5904, opened 2019-10-27
gdb does not load user init file

bustercopley:
GDB automatically loads (and treats as safe) the user init file in $HOME\.gdbinit, but HOME is not typically set on Windows, so no init file is loaded. The existing MSYS2 patch "gdb-8.3-cache-dir.patch" makes GDB fall back to USERPROFILE if HOME is not defined, in some cases but not others.

This commit makes GDB fall back to USERPROFILE if HOME is not defined, _everywhere_ it looks at HOME.


Alexpux:
@bustercopley increase pkgrel

bustercopley:
> @bustercopley <https://github.com/bustercopley> increase pkgrel
>
@Alexpux done, thanks.

Alexpux:
@bustercopley CI tell that checksums are invalid. Seems you add wrong checksum for your patch

bustercopley:
@Alexpux Forgive the newbie mistakes. Should I squash the commits into one?

PR #5907, opened 2019-10-27
libvips: Fix build with gobject-introspect enabled

larskanis:
This could previously fail in make parallel builds due to a wrong
file dependency like so:
```
x86_64-w64-mingw32-gcc -march=x86-64 -mtune=generic -O2 -pipe
-D__USE_MINGW_ANSI_STDIO=1 -pipe  ../../vips-8.8.3/libvips/introspect.c
-o introspect
../../vips-8.8.3/libvips/introspect.c:42:10: fatal error: vips/intl.h:
No such file or directory
   42 | #include <vips/intl.h>
         |          ^~~~~~~~~~~~~
	 compilation terminated.
	 make[3]: *** [<builtin>: introspect] Error 1
```

Fixes #5885


larskanis:
I sent the patch upstream: https://github.com/libvips/libvips/pull/1455

PR #5909, opened 2019-10-27
Add yices2 SMT solver package.

cr1901:
`yices2` is an SMT solver like z3, but is (_much_) faster on some problems. I've tested `yices2` locally for months, although in the most recent version one of the regression tests fails, so I [disabled](https://github.com/msys2/MINGW-packages/compare/master...cr1901:yices2?expand=1#diff-7f5afad67b37bbeb62ae102e27581256R56) `make check`.

### DIFF
--- a/mingw-w64-yices/PKGBUILD
+++ b/mingw-w64-yices/PKGBUILD
@@ -1,6 +1,6 @@
 # Maintainer: William D. Jones <thor0505@comcast.net>
 
-_realname=yices2
+_realname=yices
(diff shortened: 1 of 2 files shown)


cr1901:
Ping... what further steps should I take to get this PR moving? Can we add the `new-package` label?

lazka:
Sorry for the late response

cr1901:
No worries, at least it's in now and I know the steps for getting a package accepted.

PR #5916, opened 2019-10-31
unicode-character-database - 12.1.0 - new package.  Unicode database…

JPeterMugaas:
… for building and testing various things from unicode.org


lazka:
lgtm

PR #5920, opened 2019-11-01
Restructure DocBOok

JPeterMugaas:
Devide into docbook-xml - 4.x series and separate 5.x series for ease of maintannce and consistancy with Archlinux
also added Docbook 5.1

### DIFF
--- a/mingw-w64-docbook-xml/PKGBUILD
+++ b/mingw-w64-docbook-xml/PKGBUILD
@@ -2,12 +2,12 @@
 # Contributor: Judd Vinet <jvinet@zeroflux.org>
 
 _vers_4x=(4.{2..5})
-_vers_5x=(5.0 5.0.1)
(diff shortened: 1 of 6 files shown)


Alexpux:
@JPeterMugaas i will not accept it as you do. Last time I spend lot of time to fix this packages after your changes.
First test everything that it will work. At least this construction ````file://${_datadir}```` not work at windows. So if you just copy this from ArchLinux without understanding how it works then dont do it. I dont want to fix it again

JPeterMugaas:
I am looking at this.  But one thing I need to note is that Archlinux's docbook5 does not include the 5.0.1 doc book version.

Part of this is that docbook is cumulative.  Docbook 5.x is a ctually of docbook and it probably works a lot differently.  I eventually came to the conclusion that we should split the 4x series from the 5x series.

PR #5921, opened 2019-11-01
PostgreSQL: Split server and client packages

larskanis:
The current postgresql package is rather large and has large dependencies like tcl and python2. This makes it inconvenient to be used for many client only purposes. I therefore propose two split packages as common on Linux distros. This greatly reduces the amount of files and dependencies to be installed when no server libraries are needed.

The new package is called "postgresql-client". Package "postgresql" installs "postgresql-client" implicit, so that the package is fully backward compatible. I used the [debian package repo](https://salsa.debian.org/postgresql/postgresql/tree/12/debian) to separate which files belongs to which package (although debian splits it into a dozen packages).

Background is that I'm maintaining both [RubyInstaller](https://rubyinstaller.org/) and [ruby-pg](https://github.com/ged/ruby-pg) (the ruby client library for postgres) and I would like to install ruby-pg gem on top of the postgresql-client package.

For some reason gcc links to libxml2.dll, when linker is called with -lxml2 , although --as-needed is specified. Unfortunatelly there seems to be no easy way to disable -lxml2 for client libraries only, so that the package "postgresql-client" depends on libxml2, although it has no use there other than fulfilling this dll dependency at runtime.

I successfully tested the postgresql-client package with ruby-pg.

### DIFF
--- a/mingw-w64-postgresql/PKGBUILD
+++ b/mingw-w64-postgresql/PKGBUILD
@@ -1,28 +1,26 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
+# Maintainer: Lars Kanis <lars@greiz-reinsdorf.de>
 
 _realname=postgresql
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-client")
 pkgver=12.0
 pkgrel=1
-pkgdesc="Libraries for use with PostgreSQL (mingw-w64)"
 arch=('any')
 url="https://www.postgresql.org/"
 license=('custom:PostgreSQL')
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-gcc-libs"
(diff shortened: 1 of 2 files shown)


Alexpux:
@larskanis sorry but I dont want this changes. You need understand that supporting such scripts is not easy in future, so I prefer to not split MINGW packages. If you need this then do it in your own repo.

larskanis:
What a bummer! I understand your concerns about additional complexity. However I purposely used an easy to understand list of wildcard paths rather than some "write-only" sed scripts. And by using `mv` and `find` it's safe in that way that no file goes into two packages and no file is lost from packaging.

MySQL/MariaDB has it's own client package, MSSQL has freetds and I think that PostgreSQL should have one as well.

What do you think about the idea that I take the responsibility about this particular package (and possibly the other packages I contributed)? Being the main developer of ruby-pg since 8 years and a (MINGW-) contributor to PostgreSQL, I have a some experience in this regard.

In fact, I already use an [own repository](https://dl.bintray.com/larskanis/rubyinstaller2-packages/) for RubyInstaller. However it is used for the build process only, and I don't want to use it at runtime. IMHO there's not enough value in this small change to establish an alternative repository just for RubyInstaller. Many tools beside Ruby need the client libraries only. It has more value if it's provided in the main repository.

lazka:
Any help is appreciated, but as said above we try to avoid splitting packages (unless there is a good reason)

PR #5942, opened 2019-11-07
mecab: support MeCab 0.996

kenhys:


Alexpux:
I think it is not what I want to add to our repo. We have lot packages and I stop adding new packages without good reason. I don’t have time to build and support all of this

kenhys:
MeCab is a well known, widely used Japanese morphological analyzer, but not available as MinGW pacakge yet.
I want to support MeCab feature in mingw-w64-groonga, but because of missing MeCab, it is required to disable [explicitly](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-groonga/PKGBUILD#L39). so I've created this PR.

kenhys:
@Alexpux 

FYI: MeCab is Japanese morphological analyzer, and
https://qa.debian.org/popcon.php?package=mecab shows the numer of installed user of MeCab on Debian (Over 3000 users, install based), brew info mecab shows the numer of them on macOS (Over 12,000 users in recent one year). It indicates that there are many users about MeCab. (OT: Groonga uses MeCab as a library to implement tokenizer for improving   language specific precision in fulltext search. On macOS, MeCab is already packaged with Homebrew, It is happy if we can provide same user experience on MinGW users, by making MeCab MinGW pacakge available)

> I don’t have time to build and support all of this

Are there any way to help maintainance of MeCab?

PR #5959, opened 2019-11-12
metis: PkgConfig support

okhlybov:
This adds PkgConfig support to help the dependant packages such as parmetis.


okhlybov:
x64 build failure was due to the pacman database sync error.

PR #5962, opened 2019-11-13
New package: parmetis

okhlybov:
Parallel MPI version of the metis.


mingw-w64-parmetis/0001-mingw-w64-does-not-have-sys-resource-h.patch
```diff
@@ -0,0 +1,15 @@
...
++++ parmetis-4.0.3/metis/GKlib/gk_arch.h	2019-11-11 22:46:53.625107300 -0800
+@@ -41,8 +41,9 @@
+ #endif
+   #include <inttypes.h>
+   #include <sys/types.h>
+-  #include <sys/resource.h>
+-  #include <sys/time.h>
++  #ifndef __MINGW32__
++    #include <sys/resource.h>
++  #endif  #include <sys/time.h>
```
Alexpux: Here is seems typo error. Second include must be inside ifndef

PR #5963, opened 2019-11-14
mecab: fix failed to load mecab dicrc issue

kenhys:
This commit fixes the following error:

  ../../mecab-0.996/src/param.cpp(69) [ifs] no such file or directory:
  /mingw64/lib/mecab/dic/naist-jdic\dicrc


Alexpux:
@kenhys this is wrong change and I will not accept it.
You must probably change paths in install script that run after package install

kenhys:
@Alexpux Thank you for feedback.
I've fixed to it.

Alexpux:
@kenhys you not fix. Both commits are totally wrong.

PR #5970, opened 2019-11-17
WIP: clang: Reduce install size by using hard links

k-takata:
Clang has some files which have different names with exactly the same
contents. (E.g. clang.exe, clang++.exe, clang-cl.exe and clang-cpp.exe)
This is a waste of disk size. Use hard links to reduce the install size.

(This reduces the install size about 1.4 GB in my environment.)


k-takata:
The same method can be also applied to gcc and binutils.

Alexpux:
@k-takata I will not accept such changes as it increase scripts a lot.

PR #5976, opened 2019-11-17
nlohmann_json: Rename to nlohmann-json and update to 3.7.3

podsvirov:
Rename to `nlohmann-json` to be consistence with Arch [package](https://aur.archlinux.org/packages/nlohmann-json/).
Update to 3.7.3 to fix latest bugs.

### DIFF
--- a/mingw-w64-nlohmann-json/PKGBUILD
+++ b/mingw-w64-nlohmann-json/PKGBUILD
@@ -1,10 +1,10 @@
 # Maintainer: Konstantin Podsvirov <konstantin@podsvirov.pro>
 
 _realname=json
-_pkgname=nlohmann_json
+_pkgname=nlohmann-json
 pkgbase=mingw-w64-${_pkgname}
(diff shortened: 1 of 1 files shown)


mingw-w64-nlohmann-json/PKGBUILD
```diff
@@ -1,20 +1,22 @@
 # Maintainer: Konstantin Podsvirov <konstantin@podsvirov.pro>
 
 _realname=json
-_pkgname=nlohmann_json
```
Alexpux: Well for compatibility reason better then add also provides entry with old name package

podsvirov: I added this entry below, have you not seen?

Alexpux: Ah I see now, but you also need add replace entry to be able upgrade

podsvirov: The replaces entry added.

PR #5977, opened 2019-11-18
[WIP][RFC] Switch Mesa3D to Meson build

pal1000:
I was able to build Mesa3D with Meson using MSYS2 Mingw-w64 since 19.3 1st release candidate when Windows Meson build support debuted. There is only one known issue, Mesa3D dlls are dynamically linked against zlib and Mingw-w64 CRT ignoring LDFLAGS, see [this upstream issue report](https://gitlab.freedesktop.org/mesa/mesa/issues/2012).

PR #5981, opened 2019-11-20
Mesa3D: Implement LLVM Meson wrap generator

pal1000:
Cherry picked from #5977. This PR only includes LLVM Meson wrap generator and nothing else as it isn't bound to any Mesa3D version. I thought it's a good idea to get it done early so that switch to Meson later becomes less of a hassle.  I also improved LLVM Meson wrap generator over #5977 implementation to properly get LLVM RTTI status instead of hard-coding it.


pal1000:
This got superseded by #5983, so closing.

PR #5983, opened 2019-11-20
Mesa3D: Update to 19.2.5 and implement LLVM Meson wrap generator

pal1000:
In addition to updating to 19.2.5 this PR includes LLVM Meson wrap generator from #5977 with improvement to properly get LLVM RTTI status instead of hard-coding it. I chose to include it as it isn't bound to any Mesa3D version. I thought it's a good idea to get it done early so that switch to Meson later becomes less of a hassle.

### DIFF
--- a/mingw-w64-mesa/llvmwrapgen.sh
+++ b/mingw-w64-mesa/llvmwrapgen.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+# Get LLVM libraries
+llvmlibs=$(${MINGW_PREFIX}/bin/llvm-config --libnames engine coroutines)
+
+# Get LLVM RTTI status
+rtti=false
+if [ $(${MINGW_PREFIX}/bin/llvm-config --has-rtti) = YES ]; then
+  rtti=true
+fi
+
+# Convert llvm-config output into a Python list
+llvmlibs="${llvmlibs//.a/}"
+llvmlibs=\'"${llvmlibs// /\', \'}"\'
+
+# Get MSYS2 Mingw-w64 runtime location
+msysloc=$(${MINGW_PREFIX}/bin/python3 -c "import sys; print(sys.executable)")
+msysloc=${msysloc//\"/}
+msysloc=${msysloc//\\/\/}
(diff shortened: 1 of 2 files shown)


mingw-w64-mesa/llvmwrapgen.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Convert llvm-config output into a Python list
+llvmlibs="${llvmlibs//.a/}"
+llvmlibs=\'"${llvmlibs// /\', \'}"\'
+
+# Get MSYS2 Mingw-w64 runtime location
+msysloc=$(${MINGW_PREFIX}/bin/python3 -c "import sys; print(sys.executable)")
+msysloc=${msysloc//\"/}
+msysloc=${msysloc//\\/\/}
+msysloc=${msysloc:0:-16}
```
fabiangreffrath: Sure it's a good idea to hard-code the string length of "/bin/python3.exe" here?

pal1000: I am running ${MINGW_PACKAGE_PREFIX}-python3 there to find where MSYS2 is on storage drive as I'll be using ${MINGW_PACKAGE_PREFIX}-meson to build Mesa3D at some point.  I also noticed MSYS2 is extremely consistent about where stuff gets installed based on its type. Mingw-w64 binaries always land in `${MINGW_PREFIX}/bin`.

Alexpux: @pal1000 MINGW_PREFIX doesn’t defined in your script. You use empty variable

pal1000: The script isn't supposed to run on its own. It is supposed to only execute through PKGBUILD,
As far as I know MINGW_PREFIX is a system variable defined with right value as long as you are in a MINGW32 or MINGW64 shell.

pal1000:
Due to 19.2.6 being released so quickly I'll retract this in favor of a new PR that updates to 19.2.6.

PR #5985, opened 2019-11-22
Mesa3D: Update to 19.2.6 and implement LLVM Meson wrap generator

pal1000:
In addition to updating to 19.2.6 this PR includes LLVM Meson wrap generator from #5977 with improvement to properly get LLVM RTTI status instead of hard-coding it. I chose to include it as it isn't bound to any Mesa3D version. I thought it's a good idea to get it done early so that switch to Meson later becomes less of a hassle.

Re-spin of #5983


pal1000:
v2 - Avoid hard-codec substring value that @fabiangreffrath raised concern about in #5983.

Issue #5980, opened 2019-11-19
mpv does not start, missing libplacebo dependency


Tatsh:
mpv installs without its needed `${MINGW_PACKAGE_PREFIX}-libplacebo` dependency.

lazka:
confirmed `libplacebo-18.dll => not found`. PR welcome :)

PR #5987, opened 2019-11-24
mpv: add libplacebo dep; increment pkgrel

Tatsh:
Related to #5980


lazka:
Any idea why it fails to build?

Tatsh:
This is the error and I don't know what causes it.

```
Unable to find development files for some of the required FFmpeg/Libav libraries. Git master is recommended.
```

ffmpeg was pulled in:

```
installing mingw-w64-i686-ffmpeg...
```

I don't know if this includes development files but I presume it must because I have to assume this package worked to build before.

PR #5989, opened 2019-11-25
pyopengl 3.1.3rc1

totaam:
- removes all the workarounds we had added for python 3.8 since those have now been merged
- adds a workaround for the missing `readme.rst` file containing the long description (which is not shown anywhere)


totaam:
I have no idea why the appveyor build failed.
I have tested this build on 3 different systems, one 32-bit and two 64-bit and I am not seeing any problems anywhere building or using pyopengl.

PR #6004, opened 2019-11-28
xpra 3.0.2

totaam:
- version bump 
- fixes the launch scripts so it will correctly fallback to using the python2 build if that is the only version installed


lazka:
(Note that mingwxx/bin/python is Python 3 now)

PR #6009, opened 2019-11-30
Python 2 purge - Round 1

lazka:
Some packages which don't have any reverse dependencies.


lazka:
This should be ready now.

lazka:
@Alexpux Any thoughts on this and Python 2 removal in general?

Alexpux:
@lazka I'm ok with python2 removal. But dont increase pkgrel as removing doesnt need rebuild

lazka:
OK, I'll try to remove the pkgrel bumps as much as possible. In some cases I had to fix URLs and dependencies as well..

lazka:
On a second thought, not bumping pkgrel means the Python 3 package will all makedepend on Python 2 still and I can't track the progress :/

Alexpux:
@lazka ok but i dont have time to rebuild this untill next year

lazka:
ok, no problem

PR #6011, opened 2019-11-30
wineditline: Fix missing "extern c" in header

### LINKED ISSUE #6008: Syntax error in header from wineditline
The file `/mingw32/include/editline/readline.h` belonging to `mingw-w64-i686-wineditline` is defective under C++: contains a closing `}` but no opening `extern "C" {`.

In light of upstream not having released in nearly two years, please consider either building directly from the upstream SVN or at least backporting revision 34 from the SVN repository:

```
Index: src/editline/readline.h
===================================================================
--- src/editline/readline.h     (revision 33)
+++ src/editline/readline.h     (revision 34)
@@ -46,6 +46,10 @@
 */
 #define DEFAULT_HISTORY_SIZE    200  /* default number of history entries */
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 typedef char **rl_completion_func_t(const char *, int, int);
 typedef char *rl_compentry_func_t(const char *, int);
 typedef void rl_compentryfree_func_t(void *);
```

Thank you.

StarWolf3000:
Fixes #6008

As I don't know how to integrate SVN sources, I instead supply the patch provided by @olafmandel in the mentioned issue.

### DIFF
--- a/mingw-w64-wineditline/005-fix-missing-extern-c.patch
+++ b/mingw-w64-wineditline/005-fix-missing-extern-c.patch
@@ -1,0 +1,13 @@
+--- src/editline/readline.h	2019-07-03 09:09:00.000000000 +0200
++++ src/editline/readline.h	2019-11-30 16:56:47.213858800 +0100
+@@ -46,6 +46,10 @@
+ */
+ #define DEFAULT_HISTORY_SIZE    200  /* default number of history entries */
+ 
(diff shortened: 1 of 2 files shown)


lazka:
Here is the upstream change for future reference: https://sourceforge.net/p/mingweditline/code/34/

lgtm

PR #6039, opened 2019-12-10
New package: tnef

joankaradimov:
Mostly based on https://aur.archlinux.org/packages/tnef/, although there's nothing too fancy.

There is a patch file for the current version (1.4.18). But the next version will feature a mingw fix (https://github.com/verdammelt/tnef/pull/43) and will compile without any patches.

### DIFF
--- a/mingw-w64-tnef/PKGBUILD
+++ b/mingw-w64-tnef/PKGBUILD
@@ -1,0 +1,41 @@
+# Maintainer: Joan Karadimov <joan.karadimov@gmail.com>
+# Contributor: Luca P <meti at lplab.net>
+# Contributor: Sergej Pupykin <pupykin.s+arch@gmail.com>
+# Contributor: Jeffrey 'jf' Lim <jfs.world@gmail.com>
+
+_realname=tnef
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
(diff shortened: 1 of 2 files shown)


mingw-w64-tnef/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+prepare() {
+  cd "${srcdir}/$_realname-$pkgver"
+  patch -p1 < "${startdir}"/index-functions.patch
+  [ -x configure ] || autoreconf
+}
+
+build() {
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}"
+  cd "${srcdir}/build-${MINGW_CHOST}"
+  ../$_realname-$pkgver/configure --prefix=/usr
```
lazka, line 30: the prefix is wrong. See https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.autotools-tarball#L46

joankaradimov: @lazka I've rebased on the latest `MINGW-packages` master and I've added the `prefix`, `build`, `host` and `target` to configure script.

lazka:
Building this produces an empty package.

PR #6044, opened 2019-12-12
spdlog: Rebuild

podsvirov:
This required for binary compatibility since dependency `fmt` package updated.
The patch `001-spdlog-1.4.2-include-cassert.patch` required to fix https://github.com/gabime/spdlog/issues/1340.


podsvirov:
@Alexpux thanks to rebuild `spdlog` afster new `fmt` package.

PR #6049, opened 2019-12-15
vala: patch for glib-2.0.vapigen that allows proper compilation of Time.local()

cwrsimon:
Without this patch a simple program like this would not compile in a MINGW environment:
`void main () {
    
    time_t original_time = time_t();
    print( Time.local(original_time)
		.to_string() );
}
`


Alexpux:
@cwrsimon need update vala version to latest also maybe or increase pkgrel

cwrsimon:
@Alexpux : Thanks for the feedback. I upgraded to vala-0.46.5 (latest stable release). Could you recheck, please? I did not increase pkgrel due to the increment of the release number (cf. https://wiki.archlinux.org/index.php/PKGBUILD#pkgrel).

PR #6056, opened 2019-12-18
neon: Adding neon package to MINGW

KyleJ61782:
This removes the MSYS2 requirement for packages that don't need it.


KyleJ61782:
It's helpful to have a neon build targeting MINGW without the MSYS2 dependency. This pull request creates a package that satisfies that requirement.

PR #6061, opened 2019-12-20
mingw-w64-mesa: Switch to Meson build

pal1000:
Also add glapi shared library and OpenGL ES drivers.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,26 +4,27 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=19.3.1
-pkgrel=1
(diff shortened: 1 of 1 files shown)


mingw-w64-mesa/PKGBUILD
```diff
@@ -67,29 +68,30 @@ prepare() {
...
 
 buildcmd(){
-  if [ ${_mach} = x86 ]; then
-    scons build=release
-  fi
-  if [ ${_mach} = x86_64 ]; then
-    scons build=release swr=0
-  fi
+  ${MINGW_PREFIX}/bin/meson ./build/windows-${_mach} \
+  --default-library=static \
```
lazka: Just wondering, why static and not both?

pal1000: For starters I focused on what was working with Scons, and I already tried and failed Meson install and got shared glapi and OpenGL ES drivers to build successfully.

PR #6065, opened 2019-12-21
cmake: Update to 3.15.6

c72578:


Biswa96:
There is already a PR for cmake but with a newer version.

lazka:
> There is already a PR for cmake but with a newer version.

The other PR is broken, so this should be fine in the mean time.

Issue #4993, opened 2019-02-20
Drop Python 2 packages


lazka:
Here is a list of leaf packages containing "python2" for starters.

Any thoughts on dropping them?

```
edit: removed, outdated
```

totaam:
Can we wait until after python2 is EOLed? (2020)

lazka:
Sure, I just wanted to get the discussion started with a first target (Arch is doing the same right now).

Do you need any package of the list above?

lazka:
An alternative to dropping packages might be to stop accepting Python 2 packages without a good reason i.e. only build new Python packages for Python 3.

totaam:
> Do you need any package of the list above?

Yes - OTOH:
* `mingw-w64-gst-python2`
* `mingw-w64-python2-PyOpenGL`
* `mingw-w64-python2-comtypes`
* `mingw-w64-python2-cx_Freeze`
* `mingw-w64-python2-ldap`
* `mingw-w64-python2-ldap3`
* `mingw-w64-python2-lz4`
* `mingw-w64-python2-nvidia-ml`
* `mingw-w64-python2-paramiko`
* `mingw-w64-python2-pkginfo` (indirectly)
* `mingw-w64-python2-pyu2f`
* `mingw-w64-python2-setproctitle`
* `mingw-w64-python2-zeroconf`

And when they get removed... we'll probably have to maintain them by hand for a little while.
(same as centos6 and python2.6... some things move slowly)

> An alternative to dropping packages might be to stop accepting Python 2 packages without a good reason

I'm fine with that, got mine in just in time ;)

lazka:
OK, thanks for the list.

Let's try to be a bit stricter on adding new packages for starters and re-evaluate next year.

danyeaw:
In preparation for this, could we rename all the python3 packages to just python?

spl1inter:
Disambiguation from upcoming python 4?

lazka:
> In preparation for this, could we rename all the python3 packages to just python?

I was more thinking that this might reduce the amount of work, an renaming goes a bit against that :)

peterbud:
What is our strategy for packages whose new versions do not support anymore python2 (e.g. sphinx)? Arch has simply dropped python2, but not sure what we shall do with those?

lazka:
[...]

PR #6070, opened 2019-12-22
Add back python2-setuptools

lazka:
This adds back a python 2 version of setuptools but with all dependencies vendored
(the upstream default..). To allow users to install Python 2 packages we no longer
have in the repo etc which should help them transitioning.

This also adds conflicts between py2/3 setuptools because of the esy_install rename.

See #4993


Alexpux:
@lazka I think we need keep python 3 packages in python folders, not python3. And python2 will be in python2 folders

lazka:
ok :+1:

PR #6074, opened 2019-12-23
Add back python2-pip

lazka:
This adds back a Python 2 version of pip but with all dependencies vendored.
In combination with setuptools this allows users to install Python 2 packages
from pypi.

See #4993


lazka:
CI fails because python2-setuptools isn't in the repo yet.

PR #6086, opened 2019-12-27
Drop geoclue

lazka:
It hasn't been updated in ages, isn't used by anything in the repo
and upstream clearly targets Linux.


programmin1:
Do you mean python-geoclue or geoclue2?
I did a writeup of this some time ago, the geoclue2 has working Python call to see where the device is.
https://howtotrainyourrobot.com/building-a-mobile-app-for-linux-part-4-gps-mobile-tracking/

Is there any other library we can use in this repo to find device location? Geoclue2 works well but maybe there is something else Windows uses?

PR #6099, opened 2020-01-03
libcdio: Disable rpath as it breaks pkgconfig

KyleJ61782:
Since the pkgconfig uses @LTLIBICONV@, rpath must be disabled to prevent the -R parameter from appearing in the Libs spec in the pkgconfigs.


lazka:
1. Which .pc file contains "-R"?
2. you need to increase pkgrel

KyleJ61782:
>     1. Which .pc file contains "-R"?
>     2. you need to increase pkgrel
1) The three .pc files that contain a reference to libiconv, although my hunch is that libcdio++.pc also should probably have a reference to libiconv. Here's what libcdio.pc looks like on my system when ```--disable-rpath``` isn't used. Notice the ```-R/mingw32/lib``` on the ```Libs``` line:
```
prefix=/mingw32
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: libcdio
Description: Portable CD-ROM I/O library
Version: 2.1.0
#Requires: glib-2.0 
Libs: -L${libdir} -lcdio  -lm -lwinmm -L/mingw32/lib -liconv -R/mingw32/lib
Cflags: -I${includedir}
```
I should add that it looks the same on mingw64, just with ```-R/mingw64/lib``` instead.
2) My apologies on that screw up. I've just fixed that.

lazka:
> I should add that it looks the same on mingw64, just with `-R/mingw64/lib` instead.

It's like this here: `Libs: -L${libdir} -lcdio  -lm -lwinmm -L/mingw64/lib -liconv` even if I rebuild it myself, which is why I'm a little confused.

lazka:
Ah, I fixed the same thing some time ago with https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libcdio/PKGBUILD#L53

So maybe you haven't updated.

KyleJ61782:
Ohhh...you know what, I'm looking at the .pc files _before_ they get packaged up. I came upon this because of the issue reported and fixed in #6100 and it made me look at this package as well. I'll remove the looped sed now.

lazka:
Thanks!

PR #6100, opened 2020-01-03
libcddb: Fix pkgconfig libs spec

KyleJ61782:
When using @LIBICONV@, the entire path to libiconv is specified in Libs,
instead of using -liconv. This breaks building with gcc on mingw-w64
using response files. Furthermore, using the full path is bad form.


lazka:
Increase pkgrel

KyleJ61782:
> Increase pkgrel
Oops. My apologies for forgetting this.

PR #6105, opened 2020-01-05
dlib: update to 19.19 snapshot

kanonet:
Currently the dlib package is based on dlib release 19.19 (but build script is not updated on git). Dlib release 19.19 does not work with most current version of OpenCV, what makes it quite unusable. This is fixed in current snapshots of dlib, but a new release is probably weeks or even months away. Since our package is unusable I think this is urgent. I ask you to accept this PR, that uses current snapshot to make the package usable again.


Alexpux:
I’m not agree with this. Provide patch that fix issue instead

PR #6106, opened 2020-01-05
dlib: fix compatibility with opencv and drop python 2

kanonet:
@Alexpux
This is a follow up to #6105. You did not give any reason why you turned down my PR, but I did what you asked me to do: Now dlib release 19.19 is used again and patches (taken from dlib github) are used to fix OpenCV compatibility. That patch can get removed once dlib 19.20 gets released.
I also included changes so the python-dlib package is build instead of python3-dlib.

### DIFF
--- a/mingw-w64-dlib/002-fix-opencv-in-19.19.patch
+++ b/mingw-w64-dlib/002-fix-opencv-in-19.19.patch
@@ -1,0 +1,157 @@
+diff --git a/dlib/opencv/cv_image.h b/dlib/opencv/cv_image.h
+index 5f224d003..f6f4bbe95 100644
+--- a/dlib/opencv/cv_image.h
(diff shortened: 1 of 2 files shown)


lazka:
Somehow there is a newer version in the repo than there is in git, looks like the last change wasn't pushed. This needs to be solved first I'm afraid.

@Alexpux do you know why?

Alexpux:
@lazka, probably I forgot to push changes when renaming python3 packages. Will look today or tomorrow. We have holidays till 9th

lazka:
ok, thanks, enjoy!

kanonet:
Of cause I noticed the missing commit, that is why I named my version 19.19.2. I think I did all the needed renaming, so you may consider to simply this PR.
Enjoy your holidays!

Issue #6111, opened 2020-01-08
[gdal] Please package current gdal 3.0.2-1 with current poppler version 0.83.0-1


tuxfede:
There is a dependency missmatch.

Currently gdal 3.0.2-1 package is compiled with poppler 0.82 (which includes libpoppler-92.dll) , but poppler 0.83 is the current distributed version (which includes libpoppler-93.dll).

Thanks in advance.

lazka:
Should be fixed in 7f18d1fc81ee712e

thanks @Alexpux

PR #6113, opened 2020-01-08
gdal: rebuild against newer poppler

lazka:
See #6111


Alexpux:
@lazka need to add upstream pathces to support poppler 0.83 and later. I will push today later

Issue #6102, opened 2020-01-04
Build LLVM openmp library with mingw?


adrpo:
As far as I can see clang -fopenmp fails because of missing -lomp (libomp).
Did anybody tried to compile libomp? Is anybody interested in having clang working with openmp?

The most problematic issue seems to be that we need MASM to compile some asm code from z_Windows_NT-586_asm.asm. I don't know that much ASM to be able to translate this to YASM or GAS. 

What I could do is build uasm: https://github.com/Terraspace/UASM/tree/v2.50 as a MINGW package and then use that to compile the asm file. This seems to work, some people have tried it: https://github.com/mstorsjo/llvm-mingw/issues/16
I will give the uasm way a try and then we can continue from there.

adrpo:
UASM now available as a MINGW package  via PR: #6104

adrpo:
This is done, closing.

PR #6115, opened 2020-01-08
build clang OpenMP

adrpo:
build clang-openmp library (#6102) 
- use uasm instead of masm to build the libomp .asm file 
- add uasm as dependency to the clang package (needed by openmp) 
- include the header having xbegin on mingw as well 
- don't use weak attribute on mingw 
- sed away the windows like /flag to -flag on the call to uasm


mingw-w64-clang/PKGBUILD
```diff
@@ -406,6 +442,14 @@ package_libcxx() {
   ${_make_cmd} -C ../build-${CARCH}/projects/libcxx -j1 DESTDIR="${pkgdir}" install
 }
 
+package_openmp() {
+  pkgdesc="OpenMP library (mingw-w64)"
+  url="https://openmp.llvm.org/"
```
Alexpux: probably it need to be depends on one of the parts of clang?

PR #6119, opened 2020-01-10
fix memcpy-param-overlap in relocatable patch

nehaljwani:
Prefer memmove over memcpy for overlapping memory regions, otherwise it
triggers the address sanitizer with:

```
ERROR: AddressSanitizer: memcpy-param-overlap: memory ranges [x,y) and [a,b) overlap
```

cc @mingwandroid


lazka:
increase pkgrel (and run updpkgsums)

mingwandroid:
Thanks @nehaljwani! Happy New Year, I hope you are well?

We should fix this in other relocation patches too I think.

nehaljwani:
Wish you a very happy new year too @mingwandroid !

I have updated the merge request to target all instances of the patch, but it seems that the CI is failing due to a missing build time dependency? I am not sure how to fix that:
```bash
$  makepkg-mingw --noconfirm --noprogressbar --skippgpcheck --nocheck --syncdeps --rmdeps --cleanbuild
==> Making package: mingw-w64-bison 3.0.4-3 (Sun, Jan 12, 2020  6:56:59 AM)
==> Installing missing dependencies...
error: target not found: mingw-w64-x86_64-m4
==> ERROR: 'pacman' failed to install missing dependencies.
==> Missing dependencies:
  -> mingw-w64-x86_64-m4
==> ERROR: Could not resolve all dependencies.
```

lazka:
mingw-w64-bison isn't in the repo and probably not buildable, so this should be fine.

nehaljwani:
Isn't it the case that mingw-w64-x86_64-m4 is missing from the repos? Should I remove changes from mingw-w64-bison as it is not buildable at the moment?

Alexpux:
@nehaljwani mingw versions of bison and m4 wads never released, so remove them from commit

PR #6123, opened 2020-01-11
Qt 5.14 (CMake 3.16, cotire)

Optiligence:
Doesn’t work yet: build fails because the binaries can’t initialize the platform plugin successfully. 
`-debug-and-release` builds are also broken upstream (https://wiki.qt.io/Qt_5.14.0_Known_Issues#Windows)
haven’t tried to build static at all yet

Removes the previous static plugins support from CMake and Qt in favor of their new upstream implementation. 
Qt’s configure step wants assimp 5, which doesn’t build without patches, but master does; so this is a git package for now.


Alexpux:
@Optiligence I will not accept building from git

Optiligence:
@Alexpux do you want to backport build system patches to the v5 release, then? (because i thought that this is not only extra effort, but at some point as untested as using master…)

I’m also not sure Qt really need v5, or if it’s just unnecessarily specified somewhere (because that’s not everywhere). But especially Qt should rather work with less than more patches…

Alexpux:
@Optiligence we can wait next release of assimp if not backport patches. 
In any case there are a lot of problems with Qt5 and QtWebkit now.

Optiligence:
Waiting for any next release just cascades, because assimp blocks Qt which blocks CMake. 

But assimp actually seems to work now. Not sure if it has to do with updated CMake or what.

Alexpux:
@Optiligence as I told I will not switch assimp to build from git. Thats all. We can switch Qt build to use internal assimp

Alexpux:
@Optiligence did you try static Qt build?

Optiligence:
yes

in QtCreator it currently complains with 
```
CMake Error:
  Error evaluating generator expression:

    $<IN_LIST:Qt5::AudioCaptureServicePlugin,$<TARGET_PROPERTY:QT_PLUGINS>;$<TARGET_PROPERTY:QT_PLUGINS_mediaservice>>

  Expression did not evaluate to a known generator expression
```

and outside i’m getting 
>ninja: error: 'Qt5::Zlib-NOTFOUND', needed by 'knossos.exe', missing and no known rule to make it
for my project

Optiligence:
The shared build seems to work (i actually deleted it and can’t test it atm because i wanted the debug libs too)
The last hunk [here](https://code.qt.io/cgit/qt/qtbase.git/commit/qmake/generators/win32/winmakefile.cpp?id=ce5dc3193279cdbcc34b583151ba05464444dd5a) together with patch 0008 (which adds the import lib suffix) was breaking the plugin initialization.

Optiligence:
Both flavors have no debug libraries atm.

Alexpux:
@Optiligence assimp 5.0.1 in repo https://github.com/msys2/MINGW-packages/commit/f96a7cf63420cd70230148745bf018a671511ef7

Optiligence:
thx

the `dll.a` patch is still breaking (platform) plugin initialization for the shared build

automatic plugin linking in the static build now works

not sure debug libraries are worth getting to work still
the external debug infos (that only work in the shared build) are not installed
and debug builds don’t append the `d` suffix anymore (successfully)

@Alexpux is there a reason the static builds chooses to use none of the possible system libraries (although they have static versions available)?

Alexpux:
@Optiligence there are discussion on Qt ML about issue with debug builds, also bugreport https://bugreports.qt.io/browse/QTBUG-80792

Optiligence:
@Alexpux not sure it’s gonna get much better than it is now
see the commit messages

to get an all-static build user code now has to set
`list(PREPEND CMAKE_FIND_LIBRARY_SUFFIXES .a .lib)`
before calling `find_package(Qt5 `…

to avoid this the `find_package` calls inside Qt could be patched away and flags just forwarded to the linker, like before
not sure if desired

that’s basically the only problem i see remaining atm

Alexpux:
@Optiligence rebase changes to current master

Optiligence:
@Alexpux done

mingw-w64-qt5-static/PKGBUILD
```diff
@@ -6,18 +6,16 @@
...
 if [ "${_variant}" = "-static" ]; then
   _namesuff="-static"
+  _build_mode=("-debug-and-release")
   _with_icu=no
   _with_fontconfig=no
   _with_openssl=yes
   _with_dbus=yes
   _system_freetype=no
-  _system_zlib=no
+  _system_zlib=yes
```
Alexpux: Why you enable system zlib for static build?

mingw-w64-qt5-static/PKGBUILD
```diff
@@ -28,12 +26,13 @@ if [ "${_variant}" = "-static" ]; then
...
   _build_examples=no # examples result in > 100 GiB
   _build_tools=yes
-  _build_tests=yes
+  _build_tests=no
   _make_docs=no # docs in shared qt5 are sufficient
 #  _ltcg="yes"
   _ltcg="no"
 else
   _namesuff=
+  _build_mode=("-release" "-force-debug-info" "-separate-debug-info")
```
Alexpux: Why not debug-and-release?

mingw-w64-qt5-static/PKGBUILD
Optiligence: i don’t remember if something didn’t work (probably)
i asked in https://github.com/msys2/MINGW-packages/pull/6123#issuecomment-577068358 why it is non-system for static anyway (my guess is to test both kinds of build options for shared and static)
i then tried to _only_ use system libs which didn’t work well
so it is the way it is because that at least works

mingw-w64-qt5-static/PKGBUILD
Optiligence: this is the build option of the official release and i think it’s an improvement from the dual libraries before

PR #6129, opened 2020-01-13
rocksdb: Update to 6.5.3 and tweak build settings

EchterAgo:
This compiles rocksdb with RTTI support and in portable mode.

Portable mode is needed because otherwise the resulting binary might be unusable on older processors.

See https://bugs.archlinux.org/task/65093


lazka:
lgtm

PR #6141, opened 2020-01-25
[WIP] modular Qt

Optiligence:
i used this extensively to debug Qt 5.14
split pkg to deduplicate the patching, although that means building in the package function…
the install of qmake based builds is just hacky atm
the deps for all submodules other than base are wrong
not all modules are covered
and it would require to adjust the specific qt module deps in all packages


zgyarmati:
@Optiligence, I would like to help with this.
 Did you work on this since January, or this is the latest state?

Alexpux:
In the end of the year Qt6 will be released, so I think we can wait it as it will be switched to cmake build and maybe it will be easer to do it modular

Optiligence:
@zgyarmati i added my unpushed commits. i added you to my fork so you can directly push into the branch. 

i only worked on qt5-base (none of the other modules) for the previous fixes, and there have been changes in the big package since which have to be synced. Also i think i only applied patches to other modules if they were required for a successful build. I think there are still unused patches whose necessity needs to be evaluated (and hopefully move in the direction of getting upstreamed, if necessary). 

@Alexpux i think working on more fine grained specification of package dependencies (i.e. qt5-base or qt5-multimedia instead of qt5) for individual packages can already be very helpful. And any update will be much more easy to handle due to the lower build times (maybe even direct upload from ci), so if someone is willing to work on it i wouldn’t wave it off.

lazka:
Does anyone know what's the build time for qt5-base compared to our current package? It's annoying that qt can't be built in CI.

Optiligence:
Very little. 
Especially with a lot of cores, since e.g. qt-doc uses only a single core and takes ages to build. But i guess that matters less on CI. 
I don’t have single or dual core timings for you…

lazka:
thanks. hm, do we package any docs? can this be skipped?

Optiligence:
doc is skipped for static atm

otherwise i think declarative might be the biggest part to compile

i guess all the split packages could build fine in CI
it’s not just 1 or 2 that take their time and it just accumulates to a lot in total

lazka:
we just need each makepkg call for each arch to be under 6 hours on a dual code

lazka:
ok, CI managed to build it under 5 hours after all (after some build fixes)

lazka:
If anyone manages to split it up into two or more packages (even the docs would be an improvement) then please open a new PR. I tried some weeks ago for the docs alone, but gave up, likely due to my limited qt knowledge.

@Optiligence thanks for trying

closing for now

PR #6149, opened 2020-01-27
mpv: Update to 0.32.0

Biswa96:


Biswa96:
Should I rename `python3` packages and commands to `python` only?

PR #6151, opened 2020-01-28
ninja: Update to 1.10.0

lazka:


lazka:
Appended a CI change here so that it gets tested with a normal update.

PR #6166, opened 2020-02-04
gmp - 6.2.0 - added patch from Cygwin.

JPeterMugaas:


Biswa96:
Just curious, is the patch reported to upstream?

lazka:
Why do we need it?

JPeterMugaas:
I think it adds support for some types of chip architectures.  I found it when looking at the CygWin package for gmp.

lazka:
looking at the upstream repo it looks like the CPU detection is more up to date for x86_64 and this tries to also add it for x86. I wonder if anyone cares about x86 that much still if performance is an issue and if we should apply this if none of us understand the change.

lazka:
Also we don't build with "--enable-fat", so I'm wondering if this changes anything at all.

lazka:
I tried asking on the gmp mailing list since they don't have a bug tracker, but it blocks my messages, so I'm giving up...

JPeterMugaas:
I think we actually with "--enable-fat".  That's what I see in the script.

lazka:
> I think we actually with "--enable-fat". That's what I see in the script.

missing word? Which script?

JPeterMugaas:
Alright, I was looking at the wrong script.  It was the one for MSYS2.  But ideally, I would like to see the fat capability in MINGW.

lazka:
What about adding it to this PR?

lazka:
No response, closing.

PR #6167, opened 2020-02-05
Update libconfini (1.12.0 -> 1.13.0)

madmurphy:


lazka:
```
==> Extracting sources...
  -> Extracting libconfini-1.13.0.tar.gz with bsdtar
libconfini-1.13.0/dev/docs/baremetal.md: Can't create 'libconfini-1.13.0/dev/docs/baremetal.md'
bsdtar: Error exit delayed from previous errors.
==> ERROR: Failed to extract libconfini-1.13.0.tar.gz
    Aborting...
```

madmurphy:
Hi lazka,

That is strange, it worked just fine on my laptop. Could it be possibly due to the fact that the file it fails to extract is a symbolic link?

lazka:
yeah, that's likely the problem. See https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-grpc/PKGBUILD for example where we ignore extraction error.

madmurphy:
Thank you, lazka. Now it should work.

PR #6171, opened 2020-02-07
python-comtypes: include patch from upstream to fix missing module errors

totaam:
For details see
https://github.com/enthought/comtypes/issues/161


mingw-w64-python-comtypes/PKGBUILD
```diff
@@ -15,14 +15,18 @@ license=('OSI')
 depends=("${MINGW_PACKAGE_PREFIX}-python")
```
lazka: bump pkgrel

totaam: Oops, sorry. Done.

PR #6179, opened 2020-02-09
docbook-xsl: correct catalog directory

catcafeplus:
Supersedes #6082


lazka:
@catcafeplus this broke he gtk build, see #6194. Do you know what could be the problem?

catcafeplus:
@lazka This PR is modified from #6082 

I look at the PKGBUILD in arch, it has the following line
`pkgroot="$pkgdir/usr/share/xml/docbook/xsl-stylesheets-$pkgver$ns"`

But we have `pkgroot="$pkgdir${MINGW_PREFIX}/share/xml/docbook/xsl-stylesheets$ns-$pkgver"`

So my modifications are probably wrong, will create another PR.

lazka:
@catcafeplus thanks for looking into it :)

PR #6180, opened 2020-02-09
mecab: adjust dicdir in mecabrc

kou:
We need to use absolute path in Windows instead of absolute path in
MSYS2.


kou:
Related: #5963

PR #6184, opened 2020-02-11
GDB: Fix support for 64-bit targets on 32-bit

orgads:
If GDB is built with --enable-targets=all it should be able to connect
remotely and open core dumps from all target types.

But this change broke it for 32-bit build.

This partially reverts be921a6c47be60e25514032ba53835a98e4d5046.


Biswa96:
Is it possible to update gdb to 9.1 also? Link https://ftp.gnu.org/gnu/gdb/

orgads:
I already have a working patch for upgrading to 9.1. I just need to tweak the patch files to avoid fuzzes (not critical). I'll try to push it this evening or tomorrow.

orgads:
Done: https://github.com/msys2/MINGW-packages/pull/6190

PR #6186, opened 2020-02-11
make: Update to 4.3

jannick0:
* mingw-w64-make/PKGBUILD:
  - config: set arch to 'i686' 'x86_64', since batch build is supposed
    for pure Windows builds.
  - prepare: build with batch file build_w32.bat (instead of autotools,
     - since autotools approach is not fully supported for Windows builds).
     - overwrite default executable name 'gnumake' with 'mingw32-make'.
     - cd to src dir of package to apply package patches, since the dir
        structure was changed with make version 4.3.
  - check: move check section in package to separate function check.
  - use overall helper function defining the executable dir.

* mingw-w64-make/make-check-load-feature.mak:
  - add CFLAGS the user can pass to the makefile on the command line
  - simplify setting of default value of LIBS

### DIFF
--- a/mingw-w64-make/PKGBUILD
+++ b/mingw-w64-make/PKGBUILD
@@ -7,12 +7,12 @@ _realname=make
 #       name is part of the import library (libgnumake-1.dll.a) used to link
 #       customized dll's.
 #       General information at https://www.gnu.org/software/make/manual/html_node/Loading-Objects.html#Loading-Objects.
-_autotools=yes
+_autotools=no
 _prog_name=mingw32-make
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=4.2.1
-pkgrel=4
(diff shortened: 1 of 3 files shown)


mingw-w64-make/PKGBUILD
```diff
@@ -7,67 +7,85 @@ _realname=make
...
 _prog_name=mingw32-make
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=4.2.1
-pkgrel=4
+pkgver=4.3
+pkgrel=1
 pkgdesc="GNU make utility to maintain groups of programs (mingw-w64)"
-arch=('any')
+arch=('i686' 'x86_64')
```
Alexpux: This is wrong for mingw-packages. Revert this

jannick0:
@Alexpux The squashed version I just force-pushed should now be ready for merge from my side.

PR #6188, opened 2020-02-12
groonga: Update to 9.1.2

kou:


kou:
We can close #5964 once this is merged.

PR #6190, opened 2020-02-13
GDB: Upgrade to 9.1

orgads:


Ede123:
This update introduces a dependency on libxxhash.

Was this intentional (i.e. has advantages) so xxhash should be added to dependencies?  
Otherwise it can be disabled using `--with-xxhash=no`

orgads:
It wasn't intentional, but since it's a tiny package, I suggest just adding it as a dependency. Can you please handle it? I'm away from a Windows machine right now :)

PR #6195, opened 2020-02-13
nsis: Update to 3.05 and make it work with Python 3 scons

lazka:


lazka:
Also tested by creating an installer with it.

PR #6203, opened 2020-02-16
gmime: Update to 3.2.6

lazka:


lazka:
The build will likely fail because !6199 isn't in the repo

PR #3196, opened 2017-12-08
gdal: make postgresql dependency non optional

filnet:
see https://github.com/Alexpux/MINGW-packages/issues/3169

PR #6208, opened 2020-02-18
Add libpq client package

jeroen:
A lighter alternative for the [mingw-w64-postgres](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-postgresql) for packages that only need the client library, and not the full database server system. In particular GDAL, see: https://github.com/msys2/MINGW-packages/pull/3196

Similar package exists in [libpq.rb](https://github.com/homebrew/homebrew-core/blob/master/Formula/libpq.rb) in homebrew or [libpq-dev](https://packages.debian.org/sid/libpq-dev) in Debian.

In order to make sure packages can depend on either `postgresql` and `libpq`, we provide a virtual package `postgresql-client` similar to the [arch package](https://www.archlinux.org/packages/extra/x86_64/postgresql-libs/). So other packages should use:

```sh
makedepends=("${MINGW_PACKAGE_PREFIX}-libpq")
depends=("${MINGW_PACKAGE_PREFIX}-postgresql-client")
```

That way the package can be installed if `postgresql` is already installed.

### DIFF
--- a/mingw-w64-libpq/PKGBUILD
+++ b/mingw-w64-libpq/PKGBUILD
@@ -1,0 +1,69 @@
+# Maintainer: Alexey Pavlov <alexpux@gmail.com>
+
+_realname=libpq
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=12.2
+pkgrel=1
+pkgdesc="Postgres C API library (mingw-w64)"
+arch=('any')
+url="https://www.postgresql.org/"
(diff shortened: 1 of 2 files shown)


Alexpux:
Don’t want to add package that already contained in other package

jeroen:
Are you sure? Every other distribution does this because postgresql is the only db that does not separate the server and client. So you need to depend on the entire postgresql server system just to use the client.

mlt:
It is surprising that, say, psycopg pulls the server. I think it would be better if *postgresql* package was split into yet another package *postgresql-libs*.

PR #6223, opened 2020-02-25
qca: always link to release library

olafmandel:
We only compile in release mode and worse: the Qt library itself is only
available in release mode. So we don't want to link to a (non-existing)
debug-library.


Alexpux:
@olafmandel mow we build Qt5 with separate debug info, so when you debug application gdb automatically load debug information from separate file if it present.

PR #6227, opened 2020-02-26
python-gssapi 1.6.2

totaam:
Release notes:
https://github.com/pythongssapi/python-gssapi/releases/tag/v1.6.2


Biswa96:
Shouldn't the link be like this?

```
https://github.com/pythongssapi/${_realname}/releases/download/$pkgver/${_realname}-$pkgver.tar.gz
```

lazka:
There is more broken here: https://packages.msys2.org/base/mingw-w64-python-python-gssapi

there is "python" twice in the package name and the URL is using the wrong quotes.

Alexpux:
Done by https://github.com/msys2/MINGW-packages/commit/45344bbfb1dbce24667968b8add445a006a16db3

PR #6235, opened 2020-02-28
reduce SDL2_Mixer's dependency chain

fabiangreffrath:
I build games using SDL2_Mixer with MSYS2 and would like to distribute
them as binary packages.

However, SDL2_Mixer's hard dependency on libfluidsynth pulls in a huge
dependency chain that includes third-party libraries that apparently do
not have anything to do with music playback, such as e.g. libglib and
libreadline. This adds several MB to the binary distribution without
adding anything useful, since fluidsynth is unable to render MIDI
music anyway until an additional soundfont is also provided.

By removing the "--disable-music-midi-fluidsynth-shared" configure
parameter we will not lose fluidsynth support in SDL2_Mixer, we will
merely demote this library to an optional dependency - i.e. it will get
dlopen()ed at run-time if available.

This commit also turns modplug into an optional dependency, not
because of its own dependency chain but because I consider it a rather
uncommon music format to support. Support for the free digital music
formats, i.e. FLAC, OGG, MP3 and Opus should always remain a hard
dependency, though.


Alexpux:
@fabiangreffrath you must also change dependencies in PKGBUILD

fabiangreffrath:
> @fabiangreffrath you must also change dependencies in PKGBUILD

I have left these untouched on purpose, so the MSYS2 package works as before once installed. It is just that I want to get rid of the hard fluidsynth dependency when I copy SDL2_Mixer.dll over or package it in a ZIP file for redistribution.

Alexpux:
@fabiangreffrath increase pkgrel then

fabiangreffrath:
> @fabiangreffrath increase pkgrel then

Done.

> @fabiangreffrath you must also change dependencies in PKGBUILD

We could change fluidsynth and libmodplug to `optdepends` if you prefer.

PR #6239, opened 2020-02-28
dav1d: Add dav1d 0.5.2 PKGBUILD

1480c1:
Based on archlinux's PKBGUILD and also using the meson parts from one from @lazka

The dav1d.exe is linked to libdav1d.dll and works

Do I need a prepare function? I've set it to no-op for now

### DIFF
--- a/mingw-w64-dav1d/PKGBUILD
+++ b/mingw-w64-dav1d/PKGBUILD
@@ -1,0 +1,71 @@
+# Maintainer: Christopher Degawa <ccom@randomderp.com>
+
+_realname=dav1d
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=0.5.2
+pkgrel=1
+pkgdesc="AV1 cross-platform decoder focused on speed and correctness (mingw-w64)"
+arch=('any')
+url='https://code.videolan.org/videolan/dav1d'
+license=('BSD')
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-pkg-config"
+             "${MINGW_PACKAGE_PREFIX}-nasm"
+             "${MINGW_PACKAGE_PREFIX}-meson"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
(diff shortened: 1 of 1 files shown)


mingw-w64-dav1d/PKGBUILD
```diff
@@ -0,0 +1,86 @@
...
+}
+
+build() {
+    cd "${srcdir}"
+    [[ -d build-${MINGW_CHOST} ]] && rm -rf build-${MINGW_CHOST}
+    mkdir -p build-${MINGW_CHOST}
+    cd build-${MINGW_CHOST}
+
+    meson \
+        --default-library shared \
```
lazka: Would 'default_library=both' work instead so we only have to build once for both shared and static?

mingw-w64-dav1d/PKGBUILD
```diff
@@ -0,0 +1,86 @@
...
+        --buildtype plain \
+        -Denable_tests=false \
+        ../${_realname}-${pkgver}
+
+    ninja
+
+    for pcfile in meson-private/*.pc; do
```
lazka: Please move to this to after the install like https://github.com/msys2/MINGW-packages/blob/6febb352eb1667f19e5f745f3ca3e82e19c97879/mingw-w64-glib2/PKGBUILD#L89

We don't want to depend on meson internal things.

lazka:
Cool, thanks!

mingw-w64-dav1d/PKGBUILD
1480c1: I would assume so, but I wasn't sure why the other meson PKGBUILD I looked at had the static and shared built separately, now that I think about it, it could be because of gir?

1480c1: Done

mingw-w64-dav1d/PKGBUILD
1480c1: Done

mingw-w64-dav1d/PKGBUILD
lazka: Thank! Two reasons I know, default_library wasn't available in older meson, and some libs set various defines for the static builds so both builds aren't the same and have to be build separately.

PR #6240, opened 2020-02-29
vlc: Use binary-version to not conflict with a system installed vlc

1480c1:
Currently, when using the vlc installed by pacman, there's a chance
that vlc will pick the user's installed vlc's plugin instead of the
one installed by pacman

I'm not sure if all of the following are necessary

---

add glew to makedepends to fix missing `GL/glew.h`

Add dav1d as an av1 decoder

Add cyrus-sasl to fix missing `sasl/sasl.h`

Add post-install/upgrade vlc-cache-gen

Without running vlc-cache-gen, vlc complains about:
    `main libvlc error: stale plugins cache: modified`

### DIFF
--- a/mingw-w64-vlc/PKGBUILD
+++ b/mingw-w64-vlc/PKGBUILD
@@ -11,15 +11,17 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}-git")
 replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}-git")
 pkgver=3.0.8
-pkgrel=3
+pkgrel=4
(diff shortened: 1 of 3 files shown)


Biswa96:
`mingw-w64-dav1d` is not in pacman repository. Check out https://packages.msys2.org/new.

1480c1:
It's just been added to this repository

Biswa96:
Git repository and Pacman repository are different. The package is manually added to Pacman repo by Alexpux.

1480c1:
Yeah, I could sort of tell based on the snippets I saw previously on gitter, I'm just leaving this open until then, unless it would be better to split the changes and dav1d into two prs

lazka:
I've reran the CI tests now

Issue #6232, opened 2020-02-27
pgModeler


kassane:
It would be interesting to have this application builded.

[pgModeler source](https://github.com/pgmodeler/pgmodeler)
[Installation](https://www.pgmodeler.io/support/installation)

StarWolf3000:
Citing the website:
> On all supported platforms the first requirement is the **Qt framework** in the **LTS (Long Term Support)** modality. In this step-by-step the **Qt 5.12.x** was used so it is strongly recommended that you use the same version too. But if you like to live on the edge, you can try to compile pgModeler using the most recent version of Qt but **bear in mind that in most cases the compilation can fail due to incompatibilities between the tool's source code and the framework in its newer version or even the deployment steps below may not be enough to make the software work.**
>

This repo only provides more recent Qt versions on the 5.14 branch, and they even state that it may require a lot of extra work to make it compile.

> Second, you will need a **PostgreSQL** installation. In this document we are using the RDBMS in its version **11** so note that the location of headers and libraries in the settings below may vary depending on the version of the database system you've chosen.
>

The PostgreSQL RDBMS might work in version 12 (which is the current one in this repo).

ilhamsa1:
Qt versions 5.14 make my pgmodeller crash, how to install version 5.12?

kassane:
> 
> 
> Qt versions 5.14 make my pgmodeller crash, how to install version 5.12?

Current version only!

PR #6243, opened 2020-02-29
pgModeler

kassane:
Ref: #6232


kassane:
@Alexpux, I'm having a problem with homedrive in the patch file. Unfortunately this project is based on qmake and I can only manually direct the files directory.

Alexpux:
@kassane I cant accept this really bad patch. See other qmake based programs that we patch and try to do something acceptable. See for example qwt, qscintilla. Probably you need to rewrite PRO file for project

PR #6245, opened 2020-03-01
binutils: Install config.h for bfd

orgads:
bfd.h is useless without it.


lazka:
hm, config.h should never ever be installed. Can you go into more detail re what is broken for you?

orgads:
I'm using bfd.h for reading symbols on backtrace logging. If I include bfd.h it fails because config.h must be included before it. So either there's no reason to install bfd.h, or config.h should be installed along with it.

orgads:
From bfd.h:

```
/* PR 14072: Ensure that config.h is included first.  */
#if !defined PACKAGE && !defined PACKAGE_VERSION
#error config.h must be included before this header
#endif
```

lazka:
The developer suggests defining PACKAGE: https://sourceware.org/bugzilla/show_bug.cgi?id=14243#c10

(imo this is all super broken..)

orgads:
Accepted. Thanks.

lazka:
Thanks for testing it. I guess if there are still problems we can accept your patch, given the the weird upstream stance on this.

PR #6249, opened 2020-03-02
qwt: always link to release library

olafmandel:
Since qwt is now only compiled as a release library with separate debug
symbols, we must not advertise a non-existing debug-library.


Alexpux:
@olafmandel increase pkgrel

olafmandel:
@alexpux: Done. Feel free to squash the two commits.

PR #6252, opened 2020-03-03
wxWidgets and wxWidgets3.1 changes

stahta01:
wxWidgets: Add wxrc-${_wx_basever}.exe
wxWidgets3.1 update to version 3.1.3 and other minor changes.

Tim S.

### DIFF
--- a/mingw-w64-wxWidgets3.1/PKGBUILD
+++ b/mingw-w64-wxWidgets3.1/PKGBUILD
@@ -177,11 +177,11 @@ build() {
   make #VERBOSE=1 -j1
 }
 
-check() {
-  cd "${srcdir}/build-${CARCH}-shared/tests"    && make -k --jobs=1 || true
+#check() {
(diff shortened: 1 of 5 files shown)


mingw-w64-wxWidgets3.1/PKGBUILD
```diff
@@ -91,6 +98,27 @@ build() {
...
+
+  local -a _debug_options=()
+
+  if check_option "debug" "y"; then
+    _debug_options+=("--enable-debug=yes")
+#   _debug_options+=("--enable-debug=max")
+    _debug_options+=("--enable-debug_gdb")
+  else
+    _debug_options+=("--enable-optimise")
+    _debug_options+=("--enable-debug=yes")
```
lazka: should this be "yes" in this case?

stahta01: That was the way the code was built in the past; I decided that should be the default.
I could change it to enable-debug=no if it makes more sense. But, the default with no option is 
enable-debug=yes. Tim S.

PR #6259, opened 2020-03-08
Remove lzo2 from libarchive dependencies

laozc:
Linking against lzo2 is not recommended and libarchive is not
currently built with lzo2.
Remove it from dependencies for less confusion.


lazka:
increase pkgrel

lazka:
I can confirm that it doesn't link against it. Thanks!

PR #6265, opened 2020-03-12
mypaint: Update to 2.0.0

lazka:


lazka:
@jplloyd it seems to work (except various SVG icons don't load), but I thought I'd better wait for your approval.

lazka:
Turns out latest librsvg got stricter re namespaces in svg files, reported to upstream in IRC.

lazka:
> <jplloyd> So I've heard. Anyway, the 2.0.0 upgrade looks correct, no objections to the merge (haven't actually tested it, I currently don't even have access to a windows VM).

PR #6266, opened 2020-03-12
mingw-w64-mesa: Get static linking right. Drop all runtime dependencies.

pal1000:
This makes Meson behave exactly like Scons did before replacement.
Drop shared build. I didn't see much value in it anyway, especially with
both linking negating all storage saving and the fact that there are no
packages in mingw-packages and msys2-packages repositories depending on
mingw-w64-mesa at this point. mingw-w64-mesa has been fine with static
linking since its debut anyway.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,28 +4,25 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=20.0.1
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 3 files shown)


mingwandroid:
> I didn't see much value in it anyway especially with both linking negating
all storage saving

Only if you think about it from a single package's perspective. Ecosystem wide the savings are multiplied. MSYS2 has a strong preference for dynamic libraries.

pal1000:
> 
> 
> > I didn't see much value in it anyway especially with both linking negating
> > all storage saving
> 
> Only if you think about it from a single package's perspective. Ecosystem wide the savings are multiplied. MSYS2 has a strong preference for dynamic libraries.

There are no packages in mingw-packages and msys2-packages repositories depending on
mingw-w64-mesa at this point.

PR #6269, opened 2020-03-14
Neovim

kassane:
https://github.com/neovim/neovim/wiki/Building-Neovim


Alexpux:
@kassane format PKGBUILD. See an our cmake based PKGBUIld

Alexpux:
@kassane we have some versions of lua. Lua 5.1 and luajit was conflict with each other and cannot be installed the same time. Also some packages use 5.1 and other jit.
Most lua modules packages was outdated and no one support it. 
I will not accept any lua packages this time because fist need to resolve all problems with current lua derivedd packages

Freed-Wu:
Is it possible to support neovim nowadays?

PR #6281, opened 2020-03-19
gitignore: Add unzipped and LZMA compressed tarballs

pal1000:


Biswa96:
Is it possible to add `.tar.lz` and `.tar.lzma` also? `wget` packages has it in source.

PR #6292, opened 2020-03-22
Update gedit and gedit-plugins, and simplify build instructions

swilmet:
Note, I’m the upstream maintainer of gedit.


lazka:
(I usually explicitly enable all build options so we don't lose things if upstream defaults change, or the option names change)

swilmet:
Thanks for the quick review!

PR #6296, opened 2020-03-22
appstream-glib: update list of dependencies

swilmet:
Note that libsoup is required, but was listed two times.

git grep in appstream-glib sources gives no results for gtksourceview,
iso-codes and peas.

libarchive was missing.


lazka:
(Another way to check wrong deps is to go to our package page and follow the link to the matching Arch Linux package, which usually should have similar dependencies: https://packages.msys2.org/base/mingw-w64-appstream-glib)

swilmet:
Thanks for the reviews and information! :+1:

PR #6304, opened 2020-03-23
New package: capnproto

Xanewok:
Adapted to the best of my knowledge, please tell me if I should improve anything.

Tested the build on bare MSYS2 install, it only needs `base-devel` and `mingw-w64-{arch}-toolchain` to build.
Running `pacman -U` on the resulting archive correctly installs the package and correctly installs the `capnp` binary.

The reason for this package is to provide the `capnp` binary, which allows to compile .capnp files.
This can be used in a build step, for example by https://github.com/capnproto/capnproto-rust. See https://gitlab.com/sequoia-pgp/sequoia/-/blob/master/store/build.rs for an example Rust library, which uses it.


Biswa96:
Some thoughts:

1. `(mingw-w64)` can be added in `pkgdesc` string to differentiate mingw packages with msys2 ones.
2. The configure script searches for `-lcrypto` and `-lssl` options. Can `--with-openssl` be added explicitly with openssl dependency?
3. There are some compiler warnings. One of them redefines `_WIN32_WINNT` constant to `0x0600` but mingw-w64 headers are built with `0x601`. Would this conflict somehow?

Xanewok:
> `(mingw-w64)` can be added in `pkgdesc` string to differentiate mingw packages with msys2 ones

Done in https://github.com/msys2/MINGW-packages/pull/6304/commits/c09173d98c255796a266460ed0ca7bdfc10a27c2.

> The configure script searches for `-lcrypto` and `-lssl` options. Can `--with-openssl` be added explicitly with openssl dependency?

Do you mean declaring openssl an optional dependency? Do you mind pointing me to such a configuration or showing a simple example? Thanks 🙇 

> There are some compiler warnings. One of them redefines `_WIN32_WINNT` constant to `0x0600` but mingw-w64 headers are built with `0x601`. Would this conflict somehow?

I don't *think* so, not sure though. If I had to take an educated guess, I'd say this declares compatibility with Vista-level Windows and imports relevant API version. I think it shouldn't break if mingw-w64 headers are built with a slightly higher version.

Maybe @kentonv knows more?

kentonv:
Hmm, I thought it was the application's responsibility to define `_WIN32_WINNT` to specify the minimum API version it needs. Why is MinGW trying to inject its own definition?

That said, if the code compiles without errors, most likely everything is fine.

mati865:
@kentonv MSYS2 supports Windows 7 and the headers reflect it.
Applications that wish keep control over should `undef` it first. It's harmless warning though.

kentonv:
I mean I'm pretty sure that in the past, I had to define this, otherwise the default was that I received _older_ APIs. I don't remember getting a double-define warning in the past. Did something change? Is my understanding incorrect?

mati865:
@kentonv there is a guard in the headers so you could define `_WIN32_WINNT` before including anything.

kentonv:
Oh! I thought we _were_ defining it before including anything (which made me pretty confused how it could be already defined). If not that is indeed a bug! Sounds like no big deal for now but I'll fix it in the next release.

Xanewok:
> Can `--with-openssl` be added explicitly with openssl dependency?

Done in https://github.com/msys2/MINGW-packages/pull/6304/commits/36004a209afcf89bb773da9e3e74ab72dad99da7, hopefully that's what you meant!

> Oh! I thought we *were* defining it before including anything (which made me pretty confused how it could be already defined). If not that is indeed a bug!

To be 101% sure about `_WIN32_WINNT` I created a [patch](https://github.com/msys2/MINGW-packages/pull/6304/commits/5ff28251133ffe555c6f640379c88005589fec68#diff-813a1ccb4c42e4438e1ebd3d0bc83fc4) (@kentonv) that defines it before including anything in https://github.com/msys2/MINGW-packages/pull/6304/commits/5ff28251133ffe555c6f640379c88005589fec68.

Does this look okay now? Does it need anything else to merge (and hopefully publish)?

Xanewok:
Thank you for merging!

Xanewok:
It looks like Publish artifacts job on CI [failed](https://dev.azure.com/msys2/mingw/_build/results?buildId=4720&view=logs&j=d025390b-1dd2-5eea-69df-6ceb0b9eeced&t=f241ee06-07c5-57f3-b3e7-1cc91e6837c9) (?). Is this something to worry about?
Did I misconfigure anything?

PR #6307, opened 2020-03-24
fix qt5-static for CMake 3.17

Optiligence:
CMake 3.17 stops finding dll files for additional libraries trying to be found by Qt
replace expanding the list with the workaround that will be in 5.14.2 anyway


Optiligence:
@Alexpux something holding up the repo update? 

I’m waiting for this to fix my CI…

Alexpux:
@Optiligence I will wait 5.14.2. Don’t have time to rebuild qt5 often

Alexpux:
Also we now on carantine so maybe package updates will be once in two - three weeks

Optiligence:
Sad to hear. That makes this PR also kinda pointless.

Optiligence:
it’s not even only qt5-static that’s broken, the code runs for dynamic qt too…
i.e. non already configured CMake builds of qt projects are broken atm

PR #6326, opened 2020-03-31
openssl: Update to 1.1.1f - fix for breaking change in 1.1.1e

MSP-Greg:
1.1.1f is a 'bug fix only release', and reverts a breaking change introduced in 1.1.1e involving error handling.

See:
https://mta.openssl.org/pipermail/openssl-announce/2020-March/000168.html

### DIFF
--- a/mingw-w64-openssl/PKGBUILD
+++ b/mingw-w64-openssl/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=openssl
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-_ver=1.1.1e
+_ver=1.1.1f
 # use a pacman compatible version scheme
 pkgver=${_ver/[a-z]/.${_ver//[0-9.]/}}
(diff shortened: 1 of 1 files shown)


MSP-Greg:
The breaking change was bad enough that the below was stated in a Fedora [message](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/O6L32KT4H2ZRK4E5JLZ2FFTVCJFYUGU6/):

> It shows that this does not only break Python's own test suite, but various Python packages are failing as well (incl. python-pip).

They released 1.1.1e, but added a patch that reverted the breaking change (which is what 1.1.1f does).

OpenSSL 1.1.1e also caused failures in Ruby's tests, and in various Ruby gems.

Hence, updating to 1.1.1f would be very helpful...

MSP-Greg:
I apologize, as I left the pkgrel value unchanged, but it was 2.  With a new OpenSSL version, I assume it would normally be reset to 2.

I can submit another PR...

lazka:
ah, thanks, changed on master

PR #6336, opened 2020-04-02
cglm: New package

podsvirov:


Biswa96:
Why different package functions for each architecture? And how about combining the doc with main package?

podsvirov:
> Why different package functions for each architecture?

Please teach me how to use fewer functions.

> And how about combining the doc with main package?

I don’t think this is a good idea. Some people don't need documentation at all. Moreover, the documentation now takes 10 times more disk space compared to the main package.

lazka:
We try to not split packages, so either merge the docs or remove them.

podsvirov:
> We try to not split packages, so either merge the docs or remove them.

Ok. I merged the docs..

Biswa96:
I saw this warning:

```
WARNING: html_static_path entry '_static' does not exist
writing additional pages...  search F:/msys64/mingw64/lib/python3.8/site-packages/sphinx_rtd_theme/search.html:21: RemovedInSphinx30Warning: To modify script_files in the theme is deprecated. Please insert a <script> tag directly in your theme instead.
```

Is this related to the packaging or just external?

podsvirov:
> Is this related to the packaging or just external?

It's not critical warnings from `sphinx-build` when building documentation. May be will fixed in later `cglm` version.

Cc: @recp

Issue #6342, opened 2020-04-04
Updating Out of Date MINGW-packages


hiroakit:
## Overview

In Apr. 1th, 2020 GMT,
[Out of Date MINGW-packages](https://packages.msys2.org/outofdate) is below:
- [x] mingw-w64-python-oslotest (#6364)
- [x] mingw-w64-gtk3 (#6352, a2f59cc, 62b98e4)
- [x] mingw-w64-mesa (#6349)
- [ ] mingw-w64-python-xpra
- [x] mingw-w64-pyqt5 (20f3f09)
- [x] mingw-w64-x265 (#6343)
- [ ] mingw-w64-python-httplib2
- [ ] mingw-w64-python-cryptography
- [ ] mingw-w64-python-django
- [ ] mingw-w64-python-zeroconf
- [ ] mingw-w64-python-pep517
- [x] mingw-w64-re2 (#6361)
- [x] mingw-w64-librsvg (#6345)
- [ ] mingw-w64-python-oslo-context
- [ ] mingw-w64-python-stestr
- [x] mingw-w64-python-zope.interface (15cb482)
- [ ] mingw-w64-python-hacking
- [x] mingw-w64-lib3mf (#6398)
- [ ] mingw-w64-poppler (#6344)
- [ ] mingw-w64-python-pyperclip
- [ ] mingw-w64-waf
- [ ] mingw-w64-python-alembic
- [x] mingw-w64-python-yaml (#6365)
- [ ] mingw-w64-python-prompt_toolkit
- [ ] mingw-w64-verilator
- [x] mingw-w64-python-cliff (#6356)
- [ ] mingw-w64-dnscrypt-proxy
- [ ] mingw-w64-soci
- [ ] mingw-w64-graphviz
- [ ] mingw-w64-farstream
- [ ] mingw-w64-perl
- [ ] mingw-w64-ghostscript
- [ ] mingw-w64-jbig2dec
- [ ] mingw-w64-libelf
- [ ] mingw-w64-cogl
- [ ] mingw-w64-git-lfs
- [ ] mingw-w64-totem-pl-parser
- [ ] mingw-w64-libgphoto2
- [ ] mingw-w64-libnotify
- [ ] mingw-w64-libgusb
- [ ] mingw-w64-python-psutil
- [ ] mingw-w64-python-pysocks
- [ ] mingw-w64-scummvm
- [ ] mingw-w64-pango
- [ ] mingw-w64-lsqlite3
[...]

PR #6352, opened 2020-04-05
gtk3: Update to 3.24.17

stahta01:
Problem is wrong version reported by pkg-config.

This is a quick fix; the more proper fix is to go back to configure/make build.

Tim S.

`$ pkg-config --modversion gtk+-3.0
3.24.14

stahta01@Laptop2-tim MINGW64 ~
$ pacman -Qs gtk3
local/mingw-w64-x86_64-gtk3 3.24.16-1
    GObject-based multi-platform GUI toolkit (v3) (mingw-w64)
`

### DIFF
--- a/mingw-w64-gtk3/PKGBUILD
+++ b/mingw-w64-gtk3/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=gtk3
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.24.16
+pkgver=3.24.17
 pkgrel=1
 pkgdesc="GObject-based multi-platform GUI toolkit (v3) (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
We should just update to a newer version where this is fixed.

stahta01:
Looks like that will work this time.

Tim S.

mingw-w64-gtk3/PKGBUILD
```diff
@@ -47,6 +47,8 @@ prepare() {
   patch -p1 -i "${srcdir}"/0003-gtkwindow-Don-t-force-enable-CSD-under-Windows.patch
 
   patch -p1 -i "${srcdir}"/0004-Disable-low-level-keyboard-hook.patch
+
+  #sed -s "s|version: '3.24.17',|version: '${pkgver}',|" -i "${srcdir}/gtk+-${pkgver}/meson.build"
```
lazka, line 51: Remove this

stahta01: No, I will not do that; it is the wrong thing to do! Tim S.

lazka: Um... ok, but why?

hiroakit:
Ref #6342

lazka:
I've updated things in master: a2f59cccc30b95

PR #6377, opened 2020-04-11
qpdf: update to 10.0.1

Biswa96:


Biswa96:
`0001-fix-printf-on-mingw-w64.patch` may not be required.

PR #6378, opened 2020-04-11
mingw-w64-mesa: Switch to static linking

pal1000:
With mingw-w64-llvm linked statically the storage save with mingw-w64-mesa
shared linking is almost non-existent.
Also there is no package depending on mingw-w64-mesa in MSYS2 so
shared linking storage save would be just theoretical anyway.

Other changes:
- use Meson builtin options to pass compiler and linker flags for CRT
static linking, symbols stripping and SSSE3 optimization respectively;
- update zlib wrap to 1.2.11-4.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=20.0.4
-pkgrel=1
+pkgrel=2
 pkgdesc="Open-source implementation of the OpenGL specification (mingw-w64)"
 arch=('any')
 makedepends=("${MINGW_PACKAGE_PREFIX}-llvm"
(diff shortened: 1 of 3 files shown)


lazka:
Just wondering, why can't you use the msys2 zlib and have to built your own?

pal1000:
I tried but it doesn't link statically and end up depending on mingw-w64-zlib. I'll take another look when I have the chance.

mingwandroid:
I do not support (not that that matters too much). Your reasons are not valid. That not much uses mesa DLLs (which I dispute) is not relevant here. You are switching mesa to use static libs instead of dynamic ones. That means doubling up the storage space for those libs for everyone who installs mesa.

Worse than that, every time a CVE appears in one of the packages mesa depends on, it becomes encumbrant upon Alexey to rebuild mesa too.

Static-only linking is not what MSYS2 does, for very good reasons.

pal1000:
@mingwandroid I'll switch to both linking everything as I have all the know how. The only part that I can't both link is mingw-w64-gcc-libs.

mingwandroid:
OK, I'm fine with providing both, so long as the 3rd party build script metadata does not try to force static on downstream consumers. Thanks!

PR #6381, opened 2020-04-13
ttf-cm-unicode: New package

podsvirov:


fabiangreffrath:
What are they needed for?

podsvirov:
For text rendering, of course.

fabiangreffrath:
I mean these specific fonts. There are gazilion of free fonts available out there, but why these?

podsvirov:
Short answer: I just ported this package from [aur](https://aur.archlinux.org/packages/ttf-cm-unicode).

podsvirov:
@lazka thanks!

PR #6387, opened 2020-04-15
ffmpeg: add mpd support.

sor593:


Alexpux:
@sor593 need increase pkgrel

sor593:
okay changed.

PR #6391, opened 2020-04-19
New package: bootloadhid

fauxpark:
https://www.obdev.at/products/vusb/bootloadhid.html

BootloadHID is a USB bootloader and accompanying flashing tool for Atmel AVR microcontrollers. It leverages the HID protocol to communicate, so there are no drivers or libraries necessary.

### DIFF
--- a/mingw-w64-bootloadhid/PKGBUILD
+++ b/mingw-w64-bootloadhid/PKGBUILD
@@ -1,0 +1,39 @@
+# Maintainer: fauxpark <fauxpark@gmail.com>
+
+_realname=bootloadhid
+_realver=2012-12-08
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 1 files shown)


mingw-w64-bootloadhid/PKGBUILD
```diff
@@ -0,0 +1,39 @@
+# Maintainer: fauxpark <fauxpark@gmail.com>
+
+_realname=bootloadhid
+_realver=2012-12-08
+pkgbase=mingw-w64-${_realname}
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+pkgver=$(echo ${_realver} | sed 's/-//g')
```
lazka, line 7: you can also do `${_realver//-/}` without echo/sed

fauxpark: Ah, thanks.

PR #6403, opened 2020-04-21
Upgrade libmariadbclient to 3.1.7

thefallentree:


thefallentree:
yeah this doesn’t work I will submit another soon

Biswa96:
It has been updated. See latest commit in this repository.

thefallentree:
ah, great! The previous release didn’t work for me because they were including an broken strtoll impl , I will check if this one works better.

PR #6407, opened 2020-04-22
Remove libcroco, no longer used

lazka:
librsvg no longer uses libcroco since 2.48. Since librsvg was the last user
of libcroco we can remove it.


Biswa96:
In ArchLinux, gettext uses it.

lazka:
I think gettext includes a copy of libcroco it can use

lazka:
And in Arch it's not actaally used it seems: https://bugs.archlinux.org/task/64416?project=1&string=gettext

Ede123:
FWIW there are some ideas being thrown around in the Inkscape project to take over maintenance of libcroco (we historically include a patched version of it within the source tree), however nothing actionable yet (and I guess we can always bring libcroco back, should Inkscape actually switch to an external dependency).

lazka:
> (and I guess we can always bring libcroco back, should Inkscape actually switch to an external dependency).

Sounds good to me

PR #6410, opened 2020-04-23
mingw-w64-mesa: Update to 20.0.5

pal1000:
Other changes:
- link LLVM statically in response to #6353 disabling LLVM shared linking;
- use Meson builtin options to pass compiler and linker flags for
symbols stripping and SSSE3 optimization respectively;
- made it possible to build mingw-w64-mesa when llvm config is known to be
broken.


Alexpux:
@pal1000 you need all llvm as build dependency then

pal1000:
LLVM was always a build dependency when building Mesa llvmpipe. The only difference is that I am linking only the static libraries so that LLVM is only a build dependency and not a runtime dependency, because LLVM is huge now.

PR #6413, opened 2020-04-24
Add rime-pinyin-simp, rime-wubi, rime-double-pinyin and rime-emoji

tumashu:


tumashu:
@Alexpux Thanks very much, at the moment, rime-pinyin-simp, rime-double-pinyin and rime-emoji have been in msys2 repo,   rime-wubi is not :-)

Issue #5601, opened 2019-07-11
wxSVG: Cannot Link Statically


AntumDeluge:
I am having trouble linking statically to [wxSVG](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-wxSVG) because of <del>three</del> (these) dependencies:
- libav
- libharfbuzz
- libexpat
- libpng

I can get by the ***libav*** dependency by disabling it in wxSVG's build configuration. But I don't know how to get past the others.

Linking with `-lwxsvg` ends up linked to the dll, even if `-static` is used. So I explicitly link to the static library with `-l:libwxsvg.a` & have come to the conclusion the aforementioned libraries are related to the problem. If I try to explicitly link to the static library of each, I get multiple definitions errors:

*libharfbuzz:*
```
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x1a4): undefined reference to `gr_make_face_with_ops'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x234): undefined reference to `gr_face_destroy'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x37d): undefined reference to `gr_face_featureval_for_lang'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x3a8): undefined reference to `gr_face_find_fref'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x3bc): undefined reference to `gr_fref_set_feature_value'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x49b): undefined reference to `gr_make_seg'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x4b3): undefined reference to `gr_seg_n_slots'
C:/Development/MSYS2/mingw64/lib\libharfbuzz.a(libharfbuzz_la-hb-graphite2.o):(.text+0x5f4): undefined reference to `gr_seg_first_slot'
[...]

PR #6419, opened 2020-04-26
wxWidgets update to 3.0.5 and use builtin libraries in Static library build

stahta01:
See [wxSVG: Cannot Link Statically](https://github.com/msys2/MINGW-packages/issues/5601) for reason for using builtin library. This does not fix that issue; but, should help with that issue.

Tim S.

### DIFF
--- a/mingw-w64-wxWidgets/PKGBUILD
+++ b/mingw-w64-wxWidgets/PKGBUILD
@@ -9,8 +9,8 @@ _wx_basever=3.0
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 provides=("${MINGW_PACKAGE_PREFIX}-wxmsw${_wx_basever}")
 pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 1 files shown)


stahta01:
Changing this PR because wxWidgets 3.0.5 was just released today

Alexpux:
@stahta01 we prefer shared linking and do static linking if it not breal shared linking only.

stahta01:
So, you prefer the wxWidgets static libraries use DLLs instead of using the builtin static support libraries? Tim S.

stahta01:
@Alexpux Do you really prefer Static Libraries to use MSys2 MinGW DLLs?
This seems very wrong to me.

I do understand that Shared Libraries should use MSys2 MinGW DLLs!

Tim S.

Alexpux:
@stahta1 sorry I’m wrong. I don’t look at PKGBUILD that it build wxwidget twice. Changes is ok

Issue #6394, opened 2020-04-20
cmake 3.17 find_library() command no longer finds .dll 


peterbud:
cmake 3.17 has removed .dll as a default extension for finding libraries as you can read in the [release notes](https://cmake.org/cmake/help/v3.17/release/3.17.html):
> When using MinGW tools, the find_library() command no longer finds .dll files by default. Instead, it expects .dll.a import libraries to be available.

That leads to situations when libraries linked against system libs like gdi32, shell32 and alike are not finding anymore the appropriate library during build. Previously `cmake `has found the system libraries like with the .dll extension:
C:/Windows/System32/gdi32.dll or 
C:/Windows/System32/shell32.dll

What is the right way to handle this situation:
* adding back the .dll extension with `set(CMAKE_FIND_LIBRARY_SUFFIXES .dll ${CMAKE_FIND_LIBRARY_SUFFIXES})`
* or convincing cmake to find the approriate import libraries under `/mingw64/x86_64-w64-mingw32/lib` like `mingw64/x86_64-w64-mingw32/lib/libgdi32.a` ?

Alexpux:
@peterbud for what reason need to find system libraries? I think system libraries must be add to link without finding via “-l” switch.

peterbud:
The context:
* While building darktable, it is linking to osmgpsmap library
* we use cmake's `pkg_check_modules `and pkgconf of osmgpsmap which contains `Requires: gtk+-3.0 libsoup-2.4`
* gtk3 pulls it's dependencies based on the pkg-config, which reaches gdk-3.0, where the pkgconf contains: `Libs: -L${libdir} -lgdk-3 -lz -lgdi32 -limm32 -lshell32 -lole32 -Wl,-luuid -lwinmm -ldwmapi -lsetupapi -lcfgmgr32` 

Before cmake 3.17 that was not a problem, as eventually the linker has found C:/Windows/System32/gdi32.dll etc, but after cmake 3.17 it's not happening anymore.

I'm sure there must be something very trivial here, just I don't know what.

revelator:
Is this also the cause for inkscape not finding the openmp library anymore ?.
While not a system library it would explain some current issues im having.

peterbud:
With the new cmake you can use `-DCMAKE_FIND_DEBUG_MODE=ON` to check how the pkgconfig and `find_library()` is working/progressing or failing to find a library.

[...]

PR #6424, opened 2020-04-27
darktable: Update to 3.0.2

kmilos:


kmilos:
Sorry, should've checked, this is pending #6394 obviously

PR #6431, opened 2020-04-29
wxWidgets3.1: Fix passing of C/CXXFLAGS

stahta01:
When static build use builtin support libraries.
Also, change to disable-permissive

Tim S.

### DIFF
--- a/mingw-w64-wxWidgets3.1/PKGBUILD
+++ b/mingw-w64-wxWidgets3.1/PKGBUILD
@@ -19,7 +19,7 @@ _wx_buildver=
 pkgbase=mingw-w64-${_basename}${_wx_basever}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}${_wx_basever}"
 pkgver=${_wx_basever}.3
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


Alexpux:
@stahta01 What problem with passing flags so you add them to configure call? Also what problem with CARCH using?

stahta01:
> @stahta01 What problem with passing flags so you add them to configure call? Also what problem with CARCH using?

I am trying to reduce false warning in building wxWidgets 3.1.
The last time I used CXXFLAGS and CFLAGS; but, I found out that did not work unless I passed them to each configure command. The original reason for setting CXXFLAGS and CFLAGS was that the configure/make complained about the setting of the -O and -g flags because the configure wished to set them as result of debug or release configure commands being set.  I am trying to track down a weird wxWidgets crash by using the debug flag some of the time; but, I think the normal release should not use the max debug mode. As the result of removing the -O and -g flags I thought I should back the march flag. After I added the passing of the CXXFLAGS and CFLAGS to each configure I found that the value of $CARCH was wrong for 64 bit builds.

Tim S.

stahta01:
I am working on a different PR that removes instead of fixing the local CXXFLAGS and CFLAGS.

Tim S.

PR #6435, opened 2020-04-30
gdcm: fix build with poppler 0.87

hubutui:
1. apply this https://github.com/malaterre/GDCM/commit/be7d49cf3d13eafffdba801ba30e1c98202843a8 to build with `poppler 0.87`
2. add `docbook-xsl` to `makedepends` for building manpages
3. set `GDCM_BUILD_APPLICATIONS=ON` to build gdcm applications


Biswa96:
Generally, a patch file is placed in the same PKGBUILD folder.

hubutui:
@Biswa96 This patch is from upstream. `makepkg` could download it when building. And this patch will be available in a new release. I don't think we need to add more files to the repo.

Alexpux:
@hubutui we prefer to have all patches used to build package to be uploaded to repo.

hubutui:
@Alexpux Ok, I see. Would you like to update the sources and checksum, and then commit the master? Or should I create another pull request?

Alexpux:
@hubutui, you need just commit patch to repo and change source line. You can do it in this pull request

hubutui:
@Alexpux Done.

PR #6437, opened 2020-05-01
add python-lzo

totaam:


mingw-w64-python-lzo/PKGBUILD
```diff
@@ -0,0 +1,43 @@
+# Maintainer: Antoine Martin <totaam@xpra.org>
+
+_realname=lzo
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
```
lazka, line 8: remove provides, conflicts, replaces-

mingw-w64-python-lzo/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+build() {
+  msg "Python build for ${CARCH}"
+  cd "${srcdir}/python-build-${CARCH}"
+  LZO_DIR="${MINGW_PREFIX}" ${MINGW_PREFIX}/bin/python setup.py build
+}
+
+package() {
+  cd "${srcdir}/python-build-${CARCH}"
+  LZO_DIR="${MINGW_PREFIX}" ${MINGW_PREFIX}/bin/python setup.py install --prefix=${MINGW_PREFIX}
+#    --root="${pkgdir}" --skip-build
```
lazka, line 40: what's this about?

totaam: Removed, I had copied it from the `python-lz4` build file and forgot to clean it up.

lazka: But you need to pass both things, without root the files don't end up in the package

totaam: Ah, right. Sorry. That's why I had left it there.
And the real reason it is commented out is because the build fails on my system when it is used:
```
running install
C:/msys64/mingw64/lib/python3.8/distutils/dist.py:274: UserWarning: Unknown distribution option: 'tests_require'
  warnings.warn(msg)
  File "setup.py", line 114, in <module>
    setup(**setup_args)
  File "C:/msys64/mingw64/lib/python3.8/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "C:/msys64/mingw64/lib/python3.8/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "C:/msys64/mingw64/lib/python3.8/distutils/dist.py", line 984, in run_command
    cmd_obj.ensure_finalized()
  File "C:/msys64/mingw64/lib/python3.8/distutils/cmd.py", line 107, in ensure_finalized
    self.finalize_options()
  File "C:/msys64/mingw64/lib/python3.8/distutils/command/install.py", line 361, in finalize_options
    self.change_roots('libbase', 'lib', 'purelib', 'platlib',
  File "C:/msys64/mingw64/lib/python3.8/distutils/command/install.py", line 528, in change_roots
    setattr(self, attr, change_root(self.root, getattr(self, attr)))
  File "C:/msys64/mingw64/lib/python3.8/distutils/util.py", line 170, in change_root
    raise DistutilsChangeRootError("root and pathname not on same drive (%s, %s)"
NameError: name 'DistutilsChangeRootError' is not defined
```

So, for whatever reason, it is not happy about the path. My build tree is on the `E:` drive, this normally does not cause any problems. It also works fine if I move the build tree to the `C:` drive.

As it attempts to report this error, it errors again because the mingw hack patch is faulty (but not a big deal since at that point it's going to end up in an error anyway):
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python2/0670-msys-convert_path-fix-and-root-hack.patch

So, the latest commit re-instates `-root=${pkgdir}` and allows the build to proceed... as long as one builds on the `C:` drive.

lazka: Uh, I see

PR #6440, opened 2020-05-02
fix @libgl@ in generated glew.pc

### LINKED ISSUE #6439: Glew 2.2.0 : invalid glew.pc
pkg-config reports @libgl@ as part of the libs to be linked which breaks compilation.

oxillo:
fix #6439


lazka:
Thanks

PR #6443, opened 2020-05-02
wxWidgets3.1: When static build use builtin support libraries.

stahta01:
Also, change to disable-permissive in configure.
And, remove the local C/CXXFLAGS; this did not work well enough for
production package.


stahta01:
This change is now too outdated.

PR #6444, opened 2020-05-03
ghex: Avoid showing up a command prompt in windows.

wsgalaxy:
Add positional argument "gui_app" for executable target so that it will not show up an annoying command prompt when launching the executable.


Alexpux:
@wsgalaxy  increase pkgrel

Issue #6449, opened 2020-05-05
itk needs rebuild


hubutui:
gdcm is updated to 3.0.5, so https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-itk/PKGBUILD#L47 should be updated too. Do we need to specify `GDCM_INCLUDE_DIRS`? I assume cmake could find `gdcm`.
There are conflict cmake configure at https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-itk/PKGBUILD#L65 and https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-itk/PKGBUILD#L72.

PR #6451, opened 2020-05-06
itk: update to 5.1.0 and fix build with gdcm

hubutui:
1. update itk to 5.1.0
2. set `ITK_USE_SYSTEM_GDCM=ON`

see also https://github.com/msys2/MINGW-packages/issues/6449


lazka:
Thanks!

PR #6453, opened 2020-05-07
gcc-git: Add patch for libgomp

lhmouse:
Note there are a lot of other printf attributes that might need fixing.
But as they don't cause builds to fail, I decide to leave them alone.

Reference: https://gcc.gnu.org/pipermail/gcc-help/2020-May/138869.html


Biswa96:
`pkgrel` needs to be increased by 1.

lhmouse:
> `pkgrel` needs to be increased by 1.

It is unnecessary as this package does not exist in **pacman** repositories.

lazka:
lgtm

PR #6455, opened 2020-05-07
New package:  skyr-url 1.7.5

yumetodo:
A C++ library that implements the [WhatWG URL specification](https://url.spec.whatwg.org/#url-class)

Based on https://aur.archlinux.org/packages/skyr-url-git/ , https://github.com/microsoft/vcpkg/blob/master/ports/skyr-url/portfile.cmake

# Dependency

#6452


yumetodo:
This PR is depend on #6452 so that mark this PR as Draft.

PR #6478, opened 2020-05-11
Try to fix rust for Vista

jeroen:
Hi, just testing this on the CI, if it will work!

### DIFF
--- a/mingw-w64-rust/0006-fix-for-vista.patch
+++ b/mingw-w64-rust/0006-fix-for-vista.patch
@@ -1,0 +1,44 @@
+From c0c1c362d1ab132849cd07c5ae08ff7666588e98 Mon Sep 17 00:00:00 2001
+From: Jeroen Ooms <jeroenooms@gmail.com>
+Date: Mon, 11 May 2020 11:13:00 +0200
+Subject: [PATCH] Do not use SetThreadErrorMode for Vista
+
+---
+ src/librustc_metadata/dynamic_lib.rs | 10 +++-------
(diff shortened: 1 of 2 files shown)


lazka:
Note that we don't support Vista any more.

jeroen:
Yeah I know :( I just want to test and then maybe use it locally.

jeroen:
I just need to build a version of librsvg that can run on Windows 2008. It can be a one-off thing.

lazka:
We have lots of older versions (also pre-rustification) here: http://repo.msys2.org/mingw/x86_64/

jeroen:
@lazka I know, but those didn't always work well, for example: https://github.com/ropensci/magick/issues/240

I tested it a bit and the rendering is much bettter with the latest rsvg. The only thing that prevents me from pushing this to our users is this stupid `SetThreadErrorMode` call...

jeroen:
Actually someone else just pointed out my diagnosis was wrong. The call to `SetThreadErrorMode` is not in librsvg but in `libgmodule-2.0.a`

jeroen:
So this is not actually needed, as it turns out rustc requries Windows 7 but the resulting binaries still work on Windows Vista. The problem I have is actually in glib2.

By the way, why doesn't the rust build succeed? I was getting the same build error locally. I don't think it's related to the patch.

mati865:
> By the way, why doesn't the rust build succeed?

Hard to tell without closer look but could be another GCC 10 victim.

PR #6485, opened 2020-05-14
tepl: remove vala from makedepends

swilmet:
A *.vapi file is not generated by the Tepl library. Only the GIR
(GObject Introspection) is generated, which Vala can also use.


swilmet:
Thanks for the quick review!

PR #6486, opened 2020-05-14
appstream-glib: remove gtk2 dependency

swilmet:
Not needed normally, there is already the gtk3 dependency.

The Arch Linux package does not have the gtk2 dep either, only the gtk3
dep:
https://www.archlinux.org/packages/extra/x86_64/appstream-glib/


swilmet:
Thanks for the quick review!

PR #6493, opened 2020-05-16
gtk3: Update to 3.24.20 and port to pacman hooks

lazka:
This means any package already depending on gtk3 can drop the
.install scripts updating the icon cache.

This depends on #6492


lazka:
@Alexpux fyi, the gobject-introspection part still fails _sometimes_. I haven't figured out why yet, but usually trying to build a second time works :/

PR #6503, opened 2020-05-19
harfbuzz: Port to meson

lazka:


hessenfarmer:
unfortunately this reverted the change to have separated static and dynamic builds which doean't allow to link harfbuzz statically anymore.

lazka:
oh, I assumed this was fixed. I'll revert that part.

Issue #6122, opened 2020-01-10
otter-browser


spl1inter:
the name says it all.  I am a) recommending formal recognition b) recognizing the importance of a completely ABI-independent c) saluting the spirit of truly open web standard on windows d) drinking lots of good aussie cabsav you suckers and therefore RECOMMENDING to the world the first truly ab-independent  html5 300+ acid3 100 compliant I care about my digital privacy browser: otter on mingw64

OK someone WAY better than me should be putting this up right now guys im pretty much a side show.

QT. Otter. F bash nerds (Alexpux i already like you and hate you how is that possible) pleasde lkets do something genuinely significant together.

Otter+MinGW = future on the web WHY NOT

Alexpux:
@spl1inter Im upload rebuilded qtwebkit. So you can try to add package

stahta01:
@spl1inter: I have submitted an PR on this. #6517 

Tim S.

stahta01:
@spl1inter and others:
Do you think this package should be a git package or one that build the latest official release?
I am starting to think a git master branch package would be better while this software is this much like an beta  level project.

Tim S.

asctime:
Sorry I didn't see this.. Update: the latest Otter-browser is very good, IMO. YouTube and many Javascript-heavy sites work almost as well as in Firefox. I am really surprised. But to really make it work we need to update qtwebkit PKGBUILD:

```
-G"MinGW Makefiles" \
-DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
-DCMAKE_MAKE_PROGARME=mingw32-make \
-DCMAKE_BUILD_TYPE=Release \
-DPORT=Qt \
-DUSE_QT_MULTIMEDIA=ON \
-DSEPARATE_DEBUG_INFO=ON \
-DENABLE_CREDENTIAL_STORAGE=ON \
-DENABLE_CSS3_TEXT=ON \
-DENABLE_CSS3_TEXT_LINE_BREAK=ON \
-DENABLE_CSS_BOX_DECORATION_BREAK=ON \
-DENABLE_CSS3_BOX_DECORATION_BREAK=ON \
-DENABLE_CSS_IMAGE_ORIENTATION=ON \
-DENABLE_CSS_IMAGE_RESOLUTION=ON \
-DENABLE_ENCRYPTED_MEDIA=ON \
-DENABLE_INPUT_TYPE_DATE=ON \
[...]

PR #6517, opened 2020-05-24
[HOLD] otter-browser: New package

stahta01:
The browser works well enough to show me web pages; and, I logged into two web sites.
There does seem to be bugs; but, they match the ones I saw from the Windows download version.
I only have my name as Contributor because other than fixing build errors I can not fix run-time errors.
Is keeping the package able to be built enough to be a Maintainer?

See #6122

Tim S.


stahta01:
I am thinking that the version added is not the newest version.
So, marked as WIP till I can confirm.

stahta01:
I have decide to update the package to using the git master repo instead of the latest release.

Tim S.

stahta01:
I have gotten the project to be as good as I can make it.
But, I would say keep it as a source only package with no binary build till the project is more stable.

Tim S.

elieux:
Then maybe let's place it into mingw-packages-dev.

stahta01:
Decided to add it to mingw-packages-dev repo, instead.

lazka:
Closing in favor of https://github.com/msys2/MINGW-packages-dev/pull/10

PR #6521, opened 2020-05-25
gr: add

kou:


elieux:
I'd rather download the tarball and extract it manually (see `noextract` and `prepare`) than to bring in the Git artillery.

kou:
OK.
I've changed to use `tar.gz` instead of Git repository.

mingw-w64-gr/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+         "${MINGW_PACKAGE_PREFIX}-libpng"
+         "${MINGW_PACKAGE_PREFIX}-libtiff"
+         "${MINGW_PACKAGE_PREFIX}-qhull-git"
+         "${MINGW_PACKAGE_PREFIX}-qt5"
+         "${MINGW_PACKAGE_PREFIX}-zlib")
+makedepends=("git"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-gcc")
+options=("staticlibs" "strip" "!buildflags")
+source=("https://github.com/sciapp/gr/archive/v${pkgver}.tar.gz"
```
Alexpux, line 27: Download in named archive

mingw-w64-gr/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+      ../${_realname}-${pkgver} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+      -G "MSYS Makefiles"
+  make
+}
+
+package() {
+  cd ${srcdir}/build-${CARCH}
+  make DESTDIR="${pkgdir}" install
```
Alexpux, line 59: If sources have license file then install it. See other our packages

mingw-w64-gr/PKGBUILD
kou: I've changed to use `"${_realname}-${pkgver}.tar.gz::https://...`.

mingw-w64-gr/PKGBUILD
kou: Done.

I'll send a pull request to install license file by `make install` to GR later.

PR #6523, opened 2020-05-26
Fix Boost::locale compilation to use ICU backend

### LINKED ISSUE #6183: Boost.locale for 1.72.0-1 package is built without ICU support
As a proof, libboost_locale-mt.dll form the mingw-w64-x86_64-boost-1.72.0-1 package does not depend on icu dlls.

It breaks applications dependent on boundary analysis -- non-ICU backends [does not support it](https://www.boost.org/doc/libs/1_48_0/libs/locale/doc/html/using_localization_backends.html). 

Workaround: downgrade to mingw-w64-x86_64-boost-1.71.0-3.

indrekis:
Fixes #6183. 

Looks like boost::locale and boost::regex ICU configurations are in conflict. So, fix to compile  boost::regex with ICU ([99ffbce](https://github.com/msys2/MINGW-packages/commit/99ffbce6bffe1eed97a9429368ad286a8e717ad9)) broke boost::locale usage of ICU. 

Build variables ICU_ICUUC_NAME, ICU_ICUDT_NAME and ICU_ICUIN_NAME are honored by boost::regex but ignored by boost::locale (see libs/regex/build/Jamfile.v2 and libs/locale/build/Jamfile.v2). ICU_LINK is used by boost::locale but setting it leads to error in boost::regex. 

Though more general fix would be to modify boost::locale Jamfile, working with boost build is a daunting task (at last for me). So I have used minimal approach -- disabled error on defined	 ICU_LINK in boost::regex (and checked that this variable is really unused by the library) and re-added ICU_LINK variable for boost::locale.

### DIFF
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.73.0
 _boostver=${pkgver//./_}
-pkgrel=3
+pkgrel=4
 pkgdesc="Free peer-reviewed portable C++ source libraries (mingw-w64)"
 arch=('any')
 url="https://www.boost.org/"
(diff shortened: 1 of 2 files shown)


Alexpux:
@indrekis  increase pkgrel

indrekis:
> @indrekis increase pkgrel

I have done this. One of the tests failed with the error which looks completely unrelated to my change: "Cannot find the /usr/bin/zstd binary required for makepkg". If I have done something wrong please tell me how I can fix it.

PR #6526, opened 2020-05-27
binutils: Use xz instead of bz2

orgads:
20M instead of 30M.

I didn't increase pkgrel because it doesn't change anything in the products. Next build will benefit from it.


Biswa96:
Good idea. This may be extended to other packages also. I would like to hear from maintainers.

PR #6533, opened 2020-05-29
mingw-w64-mesa: Don't binary wrap and autodetect runtime dependency at the same time when linking LLVM

pal1000:
Use runtime dependency autodetection as it's working properly now.

Dependencies wrappping precedence can change even between point releases
in Meson, as seen in 0.54.1 when Meson was using binary wraps and now in
0.54.2 it used runtime dependency autodetection and the way build is
configured hasn't changed since then.

Because both runtime dependency autodetection and binary wrap methods link
the gigantic mingw-w64-llvm statically and didn't add it as dependency,
we don't need to rebuild mesa.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=20.1.0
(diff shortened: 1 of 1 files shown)


pal1000:
> 
> 
> bump pkgrel

Done. Only source package needs updating. Binaries package can stay at 20.1.0-1 because it's technically unchanged. That's also why I hesitated about bumping pkgrel.

lazka:
ok, that wasn't clear, thanks

PR #6534, opened 2020-05-29
Few fixes for clang package

mati865:
Tested x86_64 only.


Alexpux:
@mati865 not i get configure error for 32bit:
````
CMake Error at projects/libunwind/src/CMakeLists.txt:86 (message):
  Compiler doesn't support generation of unwind tables if exception support
  is disabled.  Building libunwind DSO with runtime dependency on C++ ABI
  library is not supported.
````

mati865:
I have no idea why it only happens for 32 bit package but https://github.com/msys2/MINGW-packages/pull/6548 fixes the build for me.

PR #6541, opened 2020-05-31
opencollada: Remove huge pointless patch

orgads:
7.7M patch for indentation?

### DIFF
--- a/mingw-w64-opencollada/PKGBUILD
+++ b/mingw-w64-opencollada/PKGBUILD
@@ -23,8 +23,7 @@ source=("${_realname}-${pkgver}.tar.gz"::"https://github.com/KhronosGroup/OpenCO
         opencollada-cmake.patch
         opencollada-disable-test-mingw.patch
         use-cpp-headers.patch
-        move-hash-function.patch
-        fix-identation.patch)
+        move-hash-function.patch)
 sha256sums=('d9db0c0a518aa6ac0359626f222707c6ca1b63a83cbf229d97a5999c9cde347b'
             '10fdc2746b83161b853df6ad4ac600252c4f964377ea0186ff97f99bd3417e34'
             '18f1bf9f57d1c67aa35927fb0d1b0b38def52b189742021f1268eb2bdbffc963'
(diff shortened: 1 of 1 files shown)


lazka:
@Alexpux do you remember why you added that?

Alexpux:
@orgads @lazka this patch solving GCC warnings about identations and guards so I will not allow to drop it

elieux:
Could we instead add something like `-Wno-indentation`?

orgads:
> Could we instead add something like `-Wno-indentation`?

That's exactly what I'm testing now.

Alexpux:
What problem with this patch? Devs write dirty code so you want to fix it by flags?

elieux:
Wow, this build is riddled with warnings; lots of case fallthroughs.

Alexpux:
as i tell i will not drop it

elieux:
@Alexpux, does the patch fix the code or just remove some warnings?

Alexpux:
@elieux it remove warnings, but also can fix some errors becasue of missinterpreting code by compiler

orgads:
I fail to see the value of having a 7M patch which effectively rewrites most of the application.

It's not like the patch actually fixes any behavioral issue. It merely keeps GCC happy. For that, we could either suppress `-Wmisleading-indentation`, or even suppress all compiler warnings.

I successfully built it without the patch without any errors (but there *are* a lot of warnings, regardless of the patch).

elieux:
Assuming it's *only* indentation and no code changes, I don't think the patch helps us in any way. The patch would make sense if gets into upstream, but we are only using the source files to build a package and then the fixed sources are essentially thrown away. *And* it forces more work on us when we upgrade to the next release of opencollada.

Alexpux:
There are no commits to it since 2018

orgads:
I just pushed a revised commit that disables the warnings.

I verified that the 7M patch does nothing but indentation changes (`git diff -w` shows nothing).

Please consider reopening.

orgads:
https://github.com/orgads/MINGW-packages/commit/37b608da83708a2b0fbb0baa965e111295ed7573

Alexpux:
cant reopen

orgads:
Created #6542.

PR #6552, opened 2020-06-03
FFmpeg: Add dav1d

1480c1:
Adds `libdav1d` as decoder option for av1


mati865:
Reason of the failure: https://github.com/msys2/MINGW-packages/issues/6553

PR #6572, opened 2020-06-07
Move gcc-git to master branch and build libgccjit package.

nicber:
Hi,

I updated the gcc-git package to build the libgccjit package.

Please give me your opinions.

Nico.


Biswa96:
How about building gccjit library as a separate package liek this [libgccjit AUR](https://aur.archlinux.org/packages/libgccjit/).

Biswa96:
How about building gccjit library as a separate package like this [libgccjit AUR](https://aur.archlinux.org/packages/libgccjit/)?

nicber:
AFAICS, that libgccjit package relies on the system GCC. It just builds the jit part.

You have given me an idea: build the minimum required for gccjit and install it in a separate directory. That way it should not interfere with the installed GCC.

lazka:
@lhmouse any thoughts on this?

lhmouse:
I don't have specific objection on this. libgccgit is really something nice to have.

lazka:
We have moved gcc-git to https://github.com/msys2/MINGW-packages-dev, so either this needs to be moved there, or ported to gcc.

lazka:
I've opened #7345, but it complains that libgccjit should be built separately.

Anyway, closing this, let's continue in #7345

lazka:
fyi, #7345 is now ready

PR #6574, opened 2020-06-07
[RFC] add GitHub Actions workflow(s)

eine:
This PR is currently a placeholder to discuss using GitHub Actions workflows in this and/or other repositories of the org. After reading [Off-load package building and uploading](https://github.com/msys2/msys2.github.io/blob/source/docs/wiki/Devtopics.md#off-load-package-building-and-uploading) I'd like to note the following points:

- Execution time limit for each job in GitHub Actions is 6h (unlike AppVeyor or Travis, where it is 50-90 min).
- GitHub Actions allows to execute up to 20 workflows/jobs at the same time.
- GitHub Actions are not limited to "plain CI testing", but can have tight integration with GitHub's API (both for triggering workflows and for interacting with it from within the workflows). Default Actions are written in JavaScript/TypeScript, but any language which can be packaged in a Docker container is suitable for jobs run on Linux.
- It is possible to cross-trigger workflows, within the same repo or between repos.

---

The first proposal in this PR is 'build'. It is an scheduled (CRON) job that is executed every 6 hours. The idea is to retrieve lists of pending tasks from [packages.msys2.org/new](https://packages.msys2.org/new) or [packages.msys2.org/queue](https://packages.msys2.org/queue) or [packages.msys2.org/outofdate](https://packages.msys2.org/outofdate). After each task is successfully completed, resulting packages can be pushed to an stagging area. I assume that 'Stagging' would be a new section in packages.msys2.org. Regarding where to store the packages/artifacts/assets for stagging, an alternative to using some external service is to use a GitHub pre-release. It can be done with [eine/tip](https://github.com/eine/tip).

Hopefully, such an approach would allow maintainers to focus on downloading and testing pre-built packages, instead of them building the packages manually.

Regarding packages that take longer than 6h, or other which cannot be built for any other reason, I think that the workflow can have exceptions. Those packages will be kept in the New Queue, Outdated or Update Queue. Alternatively, GitHub Action's caching features can be used to build a sequence of jobs.

Hence, although this proposal is not a complete solution yet, I believe it can help reduce most of the manual work.

---

The second proposal is to have a 'PR' workflow. As the name suggests, this is to test specific PRs and it is similar current Azure or AppVeyor services. In fact, I believe that GitHub Actions workflows are executed on Azure under-the-hood. Updating 'Stagging' packages from PR workflows is an optional feature. If the previous workflow is added, I believe that PRs should not push packages to stagging. However, it is also possible to only add this workflow.

---

Overall, I'd like to ask whether scripts exist already which can be used in place of the `echo "PLACEHOLDER"` commands in this PR.

---

[eine/setup-msys2](https://github.com/eine/setup-msys2) is used to set up MSYS2 in the workflows. See [msys2/msys2-installer#5](https://github.com/msys2/msys2-installer/issues/5#issuecomment-635495897).

/cc @Alexpux @elieux @lazka @Diablo-D3

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,43 @@
+name: 'build'
+
+on:
+  push:
+  schedule:
+    - cron: '0 0/4 * * *'
+
+jobs:
+
+  build:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - uses: msys2/setup-msys2@v1
+      with:
+        msystem: MSYS
+        update: true
+        cache: true
+        install: base-devel git python python-pip mingw-w64-x86_64-toolchain mingw-w64-i686-toolchain
+
+    - name: Process build queue
+      run: |
+        mkdir artifacts
+        python -m pip install pytest pytest-timeout
+        python -m pytest -v -s -ra buildqueue.py --timeout=18000
+
+    - uses: actions/upload-artifact@v2
+      with:
+        path: artifacts/*
+
+    - name: Upload packages to stagging
+      run: echo "PLACEHOLDER"
+      #uses: eine/tip@master
+      #with:
+      #  token: ${{ secrets.GITHUB_TOKEN }}
+      #  tag: 'stagging'
+      #  rm: false
+      #  files: *.pkg.zst
--- a/buildqueue.py
+++ b/buildqueue.py
@@ -1,0 +1,37 @@
+from urllib.request import urlopen
+from json import loads
+from pathlib import Path
+from subprocess import check_call
+from sys import stdout
+
+PKGEXT='.pkg.tar.zst'
+SRCEXT='.src.tar.gz'
+
+ROOT = Path(__file__).resolve().parent
+
+def test_file(pkg):
+    print('\n::group::Package %s...' % pkg)
+    stdout.flush()
+
+    def run_cmd(args):
+        check_call(['bash'] + args, cwd=str(ROOT/pkg))
+
+    try:
+        run_cmd(['makepkg-mingw', '--noconfirm', '--noprogressbar', '--skippgpcheck', '--nocheck', '--syncdeps', '--rmdeps', '--cleanbuild'])
+        run_cmd(['makepkg', '--noconfirm', '--noprogressbar', '--skippgpcheck', '--allsource', '--config', '/etc/makepkg_mingw64.conf'])
+        run_cmd(['mv', '*%s' % PKGEXT, '*%s' % SRCEXT, '../artifacts/'])
+        print('::endgroup::')
+        stdout.flush()
+        return 0
+    except:
+        print('::endgroup::')
+        stdout.flush()
+        return 1
+
+def pytest_generate_tests(metafunc):
+    pkgs = [
+        pkg['repo_path']
+        for pkg in loads(urlopen("https://packages.msys2.org/api/buildqueue").read())
+        if pkg['repo_url'] == 'https://github.com/msys2/MINGW-packages'
+    ]
+    metafunc.parametrize("pkg", pkgs)


mati865:
On Azure Pipelines which are internally the as GitHub Actions many bigger packages like Qt, LLVM, GCC fail to build because there is no enough space. So you cannot ever build a most important packages like the compiler.

eine:
@mati865, there are currently `78 + 19 + 251 = 348` packages in the queues. How many of them would you say that are impossible to handle because of space limits?

Apart from, that, as said, "*although this proposal is not a complete solution yet, I believe it can help reduce most of the manual work*". So, what do you think about the proposal(s) overall?

mati865:
> How many of them would you say that are impossible to handle because of space limits?

Not many.

> So, what do you think about the proposal(s) overall?

Packages are currently built on the machine where almost all packages are installed so if you build the same packages on the CI they will have less features enabled (since build system didn't found certain libraries). While this is good from the point of reproducibility it would break various cases if those packages were promoted from staging to main repo.
Another thing is many packages have incomplete dependencies so they would fail to build on the CI. Although that would be perfect opportunity to detect them and fix it.
Lastly, I wanted for a long time to create such automation but build packages with Clang+LLVM instead of GCC+Binutils for the sake of testing Clang so I really like the overall idea.

lazka:
> The idea is to retrieve lists of pending tasks from [packages.msys2.org/new](https://packages.msys2.org/new) or [packages.msys2.org/queue](https://packages.msys2.org/queue)

I can provide an API for this (they also need to be ordered re their dependencies), though I don't know how to tackle the other issues. One step at a time I guess.

eine:
@lazka, I'm (gratefully) surprised about https://github.com/msys2/MINGW-packages/blob/master/.github/workflows/main.yml. I spent a few days trying to achieve it without success.

First, https://github.com/eine/MINGW-packages/commit/75a952f7ae47cbf7a0f7f3e810b27ace4fecccca was required, otherwise https://github.com/eine/MINGW-packages/runs/752477144?check_suite_focus=true#step:4:4140. However, it seems to be unchanged here: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-filezilla/PKGBUILD#L6.

Then, `ci-build.sh` would hang, trying to process all the changes, due to how (auto)crlf I think.

Did you have to work around any of these issues? Is/was it because `MINGW_INSTALLS` needs to be set?

> I can provide an API for this (they also need to be ordered re their dependencies), though I don't know how to tackle the other issues. One step at a time I guess.

Agree. We can start with just being able to retrieve the list of pending tasks. Let me know when I can try it and I'll update this PR.

NOTE: in GHA it is possible to define collapsible blogs/groups in bash (as you already saw in JS). The syntax is:

```sh
printf '::group::${ANSI_COLOR}Text to be shown${ANSI_NOCOLOR}\n'
...
...
...
echo '::endgroup::'
````

Assuming that `ANSI_COLOR` and `ANSI_NOCOLOR` are defined. See https://github.com/ghdl/docker/blob/master/utils.sh and https://github.com/ghdl/docker/runs/809455786?check_suite_focus=true#step:4:68.

lazka:
> Did you have to work around any of these issues? Is/was it because `MINGW_INSTALLS` needs to be set?

I'd guess because of `fetch-depth: 0`. Our CI requires a full clone so it can see the differences compared to master.

> Agree. We can start with just being able to retrieve the list of pending tasks. Let me know when I can try it and I'll update this PR.

OK, I'll look into it.

elieux:
Diablo reminded me that we can also use our server for more than just file hosting, so if it's possible to do those large builds on Linux somehow, that could help.

Regarding the side effects of CI mentioned by Mati, I consider those definite advantages and even goals for setting up the build CI. To find mistakes before we publish, I suggest (as mentioned in some other thread) comparing the current and new package in important aspects, like the list of files contained, the exports and imports for each PE, and manually reviewing before publishing.

mati865:
> if it's possible to do those large builds on Linux somehow, that could help. 

I don't know if anything has changed in regards of Cygwin vs Wine issue but it's tracked here: https://github.com/msys2/MSYS2-packages/issues/682

eine:
I updated this PR with a draft that retrieves a JSON from https://packages.msys2.org/api/buildqueue. The script (`buildqueue.py`) is written in Python. It filters MINGW-packages from the JSON, then some specific packages are removed (those that take too long), and each of them is built. I borrowed the build commands from `ci-build.sh`. Each package is executed as a pytest test, to provide nice PASSED/FAILED results. Moreover, the log of each package is collapsed in a group.

Unfortunately, most packages fail at the end with `==> ERROR: Cannot find the strip binary required for object file stripping.` (https://github.com/eine/MINGW-packages/runs/865686828?check_suite_focus=true#step:4:623).

eine:
The content of this PR was moved to branch `gha` of repo msys2-devtools: https://github.com/msys2/msys2-devtools/tree/gha

PR #6581, opened 2020-06-11
Enable Windows hypervisor support - qemu

ZachBacon:
this should be now available within current mingw-w64 headers for use with the qemu package for those that don't want to use hax and have hyperv within their install of windows 10.


Biswa96:
Windows Hyper-V Platform libraries are only available for X86_64 only. I am not familiar with how qemu build works but should there be a check to detect X86_64 platform only e.g. `$MSYSTEM_CARCH == x86_64` in PKGBUILD?

mati865:
I'm wondering how i686 build succeed since indeed mingw-w64 provides headers and libs only for x86_64.

Biswa96:
Maybe its due to qemu uses LoadLibrary instead of linking to the export library 🤷 Found in qemu/target/i386/whpx-all.c `load_whp_dispatch_fns` function.

mati865:
`LoadLibrary` would be actually fine since it loads library from the system not mingw-64.

This is interesting:
```
$ pacman -Qo /mingw32/i686-w64-mingw32/include/winhvplatform.h
/mingw32/i686-w64-mingw32/include/winhvplatform.h is owned by mingw-w64-i686-headers-git 8.0.0.5905.066f1b3c-1
```

ZachBacon:
Well, what we could do is just enable it for 64bit if there's issues for the 32bit edition.

ZachBacon:
Any thoughts about keeping it available for 32bit or not? I rather not sit on this much longer as I have a few more changes regarding the kde5 stuff that I would like to submit a PR for

ZachBacon:
just a bump

lazka:
The question is, does it break 32bit in any way?

ZachBacon:
@lazka so while it does build, looks like I'll have to keep it 64bit since it can't load a 64bit library, I'll get that fixed up.

mati865:
It should not break 32 bit.
Qemu uses dynamic dispatch to load Hyper-V related libraries so it doesn't need any mingw-w64 libraries that provide import symbols.

ZachBacon:
> It should not break 32 bit.
> Qemu uses dynamic dispatch to load Hyper-V related libraries so it doesn't need any mingw-w64 libraries that provide import symbols.

The issue comes from the fact it's a 32bit application trying to load a 64bit dll. I've compiled it and tested it with the -accel whpx and it can't load the dll due to the differences.

Biswa96:
One query. You set `_whpx` variable but added `_mach` variable in configure command?

ZachBacon:
Whoops, I didn't pay attention, good catch, I'll fix that later.

lazka:
ping

ZachBacon:
> ping

Apologies, started a new job and didn't have time. I'm currently at work but I'll push the new changes when I get home

ZachBacon:
@lazka push made.

lazka:
meh, github got confused, I'll push this manually

PR #6586, opened 2020-06-13
[WIP] gcc-git: Fix building using GCC plugin

stahta01:
Add 0022-gcc-plugin-ada-use-def.patch and gnat1.exe.def

The gnat1.exe.def needs work; it was based on 9.3.1 exports and then I removed a lot of them.
Likely need a lot more removed and maybe a few added.

Tim S.


eine:
@stahta01, would you mind rebasing/force-pushing to have CI jobs triggered again?

stahta01:
> @stahta01, would you mind rebasing/force-pushing to have CI jobs triggered again?

Okay, sorry about late response other things detracted me.

eine:
No worries at all! Having GHDL contributed is already being a long trip 😄 

It seems that AppVeyor is failing because of a timeout. However, GitHub Actions allows up to 6h. I believe Azure too. Do you expect execution time to be above 6h?

lazka:
We have moved gcc-git to https://github.com/msys2/MINGW-packages-dev, so either this needs to be moved there, or ported to gcc proper.

stahta01:
Will work to move changes to dev repo

PR #6590, opened 2020-06-15
Update aom and libheif

lazka:
aom now provides shared libs and libheif links against them.

I also switch cmake to ninja, seems to work fine.


lazka:
I've filed the libaom issue upstream: https://bugs.chromium.org/p/aomedia/issues/detail?id=2727

PR #6596, opened 2020-06-16
ffmpeg: Update to 4.3

lazka:
move aom to depends because we link dynamically now


ZachBacon:
Since this release has vulkan support, would it be prudent to include the vulkan package as a compile time dependency? Or do we know if ffmpeg does it's vulkan support as a runtime load?

lazka:
See #6598 Not sure how to test though.

PR #6600, opened 2020-06-17
Draft: GHC 9.14.1

ofry:
For CI testing and make bug reports to GHC issue queue.


lazka:
What's the status of this?

ofry:
https://gitlab.haskell.org/ghc/ghc/-/issues/16780 prevented this compilation on 8.10.1. Need tests on 8.10.3.

ofry:
Still no luck in 8.10.3. Updated related issue on GHC issue tracker.

lazka:
Can we try backporting https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3880.patch ?

ofry:
Compiler/GHC/SysTools directory doesn't packaged in GHC releases, at least, in 8.10.x

lazka:
ok, since it's not clear if upstream will backport it, maybe ghc-9.0.1-rc1 would work?

(I don't have experience with ghc, so not sure if a major version bump would reduce its usefulness)

ofry:
GHC 9.0.1-rc1 failing with same error.

ofry:
Trying to build 9.0.1 with same patches.

bgamari:
I'll admit that I'm a bit worried about what will happen if users try to use this given that it does not use GHC's patched `gcc`. Specifically, we [patch](https://gitlab.haskell.org/ghc/win32-tarballs) GCC to shim various libc interfaces with implementations capable of working with long file paths (which the `msvcrt` libc implementation used by `gcc` currently can not handle).

This is necessary in the case of GHC since many build tools produce rather long file paths due to the "unit ID" (essentially a hash of a package's ABI) being included in various file names. Without long file path support things break quite quickly, in my experience.

mingwandroid:
We probably want this kind of stuff for our gcc!

bgamari:
> We probably want this kind of stuff for our gcc!

Yes, it has made the toolchain much more solid in our experience. However, the implementation is admittedly a bit hacky. Ideally someone would try to fix this properly upstream

hsyl20:
@ofry regarding the windres issue, https://gitlab.haskell.org/ghc/ghc/-/issues/2585 seems related

ofry:
Trying to check is windres problem related to https://github.com/msys2/MSYS2-packages/issues/2379

mingw-w64-ghc/0009-nativeGen-Don-t-assume-small-code-model.patch
```diff
@@ -0,0 +1,44 @@
+From c4b75c997ec8e7d5842d65cd75d9e2af9c8aa8b3 Mon Sep 17 00:00:00 2001
+From: GHC GitLab CI <ghc-ci@gitlab-haskell.org>
```
lazka, line 2: nice :)

ofry: I just pinged GHC devs and testing their efforts.

lazka:
https://gitlab.haskell.org/ghc/ghc/-/commit/bb690e971b02e4ae7a21a755b7104699e264fe61 maybe?

ofry:
@lazka Let's try your suggestion.

PR #6603, opened 2020-06-17
sip5: Update path in *-script.py

ustsv:


Alexpux:
@ustsv increase pkgrel

Biswa96:
One question. Aren't the install scripts be replaced with pacman hooks?

Alexpux:
@Biswa96 both are works

mati865:
Hooks are preferred but there is no strong deprecation yet.

PR #6610, opened 2020-06-19
MLT (An open source multimedia framework)

ice0:
New package:

MLT Multimedia Framework (https://www.mltframework.org)


ice0:
Hi, guys!
Is there anything else required from me? Maybe need additional description?

PR #6641, opened 2020-06-30
Update Corrade and Magnum to 2020.06.

williamjcm:
Notes:

- `mingw-w64-corrade`: the patch for 2019.10 was merged upstream, making it obsolete.
- `mingw-w64-magnum`: I had to force the path to `libvulkan`, because the official CMake `FindVulkan` module didn't work.
- `mingw-w64-magnum-plugins`: Basis Universal and meshoptimizer have to be bundled for now, according to mosra. Also, unlike the Arch Linux version of `magnum-plugins`, the Basis converter plugin can't be built on Windows due to an error in Basis Universal.
- `mingw-w64-magnum-integration`: like with Basis and meshoptimizer for `magnum-plugins`, Dear ImGui has to be bundled for the integration to work.


Biswa96:
Just a question. Is it possible to use the GitHub provided .tar.gz asset instead of cloning the repository in source?

williamjcm:
As stated in the PKGBUILDs:

```text
# The .tar.gz / .zip download contains symlinks (.travis.yml), making tar and
# unzip on Windows grumpy because dangling symlinks are created. Could be fixed
# by manually extracting everything except symlinks, but that's hard to
# maintain. Downloading a Git tag works.
```

Biswa96:
I am not sure if this right or wrong. But [noextract()][1] trick can be used here, many other PKGBUILDs also use the trick. See the following diff as a reference.

[1]: https://wiki.archlinux.org/index.php/PKGBUILD#noextract

<details><summary>Click here to see the diff in dropdown</summary>

```diff --git a/mingw-w64-corrade/PKGBUILD b/mingw-w64-corrade/PKGBUILD
index 098d8ff..48a29ce 100644
--- a/mingw-w64-corrade/PKGBUILD
+++ b/mingw-w64-corrade/PKGBUILD
@@ -16,8 +16,15 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cmake" "${MINGW_PACKAGE_PREFIX}-ninja" 'gi
 # unzip on Windows grumpy because dangling symlinks are created. Could be fixed
 # by manually extracting everything except symlinks, but that's hard to
 # maintain. Downloading a Git tag works.
-source=("${_realname}-${pkgver}"::"git+https://github.com/mosra/corrade.git#tag=v$pkgver")
-sha256sums=('SKIP')
+source=("${_realname}-${pkgver}.tar.gz::https://github.com/mosra/corrade/archive/v${pkgver}.tar.gz")
+noextract=(${_realname}-${pkgver}.tar.gz)
+sha256sums=('2a62492ccc717422b72f2596a3e1a6a105b9574aa9467917f12d19ef3aab1341')
+
+prepare() {
+  cd ${srcdir}
+  rm -rf ${_realname}-${pkgver}
+  tar -xzf ${_realname}-${pkgver}.tar.gz || true
+}
 
 build() {
     mkdir -p "${srcdir}"/build-${CARCH} && cd "${srcdir}"/build-${CARCH}
```

</details>

williamjcm:
> `tar -xzf ${_realname}-${pkgver}.tar.gz || true`

Wouldn't that cause issues if the extraction fails for a genuine reason ?

Biswa96:
`|| true` takes care of that error. As we don't have symlinks related to source code this procedure doesn't create any build error. Last decision to the maintainers.

williamjcm:
I mean, if the extraction fails due to a dying storage device, or because there's not enough space, or because of bad permissions, and the script sees the result of the expression is true due to the `|| true` part, it would lead to the script failing anyway.

Biswa96:
Then the build would fail :)

mosra:
What about using a zip archive instead? GitHub provides both and AFAICT the zip archives don't preserve symlinks, so there it would work.

Assuming makepkg can unzip on its own, I guess only `unzip` needs to be added to `makedepends`?

williamjcm:
> AFAICT the zip archives don't preserve symlinks

I think `unzip` defers symlinks until after the rest is extracted, according to this output I got when I tried:
```text
finishing deferred symbolic links:
  corrade-master/.travis.yml -> package/ci/travis.yml
```

mosra:
Oh, so that would make it work, right?

williamjcm:
Looks like it.

williamjcm:
Amended to use zip files instead of cloning the Git tags.

elieux:
Merging on the grounds of this being the author of these PKGBUILDs.

https://github.com/msys2/MINGW-packages/pull/6641/checks?check_run_id=833930388#step:4:2928 looks bad, but I assume it works outside of CI.

lazka:
> https://github.com/msys2/MINGW-packages/pull/6641/checks?check_run_id=833930388#step:4:2928 looks bad, but I assume it works outside of CI.

This was a setup-msys2 bug. We should see if it works on master -> https://github.com/msys2/MINGW-packages/runs/916244124

Issue #6607, opened 2020-06-18
clang's libc++ headers give linker's errors on duplicate RTTI


intractabilis:
Reopening #5405

Package|Version
-|-
mingw-w64-x86_64-clang|10.0.0-2
mingw-w64-x86_64-llvm |10.0.0-2
mingw-w64-x86_64-libc++|10.0.0-2
mingw-w64-x86_64-lld|10.0.0-2

If you try to build your code with libc++, you can get a lot of complains from the linker about duplicated symbols.

```cpp
// main.cpp
#include <stdexcept>

int main(int argc, char* argv[]) {
    throw std::runtime_error("Hello World!!!");
    return 0;
}
```

```
clang++ -v -stdlib=libc++ -fuse-ld=lld -o main.exe main.cpp
```
```
clang version 10.0.0 (https://github.com/msys2/MINGW-packages.git 7211ffb882cc3b7e7583c518aad45a22b278bc81)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: c:\msys64\mingw64\bin
[...]

Issue #6605, opened 2020-06-18
lldb cannot find libstdc++-6.dll


askondro:
$ lldb
C:/msys64/mingw64/bin/lldb.exe: error while loading shared libraries: libstdc++-6.dll: cannot open shared object file: No such file or directory

spaceotter:
Seeing a similar problem:
```
C:/msys64/mingw64/bin/lldb.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

spaceotter:
ldd doesn't seem to report a problem though:
```
$ ldd $(which lldb)
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7fff60560000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7fff5e900000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7fff5e080000)
        ADVAPI32.dll => /c/WINDOWS/System32/ADVAPI32.dll (0x7fff5fbb0000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7fff5e7c0000)
        liblldb.dll => /mingw64/bin/liblldb.dll (0x6ab80000)
        sechost.dll => /c/WINDOWS/System32/sechost.dll (0x7fff5e860000)
        ole32.dll => /c/WINDOWS/System32/ole32.dll (0x7fff5e9c0000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7fff5eba0000)
        combase.dll => /c/WINDOWS/System32/combase.dll (0x7fff5ed30000)
        zlib1.dll => /mingw64/bin/zlib1.dll (0x62e80000)
        dbghelp.dll => /c/WINDOWS/SYSTEM32/dbghelp.dll (0x7fff3fde0000)
        SHELL32.dll => /c/WINDOWS/System32/SHELL32.dll (0x7fff5fc60000)
        ucrtbase.dll => /c/WINDOWS/System32/ucrtbase.dll (0x7fff5d640000)
        ucrtbase.dll => /c/Windows/System32/ucrtbase.dll (0x4690000)
        bcryptPrimitives.dll => /c/WINDOWS/System32/bcryptPrimitives.dll (0x7fff5d740000)
        cfgmgr32.dll => /c/WINDOWS/System32/cfgmgr32.dll (0x7fff5e330000)
[...]

Issue #6422, opened 2020-04-26
mingw-w64-x86_64-lldb 10.0.0-1 lacks dependency for liblua53.dll


mkviatkovskii:
Hello!
When I install `mingw-w64-x86_64-lldb`, it has unmet dependency for shared library and cannot start:
```bash
$ lldb
C:/msys64/mingw64/bin/lldb.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```
With help of [Dependencies](https://github.com/lucasg/Dependencies), I found that the missing library is `liblua53.dll`:
![image](https://user-images.githubusercontent.com/4228185/80313706-90614680-87a1-11ea-9159-7bafd8373bf8.png)
And after installing `mingw64/mingw-w64-x86_64-lua`, the problem is fixed.

Could you please add this dependency for the `mingw-w64-x86_64-lldb`?

Issue #6126, opened 2020-01-12
Clang's LLD is broken, again!


HolyBlackCat:
After the last update of the llvm packages (`9.0.0-3`), LDD no longer works.
`ld.lld.exe` hangs indefinitely. (It miraculously worked once for me, but no luck since then.)

You can reproduce it simply by running
```sh
echo -e '#include <iostream>\nint main() {std::cout << "Hello!\\n";}' >a.cpp
clang++ -fuse-ld=lld a.cpp
```

This closely resembles bug https://github.com/msys2/MINGW-packages/issues/5231 that was already fixed.

mati865:
Ah I remember now, that's the same thing as in https://github.com/mstorsjo/llvm-mingw/issues/11
Backporting https://github.com/llvm/llvm-project/commit/564481aebe18a723c9cfe9ea9ca5808771f7e9d8 should fix it, otherwise #6098 has to be reverted.

cc @adrpo

adrpo:
Ouch! I guess I sort of expected some issues with allowing duplicate symbols when linking. 
I will see if i can patch with https://github.com/llvm/llvm-project/commit/564481aebe18a723c9cfe9ea9ca5808771f7e9d8 
If I can't o that I will try to link statically only lld.

mati865:
> I guess I sort of expected some issues with allowing duplicate symbols when linking.

This time it's a bug introduced somewhere around LLVM 8/9.

lazka:
@HolyBlackCat Can you test again now?

HolyBlackCat:
@lazka Works for me!

lazka:
Thanks!

mati865:
Unfortunately it's not fully fixed.
While building LLVM during Rust build I'm sometimes getting `ld.lld `hangs (over hour long).
Trying to debug hanged process unhangs it before I can get the stack trace.

HolyBlackCat:
@mati865 It sometimes hangs for me too, but very rarely (once per a few days).

mncat77:
Can confirm that ld.lld still hangs around once every three times it is invoked in an unpredictable fashion for me.

HolyBlackCat:
@lazka This probably needs to be reopened.

HolyBlackCat:
@adrpo What's the plan? It still hangs sometimes even with the patch, so the only option is to switch to static linking for LLD?

mati865:
[...]

PR #6664, opened 2020-07-09
Few LLVM fixes

mati865:
Fixes https://github.com/msys2/MINGW-packages/issues/6422
Fixes https://github.com/msys2/MINGW-packages/issues/6605
Fixes https://github.com/msys2/MINGW-packages/issues/6607

Libc++ build is based on scripts from https://github.com/mstorsjo/llvm-mingw


mati865:
This PR makes libc++ depend on libunwind instead of libgcc_s.
I think the biggest benefit from it is avoiding winpthreads (and it's bugs) but at the cost of additional DLL.
That said I'm not sure whether libc++ should use libunwind or libgcc.

With this fixed libc++ I was able to link LLVM dynamically without hangs we had previously (https://github.com/msys2/MINGW-packages/issues/6126).

ChinouneMehdi:
Is it possible to reenable dynamic linking?

mati865:
Yes but fixed libc++ has to land in the repo first. Then LLVM can be build with Clang and libc++ without hitting #6126.

PR #6666, opened 2020-07-09
gtk3: add patch from GIMP project (ported for gtk3)

Ede123:
This prevents transparent top-level windows from intercepting mouse
events as often the case with screen capturing software, see
https://bugzilla.gnome.org/show_bug.cgi?id=780979

Applies the same patch that was added for gtk2 in
  023eb8614b8f320563ecb625da334cd1392d430d


Ede123:
Confirmed to fix the issue at least partially in a downstream issue: https://gitlab.com/inkscape/inbox/-/issues/2976

PR #6667, opened 2020-07-09
Fix missing ngspice pkg-config file.

stambaughw:
I discovered that the pkg-config file for the ngspice package was missing so I updated the PKGBUILD file accordingly.  Please let me know if there is anything else I need to do to get the merged.  Thanks in advance for the help.


nickoe:
Looks good to me

stambaughw:
Thanks for merging this.  I'm assuming package will automatically be rebuilt or do I need to do something to make that happen as well?

Biswa96:
> do I need to do something to make that happen as well?

Just wait and watch here https://packages.msys2.org/updates

PR #6669, opened 2020-07-10
Fixed GnuCOBOL 3.1-rc1, please pull to fix a dependency issue

GitMensch:
* pkgrel 1 needed a "manual dependency" libxml2 before (because the build system has it, configure found it and linked it) --> made it explict
* explicit note all "optional" dependecies in the configure step to ensure nothing unexpected gets in / is missing
* let cjson be included (statically linked) by configure, after manual download (there's no package for cJSON yet)
* added `make check`and `make test` (including its source + a patch to ignore line breaks); currently without a hard exit

Note: I'm unsure how to go with newcob.val - actually it is only needed if the test is run, and then also auto-downloaded if missing (which is the reason for the checkdepends curl), but the download via source has the additional benefit of caching and SHA256 check...


mingw-w64-gnucobol/PKGBUILD
```diff
@@ -20,26 +20,46 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc"
...
+checkdepends=("${MINGW_PACKAGE_PREFIX}-curl" "${MINGW_PACKAGE_PREFIX}-perl")
+
+# local definitions
+CJSON_RELEASE_URL="https://raw.githubusercontent.com/DaveGamble/cJSON/v1.7.13"
 #source=("https://ftp.gnu.org/gnu/${_realname}/${_realname}-${pkgver}.tar.xz"{,.sig}
 source=("https://alpha.gnu.org/gnu/${_realname}/${_realname}-${pkgver_real}.tar.xz"{,.sig}
         "001-mingw-gnucobol-2.2-fixformatwarnings.patch"
-        "cobenv.sh" "cobenv.cmd")
+        "002-gnucobol-3.1-rc1-cobol85_crdiff.patch"
+        "${CJSON_RELEASE_URL}/cJSON.c"
```
lazka, line 40: Could we package cJSON instead?

mingw-w64-gnucobol/PKGBUILD
```diff
@@ -62,15 +84,15 @@ build() {
...
   make "--jobs=$(($(nproc)+1))"
 }
 
 check() {
   cd ${srcdir}/build-${CARCH}
   # note: strangely multiple tests seem to halt the complete package creation
   #make check TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"
-  make check
+  make check || echo "warning, not all internal tests passed"
+  make "--jobs=$(($(nproc)+1))" test || echo "warning, not all NIST tests passed"
```
lazka, line 95: Why do you need to pass jobs? It should work automatically.

mingw-w64-gnucobol/PKGBUILD
GitMensch: This would completely possible and likely won't harm at all, other package systems (for example brew) also have a recipe for that - this would only ease the packaging and especially allow also updating that part.
@lazka what do you think - maybe you give it a go, I have a second look at it and when it is available add pkgrel3 with the dependency?
[note: pkgrel1 must be considered broken as it does not work if libxml2 is not installed manually]?

mingw-w64-gnucobol/PKGBUILD
GitMensch: Interesting, if this is the case then I'll drop that part soon (it didn't work automatically with an old version; and actually 17 jobs with this hardware here seems to much in any case).
Note: I may try to add those to `make check` (internally to the script as parameter `TESTSUITEFLAGS="--jobs=something`"as it definitely benefits a lot from the speedup - doing so just lead to a hang years ago, didn't tried again).

Biswa96: The proper way is to edit the `MAKEFLAGS` in `/etc/makepkg.conf` file in your msys2 setup instead of hard-coding it in PKGBUILD file. This provides user to choose their own options.

GitMensch: @Biswa96 wow, where is that documented? Can I use that directly for `TESTSUITEFLAGS` somehow (otherwise I'd try to filter a possible `-j/--jobs` out of it).

Biswa96: Here https://wiki.archlinux.org/index.php/Makepkg#Improving_compile_times and here https://www.archlinux.org/pacman/makepkg.conf.5.html

mingw-w64-gnucobol/PKGBUILD
lazka: sounds good

lazka:
I've tried to create a cjson package in #7156

mingw-w64-gnucobol/PKGBUILD
GitMensch: Resolved by removing this part as it should not be used here according to the docs. I do wonder if I could pull the amount of jobs used for make out there easily to be used within the TESTSUITEFLAGS....

mingw-w64-gnucobol/PKGBUILD
GitMensch: integrated in the current version (and also removed the checkdepends curl as the external code for `make test` is now part of the files downloaded via source)

GitMensch:
CI doesn't find dependencies like `mingw-w64-i686-libxml2` or `mingw-w64-i686-db` which are listed at https://packages.msys2.org/package (it also stumbles over the missing cJSON but this isn't listed in the package list so far).

lazka:
It doesn't find mingw-w64-x86_64-cjson because it's not in the repo yet. You can work around that by bumping the pkgrel of cjson in this PR so it gets build/installed first.

GitMensch:
CI says:

> ==> Missing dependencies:
  -> mingw-w64-x86_64-cjson
  -> mingw-w64-x86_64-libxml2
  -> mingw-w64-x86_64-db

So you suggest that I increase the pkgrel for all those? I think this would be bad as it would update the libxml2 and db package for every user, wouldn't it?

lazka:
no, only cjson, the pacman error is misleading as it lists all dependencies, not just the ones not found.

lazka:
looks like it worked. ship it?

GitMensch:
> looks like it worked. ship it?

yes, please. I believe all issues with the package for rc1 were solved.

lazka:
cool, thanks!

PR #6671, opened 2020-07-11
New package: dfu-util 0.9

fauxpark:


fauxpark:
Hi, any update here?

fauxpark:
@lazka another friendly ping. Is there a group or something I can @ instead, or is this ok? Don't want to keep hassling just you :)

lazka:
I guess we could add a group/team thing. I'll have a look.

PR #6679, opened 2020-07-13
New package: libusb-win32 1.2.6.0

fauxpark:
This package is needed in order to build dfu-programmer without using libusb 1.0 - although it is possible, doing so would make it incompatible with the official release, as the MSYS2 dfu-programmer would be expecting the device to be using the WinUSB driver, while the release expects libusb0 (libusb-win32).


fauxpark:
Any update on this one?

fauxpark:
@lazka ping :)

PR #6681, opened 2020-07-14
nana: Update to v1.7.4

stickM4N:


lazka:
You have to update the checksums. Did you try to build it?

stickM4N:
I already build it and you are rigth, I forgot to comit it with the chesum,
I will do it in a few hours.

PR #6683, opened 2020-07-14
Add PKGBUILD for pdf-tools-server

svraka:
[PDF Tools](https://github.com/politza/pdf-tools) is an Emacs package for viewing PDFs. It relies on its own server, which needs to be compiled from source, which is of course [fairly complicated on Windows](https://github.com/politza/pdf-tools#compilation-and-installation-on-windows). This MINGW package – inspired by the similar [Debian](https://packages.debian.org/buster/elpa-pdf-tools-server) and [Ubuntu](https://packages.ubuntu.com/eoan/elpa-pdf-tools-server) packages – provides the binary for the PDF Tools server.


lazka:
We try to avoid packaging from git and prefer release tarballs instead. Do you think this is possible?

svraka:
Maybe. PDF Tools has [proper releases](https://github.com/politza/pdf-tools/releases) but the last one is from December 2018 and there were changes added to the git repo. 

Installing from a release seems a bit antithetical to how most Emacs users install packages. Outside [ELPA](https://elpa.gnu.org), the default package repository, [MELPA](https://melpa.org) is used most commonly, and they track source repositories directly. I opened an issue at https://github.com/politza/pdf-tools/issues/599.

APIPLM:
Which way should we go? Build the releases in the repo or create PKGBUILD file like in the msys2/MINGW-packages?
I follow [https://github.com/politza/pdf-tools#compilation-and-installation-on-window] to install, but trap on the step 5.
make: emacs: No such file or directory
Using Emacs
make: emacs: No such file or directory
make: emacs: No such file or directory
make: emacs: No such file or directory
make: cask: No such file or directory
make: *** [Makefile:44：.cask/] 错误 127

svraka:
I think those instructions refer to building the whole PDF Tools package, which requires Cask, and AFAIK there are some issues with Cask on Windows. Just do `make server/epdfinfo` and then install Lisp code from MELPA.

APIPLM:
@svraka Thanks for your help. It works as you said, build server/epdfinfo separately, copy the file epdfinfo.exe to D:/emacs/bin/, and add the directory "D:\msys64\mingw64\bin"  of its dependencies to %path%, then install Lisp code part from MELPA.
Do we issue number for Cask on Windows?

APIPLM:
Seem like there are two different projects about the cast in windows.One is in the below linker
https://cask.readthedocs.io/en/latest/guide/installation.html#windows-setup
the other one is that something you can install from MELPA.  They sound like the same one. but the description is so different.

svraka:
Thanks for the merge! Just to clarify, were there any policy changes regarding packaging straight form git?

lazka:
No, it's still preferred not to use git. I was just lazy here.

PR #6691, opened 2020-07-16
gcc: Set bugurl to "https://github.com/msys2/MINGW-packages/issues"

stahta01:
"gcc --help" displays this URL in last line of help.
Prior value was "https://sourceforge.net/projects/msys2"


Biswa96:
If this is correct increase pkgrel.

stahta01:
Bump pkgrel

PR #6696, opened 2020-07-17
libproxy: Add new patch to fix buffer overflow

lifeibiren:
https://github.com/libproxy/libproxy/pull/126


lazka:
Is this the version that was approved upstream?

mingw-w64-libproxy/PKGBUILD
```diff
@@ -38,6 +40,7 @@ prepare() {
   patch -p1 -i ${srcdir}/005-modules.patch
   patch -p1 -i ${srcdir}/006-python-site-package-dir.patch
   patch -p1 -i ${srcdir}/007-python37-support.patch
+  patch -p1 -i ${srcdir}/008-fix-buffer-overflow.patch
```
lazka, line 43: Please add a link to the upstream PR here so we know where it's coming from and when it's ok to remove.

PR #6719, opened 2020-07-26
Fix bmake package: Add Windows draft configs in mk/sys directory.

ofry:
Otherwise any library builds will fail - our LD don't support -Bforcearchive.


elieux:
Approved on the grounds of this being a new package and the pull request comes from its only contributor.

Issue #4006, opened 2018-06-24
llvm-config doesn't report shared libs correctly


RyanGlScott:
I have `mingw-w64-x86_64-llvm-6.0.0-2` installed. If I run `llvm-config --libs --link-shared`:

```
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMDemangle.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMSupport.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMBinaryFormat.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMCore.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMBitReader.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMMC.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMMCParser.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMObject.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMProfileData.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMAnalysis.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMBitWriter.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMTransformUtils.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMInstCombine.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMAsmParser.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMIRReader.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMLinker.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMScalarOpts.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMVectorize.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMInstrumentation.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMipo.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMCoroutines.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMTarget.dll
llvm-config: error: missing: C:\msys64\mingw64\bin/LLVMCodeGen.dll
[...]

PR #6721, opened 2020-07-26
Link LLVM dynamically

mati865:
Fixes https://github.com/msys2/MINGW-packages/issues/4006

There is a problem however:
```
$ llvm-config --link-shared --cxxflags
-ID:\msys64\mingw64/include -std=c++14 -stdlib=libc++  -fno-exceptions -fno-rtti -D_FILE_OFFSET_BITS=64 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
```
GCC doesn't understand `-stdlib=libc++` and will throw an error. So to link LLVM (and all related tools) dynamically at the very least we have to make GCC ignore `-stdlib`.

### DIFF
--- a/mingw-w64-clang/0001-Add-options-to-link-LLVM-with-compiler-rt-and-libunw.patch
+++ b/mingw-w64-clang/0001-Add-options-to-link-LLVM-with-compiler-rt-and-libunw.patch
@@ -1,0 +1,138 @@
+From 5dca7a33d7fbfdd4425650f060876b91885c124c Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Mateusz=20Miku=C5=82a?= <mati865@gmail.com>
+Date: Thu, 23 Jul 2020 18:03:58 +0200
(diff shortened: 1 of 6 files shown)


Biswa96:
Shouldn't we have to wait for next release or so? Patches count is going up.

mati865:
> Shouldn't we have to wait for next release or so? Patches count is going up.

That's true, 3 out of 5 patches added by this PR won't be necessary with LLVM 11.

lazka:
Because I was wondering and looked it up: LLVM 11 will be out Aug 26, so in one month.

mati865:
> LLVM 11 will be out Aug 26, so in one month

They have tendency to delay the release (it's not a bad thing though) so I'd expect it somewhere mid September.

mati865:
LLVM 11 rc3 was released recently and is expected to be promoted as final without changes.
I'll work on LLVM 11 PR instead.

Issue #6715, opened 2020-07-24
[mlpack] Build is broken


lazka:
```
 In file included from D:/a/_temp/msys/msys64/mingw64/include/armadillo:54,
                 from D:/a/MINGW-packages/MINGW-packages/mingw-w64-mlpack/src/mlpack-mlpack-3.1.1/src/mlpack/core/arma_extend/arma_extend.hpp:49,
                 from D:\a\MINGW-packages\MINGW-packages\mingw-w64-mlpack\src\mlpack-mlpack-3.1.1\src\mlpack\prereqs.hpp:108,
                 from D:\a\MINGW-packages\MINGW-packages\mingw-w64-mlpack\src\build-x86_64\src\mlpack\cotire\mlpack_CXX_prefix.cxx:4,
                 from D:/a/MINGW-packages/MINGW-packages/mingw-w64-mlpack/src/build-x86_64/src/mlpack/cotire/mlpack_CXX_prefix.hxx:4:
D:/a/_temp/msys/msys64/mingw64/include/armadillo_bits/compiler_setup.hpp:79:30: fatal error: D:/a/_temp/msys/msys64/mingw64/include/hdf5.h: No such file or directory
   79 | #define ARMA_INCFILE_WRAP(x) <x>
      |                              ^
compilation terminated.
CMake Error at D:/a/MINGW-packages/MINGW-packages/mingw-w64-mlpack/src/mlpack-mlpack-3.1.1/CMake/cotire.cmake:1885 (message):
  cotire: error 1 precompiling
  D:/a/MINGW-packages/MINGW-packages/mingw-w64-mlpack/src/build-x86_64/src/mlpack/cotire/mlpack_CXX_prefix.hxx.
Call Stack (most recent call first):
  D:/a/MINGW-packages/MINGW-packages/mingw-w64-mlpack/src/mlpack-mlpack-3.1.1/CMake/cotire.cmake:3558 (cotire_precompile_prefix_header)
```

Maybe just a missing dep?

1480c1:
seems the header file is from `mingw64/mingw-w64-x86_64-hdf5`

Biswa96:
mlpack needs version update.

Biswa96:
^ also mlpack needs version update.

lazka:
16cbb66e401531 was the problem. Now if fails with hundreds of linker errors..

revelator:
[...]

PR #6726, opened 2020-07-27
cmake: Support exporing all symbols with MinGW

mati865:
Fixes https://github.com/msys2/MINGW-packages/issues/6715


mati865:
Upstream issue https://gitlab.kitware.com/cmake/cmake/-/issues/21026

PR #6728, opened 2020-07-28
libsoxr: Remove libavcodec(FFmpeg) dependency

dofuuz:
For now, mingw-w64-x86_64-libsoxr has dependency to libavcodec(FFmpeg). 
But it's not described in package. So it's **not working without installing ffmpeg manually**.

And FFmpeg is quite huge library (and it's GPL'ed). It can cause difficulty when distributing.
Therefore, removing ffmpeg usage from libsoxr would be nice solution.

Actually removing WITH_AVFFT flag is all we have to do.
https://sourceforge.net/p/soxr/discussion/general/thread/19b1a43a/#ded0

In addition,
libsoxr's license is LGPL(2.1 or above), not GPL. So I fixed PKGBUILD and add LGPL text.
libsoxr is using PFFFT which has independent license. So I included it.


dofuuz:
@Adsun701

Biswa96:
So, the issue about the FFmpeg dependency. How about adding the dependency i.e. `${MINGW_PACKAGE_PREFIX}-ffmpeg` in `depends()` variable instead of removing the build option?

dofuuz:
> And FFmpeg is quite huge library (and it's GPL'ed). It can cause difficulty when distributing.

Also, PFFFT is faster than AVFFT. (see https://sourceforge.net/p/soxr/discussion/general/thread/19b1a43a/#ded0 )
So, removing ffmpeg dependency(WITH_AVFFT) is more clean solution.

Biswa96:
But ArchLinux also uses the same options as this mingw package and it does not mention ffmpeg as dependency. See https://github.com/archlinux/svntogit-packages/blob/packages/libsoxr/trunk/PKGBUILD.

dofuuz:
Well then, they're wrong.

libsoxr_dep
Compiling with WITH_AVFFT option requires avcodec(FFmpeg).

libsoxr is not working without FFmpeg for now.
```
*path-removed*.exe: error while loading shared libraries: libsoxr.dll: cannot open shared object file: No such file or directory
```

dofuuz:
Actually, CMake detects FFmpeg when compiling. 
libsoxr.dll compiled in FFmpeg is not installed dose not require FFmpeg.

```cmake
if (WITH_AVFFT)
  find_package (LibAVCodec REQUIRED)
  if (AVCODEC_FOUND)
    include_directories (${AVCODEC_INCLUDE_DIRS})
    set (LIBS ${LIBS} ${AVCODEC_LIBRARIES})
  endif ()
endif ()

if (WITH_AVFFT OR (CMAKE_SYSTEM_PROCESSOR MATCHES "^arm" AND SIMD32_FOUND AND WITH_CR32))
  find_package (LibAVUtil)
  if (AVUTIL_FOUND)
    include_directories (${AVUTIL_INCLUDE_DIRS})
    set (LIBS ${LIBS} ${AVUTIL_LIBRARIES})
  endif ()
endif ()
```

So, It could be OK without FFmpeg in Arch.
But, at least, libsoxr is not working without ffmpeg in MSYS for now.

lazka:
Assuming PFFFT works the same this looks good to me.

@Adsun701 any thoughts on this? Asking since you added the package.

Adsun701:
@lazka I think @dofuuz's pull request works fine. Since ffmpeg is a higher-level library it makes sense for libsoxr to use a lower-level alternative.

Biswa96:
If this is OK then I have three suggestions:
1. How about using `-DWITH_AVFFT=OFF` option? So that future contributors know that AVFFT is explicitly disabled.
2. Is it possible to add FFmpeg in `optdepends()` variable?
3. Can the license file be retrieved from source tarball instead of putting it here?

Adsun701:
@Biswa96 using the `-DWITH_AVFFT=OFF` is a good idea. But if AVFFT is explicitly disabled, it doesn't make sense to add FFmpeg in `optdepends()` since libsoxr will not be depending on FFmpeg.  For the license file, if you can find a link to it or the source tarball contains it, it makes more sense to install it from there rather than including the license in the repository.

Biswa96:
According to ArchLinux [PKGBUILD Wiki][1], optdepends is:

> An array of packages that are not needed for the software to function, but **provide additional features**...
> If the software works on multiple **alternative dependencies**, all of them can be listed here.

In this case AVFFT is an optional feature and alternative to PFFFT. Users can alternatively enable it with `-DWINT_AVFFT=ON` option. So, the reason fits well with the wiki. This is just my opinion, last decision is up to maintainer and author.

[1]: https://wiki.archlinux.org/index.php/PKGBUILD#optdepends

lazka:
Since it breaks without ffmpeg being installed it's not really optional imo

dofuuz:
@Biswa96

1. LGTM
2. No, it's not right to do so. libsoxr and FFmpeg has no direct relationship.
3. The source tarball does not have PFFFT license as separate file.
It can be obtained from [source file](https://sourceforge.net/p/soxr/code/ci/master/tree/src/pffft.c), but I think having separate file in package is more clean solution.

PR #6735, opened 2020-07-30
zbar: enable building zbarcam

itsmattkc:
I'm not sure if there was a reason to disable zbarcam, but enabling it builds fine. I've enabled DirectShow as well since it generally has much greater webcam compatibility than the default (VfW).


lazka:
increase pkgrel

itsmattkc:
Done, sorry about missing that

lazka:
Thanks!

Issue #6736, opened 2020-07-30
Weird static libraries in libzip pkgconfig


tprk77:
On my machine, the file contents are:

```
prefix=/mingw64
exec_prefix=${prefix}
bindir=${exec_prefix}/bin
libdir=${exec_prefix}/lib
includedir=${prefix}/include

zipcmp=${exec_prefix}/bin/zipcmp

Name: libzip
Description: library for handling zip archives
Version: 1.6.1
Libs:  -L${libdir} -lzip
Libs.private:  -lD:/msys64/mingw64/lib/libz.dll.a -lD:/msys64/mingw64/lib/libbz2.dll.a -lD:/msys64/mingw64/lib/liblzma.dll.a -lbcrypt -ladvapi32
Cflags: -I${includedir}
```

Note `-lD:/msys64/mingw64/lib/libz.dll.a`. I don't have a `D:` drive, so that path doesn't actually exist. Also, there's a `-l` in front of a static library. Anyway, it was breaking my build.

I was able to get my build working again by manually editing the pkgconfig file to have the following:

```
prefix=/mingw64
exec_prefix=${prefix}
bindir=${exec_prefix}/bin
libdir=${exec_prefix}/lib
includedir=${prefix}/include

zipcmp=${exec_prefix}/bin/zipcmp

Name: libzip
Description: library for handling zip archives
Version: 1.6.1
Libs:  -L${libdir} -lzip
Libs.private:  C:/msys64/mingw64/lib/libz.dll.a C:/msys64/mingw64/lib/libbz2.dll.a C:/msys64/mingw64/lib/liblzma.dll.a -lbcrypt -ladvapi32
Cflags: -I${includedir}
```

See also #6734.

lazka:
My guess would be because the static build gets installed over the shared one here (the order needs to be reversed): https://github.com/msys2/MINGW-packages/blob/52c6513cf726d9b39c29008fe2d54c32b374e01a/mingw-w64-libzip/PKGBUILD#L62

Sadly libzip doesn't build here:

```
[100%] Built target fopen_unchanged
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/ziptool.dir/objects.a(ziptool.c.obj):ziptool.c:(.text.startup+0x41b): undefined reference to `__imp__snprintf.constprop.0'
[100%] Built target in-memory
[...]

PR #6740, opened 2020-07-30
libzip: update to 1.7.3

Biswa96:


mingw-w64-libzip/0001-lib-compat-use-snprintf.patch
```diff
@@ -0,0 +1,11 @@
+--- libzip-1.7.3.orig/lib/compat.h
++++ libzip-1.7.3/lib/compat.h
+@@ -94,7 +94,7 @@
+ #if !defined(HAVE_FILENO) && defined(HAVE__FILENO)
+ #define fileno _fileno
+ #endif
+-#if defined(HAVE__SNPRINTF)
++#if !defined(HAVE__SNPRINTF)
```
mati865, line 8: This seems very wrong, you cannot use `_snprintf` if it doesn't exist.

Biswa96: " if it doesn't exist" -- example? The build and test went fine.

mati865: CMake detects if `_snprintf` exists here: https://github.com/nih-at/libzip/blob/c55cf12cec53a5ef519f98dfd11f4cb00e9ac0f1/CMakeLists.txt#L95
This patch defines `snprintf` as `_snprintf` only when it doesn't exist which is wrong.

If you want to workaround GCC bug mentioned in https://github.com/msys2/MINGW-packages/issues/6736#issuecomment-666154781 it'd be better to disable the check.

Biswa96: Done. Thank you ❤️

PR #6743, opened 2020-08-01
Sqlite 3.32.3: add extension package and split packages

jannick0:
This commits as of now provides that the extension dlls are placed into `/mingw[32|64]/share/sqlite/extensions` (sample name `xyz.dll`). Alternatively they could go into the `bin` directory, such that loading by sqlite3.exe would not need a path to the dlls, but the names should then rather be something like `mingw[64|32]-sqlite3-xyz.dll` or alike.
@lazka Any preferences? Any version of the two has pros and cons.


lazka:
I have a hard time judging what this does. I can only tell you that we don't split packages unless there is a good reason.

mingw-w64-sqlite3/PKGBUILD
```diff
@@ -62,31 +61,66 @@ build() {
...
+
   cd ${srcdir}/sqlite-doc-${_docver}
   mkdir -p ${pkgdir}${MINGW_PREFIX}/share/doc/${_realname}
   cp -R *  ${pkgdir}${MINGW_PREFIX}/share/doc/${_realname}/
-
-  install -Dm644 "${srcdir}/LICENSE" "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/LICENSE"
 }
+
+# define dynamically wrappers for package functions
+for p in "${pkgname[@]}"; do
```
lazka, line 124: uh, this is nice :D

jannick0: Yep - perhaps a candidate for a twist of `makepkg` and friends (including a check if `p` is already defined using something like `declare -F $p`)?  Unfortunately, I have no idea where this should be inserted, though.  

This could make non-MSYS2 `PKGBUILD`s of split packages cleaner and better maintanable I think.

jannick0:
> I have a hard time judging what this does.
Do you prefer to have several commits to logically split things up? 
> I can only tell you that we don't split packages unless there is a good reason.
Package size? But happy to dial back from split packages.

lazka:
> > I have a hard time judging what this does.
> 
> Do you prefer to have several commits to logically split things up?

It would be helpful, yes. One question before that: What are sqlite extensions and why should MSYS2 build them when Arch for example doesn't. And why doesn't upstream want us to build them?

> > I can only tell you that we don't split packages unless there is a good reason.
> 
> Package size? But happy to dial back from split packages.

Package size can be a reason, but for that the package would need to be 10x the size. so, yes please remove that.

jannick0:
> > Do you prefer to have several commits to logically split things up?
> 
> It would be helpful, yes.

OK. I can do that, while the commit message details out the various issues covered in a single commit. I thought this gets squashed anyways. ...   

> One question before that: What are sqlite extensions and why should MSYS2 build them when Arch for example doesn't. And why doesn't upstream want us to build them?

Sqlite extensions are shared libraries  loaded at run-time and providing additional   SQL code functions beyond `sqlite`'s standard code. One simple but powerful example is the `REGEXP` operator or the `EVAL` function; many things for more advanced SQL scripts are hardly or even not possible with sqlite's standard SQL dialect. Unfortunately, sqlite does not provide an automated build script for extensions, but instructions how to build them, BTW one c-file per extension ... more [here](https://www.sqlite.org/loadext.html), and c-files [here](https://www.sqlite.org/src/dir?ci=tip&name=ext/misc).  (BTW: Alternative extensions could be compiled into the sqlite executable which will increase its size and load time - not recommendable I believe.)  

I have no idea why other distros do not ship these valuable tools, while there is a praiseworthy exception: `MSYS2` itself! :)

(Side remark: The reason why **some** the current extensions end up in the MSYS2 sqlite package, namely the split package `extensions`,  is that they get `imported` in an untransparent and totally unmaintainable manner, namely by the [gigantic patch](https://github.com/msys2/MSYS2-packages/blob/master/sqlite/sqlite3-3.32.3-1.src.patch) which is copied from CYGWIN (I believe) once someone finds the time to create a new patch for a new sqlite release. From far sight it took me some time to understand what is going on in the patch. Unfortunately, some of new extensions are left out and do not find their way into the patch (e.g. `uuid`, remember I tried this once) for one reason or another. Sigh.)

So my suggestion: ship the extensions here using the technique (see PR) which automatically captures new extensions. Once upstream ships a build script, then we can drop that, of course.    

> Package size can be a reason, but for that the package would need to be 10x the size. so, yes please remove that.

No problem - will do.


On the MSYS sqlite extension package: Happy to draft a MSYS2 PR to apply athe same technique like here to get rid of the dependence of the gigantic 3rd-party patch which as such represents an uncomfortable bottleneck for updating the MSYS2 sqlite package once a new source package releases comes out. ... Just let me know. ;)

jannick0:
@lazka Here you are: The changes split into tiny commit steps. Split packages removed as discussed. Hope this is fine.

PR #6745, opened 2020-08-02
faudio: udpate to 20.08

Biswa96:
* Replace dependency ffmpeg to gstreamer
* Add glib2, gstreamer, gst-plugins-base as dependencies
* Use -DGSTREAMER instead of -DFFMPEG option for cmake


lazka:
Not against this, just wondering why there is the switch from ffmpeg to gstreamer?

Biswa96:
Phoronix explained it well here https://www.phoronix.com/scan.php?page=news_item&px=FAudio-20.08-Released

lazka:
perfect, thanks

PR #6747, opened 2020-08-02
[do not merge] Fix shiboken2 and pyside2 build issues

raccoonmonk:
There is at least one license issue though. PySide2 is built with Qt Data Visualization module which is not available under LGPL. https://doc.qt.io/qt-5/qtmodules.html#qt-add-ons
The question is should I exclude it from build or mark the Pyside2 package with more restrictive "GPL".

### DIFF
--- a/mingw-w64-pyside2-qt5/001-pyside.patch
+++ b/mingw-w64-pyside2-qt5/001-pyside.patch
@@ -1,18 +1,7 @@
-diff -Naur a/sources/pyside2/CMakeLists.txt b/sources/pyside2/CMakeLists.txt
---- a/sources/pyside2/CMakeLists.txt	2018-06-19 16:00:53.944984400 +0300
-+++b/sources/pyside2/CMakeLists.txt	2018-06-19 16:18:12.926410700 +0300
-@@ -232,7 +232,7 @@
- use_protected_as_public_hack()
- 
- # Build with Address sanitizer enabled if requested. This may break things, so use at your own risk.
--if (SANITIZE_ADDRESS AND NOT MSVC)
-+if (SANITIZE_ADDRESS AND NOT WIN32)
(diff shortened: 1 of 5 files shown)


raccoonmonk:
Do I need to break it into 2 separate pull requests? pyside2 has dependency on shiboken2

Biswa96:
The `pkgrel` should be 1 as the `pkgver` is increased. Also I think the CI did not build the new dependency `mingw-w64-shiboken2-qt5` first and hence the build failed.

Biswa96:
The `pkgrel` should be 1 as the `pkgver` is increased. Also I think the CI did not build the new dependency `mingw-w64-shiboken2-qt5` first and hence the build failed. Generally Ci automatically detects the dependency and build it first but not here. Handing over to CI experts here.

mati865:
If you split both packages into own commits CI should follow that order.

raccoonmonk:
@mati865 CI didn't follow the commit order.

mati865:
Packages are not built on CI so it shouldn't matter. CI is used only to indicate if package is likely build but it's too fragile to depend on it.

raccoonmonk:
I see. So even if shiboken2 will be checked first pyside2 will fail anyway since it depends on shiboken. CI will not install or even build it.
I checked both of them on my machine with both x86 and x64 arch. Tests pass.
What do we do now?

mati865:
Wait for the reviewer I suppose.

Alexpux:
@raccoonmonk is it building for you?

raccoonmonk:
@Alexpux it was at the time. Both x86 and x64. Plus I had made a separate configuration with tests which I successfully ran. Also both Arch.
Are there any issues? When I'll have time I can build it again with recent llvm and Qt versions. If that's the problem.

PR #6757, opened 2020-08-04
Fallback to normal ld if gcc ld does not work in Python distutils

### LINKED ISSUE #6744: Finding ld from Python distutils
Hi. For mingw-w64-python, there is a [patch file for Python distutils](https://github.com/msys2/MINGW-packages/blob/b1cf7cc13d656e9d06212911b0c195015c6140fa/mingw-w64-python/0630-mingw-_winapi_as_builtin_for_Popen_in_cygwinccompiler.patch#L25-L36) so that it doesn't look for 'ld', it asks gcc where ld is.

However, I find that, at least in the AppVeyor MinGW, this [fails](https://ci.appveyor.com/project/Python-pillow/pillow/builds/26416349/job/74tvtjhe0snk25a0). Python Pillow added a [fallback](https://github.com/python-pillow/Pillow/pull/4019/files) to just use 'ld' if the gcc search fails. For a more recent example, see [this passing build](https://ci.appveyor.com/project/radarhere/pillow/builds/34426479), and [this failing build](https://ci.appveyor.com/project/radarhere/pillow/builds/34426497) if the [workaround is removed](https://github.com/radarhere/Pillow/commit/cea79d0a5458326605c1335538061df1850808de)

I'm wondering if this could be integrated into the patch itself - so that if gcc doesn't correctly find ld, then 'ld' is just used?

radarhere:
Suggestion to resolve #6744

### DIFF
--- a/mingw-w64-python/0630-mingw-_winapi_as_builtin_for_Popen_in_cygwinccompiler.patch
+++ b/mingw-w64-python/0630-mingw-_winapi_as_builtin_for_Popen_in_cygwinccompiler.patch
@@ -23,17 +23,27 @@ diff -Naur Python-3.8.0-orig/Lib/distutils/cygwinccompiler.py Python-3.8.0/Lib/d
      If not possible it returns None for it.
      """
 -    commands = ['gcc -dumpversion', 'ld -v', 'dllwrap --version']
+-    return tuple([_find_exe_version(cmd) for cmd in commands])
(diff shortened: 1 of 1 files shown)


lazka:
for future reference, this was closed because it is not needed, see https://github.com/msys2/MINGW-packages/issues/6744#issuecomment-687032654

Issue #6754, opened 2020-08-03
Renaming gprbuild-gpl package


Fabien-Chouteau:
Hi @juergenpf and @reznikmm,

I think we should rename package (`gprbuild-gpr`)[https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-gprbuild-gpl] in to `gprbuild`.
The `-gpl` suffix comes from the time when this was built from the sources of the "GNAT GPL" release.
But this release doesn't exists anymore and the package is now build from the GitHub sources.

The `PKGBUILD` documentation mentions a field to make sure there is not disruption on the user side:
```shell
replaces=("${MINGW_PACKAGE_PREFIX}-gprbuild-gpl")
```

What do you think?

reznikmm:
I vote for removing `-gpl` from the package name.

juergenpf:
I’m fine with that. Haven’t tracked the GNAT stuff for a while now…

Jürgen

Von: Fabien Chouteau <notifications@github.com>
Gesendet: Montag, 3. August 2020 16:45
An: msys2/MINGW-packages <MINGW-packages@noreply.github.com>
Cc: Jürgen Pfeifer <juergen@familiepfeifer.de>; Mention <mention@noreply.github.com>
Betreff: [msys2/MINGW-packages] Renaming gprbuild-gpl package (#6754)


Hi @juergenpf<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjuergenpf&data=02%7C01%7C%7Ced55b2ac03bf4a87e8a608d837bbc8f7%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637320626949006557&sdata=pqeqjWlKYzRGe3XVZ7s3cqGLuYx8kBnLdJFlFg%2B0c%2BM%3D&reserved=0> and @reznikmm<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Freznikmm&data=02%7C01%7C%7Ced55b2ac03bf4a87e8a608d837bbc8f7%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637320626949016527&sdata=HQvPkuqN3l3gdgJG1ej8n2%2B1dZEYu7JRiHiY4b4X46I%3D&reserved=0>,

[...]

PR #6760, opened 2020-08-05
Rename gprbuild-gpl, xmlada-gpl into gprbuild, xmlada

Fabien-Chouteau:
As discussed in #6754, the `-gpl` suffix comes from the time when these packages were built from the sources of the `GNAT GPL` release. But the `GNAT GPL` release doesn't exist anymore, and the packages are now built from the GitHub repos.

This PR also updates `xmlada`, this is required for compatibility with GCC GNAT 10.

CC @juergenpf and @reznikmm

### DIFF
--- a/mingw-w64-gprbuild-bootstrap-git/PKGBUILD
+++ b/mingw-w64-gprbuild-bootstrap-git/PKGBUILD
@@ -8,6 +8,12 @@ pkgbase=mingw-w64-${_realname}-git
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}-git
 pkgver=r3576.43e55df6
 pkgrel=1
+
+_gprbuild_commit="43e55df6c26e2bd48ab6c2f4214f497c4ddbc943"
+_gprbuild_extract_dir="AdaCore-gprbuild-43e55df"
+_xmlada_commit="c16f24ebf752b7d75dfdd0eaa7dd501a08a31256"
+_xmlada_extract_dir="AdaCore-xmlada-c16f24e"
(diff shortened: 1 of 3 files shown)


Biswa96:
A little query. Is it possible to make the packages from a source tarball instead of git repository? Also the `pkgver` would be different.

1480c1:
https://docs.github.com/en/rest/reference/repos#download-a-repository-archive

you could try using github api to do something like `api.github.com/repos/AdaCore/gprbuild/tarball/43e55df6c26e2bd48ab6c2f4214f497c4ddbc943`

1480c1:
untaring that results in a folder called `AdaCore-gprbuild-43e55df/`

Biswa96:
I mean actual release/tag of that repository instead of a git commit.

1480c1:
their single tag seems a bit behind https://github.com/AdaCore/gprbuild/releases/tag/community-2019

Fabien-Chouteau:
We are not doing releases on GitHub yet.

I use the `api.github.com/repos/AdaCore/gprbuild/tarball` with sha256s if that's better.

Fabien-Chouteau:
I added a patch to use `api.github.com` tarballs.

lazka:
@Biswa96 @1480c1 can you have another look?

lazka:
thanks. Agreed, but this seems like a minor issue and forgetting updating the hash will fail the build. I'm going ahead and merge this now.

Fabien-Chouteau:
Thanks for the merge

PR #6762, opened 2020-08-05
graphviz: fix the build

lazka:
It's an ancient version, but this is easier than updating.

Use a commit hash because the git tag was removed upstream.
Build with -fcommon which was the default before gcc 10.


Biswa96:
Thinking of using Python3 🤔

lazka:
Hm, not quite sure if it uses Python at all. I'll try skipping it once this goes through.

PR #6774, opened 2020-08-10
New package: mingw-w64-aws-sdk-cpp

jeroen:
This is a dependency that a lot of other libraries are interested in. Upstream doesn't officially support mingw yet (they would like to), so this is based on open pull requests from https://github.com/aws/aws-sdk-cpp/pull/1333 and https://github.com/awslabs/aws-c-common/pull/604. I added a patch to make it build on 32-bit and also with older versions of the mingw crt.

It is recommended by upstream to build a static libs (you get linking errors when building shared): https://github.com/aws/aws-sdk-cpp/issues/1309


1480c1:
Perhaps it might be better to split the cmake changes and the c source file changes into two patches for easier rebasing and managing?

mingw-w64-aws-sdk-cpp/BuildAwsCCommon.patch
```diff
@@ -0,0 +1,188 @@
...
++#    pragma warning(push)
++#    pragma warning(disable : 4996)
++#endif
+ 
+     const char *value = getenv(aws_string_c_str(variable_name));
+ 
+-#pragma warning(pop)
++#ifndef __MINGW32__
++#    pragma warning(pop)
++#endif
```
1480c1, line 128: perhaps it's better to use
```c
#ifdef _MSVC
```
instead of `ndef __MINGW32__` since I don't think this syntax of pragmas are supported on any other compiler other than msvc

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+
+build() {
+  export CFLAGS="-D_WIN32_WINNT=0x0600"
+  export CXXFLAGS="-D_WIN32_WINNT=0x0600"
+
+  [[ -d "${srcdir}"/build-${CARCH}-static ]] && rm -rf "${srcdir}"/build-${CARCH}-static
+  mkdir -p "${srcdir}"/build-${CARCH}-static && cd "${srcdir}"/build-${CARCH}-static
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=;-DCMAKE_PREFIX_PATH=" \
+    ${MINGW_PREFIX}/bin/cmake \
+      -G'MSYS Makefiles' \
```
1480c1, line 42:
```suggestion
      -GNinja \
```
I recommend using the Ninja generator instead since MSYS Makefiles is more designed for mingw's msys, and not msys2 and also it's much faster overall

changing this will require adding ninja to the builddeps and the make at the bottom to ninja, but alternatively the make like can be changed to
```
cmake --build .
``` 
instead

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+      -DCMAKE_BUILD_TYPE=Release \
+      -DBUILD_SHARED_LIBS=OFF \
+      -DCMAKE_INSTALL_PREFIX="c:${MINGW_PREFIX}" \
```
1480c1, line 45: I'm not sure if this is correct

mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch
```diff
@@ -0,0 +1,291 @@
...
+diff --git a/aws-cpp-sdk-core/include/aws/core/utils/Array.h b/aws-cpp-sdk-core/include/aws/core/utils/Array.h
+index 1dbb9020f4..6768b6b850 100644
+--- a/aws-cpp-sdk-core/include/aws/core/utils/Array.h
++++ b/aws-cpp-sdk-core/include/aws/core/utils/Array.h
+@@ -64,7 +64,7 @@ namespace Aws
+                 {
+                     m_data.reset(Aws::NewArray<T>(m_size, ARRAY_ALLOCATION_TAG));
+ 
+-#ifdef _WIN32
++#if defined(_WIN32) && !defined(__MINGW32__)
```
1480c1, line 31: if it's this combination, it might be better to use `_MSVC` instead

mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch
```diff
@@ -0,0 +1,291 @@
...
+diff --git a/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp b/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
+index 4a0f80a161..a98635f706 100644
+--- a/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
++++ b/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
+@@ -267,7 +267,7 @@ bool WinHttpSyncHttpClient::DoQueryHeaders(void* hHttpRequest, std::shared_ptr<H
+     wmemset(contentTypeStr, 0, static_cast<size_t>(dwSize / sizeof(wchar_t)));
+     
+     WinHttpQueryHeaders(hHttpRequest, WINHTTP_QUERY_CONTENT_TYPE, nullptr, &contentTypeStr, &dwSize, 0);
+-    if (contentTypeStr[0] != NULL)
++    if (contentTypeStr[0] != 0)
```
1480c1, line 115:
```suggestion
-    if (contentTypeStr[0] != NULL)
+    if (contentTypeStr[0])
```

this could be changed to just this instead of 0 since any value that is not convertible to 0 is considered true

mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch
```diff
@@ -0,0 +1,291 @@
...
+index db054b2c1b..e1d6dce55e 100644
+--- a/cmake/compiler_settings.cmake
++++ b/cmake/compiler_settings.cmake
+@@ -11,6 +11,9 @@ else()
+         set(COMPILER_CLANG 1)
+     else()
+         set(COMPILER_GCC 1)
++        if(CMAKE_CXX_IMPLICIT_LINK_LIBRARIES MATCHES "mingw32")
++            set(COMPILER_MINGW 1)
++        endif()
```
1480c1, line 264: there's an explicit `MINGW` variable.

although it was added in 3.2, which just so happens to be point one above the min required 3.1

https://cmake.org/cmake/help/v3.2/variable/MINGW.html

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -0,0 +1,58 @@
...
+  patch -p1 -i "${srcdir}/Patch-cmake-submodules.patch"
+}
+
+build() {
+  export CFLAGS="-D_WIN32_WINNT=0x0600"
+  export CXXFLAGS="-D_WIN32_WINNT=0x0600"
+
+  [[ -d "${srcdir}"/build-${CARCH}-static ]] && rm -rf "${srcdir}"/build-${CARCH}-static
+  mkdir -p "${srcdir}"/build-${CARCH}-static && cd "${srcdir}"/build-${CARCH}-static
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=;-DCMAKE_PREFIX_PATH=" \
```
1480c1, line 41:
```suggestion
  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
```
I don't see `-DCMAKE_PREFIX_PATH` being used anywhere else, maybe this can be removed?

jeroen: OK will try!

mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch
jeroen: Done

mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch
jeroen: Done

mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch
jeroen: Done

mingw-w64-aws-sdk-cpp/PKGBUILD
jeroen: Fixed

mingw-w64-aws-sdk-cpp/PKGBUILD
jeroen: Thanks! I did not know this. I have changed it to ninja now.

mingw-w64-aws-sdk-cpp/BuildAwsCCommon.patch
jeroen: Done

jeroen:
> Perhaps it might be better to split the cmake changes and the c source file changes into two patches for easier rebasing and managing?

The patch is copied from https://github.com/aws/aws-sdk-cpp/pull/1333. I think it's ok to keep them together; there is only a single cmake change in the patch.

jeroen:
Thanks so much for your suggestions @1480c1 ! I did not know it was possible to generate ninja files from cmake, very cool.

Do you think it looks ready now?

mingw-w64-aws-sdk-cpp/BuildAwsCCommon.patch
```diff
@@ -0,0 +1,188 @@
...
+diff --git a/tests/atomics_test.c b/tests/atomics_test.c
+index 36dbbd1a5..90f0497e2 100644
+--- a/tests/atomics_test.c
++++ b/tests/atomics_test.c
+@@ -23,7 +23,9 @@
+ 
+ #ifdef _WIN32
+ #    include <malloc.h>
+-#    define alloca _alloca
++#    ifndef __MINGW32__
```
1480c1, line 183: Nit: Perhaps this can be changed to use msvc as well

jeroen: Thanks!

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
+options=('strip' 'staticlibs')
+source=("${_realname}-${pkgver}.tar.gz::https://github.com/aws/aws-sdk-cpp/archive/${pkgver}.tar.gz"
+        "aws-sdk-cpp-pr-1333.patch"
+        "Patch-cmake-submodules.patch"
+        "BuildAwsCCommon.patch"
+        "BuildAwsChecksums.patch")
+sha256sums=('95e3f40efaea7b232741bfb76c54c9507c02631edfc198720b0e84be0ebb5e9d'
+            'SKIP'
```
lazka, line 21: remove the SKIPs and run `updpkgsums`

jeroen: Thanks!

jeroen:
Awesome!! 🚀

jeroen:
cc: @ihnorton @kou @xhochy @nealrichardson, once the binaries are available, we can enable S3 support in arrow and tiledb.

Biswa96:
I am thinking to separate out the dependencies to different packages, like ArchLinux does. Any thought on that? Also ArchLinux seems to use shared libraries. If shared libraries are enabled would it create any issue with other dependent packages?

jeroen:
I think that is fine, though I'm not sure what the benefit would be. The full aws-sdk-cpp consists of many inter-dependent libraries, I think it would be quite complex to package them all separately with proper dependencies. Also I am not sure about the ABI stability; I think if you build shared libraries, you may need to rebuild arrow/tiledb for every revision of the aws-sdk-cpp.

PR #6680, opened 2020-07-13
clang: fix license

106062316:

PR #6778, opened 2020-08-11
clang: updated to 10.0.1

oscarfv:


lazka:
@mati865 ok?

mati865:
What about https://github.com/msys2/MINGW-packages/pull/6680?
It includes update but also fixes license and makedepends. We could also include https://github.com/mati865/MINGW-packages/blob/6a159fed7a510825f05d4236e723b8260c03e148/mingw-w64-clang/0016-Add-ffi-to-system-libraries-list-if-enabled.patch from https://github.com/msys2/MINGW-packages/pull/6721 which is drive-by fix for another issue.

lazka:
ah, good catch, thanks. Merged #6680 now.

As for the patch, PR welcome I guess :)

@oscarfv thanks for the PR! but closing as a duplicate of #6680

PR #6779, opened 2020-08-11
Fix bmake package (fix DLL building)

ofry:
rel=4. Fix DLL building on Windows.


elieux:
Checks have passed, @ofry is the package author&maintainer, let's merge. :)

PR #6787, opened 2020-08-12
openimageio: update to 2.1.18.1

Biswa96:


Biswa96:
I don't know what happened in CI. I built the package in my system and it went fine. The CMake option `STOP_ON_WARNING` should disable the `-Werror` option. But in CI it does not.

lazka:
Maybe only 32 bit fails? I need to make CI not fail the running builds if one fails..

Biswa96:
Investigating...

lazka:
Probably because CI sets the CI env var: https://github.com/OpenImageIO/oiio/blob/14ee3c30e4d6ef9d6a8bca85fa82f6bf1cbeb158/src/cmake/compiler.cmake#L90

Biswa96:
Done!

lazka:
unsetting CI would have been easier than patching... but OK :)

PR #6792, opened 2020-08-13
Add GN, google's metabuilder for Ninja.

Kwizatz:
Same as https://github.com/msys2/MSYS2-packages/pull/2099 but for native mingw.

This one has some issues which is why this is a draft PR.


mingw-w64-gn/PKGBUILD
```diff
@@ -0,0 +1,43 @@
+
+pkgname=gn
+pkgver=1822
+pkgrel=1
+pkgdesc="GN is a meta-build system that generates build files for Ninja."
+arch=('i686' 'x86_64')
```
mati865, line 6:
```suggestion
arch=('any')
```
MinGW packages use `any` arch.

Kwizatz:
Upstream fixes on the way:
https://gn-review.googlesource.com/c/gn/+/9720

jeroen:
Thanks for this effort. I am also very interested in a fresh mingw build of V8! Our current production version is V8 6.2 that we [built with node 8.11.1](https://github.com/rwinlib/libv8)

Kwizatz:
@jeroen I am happy that it benefits someone other than me 😁.

The v8 PKGBUILD is at
https://github.com/Kwizatz/MINGW-packages/tree/v8

I have not started a PR because I would like to have this one ready first, but that one should work without this gn package. It downloads the official win64 binary and uses that.

1480c1:
> > I think we can close this one
> 
> Why? This gn package can also be used to build other programs.

since I managed to come back here when trying to figure out why a personal liblibGLESv2.a was broken and trying to see if I could compile angle, I think gn as a package doesn't really work out as mentioned previously that each project that uses gn requires a specific hash of gn to be used, so a generic gn will not work (this is also another reason why I hope gn and other "generators" that don't even last a single patch version die out somehow since this makes it really hard for anyone other than the original creator to make and use gn stuff)

Kwizatz:
@Biswa96 closing this because as @1480c1 said, having gn as package by itself may do more harm than good, if someone else has better arguments for it, they may create a separate PR, if someone is looking on how to get gn for a PKGBUILD, look at how it is done in v8's PKGBUILD.

Biswa96:
As you wish. But GN is used in many projects as alternative build system for example, every Vulkan projects in KhronosGroup. I just want to play with it 👶

1480c1:
if you want to try it out, you might want to checkout your own hash that you want to test and use this PKGBUILD as a template, that might be the better option

Kwizatz:
@Biswa96 Sorry 😢that was my original intention, VCPKG has build-in support for it and I thought it was stable enough, but that does not seem to be the case.

On a related note, I literally just noticed skia is build with GN, also noticed there is no mingw skia package, I might give that a try that in the future.

PR #6798, opened 2020-08-15
CI: store packages as GHA artifacts

jeroen:
This makes the artifacts from available for download from GHA. This is very useful, for example when you want to download the binaries from a PR and run them locally.


lazka:
Thanks!

Biswa96:
Maybe add this in MSYS2-Packages also?

Issue #6799, opened 2020-08-15
Layer support files for mingw-w64-vulkan-validation-layers


xzn:
Trying to build https://github.com/aejsmith/vkdevicechooser but vk`vulkan/vk_layer_dispatch_table.h` is missing.

This could be worked around by changing [`PKGBUILD`](https://github.com/msys2/MINGW-packages/blob/feeb9c28b064d4e587498f0140424c5253076b8d/mingw-w64-vulkan-validation-layers/PKGBUILD)

```diff
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
     -DGLSLANG_INSTALL_DIR=${MINGW_PREFIX} \
     -DSPIRV_HEADERS_INSTALL_DIR=${MINGW_PREFIX} \
-    -DBUILD_LAYER_SUPPORT_FILES=OFF \
+    -DBUILD_LAYER_SUPPORT_FILES=ON \
+    -DCMAKE_INSTALL_INCLUDEDIR=include/vulkan \
     -DBUILD_TESTS=OFF \
     ../${_realname}-${pkgver}
```

The resulting layer seems to compile and work.

Biswa96:
Would you like to create a pull request to add the change? Contributions are always welcomed 🤗

lazka:
I guess this is fixed by #6800

PR #6800, opened 2020-08-15
vulkan-validation-layers: Build layer support files

xzn:
From #6799,

Changes are based on https://github.com/archlinux/svntogit-packages/blob/packages/vulkan-validation-layers/trunk/PKGBUILD

As a test I was able to compile [vkdevicechooser](https://github.com/aejsmith/vkdevicechooser) with both mingw64 and mingw32:

```
LDFLAGS=-static meson build-win64
meson compile -C build-win64
```

and

```
LDFLAGS='-static -Wl,--kill-at' meson build-win32
meson compile -C build-win32
```


lazka:
@Biswa96 does this look OK to you?

Biswa96:
Yup.

PR #6801, opened 2020-08-15
hub 2.14.2

tschoonj:
Long time user, first time contributor... 

Please let me know if I screwed up!


Biswa96:
Mistake! pkgrel should be 1.

tschoonj:
Ah sorry! Do I open a PR to fix this?

lazka:
oops, thanks, fixed in master

Biswa96:
Also some files installed in `/usr` folder instead of `/mingw` one.

tschoonj:
That was already the case with the previous version.

lazka:
We should really show a file tree of the package content in CI

PR #6804, opened 2020-08-15
Update innoextract to 1.9.

ofry:
Update innoextract to 1.9. Get rid of patch. It merged to upstream in https://github.com/dscharrer/innoextract/commit/1381ebdab633d827faced1b994693deda05de98d


Biswa96:
It would be helpful if a commit message is added in the commit about the removal of patch file, for example, where (link) it is merged with upstream.

ofry:
@Biswa96 I added it in PR message. Is it OK?

Biswa96:
Though I would prefer in the commit message but it's ok.

PR #6817, opened 2020-08-17
scour: add package

Biswa96:


Biswa96:
Required for Inkscape project. /cc @Ede123

lazka:
The arch package is just named "scour". Should this be too?

Biswa96:
I know. I first thought about it but can't get to any conclusion.

1. My left brain said to use 'mingw-w64-python-scour' as it makes sense, fits with other packages and it is a python package.
2. My right brain said to use 'mingw-w64-scour' so that it syncs with the package sync/update ritual with ArchLinux.

Let's have a vote 🗳️

lazka:
Or @Ede123 decides :)

Ede123:
What's more common (with or without prefix)?

I see debian splits "python-scour" (module) and "scour" (executables), but as arch doesn't seem do that and I understand we'd like to be on par with Arch (which I think we usually aim for) just "scour" would probably be the way to go.

I'm fine with both, though, if anybody else feels strongly about one or the other option.

Ede123:
What's more common (with or without prefix)?

I see debian splits "python-scour" (module) and "scour" (executables), but as arch doesn't seem do that and I understand we'd like to be on par with Arch (which I think we usually aim for) just "scour" would probably be the way to go.

I'm fine with both, though, if anybody else feels strongly about one or the other option.

Ede123:
What's more common (with or without prefix)?

I see debian splits "python-scour" (module) and "scour" (executables), but as arch doesn't seem do that and I understand we'd like to be on par with Arch (which I think we usually aim for) just "scour" would probably be the way to go.

I'm fine with both, though, if anybody else feels strongly about one or the other option.

Ede123:
What's more common (with or without prefix)?

I see debian splits "python-scour" (module) and "scour" (executables), but as arch doesn't seem do that and I understand we'd like to be on par with Arch (which I think we usually aim for) just "scour" would probably be the way to go.

I'm fine with both, though, if anybody else feels strongly about one or the other option.

Ede123:
What's more common (with or without prefix)?

I see debian splits "python-scour" (module) and "scour" (executables), but as arch doesn't seem do that and I understand we'd like to be on par with Arch (which I think we usually aim for) just "scour" would probably be the way to go.

I'm fine with both, though, if anybody else feels strongly about one or the other option.

Ede123:
What's more common (with or without prefix)?

I see debian splits "python-scour" (module) and "scour" (executables), but as arch doesn't seem do that and I understand we'd like to be on par with Arch (which I think we usually aim for) just "scour" would probably be the way to go.

I'm fine with both, though, if anybody else feels strongly about one or the other option.

Biswa96:
OK, I will change this to mingw-w64-scour.

PR #6825, opened 2020-08-19
Update glib2 to 2.65.2

taozuhong:


Alexpux:

PR #6826, opened 2020-08-19
Update vala to 0.49.90

taozuhong:


Alexpux:

PR #6828, opened 2020-08-19
Update gtksourceview4 to 4.7.90

taozuhong:
**Pass to build on my local machine:**
image


Alexpux:

PR #6840, opened 2020-08-20
libgsf: update to 1.14.47

Adsun701:
This updates libgsf to 1.14.47.


Biswa96:
As the patch is from upstream, is it possible to grab the patch from that repository? Like lazka did with fontconfig in previous pull request.

Biswa96:
As the patch is from upstream, is it possible to grab the patch from the [commit][1] in that repository? Like lazka did with fontconfig in previous pull request.

[1]: https://gitlab.gnome.org/GNOME/libgsf/-/commit/8441d5863cb90ed48199d6976bc6e2fcfbe3365a

Issue #6857, opened 2020-08-22
SDL2 static library undefined symbols


rakslice:
The SDL2 static library in the packaged mingw-w64-SDL2-2.0.12-4 has a number of undefined symbols.

FYI SDL2 2.0.12 built and installed from the upstream source on the same system with `./configure && make && make install` results in a static library that does not have this problem.

```
[...]

PR #6859, opened 2020-08-23
SDL2: fix sdl2-config --static-libs order

rakslice:
Fixes msys2/MINGW-packages#6857


lazka:
Could we use the upstream change you linked to instead?

rakslice:
That would be fine. I don't know if that is in a release yet.

rakslice:
Closing in favour of PR #6866

PR #6866, opened 2020-08-24
Add patch based on "Don't duplicate Libs in Libs.private in pkg-confi…

rakslice:
…g file" (SDL hg changeset fc03d19926ed)

Fixes a problem with `sdl2-config --static-libs` libs order #6857


rakslice:
@lazka This is the upstream change I linked to, as a patch

lazka:
increase pkgrel

lazka:
Thanks!

PR #6873, opened 2020-08-26
Add package group mingw-w64-avr-toolchain

fauxpark:
And also just tidying up the PKGBUILDs a little :)


lazka:
hm, on a second thought, shouldn't it be `groups=("=${MINGW_PACKAGE_PREFIX}-${_target}-toolchain")` ?

fauxpark:
Indeed. Give me a second :)

Biswa96:
Also if possible add the sig files and corresponding PGP keys. See packages in mingw-w64-*-toolchain group as reference.

Issue #6216, opened 2020-02-20
msmpi mpic++ 


dschwen:
```
$ mpic++ -v
...
COLLECT_GCC_OPTIONS='-I' 'C:\msys64\mingw64\include' '-v' '-LC:\msys64\mingw64\lib' '-shared-libgcc' '-mtune=generic' '-march=x86-64'
...
```

The `COLLECT_GCC_OPTIONS` output by the msmpi wrapper `mpic++` use backslashes. This causes problems with for example the nonlinear solver package PETSc, which drops the backslashes alltogether during its configure process - as does the shell

```
$ ls -la C:\msys64\mingw64\lib
ls: cannot access 'C:msys64mingw64lib': No such file or directory
```

Biswa96:
msys2 (also cygwin) uses paths forward slashes. For example, here it will be `/c/msys64/mingw64/lib`.

dschwen:
Yeah, Forward slashes would be great. What I'm saying is that `mpic++ -v` returns some arguments with backward slashes in the paths they contain.

lazka:
> The `COLLECT_GCC_OPTIONS` output by the msmpi wrapper `mpic++` use backslashes. This causes problems with for example the nonlinear solver package PETSc, which drops the backslashes alltogether during its configure process - 

Ideally those packages would just support Windows paths. 

> as does the shell

This is not the same, you are missing quotes: `ls -la 'C:\msys64\mingw64\lib'`

MehdiChinoune:
PETSc is now available as a package.

PR #6874, opened 2020-08-26
msmpi: force / as file name separator in command line paths

okhlybov:
Commit is intended to fix issue manifested in #6216


lazka:
Has anyone filed a bug with PETSc to support Windows paths properly?

PR #6878, opened 2020-08-26
binutils: Update to 2.35

lazka:
One small patch refresh. Two patches removed which are now included in the release.


lazka:
build various packages with it, including itself, looks good.

Issue #6674, opened 2020-07-11
gcc: DEP and ASLR not enabled by default


salowenh:
The libgsf is flagged by tools such as [BinSkim](https://github.com/microsoft/binskim) due to apparently not enabled safe exception handlers, DEP and ASLR security options. 
Any thoughts on how to create a more secure compilation?

Biswa96:
mingw-w64 gcc environment has many reputation of being flagged by anti virus programs. Search **"mingw-w64 virus alert"** and you will get many related issues. The alerts are mainly false positive which may not be related to disabled DEP and ASLR.

salowenh:
Unfortunately in this case I believe this is not a false positive as I can't see these flags in any of the binaries. 
Do you have any reason to think otherwise?

lazka:
Last time I looked gcc didn't support DEP and ASLR by default, maybe that has changed in the meantime.

salowenh:
@lazka support was added in gcc 10.1.0, it's indeed disabled by default (i'm not sure why)

mati865:
@salowenh because various LD versions had problem with it.

revelator:
Also gobject-introspection has problems with it, try enabling it and then watch what happens heh.

revelator:
btw you can also pretty much forget using PCH if you enable ASLR and DEP they are incompatible as is.
Im not sure how microsoft got around this problem.

jeremyd2019:
https://sourceware.org/bugzilla/show_bug.cgi?id=19011

I have no idea about GCC, but they're just flags on the PE file, controlled by the linker (ld).  I actually looked at this a bit.  You can turn the options on by default in a gcc spec file, but there is corresponding negative option to turn them back off on a case-by-case basis.  peflags.exe in the rebase package comes in handy in that case :wink:

mati865:
@jeremyd2019 we can enable it in https://github.com/msys2/MSYS2-packages/blob/master/pacman/makepkg_mingw64.conf
Then all repo packages will have it enabled.

lazka:
I've uncommented the LDFLAGS stuff and built ninja as a test:

Before:

```
Get-PESecurity -file C:\msys64\home\user\M\mingw-w64-ninja\pkg\mingw-w64-x86_64-ninja\mingw64\bin\ninja.exe

FileName         : C:\msys64\home\user\M\mingw-w64-ninja\pkg\mingw-w64-x86_64-ninja\mingw64\bin\ninja.exe
[...]

PR #6880, opened 2020-08-27
binutils: move the base addr above 4gb so we get full ASLR support if enabled

lazka:
See #6674


lazka:
@jeremyd2019 @mati865

mati865 (approved):

PR #6884, opened 2020-08-27
WIP: rocksdb: build with jemalloc

EchterAgo:
This needs a fix for rocksdb which I have included as a patch.

See https://github.com/facebook/rocksdb/pull/7320

### DIFF
--- a/mingw-w64-rocksdb/0001-Windows-disable-operator-new-in-static-build-with-je.patch
+++ b/mingw-w64-rocksdb/0001-Windows-disable-operator-new-in-static-build-with-je.patch
@@ -1,0 +1,40 @@
+From c460b7bdf08fa3dc010b92e20c10d3926af94c53 Mon Sep 17 00:00:00 2001
+From: Axel Gembe <derago@gmail.com>
+Date: Thu, 27 Aug 2020 20:45:13 +0700
+Subject: [PATCH] Windows: disable operator new in static build with jemalloc
+ >= 5
+
+Jemalloc 5 added its own operator new / delete which conflict with
+the rocksdb ones when doing a static link.
+
+See https://github.com/jemalloc/jemalloc/releases/tag/5.0.0
+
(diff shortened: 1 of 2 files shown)


Biswa96:
Please increase `pkgrel` variable by 1.

EchterAgo:
@Biswa96 oops, sorry

EchterAgo:
@Biswa96 fixed

lazka:
Any info on why this change is needed/a good idea?

EchterAgo:
Rocksdb benefits from an allocator that prevents heap fragmention. Arch linux enables it by default. That said I just realized that this needs to be improved because it might not work right with dynamic linking. I'll do more testing tomorrow.

EchterAgo:
Ok, so this crashes in the shared build, but the reason is that in a shared build `-lrocksdb -ljemalloc` actually mixes a static `rocksdb` with a dynamic `jemalloc`. Everything works fine when I use `-lrocksdb-shared -ljemalloc`.

AFAICT there should always be 2 libraries, `libreadline.a` for static and `libreadline.dll.a` for shared. In `rocksdb` the static library is `librocksdb.a` and the dynamic library is `librocksdb-shared.dll.a`, causing the linker to see default to the static library. I think the `-shared` part should be removed. I'll open up a separate PR for this.

lazka:
Yeah, looks weird given the shared lib on Linux doesn't have this suffix.

lazka:
I've created #7155

lazka:
@EchterAgo I've rebased this on top of master. Can you re-test for the crashes?

Alexpux:
@EchterAgo @lazka rocksdb updated https://github.com/msys2/MINGW-packages/commit/63392dfe785248a04a217e33a76412b5a2740fa4

lazka:
Someone will need to test this again (I don't know how), otherwise I'll close this.

lazka:
Any plans to continue working on this?

PR #6885, opened 2020-08-27
jemalloc: build without prefix as is default on other systems

EchterAgo:
This allows the library to be statically linked to replace `malloc` and other functions like is possible on other systems.

I don't know why this was made default, does anyone know a good reason? I tested it and it works.

https://github.com/jemalloc/jemalloc/commit/7cdea3973cab8640d1f44c7638ed5e30ed18be24

This is where it was changed, no reason given. I've also checked the changelog and issue tracker and found nothing.

### DIFF
--- a/mingw-w64-jemalloc/PKGBUILD
+++ b/mingw-w64-jemalloc/PKGBUILD
@@ -4,7 +4,7 @@ _realname=jemalloc
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=5.2.1
-pkgrel=1
+pkgrel=2
 pkgdesc="General-purpose scalable concurrent malloc implementation (mingw64)"
 arch=('any')
 url="http://jemalloc.net/"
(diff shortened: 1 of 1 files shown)


lazka:
@Alexpux do you know why?

Alexpux:
@lazka I'm disable build most packages with jemalloc because resulting binaries are always segfaulting for me if builded with jemalloc

lazka:
I don't know how the linker etc handle this, but maybe the prefix is there to avoid mixing with the system malloc? while on linux with LD_PRELOAD you can replace it everywhere... just a guess.

lazka:
isn't rocksdb PR you opened assuming the prefix? I see it in the patch diff at least.

EchterAgo:
@lazka yes, it is interfering when doing a shared build linking against the jemalloc dll, I just found that out now. Without prefix msvcrt will start calling jemalloc, which crashes. Making a separate static build without prefix would work.

The rocksdb PR (#6884) is independent of this, rocksdb works with the jemalloc prefix and without. Rocksdb builds with JEMALLOC_NO_DEMANGLE, which will ensure `je_malloc` are always available.

EchterAgo:
Closed because I think it is not a good idea without having a separate static build.

PR #6890, opened 2020-08-28
New packages: (pt)scotch

okhlybov:


lazka:
Why does it need to be split into two packages?

okhlybov:
scotch is a sequential code, ptscotch is a MPI parallel extension to it

lazka:
Is there a reason these can't be put into one package, like in Arch Linux?

okhlybov:
The joint package will be carrying the msmpi dependency even for those who don't parallel version.

lazka:
msmpi is just 3MB, so that's not a problem

okhlybov:
Modified to build joint package.

lazka:
Thanks!

PR #6892, opened 2020-08-28
cego : update to version 2.45.20

bjlemke:
This is a sync patch to build cego with lfcbase-1.14.4 ( header was modified with 1.14.4 )


bjlemke:
please check, if recompile of lfc dependencies possible ( lfcbase, lfcxml ). seems that header definition change in lfcbase chain class results in linkage problem ...

Von meinem iPhone gesendet

PR #6897, opened 2020-08-29
qrencode: Update to 4.1.0

jannick0:


Biswa96:
Do not add `-j` option in make command. It is selected by user or by MAKEFLAGS variable in `makepkg.conf` configuration file. Also in the `check()` function, it should be:

```sh
cd "${srcdir}\build-${MINGW_CHOST}"
make check || true
```

See other PKGBUILD files as reference.

jannick0:
> Do not add `-j` option in make command. It is selected by user or by MAKEFLAGS variable in `makepkg.conf` configuration file. 

Ok - point taken, whereas the flag is no harm to this package, while it might be to other packages (which could indicate an ill-configured build process). Here I can follow.

> Also in the `check()` function, it should be:
> 
> ```shell
> cd "${srcdir}\build-${MINGW_CHOST}"
> make check || true
> ```
Do not use backslashes in paths in shell scripts when avoidable. ;)
> See other PKGBUILD files as reference.

Why would you do that when package tests do pass (which is the case for this package)?  Separately, tests are there for a reason and should not be considered an annoying extra per se, since a test failure might indicate broken functionality or even worse. As of now tests are not run in CI, locally only.  Do you have any compelling rationale for this policy?

Biswa96:
> Do not use backslashes in paths in shell scripts when avoidable. ;)

Mere typo.

> Do you have any compelling rationale for this policy?

Sometimes tests fail and that does not mean the program has a bug/issue. If tests were failed the whole build process would fail. So, `|| true` or `|| :` does the trick. One could use `makepkg --nocheck` but it avoids all the checks. That's not a policy, it's optional 😊

PR #6899, opened 2020-08-30
gcc: specify LDFLAGS_FOR_TARGET and --with-boot-ldflags

jeremyd2019:
Allows these flags to affect binaries produced post-stage1.

After much fiddling and time spent compiling, this gets the flags applied to **most** binaries.  With what's live right now, the large address aware flag on i686 is a good indicator, as someone previously added the flag to the PKGBUILD, but it never actually took effect later than stage1.

```
$ find . -name \*.exe -o -name \*.dll | peflags -vT - | grep -v bigaddr
./mingw-w64-i686-gcc-ada-10.2.0-2-any/mingw32/bin/libgnarl-10.dll: coff(0x2306[+executable_image,+line_nums_stripped,+32bit_machine,+sepdbg,+dll]) pe(0x0000)
./mingw-w64-i686-gcc-ada-10.2.0-2-any/mingw32/bin/libgnat-10.dll: coff(0x2306[+executable_image,+line_nums_stripped,+32bit_machine,+sepdbg,+dll]) pe(0x0000)
./mingw-w64-i686-gcc-libs-10.2.0-2-any/mingw32/bin/libgcc_s_dw2-1.dll: coff(0x2306[+executable_image,+line_nums_stripped,+32bit_machine,+sepdbg,+dll]) pe(0x0000)
```

I have not yet found a way to get LDFLAGS applied to libgcc*.dll, and I have no idea about ada.

### DIFF
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -243,6 +243,7 @@ build() {
   if [ "$_enable_ada" == "yes" ]; then
     mv ${srcdir}${MINGW_PREFIX}/lib/gcc/${MINGW_CHOST}/${pkgver%%+*}/adalib/*.dll ${srcdir}${MINGW_PREFIX}/bin/
   fi
+  peflags -d0 ${srcdir}${MINGW_PREFIX}/lib/gcc/${MINGW_CHOST}/${pkgver%%+*}/*1*.exe
 }
 
 package_mingw-w64-gcc-libs() {
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -21,7 +21,7 @@ pkgver=10.2.0
 #_majorver=${pkgver:0:1}
 #_sourcedir=${_realname}-${_majorver}-${_snapshot}
 _sourcedir=${_realname}-${pkgver}
-pkgrel=1
+pkgrel=2
 pkgdesc="GCC for the MinGW-w64"
 arch=('any')
 url="https://gcc.gnu.org"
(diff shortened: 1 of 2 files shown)


jeremyd2019:
With the additional patch added in 2727ee3, the binaries are:
```
$ find . -name \*.exe -o -name \*.dll | peflags -vT -| grep -v bigaddr
./mingw-w64-i686-gcc-ada-10.2.0-2-any/mingw32/bin/libgnarl-10.dll: coff(0x2306[+executable_image,+line_nums_stripped,+32bit_machine,+sepdbg,+dll]) pe(0x0000)
./mingw-w64-i686-gcc-ada-10.2.0-2-any/mingw32/bin/libgnat-10.dll: coff(0x2306[+executable_image,+line_nums_stripped,+32bit_machine,+sepdbg,+dll]) pe(0x0000)
```

That looks good enough to me, unless any ada fan wants to try to figure out libgnat and libgnarl

mingw-w64-gcc/PKGBUILD
```diff
@@ -219,6 +224,8 @@ build() {
     --with-pkgversion="Rev${pkgrel}, Built by MSYS2 project" \
     --with-bugurl="https://github.com/msys2/MINGW-packages/issues" \
     --with-gnu-as --with-gnu-ld \
+    --with-boot-ldflags="${LDFLAGS} -static-libstdc++ -static-libgcc" \
```
mati865, line 227: Why are you linking libstdc++ and GCC libs statically?

jeremyd2019: > --with-boot-ldflags=flags
> This option may be used to set linker flags to be used when linking stage 2 and later when bootstrapping GCC. If –with- boot-libs is not is set to a value, then the default is ‘-static-libstdc++ -static-libgcc’.

mati865: Alright then.

jeremyd2019:
Pretty clear this is not sufficient, closing in favor of #6907

PR #6902, opened 2020-08-30
binutils: add upstream patches changing PE defaults.

jeremyd2019:
I don't know that this is desired, but I wanted to get this out there in draft form because there seems to be some desire to test with ASLR on by default in binutils in #6674.

### DIFF
--- a/mingw-w64-binutils/1001-Change-the-default-characteristics-of-DLLs-built-by-.patch
+++ b/mingw-w64-binutils/1001-Change-the-default-characteristics-of-DLLs-built-by-.patch
@@ -1,0 +1,611 @@
+From 3021170f1728200d1f08c2a57af625f7a1003097 Mon Sep 17 00:00:00 2001
+From: Jeremy Drake <sourceware-bugzilla@jdrake.com>
+Date: Thu, 27 Aug 2020 12:58:27 +0100
+Subject: [PATCH 1/3] Change the default characteristics of DLLs built by the
+ linker to more secure settings.
+
+	PR 19011
+	* emultempl/pe.em (DEFAULT_DLL_CHARACTERISTICS): Define.
+	(pe_dll_characteristics): Initialise to DEFAULT_DLL_CHARACTERISTICS.
+	(add_options): Add options to disable DLL characteristics.
+	(list_options): List the new options.
+	(handle_options): Handle the new options.
+	* emultempl/pep.em: Similar changes to above.
+	(NT_EXE_IMAGE_BASE): Default to an address above 4G.
+	(NT_DLL_IMAGE_BASE, NT_DLL_AUTO_IMAGE_BASE,
+	(NT_DLL_AUTO_IMAGE_MASK): Likewise.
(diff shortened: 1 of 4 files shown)


lazka:
* There has been one more commit upstream: https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=6194b866b7a89969d8c66f8a97d40acc028373b1
* If we'd add a commit reverting the default changes we could get the --disable flags which we could start using (does `LDFLAGS=--high-entropy-va; LDFLAGS+=--disable-high-entropy-va` work as I think it does?). I'd still like to see more packages getting build first before enabling this for everyone.

1480c1:
> * If we'd add a commit reverting the default changes we could get the --disable flags which we could start using (does `LDFLAGS=--high-entropy-va; LDFLAGS+=--disable-high-entropy-va` work as I think it does?). I'd still like to see more packages getting build first before enabling this for everyone.

I'm not sure this would work since it would result in 
```bash
LDFLAGS=--high-entropy-va--disable-high-entropy-va
```

would probably instead need
```bash
LDFLAGS=--high-entropy-va; LDFLAGS+=" --disable-high-entropy-va"
```

jeremyd2019:
> * There has been one more commit upstream: https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=6194b866b7a89969d8c66f8a97d40acc028373b1 Not sure if relevant
That should be 1003-Fixes-for-testsuite-failures-introduced-by-the-chang.patch

> * If we'd add a commit reverting the default changes we could get the --disable flags which we could start using (does `LDFLAGS=--high-entropy-va; LDFLAGS+=--disable-high-entropy-va` work as I think it does?).

It should be that the last one wins.

>  I'd still like to see more packages getting build first before enabling this for everyone.

Me too, which is why this is a draft, and why I had originally made 2 patches on the upstream bug, one adding the disable flags and one changing the defaults.

jeremyd2019:
I had made this branch for my own use testing, but after seeing someone going back to the old patches, I thought it'd be a good idea to make it more accessible (and as a side benefit, can grab binaries from the CI artifacts)

lazka:
> > * There has been one more commit upstream: https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=6194b866b7a89969d8c66f8a97d40acc028373b1 Not sure if relevant
> 
> That should be 1003-Fixes-for-testsuite-failures-introduced-by-the-chang.patch

ah, I missed that.

> > * If we'd add a commit reverting the default changes we could get the --disable flags which we could start using (does `LDFLAGS=--high-entropy-va; LDFLAGS+=--disable-high-entropy-va` work as I think it does?).
> 
> It should be that the last one wins.

ok

> > I'd still like to see more packages getting build first before enabling this for everyone.
> 
> Me too, which is why this is a draft, and why I had originally made 2 patches on the upstream bug, one adding the disable flags and one changing the defaults.

I'd be happy to merge this now if DEFAULT_DLL_CHARACTERISTICS was set to 0x0

PR #6905, opened 2020-08-31
Update arm-none-eabi-binutils to 2.35

fauxpark:
Technically a new package, as it's not actually in the pacman repo.

### DIFF
--- a/mingw-w64-arm-none-eabi-binutils/PKGBUILD
+++ b/mingw-w64-arm-none-eabi-binutils/PKGBUILD
@@ -1,4 +1,5 @@
 # Maintainer: Carlos Antunes <cmantunes@gmail.com>
+# Maintainer: fauxpark <fauxpark@gmail.com>
 # Contributed by: nanodude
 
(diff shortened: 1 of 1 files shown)


Biswa96:
There should be separate build directory for x86_64 and i686. So, don't remove the commands with `build-${MINGW_CHOST}` directory name.

fauxpark:
What for? `${srcdir}` gets nuked each time anyway.

Biswa96:
It is not nuked with `makepkg-mingw` which is useful to create x86_64 and i686 package in one shot. Also separate build folder is useful to repackage (i.e. `makepkg-mingw -R`) if something went wrong.

fauxpark:
> It is not nuked with makepkg-mingw which is useful to create x86_64 and i686 package in one shot.

I always use the command specified here: https://www.msys2.org/wiki/Creating-Packages/
`makepkg-mingw -sCLf` which cleans the srcdir before building. Fair point on repackaging, though.

PR #6906, opened 2020-08-31
binutils: add upstream patches but revert DEFAULT_DLL_CHARACTERISTICS to 0

jeremyd2019:
As requested in https://github.com/msys2/MINGW-packages/pull/6902#issuecomment-683582287 here are those changes with an additional patch setting the DEFAULT_DLL_CHARACTERISTICS to 0.  This still leaves the change in default --enable-reloc-section.

I haven't tested this - my machine is still busy testing builds with the flags enabled, but CI builds passed.


lazka:
I built Python with it and the new LDFLAGS, confirmed that it crashed, added `LDFLAGS+=" -Wl,--disable-high-entropy-va"` and confirmed that it built fine.

PR #6909, opened 2020-09-01
exiv2: Update to 0.27.3

kmilos:


lazka:
Why did you remove the tests?

kmilos:
GHA doesn't run check() any more AFAICT from ci-build.sh, so they're irrelevant, no?

lazka:
The maintainer can run it though (CI never ran them afaik). I don't mind much.. just wondering though

kmilos:
Besides, it seems there is no more `tests` target generated by cmake in the present configuration (EXIV2_BUILD_SAMPLES is OFF for msys2), so the maintainer would have to rebuild anyway with other flags to enable the tests...

kmilos:
With EXIV2_BUILD_SAMPLES=ON I verified that the new target `bash_tests` passes without the old patch.

But there is now a problem with (new?) python_tests, but that's a different story I guess...

kmilos:
For completeness: `EXIV_EXT=.exe make python_tests` is ok as well (env variable needed for msys2, as documented upsteram)

kmilos:
EXIV2_ENABLE_CURL was gated by EXIV2_ENABLE_WEBREADY

SSH (using libssh, not libssh2) was also never really switched on, now removing as it seems it's dropped upstream as well

lazka:
Since it seems not that straight forward to run test, removing is ok. thanks for the update!

PR #6910, opened 2020-09-01
capnproto: replace autoconf with cmake

Biswa96:


mingw-w64-capnproto/PKGBUILD
```diff
@@ -27,19 +29,34 @@ build() {
...
-    --prefix=${MINGW_PREFIX} \
-    --build=${MINGW_CHOST} \
-    --host=${MINGW_CHOST} \
-    --target=${MINGW_CHOST} \
-    --disable-shared \
-    --enable-static \
-    --with-openssl
+MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  ${MINGW_PREFIX}/bin/cmake.exe \
+    -G "MSYS Makefiles" \
```
1480c1, line 34: Prefer Ninja due to speed and also bc MSYS Makefiles is for msys, and not msys2

mingw-w64-capnproto/PKGBUILD
```diff
@@ -27,19 +29,34 @@ build() {
...
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=ON \
```
1480c1, line 37: The original used `--disable-shared` and `--enable-static`, would it be preferred to use both static and shared?

mingw-w64-capnproto/PKGBUILD
```diff
@@ -27,19 +29,34 @@ build() {
...
   mkdir -p "${srcdir}/build-${CARCH}" && cd "${srcdir}/build-${CARCH}"
 
-  ../capnproto-c++-${pkgver}/configure \
-    --prefix=${MINGW_PREFIX} \
-    --build=${MINGW_CHOST} \
-    --host=${MINGW_CHOST} \
-    --target=${MINGW_CHOST} \
-    --disable-shared \
-    --enable-static \
-    --with-openssl
```
1480c1, line 37: Is openssl not used in the cmake build?

mingw-w64-capnproto/PKGBUILD
Biswa96: `BUILD_SHARED_LIBS` enables both static and shared.

mingw-w64-capnproto/PKGBUILD
Biswa96: Nope.

mingw-w64-capnproto/PKGBUILD
Biswa96: I choose Ninja when the project is big. This is a small one, speed does not matter here.

Biswa96:
@HazardyKnusperkeks Would you like to test the artifacts from CI builds? The executables does not contain any specific paths now.

mingw-w64-capnproto/PKGBUILD
1480c1: I see, so they have it setup to generate both

HazardyKnusperkeks:
I will test it tomorrow.

HazardyKnusperkeks:
Now I get an other error. I'm using QtCreator with qbs and the capnproto example shipped with it.
For the line ``using Cxx = import "/capnp/c++.capnp";`` the error is: ``addressbook.capnp:24:20-38: error: Import failed: /capnp/c++.capnp``

But if I add manually `` -I /mingw64/include/`` to the call it succeeds.
So the include path seems to be needed in the binary, but it shouldn't be the windows absolute one.
But this one works only from within the msys shell.

Biswa96:
> For the line using Cxx = import "/capnp/c++.capnp"; the error is: addressbook.capnp:24:20-38: error: Import failed: /capnp/c++.capnp

Is that issue related to your build system or capnproto? For example, if I execute `pkg-config.exe --cflags capnp` in mingw64 environment it returns correct include directory `-pthread -IF:/msys64/mingw64/include`.

Biswa96:
> For the line using Cxx = import "/capnp/c++.capnp"; the error is: addressbook.capnp:24:20-38: error: Import failed: /capnp/c++.capnp

Is that issue related to your build system or capnproto? For example, if I execute `pkg-config.exe --cflags capnp` in mingw64 environment it returns correct include directory `-pthread -IF:/msys64/mingw64/include`. Do you have any example .capnp file to test on instead of installing whole QtCreator?

Biswa96:
> But if I add manually -I /mingw64/include/ to the call it succeeds.

I thinks that is the norm. According to the docs https://capnproto.org/cxx.html, "If you installed somewhere else, you may need to add it to the search path with the -I flag to capnp compile, which works much like the compiler flag of the same name." So, you have to add that path manually in the command line.

HazardyKnusperkeks:
That seems to be correct. But it would be nice if it could refer to ``/mingw64/include`` or ``/mingw32/include`` out of the box, otherwise you have to add these paths in the project files, which is overly complicated in my opinion. Because the paths do (most likely) not exist on linux.

Biswa96:
Your comment gives me some idea, hacking capnp...

Biswa96:
Can this change be added? At least, this change makes capnproto somewhat usable. I have opened an issue in upstream about the include directory which may take some time and better be in next version.

Biswa96:
ping @lazka.

PR #6913, opened 2020-09-02
New package: mumps

okhlybov:
Introduce a package for the MUMPS SLAE solver.
This package builds sequential, OpenMP and MPI library versions for all scalar types [sdcz].


Biswa96:
Is it possible to add the pkg-config (.pc) files as a separate source file instead of doing `echo >`?

lazka:
> Is it possible to add the pkg-config (.pc) files as a separate source file instead of doing `echo >`?

it looks like there is a lot of templating going on, so might not be that easy, right?

okhlybov:
Yes, there are many .pc variants to create, so it is better to auto-generate.

PR #6919, opened 2020-09-03
lfcxml + cego : update to lfcxml-1.2.11 and cego-2.45.21

bjlemke:


Biswa96:
Two different commits would be preferable.

Issue #6922, opened 2020-09-03
need to be rebuilt for libraw-20


sdegrande:
Libraw has been updated to version 0.20 on 2020-08-01

Freeimage tries to load libraw-19, and obviously fails.

Could you please rebuild freeimage ?

Thanks for all your great work !

lazka:
I tried in #6924, but freeimage doesn't build successfully right now.

Biswa96:
Try to add the patch file in ArchLinux for freeimage due to libraw update. See here https://github.com/archlinux/svntogit-community/blob/packages/freeimage/trunk/freeimage-libraw-0.20.patch

lazka:
should be in the next upload

sdegrande:
Great ! Thanks a lot.

sdegrande:
I just updated my env.
It works well. Thanks !

PR #6924, opened 2020-09-03
Rebuilds for libraw

lazka:
See #6922


lazka:
I removed gegl again because it fails to build currently `Gegl-0.4.gir:1770.59-1770.59: error: The symbol `Babl' could not be found` likely related to https://github.com/mesonbuild/meson/issues/7628

PR #6925, opened 2020-09-03
python: rebuild without high-entropy-va

lazka:
It crashes the build otherwise


jeremyd2019:
high-entropy-va is only available on x86_64 target, so i686 is failing due to not knowing the flag.

lazka:
thanks!

Biswa96:
Though I'm interested in why the build fails. Shouldn't this be reported to python?

mati865:
It should be reported to binutils.

lazka:
I've opened #6931 so we don't forget

PR #6926, opened 2020-09-03
Update mingw-w64-mesa to 20.1.7 and add swr driver to MINGW64 package

pal1000:


Biswa96:
Just asking. Shouldn't it be `-Dc_args=${CFLAGS} -Dcpp_args=${CXXFLAGS}` as those are configured in `makepkg.conf` by user? Also it looks like the backslashes were removed accidentally.

mingw-w64-mesa/PKGBUILD
```diff
@@ -74,23 +81,30 @@ prepare() {
...
-  -Dc_link_args='-s' \
-  -Dcpp_link_args='-s' \
-  -Dshared-glapi=true \
-  -Dgles1=true \
-  -Dgles2=true \
-  -Dllvm=true \
-  -Dosmesa=gallium \
-  -Dbuild-tests=true
+  buildconf="${MINGW_PREFIX}/bin/meson ./build/windows-${_mach}
+  --prefix=${MINGW_PREFIX}
```
lazka, line 85: missing all the backslashes?

Biswa96:
Just asking. Are those `*_args` option required? Aren't those flags configured using `makepkg.conf`?

mingw-w64-mesa/PKGBUILD
pal1000: I had to drop them to add that conditional string concatenation for swr. I have to figure something out, hopefully without relying on CFLAGS, CXXFLAGS and LDFLAGS.

pal1000:
> 
> 
> Just asking. Are those `*_args` option required? Aren't those flags configured using `makepkg.conf`?

Some of them provide performance benefit around 2-5%

pal1000:
Found a solution. I moved the args at the end of build config command which allowed me to put back the backslashes for them.

Issue #6758, opened 2020-08-05
Please upgrade rust to 1.45.2


abcfy2:
New version 1.45.2 has been released: https://blog.rust-lang.org/2020/08/03/Rust-1.45.2.html. Please upgrade.

omerakgoz34:
1.46.0 :'(
Why it's still out-dated?

1480c1:
probably because nobody has submitted an update pr and the current packagers are still busy trying to update the other packages as well

1480c1:
If you want to help, you can edit this file in a fork and open a pr for it https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-rust/PKGBUILD

The main changes should be to the pkgver and checking if the patch files are still needed, or if they need to be updated or if new patches need to be applied if something else broke. The final change of sha256sums can be autoupdated using updpkgsums 

@lazka or @Biswa96 can probably comment more since they know more about PKGBUILDS than I do

lazka:
Someone needs to do the work.

abcfy2:
I find rust support compile on MSYS2: https://github.com/rust-lang/rust#mingw

But I can't compile on my PC, anyone can help?

```shell
  ./x.py build
downloading https://static.rust-lang.org/dist/2020-08-03/rust-std-1.45.2-x86_64-pc-windows-gnu.tar.xz
extracting C:/Users/abcfy/projects/MINGW-packages/mingw-w64-rust/rustc-1.46.0-src/build/cache/2020-08-03/rust-std-1.45.2-x86_64-pc-windows-gnu.tar.xz
Traceback (most recent call last):
  File "./x.py", line 11, in <module>
    bootstrap.main()
  File "C:/Users/abcfy/projects/MINGW-packages/mingw-w64-rust/rustc-1.46.0-src/src/bootstrap/bootstrap.py", line 981, in main
    bootstrap(help_triggered)
  File "C:/Users/abcfy/projects/MINGW-packages/mingw-w64-rust/rustc-1.46.0-src/src/bootstrap/bootstrap.py", line 945, in bootstrap
    build.download_stage0()
  File "C:/Users/abcfy/projects/MINGW-packages/mingw-w64-rust/rustc-1.46.0-src/src/bootstrap/bootstrap.py", line 386, in download_stage0
    self._download_stage0_helper(filename, pattern, tarball_suffix)
[...]

PR #6932, opened 2020-09-04
Update Rust to 1.48

mati865:
Closes https://github.com/msys2/MINGW-packages/issues/6758


Biswa96:
Following the trend, I think Python commands are just `python` not `python3`.

mati865:
I can reproduce 64-bit `(exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)` locally with latest packages, trying to fix it.

abcfy2:
Hope this build failing would be fixed.

mati865:
I believe I've got the fix (still testing it to be sure).
Before https://github.com/rust-lang/rust/pull/72049 Rust was relying on the linker "to make things work". That means linker to guess dllimport/dllexport and in case it fails the linker was using runtime pseudo relocs as backup.
I haven't full investigated it but I think the recent changes to Binutils (namely increasing base address over 4G for 64-bit binaries) made runtime pseudo relocs that have 32-bit offsets to overflow and hence trigger the DEP.
That would also explain why i686 build had no issues.

@lazka I believe that would explain some of those cases:
> So far all software that also supports building with MSVC works fine (glib,gtk3,pygobject for example) so I'm wondering if this isn't just exposes some latent bugs for things only built with mingw so far. Agreed on the example search..

lazka:
So we should pass `-Wl,--no-dynamicbase,--image-base=0x400000` to revert this for rust for now?

mati865:
@lazka that would be complicated since executables and DLLs use different base addresses.
I'm backporting https://github.com/rust-lang/rust/pull/72049 and it already passed with `export LDFLAGS="-pipe"`. I'm testing default LDFLAGS now but since https://github.com/rust-lang/rust/pull/75406 works for me locally I believe this will also succeed.

jeremyd2019:
> @lazka that would be complicated since executables and DLLs use different base addresses.

Yes!  I was considering sending another patch for binutils to allow selecting whether the default bases should be "high" or "low" via command line option.  But I am not sold on that being a good idea.  It would make hacking around buggy programs a lot easier though.  Thoughts?

mati865:
> It would make hacking around buggy programs a lot easier though. Thoughts?

It would definitely help with the transition.

jeremyd2019:
> It would definitely help with the transition.

I made a patch and am testing it now (simple tests worked as intended, going to attempt building GCC with it).  I don't know if a 'transitional' option warrants upstreaming though.

lazka:
New binutils with `--default-image-base-low` option is now in the repo.

mati865:
I think using https://github.com/rust-lang/rust/pull/72049 instead of `--default-image-base-low` is better but it's up to you.

lazka:
a real upstream fix/backport is obviously better

mati865:
I have no idea why CI fails on x86_64 build, lets try again.

lazka:
Not sure if relevant, but it finds the wrong Python: `-- Found Python3: C:/hostedtoolcache/windows/Python/3.8.6/x86/python.exe (found version "3.8.6") found components: Interpreter `

mati865:
Blocked on https://github.com/msys2/MINGW-packages/pull/7116

mati865:
CI now appears to be failing due to path length but it's kinda hard to tell.

jeremyd2019:
Or command line length limit, by the look of that ar command... I copy/pasted it and saw 36618 characters there!

mati865:
I think it should have used response file but yeah sounds plausible.

filnet:
Hi,

What's holding up the rust upgrade ? A hard to solve technical problem or just lack of time from the maintainers ?
Can I help in any way ?

mati865:
Local build succeeds every time for me.
I haven't got to dive deeply into the issue yet and since the next Rust version will be released this week I just decided to wait.

Zapeth:
Looking at the build log it seems the failed ar command originates from compiling [libgit2-sys](https://github.com/rust-lang/git2-rs/tree/master/libgit2-sys) for cargo.

I haven't heard yet of command line length being an issue when compiling any kind of software, but path length could certainly be an issue (some of those paths get pretty close to the Windows [MAX_PATH](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) length, and i686 config might just fall under that limit), though I'd be curious why this wasn't an issue with earlier package releases.

In any case, 1.48 was released yesterday, so it might be worth trying a new ci build with that.

mati865:
> I haven't heard yet of command line length being an issue when compiling any kind of software, but path length could certainly be an issue

IIRC I have inspected the longest paths and they were under the limit. Command line limit is definitely a thing on Windows but it depends on how you run the executable so I won't even try to guess what is the limit in this case.

mati865:
I have no idea why it fails with `rustc-1.48.0-src.tar.gz ... FAILED (unknown public key 85AB96E6FA1BE5FE)` on CI.
The build passed for me locally before pushing.

lazka:
It means the release was signed with an ancient gnupg which doesn't include the full key in the signature. You can remove the signature to avoid it.

Zapeth:
Looks like the same issue. I did some searching online and found this -> https://github.com/alexcrichton/cc-rs/pull/496

lazka:
nice :) should I merge?

mati865:
I have made basic tests with x86_64 before pushing and it worked so I guess it's good to go.

lazka:
built librsvg with it and all looks good.

a8nJ:
So how do we get the new version?

Biswa96:
Just wait and [watch](https://packages.msys2.org/updates).

PR #6936, opened 2020-09-05
pidgin++: remove

lazka:
Depends on gtk2/python2 and is unmaintained upstream for 5-6 years. https://launchpad.net/pidgin++


lazka:
If someone wants it back just open an issue and we can look into it.

PR #6939, opened 2020-09-05
Update run-time dependencies

jannick0:
This is a byproduct of some work for automated CI dependency checks.


Biswa96:
freeimage change is done recently. It can be removed here.

jannick0:
Done ... thanks for checking. I did miss that.

PR #6945, opened 2020-09-06
Update arm-none-eabi-gcc to 8.4.0

fauxpark:


Fabien-Chouteau:
Hi @fauxpark, do you plan to update to GCC 10.2?

fauxpark:
Hopefully yes, however (and I quote from @zvecr) "it has a ton of issues, mostly due to defaulting to -fno-common and lto bad order stuff". I'm not sure exactly what the problems or ramifications are.

We at QMK have also observed somewhat larger filesizes when compiling for the AVR target on GCC 9 and 10, but given ARM microcontrollers are generally much more generous with flash space, this is probably not too big a deal. I also don't know if this issue is specific to AVR so it may need testing, unless you can confirm. I suspect it's related to it still using [CC0](https://gcc.gnu.org/wiki/CC0Transition), whereas `arm-*-*` uses MODE_CC like most other targets.

PR #6955, opened 2020-09-07
Fix libwinpthread ignoring buildflags

### LINKED ISSUE #6953: libwinpthread-1.dll under mingw-w64-winpthreads-git is missing /dynamicbase after built with --dynamicbase
Using the latest msys (20200903)
Built `mingw-w64-winpthreads-git` for both x86 & x64, verified `--dynamicbase` appears in LDFLAGS.
`/dynamicbase` is missing from `libwinpthread-1.dll` (x64 & x86)

`makepkg_mingw64.conf`:
```LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat"```

### PR DESCRIPTION
**Pull request by salowenh**
Fixes #6953

### DIFF
--- a/mingw-w64-winpthreads-git/PKGBUILD
+++ b/mingw-w64-winpthreads-git/PKGBUILD
@@ -6,7 +6,7 @@ _realname=winpthreads
 pkgbase=mingw-w64-${_realname}-git
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-git" "${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
 pkgver=8.0.0.6001.98dad1fe
-pkgrel=1
+pkgrel=2
 _commit='98dad1fefed2a07ade0b37ca6c610ad66739631a'
 pkgdesc="MinGW-w64 winpthreads library"
 url="https://mingw-w64.sourceforge.io/"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by salowenh** [comment]
#6953 6953

**Comment by lazka** [comment]
you need to increase pkgrel

File: mingw-w64-winpthreads-git/PKGBUILD [hunk 1]
```diff
@@ -20,7 +20,7 @@ makedepends=("git"
              "${MINGW_PACKAGE_PREFIX}-binutils"
              "${MINGW_PACKAGE_PREFIX}-crt-git"
              "${MINGW_PACKAGE_PREFIX}-headers-git")
-options=('strip' '!buildflags' 'staticlibs' '!emptydirs' '!debug')
+options=('strip' 'buildflags' 'staticlibs' '!emptydirs' '!debug')
```
mati865, line 23:
```suggestion
options=('strip' 'staticlibs' '!emptydirs' '!debug')
```

Biswa96:
Why does `pkgrel` need to be changed? This is a git package. So, `pkgver` will be increased to latest commit.

lazka:
good point. I wonder how we are going to automatically build such packages in the future

Biswa96:
> I wonder how we are going to automatically build such packages in the future.

Maybe CI can automatically change the `pkgver` to the latest one. Also the `pkgver()` function automatically updates the `pkgver`. So, CI can grab it and retain the changed value. Otherwise one has to clone the entire mingw-w64 git repository.

lazka:
Problem is that if the package gets built on the next day it might be different to what was tested the day before by the user.

Biswa96:
> Problem is that if the package gets built on the next day it might be different

In case of mingw-w64, it is not so fast moving repository. So, `$pkgver` will not changed, maybe even in months 🤪

>I kinda tend towards just removing the -git suffix and adding a hash.

 As you wish. It's not so different. Though I am kind of inclined towards `pkgver()` things. It helps me to test new mingw-w64 commits without touching PKGBUILD file, lazy to run commands manually.

mati865:
TBH I don't understand why mingw-w64 packages use latest commit instead of specific commit hash like described at https://www.archlinux.org/pacman/PKGBUILD.5.html#VCS

lazka:
> TBH I don't understand why mingw-w64 packages use latest commit instead of specific commit hash like described at https://www.archlinux.org/pacman/PKGBUILD.5.html#VCS

This is now done.

Someone said that `!buildflags` might have been added to avoid flags adding extra toolchain related dependencies, but let's see...

PR #6960, opened 2020-09-08
gstreamer: Update to 1.18.0

lazka:


Biswa96:
What is the error? Any pkgbuild sample?

lazka:
fixed it :)

like still dependencies missing..

Biswa96:
`glslc` program is provided by `shaderc` package in ArchLinux, see https://www.archlinux.org/packages/extra/x86_64/shaderc/

lazka:
I've created #6961

PR #6963, opened 2020-09-09
Add package Scintilla 4.4.4(GTK3)

taozuhong:


taozuhong:
@lazka could this package can merge into repo?

lazka:
Once CI passes, sure.

Biswa96:
Some queries:
1. Wouldn't the libraries be with .DLL extension instead of .SO?
2. The prefix is hard coded in pkgconfig file. Shouldn't it be `${prefix}`?

Biswa96:
Some queries:
1. Wouldn't the libraries be with .DLL extension instead of .SO?
2. The prefix is hard coded in pkgconfig file. It should be `/mingw32` for i686 packages.
3. `${MINGW_PREFIX}` already has a forwardslash in the value. So, the forwardslashes can be removed from commands.

taozuhong:
@Biswa96 Could you give me some tips about make scintilla.pc more portable?

taozuhong:
@Biswa96 done，thank for your reviewing and advice.

fauxpark:
Just a note that scintilla.pc somehow has Windows line endings and is showing up as modified in git now.

PR #6964, opened 2020-09-09
mesa-demos: add package

Biswa96:


Biswa96:
@pal1000 Every mesa demos program crash with opengl32.dll provided by mingw-w64-mesa package but does not with Sytem32/opengl32.dll. Is it necessary to provide opengl32.dll in mesa package? Can it be removed?

Biswa96:
Now what's the issue with 32bit? 🤔

Biswa96:
I think something wrong with the autoconf files. I am testing something and will let you know...

Biswa96:
Some results of my insignificant tests.

1. With autoconf, x86_64 builds and works fine with or without ASLR ✔️ But the full path of /share/mesa-demos files are hard-coded. ❌
2. With autoconf, i686 fails to test openl32 and glu32 library. ❌ If library checks are disable only osdemos*.exe are built ❌
3. With latest main branch of mesa-demos, both x86_64 and i686 can be built. ✔️ But the install directories are not correct i.e. need GNUInstallDirs. ❌

If possible I will discuss with upstream repo.

Biswa96:
Many issues to be fixed. Discussing with upstream.

PR #6966, opened 2020-09-09
graphite2: fix pkgconfig generation by cmake

Biswa96:


Biswa96:
@sheaf Would you like to test the artifacts from CI builds?

sheaf:
@Biswa96 Just tried it, works great on my end, thanks a lot!

PR #6967, opened 2020-09-09
gdb: apply upstream patch for rebasing aslr exes.

### LINKED ISSUE #6941: gdb does no work with aslr executables. Specifically, seems to assume exe loaded at its preferred base
Hi, when I compiling qemu with newest mingw-w64-binutils 2.35-4, I can running the program, but can not debug it, what's the possible issue?
And once I downgrading to mingw-w64-x86_64-binutils-2.33.1-1-any.pkg.tar.xz, it's can debugging

jeremyd2019:
Fixes #6941

Patch had to be adjusted to apply on 9.2.  It looks like more goodies for windows were added upstream than just this.

### DIFF
--- a/mingw-w64-gdb/PKGBUILD
+++ b/mingw-w64-gdb/PKGBUILD
@@ -5,7 +5,7 @@ _realname=gdb
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=9.2
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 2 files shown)


lygstate:
great work

Biswa96:
Off topic. In the patch the offset values are hard coded. But why it was not calculated from TEB itself?

lazka:
> Off topic. In the patch the offset values are hard coded. But why it was not calculated from TEB itself?

aren't those the offsets where the actual values are read from?

PR #6971, opened 2020-09-10
binutils: fix crash creating empty implib.

### LINKED ISSUE #6107: any ld.exe from binutils > v2.30 is broken
Hello,
I found that any `ld.exe` from binutils > v2.30 is broken and returns **ld returned 5 exit status** error code. So I have tried to download `mingw-w64-i686-binutils-2.30-6-any.pkg.tar.xz` from msys2 repo http://repo.msys2.org and replaced the installed `ld.exe` with the downloaded one, this times it compiled and linked without problems.

My code is very big to share it so you can reproduce it yourself. My code requires at least **wxWidgets v3.1.3** to build.

What can I do make you believe that any version > v2.30 is really broken!? Because it is working without problems for small projects and makes you believe it is working properly.

Thank you in advance.

jeremyd2019:
Fixes #6107 (most of the comments at least, the original report didn't give much to go on).

Also, added another upstream test patch, and turned my last patch into a
git-format one.

>> Just wondering. As we are importing many patches from upstream, how about using a git commit as source tarball for binutils?
>
> Are you talking about an upstream git commit, with whatever other changes might be pending for 2.36, or a branch on a fork with the desired patches cherry-picked on top of 2.35?

To maintain these patch series, I've got 2 branches going atm.  https://github.com/jeremyd2019/binutils-gdb/tree/backport-binutils-2_35 has upstream commits cherry-picked onto binutils 2.35 release tag.  https://github.com/jeremyd2019/binutils-gdb/tree/backport-binutils-2_35-revert-defaults is based on that branch but has additional patches not upstreamed (kind of unfortunately named based on the first of those patches).

### DIFF
--- a/mingw-w64-binutils/1001-Change-the-default-characteristics-of-DLLs-built-by-.patch
+++ b/mingw-w64-binutils/1001-Change-the-default-characteristics-of-DLLs-built-by-.patch
@@ -1,7 +1,7 @@
 From 3021170f1728200d1f08c2a57af625f7a1003097 Mon Sep 17 00:00:00 2001
 From: Jeremy Drake <sourceware-bugzilla@jdrake.com>
 Date: Thu, 27 Aug 2020 12:58:27 +0100
-Subject: [PATCH 1/3] Change the default characteristics of DLLs built by the
+Subject: [PATCH 1/4] Change the default characteristics of DLLs built by the
  linker to more secure settings.
 
 	PR 19011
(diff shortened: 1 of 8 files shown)


jeremyd2019:
> https://sourceware.org/bugzilla/show_bug.cgi?id=26588

My patch has been committed upstream.  The only difference from what's in this pull request is a changelog entry, so I don't see a need to make a new pkgrel with the 'official' patch, but can if desired.

lazka:
The patch likely makes the meson test suite fail: https://dev.azure.com/jussi0947/jussi/_build/results?buildId=7722&view=logs&j=73ea1edc-10d0-5f82-b73d-0a056f907b1d&t=cae5373a-8189-5a71-1e91-a2a64ed702df&l=689

Looks like it expects the importlib to be written

@jeremyd2019 any ideas?

jeremyd2019:
The condition in generate_edata was
```
  if (pe_def_file->num_exports == 0 && !bfd_link_pic (info))
```
so I'm guessing it should have been
```
      if (command_line.out_implib_filename
          && (pe_def_file->num_exports != 0
              || bfd_link_pic (&link_info)))
```

But that's getting pretty gross as far as re-checking most of the conditions of the outer if just because 
`|| pe_dll_enable_reloc_section` was added to it

PR #6973, opened 2020-09-11
Add cppgc include directory to instalation

Kwizatz:
Missed a new include directory from the installation.


Kwizatz:
Moving to live review since failures do not seem related to build.

Biswa96:
Can the `install -d` commands be removed? Doesn't `install -D` create the destination directory automatically?

Kwizatz:
@Biswa96 Looks like we do need those:
https://github.com/msys2/MINGW-packages/runs/1108785185

Biswa96:
OK, as wish. I was just asking.

Kwizatz:
> OK, as wish. I was just asking.

No problem 😁 it would had been nice if it did worked.

Issue #6961, opened 2020-09-08
package request: shaderc


lazka:
https://github.com/msys2/MINGW-packages/pull/6960#issuecomment-689103004

Required for vulkan support in gstreamer

Biswa96:
shaderc has some sort of weird dependency with glslang and spirv-tools. shaderc releases are in between releases of glslang. If I compile a specific version of shaderc either I got compiler error or linking error. shaderc has a DEPS file which mentions the git commit of glslang and spirv-tools but those commits are not the stable ones.

lazka:
ok, thanks for looking into it

PR #6985, opened 2020-09-14
shaderc: add package, required for gst-plugins-bad

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/6961


Biswa96:
Wait, I am doing the gst-plugin thing.

PR #6989, opened 2020-09-15
avisynthplus: add package

Biswa96:


Biswa96:
The file conflict issue that I wanted to fix causes the build failure. It looks like there is a circular dependency among x264-ffms2-ffmpeg. The packages may need to be built manually. /cc @lazka @1480c1

1480c1:
In mabs, we build a "light" version of ffmpeg with a small subset of libraries just for x264

qyot27:
> It looks like there is a circular dependency among x264-ffms2-ffmpeg.

Only from a certain point of view.  x264 the ***CLI tool*** is what can optionally use AviSynth, FFMS2, and LAVF as inputs.  FFmpeg only needs ***libx264***, which doesn't depend on any of those.  And what x264's LAVF and FFMS2 inputs need are an FFmpeg that only contains demuxers and decoders, not encoders and muxers.

```
libx264->FFmpeg (--enable-libx264)
AviSynth+ headers->FFmpeg (--enable-avisynth)
```
or
```
minimal FFmpeg->FFMS2->x264 CLI autodetects both LAVF and FFMS2 inputs,
                       AVS input is always automatically enabled because x264 ships a local copy of avisynth_c.h
[...]
->fully-featured FFmpeg (--enable-libx264, --enable-avisynth if the headers are present, as above)
```

Two separate builds of FFmpeg there, one with half of its stuff disabled, and one not.  And that's only if you absolutely want/need both the x264 CLI's LAVF/FFMS inputs and for FFmpeg to be able to use libx264.

Or, if you don't care about FFmpeg using libx264, you could just build a fully-featured FFmpeg, FFMS2, and then the full x264 CLI, which would be unnecessarily inflated in size due to all the stuff in the FFmpeg libraries you don't need.

AviSynth+ doesn't factor into that possible circular dependency at all.  x264 and FFmpeg don't link to AviSynth+, they dynamically load it.  Forcing a static library out of AviSynth+ won't solve anything because it will just end up getting ignored by x264 and FFmpeg, since the only thing they need are the headers (-DHEADERS_ONLY:bool=on when configuring AviSynth+, or use the GNUmakefile, and that only applies to FFmpeg 4.3 or -git, because as noted above, x264 has its own local copy of AviSynth+'s avisynth_c.h header that it can use), and they'll just end up ignoring the static library in favor of a detected system AviSynth.dll anyway.

Also, GCC-built AviSynth+ on Windows is nigh-useless because it won't be able to use any MSVC-built plugins (which is probably 98% of them).

Biswa96:
Wow! I didn't know the in-depth relationship. I have not think about the incorporation with ffmpeg now. This is just starting.

>  it won't be able to use any MSVC-built plugins

I got that impression while compiling this package. How those can be enabled or compiled?

Biswa96:
Wow! I didn't know the in-depth relationship. I have not think about the incorporation with ffmpeg now. This is just starting.

>  it won't be able to use any MSVC-built plugins

I got that impression while compiling this package. How those can be enabled or compiled? And what prevents it to be done?

qyot27:
[It's a foundational ABI incompatibility between MSVC and GCC,](https://stackoverflow.com/questions/7119588/ms-vs-non-ms-c-compiler-compatibility) because most plugins use AviSynth(+)'s C++ interface.  A small number of plugins (as well as x264 and FFmpeg, and probably a few other client programs like some avs2yuv forks) use the C interface.

You'd have to recompile all the C++ plugins with g++ in order to use them with a g++-built AviSynth+.  But that would have to mean they even *support* GCC in the first place, rather than being in an MSVC-specific dialect of C++.  Since AviSynth+ supports *nix now, that may happen *eventually*, but not at the current time.

Biswa96:
OK then. Closing this since there are many hurdles to cross.

PR #6994, opened 2020-09-16
Add package: live-chart

taozuhong:


lazka (commented):
why do we need the gtk3 suffix?

taozuhong:
make ready for gtk4 in advance.

lazka:
I see, thanks

PR #6998, opened 2020-09-17
sox: add -D__USE_MINGW_ANSI_STDIO=1 to CFLAGS

c72578:
This fixes several warnings concerning `PRIu64` and `PRIuPTR`, e.g.:
```console
  warning: unknown conversion type character 'l' in format [-Wformat=]
  warning: too many arguments for format [-Wformat-extra-args]
```

### DIFF
--- a/mingw-w64-sox/PKGBUILD
+++ b/mingw-w64-sox/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 # https://sourceforge.net/p/sox/code/ci/07de8a77a862e6800b95a8d3a61c6b4e41362755/
 pkgver=14.4.2.r3203.07de8a77
-pkgrel=1
+pkgrel=2
 pkgdesc="SoX is the Swiss Army Knife of sound processing utilities (mingw-w64)"
 arch=('any')
 url="https://sourceforge.net/projects/sox/"
@@ -69,7 +69,7 @@ build() {
     --host=${MINGW_CHOST} \
     --with-distro='MSYS2 MINGW-packages'
 
-  make
+  make CFLAGS="$CFLAGS -D__USE_MINGW_ANSI_STDIO=1"
 }
 
 package() {


1480c1:
I believe the default GCC will already have that define set

c72578:
@1480c1 Please have a look at the previous build log, which contained lots of warnings concerning `PRIu64` and `PRIuPTR`
https://github.com/msys2/MINGW-packages/runs/1106253784

e.g.:
```console
2020-09-12T16:56:28.9140811Z ../../sox/src/sox.c:3011:32: warning: unknown conversion type character 'l' in format [-Wformat=]
2020-09-12T16:56:29.0834071Z  3011 |       lsx_warn(i < input_count?"`%s' input clipped %" PRIu64 " samples" :
```
etc.

Obviously `-D__USE_MINGW_ANSI_STDIO=1` is not defined, when the build reaches the **`make`** command.

Biswa96:
Both seems to be correct. `/etc/makepkg_mingw*.conf` contains `CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"` and `makepkg-mingw` command choose that flag while building a package. But in this package, that does not work. As OP said, adding the `CFLAGS` solved those warnings.

@lazka Can `CFLAGS=${CPPFLAGS} ...` be added in makepkg conf files for this type of situation?

lazka:
> I believe the default GCC will already have that define set

I think that's only true for c99 and up. Not sure what's used here

> @lazka Can `CFLAGS=${CPPFLAGS} ...` be added in makepkg conf files for this type of situation?

I don't think we should work around upstream bugs in the makepkg config.

c72578:
In the meantime I did some further investigations. This PR turned out to be a red herring.
`-Wall`was overridden by the `CFLAGS` in the make command, so the warnings were just not shown any more ...

In general, `-D__USE_MINGW_ANSI_STDIO=1` is working properly when building sox. `PRIu64` and `PRIuPTR` are OK without warning in several cases of the code, except when they are used within `lsx_warn`, `lsx_debug` or `lsx_fail` of sox.
I have prepared a new PR #7006 now and will check with upstream.

PR #7006, opened 2020-09-20
sox: add -Wno-format to CFLAGS

c72578:
This is an update to commit 9a95842.
Silence format warnings originating from usage of `PRIu64`
and `PRIuPTR` within `lsx_warn`, `lsx_debug` and `lsx_fail`, e.g.:
```console
  warning: unknown conversion type character 'l' in format [-Wformat=]
  warning: too many arguments for format [-Wformat-extra-args]
```

### DIFF
--- a/mingw-w64-sox/PKGBUILD
+++ b/mingw-w64-sox/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 # https://sourceforge.net/p/sox/code/ci/07de8a77a862e6800b95a8d3a61c6b4e41362755/
 pkgver=14.4.2.r3203.07de8a77
-pkgrel=2
+pkgrel=3
 pkgdesc="SoX is the Swiss Army Knife of sound processing utilities (mingw-w64)"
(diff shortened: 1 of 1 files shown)


c72578:
Here is some background info from **`inttypes.h`** concerning `PRIu64` etc. [1]
```C
/* 7.8.1 Macros for format specifiers
 * 
 * MS runtime does not yet understand C9x standard "ll"
 * length specifier. It appears to treat "ll" as "l".
 * The non-standard I64 length specifier causes warning in GCC,
 * but understood by MS runtime functions.
 */
#if defined(_UCRT) || __USE_MINGW_ANSI_STDIO
#define PRId64 "lld"
#define PRIi64 "lli"
#define PRIo64 "llo"
#define PRIu64 "llu"
#define PRIx64 "llx"
#define PRIX64 "llX"
#else
#define PRId64 "I64d"
#define PRIi64 "I64i"
#define PRIo64 "I64o"
#define PRIu64 "I64u"
#define PRIx64 "I64x"
#define PRIX64 "I64X"
#endif
```

[1] https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/inttypes.h

PR #7007, opened 2020-09-20
mingw-w64: Update to 8.0.0, make build reproducible

lazka:


lazka:
CI seems to crash before the build starts...

I build a few packages and al seems fine, so mergin

lazka:
(Maybe the dependency cycle in the mingw-64 packages made it hit a recursion/memory limit)

PR #7011, opened 2020-09-20
Add package: ffmpeg-git

lmorand:


Biswa96:
`-git` packages generally went into [MINGW-packages-dev][1] repository.

[1]: https://github.com/msys2/MINGW-packages-dev.git

lmorand:
Sorry, I didn't know about it. I'll close this one and open another one there.

Biswa96:
I would suggest to wait for maintainer's reply.

lazka:
I've created #7013 to figure out what to do in such cases.

@lmorand just wondering, why do you want such a package in the repo?

lmorand:
I just thought, since building from git is explicitly encoureged in ffmpeg's site, that it would be nice to have that option. Initially, I made this as an alternative to zeranoe's builds which were getting taken down; now I see there are other options, but I still prefer pacman.

lazka:
I'm going to merge this for now, but we might end up moving those -git packages to MINGW-packages-dev in the future.

Thanks!

PR #7024, opened 2020-09-24
gmp install shared headers

### LINKED ISSUE #6983: gmp shared library consumers segfault when built with image base > 4GB, headers built for static lib
I decided to try to debug the [gcc segfault](https://github.com/msys2/MINGW-packages/issues/6674#issuecomment-686656479) when built with high image base today.  I build gcc with debug symbols, and the stacktrace disappeared into mpfr, which was not built with debug symbols.  I looked at mpfr, saw it depends on gmp, and started out by building those two packages with debug !strip.  GMP build succeeded, but after installing the built package, gcc would segfault on lines as simple as "d=1.0;" (the crash I had seen in gcc was in real_from_string).

So, it appears that this gcc issue is actually a GMP issue.

jeremyd2019:
This requires mpfr, mpc, and isl to be built shared rather than static,
as their configure scripts check that they are being built in the same
static/shared configuration as the gmp headers.

This allows gcc to no longer require the --default-image-base-low flag
on x86_64.  Unfortunately, until gcc is rebuilt with proper linkage, the
gmp dll requires that flag, or the old gcc crashes.

Fixes #6983.


jeremyd2019:
There was one more package that I wanted to rebuild, gnome-calculator, but it failed, apparently building docs.

mati865:
This doesn't look like proper solution since it will break static gmp library.
Various packages use the same headers for shared and static build, to achieve that they install static headers. 

I don't know how ld.bfd works internally but in case of LLD when dynamic symbol has no dllimport it will look for `__imp_<symbol>` and use it instead of creating pseudo reloc. Maybe LD doesn't do the same?

jeremyd2019:
I have no idea what's going on with i686 in eigen3:
```
2020-09-24T08:18:35.6338875Z ==> Starting build()...
2020-09-24T08:18:35.7435502Z ==> ERROR: A failure occurred in build().
2020-09-24T08:18:35.7695577Z     Aborting...
```

jeremyd2019:
> This doesn't look like proper solution since it will break static gmp library.
> Various packages use the same headers for shared and static build, to achieve that they install static headers.

As I quoted in https://github.com/msys2/MINGW-packages/issues/6983#issuecomment-691791584, gmp explicitly does not support that.  I have no idea what they're doing that's breaking this, because indeed most packages seem to be perfectly OK with it, but whatever it is, it is documented as unsupported.

jeremyd2019:
> I have no idea what's going on with i686 in eigen3:
> 
> ```
> 2020-09-24T08:18:35.6338875Z ==> Starting build()...
> 2020-09-24T08:18:35.7435502Z ==> ERROR: A failure occurred in build().
> 2020-09-24T08:18:35.7695577Z     Aborting...
> ```
> 
> I had previously run through CI successfully, https://github.com/jeremyd2019/MINGW-packages/runs/1156893844?check_suite_focus=true, and only edited commit messages between those commits and this pull request.

I opened #7043 for this.

jeremyd2019:
> This doesn't look like proper solution since it will break static gmp library.
> Various packages use the same headers for shared and static build, to achieve that they install static headers.
> 
> I don't know how ld.bfd works internally but in case of LLD when dynamic symbol has no dllimport it will look for `__imp_<symbol>` and use it instead of creating pseudo reloc. Maybe LD doesn't do the same?

It appears that the import library contains both `__imp_symbol` and `symbol`:
```
d000613.o:     file format pe-x86-64
 
SYMBOL TABLE:
[  0](sec  1)(fl 0x00)(ty   0)(scl   3) (nx 0) 0x0000000000000000 .text
[  1](sec  2)(fl 0x00)(ty   0)(scl   3) (nx 0) 0x0000000000000000 .idata$7
[  2](sec  3)(fl 0x00)(ty   0)(scl   3) (nx 0) 0x0000000000000000 .idata$5
[  3](sec  4)(fl 0x00)(ty   0)(scl   3) (nx 0) 0x0000000000000000 .idata$4
[  4](sec  5)(fl 0x00)(ty   0)(scl   3) (nx 0) 0x0000000000000000 .idata$6
[  5](sec  1)(fl 0x00)(ty   0)(scl   2) (nx 0) 0x0000000000000000 __gmpz_xor
[  6](sec  3)(fl 0x00)(ty   0)(scl   2) (nx 0) 0x0000000000000000 __imp___gmpz_xor
[  7](sec  0)(fl 0x00)(ty   0)(scl   2) (nx 0) 0x0000000000000000 _head_libgmp_10_dll
 
 
 
Disassembly of section .text:
 
0000000000000000 <__gmpz_xor>:
   0:   ff 25 00 00 00 00       jmpq   *0x0(%rip)        # 6 <__gmpz_xor+0x6>
   6:   90                      nop
   7:   90                      nop
 
```

That disassembly looks suspiciously like a 32-bit relative address.

mati865:
> It appears that the import library contains both __imp_symbol and symbol

LLD would use `__imp_symbol` then (even with missing dllimport in headers). The question is whether BFD does similar thing or creates pseudo relocation to `symbol` when linking library that depends on `libgmp-10.dll`. Content of `__gmpz_xor` doesn't matter here.

jeremyd2019:
> > It appears that the import library contains both __imp_symbol and symbol
> 
> LLD would use `__imp_symbol` then (even with missing dllimport in headers). The question is whether BFD does similar thing or creates pseudo relocation to `symbol` when linking library that depends on `libgmp-10.dll`. Content of `__gmpz_xor` doesn't matter here.

I wouldn't bet on that with BFD. 😬

mati865:
> I wouldn't bet on that with BFD. 😬

On second thought BFD must be creating relocation since things work without dllimport when low image base is used.

Maybe we should follow upstream and provide either shared or static gmp but not both?
The thing I fear is all the other libraries using the same trick (installing "static" headers) will break the same way.
Is high image base worth the cost of breaking either static or shared libraries?

jeremyd2019:
I'm hopeful that binutils will sort out some of this stuff now that it's the default in their code base.  In the mean time properly using dllimport works.

jeremyd2019:
> Disassembly of section .text:
>  
> 0000000000000000 <__gmpz_xor>:
>    0:   ff 25 00 00 00 00       jmpq   *0x0(%rip)        # 6 <__gmpz_xor+0x6>
>    6:   90                      nop
>    7:   90                      nop
>  
> ```
> 
> That disassembly looks suspiciously like a 32-bit relative address.

Oops, looking at disassembly in gdb (once the relocations are filled out) that is actually an indirect jump, and the 32-bit relative address is filled in with __imp___gmpz_xor.  This should actually be perfectly fine.

jeremyd2019:
> So I still haven't found _why_ this is not working, made even stranger because a test program with high image base and static gmp headers was able to call gmp dll without issue.

Took me a while but eventually a light bulb went off 💡  The crash wasn't in gcc directly calling into gmp, but was actually gcc calling **static** libmpfr which called into libgmp.  The mpfr configure script checked to make sure gmp was built static (by checking the headers), but wound up linked against dynamic libgmp because -lgmp prefers libgmp.dll.a over libgmp.a.  That at least helps me narrow things down to debug, as I recall the libmpfr tests also failed similarly.

jeremyd2019:
Not all of the symbols in libgmp.dll.a have the 'thunk' functions.  `__gmpn_add_n` has `__imp___gmpn_add_n` and `__nm___gmpn_add_n`.  Symbols which have thunks don't have `__nm_foo` symbols.

mati865:
Sounds like we could patch gmp headers to look like shared but without dllimport (so static gmp library still works) or patch mpfr build system?

jeremyd2019:
> Sounds like we could patch gmp headers to look like shared but without dllimport (so static gmp library still works) or patch mpfr build system?

I'm not sure that would work.  My current theory is that for some unknown reason, gmp's assembly functions are resulting in symbols that bfd believes are data, not functions.  Without dllimport, this is causing it to use the auto-import hack, which has the Windows loader modify addresses in the function itself, rather than using the jmp thunk which references the __imp_ symbol.  This cannot work correctly on x64 with large addresses, because the CALL instruction is taking a 32-bit relative operand.  The only way this can work is with a jmp thunk (unless the compiler knew to generate an indirect CALL, which is what dllimport does).

mati865:
Ok, so the remaining question is whether static library should be shipped since it's broken anyway because of dllimport in the header.

jeremyd2019:
gmp's configure script checks for support for .type pseudo-op, but on mingw that is only allowed inside a .def, and must take an integer not `@function`, so gmp suppresses it altogether.

If this is the issue, which I believe it is, I may be able to come up with a ~~nasty hack~~ fix for this.

jeremyd2019:
My nasty hack seems to have done the trick for mpfr's tests, gcc build of course still running.  So now the question is would we prefer breaking/dropping static libgmp support, following upstream's advice, or introducing a nasty hack so that using static headers with the dynamic libgmp works properly with large addresses.

```bash
  # nasty hack to get asm function symbols identified as functions
  sed -i 's|^define(<TYPE>, <>)$|define(<TYPE>, <.def $1; .scl 2; .type 32; .endef>)|' config.m4
```

This is nasty because a) configure should be taught how to generate that, and b) that .type is hardcoded for function, whereas the configure-generated macro gets $2 as the type of symbol (function or object I think).  For all uses of the macro on mingw/x86 it is only ever passed 'function'.

jeremyd2019:
Closing in favor of #7052, which is no longer a nasty hack

PR #7030, opened 2020-09-24
Lis solver

3rav:
New packages:
http://www.ssisc.org/lis/


lazka:
Your PKGBUILD creates an empty package.

3rav:
I update PKGBUILD

PR #7031, opened 2020-09-24
svt-hevc: Add package for gst-plugins-bad.

Biswa96:


Biswa96:
Why does this always happen if I ignore the 32bit package! 😢

Biswa96:
Why does this always happen if I ignore the 32bit package! No x86 support, I guess. 😢

1480c1:
yeah, svt-hevc doesn't support 32-bit, only 64

@lazka @eine would it be possible to make github actions check the `arch=()` line?

Biswa96:
> yeah, svt-hevc doesn't support 32-bit, only 64

Then I don't think it is necessary for only x86_64 package.

eine:
I think that tools/packages which are available on one architecture only should be supported. However, traditionally, that was not the case. See the conversation with @mati865 in https://github.com/msys2/MINGW-packages/pull/5757#issuecomment-640264853.

PR #7032, opened 2020-09-24
binutils: allow creating an empty import library

### LINKED ISSUE #7029: [binutils] Doesn't create importlib for DLLs with no exports
So we don't forget: https://github.com/msys2/MINGW-packages/pull/6971#issuecomment-692276067

jeremyd2019:
Fixes #7029


lazka:
Thanks!

jeremyd2019:
Patch was applied upstream:
https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=51dee44b37c4254e6b6a016b062c23b8e1f9ca33

lazka:
nice!

PR #7034, opened 2020-09-25
cairo: Port to meson; Update to git master

lazka:
There is no release in sight, so package the latest commit like
Arch does.


Biswa96:
Is it possible to use tarball of that specific commit? The package() will not change and fetching source will be faster that git clone.

lazka:
Like this?

Biswa96:
Hmm... Yes 👍 

This is just my personal opinion, you can skip it. Another hack I like to do to make the less diff.

```sh
prepare() {
  mv "${srcdir}/${_realname}-${_commit}" "${srcdir}/${_realname}-${pkgver}"
  patch -p1 -i ${srcdir}/0009-standalone-headers.mingw.patch
}
```

This makes the package step as previous. So, one doesn't have to change `${commit}` to `${pkgver}` back and forth at the time of stable release.

PR #7042, opened 2020-09-26
[New package] butler

Schala:
This is itch.io's game packaging/patch helper


mingw-w64-python-aiohttp/.gitignore
```diff
@@ -0,0 +1,2 @@
+aiohttp/
```
lazka, line 1: what's this about?

mingw-w64-python-typing_extensions/.gitignore
```diff
@@ -0,0 +1 @@
+typing/
```
lazka, line 1: and this?

Schala:
I partition my PRs into branches, but some use Git and the repos like to rear their ugly heads into commits on other branches

PR #7044, opened 2020-09-28
rust: Update to 1.46.0

clouds56:


lazka:
The update is currently in progress here -> #6932

PR #7046, opened 2020-09-28
Filter -D__USE_MINGW_ANSI_STDIO=1 from CPPFLAGS.

### LINKED ISSUE #7043: i686 libstdc++-6.dll built after latest mingw crt/winpthreads/etc update breaks things
After installing a freshly built gcc, things are silently failing. cmake was the first thing I saw, but then when I tried to find out what was happening I saw gdb was also silently failing.
the most likely thing that changed between when it worked and now would be the mingw crt/etc packages were updated

dependency walker is showing libstdc++ has undefined imports `__emutls_v._ZSt11_once_call` and `__emutls_v.Zst15_once_callable`. Those symbols are present in the existing/working libstdc++-6.dll.

libstdc++-6.dll from CI rebuild of gcc has the issue. Local build of gcc after downgrading mingw packages(`pacman -U /var/cache/pacman/pkg/mingw-w64-i686-*-8.0.0.5*`) ongoing, but there is already a stage3 libstdc++-6.dll, which defines those symbols.

jeremyd2019:
Fixes #7043.

Draft until my gcc build gets far enough along to verify libstdc++ build.

### DIFF
--- a/mingw-w64-winpthreads-git/PKGBUILD
+++ b/mingw-w64-winpthreads-git/PKGBUILD
@@ -6,7 +6,7 @@ _realname=winpthreads
 pkgbase=mingw-w64-${_realname}-git
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-git" "${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
 pkgver=8.0.0.6001.98dad1fe
-pkgrel=2
+pkgrel=3
 _commit='98dad1fefed2a07ade0b37ca6c610ad66739631a'
 pkgdesc="MinGW-w64 winpthreads library"
 url="https://mingw-w64.sourceforge.io/"
@@ -49,6 +49,8 @@ build() {
   mkdir -p "${srcdir}"/build-${MINGW_CHOST} && cd ${srcdir}/build-${MINGW_CHOST}
 
   declare -a extra_config
+  # https://github.com/msys2/MINGW-packages/issues/7043
+  extra_config+=("CPPFLAGS=${CPPFLAGS//-D__USE_MINGW_ANSI_STDIO=1/}")
   if check_option "debug" "y"; then
     extra_config+=("CFLAGS=-O0 -g -DWINPTHREAD_DBG")
     extra_config+=("CXXFLAGS=-O0 -g -DWINPTHREAD_DBG")


jeremyd2019:
config.log on libstdc++ shows thread-local is supported, looks like this does the trick

mingw-w64-winpthreads-git/PKGBUILD
```diff
@@ -49,6 +49,8 @@ build() {
   mkdir -p "${srcdir}"/build-${MINGW_CHOST} && cd ${srcdir}/build-${MINGW_CHOST}
 
   declare -a extra_config
+  # https://github.com/msys2/MINGW-packages/issues/7043
+  extra_config+=("CPPFLAGS=${CPPFLAGS//-D__USE_MINGW_ANSI_STDIO=1/}")
```
lazka, line 53: Wouldn't appending `-D__USE_MINGW_ANSI_STDIO=0` be better here?

mati865:
cc @1480c1 this PR will fix Rust issue.

1480c1:
Will try to locally build and test once I wake up

mingw-w64-winpthreads-git/PKGBUILD
jeremyd2019: > Wouldn't appending `-D__USE_MINGW_ANSI_STDIO=0` be better here?

With -D, that would result in a redefinition warning.

> Or alternatively just unset CFLAGS/CXXFLAGS/CPPFLAGS altogether?

That's possible.  I was trying to make as minimal impact on any potential future flags as I could, for the next person coming along wondering why their flags didn't take effect on winpthreads.

jeremyd2019: appending `-U__USE_MINGW_ANSI_STDIO -D__USE_MINGW_ANSI_STDIO=0` looks like it would work though

jeremyd2019:
```
i686-w64-mingw32-gcc -DHAVE_CONFIG_H -I. -I../mingw-w64/mingw-w64-libraries/winpthreads -I../mingw-w64/mingw-w64-libraries/winpthreads/include -DIN_WINPTHREAD -DWINPTHREAD_DBG=1 -D__USE_MINGW_ANSI_STDIO=0 -D__USE_MINGW_ANSI_STDIO=1 -U__USE_MINGW_ANSI_STDIO -D__USE_MINGW_ANSI_STDIO=0 -Wall -DWIN32_LEAN_AND_MEAN -march=i686 -mtune=generic -O2 -pipe -MT src/libwinpthread_la-rwlock.lo -MD -MP -MF src/.deps/libwinpthread_la-rwlock.Tpo -c ../mingw-w64/mingw-w64-libraries/winpthreads/src/rwlock.c  -DDLL_EXPORT -DPIC -o src/.libs/libwinpthread_la-rwlock.o
<command-line>: warning: "__USE_MINGW_ANSI_STDIO" redefined
<command-line>: note: this is the location of the previous definition
```

The existing redefinition warning still shows up, from the =0 define that winpthreads itself does and the =1 define from makepkg ☹️

mati865: It shows because winpthreads define `__USE_MINGW_ANSI_STDIO=0`, then CPPFLAGS from makepkg define it to `__USE_MINGW_ANSI_STDIO=1` then you were undefining it with `-U__USE_MINGW_ANSI_STDIO` but the warning was emitted already at this point.

jeremyd2019: correct

lazka: ok, seems it's not that easy, so leaving as is is OK with me.

PR #7047, opened 2020-09-28
mumps: dependencies fix

okhlybov:
Redistributed dependencies to fix installation issue on a pristine MSYS2.
License fix.


Biswa96:
Stack the depends and make depends vertically.

PR #7052, opened 2020-09-29
gmp asm x86_64 COFF_TYPE

### LINKED ISSUE #6983: gmp shared library consumers segfault when built with image base > 4GB, headers built for static lib

jeremyd2019:
Fixes #6983, alternate approach to #7024 .

> My current theory is that for some unknown reason, gmp's assembly functions are resulting in symbols that bfd believes are data, not functions. Without dllimport, this is causing it to use the auto-import hack, which has the Windows loader modify addresses in the function itself, rather than using the jmp thunk which references the `__imp_` symbol. This cannot work correctly on x64 with large addresses, because the CALL instruction is taking a 32-bit relative operand. The only way this can work is with a jmp thunk (unless the compiler knew to generate an indirect CALL, which is what dllimport does).

The 'unknown reason' I mentioned is that the functions were lacking `.type 32`.

> gmp's configure script checks for support for .type pseudo-op, but on mingw that is only allowed inside a .def, and must take an integer not `@function`, so gmp suppresses it altogether.

> My nasty hack seems to have done the trick. So now the question is would we prefer breaking/dropping static libgmp support, following upstream's advice, or introducing a nasty hack so that using static headers with the dynamic libgmp works properly with large addresses.
> 
> ```shell
>   # nasty hack to get asm function symbols identified as functions
>   sed -i 's|^define(<TYPE>, <>)$|define(<TYPE>, <.def $1; .scl 2; .type 32; .endef>)|' config.m4
> ```
> 
> This is nasty because a) configure should be taught how to generate that, and b) that .type is hardcoded for function, whereas the configure-generated macro gets $2 as the type of symbol (function or object I think). For all uses of the macro on mingw/x86 it is only ever passed 'function'.


jeremyd2019:
gnome-calculator failed to link.  This happened before my changes too, but does not happen on #7024.

jeremyd2019:
Was just looking at how to do this in a way that isn't a nasty hack, and apparently gmp already had and fixed this issue, on x86.  They neglected to include the fix on x86_64.  https://gmplib.org/repo/gmp-6.2/file/tip/acinclude.m4#l2128

Non-hackish patch shortly

jeremyd2019:
Assuming the CI checks succeed, I plan to try sending the patch to gmp upstream.

lazka:
@jeremyd2019 CI looks good

jeremyd2019:
> Assuming the CI checks succeed, I plan to try sending the patch to gmp upstream.

https://gmplib.org/list-archives/gmp-bugs/2020-October/004890.html

jeremyd2019:
https://gmplib.org/list-archives/gmp-bugs/2020-October/004894.html

jeremyd2019:
Is this PR waiting on anything more from me?

jeremyd2019:
https://gmplib.org/list-archives/gmp-bugs/2020-November/004929.html

It looks like a fix for this is committed.  There is a question in there that I am not equipped to answer though:
> Do you know if this might be needed also for mingw running on ARM cpus?

Maybe somebody here has enough knowledge of how Windows on ARM works?  It looks like ld still needs to be told whether a symbol is data or not to generate a proper implib, but what I don't know is whether as wants the same syntax to do so.

Biswa96:
> Maybe somebody here has enough knowledge of how Windows on ARM works? 

Guru here https://github.com/mstorsjo/llvm-mingw

jeremyd2019:
> > Maybe somebody here has enough knowledge of how Windows on ARM works?
> 
> Guru here https://github.com/mstorsjo/llvm-mingw

Thanks.  I was able to query that compiler for what I wanted to know ;)

mati865:
GNU toolchain doesn't support ARM/AArch64 as MinGW target so we would have to test it with LLVM toolchain.

PR #7053, opened 2020-09-29
Mingw w64 riscv64 unknown elf gcc

Fabien-Chouteau:
Based on mingw-w64-arm-none-eabi-gcc.

### DIFF
--- a/mingw-w64-riscv64-unknown-elf-gcc/PKGBUILD
+++ b/mingw-w64-riscv64-unknown-elf-gcc/PKGBUILD
@@ -1,0 +1,139 @@
+# Maintainer: Fabien Chouteau <chouteau@adacore.com>
+# Based on mingw-w64-arm-none-eabi-gcc
+
+_realname=gcc
+_target=riscv64-unknown-elf
+
+pkgbase=mingw-w64-${_target}-${_realname}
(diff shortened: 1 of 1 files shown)


Biswa96:
Can the target be added in arm pkgbuild files? I mean one pkgbuild file to create multiple target packages.

Fabien-Chouteau:
> Can the target be added in arm pkgbuild files? I mean one pkgbuild file to create multiple target packages.

Actually I don't know. How would this be done?

Fabien-Chouteau:
@Biswa96 I see that gcc recipes makes multiple packages: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gcc/PKGBUILD

I see multiple drawbacks for tool-chains:
 - The compilation times are already pretty big we already compile everything twice (standard and nano version) so 
 - The different build may require different switches, I think this is doable but might make the script harder to maintain
 - We might want to update to new versions of GCC or newlib at a different time

I'd like to have other opinions on these points.

lazka:
Another downside is that a person updating the arm one might not be able (or know how) to test the other arches.

Biswa96:
Ya, valid logic. I understand.

PR #7061, opened 2020-09-30
luajit: Update to 2.1.0-beta3

kou:
I want to use "JIT compiler support for x64 in GC64 mode" feature implemented in 2.1.0-beta3:

https://luajit.org/ :

> LuaJIT 2.1.0-beta3 has been released
>
> ...
>
> This is the third beta release for the new v2.1 branch. Apart from various fixes, the major new features are JIT compiler support for x64 in GC64 mode, ARM64 and MIPS64.

I know that this is a beta release not a stable release.
But It seems that there is no plan to release a new stable version: https://github.com/LuaJIT/LuaJIT/issues/563

Can we use beta release in our repository?

### DIFF
--- a/mingw-w64-luajit/PKGBUILD
+++ b/mingw-w64-luajit/PKGBUILD
@@ -3,9 +3,9 @@
 _realname=luajit
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.0.5
+pkgver=2.1.0_beta3
 pkgrel=1
-_sourcedir=${_realname}-${pkgver}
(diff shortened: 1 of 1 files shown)


mingw-w64-luajit/PKGBUILD
```diff
@@ -3,9 +3,9 @@
 _realname=luajit
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.0.5
+pkgver=2.1.0_beta3
```
kou, line 6: I use `_` not `-` here because https://wiki.archlinux.org/index.php/PKGBUILD#pkgver says so.

PR #7062, opened 2020-09-30
OpenAI dependencies tier 2

Schala:
These packages have dependencies on those committed in tier 1. I'll do my best to keep hese updated. They are as follows:

- python-ffmpeg-python (pypi package is ffmpeg-python)
- python-fire
- python-flask
- python-labelme
- python-numba
- python-oauth2client
- python-pytest-localserver
- python-responses
- python-scikit-image
- python-yarl


lazka:
tier 1 is in the repo now. I've retriggered CI here and there are some errors.

Schala:
Okay yeah I see it. Looks like pytest-localserver 404s. I'll take a look.

lazka:
One thing I forgot last time, this:

```
provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
```

is not needed anymore. It was there to let old packages transition from python3 to python naming.

Can you remove them in all packages?

Schala:
Sure, I'll do the same for tier 1 as well. There are other packages that aren't mine that have this as well.

lazka:
> Sure, I'll do the same for tier 1 as well. There are other packages that aren't mine that have this as well.

No need to touch the old ones imo.

mingw-w64-butler/PKGBUILD
```diff
@@ -1,27 +0,0 @@
-# Maintainer: Sarah Ottinger <schalaalexiazeal@gmail.com>
```
lazka, line 1: Why did you delete this?

Schala: Butler wasn't part of the OpenAI branch. I had it committed to master.

lazka: ah, sry, I missed that. I've rebased this now on top of current master.

PR #7063, opened 2020-09-30
hwloc: Use build folder, remove unwanted files.

Biswa96:


3rav:
A new version has been released (v2.3; Sep 28, 2020):
https://www.open-mpi.org/software/hwloc/v2.3/

PR #7065, opened 2020-09-30
Remove lcov

lazka:
This was using only msys perl and msys dependencies, so not a mingw package.
Newer (working) version here: https://github.com/msys2/MSYS2-packages/pull/2165


lazka:
@marknsikora fyi

PR #7074, opened 2020-10-01
[RFC] Merge embedded targets in single PKGBUILD (binutils and newlib)

Fabien-Chouteau:
Following the discussion in #7053 I tried to change the binutils and newlib recipes to build multiple targets.

CC @Biswa96 @fauxpark @lazka


lazka:
imho whatever is easier for you.

Fabien-Chouteau:
@fauxpark I'd like to have your opinion here because you are maintaining the arm-none-eabi toolchain.
Also have a look at #7053.

lazka:
@Fabien-Chouteau given the lack of feedback feel free to decide what should happen next.

Fabien-Chouteau:
@lazka I do think that it makes sens to have binutils, newlib and gdb merged. So let take this patch (pending checks).

But I would keep gcc separate.

mingw-w64-embedded-binutils/PKGBUILD
```diff
@@ -0,0 +1,81 @@
+# Maintainer: Carlos Antunes <cmantunes@gmail.com>
+# Maintainer: fauxpark <fauxpark@gmail.com>
+# Contributed by: nanodude
+
+_realname=binutils
+_all_targets="arm-none-eabi riscv64-unknown-elf"
+
+pkgbase=mingw-w64-${_realname}
```
lazka, line 8: We already have a mingw-w64-binutils

mingw-w64-embedded-binutils/PKGBUILD
```diff
@@ -0,0 +1,81 @@
...
+pkgname=($(for _target in ${_all_targets}; do
+               echo "${MINGW_PACKAGE_PREFIX}-${_target}-${_realname}"
+           done))
+pkgver=2.35
+pkgrel=1
```
lazka, line 13: You need to increase the pkgrel, the packages are already at 2.35-1

Fabien-Chouteau:
Hi @lazka, what is missing to merge this?

fauxpark:
What's the advantage to merging these PKGBUILDs? I see a few cons in the linked issue (build time, different configuration for different targets, lack of knowledge on the part of maintainers - I certainly do not know anything about RISC-V) but not really any pros.

If we are merging all embedded targets together then we should also be adding AVR here - but last I saw, I don't think this will be possible from GCC 11 onwards as it has not been ported to `MODE_CC`. There appears to be some work being done to save it from being dropped, though: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92729

Fabien-Chouteau:
> What's the advantage to merging these PKGBUILDs?

Factoring a few dozens of lines of code.

umarcor:
I'm looking into adding powerpc too. Moreover, I think it would be desirable to enable Ada as a target language on arm and powerpc toolchains (not only on riscv). The driving use case is https://github.com/antonblanchard/microwatt, an open hardware implementation of OpenPOWER ISA, which can be prototyped on FPGAs and will soon be prototyped using Google/Skywater's ASIC shuttle.

Can we please reconsider this proposal? It's not just a few dozens lines of code, as Fabien already pointed out. Keeping them separated would involve 3 recipes repeated three times each. E.g., I opened https://github.com/msys2/MINGW-packages/pull/7976 because arm and riscv toolchain were already out of sync with regard to the GCC version.

/cc @cr1901 @mithro

Fabien-Chouteau:
> I'm looking into adding powerpc too.

It should just be a matter of adding it to the `_all_targets=`.

> Moreover, I think it would be desirable to enable Ada as a target language on arm and powerpc toolchains (not only on riscv). 

There is already a PR for `arm-none-eabi` Ada support in GCC: #7976.

umarcor:
Merci Fabien.

> > I'm looking into adding powerpc too.
> 
> It should just be a matter of adding it to the `_all_targets=`.

Yes. Can you reopen? Or does it need to be handled in a new PR?

> There is already a PR for `arm-none-eabi` Ada support in GCC: #7976.

Would it make sense to ditch #7976 in favour of this one? Or anyway, reconsidering this after merging #7976?

Fabien-Chouteau:
> > It should just be a matter of adding it to the `_all_targets=`.
> 
> Yes. Can you reopen? Or does it need to be handled in a new PR?

I am ok to reopen this one, but there was no consensus in the first place so maybe it is better for you to make a powerpc binutils/newlib/gdb/gcc packages.

> > There is already a PR for `arm-none-eabi` Ada support in GCC: #7976.
> 
> Would it make sense to ditch #7976 in favour of this one? Or anyway, reconsidering this after merging #7976?

This one is about newlib and binutils, so not related to GCC in #7976.

umarcor:
> I am ok to reopen this one, but there was no consensus in the first place so maybe it is better for you to make a powerpc binutils/newlib/gdb/gcc packages.

I believe that most of Ryan's potential issues are very potential and negligible in practice. If merging is possible, I'd rather avoid creating four new packages, just for them to be removed next.

> This one is about newlib and binutils, so not related to GCC in #7976.

newlib and binutils are dependencies of GCC, which are also created for each target. Therefore, it makes sense to make GCC's recipes be coherent with this.

PR #7077, opened 2020-10-01
Update KF5 to 5.74

zrax:


zrax:
As usual, these packages are order dependent, so will probably fail CI until built manually.

mingwandroid:
Can you untangle the order and make lists of each group and PRs for each group? I think that might make things work better?

Issue #7027, opened 2020-09-24
ld link error: "ld.exe: DWARF error: could not find variable specification at offset"


ice0:
mingw-w64-x86_64-gcc **10.2.0-3**

Trying to build c++ library.
```
make VERBOSE=1 synfig
...
cd /C/msys64/home/az/projects/build-test/synfig-core/src/synfig && /C/msys64/mingw64/bin/ar.exe cr CMakeFiles/synfig.dir/objects.a @CMakeFiles/synfig.dir/objects1.rsp
cd /C/msys64/home/az/projects/build-test/synfig-core/src/synfig && /C/msys64/mingw64/bin/g++.exe  -g -fno-inline-functions -fno-inline -g -O0 -Wall  -shared -o ../../../output/bin/libsynfig.dll -Wl,--out-implib,../../../output/lib/libsynfig.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -Wl,--whole-archive CMakeFiles/synfig.dir/objects.a -Wl,--no-whole-archive   -L/C/msys64/opt/mlt-6.16.0/lib  -lsigc-2.0 -lgiomm-2.4 -lgio-2.0 -lglibmm-2.4 -lgobject-2.0 -lglib-2.0 -lintl -lsigc-2.0 -lcairo -lxml++-2.6 -lxml2 -lglibmm-2.4 -lgobject-2.0 -lglib-2.0 -lintl -lsigc-2.0 -lmlt++ -lmlt -lz -lfftw3 -lltdl -lgiomm-2.4 -lgio-2.0 -lcairo -lxml++-2.6 -lxml2 -lmlt++ -lmlt -lz -lfftw3 -lltdl -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: DWARF error: could not find variable specification at offset 4fbbf3
...
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: DWARF error: could not find variable specification at offset 4c849a
[...]

PR #7079, opened 2020-10-01
Binutils 2.35.1

jeremyd2019:
First commit is a straight update to 2.35.1, while removing the Changelog/NEWS files from the patches to reduce conflicts.  Second commit adds upstream patch for DWARF decoding, which should hopefully help with `DWARF error: could not find variable specification at offset` errors from #7027 and https://gitter.im/msys2/msys2?at=5f761f89f9cfbe19f931bbe8 (I wish we had a small reproducer for that to verify).

Both commits built in my fork's CI, but I have not tested them yet.  Is there any recommended test procedure for a new binutils version?


jeremyd2019:
I tested this building gmp (because I was also testing the gcc/libs from #7052 and gmp without --default-image-base-low) and python on both x86_64 and i686 (python because it uses LTO and PGO), and all were fine.

Alexpux:
Well let's see if it solve problem with 32bit Qt5

PR #7080, opened 2020-10-02
gtk-doc: Update to 1.33.0; Port to meson

lazka:


lazka:
(@kaiw thanks for meld! It's such a useful tool when porting packages to different build systems.)

PR #7089, opened 2020-10-06
libaec: New package

kmilos:
Advanced entropy coding library needed for e.g. future python-imagecodecs

Also [replaces SZIP which is only used by HDF5](https://gitlab.dkrz.de/k202009/libaec/-/blob/master/README.SZIP) so far.


lazka:
I wonder if instead of the "replaces" you could bump the pkgrel of hdf5 in this PR and replace the dependency there. So we know if hdf5 builds fine with libaec.

kmilos:
I've looked at the new libszip.dll from libaec and various current build hdf5 binaries using dependency walker and objdump, this really should just work: hdf5 binaries are dynamically linked against libszip.dll, and the new one exports all the necessary symbols.

But I can give a hdf5 version bump and build againt libaec a go as well (when I get around to it).

kmilos:
hdf5 indeed does not build against this new package, which I guess does not provide all the necessary devel files, just keeps runtime compatibility (static lib name is different, no cmake config files...)

I guess there are two options going forward: disabling szip as part of libaec so there is no conflict (have to check python-imagecodecs is ok with the old one), or fixing up libaec to supply all the devel artifacts as szip did (which I guess considerably more involved)... Any thoughts?

A question while we're at it: I guess if we go with the latter option to fix up libaec and rebuild hdf5, one would need to also rebuild all the packages that depend of hdf5 (quite a few it seems)?

lazka:
> hdf5 indeed does not build against this new package, which I guess does not provide all the necessary devel files, just keeps runtime compatibility (static lib name is different, no cmake config files...)

ok, so please remove the "replaces". this allows uses to install it if they want.

> I guess there are two options going forward: disabling szip as part of libaec so there is no conflict (have to check python-imagecodecs is ok with the old one), or fixing up libaec to supply all the devel artifacts as szip did (which I guess considerably more involved)... Any thoughts?

If your main motivation for adding this is for python-imagecodecs, then the first option sounds like a god idea.

> A question while we're at it: I guess if we go with the latter option to fix up libaec and rebuild hdf5, one would need to also rebuild all the packages that depend of hdf5 (quite a few it seems)?

Why do you think they need to be rebuilds?

kmilos:
> If your main motivation for adding this is for python-imagecodecs, then the first option sounds like a god idea.

I think so too. Having some unrelated build issues w/ python-imagecodecs though, so converting this to draft until I figure those out.

kmilos:
Ended up removing szlib from libaec build as discussed, will push python-imagecodecs shortly.

kmilos:
FYI, unlike w/ github, there is a difference between .diff and .patch formatting in gitlab links...

lazka:
Is this ready now?

kmilos:
I believe so; python-imagecodecs builds ok for me, waiting for this dep to get in first.

lazka:
ok, thanks!

PR #7090, opened 2020-10-07
OpenAI deps tier 3

Schala:
The following new packages added are dependent on the formerly committed tier 2 packages for OpenAI. There are more on the way than what is listed here, because I have to go back to previous tiers and add new dependencies.

- python-aiohttp
- python-google-auth
- python-resampy

### DIFF
--- a/mingw-w64-python-aiohttp/.gitignore
+++ b/mingw-w64-python-aiohttp/.gitignore
@@ -1,0 +1,2 @@
+aiohttp/
+http-parser/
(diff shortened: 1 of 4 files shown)


lazka:
Do you need help untangling this?

Schala:
Yeah. I merged previous tiers and now it's all messed up. I'm a student, so I've been busy during the week. Much appreciated!

lazka:
Thanks!

PR #7097, opened 2020-10-08
Add librdkafka version 1.5.0

dsogari:
From https://github.com/edenhill/librdkafka.

Note: the tests fail to load `libstdc++-6.dll`.

### DIFF
--- a/mingw-w64-librdkafka/PKGBUILD
+++ b/mingw-w64-librdkafka/PKGBUILD
@@ -1,0 +1,58 @@
+# Maintainer: Diego Sogari <diego.sogari@gmail.com>
+
+_realname=librdkafka
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.5.0
+pkgrel=1
+pkgdesc="The Apache Kafka C/C++ client library (mingw-w64)"
+arch=('any')
+url="https://github.com/edenhill/librdkafka"
+license=('BSD')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-gcc")
+depends=("${MINGW_PACKAGE_PREFIX}-cyrus-sasl"
+         "${MINGW_PACKAGE_PREFIX}-dlfcn"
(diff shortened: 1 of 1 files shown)


Biswa96:
In build directory, if `./src/librdkafka.dll` and `./src-cpp/librdkafka++.dll` files are copied to `./tests` directory then `ctest ./` succeed. Maybe there is some issue in the CMakeLists.txt file. Also, if possible add this command after `make install` to remove reference to MSYS2 install directory:

```sh
  for _f in "${pkgdir}${MINGW_PREFIX}"/lib/cmake/RdKafka/*.cmake; do
    sed -e "s|${PREFIX_WIN}|\$\{_IMPORT_PREFIX\}|g" -i ${_f}
  done
```

Biswa96:
Tests are not strongly required here but it's good to have. In build directory, if `./src/librdkafka.dll` and `./src-cpp/librdkafka++.dll` files are copied to `./tests` directory then `ctest ./` succeed. Maybe there is some issue in the CMakeLists.txt file. Also, if possible add this command after `make install` to remove reference to MSYS2 install directory:

```sh
  for _f in "${pkgdir}${MINGW_PREFIX}"/lib/cmake/RdKafka/*.cmake; do
    sed -e "s|${PREFIX_WIN}|\$\{_IMPORT_PREFIX\}|g" -i ${_f}
  done
```

dsogari:
@Biswa96 thanks! I applied your suggestions and was able to run the tests, but unfortunately not all of them passed in my machine. On the other hand, I see that the CI pipeline does not perform the check stage, so I think I can leave this up to you now. If everything is ok, could you give a rough estimate of when the compiled package would be available in the official repository (in days or weeks)?

Biswa96:
> unfortunately not all of them passed in my machine

It is not necessary all tests to be passed. But if a program is linked to this library the program needs to work flawlessly. 

> when the compiled package would be available in the official repository (in days or weeks)?

Generally weekly. The recently added packages are listed here https://packages.msys2.org/updates.

dsogari:
It looks like cmake build support is not officially supported, according to [this page](https://github.com/edenhill/librdkafka/blob/master/packaging/cmake/README.md). They give a hint on how to run the tests: it seems that only "local" tests (i.e., with no broker) can be run in a basic setup.

PR #7098, opened 2020-10-08
spdlog: Update to 1.8.1

podsvirov:


lazka:
This removes the static build. On purpose? I wouldn't mind, just wondering.

podsvirov:
Yes.  Previously it was impossible, but now we can try it.  Do we have any policy for this case?

lazka:
> Do we have any policy for this case?

Just trying not to break things for existing users if possible. But you are the only one changing this package and there are no reverse deps so I hope no one depends on it.

PR #7102, opened 2020-10-10
arm-none-eabi-newlib: remove workaround from gcc

jeremyd2019:
Lower imagebase is no longer needed after #7052.

Not bumping pkgrel because the built gcc is not packaged.


jeremyd2019:
cc @fauxpark

fauxpark:
👍

fauxpark:
I wonder if it's worth removing the GCC build altogether here, now that it exists as a separate package. This is what Arch does, but as it's a circular dependency, I'm not sure.

PR #7110, opened 2020-10-12
Update GTK to 3.99.4; Update pango to 1.48.0

taozuhong:


lazka:
it depends on an unreleased pango

taozuhong:
> It depends on an unreleased pango. We'll have to wait for a pango release first I think.

right exactly

lazka:
I've read on IRC that a new pango release will be out soon.

lazka:
(another thing, gtk4 no longer depends on atk)

PR #7113, opened 2020-10-13
zeromq: update zeromq to 4.3.3 and cppzmq to 4.7.1

kanonet:
Hi,
this updates mingw-zeromq to 4.3.3 and cppzmq to 4.7.1. Please review!
Note: mingw-pyzmq also needs to get updated, but so far I had no time to do this yet.
Regards.


lazka:
>  Note: mingw-pyzmq also needs to get updated, but so far I had no time to do this yet.

would not updating it break things?

kanonet:
@lazka I do not think it would break anything... anyways I just also updated pyzmq, so everything should work together now. Please review.

PR #7116, opened 2020-10-14
Upgrade to LLVM 11

mati865:


mati865:
CI bugged?
> warning: filesystem: downgrading from version 2020.10-1 to version 2020.02-7
warning: pacman-mirrors: downgrading from version 20201009-1 to version 20201007-1

jeremyd2019:
I think the sourceforge hasn't updated with the new packages yet.

lazka:
(rebased on master to get some CI changes..)

mati865:
Upstream changed a bit build process of compiler-rt. It passes for me since I have already built compiler-rt installed in the system.

I've disabled compiler-rt libs that are not built anyway (I don't remember exact issue but I think they require some features not supported by Binutils) for now.
Proper investigation will have to wait until I get time to do it.

mati865:
`Compiler-RT supported architectures: ` should say `Compiler-RT supported architectures: i386`, very funny...
Also it detects 3-rd party Python on CI.

mati865:
Okay using already existing compiler-rt is the only way to have profiler library. I'll update PR later along with few more tweaks to PKGBUILD.

BTW maybe we should consider splitting this PKGBUILD into separate packages?

oscarfv:
> BTW maybe we should consider splitting this PKGBUILD into separate packages?

I looked into this on the past and realized that the build and maintenance would become much more complex without any substantial benefit.

mingw-w64-clang/PKGBUILD
```diff
@@ -287,7 +287,8 @@ build() {
     -DLLVM_INCLUDE_EXAMPLES=OFF \
     -DLLVM_LINK_LLVM_DYLIB=OFF \
     -DLLVM_POLLY_LINK_INTO_TOOLS=OFF \
-    -DPYTHON_EXECUTABLE=${MINGW_PREFIX}/bin/python.exe \
+    -DPython3_FIND_REGISTRY=NEVER \
+    -DPython3_ROOT_DIR=${MINGW_PREFIX} \
```
lazka, line 291: shouldn't this be `${MINGW_PREFIX}/bin` ?

mati865: Docs are really bad about this thing but from what I could find it should look for Python in `${Python3_ROOT_DIR}/bin`.

lazka: ok.. (I just tried to look at the code.. 3300 lines of cmake to find python.. wow)

mati865: Yeah, did the same and gave up 😉

mati865:
Okay it works but maybe we should uninstall the other Python for other packages?



> 
> 
> > BTW maybe we should consider splitting this PKGBUILD into separate packages?
> 
> I looked into this on the past and realized that the build and maintenance would become much more complex without any substantial benefit.

It'd make easy version bumps a bit more complicated (that can be solved with script) but big upgrades would be easier to work with. Also it'd allow using faster Ninja instead of Make.

mati865:
Removed shared linking from the description, I accidentally disabled at some point and re-enabling it now causes DEP violation. I guess we have to recompile libcxx first which is not really possible with monolithic PKGBUILD.

lazka:
For some reason this fails in the autobuilder:

```
2020-10-21T06:58:19.6663357Z [  9%] [32mBuilding ASM_MASM object projects/openmp/runtime/src/CMakeFiles/omp.dir/z_Windows_NT-586_asm.asm.obj[0m
2020-10-21T06:58:19.6733194Z [  9%] [32m[1mLinking C static library ../../../../lib/libPollyISL.a[0m
2020-10-21T06:58:19.7320725Z UASM v2.50, Jan  8 2020, Masm-compatible assembler.
2020-10-21T06:58:19.7328718Z Portions Copyright (c) 1992-2002 Sybase, Inc. All Rights Reserved.
2020-10-21T06:58:19.7474819Z Source code is available under the Sybase Open Watcom Public License.
2020-10-21T06:58:19.7506188Z 
2020-10-21T06:58:19.7537191Z make[2]: *** [projects/openmp/runtime/src/CMakeFiles/omp.dir/build.make:472: projects/openmp/runtime/src/CMakeFiles/omp.dir/z_Windows_NT-586_asm.asm.obj] Error 1
2020-10-21T06:58:19.7537981Z Error A2104: No filename specified.
2020-10-21T06:58:19.7566461Z make[1]: *** [CMakeFiles/Makefile2:36325: projects/openmp/runtime/src/CMakeFiles/omp.dir/all] Error 2
2020-10-21T06:58:19.7574113Z make[1]: *** Waiting for unfinished jobs....
```

https://github.com/msys2/msys2-autobuild/runs/1285100942?check_suite_focus=true

mati865:
PR builder did succeed 🤔 
Either https://github.com/msys2/MINGW-packages/pull/7157 was the missing part or another update broke the build.

lazka:
drive by question: that libunwind depends on gcc is probably not on purpose, right?

mati865:
Yeah, libunwind really depends only on the system libraries.

PR #7121, opened 2020-10-15
glib: revert commit to fix hang when G_SLICE is set

### LINKED ISSUE #7118: Setting G_SLICE env variable to any non-empty value freezes GLib programs
While debugging https://github.com/harfbuzz/harfbuzz/issues/2697, I found that any non-empty value of this env variable freezes GLib applications. E.g.

```
#include <glib.h>
int main(int argc, char** argv) {
  g_print("Test\n");
}
```

Building with `gcc test.c -o test $(pkg-config --cflags --libs glib-2.0)`. Runnig without the env variable set is fine, setting it to even just `G_SLICE=' '` will cause the program to just freeze (nothing is printed and not returning either).

lazka:
upstream issue: https://gitlab.gnome.org/GNOME/glib/-/issues/2225

Fixes #7118


Biswa96:
How about doing `#define g_getenv getenv`?

lazka:
How do you know that this is safe?

PR #7134, opened 2020-10-18
Fix virtualenv unable to access venv setup files

### LINKED ISSUE #7014: Cannot run pip within tox on MSYS2 (MINGW64)
This was reported in tox-dev/tox#1648, then in pypa/pip#8894, and now here.

When trying to use tox on MSYS2 (MINGW64), I am getting several errors when packages are installed:

```
ERROR: Cannot uninstall 'distlib'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

The same procedure works on powershell or on GNU/Linux.

To reproduce, see https://github.com/eine/tox/actions/runs/264949167 and the corresponding commit: [eine/tox@9b702f8](https://github.com/eine/tox/commit/9b702f89d6747c0e44f729dde2bd35f095b20e0f). The error is seen in https://github.com/eine/tox/runs/1143520398?check_suite_focus=true#step:4:28.

danyeaw:
We aren't currently compiling the venvlauncher as part of python with gcc. Previous patches in #5625 removed the venvlauncher redirector by skipping "nt" platform portions in the venv module.

Although the previous patches worked to fix venv, it left us with key venvlauncher files missing that were expected to be present in Windows.

In issue #7014, tox and other tools that depend on virtualenv were unable to run. For example, there was issues with no such file or directory errors: 'mingw64/lib/python3.8/venv/scripts/nt/python.exe'. This PR closes #7014 and https://github.com/pypa/virtualenv/issues/1987.

This fixes these issues by removing some of the previous patches in venv and copying python.exe and pythonw.exe to the locations where the venvlauncher redirect files are expected.

A future solution may be to try to compile venvlauncher using gcc. If anyone has some guidance on how to go about this, I could take a look at the more proper solution as well. Currently `launcher.c` is compiled for Windows using a Visual Studio project here: https://github.com/python/cpython/blob/master/PCbuild/pylauncher.vcxproj, would we need to somehow turn this in to a Makefile?


lazka:
I get `Error: [WinError 2] The system cannot find the file specified` when trying to create a venv with this

danyeaw:
@lazka Which file does it say is missing?

lazka:
That's all I get. I haven't figured out what it's missing.

lazka:
ah, it's because I used "python3" instead of "python".

danyeaw:
@lazka Ahh, I didn't consider the use of `python3` I can update the PR to copy those files as well

lazka:
I don't see a difference in the created venv (besides missing python3 files), so I'm wondering what's the gain of removing the patch?

danyeaw:
@lazka try to `pip install virtualenv`, then `virtualenv _venv2`. Virtualenv now works because it finds the right files.

lazka:
> I'm wondering where venvlauncher is coming from? I don't see it in the official Python install.

Do you know where these launchers are in an official Python install?

danyeaw:
> Do you know where these launchers are in an official Python install?
I think they are in Python\Lib\venv\scripts\nt, that or the main python.exe and pythonw.exe files in Python\ are the wrappers around the files in the scripts directory. I notice that ones under scripts are 5x larger.

lazka:
ok, I think I get it now

mingw-w64-python/PKGBUILD
```diff
@@ -442,6 +442,13 @@ package() {
   sed -i "s|${pkgdir}${MINGW_PREFIX}|${MINGW_PREFIX}|g" \
     "${pkgdir}${MINGW_PREFIX}"/lib/python${_pybasever}/lib-dynload/_sysconfigdata*.py \
     "${pkgdir}${MINGW_PREFIX}"/lib/python${_pybasever}/smtpd.py
+
+  # Replace missing venvlauncher files (issue #7014)
+  cp "${pkgdir}${MINGW_PREFIX}"/bin/python.exe "${pkgdir}${MINGW_PREFIX}"/bin/venvlauncher.exe
+  cp "${pkgdir}${MINGW_PREFIX}"/bin/pythonw.exe "${pkgdir}${MINGW_PREFIX}"/bin/venvwlauncher.exe
```
lazka, line 448: I don't see why this is needed. venv and virtualenv also work without these. unless I'm missing something.

mingw-w64-python/2050-undo-venv-redirector.patch
```diff
@@ -18,21 +18,3 @@ diff -Naur Python-3.8.0-orig/Lib/venv/__init__.py Python-3.8.0/Lib/venv/__init__
              binname = 'Scripts'
              incpath = 'Include'
              libpath = os.path.join(env_dir, 'Lib', 'site-packages')
-@@ -156,7 +157,7 @@
```
lazka, line 21: I don't think this patch change is needed. venv is the same as before except it's missing files like python3.exe now.

mingw-w64-python/PKGBUILD
danyeaw: I was getting a venvlauncher and venvwlauncher file not found error with virtualenv.

mingw-w64-python/PKGBUILD
```diff
@@ -442,6 +442,13 @@ package() {
...
+  mkdir -p "${pkgdir}${MINGW_PREFIX}"/lib/python${_pybasever}/venv/scripts/nt
+  cp "${pkgdir}${MINGW_PREFIX}"/bin/python.exe "${pkgdir}${MINGW_PREFIX}"/lib/python${_pybasever}/venv/scripts/nt/python.exe
+  cp "${pkgdir}${MINGW_PREFIX}"/bin/pythonw.exe "${pkgdir}${MINGW_PREFIX}"/lib/python${_pybasever}/venv/scripts/nt/pythonw.exe
```
lazka, line 451: I guess this means virtualenv created venvs will fail on a host Python update because the wrong launcher is used. But better than before, so looks good.

We should look into compiling the right launcher as you said.

mingw-w64-python/PKGBUILD
lazka: When creating a new environment?

danyeaw: Yes

lazka: The only references to venvlauncher in virtualenv is here: https://github.com/pypa/virtualenv/blob/43a5e0383b33867b73eb1d6f51f9f0a0c063137e/src/virtualenv/create/via_global_ref/builtin/cpython/cpython3.py#L69 and that just references the build name of the vs project in cpython. And there is nothing in CPython. I'm a bit lost.

lazka: ah, https://github.com/python/cpython/blob/58ec58a42bece5b2804b178c7a6a7e67328465db/Lib/venv/__init__.py#L213

you probably got that error because "venv/scripts/nt/python.exe" wasn't there.

mingw-w64-python/2050-undo-venv-redirector.patch
lazka: ok, so if we ever get the right launcher working this will make it install it instead of. We just need to fix the missing python3.exe issue.

mingw-w64-python/2050-undo-venv-redirector.patch
```diff
@@ -18,21 +18,3 @@ diff -Naur Python-3.8.0-orig/Lib/venv/__init__.py Python-3.8.0/Lib/venv/__init__
...
-             if self.prompt is not None:
-                 f.write(f'prompt = {self.prompt!r}\n')
- 
--    if os.name != 'nt':
-+    if os.name != 'nt' or _POSIX_BUILD:
```
lazka, line 26: so, this is fine

mingw-w64-python/2050-undo-venv-redirector.patch
```diff
@@ -18,21 +18,3 @@ diff -Naur Python-3.8.0-orig/Lib/venv/__init__.py Python-3.8.0/Lib/venv/__init__
...
-         def symlink_or_copy(self, src, dst, relative_symlinks_ok=False):
-             """
-             Try symlinking a file, and if that fails, fall back to copying.
-@@ -230,7 +231,7 @@
-         path = context.env_exe
-         copier = self.symlink_or_copy
-         dirname = context.python_dir
--        if os.name != 'nt':
-+        if os.name != 'nt' or _POSIX_BUILD:
```
lazka, line 35: but here Windows code path does nothing what we want, like installing python3.exe etc. so I think we want the posix one.

mingw-w64-python/PKGBUILD
lazka: We need to copy python3.exe, python3w.exe and python3.8.exe as well, so the venv code you re-enabled copies them instead.

lazka:
(btw, disabling these two makes switches the python build twice as fast, it just makes the result slightly slower: https://github.com/msys2/MINGW-packages/blob/5d9054967c06cccea5632d372d2cb3ac8810f07c/mingw-w64-python/PKGBUILD#L330-L331 soi good for experimenting)

danyeaw:
@lazka I finished making the updates from your feedback. Everything tests out good for me with using python and python3.

lazka:
ok?

mingw-w64-python/smoketests.py
```diff
@@ -106,6 +106,19 @@ def test_sqlite_enable_load_extension(self):
...
+        import venv
+        import subprocess
+        import shutil
+        with tempfile.TemporaryDirectory() as tmp:
+            builder = venv.EnvBuilder()
+            builder.create(tmp)
+            assert os.path.exists(os.path.join(tmp, "bin", "activate"))
+            assert os.path.exists(os.path.join(tmp, "bin", "python.exe"))
+            assert os.path.exists(os.path.join(tmp, "bin", "python3.exe"))
+            subprocess.check_call([shutil.which("bash.exe"), os.path.join(tmp, "bin", "activate")])
```
lazka, line 120: @danyeaw if it fails, maybe remove this line and try again. I wasn't sure how if this should work in all cases.

danyeaw:
@lazka This is the error I am getting:
```
==> Starting check()...
.............Unable to copy 'C:/tools/msys64/home/dyeaw/MINGW-packages/mingw-w64-python/src/build-x86_64/venvlauncher.exe'
E
======================================================================
ERROR: test_venv_creation (__main__.Tests)
----------------------------------------------------------------------
  File "C:/tools/msys64/home/dyeaw/MINGW-packages/mingw-w64-python/src/smoketests.py", line 116, in test_venv_creation
    builder.create(tmp)
  File "C:/tools/msys64/home/dyeaw/MINGW-packages/mingw-w64-python/src/Python-3.8.6/Lib/venv/__init__.py", line 67, in create
    self.setup_python(context)
  File "C:/tools/msys64/home/dyeaw/MINGW-packages/mingw-w64-python/src/Python-3.8.6/Lib/venv/__init__.py", line 237, in setup_python
    os.chmod(path, 0o755)
FileNotFoundError: [WinError 2] The system cannot find the file specified: 'C:/Users/dyeaw/AppData/Local/Temp/tmpw_eml6lh/bin/python.exe'
```

lazka:
ok, I think I know why. will look later

lazka:
I went partially back to what you initially proposed. Using the Windows code paths, but patching them where needed. venvlauncher.exe is now available, but only during the build/testing like with the official build.

danyeaw:
@lazka Thanks!

naveen521kk:
> A future solution may be to try to compile venvlauncher using gcc. If anyone has some guidance on how to go about this, I could take a look at the more proper solution as well. Currently launcher.c is compiled for Windows using a Visual Studio project here: https://github.com/python/cpython/blob/master/PCbuild/pylauncher.vcxproj, would we need to somehow turn this in to a Makefile?

FYI, on https://github.com/msys2-contrib/cpython-mingw/pull/82 I have made the launchers built from source `PC/launcher.c`.

PR #7140, opened 2020-10-19
Add libavro version 1.10.0

dsogari:
From https://github.com/apache/avro


Biswa96:
The patch file too big. Would it be possible to upstream the changes then update the package here?

dsogari:
@Biswa96 Sorry, the patch got some text that was not intended to be there. I'll update it.

BTW, do you know the correct license name for Apache License 2.0?

Biswa96:
I do not know any hard rule about the license name in PKGBUILD. IMHO, if I need to know the license identifier I straight go to [SPDX License List][1] and copy it 🤓 

[1]: https://spdx.org/licenses/

dsogari:
@Biswa96 Thanks for the feedback! Could you give me a hint on making the x86_64 build pass?

It seems that cmake cannot `execute_process(COMMAND bash version.sh ...)`, where `version.sh` is a bash script located in the `lang/c` directory of the tarball. Do you know of any alternative to running a bash script inside cmake? I'm trying to avoid hardcoding the version numbers in the patch file, so it could be reused in new releases.

dsogari:
Everything OK now.

PR #7144, opened 2020-10-20
arrow: Update to 2.0.0

kou:


lazka:
Shouldn't we rebuild groonga as well because of the so bump?

kou:
Good catch!
I've bump `pkgrel` of Groonga.

I should not forget this next time.

lazka:
Thanks!

PR #7150, opened 2020-10-20
Add python-virtualenv

lazka:


lazka:
@danyeaw

mingw-w64-python-virtualenv/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+pkgrel=1
+pkgdesc="Virtual Python Environment builder (mingw-w64)"
+url="https://virtualenv.pypa.io/"
+arch=('any')
+license=('MIT')
+depends=("${MINGW_PACKAGE_PREFIX}-python-setuptools"
+         "${MINGW_PACKAGE_PREFIX}-python-appdirs"
+         "${MINGW_PACKAGE_PREFIX}-python-distlib"
+         "${MINGW_PACKAGE_PREFIX}-python-filelock"
+         "${MINGW_PACKAGE_PREFIX}-python-six")
```
danyeaw, line 16: It looks like the latest pyproject.toml doesn't depend on six any longer:
```
    "setuptools >= 41.0.0",
    "wheel >= 0.30.0",
    "setuptools_scm >= 2",
```

lazka: these are the deps required for building a wheel. the real deps are defined in setup.cfg

danyeaw: Thanks for the correction 👍

PR #7152, opened 2020-10-21
Med v4.1.0

3rav:
Base on: https://www.archlinux.org/packages/community/x86_64/med/


Biswa96:
As per @lazka's rule, the shared build should be after the static build in packaging step.

PR #7154, opened 2020-10-21
Add new package: Precice

3rav:


lazka:
Consider building cmake packages with ninja in the future: like this for example: https://github.com/msys2/MINGW-packages/pull/7155/commits/6dd4dcc6dde08e56ac577dacba0a32a915ae4e55

PR #7157, opened 2020-10-21
Add missing makedepend on libc++

mati865:


mati865:
At least it reproduces....

Biswa96:
pkgrel++

mati865:
No idea how it happened but previously there was `/c  /Fo` and now there is one space less: `/c /Fo`.

> pkgrel++

Is is necessary? The build failed for both architectures.

lazka:
> Is is necessary? The build failed for both architectures.

If this fixes things, yes, otherwise it wont rebuild.

lazka:
unrelated question: has anyone tried building this with ninja?

mati865:
> 
> 
> unrelated question: has anyone tried building this with ninja?

I often use Ninja for developing LLVM, it's way faster but as I said in https://github.com/msys2/MINGW-packages/pull/7116#issuecomment-711387675 we cannot use Ninja with monolithic package.

This comment still applies: https://github.com/msys2/MINGW-packages/pull/7157/files#diff-125e46f56c15d81bb2f9210bdcc4959baf788e9db4c0acfbff0148bf4416815dR251

lazka:
Ah, I missed that comment, thanks.

mati865:
> Ah, I missed that comment, thanks.

Not your fault, the comments are over the place here.

PR #7158, opened 2020-10-22
gtkwave: update to 3.3.107

umarcor:
This is a minor update, with some enhancements to the PKGBUILD file.


mingw-w64-gtkwave/PKGBUILD
```diff
@@ -43,6 +42,10 @@ build() {
   make
 }
 
+post_install() {
+  ${MINGW_CHOST}/bin/gtk-update-icon-cache -q -t -f ${MINGW_CHOST}/share/icons/hicolor
```
lazka, line 46: What does this do? I can't find it in the docs.

umarcor: According to https://wiki.archlinux.org/index.php/PKGBUILD#install:

> post_install — The script is run right after files are extracted. One argument is passed: new package version.

Alexpux: @umarcor port install scripts are separate scripts. See other packages

umarcor: @Alexpux, which would be the adequate stage in PKGBUILD to execute that command, without requiring separate scripts?

On the other hand, I compared the artifact of this PR with the one from https://packages.msys2.org/package/mingw-w64-x86_64-gtkwave?repo=mingw64, and there is not difference. Hence, either `post_install` is being used, or the command is useless anyway.

Alexpux: @umarcor look into https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-shared-mime-info for example

Alexpux:
@umarcor you cammot merge install scripts into PKGBUILD, they must be separate

Alexpux:
@umarcor back scripts as they was

PR #7168, opened 2020-10-24
mingw-w64-mesa: Update to 20.2.1

pal1000:
Automatically detect if llvm-config tool works and fallback to binary wrap if necessary
Update libraries list for binary wrap to support LLVM 11


lazka:
(I added some changes to mesa because of a meson install path change yesterday, I hope that didn't break anything)

pal1000:
Definitely #7170. I started working this update after meson fix landed. The CI errors clearly mention unresolved symbols.

PR #7198, opened 2020-10-29
openblas: rebuild

lazka:


lazka:
I can't reproduce this error locally.

lazka:
oh, now it worked. lets try again then

lazka:
same in msys2-autobuild, after a few builds it went through. so maybe racy and "make -j1" would help..

PR #7203, opened 2020-10-30
Add scons

lazka:
Required for building godot

### DIFF
--- a/mingw-w64-scons/001-no-pywin32.patch
+++ b/mingw-w64-scons/001-no-pywin32.patch
@@ -1,0 +1,10 @@
+--- scons-4.0.1/setup.cfg.orig	2020-07-17 05:25:10.000000000 +0200
++++ scons-4.0.1/setup.cfg	2020-10-30 09:39:09.875516500 +0100
+@@ -42,7 +42,6 @@
+ python_requires = >=3.5
+ install_requires =
+     setuptools
+-    pywin32 >= 1.0;platform_system=="Windows"
+ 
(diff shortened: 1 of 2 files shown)


mingw-w64-scons/001-no-pywin32.patch
```diff
@@ -0,0 +1,10 @@
+--- scons-4.0.1/setup.cfg.orig	2020-07-17 05:25:10.000000000 +0200
++++ scons-4.0.1/setup.cfg	2020-10-30 09:39:09.875516500 +0100
+@@ -42,7 +42,6 @@
+ python_requires = >=3.5
+ install_requires =
+     setuptools
+-    pywin32 >= 1.0;platform_system=="Windows"
```
akien-mga, line 7: Do you know why `pywin32` can't be installed on MSYS2?
As I mentioned on Twitter scons can be used without it, but most users will want to compile with more than one core, in which case it seems to be necessary on Windows.

lazka: See #5615. pywin32 is/was using features only provided by msvc. I'm buildin godot right now with "-j16" and so far it seems to run fine.

scons master switched from pywin32 to ctypes https://github.com/SCons/scons/pull/3810 so this should be fixed in the next release

Schala:
Finally, once this gets into master branch, I can add godot as a PR!

lazka:
you can bump the scons pkgrel in your PR to get it before that

PR #7206, opened 2020-10-30
darktable: Update to 3.2.1

lazka:
drops 32 bit package


lazka:
merging to see how the automation handles it..

PR #7208, opened 2020-10-31
med: fix DLL install locations, depend on libgfortran

lazka:


lazka:
@3rav

Biswa96:
Maybe?

```
ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```

lazka:
Where are they defined?

Biswa96:
Doc: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

lazka:
med doesn't use this module, and I don't feel confident enough changing more :/

PR #7226, opened 2020-11-02
fix lld hang (#6126) and dynlink clang to reduce size (#7190)

### LINKED ISSUE #7190: Reduce Clang package size?
Any reason we don't configure Clang package with dynamic linking to reduce its size?
I remember we had issues with lld but that one is now linked with Clang stdlib so it should not matter anymore.

Any feedback before I start looking into this again?

adrpo:
fixes #7190 and again #6126


adrpo:
Can anybody that uses lld test this properly? I don't use lld in my projects (yet).

mingw-w64-clang/0302-LLD-Allow-configuring-default-ld.lld-backend.patch
```diff
@@ -41,19 +41,65 @@ diff --git a/tools/lld/lld.cpp b/tools/lld/lld.cpp
...
 --- a/tools/lld/lld.cpp
 +++ b/tools/lld/lld.cpp
-@@ -92,7 +92,12 @@ static bool isPETarget(std::vector<const char *> &v) {
-       continue;
-     return isPETargetName(*(it + 1));
-   }
+@@ -27,0 +28,4 @@
++#if defined(__MINGW32__)
++#include <Windows.h>
++#endif
```
mati865, line 47: This patch is already merged upstream so it can be easily dropped with the next upgrade if you move your changes to another patch.

jeremyd2019:
Hmm, I was picturing providing an alternative `_Exit` implementation somehow...  It seems like it would still have issues with other targets too.

jeremyd2019:
> Maybe you could avoid disabling early exit and instead patch https://github.com/llvm/llvm-project/blob/release/11.x/llvm/lib/Support/Process.cpp#L96 ?

Better yet, https://github.com/llvm/llvm-project/blob/release/11.x/lld/Common/ErrorHandler.cpp#L70

adrpo:
Agreed. I saw the exit function in ErrorHandler after I did my patch and the code was not reached. I will move my stuff there.

jeremyd2019:
I'm going to try building llvm with the clang/llvm/lld binaries from this pull request tonight, and I'm also going to have a CI run going to try to build with this patch: https://github.com/jeremyd2019/MINGW-packages/blob/d7c70bdf3702bb56815b82c807541716709f8d41/mingw-w64-clang/0305-use-TerminateProcess-instead-of-exit.patch

mingw-w64-clang/0302-LLD-Allow-configuring-default-ld.lld-backend.patch
adrpo: Will do. I was lazy with the first fix :)

adrpo:
Nice. You already did it.

jeremyd2019:
Completely untested thus far though

adrpo:
@jeremyd2019  Should we move the lld.cpp parts from patch 302 to 305? Or is fine to apply several patches to the same file?

adrpo:
Ouch, because ErrorHandler.h is included we now have an ambiguous reference:
`C:/_/mingw-w64-clang/src/lld/tools/lld/lld.cpp:169:12: error: reference to 'wasm' is ambiguous`
I'll fix it.

adrpo:
I have a fix, but I will build it locally so I don't waste resources to find errors and then push it in.

jeremyd2019:
> I have a fix, but I will build it locally so I don't waste resources to find errors and then push it in.

Cool.  I have been working the other way around - pushing to build in GHA because I don't have the local resources to devote to building llvm (especially since I was already doing a local build of llvm with the results of your first patch).

jeremyd2019:
> I'm going to try building llvm with the clang/llvm/lld binaries from this pull request tonight

My build of llvm on x86_64 using clang/lld, and the binaries from your first patch's GHA run succeeded without any hangs. 🎉   This is what I was using to reproduce the hang in order to get a stack trace.

jeremyd2019:
I initially thought this was a pretty nasty hack, but I came across this comment in the ucrt:
```c++
// For Windows Store apps, starting in Windows 10, we use TerminateProcess
// instead of ExitProcess.  ExitProcess will allow threads to run during
// termination in an unordered fashion. This has lead to problems in various
// apps.  TerminateProcess does not allow threads to run while the process is
// being torn down.  See also CoreApplication::Run, which does the same.
//
// For non-Windows Store app processes, we continue to call ExitProcess, for
// compatibility with the legacy runtimes.
```

I guess this is Microsoft's new technique too.

pal1000:
@adrpo @jeremyd2019 you forgot to bump pkgrel in PKGBUILD, but it can wait for CI to turn green.

jeremyd2019:
Also I notice that you changed options 'strip' to '!strip' - if you did that because I had that on my test branch, I only did that to try to get a more helpful stack trace on the hang.  I wanted to enable debug symbols but that caused ld to fail.

jeremyd2019:
"failure" was uploading sources artifact.  Nothing important

jeremyd2019:
Those binaries also built llvm without hanging.  For fun, I'm going to try without the changes to lld.cpp (just the change to ErrorHandler.cpp and the lingering change passing canExitEarly through in the mingw driver).

adrpo:
@jeremyd2019 you can push directly to this branch if needed, I gave you access to the repo.

adrpo:
I will remove the !strip now.

adrpo:
@pal1000 thanks, I did that but then somehow I lost the change during merging, `pkgrel + 1` is now back.

lazka:
> "failure" was uploading sources artifact. Nothing important

We should probably remove the source upload. They fail a lot because of the size and don't have much value.

mingw-w64-clang/0305-use-TerminateProcess-instead-of-exit.patch
```diff
@@ -0,0 +1,90 @@
+--- lld/MinGW/Driver.cpp.orig	2020-11-02 21:43:24.117381600 -0800
++++ lld/MinGW/Driver.cpp	2020-11-02 21:43:31.351722700 -0800
+@@ -404,5 +404,5 @@
+   std::vector<const char *> vec;
+   for (const std::string &s : linkArgs)
+     vec.push_back(s.c_str());
+-  return coff::link(vec, true, stdoutOS, stderrOS);
++  return coff::link(vec, canExitEarly, stdoutOS, stderrOS);
```
jeremyd2019, line 8: This change doesn't really matter for the lld hang anymore, but seems like it should be made upstream.  The contribution process for llvm seems awfully complicated, though, so I'm not going to try to figure it out for something so trivial.

adrpo: I can probably make a patch, I did it before. I'll give it a try.

adrpo:
They have changed the _exit call in the master to `llvm::sys::Process::Exit(val);`
https://github.com/llvm/llvm-project/blob/master/lld/Common/ErrorHandler.cpp#L79

mati865:
> 
> 
> They have changed the _exit call in the master to `llvm::sys::Process::Exit(val);`
> [https://github.com/llvm/llvm-project/blob/master/lld/Common/ErrorHandler.cpp#L79](https://github.com/llvm/llvm-project/blob/master/lld/Common/ErrorHandler.cpp?rgh-link-date=2020-11-04T20%3A14%3A03Z#L79)

Yeah in https://github.com/llvm/llvm-project/commit/f2efb5742cc9f74ad73987760651e3d23894a416. I kind of should have connected it with hangs but I haven't checked the diff deeply.

Do we know why this was not happening when linking LLD to libc++?
From what I understood on mingw-w64 mailing list UCRT should be not affected, maybe this should be gated on MSVCRT version?

jeremyd2019:
> Do we know why this was not happening when linking LLD to libc++?

winpthreads has a pthread_mutex around a mapping between a pthread thread id and a structure that includes the thread HANDLE.  That mutex was being held by some thread when all threads other than the one calling exit/_exit were terminated.  Then when std::thread's join was called from static destructors, pthread_join tried to acquire that mutex, and hung forever as a result.

Using native primitives instead of winpthreads could have avoided this in a couple of ways.  If libc++'s std::thread implementation just keeps a thread HANDLE, and join waits on it, that would have just returned, because the thread would be dead (terminated during process shutdown).  If there were a Windows-native lock involved, Windows has workarounds to avoid such a deadlock situation.  (from what I read, a Mutex object would return WAIT_ABANDONED, and a CRITICAL_SECTION would, in Vista and later, terminate the process if it tried to enter a critial section that was held by a thread terminated during process shutdown).  winpthread's pthread_mutex does not have such a workaround, and ultimately winds up waiting on an Event object, that there's nobody left to signal.

> From what I understood on mingw-w64 mailing list UCRT should be not affected, maybe this should be gated on MSVCRT version?

UCRT changed how destructors are invoked, but I believe this would still happen there.  The comment I posted from UCRT says they only changed to TerminateProcess for windows store apps, not for 'normal' Win32 executables.

jeremyd2019:
> They have changed the _exit call in the master to `llvm::sys::Process::Exit(val);`
> https://github.com/llvm/llvm-project/blob/master/lld/Common/ErrorHandler.cpp#L79

That explains why I got confused and pointed to that instead of ErrorHandler at first 😉

mati865:
Thanks for the explanation.

> UCRT changed how destructors are invoked, but I believe this would still happen there (because of how things work on Windows, this doesn't affect DLLs, only the EXE). The comment I posted from UCRT says they only changed to TerminateProcess for windows store apps, not for 'normal' Win32 executables.

This is the comment said by Martin that I'm thinking of:
> If you're using UCRT, the callback (tls_atexit_callback in tls_atexit.c) 
registered by _register_thread_local_exe_atexit_callback gets executed. 
This is a callback that gets executed _before_ running any atexit 
functions, and allows UCRT builds to get TLS destructors executed at the 
right spot. Unfortunately, the emulation of 
_register_thread_local_exe_atexit_callback for msvcrt.dll isn't perfect - 
maybe it's possible to make it better somehow?

jeremyd2019:
> This is the comment said by Martin that I'm thinking of:
> 
> > If you're using UCRT, the callback (tls_atexit_callback in tls_atexit.c)
> > registered by _register_thread_local_exe_atexit_callback gets executed.
> > This is a callback that gets executed _before_ running any atexit
> > functions, and allows UCRT builds to get TLS destructors executed at the
> > right spot. Unfortunately, the emulation of
> > _register_thread_local_exe_atexit_callback for msvcrt.dll isn't perfect -
> > maybe it's possible to make it better somehow?

UCRT improves the order in which destructors are run (tls destructors before global destructors), but does not help the fact that DLL destructors are run LATE in process shutdown, which is the cause of this deadlock situation.

mati865:
Oh.

I'll see with current master if this is still an issue, maybe this patch should be sent upstream.

jeremyd2019:
I'm running a test right now, I think the patch can be reduced to just `exitLld`

jeremyd2019 (approved):
I'm done with the changes I wanted to make, and llvm build with the latest CI-generated binaries succeeded without deadlock.

lazka:
@adrpo  @mati865 any objections to merging this?

adrpo:
Ok from my side!

mati865:
You can add `fixes #7190` in PR body to automatically close the issue.

No objections from me to use this patch here.

lazka:
great! :partying_face:

oscarfv:
> > "failure" was uploading sources artifact. Nothing important
> 
> We should probably remove the source upload. They fail a lot because of the size and don't have much value. Or allow them to fail.

If you are talking about source uploads for all packages and not just Clang, it is important to note that it is required to comply with the GPL.

lazka:
> If you are talking about source uploads for all packages and not just Clang, it is important to note that it is required to comply with the GPL.

Yeah, I was referring to the CI pipeline artifacts, not about the source packages published on the repo.

adrpo:
Just a small question, who builds and uploads the msys2/mingw packages to the repo?

jeremyd2019:
> Just a small question, who builds and uploads the msys2/mingw packages to the repo?

Packages are built by [msys2-autobuild](https://github.com/msys2/msys2-autobuild/blob/master/README.md#automated-build-process), and uploaded manually (generally by @elieux)

PR #7235, opened 2020-11-05
binutils: Remove libintl dependency

orgads:
It was added in fbb2df5f81d3d983e006ffb97b56f34d6a890d60 (Build gold linker)
but adds no real value, and adds a lib dependency which slows down
process startup.

### DIFF
--- a/mingw-w64-binutils/PKGBUILD
+++ b/mingw-w64-binutils/PKGBUILD
@@ -6,7 +6,7 @@ _realname=binutils
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.35.1
-pkgrel=2
+pkgrel=3
 pkgdesc="A set of programs to assemble and manipulate binary and object files (mingw-w64)"
 arch=('any')
 url="https://www.gnu.org/software/binutils/"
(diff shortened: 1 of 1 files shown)


mati865:
Should gold even be enabled?
It does not support MinGW and likely never will.

orgads:
Good point. I also thought it should be removed. lld is a much better alternative anyway, and it actually works :).

I'll open a separate PR for removing it.

lazka:
> which slows down process startup.

Is this really a problem? just wondering..

orgads:
I actually need it for another use-case. I link against libbfd.a, and I don't want to depend on libintl.

orgads:
But using libbfd.a is so extreme I doubt anyone else does it, so I didn't write it in the commit message ;)

lazka:
> I don't want to depend on libintl.

I understand that ideally msys2 would provide out of the box what you need, but this isn't a very good argument.

orgads:
But it was initially disabled, and was later enabled without reasoning. Why enable it?

lazka:
> But it was initially disabled, and was later enabled without reasoning. Why enable it?

I don't know. Maybe because it's closer to the upstream defaults (it's enabled by default)

oscarfv:
> But it was initially disabled, and was later enabled without reasoning. Why enable it?

MSYS2 VC history contains lots of unexplained changes (I would say that most changes are unexplained). Most likely it was disabled while debugging some problem and it was left there unnoticed until the package required new work. Or at some point libintl stopped working...

Obviously, `--disable-nls` turns off localized output. Personally, I very much prefer tools that speak to me in English, so this change is fine with me, but it is a feature that might be appreciated by other users. Just sayin'.

orgads:
Ok, I won't insist on this PR.

Opened a separate PR for removing gold: https://github.com/msys2/MINGW-packages/pull/7253

lazka:
ok, closing for #7253

PR #7241, opened 2020-11-06
Merge GDB for embedded targets (arm-none-eabi, riscv64-unknown-elf)

Fabien-Chouteau:


mingw-w64-embedded-gdb/PKGBUILD
```diff
@@ -3,17 +3,17 @@
 # Contributed by: nanodude
 
 _realname=gdb
-_target=arm-none-eabi
+_all_targets="arm-none-eabi riscv64-unknown-elf"
 
 pkgbase=mingw-w64-${_target}-${_realname}
```
lazka, line 8: there is no _target and mingw-w64-gdb already exists

orgads:
I recently added gdb-multiarch package (didn't notice the embedded-gdb). Doesn't it work for you?

Fabien-Chouteau:
Hi @orgads, I don't think it covers the same usages. Many scripts or IDEs will try to start the binaries `arm-none-eabi-gdb` or riscv64-unknown-elf-gdb`, so they are needed.

orgads:
And can't they be just symlinks to gdb-multiarch? It should support all targets. IIRC the only thing I needed for using it with linux binaries was `set osabi GNU/Linux`.

Fabien-Chouteau:
@orgads do you know how to do that?

orgads:
How to do what? Create symbolic links?

It's a bit more complicated, since you do need to have setosabi GNU/Linux somewhere.

Fabien-Chouteau:
> How to do what? Create symbolic links?

These are MINGW packages so they should be usable outside of the `msys2` shell, as far as I understand. In that case, how do you create symlinks in Windows?

GitMensch:
What about creating those as scripts without line ending that just execute `gdb-multiarch -iex 'set osabi GNU/Linux' $*` (or for windows as `.cmd` with `%*`)? They can be directly created in the install step.

lazka:
Is there anything left to do or decide here? Or should we close this?

lazka:
It looks to me as if Linux distros also just provide gdb-multiarch? So I'd assume most tools handle that somehow.

So, if anyone ends up here, please try gdb-multiarch, and if that doesn't work for you please create an issue.

Then we can have a look if providing wrappers etc helps.

GitMensch:
Agreed, especially as mingw-w64-embedded-gdb was deleted in the meantime.
If we see that the manual setting of ABI is preferable, we can create wrapper scripts that execute gdb-multiarch with the right `-ex` options to set it.

PR #7245, opened 2020-11-06
precice: Update PKGBUILD

3rav:
https://precice.discourse.group/t/precice-and-mingw-packages/382/2


Biswa96:
Please try to use proper _precise_ commit message. It helps to identify the commit and provides a summary of what changes are being done. For example, here the commit message may be like - 'precice: Use Ninja, enable msmpi'.

3rav:
OK, I will try

PR #7246, opened 2020-11-06
New package: petsc

okhlybov:
This recipe builds {SDCZ} x {sequential, OpenMP and MPI} static and DLL PETSc builds (12 variants in total) with PkgConfig support.

### DIFF
--- a/mingw-w64-petsc/PKGBUILD
+++ b/mingw-w64-petsc/PKGBUILD
@@ -3,10 +3,13 @@
 _realname=petsc
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}" "${MINGW_PACKAGE_PREFIX}-${_realname}-build")
+# The `petsc-build` package contains the distilled PETSc build tree
+# required at compile time by some packages (such as SLEPc).
+# Regular users should use the `petsc` package.
 pkgver=3.14.1
 pkgrel=1
 arch=('any')
-pkgdesc="Sparse iterative (non)linear solver package (mingw-w64)"
+pkgdesc='Sparse iterative (non)linear solver package (mingw-w64)'
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-gcc-libgfortran"	
          "${MINGW_PACKAGE_PREFIX}-openblas"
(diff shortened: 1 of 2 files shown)


Biswa96:
`python` should be `${MINGW_PACKAGE_PREFIX}-python`

okhlybov:
> 
> 
> `python` should be `${MINGW_PACKAGE_PREFIX}-python`

It might not work with MinGW version of Python. At least it didn't for awhile.

3rav:
Do you plan to add SLEPc as well?

okhlybov:
> 
> 
> Do you plan to add SLEPc as well?

Yes, SLEPc is underway.

mingw-w64-petsc/PKGBUILD
```diff
@@ -0,0 +1,191 @@
+# Contributor: Oleg A. Khlybov <fougas@mail.ru>
+
+_realname=petsc
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}" "${MINGW_PACKAGE_PREFIX}-${_realname}-build")
```
lazka, line 5: What's this build package about? It seems to contain only source code?

lazka:
just wondering, is there an easy way to build more things shared here? The package is quite large..

mingw-w64-petsc/PKGBUILD
okhlybov: This one contains the distilled build tree for PETSc with source code, configuration data and libraries required to build the SLEPc package. It is a compile time dependency to avoid heavy patching of the SLEPc build system.
The point is, the SLEPc is so tightly bound to the PETSc to the point it uses some of its internal headers not included into the destination package.

Therefore there are two options:
1) Use an auxillary package such as petsc-build.
2) Build PETSc and SLEPc at once off a single recipe.

lazka: I see, thanks. If it really is that tightly coupled then (2) sounds like a good idea. Would the make things easier or harder for you?

okhlybov: As of now the SLEPc recipe for (1) is already done.
I can try to make a joint recipe although it will be a bit harder. The downside of (2) is a 4 hrs build time which might hinder the maintenance cycle when new package versions will come up.

okhlybov: I see no real reason against (1) because no one will need this petsc-build package except the CI building SLEPc.
The users will be perfectly happy with the base petsc package.

lazka: ok. Can you then maybe add an inline comment around "pkgname" that shortly explains what "-build" is used/needed for.

PR #7253, opened 2020-11-08
binutils: Remove gold linker

orgads:
It doesn't support MinGW, and is generally useless on Windows.

This reverts commit fbb2df5f81d3d983e006ffb97b56f34d6a890d60.


lazka:
@Alexpux any thoughts on this? Since you added that change.

mingw-w64-binutils/PKGBUILD
```diff
@@ -138,7 +126,6 @@ package() {
   #install -m644 ${srcdir}/${_realname}-${pkgver}/include/libiberty.h ${pkgdir}${MINGW_PREFIX}/include
   #install -m644 ${srcdir}/${_realname}-${pkgver}/include/demangle.h ${pkgdir}${MINGW_PREFIX}/include
 
-  find ${pkgdir}${MINGW_PREFIX}/share -type f -iname "opcodes.mo" -o -iname "bfd.mo" | xargs -rtl1 rm
```
lazka, line 141: looks like this results in file conflicts with gdb?

mingw-w64-binutils/PKGBUILD
```diff
@@ -108,10 +99,7 @@ build() {
     --enable-nls \
     --disable-rpath \
     --disable-multilib \
-    --enable-install-libiberty \
-    --enable-plugins \
-    --enable-gold \
```
lazka, line 113: What's with all the other changes?

mingw-w64-binutils/0110-binutils-mingw-gnu-print.patch
```diff
@@ -110,20 +110,6 @@ diff -urN binutils-2.27.orig/binutils/strings.c binutils-2.27/binutils/strings.c
  # else
  		printf ("%7I64x ", (unsigned long long) start);
 diff -urN binutils-2.27.orig/gas/as.h binutils-2.27/gas/as.h
---- binutils-2.27.orig/gas/as.h	2016-08-03 09:36:51.000000000 +0200
```
lazka, line 113: This doesn't seem gold related

mingw-w64-binutils/0002-check-for-unusual-file-harder.patch
```diff
@@ -1,78 +0,0 @@
-Configure scripts tend to do things like this:
```
lazka, line 1: Is this gold related?

mingw-w64-binutils/PKGBUILD
orgads: They were all part of the same commit. Are they unrelated? I thought enable-plugins was needed for enable-gold.

lazka: I don't know. I think it's clear now that the commit you reverted wasn't split up as much as it would have been :)

orgads: Ok, I'll minimize this patch.

orgads:
Better now?

lazka:
Yes, thanks. What does `enable-plugins` change?

Let's wait for @Alexpux to respond

orgads:
Reverted the `enable-plugins` too.

lazka:
Thanks!

lazka:
@mati865 ok to remove gold?

mati865:
AFAICT gold will never support non-ELF targets so there is no point in keeping it. We should wait for Alexey though.

lazka:
I'll merge this in two days if there is no response. I can't find anything re gold + Windows.

PR #7260, opened 2020-11-09
gdk-pixbuf: Update to 2.42.0; Add jp2-pixbuf-loader

lazka:


kmilos:
@lazka Since you're already registered at https://notabug.org/necklace/jp2-pixbuf-loader, could you please also raise an issue to improve some other meson recipe aspects as well: use shared module instead of shared lib, and consequently not link tests to it. These were e.g. recently done for https://github.com/aruiz/webp-pixbuf-loader/commits/mainline/meson.build

Necklaces:
@kmilos I can't promise to answer quickly, but I do have an account you can @ here as well :) 

I think I have fixed the issues you are addressing, see:
- meson.build: https://notabug.org/necklace/jp2-pixbuf-loader/src/cbd8c7a9c7608fee0cf417b1b0063a232b5af564/meson.build
- tests/meson.build https://notabug.org/necklace/jp2-pixbuf-loader/src/cbd8c7a9c7608fee0cf417b1b0063a232b5af564/tests/meson.build

Feel free to @ me if there's more missing

kmilos:
Thanks for the heads-up @Necklaces - will you be tagging and releasing 0.0.3 then?

PR #7269, opened 2020-11-11
gcc: Install libbacktrace

orgads:
Found several requests[1][2][3].

It is built anyway with gcc, but not installed.

[1] https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/oga0ug%24kno%241%40blaine.gmane.org/#msg35861445
[2] https://github.com/boostorg/stacktrace/issues/14
[3] https://sourceforge.net/p/mingw/bugs/2356/

### DIFF
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -22,7 +22,7 @@ pkgver=10.2.0
 #_majorver=${pkgver:0:1}
 #_sourcedir=${_realname}-${_majorver}-${_snapshot}
 _sourcedir=${_realname}-${pkgver}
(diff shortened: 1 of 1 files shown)


orgads:
As an alternative, we can build it from https://github.com/ianlancetaylor/libbacktrace as a separate package. WDYT?

lazka:
Upstream bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66570

Given the author of libbacktrace also hints at using the separate project I'd be in favor of using that if there are no downsides. Thoughts?

lazka:
Closing in favor of #7421

Issue #7265, opened 2020-11-10
Package-request: rtmidi


BLumia:
Hello msys2 devs,

Would like to request the `rtmidi` package.

https://github.com/thestk/rtmidi

No luck to build myself. pkg-config can find the package I built but seems not able to use, so sorry I cannot send a PR for now..

Thanks.

Biswa96:
Just did this:

```sh
git clone --depth=1 https://github.com/thestk/rtmidi.git
cd rtmidi; mkdir bin; cd bin
cmake.exe -G Ninja ..
cmake --build .
```

Make sure `echo $MSYSTEM` returns MINGW64 or MINGW32.

BLumia:
Seems the same but I'm not using ninja as generator, I will try again later. Thanks for the help!

btw is it possible to add that package to msys2 mingw repo in the future? Thanks again.

Biswa96:
> is it possible to add that package to msys2 mingw repo in the future?

IMHO, it depends on if its important or if other projects/programs need it. Last decision is up to maintainers.

BLumia:
Thank you very much for the info!

BLumia:
Just get my build working, and would like to leave a note here too for MSYS2 newbies like me.

To let the install target install cmake module files to the right place so the `cmake` in MSYS2 could find the package you just built, set `CMAKE_INSTALL_PREFIX` to `$MINGW_PREFIX`.

```
cmake .. -G Ninja -DCMAKE_INSTALL_PREFIX=$MINGW_PREFIX
```

conversy:
@Biswa96 I chose this package over other midi-related ones for my projects as it's available on multiple platforms (linux, macos, win) and there might be emscripten support in the rep. I support adding this package, if it's the right place to do so.

Biswa96:
I have added the PKGBUILD file for this package, see the linked pull request. If possible please test the package from that GitHub Actions artifacts.

PR #7271, opened 2020-11-11
rtmidi: Add package

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/7265


Biswa96:
@lazka In the pkg-config CI builds, the configure script output shows unusual error messages about header files. But those are neither present in pkgconf CI build nor in my PC's msys2 installation.

lazka:
Do you mean the `avrt.h` warnings?

PR #7274, opened 2020-11-12
Update CGAL to version 5.1.1

### LINKED ISSUE #6263: [cgal] Cmake "refuse" to use CGAL
Hello.

I get the following error when trying to generate build files for my project using `MSYS2` and `CMake`:

```
-- Visual Leak Detector (VLD) is not found.
-- Using header-only CGAL
CMake Warning at CMakeLists.txt:63 (FIND_PACKAGE):
  Found package configuration file:

    C:/tools/msys64/mingw64/lib/cmake/CGAL/CGALConfig.cmake

  but it set CGAL_FOUND to FALSE so package "CGAL" is considered to be NOT
  FOUND.


CMake Error at CMakeLists.txt:67 (MESSAGE):
  CGAL not found !


```
CGAL is installed using `mingw-w64-x86_64-cgal`. My `CMakeLists.txt` contains notably:
```
FIND_PACKAGE(CGAL)
IF(CGAL_FOUND)
    MESSAGE(STATUS "CGAL found")
ELSE()
    MESSAGE(FATAL_ERROR "CGAL not found !")
ENDIF()
```

ImperatorS79:
Fix #6263


lazka:
The package content is wrong now and looks like this: `/a/_temp/msys/msys64/mingw64/bin/`

ImperatorS79:
Strange. It does not do that on my local build.

lazka:
Probably because you removed MSYS2_ARG_CONV_EXCL which converts the prefix

ImperatorS79:
I updated the PKGBUILD accordingly. Wait and see ^^.

lazka:
I have a hard time understanding your goals. you mix make with mingw make, remove the relocation support, change some build option and update the package.

Please try to communicate what you are trying to achieve and how.

ImperatorS79:
Sorry for this. I am just trying to update CGAL to version 5.1. Also, I had trouble in march 2020 using CGAL with cmake. I did not have any problem with the build of CGAL 5.1 that I have done locally. I thus decided no to use the patch, seems I did not see how it was usefull (and since this patch reference CGAL 4.5, my guess is that it is totally outdated).

ImperatorS79:
Comments applied. Reagarding the -DCGAL_HEADER_ONLY=OFF, it allows the user to choose if he want to use CGAL as a dynamic lib or as a header only lib by building the dll. I personnaly prefer to have the choice, but it can be reverted if you want.

ImperatorS79:
So I have finaly decided that it was better to leave it header only, so that the user is not forced to install qt when he just wants to use cgal alone.

Summary of this pr:
- Update CGAL to 5.1
- remove unused options in PKGBUILD
- remove now useless cmake patch

PR #7276, opened 2020-11-12
New package: sol2

ImperatorS79:


lazka:
Same issue as with the other package. Why are you using "MinGW Makefiles"?

ImperatorS79:
I just used what I was used to use ^^. I will have a look right now.

lazka:
We use mingw32-make only in a handful of packages, so my guess is it likely not the right thing to use :)

ImperatorS79:
Comments applied.

ImperatorS79:
@lazka I think everything is ready now.

PR #7278, opened 2020-11-13
darktable: build with ninja

lazka:
@Biswa96


Biswa96:
Shouldn't we discuss this about in that upstream issue?

lazka:
Yes, good point, added something here: https://github.com/darktable-org/darktable/issues/6778#issuecomment-726796371

PR #7279, opened 2020-11-13
ode: Add package

rickertm:
Adds a new package for the Open Dynamics Engine (ODE), a library for physics simulation and collision detection.

### DIFF
--- a/mingw-w64-ode/PKGBUILD
+++ b/mingw-w64-ode/PKGBUILD
@@ -1,0 +1,64 @@
+# Maintainer: Markus Rickert <rickert@fortiss.org>
+
+_realname=ode
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=0.16.2
+pkgrel=1
+pkgdesc="A library for simulating articulated rigid body dynamics (mingw-w64)"
+arch=('any')
+url="https://bitbucket.org/odedevs/ode"
+license=('LGPL', 'BSD')
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+options=('strip' 'staticlibs')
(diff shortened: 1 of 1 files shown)


Biswa96:
Some comments:

* If there is no build issue `strip` and `buildflags` options can be enabled.
* There is no `BUILD_STATIC_LIBS` option in CMake files. It may be removed.
* With `ODE_WITH_DEMOS` enabled, the package builds fine. I may be enabled.
* Summary: In my opinion, here is the diff with current PKGBUILD file which can be added:

```diff
diff --git a/mingw-w64-ode/PKGBUILD b/mingw-w64-ode/PKGBUILD
index 01278caa7..f43dbe9a8 100644
--- a/mingw-w64-ode/PKGBUILD
+++ b/mingw-w64-ode/PKGBUILD
@@ -13,7 +13,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
-options=('!strip' 'staticlibs' '!buildflags')
+options=('strip' 'staticlibs')
 source=("https://bitbucket.org/odedevs/ode/downloads/${_realname}-${pkgver}.tar.gz")
 sha256sums=('b26aebdcb015e2d89720ef48e0cb2e8a3ca77915f89d853893e7cc861f810f22')
 
@@ -34,8 +34,8 @@ build() {
     -GNinja \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
     "${extra_config[@]}" \
-    -DBUILD_{SHARED,STATIC}_LIBS=ON \
-    -DODE_WITH_DEMOS=OFF \
+    -DBUILD_SHARED_LIBS=ON \
+    -DODE_WITH_DEMOS=ON \
     ../${_realname}-${pkgver}
 
   ${MINGW_PREFIX}/bin/cmake --build .
```

rickertm:
Thank you for the feedback, I've updated the patch accordingly. I only disabled the demos to reduce build time, as these are not installed.

Biswa96:
> I only disabled the demos to reduce build time

No problem, we've [6 hours][1] 😉 

[1]:https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

rickertm:
Should be fine then :smile:

PR #7298, opened 2020-11-18
Add various Ada libraries from AdaCore

Fabien-Chouteau:
And fix gprbuild for x86.

### DIFF
--- a/mingw-w64-gnat_util/PKGBUILD
+++ b/mingw-w64-gnat_util/PKGBUILD
@@ -1,0 +1,26 @@
+# Maintainer: Fabien Chouteau <chouteau@adacore.com>
+
+_realname=gnat_util
+pkgbase=mingw-w64-${_realname}
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+pkgver=2020.0.0
+pkgrel=1
+
+pkgdesc="Internal GNAT compiler components (mingw-w64)"
+arch=('any')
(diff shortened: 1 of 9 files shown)


Biswa96:
Some comments:

1. gnat_util package has unusual version. If possible try to use `_commit` (custom) and `pkgver` variables in commands.
2. `PROCESSORS=$(nproc)` options may be removed. The jobs count is already mentioned in `$MAKEFLAGS` in `/etc/makepkg*` files.
3. For license files, I would recommend to use `install` command instead of `cp`.

Fabien-Chouteau:
Thank you @Biswa96 

> * gnat_util package has unusual version. If possible try to use `_commit` (custom) and `pkgver` variables in commands.

I think I see what you mean. I will submit a change.

> * `PROCESSORS=$(nproc)` options may be removed. The jobs count is already mentioned in `$MAKEFLAGS` in `/etc/makepkg*` files.

`PROCESSORS` is actually used when calling the Ada builder tool `gprbuild`. So this is different than `make`'s `-j` option.

> * For license files, I would recommend to use `install` command instead of `cp`.

Will do.

Biswa96:
O, another thing. Try to use double quotes like `"${srcdir}"` and `"${pkgdir}"`.

Fabien-Chouteau:
`actions/upload-artifact@v2` failed, can someone restart the workflow?

PR #7304, opened 2020-11-20
icestorm: add 0.0

umarcor:
This PR adds package `icestorm`: https://github.com/YosysHQ/icestorm

### DIFF
--- a/mingw-w64-icestorm/PKGBUILD
+++ b/mingw-w64-icestorm/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=icestorm
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=da52117
+pkgver=0.0
 pkgrel=1
 pkgdesc="Simple tools for analyzing and creating bitstream files for Lattice iCE40 FPGAs (mingw-w64)"
 arch=('any')
@@ -13,7 +13,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-libftdi")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "flex")
 
-source=("icestorm::git://github.com/YosysHQ/${_realname}.git#commit=${pkgver}")
+source=("icestorm::git://github.com/YosysHQ/${_realname}.git#commit=da52117")
 sha256sums=('SKIP')
 
 build() {
--- a/mingw-w64-icestorm/PKGBUILD
+++ b/mingw-w64-icestorm/PKGBUILD
@@ -1,0 +1,26 @@
+_realname=icestorm
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=da52117
+pkgrel=1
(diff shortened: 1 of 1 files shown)


Biswa96:
Some comments:
1. Add a Maintainer comment at start.
2. If possible, a tarball can be used instead of git repository.
3. The `-j` option can be removed from `make` command. It is passed automatically with `$MAKEFLAGS`.

umarcor:
> 1. Add a Maintainer comment at start.

Done.

> 2. If possible, a tarball can be used instead of git repository.

Unfortunately, `codeload.github.com` does not provide tarballs for untagged commits, and the "regular" tarballs are rate limited. That's why I think that using git as a source is "safer". 

> 3. The `-j` option can be removed from `make` command. It is passed automatically with `$MAKEFLAGS`.

Done.

mingw-w64-icestorm/PKGBUILD
```diff
@@ -0,0 +1,28 @@
+# Maintainer: umarcor <unai.martinezcorral@ehu.eus>
+
+_realname=icestorm
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=da52117
```
lazka, line 6: That's not a proper version number. We need something that can be incremented over time so pacman knows when to update things.

mingw-w64-icestorm/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+         "${MINGW_PACKAGE_PREFIX}-libftdi")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "flex")
+
+source=("icestorm::git://github.com/YosysHQ/${_realname}.git#commit=${pkgver}")
+sha256sums=('SKIP')
+
+build() {
+  cd "${srcdir}/${_realname}"
+  sed -i "s/-ggdb //;" Makefile
+  make STATIC=1
```
lazka, line 22: You need to set the build prefix to MINGW_PREFIX, otherwise this installs into /usr/local

lazka:
It still installs to /usr/local...

umarcor:
:sweat: Hope it's ok now...

lazka:
It's now installing to /mingw64/usr/local instead of  /mingw64

umarcor:
Please, insert huge facepalm here, and accept my apologies for the disturbances.

lazka:
two small things:

* seems it installs various scripts using python, should python be a dependency?
* it installs a Python module to /bin instead of the standard place, I guess that's upstreams fault.. so nothing we can do about it..

umarcor:
> * seems it installs various scripts using python, should python be a dependency?

icestorm includes several tools: fpga architecture documentation, bitstream packaging tools, RAM modification tools, FPGA/board programming tools... Python is only required when the FPGA architecture documentation resources are used, and that is done by a third party tool (nextpnr), which is the one that depends on Python. Therefore, I don't think it's necessary to make Python a dependency of icestorm itself.

> * it installs a Python module to /bin instead of the standard place, I guess that's upstreams fault.. so nothing we can do about it..

I believe it is to be considered a system tool. That is, the fact that it's written in Python is expected to be ignored by users. Anyway, I can ask about it upstream and fix/improve it in a separate PR, if you are ok with that.

lazka:
ok. it contains 12 tools under /bin that depend on python.. if you are ok with those being broken..., your choice :)

PR #7328, opened 2020-11-25
Ffmpeg: build both shared and static libs

ImperatorS79:
The patch directly comes from here: http://git.videolan.org/?p=ffmpeg.git;a=commitdiff;h=7c59e1b0f285cd7c7b35fcd71f49c5fd52cf9315. It seems srt was updated and removed some deprecated constant.

I removed the pgp key check because it did not work on my end but I can add it again if someone explains me how it works ^^.

EDIT: without the patch it did not compile locally

### DIFF
--- a/mingw-w64-ffmpeg/PKGBUILD
+++ b/mingw-w64-ffmpeg/PKGBUILD
@@ -198,6 +198,9 @@ package() {
   
   cd "${srcdir}/build-${MINGW_CHOST}-shared"
   make DESTDIR="${pkgdir}" install
+  
+  rm -f ${pkgdir}/${MINGW_PREFIX}/lib/*.def
+  rm -f ${pkgdir}/${MINGW_PREFIX}/bin/*.lib
 
   #find ${pkgdir}${MINGW_PREFIX}/bin -type f -name "*.exe" -exec objcopy --subsystem console {} \;
 
--- a/mingw-w64-ffmpeg/PKGBUILD
+++ b/mingw-w64-ffmpeg/PKGBUILD
@@ -60,12 +60,12 @@ prepare() {
 }
 
 build() {
-  [[ -d "${srcdir}/build-${MINGW_CHOST}-static" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}-static"
-  mkdir -p "${srcdir}/build-${MINGW_CHOST}-static" && cd "${srcdir}/build-${MINGW_CHOST}-static"
-
   # Fix using SRT headers
   CFLAGS+=" -DWIN32"
   CXXFLAGS+=" -DWIN32"
+  
+  [[ -d "${srcdir}/build-${MINGW_CHOST}-static" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}-static"
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}-static" && cd "${srcdir}/build-${MINGW_CHOST}-static"
 
   CC=${MINGW_PREFIX}/bin/gcc \
   CXX=${MINGW_PREFIX}/bin/g++ \
@@ -125,10 +125,6 @@ build() {
   
   [[ -d "${srcdir}/build-${MINGW_CHOST}-shared" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}-shared"
   mkdir -p "${srcdir}/build-${MINGW_CHOST}-shared" && cd "${srcdir}/build-${MINGW_CHOST}-shared"
-
-  # Fix using SRT headers
-  CFLAGS+=" -DWIN32"
-  CXXFLAGS+=" -DWIN32"
 
   CC=${MINGW_PREFIX}/bin/gcc \
   CXX=${MINGW_PREFIX}/bin/g++ \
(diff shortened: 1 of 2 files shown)


Biswa96:
May I summon a VLC developer just to verify? cc @robUx4

ImperatorS79:
The patch comes from here: https://trac.ffmpeg.org/ticket/8760

robUx4:
The addition seems mostly about adding the shared version but we don't use that in VLC. We don't use SRT in ffmpeg either but as a standalone module. So I'm not sure I can be of much help here.

mingw-w64-ffmpeg/PKGBUILD
```diff
@@ -47,18 +47,19 @@ depends=("${MINGW_PACKAGE_PREFIX}-aom"
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-pkg-config"
              "${MINGW_PACKAGE_PREFIX}-nasm")
-source=(https://ffmpeg.org/releases/${_realname}-${pkgver}.tar.xz{,.asc})
-validpgpkeys=('FCF986EA15E6E293A5644F10B4322F04D67658D8')
+source=(https://ffmpeg.org/releases/${_realname}-${pkgver}.tar.xz
+        srt.patch)
```
lazka, line 51: I've moved this to #7336

mingw-w64-ffmpeg/PKGBUILD
```diff
@@ -47,18 +47,19 @@ depends=("${MINGW_PACKAGE_PREFIX}-aom"
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-pkg-config"
              "${MINGW_PACKAGE_PREFIX}-nasm")
-source=(https://ffmpeg.org/releases/${_realname}-${pkgver}.tar.xz{,.asc})
-validpgpkeys=('FCF986EA15E6E293A5644F10B4322F04D67658D8')
```
lazka, line 51: You can use `--skippgpcheck` if you don't want to verify the signature. Please don't delete this.

mingw-w64-ffmpeg/PKGBUILD
```diff
@@ -121,6 +120,69 @@ build() {
...
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}-shared" && cd "${srcdir}/build-${MINGW_CHOST}-shared"
+
+  # Fix using SRT headers
+  CFLAGS+=" -DWIN32"
+  CXXFLAGS+=" -DWIN32"
+
+  CC=${MINGW_PREFIX}/bin/gcc \
+  CXX=${MINGW_PREFIX}/bin/g++ \
+  ../${_realname}-${pkgver}/configure \
+    --prefix=${MINGW_PREFIX} \
```
lazka, line 134: Not a fan of duplicating all those build flags. Can we share them between the builds?

mingw-w64-ffmpeg/PKGBUILD
```diff
@@ -121,6 +120,69 @@ build() {
       exit 1
     }
   make
+  
+  [[ -d "${srcdir}/build-${MINGW_CHOST}-shared" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}-shared"
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}-shared" && cd "${srcdir}/build-${MINGW_CHOST}-shared"
+
+  # Fix using SRT headers
+  CFLAGS+=" -DWIN32"
```
lazka, line 128: Why is this needed again?

mingw-w64-ffmpeg/PKGBUILD
ImperatorS79: Okay thanks for the info

mingw-w64-ffmpeg/PKGBUILD
ImperatorS79: I'll merge master into my branch once it is committed.

mingw-w64-ffmpeg/PKGBUILD
ImperatorS79: Great question but no idea ^^.

mingw-w64-ffmpeg/PKGBUILD
ImperatorS79: I do not think so. You need to configure each static and shared build separetly.

mingw-w64-ffmpeg/PKGBUILD
ImperatorS79: Oh, I did not see I copy-pasted this twice! Sorry ^^'.

mingw-w64-ffmpeg/PKGBUILD
lazka: I mean the configure options, not the call to configure itself

ImperatorS79: Yes, I understood that too. But I do not of a way to do that in bash.

lazka:
@ImperatorS79 I've pushed something. Does it look OK to you?

ImperatorS79:
Since it compiles it looks good ^^.

lazka:
if that's enough for you, ok :) we don't use static libs, so don't expect any testing in that area.

PR #7333, opened 2020-11-26
OpenVR: enable static build

ImperatorS79:


mingw-w64-openvr/PKGBUILD
```diff
@@ -49,11 +49,27 @@ build() {
...
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DBUILD_SHARED=OFF \
+    ../openvr-${pkgver}
+
+  ${MINGW_PREFIX}/bin/cmake --build . #VERBOSE=1
 }
 
 package() {
-  cd ${srcdir}/build-${MINGW_CHOST}
+  cd ${srcdir}/build-${MINGW_CHOST}-shared
```
lazka, line 68: Please install the shared after the static (to avoid files from the shared build being replaced)

mingw-w64-openvr/PKGBUILD
```diff
@@ -49,11 +49,27 @@ build() {
...
+  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --build . --target install
+  
+  cd ${srcdir}/build-${MINGW_CHOST}-static
+  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --build . --target install
 
   make DESTDIR=${pkgdir} install
```
lazka, line 74: Isn't this redundant?

lazka:
pkgrel needs to be increased

mingw-w64-openvr/PKGBUILD
ImperatorS79: I do not understand. How do you want want to install both shared and static build ?

lazka: Doesn't "cmake --target install" already install everything? why is "make install" needed as well?

ImperatorS79: The two build are independant, and thus cmake --target install will only install what has been built in each directory.

Yes the make install should be removed, bad copy paste

lazka:
Thanks!

PR #7337, opened 2020-11-27
cmake: update to 3.19.1, depends on pkgconf

Astrum-polaris:
Since pkgconf replaces pkg-config, let cmake depends on pkgconf

bzip2, xz, zlib, zstd are pulled in by libarchive, so I think it is not necessary to list them in dependencies

### DIFF
--- a/mingw-w64-cmake/PKGBUILD
+++ b/mingw-w64-cmake/PKGBUILD
@@ -4,7 +4,7 @@
 _realname=cmake
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.19.0
+pkgver=3.19.1
 pkgrel=1
 pkgdesc="A cross-platform open-source make system (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mingw-w64-cmake/PKGBUILD
```diff
@@ -4,29 +4,25 @@
...
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
-         "${MINGW_PACKAGE_PREFIX}-pkg-config"
+         "${MINGW_PACKAGE_PREFIX}-pkgconf"
          "${MINGW_PACKAGE_PREFIX}-curl"
          "${MINGW_PACKAGE_PREFIX}-expat"
          "${MINGW_PACKAGE_PREFIX}-jsoncpp"
          "${MINGW_PACKAGE_PREFIX}-libarchive"
          "${MINGW_PACKAGE_PREFIX}-libuv"
-         "${MINGW_PACKAGE_PREFIX}-rhash"
-         "${MINGW_PACKAGE_PREFIX}-zlib")
```
lazka, line 29: cmake links against zlib, so I think it needs to stay.

mingw-w64-cmake/PKGBUILD
```diff
@@ -85,7 +81,6 @@ build() {
     "${srcdir}"/${_realname}-${pkgver}/configure  \
     --prefix=${MINGW_PREFIX}                      \
     --system-libs                                 \
-    --no-system-expat                             \
```
lazka, line 88: I guess OK. What motivated you to change this?

lazka:
sounds good otherwise

mingw-w64-cmake/PKGBUILD
Astrum-polaris: Reverted that part all together

mingw-w64-cmake/PKGBUILD
Astrum-polaris: It was disabled a long time ago to fix build of cmake-git
Enabling this now doesn't cause build to fail, so I remove it

lazka: ok, thanks :+1:

PR #7352, opened 2020-11-29
icesprog: get tarball instead of using git

umarcor:
Upon request, the maintainer of icesprog released a new git tag: https://github.com/wuxx/icesugar/issues/19. In this PR, the recipe is changed for retrieving a tarball, instead of using git.

/cc @Biswa96

### DIFF
--- a/mingw-w64-icesprog/PKGBUILD
+++ b/mingw-w64-icesprog/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=icesprog
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.r67.8e33c6a
+pkgver=1.0
 pkgrel=1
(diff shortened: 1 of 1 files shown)


mingw-w64-icesprog/PKGBUILD
```diff
@@ -3,7 +3,7 @@
 _realname=icesprog
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.r67.8e33c6a
+pkgver=1.0
```
lazka, line 6: 1.0 is considered "less than" 1.0.r67.8e33c6a, so that wont work sadly

umarcor: I changed it to `1.0.0`:

```sh
# vercmp 1.0 1.0.r67.8e33c6a
-1
# vercmp 1.0.0 1.0.r67.8e33c6a
1
```

lazka: :+1:

PR #7358, opened 2020-11-29
lis: update to 2.0.26 and disable building test programs

3rav:


mingw-w64-lis/PKGBUILD
```diff
@@ -37,6 +37,7 @@ build() {
     --enable-quad \
     --enable-f90 \
     --enable-fortran \
+    --disable-test \
```
Biswa96, line 40: Tests are build fine. This may not be required.

3rav: This reduces compile time and resulting size

Biswa96: Some users may want to run the tests. Also there is a `check()` function. Compile time is not so important in small packages. Are you sure compiled tests binaries install in the resultant package?

3rav: 2.0.24: 340 files vs 2.0.26: 283 files

not "bin" and "share/examples/lis" folder in 2.0.26

lazka: fair enough..

PR #7367, opened 2020-12-02
Add frei0r-plugins package

mingwandroid:


mingw-w64-frei0r-plugins/PKGBUILD
```diff
@@ -0,0 +1,96 @@
...
+  cp README.txt README.md
+  cp INSTALL.txt INSTALL.md
+  cp TODO.txt TODO
+}
+
+build() {
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}"
+  cd "${srcdir}/build-${MINGW_CHOST}"
+
+  MSYSTEM=MINGW \
```
lazka, line 79: What's this about?

mingw-w64-frei0r-plugins/PKGBUILD
```diff
@@ -0,0 +1,96 @@
...
+
+  plain "General autotools fix (upstreaming .. hopefully)"
+  apply_patch_with_msg \
+    0002-Fix-out-of-source-tree-CXX-build-failures.patch
+
+  plain "MSYS2/mingw-w64 autotools fixes (upstreaming .. hopefully)"
+  apply_patch_with_msg \
+    0003-Support-MSYS2-mingw-w64.patch
+
+  cp AUTHORS.txt AUTHORS
```
lazka, line 67: Why is this needed?

mingwandroid: Because the tarball for the release has the wrong named files for the autoconfig build to succeed. I checked that they are all the same and the are. I think it's an upstream packaging issue (but I may have overlooked something).

lazka: weird, it has "foreign" set which usually makes it ignore those files.

btw, is this missing a ./autogen.sh call since you patched those files?

(I noticed that I start forgetting autotools things.. which might be a good sign)

mingwandroid: It was regenerating it anyway from configure because of dependencies, but I made it explicit now.

Ah I have a bit of a fondness for good old autotools you know. I probably first ran it in around 1991 and it impressed me back then! It's a case of picking a poison though with built tools really. I haven't checked meson out much yet but yeah anyway... I'm having to battle bazel for my job at present so this is a welcome distraction (from the endless downloading and building of seemingly all of Google's public source code).

Is there a stated official build-system preference order? This supports CMake too and that's actually what upstream devs use for their own MSYS2 builds, still, this seems to work. Next thing is to enable it in ffmpeg hopefully.

lazka: I'd say cmake/meson > autotools > everything else. Mostly cmake/meson because it also allows building with MSVC, so more people care about it and it's less likely to regress upstream. And autotools/make is sloooowww on Windows.

PR #7372, opened 2020-12-02
Update libtorrent-rasterbar to v2.0.1

Amphaal:
@Alexpux, do you think, since libtorrent v2.0 introduces API breakage (https://libtorrent.org/upgrade_to_2.0-ref.html), that we should rather make v2.0 available from another msys package, since v1.2 is still maintained ? (https://github.com/arvidn/libtorrent/releases/tag/v1.2.11)

Cheers !

### DIFF
--- a/mingw-w64-libtorrent-rasterbar/PKGBUILD
+++ b/mingw-w64-libtorrent-rasterbar/PKGBUILD
@@ -5,7 +5,7 @@
 _realname=libtorrent-rasterbar
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.2.10
+pkgver=2.0.1
 pkgrel=1
 pkgdesc="libtorrent is a feature complete C++ bittorrent implementation focusing on efficiency and scalability (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
We only have qbittorrent depending on it. Does qbittorrent support 2.x?

Amphaal:
https://github.com/qbittorrent/qBittorrent/pull/13783, Looks like it is planned at some point.

lazka:
I dont see any mention of 2.x there

Amphaal:
Mb, @FranciscoPombal was considering a v2.x port of `libtorrent` for vcpkg, not for qBittorrent _per se_. https://github.com/microsoft/vcpkg/pull/14293, @glassez also talked there about creating a distinct package for v2.x.

lazka:
You can bump pkgrel of qbittorrent in this PR to see if it builds.

FranciscoPombal:
Thanks for bringing me in the loop.

qBittorrent is still not ready for v2. A few commits enabling support for v2 have been recently merged, but we're not there yet.

I must say I don't think it's fair for qBittorrent to hold an entire ecosystem back, but, on the other hand, I doubt we're the only project still making active use of 1.2.x, and 1.2.x is still maintained upstream.

This is why I proposed the compromise solution of making v2 available as a separate package for now, in vcpkg (there is somewhat of a precedent in other packages). Is that feasible for msys2 as well?

As far as vcpkg is concerned, it's actually not the end of the world for qBittorrent if the libtorrent package were bumped to 2.0. We can quite easily work around that with a simple "overlay port" (https://github.com/microsoft/vcpkg/blob/master/docs/specifications/ports-overlay.md). If msys2 were to move forward with a single v2 package in the short term, is there a similar escape hatch available?

lazka:
@Amphaal ah, you mentioned the right people. sorry I didn't get that.

The rules for packaging multiple version in MSYS2 are similar to Linux distros, it's possible by creating a new package (with a new name) and adding a conflict to the old one, but ideally we would just wait a bit.

If you have a need for the new version then go ahead and create a "mingw-w64-libtorrent-rasterbar2" package.

Amphaal:
I'll then close this draft an follow @lazka instructions ! Thank you guys for your insights.

Biswa96:
> If you have a need for the new version then go ahead and create a "mingw-w64-libtorrent-rasterbar2" 

I also requested this type of thing for imagemagick version 6 required for inkscape. But Alexpux denied.

lazka:
> I also requested this type of thing for imagemagick version 6 required for inkscape. But Alexpux denied.

arch seems to have a trimmed down version with just the lib: https://www.archlinux.org/packages/extra/x86_64/libmagick6/

if that helps inkscape and doesn't conflict with 7, go ahead.

PR #7376, opened 2020-12-03
New package: fritzing

jeremyd2019:
My first attempt at packaging something.

> The Fritzing application is an Electronic Design Automation software with a low entry barrier, suited for the needs of makers and hobbyists. It offers a unique real-life "breadboard" view, and a parts library with many commonly used high-level components. Fritzing makes it very easy to communicate about circuits, as well as to turn them into PCB layouts ready for production. It is particularly popular among Arduino and Raspberry Pi users, and is widely used in education and creative tinkering.

https://fritzing.org
https://github.com/fritzing/fritzing-app

Major ~~outstanding~~[solved](https://github.com/msys2/MINGW-packages/pull/7376#issuecomment-738217175) issue is that Fritzing cannot find its files out of the box.  This can be worked around by launching `Fritzing -f /mingw{32,64}/share/fritzing`.  It passes path defines to the compiler, but I found that in order for `qmake` to allow `make install` to work properly I had to use `MSYS2_ARG_CONV_EXCL="PREFIX="`.  That results in that define using a path starting with /mingw64.  Not that the alternative would have been better, since that was an absolute path on the GHA builder...  Also, the code that uses that is ifdef'd out on `QT_OS_WIN`.  Is there some common solution for mingw programs to know their prefix, so they can find their installed files?

A couple of points of oddness about this package:
* Fritzing is no longer officially supported on 32-bit.  However, it does still build and seems to run fine.
* Fritzing keeps its parts database in a git repository, and uses libgit2 to keep it up to date.  This resulted in a bit of a song-and-dance with `git` to turn the `fritzing-parts` checkout from makepkg into a self-contained git repo with the proper origin url and branch name.

### DIFF
--- a/mingw-w64-fritzing/PKGBUILD
+++ b/mingw-w64-fritzing/PKGBUILD
@@ -19,20 +19,24 @@ source=("${_realname}"::"git+https://github.com/fritzing/fritzing-app.git#tag=CD
         0001-Fix-libgit2-version-check-to-allow-building-with-ver.patch
         0002-Fix-casts-from-pointer-to-long-to-size_t-instead.patch
         0003-qmake-fixes-for-mingw.patch
+        0004-Look-in-.-share-fritzing-from-bin-dir-on-mingw.patch
         system-libgit2.patch)
 sha256sums=('SKIP'
             'SKIP'
             '98042c42cc10847f76285e0a84d258d9818571e695dd87dd6d52ef3ac387c7a7'
             '7a91295b4ab91edb4efba9d08a9f2a0cb2761da1b04e39eaca165eee53ac9e6b'
             '2d9f550ce25ddb78ac683d23c55b5b794a06eef3b022f686bad89651793c8423'
+            'fa2c0eecea654ca2f5412acec31bd0c871ddcf271886886c78e2dd07ee70bf80'
             'b9fcdd41207a5d8fd67d44c5d4b3348e25464815431a91b61111b1fd91f1495f')
 
 prepare() {
   cd "${srcdir}"/${_realname}
 
+  # upstream commit 472951243d70eeb40a53b1f7e16e6eab0588d079
   git apply "${srcdir}"/0001-Fix-libgit2-version-check-to-allow-building-with-ver.patch
   git apply "${srcdir}"/0002-Fix-casts-from-pointer-to-long-to-size_t-instead.patch
   git apply "${srcdir}"/0003-qmake-fixes-for-mingw.patch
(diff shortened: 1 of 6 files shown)


Biswa96:
> Is there some common solution for mingw programs to know their prefix

I am not sure about your desired solution. Extrapolating from the `/mingw{32,64}/share/fritzing` path, I think mingw-w64-mdloader\01-get-applet-path.patch file may help here.

Biswa96:
> Is there some common solution for mingw programs to know their prefix

I am not sure about your desired solution. Extrapolating from the `/mingw{32,64}/share/fritzing` path, I think `mingw-w64-mdloader/01-get-applet-path.patch` file may help here. BTW, if the patches are in upstream repository a little comment with the commit hash will be useful.

jeremyd2019:
> I am not sure about your desired solution. Extrapolating from the `/mingw{32,64}/share/fritzing` path, I think `mingw-w64-mdloader/01-get-applet-path.patch` file may help here.

Yeah, relative to the executable was kind of what I was thinking, too...

> BTW, if the patches are in upstream repository a little comment with the commit hash will be useful.

One was cherry-picked.  Comment added.  I'm planning to submit the other numbered patches as a pull request upstream, for mingw support.

jeremyd2019:
The patch switching to the unix behavior for looking for files on mingw, and adding `../share/fritzing` relative to to the executable's location as a candidate, seems to work well.  Also, I noticed that calling `uname -m` from qmake to determine whether it was a 32 or 64-bit build did not work correctly, despite my recollection that uname output was changed based on MSYSTEM.  I switched it to use the $MSYSTEM_CARCH environment variable instead.

mingw-w64-fritzing/PKGBUILD
```diff
@@ -0,0 +1,73 @@
...
+arch=('any')
+url="https://github.com/fritzing/fritzing-app"
+license=('GPL' 'CCPL:cc-by-sa-3.0')
+makedepends=("${MINGW_PACKAGE_PREFIX}-boost"
+             'git')
+depends=("${MINGW_PACKAGE_PREFIX}-qt5"
+         "${MINGW_PACKAGE_PREFIX}-libgit2"
+         "${MINGW_PACKAGE_PREFIX}-zlib")
+source=("${_realname}"::"git+https://github.com/fritzing/fritzing-app.git#tag=CD-498"
+        "${_realname}-parts"::"git+https://github.com/fritzing/fritzing-parts.git#branch=master"
```
lazka, line 18: needs a fixed commit

PR #7377, opened 2020-12-03
libtorrent-rasterbar2 package

Amphaal:
Following https://github.com/msys2/MINGW-packages/pull/7372, here is my first package pull request ever. @lazka, does it look A-OK to you ?


lazka:
Once things support v2 we will remove this package again.

PR #7381, opened 2020-12-05
godot_headers: fix duplication of repo for i686

Schala:
Previous behavior would put the "godot_headers-*" folder in `mingw32/include/godot`

Included godot-cpp too


lazka:
For some reason godot-cpp doesn't build anymore.

PR #7387, opened 2020-12-05
importlib-metadata: Update to 3.1.1

marcelotduarte:
cx_Freeze needs this version.


lazka:
We have Python 3.8, so in theory no one should be using it, but OK.

lazka:
ah, I guess it also adds more features from 3.9/10, never mind then.

marcelotduarte:
> We have Python 3.8, so in theory no one should be using it, but OK.

The version in py38 and py39 is buggy. I need the newest version.
https://github.com/python/importlib_metadata/issues/261

PR #7396, opened 2020-12-06
ffmpeg: Enable frei0r-plugins

mingwandroid:


mingwandroid:
Sorry @lazka, I just discovered that ffmpeg hardcodes the paths it looks for the frei0r plugins in. I am working on making that relocatable.

Do you think I should make the new search path relative to the calling DLL or the calling EXE (when it's been called by a DLL that is)?

lazka:
hm, I guess DLL would be less fragile, but not sure.

We need some documentation or relocation...

lazka:
Does it break ffmpeg, or does it just not find plugins?

Biswa96:
> Do you think I should make the new search path relative to the calling DLL or the calling EXE

Don't we already have that thingy in openssl?

mingwandroid:
No, it only breaks if you try to specify to use the plugins and it tells you what paths it looked in when it fails so the user can, for now, put them in one of those places.

mingwandroid:
No need for reversion, just, on my part, better testing.

lazka:
ok, thanks

mingwandroid:
> We need some documentation or relocation...

Agreed.

PR #7400, opened 2020-12-07
dfu-util: libwinpthread is a dep

umarcor:
Executing `dfu-util --help` fails if `*-libwinpthread-git` is not installed. See https://github.com/umarcor/MINGW-packages/runs/1509020890?check_suite_focus=true#step:4:73.

/cc @fauxpark

### DIFF
--- a/mingw-w64-dfu-util/PKGBUILD
+++ b/mingw-w64-dfu-util/PKGBUILD
@@ -4,23 +4,22 @@ _realname=dfu-util
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
 pkgver=0.10
-pkgrel=1
+pkgrel=2
 pkgdesc='Device firmware upgrade utilities (mingw-w64)'
 arch=('any')
 license=('GPL2')
 url='http://dfu-util.sourceforge.net/'
-depends=("${MINGW_PACKAGE_PREFIX}-libusb")
+depends=("${MINGW_PACKAGE_PREFIX}-libusb"
+         "${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
 source=("https://downloads.sourceforge.net/project/dfu-util/dfu-util-${pkgver}.tar.gz")
 sha256sums=('a03dc58dfc79c056819c0544b2a5970537566460102b3d82cfb038c60e619b42')
 
 build() {
     cd ${srcdir}/${_realname}-${pkgver}
-
     ./configure --prefix=${MINGW_PREFIX}
 }
 
 package() {
     cd ${srcdir}/${_realname}-${pkgver}
-
     make DESTDIR="$pkgdir" install
 }


fauxpark:
@umarcor This is very odd - the official release binary ([here](http://dfu-util.sourceforge.net/releases/)) contains no mention of libwinpthread, nor does the source code. So I don't understand where this dependency is coming from.

When trying to run the dfu-util.exe from the MSYS2 package outside of MSYS, with libusb-1.0.dll and libwinpthread-1.dll alongside, this error appears:
image

Any ideas?

fauxpark:
[Adding `LDFLAGS=-static` to the make command](http://dfu-util.sourceforge.net/build.html) fixes this, but...now libusb is statically linked too. That doesn't seem like a great solution - is there a way to only statically link libwinpthreads here?

jeremyd2019:
> [Adding `LDFLAGS=-static` to the make command](http://dfu-util.sourceforge.net/build.html) fixes this, but...now libusb is statically linked too. That doesn't seem like a great solution - is there a way to only statically link libwinpthreads here?

Completely removed from whatever you're really trying to accomplish, but:
`-Wl,-Bstatic -lpthread -Wl,-Bdynamic` ?
or `-l:libpthread.a` ?

fauxpark:
Nope :/ I still see this in dfu-util.exe, where I should only be seeing libusb-1.0.dll:
image
image

fauxpark:
I found this: https://sourceforge.net/p/msys2/mailman/message/35617060/
And amusingly, it works if I move `libpthread.dll.a` out of the way while building (no change when moving `libwinpthread.dll.a`).
image

umarcor:
@fauxpark in fact, it seems that the MINGW32 version is still failing, regardless of `mingw-w64-i686-libwinpthread-git` being installed: https://github.com/umarcor/MINGW-packages/runs/2923062947?check_suite_focus=true#step:4:76

I prepared a branch for testing. See https://github.com/umarcor/MINGW-packages/commits/dfu-util and https://github.com/umarcor/MINGW-packages/actions/runs/976450885.

fauxpark:
Interesting, when I try to run the mingw32 dfu-util in cmd I get:
image

Adding `-static-libgcc -static-libstdc++` to LDFLAGS as mentioned [here](https://stackoverflow.com/questions/4702732/the-program-cant-start-because-libgcc-s-dw2-1-dll-is-missing) appears to have no effect. Its mingw64 equivalent `libgcc_s_seh-1.dll` is nowhere to be found in the mingw64 dfu-util.exe :/

mati865:
Have you compiled all the dependencies with `-static-libgcc`?

fauxpark:
What do you mean by "all the dependencies"?

mati865:
Shared dependencies of dfu-util can require libgcc even if dfu-util was linked with static libgcc.

fauxpark:
Yes, it looks like libwinpthread is the one pulling it in, but only on i686. The reference to the libgcc DLL, and two symbols `__register_frame_info`/`__deregister_frame_info` immediately after are missing from the same area in the x86_64 build. I will see what happens if I try to compile i686 libwinpthread statically in the morning, but perhaps this gives a clue as to what is happening? I see there is a patch referencing those two symbols, but not sure what it's for.

mati865:
I don't think it's winpthreads:
```
$ ntldd /mingw32/bin/libwinpthread-1.dll
        KERNEL32.dll => not found
        msvcrt.dll => not found

$ ntldd /mingw64/bin/libwinpthread-1.dll
        KERNEL32.dll => C:\Windows\SYSTEM32\KERNEL32.dll (0x00000000004d0000)
        msvcrt.dll => C:\Windows\SYSTEM32\msvcrt.dll (0x00000000004d0000)
```

I have no idea about mentioned patch other than "it exists".

fauxpark:
Sorry, the patch I was referring to is this one: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-winpthreads-git/0001-Define-__-de-register_frame_info-in-fake-libgcc_s.patch
It actually refers to another patch for gcc which appears to have since been removed.

`libwinpthread-1.dll` from `mingw-w64-i686-libwinpthread-git 9.0.0.6246.ae63cde27-1`:
image

`libwinpthread-1.dll` from `mingw-w64-x86_64-libwinpthread-git 9.0.0.6246.ae63cde27-1`:
image

fauxpark:
Update: adding `extra_config+=("LDFLAGS=-static-libgcc")` to the PKGBUILD didn't change anything. Removing the patch also didn't work.

fauxpark:
Did some more digging in GCC itself, and found what may be the code that generates this:
https://github.com/gcc-mirror/gcc/blob/master/libgcc/config/i386/cygming-crtbegin.c#L114-L150

At this point I'm not really sure what the next step is. Is it a bug, or intentional? If it's a bug I can create a new issue to give it more visibility.

mati865:
I have no idea, crating new issue is good idea.

PR #7406, opened 2020-12-07
libiptcdata: Add package, for RawTherapee.

Biswa96:
Found that to compile RawTherapee one has to first compile libiptcdata [here][1]. What if we can provide libiptcdata as a package? /cc @Beep6581 @Desmis

[1]: https://rawpedia.rawtherapee.com/Windows#Download_and_build_libiptcdata

### DIFF
--- a/mingw-w64-libiptcdata/0001-libiptcdata-automake-remove-iptc-docs.patch
+++ b/mingw-w64-libiptcdata/0001-libiptcdata-automake-remove-iptc-docs.patch
@@ -1,0 +1,31 @@
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -2,7 +2,7 @@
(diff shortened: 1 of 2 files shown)


Beep6581:
Hi @Biswa96 

Yes, RawTherapee still depends on libiptcdata, until such a time as it adopts libexiv2 - see https://github.com/Beep6581/RawTherapee/issues/3801

I don't use Windows so can't comment on MSYS2 issues, but do let me know if there are any changes to the build process so that we can update the build instructions http://rawpedia.rawtherapee.com/Windows

lazka:
No shared library?

Biswa96:
No. autoconf shows warning about undefined symbol. Also the library is so tiny that static linking would not matter.

lazka:
fair enough

PR #7407, opened 2020-12-07
Fix the install path for CMake’s man pages

sboukortt:
I assume that this was an oversight, as `cmake` is the only package I have that seems to have installed anything directly in `/mingw64/share/man{1,7}` instead of `/mingw64/share/man/man*`.


sboukortt:
Forgot to bump `pkgrel`. This is now done.

PR #7415, opened 2020-12-08
ffmpeg: Make frei0r-plugins path look-up use a relative lookup

mingwandroid:


mingwandroid:
This also fixes some issues in the build script around the build folder names not containing the -static/-shared prefix and uses for loops instead of duplication.

mingw-w64-ffmpeg/0002-Win32-Add-path-relocation-to-frei0r-plugins-search.patch
```diff
@@ -0,0 +1,727 @@
...
++#include "pathtools.h"
+ 
+ typedef f0r_instance_t (*f0r_construct_f)(unsigned int width, unsigned int height);
+ typedef void (*f0r_destruct_f)(f0r_instance_t instance);
+@@ -181,6 +182,47 @@ static int load_path(AVFilterContext *ctx, void **handle_ptr, const char *prefix
+     return 0;
+ }
+ 
++#ifdef _WIN32
++static const char* const * get_feri0r_pathlist(AVFilterContext *ctx)
```
lazka, line 647: some s/feri0r/frei0r/ typos :)

lazka:
lgtm

mingw-w64-ffmpeg/0002-Win32-Add-path-relocation-to-frei0r-plugins-search.patch
mingwandroid: Will fix, greppability is important. thanks!

mingwandroid:
OK, fixed the transposos!

mingwandroid:
This PR serves as a useful example for adding relocation to a project too if we wanted something to point to in our docs?

mingwandroid:
Forgot to bump pkgrel. Give me 5. I will also attempt an update if any are available.

lazka:
> This PR serves as a useful example for adding relocation to a project too if we wanted something to point to in our docs?

Yeah, also having this code in a git repo would be nice. So we can add Unicode support maybe, and get rid of MAX_PATH.

mingwandroid:
The code lives in mingw-w64-pathtools

mingwandroid:
Unicode support (utf-8) happens by default from using standard C ASCII functions. The only thing we could add is translation of UCS16 or various Windows code-pages, but I do not think it is worth it. This path relocation *does* work in the face of Unicode paths though in general (if it didn't MSYS2 would be pretty broken!)

lazka:
GetModuleFileNameA returns things in the ANSI code page, which is Windows-1252 on my machine, and does not cover unicode.

Adsun701 (commented):
This looks good to me.

mingwandroid:
Ah yes, you are right @lazka. I wonder why we don't get more trouble? I guess most people install to the default location? Codepages on Windows are such a pain..

lazka:
> I guess most people install to the default location?

yeah, probably.

Issue #7256, opened 2020-11-09
lldb doesn't work with Python


orgads:
```
$ winpty lldb test.exe
(lldb) target create "test.exe"
Current executable set to 'F:\Projects\test\debug\test.exe' (i686).
(lldb) script print('hello')
Python path configuration:
  PYTHONHOME = 'D:/a/_temp/msys/msys64/mingw32/bin'
  PYTHONPATH = (not set)
  program name = 'python3'
  isolated = 0
  environment = 1
  user site = 1
  import site = 1
  sys._base_executable = 'C:/msys32/mingw32/bin/lldb.exe'
  sys.base_prefix = 'D:/a/_temp/msys/msys64/mingw32/bin'
  sys.base_exec_prefix = 'D:/a/_temp/msys/msys64/mingw32/bin'
  sys.executable = 'C:/msys32/mingw32/bin/lldb.exe'
  sys.prefix = 'D:/a/_temp/msys/msys64/mingw32/bin'
  sys.exec_prefix = 'D:/a/_temp/msys/msys64/mingw32/bin'
  sys.path = [
    'D:/a/_temp/msys/msys64/mingw32/bin/lib/python38.zip',
    'D:/a/_temp/msys/msys64/mingw32/bin/lib/python3.8',
    'D:/a/_temp/msys/msys64/mingw32/bin/lib/python3.8',
    'D:/a/_temp/msys/msys64/mingw32/bin/lib/python3.8/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00005150 (most recent call first):
<no Python frame>
```

Another attempt after manually setting `PYTHONPATH`:
```
$ export PYTHONPATH=C:/msys32/mingw32/lib/python3.8

$ winpty lldb test.exe
(lldb) target create "test.exe"
Current executable set to 'F:\Projects\test\debug\test.exe' (i686).
(lldb) script print('hello')
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:/msys32/mingw32/lib/python3.8/site-packages/lldb/__init__.py", line 95, in <module>
    import six
  File "C:/msys32/mingw32/lib/python3.8/site-packages/six.py", line 633, in <module>
    import struct
  File "C:/msys32/mingw32/lib/python3.8/struct.py", line 13, in <module>
    from _struct import *
[...]

PR #7416, opened 2020-12-08
Fix lldb's embedded PYTHONHOME

martindahlgren:
Fixes issue #7256 

lldb includes a hardcoded PYTHONHOME set to where python was during building. As can be seen in [lldb output here](https://github.com/msys2/MINGW-packages/issues/7256#issue-738833113) it was D:/a/_temp/msys/msys64/mingw32/bin.

This change makes the embedded path relative to liblldb.lldb - i.e. not absolute to the D:\ drive from build machine. Another option could be to use the build option to disable embedding the  #PYTHONHOME which is standard on non-windows targets.

As a side note: I think lldb's build scripts doesn't set this correctly. In the sys.path in the mentioned lldb output the /msys64/mingw32/bin/lib/python3.8 should rather be /msys64/mingw32/lib/python3.8. Maybe this matches the official python installer where python lib is in folder in same directory as python.exe.

I have not made commits in here before. Hope this one is ok :)

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -33,7 +33,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-openmp"
          "${MINGW_PACKAGE_PREFIX}-polly")
 pkgver=11.0.0
-pkgrel=4
+pkgrel=5
 pkgdesc="C language family frontend for LLVM (mingw-w64)"
 arch=('any')
 url="https://llvm.org/"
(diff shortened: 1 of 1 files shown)


lazka:
What about LLDB_RELOCATABLE_PYTHON  mentioned here https://lldb.llvm.org/resources/build.html#id1 ?

lazka:
you need to increase pkgrel

martindahlgren:
> What about LLDB_RELOCATABLE_PYTHON mentioned here https://lldb.llvm.org/resources/build.html#id1 ?

The LLDB_RELOCATABLE_PYTHON have been renamed. https://reviews.llvm.org/D74727 . As mentioned that would be another option. Not sure what is most appropriate.

lazka:
I see, disabling the embedding seems like the right thing, but maybe I'm missing something. But your change also makes sense, so feel free decide.

martindahlgren:
The LLDB_EMBED_PYTHON_HOME=OFF seems to work ok also. I think the end result is the same, that python looks for libraries relative to the main executable (lldb.exe). Strangely this seems to work even when lldb is not calling [Py_SetProgramName](https://docs.python.org/3/c-api/init.html#c.Py_SetProgramName) ...

PR #7419, opened 2020-12-09
Update GnuCOBOL to 3.1.1

GitMensch:
additionally switched to GNU ftp mirror as the integrity of the package can be checked via signature

Note: using json-c would be possible `--with-json=json-c` if this would be preferable because of the number of uses. The test results are identical using that option...

### DIFF
--- a/mingw-w64-gnucobol/PKGBUILD
+++ b/mingw-w64-gnucobol/PKGBUILD
@@ -6,8 +6,8 @@ _realname=gnucobol
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
 pkgver=3.1
-pkgver_real=3.1
-pkgrel=1
+pkgver_real=3.1.1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


GitMensch:
Can we re-trigger the build (seems like the CI got to a ftp-mirror which had not those files yet) or do we need a new commit for that?

Biswa96:
What is the problem with previous link https://ftp.gnu.org/gnu/ ? It has also the sig file.

GitMensch:
There's no problem with the main site which is the source for all mirrors.
There's just the general "preference" to use a nearby mirror instead of pushing all traffic to the the root.

Biswa96:
The package is not built thousands time in an hour, just twice. There would be no issue using main site.

GitMensch:
I'm reverting that part in a new PR. Thank you for pointing that out.

Biswa96:
You could just force push the commit/changes here.

GitMensch:
#7420 already has the change (I'm not on a development environment currently and therefore can only PR via browser...)

PR #7421, opened 2020-12-09
Add libbacktrace

lazka:


orgads:
Thank you!

mati865:
FYI libbacktrace is not usable with GCC because of the bug that has been just recent fixed in GCC trunk. More details: https://github.com/ianlancetaylor/libbacktrace/issues/43 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96948

lazka:
Isn't this https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gcc/0160-libbacktrace-seh.patch ?

mati865:
Right, I had missed that patch.

PR #7426, opened 2020-12-10
Bump aom and rav1e

kmilos:


kmilos:
Thanks, I have no clue about cargo and rust...

Re PGP signatures, we're removing validation in general?

lazka:
I guess we can try switching back

PR #7428, opened 2020-12-11
dlfcn: Build with CMake also for CMake files (prefer configure/make for the …

mingwandroid:
…rest)


lazka:
Can't we use cmake for everything?

mingwandroid:
Probably but I considered this lower risk. Happy to try!

mingwandroid:
Let me know (or just push something), I'll be back in about 2 hours.

lazka:
This drops the static library, is this a problem?

mingwandroid:
I think we should for this lib provide static.  The problem is it is very low level and we want to avoid it when possible. If we dynamically link it then it will get dragged in to builds and found by things that could do just a well without it but they will find it, esp. now we add the camke build files (confimed here as working in mlt framework, my next package update i hope before i tackle shotcut).

mingwandroid:
Then projects that do need it can decide whether to risk these potential pitfalls with shared or go with a static link to this instead. In general i think libs should, in this one instance prefer to link to dlfcn statically while applications should be more free to pick.. I would like @Alexpux thoughts here too!

Alexpux:
From my POV we must avoid linking with dlfcn or it will be everywhere

mingwandroid:
Ok how about we deliberately put the shared lib into a separate package, then we add dlfcn (when necessary) as a build dep. That should satisfy these goals? If everyone agrees I will make the changes.

lazka:
Why not just build the static lib with cmake as well?

mingwandroid:
I can do I guess, but are you in agreement about the shared lib being split off into its own package? That seems to be the safest way to ensure its not used inadvertentaly.

lazka:
If you really want to avoid it being used then I'd just drop it. Having builds behave differently depending on the list of packages installed should be avoided imo. It makes things hard to debug.

mingwandroid:
We would have to check everything that users the dll and rebuild them. If that ok? I personnally do prefer making it available but adding a check that it's never auto linked to a dll finally (in makepkg I guess).

lazka:
You can bump the pkgrel of defendant packages in this PR to test it

lazka:
lgtm

two issue for potentially future changes:

* ffmpeg links against libdl but is missing a dependency
* dlfcn-win32 is MIT and not LGPL2

jeremyd2019:
> We need to take care that they static link to libdl.a and not to the .dll, but since there is no .dll we should be OK with this

This does not seem to be true as of the commit this message is on (b622e69), looking at the [CI output](https://github.com/msys2/MINGW-packages/pull/7428/checks?sha=b622e694d785e1cf4f66b16a3bf9c12329366c5b#step:5:152) it is building and installing libdl.dll.

mingwandroid:
Update incoming. Thanks!

mingwandroid:
This update still builds the shared lib to ensure that works then deletes the dll and import lib. Hope its OK now?

mingwandroid:
@lazka, I think this is ready?

PR #7433, opened 2020-12-12
wt: New package

podsvirov:
For example, to launche the `Wt Widget Gallery` example like at https://www.webtoolkit.eu/widgets type:
```bash
${MINGW_PREFIX}/lib/Wt/examples/widgetgallery/widgetgallery.wt --docroot ${MINGW_PREFIX}/lib/Wt/examples/widgetgallery/docroot --approot ${MINGW_PREFIX}/lib/Wt/examples/widgetgallery/approot --resources-dir ${MINGW_PREFIX}/share/Wt/resources  --http-address localhost --http-port 8080
```
or just type:
```bash
${MINGW_PREFIX}/lib/Wt/examples/widgetgallery/widgetgallery
```
then open your browser at http://localhost:8080.


lazka:
Can we get rid of to 50 files in /bin without any namespacing? Arch doesn't have them for example.

podsvirov:
> Can we get rid of to 50 files in /bin without any namespacing? Arch doesn't have them for example.

I think we can do it, but I need to create another small patch.

lazka:
It's not that big of a problem, since it will only affect users installing this package, just something I've noticed.

I can merge as is if you want as well.

podsvirov:
I think your point is correct. Moreover, the patch I am talking about will replace `bat` scripts from the `bin` directory with `sh` compatible ones in the examples directory.

Give me a little time to create a patch, and then, after checking, you can merge.

podsvirov:
@lazka looks like we ready to merge. Please, review again.

PR #7434, opened 2020-12-12
filezilla: update

lazka:


lazka:
Forwarded the patch upstream: https://trac.filezilla-project.org/ticket/12343

PR #7435, opened 2020-12-12
Add gn generator (ninja)

ImperatorS79:
gn is used by google in a lot of their open-source project


1480c1:
https://github.com/msys2/MINGW-packages/pull/6792

ImperatorS79:
Okay, interesting info. I still would have prefer some explanations before having to spend 20 minutes reading an issue.

1480c1:
I doubt my skills would be able to pack the information there without resorting to rough english

PR #7443, opened 2020-12-14
spdlog: Update to 1.8.2

podsvirov:


podsvirov:
The changes have been merged a long time ago, but why isn't the package updated in the package repository?

lazka:
spdlog is waiting on cmake, which is waiting on qt5, which needs to be built manually because it takes longer than 6 hours to build.

umarcor:
@lazka, is it a matter of some maintainer allocating 6h of their own workstation for doing it? Or are there any other issues which are preventing it?

lazka:
the former (6 hours is per arch, so double that, but with more cores it's faster)

umarcor:
Is there any chance for me to contribute those packages? I can allocate those 12h. Yet, I don't know how could I ensure that I did not modify the recipe.

PR #7444, opened 2020-12-15
kicad: update to 5.1.9

umarcor:
This is not an update, but a problem report. Trying to run kicad fails:

When rebuilding by bumping `pkgrel`, it complains about ngspice:

```
CMake Error at D:/a/_temp/msys/msys64/mingw64/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find ngspice (missing: NGSPICE_DLL)
```

It seems that ngspice places the DLL in `/mingw64/bin/libngspice-0.dll` (https://packages.msys2.org/package/mingw-w64-x86_64-ngspice?repo=mingw64). Maybe that needs to be fixed?

/cc @orsonmmz


umarcor:
Even though ngspice's DLL seems to be mislocated, that's not the issue here. By switching from `MSYS Makefiles` to `MinGW Makefiles` and `mingw32-make`, the kicad recipe can find ngspice's DLL.

Now, it fails later, due to a mismatch between wxPython and wxWidget's versions:

```
D:/a/_temp/msys/msys64/mingw64/lib/python2.7/site-packages/wx-3.0-msw/wx/_core.py:16633: UserWarning: wxPython/wxWidgets release number mismatch
  warnings.warn("wxPython/wxWidgets release number mismatch")
-- Found wxPython 3.0.2.0/msw (wxWidgets 3.0.2.0)
CMake Error at D:/a/_temp/msys/msys64/mingw64/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find wxWidgets (missing: wxWidgets_LIBRARIES
See also "C:/_/mingw-w64-kicad/src/build-x86_64-w64-mingw32/CMakeFiles/CMakeOutput.log".
  wxWidgets_INCLUDE_DIRS) (Required is at least version "3.0.2.0")
```

orsonmmz:
Hi umarcor,

> It seems that ngspice places the DLL in `/mingw64/bin/libngspice-0.dll` (https://packages.msys2.org/package/mingw-w64-x86_64-ngspice?repo=mingw64). Maybe that needs to be fixed?

I have just checked and it seems ok: in msys2 all *.dll files are in /mingw64/bin, while *.dll.a are located in /mingw64/lib, so ngspice is no exception.

The mismatch between wxWidgets and wxPython is just a warning, the real problem is that wxWidgets is not found at all:
> Could NOT find wxWidgets (missing: wxWidgets_LIBRARIES

umarcor:
> I have just checked and it seems ok: in msys2 all *.dll files are in /mingw64/bin, while *.dll.a are located in /mingw64/lib, so ngspice is no exception.

Correct. I found dlls in `/mingw64/lib` too, and that confused me. Upon further inspection, kicad searches both: https://gitlab.com/kicad/code/kicad/-/blob/master/CMakeModules/Findngspice.cmake#L113-114.

I think that the problem was it was not searching in `/mingw64/*` because `MSYS Makefiles` were used.

> The mismatch between wxWidgets and wxPython is just a warning, the real problem is that wxWidgets is not found at all:
> 
> > Could NOT find wxWidgets (missing: wxWidgets_LIBRARIES

You are correct, again. I found that some fixes were pushed recently: https://gitlab.com/kicad/code/kicad/-/commits/master/CMakeModules/FindwxWidgets.cmake. I squashed them and applied the patch to this PR. BTW, I updated KiCad to 5.1.8 (was 5.1.5). The build starts now (so, it finds all the dependencies).

mingw-w64-kicad/PKGBUILD
```diff
@@ -60,8 +65,8 @@ build() {
   mkdir -p build-${MINGW_CHOST} && cd build-${MINGW_CHOST}
```
Biswa96, line 66: Try to add the following here to solve the build issue in X86_64:

```sh
  case "${CARCH}" in
    x86_64)
      LDFLAGS+=" -Wl,--disable-dynamicbase,--default-image-base-low"
    ;;
  esac
```

umarcor: I applied the LDFLAGS fix, and that seems to work. I'm not sure about how to use the ninja generator, though. Is that an argument to cmake? There are three calls to cmake. Should it be used in all of them?

Biswa96: No problem. It can be done later.

umarcor:
I suggest merging this PR, as it allows having a usable version in the repositories. We can fix the relocation/ASLR issues in an upcoming PR.

marekr:
Fyi, Kicad is in process of updating to 5.1.9. It's already tagged and packages built for windows/macos/some linux distros. 

You will want to skip ahead

umarcor:
@marekr, done. BTW, could you please have a look at the `0001-fix-FindwxWidgets.cmake.patch` in this PR? I created that by squashing three commits of yours to `master` (https://gitlab.com/kicad/code/kicad/-/commits/master/CMakeModules/FindwxWidgets.cmake). However, those commits were not backported to branch 5.1 (https://gitlab.com/kicad/code/kicad/-/commits/5.1/CMakeModules/FindwxWidgets.cmake). Shall we keep the patch until 6.0?

marekr:
@umarcor Those commits aren't required for MSYS2.

As part of the push to migrate fully to MSVC builds, those were a hack so one can use the VS IDE with msys2 for now.

umarcor:
@mrekr, without those commits, the build would not start because of `Could NOT find wxWidgets`. See https://github.com/msys2/MINGW-packages/pull/7444#issuecomment-746397740.

marekr:
> @mrekr, without those commits, the build would not start because of `Could NOT find wxWidgets`. See [#7444 (comment)](https://github.com/msys2/MINGW-packages/pull/7444#issuecomment-746397740).

Weird because I byuild 5.1 within msys2 along with others just fine. However we don't use the PKG build system which must not be declaring MSYS in cmake?

umarcor:
These are MINGW packages, not MSYS packages. Therefore, that might be the issue. That is, you might be executing an external process and using some MSYS binaries (some unix tools missing on Windows). However, MSYS tools are not exactly equivalent to MINGW. In this case, we want not to rely on MSYS.

marekr:
> These are MINGW packages, not MSYS packages. Therefore, that might be the issue. That is, you might be executing an external process and using some MSYS binaries (some unix tools missing on Windows). However, MSYS tools are not exactly equivalent to MINGW. In this case, we want not to rely on MSYS.

well, you will need to apply the patch for 5.1.9 then anyway since if I do backport and test (which is a painful experience under msys :( ) it won't be until 5.1.10 is released

lazka:
Can this be merged?

umarcor:
Yes, please.

PR #7445, opened 2020-12-15
qemu: Update to 5.2.0

mingwandroid:


lazka:
ok, I promptly forgot that this was missing checks...

can you open a new PR that bumps pkgrel so we can try to make it build?

PR #7447, opened 2020-12-15
Add mingw-w64-riscv64-unknown-elf-gcc 

Fabien-Chouteau:
Based on arm-none-eabi-gcc.

Also fixes libgloss_nano.a packaging in newlib.


Fabien-Chouteau:
Thank you @lazka

PR #7449, opened 2020-12-15
qemu: bump pkgrel to 2

mingwandroid:


ImperatorS79:
You should install ninja as makedepens.

mingwandroid:
Done and it seems github is happy with my email (at the moment!) so is running CI which is nice!

This package is pretty massive. I wonder if we cannot investigate using hard-links instead of the symlinks that fail and fall back to copies?

mingwandroid:
The build succeeds locally. Not sure what's wrong in ci. There are some suspicious warnings from ninja though but they appear harmless as I also got those.

ImperatorS79:
This I guess:
```
Found ninja-1.10.2 at D:/a/_temp/msys/msys64/mingw64/bin/ninja.exe
/mingw64/bin/ninja  build.ninja && touch build.ninja.stamp
ninja: no work to do.
/mingw64/bin/python -B /C/_/mingw-w64-qemu/src/qemu-5.2.0/meson/meson.py introspect --targets --tests --benchmarks | /mingw64/bin/python -B scripts/mtest2make.py > Makefile.mtest
make[1]: *** No rule to make target 'multiboot.bin', needed by 'all'.  Stop.
make: *** [Makefile:206: pc-bios/optionrom/all] Error 2
```

mingwandroid:
Yeah. Seems harmless locally though. There my build succeeded. There's a lot of symlink stuff in the logs too that didn't look too good. Maybe we should build from a different source or try setting MSYS var?

mingwandroid:
Hi @umarcor, I see you have some knowledge about QEMU. Any ideas why the build might be failing on CI but not locally?

Thanks for all your hacking on MSYS2 also!

mingwandroid:
.. although my builds happen locally, the symlink warnings when unpacking seem troublesome and I get the same locally. The build logs are the same essentially, no failures emitted from either, but mine exits 0 when CI exits something else.

umarcor:
@mingwandroid, my guess is based on previous experience contributing PRs to this repo in the context of [hdl/MINGW-packages](https://github.com/hdl/MINGW-packages). Despite maintaining [dbhi/qus](https://github.com/dbhi/qus/), I hardly ever build QEMU. dbhi/qus is about extracting and distributing static binaries from Debian/Fedora :smile:

WRT this PR, I would add `"${MINGW_PACKAGE_PREFIX}-make"` to makedepends and I would use `mingw32-make` instead of `make`. I had to do that in some other packages: https://github.com/msys2/MINGW-packages/pull/7404/commits/8eb0480a57eb2da67cd3bb2b94fdab845e3d185c.

lazka:
qemu has recently been updated in #8652 #8580 #8677 and #8696, so closing this

PR #7457, opened 2020-12-16
Ffmpeg: Move dlfcn from makedepends to depends

### LINKED ISSUE #7456: ffmpeg is broken
With the current version of MSYS2, when I type `ffmpeg`, I am getting:
```
$ ffmpeg
D:/msys64/mingw64/bin/ffmpeg.exe: error while loading shared libraries: swscale-5.dll: cannot open shared object file: No such file or directory
```
However:
```
$ ls -l /mingw64/bin/swscale-5.dll
-rwxr-xr-x 1 Olivier None 551936 Dec  9 01:27 /mingw64/bin/swscale-5.dll
```
Seems to be there...

ImperatorS79:
Fixes #7456.

### DIFF
--- a/mingw-w64-ffmpeg/PKGBUILD
+++ b/mingw-w64-ffmpeg/PKGBUILD
@@ -15,6 +15,7 @@ options=('staticlibs' 'strip')
 depends=("${MINGW_PACKAGE_PREFIX}-aom"
          "${MINGW_PACKAGE_PREFIX}-bzip2"
          "${MINGW_PACKAGE_PREFIX}-celt"
+         "${MINGW_PACKAGE_PREFIX}-dlfcn"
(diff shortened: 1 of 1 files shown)


mingwandroid:
Ping @lazka, I thought we didn't carry the DLL? In general we would like to not I think and have this as a makedepends only. I think that is @Alexpux preference too?

mingwandroid:
The fuzz is that most build systems prefer shared libraries by default so shared libraries in the install due to transitive dependencies will install libdl.dll, and those top-level builds will find and use it.

Personally I don't mind but am happy to try to keep it away from being linked dynamically (as it had been before I broke one release of dlfcn by including the DLL which we can see, was picked up and used by ffmpeg, unfortunately).

mingwandroid:
We want to merge #7428 in preference to this.

PR #7461, opened 2020-12-16
OpenAI

Schala:
New:

- python-dill: dep for upcoming package
- python-lws: lip2wav dep

Updated:
- python-scikit-image 0.18


Biswa96:
@Schala Do you check if the OpenAI packages contain any reference to msys2 installation path? For example, `skivi-script.py` has that `#!D:/a/_temp/msys/msys64/mingw64/bin/python.exe` shebang.

BTW, as there are so many OpenAI package, can we now distinguish cat vs. dogs images? 🙂

Schala:
> @Schala Do you check if the OpenAI packages contain any reference to msys2 installation path? For example, `skivi-script.py` has that `#!D:/a/_temp/msys/msys64/mingw64/bin/python.exe` shebang.
> 
> BTW, as there are so many OpenAI package, can we now distinguish cat vs. dogs images? 🙂

Hmm.... I'll look into that in my next PR

PR #7478, opened 2020-12-20
gdb: Fix array subscript when upper bound is unknown

orgads:
Backport fix from gdb-10-branch.


orgads:
Use-case:

foo.h:
```
struct Foo { int bar = 42; };
```

foo.cpp:
```
#include "foo.h"
Foo foo[10];
```

main.cpp:
```
#include "foo.h"
extern Foo foo[];
int main() {
    foo[1] = 44; // BREAK HERE
}
```

On gdb:
```
p foo[0]
```

PR #7490, opened 2020-12-22
drwmingw: fix build with gcc 10

jeroen:
See: https://github.com/msys2/MINGW-packages/commit/200852c1fe3783876ecafddd4fd736a1d04c7c16#commitcomment-45324831


Biswa96:
1. Increase `pkgrel` value by 1.
2. Add double quotes to the path in patch command.

jeroen:
I don't think it is needed to re-release this? Because it just fixes the build for newer compilers, but not the program has nog changed?

PR #7504, opened 2020-12-23
Update GnuCOBOL to 3.1.2

GitMensch:


Biswa96:
If `pkgver` is changed `pkgrel` has to be reset to 1.

GitMensch:
`pkgver` is unchanged, which is the reason that I've increased `pkgrel` [that's is a patched 3.1 and if I'd directly apply the patches then `pkgver` wouldn't be adjusted, I guess] :-)

But if you prefer I can also adjust both.

Biswa96:
I am not sure why there is another `pkgver_real` variable and it's different than `pkgver`. Maintainers may provide some guide. @lazka is busy 🏃

GitMensch:
pkgver_real has the name of the version string in the tarball and in the unpacked folder. It is not user-visible.
I suggest to pull that in and may adjust later (a newer version will always work, only going back won't) because 3.1.1 was handled 100% identical.

lazka:
I don't see the point of having an extra pkgver_real, if there is a new release with a new version we should use that. also pkgrel=1 as @Biswa96 said.

GitMensch:
Done, so `pkgver_real`'s really gone forever.

lazka:
Thanks!

PR #7506, opened 2020-12-24
libtorrent-rasterbar: Use Ninja, Remove path reference.

Biswa96:


Biswa96:
Wait! There is a version 2, forgot to change that.

PR #7507, opened 2020-12-25
groonga: Update to 10.1.0

komainu8:


Biswa96:
Just asking, any future plan to use CMake?

komainu8:
Yes, we have future plan to use CMake.
However, we need to implement some changes to use CMake. 
That tasks a lot of time.
So, we can't migration to CMake from Make soon.

komainu8:
Thanks!

PR #7509, opened 2020-12-25
prjtrellis: add 1.0.r975.2241b2a

umarcor:
This PR add package `prjtrellis`: [YosysHQ/prjtrellis](https://github.com/YosysHQ/prjtrellis). It's added to group 'eda'.


Biswa96:
Just a comment. As you are _only changing one file_ you could force push the commits. That way the pull request will be a clean one commit.

umarcor:
@Biswa96 you mean squash all of them in a single commit? That breaks GitHub's review system. If different commits are pushed, maintainers can see the difference since their last view/review. Nonetheless, when a PR is accepted, maintainers can "sqaush and merge" so that the history is clean (single commit).

umarcor:
`nextpnr` was merged some minutes ago (see #7542). nextpnr can place and route design for several architectures: generic, ice40, ecp5, nexus, gowin... Building for each architecture requires a different dependency. In the upstreamed package, ice40 was enabled only, which requires `icestorm` (uploaded already). 

Enabling ecp5 on nextpnr, depends on prjtrellis. Therefore, I converted this PR to a draft and, instead of adding prjtrellis only, I am also enabling the architecture on nextpnr.

Currently, ice40 and generic architectures can be enabled successfully on either MINGW64 or MINGW32. At the same time, prjtrellis is successfully built on either MINGW64 or MINGW32. However, nextpnr-ecp5 fails on MINGW32. See CI runs results below.

Biswa96:
Increase the `pkgrel` in nextpnr PKGBUILD because you have changed the build command which in result changes the binary.

umarcor:
> https://github.com/YosysHQ/nextpnr/issues/546#issuecomment-752733717
> The database build needs more memory than a 32bit address space can provide. Use the bba files from the 64bit build.

@Biswa96 any guess about how to do that without creating a separated package for the database (bba) files?

umarcor:
It seems not possible to add prjtrellis and enable ecp5 in nextpnr at the same time (in the same PR). Since the chipdb needs to be built on 64 bits, the MINGW32 build of nextpnr-ecp5 depends on the MINGW64 version of prjtrellis.

Therefore, I marked this PR as ready for review, and I'll handle nextpnr-ecp5 in an upcoming PR.

PR #7518, opened 2020-12-26
lis: update to 2.0.28

3rav:


lazka:
Somehow it fails to find the right fortran compiler in autobuild compared to the CI here.. I'll have a look.

lazka:
ok, fixed in 254665faf77

PR #7529, opened 2020-12-27
asio: Update to 1.18.1

podsvirov:


Biswa96:
Just asking. Why boost is disabled?

podsvirov:
> Just asking. Why boost is disabled?

Because with `boost` support, there is already in the `boost` package.

Issue #7531, opened 2020-12-27
aws-sdk build error


revelator:
fails on the git sync step for some reason i havent been able to determine.

Biswa96:
There is not git command in the PKGBUILD file of that package.

revelator:
no it seems to come from the source package itself, ill try and post the error when i do a build of it again.

revelator:
```==> Starting build()...
-- Found Git: C:/Msys64/usr/bin/git.exe (found version "2.29.2") 
-- TARGET_ARCH not specified; inferring host OS to be platform compilation target
-- Building AWS libraries as static objects
-- Generating windows build config
fatal: not a git repository: 'E:/mingw-w64-aws-sdk-cpp/src/aws-sdk-cpp-1.7.365/.git'
-- Building project version: 1.7.365
-- The C compiler identification is GNU 10.2.0
-- The CXX compiler identification is GNU 10.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Msys64/mingw32/bin/cc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Msys64/mingw32/bin/c++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: E:/mingw-w64-aws-sdk-cpp/src/build-i686-static/.deps
[1/24] Creating directories for 'AwsChecksums'
[2/24] Creating directories for 'AwsCCommon'
[3/24] Performing download step (git clone) for 'AwsChecksums'
Cloning into 'AwsChecksums'...
Note: switching to 'v0.1.5'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 519d6d9 Don't return 0 on arm please (#22)
[4/24] No update step for 'AwsChecksums'
[5/24] Performing patch step for 'AwsChecksums'
[...]

PR #7540, opened 2020-12-28
aws-sdk-cpp: Use tarball for AWSLabs dependencies instead of git.

Biswa96:
Changes:
  * Remove Patch-cmake-submodules.patch file.
  * Add 0001-cmake-change-deps-dir.patch and 0002-third-party-cmake-remove-git.patch.

Fixes https://github.com/msys2/MINGW-packages/issues/7531


mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -13,25 +16,58 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
...
   patch -p1 -i "${srcdir}/aws-sdk-cpp-pr-1333.patch"
-  patch -p1 -i "${srcdir}/Patch-cmake-submodules.patch"
+
+  # Set dependencies search to current source directory
+  patch -p1 -i "${srcdir}/0001-cmake-change-deps-dir.patch"
+
+  # Replace git command to use our extracted tarball
+  patch -p1 -i "${srcdir}/0002-third-party-cmake-remove-git.patch"
+
+  #patch -p1 -i "${srcdir}/Patch-cmake-submodules.patch"
```
kou, line 51: Can we remove this?

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -13,25 +16,58 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
...
 options=('strip' 'staticlibs')
 source=("${_realname}-${pkgver}.tar.gz::https://github.com/aws/aws-sdk-cpp/archive/${pkgver}.tar.gz"
+        "AwsCCommon.tar.gz"::"https://github.com/awslabs/aws-c-common/archive/v${_aws_common_ver}.tar.gz"
+        "AwsChecksums.tar.gz"::"https://github.com/awslabs/aws-checksums/archive/v${_aws_checksums_ver}.tar.gz"
+        "AwsCEventStream.tar.gz"::"https://github.com/awslabs/aws-c-event-stream/archive/v${_aws_stream_ver}.tar.gz"
         "aws-sdk-cpp-pr-1333.patch"
-        "Patch-cmake-submodules.patch"
+        "0001-cmake-change-deps-dir.patch"
+        "0002-third-party-cmake-remove-git.patch"
+        #"Patch-cmake-submodules.patch"
```
kou, line 25: Can we remove this?

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -13,25 +16,58 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
...
+  patch -p1 -i "${srcdir}/0001-cmake-change-deps-dir.patch"
+
+  # Replace git command to use our extracted tarball
+  patch -p1 -i "${srcdir}/0002-third-party-cmake-remove-git.patch"
+
+  #patch -p1 -i "${srcdir}/Patch-cmake-submodules.patch"
+
+  # Extract and patch AWS related dependecies
+
+  tar -xf "${srcdir}/AwsCCommon.tar.gz" || true
```
kou, line 55: Why do we need `|| true`?

mingw-w64-aws-sdk-cpp/PKGBUILD
```diff
@@ -4,7 +4,10 @@ _realname=aws-sdk-cpp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.7.365
-pkgrel=2
+_aws_common_ver=0.4.42
+_aws_checksums_ver=0.1.5
+_aws_stream_ver=0.1.5
```
kou, line 9: Should we add a comment that we need to sync these versions with the versions in https://github.com/aws/aws-sdk-cpp/blob/master/third-party/CMakeLists.txt ?

mingw-w64-aws-sdk-cpp/PKGBUILD
revelator: might be because of symlinks inside the archive msys2 tar only supports windows symlinks i think.

mingw-w64-aws-sdk-cpp/PKGBUILD
revelator: that might be a good idea :)

mingw-w64-aws-sdk-cpp/PKGBUILD
revelator: if no longer used sure :)

Biswa96: As this is not a version change, I keep the removed patch in a comment. Also it can be reviewed in next version change.

mingw-w64-aws-sdk-cpp/PKGBUILD
Biswa96: Same as before.

mingw-w64-aws-sdk-cpp/PKGBUILD
Biswa96: Yes, Ralph is right.

Biswa96:
If this is alright would anyone like to test if arrow package can be rebuild?

revelator:
sure let me at it :)

mingw-w64-aws-sdk-cpp/PKGBUILD
kou: It seems that there are no symbolic links in these archives.
If there are symbolic links in an archive, we can use GNU tar like in https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gr/PKGBUILD#L24 .

revelator: odd is the same tar command as previous ?!? tar -xf and the gr one is the same.

Biswa96: The used version of aws-c-common does not have the symlinks but the current version one has. So, I keep the `|| true` for future proof.

Biswa96:
@lazka ready to merge.

mingw-w64-aws-sdk-cpp/PKGBUILD
revelator: Np with that here 👍

PR #7542, opened 2020-12-29
nextpnr: add 0.0.r2859.5e53a182

umarcor:
This PR adds package `nextpnr` (a portable FPGA place and route tool): [YosysHQ/nextpnr](https://github.com/YosysHQ/nextpnr). It's added to group 'eda'.

Together with Yosys (#7534), it allows synthesising and implementing HDL designs targetting Lattice's ICE40 FPGAs. The GUI (based on Qt5) is enabled.

~~See YosysHQ/nextpnr#546 for a discussion about why `mingw32-make DESTDIR="${pkgdir}" install` is commented out and the assets are copied manually instead.~~

### DIFF
--- a/mingw-w64-nextpnr/PKGBUILD
+++ b/mingw-w64-nextpnr/PKGBUILD
@@ -1,0 +1,77 @@
+# Maintainer: umarcor <unai.martinezcorral@ehu.eus>
+
+_realname=nextpnr
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.0.r2859.5e53a182
+pkgrel=1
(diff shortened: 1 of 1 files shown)


mingwandroid:
Why don't you use MSYS Makefiles and MSYS2/make here?

umarcor:
@mingwandroid with some other package I had issues when using make, so I switched to mingw32-make (https://github.com/msys2/MINGW-packages/commit/8eb0480a57eb2da67cd3bb2b94fdab845e3d185c#diff-f9905f397c40e99f29ca020fc2742694360a9e8ea1daddc0dbeaf79dc3539803). Since then, I keep using it. I couldn't find any explanation about why one or the other is a better solution.

In this specific case, I did several tests and IIRC I got the same result.

umarcor:
I updated the recipe. MSYS Makefiles are used now.

PR #7553, opened 2020-12-30
starpu: add fftw (need patch to fix build)

3rav:


Biswa96:
Some queries:

* Is the patch in upstream? or is it mingw specific?
* Would you like to replace md5 with sha256/sha512?
* In `prepare()` step, it is recommended to apply the patch first then do the `autoreconf`.

3rav:
1. I think is for mingw, but it will be in next release: [https://prod-gitlab.inria.fr/starpu/starpu/-/commit/11ade2e903c49ebddc4d4b85bc63c3ead36d844a](url)
2. I can change, md5 is directly available at www: [https://files.inria.fr/starpu/](url)
3. OK

Biswa96:
As it is a upstream commit, a reference link of that commit would be useful in PKGBUILD for future updates.

PR #7568, opened 2021-01-01
nextpnr: add ARCH 'ecp5' (prjtrellis)

umarcor:
As commented in https://github.com/msys2/MINGW-packages/pull/7509#issuecomment-752734673, one of the components of nextpnr-ecp5 needs more that 4GB to be build. Therefore, it cannot be done in a 32 bit memory space.

The proposed workaround is to use a script and call it in a MINGW64 shell (even on MINGW32 builds) (see https://github.com/lazka/MSYS2-packages/blob/fa0ba92d8608a0393574eb7febd26025a2a7e45b/pacman/makepkg-mingw#L111-L113). However, doing so requires `x86_64-cmake` and `x86_64-prjtrellis` to be available on the MINGW32 build. That's because MINGW32 and MINGW64 jobs are independent in CI, so the MINGW32 build cannot reuse an artifact from the MINGW64 job.

### DIFF
--- a/mingw-w64-nextpnr/PKGBUILD
+++ b/mingw-w64-nextpnr/PKGBUILD
@@ -28,9 +28,19 @@ makedepends=(
   "git"
 )
 
+if [ "$CARCH" = "i686" ]; then
+  makedepends+=(
+    "${MINGW_PACKAGE_PREFIX}-prjtrellis"
+    "mingw-w64-x86_64-prjtrellis"
+    "mingw-w64-x86_64-cmake"
+  )
+fi
+
 _commit="fd2d4a8f"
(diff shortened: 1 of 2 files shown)


Biswa96:
Why does a separate shell script to build?

Biswa96:
Why a separate shell script to build?

umarcor:
@Biswa96:

> one of the components of nextpnr-ecp5 needs more that 4GB to be build. Therefore, it cannot be done in a 32 bit memory space.

umarcor:
This PR is now based on #8231.

lazka:
Is there anything that can be done here? Or should I close it?

PR #7563, opened 2020-12-31
 Add python-lark-parser and python-tomlkit 

naveen521kk:
Useful for python projects using PEP517, eg poetry.

PR #7572, opened 2021-01-02
Update some packages

naveen521kk:


naveen521kk:
Looks like isort also moved to poetry. So, @lazka we should bring in `poetry-core` as soon as possible. #7563

PR #7591, opened 2021-01-05
Update

naveen521kk:
https://github.com/naveen521kk/msys2-updator

### DIFF
--- a/mingw-w64-python-asgiref/PKGBUILD
+++ b/mingw-w64-python-asgiref/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=asgiref
 pkgbase=mingw-w64-python-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
-pkgver=3.2.5
+pkgver=3.3.1
 pkgrel=1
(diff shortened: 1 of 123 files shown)


lazka:
ok :)

but who checks for dependency changes?

naveen521kk:
> but who checks for dependency changes? I guess me

That looks worse though, checking through some 200 packages isn't that simple. Maybe enable checks so we can run their test suite and check it? Or let's make it work with that `updator`? Got any idea's on how to? AFAIK, PyPI API provides some info about this, like [`this`](https://pypi.org/pypi/Sphinx/json) and it can be used for checking dependency. And for now, this is only PyPi things are here.

naveen521kk:
Also, this would be the case for the first time, if it runs daily, then things would be easy :)

mati865:
For Python you can use https://github.com/wimglenn/johnnydep, I had link to similar tool for library dependencies but cannot find it right now.

naveen521kk:
This PR is only for python package, thanks :)

naveen521kk:
Ok, I think it would better to do things smaller in number and also I would try implementing a dependency handler inside it, so things don't be like this, instead would be more organized.

PR #7596, opened 2021-01-05
Fix fxc2 not found in qt build

ImperatorS79:


lazka:
fxc2-git provides fxc2, so I don't see what this changes?

ImperatorS79:
Huum, maybe it was just that fx2 was not built when qt5 was updated to use it, because the last qt PR failed because of missing fxc2.

Since I did not see qt updated in the pacman repo I guessed there was a problem but since it indeed built fine I will close this -> https://packages.msys2.org/base/mingw-w64-qt5

PR #7602, opened 2021-01-06
python: cygpty hack improvements

lazka:
In case of cygpty we want to fake a tty and at the same time make
Python behave as if it was attached to a console.

This improves two thing in that regard:

* Update the is_cygpty patch to cover more isatty() checks (likely
  they were added since the patch was created)
* Default the stdio encoding to the console output encoding by default.
  This is only in effect if stdout isn't redirected (then is uses the ANSI code page)
  and no real console is used (then it forces utf-8). Python never considers the
  case were isatty() and no console, so we have to add this.

This makes printing unicode work again in MSYS2 bash.


Biswa96:
Does it have the same purpose of conpty in conhost?

lazka:
If you mean that with winpty/conpty this all is not used/doesn't change anything, then yes. If we enabled conpty by default we can drop all these hacks.

PR #7608, opened 2021-01-07
OpenAI updates

Schala:
Updated:
- ffmpeg-python: renamed from `python-ffmpeg-python` for less redundancy, reflects AUR package name
- python-pytest-mock 3.5
- python-trimesh 3.9.1


lazka:
The AUR one is called `python-ffmpeg` not `ffmpeg-python` from what I see. Is this intentional?

Schala:
I was confused too. `python-ffmpeg` and `ffmpeg-python` are in fact different:

Screenshot 2021-01-07 125728

lazka:
huh, ok then

PR #7613, opened 2021-01-07
update fonttools

naveen521kk:


naveen521kk:
squash this while merging

PR #7622, opened 2021-01-08
Various package updates

Astrum-polaris:
One interesting thing I notice is that if I try to build zstd using ninja, the generated pc file have some extra backslash in it.
Example: `exec_prefix=\${prefix}`


lazka:
>  Example: `exec_prefix=\${prefix}`

hm, thanks, I'll have a look later

lazka:
The zstd cmake build directly depends on shell escaping in that case, so that is expected, somewhat.d

lazka:
I've filed https://github.com/facebook/zstd/pull/2462

We can port to Ninja once we get some feedback there.

PR #7625, opened 2021-01-08
plplot: add wingcc and wingdi drivers

pjb7687:
It would be awesome if wingcc and wingdi drivers for plplot is supported.


Biswa96:
The msys2 dependencies are not used in mingw packages. Any reason for that?

pjb7687:
Hi, I think actually these packages are not related to msys2 but they are header/lib files for windows dlls. They are needed because wingcc/wingdi driver requires to be linked to `gdi32.dll`, which is a standard library.

btw it seems like cmake can't find the windows kits path. I will try it again..

pjb7687:
Tested, seems to work. I just patched their cmake files so that it no longer searches for the gdi32 header files, therefore the two msys2 packages are removed. Also their `wingdi.c` seems to have typos - also corrected them.

pjb7687:
Splitted the patches into 2 files. If one of the patches is no longer needed in future releases, simply it can be commented out.

mati865:
> 
> 
> Tested, seems to work. I just patched their cmake files so that it no longer searches for the gdi32 header files, therefore the two msys2 packages are not required and removed. Also their `wingdi.c` seems to have typos - also corrected them.

What do you mean? Is there something wrong with mingw-w64 gdi32 libs like this one `/mingw64/x86_64-w64-mingw32/lib/libcomdlg32.a`?

pjb7687:
> > Tested, seems to work. I just patched their cmake files so that it no longer searches for the gdi32 header files, therefore the two msys2 packages are not required and removed. Also their `wingdi.c` seems to have typos - also corrected them.
> 
> What do you mean? Is there something wrong with mingw-w64 gdi32 libs like this one `/mingw64/x86_64-w64-mingw32/lib/libcomdlg32.a`?

Hi, you are absolutely right, there is nothing wrong with mingw-w64 libs.

It is just because the plplot lib has some custom cmake files that tries to identify windows libs, including gdi32, which requires these msys2 packages. Plplot develpers actually suggests that in the following doc.

https://github.com/PLplot/PLplot/blob/3876b262c564a3951e4b4ea60a5cda62d6f25c0b/doc/wiki_source/Configure_PLplot_for_msys2#L23-L27

As I said you are right, in fact you really don't need to search for these msys2 headers or libs, since mingw64 libs can be just linked to the plplot. Therefore I just patched the relevent cmake files so that they don't try to find these msys2 headers anymore.

pjb7687:
...and yes, without this, cmake cannot find gdi32 libs.

https://github.com/msys2/MINGW-packages/runs/1671011370#step:5:451

mati865:
Their doc talks about building for MSYS2 subsystem, not for MinGW subsystem.

> It is just because the plplot lib has some custom cmake files that tries to identify windows libs, including gdi32, which requires these msys2 packages.

They are obviously wrong and should be fixed.

pjb7687:
> Their doc talks about building for MSYS2 subsystem, not for MinGW subsystem.
> 
> > It is just because the plplot lib has some custom cmake files that tries to identify windows libs, including gdi32, which requires these msys2 packages.
> 
> They are obviously wrong and should be fixed.

Already fixed.

lazka:
Please don't comment out old code when patching things. This makes it hard to see the actual changes.

pjb7687:
I added the correct windows lib path, as suggested by @mati865, rather than commenting out several lines. It reads better, works well.

Biswa96:
The mingw-w64-plplot/fix-wingdi.patch is sent to upstream in this mail https://sourceforge.net/p/plplot/mailman/message/37796314/

PR #7629, opened 2021-01-09
Add package: hiredis 1.0.0

taozuhong:


lazka:
What about using cmake?

taozuhong:
Maybe it's a great idea to use cmake. but I'm not familiar with it.

PR #7634, opened 2021-01-10
Add new package: hiredis 1.0.0

taozuhong:


lazka:
That looks much better, thanks!

PR #7640, opened 2021-01-10
libbotan: Disable thread_utils.

dewyatt:
See https://github.com/randombit/botan/issues/2582.

### DIFF
--- a/mingw-w64-libbotan/PKGBUILD
+++ b/mingw-w64-libbotan/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libbotan
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.17.3
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


mingw-w64-libbotan/PKGBUILD
```diff
@@ -59,6 +59,7 @@ build() {
   MSYS2_ARG_CONV_EXCL="--prefix=;--libdir=;--bindir=" \
```
lazka, line 59: Please add a small comment here linking the the upstream discussion, and/or why `-disable-modules=thread_utils` is passed. So we know why it's there in the future.

lazka:
increase pkgrel

dewyatt:
Roger that, updated (squashed it down -- hope that's ok).

dewyatt:
We can grab the key from https://botan.randombit.net/pgpkey.txt if desired btw

lazka:
yeah, but pacman doesn't support that. Maybe we can think about adding those keys to the repo though.. not sure.

PR #7642, opened 2021-01-10
tklib: update to fix install, download from tcl-lang.org

oscarfv:
The install script used on the previous package was defective and
ignored several files. Upstream fixed the problem.

The github mirror lags. Now we download the sources directly from
tcl-lang.org, which contains the central repository.

### DIFF
--- a/mingw-w64-tklib/PKGBUILD
+++ b/mingw-w64-tklib/PKGBUILD
@@ -3,7 +3,12 @@
 _realname=tklib
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.7.r1974.57e720fd
+_hash="29e6d0e6da"
+# Increase every time we refresh the package with a new upstream hash:
+_releasecounter=1
+_basever=0.7.0
+pkgver="$_basever+$_releasecounter.$_hash"
+_dir="Tk_Library_Source_Code-$_hash"
 pkgrel=1
 pkgdesc="A companion to Tcllib, for Tk related packages (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mingw-w64-tklib/PKGBUILD
```diff
@@ -3,27 +3,28 @@
...
          "${MINGW_PACKAGE_PREFIX}-tcllib")
-source=("git+https://github.com/tcltk/tklib.git#commit=${_commit}")
-sha256sums=('SKIP')
+source=("https://core.tcl-lang.org/tklib/tarball/29e6d0e6da/$_dir.tar.gz")
+sha256sums=('d9fae8f16b20cf9ccb4179977468cf6dff167ee2bfe230dee16f923f4bf09a6b')
 
 pkgver() {
-  cd "${_realname}"
-  printf '0.7.r%s.%s' "$(git rev-list --count "${_commit}")" "$(git rev-parse --short "${_commit}")"
+  printf '%s.%s.%s' $_basever "$(date -u +%y%m%d%H%M)" $_hash
```
lazka, line 23: pkgver needs to return the same version every time for our automation to work (don't use date)

oscarfv: This is problematic because AFAIK we need some sort of number that increases monotonically with each release for pacman to decide which version is newer. For git we use the commit count, but in this case it is not possible. We can use a counter that is manually updated, but that's fragile.

Do you have any idea? If not, I'll use the manually updated counter or hard-coded timestamp.

lazka: >  If not, I'll use the manually updated counter or hard-coded timestamp.

both sound fine

oscarfv: Okay, now it uses a counter.

It also changed the version schema to x.x.x+c.hash. Hope it is acceptable.

Thanks.

PR #7644, opened 2021-01-10
mpv rebuild

sboukortt:
Not sure exactly which dependency update triggered the need for this, but for me, rebuilding fixes `mpv` failing at startup with code 53.


lazka:
due to the libplacebo so version bump. thanks

PR #7649, opened 2021-01-11
python-pip: update to 20.2.4

naveen521kk:


lazka:
Shouldn't there be a new resolvelib dependency?

naveen521kk:
@lazka I will fix it. I will create a package for everything missing.

lazka:
I've added cryptography because "pip check" complained that pyopenssl needs a newer one.

I've tested pip/setuptools with various packages, venv, and some install/uninstall scenarios. All looks good. Thanks!

PR #7657, opened 2021-01-12
charls: new package

kmilos:
Lossless JPEG support: https://github.com/team-charls/charls

One of the optional dependencies of python-imagecodecs.


Biswa96:
> One of the optional dependencies of python-imagecodecs.

FYI, as you are updating python-imagecodecs in another pull request you could add this dependency in that one. charls will build together with python-imagecodecs.

kmilos:
Cool, I don't have to wait until charls gets published first?

Biswa96:
> I don't have to wait until charls gets published first?

Actually you have to wait for python-imagecodecs but you don't have to wait twice.

kmilos:
Thanks, commit moved over to #7656

PR #7658, opened 2021-01-12
PySide2 and friends: Fix url and build process

podsvirov:
The page at the old address is no longer available (404 - Page Not Found).


podsvirov:
Cc: @Alexpux

lazka:
looks like shiboken needs clang as a dependency?

podsvirov:
@lazka looks like you're right, but it's worth waiting for the end of the check.

PR #7668, opened 2021-01-14
v8: rebuild

lazka:
fixes #7667

PR #7669, opened 2021-01-14
[v8] Bump pkgrel to 2 to trigger a rebuild

Kwizatz:
The current package is failing at runtime due to the upgrade from icu67 to icu68.

Not sure how this is usually handled, so I am opting for the easiest solution.


Kwizatz:
closing as a duplicate of #7668

PR #7670, opened 2021-01-15
Add libbacktrace.pc

taozuhong:


Biswa96:
1. Increase `pkgrel` variable.
2. The `prefix` variable should not be hardcoded. Maybe try something with `sed` and `${MINGW_PREFIX}`.

taozuhong:
@Biswa96 got it, thanks for your suggestion.

Biswa96:
Wait! pkgconfig file is not changed. mingw64 is hardcoded.

lazka:
pkg-config ignores prefix anyway

PR #7671, opened 2021-01-15
gtkwave: update to 3.3.108 and add gtk3 package

umarcor:
This PR updates gtkwave to 3.3.108. At the same time, instead of building the gtk2 version only, `*-gtk2` and `*-gtk3` packages are provided. Should `*-gtkwave` be set as an alias of one of those?

### DIFF
--- a/mingw-w64-gtkwave/PKGBUILD
+++ b/mingw-w64-gtkwave/PKGBUILD
@@ -1,27 +1,32 @@
 # Contributor Ricky Wu <rickleaf.wu@hotmail.com>
+# Contributor Unai Martinez-Corral <unai.martinezcorral@ehu.eus>
 
 _realname=gtkwave
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.3.107
-pkgrel=3
-pkgdesc='GTKWave is a fully featured GTK+ based wave viewer for Unix, Win32, and Mac OSX'
+pkgver=3.3.108
+pkgrel=1
+pkgdesc='GtkWave, a fully featured GTK+ based wave viewer which reads VCD, GHW, LXT, LXT2, VZT and FST files (mingw-w64)'
 arch=('any')
 url='https://gtkwave.sourceforge.io/'
 license=('GPL')
 groups=("${MINGW_PACKAGE_PREFIX}-eda")
-depends=("${MINGW_PACKAGE_PREFIX}-gtk2"
-         "${MINGW_PACKAGE_PREFIX}-tk"
-         "${MINGW_PACKAGE_PREFIX}-tklib"
-         "${MINGW_PACKAGE_PREFIX}-tcl"
-         "${MINGW_PACKAGE_PREFIX}-tcllib"
-         "${MINGW_PACKAGE_PREFIX}-gtk-update-icon-cache"
-         "${MINGW_PACKAGE_PREFIX}-adwaita-icon-theme")
-makedepends=('perlxml'
-             'intltool'
-             "${MINGW_PACKAGE_PREFIX}-gcc")
+depends=(
+  "${MINGW_PACKAGE_PREFIX}-gtk2"
+  "${MINGW_PACKAGE_PREFIX}-tk"
+  "${MINGW_PACKAGE_PREFIX}-tklib"
+  "${MINGW_PACKAGE_PREFIX}-tcl"
+  "${MINGW_PACKAGE_PREFIX}-tcllib"
+  "${MINGW_PACKAGE_PREFIX}-gtk-update-icon-cache"
+  "${MINGW_PACKAGE_PREFIX}-adwaita-icon-theme"
+)
+makedepends=(
+  'perlxml'
+  'intltool'
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+)
 source=("https://gtkwave.sourceforge.io/${_realname}-${pkgver}.tar.gz")
-sha256sums=('48fe630b0d42585d702dc967c7db8803c6ba68b465ad6697a68aabc08f8576c5')
+sha256sums=('ece447340442e7ad029713789552e8392b75dd3808c882ac5193d42fce55eb3b')
 
 prepare() {
   cd "${srcdir}/${_realname}-${pkgver}"
--- a/mingw-w64-gtkwave/PKGBUILD
+++ b/mingw-w64-gtkwave/PKGBUILD
@@ -1,9 +1,15 @@
 # Contributor Ricky Wu <rickleaf.wu@hotmail.com>
 # Contributor Unai Martinez-Corral <unai.martinezcorral@ehu.eus>
 
+# NOTE: This package installs the GTK3 build by default, but provides GTK2 versions of three binaries too:
+# - gtkwave-gtk2.exe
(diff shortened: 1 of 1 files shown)


lazka:
We are trying to get rid of gtk2, so what about just updating the existing to gtk3?

umarcor:
@lazka, that's what I thought initially. This is a failing draft on purpose, because I wanted to discuss it with you (and other maintainers).

GTK2 and GTK3 versions of GtkWave are virtually the same from a file format support and low-level feature perspective. However, there is a very annoying bug in GTK3 that has prevented the developer from officialy dropping GTK2. See https://github.com/gtkwave/gtkwave/issues/3#issuecomment-568937496. There are also issues with touchscreen support (GtkWave works on ARM). See https://github.com/gtkwave/gtkwave/issues/3#issuecomment-569139993. A explained there, "gtkwave-gtk2" in fact supports GTK1/GTK2, and "gtkwave-gtk3" supports GTK2/GTK3. In the not so far future, "gtkwave-gtk2" is expected to be deprecated and "gtkwave-gtk4" should support GTK3/GTK4.

Therefore, at least two versions will exist in the foreseeable future. On MSYS2, three different builds are possible now:

- Subdir `gtkwave` with GTK2.
- Subdir `gtkwave-gtk3` with GTK2.
- Subdir `gtkwave-gtk3` with GTK3.
- [Futue] Subdir `gtkwave-gtk4` with GTK3.
- [Futue] Subdir `gtkwave-gtk4` with GTK4.

/ccing @gtkwave here for feedback. Tony, are the scroll/touchscreen issues present when using subdir `gtkwave-gtk3` and GTK2? I.e., can we deprecate using subdir `gtkwave` and focus on `gtkwave-gtk3` only, provided that GTK1 is not supported here?

gtkwave:
Note that you can actually drop anything from the "gtkwave3" package regarding packaging support.
The gtkwave3-gtk3 version still supports gtk2. See in ./configure:

 --enable-gtk3 Causes the GTK+ frontend to be built with gtk3
 instead of gtk2.

I'd say you can safely drop the legacy "gtkwave3" package support as the annoying bugs aren't triggered in the gtk3 source tree's executables if you don't build against gtk3. The only real difference between gtk2 support in the gtkwave3 and gtkwave3-gtk3 packages is that I had to migrate drawing to use Cairo instead of x11-flavored gdk primitives. Otherwise they should be functionally equivalent.

I'd say just migrate completely to gtk3 simply because I will drop gtk2 support eventually as I am sure distros will eventually quit shipping with it, toolkit bugs or not. For all I know, those bugs don't exist in gtk4. GTK maintainers have historically been unresponsive to fixing bugs, and since these have been out in the open for something like 3 years and I can't be the only person seeing this, I don't expect they will ever be fixed. 

-Tony

     On Friday, January 15, 2021, 10:10:50 AM EST, Unai Martinez-Corral <notifications@github.com> wrote:  
 
 


@lazka, that's what I thought initially. This is a failing draft on purpose, because I wanted to discuss it with you (and other maintainers).

GTK2 and GTK3 versions of GtkWave are virtually the same from a file format support and low-level feature perspective. However, there is a very annoying bug in GTK3 that has prevented the developer from officialy dropping GTK2. See gtkwave/gtkwave#3 (comment). There are also issues with touchscreen support (GtkWave works on ARM). See gtkwave/gtkwave#3 (comment). A explained there, "gtkwave-gtk2" in fact supports GTK1/GTK2, and "gtkwave-gtk3" supports GTK2/GTK3. In the not so far future, "gtkwave-gtk2" is expected to be deprecated and "gtkwave-gtk4" should support GTK3/GTK4.

Therefore, at least two versions will exist in the foreseeable future. On MSYS2, three different builds are possible now:
   
   - Subdir gtkwave with GTK2.
   - Subdir gtkwave-gtk3 with GTK2.
   - Subdir gtkwave-gtk3 with GTK3.
   - [Futue] Subdir gtkwave-gtk4 with GTK3.
   - [Futue] Subdir gtkwave-gtk4 with GTK4.

/ccing @gtkwave here for feedback. Tony, are the scroll/touchscreen issues present when using subdir gtkwave-gtk3 and GTK2? I.e., can we deprecate using subdir gtkwave and focus on gtkwave-gtk3 only, provided that GTK1 is not supported here?

umarcor:
@gtkwave, thanks for clarifying as I had wrote `gtkwave`/`gtkwave-gtk3` instead of `gtkwave3`/`gtwave3-gtl3`.

Assuming we drop `gtkwave3` and we use subdir `gtkwave3-gtk3` exclusively: are the scroll/touchscreen issues still present with GTK3? That is, is there a technical reason to keep providing a GTK2 build?

gtkwave:
The gtk3 bugs are still present. I suppose users will simply have to deal with the annoyances. I can't work around toolkit bugs and I can't support gtk2 forever. I might get the drag-and drop -> scrollbar position one fixed, but the touchscreen-related problems can't be fixed by me.

Touchscreen for menus as well as drag and drop for touchscreen in Wayland is messed up in ANY gtk3 application. Try it with the gtk3-demo program or the gnome-shell. I really can't believe this even passes their QA.

-Tony
     On Friday, January 15, 2021, 12:23:15 PM EST, Unai Martinez-Corral <notifications@github.com> wrote:  
 
 


@gtkwave, thanks for clarifying, as I had wrote gtkwave/gtkwave-gtk3 instead of gtkwave3/gtwave3-gtl3.

Assuming we drop gtkwave3 and we use subdir gtkwave3-gtk3 exclusively: are the scroll/touchscreen issues still present with GTK3? That is, is there a technical reason to keep providing a GTK2 build?

lazka:
Feel free to ship both then.

Just note that after gimp has switched to gtk3 we might start dropping gtk2 related libraries.

umarcor:
@gtkwave, thanks for your prompt and clear response, once again. Much appreciated.

@lazka, I will handle this locally. Since Tony also explained which files can be shared (almost all of them), I will try to provide two binaries in a single package, and I will make `gtkwave.exe` be the GTK3 version. As a result:

- The name of the package will not change.
- I will check that the codebase is updated from the GTK1/2 version to the GTK2/3 version.
- The default version that users will get is GTK3.
- Only 1 executable will be added, in case some user wants to run `gtkwave-gtk2`. I might not make gtk2 a runtime dependency, but document it.
- Whenever GTK2 is broken because libraries are dropped, we just stop providing that additional binary.

I will make this PR ready for review when I'm done.

gtkwave:
That sounds like good plan. The main reason there is a gtk3 port is when gtk2 support eventually is removed. 

Of course, gtk3 does allow some interesting features and the viewer takes advantage of them (e.g., fullscreen, touchscreen support, HiDPI (I think--provided by a user), etc.)

Note that I use touchscreen support in conjunction with a mouse/trackpad. Fingers are too big to be useful as pointers, but function great for flinging, pinching, stretching, etc. I don't see aspects of it being broken as a dealbreaker.

-Tony


     On Friday, January 15, 2021, 01:33:04 PM EST, Christoph Reiter <notifications@github.com> wrote:  
 
 


Feel free to ship both then.

Just note that after gimp has switched to gtk3 we might start dropping gtk2 related libraries.

umarcor:
@lazka, I could achieve the targets set in https://github.com/msys2/MINGW-packages/pull/7671#issuecomment-761117126. I think this is ready for review.

lazka:
thanks!

PR #7682, opened 2021-01-17
serial-studio: add 1.0.8

umarcor:
This PR adds [Serial-Studio/Serial-Studio](https://github.com/Serial-Studio/Serial-Studio): a "*multi-purpose serial data visualization & processing program*".

Serial-Studio

### DIFF
--- a/mingw-w64-serial-studio/PKGBUILD
+++ b/mingw-w64-serial-studio/PKGBUILD
@@ -1,0 +1,38 @@
+# Maintainer: umarcor <unai.martinezcorral@ehu.eus>
+
+_realname=serial-studio
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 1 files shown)


Biswa96:
Some comments:

1. Use lower case for the `_realname` variable.
2. If possible, a separate build folder is preferred.
3. In `package()` step, instead of manually creating paths with `mkdir` and `cp` commands, `install -D` command with other options is better choice, IMO.

umarcor:
@Biswa96, I addressed your concerns. At the same time, a new version was tagged some minutes ago, so I updated it.

The build error is arbitray. The build needs to be restarted.

umarcor:
See successful executions in my fork (same commit):

- https://github.com/umarcor/MINGW-packages/actions/runs/492257786
- https://github.com/umarcor/MINGW-packages/actions/runs/492224111

PR #7697, opened 2021-01-19
cx_Freeze 6.5.2 - fix #7480

### LINKED ISSUE #7480: cxfreeze missing .exe extension
## Describe the issue

The `cxfreeze` binary provided by `mingw-w64-x86_64-python-cx_Freeze` and possibly other related packages do not have an exe extension.

## Steps to Reproduce the Problem

1. pacman -S mingw-w64-x86_64-python-cx_Freeze
2. cxfreeze or cxfreeze-quickstart
3. bash: /mingw64/bin/cxfreeze: cannot execute binary file: Exec format error

well, it should run

## Additional info

I tried renaming the cxfreeze to cxfreeze.exe, but the problem remains.

marcelotduarte:
closes #7480

### DIFF
--- a/mingw-w64-python-cx_Freeze/PKGBUILD
+++ b/mingw-w64-python-cx_Freeze/PKGBUILD
@@ -5,12 +5,12 @@
 
 _realname=cx_Freeze
 pkgbase=mingw-w64-python-${_realname}
-pkgname=("${MINGW_PACKAGE_PREFIX}-python-cx_Freeze")
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
 provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 pkgver=6.5.2
-pkgrel=1
+pkgrel=2
 pkgdesc="Creates standalone executables from Python scripts, with the same performance (mingw-w64)"
 arch=('any')
 license=('PSF')
@@ -19,7 +19,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-python"
          "${MINGW_PACKAGE_PREFIX}-python-importlib-metadata")
 makedepends=("${MINGW_PACKAGE_PREFIX}-python-setuptools")
 options=('staticlibs' 'strip' '!debug')
-source=(cx_Freeze-${pkgver}.tar.gz::https://pypi.io/packages/source/c/cx_Freeze/cx_Freeze-${pkgver}.tar.gz)
+source=(${_realname}-${pkgver}.tar.gz::https://pypi.io/packages/source/c/cx_Freeze/cx_Freeze-${pkgver}.tar.gz)
 sha256sums=('869731afc624b511cdad0df2d3e8fb5433da488720838c02462a25dcdd37fcd4')
 
 prepare() {
@@ -34,16 +34,14 @@ build() {
 }
 
 package() {
-  local _mingw_prefix=$(cygpath -am ${MINGW_PREFIX})
+  local _mingw_prefix=$(cygpath -wm ${MINGW_PREFIX})
   cd python-cx_Freeze-${CARCH}
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
     ${MINGW_PREFIX}/bin/python setup.py install --prefix=${MINGW_PREFIX} \
     --root="${pkgdir}" --optimize=1 --skip-build
 
   # fix python command in files
-  pushd "${pkgdir}${MINGW_PREFIX}"/bin > /dev/null
-  for filename in cxfreeze cxfreeze-quickstart; do
-    sed -e "s|${_mingw_prefix}/bin/|/usr/bin/env |g" -i ${filename}
+  for filename in "${pkgdir}${MINGW_PREFIX}/bin/cxfreeze*script.py"; do
+    sed -e "s|${_mingw_prefix}/bin/||g" -i ${filename}
   done
-  popd > /dev/null
 }


lazka:
This leaves the build time path in the shebang, so I don't think this is right.

marcelotduarte:
No, an executable is generated.
remmina_teste6-vm-win10-rdp_192 168 6 56_2021119-12154,558487

Biswa96:
The command should be like this:

```sh
  # fix python command in files
  for _f in "${pkgdir}${MINGW_PREFIX}/bin/*.py"; do
    sed -e "s|/usr/bin/env |${MINGW_PREFIX}|g" \
        -e "s|${PREFIX_WIN}|${MINGW_PREFIX}|g" -i ${_f}
  done
```

marcelotduarte:
@lazka @Biswa96 setuptools already does the job.
Please apply the PR and see with your eyes.
remmina_teste6-vm-win10-rdp_192 168 6 56_2021119-133155,407523

Biswa96:
As you have shown in the screenshot the shebang is `C:/msys64/mingw64/bin/python.exe`. But not everyone install msys2 in that specific folder `C:/msys64`. So, that part should not be in shebang. It should be like `#!/mingw64/bin/python.exe` or `#!python.exe`.

marcelotduarte:
Now I understand what you mean.

> It should be like `#!/mingw64/bin/python.exe` or `#!python.exe`.

With the change, the script is now has a shebang as the first option, but doesn't run. The error is: failed to create process.
If I edit the file to use "#!python", "#!python.exe", the error is same....

Biswa96:
Interesting! I have build the package in my PC with that change and scripts run as usual. OK, wait for maintainer reply.

naveen521kk:
> With the change, the script is now has a shebang as the first option, but doesn't run. The error is: failed to create process.
If I edit the file to use "#!python", "#!python.exe", the error is same....

Well for some reason that happens when we install the package manually. But when we install from Pacman repo things are fixed. I don't why I have noticed that happen many times. Maybe others know about that better.

lazka:
Maybe you updated setuptools with pip? that would break things.

naveen521kk:
I tested it and it worked with shebang with full paths like `#!C:\msys64\mingw64\bin\python.exe` but not with the other. I could also see that the things which were previous installed before the update also had the full path in windows format. I think there should be something going on which I don't know...

I think setuptools launcher works like how it should work...

marcelotduarte:
Later I'll verify about setuptools.
1) But with the actual PKGBUILD the error stated in #7480 is confirmed. 
2) With my PR its run ok (a least for me and others that use the default directory)
3) With the change suggest by @Biswa96 it don't run. This is part of package()
```
package() {
  cd python-cx_Freeze-${CARCH}
    ${MINGW_PREFIX}/bin/python setup.py install --prefix=${MINGW_PREFIX} \
    --root="${pkgdir}" --optimize=1 --skip-build

  # fix python command in files
  for filename in "${pkgdir}${MINGW_PREFIX}/bin/*.py"; do
    sed -e "s|${PREFIX_WIN}|${MINGW_PREFIX}|g" -i ${filename}
  done
}
```

naveen521kk:
> With the change suggest by @Biswa96 it don't run. This is part of package()

Well, that's how things are done in pretty much every python packages with scripts.

marcelotduarte:
My setuptools has errors, but it isn't updated with pip. I usually use it on projects and test it only in a virtual environment, but it may have happened. The pip package also contains errors.
So I reinstalled msys2, but the errors continued with the suggested modification.
When I run `cxfreeze --version` the error is:
`failed to create process (\mingw64\bin\python.exe "C:\msys64\ming64\bin\cxfreeze-script.py" "--version").`
Then I changed it a bit more and now it runs with the shebang now !#python.exe

mingw-w64-python-cx_Freeze/PKGBUILD
```diff
@@ -34,16 +34,14 @@ build() {
...
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
     ${MINGW_PREFIX}/bin/python setup.py install --prefix=${MINGW_PREFIX} \
     --root="${pkgdir}" --optimize=1 --skip-build
 
   # fix python command in files
-  pushd "${pkgdir}${MINGW_PREFIX}"/bin > /dev/null
-  for filename in cxfreeze cxfreeze-quickstart; do
-    sed -e "s|${_mingw_prefix}/bin/|/usr/bin/env |g" -i ${filename}
+  for filename in "${pkgdir}${MINGW_PREFIX}/bin/cxfreeze*script.py"; do
+    sed -e "s|${_mingw_prefix}/bin/||g" -i ${filename}
```
naveen521kk, line 45: well, I think the shebang should always be an absolute path to the python interpreter. So, I will strongly recommend not having `!#python.exe`
need to see how it is done in other packages...

marcelotduarte: is done like the @Biswa96 says, but apparently with the new setuptools 51.3.1 in msys2 this not run.
I tried with:
`sed -e "s|${PREFIX_WIN}/bin|${MINGW_PREFIX}/bin|g" -i ${filename}`
but an error occurs like I commented:


> When I run `cxfreeze --version` the error is:
> `failed to create process (\mingw64\bin\python.exe "C:\msys64\ming64\bin\cxfreeze-script.py" "--version").`

> Then I changed it a bit more and now it runs with the shebang now !#python.exe

I mean:
#!python.exe

naveen521kk: Well, that shebang is for getting the exact python executable and there could be many python executables...
For example I get these
```sh
$ where python
C:\msys64\mingw64\bin\python.exe
C:\msys64\usr\bin\python.exe
```
So, that means wrong python executable can be picked and it shouldn't happen.

lazka: @naveen521kk the shebang gets interpreted by setuptools, I think it looks in the same directory first

marcelotduarte: I agree. And, the first and this commit is better than the current behavior that prevents the use of the command.

naveen521kk:
I think there is something going on Pacman which we don't know here. Let me explain, I downloaded https://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-cython-0.29.21-1-any.pkg.tar.zst and extracted to check the bin folder where I could find `cython-script.py` with these contents.
```py
#!/mingw64/bin/python.exe
# EASY-INSTALL-ENTRY-SCRIPT: 'Cython==0.29.21','console_scripts','cython'
__requires__ = 'Cython==0.29.21'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('Cython==0.29.21', 'console_scripts', 'cython')()
    )
```
Then I installed it using
```sh
pacman -S mingw-w64-x86_64-cython
```
ran this command and the output was
```sh
$ cat /mingw64/bin/cython-script.py
#!C:/msys64/mingw64/bin/python.exe
# EASY-INSTALL-ENTRY-SCRIPT: 'Cython==0.29.21','console_scripts','cython'
__requires__ = 'Cython==0.29.21'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('Cython==0.29.21', 'console_scripts', 'cython')()
    )
```
So, I think there is something going on which I don't know...

lazka:
yes, some packages have install scripts that get executed after installation: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-cython/cython-x86_64.install

naveen521kk:
> yes, some packages have install scripts that get executed after installation: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-cython/cython-x86_64.install

Well, then it only works with install scripts then? Then I think I have to update quite a few packages I recently did :fearful:

Issue #7636, opened 2021-01-10
[mathgl]  Cannot use MSYS2 build with FLTK


mayd:
Even though MSYS2 packages for MathGL and FLTK is installed I cannot compile a MathGL C program to use FLTK.
The following compiler error messageis displayed:

C:/msys64/mingw64/include/mgl2/Fl_MathGL.h:34:10: fatal error: mgl2/fltk.h: No such file or directory
   34 | #include <mgl2/fltk.h>
      |          ^~~~~~~~~~~~~
compilation terminated.

I was unable to rebuild the MSYS2 MathGL package with option "enable-fltk".

Can anyone suggest how to do this?

Biswa96:
Would you like to check if the linked pull request fixes your issue?

mayd:
After rebuilding the MathGL package with the updated PKGBUILD script you created, I found the following:

I tested some of the examples built as part of the MathGL package.

Initially I did find some problems with incorrect DLLs being loaded due to the PATH setting but I identified those 
using a dependency tracker and I believe they were all fixed.

The MathGL source included example which generates a set of files containing PNG images (mgl_example.exe) worked.

None of the other examples worked. Specifically, the FLTK example program (mgl_fltk_example.exe) hangs without
displaying anything.  The remaining examples also hang.

My own FLTK example which I copied from the MathGL documentation also does not build.

Compiler command line: 
g++ mgl_fltk_ex1.cpp -o mgl_fltk_ex1 -Wl,-subsystem,windows -lmgl2 -lmgl2-fltk

Result:
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/lib/../lib/libmingw32.a(lib64_libmingw32_a-crt0_c.o): in function `main':
C:/_/M/mingw-w64-crt-git/src/mingw-w64/mingw-w64-crt/crt/crt0_c.c:18: undefined reference to `WinMain'
collect2.exe: error: ld returned 1 exit status
make: *** [Makefile:4: mgl_fltk_ex1] Error 1

Possibly, there is an error in the above ommand line but it is similar to the one given in the MathGL documentation (which
also did not work)


[...]

Issue #6623, opened 2020-06-24
MathGL simple test program hangs indefinitely


ijean:
Hi,

I've installed MathGL using MSYS2, successfully build one of their C++ examples:

```
#include <mgl2/mgl.h>

int main() {
	mglGraph gr;
	gr.FPlot("sin(x)");
	gr.WriteFrame("test.png");
}
```

Build with:

```
g++ test.cpp -lmgl
```

Launching the resulting executable seems to hang indefinitely (never ends).

Regards,
Jean

kvk1920:
I have the same issue

mati865:
Looks like another case of winpthreads bug:
```
ntdll!NtWaitForMultipleObjects+0x14
KERNELBASE!WaitForMultipleObjectsEx+0xf0
KERNELBASE!WaitForMultipleObjects+0xe
libwinpthread_1!pthread_cond_init+0x198
libwinpthread_1!sem_wait+0xb9
libgomp_1!gomp_sem_wait+0x15
libgomp_1!gomp_barrier_wait_end+0x70
libgomp_1!gomp_team_start+0x6da
libgomp_1!GOMP_parallel+0x44
libmgl!ZN7mglFont4LoadEPKcS1_+0x9c2
libmgl!ZN7mglFontC2EPKcS1_+0x1ef
libmgl!ZN11mglCanvasGL3ClfEPKc+0x46e55
libmgl!ZN11mglCanvasGL3ClfEPKc+0x61e
libmgl+0x128b
ntdll!LdrpCallInitRoutine+0x61
ntdll!LdrpInitializeNode+0x1d3
ntdll!LdrpInitializeGraphRecurse+0x42
ntdll!LdrpInitializeGraphRecurse+0xc8
ntdll!LdrpInitializeProcess+0x1fd6
ntdll!_LdrpInitialize+0x4db5b
```
Rebuilding MathGL linked to LLVM's OpenMP instead of libgomp doesn't help because it has another issue because it hangs another way.

nuxil:
I too get this bug using C, not C++
Only difference is that i am using another header.  "mgl2/mgl_cf.h" 

The example compile fine. no warnings or errors. 
It just hangs when executed.

Is there any solutions to this ?

mayd:
Could be related to my issue: #7636.

dragon-archer:
This should have been fixed by #7720

PR #7720, opened 2021-01-21
mathgl: Disabled OpenMP for MSYS2/MINGW

mayd:
Workaround for following issues:

#6623 MathGL simple test program hangs indefinitely.

#7636  [mathgl] Cannot use MSYS2 build with FLTK.


Biswa96:
Increase `pkgrel` variable by 1.

PR #7739, opened 2021-01-23
rawpy: new package, 0.16.0

kmilos:
RAW image processing for Python, a wrapper for libraw

No source available on [PyPi](https://pypi.org/project/rawpy/#files) for eventual pip install, only statically linked MSVC wheels.


lazka:
thanks

PR #7743, opened 2021-01-23
Update lazy-object-proxy to version 1.5.2

danyeaw:


Biswa96:
I did get any error while building this change. Maybe a restart in CI solve it.

lazka:
I think this needs a `export SETUPTOOLS_SCM_PRETEND_VERSION=${pkgver}`

PR #7753, opened 2021-01-24
clang: package function tweaks

jeremyd2019:
Fix depends is not an array in libc++/libc++abi

Fix name libcxx -> libc++.

Use for loop to generate package function wrappers, suggested in https://github.com/msys2/MSYS2-packages/pull/2304#issuecomment-765891946

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -541,5 +541,7 @@ package_polly() {
 
 # Wrappers
 for _name in "${pkgname[@]}"; do
-  eval "package_${_name}() { package_${_name#${MINGW_PACKAGE_PREFIX}-}; }";
+  _short="package_${_name#${MINGW_PACKAGE_PREFIX}-}"
+  _func="$(declare -f "${_short}")"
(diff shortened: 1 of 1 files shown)


Biswa96:
Don't they need some `conflicts` and `replaces` variable?

jeremyd2019:
The packages themselves are the same names, the only names that changed are the internal function names that are then wrapped, either by the explicit wrapper functions or now by functions created by eval in the for loop.

lazka:
uh, good point. "MINGW_INSTALLS=mingw64 makepkg-mingw --printsrcinfo" fails to pick up depends on both cases

lazka:
ok, even more hacky, but seems to handle all cases:

```bash
for _name in "${pkgname[@]}"; do
  _short="package_${_name#${MINGW_PACKAGE_PREFIX}-}"
  _func="$(declare -f $_short)"
  eval "${_func/$_short/package_$_name}"
done
```

lazka:
lgtm

jeremyd2019:
Superseded by #7878

PR #7770, opened 2021-01-24
Update path.py to version 12.5.0

danyeaw:


Biswa96:
upgrade or downgrade? 13.2.0 -> 12.5.0 ??

danyeaw:
@Biswa96 Whoops, you found a hole in my automation! I'm trying to get all of these Python packages up to date 👍

Biswa96:
You could just force push a commit with new version. Also https://xkcd.com/1319/

danyeaw:
@Biswa96 Yup, that is a good xkcd 😄. In this case creating the script took about 1 hour and I have created 50 PRs using it, so I think the time was well spent. Manually determining which Python packages need to be updated for the 300+ packages is not something I like to spend my time doing.

https://github.com/danyeaw/mingw-package-updater

PR #7777, opened 2021-01-24
vulkan: Update to 1.2.166

Biswa96:


Biswa96:
Let the water be settled. glslang is not updated in mirrors.

PR #7803, opened 2021-01-25
libgphoto: Update to 2.5.26

peterbud:
PS: I don't have any idea why the 32-bit doxygen generation is failing, while the 64-bt doxygen is OK:
`error: tag OUTPUT_DIRECTORY: Output directory 'doxygen-output' does not exist and cannot be created`


lazka:
looks like the build was just flaky..

PR #7805, opened 2021-01-25
libspatialite 5.0.0

jeroen:


lazka:
Any ABI breaks with this one?

jeroen:
I suspect not. The dll is still called `libspatialite-4.dll` and I didn't see a revision bump in homebrew.

lazka:
sounds good, thanks

PR #7807, opened 2021-01-26
arrow: Update to 3.0.0

kou:


kou:
This includes Groonga update proposed by #7808 because Groonga depends on Apache Arrow.
If we merge Apache Arrow first, old Groonga package is broken.
If we merge Groonga first, new Groonga packages is broken after Apache Arrow is merged.
If we merge both of them at once, Groonga isn't broken.
Suggested by @Biswa96.

PR #7808, opened 2021-01-26
groonga: Update to 10.1.1

komainu8:


Biswa96:
Just FYI. There is a separate pull request to update arrow by @kou. groonga seems to depend on arrow. Can anyone of you move your change to other one? We may prevent any possible API breakage between them due to version update.

kou:
OK. I'll merge this into #7807.

Biswa96:
Make sure @komainu8 is agreed with that 😊

komainu8:
OK. I agree with that.

PR #7809, opened 2021-01-26
binutils: update to 2.36

Astrum-polaris:
Remove patches merged upstream


Astrum-polaris:
Do we want to use the new default of dll characteristics and imagebase now?

lazka:
Thanks!

> Do we want to use the new default of dll characteristics and imagebase now?

Yes, eventually.

* I'd update to 2.36 while keeping defaults as they are now and wait a bit
* switch all defaults and write a NEWS entry on how to disable them all

@jeremyd2019 what would you do?

jeremyd2019:
I have been trying to keep an eye on issues reported to this repo, that could be attributed to these changes.  Most are due to pointer truncation bugs, and have been fixed in the package, but the main unresolved issue is [upstream binutils PR 26659](https://sourceware.org/bugzilla/show_bug.cgi?id=26659).  Later on I found that this same issue had already been encountered and reported as [upstream GCC PR 64697](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64697) for cygwin targets, since they already had default image base above 4GB.  It doesn't appear that the offer to report that to binutils was ever acted on, because I did search their bugzilla before opening a new issue.

I haven't seen any new issues in packages related to these changes for a while, that hopefully is an indication that most have been found.

lazka:
Thanks. We'll keep the extra options to revert those things, so users at least have a workaround, and then maybe upstream will have another look if it affects more people.

I gave this a quick test run with multiple packages and all looks fine

@Astrum-polaris thanks!

Issue #7207, opened 2020-10-30
Package request libmt32emu and Munt


DominusExult:
Hi,
could you add the libmt32emu to your packages?
Its repository is in https://github.com/munt/munt/tree/master/mt32emu and is CMAKE based.
It would really help if there were a msys2 package to automate everything instead of having to build the lib manual. For Exult (exult.info) we are using an old built in version of the mt32emu but we would like to switch to be dependent on the lib instead and our compile instructions for Windows use Msys2 and all its packages extensively :)

Thanks for your consideration!

DominusExult:
i've made a pkgbuild for the standalone lib, for now naming it mt32emu.
[PKGBUILD.txt](https://github.com/msys2/MINGW-packages/files/5866631/PKGBUILD.txt)

The problem with Munt is that it is a project with mutliple sub projects:
mt32emu : the library which can be standalone (shared build) for use in other projects (which I aimed at) or as a dependency of mt32emu_qt with other build rules (static build one of those)
mt32emu_qt : a QT5 graphical interface for playback, smf2wav conversion etc.
mt32emu_smf2wav: a console application for smf2wav conversion only.
(and an alsa driver, and a Windows driver)

The main CMakeLists.txt only builds _qt and _smf2wav, so my proposal would be to have a package mt32emu for the lib and a package munt for _qt and _smf2wav. The upside being that if you only need the lib you don't need to build all of QT5 and its dependencies.

I'll see if I can create the munt package and if that works with an extra lib package.

DominusExult:
[PKGBUILD.txt](https://github.com/msys2/MINGW-packages/files/5867315/PKGBUILD.txt)
This is the pkg for the munt programs. The package also creates a static lib but not a shared one (as it needs different options than the shared one provides).
The only thing these two packages provide both the same are the include header files.

DominusExult:
Thank you for accepting my pull request!

PR #7810, opened 2021-01-26
packages for munt and its mt32emu library.

DominusExult:
munt builds the two programs mt32emu-qt (a QT5 GUI program), mt32emu-smf2wav (commandline tool) as well as the static lib of libmt32emu. mt32emu builds the shared library libmt32emu for use with other programs.
I've split it up because
- the two munt programs need the static build of the lib and the lib's cmake file explicitly disables building shared when building the two programs (https://github.com/munt/munt/blob/master/mt32emu/CMakeLists.txt#L11)

- the lib has no dependencies, so requiring QT5, glib2, portaudio just for the lib seems excessive to me

Package request #7207

I understand that the policy is rather not split the package but I have found no way to do this. Maybe packages support a second build packaging phase but that is beyond me. But I'll surely look into it if you point me there.
Additionally I'm aware MSYS2 likes to be in sync with Arch Linux (https://aur.archlinux.org/packages/munt/) but from what I can tell this also does not provide the shared lib.

### DIFF
--- a/mingw-w64-mt32emu/PKGBUILD
+++ b/mingw-w64-mt32emu/PKGBUILD
@@ -1,0 +1,44 @@
+# Maintainer: Some One <some.one@some.email.com>
+
+_realname=mt32emu
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=2.4.0
+pkgrel=1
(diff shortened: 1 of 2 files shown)


DominusExult:
*NOW* I realize that both packages providing the same headers is truly not ideal. I'll see what I can do to fix that

PR #7812, opened 2021-01-26
Muntpackage

DominusExult:
Package(s) for Munt.
The first commit is an all in one solution for the Munt project.

The second commit splits off the mt32emu library for people that just need the library and can do without all the QT5, glib2 and portaudio overhead. Several linux distros are also making use of the library like this.

Done as two commits so you can easily decide the course of action you want to take.


DominusExult:
Fixing indentation, full url, depends vs makedepends and resubmitting a pull request

PR #7819, opened 2021-01-27
python: Extend MS_WINDOWS to check for mingw + Tweak patch 0440

longnguyen2004:
When building without `openssl`, `split_var('OPENSSL_LIBS', '-l')` returns an empty tuple, even though it's supposed to return a list, through list comprehension. As a result, appending a list into the empty tuple will produce a `TypeError`. I changed the list into a tuple so that it'll always succeed.

Also, some modules rely on `MS_WINDOWS` to check for Windows, which fails on mingw since we set `_PYTHON_HOST_PLATFORM=mingw`. I extended the flag to check for both `win32` and `mingw`


Biswa96:
Have you consider to put this change in upstream? If yes then a link to the PR or issue can be added.

longnguyen2004:
> Have you consider to put this change in upstream?

I haven't thought about it yet, since this is highly dependent on what we already did to get Python to build. These changes by themselves don't really mean much without the rest of the patches. Also, given that upstream is reluctant to add mingw support in the first place, I decided not to submit it.

lazka:
> Also, some modules rely on `MS_WINDOWS` to check for Windows, which fails on mingw since we set `_PYTHON_HOST_PLATFORM=mingw`. I extended the flag to check for both `win32` and `mingw`

What does this fix?

longnguyen2004:
> What does this fix?

I encountered this while cross compiling. The build system set `_PYTHON_HOST_PLATFORM=mingw`, which gets propagated to the `HOST_PLATFORM` variable inside `setup.py`. Since the `MS_WINDOWS` only checks for `win32`, various modules failed to build (`_multiprocessing`, `_testinternalcapi`, etc).

1480c1:
Could you perhaps put that in the commit message? It would help in the future if someone needs to find out why these changes were added etc

longnguyen2004:
> Could you perhaps put that in the commit message? It would help in the future if someone needs to find out why these changes were added etc

Can I just add a link to my comment here, like @Biswa96 has suggested?

lazka:
Should _PYTHON_HOST_PLATFORM be set to win32 then? There is no mingw platform.

longnguyen2004:
It is from [this patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python/0130-MINGW-configure-MACHDEP-and-platform-for-build.patch#L40), which sets `_PYTHON_HOST_PLATFORM` to `mingw` when cross-compiling. This value is then passed to the `setup.py` script, which [uses that value](https://github.com/python/cpython/blob/v3.8.7/setup.py#L33) as the host platform.

Should we modify 0130 to always use `win32` then?

lazka:
ah, I see. I'll have a look

PR #7826, opened 2021-01-28
spirv-cross: Add package

Biswa96:


Biswa96:
Do not merge now. Wait for the current mpv update. Otherwise this will be linked with older libplacebo. Also it is not required to increas pkgrel in libplace just for this.

lazka:
> Do not merge now. Wait for the current mpv update. Otherwise this will be linked with older libplacebo.

autobuild has a newer placebo, so It should work imo

Biswa96:
If I use `ntldd` command on the artifact here mpv.exe shows ` libplacebo-72.dll => not found`.

lazka:
yes, the CI builds against the pacman repo, autobuild is smarter and builds against all packages in the queue + the pacman repo.

Biswa96:
OK, feel free to whatever you want.

Biswa96:
Nope, it linked with previous libplacebo.

lazka:
mpv hasn't been built yet. Why do you think so?

Biswa96:
O! It's different than GHA builds?

lazka:
It gets build here: https://github.com/msys2/msys2-autobuild/actions/runs/521027269

The build status is here: https://packages.msys2.org/queue

Issue #7827, opened 2021-01-28
python-numpy: Illegal instruction (x86_64)


podsvirov:
Illegal instruction when importing 64-bit package, but all good with 32-bit.

```bash
user@host MSYS ~
$ pacman -Ss python-numpy
mingw32/mingw-w64-i686-python-numpy 1.19.5-2 [installed]
    Scientific tools for Python (mingw-w64)
mingw64/mingw-w64-x86_64-python-numpy 1.19.5-2 [installed]
    Scientific tools for Python (mingw-w64)
```

```bash
user@host MINGW64 ~
$ python
Python 3.8.7 (default, Jan 18 2021, 21:09:34)  [GCC 10.2.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
Illegal instruction

user@host MINGW64 ~
$
```

```bash
user@host MINGW32 ~
$ python
Python 3.8.7 (default, Jan 18 2021, 20:48:42)  [GCC 10.2.0 32 bit] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> numpy.__version__
'1.19.5'
>>>
```

naveen521kk:
I can reproduce this.
Maybe does triggering a rebuild fix this?

naveen521kk:
I wonder what does this `Illegal instruction` mean though.

podsvirov:
I rebuilt the package locally, but that didn't fix the problem.

mati865:
Cannot reproduce with up to date MSYS2. What CPUs do you have?

> 
> 
> I wonder what does this `Illegal instruction` mean though.

It often means the instruction is not supported by the CPU or panic inside panic (Rust terminology).

Biswa96:
Disclaimer: I have zero idea how python works at low level.

So, I have run `F:\msys64\mingw64\bin\python3.exe -c "import numpy"` in debugger. I follow the libpython3.8.dll, into that Py_Main -> Py_RunMain -> PyRun_SimpleStringFlags -> PyRun_StringFlags, then bam! instruction pointer stops at libopenblas.dll. It stops at `vmovaps` instruction.

```asm
libopenblas:000000006EC58E63 vmovaps xmmword ptr [rsp+170h], xmm6
libopenblas:000000006EC58E6C vmovaps xmmword ptr [rsp+180h], xmm7
[...]

PR #7830, opened 2021-01-29
openblas: Specify target CPU architecture to restrict advanced SIMD instructions.

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/7827


Astrum-polaris:
Since the make command had `DYNAMIC_ARCH=1` in it, I think you can try `ARCH=x86_64 CORE=CORE2` instead of `TARGET=CORE2`
Source: https://github.com/xianyi/OpenBLAS/blob/v0.3.13/Makefile#L88

Biswa96:
What will be the benefit of that change? I think with `TARGET` both mingw32 and mingw64 treated in same region.

lazka:
that looks good to me, can we merge?

PR #7836, opened 2021-01-29
setuptools-scm: rework patches

lazka:
This patches setuptools-scm to make its git test suite pass.
Instead of converting absolute paths we just avoid them.

The upside here is that it is potentially upstreamable.. though not sure


lazka:
@naveen521kk

naveen521kk:
Possibly send this patch to upstream and see their response?

lazka:
yeah, upstream still supports python2 so that wont work directly, after they are py3 only I'll try

naveen521kk:
Let me try porting this to py2 and send them a PR.

mingw-w64-python-setuptools-scm/0001-git-Use-show-prefix-instead-of-show-toplevel-for-fin.patch
```diff
@@ -0,0 +1,67 @@
...
++        from pathlib import Path
++        import os
++        wd = os.path.abspath(wd)
++        real_wd, _, ret = do_ex("git rev-parse --show-prefix", wd)
+         if ret:
+             return
++        if not real_wd:
++            real_wd = wd
++        else:
++            real_dw = str(Path(wd).parents[len(Path(prefix).parents) - 1])
```
jeremyd2019, line 61:
```suggestion
+            real_wd = str(Path(wd).parents[len(Path(prefix).parents) - 1])
```

lazka: uh

lazka: pushed to master

naveen521kk, line 61: looks like prefix isn't defined no?

lazka: yes. i changed that in master. thanks

PR #7843, opened 2021-01-30
Fix launcher for manim

naveen521kk:


Biswa96:
FYI, if the `${MINGW_PREFIX}` is removed the result may also be same.

naveen521kk:
> FYI, if the ${MINGW_PREFIX} is removed in the sed command then the result may also be same.

Can you give the exact sed command?

Biswa96:
> Can you give the exact sed command?

Here:

```sh
  # fix python command in files
  for _f in "${pkgdir}${MINGW_PREFIX}"/bin/*.py; do
    sed -e "s|/usr/bin/env ||g" \
        -e "s|${PREFIX_WIN}/bin||g" -i ${_f}
  done
```

naveen521kk:
@Biswa96 it creates a shebang like `#!/python.exe`. My aim was to not have any shebang.

naveen521kk:
please squash while merging

PR #7845, opened 2021-01-30
serial-studio: update to 1.0.11

umarcor:
The build error is arbitrary, as commented in #7682.


lazka:
Why the git change?

umarcor:
@lazka, because git submodules were added upstream. I believe that retrieving the archive (tarball/zipfile) downloads the repo only, but it does not retrieve submodules recursively.

lazka:
ok, thanks

PR #7848, opened 2021-01-30
scons: Update to 4.1.0

lazka:


lazka:
@Schala godot-cpp should work after this

revelator:
nice one :) 💯

PR #7870, opened 2021-02-01
binutils: fix windres quoting issues

lazka:
our space handling patch conflicts with https://sourceware.org/bugzilla/show_bug.cgi?id=26865
which results in ffmpg failing to build with something like
"fatal error: config.h: No such file or directory"

Drop that patch and backport https://sourceware.org/bugzilla/show_bug.cgi?id=4356
instead which seems to do something similar.

Fixes the ffmpeg build.


ddennedy:
Compiling ffmpeg (not the msys2 package) is still failing for me with build 11 hours ago:
https://github.com/mltframework/shotcut/runs/1813003452?check_suite_focus=true

Need to wait for something to rebuild and update?

naveen521kk:
> Need to wait for something to rebuild and update?

Yeah, you should wait for the package to be signed and uploaded...
you can see the status at https://packages.msys2.org/queue

PR #7879, opened 2021-02-03
crt/headers/winpthreads: clang prefix handling

jeremyd2019:
My experiments with clang on aarch64 have been going well, I figured it was time to start getting changes reviewed for inclusion back into MINGW-packages.  These packages seemed like a good place to start, being moderately complicated (not like mingw-w64-clang!), but containing representative techniques.

The packages for the existing MINGW_INSTALLS should be unchanged.

If the MINGW_PACKAGE_PREFIX contains `-clang-` (a "clang prefix"), default to ucrt instead of msvcrt.  (This might also be a handy way to detect, say, `-ucrt-` in MINGW_PACKAGE_PREFIX for a gcc+ucrt variant).

If the CARCH is not i686 or x86_64, default WIN32_WINNT to 0xA00 (Windows 10), under the assumption that any new arches will not be relevant for any older versions.

Appropriate crt configure options for armv7 and aarch64 are added.

winpthreads now use the dynamically-generated package wrapper functions.

### DIFF
--- a/mingw-w64-crt-git/PKGBUILD
+++ b/mingw-w64-crt-git/PKGBUILD
@@ -42,6 +42,7 @@ prepare() {
 
 build() {
   msg "Building ${MINGW_CHOST} CRT"
+  local _crt_configure_args
   case "$CARCH" in
     i686)
       _crt_configure_args="--disable-lib64 --enable-lib32"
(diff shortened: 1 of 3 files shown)


Biswa96:
Thank you very much for this apporach. One little request. The `--with-default-win32-winnt` option can be removed now if you use latest commit of mingw-w64. It is now set to Win10 by default.

Biswa96:
Thank you very much for this apporach. One little info. The `--with-default-win32-winnt` option can be removed now if you use latest commit of mingw-w64. It is now set to Win10 by default.

lazka:
>  The `--with-default-win32-winnt` option can be removed now if you use latest commit of mingw-w64. It is now set to Win10 by default

I don't mind the important bits being explicit and not depending on mingw defaults.

Biswa96:
Is it possible to add those arm and arm64 packages in Pacman repo?

PR #7880, opened 2021-02-03
ghdl: update to 1.0.0

umarcor:
After almost 20 years of development, GHDL 1.0.0 was tagged! :tada:

I didn't remove `pkgver()`, because I expect it to be required in future updates.

@lazka, is there any problem with releasing `1.0.0` and the next update being something such as `1.0.0.r75.g55555555`?

### DIFF
--- a/mingw-w64-ghdl/PKGBUILD
+++ b/mingw-w64-ghdl/PKGBUILD
@@ -1,7 +1,7 @@
 _realname=ghdl
 pkgbase="mingw-w64-${_realname}"
 pkgname='__placeholder__'
-pkgver=0.37.0.r1370.g7135caee
+pkgver=1.0.0
 pkgrel=1
 pkgdesc='GHDL: the open-source analyzer, compiler, simulator and (experimental) synthesizer for VHDL'
 arch=('any')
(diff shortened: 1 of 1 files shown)


Biswa96:
> is there any problem with releasing 1.0.0 and the next update being something such as 1.0.0.r75.g55555555?

If you just follow the basic [PKGBUILD version][1] rules there will be no issue. Also there is a [vercmp][2] tool which can help to compare version numbers to check if it is downgrade or upgrade.

[1]: https://wiki.archlinux.org/index.php/PKGBUILD#Version
[2]: https://man.archlinux.org/man/vercmp.8

lazka:
What's the point of switching between a dynamic pkgver and a static one?

If ghdl does proper releases you could just switch to building tarballs like other packages.

umarcor:
@lazka, GHDL is tagged once a year. Meanwhile, commits shas are used. We hope to have some more structured tagging in the future, but it is not ready yet. That's why we've been using dynamic pkgver for the last months. However, this specific commit is tagged nicely, so I thought it'd be good to have `1.0.0` instead of `1.0.0.r0.g2fb2384d` (which is the output of pkgver). The next tag, in a couple of weeks or month, will use `pkgver()` again.

@Biswa96, thanks for the reference to vercmp. I was told about it before, but I forgot. Sorry about that.

lazka:
ok, fair enough.

PR #7882, opened 2021-02-03
mingw-w64: Update

lazka:


jeremyd2019:
Good timing - I was cherry-picking [97806c](https://sourceforge.net/p/mingw-w64/mingw-w64/ci/97806c0ecf06b25295200d636978b06e28aa89bc/) for arm

lazka:
built glib, ffpmeg, python, all looks good.

PR #7884, opened 2021-02-03
pyqt3d: new package

melven:
New package for mingw-w64-pyqt3d very much copied from mingw-w64-python-pyqt5.

Tested and worked fine for me...<br>

Locally, I had to remove the --confirm-licence for sip-build though.
In this pull request, the flags are identical to those in the python-qt5 package.


melven:
Ok, the CI also does not like the `--confirm-license` flag. Should I remove it?

lazka:
hm, thanks. It would be nice if only those things were copied which are actually needed for this package.

I can try to help if you want though.

melven:
What should a change / remove? I just assumed that most things for building PyQt3d are very similar to PyQt5.

lazka:
ok, I'll have a look tomorrow

PR #7885, opened 2021-02-03
tixi: new package

melven:
mingw-w64 package for the samll XML library interface tixi (https://github.com/DLR-SC/tixi)


lazka:
Why such an old version?

The linux packages are all called "tixi3", it would be nice if we'd have a similar naming.

melven:
Sorry, our software still uses the old interface...

But I can also add a tixi3 package - they should be independent (everything prefixed with tixi3 to avoid file/name clashes as far as I know!)

lazka:
> Sorry, our software still uses the old interface...

puh... 3.x is 3 years old :)

> But I can also add a tixi3 package - they should be independent (everything prefixed with tixi3 to avoid file/name clashes as far as I know!)

If you want to, sure. I'd rename this to "tixi", and if you package 3.0 then call it "tixi3", to be in line with other distros.

thanks

PR #7886, opened 2021-02-04
libraw: force use of OpenMP

kmilos:
See https://github.com/LibRaw/LibRaw/pull/378#issuecomment-770166755

This takes only the relevant part of the upstream commit.


Biswa96:
If the change is an upstream commit would you like to add that commit link as a comment above the `patch` command? Also don't forget to add quotes in the `install` command.

kmilos:
> If the change is an upstream commit would you like to add that commit link as a comment above the `patch` command?

Ordinarily yes, but this upstream commit has touched some 50 files, while I need just this one hunk.

> Also don't forget to add quotes in the install command.

Copied from the [template](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates), sorry. Maybe it should be updated?

PR #7894, opened 2021-02-05
uasm: fix build

### LINKED ISSUE #7893: uasm: should this be a git package?
I just ran into a source file verification error trying to build uasm.  It was downloading a github archive, so I figured that something internally to github changed, but then I couldn't even find the tag.  Turns out that the PKGBUILD is downloading an archive for a branch, not a tag.  It seems like this should really be a git package (not sure on the current best practice for renaming to -git though).

jeremyd2019:
Change sources to git repo, pinned to same commit that the package was last built against.

Fix multiple definition errors linking by adding -fcommon to compiler flags.

Fixes #7893

Probably still needs some pkgver adjustment, not sure how to go about that on something that uses branches for versions instead of tags.

### DIFF
--- a/mingw-w64-uasm/PKGBUILD
+++ b/mingw-w64-uasm/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=uasm
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=v2.50
+pkgver=2.50.0.623.8d1afb0
 pkgrel=1
 _commit='8d1afb0a040217f87106b0c6b924bf9eb0254a2b'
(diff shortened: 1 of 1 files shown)


Biswa96:
Add the `pkgver()` function like in ghdl or mingw-w64-headers,crt etc. Then `makepkg` will handle the rest.

jeremyd2019:
Those use `git describe --long` which require there to be a tag.  `git describe --long --tags` works for this, but goes back to v213, not referring to v2.50.  It looks like H/globals.h has a `_UASM_VERSION_STR_` that might get the first part, but don't know where the revision count would be counted from.

jeremyd2019:
The mingw-w64 packages turned out to be a better example than the ghdl one, was able to put something together.

lazka:
(I hope that change is OK)

Issue #7887, opened 2021-02-04
GTKMM4 packages missing from repository


psousa8:
Version 4 of GTk+ and GTKMM packages were released and are missing from MSYS2/MINGW packages.

Version 4 of GTKMM library bring library improvements and multiples applications are making the shift for this new version. Support on GTK mailling lists are pushing for the installation of this new version and novel developments in previous versions will be reduced over time.

Biswa96:
Would you like to check the packages I have added in the linked pull request? You can download the unsigned packages from GitHub Actions' artifacts.

psousa8:
Hi. I have tested the packages and they were correctly installed and tested with simple test examples.

From the small tests I have made, there is a point I want to emphasis. In order to compile the example I need
`pkg-config --cflags gtkmm-4.0 libxml++-3.0 glibmm-2.68`
Should glibmm be a dependency from gtkmm pc file?

I not an expert on packaging, but as glibmm increases its numbering, will package name glibmm-2.68 will also change name?

Biswa96:
> will package name glibmm-2.68 will also change name?

Package name is not generally changed and when it is changed user (generally) don't have to do anything, maintainer and Pacman will take care of it. I copied the package name from AUR so that we'll snyc with them.

lazka:
gtkmm4 is in the repo now: https://packages.msys2.org/base/mingw-w64-gtkmm4

psousa8:
thanks for your work.

PR #7895, opened 2021-02-05
gtkmm4: Add package

Biswa96:
Related issue https://github.com/msys2/MINGW-packages/issues/7887


naveen521kk:
I would love to do something similar for Python also.

lazka:
I think you can drop atk, gtk4 doesn't depend on it.

PR #7903, opened 2021-02-06
glib2: Update to 2.66.6

Biswa96:


lazka:
re cflags, alternative is to set the buildtype like in gtk3

Biswa96:
Apology Sir, I can not understand that. According to the `meson.build` file, the `G_DISABLE_CAST_CHECKS` cflag is added when optimization option is enable, not the build type 🤔 ALthough that cflag is not so important, should I remove it?

lazka:
optimization is set according to build type. I'd just copy what gtk3 does atm. I think it's a good idea.

lazka:
ah, that buildtype thing was already in place. never mind then.

Biswa96:
> that buildtype thing was already in place.

Though your comment gave me a nudge. I will add thosse buildtype option in the new gtkmm4 PR.

lazka:
Note that gtk and glib are the only projects I know that change the build config according to buildtype. But it can't hurt, so sure.

Biswa96:
Interesting, if I add release build type in glibmm268 (in my other PR) the project is built three times (build, check, install). SO, scratch my previous idea.

Issue #2745, opened 2017-08-05
pip vs pacman when installing Python modules / packages


Ede123:
Many packages containing binary components are available as MINGW-packages (`mingw-w64-python-...`).

If **pacman** is used to install them the files are copied to Python's "site-packages" directory. They work and can be used. So far so good...

If I know accidentally use **pip** to upgrade such a package I'm unsure what will happen, but at the very least it will result in modified files on disk pacman knows nothing about and potential update conflicts in future.

My questions therefore are:
* Is there a possibility to "distinguish" Python modules installed by pip and pacman, so I do not accidentally use pip to update a package that should be managed by pacman?
* Is there a possibility to "fix" this at the start by changing how python modules from MINGW-packages install themselves?

lazka:
You can use "pip install --user" to install them to you user directory but then you get conflicts for i686 vs x86_64

Or use virtualenv

mikofski:
@lazka actually I am having issues with virtualenv (installed via either pip2 or pip3), I'll post the full traceback shortly (sorry) but it raises and `ImportError` exception, "no module called pkgutils" which _does_ exist in both python2 and python3, it's part of setuptools (you can not install it separately, it's a dummy package). From what I can tell, it creates the folder structure for the new virtualenv, and starts to install pip, setuptools, and wheel, but then raises the error in a subprocess call, where it executes a script that contains the import. Execute the script in the base environments of either python2 or 3 and the imports work fine, so something in setting up the folder structure of the virtualenv is failing, limited by a configuration limitation or other issue in my msys2 environments. I will investigate and post back more when I have time.

If anyone thinks this is a legitimate issue, and should be posted separately, I'm happy to do so, this is the only issue containing "virtualenv" that I could find.

thanks!

lazka:
@mikofski I guess I've never used virtualenv under msys2, yeah, seems something is broken there.

I agree that switching to "--user" by default would improve the situation. Here is the current debian patch: https://anonscm.debian.org/cgit/python-modules/packages/python-pip.git/tree/debian/patches/set_user_default.patch

(looking into distutils I see we could even have different user directories per platform, but that would break existing user installs and not sure all packages can deal with it, seems uncommon)

KarimReefat:
> @lazka actually I am having issues with virtualenv (installed via either pip2 or pip3), it raises and `ImportError` exception, "no module called pkgutils" which _does_ exist in both python2 and python3 .
> 
[...]

PR #7911, opened 2021-02-07
python-pip: default to install --user by default

### LINKED ISSUE #6158: pip: default to user install
This has come up before, and I kinda think we should try to follow what debian does, default to user, allow override with a new "--system" option.

patch: https://salsa.debian.org/python-team/modules/python-pip/blob/master/debian/patches/set_user_default.patch

This makes sure pip doesn't mess with pacman installed files when installing or upgrading packages.

Newer pip now also does something similar in case the target isn't writable, so this behavior change is no longer that special, even with an unpatched pip, which should reduce the risk of it breaking unrelated use cases: https://github.com/pypa/pip/pull/7002

lazka:
This makes `pip install` not install things to pacman managed files by default.

This is what Debian uses.

Adds a new `--system` option which gives the old behavior.

Fixes #6158

### DIFF
--- a/mingw-w64-python-pip/PKGBUILD
+++ b/mingw-w64-python-pip/PKGBUILD
@@ -7,7 +7,7 @@ provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 pkgver=21.0
-pkgrel=1
+pkgrel=2
 pkgdesc="An easy_install replacement for installing pypi python packages (mingw-w64)"
 arch=('any')
 license=('MIT')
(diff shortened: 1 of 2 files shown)


lazka:
This will likely have some fallout in some CI systems since the user /bin is not in PATH by default.

mingwandroid:
Cool.

naveen521kk:
I would like it to be inside on `MINGW_PREFIX`. I don't want it to write some random folders on the computer. Thought?

kmilos:
I might be wrong, but isn't `~/` shared between i686 and x86_64 environments, how is that handled?

naveen521kk:
> I might be wrong, but isn't ~/ shared between i686 and x86_64 environments, how is that handled? Isn't there a danger of mixing binary Python extensions?

obviously... 

> As if we need ${MINGW_PREFIX}/local to be the equivalent of /usr/local...

should be a lot better idea imo...

kmilos:
I guess this is not only related to Python, e.g. `~/bin` is not unambiguous in general...

lazka:
I wonder how CPython solves this (32bit vs 64bit) on Windows for --user installs.

kmilos:
Weirdly enough, platform is not mentioned here either: https://www.python.org/dev/peps/pep-0370/

My Anaconda 64-bit CPython indeed points to just `C:\Users\kmilos\AppData\Roaming\Python`

naveen521kk:
> I wonder how CPython solves this (32bit vs 64bit) on Windows for --user installs.

Weird it uses the same folder...

naveen521kk:
I think we should at least make it not happen with our python at least. Also, open an issue upstream.

kmilos:
Btw, looks like this was discussed before here: https://github.com/msys2/MINGW-packages/issues/2745

naveen521kk:
Also on https://github.com/msys2/MINGW-packages/discussions/7619

naveen521kk:
something related https://bugs.python.org/issue43195

lazka:
Debian has dropped the patch in favor of an upstream change which makes pip default to "user" in case the system isn't writable i.e. you try to pip install not being root. That doesn't really help us because in the default setup the user has write rights to the system dir in MSYS2.

I don't really want to maintain this patch though, and handle the fallout of the behavior differences and CLI options if we are the only distro doing this...

So I'm going to close this for now.

PR #7916, opened 2021-02-08
glibmm: Update to 2.66.0, use meson.

Biswa96:


Biswa96:
Forgot this in previous batch. Are those install files required now?

lazka:
>  Are those install files required now?

I think you can remove them. The .pc files look good now.

Issue #7580, opened 2021-01-03
Wrong dependencies for python-pip


Paebbels:
Since when does python-pip depend on so many Python packages?

```python
depends=("${MINGW_PACKAGE_PREFIX}-python-appdirs
         "${MINGW_PACKAGE_PREFIX}-python-cachecontrol"
         "${MINGW_PACKAGE_PREFIX}-python-colorama"
         "${MINGW_PACKAGE_PREFIX}-python-contextlib2"
         "${MINGW_PACKAGE_PREFIX}-python-distlib"
         "${MINGW_PACKAGE_PREFIX}-python-html5lib"
         "${MINGW_PACKAGE_PREFIX}-python-lockfile"
         "${MINGW_PACKAGE_PREFIX}-python-msgpack"
         "${MINGW_PACKAGE_PREFIX}-python-packaging"
         "${MINGW_PACKAGE_PREFIX}-python-pep517"
         "${MINGW_PACKAGE_PREFIX}-python-progress"
         "${MINGW_PACKAGE_PREFIX}-python-pyparsing"
         "${MINGW_PACKAGE_PREFIX}-python-pytoml"
         "${MINGW_PACKAGE_PREFIX}-python-requests"
         "${MINGW_PACKAGE_PREFIX}-python-retrying"
         "${MINGW_PACKAGE_PREFIX}-python-six"
         "${MINGW_PACKAGE_PREFIX}-python-webencodings")
```
Source: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-pip/PKGBUILD#L15-L31

When I install pip in e.g. a Debian or a Windows, only `colorama` is installed. Not even `wheel` is installed, even when nowadays most packages are only delivered in wheel format.

Can you please explain, why a pip in MinGW needs so many dependencies it's not using at all?

-----------------------------
When checking the sources of [pip](https://github.com/pypa/pip), there is no requirement specified, not even `colorama`.

Biswa96:
How many of them do you find here https://github.com/archlinux/svntogit-packages/blob/packages/python-pip/trunk/PKGBUILD ?

Paebbels:
@Biswa96 Arch Linux is not a reference (maybe an indication). Only the source code of a package define what is needed and was is unneeded. So if the source code of pip (see link above) doesn't define dependencies, why is MinGW defining so many dependencies?

If you're saying MinGW copied dependencies from Arch Linux, then the questions are:
[...]

PR #7919, opened 2021-02-09
python-pip: update to 21.0.1

naveen521kk:
uses vendored dependencies rather than removing them. works locally.
https://github.com/pypa/pip/tree/21.0.1/src/pip/_vendor#vendoring-policy seems satisfying that we don't need to devendor it.
Also, remove `*.install` script and also remove shebang from script files.
/cc @lazka


lazka:
Thanks! I'll try to test it.

naveen521kk:
Fixes https://github.com/msys2/MINGW-packages/issues/7580

PR #7921, opened 2021-02-09
ffms2: Update to 2.40

Biswa96:


Biswa96:
Wait! Need to discuss something with mingw-w64 devs.

PR #7922, opened 2021-02-09
scipy: update to 1.6.0

naveen521kk:


naveen521kk:
@lazka can you cancel the CI? Didn't notice that while loop.

PR #7927, opened 2021-02-10
gtkmm3: rebuild against newer glibmm

lazka:


Biswa96:
glibmm may not require pkgrel++, it is now in pacman repo.

PR #7929, opened 2021-02-10
add mathml headers to install.  Update version

dsisnero:
mathematical needs mathml headers.  Add them to install. Also
update the version to the latest.


Biswa96:
Just asking. Have you tried to send the patches to upstream?

lazka:
I don't see any change in the resulting packages. What does this change/fix?

PR #7930, opened 2021-02-11
qt5-static: patch schannel

Optiligence:
since 5.15 it was chosen to build tls support via schannel
unfortunately this is broken in some cases
add a patch to mitigate that while keeping on using it
(from https://codereview.qt-project.org/c/qt/qtbase/+/333051)

sync up PKGBUILDs


Optiligence:
i thought bumping shared qt wasn’t necessary since it isn’t affected and pkgrels  aren’t in sync anyway
but fair enough

lazka:
ah, I didn't see the different build config. but yeah, making sure it still builds can't hurt.

PR #7931, opened 2021-02-11
Split package template

jeremyd2019:
This adds a template, `mingw-w64-splitpkg-wrappers`, to mingw-w64-PKGBUILD-templates, and converts a selection of small-ish packages to use it.  This was done by adding the line
```bash
# template input; name=mingw-w64-splitpkg-wrappers; version=1.0;
```
to each PKGBUILD file, and then running
```bash
git diff --name-only | while IFS= read -r line; do
  pushd `dirname $line`
  makepkg-template --template-dir ../mingw-w64-PKGBUILD-templates/
  popd
done
```

If this is an acceptable approach, I can do the remaining packages, split into CI-sized pull requests.


jeremyd2019:
huh, ocaml was downloading and building for me yesterday

lazka:
lgtm

PR #7933, opened 2021-02-11
Some updates

lazka:


Biswa96:
Are the install files required in glib package? Wasn't there a movement to remove mingw prefix in python shebang?

PR #7937, opened 2021-02-12
gtkmm3: backport fix for inkscape

lazka:


lazka:
@Ede123

Ede123:
Thanks!

Stupid me even tried patching the header manually yesterday, but I did not patch the source, but the installed file. 🤦

PR #7940, opened 2021-02-12
pkgconf: Remove leading slashes in --list-all output

RyanGlScott:
This fixes a `pkgconf`, observed in pkgconf/pkgconf#209, where the output of `pkgconf --list-all` would mistakenly print out package names with leading slashes, e.g.,

```
$ ./pkgconf.exe --list-all
/bzip2                         bzip2 - Lossless, block-sorting data compression
/expat                         expat - expat XML parser
...
```

This interferes with tools like `cabal-install`, which parse the output of `pkgconf --list-all` (or `pkg-config --list-all`) and assume that each line begins with an alphanumeric package name.

pkgconf/pkgconf#209 was fixed upstream in pkgconf/pkgconf#210, but it may be a while for another version of `pkgconf` is released (see https://github.com/pkgconf/pkgconf/pull/210#issuecomment-776663093). In the meantime, this incorporates the patch from pkgconf/pkgconf#210 into `MINGW-packages`.


lazka:
Thanks!

PR #7949, opened 2021-02-13
darktable: Update to 3.4.1

Biswa96:
Remove upstream patches darktable-libavif-0.8.2.patch and ninja-build-fixes.patch.


lazka:
> Enable strip binaries.

Isn't it doing that on purpose so it can write backtraces with drmingw?

Biswa96:
O, I didn't know that. guy blame does not show any message about it.

Biswa96:
O, I didn't knwo that. I'll revert that change.

PR #7963, opened 2021-02-14
cairo: fix issue with illegal font names in PDF output

Ede123:
See https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/125


Ede123:
(patch confirmed working downstream in Inkscape, where it resulted in unreadable PDF exports, see https://gitlab.com/inkscape/inbox/-/issues/4413)

Biswa96:
Need to add the link of that MR.

PR #7967, opened 2021-02-15
iverilog: update to 11.0.r8891.18e7406d

umarcor:
This should be just a regular update. However, I found that I was incorrectly using `12.0` as the base version. It should be `11.0`. Otherwise, when `12.0` is tagged, existing versions will be considered larger by `vercmp`.

Therefore, I decided to handle it here. What needs to be done so that this "downgrade" is acceptable?

### DIFF
--- a/mingw-w64-iverilog/PKGBUILD
+++ b/mingw-w64-iverilog/PKGBUILD
@@ -5,6 +5,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=11.0.r8891.18e7406d
 pkgrel=1
+epoch=1
 pkgdesc="Icarus Verilog, is a Verilog simulation and synthesis tool (mingw-w64)"
 arch=('any')
 url="http://iverilog.icarus.com/"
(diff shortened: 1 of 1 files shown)


lazka:
You can increase the "epoch", or leave it at 12 for now. A downgrade isn't possible.

umarcor:
Thanks! I increased the epoch.

mingw-w64-iverilog/PKGBUILD
```diff
@@ -3,8 +3,10 @@
 _realname=iverilog
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=12.0.r8793.96df23c5
-pkgrel=3
+pkgver=11.0.r8891.18e7406d
+pkgrel=1
+# Required until 12.0 is tagged and >r8793
```
lazka, line 8: the epoch will always be required if we start using it.

umarcor: I misunderstood. I now removed the comment, which didn't make sense. Thanks for being so explicative, once again :wink:

PR #7970, opened 2021-02-15
meson: update to 0.57.0

naveen521kk:


lazka:
There are quite a few regressions in 0.57.0. I think we should wait for 0.57.1

naveen521kk:
Then I will make a PR when it is released. 
Thanks.

PR #7972, opened 2021-02-15
Passing NO_AVX512=1 Instead of TARGET=CORE2 should allow AVX2 and AVX to be used by OpenBLAS yet forbid AVX512

aadler:
Prevents AVX512 without destroying other instruction sets.

I cannot test directly on my machine (stuck using GCC 8.3 for R reasons) but having built OpenBLAS from source for many years, I do not think this will be problematic and will save many advanced instructions set outside of AVX512

### DIFF
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -5,7 +5,7 @@ _realname=OpenBLAS
 pkgbase=mingw-w64-openblas
 pkgname="${MINGW_PACKAGE_PREFIX}-openblas"
 pkgver=0.3.13
-pkgrel=2
+pkgrel=3
 pkgdesc="An optimized BLAS library based on GotoBLAS2 1.13 BSD, providing optimized blas, lapack, and cblas (mingw-w64)"
 arch=('any')
 url="https://www.openblas.net/"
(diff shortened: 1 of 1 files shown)


mingw-w64-openblas/PKGBUILD
```diff
@@ -49,7 +49,7 @@ build() {
        CC=${MINGW_PREFIX}/bin/gcc                                      \
        FC=${MINGW_PREFIX}/bin/gfortran                                 \
        OPENBLAS_INCLUDE_DIR=${MINGW_PREFIX}/include/${_realname}       \
-       USE_THREAD=1 NUM_THREADS=64 TARGET=CORE2
+       USE_THREAD=1 NUM_THREADS=64 NO_AVX512 = 1
```
1480c1, line 52: would it have spaces around the `=`? the other options lack the spaces

aadler: Probably not. Need to change it.

1480c1:
Could you perhaps make the title a bit more descriptive such as marking the PR with OpenBLAS or some identifier to know what package it's related to?

jeremyd2019:
So what is the minimum supported processor after this change?

aadler:
@1480c1 Done
@jeremyd2019 As you pass DYNAMIC_ARCH, it should still go back all the way to CORE2, but it should not prevent AVX or AVX2 for SANDYBRIDGE and HASWELL/BROADWELL. At least to the best of my understanding. @xianyi or @wernsaar would know better.

Biswa96:
@podsvirov Does python-numpy work with this change? I have no issue with this.

lazka:
What's the point of this with DYNAMIC_ARCH being active anyway?

The docs say to use TARGET and DYNAMIC_ARCH  and nothing about NO_AVX512, so I'm wary.

mati865:
#7827 was about normal AVX instructions so reverting https://github.com/msys2/MINGW-packages/pull/7830 will bring it back.

lazka:
As this can't go in as is, and there hasn't been any further info -> closing.

If something changes we can reopen or create a  new PR though.

PR #7465, opened 2020-12-17
mingw-w64-arm-none-eabi-gcc: update gcc version and enable Ada support

Fabien-Chouteau:

PR #7976, opened 2021-02-16
arm-none-eabi-gcc: update to 10.1.0

umarcor:


lazka:
@fauxpark ok?

fauxpark:
Apparently there are still issues with LTO - trying to compile QMK causes this warning at linking stage:

```
Linking: .build/planck_rev6_default.elf                                                             [WARNINGS]
 | 
 | lto-wrapper: warning: Options to Xassembler do not match: -adhlns=.build/obj_planck_rev6_default/muse.lst, -adhlns=.build/obj_planck_rev6_default/keyboards/planck/planck.lst, dropping all -Xassembler and -Wa options.
```
and does not reduce file size as expected.

Additionally there is some problem with the change in behaviour of `-fcommon`? I think it was implied before but is not now. This may be something we should fix on our side, though.

Fabien-Chouteau:
There is already #7465 that updates GCC to `10.1.0` and enables Ada support.
The PR is stalled since two months ago because the build takes too long for GH actions runs.

umarcor:
Thanks @Fabien-Chouteau. I overlooked that. Closing this as a dup.

lazka:
> Additionally there is some problem with the change in behaviour of `-fcommon`? I think it was implied before but is not now. This may be something we should fix on our side, though.

yeah, that changed with gcc 10 to be more strict. Maybe fixing it will also help your LTO case.

PR #7982, opened 2021-02-18
live-media: Update to 2021.02.11

mingwandroid:


Biswa96:
So many patches 🤯

mingwandroid:
They all come from the VLC project. I can look into upstreaming them, but I assume the VLC project would do so if it were easy.

These patches are important. They add e.g. pkg-config support which VLC relies on to find livemedia555. With this done, we can support RTSP streams on MSYS2 (at least in VLC).

lazka:
This installs lots of files to `/usr/local/` (see the CI output under [diff])

mingwandroid:
> This installs lots of files to /usr/local/ (see the CI output under [diff])

Could this be made into a failure condition?

mingwandroid:
@lazka I think that's fixed now. Thanks.

mingwandroid:
Thanks!

PR #7986, opened 2021-02-18
Add poetry-core and python-build

naveen521kk:
https://github.com/msys2/MINGW-packages/pull/7563


Biswa96:
Is git required as make depends?

naveen521kk:
No, it isn't required let me remove

naveen521kk:
I have added isort to test whether this works.

PR #7991, opened 2021-02-19
[muparser] Build project using its own cmake file.

### LINKED ISSUE #7990: mupraser.pc file is not installed with mupraser.
the package doesn't contain the mentioned file.

ChinouneMehdi:
Fixes #7990


mingw-w64-muparser/PKGBUILD
```diff
@@ -36,7 +33,7 @@ build() {
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
     ../${_realname}-${pkgver}
 
-  make
+  make -j$(nproc)
```
mati865, line 36: Pacman does use `MAKEFLAGS` configured in `makepkg*.conf`, you should *not* specify it here.

PR #7996, opened 2021-02-20
github-cli: new package

naveen521kk:


lazka:
This installs completions to /usr/share instead of /mingw64/share, which will leads to conflicts.

naveen521kk:
> This installs completions to /usr/share instead of /mingw64/share, which will leads to conflicts.

That was on purpose. Having competitions in `/mingw64/share` doesn't work... also see
https://github.com/msys2/MINGW-packages/blob/1a14a5d8b54b386b2be12e2430b38b9b93e4c8a0/mingw-w64-hub/PKGBUILD#L33-L34

lazka:
ok, then we need to fix bash-completion. Maybe extending XDG_DATA_DIRS is enough: https://github.com/scop/bash-completion/blob/223762c744caffba1bcc8f94eec131062aa0c604/bash_completion#L2175

1480c1:
Perhaps maybe it might be better to make a separate msys2/ package for the bash-completion addon? It doesn't really make that much sense to me to have it in the mingw* ones since you don't have bash in the mingw* repos

naveen521kk:
what should I do then?

lazka:
>  It doesn't really make that much sense to me to have it in the mingw* ones since you don't have bash in the mingw* repos

we also don't have "man" as a mingw package and make man pages work for mingw packages.

> what should I do then?

install to the right prefix and we'll look into fixing bash-completion

naveen521kk:
@lazka done :)

PR #7997, opened 2021-02-20
CMake: Executables need to generate import libraries

bradh352:
It is not uncommon for executables to allow the use of plugins, and
plugins need to access symbols exported from the executable.  On windows,
things require explicit linking, therefore this import library is
necesary to be generated in order to be able to compile the plugins.

At one point in time, MinGW had some linking issues, and it appears
this patch was put in place to disable generation of import libraries
for executables due to this oversight of the requirement in relation
to plugins.

CMake's author's agree with this assessment as they did not accept
this merge request:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5194


bradh352:
@Alexpux @mingwandroid comments?

lazka:
I don't know enough about this, but we had some related fixes in binutils for crashes (like https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=9cdc5bacddc3776a455cc918e034b01dfb31c75b) , which might have been the reason this patch exists. So I'm open to giving this a try.

jeremyd2019:
I wish that cmake had some option to generate an import library for exe's, off by default.  In my experience, they're relatively rarely needed.  Anyway, the commit linked and a later one that actually fixed making an empty import lib for a dll, are included now, so this should be fine.

bradh352:
Yeah, I seem to recall the aforementioned issue with a crash when no symbols were being exported, but I think there was another hack/workaround for that we had put in place that worked (though I can't recall what it was).  Seems like that issue is long gone though, it wasn't until I started switching to MSYS's MinGW and CMake that I ran into this issue that caused some blockers for me, which this PR definitely fixes...

PR #8007, opened 2021-02-23
lfcbase + cego : new versions

bjlemke:


MehdiChinoune:
@bjlemke the packages you were added weren't updated for a long time. they start now failing to build with latest compilers and package. I think they will be dropped.

lazka:
newer versions of lfcbase also don't build with lots of errors

PR #8025, opened 2021-02-27
gr: Use Ninja, fix version generation.

Biswa96:
Manually generate version number otherwise CMake fails for not being git repository.

### DIFF
--- a/mingw-w64-gr/PKGBUILD
+++ b/mingw-w64-gr/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gr
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.55.0
-pkgrel=1
+pkgrel=2
 pkgdesc="A graphics library for visualisation applications (mingw-w64)"
 arch=("any")
 url="https://gr-framework.org/"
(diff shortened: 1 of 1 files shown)


Biswa96:
1. Use Ninja for faster build.
2. Create `version.txt` using `pkgver` variable. Otherwise, CMake shows this warning and pkgconfig files have zero version number.

```
CMake Warning at cmake/GetVersionFromGit.cmake:29 (message):
    Could not read version from git or version.txt
    CMakeLists.txt:5 (get_version_from_git)
```

3. Also remove the `-Werror` which shows warnings like these:

```
cc1plus.exe: warning: '-Werror=' argument '-Werror=implicit' is not valid for C++
```


/cc @kojix2 @kou Any opinions?

kojix2:
@Biswa96
Awesome. 
I can do simple things like updating the version number or updating the sha 256, but honestly, I don't know much about difficult things. I think @kou will take a look. It's good that the time required to build is reduced. 
Thanks anyway!

mingw-w64-gr/PKGBUILD
```diff
@@ -20,39 +20,51 @@ depends=("${MINGW_PACKAGE_PREFIX}-bzip2"
...
-  cd "${srcdir}/"
   rm -rf "${_realname}-${pkgver}"
   tar -xf "${_realname}-${pkgver}.tar.gz" || true
+  cd "${srcdir}/${_realname}-${pkgver}"
+
+  # Set version manually, otherwise it resets to zero
+  echo "${pkgver}" > version.txt
+
+  # Remove invalid C++ compiler option
+  sed -e "s|-Werror=implicit||g" -i CMakeLists.txt
```
kou, line 41: Could you report this to upstream?
https://github.com/sciapp/gr/issues

kou (approved):
+1

kojix2:
Biswa96 has the right to report.
However, I think Biswa96 is just volunteering to help us this time, so I don't want to bother him.
I've posted a link to an existing issue on sciapp/gr. 
But you can still report it as a new issue.

Biswa96:
@lazka ready to merge.

mingw-w64-gr/PKGBUILD
FlorianRhiem: The compiler flag support detection should be [fixed](https://github.com/sciapp/gr/commit/a231e58df9b5da2092568c5a2d87b3a3e3ddd08a) now, so that CMakeLists.txt won't assume that all GCC and Clang versions support this flag.

PR #8026, opened 2021-02-27
python-language-server: New package

podsvirov:
Note: some tests skipped and not all checks passed.


podsvirov:
I managed to build and install the package. The `pyls` command works, but I'm not an advanced user.
I wanna use this with a [qt-creator](https://packages.msys2.org/base/mingw-w64-qt-creator) to develop python-based projects.
Can anyone check this integration? Or integration with other editors?

Biswa96:
AFAIK, the .install files are not required anymore. mingw-w64-python is picked up from simple python.exe in shebang.

PR #8027, opened 2021-02-27
soqt: Fix SoQt.pc

podsvirov:
The missing -lSoQt part will be added to Libs section.
The missing -DSOQT_DLL part will be added to Cflags section.

### DIFF
--- a/mingw-w64-soqt/PKGBUILD
+++ b/mingw-w64-soqt/PKGBUILD
@@ -4,7 +4,7 @@ _realname=soqt
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.6.0
-pkgrel=2
+pkgrel=3
 pkgdesc="A Qt Gui-toolkit binding for Coin (mingw-w64)"
 arch=('any')
 url="https://github.com/coin3d/soqt"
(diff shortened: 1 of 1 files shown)


Biswa96:
Shouldn't this be done from upstream or by patching the .pc.in file?

podsvirov:
These issues have already been resolved in the master branch. I think that the fixes are superfluous - the additional configuration is enough. All this will not be needed in the new version.

Biswa96:
Just asking. How about adding that upstream change as a patch file? It would make easy to track in future.

podsvirov:
This is the correct thought. But in this case, I don't know where this patch is located. If you don't mind, I'll leave this question until the next release of SoQt.

podsvirov:
In pair with #8029 we can build [test-code/components](https://github.com/coin3d/soqt/tree/master/test-code/components) by simple `Makefile`:
```make
TARGETS := $(filter-out coloreditor,$(patsubst %.cpp,%,$(wildcard *.cpp)))

CXXFLAGS := $(shell pkg-config SoQt --cflags)
LDFLAGS := $(shell pkg-config SoQt --libs)

.PHONY: all
all: ${TARGETS}

%: %.cpp
        ${CXX} ${CXXFLAGS} -o $@ $< ${LDFLAGS}

.PHONY: clean
clean:
        ${RM} ${TARGETS}
```

PR #8029, opened 2021-02-27
coin: Fix Coin.pc

podsvirov:
The missing -DCOIN_DLL part will be added to Cflags section.


podsvirov:
@lazka thanks for quick reaction :-)

Issue #8030, opened 2021-02-27
[mingw-w64-vulkan-validation-layers] Failed to find 'vkGetInstanceProcAddr' in layer


Amphaal:
I've been trying to use this package instead of the LunarG's MSVC binaries package, and I've had 2 issues with it so far : 

- The generated `VkLayer_khronos_validation.json` should be copied in the `bin` directory, along its associated `.dll` file (like in the MSVC package) and not in the `lib` directory.
- Trying to use `VkLayer_khronos_validation` in a Vulkan app has no effect, and outputs : 
```
loader_create_instance_chain: Failed to find 'vkGetInstanceProcAddr' in layer C:\msys64\mingw64\bin\.\libVkLayer_khronos_validation.dll
loader_create_device_chain: Failed to find 'vkGetInstanceProcAddr' in layer C:\msys64\mingw64\bin\.\libVkLayer_khronos_validation.dll.  Skipping layer.
```

Has anyone encountered this problem yet ?

lazka:
We have two patches for this package: one to change symbol exports and one to fix the install location, so likely both things need further improvements.

(and you might be the first to use it)

Amphaal:
I wish I could help on tweaking the script here, but I think I am not seasoned enough with mingw intrinsic mechanisms to be of any help with dynamic linkage issues :( 

This could be related to symbols beings exported to the habitual sister `dll.a` file, and not directly included into the `.dll` file, as the Vulkan loader might expect it ?

I'll have a look on fixing the `install()` destination location for the `.json` manifests.

Biswa96:
`vkGetInstanceProcAddr` seems to be exported from libvulkan-1.dll. Though I am curious about if the vulkan package has that issue why the dependent package(s) can be built without any issue.

Amphaal:
I ran both binaries into `dllexp.exe`, and here are the outputs [Exported Functions Lists.zip](https://github.com/msys2/MINGW-packages/files/6062090/Exported.Functions.Lists.zip)

Well, it looks like the expected functions layout from MSVC binaries differs radically from what is generated for MinGW. I wonder where the heck that comes from !

Biswa96:
Can you provide a sample program to reproduce your issue?

Amphaal:
[...]

PR #8041, opened 2021-03-01
[mingw-w64-vulkan-validation-layers] Fixes validation layers manifests install folder + shared library exposed functions

Amphaal:
Solves #8030 in its entirety.


lazka:
ok, so this fixes the exports for 64bit, but breaks the symbol naming for 32bit because it removes the stdcall name mangling. So a step in the right direction, but not ideal. I'll see if I can figure this out.

lazka:
Turns out VK_LAYER_EXPORT was also exported in vulka-headers, making our patch not work as intended... (maybe that got added later). Patched there too now.

lazka:
It exports correctly now, but more than on Linux since they use `-Wl,--version-script` there to filter out symbols and we don't have that. Hopefully not that big of a problem. If not we can have another look.

Amphaal:
Many thanks for your work !

PR #8044, opened 2021-03-01
x11docker: add 6.7.0

umarcor:
This PR adds [x11docker](https://github.com/mviereck/x11docker), a utility for running GUI applications in Docker containers. It allows starting X servers and setting container options for using them automatically. For instance, `x11docker --runx --no-auth ghdl/ext gtkwave` or `x11docker --runx --no-auth aptman/dbhi:bionic-octave octave`. It requires Docker Desktop and VcxSrv or Cygwin/X or xpra.

### DIFF
--- a/mingw-w64-x11docker/PKGBUILD
+++ b/mingw-w64-x11docker/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: umarcor <unai.martinezcorral@ehu.eus>
+
+_realname=x11docker
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=6.7.0
+pkgrel=1
+pkgdesc="Run GUI applications and desktops in docker (mingw-w64)"
+arch=('any')
+url="https://github.com/mviereck/x11docker"
+license=("MIT")
+
+source=(
+  "${_realname}.tgz::https://codeload.github.com/mviereck/${_realname}/tar.gz/v${pkgver}"
+  "runx::git://github.com/mviereck/runx.git#commit=908ee52d"
+)
+sha256sums=(
+  'fd41bfb86bd4db20dc5de5132a03caf426db6f940475b933126af926e5532995'
+  'SKIP'
+)
+
+package() {
+  _bin="${pkgdir}${MINGW_PREFIX}/bin"
+  mkdir -p "${_bin}"
+
+  _x11docker="${srcdir}/${_realname}-${pkgver}"
+  install -m 755 "${_x11docker}"/x11docker "${_bin}"
+  install -m 755 "${_x11docker}"/x11docker-gui "${_bin}"
+
+  install -m 755 "${srcdir}"/runx/runx "${_bin}"
+
+  _licenses="${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}"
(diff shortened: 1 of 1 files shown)


Biswa96:
Are not those just some shell scripts? Do we need a package for it?

1480c1:
I'm voting -1 for this one since it requires a package/program that we do not provide, yet. If we could package something like vcxsrv or Cygwin's X11 or one of the other ones in a way that it's available inside the mingw{64,32} repository, then it's a maybe, but since installing this package by itself does not allow it to be useful, I don't think it's worth packaging atm.

umarcor:
@Biswa96 many Python packages are "just some scripts". The criteria for packaging is the installation complexity and the effort by users for keeping software up to date. These are two different scripts, and one of them depends on the other being available in the PATH. At the same time, versions need to be picked carefully, because runx is not submoduled in x11docker. Therefore, I believe it makes sense to package it.

@1480c1, per reductio ad absurdum, since we don't provide Docker, it doesn't make sense to install any tool for helping with that. I beg to disagree. Docker Desktop is usable on MSYS2 already. It is sensible not to force users to duplicate their setup through either Cygwin or WSL2, unless they want to do so. Yet, adding this package is the first step for justifying the addition of `xauth` from Xorg. See https://github.com/mviereck/x11docker/issues/325#issuecomment-782898930. Ideally, MSYS2 would provide a useful X server as a package, so that Cygwin or VcxSrv are not required. However, that's rather disruptive because it is non-trivial and MSYS2 per se does not require it, due to MSYS2 tools using native GUI toolkits. This package is useful as-is for users of MSYS2 and Docker which require running GUI tools inside containers.

1480c1:
> yet. If we could package something like vcxsrv or Cygwin's X11 or one of the other ones in a way that it's available inside the mingw{64,32} repository, then it's a maybe

> Yet, adding this package is the first step for justifying the addition of xauth from Xorg. See mviereck/x11docker#325 (comment). Ideally, MSYS2 would provide a useful X server as a package, so that Cygwin or VcxSrv are not required. However, that's rather disruptive because it is non-trivial and MSYS2 per se does not require it, due to MSYS2 tools using native GUI toolkits. This package is useful as-is for users of MSYS2 and Docker which require running GUI tools inside containers.

ideally, we would package an x server and other tools required for a proper x11 session before packaging a tool that requires an x server and other tools. This PR as it is is buying a GPU before you have the rest of the system, you are missing a lot of integral components that would make it useful at the moment.


> per reductio ad absurdum, since we don't provide Docker, it doesn't make sense to install any tool for helping with that. I beg to disagree. Docker Desktop is usable on MSYS2 already. It is sensible not to force users to duplicate their setup through either Cygwin or WSL2, unless they want to do so

this is sort of irrelevant to my argument as I'm mainly pointing out the x11 part that is missing, docker I understand is almost never packaged on windows than their prebuilt binaries etc. Although in a similar vein with the x server, it may be possible if someone tries to package https://github.com/docker/engine and/or https://github.com/docker/cli

lazka:
Hm, my 2c, since this doesn't depend on any mingw packages it should be a msys package. I can't really tell the usefulness re our goal to only allow tools that help us developing mingw packages. But since these are just some scripts I'd be OK with making an exception..

lazka:
-> msys2-packages

Issue #7650, opened 2021-01-11
CTRL+C doesn't kill Python Executables(aka pip)


naveen521kk:
Just tried with [this](https://github.com/naveen521kk/msys2-updator/blob/main/requirements.txt)
```sh
pip install -r requirements.txt
```
pressed CTRL+C expecting the **oh no!** thing I was doing(forgot to add `--user`). What happened was that the program didn't terminate. Bash returned to prompt but `pip` was adding outputs over that. Finally, I ruined my msys2-MinGW environment.
Overwrote all the Pacman install libraries and I am here don't know what to do. lazka told me he would help on discord...

Now the problem is why doesn't pip terminate on pressing CTRL+C?  I am sure that I installed pip through Pacman didn't mess it up with anything else. I have seen this happen in other setuptools produced launcher. So, I suspect it is something wrong with `python-setuptools`.

naveen521kk:
https://github.com/msys2/MSYS2-packages/pull/2547 has fixed this

PR #8048, opened 2021-03-02
python-pip: use winpty

naveen521kk:
Fixes https://github.com/msys2/MINGW-packages/issues/7650

A temporary workaround that makes pip use winpty which handle keyboard interrupts. This should fix the child process not terminating error as described in https://github.com/msys2/MINGW-packages/issues/7650 also discussed in [discord](https://discord.com/channels/792780131906617355/794220101741838401/815593216329515050).


lazka:
hm, not sure if this is worth it. I'll think about it.

PR #8055, opened 2021-03-03
[libgme] Add static library

drfiemost:
Build and install the static gme library too.
Also added zlib dependency.


Biswa96:
Should not that static library be automatically install with `make install` command instead of explicit `install` command?

drfiemost:
> Should not that static library be automatically install with `make install` command instead of explicit `install` command?

But this would overwrite the already installed files . Maybe there's a make target to install only the library? Will check...

Biswa96:
Technically yes but there is a trick. Repeat after me "shared after static" :) If `make install` command is run from shared folder after static folder it'll do the trick. See my very previous faudio pull request or any PKGBUILD file which builds both shared and static libraries separately.

drfiemost:
Ok, will rework the change, thanks. Should I switch to Ninja while at it?

Biswa96:
As you wish.

PR #8057, opened 2021-03-03
Add libopenmpt package 

drfiemost:
A mod player library, evolved from libmodplug and actively maintained.
I've included the command line frontend which adds extra dependency on portaudio, libsndfile and flac, can be removed if undesired.


Biswa96:
There seems to be a newer version 0.5.5. Can this be updated to latest version?

drfiemost:
I've tried compiling my project against 0.5 and encountered some issues so I feel hesitant about upgrading, considering also that both versions are still maintained. Unfortunately I had no time to dig further into the problems yet but it's on my TODO list.

PR #8063, opened 2021-03-04
Add libcue static lib

drfiemost:
Build both static and shared libraries
Switch to Ninja build
Build for release
Install license file


Biswa96:
No, the LICENSE should be in license folder. The README should be in doc folder.

drfiemost:
Got it, fixed!

PR #8075, opened 2021-03-06
add setuptools_scm dependency

naveen521kk:


naveen521kk:
It looks like I should add all the package all together like #8071 will fix everything which fails in this PR.

PR #8080, opened 2021-03-07
Add more setuptools_scm dependency

naveen521kk:
Fixes the packages which failed to build in autobuild

on a side note, it seems weird that this CI over here doesn't check for these. `setuptools_scm` was installed in one package and is reused everywhere in this CI while not in the autobuild which seems to be weird, and needs to be fixed.


lazka:
yeah, it's on my todo list. thanks for fixing!

PR #8081, opened 2021-03-07
gcc: backport some more fixes

### LINKED ISSUE #8079: gcc bug ?!?
building binutils after the last gcc patch level now yields this ->
../../binutils-2.36.1/opcodes/i386-dis.c:12308:1: internal compiler error: in seh_cfa_adjust_cfa, at config/i386/winnt.c:1104
12308 | }

tried earlier versions like 2.35 and 2.30 and same error.

lazka:
Fixes #8079


revelator:
AH good :)

PR #8083, opened 2021-03-07
CI: read mingw_arch array to decide which builds the PKGBUILD supports

lazka:


lazka:
Blocked on #8081

Biswa96:
The new keyword should be in template pkgbuild and/or in docs. Though I have some mixed feeling about not using [standard pkgbuild keywords][1].

[1]: https://wiki.archlinux.org/index.php/Arch_package_guidelines

lazka:
yeah, once it works. Related: https://github.com/msys2/MSYS2-packages/pull/2363

Biswa96:
Will `arch=('x86_64')` work instead of `arch=('any')`?

lazka:
It would limit the package to 64bit msys envs, but that doesn't change anything since that's all we have anyway right now.

PR #8070, opened 2021-03-06
aws-sdk-cpp: Update to 1.8.149, separate dependencies.

Biswa96:

PR #8085, opened 2021-03-07
New package: TileDB

jeroen:
@Biswa96 Here is a new package which also depends on aws-sdk-cpp, so we can test if #8070 actually works. You can merge this after or together with #8070. 
cc @ihnorton

### DIFF
--- a/mingw-w64-tiledb/PKGBUILD
+++ b/mingw-w64-tiledb/PKGBUILD
@@ -28,6 +28,12 @@ prepare() {
 }
 
 build() {
+  # Enable SSE2 (required) on i686 (these are enabled by default on x64)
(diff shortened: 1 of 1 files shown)


Biswa96:
The install step has gone in horrible way 😨 Also why are every AWS related package huge in size that it requires +4GB memory to build?

Biswa96:
The binaries are installed in wrong path. Also this requires spdlog and catch package as dependency otherwise CMake download those externally.

jeroen:
> The binaries are installed in wrong path.

Good catch. That should be fixed.

I have to go now, feel free to take it from here or make changes if you like.

Biswa96:
OK, I'll add the dependencies tomorrow. Would you like to squash the commits into one?

PR #8090, opened 2021-03-08
[mingw-w64-tclx] Remove command that moves tclx86.dll to bin and add --tcl-prefix to configure command line

mayd:


Biswa96:
The DLLs should be in bin folder according to the file system hierarchy in mingw.

mayd:
Please refer to TCl developer documentation for Tcl extensions and packages (TEA).
DLLs in Tcl package CANNOT be simply moved around.

PR #8091, opened 2021-03-08
tiledb: Fix dependencies, Use Ninja.

Biswa96:


Biswa96:
@jeroen Your opinion please.

jeroen:
I had to use `export MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX="` to fix the install dir, is that no longer needed when using ninja?

Biswa96:
Seems not to. Maybe it's make vs ninja thing.

Biswa96:
The build has complete. Now you can check the difference with previous commit.

jeroen:
Why does the static library increase from 27 to 45mb? Does ninja include debug symbols?

Biswa96:
No, debug is disabled. See the negation `options` list. AFAIK, the static library is not stripped.

mingw-w64-tiledb/PKGBUILD
```diff
@@ -12,69 +14,87 @@ depends=("${MINGW_PACKAGE_PREFIX}-aws-sdk-cpp"
...
-  export MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX="
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
   ${MINGW_PREFIX}/bin/cmake.exe \
-    -G"MSYS Makefiles" \
+    -Wno-dev \
+    -G "Ninja" \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
-    -DCMAKE_BUILD_TYPE=Release \
+    "${extra_config[@]}" \
     -DTILEDB_STATIC=ON \
```
lazka, line 78: TILEDB_STATIC is needed for the shared build as well?

jeroen: Ah probably not... (I'm only using static builds myself so I didn't notice).

PR #8092, opened 2021-03-08
libavif: update to 0.9.0

kmilos:


Biswa96:
Would you like to update the dav1d also?

PR #8101, opened 2021-03-09
Update 0, 013-further-fixes.patch, and PKGBUILD

revelator:
we can now build the gdiplus plugin with mingw-w64.
swig plugins now work for go and python.
gvedit can also be built and i corrected a small oversight where it was set to look for a non existing attributes.txt which is now named attrs.txt.


revelator:
wrong checksum on one patch and not even one of my own Oo, pushed the corrected PKGBUILD.

PR #8104, opened 2021-03-10
simplified a few things while keeping compatibility with the nix code.

revelator:
old style linking to python can be enabled now but cannot handle multiple versions of python, links to the first one found which in case it is enabled is python3, use the swig loader instead :).

loads of old code left in this that might have been kept as a compatibility fallback for old or unusual compilers, leave it as is for now but might be time to clean up this mess...

ruby plugin code needs some work cannot find ruby.pc in pkgconfig and is not looking for ruby$version.pc.

lua plugin has problems with the two versions we sport it does find it but tries to link in the wrong version groan.

Signed-off-by: revelator <ralphengels@gmail.com>


mingwandroid:
@revelator many thanks for making the PRs! Super stuff :-)

PR #8108, opened 2021-03-10
New package glslviewer

naveen521kk:
I was reading https://thebookofshaders.com/ where there was a mention of [glslViewer](https://github.com/patriciogonzalezvivo/glslViewer), sadly I couldn't find windows build there. so build it, will try and push the patches upstream later :)


Biswa96:
Though I didn't face the issue, the libraries name should be in lower case.

naveen521kk:
> the libraries name should be in lower case.

I don't think it installs any library?  Can you elaborate?

mingw-w64-glslviewer/0001-add-mingw.patch
```diff
@@ -0,0 +1,42 @@
...
++INCLUDES +=	-Isrc/ -Iinclude/ -Iinclude/oscpack
+ CFLAGS += -Wall -O3 -std=c++11 -fpermissive
+ 
+ ifeq ($(DRIVER),vc)
+@@ -78,6 +78,9 @@
+ 
+ endif
+ 
++CFLAGS += -DPLATFORM_WINDOWS -D_USE_MATH_DEFINES -DGLFW_INCLUDE_GLCOREARB -DDRIVER_GLFW $(shell pkg-config --cflags glfw3 glew) 
++LDFLAGS += $(shell pkg-config --libs glfw3 glew) -lpthread -lopengl32 -lWs2_32 -lWinmm
```
Biswa96, line 29: This would be useful if compiled from case sensitive fs:

```suggestion
+LDFLAGS += $(shell pkg-config --libs glfw3 glew) -lpthread -lopengl32 -lws2_32 -lwinmm
```

naveen521kk: it works atm right? any way thanks.

PR #8113, opened 2021-03-12
Update pango python-faker python-pep517

naveen521kk:
- pango: update to 1.48.3
- python-faker: update to 6.6.0
- python-pep517: update to 0.10.0 

I added pango to that automatic updator and it seems to work :)


naveen521kk:
Also, added `gi-docgen` and it's dependency `typogrify`

naveen521kk:
I will squash this in another PR because this is prone to force pushes.

PR #8122, opened 2021-03-13
mingw-w64-mesa: Update to 21.0.0 and add zink driver

pal1000:


mingw-w64-mesa/PKGBUILD
```diff
@@ -86,12 +90,12 @@ build() {
   -Dllvm=enabled
   -Dshared-llvm=enabled
   -Dosmesa=gallium
```
Biswa96, line 92: Try with this:

```suggestion
  -Dosmesa=true
```

Issue #8099, opened 2021-03-09
[arm-none-eabi-gcc] gcc always exists with exit code 1 and produces no object file


jgressma:
1. Install the package using pacman -S mingw64/mingw-w64-x86_64-arm-none-eabi-gcc
2. touch foo.c
2. Compile file i.e. arm-none-eabi-gcc -c foo.c -o foo.o  

Installed version was 
mingw64/mingw-w64-x86_64-arm-none-eabi-gcc 10.1.0-1 (mingw-w64-x86_64-arm-none-eabi-toolchain) 

The previous version works fine.
https://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-arm-none-eabi-gcc-8.4.0-2-any.pkg.tar.zst

Amphaal:
![image](https://user-images.githubusercontent.com/13903151/110622140-5b809700-819b-11eb-99e9-ec9b364135db.png)

No issues on my side :(

jgressma:
Tjhese steps reproduce the issue on my system:

1. Install MSYS2 from https://www.msys2.org/
2. Start the 64 bit MinGW terminal
3. Update
```
pacman -Syu
```

4. Install ARM EABI toolchain
```
pacman -S git make python3 mingw-w64-x86_64-arm-none-eabi-toolchain
```

5. Compile an example program
```
cat <<EOF >hello.c
#include <stdio.h>

int main(void)
{
    printf("Hello, world!\n");
    return 0;
}
EOF

arm-none-eabi-gcc -c -o hello.o hello.c
```

Amphaal:
I can confirm than a clean and simple install of said toolchain results in no output, BUT installing `mingw-w64-x86_64-cmake` fixes your issue...

I checked linked dependencies to the `arm-none-eabi-gcc`; thoses are quite simple and already installed with full toolchain.

![image](https://user-images.githubusercontent.com/13903151/111028721-a77c4780-83f8-11eb-9c47-829715cda232.png)

So I think that dynamic loading or library checking happens in said wrapper.

My guess is that an hidden dependency of `mingw-w64-x86_64-arm-none-eabi-gcc` lies in thoses of `mingw-w64-x86_64-cmake`, eg :

```
( 1/20) installing mingw-w64-x86_64-pkg-config     [#####################] 100%
( 2/20) installing mingw-w64-x86_64-c-ares         [#####################] 100%
( 3/20) installing mingw-w64-x86_64-brotli         [#####################] 100%
[...]

PR #8128, opened 2021-03-15
[arm-none-eabi-gcc] Add zstd as required dependency

Amphaal:
...Although it is only required at runtime. See #8099. Once again, maybe a too-straightforward patch, might need further investigation on the cause of said dependency.


lazka:
Thanks. I'd prefer we we figure out what needs it realls. https://packages.msys2.org/package/mingw-w64-x86_64-ntldd-git?repo=mingw64 can be helpful here.

lazka:
ah, arm-none-eabi-lto-dump.exe links against it. thanks!

PR #8130, opened 2021-03-15
Update GTK to 4.1.2

taozuhong:


mingwandroid:
Appropriate for https://github.com/msys2/MINGW-packages-dev perhaps?

Biswa96:
Which lunatic decides the version of GTK? The version 4.1.2 literally just belongs to gtk4.

lazka:
odd minor releases are unstable for all GNOME libraries

mingwandroid:
odd-sem-ver I like to call it.

taozuhong:
GTK 4 is a big change version and needs more testing, release these packages for testing.

mingwandroid:
If we release packages that need testing the we risk breaking a lot of people though, we'd probably love some testing repos? but we are not there yet.

taozuhong:
I  respect your thoughts, 

but I should point out: there is no app that supports GTK 4 now, just for testing only.

lazka:
OK, fair enough. Let's allow it for 4.1.x only to get more testing.

PR #8137, opened 2021-03-16
igraph: Update to 0.9.0

jannick0:
* PKGBUILD
  - upgrade to 0.9.0
  - move mandatorily from build system autoconf to cmake
  - use internal libraries to reduce package dependencies
  - remove unneccessary make dependencies 'bison' and 'flex'
  - fix link bug by removing explicit link against libm (using patch)
* 0001-igraph-0.9.0-link-libm-only-when-needed.patch
  - patch checking if explicit link against libm is needed fixing link error

### DIFF
--- a/mingw-w64-igraph/0001-igraph-0.9.0-link-libm-only-when-needed.patch
+++ b/mingw-w64-igraph/0001-igraph-0.9.0-link-libm-only-when-needed.patch
@@ -1,0 +1,35 @@
+--- a/etc/cmake/dependencies.cmake	2021-03-16 09:36:45.888000200 +0100
++++ b/etc/cmake/dependencies.cmake	2021-03-16 09:38:50.702227000 +0100
+@@ -4,6 +4,8 @@
+ # for tests
+ include(FindThreads)
+
++include(CheckSymbolExists)
++
+ macro(find_dependencies)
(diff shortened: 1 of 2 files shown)


lazka:
This lost `libigraph-0.dll`. Is that on purpose?

Biswa96:
Need to add `-DBUILD_SHARED_LIBS=ON` with cmake command.

jannick0:
> This lost `libigraph-0.dll`. Is that on purpose?

@lazka Well, the `cmake` implementation and `CMakeLists.txt` do not allow simultaneous builds of shared and static libraries as of now, while `libtool` is able to do that out of the box. There is some deep on `CMakeLists.txt` work to be done upstream for this to happen.

Suggestion: Let's go with the static library for now and wait for upstream to provide a parallel build without any issues with the import library, header files etc.

lazka:
ok, normally we just build twice and install shared over the static build. But since this doesn't have any reverse deps, fair enough.

jannick0:
> ok, normally we just build twice and install shared over the static build. 

A bit of caution needed here I think: `cmake` might produce incompatible header and/or pc files for static and shared libraries which could cause trouble. So installing shared and static builds on top of one another might work, but only by chance, if I understand the current `cmake` mechanism correctly.

Issue #8141, opened 2021-03-17
adns old git path needs to be updated


revelator:
it is now https://github.com/gpg/adns.git

revelator:
also need to change the definition of struct pollfd in config.in.h or it will error out because it is allready defined in winsock2.h

Biswa96:
That [gnupg adns repository][1] was "retired" many years ago. Is it possible to use the [actual git repo][2] here? It does not support Win32 or mingw platform but can be modified?

[1]: https://git.gnupg.org/cgi-bin/gitweb.cgi
[2]: http://www.chiark.greenend.org.uk/ucgi/~ianmdlvl/git?p=adns.git

revelator:
possibly though it would need a lot of patching i reckon.

lazka:
Looks like it was dropped in 2016: https://github.com/gpg/gnupg/blob/a788f2e8306d80f7f3df34eb62ec7ce1a62d48e1/NEWS#L1251

So we can remove adns

lazka:
@revelator do you have any actual use for adns as it is now?

revelator:
if it is not used kill it with fire :)

PR #8151, opened 2021-03-18
gnupg: Remove unused adns dependency

lazka:


Biswa96:
Also remove the adns package?

lazka:
Depends on https://github.com/msys2/MINGW-packages/issues/8141#issuecomment-802186869

PR #8162, opened 2021-03-20
Update sip5 to sip 6.0.3 and Move sip to sip4

MehdiChinoune:
See:
https://archlinux.org/packages/?sort=&arch=x86_64&repo=Extra&q=sip&maintainer=&flagged=


lazka:
Do you think we still need sip4 at all?

MehdiChinoune:
> Do you think we still need sip4 at all?

Yes, Some do.
sip6 doesn't have a standalone tool to generate bindings.

See:
https://www.riverbankcomputing.com/news/SIP_v6.0.0_Released
> All features deprecated in SIP v5 have been removed.

lazka:
ok, thanks

PR #8183, opened 2021-03-23
rhash: adjust MINGW uname detection, and link import library instead of dll

jeremyd2019:
rhash's configure script checks for `MINGW32*` and `MINGW64*` from `uname -s`, handles them separately for a bit, then checks if it's either mingw32 or mingw64 (or several other things) to tell if it's win32.  This is the only check for either.  Treat all MINGW* as `MINGW32` instead, as that functions as a sort of overarching 'project name'. Fixes msys2/CLANG-packages#15.  See also https://github.com/msys2/msys2-runtime/pull/33#issuecomment-805168564.

Also, as long as I was making a PR for rhash, include patch to link against its implib instead of the DLL, as `lld` does not support linking against a DLL directly.  Fixes msys2/CLANG-packages#8.


lazka:
thanks!

PR #8185, opened 2021-03-24
Seems like daily Boto3 patches now...

Schala:


lazka:
(you don't have to feel obligated to keep them up to date all the time.. I don't mind though)

PR #8190, opened 2021-03-24
Some small ucrt fixes

lazka:


lazka:
ok, so this wasn't just a ucrt thing. libgomp DLL isn't created

PR #8193, opened 2021-03-25
[new-package] Qt6 6.0.2

MehdiChinoune:


Biswa96:
Wasn't there a discussion about making modular Qt packages?

MehdiChinoune:
> Wasn't there a discussion about making modular Qt packages?

Yes There was.
I didn't any efficient way to build all packages within the same PKGBUILD file.

I tried two ways:

1st - Build each package from its own archive (submodule files). but the problem is that each package depend on other so build qt-svg for example qt-base must be available through pacman. because it needs some files from qt-base only available after installation.

2nd - Build all packages from the single file, then package each component separately.
It builds, packages but fails to install them because there is a conflict between them which I can't figure out how to resolve.

I am open to any suggestion to resolve these issues.

Biswa96:
Using submodules tarball is a superior choice. There are some mingw-w64-Qt6 packages in AUR. I would like to test those.

MehdiChinoune:
> Using submodules tarball is a superior choice. There are some mingw-w64-Qt6 packages in AUR. I would like to test those.

Well, I hope that's not a blocker to this pr. You push your changex when you succeed.

Biswa96:
Can this option be added `-DFEATURE_system_harfbuzz=ON`? Otherwise the build uses bundled harfbuzz.

MehdiChinoune:
> Can this option be added `-DFEATURE_system_harfbuzz=ON`? Otherwise the build uses bundled harfbuzz.

No, cmake can't find it.

It's "-DQT_FEATURE_system_harfbuzz".

MehdiChinoune:
superseded by #8199

PR #8194, opened 2021-03-25
Update Rust to 1.51.0

Zapeth:


lazka:
@mati865 OK?

mati865:
FWIW `0002-fix-install-msys2.patch` is unnecessary since https://github.com/rust-lang/rust/pull/80797 and should be removed from both PKGBUILD and repository.

Otherwise LGTM, I assume you were able to build at least one crate with cargo?

lazka:
Thanks. I built librsvg and it worked fine.

PR #8201, opened 2021-03-25
emacs: update to 27.2

oscarfv:


oscarfv:
Are the checkers broken? They fail with

```
:: Starting full system upgrade...
error: failed to prepare transaction (could not find database)
```

lazka:
Should be fixed now.

Looks like the new release is signed with a different key: `E6C9029C363AD41D787A8EBB91C1262F01EB8D39`. That one isn't listed as a emacs signing key though: https://www.gnu.org/software/emacs/download.html

oscarfv:
Thank you for the heads-up.

The checks have passed now but auto-merge was previously disabled because `Head branch was pushed to by a user without write access`. Was this caused by the force-push, or auto-merge only works for people with write access?

lazka:
Found the key here https://savannah.gnu.org/users/eliz so looks ok.

PR #8207, opened 2021-03-26
New package: htslib

teepean:


teepean:
I excluded 32 bit builds as I don't think anyone use 32 bit version of bioinfomatic tools.

Biswa96:
Generally, we tried to build for both 32 and 64 bit unless the build failed. Yes, there are many 32 bit users and developers who still need the same support as 64 bit.

teepean:
Looks like 32 bit version cannot find recv() from libws2_32. 

Same issue as this person:

https://sourceforge.net/p/mingw-w64/discussion/723798/thread/a44f6350af/?limit=25

Biswa96:
If I add `__stdcall` to that recv function definition the test works. Weird!

teepean:
I have an old TDM-GCC-64 (gcc 5.1.0) installed and it gives the same error so the problem might be in the compiler.

valeriuo:
Can you rerun your tests on the latest HTSlib version (1.13)?

lazka:
Enabled for mingw64/ucrt64 only for now. mingw32 fails with some weird error, clang because of missing dllexports in some places (and possibly more after that).

Feel free to fix them if wanted.

PR #8210, opened 2021-03-27
Update pango python-faker python-importlib-metadata python-jaraco.functools python-nuitka

naveen521kk:


naveen521kk:
@lazka I am ignoring docs for Pango this time also. gi-docgen didn't have a release still.

PR #8212, opened 2021-03-27
mingw-w64-headers: default to win10 on ucrt

Astrum-polaris:
I noticed msys2 now has a ucrt repo, that's great news!

This PR changes the default to win10 when building mingw-w64-headers for ucrt,
based on the inspiration of recent install files fixup

### DIFF
--- a/mingw-w64-headers-git/PKGBUILD
+++ b/mingw-w64-headers-git/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgdesc="MinGW-w64 headers for Windows"
 pkgver=9.0.0.6128.07922837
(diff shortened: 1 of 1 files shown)


lazka:
hmm, in theory ucrt works fine on win7 too. I'm not sure...

Astrum-polaris:
In theory ucrt works on vista too.
Given the fact that you need to manually install ucrt on win7, I think it is reasonable that we target win10 on ucrt

lazka:
I don't seem much value in bumping this. Most projects that care already define their minimum version so this doesn't change anything there. There are also no new features enabled by this in mingw-w64 libraries.

If there is a package which gets new features with win10+ we can enable it there for ucrt builds explicitly.

Astrum-polaris:
Got it.
✨ 🍰 ✨

lazka:
If we get to a point where "core" packages start to require win10 we can revisit this though.

PR #8215, opened 2021-03-27
Update glib2 python-hypothesis

naveen521kk:


naveen521kk:
@lazka can you merge this?

lazka:
The dependency on libelf seems weird. Why is it needed?

naveen521kk:
> The dependency on libelf seems weird. Why is it needed?

It failed https://github.com/msys2/MINGW-packages/runs/2211095425#step:8:476, maybe if somebody doesn't use that feature that can be disabled.

lazka:
yes, please build with `-Dlibelf=disabled` and remove the dep

lazka:
Thanks!

PR #8221, opened 2021-03-28
Python deps for Wind Waker Randomizer

Schala:


1480c1:
I think pinned git packages are preferred in this repo, but I could be wrong

Schala:
Pinned?

Biswa96:
Git packages are prohibited/discouraged untill there is no alternative way. If you really want git packages, add the sepcific commit, see mingw headers, crt etc. as examples.

Schala:
Ahh alright. I mean I could put them in with the package I'm trying to commit, I guess.... pinned, that is

lazka:
I added `0.r` to the version so we can easily change the version schema if upstream decides to tag a release or something.

PR #8235, opened 2021-03-29
Fix libtommath, add libtomcrypt

Biswa96:
* libtommath: Fix pkgconfig file.
  * Use .tar.xz source tarball.
  * Use upstream PR as patch file to add -no-undefined option.
  * Add PREFIX variable with make.
* libtomcrypt: Add package for firebird/libfbclient.


Biswa96:
If required I can add the `mingw_arch` array.

PR #8241, opened 2021-03-30
[new-package] PyQt6 6.0.3

MehdiChinoune:


Biswa96:
The error is weird. Linker can't find `QFileInfo::_stat64()` but that does neither exist in source code nor in Qt6 libs or headers.

PR #8243, opened 2021-03-30
aom: update to 3.0.0

kmilos:


lazka:
Does this break ABI? https://aomedia.googlesource.com/aom/+/v3.0.0 isn't very clear .. hmm

kmilos:
I can't be 100% sure either...

libavif change after they updated their external reference seems only API related? https://github.com/AOMediaCodec/libavif/commits/master

Happy to let this one sit until a new libavif release though...

kmilos:
We can wait at least for the next libavif release that uses this new aom version and rebuild all then...

Do you think it's worth patching CMakeLists to [make the SOVERSION part of the DLL filename](https://gitlab.kitware.com/cmake/cmake/-/issues/21716)?

lazka:
> We can wait at least for the next libavif release that uses this new aom version and rebuild all then...

sure

> Do you think it's worth patching CMakeLists to [make the SOVERSION part of the DLL filename](https://gitlab.kitware.com/cmake/cmake/-/issues/21716)?

Ideally upstream would decide that, so everyone uses the same DLL name (at least for the same type of toolchain). Do you think it's worth asking upstream?

lazka:
rebuilds worked (except ucrt64, but that is WIP), so I'm merging this.

lazka:
Thanks!

kmilos:
Great, thanks!

PR #8246, opened 2021-03-31
Python stuff

Schala:


lazka:
How can wwrando be used without it being in site-packages and without any launcher?

Schala:
I honestly don't know how to symlink on MSYS2. I was going to commit what I have and then look into it today. I did this late at night.

naveen521kk:
> I honestly don't know how to symlink on MSYS2. I was going to commit what I have and then look into it today. I did this late at night.

Maybe write a think layer of setuptools script, or use distlib or do something like https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-glib2/pyscript2exe.py

Schala:
Hmmm... okay. I mean, I made a shell script, but ok

PR #8249, opened 2021-03-31
Update GTK4 to 4.2

taozuhong:


mingw-w64-gtk4/PKGBUILD
```diff
@@ -56,8 +56,12 @@ build() {
     "${extra_config[@]}" \
     -Dbuild-examples=false \
     -Dbuild-tests=false \
-    -Dman-pages=true \
+    -Dman-pages=false \
     -Dmacos-backend=false \
+    -Dx11-backend=false \
+    -Dwayland-backend=false \
+    -Dwin32-backend=true \
+    -Dvulkan=disabled \
```
ZachBacon, line 64: any reason the vulkan backend is disabled?

taozuhong: Just copy them from GTK's CI script.

ZachBacon: I think it's safe to add vulkan support since we do have the vulkan package and windows vulkan support iirc is fairly mature. But it's not my call.

taozuhong: Ok, done

lazka: I think vulkan was broken when I added the gtk CI script. But it should work now.

ZachBacon: I just did a local test with it enabled explicitly, I mean, it compiles, but I have yet to throw the actual environment switch on for GSK to really test it

taozuhong:
GTK 4 has a new doc tool: gi-doc, it builds docs from girs, so it could be disabled.

lazka:
yeah, we have it here: https://packages.msys2.org/base/mingw-w64-gi-docgen

feel free to make it work

naveen521kk:
I updated it and made it work with Pango on https://github.com/msys2/MINGW-packages/pull/8256

PR #8250, opened 2021-04-01
Update Qt6 to 6.0.3

MehdiChinoune:
Use submodule files.


MehdiChinoune:
@Biswa96 @lazka
Can you guide how to build these dependent packages with the same pull-request.

Biswa96:
Do we need to add `=${_qtver}` in depends?

MehdiChinoune:
> Do we need to add `=${_qtver}` in depends?

Yes, to assure it doesn't build against the previous version.

lazka:
> Do we need to add `=${_qtver}` in depends?

Removing that might fix it. We build them in the right order anyway.

lazka:
Or I can look into fixing our CI script

PR #8261, opened 2021-04-02
groonga: Update to 11.0.1

komainu8:


komainu8:
Tanks!

Biswa96:
BTW, do you guys support ARM architecture? There is a todo in this AUR https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=groonga

komainu8:
@Biswa96 Groonga has already supported the ARM architecture.
So, we think that Groonga will work if MSYS2 will support the ARM architecture in the near future.

PR #8264, opened 2021-04-02
python: update to 3.8.9

naveen521kk:


MehdiChinoune:
Why not 3.9.3 ?

mingwandroid:
3.9.3 will take more patch reworking I guess?

Biswa96:
AFAIK, Python 3.9.3 uses a WinAPI which is not present in Microsoft Windows 7. But it can be modified with a simple reverse patching (as per lazka).

mingw-w64-python/PKGBUILD
```diff
@@ -40,7 +40,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-tk"
          "${MINGW_PACKAGE_PREFIX}-zlib"
          "${MINGW_PACKAGE_PREFIX}-xz")
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-pkg-config")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-pkg-config" "autoconf-archive")
```
naveen521kk, line 43: no idea why I should add `autoconf-archive` here. Thoughts?

It fails without this https://github.com/msys2/MINGW-packages/runs/2256211128#step:8:645

Biswa96: Assume autoconf-archive as an external library of shell scripts. Probably the configure.ac uses some function from that.

naveen521kk: Ah, this commit https://github.com/python/cpython/commit/e516290976626cf8535b88a14b1b34e37f88a78a#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810

lazka:
and we would have to rebuild lots of packages

naveen521kk:
> and we would have to rebuild lots of packages

why? it is a minor update, right?

MehdiChinoune:
Python 3.9.0 released 6 months ago.

lazka:
I meant for 3.9

naveen521kk:
> Python 3.9.0 released 6 months ago.

yeah, that requires lots of work. Nearly every py package needs to be rebuild. Not just that, it seems upstream dropped support for Windows 7, which is :(.

> I meant for 3.9

:+1:

MehdiChinoune:
> yeah, that requires lots of work. Nearly every py package needs to be rebuild. Not just that, it seems upstream dropped support for Windows 7, which is :(.

The upstream dropped support for Windows 7 (12 years old OS) because Microsoft dropped it first. and MSYS2 should drop it too.
Why should msys2/mingw-w64 support an OS not supported by its owner. I don't see any logic about running an outdated OS and asking for latest software releases.

Biswa96:
> and we would have to rebuild lots of packages

With GHA, it wouldn't be hard as it is pumping the ucrt builds now :)

> why? it is a minor update, right?

The /lib/python folder will be changed.

lazka:
> With GHA, it wouldn't be hard as it is pumping the ucrt builds now :)

yes, definitely.

naveen521kk:
> The /lib/python folder will be changed.

I was saying about 3.8.8 -> 3.8.9 one. I wasn't referring to the 3.9 bumps.

> The upstream dropped support for Windows 7 (12 years old OS) because Microsoft dropped it first. and MSYS2 should drop it too.
> Why should msys2/mingw-w64 support an OS not supported by its owner. I don't see any logic about running an outdated OS and asking for latest software releases.

But still, Python didn't make 3.8.x series EOL right, it still supports it and works on Windows 7? I think doing this will make all the software that uses gtk or pygobject, unsupported on Windows 7 and things like Inkscape would not work on Windows 7. Also, it is fine to do this, but someone would have to work on reworking the patches and testing them out (which is difficult actually). If you would like to work go ahead :).

PR #8269, opened 2021-04-03
poppler: update to 21.04.0

naveen521kk:


mingw-w64-poppler/PKGBUILD
```diff
@@ -3,8 +3,8 @@
 _realname=poppler
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=20.12.1
-pkgrel=2
+pkgver=21.04.0
```
naveen521kk, line 6: should I bump pkgrel of all reverse deps?

naveen521kk: ig, I should

naveen521kk:
*has no idea why things fail to build*

naveen521kk:
*I think the update can be done later*

PR #8271, opened 2021-04-03
Add TexLive

naveen521kk:
Bring in LaTeX. My plan is to follow Archlinux. First, this PR will add `texlive-bin` package, which in itself isn't useful. Then `texlive-core` must be added, I have no idea why Archlinux uses `"https://sources.archlinux.org/other/texlive/$pkgname-$pkgver-src.zip"` https://github.com/archlinux/svntogit-packages/blob/packages/texlive-core/trunk/PKGBUILD#L32 as a source, need to look into that. Later every collection can be added as a new package.


lazka:
nice!

Biswa96:
Isn't it possible to build the programs here instead of using prebuilt binaries?

Biswa96:
Why is there -bin in the package name?

naveen521kk:
> Why is there -bin in the package name?

As I told I am following Archlinux here. They have the core binaries in `texlive-bin`.

naveen521kk:
I made https://github.com/msys2/msys2-texlive so that `texlive-core` can be built. I am slightly confused about the other setup's like configuration and how it is expected to be on Windows and I have contacted their [mailing list](https://tug.org/pipermail/tldistro/2021q2/000387.html) they could help me.

naveen521kk:
Looks like it's going to be *hard* to get things right, I don't know much about configuration, which we can copy from Arch if they had texlive 2021, maybe let's wait for that to happen.

mingw-w64-texlive-bin/PKGBUILD
```diff
@@ -0,0 +1,430 @@
...
+   cd Work
+   echo "--> Initial configuration..."
+   # we use temporary prefix to avoid messing the existing
+   # $pkgdir/usr/share/texmf tree
+   ../configure --prefix="${MINGW_PREFIX}" -C \
+     --build=${MINGW_CHOST} \
+     --host=${MINGW_CHOST} \
+     --target=${MINGW_CHOST} \
+     --disable-native-texlive-build \
+     --with-banner-add="/MSYS2 MINGW" \
```
naveen521kk, line 72: @lazka what do you think this banner should be?
It will be shown each time when launching latex.

lazka: no idea..

naveen521kk: like how we have something in gcc? can I copy the same?

lazka: sure

naveen521kk:
If the checks passed or it build sucessfully this should be ready for testing.

naveen521kk:
for ucrt build, Perl is missing no idea what to do.

naveen521kk:
Looks like this requires quite some Perl modules for texlive utilities to work. I found https://tug.org/svn/texlive/trunk/Master/source/tlperl/tlperl.README?view=markup that needs to be added first.

naveen521kk:
After adding those Perl modules this should be ready

naveen521kk:
This is pretty much ready, AFAIK. I will check for the CI run I triggered in https://github.com/msys2/MINGW-packages/pull/8271/commits/e719cc73345a0776f7a168c304ded4c43a55074f and then do a local test, finally squash some commits so that this can be merged.

naveen521kk:
Tested this PR locally, by downloading the artefacts and things are working, so time to clean.
TODO:
- [ ] Reduce output of the hooks
- [ ] Cleans things in PKGBUILD
- [ ] Squash some commits

naveen521kk:
@lazka this works locally, review and merge(please squash while merging.)

naveen521kk:
Thanks.

podsvirov:
@naveen521kk, thanks for working to make this possible!

PR #8279, opened 2021-04-04
add webm

hgdagon:
Master branch builds dynamic library. The latest release is static only. A lot of stuff copied over straight from arch (aur).


mingwandroid:
Would we not prefer shared here?  Can it be re enabled?

hgdagon:
It seems like, we can, but I'd prefer to do both in that case.

PR #8281, opened 2021-04-04
Emacs native compilation

WolfgangRamos:
I added a `PKGBUILD` for building Emacs `native-compilation` development branch. I am building it from the most recent commit on the `native-comp` branch available at the time I created the `PKGBUILD` file  (instead of building the branch head itself). I chose to do it that way in order to get reproducible build results.

I have almost no experience with writing `PKGBUILD` files. So any feedback is welcome. I created my `PKGBUILD` by adjusting the Emacs release version `PKGBUILD` file. Thus I hope it got most things right (and of course I already successfully built and started Emacs with native-compilation on my machine).

I would appreciate it if this recipe could make it into the official repository :)

### DIFF
--- a/mingw-w64-emacs-native-compilation/PKGBUILD
+++ b/mingw-w64-emacs-native-compilation/PKGBUILD
@@ -62,11 +62,17 @@ build() {
     --build="${MINGW_CHOST}" \
     --with-modules \
     --without-dbus \
+    --without-pop \
     --with-native-compilation \
     --without-compress-install
 
   # --without-compress-install is needed because we don't have gzip in
   # the mingw binaries.
+
+  # --without-pop is used to prevent emacs from installing a an auxiliary
+  # limited and insecure substitute 'movemail' program. This we be unnessecary
+  # when GNU mailutils become available for MINGW64 and MINGW32. Then 
+  # mailutils should be included in the depends list.
 
   make
 }
--- a/mingw-w64-emacs-native-compilation/PKGBUILD
+++ b/mingw-w64-emacs-native-compilation/PKGBUILD
@@ -19,7 +19,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-universal-ctags-git"
          "${MINGW_PACKAGE_PREFIX}-jansson"
          "${MINGW_PACKAGE_PREFIX}-gnutls"
          "${MINGW_PACKAGE_PREFIX}-libwinpthread"
-		 "${MINGW_PACKAGE_PREFIX}-libgccjit")
+         "${MINGW_PACKAGE_PREFIX}-libgccjit")
 optdepends=("${MINGW_PACKAGE_PREFIX}-giflib"
             "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo"
             "${MINGW_PACKAGE_PREFIX}-libpng"
@@ -37,7 +37,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "autoconf"
              "texinfo"
              "patch"
-			 "git"
+             "git"
              "${optdepends[@]}")
 provides=("${MINGW_PACKAGE_PREFIX}-emacs-native-compilation")
 conflicts=("${MINGW_PACKAGE_PREFIX}-emacs")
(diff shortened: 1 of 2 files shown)


WolfgangRamos:
Hi, why were the builds for `mingw32` and `mingsw64` cancelled?

Shall I remove `ucrt64` from the list of supported architectures? I think this isn't even supported by Emacs release version.

@lazka, can you help here?

lazka:
It was cancelled because it took longer than 6 hours. After that time github stops the jobs.

WolfgangRamos:
Hi @lazka, thank you for clarifying.

I have absolutely no idea, why the build should take that long. On my machine it was completed within approx. 1 hour. Are we running out of memory in the build environment?

I would like to retrigger the CI to see if this failure is reproducible, but I don't know how to do you do this. Perhaps you can do this?

lazka:
looking at the build log it seems it simply gets stuck after 30 minutes. Maybe it's waiting on some user input, do you remember something like this during the build?

WolfgangRamos:
Hi @lazka, there is no need for user input during the build process.

I made GitHub rerun the CI by creating PR #8292 for new branch (`emacs-native-compilation-2`) pointing to **the same commit** as this branch (`emacs-native-compilation`). The rerun had the number [4125](https://github.com/msys2/MINGW-packages/actions/runs/720609321). The result was: 

- the build for `mingw32` succeeded.
- the build for `mingw64` again stopped without detectable reason ... at some other point than before.

To rule out some concurrency-related issues I will change the `PKGBUILD` to call `make` with `--jobs=1` to force `make` to use only one job at a time. The number of this workflow run is [4126](https://github.com/msys2/MINGW-packages/actions/runs/720732426).

WolfgangRamos:
Hi @lazka, using only one job at a time for building did the trick :) Just to be sure, I added a comment to trigger the CI one more time to see if this build succeeds as well. Number is [4131](https://github.com/msys2/MINGW-packages/actions/runs/721261559).

WolfgangRamos:
Hi @lazka, the builds for `mingw32` and `mingw64` succeeded two times in a row :) Can this PR be merged like this or is there something else to be done?

lazka:
Thanks. I'll have a look later.

oscarfv:
Emacs' `native-comp` is a not-yet-merged feature in Emacs 28, the development version, which is not planned to be released anytime soon. Creating a package for it here is the same as having `mingw-w64-emacs-git`, something that was rejected on the past by Alexey.

In principle, adding `native-comp` support to https://github.com/msys2/MINGW-packages-dev/blob/master/mingw-w64-emacs-git/PKGBUILD should be easy enough. There is no hurry, though.

WolfgangRamos:
Hi @oscarfv, thanks for taking interest in this MR :) I know  that `native-comp` is still an experimental Emacs feature. I wanted to be able to use it under windows anyway (you can think of it as doing beta testing :D).

With the latest changes, the `PKGBUILD` is now complete IMHO (I installed it successfully on two system so far). 

@oscarfv I added GnuPG as dependency, because downloading packages from MELPA failed with strange errors when GnuPG wasn't installed. Maybe GnuPG should also be a dependency for the Emacs release version?

I close this MR because it is likely going to be rejected anyway. If you want me to reopen it feel free to contact me. No hard feelings from my side :) I will maintain Emacs native compilation `PKGBUILD` on my fork of this repo.

oscarfv:
@WolfgangRamos I'm just another contributor and my word has the same weight as yours. I was just mentioning an old policy on this project about -git packages. Long time ago MSYS2/MinGW-packages used to have an emacs-git binary package, maintained by yours truly, that was removed because of this.

At this point native-comp is being recommended by several prominent Emacs packages (such as emacs-lsp) because it makes a noticeable difference on CPU-intensive tasks. Maybe the MSYS2 maintainers could make an exception on this case, I don't know.

As for GnuPG, I don't see it listed on the Emacs documentation as a dependency, nor can I see how it could affect the operation of the package manager. Maybe you could open an Emacs bug report (M-x report-emacs-bug) about the issues it resolves? If you do, please name the bug# on this discussion so I can notice. Thanks.

PR #8282, opened 2021-04-05
Update vpx to 1.10.0

hgdagon:


ferdnyc:
Unfortunately this upgrade broke vp9 encoding in `ffmpeg`, because it wasn't rebuilt against the new `libvpx` ABI. See #8404.

lazka:
thanks

lazka:
I wonder why. the chaangelig doesn:t say anything about an ABI break

PR #8286, opened 2021-04-05
gtk4: build docs

naveen521kk:


naveen521kk:
upstream does something weird when building docs.

PR #8291, opened 2021-04-05
libtheora: rebuild

lazka:


lazka:
libtheoraenc-1.dll is missing now for some reason..

lazka:
turns out the parallel build is racy

PR #8292, opened 2021-04-05
Emacs native compilation 2

WolfgangRamos:
I created this PR to trigger Github CI builds in order to see if the builds fail again at the same point they did for #8281.
This branch and the one that was build for #8281 point to the same commit.


lazka:
you can always force push to retrigger ci

WolfgangRamos:
Closed since this was only a dummy PR.

PR #8295, opened 2021-04-06
Update Ruby version to 3.0.1

sinankeskin:
Hi.

Notice: I'm not expert but with this changes I'm able to compile and install successfully.

Also there is no `tar.bz2` file on the ruby file server anymore.

Thank you.


lazka:
Have you used the resulting package in any way? So we have at least some testing...

sinankeskin:
Yes I have. Everything is ok.

I'm speaking only for x64 file.

lazka:
Wonderful thanks. We still need to rebuild all ruby packages against this new one first. I'll give that a try.

PR #8312, opened 2021-04-08
libtasn1: makedepend on gtk-doc

jeremyd2019:
Apparently `autoreconf` now ends up trying to call `gtkdocize`, even though docs are not built.  I added makedepend on `gtk-doc` instead of `${MINGW_PACKAGE_PREFIX}-gtk-doc`, seems unclear why the mingw variant would be needed here, and in the process of bootstrapping I haven't gotten to building gtk-doc by the time I build this package.


jeremyd2019:
I split the commits like that to demonstrate in CI the breakage I was fixing.

lazka:
if its not actually used then the msys variant is fine I guess.

I remember a similar fix recently. maybe due to the autoconf update.

Issue #8094, opened 2021-03-09
Crash after winpthread update


omichel:
Since the update to mingw-w64-x86_64-libwinpthread-git-9.0.0.6128.07922837-1 / mingw-w64-x86_64-winpthreads-git-9.0.0.6128.07922837-1, my app is crashing with the following gdb stack:
```
Thread 43 received signal ?, Unknown signal.
[Switching to Thread 2160.0x21f0]
0x00007ffea5b31f46 in ntdll!RtlRaiseStatus ()
   from C:\WINDOWS\SYSTEM32\ntdll.dll
(gdb) #0  0x00007ffea5b31f46 in ntdll!RtlRaiseStatus ()
   from C:\WINDOWS\SYSTEM32\ntdll.dll
#1  0x00007ffea5ae0052 in ntdll!memset () from C:\WINDOWS\SYSTEM32\ntdll.dll
#2  0x00007ffea55a2eed in msvcrt!_setjmpex ()
   from C:\WINDOWS\System32\msvcrt.dll
#3  0x00007ffe9d8a5c31 in ?? () from D:\msys64\mingw64\bin\libwinpthread-1.dll
#4  0x00007ffe6cb88330 in ode!_ZN22dxThreadManagerPthread12simLoopChildEPv ()
   from D:\msys64\home\Olivier\webots\msys64\mingw64\bin\ode.dll
#5  0x00007ffe9d8a4f35 in ?? () from D:\msys64\mingw64\bin\libwinpthread-1.dll
#6  0x00007ffea556af5a in msvcrt!_beginthreadex ()
   from C:\WINDOWS\System32\msvcrt.dll
#7  0x00007ffea556b02c in msvcrt!_endthreadex ()
   from C:\WINDOWS\System32\msvcrt.dll
#8  0x00007ffea43b6fd4 in KERNEL32!BaseThreadInitThunk ()
   from C:\WINDOWS\System32\kernel32.dll
#9  0x00007ffea5a7cec1 in ntdll!RtlUserThreadStart ()
   from C:\WINDOWS\SYSTEM32\ntdll.dll
#10 0x0000000000000000 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) A debugging session is active.

        Inferior 1 [process 2160] will be killed.
```

lazka:
[...]

PR #8320, opened 2021-04-09
gcc: Update to 11.1.0

Astrum-polaris:


mingw-w64-gcc/PKGBUILD
```diff
@@ -145,7 +143,6 @@ prepare() {
     0011-Enable-shared-gnat-implib.patch \
     0012-Handle-spaces-in-path-for-default-manifest.patch \
     0014-gcc-9-branch-clone_function_name_1-Retain-any-stdcall-suffix.patch \
-    0015-makedeps-properly-handle-win-paths.patch \
```
lazka, line 148: Was this fixed upstream? if yes, the file needs to be removed.

mingw-w64-zstd/PKGBUILD
```diff
@@ -14,18 +14,13 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
...
+        zstd-1.4.0-fileio-mingw.patch)
+sha256sums=('29ac74e19ea28659017361976240c4b5c5c24db3b89338731a6feb97c038d293'
+            '210e57fc4bff62e1af1ca8a3da3960de32a91255ba62b057c6bed5b3a7ebb3d5')
 
 prepare() {
   cd "${srcdir}/${_realname}-${pkgver}"
   patch -Np1 -i "${srcdir}/zstd-1.4.0-fileio-mingw.patch"
-
-  # https://github.com/facebook/zstd/pull/2462
-  patch -Np1 -i "${srcdir}/0001-fix-pkgconf-escaping.patch"
```
lazka, line 28: the patch file needs to be removed

mingw-w64-gcc/PKGBUILD
```diff
@@ -243,17 +241,14 @@ build() {
     --enable-libstdcxx-time=yes \
     --disable-libstdcxx-pch \
     ${extra_config} \
-    --disable-isl-version-check \
```
lazka, line 246: what's this about?

mingw-w64-gcc/PKGBUILD
```diff
@@ -243,17 +241,14 @@ build() {
...
     --enable-lto \
     --enable-libgomp \
     --disable-multilib \
-    --enable-checking=release \
     --disable-rpath \
     --disable-win32-registry \
     --disable-nls \
     --disable-werror \
     --disable-symvers \
-    --disable-plugin \
```
lazka, line 256: what's this about?

mingw-w64-gcc/PKGBUILD
Astrum-polaris: - [x] Yes

mingw-w64-gcc/PKGBUILD
Astrum-polaris: This is probably a workaround used in the old days when the isl version does not pass the check so it is disabled.

Now the check pass successfully so there is no need to disable it.

lazka: ok, thanks

mingw-w64-gcc/PKGBUILD
Astrum-polaris: It was disabled because linking fail with "export ordinals too large"
Now the build pass without disabling it, similar to the isl version check thing.

lazka: perfect, thanks

lazka:
I'd like to wait a bit for a response to https://github.com/msys2/MINGW-packages/issues/8094#issuecomment-817318995 before updating gcc

lazka:
Thanks. Looks good. We need to report this and debug this eventually :/ I've cherry picked the zstd/curl changes to master for now.

dscho:
> I've cherry picked the zstd/curl changes to master for now.

Heh, just when you did that, cURL v7.76.1 came out ;-) I'll open a separate PR for that.

dscho:
> I'll open a separate PR for that.

I [just did](https://github.com/msys2/MINGW-packages/pull/8377), and for good measure I [also opened a corresponding `MSYS-packages` PR](https://github.com/msys2/MSYS2-packages/pull/2445).

revelator:
btw you ask how to bootstrap ada with ucrt, you would probably need a cross compiler which initially might be easier to setup on linux.

lazka:
no clear error message for mingw32, maybe https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98341 ?

lazka:
I've cherry-picked your 10.3 changes to #8532

Astrum-polaris:
I disabled ada in MINGW32, the build was successful but install failed because new gcc conflicts with old ada.

Do we want to disable ada in MINGW32 for now (and fix CI) or try reverting more commit or waiting for potential fixes?

oscarfv:
We already released gcc without Ada support on the past, after trying hard to solve the blocking issue and then waiting a long time for a fix from upstream, which came shortly after the binaries were released when some users realized that Ada was missing and poked the Ada developers.

adisuciu:
Does this mean that you can't have ADA and latest GCC on your system ? I'm getting conflicts in my CI, beacuse of this and I'm not sure what the correct resolution is - remove ADA ? https://github.com/appveyor/ci/issues/3672

lazka:
> Does this mean that you can't have ADA and latest GCC on your system ? I'm getting conflicts in my CI, beacuse of this and I'm not sure what the correct resolution is - remove ADA ? [appveyor/ci#3672](https://github.com/appveyor/ci/issues/3672)

We haven't merged anything, so this can't affect you. You are doing partial upgrades ("-Sy") which isn't supported. Either just do "-S" or follow the instructions here: https://www.msys2.org/docs/ci/#appveyor

lazka:
Filed a bug for ada: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100486

oscarfv:
> Filed a bug for the ada error: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100486

A gcc developer requested some extra info there. I provided what I know. The actual configure command is not shown on the CI build log.

mati865:
Building with `-j1` could be useful for finding the error/failing command but it will take long long time.

oscarfv:
> Building with `-j1` could be useful for finding the error/failing command but it will take long long time.

The error is reported here:

```
make[3]: *** [../../gcc-11.1.0/gcc/ada/gcc-interface/Make-lang.in:1052: ada/libgnat/a-except.o] Error 1
```

and that file is compiled just a few lines above:

```
/C/_/mingw-w64-gcc/src/build-i686-w64-mingw32/./prev-gcc/xgcc -B/C/_/mingw-w64-gcc/src/build-i686-w64-mingw32/./prev-gcc/ -B/mingw32/i686-w64-mingw32/bin/ -L/mingw32/i686-w64-mingw32/lib -L/mingw32/lib -isystem /mingw32/i686-w64-mingw32/include -isystem /mingw32/include -B/mingw32/i686-w64-mingw32/bin/ -B/mingw32/i686-w64-mingw32/lib/ -isystem /mingw32/i686-w64-mingw32/include -isystem /mingw32/i686-w64-mingw32/sys-include   -fno-checking -c -g -O2 -D__USE_MINGW_ACCESS -Wno-pedantic-ms-format -fno-checking -gtoggle  -gnatpg  -W -Wall -g -O1 -fno-inline \
2021-05-08T11:34:54.9009139Z  -nostdinc -I- -I. -Iada/generated -Iada -Iada/gcc-interface -I../../gcc-11.1.0/gcc/ada -I../../gcc-11.1.0/gcc/ada/gcc-interface -Iada/libgnat -I../../gcc-11.1.0/gcc/ada/libgnat ../../gcc-11.1.0/gcc/ada/libgnat/a-except.adb -o ada/libgnat/a-except.o
```

IIUC this means that the compiler crashes, right?

jeremyd2019:
> unless unwinding is broken...

Which it seems to frequently be. ☹️

oscarfv:
Can we go ahead with this, please? Let's just release it without Ada on i686.

mati865:
If 32-bit unwinding is broken (maybe it uses dwarf 5 now?) then it's bigger blocker than missing Ada.

revelator:
seems to be related to dwarf unwinding indeed i use sjlj in my personal build and it works there so :S

oscarfv:
> If 32-bit unwinding is broken (maybe it uses dwarf 5 now?) then it's bigger blocker than missing Ada.

Can anyone confirm that 32-bit unwinding is broken (outside of Ada)? And, if it is broken, is there a bug report on gcc's bugzilla?

oscarfv:
Upstream [is asking again](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100486#c3) for the configure line and mentions that a recent change on the gcc-11 branch could help.

revelator:
odd that it only seems to affect the 32 bit build when using dwarf exceptions, i can build it no problem here using sjlj.

revelator:
just tried building it with differing exception models, only dwarf builds fail :S seh and sjlj work just fine.
Did they change anything recently in the exception code for dwarf ? i tried making a diff against the previous libgcc but nothing appareant seems to have changed so im a bit miffed.

mati865:
This is the change I was talking about: 
> For targets that produce DWARF debugging information GCC now defaults to DWARF version 5 (with the exception of VxWorks and Darwin/Mac OS X which default to version 2 and AIX which defaults to version 4). This can produce up to 25% more compact debug information compared to earlier versions.

lazka:
Could this upstream fix be related https://github.com/gcc-mirror/gcc/commit/8ce35e4c066b68d0cbc656b000ece84f7ea7741a ?

revelator:
lets cross fingers :)

oscarfv:
> Could this upstream fix be related [gcc-mirror/gcc@8ce35e4](https://github.com/gcc-mirror/gcc/commit/8ce35e4c066b68d0cbc656b000ece84f7ea7741a) ?
> 
> edit: I'm going to backport that, it seems relevant -> #9085

I tried with that change backported. No dice, the build fails the same.

revelator:
ya thinking abit about it this patch only seems to affect PECOFF so if it was broken it would affect all exception models i think ?.

mati865:
> 
> 
> ya thinking abit about it this patch only seems to affect PECOFF so if it was broken it would affect all exception models i think ?.

I don't think so, Clang uses different PIC mode between 32-bit and 64-bit Windows and I'd expect GCC to do the same.

oscarfv:
So @lazka merge of the EH pointer fix [worked](https://github.com/msys2/MINGW-packages/runs/2982714668?check_suite_focus=true)?

This is great news, but now I wonder why mine failed.

revelator:
ofc i ment in regards to the exception model :), on a hunch i tried bootstrapping gcc-11 with an older sjlj exception based model
and the result is that it builds Oo atleast this far as the bootstrap is not yet done (how far did the ci build go before it bonked out ?), im at stage2 and the gnat compiler itself has been built, so waiting to see if it fails on the ada runtime now.

lazka:
> So @lazka merge of the EH pointer fix [worked](https://github.com/msys2/MINGW-packages/runs/2982714668?check_suite_focus=true)?

I backported it for 10.x. I assumed it might be helpful based on the upstream comment, but as always I have no clue in this area..:

> AFAICS the failure occurs during stage #2 so the stage #1 compiler might be miscompiled by the base compiler

revelator:
if i manage to build it would it be possible to use it in a ci build to bootstrap itself ? i can upload the packages once done.

mati865:
It'd be worth to try building with https://github.com/msys2/MINGW-packages/pull/9085 first.

revelator:
allready am ;)

oscarfv:
> > So @lazka merge of the EH pointer fix [worked](https://github.com/msys2/MINGW-packages/runs/2982714668?check_suite_focus=true)?
> 
> I backported it for 10.x. I assumed it might be helpful based on the upstream comment, but as always I have no clue in this area..:
> 
> > AFAICS the failure occurs during stage #2 so the stage #1 compiler might be miscompiled by the base compiler

Oh, I should look harder. But your motivation is interesting.

For the record, passing `-dwarf-4` on `CFLAGS/CXXFLADS/BOOT_CFLAGS` changes nothing.

revelator:
stage 3 still going :D

lazka:
oh, I tried to close/reopen to trigger CI again, but your branch is no longer there :(

oscarfv:
@lazka : who is "you" in "your branch"?

Right now I'm building 10 with the PEH patch on C.I., then I'll download and install the packages and locally build 11 with the PEH patch.

Just in case you ( @lazka ) already uploaded the patched packages to C.I., I'm re-running the 11+PEH patch [on my C.I.](https://github.com/oscarfv/MINGW-packages/actions/runs/995434750)

revelator:
build successfully completed here, if the +PEH patch indeed does fix it for the ci build i wont go further, if not ill upload my build and you can try bootstrapping with that.

lazka:
I've re-created this and rebased in #9088

PR #8321, opened 2021-04-09
Add install scripts for CLANG MSYSTEMs.

jeremyd2019:
Generated by:
```bash
for m in CLANG32 CLANG64 CLANGARM64; do
  for f in $(find . -name \*UCRT64.install); do
    n="$(echo "$f" | sed -e "s/UCRT64/${m}/")"
    sed -e "s/ucrt64/${m,,}/g" < "$f" > "$n"
    git add "$n"
  done
done
```


Biswa96:
Wasn't there a movement to replace install scripts with pacman hooks?

mati865:
It's not always that easy.

PR #8324, opened 2021-04-09
clang libtool fixes

jeremyd2019:
autoreconf in several packages to get updated libtool which supports clang.  Fixes msys2/CLANG-packages#9

### DIFF
--- a/mingw-w64-libssh2/PKGBUILD
+++ b/mingw-w64-libssh2/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libssh2
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.9.0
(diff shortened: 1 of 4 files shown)


jeremyd2019:
I converted this to draft due to an alternate (more proper) fix for the `lt_cv_deplibs_check_method='pass_all'` workaround being discussed over on msys2/CLANG-packages#9

jeremyd2019:
D'oh
```
patching file libssh2.pc.in
  configure.ac:130: error: m4_undefine: undefined macro: backend
  configure.ac:130: the top level
```
Looks like they are trying to undefine something that's not defined.  A quick look suggests that it's the variable in a for loop, maybe scoping rules changed?

jeremyd2019:
I ran through building a bunch of packages, on i686 and aarch64, and this seems to be working now after msys2/MSYS2-packages#2439

PR #8325, opened 2021-04-09
Disable assembly on new arches

jeremyd2019:
Arches can be added to case statements once assembly support is added.

Openssl is a little different, and specifically handles arm because it 'needs' to know 32 vs 64-bit.

### DIFF
--- a/mingw-w64-libgcrypt/PKGBUILD
+++ b/mingw-w64-libgcrypt/PKGBUILD
@@ -5,7 +5,7 @@ _realname=libgcrypt
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.9.2
-pkgrel=1
+pkgrel=2
 pkgdesc="General purpose cryptographic library based on the code from GnuPG (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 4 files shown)


Biswa96:
Does `mingw64 aarch64_asm` work in openssl for aarch64?

jeremyd2019:
No.  I believe the reason is that Windows AArch64 uses the Thumb instruction set, while pretty much everybody else uses the ARM instruction set on AArch64.

jeremyd2019:
I had hoped my patch to add new configs to openssl for arm and aarch64 wouldn't be needed, since I was disabling asm anyway, but it turns out that `mingw64` config was passing an argument when building the rc that made it try to use x86_64.

Biswa96:
I forgot to tell that openssl _recently_ changes something to fix AArch64 build. See my upstream issue https://github.com/openssl/openssl/pull/12620

Biswa96:
I forgot to tell that openssl _recently_ changes something to fix AArch64 build. See my upstream issue https://github.com/openssl/openssl/issues/13414

jeremyd2019:
I would certainly welcome anything better, this was just to get things to build for me.  But I found that upstream issue hard to follow - there was a comment about adding aarch64, a reply to open a new pull request, but I didn't see anything more than that.

mati865 (approved):

PR #8326, opened 2021-04-09
libffi: backport support for aarch64 mingw

jeremyd2019:
This applies 4 upstream commits related to aarch64 mingw support.


Biswa96:
A little comment (e.g. delete in next ver) above the patch command will be useful in future.

PR #8329, opened 2021-04-10
cmake: bootstrap using system libs.

jeremyd2019:
This causes cmake to use system libs rather than bundled sources when it bootstraps itself.  This probably doesn't matter much, unless one of those libs has been patched for the system (as I need to do for libuv on aarch64).

Also disable some makedepends for ucrt and clang.  I guess I'm not the only one who hasn't gotten qt5 and emacs built 😜

### DIFF
--- a/mingw-w64-cmake/PKGBUILD
+++ b/mingw-w64-cmake/PKGBUILD
@@ -5,7 +5,7 @@ _realname=cmake
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.20.0
-pkgrel=1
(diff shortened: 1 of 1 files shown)


lazka:
yeah, here is a list of things where I skipped some deps and need a rebuild some time:

```
freetype
swig
cairo
librsvg
adwaita-icon-theme
jasper
postgresql
poppler
doxygen
graphviz
cmake
```

jeremyd2019:
I think of those I only built swig and cmake.  I thought swig didn't need the actual languages to be built, only for testing (so maybe some of those makedepends could be changed to checkdepends).  Also, I thought librsvg was another thing that needed rust to build.

PR #8330, opened 2021-04-10
fritzing: update to 0.9.6

jeremyd2019:
I don't know what to do about UCRT... Qt5 isn't available there.


mingwandroid:
Is that for cmake? We need a cmake without any qt5 dep if that's the issue.

jeremyd2019:
no, this is entirely unrelated to my other pull requests

PR #8333, opened 2021-04-10
libimobiledevice: Update to 1.3.0'

nisargjhaveri:


Biswa96:
Python2 is out-of-world. Can't it be used with Python3? Also python is not in dependency list.

nisargjhaveri:
@Biswa96 This version does not yet support. Though there are some fixes after this version, I'll try to backport if possible.

nisargjhaveri:
Fixed the python issue. Missed that it was already patched earlier as well.

PR #8334, opened 2021-04-10
CI: move continue-on-error from steps to job

jeremyd2019:
See
https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/#expressions-in-jobcontinue-on-error

https://github.com/msys2/MINGW-packages/commit/4e952a7445daa0fb9df229ea61b69ccee3264d4a#r49357293

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,7 @@ jobs:
           { msystem: UCRT64, arch: ucrt-x86_64, can-fail: false }
         ]
     name: ${{ matrix.msystem }}
+    continue-on-error: ${{ matrix.can-fail }}
     steps:
 
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
That means CI will never report a failure, Is that acceptable for you?
What is the purpose of CI then? Isn't it to check if changes pass or not?

jeremyd2019:
Apparently specifying `continue-on-error` on the job results in the job still reporting as failed in the lists, and also (perhaps more importantly) in the icon on the commit and PR being the red x rather than the green check.  So this may not be desired.

jeremyd2019:
It does show the green icon on the Actions page though, for what that's worth.

lazka:
yeah, that's why I removed it for ucrt64 again. Too hard to see if it failed or not.

PR #8346, opened 2021-04-11
libtool: support `llvm-objdump -f` output

mati865:
More info in https://github.com/msys2/CLANG-packages/issues/9


lazka:
Do you happen know the difference between the mingw libtool and the msys one and when which is used?

mati865:
I have no idea but usually both are patched at same time so I just followed that.

lazka:
ah ok

PR #8348, opened 2021-04-12
swig: move languages to checkdepends.

jeremyd2019:
The swig docs say that the languages are only necessary to run the test
suite or build the examples, so move them from `makedepends` to
`checkdepends`.

So that `checkdepends` actually means something, add a check function to
run swig's test suite, which takes a long time and is expected to give errors.
Some thought could be given to expanding `checkdepends` in the future,
adding more languages, and also boost.


lazka:
nice!

PR #8349, opened 2021-04-12
fixes for LLD support

jeremyd2019:
- mpdecimal: remove `-fPIC`, unsupported on Windows. (I guess that's for clang, not lld.  Oops)  Fixes msys2/CLANG-packages#5
- sqlite3: remove `--no-undefined` linker option, which is unknown to lld on Windows (Windows doesn't allow undefined symbols in shared libraries)  Fixes msys2/CLANG-packages#10
- libxml2: patch out version script linker option, unsupported on Windows.  Fixes msys2/CLANG-packages#19
- lua: link to import library, lld does not support linking directly to a DLL.
- icu: remove `-Bsymbolic` linker flag, which lld on Windows does not accept.  It appears to be an ELF-specific option.  Fixes msys2/CLANG-packages#22

### DIFF
--- a/mingw-w64-icu/0024-mingw-no-Bsymbolic.patch
+++ b/mingw-w64-icu/0024-mingw-no-Bsymbolic.patch
@@ -1,0 +1,22 @@
+--- icu/source/config/mh-mingw64.orig	2021-01-31 17:25:18.521418000 -0800
++++ icu/source/config/mh-mingw64	2021-01-31 17:25:53.240156000 -0800
+@@ -52,7 +52,7 @@
+ #LINK.cc=      $(CXX) $(CXXFLAGS) $(LDFLAGS)
+ 
+ ## Shared library options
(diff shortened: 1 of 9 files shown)


mati865:
Do you know how `libxml2` used version script? Some projects use it to control which symbols to export on Windows.

lazka:
* https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/Makefile.am#L25
* https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/libxml2.syms

lazka:
checking the dll exports and seems nothing changed, so looks fine I guess

jeremyd2019:
A better fix rather than a hack would be to fix configure to make sure version script is actually supported, rather than trying to detect GNU ld and assuming it is if so.  I've seen other configure scripts that work that way.

PR #8351, opened 2021-04-12
python-jsonschema: add six dep, sync w/ template

### LINKED ISSUE #8275: mingw-w64-python-jsonschema missing dependency on six
Building darktable in msys2 fails because of six being required by jsonschema, but not installed through mingw-w64-python-jsonschema:

```
pkg_resources.DistributionNotFound: The 'six>=1.11.0' distribution was not found and is required by jsonschema
```
https://github.com/darktable-org/darktable/runs/2261989163?check_suite_focus=true#step:4:1094

jsonschema has dropped python2 support and thus doesn't require six anymore on their main branch, but the latest release (which is packaged by mingw) still requires it.

Adding an explicit dependency on `mingw-w64-x86_64-python3-six` fixes the problem as expected.

kmilos:
Fixes #8275


mingw-w64-python-jsonschema/PKGBUILD
```diff
@@ -17,38 +17,43 @@ depends=("${MINGW_PACKAGE_PREFIX}-python"
...
+  ${MINGW_PREFIX}/bin/python setup.py check
+}
 
+package() {
+  cd "${srcdir}/python-build-${CARCH}"
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
-  ${MINGW_PREFIX}/bin/python setup.py install \
-    --prefix=${MINGW_PREFIX} --root=${pkgdir} --optimize=1 --skip-build
-
-  find "${pkgdir}/" -name "*.pyc" -delete
```
lazka, line 45: this line needs to stay, otherwise users will get update errors due to files already existing.

PR #8356, opened 2021-04-12
jo: add new package

1480c1:
Signed-off-by: Christopher Degawa <ccom@randomderp.com>

### DIFF
--- a/mingw-w64-jo/PKGBUILD
+++ b/mingw-w64-jo/PKGBUILD
@@ -1,0 +1,37 @@
+# Maintainer: Christopher Degawa <ccom@randomderp.com>
+
+_realname=jo
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=1.4
+pkgrel=1
+pkgdesc='JSON output from a shell (mingw-w64)'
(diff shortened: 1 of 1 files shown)


1480c1:
bash-completion is in the wrong place

1480c1:
left bash-completion stuff up to the msys2 version of the package

mingw-w64-jo/PKGBUILD
```diff
@@ -0,0 +1,37 @@
+# Maintainer: Christopher Degawa <ccom@randomderp.com>
+
+_realname=jo
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=1.4
+pkgrel=0
```
Biswa96, line 7: I think not everything start with zero, right? 🤔

1480c1: https://wiki.archlinux.org/index.php/PKGBUILD#pkgrel
> The release number. This is usually a positive integer number ... When a new version of the software is released, this value must be reset to 1

You're right

lazka:
There was a plan at one point to make bash-completion also look in the mingw prefix, so this is fine actually.

PR #8373, opened 2021-04-15
Fix ucrt builds

Biswa96:


mingw-w64-gc/PKGBUILD
```diff
@@ -23,9 +25,10 @@ prepare() {
 }
 
 build() {
-  mkdir -p "${srcdir}/build-${MINGW_CHOST}"
-  cd "${srcdir}/build-${MINGW_CHOST}"
-  export lt_cv_deplibs_check_method='pass_all'
+  [[ -d "${srcdir}/build-${MINGW_CHOST}" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}"
+  mkdir -p "${srcdir}/build-${MINGW_CHOST}" && cd "${srcdir}/build-${MINGW_CHOST}"
+
+  CFLAGS+=" -fcommon" \
```
ivmai, line 31: Could you please tell me the error/waring details (which you workarounded with -fcommon) so that I could fix the issue upstream?

MehdiChinoune: GCC 10 default to `fno-common` which broke many projects at the time.
I think a the time `gc` wasn't updated to build without `-fcoomon`.

PR #8381, opened 2021-04-16
Upgrade to LLVM 12

Astrum-polaris:
Use single source to simplify PKGBUILD
Dropped half of the patched fixed upstream


Astrum-polaris:
CI succeed ✨ 

openmp is disabled in 32bit build since it failed with
`mingw-w64-clang/src/openmp/runtime/src/kmp_atomic.cpp:666:16: error: could not convert 'std::operator+<__float128>(lhs.kmp_cmplx128_a16_t::q.__kmp_cmplx128_t::<anonymous>, rhs.kmp_cmplx128_a16_t::q.__kmp_cmplx128_t::<anonymous>)' from 'std::complex<__float128>' to 'kmp_cmplx128_a16_t'`
Any idea how to fix this?

mingw-w64-clang/0305-use-TerminateProcess-instead-of-exit.patch
```diff
@@ -1,23 +1,23 @@
...
  using namespace llvm;
  using namespace lld;
  
-@@ -67,6 +71,9 @@
+@@ -80,6 +84,9 @@
      lld::outs().flush();
      lld::errs().flush();
    }
 +#ifdef _WIN32
 +  TerminateProcess(GetCurrentProcess(), val);
```
jeremyd2019, line 19: IIRC from the time this was added, there was discussion that the process shutdown was changed and this patch may no longer be required.

Astrum-polaris: Maybe.
I would like to hear @mati865's opinion on this also

mati865: IIRC there was change to LLVM's threading library that should make it obsolete.
The best way to test is running LLD test suite, I'll do that when I find time for the review.

mati865: The patch is still necessary when building LLVM with GCC, trace of the hang:
```
ntdll!NtWaitForSingleObject+0x14
KERNELBASE!WaitForSingleObjectEx+0x8e
libwinpthread_1!pthread_mutex_lock+0xaf
libwinpthread_1!pthread_join+0x23
libstdc___6!ZNSt6thread4joinEv+0x17
libLLVM!ZN4llvmlsERNS_11raw_ostreamENS_8NoneTypeE+0x54d
libLLVM!ZN4llvmlsERNS_11raw_ostreamENS_8NoneTypeE+0xa0a
libLLVM!ZN4llvm16windows_manifest21WindowsManifestMerger5mergeERKNS_12MemoryBufferE+0xf9a5
libLLVM+0x117c
libLLVM+0x125d
ntdll!LdrpCallInitRoutine+0x61
ntdll!LdrShutdownProcess+0x141
ntdll!RtlExitUserProcess+0xad
kernel32!ExitProcessImplementation+0xb
msvcrt!_crtExitProcess+0x15
msvcrt!doexit+0x171
libLLVM!ZN4llvm3sys7Process4ExitEib+0x2d
ld_lld+0x3b4d
ld_lld+0x1a320
ld_lld+0x1c8fac
ld_lld+0x237d
ld_lld+0x3b4b6e
ld_lld+0x13c1
ld_lld+0x14f6
kernel32!BaseThreadInitThunk+0x14
ntdll!RtlUserThreadStart+0x21
```

mati865:
@Astrum-polaris the dust after merging Clang subsystem hast mostly settled. Would you like to rebase your PR or should I do it?

mingw-w64-clang/PKGBUILD
```diff
@@ -167,27 +112,18 @@ apply_patch_with_msg() {
...
-  [[ -d ${srcdir}/libcxx-${pkgver} ]] && rm -rf ${srcdir}/libcxx-${pkgver}
-  [[ -d ${srcdir}/libcxx-${pkgver} ]] || tar -xJvf ${srcdir}/libcxx-${pkgver}.src.tar.xz -C ${srcdir} || true
-
-  plain "Extracting lldb-${pkgver}.src.tar.xz due to symlink(s) without pre-existing target(s)"
-  [[ -d ${srcdir}/lldb-${pkgver} ]] && rm -rf ${srcdir}/lldb-${pkgver}
-  [[ -d ${srcdir}/lldb-${pkgver} ]] || tar -xJvf ${srcdir}/lldb-${pkgver}.src.tar.xz -C ${srcdir} || true
+  plain "Extracting llvm-project-${pkgver}.src.tar.xz due to symlink(s) without pre-existing target(s)"
+  cd "${srcdir}"
+  rm -rf llvm clang clang-tools-extra compiler-rt lld lldb libcxx libcxxabi libunwind openmp polly || true
+  tar --strip-components=1 -xJvf ${srcdir}/llvm-project-${pkgver}.src.tar.xz -C ${srcdir} || true
```
mati865, line 118: Maybe we should use `tar` without `-v` here to avoid gazillion lines of output?

mingw-w64-clang/PKGBUILD
```diff
@@ -196,56 +132,25 @@ prepare() {
   apply_patch_with_msg \
       "0007-getLoadAddressWithOffset-assertion.patch"
```
mati865, line 133: Let's drop these two patches. They are really old and patch similar to the first one was reverted shortly after landing because of the regressions.

mingw-w64-clang/0504-fix-lldb-library-name-mingw.patch
```diff
@@ -1,7 +1,7 @@
 --- lldb/source/API/CMakeLists.txt.orig      2020-04-01 11:55:48.460981900 +0200
 +++ lldb/source/API/CMakeLists.txt   2020-04-01 12:03:55.295954500 +0200
 @@ -19,7 +19,14 @@
-   set(option_framework FRAMEWORK)
+   set(lldb_lua_wrapper ${lua_bindings_dir}/LLDBWrapLua.cpp)
  endif()
 
 +if(MSVC)
```
mati865, line 7: This patch should be unnecessary in LLVM 12: https://github.com/llvm/llvm-project/commit/cc76965b19085519278bff1052059e03769b71e8

mingw-w64-clang/PKGBUILD
```diff
@@ -299,12 +207,14 @@ build() {
...
 
-  # sed away the bad windows style command line parameters (they are translated by msys from /bad to c:\msys\bad
-  # Ninja doesn't need fixing them.
-  sed -i.orig 's/\/c \/Fo/-c -Fo/' projects/openmp/runtime/src/CMakeFiles/omp.dir/build.make
-  sed -i.orig 's/\/safeseh \/coff/-safeseh -coff/' projects/openmp/runtime/src/CMakeFiles/omp.dir/build.make
+  if [ "${CARCH}" == "x86_64" ]; then
+    # sed away the bad windows style command line parameters (they are translated by msys from /bad to c:\msys\bad
+    # Ninja doesn't need fixing them.
+    sed -i.orig 's/\/c \/Fo/-c -Fo/' projects/openmp/runtime/src/CMakeFiles/omp.dir/build.make
+    sed -i.orig 's/\/safeseh \/coff/-safeseh -coff/' projects/openmp/runtime/src/CMakeFiles/omp.dir/build.make
```
mati865, line 216: Is it still necessary?

mingw-w64-clang/PKGBUILD
```diff
@@ -299,12 +207,14 @@ build() {
...
+  fi
```
mati865, line 217: Is it still necessary?

mati865:
While at it we could really use two significant fixes for GC of the sections: https://reviews.llvm.org/rG2b01a417d7ccb001ccc1185ef5fdc967c9fac8d7 and https://reviews.llvm.org/D101615

I'll take care of other backports in another PR since some of them depend on the patches I mentioned above.

mingw-w64-clang/PKGBUILD
Astrum-polaris: Good Point.
I already do this in https://github.com/msys2/MINGW-packages/pull/8393 to reduce log size

mingw-w64-clang/PKGBUILD
Astrum-polaris: 🆗

mingw-w64-clang/0504-fix-lldb-library-name-mingw.patch
Astrum-polaris: 🆗

mingw-w64-clang/PKGBUILD
Astrum-polaris: `/safeseh /coff` seems only present in 32bit build
I still see `/c /Fo` in 64bit build but not sure if the sed is still needed

Astrum-polaris:
@mati865 Feel free to open another PR to supersede this
I might not have enough time to rebase this in the next few days, and I think you can do better than me

mingw-w64-clang/PKGBUILD
mati865: Okay, I'll send patches to the upstream for it one we finish upgrading.

jeremyd2019:
> openmp is disabled in 32bit build since it failed with
...

I think this is gcc-specific.  I added a CLANG32 job to the CI matrix and it has gotten past the point in the build where MINGW32 failed.  I don't know if that helps much, other than narrowing down the issue.

Astrum-polaris:
> I think this is gcc-specific. 

I think so too, maybe gcc 11.1 will help, but it has its own issue (ada bootstrap failed in 32bit)

> Has this been reported upstream?

No, I do not have a Bugzilla account.

lazka:
happy to merge is someone OKs this.

jeremyd2019:
I would prefer either a fix for i686 openmp, or a better disablement (it works fine on clang32, just mingw32 is borked)

mati865:
Thank you for rebasing and sorry for the delay.
I've inspected openmp code and I'm convinced this is odd GCC bug, I have patch here: https://github.com/mati865/MINGW-packages/commit/618aaab74b91b0e5c0e185e71755b24a2f4cccbb

I did 32-bit build only and noticed LLD hang during `check()` so I'll have to check it before we can merge.

mati865:
Just confirmed the hang doesn't happen with `clang64` subsystem. We'd probably need trace with debuginfo. I'll try to obtain it tomorrow.

jeremyd2019:
Might just be a new/different place that lld calls an `exit` function rather than the one that was patched to use TerminateProcess

PR #8376, opened 2021-04-15
github cli v1.9.0

dscho:
This upgrades `gh` to v1.9.0, and while at it also allows running in WSL.

PR #8383, opened 2021-04-16
github cli v1.9.1

dscho:
That was fast. The .0 version came out only yesterday...

And we may want to wait a couple of days, as [v1.9.2 might be needed soon, too](https://github.com/cli/cli/issues/3431).

[EDIT] @lazka [suggested](https://github.com/msys2/MINGW-packages/pull/8376#issuecomment-820988630) that it's not a burden to have a quick release cadence, so I'll mark this as ready to review.


naveen521kk:
what did the PR do? I already updated it in https://github.com/msys2/MINGW-packages/pull/8382 ?

lazka:
not much then I guess :)

PR #8390, opened 2021-04-16
ripgrep: add ripgrep

1480c1:
Uses git tag because tar fails with symlinks

### DIFF
--- a/mingw-w64-ripgrep/PKGBUILD
+++ b/mingw-w64-ripgrep/PKGBUILD
@@ -1,0 +1,56 @@
+# Maintainer: Christopher Degawa <ccom@randomderp.com>
+
+_realname=ripgrep
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=12.1.1
+pkgrel=1
+pkgdesc='line-oriented search tool that recursively searches your current directory for a regex pattern (mingw-w64)'
+arch=(any)
+mingw_arch=(mingw32 mingw64 ucrt64)
+url=https://github.com/BurntSushi/$_realname
+license=(MIT custom:UNLICENSE)
+depends=()
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
+             git)
+source=("${_realname}-${pkgver}::git+https://github.com/BurntSushi/ripgrep.git#tag=12.1.1?signed")
+sha256sums=(SKIP)
+validpgpkeys=(2D8111E141151C072EE9AB9DB2E3A4923F8B0D44) # Andrew Gallant (aka burntsushi)
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}" || return 1
+
+  cargo fetch \
+    --locked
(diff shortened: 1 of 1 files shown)


Biswa96:
If the symlink does not link to source/build files we've `noextract` trickery 🧚

1480c1:
I think it might be fine in this case since it seems he signed the git tag with a gpg key, so we at least have that going for it

1480c1:
Oh wait, I need to add git to the make depends don't I?

mingw-w64-ripgrep/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+pkgrel=1
+pkgdesc='line-oriented search tool that recursively searches your current directory for a regex pattern (mingw-w64)'
+arch=(any)
+mingw_arch=(mingw32 mingw64 ucrt64)
+url=https://github.com/BurntSushi/$_realname
+license=(MIT custom:UNLICENSE)
+depends=()
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
+             git)
+source=("${_realname}-${pkgver}::git+https://github.com/BurntSushi/ripgrep.git#tag=12.1.1?signed")
```
lazka, line 16: just wondering, where is `?signed` documented?

mingw-w64-ripgrep/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+sha256sums=(SKIP)
+validpgpkeys=(2D8111E141151C072EE9AB9DB2E3A4923F8B0D44) # Andrew Gallant (aka burntsushi)
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}" || return 1
```
lazka, line 21: what's with the return 1?

lazka:
lgtm

mingw-w64-ripgrep/PKGBUILD
1480c1: https://man.archlinux.org/man/PKGBUILD.5#USING_VCS_SOURCES
> query
(optional) Allows specifying whether a VCS checkout should be checked for PGP-signed revisions. The source line should have the format source=(url#fragment?signed) or source=(url?signed#fragment). Currently only supported by Git.

mingw-w64-ripgrep/PKGBUILD
1480c1: in case somehow we manage to fail to cd, theoretically set -x should catch it, but just to be a bit more verbose

mingw-w64-ripgrep/PKGBUILD
lazka: ah, thanks.. I was looking at the wiki..

mingw-w64-ripgrep/PKGBUILD
lazka: maybe we will get shellcheck one day. then we can use a wrapper like https://github.com/msys2/msys2-devtools/blob/main/experiments/shellcheck-pkgbuild

PR #8393, opened 2021-04-17
LLVM 12

Astrum-polaris:
Open this mainly to test building clang with clang
And to see if 32bit openmp will succeed


jeremyd2019:
FYI, you can enable actions to run in your own fork from the actions tab, and then CI will run on push without having to open a pull request to run it.

Issue #8341, opened 2021-04-11
[perl] How to create packages?


naveen521kk:
I need https://metacpan.org/release/Win32-API for bootstrapping TexLive on https://github.com/msys2/MINGW-packages/pull/8271. Is there any template that I can follow to create a package for that because I couldn't see any Perl module in MinGW repo?

naveen521kk:
looks like `make` directly isn't support for some reason and `mingw32-make` is required.

naveen521kk:
looks like makefile from `perl Makefile.pl` have a prefix in them and can be installed only there, so all DESTDIR needs to be passed manually or we should try to sed it.

PR #8397, opened 2021-04-17
Add some Perl modules

naveen521kk:
It is needed for https://github.com/msys2/MINGW-packages/pull/8271
Also, fixes https://github.com/msys2/MINGW-packages/issues/8341


naveen521kk:
*ucrt* builds fail, they don't have perl

Biswa96:
Instead of copying patchmakefile.py file multiple times, it can go into mingw-w64-PKGBUILD-common folder. Speaking of UCRT, we could wait for perl in UCRT.

naveen521kk:
> Instead of copying patchmakefile.py file multiple times, it can go into mingw-w64-PKGBUILD-common folder.

How can it be added as a source then?

Biswa96:
> How can it be added as a source then?

I have not test this trickery in this packages, just the idea here to avoid the file duplication:

0. It does not need to be in source, git does the file verification.
1. Grab the actual path of PKGBUILD, convert it to `$pkgbuild_path/../mingw-w64-PKGBUILD-common/patchmakefile.py` and store it in a variable.
2. Place the py file in mingw-w64-PKGBUILD-common folder.
3. Run as `python ${_thepyfile}`.

See any mingw-w64 KDE package as example. Though I would like to wait for maintainers' opinion.

naveen521kk:
Rather I will put that in a gist and fetch it as a URL. Your method looks rather complicated to me.

Biswa96:
I changed one PKGBUILD from here and it's just a two line change.

naveen521kk:
> I changed one PKGBUILD

Can you commit that here? Also, that will cause problems if the repository isn't cloned as a whole and only that package's PKGBUILD was build, is that fine?

Biswa96:
> is that fine?

Maintainers' decision.

naveen521kk:
> Maintainers' decision.

In that case I am just leaving it as is.

naveen521kk:
This is ready.

naveen521kk:
ping for review and merge

naveen521kk:
Thanks :)

PR #8401, opened 2021-04-17
openblas: set NO_AFFINITY=1 USE_OPENMP=1

lazka:
To follow what arch does. Also remove install file since the mentioned
env var has no effect with openmp


lazka:
I'm afraid I don't know enough about this to decide if this should go in. If anyone ends up here and thinks this is a good idea please comment and we can try.

PR #8402, opened 2021-04-17
Update python packages

naveen521kk:


mingw-w64-python-ftfy/PKGBUILD
```diff
@@ -3,17 +3,17 @@
...
 pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
-pkgver=6.0
+pkgver=6.0.1
 pkgrel=1
 pkgdesc='Fixes some problems with Unicode text after the fact (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
 url="https://github.com/LuminosoInsight/python-ftfy"
 license=('MIT')
-depends=("${MINGW_PACKAGE_PREFIX}-python-wcwidth")
```
lazka, line 13: I think this is still needed: https://github.com/LuminosoInsight/python-ftfy/blob/master/setup.py#L34

naveen521kk: weird, https://pypi.org/pypi/ftfy/json, PyPI doesn't have them...
```json
"requires_dist":null,
```
maybe I will add manually.

naveen521kk:
@lazka it would nice if we could run the test suite so that we can be sure that things work. Thoughts?

mingw-w64-python-ftfy/PKGBUILD
naveen521kk: should I complain to PyPI?

lazka: I don't think the pypi API exposes all dependencies in all cases (that's why pip and poetry have to download things before knowing)

lazka: This seems to work:

```shell
$ PYTHONUTF8=1 python3 -m johnnydep --output-format json ftfy
2021-04-18 08:18:05 [info     ] init johnnydist                [johnnydep.lib] dist=ftfy parent=None
2021-04-18 08:18:07 [info     ] init johnnydist                [johnnydep.lib] dist=wcwidth parent=ftfy
2021-04-18 08:18:09 [info     ] resolved                       [johnnydep.lib] dist=ftfy name=ftfy required_by=[] spec=ANY v=6.0.1
2021-04-18 08:18:10 [info     ] resolved                       [johnnydep.lib] dist=ftfy name=wcwidth required_by=['ftfy'] spec=ANY v=0.2.5
[
  {
    "name": "ftfy",
    "summary": "Fixes some problems with Unicode text after the fact"
  },
  {
    "name": "wcwidth",
    "summary": "Measures the displayed width of unicode strings in a terminal"
  }
]

```

lazka: I've asked https://github.com/wimglenn/johnnydep/issues/64

lazka:
> @lazka it would nice if we could run the test suite so that we can be sure that things work. Thoughts?

yes, I'm not quite sure how to best do it right now :/

naveen521kk:
> yes, I'm not quite sure how to best do it right now :/

Maybe I will try to get something like a checkbox in the PR template for running the checks. Let's see.

mingw-w64-python-ftfy/PKGBUILD
naveen521kk: Yeah, that's a problem :(
If we could the dependency by some means it would be great. And if johnny dep seems to work, it should be simple to replace the current one.

naveen521kk:
@lazka I think you can merge this, getting the correct list of dependency can be done later.

lazka:
yes, thanks!

mingw-w64-python-ftfy/PKGBUILD
naveen521kk: I think modifying https://github.com/python-poetry/poetry/blob/master/poetry/repositories/pypi_repository.py to suit our needs should be a nice idea. I will see if that works.

lazka: I think poetry also downloads and builds wheels to get all the info in some cases. But I'm not sure how often this is needed and if it's worth it.

naveen521kk: > I think poetry also downloads and builds wheels to get all the info in some cases.

Yeah it does so.

> But I'm not sure how often this is needed and if it's worth it.

If johnnydep works, I will simply use it. Or if it is as simple as copying the above-linked script it is also fine. If both don't work, I would leave it as is.

naveen521kk: Or even better is to directly depend on poetry. let me see.

lazka:
```python
def get_deps(name):
    from johnnydep import JohnnyDist
    return [d.name for d in JohnnyDist(name).children]
```

though it might fail if building a wheel fails (because of some C deps)... so maybe have a fallback if this fails.

naveen521kk: Can we make it to use wheels from PyPI if available than build a wheel ourselves?

naveen521kk: Or does it do that automatically? Anyway thanks, I will change that part of code.

naveen521kk: Or does it do that automatically? Anyway thanks, I will change that part of code.

naveen521kk: Or does it do that automatically? Anyway thanks, I will change that part of code.

lazka: I don't know

naveen521kk: I tried https://github.com/naveen521kk/msys2-updator/commit/8b2b2089e638f41ce2b7e34c0a870d178b925f14 let's see how it goes

lazka: I got a response: `python -m johnnydep ftfy --fields name requires -o json --no-deps`

naveen521kk: I am using the python version and seems to work.
Ah, for C extension it's really worse.

naveen521kk: Maybe I should use normal python rather than the one from MSYS2?

lazka: you should set the following env var: `PYTHONUTF8=1` (without a windows console python falls back to the ANSI codepage for output..)

lazka: ah, you already set PYTHONIOENCODING.. weird

naveen521kk: I know that error is about printing emoji, which for some reason weirdly error's out.
I changed to ubuntu runner and this happened, https://github.com/naveen521kk/msys2-updator/runs/2375589439?check_suite_focus=true#step:9:10

lazka: Aren't dependencies different depending on the Python version and operating system? of course if there is no other way...

naveen521kk: Maybe I should try Git Bash + MSVC python, that should better match.

lazka: Does PYTHONUTF8  not help?

naveen521kk: > Does PYTHONUTF8 not help?

Didn't try yet. Let me see.

naveen521kk: It looks like I can specify the environment myself as a parameter when calling that function and I am gonna do that.

naveen521kk: > Does PYTHONUTF8 not help?

it worked. :blush:

naveen521kk: What I did was to call `py.exe` from msys2 shell, at least that should work until MSYS2 itself provides a `py.exe`.

PR #8413, opened 2021-04-18
cx_Logging - install include with the package

marcelotduarte:


Biswa96:
Shouldn't the patch command go after the `cp` command in `prepare()` step? And any reason for the removal of `build()`?

marcelotduarte:
> Shouldn't the patch command go after the `cp` command in `prepare()` step? 

No.

> And any reason for the removal of `build()`?

Yes. I cannot use --skip-build in package() - throws errors -, so the build step is unnecessary.

PR #8414, opened 2021-04-18
prjtrellis: switch from gcc to clang

umarcor:
In this PR, prjtrellis is built with clang, instead of using gcc. By the way, it's updated.

### DIFF
--- a/mingw-w64-prjtrellis/PKGBUILD
+++ b/mingw-w64-prjtrellis/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=prjtrellis
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.r993.7454564
-pkgrel=2
+pkgver=1.0.r995.da82093
+pkgrel=1
 pkgdesc="Documenting the Lattice ECP5 bit-stream format (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


mingw-w64-prjtrellis/PKGBUILD
```diff
@@ -38,8 +38,17 @@ prepare () {
 
 build() {
   cd "${srcdir}/${_realname}"/libtrellis
-  MSYS2_ARG_CONV_EXCL=- cmake \
+
+  case "${CARCH}" in
+    i686)
+      sed -s 's|__GNUC__|__clang__|g' -i /mingw32/i686-w64-mingw32/include/intrin.h
```
mati865, line 44: This is totally wrong, you should never modify MSYS2 sysroot inside PKGBUILD.

lazka: nice catch, reverted -> cdd9f630a734c4e93a60b3

@umarcor ^

mati865: This error should be fixed by https://gitlab.com/mati865/mingw-w64-mirror/-/commit/c1804804aff14f1b0333d782c9fac1cc904fcd9f
BTW what was the reason for using Clang?

umarcor: So, just removing L42-L46 of this PR should work as soon as that patch arrives in MSYS2?

> BTW what was the reason for using Clang?

Consistency and testing compatibility. Several projects are built using clang on Linux as the main testing environment.

umarcor: > So, just removing L42-L46 of this PR should work as soon as that patch arrives in MSYS2?

Replying to myself: it seems to work #8466.

Issue #8435, opened 2021-04-20
mingw-w64-libopenmpt-modplug?


mxmilkiib:
I see there is https://packages.msys2.org/base/mingw-w64-libopenmpt and https://packages.msys2.org/base/mingw-w64-libmodplug but there is also a shim method of using libopenmpt as libmodplug, which is used for https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libopenmpt-modplug and https://packages.debian.org/sid/libopenmpt-modplug-dev.

If this there were a libopenmpt-modplug package, I am informed that Mixxx should be able to build with module playback support for Windows for the first time.

pinging @drfiemost

drfiemost:
Sure it can be done. But it's just a wrapper around libopenmpt that replaces libmodplug.
I don't get why Mixxx can't be built against the current libmodplug package.

mxmilkiib:
Because it's not thread safe, where as libopenmpt is.

drfiemost:
Package is available: https://packages.msys2.org/base/mingw-w64-libopenmpt-modplug

PR #8442, opened 2021-04-21
Add libopenmpt-modplug

drfiemost:
A wrapper around libopenmpt that replaces libmodplug, as requested in #8435
As noted by upstream:

> CAUTION: The emulation of the libmodplug interface is not complete as libmodplug exposes lots of internal implementation details. If any of those is used by an application, the emulation via libopenmpt will fail and/or crash


drfiemost:
Should libmodplug be marked as conflicting with this one?

lazka:
> Should libmodplug be marked as conflicting with this one?

yes, "conflicts" and "provides"

Biswa96:
The provides and conflicts should also contain `${MINGW_PACKAGE_PREFIX}-` string in package name. Also increase the `pkgrel` of libmodplug.

Biswa96:
With double quote please 😁

lazka:
not sure if conflicts in both directions add much, but can't hurt I guess. Thanks!

PR #8449, opened 2021-04-22
argtable: fix build

jeroen:
This package defines an internal enum to `EOVERFLOW` which already exists. Simply renaming it fixes the build.


Biswa96:
Please increase the `pkgrel` value by 1.

jeroen:
ok done

PR #8452, opened 2021-04-22
Update mingw-w64

lazka:


Biswa96:
Wait 🛑 I think this skips the "crt: Regenerate Makefile.in" commit. Without that the new ucrt files may not be built*.

<sub><sup>*If automake is not executed explicitly</sup></sub>

mingw-w64-headers-git/PKGBUILD
```diff
@@ -4,9 +4,9 @@ _realname=headers
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgdesc="MinGW-w64 headers for Windows"
-pkgver=9.0.0.6158.1c773877
+pkgver=9.0.0.6166.0a00d0ea
```
jeremyd2019, line 7: Why is this one 6166 and the other packages 6165?  Manual munging?

lazka: that's what I get when I try to quickly change things without thinking :)

PR #8454, opened 2021-04-23
[v8] Update to stable version 9.0.257.17

Kwizatz:
This patch includes a change to fetch the v8 source tar
from the github mirror in order to bring build time down.

### DIFF
--- a/mingw-w64-v8/0000-add-mingw-main-code-changes.patch
+++ b/mingw-w64-v8/0000-add-mingw-main-code-changes.patch
@@ -1,8 +1,8 @@
 diff --git a/BUILD.gn b/BUILD.gn
-index 1dc6c7d5ac..fb8be3f42f 100644
+index a9ab6783fa..5e144b24e0 100644
 --- a/BUILD.gn
 +++ b/BUILD.gn
-@@ -427,6 +427,11 @@ config("internal_config") {
+@@ -444,6 +444,11 @@ config("internal_config") {
    if (is_component_build) {
      defines += [ "BUILDING_V8_SHARED" ]
    }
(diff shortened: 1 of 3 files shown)


Biswa96:
As we are using a specific tarball we don't need git and patch in makedepends.

Kwizatz:
Whats the difference between a MINGW64 and a UCRT64 build? seems like the UCRT64 build is trying to compile with MSVC, which is probably an error on the gn side.

1480c1:
ucrt64 build is using a mingw-w64 toolchain that uses ucrt instead of msvcrt as the c runtime library, but in this case it seems it might be a problem on gn's side on detecting ucrt64 as still mingw-w64 gcc

Kwizatz:
It shouldn't be too hard to add support for it, I assume there is a compiler and/or linker flag to use it? let me know what is it and I'll see what can I do, I'll also remove patch and git from the makedepends.

Thanks!

lazka:
You don't need to pass any flags to use it. The build seems to default to msvc though, so I'd guess the mingw detection needs to be fixed/improved.

Kwizatz:
No, I have an idea of what the problem might be, the build determines that it is on a msys/mingw system based on the MSYSTEM enviroment variable, so I'll just have to check what ENV vars are set for UCRT64, is it MSYSTEM=UCRT64?.

I also want to remove git and patch from de dependencies so the PKGBUILD is cleaner.

lazka:
`(MSYSTEM != "" && MSYSTEM != "MSYS")` would be more future proof

Kwizatz:
Got it, thanks!

lazka:
Thanks!

PR #8192, opened 2021-03-24
clang: merge with CLANG-packages PKGBUILD

jeremyd2019:
Add posix style path separators patches, and disable pthread and emutls patches on clang prefix.

I've been meaning to go through my branch of things I changed for clang/arm64 builds and split it into pull requests.  This is the package with the most complexity.

Also went through package dependencies, fixes msys2/CLANG-packages#36

PR #8456, opened 2021-04-23
boost: detect windows-gnu as mingw

jeremyd2019:
tools/build/engine/build.sh uses `gcc -dumpmachine` to determine the
CHOST, but clang reports <arch>-w64-windows-gnu not <arch>-w64-mingw32
like gcc.

Fixes msys2/CLANG-packages#24


jeremyd2019:
The boost PKGBUILD will require additional tweaks for aarch64, to disable libraries that don't build (context and coroutine I think), but my aarch64 system is going to be busy for a while so I thought I'd get this change in a pull request as it's the last thing on my clang branch other than #8192.

PR #8459, opened 2021-04-23
liblastfm: add proper suffix for the Qt5 version as done in Arch

drfiemost:
The include dir and the library name should have a '5' suffix, see https://archlinux.org/packages/extra/x86_64/liblastfm-qt5/

This is what other software expect, i.e. Clementine player:
https://github.com/clementine-player/Clementine/blob/8258c78c0b930e2c227c35075b6ad4b2ac615f17/CMakeLists.txt#L84


lazka:
ok, fair enough.

PR #8461, opened 2021-04-23
add clang64 to boost and icu

jeremyd2019:
These are packages that I have built using clang but were not required for bootstrapping.


jeremyd2019:
This is the extent of the packages I've tried building with clang thus far, and all build on i686 x86_64 and aarch64, with the exception of boost which does not build on aarch64 yet due to msys2/CLANG-packages#26.

Issue #3617, opened 2018-04-18
Build libcurl with http/2 over winssl


jeroen:
The current [mingw-w64-curl](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-curl/PKGBUILD) has a variant to build with nghttp2+openssl that supports HTTP/2, and a variant to build with winssl that does not support HTTP/2

However I would like to build libcurl to use HTTP/2 (nghttp2) over winssl rather than openssl. According to [this post](https://curl.haxx.se/mail/lib-2018-04/0077.html) on the mailing list this can be done:

> You need to build curl with a new enough SDK ("version 8.1 of the Windows SDK"), run on a new enough Windows ("Windows 8.1 / Server 2012 R2 and above") and of course nghttp2.

How would I do this with msys2?

Alexpux:
@jeroen you need specify CFLAGS to target Windows 8.1 and try to rebuild.

jeroen:
Which CFLAGS would I need to set?

MehdiChinoune:
Fixed by #8469

Issue #950, opened 2015-12-25
Cannot statically link libcurl


kryptan:
``` c
#include <curl/curl.h>

void main() {
    curl_easy_init();
}
```

If I try to compile my program with

``` bash
gcc -o crl crl.c -static -DCURL_STATICLIB $(pkg-config --libs --cflags --static libcurl)
```

I get a lot of undefined reference errors. But even if I add missing libraries manually there are still missing symbols:

```
$ gcc -o crl crl.c -static -DCURL_STATICLIB $(pkg-config --libs --cflags --static libcurl) -lgnutls.dll -lnettle -lgmp -liconv -lintl -lwinmm
C:/msys64/mingw64/lib\librtmp.a(rtmp.o):(.text+0xeba): undefined reference to `nettle_mpz_get_str_256'
C:/msys64/mingw64/lib\librtmp.a(rtmp.o):(.text+0x111b): undefined reference to `nettle_mpz_set_str_256_u'
C:/msys64/mingw64/lib\librtmp.a(rtmp.o):(.text+0x122a): undefined reference to `nettle_mpz_get_str_256'
C:/msys64/mingw64/lib\libintl.a(relocatable.o):(.text+0x46): undefined reference to `libiconv_set_relocation_prefix'
collect2.exe: error: ld returned 1 exit status
```

Alexpux:
I build it successfully using ````curl-config````
See:
````
gcc crl.c -o crl -DCURL_STATICLIB $(curl-config --static-libs)
````

kryptan:
Your command links only curl as static lib but all of its dependencies are linked dynamically.
What I want is to produce a self-contained executable which doesn't depend on anything except system windows dlls. If I add `-static` and missing libraries:

    gcc crl.c -o crl -static -DCURL_STATICLIB $(curl-config --static-libs)  -lgnutls.dll -lnettle -lgmp -liconv -lintl -lwinmm

then I get the same errors as above.

Alexpux:
Today I'm upload new gnutls package with static libraries. Feel free to test

elieux:
No response. Closing.

AndrewBelt:
I'm still experiencing the problem above. Using `--static` gives

	C:/msys64/mingw32/lib\libcurl.a(libcurl_la-http2.o):(.text+0x1c): undefined reference to `_imp__nghttp2_session_del'
[...]

Issue #436, opened 2015-01-25
curl missing static lib


Optiligence:
I guess this is unintentional. 
To solve, run configure explicitly with `--enable-static`.

Optiligence:
w8, the flag does not seem to be enough though

Optiligence:
Ok, so i built the static version using cmake (without tests – because there are missing `CMakeLists`, without the exe – because it broke compilation and with patching `FindOpenSSL` – because it was regexing a changed line of code) and despite issuing a »Static linking is broken!« it built fine and seems to work as far as i can tell. 
I just would like having this available right away.

Optiligence:
The intention for this was, that i wanted to use `curl-config --static-libs` so i wouldn’t have to specify all the dependecies myself. I configured a static package with `--enable-static --disable-shared` as it doesn’t seem to be possible to build both together. The resulting `curl-config` executable provided a hand full of link-flags, but not enough, so this all seemed kind of pointless.

Alexpux:
I'm upload yesterday new cur package with both shared and static libraries.

Optiligence:
Yeah, thanks for that. 
As i mentioned above, the thing is, how am i supposed to link to that? 
Compiling the following code with `g++ main.cpp -static -DCURL_STATICLIB $(curl-config --static-libs)`
is not sufficient to link correctly. 
```
#include <curl/curl.h>
int main() {
	CURL * const curl = curl_easy_init();
	if (!curl) {
		curl_easy_cleanup(curl);
	}
}
```

andreldm:
I'm kind of having the same trouble with libcurl. Before migrating to MSYS2/MinGW-64, I used to compile it and all I needed to ship with the app was libcurl.dll. Now it depends on lots of others libs:
libcurl-4.dll
    LIBEAY32.dll
    libidn-11.dll
        libiconv-2.dll
        libintl-8.dll
    librtmp-1.dll
        libgmp-10.dll
        libgnutls-30.dll
            libhogweed-4-0.dll
            libp11-kit-0.dll
            libtasn1-6.dll
        libhogweed-4-0.dll
        libnettle-6-0.dll
    libssh2-1.dll
    SSLEAY32.dll
    zlib1.dll
(duplicates removed)

I've tried to set the "-static" flag, but no cigar:
http://pastebin.com/e2F1DbYc

Strangely the executable went from ~2mb to ~6mb, but was requesting the same dlls.

PR #8469, opened 2021-04-25
curl: Split package to openssl, winssl and gnutls

orgads:
openssl keeps the default name. The other packages have suffixes.

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -40,7 +40,7 @@ for package in "${packages[@]}"; do
     echo "::endgroup::"
 
     echo "::group::[install] ${package}"
-    execute 'Installing' yes:pacman --noprogressbar --upgrade '*.pkg.tar.*'
+    execute 'Installing' install_packages
     echo "::endgroup::"
 
     echo "::group::[diff] ${package}"
--- a/mingw-w64-curl/PKGBUILD
+++ b/mingw-w64-curl/PKGBUILD
@@ -1,49 +1,36 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
-_variant=-openssl
-#_variant=-winssl
-#_variant=-gnutls
-
-if [ "${_variant}" = "-openssl" ]; then
-  _namesuff=
-else
-  _namesuff="${_variant}"
-fi
-
 _realname=curl
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}${_namesuff}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-gnutls"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-winssl")
 pkgver=7.77.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Command line tool and library for transferring data with URLs. (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://curl.se/"
 license=("MIT")
+_cert_depends=("${MINGW_PACKAGE_PREFIX}-ca-certificates"
+               "${MINGW_PACKAGE_PREFIX}-libssh2")
+_openssl_depends=("${MINGW_PACKAGE_PREFIX}-openssl"
+                  "${MINGW_PACKAGE_PREFIX}-nghttp2")
+_gnutls_depends=("${MINGW_PACKAGE_PREFIX}-rtmpdump"
+                 "${MINGW_PACKAGE_PREFIX}-gnutls")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
-             "${MINGW_PACKAGE_PREFIX}-pkg-config")
+             "${MINGW_PACKAGE_PREFIX}-pkg-config"
+             "${_cert_depends[@]}"
+             "${_openssl_depends[@]}"
+             "${_gnutls_depends[@]}")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-c-ares"
          "${MINGW_PACKAGE_PREFIX}-brotli"
          "${MINGW_PACKAGE_PREFIX}-libidn2"
          "${MINGW_PACKAGE_PREFIX}-libmetalink"
          "${MINGW_PACKAGE_PREFIX}-libpsl"
-         "${MINGW_PACKAGE_PREFIX}-libssh2"
          "${MINGW_PACKAGE_PREFIX}-zlib"
-         "${MINGW_PACKAGE_PREFIX}-zstd"
-         $([[ "$_variant" == "-openssl" ]] && echo \
-            "${MINGW_PACKAGE_PREFIX}-ca-certificates" \
-            "${MINGW_PACKAGE_PREFIX}-openssl" \
-            "${MINGW_PACKAGE_PREFIX}-nghttp2")
-         $([[ "$_variant" == "-gnutls" ]] && echo \
-            "${MINGW_PACKAGE_PREFIX}-rtmpdump" \
-            "${MINGW_PACKAGE_PREFIX}-ca-certificates" \
-            "${MINGW_PACKAGE_PREFIX}-gnutls")
-         )
-if [ -n "${_namesuff}" ]; then
-  provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-  conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-fi
+         "${MINGW_PACKAGE_PREFIX}-zstd")
 options=('staticlibs')
 source=("https://github.com/curl/curl/releases/download/${_realname}-${pkgver//./_}/${_realname}-${pkgver}.tar.xz"{,.asc}
         "0001-Make-cURL-relocatable.patch"
@@ -67,7 +54,12 @@ prepare() {
   autoreconf -vfi
 }
 
-build() {
+do_build() {
+  _variant=$1
+  _destdir="${srcdir}/build-${MSYSTEM}"
+  if [ "${_variant}" != "-openssl" ]; then
+    _destdir="${_destdir}${_variant}"
+  fi
   local -a extra_config
   if check_option "debug" "y"; then
     extra_config+=( --enable-debug )
@@ -98,7 +90,7 @@ build() {
     _variant_config+=("--without-librtmp")
   fi
 
-  mkdir -p "${srcdir}/build-static-${MSYSTEM}" && cd "${srcdir}/build-static-${MSYSTEM}"
+  mkdir -p "${_destdir}-static" && cd "${_destdir}-static"
   ../${_realname}-${pkgver}/configure \
     --prefix=${MINGW_PREFIX} \
     --build=${MINGW_CHOST} \
@@ -122,7 +114,7 @@ build() {
   sed -i "s|\/curl > \$\@|\/curl\$\{EXEEXT\} > \$\@|" scripts/Makefile
   make
 
-  mkdir -p "${srcdir}/build-shared-${MSYSTEM}" && cd "${srcdir}/build-shared-${MSYSTEM}"
+  mkdir -p "${_destdir}-shared" && cd "${_destdir}-shared"
   ../${_realname}-${pkgver}/configure \
     --prefix=${MINGW_PREFIX} \
     --build=${MINGW_CHOST} \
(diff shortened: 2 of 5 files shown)


lazka:
Why do you need this?

orgads:
For deploying ScummVM without ca dependencies. See the complex instructions [here](https://wiki.scummvm.org/index.php/Compiling_ScummVM/MinGW-w64).

lazka:
@dscho I see git4win uses winssl and you added mutli backend support to curl some time ago (https://curl.se/mail/lib-2017-08/0118.html). Do you have any thoughts on this change? What would make your life easier (or not harder)?

.ci/ci-build.sh
```diff
@@ -40,7 +40,7 @@ for package in "${packages[@]}"; do
     echo "::endgroup::"
 
     echo "::group::[install] ${package}"
-    execute 'Installing' yes:pacman --noprogressbar --upgrade '*.pkg.tar.*'
+    execute 'Installing' install_packages
```
orgads, line 43: I had to change this because the 3 packages conflict with each other, so they cannot be installed in a single command.

I can split it to a separate commit if you'd prefer that.

lazka: thanks, this is fine as is.

Biswa96: I think adding `replaces` with same value as `conflicts` in PKGBUILD solve the issue. ci change will not be required.

orgads: I'll try.

lazka: no, replaces takes affect on DB update, not at install time. This means you can't use the package it replaces anymore.

orgads: Right. It didn't help.

orgads:
I don't know how to fix (or test locally) clang64. It fails when linking gnutls. Is it possible to opt out some of the sub-packages on clang?

mingw-w64-curl/PKGBUILD
```diff
@@ -1,27 +1,27 @@
...
-else
-  _namesuff="${_variant}"
-fi
-
 _realname=curl
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}${_namesuff}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-gnutls"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-winssl")
```
jeremyd2019, line 7: The gnutls issue is under active investigation.
> Is it possible to opt out some of the sub-packages on clang?

```suggestion
pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
         $( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-${_realname}-gnutls")
         "${MINGW_PACKAGE_PREFIX}-${_realname}-winssl")
```

untested...

mingw-w64-curl/PKGBUILD
```diff
@@ -126,10 +118,53 @@ build() {
   make
 }
 
-package() {
-  cd "${srcdir}/build-${CARCH}"
+build() {
+  do_build -openssl
+  do_build -winssl
+  do_build -gnutls
```
jeremyd2019, line 124:
```suggestion
  if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]]; then
    do_build -gnutls
  fi
```

untested

mingw-w64-curl/PKGBUILD
```diff
@@ -1,27 +1,27 @@
...
+_openssl_depends=("${MINGW_PACKAGE_PREFIX}-openssl"
+                  "${MINGW_PACKAGE_PREFIX}-nghttp2")
+_gnutls_depends=("${MINGW_PACKAGE_PREFIX}-rtmpdump"
+                 "${MINGW_PACKAGE_PREFIX}-gnutls")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
-             "${MINGW_PACKAGE_PREFIX}-pkg-config")
+             "${MINGW_PACKAGE_PREFIX}-pkg-config"
+             "${_common_depends[@]}"
+             "${_openssl_depends[@]}"
+             "${_gnutls_depends[@]}")
```
jeremyd2019, line 24:
```suggestion
             $( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${_gnutls_depends[@]}"))
```

untested...

dscho:
> 
> 
> @dscho I see git4win uses winssl and you added mutli backend support to curl some time ago (https://curl.se/mail/lib-2017-08/0118.html).

Right. In many ways, what I did is the opposite of what this change does. My intention was to have a single package providing all SSL backends, while this PR intends to host the different SSL backends in different packages.

> Do you have any thoughts on this change?

I can see how it makes more sense for MSYS2 to have the three separate packages. After all, users specify precisely what packages they want, and they can then specify which backend they want by using the appropriate package name.

The main problem I see with this is _dependencees_. I.e. if there is a package that depends on `mingw-w64-curl` (such as `mingw-w64-git`), and the user installs `mingw-w64-gnutls`, that dependencee will now have an unfulfilled dependency.

I am not familiar enough with `pacman` to know whether there is a correct way to specify "we have multiple alternative variants of this package; any will do as dependency of XYZ".

> What would make your life easier (or not harder)?

Since Git for Windows builds its own `curl` packages, this PR has no impact on my life that would it either harder or easier ;-)

orgads:
Regarding the dependency issue, it is already handled. The alternative packages have `provides`, which means they can serve as a dependency for packages that depend on `mingw-w64-curl`.

dscho:
> 
> 
> Regarding the dependency issue, it is already handled. The alternative packages have `provides`, which means they can serve as a dependency for packages that depend on `mingw-w64-curl`.

Thanks, for some reason I had missed that (I did see the `conflicts` entries in the `package_*()` functions, though...)

Biswa96:
Let me solve the gnutls build issue if possible.

orgads:
I don't mind. But anyway, you can always pull out the exclusions here in the same commit when you fix the gnutls build. Either way is fine with me.

orgads:
@Biswa96 did you report to lld? From my experience, they're glad to help with reported bugs.

Biswa96:
Nope. Experts are discussing things here https://github.com/msys2/CLANG-packages/issues/39

orgads:
Waiting for review then. @lazka?

msink:
Can `curl-winssl` package here be configured to use default Windows cert store?

https://stackoverflow.com/a/37553616:
> Starting with libcurl 7.71.0, due to ship on June 24, 2020, it will get the ability to use the Windows CA cert store when built to use OpenSSL. You then need to use the CURLOPT_SSL_OPTIONS option and set the correct bit in the bitmask: CURLSSLOPT_NATIVE_CA.

Or it should "curl with default Windows certs" be yet another package, and separate PR?

orgads:
> Will `curl-winssl` package use default Windows cert store?
> 
> https://stackoverflow.com/a/37553616:
> 
> > Starting with libcurl 7.71.0, due to ship on June 24, 2020, it will get the ability to use the Windows CA cert store when built to use OpenSSL. You then need to use the CURLOPT_SSL_OPTIONS option and set the correct bit in the bitmask: CURLSSLOPT_NATIVE_CA.
> 
> Or "curl with default Windows certs" should be yet another package, and separate PR?

Looks like it does. No further changes are needed.

dscho:
I think there is some confusion here: `-winssl` means the Secure Channel backend, while the StackOverflow comment is about the OpenSSL backend. Two pretty different backends. Secure Channel uses the Windows-native certificate store _already_.

msink:
And one more question - '-winssl' still depends on `libssh2`, and it depends on `openssl`.
Will it really work?

orgads:
Well?

msink:
@orgads  Did you managed to link `libcurl-winssl` statically?

I tried to download artifacts from CI, and add dependencies from `pkgconfig`
```
            linkerOpts(
                "-L${mingwPath.resolve("lib")}",
                "-Wl,-Bstatic",
                "-lstdc++",
                "-static",
                "-lcurl-winssl",
                "-lssh2-wincng",
                "-ladvapi32",
                "-lbrotlidec-static",
                "-lidn2",
                "-lpsl",
                "-lwldap32",
                "-lws2_32",
                "-lz",
                "-lzstd"
            )
```
but got many link errors.

orgads:
@msink The library name is not `-lcurl-winssl`, but `-lcurl`. Same goes for `-lssh2`.

msink:
@orgads Yes, I just didn't want to overwrite default ones, so renamed them before copying to `/msys64/mingw/64/lib`. 
Anyway linker did found them, but not found many other things.

orgads:
I can link with the static libcurl, but some libraries don't behave well (it's the same with the default `curl`):

`gcc -DIDN2_STATIC -DNGHTTP2_STATICLIB -DCURL_STATICLIB -static -Wl,-Bstatic test.c -lcurl -lidn2 -lssh2 -lzstd -lbrotlidec-static -lbrotlicommon-static -lpsl /mingw64/lib/libunistring.dll.a -liconv -lidn2 /mingw64/lib/libnghttp2.dll.a -lws2_32 -liconv -lz -lbcrypt -ladvapi32 -lcrypt32 -lwldap32`

msink:
Well, maybe unrelated to this PR and should be discussed in separate issue, but do you know what exactly is broken?

Tried this:
```
            compilerOpts = -DIDN2_STATIC -DNGHTTP2_STATICLIB -DCURL_STATICLIB
            linkerOpts(
                "-L${mingwPath.resolve("lib")}",
                "-Wl,-Bstatic",
                "-lstdc++",
                "-static",
                "-lcurl-winssl",
                "-lssh2-wincng",
                "-ladvapi32",
                "-lbcrypt",
                "-lbrotlidec-static",
                "-lbrotlicommon-static",
                "-liconv",
                "-lcrypt32",
                "-lidn2",
                "-lpsl",
                "-lunistring",
                "-lwldap32",
                "-lws2_32",
                "-lz",
                "-lzstd"
            )
```

output:
```
ld: f:\msys64\mingw64\lib\libidn2.a(lookup.o):(.text+0x709): undefined reference to `__imp_uninorm_nfc'
ld: f:\msys64\mingw64\lib\libidn2.a(idna.o):(.text+0xaf): undefined reference to `__imp_uninorm_nfc'
ld: f:\msys64\mingw64\lib\libidn2.a(idna.o):(.text+0x38b): undefined reference to `__imp_UC_CATEGORY_M'
ld: f:\msys64\mingw64\lib\libidn2.a(idna.o):(.text+0x3c0): undefined reference to `__imp_uninorm_nfc'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x1173): undefined reference to `libiconv_open'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x11e0): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x1200): undefined reference to `libiconv_close'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x128b): undefined reference to `__imp_uninorm_nfkc'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x12e1): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x1307): undefined reference to `__imp_uninorm_nfkc'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0xcb): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x3e4): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x5a8): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x6de): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x707): undefined reference to `libiconv'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x86e): more undefined references to `libiconv' follow
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1607): undefined reference to `libiconv_open'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x165b): undefined reference to `libiconv_open'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1703): undefined reference to `libiconv_open'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1723): undefined reference to `libiconv_close'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1731): undefined reference to `libiconv_close'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1756): undefined reference to `libiconv_close'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1785): undefined reference to `libiconv_close'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x1798): undefined reference to `libiconv_close'
ld: f:\msys64\mingw64\lib\libunistring.a(striconveh.o):(.text+0x17ac): more undefined references to `libiconv_close' follow
```
Weird why it didn't found `libiconv`.
And what is `__imp_uninorm_*`...

orgads:
The order matters. Looks like `iconv` should come after `unistring`. And the `__imp_uninorm_*` references mean that the `unistring` references had `__declspec(dllimport)` when `psl` was compiled.

jeremyd2019:
Kind of a moving target, but gnutls should now be available for clang (at least clang64, clang32 could build it if it tried, and I bet clangarm64 could too with the addition of a configure option to not use assembly)

mingw-w64-curl/PKGBUILD
```diff
@@ -126,10 +118,59 @@ build() {
...
+package_curl-gnutls() {
+  provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+  depends=("${depends[@]}"
+           "${_cert_depends[@]}"
+           "${_gnutls_depends[@]}")
+  conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+             "${MINGW_PACKAGE_PREFIX}-${_realname}-winssl")
+  if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]]; then
+    do_package -gnutls
+  fi
```
jeremyd2019, line 164: This one wasn't necessary before, as the package function would not be called if the corresponding package wasn't listed in `pkgname`
```suggestion
  do_package -gnutls
```

jeremyd2019 (commented):
gnutls and rtmpdump are now in staging for clang32, so it is now safe to remove the `*-clang-*` exclusion for the `-gnutls` variant.  clangarm64 is also built.

jeremyd2019:
BTW, if you rebase to master you'll get a CLANG32 CI job to verify that works.

.ci/ci-build.sh
jeremyd2019: I'm concerned about this in the general case (in packages other than those in this PR, since this changes the CI script for everybody) when the packages in a CI run depend on each other, this could fail if the depending package is installed before the package it depends on (mingw-w64-clang depends on mingw-w64-llvm of the matching version, and clang sorts before llvm in a glob).

orgads:
Deon

.ci/ci-build.sh
orgads: Changed to to install sequentially only when a file named `.ci-sequential` exists in the package directory. Do you approve?

jeremyd2019 (approved):
the changes LGTM technically, but I have no opinion as to their desirability

shithappens2016:
@jeremyd2019
please also fix this #8774. it's been there really long time. and it's really annoying.

> libcurl.a should link against libnghttp2.a not libnghttp2.dll.a
libidn2.a should link against libunistring.a not libunistring.dll.a
libpsl.a should link against libunistring.a not libunistring.dll.a

orgads:
Time to merge?

msink:
@jeremyd2019 ability to statically link libcurl to single executable without any .dll and certificates folder dependencies is desired for years: #436 #823 #950 #3617

msink:
Thinking it over once more - maybe better choice was to leave packages `mingw-w64-curl` and `mingw-w64-libssh2` (almost) as is, and just clone them to completely new packages `mingw-w64-curl-winssl`, `mingw-w64-curl-gnutls` and `mingw-w64-libssh2-wincng`, in separate folders and their own `PKGBUILD`.
It would be less potentially breaing change for build infrastructure.
But not sure here.

orgads:
But all the build logic and version number is exactly the same. What's the point in duplicating it? This will require the maintainers to remember to update 3 different packages on every new release, or when the build script should change.

What's the advantage?

msink:
Well, as it done now - it have to change something `.ci` in not tested and potentially breaking way.
With cloned packages - no need to change `.ci`.
But again - I'm not sure it's good idea.

jeremyd2019:
I have no objection to this, but it's complicated enough that I leave it to a more experienced member/collaborator to review and merge.

orgads:
ping? 7.77 is released...

orgads:
Rebased and adapted to static/shared separation (hopefully).

lazka:
thanks, I'll have another look shortly

orgads:
ping?

orgads:
Hello?

orgads:
@lazka? Anybody else?

PR #8475, opened 2021-04-26
odt2txt: fix messed up depends

jeremyd2019:
was not using ${MINGW_PACKAGE_PREFIX}


jeremyd2019:
I wasn't sure if this needed a pkgrel bump or not

Biswa96:
Anything that makes changes in resultant package needs a pkgrel bump. So, yes in this case. But the `mingw-w64-${CARCH}` looks similar with `${MINGW_PACKAGE_PREFIX}`.

PR #8479, opened 2021-04-26
clang: enable sanitizers on clang prefix, and large-address-aware on 32-bit

jeremyd2019:
Fixes msys2/CLANG-packages#37.  May help msys2/CLANG-packages#34


jeremyd2019:
I don't know if there are any more clang changes that might be needed

PR #8480, opened 2021-04-26
remove clang64 from binutils, it's likely to conflict

jeremyd2019:
Might need to remove it from the repo too, if it make it there.


lazka:
I've removed it from the repo

PR #8482, opened 2021-04-26
libssp: new package.

jeremyd2019:
Provides gcc's libssp for clang prefixes.  Fixes msys2/CLANG-packages#40

Borrowed heavily from llvm-mingw.  Thanks to @mstorsjo!

My thought is that gcc-libs should provide libssp, and then packages that need it can depend on libssp (or makedepend if they link it statically).

gcc prefixes blow up trying to install the package as it conflicts with gcc files.  Don't know what if anything I can do about that.  I tried making it conflict with gcc-libs, but libc++ provides that so it conflicted with libc++ ☹️

### DIFF
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -28,7 +28,7 @@ pkgver=10.2.0
 #_majorver=${pkgver:0:1}
 #_sourcedir=${_realname}-${_majorver}-${_snapshot}
 _sourcedir=${_realname}-${pkgver}
-pkgrel=10
+pkgrel=11
 pkgdesc="GCC for the MinGW-w64"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
@@ -288,6 +288,7 @@ package_gcc-libs() {
            "${MINGW_PACKAGE_PREFIX}-mpc"
            "${MINGW_PACKAGE_PREFIX}-mpfr"
            "${MINGW_PACKAGE_PREFIX}-libwinpthread")
+  provides=("${MINGW_PACKAGE_PREFIX}-libssp")
 
   # Licensing information
 
--- a/mingw-w64-libssp/PKGBUILD
+++ b/mingw-w64-libssp/PKGBUILD
@@ -1,0 +1,59 @@
+# Maintainer: Jeremy Drake <github@jdrake.com>
+# borrowed heavily from https://github.com/mstorsjo/llvm-mingw/blob/22ef26848995f2a3ccd0f1c13a2fcff60613a54b/build-libssp.sh
+
+_realname=libssp
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=7.3.0
+pkgrel=1
+pkgdesc="A library for stack smashing protection (mingw-w64)"
+arch=('any')
+mingw_arch=('clang32' 'clang64' 'clangarm64')
+url="https://gcc.gnu.org"
+license=('LGPL' 'custom')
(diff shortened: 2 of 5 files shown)


jeremyd2019:
Feedback from people with more packaging experience would be appreciated.

revelator:
i rewrote the clang PKGBUILD to build libssp as part of the toolchain, it will only install in /clang* prefix now :)

jeremyd2019:
That's another option, thoughts on which approach would be better?

revelator:
huh are those failures from my updated PKGBUILD ?!?.

Btw a list of things the toolchain would benefit from >
-fPIC < does not and newer has worked on windows, unfortunatly a ton of sources still define it and gcc just ignores it heh.
-rpath < well not so sure about this one but clang will error out if it is defined, if i remember correctly it hardlinks the runtime path's to the library.
--no-undefined < error error error!!! strangely removing it from sources which gcc cannot link dynamically without it set works just fine with clang so huh ?.
and the aforementioned __GNUC__ etc.

had to literally butcher the sources to get around these which probably means i have a ton of packages now which you might be missing, the downside is if i have to make these reliable for building with both gcc and clang it would take a massive rewrite of the PKGBUILD files.

jeremyd2019:
If you are talking about a commit I made that I @ mentioned you on, yeah, the way you quoted in pkgname didn't work out for non-clangprefix.  I've got another commit running CI now.  I think I see a couple more tweaks to do though.  That's what I see as the downside of including it in the clang PKGBUILD: it takes soo long to build already, it doesn't need any more things included in it that could reasonably be separated.

revelator:
must be a ci thing i can build it no problem with my local compiler for those arch's.

Yeah it is a huge package to build not sure if it would help any splitting it into a seperate package though, unless there is some way to limit that package to just /clang*

jeremyd2019:
That's what the new `mingw_arch` variable is supposed to do, iirc

revelator:
AH ok :)

lazka:
> My thought is that gcc-libs should provide libssp, and then packages that need it can depend on libssp (or makedepend if they link it statically).

Sounds good.

lazka:
> My thought is that gcc-libs should provide libssp, and then packages that need it can depend on libssp (or makedepend if they link it statically).

Sounds good.

revelator:
you mean link to a gcc built libssp statically or ?.
not sure that would work as the gcc builds are linked with msvcrt and clang uses ucrt so it might include stuff we cannot use.

jeremyd2019:
> > My thought is that gcc-libs should provide libssp, and then packages that need it can depend on libssp (or makedepend if they link it statically).
> 
> Sounds good.

Does that mean I should continue pursuing the separate package here?  Or would you prefer @revelator's approach of merging it into the clang PKGBUILD?

lazka:
Is there any upside to building it with clang?

I meant this separate package is fine and gcc-libs can have a `provides=("${MINGW_PACKAGE_PREFIX}-libssp")`.

jeremyd2019:
> Is there any upside to building it with clang?

That you don't have the CI fail on the non-clang prefixes is all I think...

lazka:
Then I finally need to fix the CI :)

jeremyd2019:
This PR is going to be a CI killer anyway, as GCC won't build under clang, and libssp won't install under non-clang 💣

jeremyd2019:
I've built libssp and xapian-core from this branch on clang32 successfully.  CI is building gcc...

jeremyd2019:
GCC built on the MINGW* and UCRT64, libssp failed due to clang/llvm disappearing from staging... we know it would have failed anyway at install so no big loss there...

jeremyd2019:
So unless you want to fix CI first, this should be ready

PR #8490, opened 2021-04-27
GIFLIB remove -fPIC to allow compiling with clang

ZachBacon:


jeremyd2019:
You'll need to edit the PKGBUILD as well, to update the sha256sums and bump the pkgrel at least

ZachBacon:
I did, but I forgot to include the PKGBUILD. That's what I get for tired coding

PR #8502, opened 2021-04-29
flac: depend on libssp, and work around no MIN/MAX

jeremyd2019:
flac expected that if sys/param.h exists, that it contains macros MIN and
MAX.  mingw-w64's doesn't appear to, so backport a commit that fixes
that assumption.


jeremyd2019:
D'oh... https://github.com/xiph/flac/commit/4a43f2bd4d7e12b5d994f48e18cf014008d204ce
Could backport that instead, probably be nicer

PR #8505, opened 2021-04-29
Akira 0.0.14

albfan:
akira-0 0 14-windows

Packages needed to run Akira on windows 

https://github.com/akiraux/Akira

- Review needed: 
   - Is gtk-update-icon-cache needed? I see in other themes
   - xcursorgen is not available so some cursors are not generated. Should I include a tar with generated sources (they are png files)

### DIFF
--- a/mingw-w64-akira/PKGBUILD
+++ b/mingw-w64-akira/PKGBUILD
@@ -1,0 +1,49 @@
+# Maintainer: Alberto Fanjul Alonso <albfan@gnome.org>
+
+_realname=akira
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.0.14
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64')
+pkgdesc="git repository viewer for GTK+/GNOME (mingw-w64)"
+depends=( "${MINGW_PACKAGE_PREFIX}-granite"
+          "${MINGW_PACKAGE_PREFIX}-elementary-icon-theme"
+          "${MINGW_PACKAGE_PREFIX}-gtk-elementary-theme"
+          "${MINGW_PACKAGE_PREFIX}-libarchive"
+          "${MINGW_PACKAGE_PREFIX}-goocanvas")
+makedepends=("${MINGW_PACKAGE_PREFIX}-meson"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-pkg-config"
+             "${MINGW_PACKAGE_PREFIX}-vala"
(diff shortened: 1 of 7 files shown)


albfan:
@Alexpux some CI task failed due to order building, should I open separate PR?

Biswa96:
It looks like you named the package "goocanvas3" but akira has "goocanvas" in dependencies. Also make sure the package name is similar with ArchLinux or AUR packages. It makes thing easier while comparing versions.

Biswa96:
Also if available, use source tarball instead of git repository.

albfan:
@Biswa96 on themers I need to use git repo as tarballs has symlinks and that do not extract correctly on windows.

goocanvas3 is due to change on release for goocanvas, not sure if we can update goocanvas to version 3 and break some packages depending on goocanvas (version 2.0.4)

Biswa96:
No no, I mean make the package name as "goocanvas" but the version 3, like Arch does. 

Speaking of symbolic links, if the symlinks does not affect the source code or assets then the `noextract` trick can be use. May the grep be with you!

albfan:
I can update that, I see only libgda depending on goocanvas. pkg-config changes to provide goocanvas-3 so probably it will break other packages.

In this case, symbolic links are needed, they are icons like cdrom.png symlinked to optical.png or similar, so symlinks are needed.

Will rework things and recheck again.

About elementary icon themes, I will add a tar of cursors to avoid build xcursorgen, because is an endless package dependency just to turn png into xcursor format

Biswa96:
The `pkgname` and `pkgbase` variables does not interfere with pkg-config. You can use any `pkgname` you want, maybe name it like honolulu :)

albfan:
pkg-config is different, hope this clarifies:

- https://gitlab.gnome.org/GNOME/libgda/-/blob/master/meson.build#L62
- https://github.com/akiraux/Akira/blob/master/meson.build#L21

goocanvas 2.0.4 is goocanvas-2.0 on pkg-config and goocanvas 3.0 and above is goocanvas-3.0 on pkg-config

If we are ok with breaking other packages dependencies great.

We plan some ABI breaks on goocanvas 3.0

lazka:
I've split out the goocanvas update to #8528

lazka:
You can depend on goocanvas v3 now

lazka:
> * Is gtk-update-icon-cache needed? I see in other themes

Yeah, on the theme, and the app if it installs icons.

> * xcursorgen is not available so some cursors are not generated. Should I include a tar with generated sources (they are png files)

Afaik the cursor format is different on Windows anyway, so xcursorgen wouldn't help. (I think adwaita uses this: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/master/src/cursors/anicursorgen.py and ).

albfan:
Anycursongen, cool! Will rework that and finish this

lazka:
(I'm fine with skipping cursors for now..)

albfan:
@lazka I agree this would be better fixed upstream: https://github.com/elementary/icons/issues/1026

We can merge this way by now

PR #8508, opened 2021-04-29
meson: don't pass rpath to gtk-doc on Windows

jeremyd2019:
Unfortunately, meson parses -Wl,-rpath args itself to set the PATH, so I
couldn't keep it from generating them, I had to filter them out after it
did its handling.

This allows graphene and json-glib to build for me on clang64 (which failed in autobuild recently due to this issue)


jeremyd2019:
I don't know if I need to bump pkgrels for packages that are unblocked by this, or if their failure state can be nuked manually.  There are also several packages that failed early on, before the srcinfo was completely done being updated, due to missing dependencies that may now build (at the end of [my gist of notes](https://gist.github.com/jeremyd2019/c445a5312bfe49fb8b0192b10bd5710d), with the exception of freetype)

lazka:
I've cleared all failure states manually now to let it try rebuild everything.

jeremyd2019:
Shook a few more packages out that way, in addition to the ones I expected.

PR #8510, opened 2021-04-29
neon: fix configure test for stdcall function

jeremyd2019:
They were applying __stdcall to an implicitly declared function call.
Instead, include the header and call the function normally.


1480c1:
Perhaps https://github.com/notroj/neon/commit/43a9b7e0d72ee97dcdbc102254ada20278eac80b#diff-0a5b6131a8771bb18b69e97d77cbe4f8a90e8e657ec0a8d43060ab613b14fb55 ?

1480c1:
As a note, if this should be changed, it probably should be improved on upstream's side as well as tbh, I do not know a lot of about writing and modifying autotools past plain bash

jeremyd2019:
Nice, that looks pretty familiar :smile:

PR #8520, opened 2021-04-30
clang: change LLVM_HOST_TRIPLE

### LINKED ISSUE #8517: Wrong default target in `mingw-w64-x86_64-llvm-11.0.0-11`
The default target changed from previous builds of the package, I think starting from `11.0.0-10`, although I'm not 100% on this. Previously it was `x86_64-w64-windows-gnu`, now it's `x86_64-w64-mingw32`.
Running `/mingw64/bin/llc --version` yields:
```
LLVM (http://llvm.org/):
  LLVM version 11.0.0
  Optimized build.
  Default target: x86_64-w64-mingw32
  Host CPU: haswell
```
Getting the default target via `llvm::sys::getDefaultTargetTriple()` also results in `x86_64-w64-mingw32`.

The `mingw-w64-x86_64-clang-11.0.0-11` package doesn't seem to have this problem, running `/mingw64/bin/clang --version` yields:
```
clang version 11.0.0
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:/msys64/mingw64/bin
```

jeremyd2019:
LLVM seems to expect `${CARCH}-w64-windows-gnu`, while `${MINGW_CHOST}` is `${CARCH}-w64-mingw32`

Fixes #8517

It may be better to remove setting LLVM_HOST_TRIPLE altogether, but I want to test a native aarch64 build first (and it'll be a while before my Pi is ready to do that) so proposing this in the meantime.

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -38,7 +38,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
            "${MINGW_PACKAGE_PREFIX}-openmp")
          "${MINGW_PACKAGE_PREFIX}-polly")
 pkgver=11.0.0
-pkgrel=11
+pkgrel=12
 pkgdesc="C language family frontend for LLVM (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
@@ -353,7 +353,7 @@ build() {
     -DLLVM_ENABLE_PROJECTS="${_projects}" \
     -DLLVM_ENABLE_SPHINX=ON \
     -DLLVM_ENABLE_THREADS=ON \
-    -DLLVM_HOST_TRIPLE="${MINGW_CHOST}" \
+    -DLLVM_HOST_TRIPLE="${CARCH}-w64-windows-gnu" \
     -DLLVM_INCLUDE_EXAMPLES=OFF \
     -DLLVM_LINK_LLVM_DYLIB=ON \
     -DLLVM_POLLY_LINK_INTO_TOOLS=OFF \
@@ -409,7 +409,7 @@ build() {
     -DLIBUNWIND_USE_COMPILER_RT=ON
     -DLLVM_ENABLE_LLD=ON
     -DLLVM_ENABLE_PROJECTS="libcxx;libcxxabi;libunwind"
-    -DLLVM_HOST_TRIPLE="${MINGW_CHOST}"
+    -DLLVM_HOST_TRIPLE="${CARCH}-w64-windows-gnu"
     -DPython3_FIND_REGISTRY=NEVER
     -DPython3_ROOT_DIR=${MINGW_PREFIX})
 


revelator:
hmm but is that not what this patch allready does ? -> 0002-Fix-GetHostTriple-for-mingw-w64-in-msys.patch or is that only when detecting mingw gcc.

jeremyd2019:
I'm no great expert, but I think that patches the logic for determining the _default_ host triple, and when I added aarch64 support I set it explicitly (to `$MINGW_CHOST`).  Looking at the whole file, it seems to only know about x86_64 and i686, which seems to be the reason I set it explicitly (I think it would pick `x86_64-w64-windows-gnu` on any Windows where `sizeof(void *) == 8`).

revelator:
my C++ fu is greatly lacking but the cmake module seems to only enable it if mingw is detected as the compiler, question is does it also detect it as mingw when we build the package with the clang* compilers due to the host being set in the target triple ?. Maybe mstorsjo can shed some light.

jeremyd2019:
Pretty sure cmake built with clang still detects as mingw (I think that's a condition built into cmake rather than provided by llvm)

revelator:
ah so the patch might actually overwrite it even if we dont specify the mingw tripple as the target, hmm guess we can only wait and see if the person who complained about it not returning <arch>-w64-windows-gnu still has trouble with it after this.

revelator:
ok it returns arch-w64-windows-gnu after the patch :) so it seems to be ok.

btw i was wondering if chainging this to a warning instead of an error would allow building the stuff that compains about -rpath etc.

in lld MinGW\Driver.cpp

`opt::InputArgList MinGWOptTable::parse(ArrayRef<const char *> argv) {
  unsigned missingIndex;
  unsigned missingCount;

  SmallVector<const char *, 256> vec(argv.data(), argv.data() + argv.size());
  cl::ExpandResponseFiles(saver, getQuotingStyle(), vec);
  opt::InputArgList args = this->ParseArgs(vec, missingIndex, missingCount);

  if (missingCount)
    error(StringRef(args.getArgString(missingIndex)) + ": missing argument");
  for (auto *arg : args.filtered(OPT_UNKNOWN))
    "change this to a warn" -> error("unknown argument: " + arg->getAsString(args));
  return args;
}`

jeremyd2019:
After reading `GetHostTriple.cmake` it seems that the detected `LLVM_HOST_TRIPLE` would only ever be i686 or x86_64, so setting it explicitly would be required for aarch64.

revelator:
because we crossbuild it with a mingw compiler ? well i can see where that might lead to some problems.

jeremyd2019:
> btw i was wondering if chainging this to a warning instead of an error would allow building the stuff that compains about -rpath etc.

It probably would, but would also make it more difficult to notice when options that are significant are not supported.  Recently looked at `--add-stdcall-alias`/`--enable-stdcall-fixup` and still not sure what to do about those options.  Also mentioned was `--disable-dynamicbase` (that one was my fault, I should have looked at what lld had when I made my binutils patch, but adding a disable option for all of those flags would have resulted in `--no-no-seh` and that was more than I could stand).

revelator:
Hmm i wonder if we could just return an empty string after the warning (i guess args is a string value ?) that might atleast skip such unpleasantries i would reckon.

mati865:
Yup, we can drop `0002-Fix-GetHostTriple-for-mingw-w64-in-msys.patch` since we are setting the tripple manually.

> btw i was wondering if chainging this to a warning instead of an error would allow building the stuff that compains about -rpath etc.

That would terribly break build systems because they'd think LLD supports any flag since it didn't return an error.
Ignoring specific options is trivial and doesn't even require changing C++ code.

I make Clang/LLD ignore any flags but we want to be careful since that could cause problems. Autotools check for flags using `--help` and then grepping through the output. If the build system finds the option it's looking for but doesn't really test if it works (staring at autotools again) it may not look for alternative options.

revelator:
Aye have to be carefull i was only contemplating it since this seems to be the option parser and returning an empty string should tell the tools that it is not supported. But if it is trivial to implement i look forward to seing the result :)

PR #8531, opened 2021-05-02
delta: new package

naveen521kk:


Biswa96:
Those `msg` commands, are internal to makepkg, aren't necessary.

PR #8532, opened 2021-05-02
gcc: update to 10.3.0

lazka:


lazka:
(for future reference: some discussion regarding these changes happened in https://github.com/msys2/MINGW-packages/pull/8320)

PR #8535, opened 2021-05-02
freetype: port to meson (using git master)

lazka:


mingw-w64-freetype/PKGBUILD
```diff
@@ -21,67 +21,59 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
...
 fi
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc" "${MINGW_PACKAGE_PREFIX}-pkg-config")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-meson"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-pkg-config")
 options=('strip' '!libtool' 'staticlibs')
-source=(https://downloads.sourceforge.net/sourceforge/freetype/freetype-${pkgver}.tar.xz{,.sig}
+_commit="d911cb5392bd558642fcd6e85d6224618b8e1421"
+source=("git+https://gitlab.freedesktop.org/freetype/freetype.git#commit=${_commit}"
```
naveen521kk, line 30: You can get an archive from GitLab rather than using git?

Biswa96:
BTW, Arch uses tarball from GNU mirror.

naveen521kk:
> BTW, Arch uses tarball from GNU mirror.

I think meson support is something unreleased.

lazka:
yeah, normally I would wait for the next release, but this fixes the clang build as well.

naveen521kk:
Or can we backport the commits that brings meson support?

lazka:
there are about 20 commits..

naveen521kk:
looks like what currently is a lot easier than bringing 20 commits.

revelator:
breaks texlive-bin because of a missing header 
[missing-header.zip](https://github.com/msys2/MINGW-packages/files/6520432/missing-header.zip)

hhromic:
Hi @lazka , this PR unfortunately breaks libfreetype for us at [Solarus Games](https://www.solarus-games.org).

The last working version is `mingw-w64-x86_64-freetype-2.10.4-1-any.pkg.tar.zst`, which was using a stable from upstream.

One of our games uses [this font](https://sourceforge.net/projects/wqy/files/wqy-zenhei/) for Chinese languages which no longer works since `mingw-w64-x86_64-freetype-2.10.4-2-any.pkg.tar.zst` with "Glyph not found" errors from `SDL2_ttf`.

Simply downgrading to the previous stable version mentioned makes it work again.
Freetype is a rather fundamental package, maybe is better to stay at stable versions instad of Git commits?

1480c1:
Could you try using building from upstream and seeing if that breaks it as well, it could be an actual upstream issue that would need to be reported either way

hhromic:
> Could you try using building from upstream and seeing if that breaks it as well, it could be an actual upstream issue that would need to be reported either way

I just re-built `mingw-w64-freetype` using latest `master` (commit `8336d53c`) from upstream and indeed it still breaks :/

While I agree that this should be reported upstream (and I will do so), I also think that MSYS2 should not be shipping broken packages either :(

PR #8538, opened 2021-05-02
libtiff: re-enable libwebp for clang

jeremyd2019:
Now that it's built.


jeremyd2019:
I don't know if some "breadcrumbs" can/should be left for clang32, if that should get this far

PR #8541, opened 2021-05-02
Initial version of package spice-server 0.14.3

heljkon:
Hi,

thanks for your project! I have been working with msys2 since 2 months and I like your distribution. I also like your approach, which allows to get involved easily.

I created this package to run qemu based on msys2 with spice. A few days ago a newer version was released, I will work on it later. Using this package qemu can be run with spice. I use qemu/spice on Linux in my daily work and was happy to see, spice started supporting windows last year.
Sadly there is a bug in qemu, which influences the performance when activating qemu features spice needs.

This is my first package for msys2, but I hope it is acceptable. If there are any issues, I will fix them.

I had to deactivate main parts of test-qxl-parsing. I hope I can fix this in future. For some gstreamer plugins I defined optional dependencies, because they introduce a huge amount of transitional dependencies. As far as I know, the plugins should be usable, if installed

Regards.

### DIFF
--- a/mingw-w64-spice-server/PKGBUILD
+++ b/mingw-w64-spice-server/PKGBUILD
@@ -1,0 +1,87 @@
+_basename=spice
+_realname=${_basename}-server
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.14.3
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64')
+pkgdesc="SPICE: Simple Protocol for Independent Computing Environments"
+license=("LGPLv2.1")
+url="https://www.spice-space.org/"
+depends=(
+             "${MINGW_PACKAGE_PREFIX}-glib2"
+             "${MINGW_PACKAGE_PREFIX}-gstreamer"
+             "${MINGW_PACKAGE_PREFIX}-gst-plugins-base"
+             "${MINGW_PACKAGE_PREFIX}-gst-plugins-good"
+             "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo"
+             "${MINGW_PACKAGE_PREFIX}-lz4"
+             "${MINGW_PACKAGE_PREFIX}-openssl"
+             "${MINGW_PACKAGE_PREFIX}-opus"
+             "${MINGW_PACKAGE_PREFIX}-pixman"
+             "${MINGW_PACKAGE_PREFIX}-spice-protocol"
+             "${MINGW_PACKAGE_PREFIX}-zlib"
+)
+optdepends=(
+             "${MINGW_PACKAGE_PREFIX}-gst-plugins-ugly: for the x264enc GStreamer element"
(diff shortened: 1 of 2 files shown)


neheb:
For faster compilation, you could try the meson build.

heljkon:
Well, I am just using the tool chain provided in the source package. I did not create the source package, so I can't repackage. How should this work?

mingw-w64-spice-server/PKGBUILD
```diff
@@ -0,0 +1,87 @@
+_basename=spice
+_realname=${_basename}-server
```
lazka, line 2: We usually follow the package naming of Arch Linux where this is just called "spice". Could you rename it to that? (also the directory)

lazka:
looks good, thanks

mingw-w64-spice-server/PKGBUILD
```diff
@@ -0,0 +1,87 @@
...
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.14.3
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64')
```
lazka, line 8: change this to `mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')`

heljkon: I can do so.

But I got no clue what ucrt64 and clang64 are, is there any location to read about on your website?

neheb:
> Well, I am just using the tool chain provided in the source package. I did not create the source package, so I can't repackage. How should this work?

hmm? the source contains meson.build.

mingw-w64-spice-server/PKGBUILD
lazka: https://www.msys2.org/docs/environments/

lazka:
meson is preferred over autotools, but we can always change that later.

mingw-w64-spice-server/PKGBUILD
heljkon: Debian calls the components libspice-server-1 and libspice-server-dev

Each component in this package is put to spice-server directories or is called spice-server (include/spice-server, libspice-server-1.dll) by spice project. I'd rather not rename these directories and keep them at expected locations

If it is really necessary, I would rename it to spice and rename the directories I created on packaging (share/doc/spice-server, share/licenses/spice-server).

But I'd rather leave it spice-server, because it is more specific.

lazka: Yeah, I understand. We follow Arch here because it allows us to easily link them up with Arch packages: https://packages.msys2.org/outofdate

> If it is really necessary, I would rename it to spice and rename the directories I created on packaging (share/doc/spice-server, share/licenses/spice-server).

This is what I mean. You can leave the doc/llicenses directories as is if you want, so just the package name.

heljkon: OK, I will do so. And thanks for the explanation.

heljkon:
> hmm? the source contains meson.build.

I just followed the README in source package.

Thanks for pointing to this potential alternative.

PR #8543, opened 2021-05-03
zeromq: update to 4.3.4

neheb:
Added patch to fix compilation with clang.

Added patch stuff from pupnp for ease of patching.

### DIFF
--- a/mingw-w64-zeromq/PKGBUILD
+++ b/mingw-w64-zeromq/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=zeromq
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=4.3.3
+pkgver=4.3.4
 _cppzmqver=4.7.1
 pkgrel=1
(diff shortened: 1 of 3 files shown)


mingw-w64-zeromq/PKGBUILD
```diff
@@ -17,15 +17,28 @@ options=('staticlibs' 'strip')
...
+    patch -Nbp1 -i "${srcdir}/${_patch}"
+  done
+}
 
 prepare() {
   cd ${_realname}-${pkgver}
-  patch -Np1 -i "${srcdir}/001-testutil_different_signedness-fix.patch"
+  apply_patch_with_msg \
+    001-testutil_different_signedness-fix.patch \
+    002-unused.patch
```
lazka, line 41: while we are at it, can you add a "autoreconf -fvi" here so clang works as well?

neheb: What do you mean? It currently compiles.

lazka: If you look at https://github.com/msys2/MINGW-packages/pull/8543/checks?check_run_id=2489384489 under CI-Build -> "[diff]" you can see that the DLL is missing.

PR #8550, opened 2021-05-03
fluidsynth: update to 2.2.0

neheb:
Fix compilation with clang.


Biswa96:
Interesting! Those compiler flags are specifically added for clang.

neheb:
They're not even correct. Those are linker flags that should be matched based on the linker, not the compiler.

Biswa96:
Would you like to upstream your changes for this package and from other pull requests?

lazka:
There is an abi version bump, so needs rebuilds:

````
        mingw-w64-alure
        mingw-w644-gst-plugins-bad
        mingw-w64-openal
        mingw-w64-scummvm
        mingw-w644-SDL2_mixer
        mingw-w64-SDL_mixer
        mingw-w64-vlc
````

neheb:
Seems like scummvm is the issue.

PR #8555, opened 2021-05-03
Manim: update to 0.6.0

naveen521kk:


Biswa96:
1. Those msg commands, are internal to makepkg, aren't necessary.
2. Having a comment in optdepends is preferred.

naveen521kk:
> Those msg commands, are internal to makepkg, aren't necessary.

I have seen them used in many places so reluctant to remove them.

> Having a comment in optdepends is preferred.

Sure

PR #8562, opened 2021-05-04
baobab: update to 40.0

rautesamtr:
Updates baobab to 40.0 and adds patches to fix icon and to build a Windows GUI application instead of a Windows TUI application.


Biswa96:
If the patches are in upstream then it would be helpful to add a little comment above the two `patch` commands.

rautesamtr:
> If the patches are in upstream then it would be helpful to add a little comment above the two `patch` commands.

They aren't. But i already had a PR for the first one and added one for the second patch. I added the corresponding urls as comment to the patch command.

PR #8565, opened 2021-05-04
cairo: remove librsvg dep, fix ucrt build

lazka:
librsvg is only needed for tests which we don't build

### DIFF
--- a/mingw-w64-cairo/0030-ucrt-clang-fixes.patch
+++ b/mingw-w64-cairo/0030-ucrt-clang-fixes.patch
@@ -1,0 +1,11 @@
+--- cairo-1.17.4/src/cairo-compiler-private.h.orig	2021-05-04 17:46:38.073000000 +0200
++++ cairo-1.17.4/src/cairo-compiler-private.h	2021-05-04 17:50:50.920604000 +0200
+@@ -203,7 +203,7 @@
+ #define popen _popen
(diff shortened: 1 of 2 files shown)


naveen521kk:
Also, `gtk-doc` because using meson build with Cairo doesn't use it.

mingw-w64-cairo/0030-ucrt-clang-fixes.patch
```diff
@@ -0,0 +1,11 @@
+--- cairo-1.17.4/src/cairo-compiler-private.h.orig	2021-05-04 17:46:38.073000000 +0200
++++ cairo-1.17.4/src/cairo-compiler-private.h	2021-05-04 17:50:50.920604000 +0200
+@@ -203,7 +203,7 @@
+ #define popen _popen
+ #define strdup _strdup
+ #define unlink _unlink
+-#if _MSC_VER < 1900
++#if (defined(_MSC_VER) && _MSC_VER < 1900) && (defined(__MINGW32__) && !defined(_UCRT))
```
mati865, line 8:
```suggestion
+#if (defined(_MSC_VER) && _MSC_VER < 1900) || (defined(__MINGW32__) && !defined(_UCRT))
```

lazka: ehh

PR #8574, opened 2021-05-05
cairomm, libsigc++3 autoreconf for clang

jeremyd2019:
This (make)depends on new msys2 package 'mm-common' (msys2/MSYS2-packages#2478) to succeed, since that
contains necessary autoconf macros.


Biswa96:
You could add these changes in that PR so that these packages can be built with mm-common.

PR #8579, opened 2021-05-05
harfbuzz: Update to 2.8.1

lazka:


Biswa96:
Any chance if chafa addition will make harfbuzz more attractive? I was looking into chafa project which also depends on imagemagick.

lazka:
> Any chance if chafa addition will make harfbuzz more attractive?

What do you mean by more attractive? I mostly "use" it via gtk.

> I was looking into chafa project which also depends on imagemagick.

Yeah, I tried building it, but it has some unix only parts. Not sure how that would work...

Issue #8567, opened 2021-05-04
Proposal to enhance mingw-w64-qemu


heljkon:
Hi,

recently I prepared the mingw-w64-spice recipe. Of course I did this, because I want qemu to run with spice.

The current state of mingw-w64-qemu is not very good - it doesn't build. 

And even the last version has some issues in my opinion, because everything is just put into /mingw64/bin 
I guess the reason for /mingw64/bin is the strange behaviour of qemu configure. I presuse it works this way, because for windows configure is indented to prepare a windows package - make install puts it all in one dir, make installer uses this dir.

I would propose to create a recipe which would do the following
- Integrate more features by using build deps to activate and optional deps to select by user: this way the features wouldn't cause a huge amount of deps (see https://github.com/msys2/MINGW-packages/issues/8506)
- Use a different base location: put most binary stuff to /mingw64/lib/qemu and provide wrappers in /mingw64/bin, move other stuff to suitable locations in /mingw64/share/
- split the package in several binary packages
- repair the build problems: my local tests work 

This proposal wouldl turn the recipe upside down, which may not make everyone happy:
I would like to hear your opinion before, because I don't want to work for the bin.

Regards

lazka:
Any fix to make it build/update are welcome. Note that we generally try to avoid splitting packages to keep things simple.

lazka:
I'm going to revert qemu to 5.1 for starters, so we have a working base.

heljkon:
> I'm going to revert qemu to 5.1 for starters, so we have a working base.

I would start with 5.2, because I already built it. Never tried 5.1

lazka:
ok, I've create #8568 just to see if/how it works.

PRs welcome.

PR #8580, opened 2021-05-05
mingw-w64-qemu: Enable 5.2.0, spice support, refactor install locations, documentation

heljkon:
As already discussed in https://github.com/msys2/MINGW-packages/issues/8567 here my PR for qemu 5.2.0

**Relocation of installation**
{MINGW_PREFIX}/bin/<QEMU_STUFF>
is splitted into
- {MINGW_PREFIX}/lib/qemu
- 	{MINGW_PREFIX}/share/man
- 	{MINGW_PREFIX}/share/locale
- 	{MINGW_PREFIX}/share/icons
- 	{MINGW_PREFIX}/share/applications
- 	{MINGW_PREFIX}/share/doc
- 	{MINGW_PREFIX}/share/licenses
- 	{MINGW_PREFIX}/share/qemu/firmware

Relevant binaries in 
{MINGW_PREFIX}/lib/qemu 
are invocated by identically named wrappers in
{MINGW_PREFIX}/bin

**New feature**
spice is included and can be used if optional dep spice (adds 150MiB extra) is installed
	
**Package split was not done**
qemu binary package adds 750MiB which includes 250MiB for GDI-doubles (e.g. qemu-system-x86_64.exe and its double qemu-system-x86_64w.exe)

Hopefully the builds pass as expected...

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -1,5 +1,3 @@
-# Maintainer: Martell Malone <martellmalone@gmail.com>
-
 _realname=qemu
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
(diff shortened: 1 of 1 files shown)


lazka:
Thanks.

Just wondering, what are you using, or planing to use, qemu for?

mingw-w64-qemu/PKGBUILD
```diff
@@ -1,94 +1,168 @@
...
+  "${MINGW_PACKAGE_PREFIX}-curl"
+  "${MINGW_PACKAGE_PREFIX}-cyrus-sasl"
+  "${MINGW_PACKAGE_PREFIX}-glib2"
+  "${MINGW_PACKAGE_PREFIX}-gnutls"
+  "${MINGW_PACKAGE_PREFIX}-gtk3"
+  "${MINGW_PACKAGE_PREFIX}-libjpeg"
+  "${MINGW_PACKAGE_PREFIX}-libnfs"
+  "${MINGW_PACKAGE_PREFIX}-libpng"
+  "${MINGW_PACKAGE_PREFIX}-libslirp"
+  "${MINGW_PACKAGE_PREFIX}-libssh"
```
jeremyd2019, line 33: For clang when dependencies are eventually available.

```suggestion
  "${MINGW_PACKAGE_PREFIX}-libssh"
  "${MINGW_PACKAGE_PREFIX}-libssp"
```

mingw-w64-qemu/PKGBUILD
```diff
@@ -1,94 +1,168 @@
...
+  # For executables in ${MINGGW_PREFIX}/lib/qemu create wrappers in ${MINGGW_PREFIX}/bin
+  local P_M_LIBEXES=$(find $P_M_LIB -name "*.exe" ! -name "*w.exe")
+  local P_M_LIBEXE=""
+  for P_M_LIBEXE in $P_M_LIBEXES
+  do
+    local EXE=$(basename $P_M_LIBEXE)
+    local WRAPPER=$P_M_BIN/$EXE
+    echo "#!/bin/bash"                       >> $WRAPPER
+    echo "export PATH=\"$M_LIB/qemu:$PATH\"" >> $WRAPPER
+    echo "exec $M_LIB/qemu/$EXE \$@"         >> $WRAPPER
```
jeremyd2019, line 157:
```suggestion
    echo "exec $M_LIB/qemu/$EXE \"\$@\""         >> $WRAPPER
```

jeremyd2019 (commented):
Couple small nits

heljkon:
Well, personally I won't use it that much right now, because I am working most of the time with Linux.

But I would like to see a good alternative to VirtualBox which I have to deal with - most of my colleagues use this. And in my opinion the combination qemu/spice is better than VirtualBox or Hyper-V - it is better than VirtualBox, because the VMs tend to break on guest addition update and it is better than Hyper-V because of the spice desktop integration.

Spice on Windows just started, so I hope it improves, but this need some interest - in form of contributions.

Moreover my company is heading more and more in Windows/Mac direction because of all the MS and SAP tools.
If I try to contribute now, I may use Qemu/Spice on Windows in the near future.

That all.

mingw-w64-qemu/PKGBUILD
heljkon: Yeah, I search for it while preparing the recipe, but https://packages.msys2.org/base/mingw-w64-libssp shows only clang packages. Won't this break the other builds?

Biswa96:
Just one-liner. Before I had Raspberry Pi 4B, I was using mingw-w64-qemu with Win10 PE* ARM64 to compile and test little programs. We may explore clangarm64 in that environment.

<sup><sub>*due to my low system memory</sub></sup>

mingw-w64-qemu/PKGBUILD
jeremyd2019: it's provided by mingw-w64-gcc-libs too to prevent that (and it's the truth: it does in fact also provide libssp)

mingw-w64-qemu/PKGBUILD
heljkon: Thought about this... couldn't remember the difference.
Could you explain the advantage?

jeremyd2019: a better bash expert than I could weigh in, but my understanding is that `"$@"` expands to the all of the arguments properly quoted, so if you ran the script as `qemu a "b c" d` it'd expand to 3 args not 4

heljkon: Good point, thanks

jeremyd2019:
> Just one-liner. Before I had Raspberry Pi 4B, I was using mingw-w64-qemu with Win10 PE* ARM64 to compile and test little programs. We may explore clangarm64 in that environment.

I tried that, and found it _even slower_ than my raspberry pi 3B+ (amazingly)

mingw-w64-qemu/PKGBUILD
heljkon: Will test right now

mingw-w64-qemu/PKGBUILD
heljkon: Would be nice, but my test shows no difference - is this test OK?

`helge@win10pro MSYS ~
$ cat test.sh
#!/bin/bash
echo with quotes
./test2.sh "${@}"
echo without quotes
./test2.sh ${@}

helge@win10pro MSYS ~
$ cat test2.sh
#!/bin/bash
for A in ${@}
do
        echo $A
done

helge@win10pro MSYS ~
$  ./test.sh a "s d" f
with quotes
a
s
d
f
without quotes
a
s
d
f
`

lazka:
Had  a quick look. Thanks for making it work. The build itself is still a bit of a mess, since configure calls meson, so our auto unix/Windows path conversion breaks various install paths. I think most of the manual moving around isn't actually needed if fixed at that place.

But this is a good start, so let's improve this later.

heljkon:
Hi,

I just did some more tests (using the qemu adventcalendar) and saw failures on VMs on 5.2.0 which did not fail on 5.1.0.

Because I would like to examine these I would like to ask you to postpone the upload of 5.2.0

Sorry for the inconvinience!

Regards.

heljkon:
Finally I finished my testing activity and I found no major regressions which were introduced by this refactored build.

One of my first VMs tested shows minor differences, which I did not interpret right. So I wanted to keep it secure.

I tested more than 80 VMs provided by https://www.qemu-advent-calendar.org and of course my own use cases. 

I'll attach my protocol, so the tests can be reviewed:
[regression-tests.txt](https://github.com/msys2/MINGW-packages/files/6445767/regression-tests.txt)

Biswa96:
@heljkon Why did you add `--cross-prefix` option? qemu compiles fine without that.

heljkon:
It is documented in https://wiki.qemu.org/Hosts/W32 

IMHO it solved the issues building Qemu 5.2.0 which didn't succeed before I introduced it.

Qemu 6.1rc0 is out - I will try to build without. You'll see it in the PR Draft

Biswa96:
Qemu is now using meson and it handles things better than autoconf.

heljkon:
> Qemu is now using meson and it handles things better than autoconf.

I know they introduced it and migrated many tests, but I yet haven't seen it.

The configure still is the skript, which has to be used I guess, but I'll have a look at the weekend I guess...

jeremyd2019:
I was just reading about this, as I've been having a *fun* time with trying to use emulated TPM support (on Linux; I had a quick poke at building swtpm on mingw-w64, but it seems to assume too much unixy stuff... looks like it may work on cygwin though).

I guess they are in the process of migrating to meson, and use a script called `configure` as a wrapper, but I don't think it's autoconf anymore.

heljkon:
AFAIK configure is highly customized for a long time. I do not think this is going to change soon.

PR #8581, opened 2021-05-05
dbus-glib: update to 0.112

neheb:


jeremyd2019:
well, at least the error I saw with clang is showing up everywhere, so it's not some obscure clang-specific thing 😉

neheb:
```
 Compilation failed: Failed to execute child process (Permission denied)
```

Locally this compiles under MinGW64

jeremyd2019:
oh, I guess the error I was seeing was only UCRT64 and CLANG64.

neheb:
added random autoreconf.

Biswa96:
@lazka Would you like to manually build/update this package fix? It can be built without any issue, at least in my setup.

PR #8587, opened 2021-05-06
helics: Update to 2.7.0

nightlark:
Update to the HELICS 2.7.0 release

### DIFF
--- a/mingw-w64-helics/PKGBUILD
+++ b/mingw-w64-helics/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=helics
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.6.1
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Any idea why lgamma_r would be undefined for clang?

nightlark:
No, it works fine on our Linux clang CI builds. Must be something wonky with the clang build environment used by the msys2 CI job.

Biswa96:
> Must be something wonky with the clang build environment

Agreed. Same error with older version of helics.

jeremyd2019:
Well, it's right - lgamma_r is not present in the mingw-w64 libraries, but also not on the mingw64 variant

jeremyd2019:
not in the headers, either, but seems to be referencd from libc++.  Sounds like an upstream bug to me. (either mingw-w64 needs to provide it, or libc++ needs to avoid using it)

PR #8591, opened 2021-05-07
Fix packages

Biswa96:


mingw-w64-tktable/PKGBUILD
```diff
@@ -7,14 +7,15 @@ pkgbase=mingw-w64-${_realname}
...
+pkgrel=5
 pkgdesc="A full-featured 2D table widget for Tk (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 url="https://tktable.sourceforge.io/"
 license=('custom')
 depends=("${MINGW_PACKAGE_PREFIX}-tk")
-source=("https://sourceforge.net/projects/tktable/files/${_realname}/${pkgver}/${_pkgnameactual}${pkgver}.tar.gz"
+#"https://sourceforge.net/projects/tktable/files/${_realname}/${pkgver}/${_pkgnameactual}${pkgver}.tar.gz"
+source=("https://pilotfiber.dl.sourceforge.net/project/tktable/tktable/${pkgver}/Tktable${pkgver}.tar.gz"
```
oscarfv, line 18: Sourceforge automatically chooses a mirror. It is far from optimal, but explicitly pointing to a specific mirror is asking for trouble.

Do you realize that a mirror that worked fine for you when you tried it does not necessarily work well for everyone else the rest of the time?

Biswa96: At least the package can be built. Does the previous link work for you?

oscarfv: Just tried `wget https://sourceforge.net/projects/tktable/files/tktable/2.10/Tktable2.10.tar.gz` and it worked fine.

OTOH, may be we should switch to https://github.com/wjoye/tktable as Debian and did long time ago.

Issue #3039, opened 2017-10-12
mingw-ccache does not work with CCACHE_BASEDIR and full path sources


peterbud:
mingw64 version of `ccache `is not working if `CCACHE_BASEDIR` environment variable is set and source file name contains full path.

Repro:
Turn on logging with 
`ccache --set-config=log_file=/c/msys64/home/<username>/.ccache/ccache.log`

then try to compile something and use full path to the source file:
`ccache gcc -g -Wall -c /c/msys64/home/<username>/test/main.c -o helloworld`

Open the log, you will see something like this:
```
[.520489 37960] === CCACHE 3.3.4 STARTED =========================================
[.520489 37960] Config: (default) base_dir = 
....
[.954004 37960] Result: cache miss
```
Compile again with the same command:
`ccache gcc -g -Wall -c /c/msys64/home/<username>/test/main.c -o helloworld`

Open the log, you will see something like this:
```
[.520489 37960] === CCACHE 3.3.4 STARTED =========================================
[.520489 37960] Config: (default) base_dir = 
....
[.581635 36724] Result: cache hit (direct)
```
So far so good.

Now compile but set `CCACHE_BASEDIR` environment variable before:
`CCACHE_BASEDIR=/c/msys64/home/<username>/test ccache gcc -g -Wall -c /c/msys64/home/<username>/test/main.c -o helloworld`

and it fails, and in the logs you see:
```
[.692203 28232] === CCACHE 3.3.4 STARTED =========================================
[.692203 28232] Config: (environment) base_dir = C:/msys64/home/<username>/test
...
[.692203 28232] Working directory: C:/msys64/home/<username>/test
[.692203 28232] Unsupported source extension: .
[.692203 28232] Failed; falling back to running the real compiler

```
Interestingly, if the source file name does not contain full path, just `main.c`, it works even with `CCACHE_BASEDIR`.

PR #8594, opened 2021-05-07
ccache: Update to 4.2.1

lazka:
@peterbud do you know what `01-directory-separator.patch` fixed? See https://github.com/msys2/MINGW-packages/pull/3477 It doesn't apply and it's not straight forward to rebase without understanding the issue.

I tried your steps in #3039 but they "fail" before and after this change: fail, in that if you set CCACHE_BASEDIR you get a miss, but also get a hit if you do it again.

(if anyone wonders: build with ninja is broken)


peterbud:
Hi @lazka - I'm not following closely ccache development anymore, but I have seen that version 4 is a major re-write, and also I have seen a more active Windows development on the project. In the past the windows version failed in many test cases due to the way it handled or not handled directory separators (like backward and forward slashes), and the patch meant to fix those shortcomings. Honestly, my best advice would be to go without the patch, and see where the tests are failing.

lazka:
Sounds good, thanks!

lazka:
unittests are passing, but the new test system seems linux specific and doesn't really work.. :/

lazka:
hm, this broke ccache on 32bit.. Not output, just doesn't do anything and returns "3"

PR #8595, opened 2021-05-07
libimobiledevice: fix clang compilation

neheb:


Biswa96:
Huge 🌋

lazka:
Did you file this upstream?

neheb:
https://github.com/libimobiledevice/libimobiledevice/pull/1156

Upstream is fairly slow in my experience. In terms of releases at least.

neheb:
https://github.com/libimobiledevice/libirecovery/pull/83

Biswa96:
Instead of storing the huge patches, we could add the raw patch link from those pull request in pkgbuild sources.

neheb:
Not possible with the libirecovery patch. Upstream API changed slightly.

libimobiledevice (the bigger one) is possible I think.

PR #8598, opened 2021-05-07
mingw-w64-mesa: Reproducer for lavapipe build failure with static CRT

pal1000:


pal1000:
[This patch](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/10305) makes it work but I think there is something wrong with the pc files of the packages referred in the patch. There should be no need for Mesa patch.

Mesa3D bug report: https://gitlab.freedesktop.org/mesa/mesa/-/issues/4665

pal1000:
Analyzing the root cause for this I discovered some issues with `${MINGW_PACKAGE_PREFIX}-{libtre,gettext}` packages:
### ${MINGW_PACKAGE_PREFIX}-libsystre
- regex.pc
````
Name: regex
Requires: tre
Libs: -L${libdir} -lregex
````
### ${MINGW_PACKAGE_PREFIX}-libtre
- tre.pc
````
Name: TRE
Libs: -L${libdir} -ltre -pipe -lintl
````
OOPS: Does `-lintl` belongs here or rather it should be in requires field and the uppercase name nay be an issue
### ${MINGW_PACKAGE_PREFIX}-gettext
OOPS: pc files missing, intl.pc needs to advertise `-lintl` according to package contents and `intl` depends on `iconv`
### ${MINGW_PACKAGE_PREFIX}-libiconv
- iconv.pc
````
Name: iconv
Libs: -L${libdir} -liconv
````

MehdiChinoune:
Is this a PR or a bug report?

pal1000:
The discussion with this moved in other places, see #11916. I'll close it.

PR #8603, opened 2021-05-07
ccache: fix compilation with clang

neheb:
Disable testing as it has some linking problem.

### DIFF
--- a/mingw-w64-ccache/010-ucrt.patch
+++ b/mingw-w64-ccache/010-ucrt.patch
@@ -1,0 +1,12 @@
+--- ccache-3.7.9/src/system.h.orig	2020-03-29 16:11:45.000000000 +0200
++++ ccache-3.7.9/src/system.h	2021-05-09 20:11:02.291660400 +0200
+@@ -55,7 +55,9 @@
+ extern int usleep(useconds_t);
+ #endif
+ 
++#ifndef environ
+ extern char **environ;
++#endif
+ 
+ #ifndef ESTALE
+ #define ESTALE -1
--- a/mingw-w64-ccache/PKGBUILD
+++ b/mingw-w64-ccache/PKGBUILD
@@ -5,7 +5,7 @@ _realname=ccache
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.7.9
-pkgrel=1
+pkgrel=2
 pkgdesc="Compiler cache that speeds up recompilation by caching previous compilations (mingw-w64)"
 arch=('any')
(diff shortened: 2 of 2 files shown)


neheb:
```
undefined reference to `__imp___imp__environ'
```

strange.

jeremyd2019:
ccache: C:/_/M/mingw-w64-ccache/src/ccache-4.2.1/src/system.hpp:190:25: error: '__p__environ' redeclared without 'dllimport' attribute: previous 'dllimport' ignored [-Werror,-Winconsistent-dllimport]

I would check that file/line, I think it's doing something stupid there

neheb:
That warning the patch fixes.

jeremyd2019:
https://github.com/ccache/ccache/blob/master/src/system.hpp#L193
`DLLIMPORT extern char** environ;`

But, that shows that `environ` is actually a macro, so it expands into something unexpected

jeremyd2019:
like `#define _environ (* __MINGW_IMP_SYMBOL(_environ))` and then `#define environ _environ`

neheb:
??? seriously?

https://github.com/msys2/MINGW-packages/blob/f15a21c517454c8713ded54e7fc950d4b5cf8d2c/mingw-w64-ccache/010-clang.patch

this macro does not expand into anything as Visual C++ is not being used.

jeremyd2019:
I'm talking about `environ` in mingw-w64... that's a macro, and may or may not expand into something that can be successfully used as a variable name.
such as
https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/stdlib.h#L323
and
https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/stdlib.h#L704

neheb:
sounds like a bug.

Anyway, is there a macro to match against UCRT?

jeremyd2019:
`_UCRT`  but in either (UCRT or MSVCRT) case environ doesn't need to be explicitly declared extern, just include stdlib.h

jeremyd2019:
this patch works for me:
```diff
--- ccache-4.2.1/src/system.hpp.orig    2021-05-07 21:56:05.544685900 -0700
+++ ccache-4.2.1/src/system.hpp 2021-05-07 21:56:19.060320500 -0700
@@ -187,7 +187,9 @@
 #  define PATH_DELIM ":"
 #endif

+#ifndef environ
 DLLIMPORT extern char** environ;
+#endif

 // Work with silly DOS binary open.
 #ifndef O_BINARY
```

neheb:
Builds now.

lazka:
ccache 4 was broken under mingw32 so I had to revert it. rebase accordingly.

PR #8604, opened 2021-05-07
mxml: fix compilation with clang

neheb:


Biswa96:
Give the patch a file a good name that can be understan in future.

neheb:
lld maybe?

Biswa96:
0001-autoconf-remove-options-for-clang.patch

PR #8606, opened 2021-05-07
libmysofa: update to 1.2

neheb:
Add patch to fix clang compilation.

### DIFF
--- a/mingw-w64-libmysofa/004-exports.patch
+++ b/mingw-w64-libmysofa/004-exports.patch
@@ -1,0 +1,131 @@
+--- a/src/hrtf/mysofa.h
++++ b/src/hrtf/mysofa.h
+@@ -28,6 +28,12 @@ extern "C" {
+ #define mylog(...)
+ #endif
+ 
++#ifdef _WIN32
++#define MYSOFA_EXPORT __declspec( dllexport )
++#else
++#define MYSOFA_EXPORT
++#endif
++
+ /** attributes */
+ struct MYSOFA_ATTRIBUTE {
+   struct MYSOFA_ATTRIBUTE *next;
+@@ -126,43 +132,43 @@ enum {
+   MYSOFA_ONLY_SOURCES_WITH_MC_SUPPORTED
+ };
+ 
+-struct MYSOFA_HRTF *mysofa_load(const char *filename, int *err);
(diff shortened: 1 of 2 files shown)


mingw-w64-libmysofa/004-exports.patch
```diff
@@ -0,0 +1,131 @@
+--- a/src/hrtf/mysofa.h
++++ b/src/hrtf/mysofa.h
+@@ -28,6 +28,12 @@ extern "C" {
+ #define mylog(...)
+ #endif
+ 
++#ifdef _WIN32
++#define MYSOFA_EXPORT __declspec( dllexport )
```
jeremyd2019, line 8: If this header is installed, you don't want to use dllexport on all _WIN32 in it, because consumers of the library would want dllimport instead.

neheb: note that MYSOFA_EXPORT is already defined for these functions.

jeremyd2019: Then I think you don't need to do this, cmake should already be generating the proper definition in the mysofa_export.h

neheb: I mean they are defined in the src files but not the header. Meaning that all this patch does is move MYSOFA_EXPORT to the header.

jeremyd2019:
I think CMake needs to be told to install the `mysofa_export.h` header too.  I don't really know about that though.  I *think* it needs to be declared a `PUBLIC_HEADER` (something like https://github.com/dev-cafe/cmake-cookbook/blob/v1.0/chapter-10/recipe-02/cxx-example/src/CMakeLists.txt#L39).

neheb:
It's probably meant to be an internal only header.

jeremyd2019:
I'm no cmake expert, but it seems like the export header needs to be public/installed.  Most of what I found online seems to think so.  Any cmake experts care to weigh in?

jeremyd2019:
see for example https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html installs the generated export header

jeremyd2019:
Updated "suggestion" with patch to CMakeLists.txt to install export header

Biswa96:
Did we reach any conclusion here?

ZachBacon:
I believe so since the branch has been deleted and the pr closed

PR #8614, opened 2021-05-09
faac: fix compilation with lld

neheb:
lld doesn't support these linker flags.

### DIFF
--- a/mingw-w64-faac/010-lld.patch
+++ b/mingw-w64-faac/010-lld.patch
@@ -1,0 +1,12 @@
+--- a/libfaac/Makefile.am
++++ b/libfaac/Makefile.am
+@@ -19,9 +19,4 @@ libfaac_la_LIBADD = ${common_LIBADD}
+ libfaac_la_CFLAGS = ${common_CFLAGS}
+ endif
+ 
+-libfaac_la_LDFLAGS = -no-undefined
+-if MINGW
+-libfaac_la_LDFLAGS += -Wl,--add-stdcall-alias
+-endif
+-
+ AM_CPPFLAGS = -I$(top_srcdir)/include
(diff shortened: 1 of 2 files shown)


mingw-w64-faac/010-lld.patch
```diff
@@ -0,0 +1,12 @@
+--- a/libfaac/Makefile.am
++++ b/libfaac/Makefile.am
+@@ -19,9 +19,4 @@ libfaac_la_LIBADD = ${common_LIBADD}
+ libfaac_la_CFLAGS = ${common_CFLAGS}
+ endif
+ 
+-libfaac_la_LDFLAGS = -no-undefined
```
jeremyd2019, line 7: `-no-undefined` is usually ok, cause libtool interprets that

mingw-w64-faac/010-lld.patch
```diff
@@ -0,0 +1,12 @@
...
+-if MINGW
+-libfaac_la_LDFLAGS += -Wl,--add-stdcall-alias
```
jeremyd2019, line 9: I have been concerned about `--add-stdcall-alias` and its friend `--enable-stdcall-fixup` as these are not no-ops on PE-COFF.  I don't know whether they are required or not (in the case of `--add-stdcall-alias` it would be required by consumers of the library)

mingw-w64-faac/010-lld.patch
neheb: It's easier to just patch out both of them.

mingw-w64-faac/010-lld.patch
lazka: yeah, likely an ABI break, needs to be checked.

lazka:
This happened in #8971

PR #8616, opened 2021-05-09
range-v3: Add mingw_arch.

Biswa96:


Biswa96:
Do I need to increase pkgrel? This package is new.

jeremyd2019:
I don't believe so, when new items were added in bulk pkgrel was not changed.

PR #8618, opened 2021-05-09
igraph: Upgrade to 0.9.3

jannick0:
* PKGBUILD:
  - upgrade igraph version to 0.9.3
  - remove references to obsolete patch (issue covered upstream)
  - use cmake generator 'Ninja' instead of 'MSYS Makefiles' and use
    cmake commands for build and install (instead of calling make)
  - add cmake config -DIGRAPH_ENABLE_TLS=1
  - remove cmake config -DIGRAPH_VERIFY_FINALLY_STACK=1 (devel only)
  - build static and shared library separately, and install one over
    another (where order is important to ensure to have the correct
    header file 'igraph_export.h' working for both static and shared
    linking).
  - add generation of .pc files igraph-shared.pc and igraph-static.pc
  - remove cmake files to avoid confusion when both shared and static
    libraries are installed (implementation assumes that one of them only
    is installed).

* 0001-igraph-0.9.0-link-libm-only-when-needed.patch:
  - remove since obsolete after issue was resolved upstream.

NB: All tests succeeded for MINGW64 locally.


lazka:
side note: since we moved from pkg-config to pkgconf you could now use `Cflags.private: -DIGRAPH_STATIC=1`

PR #8624, opened 2021-05-09
igraph 0.9.3: simplify igraph.pc using Cflags.private

jannick0:


Biswa96:
Is the patch from any upstream pull request? If yes please add the link of that PR as a comment right above the patch command. It helps to track things.

jannick0:
> Is the patch from any upstream pull request? 

No. Consider it as a hotfix on Windows. Highly unlikely that the patch will go into the upstream repo as is.

PR #8628, opened 2021-05-10
pyside2-qt5: Remove install scripts, use ninja.

Biswa96:


Biswa96:
Do we need install scripts here? Replacing `${MINGW_PREFIX}` with `${_IMPORT_PREFIX}` in CMake files does the fix.

Biswa96:
Do we need install scripts here? Replacing `${MINGW_PREFIX}` with `${_IMPORT_PREFIX}` in CMake files does the fix. @lazka

PR #8632, opened 2021-05-10
Entt

fejoa:
A fast and reliable entity-component system.
https://github.com/skypjack/entt


Biswa96:
1. Please add your name as a maintainer.
2. If the project have some test you can add the `check()` function with gtest in `checkdepends` array.

ImperatorS79:
@fejoa Would be great if you could update this.

PR #8634, opened 2021-05-11
emacs: Add upstream patch to fix build with latest mingw-w64-headers-git

yasuhirokimura:
Add upstream patch to fix build with latest mingw-w64-headers-git.


Biswa96:
To big branch name 😟

PR #8636, opened 2021-05-11
libsignal-protocol-c: rebuild

lazka:


Biswa96:
skr-url may need a simple pkgrel bump to trigger msys2-autobuild.

jeremyd2019:
Any idea what the issue was (that only appeared on autobuild and not in CI checks)?

lazka:
my current guess is cmake tends to find msys libm if it is installed.

lazka:
but that's also installed here.. so no idea.

Issue #8640, opened 2021-05-11
Can't build Qt5 with latest Clang 12.0.0


MehdiChinoune:
yesterday, I have pushed some changes #8630 to fix building Qt5 on Clang64, the building completed successfully with clang 11.0.0
This morning after @lazka merged my changes, the building fails with Clang 12.0.0 throwing messages like:
```
lld-link: error: undefined symbol: pthread*
```

See: https://github.com/msys2/msys2-autobuild/commit/4ba4930f7e8893333ebf1d65c04342c7fa128faf/checks/2553346053/logs

mati865:
It's a bug with libc++ configuration, I'll fix it in the evening.

PR #8641, opened 2021-05-11
clang: Force win32 threads for libc++abi

lazka:
This was removed with the update to 12 but is still needed.

See #8640


mati865:
@Astrum-polaris `LIBCXXABI_HAS_WIN32_THREAD_API` has been added for LLVM 13, ofc we can backport it: https://github.com/llvm/llvm-project/commit/05b3716ddbd536c130caa79cf8486f62548f05fc

PR #8643, opened 2021-05-11
Rust: Update to 1.52

peterbud:
Separate rust-docs

@mati865 @Zapeth pls have a look.


mati865:
From Rust point of view this doesn't change anything so it should be safe.

peterbud:
Thanks @mati865 . It's interesting that 1.52.1 is not marked as "latest release". Anyhow now I have bumped to 1.52.1

PR #8644, opened 2021-05-11
clang: patch to add version option to nm.

### LINKED ISSUE #8646: clang: library search path with backslashes confuses libtool
Discovered when testing #8644 against gnutls.  Libtool was not able to find what was necessary to build a DLL for gnutls, while it could with clang-11.0.0-12.  It seems that the library search path extracted from clang-12.0.0-1 has some paths with backslashes in them, while clang-11.0.0-12 has all forward slashes.  The presence of backslashes seems to confuse libtool into not being able to find the import libraries its looking for.

jeremyd2019:
Fixes msys2/CLANG-packages#39

gnutls is having another issue now with LLVM 12 (which I'm attempting to track down now), but this does fix libidn which had the same nm-parsing issue.

The new gnutls issue is #8646, and the second commit on this PR fixes #8646.


jeremyd2019:
Confirmed that this fixes gnutls and libidn.

PR #8648, opened 2021-05-12
Backport patch for libc++abi

Astrum-polaris:


lazka:
I'll wait a bit with this in case there are more fixes

mati865:
I was thinking about opening draft PR this weekend to collect backports/tweaks and wait week or two to merge them.
This patch fixes some of the warnings (which is always nice) but I don't think it's worth to rebuild such heavy package for this purpose only.

mingw-w64-clang/PKGBUILD
```diff
@@ -204,8 +210,11 @@ build() {
   if [ "${_compiler}" == "clang" ]; then
     export CC='clang'
     export CXX='clang++'
+    export AR='llvm-ar'
+    export RANLIB='llvm-ranlib'
```
jeremyd2019, line 214: We've discovered elsewhere that setting `AR` and `RANLIB` environment variables no longer seems to affect cmake (at least the msys variant).  Another interesting discovery from that adventure was that the `CMAKE_AR`/`CMAKE_RANLIB` options don't seem to search the path so you have to give a full path to them.  (comments starting from https://github.com/msys2/MSYS2-packages/pull/2294#issuecomment-819330140)

mingw-w64-clang/PKGBUILD
```diff
@@ -249,7 +257,6 @@ build() {
     -DCMAKE_SYSTEM_IGNORE_PATH=/usr/lib \
     -DFFI_INCLUDE_DIR="${FFI_INCLUDE_DIR}" \
     -DLIBCLANG_BUILD_STATIC=ON \
-    -DLIBOMP_FORTRAN_MODULES=$( (( _clangprefix )) && echo "OFF" || echo "ON" ) \
```
jeremyd2019, line 252: I was running into an issue with cmake picking up the msys2 fortran compiler.

Astrum-polaris: The reason why I remove this is `-DLLVM_ENABLE_LIBCXX=ON` pass `-stdlib=libc++` to gfortran and break the configure test
So only one can be enabled at the same time, or need to find a way to filter flags passed to gfortran

mingw-w64-clang/PKGBUILD
Astrum-polaris: This is more like the continuation of https://github.com/msys2/MINGW-packages/pull/8393

I discovered that building clang with itself without lld, the new compiler just segfault on MINGW64/UCRT64 but not MINGW32

I will add `CMAKE_AR`/`CMAKE_RANLIB` as needed and remove the two environment variables

mati865:
Here is my WIP branch, patches have been tested only on the trunk so they might not work with LLVM 12.
There is no point on pulling the changes yet since there will be 2 more fixes incoming and I need to test if it works.

What is the reasoning for `_compiler=clang`? I think it will lead to a problems within GNU based subsystems.

Astrum-polaris:
> What is the reasoning for `_compiler=clang`? I think it will lead to a problems within GNU based subsystems.

I thought it is preferable (or was) to compile clang with itself

we can move `LLVM_ENABLE_LIBCXX` and `COMPILER_RT_USE_BUILTINS_LIBRARY` from `compiler=clang` to `_clangprefix`

`LLVM_ENABLE_LLD` is required to build clang with iteslf on MINGW64 / UCRT64, otherwise the new compiler just segfault on CI, so here is at least one problem with building clang with itself on GNU based subsystems.

I can drop the commit if it is not preferred anyway.

mati865:
> I thought it is preferable (or was) to compile clang with itself

There were hopes to do it in the past but it always caused troubles. I think now when there are LLVM based subsystems there are even less reasons to do it.

LLVM can only produce short import libraries which are teorethically supported by BFD but it has been broken at least for a few years and still nobody fixed it. I'll look later how LLD produces them but I'm convinced they are short import libs.
For the reference BFD produces long import libraries which work nicely in both linkers.

> `LLVM_ENABLE_LLD` is required to build clang with iteslf on MINGW64 / UCRT64, otherwise the new compiler just segfault on CI, so here is at least one problem with building clang with itself on GNU based subsystems.

My gut says it's buggy weak symbols support in BFD but it's hard to tell for sure without stack traces.

> I can drop the commit if it is not preferred anyway.

Yes, we cannot use it right now without breaking other things and I'm not sure we will ever be.
I feel sorry for your time but you can ask about such big changes on our Discord to know whether they are feasible.

Astrum-polaris:
No worries.

This PR is ready to merge

jeremyd2019:
What about the python2 hunk? that seemed like a good idea.  Is python2 workaround really still necessary?
https://github.com/msys2/MINGW-packages/commit/9c62466f1597a1546d0727d0256f6848be80b39e#diff-125e46f56c15d81bb2f9210bdcc4959baf788e9db4c0acfbff0148bf4416815dL410-L415

jeremyd2019:
> I just noticed that I haven't pasted link to my branch: https://github.com/mati865/MINGW-packages/commits/llvm-tweaks.

FYI, I've had [this commit](https://github.com/jeremyd2019/MINGW-packages/commit/485ef233ecc1561d516c5275093dd995d813951c) I've been cherry-picking on my own llvm change branches to enable building clang32 as well.  It's a bit of a hack so I've not proposed it as a PR, but may be helpful for testing.

mati865:
I think this is ready now: https://github.com/mati865/MINGW-packages/commits/llvm-tweaks

There is problem with clang32 however. It failed 3 out of 3 times on the CI:
- https://github.com/mati865/MINGW-packages/actions/runs/848020018
- https://github.com/mati865/MINGW-packages/actions/runs/848995781
- clicking on re-run button removed this log which is a bit shame since it had slightly different backtrace (but still allocation failed)

@Astrum-polaris my branch should apply cleanly onto yours.
How do you think we should merge? You could pull my branch, apply [this patch](https://github.com/Astrum-polaris/MINGW-packages/compare/backport...mati865:llvm-tweaks.patch) or I can open new PR which contains your and mine commits.

jeremyd2019:
> There is problem with clang32 however. It failed 3 out of 3 times on the CI:

☹️ It had been doing pretty well after the large-address-aware flag was added.  Do you think switching to ninja somehow made it worse? (sort of like how switching to -j1 made it better for openssl when I was bootstrapping)

jeremyd2019:
> > There is problem with clang32 however. It failed 3 out of 3 times on the CI:
> 
> Do you think switching to ninja somehow made it worse? (sort of like how switching to -j1 made it better for openssl when I was bootstrapping)

It kind of looks that way.  All ninja builds failed ,while make builds succeeded.  Kind of sucks that being more efficient about building has had the side-effect of making it more likely to run out of resources on 32-bit.

Astrum-polaris:
@mati865 I have merged your branch
Thanks! 🎉

jeremyd2019:
CLANG32 is having issues with Ninja.  I have this change

https://github.com/jeremyd2019/MINGW-packages/compare/f5b784a..80a0768

lazka:
Could `ninja -j1` help? I can't think of why, but maybe worth a try?

jeremyd2019:
Possibly.  -j1 helped openssl

jeremyd2019:
I built this (actually my version of this branch) on arm64 today successfully.  Since there is no clangarm64 CI

lazka:
> Possibly. make -j1 helped openssl. But I don't know that ninja -j1 would be faster than make -jwhatever

it seems to default to 3 for CI, https://github.com/ninja-build/ninja/blob/0a632d11eb1e8b7becbd9756c3e3439777924f6d/src/ninja.cc#L238 so `-j"$(nproc)"` might be an option.

jeremyd2019:
That also seems to be the default for make:
```
cat /etc/makepkg_mingw.conf | grep -- -j
MAKEFLAGS="-j$(($(nproc)+1))"
```

jeremyd2019:
I've got test runs going setting [`CMAKE_BUILD_PARALLEL_LEVEL`](https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html):
* [to 1](https://github.com/jeremyd2019/MINGW-packages/runs/2622765215?check_suite_focus=true)
* [to 2](https://github.com/jeremyd2019/MINGW-packages/runs/2622770641?check_suite_focus=true)

jeremyd2019:
`CMAKE_BUILD_PARALLEL_LEVEL=2` still failed.  `CMAKE_BUILD_PARALLEL_LEVEL=1` hit the 6 hour time limit, so it at least didn't fail in the first 6 hours.  For comparison the last successful job that ran CLANG32 using `MSYS Makefiles` succeeded in 4:39:20, using the default `MAKEFLAGS="-j$(($(nproc)+1))"`

mati865:
`CMAKE_BUILD_PARALLEL_LEVEL=1` didn't even made it to the point where every other build failed (linking libclang-cpp).

jeremyd2019:
I recommend applying the changes to use 'MSYS Makefiles' for clang32 and Ninja everywhere else, and rebasing to pick up the clang32 CI job.

mati865:
> I recommend applying the changes to use 'MSYS Makefiles' for clang32 and Ninja everywhere else

Already done on https://github.com/mati865/MINGW-packages/commits/llvm-tweaks + few more backports.

> rebasing to pick up the clang32 CI job

Damn, builds are already running locally and on CI but without Clang32.

mati865:
Once all CIs at https://github.com/mati865/MINGW-packages/commits/llvm-tweaks pass (keep in mind to manually inspect logs for `clang*` jobs since they are allowed to fail) it can be merged here. There were big changes to the patches so something like `git reset --hard mati865/llvm-tweaks` will be easiest way to handle it.
Then this is ready to ship.

jeremyd2019:
> Once all CIs at https://github.com/mati865/MINGW-packages/commits/llvm-tweaks pass

My tests have shown the new python check job fails spuriously due to the pinned version of llvm in the split packages.  This failure can be safely ignored.  @naveen521kk want to take a look at [that](https://github.com/jeremyd2019/MINGW-packages/runs/2626728644?check_suite_focus=true)?

Astrum-polaris:
I found this at https://llvm.org/docs/GettingStarted.html

> -DLLVM_PARALLEL_LINK_JOBS Set this equal to number of jobs you wish to run simultaneously. This is similar to the -j option used with make, but only for link jobs. This option can only be used with ninja. You may wish to use a very low number of jobs, as this will greatly reduce the amount of memory used during the build process. If you have limited memory, you may wish to set this to 1.

Maybe we should give a try? (`-DLLVM_PARALLEL_LINK_JOBS=1`)

jeremyd2019:
Worth a try anyway

jeremyd2019:
Nope ☹️

mati865:
We have tested that option before adding large address flag and it doesn't help.

mati865:
This is ready to go, all key checks are green.

jeremyd2019 (approved):

jeremyd2019:
I don't really want to delay this any more, but I just noticed the [windres wrapper](https://github.com/mstorsjo/llvm-mingw/blob/master/wrappers/windres-wrapper.c) had a couple of updates since the version included here.

https://github.com/jeremyd2019/MINGW-packages/commit/b307c032421fae0ac653bd99a95260c3dc3ee0a8

jeremyd2019:
I think something 'funny' happened to the commit history when you merged my commit in, all the prior commits are now doubled

jeremyd2019 (approved):

PR #8654, opened 2021-05-13
gnutls: update to 3.7.1

neheb:
Add two patches fixing compilation with clang.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

PR #8656, opened 2021-05-13
gnutls: bump to 3.7.1

jeremyd2019:
https://lists.gnupg.org/pipermail/gnutls-help/2021-March/004698.html

Notably a (low-severity) security fix, and a performance improvement for W32.

Extracted from #8654


neheb:
I would remove patch 0001 as it fixes a bug no longer present.

PR #8666, opened 2021-05-13
package fixes for clang

jeremyd2019:
Several more were turned up by the last autobuild run


jeremyd2019:
libmicrohttpd additionally seems to try to use dlltool in a way that llvm doesn't accept.

PR #8667, opened 2021-05-14
Fix building qemu

ZachBacon:
since the last author didn't use "${pkgdir}"/${MINGW_PREFIX}
for the readme he added or the shell script, and this caused the build to fail


Biswa96:
1. No need to add `/` before $MINGW-PREFIX, it already has.
2. _By the law of pacman_, this changes the packages, so need a pkgrel bump. But we can skip that here, can't we?

jeremyd2019:
? The cwd should have been `${pkgdir}/${MINGW_PREFIX}` at the time

ZachBacon:
> 1. No need to add `/` before $MINGW-PREFIX, it already has.
> 2. _By the law of pacman_, this changes the packages, so need a pkgrel bump. But we can skip that here, can't we?

I just got that from another pkgbuild and I didn't think it was necessary to bump the rel since it didn't build in the first place, but I can do that if I need to. 

Which I'm assuming probably a yes.

jeremyd2019:
Yes, without manual intervention autobuild will not retry building a package unless it has a newer version (ie, pkgrel bump)

ZachBacon:
> Yes, without manual intervention autobuild will not retry building a package unless it has a newer version (ie, pkgrel bump)

I'll do it in the morning, it's already past midnight and I have work tomorrow.

heljkon:
> since the last author didn't use "${pkgdir}"/${MINGW_PREFIX}
> for the readme he added or the shell script, and this caused the build to fail

Sorry for the problems you have experienced

I do not understand why this fails because currenty working dir is "${pkgdir}"/${MINGW_PREFIX}. I am using relative paths for readablity. 
Since I had already issues with transition to Sphinx 4 the problem could be found there, too. I registered a huge amount of changes...

Nevertheless I am also preparing a PR for other issues I recognized yesterday which I already implemented. I have to start work now, so I will do this later this day. Then of course I will build once more locally before issueing the PR. My local build a few hours ago was successful.

For the sake of not bumping every other day, let's solve this in my PR to come.

Regards

heljkon:
OK, you are referring to pending updates, I guess.
I have got no insight into logs of these as far as I know. 
Since the CI builds for my update PR were OK, it was merged.
It would be interesting to see, why the builds for repo updates failed. Are there any logs?

jeremyd2019:
View the logs from https://packages.msys2.org/queue on the row for mingw-w64-qemu

heljkon:
Thanks.

That's it.

```
2021-05-13T21:20:26.1724742Z cp: cannot create regular file 'share/doc/qemu/': No such file or directory
2021-05-13T21:20:26.2235717Z ==> ERROR: A failure occurred in package().
```

A `mkdir -p share/doc/qemu` should do

But I guess the problem arises much earlier. For some reason share/doc/qemu/ was not created by `make install`. No documentations was put there although is was configured to do so (Doc directory: D:/a/_temp/msys/msys64/ucrt64/share/doc/qemu).

In my opinion more updates like transition to Sphinx may have caused this. There is more work to be done.

lazka:
I assume this is fixed already

PR #8669, opened 2021-05-14
gtkwave: do not install ghwdump

umarcor:
As discussed in https://github.com/ghdl/ghdl/issues/1756, recently ghwdump was added to the GHDL package. The same binary was provided by package GTKWave already. Therefore, there is now a conflict that prevents users of these packages from updating their MSYS2 installations.

After getting feedback from GHDL and GTKWave maintainers and contributors, this PR removes the ghwdump binary from the GTKWave package.


Biswa96:
Just thoughts. If some projects conflict with a simple executable file one can rename the exe (like project1-foo.exe and project2-foo.exe) rather than removing it. After asking upstream, of course.

umarcor:
@Biswa96, see https://github.com/ghdl/ghdl/issues/1756.

PR #8670, opened 2021-05-14
CMake: Build cmake-gui for clang64 and ucrt

MehdiChinoune:
Enable Qt GUI on clang64 and ucrt

### DIFF
--- a/.ci/ci-check.py
+++ b/.ci/ci-check.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__file__)
 
 ARTIFACTS_LOCATION = Path("C:/_/artifacts")
 RDEPS_REGEX = re.compile(
-    r"Required By\s*:\s*(?P<rdeps>[\s\-_A-za-z1-9]*)Optional For\s*:\s*None"
+    r"Required By\s*:\s*(?P<rdeps>[\s\S]*)Optional For\s*:\s*"
 )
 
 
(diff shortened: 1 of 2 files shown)


jeremyd2019:
We still have clang32 and clangarm64 which have not gotten qt5

MehdiChinoune:
> We still have clang32 and clangarm64 which have not gotten qt5

You can update it later.

MehdiChinoune:
- Fix default generator
cmake will default to "Unix Makefiles"

lazka:
> > We still have clang32 and clangarm64 which have not gotten qt5
> 
> You can update it later.

Ideally this PR wouldn't break it in the first place.

jeremyd2019:
>   cmake will default to "Unix Makefiles"

Is this the most sensible default?  It seems like if any makefiles would be default, "MSYS Makefiles" would be the one to use.

1480c1:
> > cmake will default to "Unix Makefiles"
> 
> Is this the most sensible default? It seems like if any makefiles would be default, "MSYS Makefiles" would be the one to use.

`MSYS Makefiles` is supposed to be for msys provided from MinGW and `MinGW Makefiles` is supposed to be for `mingw32-make` provided by MinGW, `Unix Makefiles` works fine for `make`, `Ninja` is a bit preferable though.

As a personal note for when actually building with cmake, I prefer to use just `cmake --build $builddir` as the user can set `CMAKE_GENERATOR` to their own and that will override the default (in the case of no `-G`) and assuming one way or the other without explicit `-G` can break `PKGBUILD`s, as it does for me on manjaro

MehdiChinoune:
> >   cmake will default to "Unix Makefiles"
> 
> Is this the most sensible default?  It seems like if any makefiles would be default, "MSYS Makefiles" would be the one to use.

That needs more work.

lazka:
>  cmake will default to "Unix Makefiles"

What did it use before? We don't use "Unix Makefiles"

1480c1:
> > cmake will default to "Unix Makefiles"
> 
> What did it use before? We don't use "Unix Makefiles"

cmake defaults to checking visual studio installs, then NMake if vs isn't there for windows

lazka:
I would either leave it as is or switch to ninja.

PR #8672, opened 2021-05-14
libass: Update to 0.15.1

lifegpc:


Biswa96:
Agreed, harfbuzz is now a hard dependency with this upstream change https://github.com/libass/libass/commit/f3e2c97

PR #8675, opened 2021-05-14
[CI] export CC and CXX variables.

MehdiChinoune:
It doesn't change much, but at least it will show explicitly the used compiler instead of the alias.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,11 +12,14 @@ jobs:
       fail-fast: false
       matrix:
         include: [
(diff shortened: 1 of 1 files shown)


Biswa96:
But some packages set those environment variable if they required, not for all.

lazka:
The PKGBUILDs should be self contained and not depend on the user setting the right env vars.

MehdiChinoune:
I will push tomorrow a pull-request to MSYS-packages to export these variables from makepkg_mingw.conf file.

lazka:
That might be worth a discussion, yeah.

PR #8676, opened 2021-05-14
add scopehal-apps

umarcor:
This PR adds [azonenberg/scopehal-apps](https://github.com/azonenberg/scopehal-apps): glscopeclient and other client applications for libscopehal. [azonenberg/scopehal](https://github.com/azonenberg/scopehal) is an oscilloscope / logic analyzer platform abstraction library. scopehal-apps depends on [anthonix/ffts](https://github.com/anthonix/ffts). Therefore, in this PR package ffts is added too.

Build instructions on MSYS2 are provided in https://www.antikernel.net/temp/glscopeclient-manual.pdf (sec 3.4.2).

I'm unsure about some decisions, so I'll mark this as a draft for now.

@azonenberg would you mind helping me answer the following questions?

- Are the `depends` and `makedepends` lists of scopehal-apps correct?
- ~~Why is FTTS built with `DENABLE_SHARED=ON`, if scopehal-apps uses the static file?~~ (azonenberg/scopehal-docs#28)
- ~~The `lib/scopeprotocols/kernels` shown in step 6. a) of the  scopehal-apps documentation seems not to exist?~~
- ~~Would it be possible to copy all the exes/dlls as part of `make install`, instead of having to copy them manually afterwards?~~ azonenberg/scopehal-apps#355
- ~~Is it worth distributing scopehal as a package? Currently, this PR includes a package for ffts and a package for scopehal-apps.~~ https://github.com/msys2/MINGW-packages/pull/8676#issuecomment-842185936

### DIFF
--- a/mingw-w64-ffts/PKGBUILD
+++ b/mingw-w64-ffts/PKGBUILD
@@ -1,0 +1,51 @@
+# Maintainer: umarcor <unai.martinezcorral@ehu.eus>
+
+_realname=ffts
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.0.0.r799.gfe86885
+pkgrel=1
+pkgdesc="ffts: The Fastest Fourier Transform in the South  (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
+url="https://github.com/anthonix/ffts"
+license=('custom')
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+  "${MINGW_PACKAGE_PREFIX}-cmake"
+)
+
+_commit='fe86885e'
+source=("${_realname}::git://github.com/anthonix/ffts.git#commit=${_commit}")
+sha256sums=('SKIP')
+
+pkgver() {
+  cd "${_realname}"
+  echo '0.0.0.r'"$(git rev-list --count HEAD)"'.g'"$(git describe --all --long | sed 's/^.*-g\(.*\)/\1/')"
+}
+
+build() {
+  cd "${srcdir}/${_realname}"
+  mkdir build
+  cd build
+  MSYS2_ARG_CONV_EXCL=-DCMAKE_INSTALL_PREFIX= cmake \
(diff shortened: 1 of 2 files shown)


azonenberg:
libcopehal and libscopeprotocols should be part of the glscopeclient package until version 1.0 to discourage independent use before the ABI is stable.

After v1.0 we will be maintaining ABI compatibility within each major version of the library, so the libraries should be packaged independently.

azonenberg:
Copying the files as part of "make install" is blocked on https://github.com/azonenberg/scopehal-apps/issues/142.

umarcor:
I think this PR is ready to merge. Note that the MINGW32 job fails because scopehal-apps is for 64 bits only.

@azonenberg, when executing glscopeclient, I see the following:

```sh
GetConsoleScreenBufferInfo() failed (6);
Warning: glscopeclient works best with the OMP_WAIT_POLICY environment variable
set to PASSIVE
    Warning: AVX2/AVX512 detected but disabled on MinGW64/GCC (see https://githu
    b.com/azonenberg/scopehal-apps/issues/295)

```

At exit:

```sh
clFFT support: not present at compile time
```

I guess those warnings are acceptable?

azonenberg:
Not sure what the deal about GetConsoleScreenBuffer is, but it should be harmless. I'll have to see who wrote that part of the logtools code and check.

The OMP_WAIT_POLICY warning is expected. It's an environment variable that unfortunately can't be controlled any other way, so the user is expected to set it. On Linux we setenv() and exec() to force it on if it's not set  right, but nobody's written the code to do that on Windows so you get the warning.

We've experienced crashes with AVX2/AVX512 instructions passing vectors across function boundaries on Windows, so it's disabled until we get to the bottom of the issue. For now it will run slower than on Linux, but it should behave correctly.

The lack of clFFT support is expected as well. FFTs will run slower without GPU acceleration being available, but everything should function correctly.

jeremyd2019:
The AVX issue may be #8317

ffts fails to  build on clang64 with `error: always_inline function '_mm_moveldup_ps' requires target feature 'sse3', but would be inlined into function 'ffts_execute_1d_real' that is compiled without support for 'sse3'`.  I wonder if that might cause alignment issues too in gcc which doesn't seem to error for that

PR #8677, opened 2021-05-14
mingw-w64-qemu: Fix build with sphinx 4.0.1 and other fixes/enhancements

heljkon:
Build of qemu-6.0.0 fails on staging build because of switch to sphinx 4.0.1:

doc generation fails, because qemu-6.0.0 meson.build can't handle sphinx 4.0.1

qemu build fails because directory share/doc/qemu not created by doc generation

This is fixed by creating each dir in package with mkdir -p before moving/copying.

Other fixes are included as well.

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -4,7 +4,7 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 _base_ver=6.0.0
 _rc=
 pkgver=${_base_ver}${_rc//-/.}
-pkgrel=1
+pkgrel=2
 pkgdesc="QEMU - a generic and open source machine emulator and virtualizer (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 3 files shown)


jeremyd2019:
I wonder if the sphinx thing could be fixed by something like https://github.com/msys2/MINGW-packages/pull/8670/files#diff-dd78400f2ef7470d57de1fbd7ad21887c1907fd3c27c5efe311f82381ed34fd0

heljkon:
Thanks, I will try to fix it. Maybe this helps.
Honestly it is not clear, if Sphinx would really fail. configure tests something using meson and deactivates doc build.

I'll test it locally...

jeremyd2019:
Yeah I have no idea about sphinx or meson... just saw you were moving things around and that option seemed like a way to stop it from putting the man pages into directories.

heljkon:
No, the way to fix the build with Sphinx 4.0.1 is to provide the directories, which were autocreated before (escecially share/doc/qemu).

I moved things around from 5.2.0 to 6.0.0 because I've understood, the build was to complicate. I think now it is better readable

PR #8678, opened 2021-05-14
gimp: Update to 2.10.24

lazka:


Biswa96:
In ArchLinux, gimp does not seem to be dependent on python2 packages.

PR #8680, opened 2021-05-14
emacs: Define _snprintf for ucrt.

Biswa96:
Run make three times for unknown failures. At least it builds. Suggestions are welcomed!


mingw-w64-emacs/PKGBUILD
```diff
@@ -69,6 +65,10 @@ build() {
   # --without-compress-install is needed because we don't have gzip in
   # the mingw binaries.
 
+  # Workaround: make fails with bootstrap-emacs.pdmp and emacs.pdmp
+  # for unknown reason.
+  make || true
+  make || true
   make
```
jeremyd2019, line 72:
```suggestion
  make || make || make
```

??  I've had to do things like that before ☹️

Biswa96:
Hope I am allowed to copy things from mingw-w64! Am I? @lhmouse

lhmouse:
Yes of course.

Biswa96:
Closed! emacs crashes in UCRT. May need special treatment.

PR #8689, opened 2021-05-15
Add a pip check

naveen521kk:
lately, there have been many issues with broken deps for py packages. add a check for that.

I have rebuild isort as a test.


naveen521kk:
This works, but something is broken when installing isort.

naveen521kk:
ready for review

lazka:
https://github.com/msys2/MINGW-packages/pull/8712/checks?check_run_id=2613560448

naveen521kk:
That log is missing?

PR #8690, opened 2021-05-15
Python updates

Schala:
- python-boto3 1.17.73
- python-botocore 1.20.73
- python-click 8
- python-flask 2
- python-gsutil 4.62
- python-itsdangerous 2
- python-werkzeug 2

### DIFF
--- a/mingw-w64-python-boto3/PKGBUILD
+++ b/mingw-w64-python-boto3/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=boto3
 pkgbase=mingw-w64-python-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
-pkgver=1.17.70
+pkgver=1.17.73
 pkgrel=1
 pkgdesc='The AWS SDK for Python (mingw-w64)'
 arch=('any')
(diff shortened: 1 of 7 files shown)


lazka:
the jinja update is missing

naveen521kk:
> the jinja update is missing

irrc, I updated it and it caused the sphinx problem :facepalm:

lazka:
ah, right, missed that.

naveen521kk:
now, https://github.com/msys2/MINGW-packages/pull/8690/commits/174a45b050b6efaf31434224f5ca498e4a53cc0a broke [manim](https://packages.msys2.org/package/mingw-w64-x86_64-manim?repo=mingw64), again :facepalm:

lazka:
the downsides of having a proper dependency resolver in pip, people start to depend on it and use it :)

naveen521kk:
I think getting in https://github.com/msys2/MINGW-packages/pull/8689 should at least inform us which are broken.

jeremyd2019:
It's frustrating when people use it to pin exact versions or set maximum versions without any actual known breakage with the other versions they're blocking though.

naveen521kk:
it's always an expectation that a major version would cause breakage(and IMO it is good practice) and quite some time it does(for example jinja did).

PR #8695, opened 2021-05-16
hdf4: Fix building with gcc 10.

Biswa96:
Add -fallow-argument-mismatch flag for fortran compiler.

### DIFF
--- a/mingw-w64-hdf4/PKGBUILD
+++ b/mingw-w64-hdf4/PKGBUILD
@@ -4,7 +4,7 @@ _realname=hdf
 pkgbase=mingw-w64-${_realname}4
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}4")
 pkgver=4.2.15
-pkgrel=2
+pkgrel=3
 pkgdesc="Legacy HDF4 driver for GDAL (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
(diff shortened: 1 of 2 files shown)


Biswa96:
GHA sleeping!

MehdiChinoune:
> GHA sleeping!

https://www.githubstatus.com/

mingw-w64-hdf4/PKGBUILD
```diff
@@ -33,9 +34,14 @@ prepare() {
 build() {
   [[ -d "${srcdir}"/build-${CARCH} ]] && rm -rf "${srcdir}"/build-${CARCH}
   mkdir -p "${srcdir}"/build-${CARCH} && cd "${srcdir}"/build-${CARCH}
+
+  if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]]; then
+    export FFLAGS="-fallow-argument-mismatch"
+  fi
```
MehdiChinoune, line 40: I don't think there is another mature Fortran compiler to be used other than gfortran.
Why excepting *-clang-*

Biswa96: As far as I have seen, most of the packages don't build in clang folders because clang doesn't have the compiler options like gcc. So, I'm 80% sure that clang+fortran thingy will not have that option. Though we could remove the check in future.

MehdiChinoune: That option is specific to gfortran.

Biswa96: Hence _not_ equal to `*-clang-*`.

PR #8698, opened 2021-05-16
Retro68 toolchain for Mac OS 68k, PPC and Carbon: binutils phase

Schala:


Biswa96:
`pkgver` variable should be deduced by `pkgver()` function. See mingw-w64-headers-git package as reference.

Schala:
There is no .git meta directory though

PR #8699, opened 2021-05-16
Some Python package fixes

lazka:


naveen521kk:
wondering how did it python-colour build at first...?

lazka:
older setuptools downloaded/installed build dependencies if they were missing, newer one does the same, but asks pip instead, which isn't installed in CI.

Issue #8704, opened 2021-05-17
stlink new release


valentinpi:
A new version 1.7.0 is out: [https://github.com/stlink-org/stlink/releases/tag/v1.7.0](https://github.com/stlink-org/stlink/releases/tag/v1.7.0)

Biswa96:
Hi! @valentinpi Thank you for the report. Would you like to test the package from the linked pull request? The packages can be found in the GitHub Actions artifacts in a zip archive.

Biswa96:
Any reply?

valentinpi:
Ah yes so sorry I've been busy, I'll try it in a sec

PR #8708, opened 2021-05-18
stlink: Update to 1.7.0

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/8704


Biswa96:
No replay from @valentinpi, so just see what happen.

PR #8709, opened 2021-05-18
libgit2-glib: stage/unstage correctly on mingw

albfan:
Paths with directories need to translate windows to unix paths to work


Biswa96:
If this is a upstream commit or in a upstream merge request a little comment upon the patch command would be helpful for future maintenance.

albfan:
The upstream patch is still not done, at that moment I will remove this patch and rework the PKGBUILD with proper meson setup

albfan:
@Biswa96 related merge request: https://gitlab.gnome.org/GNOME/libgit2-glib/-/merge_requests/27 created and linked in patch

PR #8714, opened 2021-05-18
ci-check: don't run on rebuilding python itself

naveen521kk:


naveen521kk:
@lazka does these changes look good to you?

PR #8716, opened 2021-05-19
Update mingw-w64

lazka:


jeremyd2019:
I see that extra 'digit' in short revision hash finally showed up

PR #8718, opened 2021-05-19
libirecovery + libidevicerestore: fix with clang64

neheb:


neheb:
Still have no idea why libirecovery didn't fail compilation before this PR.

jeremyd2019:
Probably libtool... it likes to not fail to build, but instead degrade to just building static libraries because Windows doesn't allow undefined symbols in DLLs

Biswa96:
Is this patch in upstream?

neheb:
No. Upstream's not that active.

PR #8733, opened 2021-05-20
libavif: update to 0.9.1

kmilos:


Biswa96:
`git push --force` best for hiding mistakes.

PR #8737, opened 2021-05-20
Update mingw-w64

lazka:
The at_quick_exit issue might be fixed now, needs testing by building gcc once this is in staging.


1480c1:
misspelled mingw-w64 :p

PR #8739, opened 2021-05-20
enable gnutls and rtmpdump for clang32 and clangarm64

jeremyd2019:
Add `--disable-hardware-acceleration` for gnutls on new arches.  That disables the assembly routines which don't yet work for mingw aarch64.

These should be the requirements for `-gnutls` curl, from #8469.

Once this is merged I'll build and provide clangarm64 binaries


jeremyd2019:
Guess I'll make this my inaugural merge

PR #8741, opened 2021-05-21
vulkan: update to 1.2.178

neheb:


mingw-w64-vulkan-loader/002-proper-def-files-for-32bit.patch
```diff
@@ -60,19 +60,6 @@
          target_link_libraries(vulkan OneCoreUAP.lib LIBCMT.LIB LIBCMTD.LIB LIBVCRUNTIME.LIB LIBUCRT.LIB)
          set_target_properties(vulkan PROPERTIES LINK_FLAGS "/NODEFAULTLIB")
      else()
---- Vulkan-Loader-1.2.158.orig/loader/loader.h
```
lazka, line 63: what's this about?

neheb: Breaks compilation with clang as clang expects dllexport in the declaration only.

lazka: this breaks the ABI for 32bit I think, at least the symbols are different here

neheb:
bleh too many.

Biswa96:
I have fixed the issue. I will try to discuss this with upstream. The whole patch is not wrong. The trick is to define `__declspec(dllexport)` at the time of building DLLs _only_. clang is right here. Thanks @mati865 for the hint 😍

mingw-w64-vulkan-loader/002-proper-def-files-for-32bit.patch
Biswa96: The patch file name is somewhat misleading. It does not do anything with 32bit building. Will try upstreaming.

Issue #8749, opened 2021-05-21
python check fails due to pinned package versions


jeremyd2019:
My tests have shown the new python check job fails spuriously due to the pinned version of llvm in the split packages.  This failure can be safely ignored.  @naveen521kk want to take a look at [the logs](https://github.com/msys2/MINGW-packages/pull/8648/checks?check_run_id=2637758425) from #8648?

_Originally posted by @jeremyd2019 in https://github.com/msys2/MINGW-packages/issues/8648#issuecomment-845601678_

naveen521kk:
Maybe simply ignore the failure?

jeremyd2019:
I did a mingw-w64-clang run again and it succeeded:
https://github.com/jeremyd2019/MINGW-packages/runs/2714325583?check_suite_focus=true

PR #8750, opened 2021-05-21
ci-check: don't fail if pacman install failed

naveen521kk:
Fixes https://github.com/msys2/MINGW-packages/issues/8749

### DIFF
--- a/.ci/ci-check.py
+++ b/.ci/ci-check.py
@@ -32,7 +32,7 @@ def convert_win2unix_path(winpath: typing.Union[Path, str]):
     return p.stdout.strip()
 
 
-def install_package(pkg: typing.Union[Path, str], local: bool) -> None:
+def install_package(pkg: typing.List[typing.Union[str, Path]], local: bool) -> None:
     logger.info("Installing %s", pkg)
     if local:
         command = [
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Wouldn't the rest of the check then potentially be invalid?  If it's going to punt, it might make more sense to skip the rest of the check if an install fails.

naveen521kk:
maybe how about running Pacman twice? That should work. At first, it will install all the things it could and then in round 2 it will install things it failed at first.

Also, btw did you PR had python modules? If not it should have ideally skipped them.

jeremyd2019:
> Also, btw did you PR had python modules? If not it should have ideally skipped them.

Yes

jeremyd2019:
> maybe how about running Pacman twice? That should work. At first, it will install all the things it could and then in round 2 it will install things it failed at first.

Maybe.  It seems like in degenerate cases it may need more rounds, or may even never resolve (such as the mutually-conflicting packages from #8469).  For clang/llvm, running pacman to install all packages at once would work, but that would fail for those new curl split packages (though luckily I don't think they have any python).  I don't know of any approach that would work in all cases

naveen521kk:
> I don't know of any approach that would work in all cases

Or how about passing all the packages to Pacman together? Finally, calculate rdeps for each package and not failing if failed to install?

jeremyd2019:
> Or how about passing all the packages to Pacman together?

That would work for pinned deps, or circular deps, but if there are conflicting packages in the group that would fail.
 
> Finally, calculate rdeps for each package and not failing if failed to install?

would you skip the check on install failure?  I think you'd have to if installing all packages in one pacman call, as that would probably mean none had been installed so the check would be meaningless.

naveen521kk:
@lazka I could see https://github.com/msys2/MINGW-packages/commit/e1a97686b4d5c636e42e36d0a726f6421da01989, FYI, `-Qi` only prints the reverse dependency of packages installed, but we need from the whole database, so `-Sii` should be used. I made it so that it doesn't fail on new packages. https://github.com/msys2/MINGW-packages/pull/8750/commits/ac4cfa61ece71fc1b9a03b730daf827bb2613881

lazka:
ah, ok, sry.

lazka:
one possible fix, if I'm not mistaken, is put all packages in a local repo, install/uninstall one by one.

naveen521kk:
I think it still can't fix conflicts and it works the same as passing everything to pacman.

jeremyd2019:
I think it wouldn't "fix" conflicts, but it would "succeed" by automatically uninstalling the package it conflicts with before installing the package.  So you'd wind up with the "last" package of mutually-conflicting packages, rather than a failure to install.  Don't know what's preferable here (with the failure you know something funny's going on and can skip the rest of the check)

naveen521kk:
I think it's simply fine to skip the tests if the installation failed(*usually* python packages we care about don't have that problem).

naveen521kk:
I will try to create a temp repo and working and that seems like an effective idea.

PR #8751, opened 2021-05-22
fix alure 1 build and move alure2 it's own package

### LINKED ISSUE #8738: recent update to the package have appended the number 2 to headers and libraries causing build errors in my other [alure]
### **description : recent update to the package have appended the number 2 to headers and libraries causing build errors in my projects**

```
Files:

    /mingw64/bin/libalure2.dll
    /mingw64/include/AL/alure2-alext.h
    /mingw64/include/AL/alure2-aliases.h
    /mingw64/include/AL/alure2-typeviews.h
    /mingw64/include/AL/alure2.h
    /mingw64/lib/cmake/Alure2/Alure2Config-noconfig.cmake
    /mingw64/lib/cmake/Alure2/Alure2Config.cmake
    /mingw64/lib/libalure2.dll.a
    /mingw64/lib/libalure2_s.a
    /mingw64/share/licenses/alure/LICENSE
```

fundies:
fixes #8738

### DIFF
--- a/mingw-w64-alure/build.patch
+++ b/mingw-w64-alure/build.patch
@@ -1,0 +1,12 @@
+diff -rupN a/src/codec_fluidsynth.cpp b/src/codec_fluidsynth.cpp
+--- a/src/codec_fluidsynth.cpp	2011-07-29 08:37:48.000000000 +0000
++++ b/src/codec_fluidsynth.cpp	2014-01-03 10:42:13.280792235 +0000
+@@ -32,7 +32,7 @@
+ #endif
+ 
+ #include <istream>
+-
++#include <unistd.h>
+ #include <fluidsynth.h>
+ 
+ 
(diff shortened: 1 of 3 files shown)


Biswa96:
Why is ninja replaced with make?

jeremyd2019:
Does the alure package require an 'epoch' in this case?  Or is the downgrade acceptable

Biswa96:
Do we? The pkgver seems to be same.

jeremyd2019:
Nevermind.  🤦

fundies:
> Why is ninja replaced with make?
I only reverted the pkgbuild to before it was changed to build alure2 it originally called make

fundies:
> Does the alure package require an 'epoch' in this case? Or is the downgrade acceptable

I bumped alure to PKG release 4 but 3 was an invalid release and I just set alure2 pkgbuild to 1 because it'd be the first time the package was named properly

lazka:
perfect, thanks!

Issue #7607, opened 2021-01-07
python-distlib: has msvc compiled executables


naveen521kk:
for some reasons, [this](https://files.pythonhosted.org/packages/2f/83/1eba07997b8ba58d92b3e51445d5bf36f9fba9cb8166bcae99b9c3464841/distlib-0.3.1.zip) source archive contains msvc compiled executable, which are use in `pip` .
[source](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-distlib/PKGBUILD) doesn't compile the executable though.

Biswa96:
In ArchLinux's [PKGBUILD][1], there is `rm distlib/*.exe` in prepare step.

[1]: https://github.com/archlinux/svntogit-community/blob/packages/python-distlib/trunk/PKGBUILD

naveen521kk:
Those exe files are useless in Linux but those are used by Pip in windows.

Biswa96:
Where are the source code of those executables? Is it possible to download and compile those source code in build step?

naveen521kk:
It is included inside the tar archive just that we aren't compiling it.

lazka:
We have a similar issue in setuptools. Note that we patch the launcher in setuptools to handle our directory layout, we might need to do something similar here.

@naveen521kk do you know when these executables are installed?

naveen521kk:
> @naveen521kk do you know when these executables are installed?

`pip` uses this library internally, I think, I don't know exactly. What I can link now is this doc [here](https://distlib.readthedocs.io/en/stable/reference.html#the-distlib-scripts-package). I don't know anything more than this.

lazka:
ok, thanks

lazka:
Looks like it's used when installing wheels which have entry points, to create corresponding launchers.

mingwandroid:
These are the 'standard' python launcher executables.

mingwandroid:
On Anaconda Distro on Windows we build these exes (with GCC because well, I did it) and package them with conda-build: https://github.com/conda/conda-build/tree/master/conda_build/launcher_sources

When conda-build then builds a python package with entry points, these are used instead.

In general, on Windows, a Python entry point consists of two parts, <foo>-script.py and <foo>.exe. This is the foo.exe and the name of the excecutable file is used to determine the name of the script file which is then launched through python.

mingwandroid:
[...]

PR #8755, opened 2021-05-23
distlib: build executables from source

naveen521kk:
Possibly it should be done for pip also
Fixes: https://github.com/msys2/MINGW-packages/issues/7607

### DIFF
--- a/mingw-w64-python-distlib/PKGBUILD
+++ b/mingw-w64-python-distlib/PKGBUILD
@@ -1,4 +1,5 @@
 # Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
+# Maintainer: Naveen M K <naven@syrusdark.website>
 
 _realname=distlib
 pkgbase=mingw-w64-python-${_realname}
@@ -7,7 +8,7 @@ provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 pkgver=0.3.1
-pkgrel=1
+pkgrel=2
 pkgdesc="Low-level components of distutils2/packaging (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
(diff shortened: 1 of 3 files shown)


naveen521kk:
@lazka How can we name the launcher for arm?

lazka:
I'd just name it the same for now. Otherwise they wont be used.

mingw-w64-python-distlib/PKGBUILD
```diff
@@ -48,8 +30,39 @@ prepare() {
...
+  local _gui_lflags="-lshlwapi"
+  local _cli_lflags="-lshlwapi"
+  if check_option "strip" "y"; then
+    _gui_cflags+=" -s"
+    _cli_cflags+=" -s"
+  fi
+  case "$MSYSTEM" in
+  MINGW*)
+    _gui_cflags+=" -mcrtdll=msvcr120"
+    _cli_cflags+=" -mcrtdll=msvcr120"
```
lazka, line 47: What's this about?

naveen521kk: it fails without them. https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/dupenv-s-wdupenv-s?view=msvc-160
google got me to https://stackoverflow.com/a/61556667/12858827

naveen521kk:
> Otherwise they won't be used.

It's simple enough to patch. But I can name it the same as `t64` also, if we could decide on a name.

mingw-w64-python-distlib/PKGBUILD
lazka: ok

naveen521kk:
ready!

jeremyd2019:
What I had done with setuptools is https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-setuptools/0007-windows-arm64.patch but I only did that during the time when it tried to include *all* arches in the package, before that I just overloaded the 64 variant as you are talking about here.

naveen521kk:
If you give a nice executable name I can patch like that, or this would anyway work.

mingw-w64-python-distlib/PKGBUILD
```diff
@@ -48,8 +30,43 @@ prepare() {
...
+    gcc $_cli_cflags -o distlib/t32.exe PC/launcher.c PC/resources.res $_cli_lflags
+    gcc $_gui_cflags -o distlib/w32.exe PC/launcher.c PC/resources.res $_gui_lflags
+    ;;
+  x86_64-w64-mingw32)
+    gcc $_cli_cflags -o distlib/t64.exe PC/launcher.c PC/resources.res $_cli_lflags
+    gcc $_gui_cflags -o distlib/w64.exe PC/launcher.c PC/resources.res $_gui_lflags
+    ;;
+  aarch64-w64-mingw32)
+    gcc $_cli_cflags -o distlib/t64.exe PC/launcher.c PC/resources.res $_cli_lflags
+    gcc $_gui_cflags -o distlib/w64.exe PC/launcher.c PC/resources.res $_gui_lflags
```
mati865, line 63: All these should probably use `cc`. Clang subsystems use `gcc` as `clang` but hopefully it's temporary measure.

naveen521kk: ok done.

naveen521kk:
@lazka can this be merged?

PR #8757, opened 2021-05-23
Update mingw-w64

lazka:
hardcode the pkgver, to bring us one step closer to "git describe" output


Biswa96:
There is v9.0.0 release in mingw-w64, so shouldn't this be exactly v9 because there are no extra commit after v9?

lazka:
the old version is already larger than 9.0.0

jeremyd2019:
I was wondering if it was going to change to 10.0.0.1 now that 9.0.0 was released 😉

PR #8761, opened 2021-05-25
libnfs: fix compilation with clang

neheb:
Remove faulty patch


Biswa96:
Does removing the patch make any difference in GCC builds?

neheb:
The CI says no.

PR #8762, opened 2021-05-25
mingw-w64-cross-clang: prefixed cross-compiling wrappers for clang

jeremyd2019:
Related to discussion #8736.

Based on wrappers from [llvm-mingw](https://github.com/mstorsjo/llvm-mingw).  Also adds a package that provides "foreign" compiler-rt libs.

The clang wrapper could be simplified (have some of the options removed) if the package is only provided for CLANG*.  I haven't tested them under MINGW*.

### DIFF
--- a/mingw-w64-cross-clang/PKGBUILD
+++ b/mingw-w64-cross-clang/PKGBUILD
@@ -1,0 +1,93 @@
+declare -g -A _cross_arches=(
+  ["/clang64"]="x86_64-w64-mingw32"
+  ["/clangarm64"]="aarch64-w64-mingw32"
+)
+pkgbase=mingw-w64-cross-clang
+pkgname=($(for _pfx in "${!_cross_arches[@]}"; do
+             if [[ "${_cross_arches[$_pfx]%%-*}" != "${CARCH}" ]]; then
+               echo "${MINGW_PACKAGE_PREFIX}-cross-clang-${_cross_arches[$_pfx]%%-*}"
+             fi
+           done))
+pkgver=19.1.6
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
+license=('ISC')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc")
+# this is unused in the packages but necessary to make CI happy
+depends=("${MINGW_PACKAGE_PREFIX}-clang=${pkgver}"
+         "${MINGW_PACKAGE_PREFIX}-cross-compiler-rt=${pkgver}")
+source=("native-wrapper.h"
+        "llvm-wrapper.c"
+        "clang-target-wrapper.c"
+        "clang-scan-deps-wrapper.c"
+        "dlltool-wrapper.c"
+        "windres-wrapper.c")
+sha256sums=('c9b63d78a6e45b7ececfcc354d84ae4d452df8afc738c151f3aa37b843f6a338'
+            'b807f9b7ce3c205afc61f917a19d18e007b994b4bb2e8042f0db65976e96a4ec'
+            '67fd935417e9f4c69f44571e0cd219cab06699465a089e3149566c17ef8456ab'
+            'f9c8223a1b9813de4d09f914121ec874568296cf0bf65569e1802b66e784de51'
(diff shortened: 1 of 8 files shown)


jeremyd2019:
This seems to be working quite well.  I can use this to cross-compile python.  The packaging is a bit odd, but it seems to do the job.

driver1998:
Wonder how will this work with things like cmake, meson and ninja? Can I use mingw64 cmake with aarch64-w64-mingw32-clang?

jeremyd2019:
> Wonder how will this work with things like cmake, meson and ninja? Can I use mingw64 cmake with aarch64-w64-mingw32-clang to cross compile ARM64 binaries on x86_64?

Note I haven't bothered tested the GCC-based MSYSTEMs with this, they should work but who knows...

But with clang64, absolutely!  All the major build systems have a mechanism to configure them to cross-compile.  In fact, I had good luck with cmake not even bothering configuring a toolchain file, and instead just specifying options:

```
        -DCMAKE_{ASM,C,CXX}_COMPILER_TARGET="aarch64-w64-mingw32" \
        -DCMAKE_{ASM,C,CXX}_COMPILER_WORKS=ON \
        -DCMAKE_SYSROOT="/clangarm64" \
        -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
        -DCMAKE_FIND_ROOT_PATH_MODE_{LIBRARY,INCLUDE,PACKAGE}=ONLY \
```

jeremyd2019:
The way I'm doing a "standalone" compiler-rt build here is apparently deprecated.  I'll have to investigate how they intend it to be done now, hopefully it doesn't require downloading the `llvm` sources.

mati865:
How about installing all architectures supported by us on every architecture instead of creating new compiler-rt packages?
It's 35 MiB right now so that would add 70 MiB which doesn't seem so bad considering size and price of today's storage.

jeremyd2019:
How difficult would that be?  Is that possible from the existing mingw-w64-headers-git install or would we need to install the 'foreign' -headers-git like I do for the cross-compiler-rt package here? (compiler-rt only seems to need the headers, thankfully, as it is a static library.  Don't know about the sanitizers, haven't tried to do those for this cross compilation thing yet).

mati865:
Right, forgot about mingw-w64. That would make it even harder than current approach...

jeremyd2019:
I set up a github repository with a release to keep binaries of this, and set up a pacman repo for the latest version.

The files are in the release https://github.com/msys2-arm/cross-clang/releases/repo.  To use, add the following in `/etc/pacman.conf`:

```ini
[cross-clang]
Server = https://github.com/msys2-arm/cross-clang/releases/download/repo
SigLevel = Never
```

MehdiChinoune:
> I set up a github repository with a release to keep binaries of this, and set up a pacman repo for the latest version.
> 
> The files are in the release https://github.com/msys2-arm/cross-clang/releases/repo. To use, add the following in `/etc/pacman.conf`:
> 
> ```ini
> [cross-clang]
> Server = https://github.com/msys2-arm/cross-clang/releases/download/repo
> SigLevel = Never
> ```
```
$ pacman -S mingw-w64-clang-x86_64-cross-clang-aarch64
resolving dependencies...
warning: cannot resolve "mingw-w64-clang-aarch64-crt", a dependency of "mingw-w64-clang-x86_64-cross-clang-aarch64"
warning: cannot resolve "mingw-w64-clang-aarch64-headers", a dependency of "mingw-w64-clang-x86_64-cross-clang-aarch64"
warning: cannot resolve "mingw-w64-clang-aarch64-libc++", a dependency of "mingw-w64-clang-x86_64-cross-clang-aarch64"
warning: cannot resolve "mingw-w64-clang-aarch64-libunwind", a dependency of "mingw-w64-clang-x86_64-cross-clang-aarch64"
warning: cannot resolve "mingw-w64-clang-aarch64-winpthreads-git", a dependency of "mingw-w64-clang-x86_64-cross-clang-aarch64"
:: The following package cannot be upgraded due to unresolvable dependencies:
      mingw-w64-clang-x86_64-cross-clang-aarch64
```

jeremyd2019:
Unfortunately, I had to drop i686 targets for this because there are no more clang32 packages to base it on ☹️.  I have extracted past versions from this PR and put them in https://github.com/msys2-arm/cross-clang, and added a CI to build and upload them to the `repo` release.  Thus, I am going to close this PR and all future versions will happen there.

PR #8763, opened 2021-05-25
liblo: fix compilation with lld

neheb:


Biswa96:
Another thing in the list of GCC can but LLVM can't.

neheb:
lld is a linker. gcc is a compiler. You're thinking of GOLD.

jeremyd2019:
BFD if you want to be pedantic :wink:

neheb:
:)

mati865:
> Another thing in the list of GCC can but LLVM can't.

Opened https://github.com/msys2/CLANG-packages/issues/45

PR #8764, opened 2021-05-25
gss: add missing make dependency

neheb:


lazka:
bump pkgrel

neheb:
why? this is just a make dependency?

lazka:
What does adding this makedep change in regards to the final package?

no pkgrel bump, means no rebuild. If you don't need a rebuild then I wonder what this does.

neheb:
added.

PR #8766, opened 2021-05-25
ci-check: use temporary repo's

naveen521kk:


naveen521kk:
@lazka does this look good?

.ci/ci-check.py
```diff
@@ -116,9 +116,9 @@ def install_package(pkg: typing.Union[typing.List[str], Path]) -> typing.Generat
         ret = subprocess.run(command)
     if ret.returncode != 0:
         logger.error(
-            "%s failed to run. Returned return code: %s", command, ret.returncode
+            "[WARNING] %s failed to run. Returned return code: %s", command, ret.returncode
```
jeremyd2019, line 119: I think you want https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message

It says file/line/col are optional, I've used `::error ::<message>` don't know if the space after error is required or not

.ci/ci-check.py
```diff
@@ -116,9 +116,9 @@ def install_package(pkg: typing.Union[typing.List[str], Path]) -> typing.Generat
         ret = subprocess.run(command)
     if ret.returncode != 0:
         logger.error(
-            "[WARNING] %s failed to run. Returned return code: %s", command, ret.returncode
+            "::error :: %s failed to run. Returned return code: %s", command, ret.returncode
```
jeremyd2019, line 119: these should have been `::warning::` I think

naveen521kk: error looks right here

PR #8767, opened 2021-05-25
freerdp: Add package.

Biswa96:


Biswa96:
Pinging some random bots

* @vityank and @astrand - who started this
* @akallabeth - freerdp member
* @ohardening - freerdp contributor
* freerdp-bot - without whom this would not be possible 😘 

Any comments?

Biswa96:
Pinging some random bots

* @vityank and @astrand - who started this
* @akallabeth - freerdp member
* @hardening - freerdp contributor
* freerdp-bot - without whom this would not be possible 😘 

Any comments?

mingw-w64-freerdp/0005-rename-interface-variable.patch
```diff
@@ -0,0 +1,67 @@
...
+ 
+--- a/winpr/include/winpr/stream.h
++++ b/winpr/include/winpr/stream.h
+@@ -149,7 +149,7 @@
+ #define Stream_Read_UINT32_BE(_s, _v) _stream_read_n32_be(UINT32, _s, _v, TRUE)
+ #define Stream_Read_INT32_BE(_s, _v) _stream_read_n32_be(INT32, _s, _v, TRUE)
+ 
+-#define Stream_Read_UINT64(_s, _v) _stream_read_n64_le(UINT64, _s, _v, TRUE)
++#define Stream_Read_UINT64(_s, _v)
+ #define Stream_Read_INT64(_s, _v) _stream_read_n64_le(INT64, _s, _v, TRUE)
```
hardening, line 65: This change makes the compilation work, but will fail at runtime for sure...

mingw-w64-freerdp/0002-fix-build-errors-mingw.patch
```diff
@@ -0,0 +1,308 @@
...
+ 
+ LPWSTR devStr = NULL;
+ wfPeerContext* latestPeer = NULL;
+@@ -162,7 +164,7 @@ int wf_wasapi_get_device_string(LPWSTR pattern, LPWSTR* deviceStr)
+ 	return 0;
+ }
+ 
+-static DWORD WINAPI wf_rdpsnd_wasapi_thread(LPVOID lpParam)
++DWORD WINAPI wf_rdpsnd_wasapi_thread(LPVOID lpParam)
+ {
```
hardening, line 260: just curious, why can't it be ` static`  ?

mingw-w64-freerdp/0005-rename-interface-variable.patch
Biswa96: I didn't get any proper solution. There is confusion b/w LARGE_INTEGER and long long. But wfreerdp works with VBox's own RDP server and xrdp server in ArchLinux VM.

mingw-w64-freerdp/0002-fix-build-errors-mingw.patch
Biswa96: That function is in a header file and I'm not sure if it will be used in future.

mingw-w64-freerdp/0005-rename-interface-variable.patch
Biswa96: O, found the solution.

jeremyd2019:
I've tested the mingw32 variant and it works for me.

Biswa96:
@akallabeth You can find those countless warnings in the CI build log here.

PR #8770, opened 2021-05-26
Update Packages

naveen521kk:


naveen521kk:
hmm, pip itself gets uninstalled if pip is rebuilt. should find a fix.

mingw-w64-python-slugify/PKGBUILD
```diff
@@ -13,6 +13,7 @@ url="https://github.com/un33k/python-slugify"
 license=('MIT')
 depends=(
   "${MINGW_PACKAGE_PREFIX}-python"
+  "${MINGW_PACKAGE_PREFIX}-python-text-unidecode"
```
lazka, line 16: missing a pkgrel bump?

naveen521kk: yeah, fixed

PR #8772, opened 2021-05-27
libmodplug: fix package

neheb:
Remove faulty pkgconfig patch that made sndfile.h available without a
prefix. This breaks compilation for packages that use both libsndfile
and libmodplug like MPD.

Added small patch to fix compilation with MinGW. lstrcpyn is defined
to the W variant in MinGW's headers, which breaks compilation. Just
make it explicit.


lazka:
lgtm

Issue #8724, opened 2021-05-19
Static linking of libcurl (again)


msink:
When testing #8469 tried to fully static link of new `libcurl-winssl` package:
```
            compilerOpts = -DIDN2_STATIC -DNGHTTP2_STATICLIB -DCURL_STATICLIB
            linkerOpts(
                "-L${mingwPath.resolve("lib")}",
                "-Wl,-Bstatic",
                "-lstdc++",
                "-static",
                "-lcurl-winssl",
                "-lssh2-wincng",
                "-ladvapi32",
                "-lbcrypt",
                "-lbrotlidec-static",
                "-lbrotlicommon-static",
                "-lcrypt32",
                "-lidn2",
                "-lpsl",
                "-lunistring",
                "-liconv",
                "-lwldap32",
                "-lws2_32",
                "-lz",
                "-lzstd"
            )
```
output:
```
ld: f:\msys64\mingw64\lib\libidn2.a(lookup.o):(.text+0x709): undefined reference to `__imp_uninorm_nfc'
ld: f:\msys64\mingw64\lib\libidn2.a(idna.o):(.text+0xaf): undefined reference to `__imp_uninorm_nfc'
ld: f:\msys64\mingw64\lib\libidn2.a(idna.o):(.text+0x38b): undefined reference to `__imp_UC_CATEGORY_M'
ld: f:\msys64\mingw64\lib\libidn2.a(idna.o):(.text+0x3c0): undefined reference to `__imp_uninorm_nfc'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x128b): undefined reference to `__imp_uninorm_nfkc'
ld: f:\msys64\mingw64\lib\libpsl.a(psl.c.obj):(.text+0x1307): undefined reference to `__imp_uninorm_nfkc'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

https://github.com/msys2/MINGW-packages/pull/8469#issuecomment-843978186
> the __imp_uninorm_* references mean that the unistring references had __declspec(dllimport) when psl was compiled.

So fully static `libcurl` is almost passible, only static `unistring` is broken?

lazka:
we link it statically for the cygwin build where it comes down to https://github.com/msys2/MSYS2-packages/blob/master/curl/0001-more-static-fixes.patch

msink:
>  Upstream doesn't support static builds, so patches welcome

Upstream of what, `libunistring` ?

[...]

PR #8773, opened 2021-05-27
Fix static libunistring and rebuild libidn2 and libpsl

Astrum-polaris:
Try to address #8724 by separating static and shared builds


msink:
Confirmed that it fixes #8724

PR #8778, opened 2021-05-29
popt: update to 1.18

jeremyd2019:
Rebase patches, drop one that was already applied.

I just packaged popt for msys and noticed there was a newer version than what was in mingw-w64-popt


jeremyd2019:
I installed mingw-w64-i686-librsync and tested that `--help` and `-V` (version) still worked

PR #8779, opened 2021-05-29
Update packages

Biswa96:


jeremyd2019:
fyi, clang64 failure on glibmm:
```
  ../glibmm-2.66.1/glib/glibmm/ustring.h:1079:14: error: attribute 'dllexport' cannot be applied to a deleted function
    GLIBMM_API FormatStream(const ustring::FormatStream&) = delete;
               ^
  ../glibmm-2.66.1/glib/glibmm/ustring.h:1080:28: error: attribute 'dllexport' cannot be applied to a deleted function
    GLIBMM_API FormatStream& operator=(const ustring::FormatStream&) = delete;
```

lazka:
Reported upstream: https://gitlab.gnome.org/GNOME/glibmm/-/issues/92

Biswa96:
Is that C++ thing or clang/llvm specific issue? Not judging by gcc's success or faliure.

lazka:
I'd guess clang being picky.

jeremyd2019:
Yeah I'd think so too.  `= delete` means there *isn't* a function there (as opposed to an automatically-defined one), and clang says you can't dllexport/import that.

Biswa96:
Yes, that's intuitive. Shouldn't gcc check that as well? /cc @lhmouse

jeremyd2019:
> Reported upstream: https://gitlab.gnome.org/GNOME/glibmm/-/issues/92

This appears to have been added in https://gitlab.gnome.org/GNOME/glibmm/-/merge_requests/51/diffs#103176574bfa8e2d79a565c03f0264b03aad7576_986_985

jeremyd2019:
> Yes, that's intuitive. Shouldn't gcc check that as well?

I don't know if there's an intention behind it, but I've found when gcc/binutils is asked to do something that doesn't make sense on Windows, it generally just ignores it, while when llvm/clang/lld is asked to do the same thing it throws an error.  Somewhat less often now as of mingw-w64-clang-12.0.0-5 😉

PR #8782, opened 2021-05-29
Python updates

Schala:
- python-boto3 1.17.84
- python-botocore 1.20.84
- python-cattrs 1.7.1
- python-google-auth 1.30.1
- python-requests-mock 1.9.3


lazka:
cattrs is missing a dependency on attrs

PR #8792, opened 2021-05-30
ci: explicitly install pactoys-git

jeremyd2019:
For some reason it seems to have stopped being installed, so do it explicitly.

Fixes msys2/MSYS2-packages#2524


lazka:
Because of https://github.com/msys2/MSYS2-packages/commit/19bd06b131d2b579de00489374a8c0d853caf812#diff-1bb39ad9326f0bf25436a7eea61cf22ed04219ec60f31fd5576a9858ca592a46

PR #8795, opened 2021-05-30
glibmm: fix build with clang

jeremyd2019:
Don't try to dllexport deleted methods.

https://gitlab.gnome.org/GNOME/glibmm/-/issues/92


jeremyd2019:
I would have waited for upstream but the queue page looked like a lot of things were being blocked by this.

PR #8803, opened 2021-06-01
clang: update 12.0.0-6

### LINKED ISSUE #8706: `std::call_once` from libstdc++ always segfaults on Clang 12 - patch included
I can reproduce on both MSYS2 Clang, and the official Windows and Ubuntu binaries.

It's a regression from Clang 11, where it used to work. It seems Clang 12 doesn't like namespace-scope TLS on MinGW.

A minimal code that crashes:
```cpp
#include <iostream>
#include <mutex>

int main()
{
    std::once_flag flag;
    std::call_once(flag, []{std::cout << "Once!\n";});
}
```

A proper fix would be on Clang's side, but as a workaround we can patch libstdc++ (`/mingw64/include/c++/10.3.0/mutex`).
```cpp
@@ -691,8 +691,8 @@ _GLIBCXX_BEGIN_NAMESPACE_VERSION

   /// @cond undocumented
 #ifdef _GLIBCXX_HAVE_TLS
-  extern __thread void* __once_callable;
-  extern __thread void (*__once_call)();
+  inline void *&__once_callable_get() {static __thread void *__ret; return __ret;}
+  inline void (*&__once_call_get())() {static __thread void (*__ret)(); return __ret;}
 #else
   extern function<void()> __once_functor;

@@ -703,7 +703,7 @@ _GLIBCXX_BEGIN_NAMESPACE_VERSION
   __get_once_mutex();
 #endif

-  extern "C" void __once_proxy(void);
+  extern "C" inline void __once_proxy_inline(void) {__once_call_get()();}
   /// @endcond

   /// Invoke a callable and synchronize with other calls using the same flag
@@ -718,15 +718,15 @@ _GLIBCXX_BEGIN_NAMESPACE_VERSION
            std::forward<_Args>(__args)...);
       };
 #ifdef _GLIBCXX_HAVE_TLS
-      __once_callable = std::__addressof(__callable); // NOLINT: PR 82481
-      __once_call = []{ (*(decltype(__callable)*)__once_callable)(); };
+      __once_callable_get() = std::__addressof(__callable); // NOLINT: PR 82481
+      __once_call_get() = []{ (*(decltype(__callable)*)__once_callable_get())(); };
 #else
       unique_lock<mutex> __functor_lock(__get_once_mutex());
       __once_functor = __callable;
       __set_once_functor_lock_ptr(&__functor_lock);
 #endif

-      int __e = __gthread_once(&__once._M_once, &__once_proxy);
+      int __e = __gthread_once(&__once._M_once, &__once_proxy_inline);

 #ifndef _GLIBCXX_HAVE_TLS
       if (__functor_lock)
```

### LINKED ISSUE #8830: cannot compile with clang (clang64)
testcase:

```C
#include<stdio.h>

int main() {
        printf("Hello World\n");
        return 0;
}
```
Using `clang` from `mingw64` or `ucrt64` compiles this simple test case.

`clang` from `clang64` can't find the includes files:

```
$ clang hello.c
hw.c:1:9: fatal error: 'stdio.h' file not found
#include<stdio.h>
        ^~~~~~~~~
1 error generated.
```
or
```
$ clang++ hello.c
clang++: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
In file included from hw.c:1:
C:/Users/User/AppData/Local/msys2/clang64/include/c++/v1/stdio.h:107:15: fatal error:
      'stdio.h' file not found
#include_next <stdio.h>
              ^~~~~~~~~
1 error generated.
```

jeremyd2019:
Done as a branch on msys2/MINGW-packages with the hope that that makes it easier to collaborate on than in a personal repo like it was last time.

clang: apply backport of llvm-windres in lieu of windres wrapper.  Path from @mati865 from https://github.com/msys2/CLANG-packages/issues/44#issuecomment-846384245

This turned out not to work because it included a patch that touched a lot more things than just `llvm-rc`, and changed an API in such a way that it still managed to compile, but didn't do the right thing anymore (having to do with opening files in text/binary mode).

Therefore, I updated the windres backport patch, excluding llvm/llvm-project@c83cd8f as it touches a lot more than llvm-rc.
This required fixing one other commit that conflicted after that was removed.  Fixes msys2/CLANG-packages#44

Since these commits were cherry-picked, and one conflicted, the commit hashes in the patch file no longer match upstream.

Backport commit llvm/llvm-project@0e4cf80 which fixes #8706

Make -clang package depend on crt and headers for clangprefix, like GCC does.  Fixes #8830.

Patch llvm-dlltool to default to LLVM's default target rather than throw an error that no target was specified.  Fixes msys2/CLANG-packages#43.


jeremyd2019:
https://github.com/llvm/llvm-project/commit/0e4cf807aeaf54a10e02176498a7df13ac722b37 for #8706?

mati865:
> llvm/llvm-project@0e4cf80 for #8706?

Sure.

> @mati865 what was the cmake option required for building a libunwind that allows rust to work? 

[0001-Fix-libunwind.patch.txt](https://github.com/msys2/MINGW-packages/files/6578378/0001-Fix-libunwind.patch.txt)

> Do you have any other patches?

I don't know if it will apply cleanly and it's not yet accepted but https://reviews.llvm.org/D103012 is important for any code using LTO together with `--export-all-symbols`. This should be rare so maybe don't even have such package in the repo.

---

You should include something like this in the description so relevant issues are closed automatically:
```
Fixes https://github.com/msys2/CLANG-packages/issues/44
Fixes #8706
```

jeremyd2019:
> 
> You should include something like this in the description so relevant issues are closed automatically:
> 
> ```
> Fixes https://github.com/msys2/CLANG-packages/issues/44
> Fixes #8706
> ```

I have, well enough for github to see them, but I kind of did them in 'prose' rather than tacked on at the end

mati865:
Oh, dotted underline isn't really visible with Dark GiHub theme.

jeremyd2019:
I grabbed the artifacts from this state (bc68f3e), installed the mingw-w64-clang-x86_64 variant, and used it to build python successfully (uses PGO, LTO, rc file with icon, basically a good exercise of a toolchain)

Astrum-polaris:
Clang should depend on mingw-w64 crt, headers, winpthreads like gcc does

Otherwise clang fails to compile `Hello World`  if the user does not install the full toolchain
https://github.com/msys2/MINGW-packages/issues/8830#issuecomment-854651093

mati865:
Winpthreads are unnecessary for Clang.

mati865:
Patch that fixes https://github.com/msys2/CLANG-packages/issues/43 (default target for dlltool).
[Fix-dlltool-default-target.patch.txt](https://github.com/msys2/MINGW-packages/files/6612614/Fix-dlltool-default-target.patch.txt)

Tested only with LLVM trunk.

jeremyd2019:
Artifacts from 26a7ef4 were still able to build python.  Also, was able to build wined3d (which was a package failing due to not setting a target machine for dlltool), after adding `-fcommon` to the CFLAGS (which should be required for GCC now too)

I would have expected the dlltool patch to be upstream-able.  Especially if it were enhanced with some of the extra logic from llvm-rc: https://github.com/llvm/llvm-project/blob/main/llvm/tools/llvm-rc/llvm-rc.cpp#L417-L431  namely, get from argument, or from prefix on executable name, or default.

mati865:
> 
> 
> Winpthreads are unnecessary for Clang.

I take my words back, some headers try to include winpthreads headers (since we have removed stubs since they conflict with winpthreads...) and that fails if winpthreads packages is not installed.

jeremyd2019:
> > Winpthreads are unnecessary for Clang.
> 
> I take my words back, some headers try to include winpthreads headers (since we have removed stubs since they conflict with winpthreads...) and that fails if winpthreads packages is not installed.

Should I add winpthreads to clang depends too then?

mati865:
> Should I add winpthreads to clang depends too then?

Yeah, I see you already did it.

> It seems gcc also depends on windows-default-manifest but that doesn't
build with llvm-rc.

AFAIK Clang doesn't even link it anyway.

jeremyd2019:
#8868 still happens with the assets from the latest run (though lldb from CLANG64 works fine).  @mati865 can you look at this?  I've had 0 luck trying to get RelWithDebInfo builds of llvm to work

mati865:
Building shared LLVM on Windows with debug info is doomed to fail because of DLL limitations. Fully static could be possible though.

LLDB from Clang subsystem doesn't help either:

<details>
<summary>Details</summary>

```
$ winpty lldb /mingw64/bin/lldb.exe
(lldb) target create "D:/msys64/mingw64/bin/lldb.exe"
Current executable set to 'D:\msys64\mingw64\bin\lldb.exe' (x86_64).
(lldb) r
Process 10100 launched: 'D:\msys64\mingw64\bin\lldb.exe' (x86_64)
Process 10100 stopped
* thread #1, stop reason = Exception 0xc0000005 encountered at address 0x7ffbfbc00000: User-mode data execution prevention (DEP) violation at location 0x7ffbfbc00000
    frame #0: 0x00007ffbfbc00000
error: Only part of a ReadProcessMemory or WriteProcessMemory request was completed.
(lldb) bt
* thread #1, stop reason = Exception 0xc0000005 encountered at address 0x7ffbfbc00000: User-mode data execution prevention (DEP) violation at location 0x7ffbfbc00000
  * frame #0: 0x00007ffbfbc00000
    frame #1: 0x00007ffc5bff35f5 liblldb.dll`lldb::SBData::SetDataFromDoubleArray(double*, unsigned long long) + 1477
    frame #2: 0x00007ffc5c00c9b8 liblldb.dll`lldb::SBDebugger::InitializeWithErrorHandling() + 408
    frame #3: 0x00007ff6da80db74 lldb.exe
    frame #4: 0x00007ff6da8013c1 lldb.exe
    frame #5: 0x00007ff6da8014f6 lldb.exe
    frame #6: 0x00007ffd16fa7034 kernel32.dll`BaseThreadInitThunk + 20
    frame #7: 0x00007ffd17d42651 ntdll.dll`RtlUserThreadStart + 33
```

</details>

jeremyd2019:
That's pretty much what I saw with windbgx as well (though without the demangling)

jeremyd2019:
I was able to (eventually) get liblldb.dll built RelWithDebInfo, and `cv2pdb`-ed.  I put the windbgx `!analyze -v` output on the issue rather than on this PR since I think it's more relevant there.

jeremyd2019:
Is there a point to continue holding this up on #8868?

mati865:
I don't think so, this seems like LLVM 12 or GCC regression and we probably won't be able to quickly fix it.

mati865 (approved):
You can merge it if you think this is ready.

jeremyd2019:
Yes, I just gave it a test on ARM64 and it seems good there too.  I am only concerned now with the scheduling with autobuild since this package takes so long to build.

PR #8804, opened 2021-06-01
curl: update to 7.77.0

### LINKED ISSUE #8774: About  libcurl.a is link against libnghttp2.dll.a
it's been there for quite a long time, please fix it.

#5440
#4404
#3538
#950

Astrum-polaris:
adapt `_variant_config` to https://github.com/curl/curl/pull/6897

also separate static and shared builds because `0002-nghttp2-static.patch` only works with shared build disabled.  Fixes #8774

### DIFF
--- a/mingw-w64-curl/PKGBUILD
+++ b/mingw-w64-curl/PKGBUILD
@@ -13,7 +13,7 @@ fi
 _realname=curl
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}${_namesuff}"
-pkgver=7.76.1
+pkgver=7.77.0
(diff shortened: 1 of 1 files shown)


jeremyd2019:
How should this relate with #8469?  They seem likely to conflict

Astrum-polaris:
This has no conflict with #8469 just that it seems like it is not going to be merged anytime soon so I make a separate PR

If you want to merge #8469 first I can rebase the change pretty easily.

jeremyd2019:
Does this fix #8774?

Astrum-polaris:
It should, updated the commit message.

msink:
Checked files from CI - it really fixes #8774

PR #8808, opened 2021-06-01
python-imagecodecs: update to 2021.5.20

kmilos:


kmilos:
Something's strange with the `mingw-w64-cfitsio` package, symbols from `ricecomp.c` don't seem to be exported @Adsun701 @lock042

kmilos:
Ah, none of the symbols in `fitsio2.h` are marked as exportable/public with `CFITS_API`, but can't say if this is a bug upstream or these are private for a reason...

Adsun701:
@kmilos seeing as CFITS_API is only used in two files, `fitsio.h` and `iraffits.c`, I think those symbols in `fitsio2.h` may be private for some reason. Not really sure why though.

kmilos:
Thanks for checking - that's right, there's no way to know, so I've asked the `imagecodecs` author to try and figure this out with upstream...

I guess the most sensible thing is to just not enable the feature ATM.

PR #8812, opened 2021-06-01
mpv: Update to 0.33.1

Biswa96:


Biswa96:
Merge this after we fix the x264 dependency. VLC also need x264.

PR #8825, opened 2021-06-03
extra-cmake-modules: Update to 5.82.0

Biswa96:


Biswa96:
CI can't detect the dependency to build it first.

PR #8826, opened 2021-06-03
Update Packages

naveen521kk:


lazka:
ignore the error in this case I guess?

naveen521kk:
Yeah, irrc https://github.com/msys2/MINGW-packages/pull/5615

PR #8827, opened 2021-06-03
python3.9: reserve 2MB of stack space

jeremyd2019:
https://bugs.python.org/issue40458 was happening on CLANG64 (test_pickle was crashing due to Windows stack overflow exception)

Also, get rid of no-longer-needed setting of CC and CXX for clang support (no longer needed after msys2/MSYS2-packages#2501)


lazka:
Thanks!

PR #8828, opened 2021-06-04
libkml clang fix.=

neheb:


mati865:
`strncasecmp` is available for all targets with mingw-w64 so I think those ifdefs should be like this:
```
#if defined(__MINGW32__) && !defined(__MINGW64_VERSION_MAJOR)
```

MehdiChinoune:
libkml is fixed for clang.

PR #8829, opened 2021-06-04
python3.9: add windows 7 support

naveen521kk:


Biswa96:
It may be better to track if a comment is added in the patch file with the link you've provided me tomorrow in Discord.

lazka:
How hard would it be to have all win7 changes in one patch?

naveen521kk:
Hmm, ok I will do those.

naveen521kk:
ping

lazka:
tested on win7, works nicely.

Issue #8824, opened 2021-06-03
static FFmpeg (again)


msink:
I asked it years ago #4234 ...

Now it is formally supported, but of course it is broken.

To start with: it links to `-l:libx264.dll.a` and `-l:libxml2.dll.a`, instead of `-lx264` and `-lxml2`
```
ld: libavformat.a(dashdec.o):(.text+0x10bc): undefined reference to `_imp__xmlGetProp'
...
ld: libavcodec.a(libx264.o):libx264.c:(.text+0x30): undefined reference to `_imp__x264_picture_init'
...
```

libx265 is broken in exotic way:
```
ld: libx265.a(api.cpp.obj):(.text+0x3c30): undefined reference to `x265_12bit::x265_api_get_192(int)'
ld: libx265.a(api.cpp.obj):(.text+0x3c49): undefined reference to `x265_10bit::x265_api_get_192(int)'
ld: libx265.a(api.cpp.obj):(.text+0x3cc1): undefined reference to `x265_10bit::x265_api_query(int, int, int*)'
ld: libx265.a(api.cpp.obj):(.text+0x3cd9): undefined reference to `x265_12bit::x265_api_query(int, int, int*)'
```
There was fix for this in vcpkg https://github.com/microsoft/vcpkg/pull/18145 but I just don't indestand what it does.

msink:
Well, my test is not minimal and not in C...
Googled relatively small videoplayer, should be good enough as integration test.
https://raw.githubusercontent.com/rambodrahmani/ffmpeg-video-player/master/tutorial07/tutorial07.c

But - I feel dumb but cannot compile it as static, everything from windows impibs is unresolved.
Definitely missed something obvious.

```
gcc -std=c99 tutorial07.c -o tutorial07.exe -static\
 -lSDL2 -lavformat -lavcodec -lavutil -lswscale -lswresample\
 -laom -ldav1d -lgsm -lmfx -lmodplug -lmp3lame -lopencore-amrnb -lopencore-amrwb\
 -lspeex -lopus -lvorbis -lvorbisenc -ltheoraenc -ltheoradec -logg -lsoxr -lgomp\
 -lwebpmux -lwebp -lvpx -l:xvidcore.a -lrtmp -lhogweed -lnettle -lgmp -lsrt\
 -liconv -lbz2 -llzma -lz\
 -l:libbluray.dll.a\
 -l:libcelt0.dll.a\
 -l:libgnutls.dll.a\
[...]

PR #8835, opened 2021-06-05
X264.pc.ffmpeg

shithappens2016:
> #8824

fix static libraries of ffmpeg link against libx264.dll.a


Biswa96:
Please merge the commits into one or two with meaningful commit message. And increase the `pkgrel` variable of the affected packages.

PR #8840, opened 2021-06-06
Fix libass.a link against libfribidi.dll.a

shithappens2016:
Fix libass.a link against libfribidi.dll.a


ImperatorS79:
I think this can be merged too.

Issue #7223, opened 2020-11-02
[indent]


bbostjanv:
Hello,
I use mingw64 for running Linux software on Windows10. So far I haven't had any big issues. I use gcc, and many other standard Linux tools all the time. However, lately I've run into problems with gnu indent. The usual way to use indent is on the command line, like this

indent <file name>

and until recently that worked for me. However, starting recently, when I write the above command line, and press RETURN, nothing happens. It seems indent simply hangs. This must be a stupid setup problem, but I can't figure out which.

I might add that after this problem appeared, I installed the gnuwin32 version of indent (version 2.2.10), and it works without problems. But I would still like to have all my Linux software in one place, and that's the reason I would like to fix my mingw64 indent.

Any answer would be appreciated.
Regards,
bbostjanv

bbostjanv:
my original post needs a correction. The line

indent

should be changed to

indent filename.c

bbostjanv

bbostjanv:
I must have done something wrong to "close" the issue. I still don't have a (re)solution.
bbostjanv

ejbergdk:
I can confirm this. I am using indent 2.2.12-2 (64-bit) on Windows 10 20H2.
When indenting a file with LF-only line endings, it works. The indented file is then written with CRLF (as expected under Windows).
But when indenting the same file again (which now has CRLF line endings), indent just hangs unresponsively until pressing Ctrl-C.

MartenH:
Having the same problem. Can verify that it have something to do with line-endings. LF-only works fine. CRLF hangs.
Piping to it also works:

`cat foo.c | indent.exe `

ejbergdk:
I think I'm on to something. In indent 2.2.12 file: src/code_io.c function: read_file.
There is a while loop, where the file is read. This while loop never exits, as read first return a size smaller than requested (due to CR/LF being converted to LF). All subsequent reads return 0 (as there is no more to read).

ejbergdk:
I believe I have a fix for this problem (it works for me at least). I just have to read up on how I proceed to get this fix sent through the system. Never tried it before.

PR #8841, opened 2021-06-06
indent: Fix endless file read loop

ejbergdk:
Indent 2.2.12 would end up in an endless loop if indenting a file with CR/LF line endings.

This happens when the file is read, where all CR/LF line endings are converted to LF only. This results in fewer bytes read than the file size indicates. A while loop keeps trying to read the "missing" bytes, but read returns 0.
This fix stops the while loop when there is no more to read.

### DIFF
--- a/mingw-w64-indent/PKGBUILD
+++ b/mingw-w64-indent/PKGBUILD
@@ -6,7 +6,7 @@ _realname=indent
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.2.12
-pkgrel=2
+pkgrel=3
 pkgdesc="C language source code formatting program (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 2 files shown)


ejbergdk:
This PR could be linked with issue #7223

lazka:
Was this reported upstream?

ejbergdk:
> Thanks. Was this reported upstream?

No, not yet at least. I'll have to read up on how to report problems to them. It took me long enough to get this far here (first time trying to contribute with something)

lazka:
I can reproduce the issue, but no on Linux, but there it creates garbage output when the input is CRLF

ejbergdk:
I don't think Linux file operations silently discards CR, so there it can read all the expected bytes in the file, without getting stuck in a loop.

lazka:
yeah, your changes make sense (and the existing code not so much). And savannah.gnu.org doesn't let me create a user account... so I guess this is fine as is.

PR #8844, opened 2021-06-06
mingw-w64-headers-git: add patch for av1 with dxva

1480c1:
Adds a patch that is currently not in mingw-w64 nor wine for adding av1
decoding support via dxva. This is useful for FFmpeg to decode using
their hwaccel.

From https://www.winehq.org/pipermail/wine-devel/2020-September/173175.html and used in https://github.com/BtbN/FFmpeg-Builds/commit/0df2644223a5d3f775b4e25d027a3dc4446caefe

### DIFF
--- a/mingw-w64-crt-git/0001-headers-add-AV1-support-to-dxva.h.patch
+++ b/mingw-w64-crt-git/0001-headers-add-AV1-support-to-dxva.h.patch
@@ -1,0 +1,309 @@
+From 4ede80067d321e75ea082f8c7e66fb167052fde9 Mon Sep 17 00:00:00 2001
+From: Steve Lhomme <robux4@ycbcr.xyz>
+Date: Sat, 15 May 2021 15:26:42 +0200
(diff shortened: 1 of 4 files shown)


Biswa96:
mingw-w64-crt need a bump. The GUIDs have to go into libuuid.a.

Biswa96:
crt does not need the patch as the patch is applied in mingw-w64-headers folder. By bump, I mean pkgrel++ Apology for confusion.

1480c1:
Oh, I just applied it since it seems the other patches that were the same were also applied

Biswa96:
It's OK to apply for both package.

1480c1:
found out that pkgver does indeed run after prepare, so the git patch being there will properly change the pkgver

Biswa96:
A new version of Windows SDK has been released and it has these definitions. I'd like to wait for some days before poking in wine mailing list.

PR #8850, opened 2021-06-07
perl: update to 5.32.1 

naveen521kk:


naveen521kk:
I am also going to make all the modules here to lowercase.

Biswa96:
> I am also going to make all the modules here to lowercase.

Is there any issue with current naming scheme? I agree it's different. Doesn't that create package conflict? And to mitigate that there'll be more conflicts...replaces...provides... array.

naveen521kk:
> Is there any issue with current naming scheme? I agree it's different. Doesn't that create package conflict? And to mitigate that there'll be more conflicts...replaces...provides... array.

Hmm, either we should make this lower case or revert https://github.com/msys2/MINGW-packages/pull/8430

Biswa96:
OK, seems reasonable. The `provides` list is created by `updpkgprovs` script. It has to be re-run after perl version change. It may need some change also for lowercase.

Biswa96:
Is it possible not to rename patch files? Just to make the diff cleaner and less.

naveen521kk:
Ok https://www.cpan.org/src/5.0/perl-5.32.1.tar.xz is down

naveen521kk:
@Biswa96 does this look good to you?

PR #8858, opened 2021-06-08
texlive-bin: improve quoting patch

naveen521kk:
make things a bit better.


Biswa96:
Can the `mingw_arch` array be added in this rebuild?

naveen521kk:
> The perl update changes file names so if this mereged after perl update we could skip another mass rebuild.

I don't think the texlive packages needs rebuild for perl update.

PR #8862, opened 2021-06-08
Update Tier 1 KF5 frameworks to 5.82.0

Biswa96:


Biswa96:
Ignore the failure in ucrt64. CI can build kdnssd-qt5 in ucrt64 if done separately.

PR #8883, opened 2021-06-12
Update bleach to version 3.3.0

danyeaw:


mingw-w64-python-bleach/PKGBUILD
```diff
@@ -17,7 +17,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-python"
          "${MINGW_PACKAGE_PREFIX}-python-html5lib")
```
lazka, line 17: Add a depends on "...-python-packaging"

PR #8888, opened 2021-06-12
leptonica: Upgrade to 1.18.1

jannick0:
@lazka Ok that the CLANG32 dependencies are not available? Or are they to be regenerated?

### DIFF
--- a/mingw-w64-leptonica/PKGBUILD
+++ b/mingw-w64-leptonica/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=leptonica
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=1.81.0
+pkgver=1.81.1
 pkgrel=1
 pkgdesc="Leptonica library (mingw-w64)"
 arch=('any')
@@ -20,7 +20,7 @@ depends=(${MINGW_PACKAGE_PREFIX}-gcc-libs
          ${MINGW_PACKAGE_PREFIX}-openjpeg2
          ${MINGW_PACKAGE_PREFIX}-zlib)
 source=(https://github.com/DanBloomberg/leptonica/releases/download/${pkgver}/${_realname}-${pkgver}.tar.gz)
-sha256sums=('d192b055e9bd60b84111023cc980c37390e6d427b194a8fd2bd611543a3bddad')
+sha256sums=('0f4eb315e9bdddd797f4c55fdea4e1f45fca7e3b358a2fc693fd957ce2c43ca9')
 
 prepare() {
   cd "${srcdir}/${_realname}-${pkgver}"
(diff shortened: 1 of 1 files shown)


mingw-w64-leptonica/PKGBUILD
```diff
@@ -29,28 +29,25 @@ prepare() {
...
+  test -d "${_builddir}" && rm -rf "${_builddir}"
+  mkdir -p "${_builddir}" && cd "${_builddir}"
   CFLAGS+=" -DMINIMUM_SEVERITY=L_SEVERITY_WARNING"
   ../${_realname}-${pkgver}/configure -C \
-    --disable-dependency-tracking \
-    --disable-silent-rules \
-    --build=${MINGW_CHOST} \
-    --host=${MINGW_CHOST} \
-    --target=${MINGW_CHOST} \
-    --prefix=${MINGW_PREFIX}
```
lazka, line 41: why these changes?

mingw-w64-leptonica/PKGBUILD
```diff
@@ -29,28 +29,25 @@ prepare() {
...
-  make
+    --disable-dependency-tracking
+  make V=0 || make V=1
```
lazka, line 38: does the build fail the first time?

lazka:
> @lazka Ok that the CLANG32 dependencies are not available? Or are they to be regenerated?

yes, please ignore

mingw-w64-leptonica/PKGBUILD
jannick0: These are set by `makepkg`, except for `--build` and `--target` which are redundant and only needed if cross-compiling and compiling compilers, resp.

Happy to add back
```
--host=${MINGW_CHOST}
--prefix=${MINGW_PREFIX}
```
the other stuff should go as in the commit.

mingw-w64-leptonica/PKGBUILD
jannick0: No, but after it failed in non-verbose mode, `make` is re-run to show the failing command line.  Droppable if you prefer. Either way I am fine. Locally I am running things as given in the commit.

lazka: ok, we've had this pattern when "make" had missing dependencies, or crashes which could be worked around by running it twice. This means I'd be afraid to change it in the future without knowing your motivation.

But I guess this makes sense.

mingw-w64-leptonica/PKGBUILD
lazka: ok, thanks, makes sense. We likely use it everywhere because prior to https://github.com/msys2/MSYS2-packages/issues/2285 CONFIG_SITE wasn't always set. But that's no longer the case now.

lazka: I'm not sure if `--prefix=${MINGW_PREFIX}` still has value for consistency and documentation, since we need to set it with all other build tools.

jannick0: to be on the safe side let me add back the defs of `prefix` and `host` ... together with a simple call of `make`.  Then everything is golden I guess.

lazka: We are definitely cargo culting various things for no good reason :) I'm fine with removing host if you think it's not needed. You definitely know much more about autotools then me

lazka: > together with a simple call of `make`

or add a small comment maybe `# call make again to see debug output if the first fails for some reason`

jannick0: All good - thanks. ... Force-pushed new more explicit version ready to be merged once CI is happy.

Issue #6538, opened 2020-05-30
Aspell 0.60.8 is out.


Eason0210:
[Aspell ](http://aspell.net/)0.60.8 is already released，https://ftp.gnu.org/gnu/aspell/aspell-0.60.8.tar.gz
Do you have plan to update to the latest version. This version support camel case word check.

mardukbp:
This version also adds support for Markdown files. Please update :)

Eason0210:
Is there any news about updating Aspell to 0.60.8 on Mingw64?

ZachBacon:
Feel free to submit a pull request.

Biswa96:
Would anyone like to test the updated package from the GitHub Actions artifacts in the linked pull request?

PR #8889, opened 2021-06-12
aspell: Update to 0.60.8

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/6538

### DIFF
--- a/mingw-w64-aspell/0002-printf.mingw.patch
+++ b/mingw-w64-aspell/0002-printf.mingw.patch
@@ -10,6 +10,18 @@
  #include "gettext.h"
  
  #include "iostream.hpp"
+--- a/common/convert.cpp
++++ b/common/convert.cpp
+@@ -22,6 +22,9 @@
+ 
+ #include "iostream.hpp"
+ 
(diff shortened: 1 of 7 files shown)


mingw-w64-aspell/0002-printf.mingw.patch
```diff
@@ -10,6 +10,18 @@
...
+--- a/common/convert.cpp
++++ b/common/convert.cpp
+@@ -22,6 +22,9 @@
+ 
+ #include "iostream.hpp"
+ 
++#define printf printf
++#include "libintl.h"
++#undef printf
+ #include "gettext.h"
```
lazka, line 22: just fyi, there is also:

```
#pragma push_macro("printf")
#undef printf
...
#pragma pop_macro("printf")
```

Biswa96: I just followed other hunks of the patch file.

lazka: sure, just something I've found out recently..

PR #8893, opened 2021-06-12
Update isort to version 5.8.0

danyeaw:


lazka:
I don't know if it makes things harder for you, but our new pip dependency checks would have an easier time if related Python updates were grouped in one PR. With them all separate I have to guess which version dependency errors are fixed by other PRs that are also ready to merge. Not that big of an issue though., just fyi.

Issue #8896, opened 2021-06-12
mingw-w64-waf: uses msys2 python


jeremyd2019:
This is pretty clearly wrong.

jeremyd2019:
Or not... I can't get it to work with mingw python because it doesn't think 'gcc' is executable.  And telling it to use 'gcc.exe' confuses it because it doesn't have a module for 'gcc.exe'

jeremyd2019:
if it does actually depend on msys python, it may be better as a msys package

shithappens2016:
same problem

MehdiChinoune:
> @MehdiChinoune
> 
> Please fix or close this issue if it has already been fixed

I wish I could, but I don't have access to this repo

mati865:
Fixed in https://github.com/msys2/MINGW-packages/pull/11901

PR #8897, opened 2021-06-12
wxPython: update to 4.1.1

### LINKED ISSUE #7969: Update wxpython to 4.x
There is now support upstream: https://github.com/wxWidgets/Phoenix/issues/1582

### LINKED ISSUE #8863: Can not "import wx" in python 3 after install  "mingw-w64-x86_64-wxPython"
i'm new user msys2, i don't know what i did wrong

Python 3.8.9 (default, Apr 13 2021, 15:54:59)  [GCC 10.2.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import wx
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'wx'

jeremyd2019:
Based on @ustsv's  https://bitbucket.org/ustsv/testpkg_msys2/src/master/mingw-w64-wxpython-phoenix-git/PKGBUILD

Fixes #7969, fixes #8863

Update kicad to use python3 and wxPython 4 (aka "Phoenix"), and work around it preferring the newer msys2 python (3.9) to the older mingw python (3.8).

Update oce to use `${MSYSTEM}` rather than `${CARCH}` for its install scripts, and add proper install scripts for ucrt64 and clang64.  This allows ucrt64 kicad to build.

TODO: if `waf` were taught how to use `windres`, wxPython's rc files could be switched on (by patching `isWindows` to `isWindowsLike` circa https://github.com/wxWidgets/Phoenix/blob/63b8099bc295e767923a9dd05693aa8a23e8c6b1/wscript#L706)


jeremyd2019:
```
  Pip Check Failed for mingw-w64-x86_64-wxPython
  Here is what it failed.
  wxpython 4.1.1 requires numpy, which is not installed.
  wxpython 4.1.1 requires pillow, which is not installed.
```
Guess it doesn't like `optdepends`.

jeremyd2019:
kicad will almost certainly require at least a rebuild, if not some new option to tell it to use python3 and wxPython4 aka "Phoenix"

mingw-w64-wxPython/0001-wxWidgets-win-from-cygwin.patch
```diff
@@ -0,0 +1,28 @@
+--- wxPython-4.1.1/wscript.ORIG	2021-06-12 13:34:52.919243800 -0700
++++ wxPython-4.1.1/wscript	2021-06-12 13:41:18.778905500 -0700
+@@ -26,6 +26,7 @@
+ VERSION = cfg.VERSION
+ 
+ isWindows = sys.platform.startswith('win')
++isWindowsLike = isWindows or sys.platform == 'cygwin'
```
jeremyd2019, line 7: another thought, which I came across somewhere today, would be
```suggestion
+isWindowsLike = isWindows or (sys.platform == 'cygwin' and 'MSYSTEM' in os.environ)
```

jeremyd2019:
```
In file included from C:/_/mingw-w64-kicad/src/kicad-5.1.10/pcbnew/swig/python_scripting.h:37,
                 from C:\_\mingw-w64-kicad\src\kicad-5.1.10\pcbnew\pcbnew.cpp:32:
D:/a/_temp/msys/msys64/usr/include/python3.9/Python.h:44:10: fatal error: crypt.h: No such file or directory
   44 | #include <crypt.h>
      |          ^~~~~~~~~
compilation terminated.
```

Need to figure out why it's picking up msys python...

jeremyd2019:
I guess I have to try it locally to poke at logs and see what's happening.

mingw-w64-wxPython/PKGBUILD
```diff
@@ -3,85 +3,95 @@
...
+         "${MINGW_PACKAGE_PREFIX}-wxWidgets3.1")
+makedepends=("${MINGW_PACKAGE_PREFIX}-doxygen"
+             "${MINGW_PACKAGE_PREFIX}-python-pillow"
+             "${MINGW_PACKAGE_PREFIX}-python-requests"
+             "${MINGW_PACKAGE_PREFIX}-sip4"
+             "${MINGW_PACKAGE_PREFIX}-waf"
+             # for waf
+             "python" "python-setuptools")
+optdepends=("${MINGW_PACKAGE_PREFIX}-python-numpy: floatcanvas and plot widgets"
+            "${MINGW_PACKAGE_PREFIX}-python-pillow: a few agw widgets")
```
jeremyd2019, line 24: Is this OK (despite the pip check failure) or should I convert these to 'hard' depends?

jeremyd2019:
It looks like kicad's cmake code iterates python version (3.10 3.9 3.8 ...) and within that iterates directories (user-specified ones first).  So it finds a version 3.9 in msys, before it tries looking for a 3.8 mingw.  Short of major patching of this cmake module, it seems the most reasonable course is to explicitly tell it where the mingw python include dir is.

jeremyd2019:
I opened https://github.com/wxWidgets/Phoenix/issues/1971 to discuss `0002-handle-cast.patch`

jeremyd2019:
It builds!

mingw-w64-wxPython/PKGBUILD
umarcor: Being numpy and pillow, I think it's ok to make them hard dependencies.

jeremyd2019: Except that numpy is not (yet) available for CLANG64 (due to lack of a fortran compiler), and I wanted to get wxPython to build under clang.  I expect it should be possible to build numpy without fortran support, but I don't think anybody's tried yet.

umarcor: Fair point. What about making it hard on other environments but soft on CLANG64?

jeremyd2019: possible, but is it worth the complexity?

umarcor: Up to you... This is my experience with setting different deps depending on the env: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ghdl/PKGBUILD#L75-L100

jeremyd2019: pillow is available for CLANG64, so that can easily be a hard dependency.  I only did it as `optdepends` because I saw it was only used in a few widgets, like numpy.

umarcor: My understanding is that someone using wxPython might want to manipulate images in Python, which are then to be shown in the GUI. Hence, I think not many people would complain about pillow being a hard dependency. Anyway, I understand if it's left as optional. numpy might be slightly arguable, since it is not explicitly for images/visual features.

jeremyd2019:
wxPython commit in isolation passed the pip check now.

lazka:
nice!

mingw-w64-wxPython/PKGBUILD
```diff
@@ -3,85 +3,96 @@
...
+  patch -Np1 -i "${srcdir}/0001-wxWidgets-win-from-cygwin.patch"
+  # clang doesn't like casting from pointer to smaller integer type (long)
+  # https://github.com/wxWidgets/Phoenix/pull/1972
+  patch -Np1 -i "${srcdir}/0002-handle-cast.patch"
+  patch -Np1 -i "${srcdir}/0003-wxWidgets.Doxyfile.patch"
+  patch -Np1 -i "${srcdir}/0006-grid_h.patch"
+  # https://github.com/wxWidgets/Phoenix/commit/3500ac7a9e7377c154a507dd7ea1b5b7bfda8c09
+  patch -Np1 -i "${srcdir}/0100-wxPython-wxHandleFatalExceptions.patch"
+  # windows crap is WRONG
+  #sed -i -e 's/^\(isWindows\s*=.*\)$/\1 and "GCC" not in sys.version/' wscript
```
mati865, line 56: Leftover?

jeremyd2019: yeah.  It *is* wrong, though 😉 Just even more wrong than I expected (#8896)

jeremyd2019: looks like I got an opportunity to remove that ☹️

jeremyd2019: To be clear, the "windows crap is wrong" in that it assumes `isWindows` means `isMSVC`.  But, it was even more wrong because `mingw-w64-waf` actually uses msys2's python, so sys.platform will not be `win32` (but `cygwin`), and `isWindows` will be `False`.  So I had to patch the file more extensively to treat `cygwin` as "Windows-like", enabling the pieces that are Windows and not MSVC.

jeremyd2019:
UCRT64 got an error downloading kicad sources.  Luckily that's not marked as 'required' so should still merge if the others succeed

jeremyd2019:
Note for the future: post-4.1.1, wxPython updated its required sip version from 4.x to 5.5.  There has not been a release since then, but expect the next release to require sip 5.5, while in this repo there is only `sip4` and `sip` which has since moved on to 6.x.

PR #8909, opened 2021-06-14
Update hacking to version 4.1.0

danyeaw:


lazka:
Does it work despite the broken dep requirements?

lazka:
Seems to work fine here.

PR #8919, opened 2021-06-14
Fix ucrt builds

Biswa96:


mingw-w64-grep/PKGBUILD
```diff
@@ -4,25 +4,34 @@
...
-pkgver=3.4
+pkgver=3.6
 pkgrel=1
 pkgdesc="Grep searches one or more input files for lines containing a match to a specified pattern (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 url='https://www.gnu.org/software/grep/'
 license=('GPL3')
 depends=("${MINGW_PACKAGE_PREFIX}-pcre")
+makedepends=("${MINGW_PACKAGE_PREFIX}-winpthreads")
```
mati865, line 15: I think this should also have runtime dependency on `${MINGW_PACKAGE_PREFIX}-libwinpthread-git`.

mingw-w64-sed/PKGBUILD
```diff
@@ -4,27 +4,32 @@
...
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=4.2.2
-pkgrel=2
+pkgver=4.4
+pkgrel=1
 pkgdesc="Sed is a stream editor (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 url='https://www.gnu.org/software/sed/'
 license=('GPL3')
```
jeremyd2019, line 13:
```suggestion
license=('GPL3')
depends=("${MINGW_PACKAGE_PREFIX}-libssp")
```

mingw-w64-grep/PKGBUILD
jeremyd2019: the commit message (and my recollection of the issue) refers to needing to link winpthreads for *tests*, so I would guess the actual installed binary does not depend on libwinpthread.

Biswa96: Yup, winpthreads in grep are for tests, not for any installed binaries.

mingw-w64-sed/PKGBUILD
Biswa96: Done.

Biswa96:
CI tries to use the grep from previous build and fails. But sed and grep build fine.

mingw-w64-grep/PKGBUILD
mati865: There is special array for that: https://wiki.archlinux.org/title/PKGBUILD#checkdepends

Biswa96: According to it `These dependencies are only considered when the check() function is present and is to be run by makepkg`. Those tests are built in `build()` function.

Biswa96:
ping!

Biswa96:
@lazka ping!

PR #8920, opened 2021-06-14
curl: preserve UNC paths when normalizing/simplifying

dscho:
The code to make cURL relocatable simplifies/normalizes paths, e.g. reducing multiple forward slashes to a single one.

This is okay in general, but at the beginning of the path, a double slash has a special meaning: it denotes a UNC path of the form
`//<host>/<path>`. It would therefore be a mistake to reduce those first two slashes to a single one: otherwise it would refer to the absolute path relative to the current directory's drive.

Original report: https://github.com/git-for-windows/git/issues/3266

So sad that I cannot Cc: the original author of the `pathtools.c` file. 😞

### DIFF
--- a/mingw-w64-curl/0001-Make-cURL-relocatable.patch
+++ b/mingw-w64-curl/0001-Make-cURL-relocatable.patch
@@ -22,10 +22,10 @@ Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
  configure.ac         |   1 +
  lib/Makefile.inc     |   5 +-
  lib/curl_config.h.in |   3 +
- lib/pathtools.c      | 533 +++++++++++++++++++++++++++++++++++++++++++++++++++
+ lib/pathtools.c      | 539 +++++++++++++++++++++++++++++++++++++++++++++++++++
  lib/pathtools.h      |  53 +++++
  lib/url.c            |  26 ++-
- 6 files changed, 618 insertions(+), 3 deletions(-)
+ 6 files changed, 624 insertions(+), 3 deletions(-)
  create mode 100644 lib/pathtools.c
  create mode 100644 lib/pathtools.h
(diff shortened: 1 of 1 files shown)


Biswa96:
If this is correct shouldn't this be done in mingw-w64-pathtools/pathtools.c also?

off-topic. I was thinking about separating pathtools.c and pathtools.h from the patch file and copy them from mingw-w64-pathtools at build time. There are many packages which have same code copied multiple times.

dscho:
> If this is correct shouldn't this be done in mingw-w64-pathtools/pathtools.c also?

Probably, but I have no insight about all the places where this file was copied.

> I was thinking about separating pathtools.c and pathtools.h from the patch file and copy them from mingw-w64-pathtools at build time. There are many packages which have same code copied multiple times.

I think that would be a fine thing, although I would recommend not copying it but instead making it a header-only package. That way, it would be much easier to see which packages need rebuilding after an update to `pathtools.h`.

lazka:
I've created #8955 so we (hopefully) don't forget.

jeremyd2019:
This needed sha256sums updated in the PKGBUILD

lazka:
thanks, fixed in master

lazka:
This couldn't have built, reverted everything for now. Sorry for missing this in the review.

dscho:
> 
> 
> This couldn't have built, reverted everything for now. Sorry for missing this in the review.

Bummer. How did it pass the CI build, then?

_clicketyclick_

Oh. [I see](https://github.com/msys2/MINGW-packages/runs/2821409437#step:9:20). It mistook the change outside of `PKGBUILD` for "don't need to rebuild". Bummer. I guess I should have _at least_ incremented `pkgrel`, eh?

lazka:
yeah

PR #8923, opened 2021-06-14
Update botocore to version 1.20.94

danyeaw:


danyeaw:
Merging with boto3 update

PR #8953, opened 2021-06-17
python3.9: autogenerate patches from git

naveen521kk:
https://github.com/msys2-contrib/cpython-mingw this is a similar setup to msys2-runtime


lazka:
bump pkgrel

PR #8956, opened 2021-06-18
Update Yosys, prjtrellis and nextpnr

umarcor:
This PR updates Yosys, prjtrellis and nextpnr. Since Yosys depends on GHDL, it is desirable to wait until that one is uploaded (https://packages.msys2.org/package/mingw-w64-x86_64-ghdl-llvm?repo=mingw64) before merging this PR. For that reason, I'll keep it as a draft.

### DIFF
--- a/mingw-w64-nextpnr/PKGBUILD
+++ b/mingw-w64-nextpnr/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=nextpnr
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.0.r3229.326b3488
+pkgver=0.0.r3600.9df05c4f
 pkgrel=1
 pkgdesc="Portable FPGA place and route tool (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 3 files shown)


Biswa96:
> it is desirable to wait until that one is uploaded

That "wait" is not required now. The updated dependencies are stored in "staging" section. CI uses those to build the dependent packages. See the "update using staging" section in CI build log.

BTW, wish ucrt64 build issues are fixed here.

umarcor:
> > it is desirable to wait until that one is uploaded
> 
> That "wait" is not required now. The updated dependencies are stored in "staging" section. CI uses those to build the dependent packages. See the "update using staging" section in CI build log.

In theory. In practice, GHDL's is a peculiar PKGBUILD recipe and there have been ordering issues before. Although those were fixed, I prefer to be on the safe side and wait a few days until GHDL is uploaded. There are no such issues with the packages in this PR, that's why I put them together.

> BTW, wish ucrt64 build issues are fixed here.

There are no build issues that I can see. There is a dependency issue only. Yosys depends on GHDL, which depends on Ada (https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ghdl/PKGBUILD#L80). As we discussed before, Ada is not available on UCRT64 yet.

umarcor:
I think this can be merged now.

PR #8957, opened 2021-06-18
libxslt: Fix build failure and date/time issue in libxslt

larskanis:
This new patch was applied upstream almost two years ago here:
  https://gitlab.gnome.org/GNOME/libxslt/-/commit/90c34c8bb90e095a8a8fe8b2ce368bd9ff1837cc

Without the patch a clean UCRT64 build (no previously installed UCRT64 packages) fails with:
```
  checking for x86_64-w64-mingw32-xml2-config... no
  checking for xml2-config... /ucrt64/bin/xml2-config
  configure: error: Could not find libxml2 anywhere, check ftp://xmlsoft.org/.
```

I verified, that the build issue is currently present on my local system as well as on github: https://github.com/larskanis/MINGW-packages/runs/2861341540?check_suite_focus=true

With this patch the build succeeds: https://github.com/larskanis/MINGW-packages/runs/2858716280?check_suite_focus=true

In addition to the build failure the current UCRT64 package `mingw-w64-ucrt-x86_64-libxslt-1.1.34-3-any.pkg.tar.zst` doesn't support date/time functions, which are usually part of libxslt. I don't know how this package was built and why this package doesn't support them, but I verified that both the local package and the github action artifact produced by this PR fix the issue. Therefore I'd like to ask for merging this PR and rebuilding the package libxslt.

The MINGW 32/64 packages support date/time functions in contrast to the UCRT64 package.

The original issue was raised by the nokogiri project in the following issue: https://github.com/sparklemotion/nokogiri/pull/2272#issuecomment-863283296

### DIFF
--- a/mingw-w64-libxslt/0009-fix-xml2-config-check-in-configure.patch
+++ b/mingw-w64-libxslt/0009-fix-xml2-config-check-in-configure.patch
@@ -1,0 +1,24 @@
+From 90c34c8bb90e095a8a8fe8b2ce368bd9ff1837cc Mon Sep 17 00:00:00 2001
+From: Nick Wellnhofer <wellnhofer@aevum.de>
+Date: Fri, 15 Nov 2019 11:53:11 +0100
+Subject: [PATCH] Fix xml2-config check in configure script
+
+A 'print' option has never been supported. After a recent change to
+libxml2, invalid options cause xml2-config to fail.
+---
+ configure.ac | 2 +-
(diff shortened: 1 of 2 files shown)


Biswa96:
If the patch file is an upstream commit please add a comment with that commit link. It'll help in future.

larskanis:
The link is in the commit message. Isn't that the right place?

Biswa96:
All right.

PR #8958, opened 2021-06-19
clang: llvm-rc don't rewrite default arch

jeremyd2019:
Fixes msys2/CLANG-packages#46

NOTE: this bug causes `CLANG32` to fail to build clang.  It can be built by first downgrading to 12.0.0-5.  @lazka how can this be handled?


Biswa96:
You could manually updated the package like clangarm64 and then rebuild in CI.

lazka:
I can build/upload manually if needed

lazka:
> I can build/upload manually if needed

done

jeremyd2019:
thanks

PR #8966, opened 2021-06-20
libdeflate: fix clang build

jeremyd2019:
Since CC was changed to "clang" in makepkg, the makefile got confused
and tried to use "clang" as AR.  This was not noticed because nobody
tried to build libdeflate under clang until CLANG32.


Biswa96:
pkgrel++ 💭

PR #8970, opened 2021-06-21
gtest: update to 1.11.0

MehdiChinoune:


mingw-w64-gtest/PKGBUILD
```diff
@@ -3,59 +3,54 @@
 _realname=googletest
```
lazka, line 3: I was recently wondering if we should rename this to "gtest". Any thoughts on that?

PR #8976, opened 2021-06-22
libaec: update to 1.0.5

kmilos:
Also enable SZIP replacement and rebuild HDF5


lazka:
For context, I guess this is a continuation of https://github.com/msys2/MINGW-packages/pull/7089#issuecomment-706120349

Thanks!

kmilos:
That's correct, the new version improves the CMake build recipe to include all the necessary files to replace szip.

PR #8978, opened 2021-06-22
new package bat and dust

naveen521kk:
https://github.com/sharkdp/bat

### DIFF
--- a/mingw-w64-bat/PKGBUILD
+++ b/mingw-w64-bat/PKGBUILD
@@ -1,0 +1,60 @@
+# Maintainer: Naveen M K <naveen521kk@gmail.com>
+
+_realname=bat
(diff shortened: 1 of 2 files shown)


Biswa96:
Another Rust clone 😩

naveen521kk:
> Another Rust clone

hmm, what do you mean by that?

Biswa96:
So, here are the clones we've
* ripgrep -- grep
* sccahce -- ccache
* rav1e -- dav1d
* bat -- cat
* dust -- du

naveen521kk:
what's the problem with that?

mati865:
Some of them aren't exactly clones:
- sccache - cloud based storage is rather big feature
- rav1e - it's the encoder, dav1d is the decoder

PR #8979, opened 2021-06-22
libgit2: Use abs path and ninja for cmake and use env for PKG_CONFIG_WITHOUT_PREFIX

1480c1:


lazka:
Alternative :P #8980

1480c1:
Either is fine with me, just wanted to get rid of the cmake patch if possible

Issue #8981, opened 2021-06-23
ffmpeg.exe is broken (again)


omichel:
The last update of ffmpeg broke it:
```bash
$ ffmpeg
ffmpeg version 4.4 Copyright (c) 2000-2021 the FFmpeg developers
  built with gcc 10.3.0 (Rev2, Built by MSYS2 project)
  configuration: --prefix=/mingw64 --target-os=mingw32 --arch=x86_64 --disable-debug --enable-dxva2 --enable-d3d11va --enable-fontconfig --enable-frei0r --enable-gnutls --enable-gpl --enable-libaom --enable-libass --enable-libbluray --enable-libcaca --enable-libcelt --enable-libdav1d --enable-libfreetype --enable-libgsm --enable-libmfx --enable-libmodplug --enable-libmp3lame --enable-libopencore_amrnb --enable-libopencore_amrwb --enable-libopenjpeg --enable-libopus --enable-librtmp --enable-libsoxr --enable-libspeex --enable-libsrt --enable-libtheora --enable-libvorbis --enable-libx264 --enable-libx265 --enable-libxvid --enable-libvpx --enable-libwebp --enable-libxml2 --enable-openal --enable-pic --enable-postproc --enable-runtime-cpudetect --enable-swresample --enable-version3 --enable-vulkan --enable-zlib --disable-doc --logfile=config.log --enable-shared
  libavutil      56. 70.100 / 56. 70.100
  libavcodec     58.134.100 / 58.134.100
  libavformat    58. 76.100 / 58. 76.100
  libavdevice    58. 13.100 / 58. 13.100
  libavfilter     7.110.100 /  7.110.100
  libswscale      5.  9.100 /  5.  9.100
  libswresample   3.  9.100 /  3.  9.100
  libpostproc    55.  9.100 / 55.  9.100
Hyper fast Audio and Video encoder
usage: ffmpeg [options] [[infile options] -i infile]... {[outfile options] outfile}...

Use -h to get full help or, even better, run 'man ffmpeg'

$ pacman -Syuu
:: Synchronizing package databases...
 mingw32              1204.5 KiB  1064 KiB/s 00:01 [#####################] 100%
 mingw64              1210.6 KiB  1085 KiB/s 00:01 [#####################] 100%
 ucrt64               1463.1 KiB  1100 KiB/s 00:01 [#####################] 100%
 clang64              1059.6 KiB   884 KiB/s 00:01 [#####################] 100%
[...]

PR #8982, opened 2021-06-23
aom: Fix library name in import library.

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/8981


Biswa96:
@1480c1 Any thoughts?

ImperatorS79:
Since ffmpeg has a problem, should not you bump ffmpeg pkgrel ?

1480c1 (approved):

PR #8988, opened 2021-06-24
cotire: update for CMake 3.20

Optiligence:


Biswa96:
Please increment `pkgrel` variable for this rebuild.

Optiligence:
there is no need since the CMake version that gets bumped is part of the version
(`1.8.1_3.19-1` → `1.8.1_3.20-1`)

Biswa96:
Just asking. Should the package version append another package version number?

Optiligence:
Other packages sync their versioning or have it in their name. (which wasn’t done here, i’m not the author of `cotire`)
I don’t think there is a »should«. I think it fits the way it is. It’s the primary and only dependency and wont work if the minor version doesn’t match.

Biswa96:
All right. Thanks for sharing.

PR #8992, opened 2021-06-25
curl: preserve UNC paths when normalizing/simplifying

dscho:
This is a fixed version of https://github.com/msys2/MINGW-packages/pull/8920.

The code to make cURL relocatable simplifies/normalizes paths, e.g. reducing multiple forward slashes to a single one.

This is okay in general, but at the beginning of the path, a double slash has a special meaning: it denotes a UNC path of the form
`//<host>/<path>`. It would therefore be a mistake to reduce those first two slashes to a single one: otherwise it would refer to the absolute path relative to the current directory's drive.

Original report: https://github.com/git-for-windows/git/issues/3266

So sad that I cannot Cc: the original author of the `pathtools.c` file. 😞

### DIFF
--- a/mingw-w64-curl/0001-Make-cURL-relocatable.patch
+++ b/mingw-w64-curl/0001-Make-cURL-relocatable.patch
@@ -22,10 +22,10 @@ Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
  configure.ac         |   1 +
  lib/Makefile.inc     |   5 +-
  lib/curl_config.h.in |   3 +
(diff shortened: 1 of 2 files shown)


dscho:
Let me make _real_ sure this time that the CI _actually_ built the package correctly.

dscho:
Looks like it is building it: https://github.com/msys2/MINGW-packages/pull/8992/checks?check_run_id=2915756901#step:9:25

dscho:
Seems this PR is actually good to go. @lazka my apologies for the previous PR that left things in an unbuildable state.

PR #8997, opened 2021-06-27
goffice: Add package

stahta01:
I am trying to get gnumeric to build and it needs goffice.

Tim S.

### DIFF
--- a/mingw-w64-goffice/PKGBUILD
+++ b/mingw-w64-goffice/PKGBUILD
@@ -1,0 +1,117 @@
+# Contributor: Tim Stahlhut <stahta01@gmail.com>
+
+_realname=goffice
+_sourcedir=${_realname}-git
+pkgbase=mingw-w64-${_realname}
+pkgname=(
+  "${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+)
+pkgver=0.10.49.0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


Biswa96:
A stable release would be preferred. Like this https://github.com/archlinux/svntogit-packages/blob/packages/goffice/trunk/PKGBUILD

stahta01:
I consider a git tag to be more stable than a git commit; if you think differently please explain!

Tim S.

Biswa96:
That's what I just said 😊 But this package uses git commit instead of tag, doesn't it?

stahta01:
I edited the pkgver() to reduce the confusion.

stahta01:
> That's what I just said  But this package uses git commit instead of tag, doesn't it?

To clarify for the other readers the Arch Linux package uses an plain git commit; while my package is using a git tag.

Tim S.

Biswa96:
And why do we need those bunch of variables and if...else statement?

stahta01:
I am guessing that some times that a git commit is needed to be used instead of a git tag.
I can get rid of some of them. But, it might be needed to add some of them back in the future.
I still do not under stand why the Arch Linux package uses a git commit instead of a git tag.

Tim S.

stahta01:
Decided to work on this package in the [MINGW-packages-dev repo](https://github.com/msys2/MINGW-packages-dev)

PR #8998, opened 2021-06-27
libaec: Fix cmake files install path.

Biswa96:


Biswa96:
@kmilos Any thoughts?

kmilos:
Sure, sorry for not catching that...

I'd just prefer to keep the sequential numbering of patches, so 5 in this case (there was a 1 that got dropped/upstreamed...)

Biswa96:
IMO, patches number are the count of current patches. git history shows when a patch was removed. But it's OK, as you wish ☺️

PR #9000, opened 2021-06-27
sip: remove replaces sip4, sip4 still exists as a package

jeremyd2019:
My PKGBUILD knowledge is somewhat lacking, but it seems to me that sip should not `replace` sip4, as sip4 still exists as a package (and is necessary to build the latest release of wxPython at least).  Note that sip declares to `conflict` with sip4, but not vice-versa, hopefully that is ok.

Also, the current development branch of wxPython now depends on sip 5.5, so if/when this winds up being released we may need to resurrect the sip5 package, and then remove the `replaces` sip5 here as well.

### DIFF
--- a/mingw-w64-sip/PKGBUILD
+++ b/mingw-w64-sip/PKGBUILD
@@ -6,10 +6,9 @@ _realname=sip
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-sip4")
-replaces=("${MINGW_PACKAGE_PREFIX}-sip4"
-          "${MINGW_PACKAGE_PREFIX}-sip5")
+replaces=("${MINGW_PACKAGE_PREFIX}-sip5")
 pkgver=6.1.0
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I've thought a couple of times now about resurrecting sip5 now, but it's been quite a while since a wxPython release, and maybe by the time they get around to a new release they'll have bumped the requirement again to sip 6.

Biswa96:
Let's invite @MehdiChinoune for his thoughts.

jeremyd2019:
It looks like at least something would need to change for sip6 support: wxWidgets/Phoenix#1916

lazka:
Thanks. Why does pyqt5-sip also replace python-sip(4)?

lazka:
(if there is a good chance that sip4 wont be needed in the future I'd welcome any work to remove it)

MehdiChinoune:
> Thanks. Why does pyqt5-sip also replace python-sip(4)?

I don't know, It was like that before I updated it.

MehdiChinoune:
> (if there is a good chance that sip4 wont be needed in the future I'd welcome any work to remove it)

When all the dependant packages update to use sip6 (Which is not straightforward).

PR #9001, opened 2021-06-27
try limiting clang32 to -j1 for packages known to have address space issues with lld

jeremyd2019:
This passed in CI on my fork.  The PR will give it another test in CI, and then hopefully it will work under autobuild.  (`aom` without these changes would succeed in CI but fail under autobuild.)


jeremyd2019:
All of the packages built in CI on the pull request too

PR #9003, opened 2021-06-28
Update mingw-w64

lazka:


Biswa96:
So quick! I was thinking to update this after pushing some changes. e.g. python-pywin32 and qt6-qtmultimedia needs bunch of symbols. We'd skip one update.

lazka:
after the update is before the update :)

PR #9008, opened 2021-06-29
reduce SDL2_Image's dependency chain

fabiangreffrath:
For the same reasons as outlined in https://github.com/fabiangreffrath/MINGW-packages/commit/53ce00433d4f41f6fb09470fdf5e8ec9b1ee2c42, this commit reduces SDL2_Image's dependency chain to only one hard dependency on libpng.

While PNG support is absolutely expected from an image processing library, support for TIF, JPG and WEBP is demoted to soft dependencies which rely on the corresponding libraries being installed and available to get opened at run-time.


Biswa96:
Why?

fabiangreffrath:
> Why?

For the same reasons as outlined in fabiangreffrath@53ce004

Biswa96:
That sounds like specific to your use case. Debian and ArchLinux both do the same as this package.

lazka:
> Debian and ArchLinux both do the same as this package.

Our target audience is more focused on bundling our build results compared to Arch/Debian, so imo we should consider it at least.

fabiangreffrath:
> You need to bump pkgrel

Done, thanks!

lazka:
If anyone hits any problems because of this please let us know.

PR #9021, opened 2021-07-01
gdcm: Fix build with clang, use ninja.

Biswa96:
Use CMake to disable getopt linking conflict which replaces the multiple
definition workaround in 28016b68b3c7057c4cee10c66e48402f0396042b commit.


Biswa96:
@emmenlau (creator of this package) Any thought on this change?

PR #9023, opened 2021-07-01
Update

naveen521kk:


kmilos:
There seems to be a problem w/ Pillow 8.3.0 and NumPy, so might want to hold of pushing it out to repo @lazka , I guess there will be and 8.3.1 soon...

PR #9025, opened 2021-07-01
python3.9: update to v3.9.6

naveen521kk:
Fixes https://github.com/msys2-contrib/cpython-mingw/issues/5


mingw-w64-python3.9/PKGBUILD
```diff
@@ -28,8 +28,8 @@ else
   pkgbase="mingw-w64-${_realname}${_pybasever}"
   pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}${_pybasever}")
 fi
-pkgver=${_pybasever}.5
-pkgrel=7
+pkgver=${_pybasever}.6
+pkgrel=2
```
jeremyd2019, line 32: Silly question, why start pkgrel at 2?

naveen521kk: Didn't notice. I set this as 1 and then run `update-pactches.sh` which bumped it again 😄

PR #9027, opened 2021-07-02
qt6: update to 6.1.2

MehdiChinoune:


Biswa96:
FYI, there is a build error in ucrt64 environment for qt6-virtualkeyboard. Build log can be found here https://packages.msys2.org/queue

PR #9031, opened 2021-07-02
mingw-w64-mesa: Update to 21.1.4

pal1000:


Biswa96:
If you would like to wait there is something going on with cpuid being redeclared. @1480c1

pal1000:
With 5fc97ca being quickly [reverted](https://github.com/msys2/MINGW-packages/commit/a075432182c5a0a4cde28b7d34b20723e21a4300#diff-d211bc7ba2b51ca6e8996b94c7718d6589a01aefc091330e2e6a9edcfe4b58fb), I think this should be merged as is.

pal1000:
Removed swr disabling commit. Thanks for heads up regarding the cause, @Biswa96.

pal1000:
Superseded by #9155.

PR #9032, opened 2021-07-02
gcc: backport patch for cpuid.h header guard

1480c1:
/cc @Biswa96

### DIFF
--- a/mingw-w64-gcc/0204-x86-Add-__cpuidex-and-include-guard-to-cpuid.h.patch
+++ b/mingw-w64-gcc/0204-x86-Add-__cpuidex-and-include-guard-to-cpuid.h.patch
@@ -1,0 +1,90 @@
+From 29e1039ca211c4e3f1053925eda0a81c57cbcb0c Mon Sep 17 00:00:00 2001
+From: "H.J. Lu" <hjl.tools@gmail.com>
+Date: Sat, 18 Jul 2020 04:43:10 -0700
+Subject: [PATCH] x86: Add __cpuidex and include guard to <cpuid.h>
+
+Add
+
+void __cpuidex (int __cpuid_info[4], int __leaf, int __subleaf);
+
+as well as include guard to <cpuid.h>.
+
+gcc/
+
+	PR target/95973
+	PR target/96238
+	* config/i386/cpuid.h: Add include guard.
+	(__cpuidex): New.
+
+gcc/testsuite/
+
+	PR target/95973
+	PR target/96238
+	* gcc.target/i386/pr95973.c: New test.
+---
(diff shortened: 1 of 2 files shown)


1480c1:
waiting for CI to tell if it actually applies, haven't tested locally, although idk build time of gcc

mati865:
You forgot to attach the patch.

1480c1:
darn

lazka:
> waiting for CI to tell if it actually applies, haven't tested locally, although idk build time of gcc

around 2h50 all together

lazka:
ok to merge?

1480c1:
Yeah, CI didn't complain, so it's probably fine

jeremyd2019:
I think this broke in combination with mingw-w64 version of `__cpuidex`:
https://github.com/msys2/MINGW-packages/pull/9082/checks?check_run_id=2981533749#step:9:496

/cc @lhmouse

lazka:
yeah, https://github.com/mirror/mingw-w64/commit/d2374f898457b0f4ea8bd4084a94f2dafc87a99a

So we need to revert this until gcc 11 from what I see.

lazka:
reverted in master

jeremyd2019:
Now it fails with:
```
  In file included from C:/_/mingw-w64-openal/src/openal-soft-openal-soft-1.21.0/alc/cpu_caps.cpp:10:
  D:/a/_temp/msys/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/10.3.0/include/cpuid.h:228:1: error: redefinition of 'unsigned int __get_cpuid_max(unsigned int, unsigned int*)'
    228 | __get_cpuid_max (unsigned int __ext, unsigned int *__sig)
        | ^~~~~~~~~~~~~~~
  In file included from D:/a/_temp/msys/msys64/mingw64/x86_64-w64-mingw32/include/intrin.h:70,
                   from C:/_/mingw-w64-openal/src/openal-soft-openal-soft-1.21.0/alc/cpu_caps.cpp:7:
  D:/a/_temp/msys/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/10.3.0/include/cpuid.h:228:1: note: 'unsigned int __get_cpuid_max(unsigned int, unsigned int*)' previously defined here
    228 | __get_cpuid_max (unsigned int __ext, unsigned int *__sig)
        | ^~~~~~~~~~~~~~~
```

I think this is due to the lack of an include guard, which was added in the patch reverted, but that patch also added a new function that caused a conflict with mingw-w64.  I wonder if it would be possible to "split" that patch and only apply the include guard, not the new function.

lazka:
>  I wonder if it would be possible to "split" that patch and only apply the include guard, not the new function.

worth a try I guess

PR #9036, opened 2021-07-03
curl: Separate pathtools from patch files

Biswa96:


Biswa96:
@dscho Please review.

mingw-w64-curl/PKGBUILD
```diff
@@ -46,12 +46,16 @@ if [ -n "${_namesuff}" ]; then
 fi
 options=('staticlibs')
 source=("https://github.com/curl/curl/releases/download/${_realname}-${pkgver//./_}/${_realname}-${pkgver}.tar.xz"{,.asc}
+        "https://raw.githubusercontent.com/msys2/MINGW-packages/master/mingw-w64-pathtools/pathtools.c"
+        "https://raw.githubusercontent.com/msys2/MINGW-packages/master/mingw-w64-pathtools/pathtools.h"
```
jeremyd2019, line 50: I would suggest using a commit hash instead of 'master' here, sort of the same premise as using a commit hash instead of 'master' in a -git package.

Biswa96: Done.

mingw-w64-curl/PKGBUILD
```diff
@@ -60,10 +65,15 @@ validpgpkeys=('914C533DF9B2ADA2204F586D78E11C6B279D5C91'  # Daniel Stenberg
 
 prepare() {
   cd "${srcdir}/${_realname}-${pkgver}"
-  rm -f lib/pathtools.h lib/pathtools.c > /dev/null 2>&1 || true
+
+  rm -f lib/pathtools.h lib/pathtools.c
+  cp -v "${srcdir}/pathtools.c" lib/pathtools.c
+  cp -v "${srcdir}/pathtools.h" lib/pathtools.h
```
dscho, line 71: The `rm` is superfluous. Just use `cp -f "${srcdir}"/pathtools.[ch] lib/`

mingw-w64-curl/PKGBUILD
```diff
@@ -60,10 +65,15 @@ validpgpkeys=('914C533DF9B2ADA2204F586D78E11C6B279D5C91'  # Daniel Stenberg
...
+
```
dscho, line 72: I'd rather leave whitespace alone in this patch (its mission is not to change whitespace).

mingw-w64-curl/PKGBUILD
```diff
@@ -60,10 +65,15 @@ validpgpkeys=('914C533DF9B2ADA2204F586D78E11C6B279D5C91'  # Daniel Stenberg
...
   patch -p1 -i "${srcdir}/0001-Make-cURL-relocatable.patch"
   patch -p1 -i "${srcdir}/0002-nghttp2-static.patch"
   patch -p1 -i "${srcdir}/0003-libpsl-static-libs.patch"
+
```
dscho, line 76: I'd rather leave whitespace alone in this patch (its mission is not to change whitespace).

mingw-w64-curl/PKGBUILD
```diff
@@ -156,4 +166,5 @@ package() {
 
   local PREFIX_DEPS=$(cygpath -am ${MINGW_PREFIX})
   sed -s "s|${PREFIX_DEPS}|${MINGW_PREFIX}|g" -i ${pkgdir}${MINGW_PREFIX}/bin/curl-config
+  sed -s "s|${PREFIX_DEPS}|${MINGW_PREFIX}|g" -i ${pkgdir}${MINGW_PREFIX}/lib/pkgconfig/libcurl.pc
```
dscho, line 169: Why is this needed?

mingw-w64-curl/PKGBUILD
```diff
@@ -45,13 +45,18 @@ if [ -n "${_namesuff}" ]; then
   conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 fi
 options=('staticlibs')
+_commit="3f788969a9ef371633341f36e57461c2b7586b3f"
```
dscho, line 48: I know @jeremyd2019 suggested to hard-code the commit, but I would rather suggest to do this _completely_ differently: use the path `../mingw-w64-pathtools/pathtools.c`. There is absolutely no need to

1. go stale (as a hard-coded commit would be prone to do)
2. go via https:// when we already have the file locally.

dscho (commented):
I like the direction this is going.

mingw-w64-curl/PKGBUILD
dscho: I guess we could even go one step further and copy directly from `${srcdir}/../../mingw-w64-pathtools/`. That would also avoid having to update the SHA-256 sum whenever the `pathtools.[ch]` file(s) change.

mingw-w64-curl/PKGBUILD
jeremyd2019: I'd agree with that approach too.  I just didn't like referring to 'master' in the url.

What about sha256sums?  They will still need updating when pathtools gets updated (or will they be 'SKIP'?).

What about source package?  Will ../mingw-w64-pathtools/pathtools.c get included in the source package?  Will it still be reasonable to reproduce an old build if necessary?

mingw-w64-curl/PKGBUILD
Biswa96: That pkgconfig file contains full path reference of msys2 installation. That command replaces that with relative root path e.g. /mingw64 etc.

mingw-w64-curl/PKGBUILD
Biswa96: But one has to clone this whole repository to build this package. With the pathtools link as source, one has to only download the PKGBUILD file and patches for building.

mingw-w64-curl/PKGBUILD
dscho: So this is a bug fix in and of itself, and should not be mixed in with the pathtools patch, right?

mingw-w64-curl/PKGBUILD
dscho: > 
> 
> But one has to clone this whole repository to build this package. With the pathtools link as source, one has to only download the PKGBUILD file and patches for building.

I really don't think that this is a use case we need to be supporting.

> What about sha256sums?

If the `pathtools.[ch]` are (force-)copied as part of the `prepare` function in the `PKGBUILD` then no, they do not need to be copied.

> Will ../mingw-w64-pathtools/pathtools.c get included in the source package?

Isn't everything in `src/` bundled in the source tar? Then yes, `pathtools.[ch]` is very much part of that source package. That's the same thing as with the current process, where those two files are created by a `patch` call, by the way.

Biswa96: I have not problem with the suggestion. @lazka or @elieux agree with this idea?

jeremyd2019: I think that only things in the sources array are included in the source package, not the contents of `src/ ` after `prepare`.

dscho: Oh? @Biswa96 could you test that? If that is true, then my suggestion is no good, and you will probably want to add something like `"../mingw-w64-pathtools/pathtools.c" "../mingw-w64-pathtools/pathtools.h"` to the `sources` array. I am not sure what to do about the `sha256sums`. Part of me wants to suggest `SKIP`, to avoid it from going stale whenever `mingw-w64-pathtools` gets updated, the other part of me wants to require wholesale updates to all dependencees' `sha256sums` whenever `mingw-w64-pathtools` is updated.

jeremyd2019: I think updating all dependencies is a good thing, because they should be rebuilt with any update anyway (this is effectively a statically linked library)

Biswa96: OK, I am going with Jeremy's suggestion.

dscho: You mean the suggestion where the relative path is used, right? Because that will make it easier to automate updating all the dependencees.

Biswa96: I mean the current commit. IMO, I am not completely agree with the relative path. And when pathtools will be changed the dependent packages can be updated just like other dependencies.

dscho: > I mean the current commit. IMO, I am not completely agree with the relative path. And when pathtools will be changed the dependent packages can be updated just like other dependencies.

They can be updated, but it's much harder to

1. Find the dependencees using a concise pattern
2. Update them in an automated way because you not only have to update the SHA-256, but _also_ have to update the hardcoded commit hash _and_ that commit might be in a PR at the time of this change, i.e. subject to rewriting (in which case that painfully hardcoded commit hash will become invalid).

It also won't work well with forked repositories (although it is currently possible to download objects in GitHub repositories that actually only exist in forks, but this is technically incorrect behavior on which you rely).

Also, it will be much harder to ensure that all dependencees are up to date.

I find no upsides to that approach.

dscho:
For the record, [I still think that this PR took the wrong turn](https://github.com/msys2/MINGW-packages/pull/9036#discussion_r665310130) when it was decided to use https:// URLs instead of relative paths for sources that are shipped in the very same repository and whose revisioning should be tightly coupled.

In the current shape, it would set things up to be a real pain to maintain.

Biswa96:
Understood ✌️

PR #9043, opened 2021-07-03
box2d: Update to 2.4.1

podsvirov:


lazka:
This removes the static build. Is that a problem?

podsvirov:
As I understood both are not supported now.  There are no dependent packages in the repository - I can't say for sure about the problems. Overall, this update contains a series of changes that partially break backward compatibility.

lazka:
ok, sounds good.

(if anyone finds this because static libs broke, please open an issue here)

PR #9083, opened 2021-07-04
libmysofa: Update to 1.2, use ninja

Biswa96:
The idea is shamelessly copied/stolen from https://github.com/msys2/MINGW-packages/pull/8606

Pinging some bots @jeremyd2019 @neheb @ZachBacon

### DIFF
--- a/mingw-w64-libmysofa/001-mingw-dont-use-nuget.patch
+++ b/mingw-w64-libmysofa/001-mingw-dont-use-nuget.patch
@@ -1,7 +1,7 @@
 --- a/src/CMakeLists.txt
 +++ b/src/CMakeLists.txt
-@@ -13,28 +13,9 @@
- configure_file (config.h.in ${PROJECT_SOURCE_DIR}/src/config.h)
+@@ -13,28 +13,9 @@ use_c99()
+ configure_file(config.h.in ${PROJECT_SOURCE_DIR}/src/config.h)
  
  if(NOT MSVC)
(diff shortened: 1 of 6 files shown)


jeremyd2019:
I believe this just needs some pkgconf magic to define `MYSOFA_STATIC_DEFINE` for static builds, and it'd be good for both static and dynamic builds (assuming whoever is doing static builds uses pkgconf).

Biswa96:
Why? There is already a static library built with shared library. And dependent packages are linked with dlls. Aren't they?

jeremyd2019:
If somebody wanted to *use* the static library, they'd need to define `MYSOFA_STATIC_DEFINE` otherwise they'd get the API's with dllimport attributes, and the linker would be sad.  If you add the line `Cflags.private: -DMYSOFA_STATIC_DEFINE=1` to the .pc file, users of pkgconf who ask for `pkg-config --cflags --static` will get that.

Biswa96:
And where did you get the `MYSOFA_STATIC_DEFINE` name from?

jeremyd2019:
From the generated mysofa_export.h.  https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html 
> If the same sources are used to create both a shared and a static library, the uppercased symbol ${BASE_NAME}_STATIC_DEFINE should be used when building the static library
This will cause the export macros to expand to nothing when building the static library.

neheb:
I hold no opinion.

PR #9086, opened 2021-07-04
Rust fixes

### LINKED ISSUE #9010: rust brokenness re `mingw_initcharmax`
As reported on Discord:

```
error: linking with `x86_64-w64-mingw32-gcc`...
undefined reference to `mingw_initcharmax'
```

likely related to https://www.mail-archive.com/mingw-w64-public@lists.sourceforge.net/msg18871.html

> mati865: Oh, it's Rust bug. It should not use self contained mode.
As a workaround using RUSTFLAGS=-Clink-self-contained=no should help.

>  mati865: We should remove self-contained directory from our package so this is only a temporary workaround.

>  mati865: Quickest workaround for this bug from msys2 would be rebuilding Rust package.

### LINKED ISSUE #9047: rust: hardcodes linking libgcc, doesn't build on clang prefix
```
lld: error: unable to find library -lgcc_eh
lld: error: unable to find library -lgcc
```

It also seems to hardcode libpthread.a, don't know if that's necessary/desired on a clang/ucrt prefix or if it would be better off using Windows thread API directly.

mati865:
Fixes #9010
Fixes #9047

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -35,13 +35,17 @@ source=("https://static.rust-lang.org/dist/${_realname}c-${pkgver}-src.tar.gz"
         "0001-add-missing-libs.patch"
         "0004-unbundle-gcc.patch"
         "0005-win32-config.patch"
-        "0006-Update-cc.patch")
+        "0006-Update-cc.patch"
+        "0006-Update-cc.patch"
+        "0007-clang-subsystem.patch")
 sha256sums=('5cf7ca39a10f6bf4e0b0bd15e3b9a61ce721f301e12d148262e5ba968ab825b9'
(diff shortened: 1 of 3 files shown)


mingw-w64-rust/PKGBUILD
```diff
@@ -35,13 +35,19 @@ source=("https://static.rust-lang.org/dist/${_realname}c-${pkgver}-src.tar.gz"
         "0001-add-missing-libs.patch"
         "0004-unbundle-gcc.patch"
         "0005-win32-config.patch"
-        "0006-Update-cc.patch")
+        "0006-Update-cc.patch"
+        "0006-Update-cc.patch"
```
jeremyd2019, line 39: Oops, listed twice.  But sha256sum listed twice too, so no error

mati865: Good catch.

PR #9088, opened 2021-07-04
gcc: update to 11.1.0

lazka:


oscarfv:
I don't see the i386 PEH fix here. Is that intentional?

lazka:
> I don't see the i386 PEH fix here. Is that intentional?

Are you sure? it's rebased on master.

oscarfv:
> > I don't see the i386 PEH fix here. Is that intentional?
> 
> Are you sure? it's rebased on master.

I missed that bit. Sorry for the noise.

oscarfv:
My build failed at the usual point, so it would be a surprise if this PR succeeds.

I suggest either going through the path @revelator suggests or drop Ada on i686 altogether. It is a dying platform that attracts less and less attention from upstream and it is very likely that future releases will be problematic too.

There are about a dozen packages that depend on Ada, mostly developer tools. It doesn' t look as if dropping Ada on i686 would cause any major problem to our users or to the MSYS2 ecosystem.

mati865:
It's not only about Ada. We should not ship compiler with broken exception handling.

oscarfv:
> It's not only about Ada. We should not ship compiler with broken exception handling.

But again: is there proof that the exception handling is broken beyond Ada? (supposing that the broken thing is the EH) The Ada compiler/runtime might be using on a specific way that does not affect C/C++.

And are we sure that the broken thing is on 11 but not on 10? Because the fact that the bootstrap completes when initiated with gcc-10 sjlj points to gcc-10's DWARF EH as the culprit. Let's remember that gcc-11 successfully compiles itself on the bootstrap process, so we have an indication that 10 is defective while 11 is fine.

It would be interesting to bootstrap 11 from @revelator's 11 packages. If it succeeds, IMO we would have a strong indication that whatever is broken in 10 was fixed on 11.

lazka:
> It would be interesting to bootstrap 11 from @revelator's 11 packages. If it succeeds, IMO we would have a strong indication that whatever is broken in 10 was fixed on 11.

I can try that if I get the packages

revelator:
uploading now :)

revelator:
and he is right if it turns out we can bootstrap it again using my packages it indicates that something might have been broken in the earlier version (not the first time either, it does happen).

revelator:
https://sourceforge.net/projects/cbadvanced/files/bootstrap/

give it a little if all the packages are not there yet, sourceforge can be a bit slow at times to show newly uploaded items.

lazka:
thanks, I'll try tomorrow.

oscarfv:
Just tried it with PR #8320. It fails **earlier** (stage 1) on Ada too. @lazka : please go ahead with your build, maybe my setup is botched.

```
echo timestamp > s-i386-bt
cp -p ../../gcc-11.1.0/gcc/ada/sinfo.ads ../../gcc-11.1.0/gcc/ada/sinfo.adb ../../gcc-11.1.0/gcc/ada/xsinfo.adb ../../gcc-11.1.0/gcc/ada/csinfo.adb ada/bldtools/sinfo
(cd ada/bldtools/einfo; gnatmake -q xeinfo ; ./xeinfo einfo.h )
mkdir -p ada/bldtools/nmake
(cd ada/bldtools/sinfo; gnatmake -q xsinfo ; ./xsinfo sinfo.h )
rm -f ada/bldtools/nmake/sinfo.ads ada/bldtools/nmake/nmake.adt ada/bldtools/nmake/xnmake.adb ada/bldtools/nmake/xutil.ads ada/bldtools/nmake/xutil.adb
mkdir -p ada/bldtools/treeprs
cp -p ../../gcc-11.1.0/gcc/ada/sinfo.ads ../../gcc-11.1.0/gcc/ada/nmake.adt ../../gcc-11.1.0/gcc/ada/xnmake.adb ../../gcc-11.1.0/gcc/ada/xutil.ads ../../gcc-11.1.0/gcc/ada/xutil.adb ada/bldtools/nmake
rm -f ada/bldtools/treeprs/treeprs.adt ada/bldtools/treeprs/sinfo.ads ada/bldtools/treeprs/xtreeprs.adb
mkdir -p ada/bldtools/snamest
(cd ada/bldtools/nmake; gnatmake -q xnmake ; ./xnmake -b nmake.adb ; ./xnmake -s nmake.ads)
rm -f ada/bldtools/snamest/snames.ads-tmpl ada/bldtools/snamest/snames.adb-tmpl ada/bldtools/snamest/snames.h-tmpl ada/bldtools/snamest/xsnamest.adb ada/bldtools/snamest/xutil.ads ada/bldtools/snamest/xutil.adb
cp -p ../../gcc-11.1.0/gcc/ada/treeprs.adt ../../gcc-11.1.0/gcc/ada/sinfo.ads ../../gcc-11.1.0/gcc/ada/xtreeprs.adb ada/bldtools/treeprs
cp -p ../../gcc-11.1.0/gcc/ada/snames.ads-tmpl ../../gcc-11.1.0/gcc/ada/snames.adb-tmpl ../../gcc-11.1.0/gcc/ada/snames.h-tmpl ../../gcc-11.1.0/gcc/ada/xsnamest.adb ../../gcc-11.1.0/gcc/ada/xutil.ads ../../gcc-11.1.0/gcc/ada/xutil.adb ada/bldtools/snamest
touch ada/GNAT_DATE
(cd ada/bldtools/treeprs; gnatmake -q xtreeprs ; ./xtreeprs treeprs.ads )
(cd ada/bldtools/snamest; gnatmake -q xsnamest ; ./xsnamest )
mkdir -p ada/libgnat
mkdir -p ada/libgnat
cp -p ../../gcc-11.1.0/gcc/ada/libgnat/s-excmac__gcc.ads ada/libgnat/s-excmac.ads
cp -p ../../gcc-11.1.0/gcc/ada/libgnat/s-excmac__gcc.adb ada/libgnat/s-excmac.adb
echo "pragma Style_Checks (Off);" >tmp-sdefault.adb
gnatmake: "xsinfo.adb" compilation error
echo "with Osint; use Osint;" >>tmp-sdefault.adb
cp ../../gcc-11.1.0/gcc/gcc-ar.c gcc-nm.c
/bin/sh: line 1: ./xsinfo: No such file or directory
make[3]: *** [../../gcc-11.1.0/gcc/ada/Make-generated.in:45: ada/sinfo.h] Error 127
make[3]: *** Waiting for unfinished jobs....
```

revelator:
hmm does indeed sound like something might have broken on your end :S but just to be sure ill try setting up a local version to bootstrap from.

revelator:
local bootstrap also fails here in the exact same spot, so it seems this is indeed an upstream bug :S 

```
echo "pragma Style_Checks (Off);" >tmp-sdefault.adb
echo "with Osint; use Osint;" >>tmp-sdefault.adb
cp ../../gcc-11.1.0/gcc/gcc-ar.c gcc-nm.c
cp ../../gcc-11.1.0/gcc/gcc-ar.c gcc-ranlib.c
echo "package body Sdefault is" >>tmp-sdefault.adb
rm -f mm_malloc.h
TARGET_CPU_DEFAULT="" \
HEADERS="auto-host.h ansidecl.h config/i386/xm-mingw32.h" DEFINES="USED_FOR_TARGET " \
/bin/sh ../../gcc-11.1.0/gcc/mkconfig.sh tconfig.h
gnatmake: "xsinfo.adb" compilation error
echo "   S0 : constant String := \"/mingw32/\";" >>tmp-sdefault.adb
cat ../../gcc-11.1.0/gcc/config/i386/gmm_malloc.h > mm_malloc.h
/bin/sh: line 1: ./xsinfo: No such file or directory
(echo "@set version-GCC 11.1.0"; \
 if [ "" = "experimental" ]; \
 then echo "@set DEVELOPMENT"; \
 else echo "@clear DEVELOPMENT"; \
 fi) > gcc-vers.texiT
make[3]: *** [../../gcc-11.1.0/gcc/ada/Make-generated.in:45: ada/sinfo.h] Error 127
make[3]: *** Waiting for unfinished jobs....
CMD //c echo @set srcdir `echo /D/mingw-w64-gcc/src/build-i686-w64-mingw32/gcc/../../gcc-11.1.0/gcc | sed -e 's|\\([@{}]\\)|@\\1|g'` >> gcc-vers.texiT
```

oscarfv:
Tried `make -j1` to discard a Makefile race. Failed at `xsinfo.adb` too.

Then compiled a "Hello world" Ada program to check that the gcc 11 Ada toolset provided by @revelator's packages is minimally functional. It worked fine.

Then tried building today's gcc 11 snapshot. Failed at `xsinfo.adb` too.

Some web searching hinted at a filesystem related problem (file path length exceeded some limit). Launched the build in `/d/g` directory. Same result.

oscarfv:
OTOH a simple `/path/to/src/configure --disable-multilib && make` on a MINGW32 shell goes along happily until stage 2.

Maybe our `makepkg-mingw` and `PKGBUILD` are a bit too much over-engineered?

Configuring like this:

```
 ../gcc-11-20210703/configure --disable-multilib --host=i686-w64-mingw32 --target=i686-w64-mingw32 --build=i686-w64-mingw32 --enable-languages=c,lto,c++,fortran,ada,jit --disable-rpath --disable-win32-registry --disable-nls --disable-werror --disable-symvers --enable-threads=posix --enable-shared --enable-static
```

triggers the `xsinfo.adb` error. Looks like an opportunity for a command-line bisection ;-)

revelator:
guess we need to report this upstream, maybe fixed in the next version then :/.
besides the same options oscarfv used i also tried with differing -gdwarf-versions sadly to no avail.

revelator:
interresting, though odd unless something changed recently in the autotools build scripts.
there was a patch added in 2019 that did away with some duplicated build lines in ada not sure if this might have triggered it, though i suspect version 10 would also have been affected then.

revelator:
there was a similar bug back on gcc-4.8.0 i dont remember how it was fixed unfortunatly, though at that time the package was built using mingw.orgs old msys shell which sadly does not work to well with win10, it did have one plus though as it did not hide some errors which made finding the bugger a lot easier ugh...

revelator:
sorry was gcc-4.1.0 so long ago my memory fades :S the fix was compiling it with the old gcc-3.3.6 sadly this does not apply here as that build was 32 bit only.  The culprit was that 4.1.0 produced corrupt code when bootstrapping https://gcc.gnu.org/bugzilla/show_bug.cgi?id=23894

oscarfv:
> OTOH a simple `/path/to/src/configure --disable-multilib && make` on a MINGW32 shell goes along happily until stage 2.

Of course it does: Ada is an optional language, so the above command does not build it :-/

Tried `configure --disabe-multilib --enable-languages=c,ada` with and without `CPPFLAGS=-D__USE_MINGW_ANSI_STDIO=1`. Both ways keep failing at `xsinfo.adb`. The `--disable-multilib` bit is just because I know the build will fail without it.

At least we can file a bug report upstream without throwing at them an screenfull of configure options.

revelator:
Aye

revelator:
i commented in the upstream bug report about the new build error with the successfull bootstrap.

revelator:
https://aur.archlinux.org/packages/gcc-ada-git/

huh arch is also running into problems with ada :S

mati865:
> https://aur.archlinux.org/packages/gcc-ada-git/
> 
> huh arch is also running into problems with ada :S

This was almost 2 years ago, this AUR package uses latest commit again.

revelator:
must have been some heavy changes to cause so many problems heh. noticed it now also has preliminary support for C++20 modules via libcody which still seems to be in a preliminary state. i wonder if the best thing to do would be waiting untill the next version of it is out ? we could still report bugs upstream to make things happen sooner. unless someone has an ephiphany and spots the culprit.

oscarfv:
Building i686 10.3 PKGBUILD with @revelator's 11.1 packages also fails with

```
gnatmake: "xsinfo.adb" compilation error
```

Seems that the problem is on 11.1 itself (either on those packages or on upstream, dunno).

revelator:
it would seem so indeed :(

oscarfv:
For the record:

A new [build bug was fixed upstream](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101377). Applying that patch to the latest GCC 11 snapshot does not fix our problem.

On that bug report, the reporter says `I have bootstrapped GCC 11.1.1 on both {x86_64,i686}-w64-mingw32 and seen no problem so far.` The maintainer that responded to our bug report also says that the bootstrap builds fine for him. I'm mystified about why we seem to be the only ones having problems building 11 on i686. One of the things I tried past week was to build 11 without local patches, and it failed the same way.

Maybe the problem is our binutils?

revelator:
could be any number of things... but as an experiment we could try building it without aslr/dep using the last stable binutils version which did not have this option.

stahta01:
Has anyone tried my patch to see if it helps.
[0025-gcc-11-Set-default-dwarf-version-to-4.zip](https://github.com/msys2/MINGW-packages/files/6836775/0025-gcc-11-Set-default-dwarf-version-to-4.zip)

Tim S.

mati865:
^ just tried it and got infamous `Error 1` without any useful info.

revelator:
anyone tried bootstrapping it with the builds from winlibs ?, seems he has no problems building it with dwarf exceptions so maybe one of our patches ?.

oscarfv:
@revelator : is Ada included in winlib's distribution? It is not mentioned on the web page and a quick look at the i686 binary package seems to indicate that it is missing.

However, the upstream maintainer claims that the bootstrap works fine on his tests.

We had problems on the past with MSYS2 bugs spoiling the build on subtle ways.  It would be interesting to see if the build fails on Cygwin with our configuration. At least we need to know the details about the setup of those who successfully build gcc with Ada on i686.

BTW, One of the multiple things I tried was building with patches removed, same problem.

revelator:
aye ada is missing in the winlibs build, so it might be something from msys2 that is causing it to fail ? hmm interresting.

revelator:
[gcc-11.1.0-upstream_fixes-1.zip](https://github.com/msys2/MINGW-packages/files/6875994/gcc-11.1.0-upstream_fixes-1.zip)

found this while scouring other sites for problems, not sure if it is any help but since it is an upstream fix i doubt it hurts.

lazka:
(11.2 should be out this week)

Astrum-polaris:
### Status update

I rebuild MINGW32 gcc-10.3 with **sjlj exception** and successfully bootstrapped **gcc-11.2-sjlj** from **gcc-10.3-sjlj**
So it seems like dwarf exception handling is broken on MINGW32

The bad news is that SJLJ has a performance penalty and `libgcc_s_dw2-1.dll` has to keep around for some time otherwise the new compiler will not work (I just do `cp libgcc_s_{sjlj,dw2}-1.dll` on package step)

And packages depend on gcc-libs will need a rebuild as well.

revelator:
aye it is pretty strange though that you can successfully bootstrap the dwarf version with the sjlj version but you cannot do the same if bootstrapping it with the dwarf version. something is definatly screwy here...

revelator:
btw using sjlj will break building rust and luajit as both depend on the dwarf version so :S

mati865:
Yup, there are multiple packages which will not work with SJLJ unless somebody modifies/rewrites their exception handling.

revelator:
guess we need to try a mingw-w64 bootstrap in another environment, if it works there we can narrow it down some.

revelator:
seems like arch made a change to gnats exception handling to allow older gcc packages to bootstrap it since gnats exception model 
changed starting from version 10. it might be a guess but im backporting the change now to see if it changes anything.

the original patch which was accepted upstream is here ->
[571eda6.zip](https://github.com/msys2/MINGW-packages/files/6911079/571eda6.zip)

revelator:
here is a backport of the above, if someone has the time try bootstrapping gcc-11 with atleast gcc-10  as older versions dont support the new exception model used in gnat.

[0009-backport-571eda6.zip](https://github.com/msys2/MINGW-packages/files/6911110/0009-backport-571eda6.zip)

Astrum-polaris:
Well, I will call it a revert, not a backport

The original commit is here
https://github.com/gcc-mirror/gcc/commit/5d733372faa97c1c3943a20a252d000db37c738b

revelator:
heh sorry tired been up for 36 hours :P

Astrum-polaris:
Upon seeing #9289, I decided to give this another try
This time I purge `-static-libgcc` from stage1/boot ldflags and CI succeeded  🎉
https://github.com/Astrum-polaris/MINGW-packages/runs/3248922992

Ajaja:
Any status update?

lazka:
> Any status update?

The remove `-static-libstdc++ -static-libgcc` workaround works, but a rebuild using that gcc11 still fails without it, so we have two options:

* Remove `-static-libstdc++ -static-libgcc` from now on and depend on it for mingw32 (I have no ideas what problems this can cause)
* Have a gcc10 package for mingw32 and gcc11 for all others

MeerMusik:
Removing ``-static-libstdc++ -static-libgcc`` would disable support for static Libs / disallow the "End User" like me to build static Libs etc. if i remember correctly?

If that is the case: Yikes. But better a updated GCC Version than an outdated one. And therefore possibly insecure one: I am not sure how much back porting you are doing to be honest, with you being manpower and time limited.

DISCLAIMER: Private User. Hobby Projects. No Enterprise / Business etc. needs.

revelator:
sounds like more static exception problems :S well i can rebuild gcc-11.1.0 with itself with just --static-libstdc++ but any notion of --static-libgcc breaks it totally. also seems to break exceptions on other projects if using static libgcc so id say it is definatly broken...

crillion:
> 
> 
> Removing `-static-libstdc++ -static-libgcc` would disable support for static Libs / disallow the "End User" like me to build static Libs etc. if i remember correctly?
> 
> If that is the case: Yikes. But better a updated GCC Version than an outdated one. And therefore possibly insecure one: I am not sure how much back porting you are doing to be honest, with you being manpower and time limited.
> 
> DISCLAIMER: Private User. Hobby Projects. No Enterprise / Business etc. needs.

If then I couldn't build static libs, I'd prefer to wait more time (even some months) rather than get an updated version of the compiler. Searching around I found this regression tracking, don't know if it's exactly the problem with ada exceptions but maybe it's related or useful :
[https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101947](url)

oscarfv:
> If then I couldn't build static libs,

You can build static libs, but they will depend on the dynamic gcc runtime.

> I'd prefer to wait more time (even some months) rather than get an updated version of the compiler.

There is no guarantee that this will be fixed anytime soon. And if a new gcc package is released you are not obliged to upgrade.

>  Searching around I found this regression tracking, don't know if it's exactly the problem with ada exceptions but maybe it's related or useful :
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101947

That was introduced about ten days ago, so no, that's not the problem.

revelator:
indeed this problem is a bit older than that, first started to crop up about gcc 8 allthough it was no near as broken as it is now.

MehdiChinoune:
Why not just disable ada for mingw32?
How many users use ada 3% at max.
How many users still relying on (32bit) mingw32 5% at max I assume.
So the upgraded is holded because of (0.03*0.05) 0.15% of users.

mati865:
What if Ada isn't the cause and it only shows static GCC libraries are broken even more than ever?
IMO we could go with option "Have a gcc10 package for mingw32 and gcc11 for all others" and if somebody wants GCC 11 for i686 they should get in touch with GCC devs on how to fix static libraries.

MehdiChinoune:
> What if Ada isn't the cause and it only shows static GCC libraries are broken even more than ever?
> IMO we could go with option "Have a gcc10 package for mingw32 and gcc11 for all others" and if somebody wants GCC 11 for i686 they should get in touch with GCC devs on how to fix static libraries.

Time will tell, but for now the only certain thing is that gcc fails to build **Ada** on **mingw32**

oscarfv:
> > What if Ada isn't the cause and it only shows static GCC libraries are broken even more than ever?
> > IMO we could go with option "Have a gcc10 package for mingw32 and gcc11 for all others" and if somebody wants GCC 11 for i686 they should get in touch with GCC devs on how to fix static libraries.
> 
> Time will tell, but for now the only certain thing is that gcc fails to build **Ada** on **mingw32**

Furthermore, the failure happens while running an executable generated by gcc 10. This points to an existing problem since at least version 10.

BTW, problems with exceptions and static libraries were already a thing when I started using MinGW32 20 years ago, on the gcc 2.95 era. Eventually a hack was created to allow throwing exceptions across modules, but it depended on a shared library. AFAIK today's dynamic gcc runtime implements that same hack. Please note that a monolithic executable that throws an exception from code called by a Win32 callback (a common scenario) also depends on the capability of the gcc runtime to unwind through modules. Worse: modules not compiled by gcc itself. IIRC the problem was revived when the exception model changed from SJLJ to DWARF.

revelator:
huh sounds like this stems back from the days of the mingw.org runtime dll days. Kinda odd that it only started getting this bad recently we had dwarf exceptions for quite some time.

MehdiChinoune:
I tried to make a new package gcc10 but it conflicts with gcc-11.
Passing "--program-suffix=10" removes the conflicts between of the executables, but I didn't find a way to remove the conflict between these runtime libraries.
```
/mingw32/bin/libatomic-1.dll
/mingw32/bin/libgcc_s_seh-1.dll
/mingw32/bin/libgomp-1.dll
/mingw32/bin/libquadmath-0.dll
/mingw32/bin/libssp-0.dll
/mingw32/bin/libstdc++-6.dll
```
Are these libraries backward-compatible?, If that is true I could make "mingw32-gcc-libs" replaces and provides "mingw32-gcc10-libs".

revelator:
hmm libstdc++ has added C++ module support with gcc-11 so probably not :/

mati865:
> Are these libraries backward-compatible?, If that is true I could make "mingw32-gcc-libs" replaces and provides "mingw32-gcc10-libs".

Yes, they are backwards compatible.

revelator:
im at a point where i backported the old mingw.org patch used for allowing to throw exceptions across dll boundaries, together with
the TDM developer we have recently updated this patch considerably so it now works rather well with 64 bit also. The TDM patch would be fine to use for this but it defaults to linking to the static gcc runtimes so i removed those parts of it. The shared memory unwinder only gets called when using static linking anyway, so should work just fine without all the voodoo :)

revelator:
allready tried it out on my own fork and it works fine for bootstrapping the 32 bit build there, but since mine uses sjlj exceptions i need somone to try bootstrapping a dwarf build with the variant im about to build.

revelator:
[0160-shared-memory-unwinder.zip](https://github.com/msys2/MINGW-packages/files/7117763/0160-shared-memory-unwinder.zip)

this is the patch that adds the ability for static runtime libraries to throw exceptions across dll boundaries, originally it stems from the earliest mingw compiler which at that time could not bulid the dll versions of the runtimes anyway. it has been changed quite a lot in the meantime to also support 64 bit and a bug that cropped up in later versions of gcc building the gnat compilers.
it works by using a shared memory interface allowing the static runtimes to call on exceptions across external dll boundaries adding a few exports which are newer used anywhere except internally so it is pretty safe to use.

first bootstrap must have the -static-libgcc omitted for it to build correctly, but bootstrapping a second build can be done without removing it (atleast it worked here but im using sjlj exceptions in my fork).

mati865:
> it has been changed quite a lot in the meantime to also support 64 bit

Haven't it always worked on 64 bits?

> atleast it worked here but im using sjlj exceptions in my fork

Astrum-polaris has successfully built GCC 11 with SJLJ without any patching.

revelator:
there where a few bugs that needed ironing out in the original, most which became appareant with newer gcc versions int vs size_t that kinda stuff. Also there was a bug which broke ASLR in it but this new version works just fine.

Could astrum-polaris also build the 32 bit ada compiler using sjlj ?, asking cause i had some problems doing that when using -static-libgcc even with sjlj exceptions but seems to be working now though i did update our gcc package to gcc-11.2.0 so maybe they fixed some stuff.

revelator:
p.s a word of caution since this patch uses some libtool magic to split the interface for the shared memory exception handler so that it is only active in the static version, so you need to rebuild libtool as well if you intend to bootstrap gcc with a build using this patch.

revelator:
[0160-shared-memory-unwinder.zip](https://github.com/msys2/MINGW-packages/files/7123570/0160-shared-memory-unwinder.zip)
ugh stupid mistake in unwind-generic.h fixed in this one.

revelator:
just bootstrapped a full build of the dwarf compiler using this with no problem :) seems it works.

oscarfv:
> just bootstrapped a full build of the dwarf compiler using this with no problem :) seems it works.

I'm afraid that we will need much more than a successful bootstrap for saying that it works. It is very nice to have this possibility, though.

revelator:
https://sourceforge.net/projects/cbadvanced/files/bootstrap/07-09-2021/

build is here, try it out together with the patch. remember to rebuild libtool if upgrading from an older gcc as else you will get a build error.

revelator:
darn still fails if using -static-libgcc damn thing is broken beyond repair :( odd thing though if bootstrapped using an sjlj compiler it works just fine and the resulting gnat compiler happily compiles anything i throw at it just not itself sigh...

revelator:
hmm well seems something did change though as i now got a bit more info on the crash, still not sure what happens there ? but here is the output ->

```
gnatmake: "xsinfo.adb" compilation error
echo "with Osint; use Osint;" >>tmp-sdefault.adb
cp ../../gcc-11.2.0/gcc/gcc-ar.c gcc-nm.c
cp ../../gcc-11.2.0/gcc/gcc-ar.c gcc-ranlib.c
/bin/sh: line 1: ./xsinfo: No such file or directory
make[3]: *** [../../gcc-11.2.0/gcc/ada/Make-generated.in:45: ada/sinfo.h] Error 127
make[3]: *** Waiting for unfinished jobs....
echo "package body Sdefault is" >>tmp-sdefault.adb
echo "   S0 : constant String := \"/mingw32/\";" >>tmp-sdefault.adb
echo "   S1 : constant String := \"/mingw32/lib/gcc/i686-w64-mingw32/11.2.0/adainclude/\";" >>tmp-sdefault.adb
echo "   S2 : constant String := \"/mingw32/lib/gcc/i686-w64-mingw32/11.2.0/adalib/\";" >>tmp-sdefault.adb
echo "   S3 : constant String := \"i686-w64-mingw32/\";" >>tmp-sdefault.adb
/bin/sh ../../gcc-11.2.0/gcc/../move-if-change tmp-mlib.h multilib.h
echo "   S4 : constant String := \"/mingw32/lib/gcc/i686-w64-mingw32/11.2.0/\";" >>tmp-sdefault.adb
echo "   function Include_Dir_Default_Name return String_Ptr is" >>tmp-sdefault.adb
echo timestamp > s-mlib
echo "   begin" >>tmp-sdefault.adb
echo "      return Relocate_Path (S0, S1);" >>tmp-sdefault.adb
echo "   end Include_Dir_Default_Name;" >>tmp-sdefault.adb
echo "   function Object_Dir_Default_Name return String_Ptr is" >>tmp-sdefault.adb
echo "   begin" >>tmp-sdefault.adb
echo "      return Relocate_Path (S0, S2);" >>tmp-sdefault.adb
echo "   end Object_Dir_Default_Name;" >>tmp-sdefault.adb
echo "   function Target_Name return String_Ptr is" >>tmp-sdefault.adb
echo "   begin" >>tmp-sdefault.adb
echo "      return new String'(S3);" >>tmp-sdefault.adb
echo "   end Target_Name;" >>tmp-sdefault.adb
echo "   function Search_Dir_Prefix return String_Ptr is" >>tmp-sdefault.adb
echo "   begin" >>tmp-sdefault.adb
echo "      return Relocate_Path (S0, S4);" >>tmp-sdefault.adb
echo "   end Search_Dir_Prefix;" >>tmp-sdefault.adb
echo "end Sdefault;" >> tmp-sdefault.adb
mv -f tmp-sdefault.adb ada/sdefault.adb
touch ada/stamp-sdefault
mv -f ada/bldtools/snamest/snames.ns ada/snames.ads
mv -f ada/bldtools/treeprs/treeprs.ads ada/treeprs.ads
mv -f ada/bldtools/nmake/nmake.ads ada/nmake.ads
mv -f ada/bldtools/snamest/snames.nb ada/snames.adb
mv -f ada/bldtools/nmake/nmake.adb ada/nmake.adb
mv -f ada/bldtools/snamest/snames.nh ada/snames.h
touch ada/stamp-nmake
touch ada/stamp-snames
```

revelator:
findings so far...
bootstrap with dwarf compiler = fail removing -static-libgcc does not work anymore either (fails in gnat build).
bootstrap with sjlj compiler = success even works with both -static-libgcc and -static-libstdc++ now (gnat also builds just fine).
the shmem patch i posted uses a debug trap which would be triggered if something in the unwinders was crashing, but it newer is.

might actually be a path length problem but then would that not also affect an sjlj build ?.
dwarf5 might not work as intended with gnat on windows maybe, but the sjlj bootstrapped dwarf gnat works just fine for everything other than bootstrapping itself ?.

so atm if we want gnat id suggest using an sjlj gcc build as the bootstrap compiler atleast untill someone finds out where things break

revelator:
https://gcc.gnu.org/legacy-ml/gcc-patches/2019-10/msg00068.html

oh boy... i suggest someone has a look at the above :S

revelator:
[mingw-w64-gcc.zip](https://github.com/msys2/MINGW-packages/files/7139360/mingw-w64-gcc.zip)

can anyone try building it with this ? i might have gone lucky but the after i added a small patch from the ada maintainer which is not yet upstream it seems to build gnat with dwarf exceptions again.

there are some local changes in the PKGBUILD which are specific to my own fork, feel free to use it if you find it interresting.
i also added a few files that where missing in the package installers.

revelator:
```
diff -urN gcc-11.2.0.a/gcc/dwarf2out.c gcc-11.2.0.b/gcc/dwarf2out.c
--- gcc-11.2.0.a/gcc/dwarf2out.c	2021-07-28 08:55:07.000000000 +0200
+++ gcc-11.2.0.b/gcc/dwarf2out.c	2021-09-08 22:41:07.552294600 +0200
@@ -18990,7 +18990,10 @@
     case ROUND_DIV_EXPR:
     case TRUNC_DIV_EXPR:
     case EXACT_DIV_EXPR:
-      if (TYPE_UNSIGNED (TREE_TYPE (loc)))
+      /* The sign bit is expected to be never set for sizetypes.  */
+      if (TYPE_UNSIGNED (TREE_TYPE (loc))
+	  && TREE_TYPE (loc) != sizetype
+	  && TREE_TYPE (loc) != bitsizetype)
 	return 0;
       op = DW_OP_div;
       goto do_binop;
```

this is the small patch for ada, atm it is conjoined with the shared memory unwinder patch but if it works standalone ill make a new one just for this.

mati865:
Extracted the zip without making any changes into clean master branch checkout and the build failed with no specific error: [gcc-err.txt](https://github.com/msys2/MINGW-packages/files/7139951/gcc-err.txt)

revelator:
hmm odd the first error points to this line in make-lang.in

```
b.o:
	-> mkdir -p $(dir $@) // wtf!!!
	$(CC) -c $(ALL_ADAFLAGS) $(ADA_INCLUDES) $< $(ADA_OUTPUT_OPTION)
	@$(ADA_DEPS)
```

strangely it builds here but i bootstrapped my dwarf version using the sjlj compiler to get the missing files installed first.
then i rebuilt the dwarf version using that.

revelator:
well heres something mighty interresting bootstrapping 32 bit dwarf gcc-10.3.0 with 32 bit dwarf gcc-11.2.0 works with ada Oo

revelator:
yep gnat with dwarf exceptions is broken, compiling gprbuild-bootstrap with the sjlj bootstrapped dwarf version aborts at the first ada source file. Even the simplest ada example will segafult it so there is something not right with it.

oscarfv:
I build an Ada-less gcc 11.2.0 and on i686 the most trivial test that throws an exception fails:

```
int main() {
  try {
    throw 13;
  }
  catch(...) {
    return 1;
  }
  return 0;
}
```

```
$ g++ foo.cpp
$ ./a.exe
terminate called after throwing an instance of 'int'
terminate called recursively
```

Using a gcc built with debug info, I see on `_Unwind_RaiseException` that it returns here:

```
      _Unwind_FrameState fs;

      /* Set up fs to describe the FDE for the caller of cur_context.  The
	 first time through the loop, that means __cxa_throw.  */
      code = uw_frame_state_for (&cur_context, &fs);

      if (code == _URC_END_OF_STACK)
	/* Hit end of stack with no handler found.  */
	return _URC_END_OF_STACK;
```

This causes a call to `terminate` on the calling function and then another call to `terminate`. Obviously the next question is what happens in `uw_frame_state_for`, but I know zilch about this code and can't assign meaning to what I'm seeing.

BTW, mingw32 gdb is broken (`catch throw` runs the program and gdb exits.) Had to use MSYS2 gdb.

oscarfv:
One more detail: compiling with our clang has the same effect, so the problem is in the gcc support libraries which are also used by clang.

This could be a great bisection project, but bootstrapping gcc takes two hours here, so...

revelator:
stack handler broken maybe ?, hmm there was a report on the gcc bug list for m68k which seemed to have a similar problem with memory corruption, i think lazka mentioned our problem there.

jeremyd2019:
Weirdly reminiscent of #9091's unwinder issues, though that's LLVM's libunwind rather than libgcc (though the function names look much the same - did they end up both using libunwind or something?)

revelator:
does clang even get installed beforehand in a ci build ?.

revelator:
thinking about it i tried bootstrapping it with a different gcc than ours and hit the same problem (no libunwind present in that build) so that cant be it :/.

jeremyd2019:
No, looking at the code I think it's just that the function names are similar because they're doing similar things

revelator:
aye clangs libunwind uses several unwinders, the main difference compared to the gcc unwinders is that all clangs unwinders are present in the unwinder library though it defaults to one specific model you can actually select a totally different one.

revelator:
@oscarfv seems it is crashing in some newer code that was added as far back as gcc-7 in the atomic fde path...

specifically in this function _Unwind_Find_FDE where it returns a null pointer ->

```
  if (__builtin_expect (!__atomic_load_n (&any_objects_registered,
					  __ATOMIC_RELAXED), 1))
    return NULL;
```

jeremyd2019:
That check *should* only be returning NULL if *no* `.eh_frame` sections were registered with libgcc (which would indeed mean that nobody would be able to catch an exception).  That registration is supposed to happen early in the module load process.

revelator:
might be it then :) though something must have broken it after as it worked in gcc-7 as far as im aware.
lets see what disabling it nets us, if it starts working after then we actually have something to present to the gcc devs.

jeremyd2019:
I would also try setting a breakpoint/stepping through `__gcc_register_frame`.

revelator:
good point, probably need to build the first dwarf version with this disabled using the sjlj compiler to avoid any nastiness, then ill try bootstrapping it using that.

revelator:
damnation still fails in the exact same spot when building gnat :(

oscarfv:
Replacing gcc-11 `libstdc++.dll` with gcc-10 makes the simple test case work. Replacing `libgcc_s_dw2-1.dll` has no effect. So either this is a different problem from Ada's failure or `libstdc++.dll` has the problematic code inlined.

Anyway, a quick method for pinpointing the source of the problem is simply stepping through all the exception-related code in parallel for executables built with each version, noting where the execution diverges, reverting the associated commit, checking if the problem gets fixed, rinse, repeat. The EH-related machinery has very few changes from version 10 to 11, so it is faster than `git-bisect`.

Actually, reverting all EH-related changes from 10 to 11 in libgcc probably would fix the problem without introducing regressions ;-)

jeremyd2019:
Based on what @revelator said, I'm guessing it's failing to register its `.eh_frame` with the unwinder

revelator:
one mighty interresting thing is if i build it using a very old gcc-4.9.1 it actually fails in the libstdc++ build but not in gnat, strangely it first fails on stage2 of the bootstrap so i guess oscarfv might be onto something with the problem being in the libstdc++ code.
gcc-4.9.1 uses dwarf version 2 not 4.

revelator:
aye unfortunatly it still fails so that was not it :/ in fact the code there was made as part of an effort to fixup dwarf exceptions for mingw* builds i found out, so disabling it would lead to even more problems.

so far any effort has been in vain and i dont know enough about exception models to do much further :(, guess it will be upto upstream to get this one sorted out.

GitMensch:
Yay, so I guess that means building 11.2 release + that patch integrated for this single release, correct? Note: that may not worth it if the 11.3 will stay in the 2-4 months release., because then 11.3 is likely to be released in the next two weeks and this can be adjusted to build the fixed 11.3 instead.
Any insights on 11.3?

lazka:
Closing in favor of #9825

jakitliang:
A history issue:

https://salsa.debian.org/toolchain-team/gcc/-/blob/5ac26f96b568a7258b4a2900f8d99799d94bb8a9/debian/patches/dwarf5-revert-default.diff

Never applyed to msys's gcc and mingw-w64-gcc

So that in around 2021-05 to 2023-01, gdb can not correctly debug the program when use MSYS2's MINGW

See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98860

And also: https://github.com/skeeto/w64devkit/commit/6fff86fc4bce8a734036db69b6f6585a85f662aa

I mean, **The entire MSYS2 was unfit to serve as a reliable desktop toolchain during the 2021–2023 period.**

PR #9096, opened 2021-07-05
Update libvips to 8.11.2

larskanis:
This removes the dependency to `opencl-icd` for two reasons:
- `opencl-icd` is not available in the UCRT64 repository
- I could not find any use of that library within libvips

I hope the UCRT64 build can be triggered that way (in addition to MINGW32+MINGW64).

### DIFF
--- a/mingw-w64-libvips/PKGBUILD
+++ b/mingw-w64-libvips/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=libvips
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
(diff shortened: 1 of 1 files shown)


larskanis:
I also tried to build a recent version of [`opencl-icd-git`](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-opencl-icd-git/), but to no avail. The header files of OpenCL (`c:\msys64\ucrt64\include\CL\cl_icd.h`) seems to be broken on MINGW - but I know too less about OpenCL to fix this.

lazka:
I also can't find anything referencing it upstream, so looks good to me.

PR #9104, opened 2021-07-06
libssh: Remove install script, use ninja

Biswa96:
This fixes installation in ucrt and clang environments.
The cmake files are now modified with sed command to
add _IMPORT_PREFIX replacing full mingw prefix path.


Biswa96:
@heljkon Did qemu work in your setup? Without the upstream qemu patch in this PR, I got

```
ERROR:../qemu-6.0.0/util/oslib-win32.c:61:qemu_try_memalign: assertion failed: (is_power_of_2(alignment))
```

Any thoughts on this change?

heljkon:
This error line doesn't look unfamiliar to, even if I don't remember the context. But I compiled qemu without external libslirp and capstone and on cygwin, too. And I had a look at qemu commits as well.

I can't remember experiencing it in my tests, but if the deps evolve...

Could you give me a command line to reproduce it?

BTW I guess the next qemu release will have its RCs soon

Biswa96:
I was trying to boot Win10 PE ARM64 with Qemu using [this article][1]. Though it shows the boot logo but does not boot. I forgot what I did a year ago. But it is not Qemu's fault.

[1]: https://withinrafael.com/2018/02/12/boot-arm64-builds-of-windows-10-in-qemu/

PR #9114, opened 2021-07-09
GTK 2/3: Fix keyboard input when using IMEs

jtanx:
Upstream issues: 
* https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3741/
* https://github.com/fontforge/fontforge/issues/4577

Definitely required for GTK2, GTK3 seems to deal with IMEs differently and avoids this particular code path, but the underlying GDK calls are still affected. 

e.g. to reproduce (this is against the 32 bit mingw builds - may be relevant because behaviour is dependent on uninitialised state):

* Set keyboard layout to the Japanese IME in half-width alphanumeric/direct input mode
* Open gtk-demo (GTK2 demo)
* Open the 'Dialog and Message boxes' demo
* Type some characters in the text boxes - observe half-width katakana instead of alphanumeric input

### DIFF
--- a/mingw-w64-gtk2/0023-gdkkeys-win32.c-fix-initialisation-of-key_state-in-u.patch
+++ b/mingw-w64-gtk2/0023-gdkkeys-win32.c-fix-initialisation-of-key_state-in-u.patch
@@ -1,0 +1,28 @@
+From c9753b78854a188ffeeced8fd997e6e5ff7fde70 Mon Sep 17 00:00:00 2001
+From: Jeremy Tan <jtanx@outlook.com>
+Date: Fri, 9 Jul 2021 12:35:44 +0800
+Subject: [PATCH] gdkkeys-win32.c: fix initialisation of key_state in
+ update_keymap
+
+It apparently worked by chance in the past, but now causes e.g.
+alphanumeric characters to be interpreted as half-width katakana
+when using the Japanese IME.
+---
+ gdk/win32/gdkkeys-win32.c | 1 +
+ 1 file changed, 1 insertion(+)
+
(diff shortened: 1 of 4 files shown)


Biswa96:
Separate commits for gtk2 and gtk3 would be better. Also add the comment in gtk2 pkgbuild like in gtk3 pkgbuild.

jtanx:
Done.

Jehan:
GTK 3.24.30 just got released but it had a tarball bug (a missing file for Windows builds), which just got fixed. As a consequence, there will be a followup release soon. Cf. maintainer comment: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3743

Since this other patch got merged just after 3.24.30, I am guessing it will be in the fixup release (unless they do one with just the tarball fix). So it might be even simpler/nicer to just wait a day or 2 for this fixup release, and update the GTK3 version. What is in this new GTK 3.24.30 is huge for us at GIMP (with Windows Ink support, finally! 🥳) so I am really looking forward having an updated MSYS2 package for our next dev release. 🙂

Biswa96:
IMO, this small bugfix would not cost much. Imagine next version has some bugs. We can check if the previous package build has that issue or not. There are many dependencies and compiler changes happened recently, so we could also eliminate any dependency issue for any bugs in next gtk2/3 versions. Just thoughts 💭

lazka:
@Jehan feel free to ping (or ask in discord) for updates

Jehan:
Anyway this specific issue was also affecting us apparently, so at the very least it's very useful that we get the GTK+2 backport! Thanks all! 🙂

For the GTK3 side, I mostly hope MSYS2 package will update soon. I'll ping if necessary as proposed. 😉

PR #9116, opened 2021-07-09
drmingw: Update to 0.9.3, use ninja

Biswa96:


jeremyd2019:
Did you have any luck with clang?  I looked at this at one point and got discouraged.

Biswa96:
Coming soon... 😉

Biswa96:
I would like to hear some words from @jannick0. Can this workaround be unstreamed?

jeremyd2019:
by default clang identifies itself as some old version of GCC, apparently too old for that version check.  There's also a commandline option to change what version of GCC it claims to be, but I don't know what other fallout might result from that.

I seem to recall hitting some other issue once I got past that cmake check, but maybe the newer version didn't have that problem...  That's nice

Biswa96:
I have tried to build CodeLite package with this dependency but it fails with clang due to some C++ STL reason.

PR #9117, opened 2021-07-09
sundials: New package

gharveymn:
We currently have the CVODE subproject, but not the main SUNDIALS project. The SUNDIALS project is used by GNU Octave, for example.

Please take a look at the check function and comment it out if needed. The tests for MPI require an executable which I'm not sure potential automatic build systems will have.

This is my first attempt at packaging, so maybe look a bit more carefully for mistakes.

### DIFF
--- a/mingw-w64-sundials/0001-sundials-link-errors.patch
+++ b/mingw-w64-sundials/0001-sundials-link-errors.patch
@@ -1,0 +1,484 @@
+diff -Naur sundials-5.7.0.orig/cmake/SundialsBuildOptionsPost.cmake sundials-5.7.0/cmake/SundialsBuildOptionsPost.cmake
+--- sundials-5.7.0.orig/cmake/SundialsBuildOptionsPost.cmake	2021-01-29 15:14:52.000000000 -0600
++++ sundials-5.7.0/cmake/SundialsBuildOptionsPost.cmake	2021-07-08 18:20:57.773489500 -0500
+@@ -204,3 +204,24 @@
+                 DEPENDS_ON ENABLE_PETSC PETSC_FOUND
+                 ADVANCED)
+ list(APPEND SUNDIALS_BUILD_LIST "BUILD_SUNNONLINSOL_PETSCSNES")
++
++set(SUNDIALS_NVECTOR_F2C_TYPES)
++if(BUILD_NVECTOR_SERIAL)
++  list(APPEND SUNDIALS_NVECTOR_F2C_TYPES "serial")
++  set(SUNDIALS_NVECTOR_F2C_serial_SUFFIX "")
++endif()
++
++if(BUILD_NVECTOR_PARALLEL)
++  list(APPEND SUNDIALS_NVECTOR_F2C_TYPES "parallel")
++  set(SUNDIALS_NVECTOR_F2C_parallel_SUFFIX "_parallel")
++endif()
++
++if(BUILD_NVECTOR_OPENMP)
++  list(APPEND SUNDIALS_NVECTOR_F2C_TYPES "openmp")
++  set(SUNDIALS_NVECTOR_F2C_openmp_SUFFIX "_openmp")
++endif()
++
++if (BUILD_NVECTOR_PARALLEL)
++  list (APPEND SUNDIALS_NVECTOR_F2C_TYPES "pthreads")
++  set (SUNDIALS_NVECTOR_F2C_pthreads_SUFFIX "_pthreads")
++endif ()
+diff -Naur sundials-5.7.0.orig/src/arkode/fcmix/CMakeLists.txt sundials-5.7.0/src/arkode/fcmix/CMakeLists.txt
+--- sundials-5.7.0.orig/src/arkode/fcmix/CMakeLists.txt	2021-01-29 15:14:53.000000000 -0600
++++ sundials-5.7.0/src/arkode/fcmix/CMakeLists.txt	2021-07-08 14:00:57.312960900 -0500
+@@ -38,35 +38,37 @@
+   farksparsemass.c
+   )
+ 
+-sundials_add_library(sundials_farkode
+-  SOURCES
+-    ${farkode_SOURCES}
+-  OBJECT_LIBRARIES
+-    sundials_fnvecserial_obj
+-    sundials_fsunmatrixband_obj
+-    sundials_fsunmatrixdense_obj
+-    sundials_fsunmatrixsparse_obj
+-    sundials_fsunlinsolband_obj
+-    sundials_fsunlinsoldense_obj
+-    sundials_fsunlinsolspbcgs_obj
+-    sundials_fsunlinsolspfgmr_obj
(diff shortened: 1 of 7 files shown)


lazka:
lgtm, besides there being lots of build patches, but I guess you know what you are doing.

Does this mean this supersedes cvode and we can drop it?

@okhlybov do you have any comments re the petsc and superlu_dist comments in the PKGBUILD here?

okhlybov:
1. The SuperLU's index size must match that of the libraries it depends upon. As everything (OpenBLAS to start with) in MSYS2 is built with 32-bit index, there is no option without breaking the existing packages.
BTW why one might need the 64-bit indices on Windows anyway?

2. @gharveymn what kind of documentation do you need besides the PKGBUILD itself? However, some clarification notes on the build suffix ofc should be introduced.

3. On to the integration issues. Take a look at the supplementary compile-time package `petsc-build` which contains source & build tree which retains configuration files from the actual PETSc build. There is the [SLEPc](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-slepc) package which is tightly integrated with PETSc and as such is strictly dependent on the `petsc-build`.

gharveymn:
@okhlybov
1. I see, I misunderstood the issue. I've updated the build configuration to reflect that and also added additional support for SuperLU_DIST and PETSc (with some small patches). That also fixes the issues that were occurring with the LAPACK tests. 
As to why 64-bit indexing would be nice? Well it's not _that_ difficult to hit 4B elements. And sure, scientific computing is probably better done on Linux, but some students might need access to those tools without learning a new OS. MATLAB for example ships with 64-bit BLAS.

2. Yeah, so the build suffix is required because of issues with the `noundefined` requirement. The libraries which are patched were intended to be mix-in libraries which have some symbols that are undefined, and are to be provided by a few different libraries (namely, nvecserial, nvecopenmp, etc.). This is fine on other systems, but doesn't work with DLLs, so for shared libraries we need to explicitly provide different flavors.  
My original solution was poorly thought out and would have silently tied some people to the `serial` flavor, and also produced redundant static libraries. The updated solution is to provide a mix-in static library (the same as other systems), as well as specific shared libraries (\<name\>_{serial,openmp,parallel,pthreads}_shared.dll[.a]). If a user wants to use the shared library they must specify which flavor.
Is there a specific place where we should put this information or should I just put a comment in the PKGBUILD?

3. Updated the PKGBUILD with support for PETSc. I chose the `dmo` flavor because SUNDIALS requires MPI support and prefers a real scalar interface over complex.
Thanks for the tip, and apologies for the rude comment. I was just a bit frustrated and intended to remove it before committing.

4. Done.

@lazka 
Most of the patches are pretty small. Would you prefer it if I were to consolidate them?

While it does supercede CVODE, it also has a lot of additional dependencies that people might not want. Maybe just keep it around?

okhlybov:
> The updated solution is to provide a mix-in static library (the same as other systems), as well as specific shared libraries (<name>_{serial,openmp,parallel,pthreads}_shared.dll[.a]). If a user wants to use the shared library they must specify which flavor.

The `_shared` suffix is redundant as static, dynamic and import libraries do not coinicide anyway (`lib.a`, `lib.dll`, `lib.dll.a`).
Moreover, it will put extra burden on users as they will have to modify the command line's `-l` flag depending on library type.

As for naming, it is better to have fully qualified library names for the sake of consistency while providing a kind of alias to generic library. On UNIX this would have normally been done with symlinks, but this is not the way on Windows for obvious reasons, so there is an alternative way of using the archiver's neat `T` feature which creates a kind of import library for static libraries.

Take a looks at my [MUMPS](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-mumps/PKGBUILD) package where I faced the same problem.

gharveymn:
> The _shared suffix is redundant as static, dynamic and import libraries do not coinicide anyway (lib.a, lib.dll, lib.dll.a).
Moreover, it will put extra burden on users as they will have to modify the command line's -l flag depending on library type.

Sorry, the `_shared` suffix there was a typo. The actual patch doesn't include that in the library names.

That thin archive trick is really cool. I've added the commands to the PKGBUILD to create thin archives for each of the flavors. Linking works as expected.

lazka:
small patches is fine.

Thanks everyone.

lazka:
something failed:

```
2021-07-19T20:05:21.3644068Z -- Checking if SuperLU_DIST works with SUNDIALS... OK
2021-07-19T20:05:21.3688625Z -- Found KLU: D:/a/_temp/msys/msys64/mingw64/lib/libklu.dll.a  
2021-07-19T20:05:21.3690511Z -- KLU_LIBRARIES:   D:/a/_temp/msys/msys64/mingw64/lib/libklu.dll.a;D:/a/_temp/msys/msys64/mingw64/lib/libamd.dll.a;D:/a/_temp/msys/msys64/mingw64/lib/libcolamd.dll.a;D:/a/_temp/msys/msys64/mingw64/lib/libbtf.dll.a;D:/a/_temp/msys/msys64/mingw64/lib/libsuitesparseconfig.dll.a
2021-07-19T20:05:21.3697563Z -- KLU_INCLUDE_DIR: D:/a/_temp/msys/msys64/mingw64/include
2021-07-19T20:05:22.1179892Z -- Checking if KLU works... OK
2021-07-19T20:05:22.1215237Z -- Finding PETSC using PETSC_INCLUDES and PETSC_LIBRARIES
2021-07-19T20:05:22.1216498Z CMake Error at cmake/tpl/FindPETSC.cmake:322 (message):
2021-07-19T20:05:22.1223151Z   Could not determine PETSc include directory from PETSC_INCLUDES:
2021-07-19T20:05:22.1223748Z 
2021-07-19T20:05:22.1225580Z     PETSC_INCLUDES=D:/a/_temp/msys/msys64/mingw64/src/petsc-3.14.5/dmo/includeD:/a/_temp/msys/msys64/mingw64/src/petsc-3.14.5/include
2021-07-19T20:05:22.1226703Z 
2021-07-19T20:05:22.1227167Z Call Stack (most recent call first):
2021-07-19T20:05:22.1228560Z   cmake/tpl/SundialsPETSC.cmake:52 (find_package)
2021-07-19T20:05:22.1233963Z   cmake/SundialsSetupTPLs.cmake:112 (include)
2021-07-19T20:05:22.1234877Z   CMakeLists.txt:167 (include)
2021-07-19T20:05:22.1235211Z 
2021-07-19T20:05:22.1235424Z 
2021-07-19T20:05:22.1235929Z -- Configuring incomplete, errors occurred!
2021-07-19T20:05:22.1239797Z See also "C:/_/M/mingw-w64-sundials/src/build-mingw-w64-x86_64/CMakeFiles/CMakeOutput.log".
2021-07-19T20:05:22.1241034Z See also "C:/_/M/mingw-w64-sundials/src/build-mingw-w64-x86_64/CMakeFiles/CMakeError.log".
2021-07-19T20:05:22.2105225Z ==> ERROR: A failure occurred in build().
2021-07-19T20:05:22.2418089Z     Aborting...
```

PR #9118, opened 2021-07-10
vala: disable graphviz dependency for clang

jeremyd2019:
This requires disabling building valadoc.

Also enabled clang32.


jeremyd2019:
I was trying to get librsvg to build for clang64, but unfortunately that failed for all MSYSTEMs (even after adding a makedepends on `${MINGW_PACKAGE_PREFIX}-gtk-doc).

mati865:
What was the error for librsvg?

lazka:
librsvg is failing for some weeks now

PR #9124, opened 2021-07-10
python: Update to 3.9.6

lazka:
* [x] Figure out list of packages to rebuild
* [x] bump pkgrel for those packages


jeremyd2019:
I think @danyeaw had a script for that.

danyeaw:
```
pacman -Fqx "/python3.8/" | grep 'mingw64/' | cut -f2- -d/ | \
	cut --complement -f3 -d- | while read -r dir; do
	if [ -d "$dir" ]; then
		echo "Updating pkgrel for $dir"
		cd "$dir"
		awk -F"=" 'BEGIN{OFS="="} { if ($1=="pkgrel") {print $1,$2=$2+1} else {print $0}}' \
		PKGBUILD > temp && mv temp PKGBUILD
		cd ..

	else 
		echo "$dir does not exist"
	fi
done
```

jeremyd2019:
Hmm, and that's sufficient?  I would be trying to overengineer something, to try to get from package to %PKGBASE% to source dir.  https://gist.github.com/jeremyd2019/dbdabd1f91dd87d97f58ba25497ce851#file-clang32-py-L103-L113

lazka:
Thanks, I went with `pacman -Fqx '/python3\.8/' '.*\.py.*' '.*-cpython-.*'` and did the pkgbase ones manually

lazka:
let's go

PR #9140, opened 2021-07-12
OpenCV: Make SFM patch more robust

ferdnyc:
(Apologies for the noise, I guess this _is_ still needed! Not sure why, it was thought to be addressed upstream. Well, back to updating the patch...)

This PR updates the 0103 patch ("sfm module linking') to be compatible with OpenCV 4.5.3, and just generally more robust (due to it inserting a new line of code, instead of modifying an existing one).

I don't have an OpenCV 4.5.3 PR ready yet, but the updated patch is safer and works just fine with the existing 4.5.2 as well. The current patch **will** start failing with OpenCV 4.5.3, whereas this updated one will not.

No real need for a rebuild, I'd say, if it's possible to commit without one. Applying the new patch is a build-time-only change that can wait until the next time a build is necessary for other reasons.


lazka:
thanks

PR #9146, opened 2021-07-15
svt-av1: new package

kmilos:
https://gitlab.com/AOMediaCodec/SVT-AV1

Seems restricted to x86_64

### DIFF
--- a/mingw-w64-svt-av1/PKGBUILD
+++ b/mingw-w64-svt-av1/PKGBUILD
@@ -1,0 +1,54 @@
+# Maintainer: Miloš Komarčević <miloskomarcevic@aim.com>
+
+_realname=SVT-AV1
+pkgbase=mingw-w64-svt-av1
+pkgname=("${MINGW_PACKAGE_PREFIX}-svt-av1")
+pkgver=0.8.7
+pkgrel=1
+pkgdesc="Scalable Video Technology AV1 encoder and decoder (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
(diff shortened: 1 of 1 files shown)


mingw-w64-svt-av1/PKGBUILD
```diff
@@ -0,0 +1,53 @@
...
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    ${MINGW_PREFIX}/bin/cmake \
+      -GNinja \
+      -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+      "${extra_config[@]}" \
+      -DBUILD_SHARED_LIBS=ON \
```
1480c1, line 40: might I suggest `-DREPRODUCIBLE_BUILDS=ON`? This will disable the usage of the `__TIME__` and `__DATE__` macros so the resulting hash of the binaries should match

kmilos: Done, thanks.

Biswa96: Why is it enabled by default? Shouldn't it be on iff something like developer/maintainer mode enabled?

kmilos: No idea, [here's the upstream line](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/b0c1e37506f5ebb4346f97dadd8f80533dd59c55/CMakeLists.txt#L127).

PR #9148, opened 2021-07-16
openshadinglanguage: Update to 1.11.14.2

lazka:


lazka:
I don't know what the problem is, but this gets at least further than the old version, so merging.

mati865:
We need to backport yet another patch: https://github.com/llvm/llvm-project/commit/54bff1522fc863329894d875d54c2fe4cd1b4f3f

lazka:
I can confirm that this fixes the build.

PR #9151, opened 2021-07-16
initial packaging for Aravis library (industrial cameras) and related utils

eudoxos:


mingw-w64-aravis/PKGBUILD
```diff
@@ -0,0 +1,55 @@
+# Maintainer: Václav Šmilauer <eu@doxos.eu>
+
+_realname=aravis
+_REALNAME=ARAVIS
```
lazka, line 4: remove

lazka:
lgtm

mingw-w64-aravis/PKGBUILD
eudoxos: Being new to msys2, I need perhaps more explanation. I used another package (can't remember which one) and most of them use `_realname` and have the maintainer comment. When I remove `_realname`, I will just write `aravis` verbatim everywhere where it appears, that's the way it whould be? Thanks.

lazka: I meant the uppercase `_REALNAME`. Why do you think it is needed?

eudoxos: okay, a leftover from the package I started from. deleted.

Issue #9126, opened 2021-07-10
poppler-qt6


sunderme:
Are there any plans to enable poppler-qt6 interface ?

Biswa96:
It is possible to do that but there are some hurdles.
1. In same pkgbuild: poppler has qt5 as make depends. If we add qt6-base as make depends then qt6-base will conflict with qt5. 
2. In different pkgbuild no issue. But one has to change two pkgbuild files if there is an update or fix.

CC. @MehdiChinoune (qt6 package maintainer) Any thoughts?

ZachBacon:
> It is possible to do that but there are some hurdles.
> 1. In same pkgbuild: poppler has qt5 as make depends. If we add qt6-base as make depends then qt6-base will conflict with qt5. 
> 2. In different pkgbuild no issue. But one has to change two pkgbuild files if there is an update or fix.
> 
> CC. @MehdiChinoune (qt6 package maintainer) Any thoughts?

Based on Arch's pkgbuild for qt6 you can change the bin directory for qt6 so that it doesn't conflict with qt5

Biswa96:
> you can change the bin directory for qt6 so that it doesn't conflict with qt5

It's possible if qt6-base does not conflict with qt5. That trick was also done py2+py3 days. Those makedepends need to be installed before build. See the conflict here https://packages.msys2.org/package/mingw-w64-x86_64-qt6-base?repo=mingw64

sunderme:
if qt6 can't be installed along with qt5, it may be simpler to have a poppler-qt6 package ?

sunderme:
so, I have got homebrew to have poppler and poppler-qt5. (https://github.com/Homebrew/homebrew-core/pull/81248#event-50287520179)
Now, how do we proceed here ?

MehdiChinoune:
I removed the conflcit between qt6-base and qt5.
#9268

sunderme:
Thanks.

PR #9157, opened 2021-07-17
poppler-qt6: Add package

Biswa96:
Just an experiment for this issue https://github.com/msys2/MINGW-packages/issues/9126

### DIFF
--- a/mingw-w64-poppler-qt6/010-clang.patch
+++ b/mingw-w64-poppler-qt6/010-clang.patch
@@ -1,0 +1,11 @@
+--- a/glib/poppler-private.h
++++ b/glib/poppler-private.h
+@@ -150,7 +150,7 @@ GooString *_poppler_convert_date_time_to_pdf_date(GDateTime *datetime);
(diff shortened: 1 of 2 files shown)


Biswa96:
@sunderme Would you like to try the packages from the GitHub Actions artifacts?

sunderme:
I can try. Is there a comfortable way, to get them on a CD machine ?

sunderme:
I think it is much easier to have this PR merged.
As a new package, it does not break anything else and I can set-up the txs CD build against this easily then.

Biswa96:
I'd like to wait if someone comments or approves this change. Meanwhile you could test the package in your macine.

Biswa96:
OK, no comment so far, merging...

sunderme:
thanks

sunderme:
Now that the package is build, the pkgconfig file is missing.
in poppler it is (among others): e.g.  /mingw64/lib/pkgconfig/poppler-cpp.pc

Biswa96:
I was expecting this reply. The pkgconfig file is added in new version of poppler. For that, we'll need to update poppler. But it also needs a rebuild of plethora of packages.

sunderme:
okay, thanks for the quick reply

PR #9159, opened 2021-07-17
Build windows-default-manifest on CLANG64/32

MehdiChinoune:


jeremyd2019:
I'm curious, is there some particular reason you need this?

MehdiChinoune:
> I'm curious, is there some particular reason you need this?

I want to build gfortran on clang*

mati865:
Is there Thread Local Storage in Fortran? GCC doesn't support native TLS for Windows so libraries/executables produced by GCC wouldn't even link with ones made by Clang if TLS is used.

MehdiChinoune:
> Is there Thread Local Storage in Fortran? GCC doesn't support native TLS for Windows so libraries/executables produced by GCC wouldn't even link with ones made by Clang if TLS is used.

I don't know!

PR #9161, opened 2021-07-17
[NFC] Remove explicit make-dependency on gcc.

MehdiChinoune:


jeremyd2019:
Why?!?

https://discord.com/channels/792780131906617355/794220101741838401/865621103539912774

MehdiChinoune:
> Why?!?

To make "gcc" an independent package on clang-* (like clang on mingw64/32 and ucrt). so if a package explicitly requires gcc, clang won't provide it.

> 
> https://discord.com/channels/792780131906617355/794220101741838401/865621103539912774

I don't use/have discord/discord account.

jeremyd2019:
> To make "gcc" an independent package on clang-* (like clang on mingw64/32 and ucrt). so if a package explicitly requires gcc, clang won't provide it.

This is likely doomed to fail.  Clang knows about gcc's libgcc and libstdc++, and can generate code that uses them, which is how clang on mingw64/32 and ucrt works.  Gcc does not know about clang's compiler-rt and libc++, so to get gcc to propery interact with clang, you're probably going to wind up with mingw64/32 by the time you're done (unless you are planning to do some major changes to gcc to support clang's runtime)

jeremyd2019:
> I don't use/have discord/discord account.

> _lazka — 07/15/2021_
> any thoughts on having a ${MINGW_PACKAGE_PREFIX}-c-compiler meta package?
> which would depend on the default compiler/linker/mingw etc
> _Biswa96 — 07/15/2021_
> Reason?
> Does not toolchain package group solve that purpose?
> _lazka — 07/15/2021_
> so tools for automation know the real dependencies
> and you don't need fortran/gdb/make in all cases
> it would be an alternative to the toolchain group
> _Elieux — Yesterday at 4:53 AM_
> How likely is having the other compiler in a subsystem?
> Like mingw64/clang or clang64/gcc?
> 2) Does clang provide gcc? 3) Does gcc provide clang?
> _mati865 — Yesterday at 7:01 AM_
> Clang is popular within mingw64 beause it's special version that creates binaries compatible with GCC.
> GCC in clang64 is impossible right now, possibly ever.
> 2) in clang* subsystems
> 3) never
> _Elieux — Yesterday at 8:21 AM_
> Maybe the compiler package then can be called gcc.
> _lazka — Yesterday at 8:49 AM_
> we kinda do that already, clang64/clang provides gcc
> (and even "gcc" binaries)
> so, maybe to generalize my question: what about making all mingw dependencies explicit one way or another?

MehdiChinoune:
> > I don't use/have discord/discord account.
> 
> > _lazka — 07/15/2021_
> > any thoughts on having a ${MINGW_PACKAGE_PREFIX}-c-compiler meta package?
> > which would depend on the default compiler/linker/mingw etc
> > _Biswa96 — 07/15/2021_
> > Reason?
> > Does not toolchain package group solve that purpose?
> > _lazka — 07/15/2021_
> > so tools for automation know the real dependencies
> > and you don't need fortran/gdb/make in all cases
> > it would be an alternative to the toolchain group
> > _Elieux — Yesterday at 4:53 AM_
> > How likely is having the other compiler in a subsystem?
> > Like mingw64/clang or clang64/gcc?
> > 2) Does clang provide gcc? 3) Does gcc provide clang?
> > _mati865 — Yesterday at 7:01 AM_
> > Clang is popular within mingw64 beause it's special version that creates binaries compatible with GCC.
> > GCC in clang64 is impossible right now, possibly ever.
> > 2) in clang* subsystems
> > 3) never
> > _Elieux — Yesterday at 8:21 AM_
> > Maybe the compiler package then can be called gcc.
> > _lazka — Yesterday at 8:49 AM_
> > we kinda do that already, clang64/clang provides gcc
> > (and even "gcc" binaries)
> > so, maybe to generalize my question: what about making all mingw dependencies explicit one way or another?

I think they should move that discussion to "Discussions"

MehdiChinoune:
> > To make "gcc" an independent package on clang-* (like clang on mingw64/32 and ucrt). so if a package explicitly requires gcc, clang won't provide it.
> 
> This is likely doomed to fail. Clang knows about gcc's libgcc and libstdc++, and can generate code that uses them, which is how clang on mingw64/32 and ucrt works. Gcc does not know about clang's compiler-rt and libc++, so to get gcc to propery interact with clang, you're probably going to wind up with mingw64/32 by the time you're done (unless you are planning to do some major changes to gcc to support clang's runtime)

It's all about gfortran. Imagine I successfully build gfortran (which is a part of GCC). when the user try to install gfortran on clang-, pacman will not install the original gcc since it is provided by clang.

jeremyd2019:
> It's all about gfortran. Imagine I successfully build gfortran (which is a part of GCC). when the user try to install gfortran on clang-, pacman will not install the original gcc since it is provided by clang.

Even if you successfully *build* gfortran, you would essentially have a clang64 to ucrt64 cross-compiler.  GCC *does not* support the clang runtime libs, or native TLS, or native threading (just winpthreads).  These are all differences between clang64 and ucrt64.

MehdiChinoune:
> > It's all about gfortran. Imagine I successfully build gfortran (which is a part of GCC). when the user try to install gfortran on clang-, pacman will not install the original gcc since it is provided by clang.
> 
> Even if you successfully _build_ gfortran, you would essentially have a clang64 to ucrt64 cross-compiler. GCC _does not_ support the clang runtime libs, or native TLS, or native threading (just winpthreads). These are all differences between clang64 and ucrt64.

I didn't understand what are you trying to say, could you elaborate more.

jeremyd2019:
It's basically what @mati865 said in the discussion above:

> Clang is popular within mingw64 beause it's special version that creates binaries compatible with GCC.
> GCC in clang64 is impossible right now, possibly ever.

The difference between the "clang" subsystem/prefix and the mingw/ucrt is that clang is not patched/configured to create binaries compatible with GCC. I rattled off a list the major things that make them incompatible.  While clang can be made to make binaries compatible with GCC's practices, GCC does not have support to make binaries compatible with what the clang in a "clang" subsystem does.  This is the major reason that the new subsystems were created

MehdiChinoune:
> It's basically what @mati865 said in the discussion above:
> 
> > Clang is popular within mingw64 beause it's special version that creates binaries compatible with GCC.
> > GCC in clang64 is impossible right now, possibly ever.
> 
> The difference between the "clang" subsystem/prefix and the mingw/ucrt is that clang is not patched/configured to create binaries compatible with GCC. I rattled off a list the major things that make them incompatible. While clang can be made to make binaries compatible with GCC's practices, GCC does not have support to make binaries compatible with what the clang in a "clang" subsystem does. This is the major reason that the new subsystems were created

Anyway, I don't think this pr will affect something. It removes the explicit make-dependency on gcc which is part of the build-toolchain on mingw64/32 and ucrt64. while it's not needed on clang subsystems.

jeremyd2019:
Except there was just a discussion that it would be helpful to autobuild to have an explicit makedepends on the compiler, and that gcc already kind of serves as that because clang provides gcc on clang prefixes.

If there's going to be a wholesale change of PKGBUILDs, it would probably be to switch to a new meta-package, rather than just removing gcc.

mati865:
Isn't `mingw-w64-*-toolchain` expected to be installed by the default? `makepkg-mingw` even warns when it's not available:
```
  if [ ! -f "${_compiler}" ]; then
    print_warning "You don't have the required toolchain installed for ${_mingw}."
    print_warning "To install it run: 'pacman -S ${_toolchain}'"
  fi

```

Issue #9049, opened 2021-07-03
LLD fails on `DESCRIPTION` in `.def` files


mati865:
https://github.com/msys2/MINGW-packages/pull/8763

jeremyd2019:
Found this in perl's code:
```perl
    print "LIBRARY $dll\n";
    # The DESCRIPTION module definition file statement is not supported
    # by VC7 onwards.
    if ($ARGS{CCTYPE} eq 'GCC') {
        print "DESCRIPTION 'Perl interpreter'\n";
    }
    print "EXPORTS\n";
```

PR #9164, opened 2021-07-18
perl: fix build on clang

### LINKED ISSUE #9175: perl: mingw64 vs ucrt64 io test failures
@Biswa96 asked about IO with his ucrt io patch, so here's the results of the `check` phase of the PKGBUILD for mingw64 diffed with ucrt64, with irrelevant differences edited out for brevity.  It looks like they're mostly socket related.

<details>
<summary>Diff of mingw64 vs ucrt64 perl tests</summary>

```diff
364d363
< ..\miniperl.exe -I..\lib ..\pod\perlmodlib.PL -q ..
365a365
> ..\miniperl.exe -I..\lib ..\pod\perlmodlib.PL -q ..
565,566c565,570
< io/socket.t ........................................................ ok
< io/socketpair.t .................................................... ok
---
> io/socket.t ........................................................ 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
> io/socketpair.t .................................................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1242c1246,1249
< ../cpan/File-Fetch/t/01_File-Fetch.t ............................... ok
---
> ../cpan/File-Fetch/t/01_File-Fetch.t ............................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 119 subtests passed 
> 	(less 12 skipped subtests: 107 okay)
1281c1288,1290
< ../cpan/IPC-Cmd/t/01_IPC-Cmd.t ..................................... ok
---
> ../cpan/IPC-Cmd/t/01_IPC-Cmd.t ..................................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 18 subtests passed 
1286c1295,1297
< ../cpan/IO-Socket-IP/t/01local-client-v4.t ......................... ok
---
> ../cpan/IO-Socket-IP/t/01local-client-v4.t ......................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1294c1305,1307
< ../cpan/IO-Socket-IP/t/02local-server-v4.t ......................... ok
---
> ../cpan/IO-Socket-IP/t/02local-server-v4.t ......................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1302c1315,1317
< ../cpan/IO-Socket-IP/t/03local-cross-v4.t .......................... ok
---
> ../cpan/IO-Socket-IP/t/03local-cross-v4.t .......................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 6 subtests passed 
1306d1320
< getaddrinfo() - ai_family not supported at t/04local-client-v6.t line 38.
1312c1326
< Dubious, test returned 25 (wstat 6400, 0x1900)
---
> Dubious, test returned 5 (wstat 1280, 0x500)
1320,1326d1333
< 
< #   Failed test 'IO::Socket::IP->new constructs a SOCK_STREAM socket'
< #   at t/05local-server-v6.t line 41.
< #   error was nodename nor servname provided, or not known
< Can't call method "sockdomain" on an undefined value at t/05local-server-v6.t line 44.
< # Tests were run but no plan was declared and done_testing() was not seen.
< # Looks like your test exited with 22 just after 1.
1329,1330c1336,1337
< Dubious, test returned 22 (wstat 5632, 0x1600)
< Failed 1/1 subtests 
---
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1356,1357c1363,1368
< ../cpan/IO-Socket-IP/t/11sockopts.t ................................ ok
< ../cpan/autodie/t/recv.t ........................................... ok
---
> ../cpan/IO-Socket-IP/t/11sockopts.t ................................ 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 1 subtests passed 
> ../cpan/autodie/t/recv.t ........................................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 8/8 subtests 
1364c1375,1377
< ../cpan/IO-Socket-IP/t/12port-fallback.t ........................... ok
---
> ../cpan/IO-Socket-IP/t/12port-fallback.t ........................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1372c1385,1387
< ../cpan/IO-Socket-IP/t/13addrinfo.t ................................ ok
---
> ../cpan/IO-Socket-IP/t/13addrinfo.t ................................ 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 2 subtests passed 
1380c1395,1397
< ../cpan/IO-Socket-IP/t/14fileno.t .................................. ok
---
> ../cpan/IO-Socket-IP/t/14fileno.t .................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1388c1405,1407
< ../cpan/IO-Socket-IP/t/15io-socket.t ............................... ok
---
> ../cpan/IO-Socket-IP/t/15io-socket.t ............................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 1 subtests passed 
1404c1423,1425
< ../cpan/IO-Socket-IP/t/17gai-flags.t ............................... ok
---
> ../cpan/IO-Socket-IP/t/17gai-flags.t ............................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1412c1433,1435
< ../cpan/IO-Socket-IP/t/18fdopen.t .................................. ok
---
> ../cpan/IO-Socket-IP/t/18fdopen.t .................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 2 subtests passed 
1420c1443,1445
< ../cpan/IO-Socket-IP/t/19no-addrs.t ................................ ok
---
> ../cpan/IO-Socket-IP/t/19no-addrs.t ................................ 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 4 subtests passed 
1428c1453,1455
< ../cpan/IO-Socket-IP/t/20subclass.t ................................ ok
---
> ../cpan/IO-Socket-IP/t/20subclass.t ................................ 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 1 subtests passed 
1436c1463,1465
< ../cpan/IO-Socket-IP/t/21as-inet.t ................................. ok
---
> ../cpan/IO-Socket-IP/t/21as-inet.t ................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 3 subtests passed 
1444c1473,1475
< ../cpan/IO-Socket-IP/t/22timeout.t ................................. ok
---
> ../cpan/IO-Socket-IP/t/22timeout.t ................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 3 subtests passed 
1452c1483,1485
< ../cpan/IO-Socket-IP/t/30nonblocking-connect.t ..................... ok
---
> ../cpan/IO-Socket-IP/t/30nonblocking-connect.t ..................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1460c1493,1495
< ../cpan/IO-Socket-IP/t/31nonblocking-connect-internet.t ............ ok
---
> ../cpan/IO-Socket-IP/t/31nonblocking-connect-internet.t ............ 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
1942,1943c1977,1983
< ../cpan/Socket/t/Socket.t .......................................... ok
< ../cpan/Socket/t/socketpair.t ...................................... ok
---
> ../cpan/Socket/t/Socket.t .......................................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 6 subtests passed 
> 	(less 4 skipped subtests: 2 okay)
> ../cpan/Socket/t/socketpair.t ...................................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 45/45 subtests 
2677,2678c2717,2724
< ../dist/IO/t/cachepropagate-tcp.t .................................. ok
< ../dist/IO/t/cachepropagate-udp.t .................................. ok
---
> ../dist/IO/t/cachepropagate-tcp.t .................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 9 subtests passed 
> 	(less 1 skipped subtest: 8 okay)
> ../dist/IO/t/cachepropagate-udp.t .................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 7 subtests passed 
> 	(less 1 skipped subtest: 6 okay)
2706c2752,2755
< ../dist/IO/t/io_multihomed.t ....................................... ok
---
> # Test process timed out - terminating
> ../dist/IO/t/io_multihomed.t ....................................... 
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 5/8 subtests 
2712c2761,2764
< ../dist/IO/t/io_sel.t .............................................. ok
---
> ../dist/IO/t/io_sel.t .............................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 3/27 subtests 
> 	(less 6 skipped subtests: 18 okay)
2715c2767,2769
< ../dist/IO/t/io_sock.t ............................................. ok
---
> ../dist/IO/t/io_sock.t ............................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 22/26 subtests 
2718c2772,2776
< ../dist/IO/t/io_udp.t .............................................. ok
---
> # Test process timed out - terminating
> ../dist/IO/t/io_udp.t .............................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 15 subtests passed 
> 	(less 3 skipped subtests: 12 okay)
2720c2778,2780
< ../dist/IO/t/io_unix.t ............................................. ok
---
> ../dist/IO/t/io_unix.t ............................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> No subtests run 
2758c2818,2820
< ../dist/Net-Ping/t/010_pingecho.t .................................. ok
---
> ../dist/Net-Ping/t/010_pingecho.t .................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 1/2 subtests 
2774c2836,2838
< ../dist/Net-Ping/t/120_udp_inst.t .................................. ok
---
> ../dist/Net-Ping/t/120_udp_inst.t .................................. 
> Dubious, test returned 5 (wstat 1280, 0x500)
> All 2 subtests passed 
2863,2868d2926
< 
< #   Failed test 'second port is reachable'
< #   at t/450_service.t line 82.
< #          got: '0'
< #     expected: '1'
< # Looks like you failed 1 test of 26.
2870,2871c2928,2929
< Dubious, test returned 1 (wstat 256, 0x100)
< Failed 1/26 subtests 
---
> Dubious, test returned 5 (wstat 1280, 0x500)
> Failed 20/26 subtests 
3361,3362c3419,3441
< ../cpan/IO-Socket-IP/t/04local-client-v6.t                       (Wstat: 6400 Tests: 0 Failed: 0)
<   Non-zero exit status: 25
---
> io/socket.t                                                      (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> io/socketpair.t                                                  (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/File-Fetch/t/01_File-Fetch.t                             (Wstat: 1280 Tests: 119 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IPC-Cmd/t/01_IPC-Cmd.t                                   (Wstat: 1280 Tests: 18 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/01local-client-v4.t                       (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/02local-server-v4.t                       (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/03local-cross-v4.t                        (Wstat: 1280 Tests: 6 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/04local-client-v6.t                       (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
3364,3366c3443,3444
< ../cpan/IO-Socket-IP/t/05local-server-v6.t                       (Wstat: 5632 Tests: 1 Failed: 1)
<   Failed test:  1
<   Non-zero exit status: 22
---
> ../cpan/IO-Socket-IP/t/05local-server-v6.t                       (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
3367a3446,3495
> ../cpan/IO-Socket-IP/t/11sockopts.t                              (Wstat: 1280 Tests: 1 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/autodie/t/recv.t                                         (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 8 tests but ran 0.
> ../cpan/IO-Socket-IP/t/12port-fallback.t                         (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/13addrinfo.t                              (Wstat: 1280 Tests: 2 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/14fileno.t                                (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/15io-socket.t                             (Wstat: 1280 Tests: 1 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/17gai-flags.t                             (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/18fdopen.t                                (Wstat: 1280 Tests: 2 Failed: 0)
>   Non-zero exit status: 5
> ../cpan/IO-Socket-IP/t/19no-addrs.t                              (Wstat: 1280 Tests: 4 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/20subclass.t                              (Wstat: 1280 Tests: 1 Failed: 0)
>   Non-zero exit status: 5
> ../cpan/IO-Socket-IP/t/21as-inet.t                               (Wstat: 1280 Tests: 3 Failed: 0)
>   Non-zero exit status: 5
> ../cpan/IO-Socket-IP/t/22timeout.t                               (Wstat: 1280 Tests: 3 Failed: 0)
>   Non-zero exit status: 5
> ../cpan/IO-Socket-IP/t/30nonblocking-connect.t                   (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/IO-Socket-IP/t/31nonblocking-connect-internet.t          (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../cpan/Socket/t/Socket.t                                        (Wstat: 1280 Tests: 6 Failed: 0)
>   Non-zero exit status: 5
> ../cpan/Socket/t/socketpair.t                                    (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 45 tests but ran 0.
> ../dist/IO/t/cachepropagate-tcp.t                                (Wstat: 1280 Tests: 9 Failed: 0)
>   Non-zero exit status: 5
> ../dist/IO/t/cachepropagate-udp.t                                (Wstat: 1280 Tests: 7 Failed: 0)
>   Non-zero exit status: 5
> ../dist/IO/t/io_multihomed.t                                     (Wstat: 1280 Tests: 3 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 8 tests but ran 3.
3371,3373c3499,3517
< ../dist/Net-Ping/t/450_service.t                                 (Wstat: 256 Tests: 26 Failed: 1)
<   Failed test:  9
<   Non-zero exit status: 1
---
> ../dist/IO/t/io_sel.t                                            (Wstat: 1280 Tests: 24 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 27 tests but ran 24.
> ../dist/IO/t/io_sock.t                                           (Wstat: 1280 Tests: 4 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 26 tests but ran 4.
> ../dist/IO/t/io_udp.t                                            (Wstat: 1280 Tests: 15 Failed: 0)
>   Non-zero exit status: 5
> ../dist/IO/t/io_unix.t                                           (Wstat: 1280 Tests: 0 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: No plan found in TAP output
> ../dist/Net-Ping/t/010_pingecho.t                                (Wstat: 1280 Tests: 1 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 2 tests but ran 1.
> ../dist/Net-Ping/t/120_udp_inst.t                                (Wstat: 1280 Tests: 2 Failed: 0)
>   Non-zero exit status: 5
> ../dist/Net-Ping/t/450_service.t                                 (Wstat: 1280 Tests: 6 Failed: 0)
>   Non-zero exit status: 5
>   Parse errors: Bad plan.  You planned 26 tests but ran 6.
```
</details>

jeremyd2019:
Fixes build on clang prefixes.

Reasons for the patch files
* 008 - was causing issues with sockets.  Adjusted to more closely match what perl was already doing for MSVC (see ex. https://github.com/Perl/perl5/blob/blead/win32/GNUmakefile#L1389-L1396).  Fixes #9175.
* 009 - imported from @Biswa96
* 010 - imported from @Biswa96 
* 011 - imported from @Biswa96 
* 012 - was triggering `#error`s from headers that said not to include them directly, so filtered out only the errno ones and rely that they then include whatever is needed for them.
* 013 - llvm-dlltool doesn't support .exp files, and they are unneeded in modern toolchains anyway (there was a comment elsewhere in perl where they removed the .exp file dlltool dance that said they can assume a toolchain newer than 2002)
* 014 - llvm doesn't like the `DESCRIPTION` in .def file (#9049)
* 015 - apparently clang thinks the `__attribute__noreturn__` is part of the function pointer type, and complained about a type mismatch.  Gcc for its part gives an error on `__attribute__noreturn__` on function definition
* 016 - lovely crash due to clang on x86_64 generating an instruction that requires 16-byte alignment, and this 'header' tacked onto the front of memory allocations was only 24 bytes long on x86_64.  Now it's 32 bytes on 64-bit and 16 bytes on 32-bit, maintaining 16-byte alignment for both.
* 017 - build on Windows ARM64
* 018 - copy clang runtime dlls in addition to gcc runtime dlls, to allow taint test to work.

Also, hooked up 'debug' option, and disabled a *very noisy* clang warning about perl's macro tricks that was making it hard to find errors in the build output. (this warning was the subject of an upstream issue and pull request, but seems to have caused some other issues and seems best to just shut it up for now rather than trying to backport)


jeremyd2019:
Reasons for the patch files
* 009 - was triggering `#error`s from headers that said not to include them directly, so filtered out only the errno ones and rely that they then include whatever is needed for them.
* 010 - llvm-dlltool doesn't support .exp files, and they are unneeded in modern toolchains anyway (there was a comment elsewhere in perl where they removed the .exp file dlltool dance that said they can assume a toolchain newer than 2002)
* 011 - llvm doesn't like the `DESCRIPTION` in .def file (#9049)
* 012 - `-mdll` seems to turn into `--dll` which LLD doesn't know what to do with
* 013 - apparently clang thinks the `__attribute__noreturn__` is part of the function pointer type, and complained about a type mismatch.  Gcc for its part gives an error on `__attribute__noreturn__` on function definition
* 014 - lovely crash due to clang on x86_64 generating an instruction that requires 16-byte alignment, and this 'header' tacked onto the front of memory allocations was only 24 bytes long on x86_64.  Now it's 32 bytes on 64-bit and 16 bytes on 32-bit, 16-byte aligned for both.

Biswa96:
How did you figure out the 014 patch 🤩

jeremyd2019:
stared at the instruction in the disassembler for a while:
```
perl532!perl_construct+0x3c:
00007ffe`1fa43b6c 66410f7f7550    movdqa  xmmword ptr [r13+50h],xmm6 ds:000001eb`f61b9fe8=00000000000000000000000000000000
```
poked the memory it was referencing and went "hey, that's a valid pointer, what gives?", and finally googled the instruction (`movdqa`) which says:
> When the source or destination operand is a memory operand, the operand must be aligned on a 16-byte boundary or a general-protection exception (#GP) will be generated.

saw that the pointer ended in 8 (not 16-byte aligned), so set a breakpoint on the allocation function (PerlMemMalloc), and found that vmem.h linked list implementation messing up the alignment.

jeremyd2019:
On second thought, it might be better to keep the patches separate.

jeremyd2019:
<details>
<summary>Diff of Errno.pm between 'live' mingw64 and built clang64</summary>

```diff
$ diff -u /mingw64/lib/perl5/core_perl/Errno.pm pkg/mingw-w64-clang-x86_64-perl/
clang64/lib/perl5/core_perl/Errno.pm
--- /mingw64/lib/perl5/core_perl/Errno.pm       2021-07-01 13:04:52.000000000 -0700
+++ pkg/mingw-w64-clang-x86_64-perl/clang64/lib/perl5/core_perl/Errno.pm       2021-07-17 22:36:09.000000000 -0700
@@ -10,8 +10,8 @@

 use Config;
 "$Config{'archname'}-$Config{'osvers'}" eq
-"MSWin32-x64-multi-thread-10.0.17763.1999" or
-       die "Errno architecture (MSWin32-x64-multi-thread-10.0.17763.1999) does not match executable architecture ($Config{'archname'}-$Config{'osvers'})";
+"MSWin32-x64-multi-thread-10.0.19043.1110" or
+       die "Errno architecture (MSWin32-x64-multi-thread-10.0.19043.1110) does not match executable architecture ($Config{'archname'}-$Config{'osvers'})";

 our $VERSION = "1.30";
 $VERSION = eval $VERSION;
@@ -2554,17 +2554,17 @@
        ETXTBSY EUSERS EWOULDBLOCK EXDEV
     )],
     WINSOCK => [qw(
-       WSAEMSGSIZE WSAEADDRNOTAVAIL WSAENOMORE WSAESHUTDOWN WSAEREMOTE
-       WSAEPROVIDERFAILEDINIT WSAELOOP WSAEPROTONOSUPPORT WSAEAFNOSUPPORT
-       WSAENAMETOOLONG WSAEACCES WSAEINTR WSAENETRESET WSAEDISCON
-       WSAETIMEDOUT WSAEPFNOSUPPORT WSAEHOSTDOWN WSAETOOMANYREFS WSAENOTSOCK
-       WSAEOPNOTSUPP WSAEPROCLIM WSAENOBUFS WSAEINVALIDPROCTABLE WSAEFAULT
-       WSAECANCELLED WSAEALREADY WSAEWOULDBLOCK WSAEADDRINUSE WSAEREFUSED
-       WSAEDESTADDRREQ WSAENOPROTOOPT WSAECONNREFUSED WSAEINVAL WSAEBADF
-       WSAENETDOWN WSAEPROTOTYPE WSAECONNABORTED WSAESOCKTNOSUPPORT WSAEMFILE
-       WSAESTALE WSAENOTCONN WSAENETUNREACH WSAEUSERS WSAEISCONN WSAENOTEMPTY
-       WSAEDQUOT WSAEHOSTUNREACH WSAEINVALIDPROVIDER WSAEINPROGRESS
-       WSAECONNRESET
+       WSAEBADF WSAECONNREFUSED WSAECONNABORTED WSAEOPNOTSUPP WSAENOTSOCK
+       WSAETIMEDOUT WSAENOMORE WSAEINVALIDPROCTABLE WSAEFAULT WSAEHOSTDOWN
+       WSAENAMETOOLONG WSAEAFNOSUPPORT WSAEHOSTUNREACH WSAEDESTADDRREQ
+       WSAEINTR WSAEREFUSED WSAENETUNREACH WSAENETRESET WSAEINVAL WSAENETDOWN
+       WSAEREMOTE WSAESOCKTNOSUPPORT WSAEPROTONOSUPPORT WSAELOOP WSAESHUTDOWN
+       WSAEALREADY WSAENOBUFS WSAEPROVIDERFAILEDINIT WSAEWOULDBLOCK
+       WSAENOTCONN WSAEISCONN WSAEPROTOTYPE WSAESTALE WSAEDQUOT
+       WSAEPFNOSUPPORT WSAEMFILE WSAECANCELLED WSAENOPROTOOPT WSAETOOMANYREFS
+       WSAEUSERS WSAEPROCLIM WSAEDISCON WSAEADDRINUSE WSAEINVALIDPROVIDER
+       WSAEACCES WSAENOTEMPTY WSAEADDRNOTAVAIL WSAEMSGSIZE WSAECONNRESET
+       WSAEINPROGRESS
     )],
 );

```
</details>

jeremyd2019:
Let me know if it's preferred to change anything here.

Biswa96:
1. Would you like to upstream these these changes? I made a pull request here https://github.com/Perl/perl5/pull/18879 but maintainers asked me to redo it.
2. Have you checked if basic file r/w works? I'm afraid of the usage of the ucrt hack.
3. If you follow the link in my PR link there is an 'actual' fix fo that clang compound token warnings.

jeremyd2019:
1. I could, I guess...
2. Well, it works well enough to build itself.  I have not given as much testing as I would have liked.  I should probably at least try the `check()` between mingw64 and clang64 and see if there's a difference.
3. Looks like there are still [issues with that](https://github.com/Perl/perl5/pull/18984#issuecomment-881922991).  Would be nice for them to be fixed properly, but suppressing them works for now.  This seems like a pretty core thing to try to patch.

jeremyd2019:
I like your approach better on GNUmakefile than what I did in the arm patch, might have to revisit that.

jeremyd2019:
Speaking of the half-assed fix for .exp files, it seems from check output I missed something that is necessary for building extensions.

jeremyd2019:
Diff between check output for mingw64 from master and my branch on clang64.

[mingw64-clang64-check-diff.txt](https://github.com/msys2/MINGW-packages/files/6838154/mingw64-clang64-check-diff.txt)

jeremyd2019:
Oh I remember now why I ran screaming from this file:
https://github.com/Perl/perl5/blob/blead/dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/Windows/GCC.pm

also
https://github.com/Perl/perl5/blob/blead/dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/Windows.pm#L116

jeremyd2019:
> 2. Have you checked if basic file r/w works? I'm afraid of the usage of the ucrt hack.

Opened #9175 for that specifically since that's in master rather than these changes.

jeremyd2019:
Now the only test that seems to fail in clang64 but succeed in ucrt64 is taint.  Need to look into why that is.

jeremyd2019:
That fails because when prepping for running tests, perl's build system "knows" to copy over gcc-libs dlls (libgcc, libstdc++), but doesn't know about libunwind or libc++, so when it tries to run the perl under test again within the test, after it's screwed around with PATH, it doesn't work.

jeremyd2019:
With e1a7b0d, ucrt64 tests match mingw64 tests before #9164.  I still need to run tests on clang64 and probably build/test on clangarm64 too after these updates, but if that all checks out I'll rebase/squash some of this and it should be ready

jeremyd2019:
clang64 tests match too.

jeremyd2019:
clang32 has a number of time-related errors.  I wonder if there's a 64-bit vs 32-bit `time_t` thing going on.

jeremyd2019:
The tests worked on mingw32, so I'm very suspicious of `_USE_32BIT_TIME_T` going to try defining that

jeremyd2019:
`_USE_32BIT_TIME_T` doesn't seem to have helped.  Which is good, because 64-bit time_t is supposed to work in msvc, but bad because I have no idea what the issue might be.

jeremyd2019:
clangarm64 tests look pretty ok, except one error about getting chip name from `Win32::GetChipName`...  Probably never heard of ARM64.

jeremyd2019:
Unless somebody wants to investigate why clang32 (probably ucrt32 but there is no such thing currently for gcc) has time-related issues, I'll plan to squash some commits and then mark this ready for review again.

jeremyd2019:
Created #9178 for the clang32 time issues.

jeremyd2019:
I've removed clang32 from mingw_arch for perl for now, until #9178 is fixed.

jeremyd2019:
My debugging on #9178 looks like a mingw-w64 issue, not a perl issue, so this should be ready to go.

Biswa96:
> saw that the pointer ended in 8 (not 16-byte aligned)

About the memory alignment in previous comment. Shouldn't compile take care of that padding in structure?

jeremyd2019:
If the compiler knew what the heck was going on, it should either add padding or refrain from using the 16-byte-alignment-requiring instruction.  But it doesn't.  This code is a malloc wrapper, that mallocs the size requested plus the size of this header struct, does stuff with the header, and then returns a void pointer past the header.  The compiler does not know the alignment requirement of anything whose pointer type that `void*` might be cast to so cannot know the proper amount of padding to add.  More concerning to me is that the compiler seems to assume the resulting void* is 16-byte-aligned.

PR #9169, opened 2021-07-19
use consistent pathtools

dscho:
This is my counterproposal to the abandoned https://github.com/msys2/MINGW-packages/pull/9036

It essentially consists of four parts:

- dropping the unused patch from the `gnupg` package that would mislead the occasional reader that this package uses the `pathtools` at all
- updating the existing `pathtools` users (including, but not limited to, cURL) to use a copy of `mingw-w64-pathtools`
- adding a script to help maintaining the current setup and to avoid future divergences
- [BONUS TRACK] fixing the warnings about missing `(mingw-w64)` suffixes in the package description of two of the touched packages

The way this is set up not only makes it easy to update the `pathtools` users (by running `mingw-w64-pathtools/update-dependencees.sh`), but also ensures that stale/divergent copies are detected when building packages.

It also does away with the awfully error-prone method to use `mingw-w64-pathtools` via a hardcoded commit and downloading via the GitHub "raw file" URLs that was used in https://github.com/msys2/MINGW-packages/pull/9036.

Cc: @jeremyd2019 @Biswa96


Biswa96:
Thank you for this awesome work 😍 BTW, didn't you proposed to use relative path for those two pathtools files? Here we've multiple copies of same files.

dscho:
> 
> 
> Thank you for this awesome work 😍

You're welcome.

> BTW, didn't you proposed to use relative path for those two pathtools files? Here we've multiple copies of same files.

Oh, I thought you had tried this yourself before giving up... But I did document the reason in [the commit message](https://github.com/msys2/MINGW-packages/pull/9169/commits/31374aacd434170b03a41bf4bb45fa9d88d2d348):

```
 Sadly, we cannot use relative paths (which is forbidden by `PKGBUILD`)
 nor symbolic links (because of MSYS2's inadequate support for symbolic
 links). Therefore we do the next best thing: we copy the code and add a
 check to ensure that the code hasn't gone stale.
```

Biswa96:
I thought of doing something like KDE packages where PKGBUILD `source` a shell script from another folder like this

```sh
source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
```

That seems kind of abnormal way to me. As this current approach helps maintenance, 👌 from me. Let see what maintainers say.

dscho:
@Biswa96 The _shell_ directive `source` has no problems with relative or absolute paths. Pacman's `makepkg` disallows relative paths in the `source` _array_. It only allows files in the same directory as `PKGBUILD` and URLs, AFAICT.

mingw-w64-curl/PKGBUILD
```diff
@@ -46,21 +46,30 @@ if [ -n "${_namesuff}" ]; then
...
 
 prepare() {
+  test ! -d "${startdir}/../mingw-w64-pathtools" || {
+    cmp "${startdir}/../mingw-w64-pathtools/pathtools.c" "${srcdir}/pathtools.c" &&
+    cmp "${startdir}/../mingw-w64-pathtools/pathtools.h" "${srcdir}/pathtools.h"
+  } || exit 1
+
   cd "${srcdir}/${_realname}-${pkgver}"
-  rm -f lib/pathtools.h lib/pathtools.c > /dev/null 2>&1 || true
+  cp -fHv "${srcdir}"/pathtools.[ch] lib/
```
jeremyd2019, line 72: I wonder if it would be worthwhile to use a makepkg-template for few duplicated lines in each PKGBUILD, or if they are few enough to not be worth it...  The destination/extension could be specified by variables set before the template if the `cp` were included.

dscho: With this few lines (and subtle differences e.g. when the destination _requires_ a `.cxx` extension), introducing a template might very well end up costing more lines than are saved. I'm not totally opposed, but mildly in favor of keeping it unrolled.

lazka:
Thanks! Do you still think we should move this to its own package pulled in via makedepends in the long run? (I might have missed some discussion on this elsewhere..)

dscho:
> 
> 
> Thanks! Do you still think we should move this to its own package pulled in via makedepends in the long run? (I might have missed some discussion on this elsewhere..)

It would complicate things a bit because `mingw-w64-pathtools` is designed not as a static library, but as source code that is copied into the respective projects. Changing this paradigm would take more effort than I can afford spending.

lazka:
OK, thanks.

dscho:
And _of course_, after _yet another_ long PR build (3.5h!), it [failed](https://github.com/msys2/MINGW-packages/pull/9169/checks?check_run_id=3107949651#step:9:73605), but this time with a transient network error (and to add insult to injury, it [the same network error caused the UCRT64 job to fail, too](https://github.com/msys2/MINGW-packages/runs/3107949712?check_suite_focus=true#step:9:73003)!

dscho:
Uhm. What's [this](https://github.com/msys2/MINGW-packages/pull/9169/checks?check_run_id=3112941340#step:6:1993) supposed to mean?

```
  File ".ci/ci-check.py", line 248, in <module>
    main()
  File ".ci/ci-check.py", line 196, in main
    rdeps = get_rdeps(pkgname)
  File ".ci/ci-check.py", line 152, in get_rdeps
    raise ValueError(f"Unable to parse Pacman output: {ret}")
ValueError: Unable to parse Pacman output: Repository      : ci
```

(Of _course_ this failure _had_ to happen _after 4 hours_!)

lazka:
Feel free to ignore for now.

dscho:
> This only checks python dependency chains.

So is it known that this check is broken?

lazka:
lgtm

lazka:
> So is it known that this check is broken?

yes, but not recorded somewhere I think. Naveen is maintaining it, but he is away/offline for a few weeks.

dscho:
@lazka can this PR move forward independently, or do I have to wait for that bug to be resolved?

lazka:
it can be ignored. merge?

dscho:
> it can be ignored. merge?

@lazka I'd love to see this merged, yes. (I don't have permission to do that myself.)

lazka:
thanks!

PR #9176, opened 2021-07-19
cmake: default to ninja

lazka:
It defaults to msvc/nmake otherwise which in MSYS2 is never a good default imo.


lazka:
@orgads @MehdiChinoune any thoughts on this?

1480c1:
Would be nice

orgads:
+1

lazka:
ok, let's try this then. I'd still advice everyone to explicitly pass a generator for your automated builds.

1480c1:
Wait, I just realized cmake doesn't have a dependency set for ninja, is that okay?

lazka:
Not sure. You mean for convenience?

1480c1:
Yeah, I mean, I guess it's the same situation without the patch because msvc isn't guaranteed to be present, so the default would fail in most cases

lazka:
yeah, probably a good idea.

PR #9177, opened 2021-07-20
[v8] Update to 9.1.269.39

Kwizatz:
Just a regular stable version update.

### DIFF
--- a/mingw-w64-v8/0000-add-mingw-main-code-changes.patch
+++ b/mingw-w64-v8/0000-add-mingw-main-code-changes.patch
@@ -1,8 +1,8 @@
 diff --git a/BUILD.gn b/BUILD.gn
-index a9ab6783fa..5e144b24e0 100644
+index d2bfb6129d..2c40c9f506 100644
 --- a/BUILD.gn
 +++ b/BUILD.gn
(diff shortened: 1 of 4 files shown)


jeroen:
Would it be possible to update again to the a new stable for example 9.6.180.12

Kwizatz:
@jeroen It should be. I have not abandoned ship, but keeping up with Google's updates is not so simple.

Each update brings new issues and it takes around 2 hours to build the library, add to that that I need to support both static and dynamic builds for Windows and static for Linux for the VCPKG version (I have to remove all MSYS related stuff for VCPKG, because it is seen as clutter and clean patches are better than faster delivery
ConfusesHandsupGIF
)

My own project using v8 is in hiatus right now, so updates are kind of a labor of love and a way for me to not to entirely  drop that project, so I decided to do the updates around every 6 months given my perception of the size of the public using it.

That said, I'll try and start the move to incorporate at least the MSVC patches upstream, not having to maintain patches for VCPKG would really help things moving along faster.

jeroen:
AH okay I wasn't aware you were manually writing these patches. Thanks for all the hard work!

PR #9190, opened 2021-07-22
[mingw-w64-sundials] fix build petsc-build error

gharveymn:
Should fix: https://github.com/msys2/MINGW-packages/pull/9117#issuecomment-883589541

mingw-w64-petsc-build uses the installation directory `${MINGW_PREFIX}/src/petsc-<version>` and I had hardcoded that version number before. The change makes the package use a `pacman -Qi` query to obtain that version number.

Is there a canonical way of doing this?


Biswa96:
`pacman` command should not be used in PKGBUILD file. What is the purpose of this change?

lazka:
puh, this is all hacky..  maybe something like ` pkgconf.exe --modversion /ucrt64/src/petsc-*/sso/lib/pkgconfig/PETSc.pc` ?

gharveymn:
@Biswa96 fixing a mistake on the initial merged package which is preventing it from building since PETSc got upgraded.

Sorry for the mistakes, trying to learn.

How about this. I took a look at what SLEPc is doing, so I think this latest commit should be the preferred way to do it.

lazka:
bump pkgrel

PR #9191, opened 2021-07-23
numpy: fix clang build

jeremyd2019:
Disable BLAS and fortran.  Numpy states both are optional to build, but had an issue in their build scripts if fortran wasn't found.

Add an explicit clang check to a GCC bug workaround for GCC < 8.4, because clang pretends to be a really old GCC version (like 4.x).

We could conceivably use a different BLAS library, if there's a compatible one that doesn't require fortran, but I'm not familiar with them.


jeremyd2019:
Seeing error including `_numpyconfig.h`.  Weird, I noticed that a couple of files were 'removed' in the diff even on mingw64, but I don't think anything I changed should have caused that... unless it was the `--skip-build` on the install step.  I will try taking that out and see if `_numpyconfig.h` comes back

PR #9194, opened 2021-07-23
numpy: remove --skip-build from install step

jeremyd2019:
The install was missing a few files, including `_numpyconfig.h`.  This is the only thing I can think of that could have caused it.

This does indeed seem to restore the missing files.  Sorry for the breakage, should have paid more attention to the files diff 🤦


jeremyd2019:
### Packages that failed due to _numpyconfig.h:
* python-lws
* python-numexpr
* python-pywavelets
* python-pandas
* python-pesq
* nlopt
* python-imagecodecs
* python-matplotlib
* openexr
* mypaint
* audaspace

### Packages that failed for other missing file from numpy
* python-numba

Issue #9147, opened 2021-07-16
qemu-system-arm segfaults: can't find libzstd.dll


BrendanSimon:
$ /c/msys64/mingw64/bin/qemu-system-arm.exe --help
C:/msys64/mingw64/lib/qemu/qemu-system-arm.exe: error while loading shared libraries: libzstd.dll: cannot open shared object file: No such file or directory

$ find /c/msys64 -name "libzstd.dll"
/c/msys64/mingw64/bin/libzstd.dll

Is there possibly a path or environment variable issue?  Or a build bug?

ZachBacon:
probably needs a rebuild

BrendanSimon:
I also just discovered the ucrt environment and thought I'd give that a crack -- unfortunately that too had the same issue.

$ /ucrt64/bin/qemu-system-arm.exe --help
C:/msys64/ucrt64/lib/qemu/qemu-system-arm.exe: error while loading shared libraries: libzstd.dll: cannot open shared object file: No such file or directory

jeremyd2019:
the report of which library fails to load has been known to lie.  Try using ntldd (mingw-w64-*-ntldd)

BrendanSimon:
Not sure how to use "mingw-w64-*-ntldd".

I installed with `pacboy sync ntldd:x"

$ /mingw64/bin/ntldd.exe -R /mingw64/bin/qemu-system-arm.exe
C:/msys64/mingw64/bin/qemu-system-arm.exe: not found

BrendanSimon:
I tried a rebuild using the instructions here:

https://www.msys2.org/wiki/Creating-Packages/

It fails to build :(

==> Verifying source file signatures with gpg...
    qemu-6.0.0.tar.xz ... FAILED (unknown public key 3353C9CEF108B584)
==> ERROR: One or more PGP signatures could not be verified!

BrendanSimon:
`--skipinteg` is getting further ...

jeremyd2019:
the missing dll is `libspice-server-1.dll`, part of `mingw-w64-x86_64-spice` package.  I guess mingw-w64-qemu needs to declare a dependency on that package.

jeremyd2019:
It seems to be declared an optional dependency, but it seems a little more than optional.

BrendanSimon:
I already have that installed.  I installed again, but still get the same error.  Must be another package perhaps ??

```
$ pacboy sync spice:x
[...]

PR #9201, opened 2021-07-24
qemu: Update to 6.1.0

heljkon:
Hello,

I've prepared the switch to 6.1.0. The PR currently builds 6.1.0-rc0, Qemu calls it version 6.0.90. So I assigned version number 6.0.90.rc0 to it. sha256sum are temporarily skipped.

My tests (included along PKGBUILD as a script) discovered 3 regressions. I will create issues in https://gitlab.com/qemu-project/qemu for these.

Along with the version change the build was changed a little bit, too:
- Patch https://gitlab.com/qemu-project/qemu/-/commit/1c96386.patch was removed, because 6.1.0 contains it (https://gitlab.com/qemu-project/qemu/-/commit/e2abbeac7a0c19e7757370da4c5164c862016d31)
- configure-Parameter "-cross-prefix" couldn't be dropped, it still is essential for the build to succeed
- configure-Parameter "-python" was dropped, which results in using system meson instead of qemu-tar meson

Most important change:
Binaries enter {MINGW}/bin and firmware files enter {MINGW}/share/qemu instead of putting all to {MINGW}/lib/qemu and creating wrappers in {MINGW}/bin as I've done before.
Are there any objections? My tests observed no problems with this. I introduced the wrappers in 5.2 to keep all binaries together as it was done in 5.1 (in {MINGW}/bin!!)


Biswa96:
> configure-Parameter "-cross-prefix" couldn't be dropped, it still is essential for the build to succeed

What is the error did you see after removing it?

Biswa96:
> configure-Parameter "-cross-prefix" couldn't be dropped, it still is essential for the build to succeed

What was the error after removing it?

heljkon:
> What was the error after removing it?

```
make[1]: *** No rule to make target 'multiboot.bin', needed by 'all'.  Stop.
make: *** [Makefile:189: pc-bios/optionrom/all] Error 2
...
[1599/1599] Generating QAPI rST doc reference newline-sanitized with a custom command (wrapped by meson to capture output)
    Aborting...
```

jeremyd2019:
Have you looked at #9147?  Specifically I was unable to run `qemu-system-aarch64` without installing the "optional" spice dependency.  So, it doesn't seem to be all that optional (as in, loaded at runtime if required rather than linked in so the OS loader tries to load it at start).

heljkon:
> Have you looked at #9147

Thanks to pointing me at it. I guess adding zstd to qemu should do! I will test this. spice is a very heavy dep, I would like to keep optional.

jeremyd2019:
zstd was a red herring.  The code in cygwin that tells you what dll is missing has issues (I believe since the split API set DLLs came along)

heljkon:
So, after reading the named thread more carefully I guess two tasks are mentioned

1. remove wrappers and place exe to bin, what I've already done anyway
2. make spice non optional

OK then let's do it...

jeremyd2019:
> 2. make spice non optional
> 
> OK then let's do it...

Either that or make it actually optional.  But I don't know how that could be accomplished

heljkon:
> My tests (included along PKGBUILD as a script) discovered 3 regressions. I will create issues in https://gitlab.com/qemu-project/qemu for these.

- https://gitlab.com/qemu-project/qemu/-/issues/500
- https://gitlab.com/qemu-project/qemu/-/issues/501
- https://gitlab.com/qemu-project/qemu/-/issues/502

jeremyd2019:
CLANG64 build runs Windows 10 ARM64 21354 :grin:

jeremyd2019:
Something I've noticed with the clang64 build at least is that when something is wrong with `qemu-system-aarch64` arguments, it doesn't print anything to the console to say what's wrong.  Also, `-serial stdio` is not printing anything.

jeremyd2019:
It seems that TPM support is not enabled in qemu.  I have been trying to get this working for Windows 11.  Is it reasonably easy to enable, or does it require extra libraries that aren't packaged yet?

jeremyd2019:
```sh
##########################################
# TPM emulation is only on POSIX

if test "$tpm" = ""; then
  if test "$mingw32" = "yes"; then
    tpm=no
  else
    tpm=yes
  fi
elif test "$tpm" = "yes"; then
  if test "$mingw32" = "yes" ; then
    error_exit "TPM emulation only available on POSIX systems"
  fi
fi
```

well that's disappointing.  wonder why?

heljkon:
I cannot confirm your observation.

> In fact, `--version` doesn't even print anything. I'm kind of thinking qemu-system-aarch64 is flagged as subsystem windows instead of console (and not just the `w` variant)

```
Normalo@DESKTOP-6MOKBL6 CLANG64 ~/qemu-tests/qemu-advent-calendar/2014
$ qemu-system-x86_64 -version
QEMU emulator version 6.0.90
Copyright (c) 2003-2021 Fabrice Bellard and the QEMU Project developers
```

> Something I've noticed with the clang64 build at least is that when something is wrong with `qemu-system-aarch64` arguments, it doesn't print anything to the console to say what's wrong. Also, `-serial stdio` is not printing anything.

I tested this using https://www.qemu-advent-calendar.org/2014/download/arm64.tar.xz (on of the tests of  /clang64/share/doc/qemu/msys2.examples.tests.sh).
After download and extraction, I executed:

```
User@DESKTOP-6MOKBL6 CLANG64 ~/qemu-tests/qemu-advent-calendar/2014
$  export MSYS2_ARG_CONV_EXCL="*"

User@DESKTOP-6MOKBL6 CLANG64 ~/qemu-tests/qemu-advent-calendar/2014
$  qemu-system-aarch64 -m 1024 -cpu cortex-a57 -machine virt -monitor none -kernel arm64/Image -append 'root=/dev/vda2 rw rootwait mem=1024M console=ttyAMA0,38400n8' -drive if=none,id=image,file=arm64/armv8.qcow2 -netdev user,id=user0,hostfwd=tcp::5555-:22 -device virtio-net-device,netdev=user0 -device virtio-blk-device,drive=image -serial stdio

(qemu:1616): Gtk-WARNING **: 10:12:04.098: Could not load a pixbuf from icon theme.
This may indicate that pixbuf loaders or the mime database could not be found.
[    0.000000] Linux version 3.18.0-1-linaro-vexpress64 (buildslave@x86-64-07) (gcc version 4.8.3 20140401 (prerelease) (crosstool-NG linaro-1.13.1-4.8-2014.04 - Linaro GCC 4.8-2014.04) ) #1ubuntu1~ci+141120142552 SMP PREEMPT Thu Nov 20 14:26:36 UTC 20
[    0.000000] CPU: AArch64 Processor [411fd070] revision 0
[    0.000000] Detected PIPT I-cache on CPU0
[    0.000000] Memory limited to 1024MB
```

heljkon:
> well that's disappointing. wonder why?

The snippet from `qemu-6.1.0-rc0/configure` shows once more that Windows is not the primary platform of Qemu.

There are other features / optimizations I really would like to have, too...

jeremyd2019:
What I saw was that `qemu-system-aarch64.exe` was flagged as `subsystem:windows` in the PE header.  It's pretty clear that upstream qemu intended it to be `subsystem:console`, and the variants with `w` appended to be `subsystem:windows`, but for some reason at least the `CLANG64` version both are `subsystem:windows`

jeremyd2019:
Looking at the .rsp file, it seems to pass first `-mwindows`  (possibly from SDL pkg-config?), and later `-mconsole`.  It appears the later `-mconsole` doesn't override the earlier `-mwindows`.

jeremyd2019:
> Looking at the .rsp file, it seems to pass first `-mwindows` (EDIT: twice!, seemingly once from SDL and once from SDL_image pkg-config?), and later `-mconsole`. It appears the later `-mconsole` doesn't override the earlier `-mwindows`.
> 
> https://bugs.llvm.org/show_bug.cgi?id=51296

https://github.com/llvm/llvm-project/commits/ce49fd024b43bd76b149f984b8f0d16e92b9bb06

Biswa96:
There are some huge additions in Hyper-V headers. If you test the new headers & crt from mingw-w64 it'd be great.

heljkon:
qemu-6.1.0-rc4 will be available soon. After rebuild I'll perform my tests with qemu-6.1.0-rc4.

heljkon:
I performed all tests on MINGW64 after local build of 6.0.94 and everything works as expected.

No problems with new headers & crt.

Biswa96:
The new changes are not in current headers and crt packages in msys2. You have to build headers and crt packages from latest mingw-w64 git repository. Then build qemu. I have build qemu without any issue. As you are testing qemu frequently it will be easier to catch any issue with more 👀

Biswa96:
Just asking. I shall do it myself.

heljkon:
I haven't understood the task in the first run as you may have read in my post.

I would like to help, but also don't want to do tasks, you have prepared already and which I could do wrong, resulting in testing wrong things.

Draft PRs are very helpful in this case I guess.

PR #9206, opened 2021-07-25
RapidJSON: Fix `PrettyWriter` feature with current compilers

mmuetzel:
The `PrettyWriter` functions of RapidJSON don't compile correctly with current compiler versions.

Cherrypick some patches from upstream that fix that feature.

Is it ok to just create a pull request? Or would you prefer if I file an issue first?


Biswa96:
If the patches are form upstream please add the upstream links as a comment above the `patch` command.

mmuetzel:
OK like this?

Biswa96:
# 👌

PR #9207, opened 2021-07-25
Update Pandas to 1.3.0

SirToffski:
Update Pandas to 1.3.0 - which is the latest version used by Arch. Tested build and the package for a couple on days - no issues to report.


Biswa96:
There is 1.3.1

SirToffski:
> There is 1.3.1

Ah - I always use the latest version in Arch repos - which is 1.3.0. 

1.3.1 only appeared on PYPI two hours ago - I will try building and testing it. If you'd rather wait, feel free to ignore this one.

lazka:
There are quite a few regression fixes in 1.3.1, see https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.3.1.html#fixed-regressions, so might be worth updating to 1.3.1 right away

SirToffski:
Sounds good, 1.3.1 works as expected - I am closing this pull and opening one for 1.3.1.

Cheers

PR #9210, opened 2021-07-26
[new-package] qwt-qt6

MehdiChinoune:


mingw-w64-qwt-qt6/PKGBUILD
```diff
@@ -0,0 +1,64 @@
...
+
+prepare() {
+  cd ${_realname}-${pkgver}
+
+  patch -p1 -i ${srcdir}/001-separate-debug-info-build.patch
+  patch -p1 -i ${srcdir}/002-link-to-release.patch
+  patch -p1 -i ${srcdir}/003-qwtconfig-msys2.patch
+}
+
+package() {
```
lazka, line 38: Any reason why there is no `build()` step?

MehdiChinoune: I have just copied the qwt file!

lazka: ah, I see, it uses `pkgdir` during the build. Seems weird but, OK I guess.

PR #9211, opened 2021-07-26
[new-package] PyQt6 6.1.1

MehdiChinoune:
Somehow the reference to QFileInfo::stat was missing from QtCore shared library.
The only way I have found to fix it is to remove its mention from the sip file.

### DIFF
--- a/mingw-w64-pyqt6-sip/PKGBUILD
+++ b/mingw-w64-pyqt6-sip/PKGBUILD
@@ -1,0 +1,34 @@
+# Contributor: Mehdi Chinoune <mehdi.chinoune@hotmail.com>
+
+_realname=pyqt6-sip
+pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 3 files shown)


mingw-w64-pyqt6-sip/PKGBUILD
```diff
@@ -0,0 +1,34 @@
+# Contributor: Mehdi Chinoune <mehdi.chinoune@hotmail.com>
+
+_realname=pyqt6-sip
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
```
lazka, line 5: Can we follow the arch naming here too? https://archlinux.org/packages/extra/x86_64/python-pyqt6-sip/

(I know it's not consistent with other packages here, but we try to match Arch when possible)

MehdiChinoune: Thanks.

PR #9215, opened 2021-07-26
bump pkgrel on numpy dependees that failed on clang64

jeremyd2019:


Biswa96:
I have updated some of those like - nlopt, python-pesq etc. See the pending page.

PR #9216, opened 2021-07-26
Drop gtksourceview2

lazka:
We want to get rid of gtk2 in the long run, and this package has no users.

If you still need it please open an issue and we can revert this.


lazka:
@Jehan OK with you? (since this is a gtk2 package)

Jehan:
@lazka I don't think we need it in GIMP, at least not as a direct dependency. Is it a GTK2 dependency maybe, which would allow some specific feature (which we might be using)? If not, we probably don't need/use it either.

lazka:
thanks  this doesn't have any reverse dependencies in msys2, so you can't pull this in by accident.

PR #9221, opened 2021-07-26
qwt: Enable generating pkgconfig file.

MehdiChinoune:


MehdiChinoune:
Is it OK to rename qwt to qwt-qt5?

MehdiChinoune:
The CI fails because of conflicting between qwt (qt5) and qwt (qt6-base) dependencies.

Biswa96:
You can separate any one of the packages here in different pull request.

MehdiChinoune:
> You can separate any one of the packages here in different pull request.

Done.

lazka:
Isn't this the same as #9225 now?

MehdiChinoune:
> Isn't this the same as #9225 now?

It's the same, but for qwt-qt5

lazka:
> Is it OK to rename qwt to qwt-qt5?

It was named that before (see the replaces/conflict/provides). I'd have to coordinate removing the old package after the rename to avoid a rename loop, but sure.

lazka:
> It's the same, but for qwt-qt5

I only see qt6 changes here.

MehdiChinoune:
> > It's the same, but for qwt-qt5
> 
> I only see qt6 changes here.

Are you sure!

lazka:
yes

PR #9223, opened 2021-07-26
autoreconf for clang packages missing DLLs, continued

jeremyd2019:
This continues #9217.  This should hopefully be all of them.  Once these hit the repo, I will update the files diff and see what else may need fixing.


jeremyd2019:
Oddly, this difference showed up on the gcc prefixes:
   ```diff
  -/mingw64/bin/libnpth6-0.dll
  +/mingw64/bin/libnpth-0.dll
  ```

jeremyd2019:
Well, I think I found where the `6-` came from, but I don't really understand why...
```diff
--- build-aux/ltmain.sh	2014-01-10 06:53:27.000000000 -0800
+++ /usr/share/libtool/build-aux/ltmain.sh	2021-04-12 15:28:01.000000000 -0700
 	    elif test -n "$soname_spec"; then
 	      # bleh windows
 	      case $host in
-              x86_64-*mingw32*)
-                func_arith $current - $age
-		major=$func_arith_result
-		versuffix="6-$major"
-		;;
-	      *cygwin* | mingw* | *cegcc*)
+	      *cygwin* | *msys* | mingw* | *cegcc* | *os2*)
 	        func_arith $current - $age
 		major=$func_arith_result
-		versuffix="-$major"
+		versuffix=-$major
 		;;
 	      esac
 	      eval soname=\"$soname_spec\"
```

jeremyd2019:
I added a commit to rebuild the *two* dependees of npth (gnupg and gpgme), due to the 64-bit DLL name having changed.  The other option would be to hack the `6-` case back in after autoreconf-ing.

jeremyd2019:
>```diff
> -	      *cygwin* | mingw* | *cegcc*)
> +	      *cygwin* | *msys* | mingw* | *cegcc* | *os2*)
> ```

This is particularly screwy, because it's switching on `$host` (the whole triple), which should never start with `mingw`.  I wonder if they're missing a leading `*` on that

lazka:
looks like a (very old) typo to me: https://github.com/autotools-mirror/libtool/commit/448a197074e7badaf0d932ab2ec1820d8b7d3cf7

lazka:
> The other option would be to hack the `6-` case back in after autoreconf-ing.

Looks fine to me as is. I don't see a reason for the "6" there.

PR #9233, opened 2021-07-28
binutils: Update to 2.37

lazka:
rebased patches; removed the ones in the release; backported one build fix from upstream

### DIFF
--- a/mingw-w64-binutils/0010-bfd-Increase-_bfd_coff_max_nscns-to-65279.patch
+++ b/mingw-w64-binutils/0010-bfd-Increase-_bfd_coff_max_nscns-to-65279.patch
@@ -1,7 +1,7 @@
-From 1316395a096fd327efe88c171a08ba077f6f02df Mon Sep 17 00:00:00 2001
+From 520d5017366ff334c8a866aa15abb8cce22fb952 Mon Sep 17 00:00:00 2001
 From: Ray Donnelly <mingw.android@gmail.com>
 Date: Mon, 23 Nov 2015 22:42:53 +0000
-Subject: [PATCH 5/5] bfd: Increase _bfd_coff_max_nscns to 65279
+Subject: [PATCH] bfd: Increase _bfd_coff_max_nscns to 65279
 
 .. from 32768. This is the value that llvm uses. There's no
 indication in the PECOFF specs that 32768 is the limit.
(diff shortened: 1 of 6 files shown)


Astrum-polaris:
`0410-windres-handle-spaces.patch` got reverted (partially) again in 2.37, do we still need it?

https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=749c700282097cf679ff019a9674d7c762f48619

https://sourceware.org/bugzilla/show_bug.cgi?id=27594#c7

In msys2 binutils we have another commit reverted to fix mintty build, do mingw binutils need that?

lazka:
From what I remember we needed 0410-windres-handle-spaces.patch for building ffmpeg, but that is inlcluded in the release now (or reverted again, I don't know). I tested building ffmpeg with this and everything seemed to work.

For the msys one, I suspect we can drop the revert then (we will get a conflict anyway)

Astrum-polaris:
> And in case you missed it, this removes 0410-windres-handle-spaces.patch

Yeah. I am talking about the `if defined (_WIN32) && !defined (__CYGWIN__)` part we patched in 2.36.1 got reverted again in 2.37
So I am wondering if we still need it.
If the original problem solved without patching that's fine too.

lazka:
Ah, that got reverted too.. ok, let me check.

lazka:
Now I'm confused why ffmpeg still built fine here with the fix removed.

lazka:
Ah, it was fixed here: https://github.com/FFmpeg/FFmpeg/commit/f9626d1065c43f1d51afe66bdf988b9f33729440 (thanks @mstorsjo !)

I'm wondering if we should just use upstream binutils now or revert like in the msys binutils: https://github.com/msys2/MSYS2-packages/blob/master/binutils/2002-Allow-spaces-in-the-name-of-the-external-preprocesso.patch

lazka:
I think I'll go for the revert for now, and maybe drop it next version.

mstorsjo:
FYI, this is my take of the situation:

https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=21c33bcbe36377abf01614fb1b9be439a3b6de20 - (1) Changed the `--preprocessor` option to treat it as one single path with potential spaces, instead of a string of multiple command line options (as it was documented before). This broke a couple projects that used the `--preprocessor` option as it was documented. In practice, this didn't seem to break things when running on native windows though (due to other details in how shells interpret some quoted paths), only when run in msys/cygwin mode or on actual unix. This issue is what my ffmpeg patch avoids, by just not using this option.

https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=cc3edc52747fd8b184ee48f1b0cc1ac0aca7832e - (2) Fixed handling of options like `-I` that are given paths with spaces and such. This commit is all good afaik, and this one corresponds to something that was patched manually in msys2 before, if I remember correctly.

In response to reports about problems caused by (1), but due to other confusion, (2) was in practice reverted, in https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=749c700282097cf679ff019a9674d7c762f48619.

And the discussion in https://sourceware.org/bugzilla/show_bug.cgi?id=27594 hasn't gotten any consensus regarding what to do, so 2.37 was released in this state, with the seemingly wrong commit reverted and a deadlock in the discussion.

So if you want the quoting fix that you used to have, you should keep patching it in, either in the form of (2) or in the form of what you had before.

lazka:
Thanks for the summary, that helps a lot.

PR #9244, opened 2021-07-30
libftdi: Remove check() since it was failing

DavidEGrayson:
Someone reported to me that the libftdi PKGBUILD script was failing during the `check` step and I was able to reproduce the issue. 
 There is no 'check' target anymore in the Makefile for libftdi that is generate by CMake.  Also, I don't know of any easy way to run tests for libftdi.  CMake gives an error if I pass `-DBUILD_TESTS=ON` and I didn't bother to look into fixing it.

This pull request fixes the PKGBUILD script by removing the `check` function.

### DIFF
--- a/mingw-w64-libftdi/PKGBUILD
+++ b/mingw-w64-libftdi/PKGBUILD
@@ -48,7 +48,7 @@ build() {
 
 check() {
   cd "${srcdir}/build-${MINGW_CHOST}"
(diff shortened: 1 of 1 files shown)


Biswa96:
I would suggest not to remove the `check` function. It may be useful for someone. Append `|| true` after that make command.

DavidEGrayson:
OK,  I changed it to just add `|| true`.

Issue #9240, opened 2021-07-29
some packages pull in avoidable GPL dependencies (when packaging non-GPL code)


eudoxos:
I am in the (for me) unusual position of packaging commercial software for windows; the software itself uses no GPL libraries, but some packages it depends on pull in GPL dependencies which are not strictly necessary.

Some of the dependencies are optional at build-time, e.g.:

* the BSD-licensed [ceres-solver](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-ceres-solver) which can be optionally built with GPL-licensed [suitesparse](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-suitesparse) (the package does include this dependency at build-time).

There are other cases where only some part of a package pulls in GPL code, but the package may well be usable for other purposes without it, e.g.:

* the LGPL-licensed [aravis](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-aravis/), packaged by myself, includes the library itself, but also a GStreamer plugin, which depends on [gst-plugins-bad](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-gst-plugins-bad) which pulls in a bunch of GPL-licensed libraries in turn.

* public-domain [sqlite3](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-sqlite3), of which terminal client (`sqlite3.exe`) links against GPL-licensed [readline](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-readline).

* not speaking about Qt (where some modules are GPL0only bust most are LGPL) or OpenCV (BSD/Apache2, but some dependencies are GPL).

So far just describing the situation, now what could be done?

1. For the package which have optional build-time dependency, I could create an alternative package (e.g. `ceres-solver-nogpl`) which would be built without that dependency. The `PKGBUILD` would specify it as conflicting with `ceres-solver` so that headers could be in the same location in both cases. Now, would the maintainers be willing to upstream such a package, to have them in the repository both? And if not, is there some documentation on creating private pacman repositories?

2. For the other packages, it would be at least useful to have a tool to check licenses of DLLs about to be distributed. In fact, I  wrote a simple script for this (it queries which packages they belong to, and then looks at the license of the package), but it is quite rudimentary.

I assume I am not the first one in this situation. Is that something which others would like to coordinate?

lazka:
I'm afraid with most of use doing this in our free time and/or working on open source software this isn't much of a priority.

[...]

PR #9247, opened 2021-07-30
split aravis to aravis (little deps) & aravis-gst (pulls in GStreamer) + update

eudoxos:
Please someone review the way how the split is done. I tested locally
and it is okay. Also see #9240 for context.


eudoxos:
I have to figure out minimal dependencies in private repo, pulling all GST plugins is perhaps not necessary. Sorry for noise.

PR #9248, opened 2021-07-30
Update aravis version, modularize, minimize deps

eudoxos:
Split packages for low-level library+tools and GStreamer plugin and GUI viewer.


eudoxos:
Thanks for merging. Is this a reason for concern? (from https://packages.msys2.org/queue) When I click on the build hyperlink, I can't identify any errors there...
image

Biswa96:
Nope. There is only 1 or 2 person amogus using clang32.

PR #9257, opened 2021-08-01
texlive-bin: autoreconf for clang support

### LINKED ISSUE #9256: texlive-bin: undefined symbol stpcpy on clang64
Interesting configure results:
```
configure:13653: checking whether stpcpy is declared
configure:13653: clang -c -march=x86-64 -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 conftest.c >&5
conftest.c:93:10: warning: implicitly declaring library function 'stpcpy' with type 'char *(char *, const char *)' [-Wimplicit-function-declaration]
  (void) stpcpy;
         ^
conftest.c:93:10: note: include the header <string.h> or explicitly provide a declaration for 'stpcpy'
1 warning generated.
configure:13653: $? = 0
configure:13653: result: yes
```

It may need to test linking to get a real answer.

jeremyd2019:
Also, autoreconf in texk/chktex specifically because it didn't seem to happen with the top-level autoreconf.  Fixes #9256 

Fix building wrappers with clang, which wants to use import libraries rather than dlls directly.

### DIFF
--- a/mingw-w64-texlive-bin/PKGBUILD
+++ b/mingw-w64-texlive-bin/PKGBUILD
@@ -4,7 +4,7 @@ pkgbase=mingw-w64-texlive-bin
 pkgname=("${MINGW_PACKAGE_PREFIX}-texlive-bin" "${MINGW_PACKAGE_PREFIX}-libsynctex")
 pkgver=2021.20210424
 pkgdesc="TeX Live binaries (mingw-w64)"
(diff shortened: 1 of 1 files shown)


mati865:
Autotools being autotools 😬 

FWIW LLD 13 will support linking DLL directly but as always import libs are strongly preferred.

PR #9259, opened 2021-08-01
Update mingw-w64

lazka:


mingw-w64-winstorecompat-git/PKGBUILD
```diff
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgver=9.0.0.6246.ae63cde27
```
Biswa96, line 6: pkgver is same.

MehdiChinoune:
Why not using [pkgver()](https://wiki.archlinux.org/title/creating_packages#pkgver())?

Biswa96:
All the mingw-w64 pkgbuild files are using `pkgver()` function.

jeremyd2019:
Note that commit 0e0eed623 *fixes* the intended behavior that `time_t` on 32-bit UCRT defaults to 64 instead of 32 bits.  This is potentially an ABI breakage for clang32, the only 32-bit UCRT currently in existence here.  /cc @revelator since you seem to be using clang32 as well.

lazka:
Not sure if that warrants rebuilds since clang32... I'd be fine with leaving things as is for now.

jeremyd2019:
Sounds good to me.  I can rebuild things if/when there's a failure, or maybe grep include files for `time_t` to try to find packages where `time_t` is in their external interface.

Biswa96:
😈 

1. Delete `/clang32` in repo
2. rebuild everything in it.

PR #9260, opened 2021-08-01
Modular Qt5

MehdiChinoune:
This the best solution I can come with.

I tried to build each submodule individually but qtdeclarative fails to build and I didn't find a solution. So I decided to build qtbase and qtdeclarative only from the single module and then build the other submodules individually.
Hope that works.

### DIFF
--- a/mingw-w64-qt5-3d/PKGBUILD
+++ b/mingw-w64-qt5-3d/PKGBUILD
@@ -1,0 +1,78 @@
+# Contibutor: Mehdi Chinoune <mehdi.chinoune@hotmail.com>
+
+_realname=qt5-3d
+pkgbase="mingw-w64-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+_qtver=5.15.2
+pkgver=${_qtver/-/}
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+pkgdesc="C++ and QML APIs for easy inclusion of 3D graphics (mingw-w64)"
+url='https://www.qt.io/'
+license=('GPL3' 'LGPL' 'FDL' 'custom')
+depends=("${MINGW_PACKAGE_PREFIX}-qt5-declarative"
+         "${MINGW_PACKAGE_PREFIX}-assimp")
+optdepends=("${MINGW_PACKAGE_PREFIX}-qt5-imageformats"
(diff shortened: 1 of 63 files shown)


MehdiChinoune:
Libraries built from submodule files:
    - 3d
    - activeqt
    - connectivity
    - datavis3d
    - imageformats
    - gamepad
    - graphicaleffects
    - lottie
    - multimedia
    - networkauth
    - purchasing
    - quickcontrols
    - quicktimeline
    - remoteobjects
    - scxml
    - sensors
    - serialbus
    - serialport
    - speech
    - svg
    - tools
    - translations
    - webglplugin
    - websockets
    - winextras
    - xmlpatterns
    - webchannel
    - webview
  
Libraries fails to:
    - declarative
    - location
    - quick3d
    - quickcontrols2
    - charts
    - script

MehdiChinoune:
It should be ready now.

jeremyd2019:
what's the plan with the existing monolithic qt5 package?  I expected it would have been removed in this pull request but apparently not...  Do I still need to update it too like #9357 ?

lazka:
We need to move all reverse deps away (like #9361) and then tell people to uninstall it, because it will conflict on upgrades :(

MehdiChinoune:
> what's the plan with the existing monolithic qt5 package? I expected it would have been removed in this pull request but apparently not... Do I still need to update it too like #9357 ?

My plan is:
1- Mark qt5 as a provider of all its components, so we the existing users can still use until we rebuild all the qt5-dependent packages.
2- Rebuild qt5-dependent packages one by one until we got no dependency on monolithic qt5
3- Remove monolithic qt5

lazka:
>  3- Remove monolithic qt5

can we make it a dummy package depending on all components instead?

jeremyd2019:
> 1- Mark qt5 as a provider of all its components, so we the existing users can still use until we rebuild all the qt5-dependent packages.

OK, then I won't worry about rebuilding it, as this will cause a rebuild (and it takes some time to build, so no reason to do it twice)

MehdiChinoune:
> > 3- Remove monolithic qt5
> 
> can we make it a dummy package depending on all components instead as step 3?

I think this is better.

jeremyd2019:
Question though... where did `fad2dc11.diff.zip::https://codereview.qt-project.org/changes/qt%2Fqtbase~333051/revisions/1/patch?zip` go in the new modular packages?  Is it still necessary?

Alexpux:
Probably better create Qt5-meta package and add to it conflicts and replaces for Qt5 package

Issue #9271, opened 2021-08-02
ruby: STATUS_BAD_STACK on longjmp with clang64


jeremyd2019:
The attempt to longjmp back to the _setjmpex at https://github.com/ruby/ruby/blob/378e8cdad69e6ba995a024da2957719789f0679e/thread.c#L823 is raising `STATUS_BAD_STACK` from `RtlUnwindEx`.

> An invalid or unaligned stack was encountered during an unwind operation.

In order to get ruby to not crash during its own bootstrapping with clang, I had to switch it from trying to use `__builtin_setjmp` to use mingw-w64's setjmp/longjmp, so this may be an issue on GCC if using UCRT's implementation.

jeremyd2019:
This works with GCC UCRT64 when switched to use `setjmp` rather than `__builtin_setjmp`

jeremyd2019:
Aha, https://github.com/msys2-contrib/mingw-w64/blob/master/mingw-w64-headers/crt/setjmp.h#L233-L240.  clang is using the mingw_getsp variant, which is returning the sp at the time, not the bp which is different (there's an alloca call earlier in the function).  Going to try enabling the `__builtin_frame_address` branch for clang, hopefully they implement that.

jeremyd2019:
Yep that fixes it.  Time to email a patch to mingw-w64!

jeremyd2019:
https://sourceforge.net/p/mingw-w64/mailman/message/37329802/

Biswa96:
Love your explanation 😍

jeremyd2019:
This was fixed upstream (mingw-w64/mingw-w64@1db0f476fbfdbb54dc089284fc278957c73cdd05 as github linked above).

PR #9265, opened 2021-08-02
ruby: fix clang build

jeremyd2019:
`__builtin_setjmp`/`__builtin_longjmp` are only meant for internal exception handling usage IIRC, use mingw-w64's setjmp instead.
> The built-in versions are used internally by GCC’s libraries to implement exception handling on some targets. You should use the standard C library functions declared in <setjmp.h> in user code instead of the builtins.

Pass full options --extern-only and --defined-only to nm, llvm-nm doesn't accept the truncated variant.

PR #9272, opened 2021-08-02
Revert "ruby: fix clang build"

jeremyd2019:
Reverts msys2/MINGW-packages#9265.  There's a mingw-w64 issue with setjmp/longjmp (#9271) that makes threading crash on it .


jeremyd2019:
These few packages built against clang64 ruby, but will likely need to be rebuilt especially with the changing of the soname for clang in #9274:
* mingw-w64-clang-x86_64-mruby-2.1.2-1
* mingw-w64-clang-x86_64-ngraph-gtk-6.09.03-1
* mingw-w64-clang-x86_64-ruby-hpricot-0.8.6-3
* mingw-w64-clang-x86_64-ruby-rdiscount-2.2.0.1-3

PR #9275, opened 2021-08-03
groonga: fix clang64 build

komainu8:
This PR fixes the following error.

```
../../groonga-11.0.2/lib/window_function_executor.cpp:392:22: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
  constexpr uint32_t unpack_source_id_record_id(uint64_t source_id) {
                     ^
../../groonga-11.0.2/lib/window_function_executor.cpp:393:30: note: shift count 32 >= width of type 'unsigned long' (32 bits)
    return source_id & ((1UL << 32) - 1);
                             ^
../../groonga-11.0.2/lib/window_function_executor.cpp:393:30: warning: shift count >= width of type [-Wshift-count-overflow]
    return source_id & ((1UL << 32) - 1);
                                 ^  ~~
```

### DIFF
--- a/mingw-w64-groonga/PKGBUILD
+++ b/mingw-w64-groonga/PKGBUILD
@@ -4,14 +4,15 @@ _realname=groonga
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=11.0.2
-pkgrel=2
+pkgrel=3
 pkgdesc="An opensource fulltext search engine (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 url="http://groonga.org/"
 license=('LGPL2')
 options=('strip' 'staticlibs')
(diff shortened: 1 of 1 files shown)


Biswa96:
Try not to add any GitHub Actions link in commit message. Those are removed after 90 days. Maybe add the error message or what you've fixed.

jeremyd2019:
> Try not to add any GitHub Actions link in commit message. Those are removed after 90 days.

Those particular signed URLs expire sooner than that.

komainu8:
@Biswa96 @jeremyd2019 Thank you for your comments.
I've fixed it.

Biswa96:
BTW, groonga seems to use bundled mruby. Is it posible to use the mruby mingw package?

komainu8:
@Biswa96 
> BTW, groonga seems to use bundled mruby. Is it posible to use the mruby mingw package?

No, it is not.
Groonga  uses some mrbgem. mrbgem need to add into mruby when we build it.
Because we can't install mrbgem after mruby install.

komainu8:
Thanks!

PR #9281, opened 2021-08-03
Allow building conficting packages by the same CI

MehdiChinoune:
This PR change the way ci build packages.

Previously the ci sort the packages by dependency and then build them
one after another. The upside of that approach is to be able to build
dependent packages within the same job, the downsides is that approach
makes it impossible to build conflicting packages within the same job,
and also doesn't offer a clean environment to build the independent packages.

With this approach we uninstall every package after a successful build
while adding a local (ci) repository from where the following packages
can install their dependencies. This approach will allow to build conflicting
packages within the same job and offering a clean environment to build
the independent packages, so we can discover missing make-dependencies.

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -31,6 +31,11 @@ define_build_order || failure 'Could not determine build order'
 message 'Building packages' "${packages[@]}"
 execute 'Approving recipe quality' check_recipe_quality
 
+message 'Adding an empty local repository'
+repo-add $PWD/artifacts/ci.db.tar.gz
+sed -i '1s|^|[ci]\nServer = file://'"$PWD"'/artifacts/\nSigLevel = Never\n|' /etc/pacman.conf
+pacman -Sy
+
 message 'Building packages'
 for package in "${packages[@]}"; do
     echo "::group::[build] ${package}"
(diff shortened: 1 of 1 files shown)


.ci/ci-build.sh
```diff
@@ -31,11 +31,16 @@ define_build_order || failure 'Could not determine build order'
...
+repo-add $PWD/artifacts/ci.db.tar.gz
+sed -i '1s|^|[ci]\nServer = file://'"$PWD"'/artifacts/\nSigLevel = Never\n|' /etc/pacman.conf
+pacman -Sy
+
 message 'Building packages'
 for package in "${packages[@]}"; do
     echo "::group::[build] ${package}"
     execute 'Fetch keys' "$DIR/fetch-validpgpkeys.sh"
-    execute 'Building binary' makepkg-mingw --noconfirm --noprogressbar --nocheck --syncdeps --rmdeps --cleanbuild
+    execute 'Building binary' makepkg-mingw --noconfirm --noprogressbar --needed --nocheck --syncdeps --rmdeps --cleanbuild
```
jeremyd2019, line 43: why is this, uh, needed?

.ci/ci-build.sh
```diff
@@ -53,6 +58,18 @@ for package in "${packages[@]}"; do
     cd - > /dev/null
     echo "::endgroup::"
 
+    echo "::group::[uninstall] ${package}"
+    repo-add $PWD/artifacts/ci.db.tar.gz "${package}"/*.pkg.tar.*
+    pacman -Sy
+    cd "$package"
+    for pkg in *.pkg.tar.*; do
+        message "Uninstalling ${pkg}"
+        pkgname="$(echo "$pkg" | rev | cut -d- -f4- | rev)"
```
jeremyd2019, line 67: Is this a safe way to get the package name from the filename?

.ci/ci-build.sh
```diff
@@ -31,11 +31,16 @@ define_build_order || failure 'Could not determine build order'
 message 'Building packages' "${packages[@]}"
 execute 'Approving recipe quality' check_recipe_quality
 
+message 'Adding an empty local repository'
+repo-add $PWD/artifacts/ci.db.tar.gz
```
jeremyd2019, line 35: I thought @lazka had to add a dummy package to ensure that a db was created?

.ci/ci-build.sh
MehdiChinoune: To not reinstall installed packages, saves time.

jeremyd2019: I've never seen makepkg install packages it didn't need to.

MehdiChinoune: Just type makepkg -h

.ci/ci-build.sh
lazka: I don't recall that. maybe elieux? anyway, this seems to work.

lazka:
perfect, thanks

.ci/ci-build.sh
jeremyd2019: https://github.com/msys2/msys2-main-server/blob/main/staging/runner.sh#L30 is what I was remembering

PR #9286, opened 2021-08-04
python-{imagecodecs,tifffile}: update to 2021.7.30

kmilos:


kmilos:
It was about insignificant API change, now imagecodecs is up to date w/ libavif.

The rest of libavif patches were also not impacting the function significantly (upstream forgot version bump and some tweaks to variables in .pc file). They're merged upstream and can wait until next release.

PR #9287, opened 2021-08-04
[CI] Remove dependent packages.

MehdiChinoune:
We should add "--cascade", as sometimes a single PKGBUILD produce many packages, which make it not possible to remove an earlier (alphabetically) package before its dependent package. and that package will stay installed.


jeremyd2019:
I thought of this when reviewing the earlier PR, but then thought it might remove something that should remain (ie, part of toolchain) if something "core" were being built.

MehdiChinoune:
> I thought of this when reviewing the earlier PR, but then thought it might remove something that should remain (ie, part of toolchain) if something "core" were being built.

I have added some commands to ensure that the toolchain is installed before building.

Issue #9289, opened 2021-08-04
Exception handling is broken on mingw32 when using static runtime libraries.


rvogg:
If you build ccache >= 4.2 for mingw32, it crashes at runtime. 
This bug started to occur when static linking was enabled for the gcc runtime libraries.  (https://github.com/ccache/ccache/pull/732)

The error almost always occurs in the same place: 
In Util.cpp read_file rais an exception:
```
std::string
read_file(const std::string& path, size_t size_hint)
{
  if (size_hint == 0) {
    auto stat = Stat::stat(path);
    if (!stat) {
>>    throw Error(strerror(errno));
    }
    size_hint = stat.size();
  }
```

But the catch block in the calling function is never reached.
```
std::string data;
  try {
>>  data = Util::read_file(path);
  } catch (const Error&) {
    // Ignore.
    return counters;
  }
```

Here is the callstack of "ccache -s" : 
```
msvcrt.dll!msvcrt!_exit (Unbekannte Quelle:0)
msvcrt.dll!msvcrt!abort (Unbekannte Quelle:0)
uw_init_context_1(struct _Unwind_Context * context, void * outer_cfa, void * outer_ra) (c:\_\M\mingw-w64-gcc\src\gcc-10.3.0\libgcc\unwind-dw2.c:1593)
_Unwind_RaiseException(struct _Unwind_Exception * exc) (c:\_\M\mingw-w64-gcc\src\gcc-10.3.0\libgcc\unwind.inc:93)
__cxxabiv1::__cxa_throw(void * obj, std::type_info * tinfo, void (*)(void *) dest) (c:\_\M\mingw-w64-gcc\src\gcc-10.3.0\libstdc++-v3\libsupc++\eh_throw.cc:90)
Util::read_file(const std::string & path, size_t size_hint) (...\cache-4.3\src\Util.cpp:1164)
Statistics::read(const std::string & path) (...\cache-4.3\src\Statistics.cpp:196)
operator()(const struct {...} * const __closure, const std::string & path) (...\cache-4.3\src\Statistics.cpp:102)
std::__invoke_impl<void, collect_counters(const Config&)::<lambda(const string&)>&, const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&>(std::__invoke_other, struct {...} &)(struct {...} & __f) (c:\msys64\mingw32\include\c++\10.3.0\bits\invoke.h:60)
[...]

PR #9290, opened 2021-08-04
ccache: update to 4.3

rvogg:
I reused 19b5e0678a2b4e3e02536494ad324f3fb7deabe2 and updated to version 4.3. 
After investigating the bug in the 32-bit version, I found a quick solution by disabling the static linking.  

The bug itself is described at https://github.com/msys2/MINGW-packages/issues/9289


mingw-w64-ccache/PKGBUILD
```diff
@@ -4,47 +4,65 @@
...
+    extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  # Note: enabling static linking as default has brought a bug to light in mingw32
+  #       "-DSTATIC_LINK=OFF" can be removed as soon as the bug is fixed
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    ${MINGW_PREFIX}/bin/cmake \
+      -G"MSYS Makefiles" \
```
1480c1, line 46: Can you try Ninja?

lazka: (it was broken with 4.2.1, but maybe that changed)

rvogg: Ninja build is broken.
(failed on generating MANUAL.xml)

PR #9291, opened 2021-08-04
Clang update 12.0.1-3

### LINKED ISSUE #9285: clang: undefined symbol `std::__1::chrono::system_clock::from_time_t(long long)`
Got this while trying to build clang on clang32:

```
  lld-link: error: undefined symbol: std::__1::chrono::system_clock::from_time_t(long long)
  >>> referenced by libLLVMSupport.a(VirtualFileSystem.cpp.obj):(llvm::vfs::InMemoryFileSystem::addFile(llvm::Twine const&, long long, std::__1::unique_ptr<llvm::MemoryBuffer, std::__1::default_delete<llvm::MemoryBuffer> >, llvm::Optional<unsigned int>, llvm::Optional<unsigned int>, llvm::Optional<llvm::sys::fs::file_type>, llvm::Optional<llvm::sys::fs::perms>, llvm::vfs::detail::InMemoryFile const*))
  >>> referenced by libLLVMSupport.a(VirtualFileSystem.cpp.obj):(llvm::vfs::InMemoryFileSystem::addFile(llvm::Twine const&, long long, std::__1::unique_ptr<llvm::MemoryBuffer, std::__1::default_delete<llvm::MemoryBuffer> >, llvm::Optional<unsigned int>, llvm::Optional<unsigned int>, llvm::Optional<llvm::sys::fs::file_type>, llvm::Optional<llvm::sys::fs::perms>, llvm::vfs::detail::InMemoryFile const*))
```

Looks to be fallout from #9259 change in `time_t` from 32 to 64 bits.

jeremyd2019:
- Removed switch for using `MSYS Makefiles` generator instead of `Ninja`.  Now that clang32 linking is more reliable this shouldn't be necessary anymore.
- backport patches to allow `--stack` and `--image-base` to work with = as well as space between them and their value
- backport patch to allow a later `-mconsole` to override `-mwindows`, for https://github.com/msys2/MINGW-packages/pull/9201#issuecomment-890403294
- backport patch to allow `0x` prefix to language ID on `llvm-rc` and `llvm-windres` command line, for #9253
- cleaned up cmake arg variables.  There were several, that were not entirely clear how they were used.  Now, there are 
  - `common_cmake_args` are args common to all cmake invocations,
  - `platform_config` are args specific to the main llvm build (generally platform-specific, hence the name), and
  - `libcxx_config` are args common to the libc++/libunwind shared and static builds.

**HACK** added to libc++ to export overloads for 32-bit time_t when building with 64-bit time_t on clang32.  This is required to allow both new binaries built with 64-bit time_t and old ones built with 32-bit time_t to use the same libc++.dll.  This situation happens when building clang itself, so build a shared libc++ first, and override PATH and LDFLAGS to get it used during the build.  This is a **temporary** workaround for the broken state of ABI that we are currently in.  Fixes #9285

### DIFF
--- a/mingw-w64-clang/0020-llvm-rc-Allow-specifying-language-with-a-leading-0x-.patch
+++ b/mingw-w64-clang/0020-llvm-rc-Allow-specifying-language-with-a-leading-0x-.patch
@@ -1,0 +1,101 @@
+From 46020f6f0c8aa134002208b2ecf0593b04c46d08 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Martin=20Storsj=C3=B6?= <martin@martin.st>
+Date: Mon, 2 Aug 2021 14:10:22 +0300
+Subject: [PATCH] [llvm-rc] Allow specifying language with a leading 0x prefix
+
+This option is always interpreted strictly as a hexadecimal string,
+even if it has no prefix that indicates the number format, hence
+the existing call to StringRef::getAsInteger(16, ...).
+
+StringRef::getAsInteger(0, ...) consumes a leading "0x" prefix is
+present, but when the radix is specified, the radix shouldn't
+be included.
+
+Both MS rc.exe and GNU windres accept the language with that
+prefix.
+
+Also allow specifying the codepage to llvm-windres with a different
+radix, as GNU windres allows that (but MS rc.exe doesn't).
+
+This fixes https://llvm.org/PR51295.
+
+Differential Revision: https://reviews.llvm.org/D107263
+---
+ test/tools/llvm-rc/codepage.test |  2 ++
+ test/tools/llvm-rc/language.test |  2 ++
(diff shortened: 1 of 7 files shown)


jeremyd2019:
Looking forward to 13.0 :grin:

mati865:
It's been branched already and honestly I can't wait to bring patches count to something more "reasonable" in terms of packaging.

MehdiChinoune:
I am thinking about modularizing llvm packages (build each component individually).
What do you think?

jeremyd2019:
I've confirmed the fix for --image-base on openldap and meanwhile, and I've confirmed the fix for `0x` language id with graphviz.  I have not tried rebuilding qemu with this version of clang for the `-mwindows`/`-mconsole` fix.

jeremyd2019:
I've confirmed that qemu is producing both console and windows subsystem executables as intended.

PR #9296, opened 2021-08-05
qt6-tools: remove conflict with qt5 (qt5-tools).

MehdiChinoune:
I haven't bump the pkgrel as the package is not uploaded yet to the stable repos.


lazka:
> I haven't bump the pkgrel as the package is not uploaded yet to the stable repos.

I guess this isn't true now?

MehdiChinoune:
> > I haven't bump the pkgrel as the package is not uploaded yet to the stable repos.
> 
> I guess this isn't true now?

Done.

PR #9300, opened 2021-08-06
qt6-declarative: remove conflict with qt5 (qt5-declarative)

MehdiChinoune:


lazka:
Just wondering, Arch just has "6" as a suffix: https://archlinux.org/packages/extra/x86_64/qt6-declarative/ Is this maybe the preferred upstream way to avoid conflicts?

MehdiChinoune:
> Just wondering, Arch just has "6" as a suffix: https://archlinux.org/packages/extra/x86_64/qt6-declarative/ Is this maybe the preferred upstream way to avoid conflicts?

qt6 offer wo destinations for binaries the first is INSTALL_BINDIR where it installs the executables and the second is INSTALL_PUBLICBINDIR where it installs symlinks (suffixed with 6 to avoid conflict) to those executables.

In archlinux they chose INSTALL_BINDIR=lib/qt6/bin and INSTALL_PUBLICBINDIR=bin which works perfectly for them.
https://github.com/archlinux/svntogit-packages/blob/packages/qt6-base/trunk/PKGBUILD#L42

The problem on MINGW-w64 is that executables and shared libraries (dlls) share the same installation destination. Installing shared libraries on "lib/qt6/bin" or any destination other than "bin" will make them unavailable at runtime.

lazka:
ok

PR #9304, opened 2021-08-06
libglade: update for python3 support

jeremyd2019:
There's just one standalone python script, so just run 2to3 over it, and fix up mixed tabs and spaces indentation errors.

configure has an error about missing `GNOME_COMMON_INIT`, but it has that when gnome-common is installed too, so didn't add it to makedepends.

### DIFF
--- a/mingw-w64-libglade/PKGBUILD
+++ b/mingw-w64-libglade/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libglade
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.6.4
-pkgrel=5
+pkgrel=6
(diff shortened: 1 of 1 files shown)


Biswa96:
Do we need libglade package? graphviz can be built without it.

jeremyd2019:
🤷

revelator:
libglade is old and deprecated but if we can still build it and it does not hurt anything else i dont see why not :)

jeremyd2019:
D'oh, forgot graphviz depends on #9291.  Will let it fail on clang64, then force-push back to just libglade, and make a new PR for graphviz, on the assumption that clang and libglade can be merged first.

Biswa96:
python2 was removed from graphviz recently here https://gitlab.com/graphviz/graphviz/-/merge_requests/1987. So, this may not be required in later graphviz released.

PR #9309, opened 2021-08-07
libspatialite: Fix build for clang

Biswa96:
@revelator @podsvirov @jeroen Review please.

### DIFF
--- a/mingw-w64-libspatialite/0002-use-win32-mutex-locks.patch
+++ b/mingw-w64-libspatialite/0002-use-win32-mutex-locks.patch
@@ -1,0 +1,38 @@
+--- a/src/connection_cache/alloc_cache.c
++++ b/src/connection_cache/alloc_cache.c
+@@ -51,7 +51,7 @@
+ #include <libloaderapi.h>
+ #endif
+ 
+-#if defined(_WIN32) && !defined(__MINGW32__)
++#if defined(_WIN32)
+ #include <windows.h>
+ #else
+ #include <pthread.h>
+@@ -110,7 +110,7 @@
+ 
+ /* GLOBAL semaphores */
+ int gaia_already_initialized = 0;
+-#if defined(_WIN32) && !defined(__MINGW32__)
++#if defined(_WIN32)
+ static CRITICAL_SECTION gaia_cache_semaphore;
+ #else
+ static pthread_mutex_t gaia_cache_semaphore = PTHREAD_MUTEX_INITIALIZER;
+@@ -1262,7 +1262,7 @@
+ SPATIALITE_PRIVATE void
+ splite_cache_semaphore_lock (void)
+ {
+-#if defined(_WIN32) && !defined(__MINGW32__)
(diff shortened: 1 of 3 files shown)


revelator:
looks good to me :)

revelator:
not related though i found out why smpeg does not build with clang also, it seems to rely on gcc as the build horse and does not link by default to the clang rt builtins library leading to undefined symbols. I got around it temporarily by doing this ->

```
build() {
  [[ -d ${srcdir}/build-${MINGW_CHOST} ]] && rm -rf ${srcdir}/build-${MINGW_CHOST}
  cd "${srcdir}/build-${MINGW_CHOST}/" 
  case "${CARCH}" in
    i686)
      export LIBS+=" -lclang_rt.builtins-i386"
    ;;

    x86_64)
      export LIBS+=" -lclang_rt.builtins-x86_64"
    ;;
  esac
  CFLAGS+=" -Wno-error=narrowing"
  ../${_realname}/configure \
    --prefix=${MINGW_PREFIX} \
    --build=${MINGW_CHOST} \
    --host=${MINGW_CHOST} \
    --target=${MINGW_CHOST} \
    --mandir=${MINGW_PREFIX}/share/man \
    --enable-shared \
    --enable-static

  make
}
```

podsvirov:
LGTM too :-)

Biswa96:
Is it possible to upstream these changes? Not exactly same changes but what upstream can accept. I have seen the upstream repository but it seems not to use git as SCM.

podsvirov:
> Is it possible to upstream these changes? Not exactly same changes but what upstream can accept. I have seen the upstream repository but it seems not to use git as SCM.

You're right. The author uses [fossil](https://fossil-scm.org/). I have experience with this `SCM`. If no one protested, will I try to promote these patches on my own behalf?

podsvirov:
@Biswa96, looks like I need you help in [SpatiaLite-Users](https://groups.google.com/g/spatialite-users/c/g5uViVYty9k) list.

Biswa96:
What kind of help? I am not familiar with the project in-depth.

podsvirov:
Help to response on question:

> 3. I've never used the clang compiler on Windows, but I 
   easily imagine that it has problems in supporting
   libpthreads; do you confirm ?

PS: I've never used the clang compiler on Windows too

jeremyd2019:
clang on windows can support libpthreads, via the winpthreads package from mingw-w64, same as GCC.  But, the clang prefixes here are configured so that libc++ uses native windows threads instead, so winpthreads is not pulled along automatically by things linking to the C++ runtime as it is with gcc.

podsvirov:
The [0002-use-win32-mutex-locks.patch](https://github.com/msys2/MINGW-packages/pull/9309/files#diff-ec663eff12f52509553a9e1842b4c1afa1378cf9223424d64a83cde5bab30804) path applyed [there](https://www.gaia-gis.it/fossil/libspatialite/info/08517f176a819de2).

PR #9310, opened 2021-08-07
Add package "GNU Octave"

mmuetzel:
This is the first time, I try to get a package added to MSYS2. So, I hope this is ok. Any feedback is very welcome.

GNU Octave is a high-level interpreted language, primarily intended for numerical computations. It provides capabilities for the numerical solution of linear and nonlinear problems, and for performing other numerical experiments. It also provides extensive graphics capabilities for data visualization and manipulation. Octave is normally used through its interactive command line interface, but it can also be used to write non-interactive programs. The Octave language is quite similar to Matlab so that most programs are easily portable.

See: https://www.octave.org

I hope that it is eligible to be included in MSYS2.

I tried to keep the `PKGBUILD` file in line with the ones I found for other packages.

I tried building it for mingw64 locally and it seems to be working for me. I haven't tried the other architectures.

### DIFF
--- a/mingw-w64-octave/0001-binary-install-location.patch
+++ b/mingw-w64-octave/0001-binary-install-location.patch
@@ -1,0 +1,45 @@
+# HG changeset patch
+# User Markus Mützel <markus.muetzel@gmx.de>
+# Date 1628329289 -7200
+#      Sat Aug 07 11:41:29 2021 +0200
+# Branch stable
+# Node ID 5a2de6e9d0f66361abfee17823030cd56fe51ef3
+# Parent  116dbaba5a74d79892b8f7d97d0d28a8a0686a7d
+Install all binaries in bin directory.
+
+If these binaries are installed in the archlib directory, it looks like they
+cannot find the libraries that are installed in the bin directory. Install
+them in the bin directory as well.
+
+diff -r 116dbaba5a74 -r 5a2de6e9d0f6 build-aux/subst-config-vals.in.sh
+--- a/build-aux/subst-config-vals.in.sh	Sat Jul 31 16:07:45 2021 +0200
++++ b/build-aux/subst-config-vals.in.sh	Sat Aug 07 11:41:29 2021 +0200
+@@ -250,7 +250,7 @@
+   -e "s|%NO_EDIT_WARNING%|DO NOT EDIT!  Generated automatically by subst-config-vals.|" \
+   -e "s|%NO_OCT_FILE_STRIP%|${NO_OCT_FILE_STRIP}|" \
+   -e "s|%OCTAVE_API_VERSION%|\"${api_version}\"|" \
+-  -e "s|%OCTAVE_ARCHLIBDIR%|\"${archlibdir}\"|" \
++  -e "s|%OCTAVE_ARCHLIBDIR%|\"${bindir}\"|" \
+   -e "s|%OCTAVE_BINDIR%|\"${bindir}\"|" \
+   -e "s|%OCTAVE_BINDIR%|\"${bindir}\"|" \
+   -e "s|%OCTAVE_CANONICAL_HOST_TYPE%|\"${canonical_host_type}\"|" \
+diff -r 116dbaba5a74 -r 5a2de6e9d0f6 src/module.mk
+--- a/src/module.mk	Sat Jul 31 16:07:45 2021 +0200
++++ b/src/module.mk	Sat Aug 07 11:41:29 2021 +0200
+@@ -43,14 +43,14 @@
+ OCTAVE_VERSION_LINKS += %reldir%/octave-cli-$(version)$(EXEEXT)
+ 
+ if AMCOND_BUILD_QT_GUI
+-  archlib_PROGRAMS += %reldir%/octave-gui
++  bin_PROGRAMS += %reldir%/octave-gui
+   OCTAVE_VERSION_LINKS += %reldir%/octave-gui-$(version)$(EXEEXT)
+ 
+   OCTAVE_INTERPRETER_TARGETS += %reldir%/octave-gui$(EXEEXT)
+ endif
+ 
+ if AMCOND_BUILD_QT_GUI
+-  archlib_PROGRAMS += %reldir%/octave-svgconvert
++  bin_PROGRAMS += %reldir%/octave-svgconvert
+   OCTAVE_INTERPRETER_TARGETS += %reldir%/octave-svgconvert$(EXEEXT)
+ endif
+ 
--- a/mingw-w64-octave/PKGBUILD
+++ b/mingw-w64-octave/PKGBUILD
@@ -1,0 +1,74 @@
+_realname=octave
+pkgbase=mingw-w64-${_realname}
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+pkgver=6.3.0
+pkgrel=1
+pkgdesc="GNU Octave: Interactive programming environment for numerical computations."
+url="https://www.octave.org"
+license=('GPL3')
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libgfortran"
+         "${MINGW_PACKAGE_PREFIX}-lapack"
+         "${MINGW_PACKAGE_PREFIX}-arpack"
+         "${MINGW_PACKAGE_PREFIX}-curl"
+         "${MINGW_PACKAGE_PREFIX}-fftw"
+         "${MINGW_PACKAGE_PREFIX}-ghostscript"
+         "${MINGW_PACKAGE_PREFIX}-gl2ps"
+         "${MINGW_PACKAGE_PREFIX}-glpk"
+         "${MINGW_PACKAGE_PREFIX}-graphicsmagick"
+         "${MINGW_PACKAGE_PREFIX}-hdf5"
+         "${MINGW_PACKAGE_PREFIX}-libsndfile"
+         "${MINGW_PACKAGE_PREFIX}-qhull"
+         "${MINGW_PACKAGE_PREFIX}-suitesparse"
+         "${MINGW_PACKAGE_PREFIX}-sundials"
+         "${MINGW_PACKAGE_PREFIX}-qrupdate"
+         "${MINGW_PACKAGE_PREFIX}-qscintilla"
+         "${MINGW_PACKAGE_PREFIX}-qt5")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc-fortran"
+             "${MINGW_PACKAGE_PREFIX}-gnuplot"
+             "${MINGW_PACKAGE_PREFIX}-fltk"
+             "${MINGW_PACKAGE_PREFIX}-portaudio")
+optdepends=('texinfo: for help-support in Octave'
+            "${MINGW_PACKAGE_PREFIX}-fltk: alternative plotting and dialogs"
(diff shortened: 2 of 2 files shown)


1480c1:
For clang64, could you list the dependencies in the commit, so it's easier to track without needing to go back to this pr every time?

jeremyd2019:
Looks like fortran.  There is no fortran compiler for clang subsystem (yet -- maybe someday flang will be up to speed)

mmuetzel:
Thank you very much for the swift addition.

Afaict, the missing dependencies for the clang subsystem indeed are most importantly a Fortran compiler and Fortran math libraries like BLAS and LAPACK.

gharveymn:
This exactly what I was hoping for when I PR'd SUNDIALS! I'm currently working on a port to CMake, so I I'm pretty familiar with the issues surrounding building/using with MSYS2. Here's some of them:

- GraphicsMagick is 8-bit when it would be preferred to be > 16-bit. Should just be a one-line PR.
- No MinGW packages currently for GZip, Zip, and Unzip. I'm sitting on PKGBUILDs for all three, but I'm not sure why MSYS2 doesn't already have packages for them, so I haven't sent a PR yet.
- No MinGW Texinfo package, so `help` doesn't work. I see the fix in your PKGBUILD but it seems to not be working on my system. I think it's because `MSYSTEM_PREFIX` isn't set without firing up the MSYS2 shell. I currently have a method for creating a standalone Makeinfo executable instead.

Issues with the current package:

- PKGBUILD should patch octave.pc for some of the -L flags which are leaking the build environment currently.
- Again, `help` isn't really working.
- Using octave.exe from cmd doesn't work as expected because of the usage of `P_OVERLAY` in `octave_execv_wrapper`. octave-cli and octave-gui work fine.
- `mkoctfile` isn't working. Fails with "The system cannot find the path specified."

gharveymn:
> Without the correct environment, various things may and will (sometimes silently) break. The exception is using mingw subsystems from pure Windows, which shouldn't require any special environment apart from an entry in PATH. Do not set MSYSTEM outside of the shells, because that will also break things.

[ref](https://www.msys2.org/wiki/MSYS2-introduction/#:~:text=Without%20the%20correct,also%20break%20things.) 

So yes, generally things should work from native Windows (CMD) when using the MinGW environments. Other than compiling the Perl code of Makeinfo into a private standalone executable for our purposes, I don't really see an easy solution. Packaging Texinfo for MinGW isn't really possible at the moment because large parts of it are written in shell code, so it would be missing significant parts of the distribution.

On mkoctfile: Perhaps add an option to set the slash type (or full conversion to POSIX/Windows style path?), otherwise just convert to forward slashes or inspect the `CC` variable beforehand and convert if needed? Windows usually does just fine with forward slashes, GCC evidently requires them, and I just tested that MSVC (cl.exe) is fine with them, so maybe make conversion to forward slashes the default? This is probably something to discuss with the maintainers.

mmuetzel:
mkoctfile uses `getenv` to query the value of the `TEMP` environment variable. For some reason that returns `C:\msys64\tmp` (i.e. a path including backslashes) while I see the following at the prompt:
```
$ echo $TEMP
/tmp
```
Is this expected?

gharveymn:
[Yes, I believe so](https://www.msys2.org/wiki/Porting/#:~:text=when%20building%20native%20software%20using%20MSYS2%27s%20tools%2C%20you%20must%20be%20mindful%20that%20when%20msys2%20executes%20a%20native%20program%2C%20it%20will%20translate%20environment%20variables%20and%20command%20arguments%20from%20msys2-form%20to%20native-form). This is because you are executing native software from inside the MSYS2 shell, and so the environment variables are translated before execution. This discrepancy does not occur when you invoke from `cmd`.

mmuetzel:
It looks like forward vs. backward slashes was a red herring.
The following command works for me:
```
$ CXX=x86_64-w64-mingw32-g++ mkoctfile is_system_function.cc -v
x86_64-w64-mingw32-g++ -c -DPSAPI_VERSION=1 -D__USE_MINGW_ANSI_STDIO=1  -IC:\msys64\mingw64\include\octave-6.3.0\octave\.. -IC:\msys64\mingw64\include\octave-6.3.0\octave -IC:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe    is_system_function.cc -o C:\msys64\tmp/oct-krgIcm.o
x86_64-w64-mingw32-g++ -IC:\msys64\mingw64\include\octave-6.3.0\octave\.. -IC:\msys64\mingw64\include\octave-6.3.0\octave -IC:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe  -o is_system_function.oct  C:\msys64\tmp/oct-krgIcm.o  -shared -pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high  -LC:\msys64\mingw64\lib\octave\6.3.0 -LC:\msys64\mingw64\lib -LC:\msys64\mingw64\lib\octave\6.3.0 -loctinterp -loctave  -pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high
```
So, it looks like the default compiler messes this up:
```
$ mkoctfile -p CXX
/mingw64/bin/g++
```

`mkoctfile` uses the same compilers that were used to compile Octave by default. It creates a string with the command and sends it to the command interpreter using [`system`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem?view=msvc-160). Would that require converting from POSIX to Windows paths?

Alternatively, would it be ok to configure with `CXX=${MINGW_CHOST}-g++` and Co. when building Octave? Most other `PKGBUILD` rules seem to use `CXX=${MINGW_PREFIX}/bin/g++`. Is there a particular reason for that?

mmuetzel:
> * Using octave.exe from cmd doesn't work as expected because of the usage of `P_OVERLAY` in `octave_execv_wrapper`. octave-cli and octave-gui work fine.

Could you please elaborate why it is expected that spawning a process with `P_OVERLAY` doesn't work? How should that wrapper program start the actual program instead to make it work?

mmuetzel:
I believe I can reproduce the error with `makeinfo` that you are seeing. This occurs for me if the current drive differs from the frive where MSYS2 is installed.
I'll try and add a fix for that in #9372.

gharveymn:
> CXX=x86_64-w64-mingw32-g++
Great, this also works for me when using cmd:

```
> set CXX=x86_64-w64-mingw32-g++
> set CC=x86_64-w64-mingw32-gcc
> mkoctfile bug_54096.c -v
x86_64-w64-mingw32-gcc -c -DPSAPI_VERSION=1 -D__USE_MINGW_ANSI_STDIO=1  -Ic:\msys64\mingw64\include\octave-6.3.0\octave\.. -Ic:\msys64\mingw64\include\octave-6.3.0\octave -Ic:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe    bug_54096.c -o C:\Users\Gene\AppData\Local\Temp/oct-lDZMVa.o
x86_64-w64-mingw32-g++ -Ic:\msys64\mingw64\include\octave-6.3.0\octave\.. -Ic:\msys64\mingw64\include\octave-6.3.0\octave -Ic:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe  -o bug_54096.oct  C:\Users\Gene\AppData\Local\Temp/oct-lDZMVa.o  -shared -pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high  -Lc:\msys64\mingw64\lib\octave\6.3.0 -Lc:\msys64\mingw64\lib -Lc:\msys64\mingw64\lib\octave\6.3.0 -loctinterp -loctave  -pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high
```

> `CXX=${MINGW_CHOST}-g++`

If you want to do this you'll need to add `mingw-w64-cross-gcc` as a dependency because packages are built using the plain MSYS2 shell, without an environment. So, for example `x86_64-w64-mingw32-c++` will need to resolve to `/opt/bin/x86_64-w64-mingw32-c++.exe` because `/mingw64/bin` isn't in the PATH by default. Maybe ping someone with more expertise for better advice on this.

> Could you please elaborate why it is expected that spawning a process with P_OVERLAY doesn't work?

Unfortunately in `cmd` the behavior of `P_OVERLAY` is such that instead of replacing the process memory while using the same handle, it creates a new subprocess for octave-cli/gui and exits thereafter. This returns the user to the prompt while the subprocess is still running in the background. It doesn't look like the same behavior occurs when using an MSYS2-MinGW shell, but regardless, the MinGW packages are intended to be Windows native, with the ability to be plopped into your PATH and ran from `cmd`. This behavior can also be unexpected for users doing things like debugging Octave with GCC because GCC will try to track the process handle. [Further reading](https://stackoverflow.com/questions/26815733/calling-exec-or-spawnp-overlay-unblocks-the-console-any-way-to-keep-the-con).

mmuetzel:
I tried starting Octave from a CMD shell and it is working for me after I added the respective folders to `PATH`:
```
set PATH=C:\msys64\mingw64\bin;C:\msys64\usr\bin;%PATH%
octave --gui
```

The input at the CLI seems to be quite erratic though. (But that's the same if I start without the wrapper program: `octave-gui --cli`.)
The GUI seems to work fine.
That might be an issue with `readline` on Windows.
@lostbard added a bunch of patches to MXE Octave that we "normally" use to package for Windows. It might be that some of those would be helpful here, too.
See the `readline-*.patch` files here: https://hg.octave.org/mxe-octave/file/tip/src

mmuetzel:
I submitted PR #9394 with a potential fix for the apparent issues with `readline`.

mmuetzel:
I ran all my tests in the Windows Terminal setup like described [here](https://www.msys2.org/docs/terminals/).
It seems to be working correctly in that terminal.

However, if I start the CLI (by executing `octave-gui`) with the minTTY (i.e. the terminal that opens by running "MSYS2 MinGW 64-bit" from the start menu, Octave seems to ignore any input (apart from Ctrl+C).
Also, executing `octave` on that shell spawns a background process that seems to detach from the terminal. (That is not what is wanted.)
Are there any known incompatibilities of minTTY with respect to spawning subprocesses and redirecting console in/out?

Biswa96:
Setting `MSYS=enable_pcon` environment variable in may help. It enables ConPTY features.

mmuetzel:
Thanks for the hint. Unfortunately, running `MSYS=enable_pcon octave-gui` at the minTTY prompt doesn't seem to make a difference.
Or do I have to set that environment variable before starting minTTY?

Biswa96:
> do I have to set that environment variable before starting minTTY?

Yes.

mmuetzel:
Setting `MSYS=enable_pcon` before starting minTTY (from the start menu) helps when the Octave CLI is started with `octave-gui`. Input at the Octave prompt seems to be correctly redirected to Octave's interpreter. :+1:

Can we do something to help it work correctly if the user doesn't set that variable?

Trying to start the program with the wrapper `octave` still doesn't work. But in contrast to before, it no longer seems to spawn a background process (or it dies immediately after it is spawned).

Some background: The wrapper program `octave` is supposed to detach to either `octave-gui` or `octave-cli` depending on the command line parameters the user passed.

zklhp:
Hello,

For me, in Octave GUI and in Emacs octave-mode, install packages still gives

pkg: unable to find the octave executable, Octave installation is incomplete

Could you please give me some hints for debugging.

Thank you very much. :heart:

mmuetzel:
Thanks for checking and reporting. See PR #9494 for a possible fix.

PR #9311, opened 2021-08-07
Build qwt-qt5 and qwt-qt6 with the same PKGBUILD file.

MehdiChinoune:


Biswa96:
Would you like to do this same thing with poppler?

PR #9314, opened 2021-08-07
Fix CI when the same PKGBUILD build conflicting packages.

MehdiChinoune:
Fixes #9313

Issue #9295, opened 2021-08-05
clang32 packages with time_t references


jeremyd2019:
I'm doing a grep through clang32 packages looking for potential trouble - either `time_t` in includes or `time_tEl` in bin or lib (that's part of the mangling of the problematic function in libc++).  Hitting quite a few.  And the time_t include grep doesn't necessarily catch consumers of problematic APIs only providers...

_Originally posted by @jeremyd2019 in https://github.com/msys2/MINGW-packages/issues/9285#issuecomment-893173343_

The actual mangled name for 32-bit time_t from libc++ is `__ZNSt3__16chrono12system_clock11from_time_tEl` (the last letter is `x` for 64-bit time_t)

<details>
<summary>Quick grep script</summary>

```bash
#!/bin/bash

for pkg in ../clang32mirror/clang32/*.pkg.tar.zst; do
        flag=
        mkdir t
        tar -C t -xf "$pkg"
        grep -sqr time_tEl t/clang32/bin t/clang32/lib && flag=y
        grep -sqrE '\<time_t\>' t/clang32/include t/clang32/i686-w64-mingw32/include && flag=y
        if [[ -n "$flag" ]]; then
                basename "$pkg"
        fi
        rm -rf t
done
```
</details>

## List of packages from grep

- [x] mingw-w64-clang-i686-apr-1.6.5-3-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-aria2-1.35.0-2-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-aribb24-1.0.3-4-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-assimp-5.0.1-9-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-aws-c-common-0.5.5-1-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-boost-1.75.0-6-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-catch-2.13.7-1-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-cgal-5.2-2-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-chicken-4.12.0-1-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-clang-12.0.1-2-any.pkg.tar.zst
- [x] mingw-w64-clang-i686-clustal-omega-1.2.4-1-any.pkg.tar.zst
[...]

PR #9316, opened 2021-08-08
bump packages for clang32 time_t rebuild

jeremyd2019:
These are packages that were part of "bootstrapping" clang prefixes.
For #9295.  I don't know if their dependencies also need rebuilding, but
this is a start.

Includes commit from #9314 so CI can succeed.


jeremyd2019:
Manually applied just the second commit in d93748c8e.

PR #9324, opened 2021-08-08
fix typecode decoding DoS

totaam:
As per the report upstream: https://github.com/aresch/rencode/pull/29
I have requested a CVE for this bug - pending.
It affects any software that uses rencode. ie: deluge, xpra and many others. Many of those will be available over the network. Ouch!


Biswa96:
Increase `pkgrel` variable.

PR #9327, opened 2021-08-08
smpeg{,2}: libtool hack

### LINKED ISSUE #9317: smpeg does not build with clang
not related though i found out why smpeg does not build with clang also, it seems to rely on gcc as the build horse and does not link by default to the clang rt builtins library leading to undefined symbols. I got around it temporarily by doing this ->

```
build() {
  [[ -d ${srcdir}/build-${MINGW_CHOST} ]] && rm -rf ${srcdir}/build-${MINGW_CHOST}
  cd "${srcdir}/build-${MINGW_CHOST}/" 
  case "${CARCH}" in
    i686)
      export LIBS+=" -lclang_rt.builtins-i386"
    ;;

    x86_64)
      export LIBS+=" -lclang_rt.builtins-x86_64"
    ;;
  esac
  CFLAGS+=" -Wno-error=narrowing"
  ../${_realname}/configure \
    --prefix=${MINGW_PREFIX} \
    --build=${MINGW_CHOST} \
    --host=${MINGW_CHOST} \
    --target=${MINGW_CHOST} \
    --mandir=${MINGW_PREFIX}/share/man \
    --enable-shared \
    --enable-static

  make
}
```

_Originally posted by @revelator in https://github.com/msys2/MINGW-packages/pull/9309#issuecomment-894637478_

jeremyd2019:
These do not work when updating their bundled libtool files with the newer ones from MSYS2-packages, so instead apply the minimal patch to let it use compiler-rt, and use `lt_cv_deplibs_check_method=pass_all` to work around the other clang issues with libtool.  Fixes #9317

Note CI fails building the source package, apparently because the sources are from svn.

### DIFF
--- a/mingw-w64-smpeg/PKGBUILD
+++ b/mingw-w64-smpeg/PKGBUILD
@@ -4,7 +4,7 @@ _realname=smpeg
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.4.5
-pkgrel=2
+pkgrel=3
 pkgdesc="SDL MPEG Player Library (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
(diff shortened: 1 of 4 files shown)


revelator:
probably the easiest way around it i agree :)

PR #9330, opened 2021-08-09
coin: fix clang build

jeremyd2019:
It was using a C-style idiom of `typedef struct {...} foo;` in C++, which clang
didn't like because they were also trying to export that struct.  Use
C++ style `struct foo {...};` instead.


Biswa96:
@podsvirov Any comment on this change?

podsvirov:
It's cool becouse this already [fixed](https://github.com/coin3d/coin/commit/46d61fd40609429e42d73996329fec17a3690fc9) in `master`.

PR #9331, opened 2021-08-09
libtiff: enable lerc codec

kmilos:


kmilos:
Hm, not sure why the clang32 build of lerc is missing in the repo, it seemed fine [here](https://github.com/msys2/MINGW-packages/pull/9292/checks?check_run_id=3250110637).

kmilos:
Duh, wasn't in the arch list originally, sorry or the noise...

PR #9344, opened 2021-08-10
baobab: Open directory in explorer patch

### LINKED ISSUE #9343: baobab cannot open directories
Context menu -> Open Folder raises "Failed to open file" "No application is registered as handling this file" on directories.

mugiseyebrows:
Fixes #9343


Biswa96:
1. Increase `pkgrel` variable.
2. If this patch is from upstream put a link of that commit or merge request as a comment.

mugiseyebrows:
No, it's not from upstream, it's windows-specific, I think it wont be merged.

Biswa96:
BTW, I am not familiar with GNOME environment but does it have some equivalent of LOAD_LIBRARY_SEARCH_SYSTEM32 flag? I mean instead of using "explorer" string can it be more specific like System32 folder only?

lazka:
It's not ideal that we carry functional patches that wont make it upstream... but this is small enough I guess.

PR #9345, opened 2021-08-10
cotire: update for CMake 3.21

Optiligence:


lazka:
pkgrel?

Optiligence:
see https://github.com/msys2/MINGW-packages/issues/8988#issuecomment-867906070

i was wondering if i should change that to avoid confusion but i like it the way it is
next time i could add a comment that stays visible in future diffs, i guess

lazka:
ah, yes, sorry

PR #9348, opened 2021-08-11
libftdi: Fix libdir in pkgconfig

Biswa96:
@DavidEGrayson Review please.


DavidEGrayson:
This change looks good to me.  I confirmed you're just fixing the `.pc` file, not actually moving any files from `bin` to `lib`.

PR #9371, opened 2021-08-13
pango: Add rounding code patch

lb90:
Backport a commit fixing https://gitlab.gnome.org/GNOME/pango/-/issues/578


lb90:
@Biswa96 I have dropped [that patch](https://github.com/msys2/MINGW-packages/blob/48bec78eb23a46d5905c1d3bbec6200addbe5d8f/mingw-w64-pango/0001-fix-stdio-include.patch) because I think that it's not necessary anymore. Can you confirm that?

Biswa96:
I have not built the current version. So, I am not sure if that patch is required here. But I have already upstreamed the required hunk https://gitlab.gnome.org/GNOME/pango/-/merge_requests/381

lazka:
Thanks!

PR #9372, opened 2021-08-13
octave: Fix mkoctfile with default settings. Fix pkgconfig files.

mmuetzel:
Like @gharveymn noticed in #9310, `mkoctfile` doesn't work with default settings and `octave.pc` leaks paths from the build system.

The proposed change fixes both issues for me.

Is it ok to configure with `CXX=${MINGW_CHOST}-g++` and Co. when building Octave? Most other PKGBUILD rules seem to use `CXX=${MINGW_PREFIX}/bin/g++`. Is there a particular reason for that?

### DIFF
--- a/mingw-w64-octave/PKGBUILD
+++ b/mingw-w64-octave/PKGBUILD
@@ -2,7 +2,7 @@ _realname=octave
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
 pkgver=6.3.0
-pkgrel=1
+pkgrel=2
 pkgdesc="GNU Octave: Interactive programming environment for numerical computations."
 url="https://www.octave.org"
 license=('GPL3')
@@ -56,12 +56,10 @@ build() {
     --libexecdir=${MINGW_PREFIX}/lib \
     --sbindir=${MINGW_PREFIX}/bin \
     --enable-shared --disable-static \
+    --enable-relocate-all \
     --with-blas="-lblas" \
     ac_cv_search_tputs=-ltermcap \
-    JAVA_HOME="" \
-    CC=${MINGW_PREFIX}/bin/gcc \
-    CXX=${MINGW_PREFIX}/bin/g++ \
-    F77=${MINGW_PREFIX}/bin/gfortran
+    JAVA_HOME=""
 
   make all
 }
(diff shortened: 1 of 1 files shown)


Biswa96:
AFAIK, the prefixed compiler are for cross compiling. Those are in mingw toolchain for compatibility.

mmuetzel:
Does that mean they shouldn't be used in the build rules?

IIUC, the current issue with `mkoctfile` in Octave is that POSIX paths can't be used when running a command with [`system`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem?view=msvc-160).
By default, `mkoctfile` uses the same compilers that were found during configure. Those are used literally (e.g. `/mingw64/bin/g++`). That seems to cause an error.
Alternatively, a user can override which compilers should be used by setting corresponding environment variables. If I set these environment variables (even to the exact same value), it is working because the path seems to be translated to a Windows path.
E.g.:
```
$ mkoctfile is_system_function.cc -v
/mingw64/bin/g++ -c -DPSAPI_VERSION=1 -D__USE_MINGW_ANSI_STDIO=1  -IC:\msys64\mingw64\include\octave-6.3.0\octave\.. -IC:\msys64\mingw64\include\octave-6.3.0\octave -IC:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe    is_system_function.cc -o C:\msys64\tmp/oct-9O0sJP.o
Das System kann den angegebenen Pfad nicht finden.
$ CXX=/mingw64/bin/g++ mkoctfile is_system_function.cc -v
C:/msys64/mingw64/bin/g++ -c -DPSAPI_VERSION=1 -D__USE_MINGW_ANSI_STDIO=1  -IC:\msys64\mingw64\include\octave-6.3.0\octave\.. -IC:\msys64\mingw64\include\octave-6.3.0\octave -IC:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe    is_system_function.cc -o C:\msys64\tmp/oct-2rarxw.o
C:/msys64/mingw64/bin/g++ -IC:\msys64\mingw64\include\octave-6.3.0\octave\.. -IC:\msys64\mingw64\include\octave-6.3.0\octave -IC:\msys64\mingw64\include   -fopenmp -march=x86-64 -mtune=generic -O2 -pipe  -o is_system_function.oct  C:\msys64\tmp/oct-2rarxw.o  -shared -pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high  -LC:\msys64\mingw64\lib\octave\6.3.0 -LC:\msys64\mingw64\lib -LC:\msys64\mingw64\lib\octave\6.3.0 -loctinterp -loctave  -pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high
```
I'd like that `mkoctfile` would work "out of the box" (without the need to set environment variables).

With the "plain" executable names, that error doesn't occur (no substitution between POSIX and Windows paths is necessary).
Is there some (easy) way I could patch Octave such that it would do the same substitution that seems to be done for environment variables? Some system call maybe?

lazka:
Can't you just use `CC=gcc` etc?

mmuetzel:
Perfect! That was too easy! 👍 
I'll try that and update the PR if it does the trick.

mmuetzel:
Afaics, `configure` choses the non-prefixed compilers anyway. So, I removed setting those variables for `configure` completely.
`mkoctfile` is working for me with that change.

Also, `help` is working for me now also when the current disc differs from the disc where MSYS2 is installed.

PR #9373, opened 2021-08-13
fritzing: use new split qt5 packages

jeremyd2019:
Figured I'd help out on the effort on the one package I know about building.  This uses `lrelease` as part of building, which seems to be part of qt5-tools, so added that as a makedepends.


jeremyd2019:
Changes contained in #9376

Issue #4317, opened 2018-08-26
"Too many open files" and binutils 2.31.1-1


slomp:
I'm getting "Too many open files" when building LLVM/clang (release_60 branch) with the most recent version of binutils (2.31.1-1).
Running ulimit -n 3200 did not help.
Downgrading to [2.30-2](http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-binutils-2.30-2-any.pkg.tar.xz) solved the problem (without even needing to tweak the limit, which defaults to 256).

mingwandroid:
You probably need to reduce the level of parallelism in the call to make

slomp:
Already tried doing that to with `-j1` and the same problem persisted.

mingwandroid:
Oh weird.

slomp:
For reference, here's a similar 2016 discussion on the issue:
https://sourceforge.net/p/msys2/mailman/message/34988811/
That's where I got the hint that the problem was related to binutils.

mingwandroid:
Yeah, it does seem to be binutils running into the limit.

mingwandroid:
You should report this at binutil's bugzilla really.

slomp:
Oh, I see, so does that mean that you managed to also reproduce the issue on a non-MinGW system?

If that's the case, I'll go ahead and file a bug on GNU binutils, providing this thread as a reference.

Alexpux:
Will downgrade binutils package to 2.30

slomp:
FYI, a new post on the bug report thread suggests that the bug has been fixed on binutils master:
https://sourceware.org/bugzilla/show_bug.cgi?id=23573

Alexpux:
Will see when new binutils will be released

Alexpux:
@slomp at least no one patch mentioned in bug report doesnt resolve the issue

revelator:
the fix on linux is setting /etc/security limits higher, but i doubt this would work here.
Also i just tested the new binutils-2.32 and it still fails on building clang with too many open files despite the dev saying its fixed :(.  Back to the drawing board with this one it seems.

slomp:
Yeap, clang is resource hog when it comes to building it :(

manisandro:
FYI, this is fixed in master, and the commit which fixes this is https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commit;h=999d6dff80fab12d22c2a8d91923db6bde7fb3e5

[...]

PR #9374, opened 2021-08-13
qtwebkit: update dependencies

MehdiChinoune:


MehdiChinoune:
It fails on MinGW64 with messages like
`D:/a/_temp/msys/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find *a library*: Too many open files`
and on MinGW32 with a message indicating a mlformed archive
`D:/a/_temp/msys/msys64/mingw32/lib/gcc/i686-w64-mingw32/10.3.0/../../../../i686-w64-mingw32/bin/ld.exe: ../../lib/libWTF.a: error adding symbols: malformed archive`
Any Idea?

MehdiChinoune:
It looks like It is a binutils related problem, as It happened before #4317

lazka:
We should try backporting https://sourceware.org/pipermail/binutils/2021-July/117264.html

lazka:
I've added a comment here: https://sourceware.org/bugzilla/show_bug.cgi?id=24723

PR #9378, opened 2021-08-13
qt5-multimedia: Enable WMF plugin

Biswa96:
@MehdiChinoune Review please.


MehdiChinoune:
I couldn't say It is a good ar bad, but since It is just an extra plugin, so It won't the main library.
OK from me.

PR #9390, opened 2021-08-15
Update dependencies for PyQt5 and QScintilla

MehdiChinoune:
I have disabled qtwebkit for pyqt5, until we resolve the failure in #9374


MehdiChinoune:
Failure because of pywin32, which doesn't exist in MinGW-w64.

PR #9394, opened 2021-08-15
readline: Update to version 8.1 patch level 001.

mmuetzel:
Also adopt patches from MXE Octave.

That resolves the erratic behavior of input at the Octave CLI for me.


mingw-w64-readline/PKGBUILD
```diff
@@ -19,34 +19,46 @@ options=('staticlibs' 'strip')
...
+            'dd11d64101b1b51ad42c61a78c72247440f1c18a67876dc36d2e6f058a2f1894'
+            '0242c9311fb97383d59627df33c66099c582401174a9bc6c80c52fbb1cc1b66f'
+            '917b65f9abb63161aa2469fa3a8dec8f4868f230d04b56fb824520c46c5870e4'
+            'fd511721455814fe760beb8696b58fd93fbc7e52fccd999c7b514160370d1248'
+            '682a465a68633650565c43d59f0b8cdf149c13a874682d3c20cb4af6709b9144'
             'SKIP')
 
 prepare() {
   cd "${srcdir}/readline-${_basever}"
-  patch -p1 -i ${srcdir}/0001-Add-private-history_rename-function-and-in-WIN32-use.patch
```
lazka, line 43: Why was this removed?

mingw-w64-readline/PKGBUILD
```diff
@@ -19,34 +19,46 @@ options=('staticlibs' 'strip')
...
+  # Patch adopted from https://hg.octave.org/mxe-octave/file/41e50d658de0/src/readline-1-sigwinch.patch
+  patch -p1 -i ${srcdir}/0001-sigwinch.patch
+  # Patch adopted from https://hg.octave.org/mxe-octave/file/41e50d658de0/src/readline-1-fixes.patch
+  patch -p1 -i ${srcdir}/0002-S_ISxxx.patch
+  # Patch adopted from https://hg.octave.org/mxe-octave/file/41e50d658de0/src/readline-1-input.patch
+  patch -p1 -i ${srcdir}/0003-input.patch
+  # Patch adopted from https://hg.octave.org/mxe-octave/file/41e50d658de0/src/readline-1-display.patch
+  patch -p1 -i ${srcdir}/0004-display.patch
+  # Patch adopted from https://hg.octave.org/mxe-octave/file/41e50d658de0/src/readline-2-event-hook.patch
+  patch -p1 -i ${srcdir}/0005-event-hook.patch
```
lazka, line 60: puh... can you give any more context as to why you maintain those patches and where they are coming from?

mingw-w64-readline/PKGBUILD
mmuetzel: The bug that this patch was addressing is fixed upstream. So, it is no longer needed.

mingw-w64-readline/PKGBUILD
mmuetzel: The Octave community is providing an installer for Windows that is cross-built on Linux with MXE Octave. It bundles Octave with its dependencies and some development tools.
MXE Octave is a forge of [MXE](https://mxe.cc/).

These patches were originally added to MXE Octave here:
https://hg.octave.org/mxe-octave/rev/2cf09b449d88
And here:
https://hg.octave.org/mxe-octave/rev/098ca5276ab8
And have since been updated occasionally to still apply when readline was updated.

I'm not the original author of these two patches. But I can try and reach out to the community for comments as to why we are using them.

mingw-w64-readline/PKGBUILD
lazka: ok, thanks

mingw-w64-readline/PKGBUILD
mmuetzel: No one seems to remember why we are using these particular two patches in MXE Octave. So, I tried to build readline without them. 
The erratic behavior of the command line interface of Octave is resolved also without these patches.

Should I just drop them? 
Should I force push a version without these patches? Or do you prefer to keep the history of this pull request?

mmuetzel:
Tbh, I was blindly assuming that the patches that we had in MXE Octave are necessary for something.
But after some testing, the only patch that is obviously necessary (or readline wouldn't even compile) is the first one (signal handling on mingw).
The other patches from MXE Octave are still available if other issues should arise in the future. But for the time being, I removed them from this pull request.

PR #9421, opened 2021-08-19
qt5 meta-package

MehdiChinoune:


Biswa96:
meta package? I thought the "qt5" name would be a [group][1] like toolchain, qt6 etc.

[1]: https://packages.msys2.org/group/

MehdiChinoune:
> meta package? I thought the "qt5" name would be a [group](https://packages.msys2.org/group/) like toolchain, qt6 etc.

@lazka suggested to make a meta-package for qt5 https://github.com/msys2/MINGW-packages/pull/9260#issuecomment-897855181

Alexpux:
I'm also for meta-package

Alexpux:
But I don't see in this commit meta-package.
You must create **mingw-qt5-metapackage** folder with PKGBUILD.
In this PKGBUILD you need add all qt5 modules as depends, add to conflicts **qt5** and **qt5-debug** and add the same to provide and replaces entries.

@Biswa96 this is needed for proper migration user from monolith qt5 to qt5-modules

MehdiChinoune:
> But I don't see in this commit meta-package.
> You must create **mingw-qt5-metapackage** folder with PKGBUILD.
> In this PKGBUILD you need add all qt5 modules as depends, add to conflicts **qt5** and **qt5-debug** and add the same to provide and replaces entries.
> 
> @Biswa96 this is needed for proper migration user from monolith qt5 to qt5-modules

Should we make qt5 or qt5-meta ?

Alexpux:
qt5-meta.
Add to conflicts and replaces **qt5-debug** packages
Also add entry **provides** with qt5

MehdiChinoune:
> qt5-meta.
> Add to conflicts and replaces **qt5-debug** packages
> Also add entry **provides** with qt5

done

Alexpux:
Ray cannot be maintainer of package. Use your own name and email

MehdiChinoune:
Marking qt5-meta to provide qt5 make it conflicts with qt5-base.
Does "provides" make the provider inherits the "conflicts" of the provided?

Alexpux:
Remove then provides from qt5-base

MehdiChinoune:
> Remove then provides from qt5-base

provides exists on qt5 not qt5-base, should I remove it?

lazka:
The only reason I suggested a meta package was so we can get rid of the qt5 package and have an upgrade path for users. I don't see the value without that tbh.

Alexpux:
This will drop qt5-static package

lazka:
@Alexpux is there any reason why you want to keep the full qt5 package? Maybe we can improve the other packages to cover your use cases?

Alexpux:
I don't have reason to keep monolith Qt5,just not drop Qt5-static

MehdiChinoune:
> I don't have reason to keep monolith Qt5,just not drop Qt5-static

qt5-statix has its own package under mingw-w64-qt5-static

PR #9423, opened 2021-08-19
Update newton-dynamics: fix compile error in Newton.h

ImperatorS79:
This fix has been committed in newton main repository but no new version yet.

### DIFF
--- a/mingw-w64-newton-dynamics/PKGBUILD
+++ b/mingw-w64-newton-dynamics/PKGBUILD
@@ -27,6 +27,7 @@ prepare() {
   cd "${srcdir}"/${_realname}-${pkgver}
   patch -p1 -i "${srcdir}"/0001-fix-arch-detection.patch
   patch -p1 -i "${srcdir}"/0002-use-msys-tinyxml.patch
+  # https://github.com/MADEAPPS/newton-dynamics/commit/a6782a4c50c68964d20ca8ebf9621da14bf8ebf0
   patch -p1 -i "${srcdir}"/0003-fix-newton-h.patch
 }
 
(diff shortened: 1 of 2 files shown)


Biswa96:
As this change is a upstream commit, a comment with that commit link in PKGBUILD file would be useful to maintain this patch.

ImperatorS79:
I have no idea which commit it is unfortunately. The history is broken because the folder structure was changed and I do not know how to search for that in github.

Biswa96:
Alright. The repository seems to be weird. Here is the commit https://github.com/MADEAPPS/newton-dynamics/commit/a6782a4c50c68964d20ca8ebf9621da14bf8ebf0

ImperatorS79:
Obviousy the sha256sum 🤦.

mingw-w64-newton-dynamics/PKGBUILD
```diff
@@ -56,6 +60,10 @@ build() {
     -DCMAKE_BUILD_TYPE="Release"\
     -G"MSYS Makefiles" \
     -DNEWTON_DOUBLE_PRECISION=ON \
+    -DNEWTON_WITH_SSE4_PLUGIN=OFF \
+    -DNEWTON_WITH_AVX2_PLUGIN=OFF \
+    -DNEWTON_WITH_AVX_PLUGIN=ON \
+    -DNEWTON_WITH_REFERENCE_GPU_PLUGIN=ON \
```
lazka, line 66: Why these changes?

ImperatorS79: I tried to enable avx2 and sse4 plugin but it failed and I forgot those modifications in the PKGBUILD.

lazka: ok, I see they are the default anyway, thanks

PR #9437, opened 2021-08-20
tulip: fix for clang

jeremyd2019:
* template fixes
* handle linking libomp like libgomp.
* dll import/export fixes
* tell cmake to ignore `/usr/lib` to avoid it picking up msys2 libbfd.


jeremyd2019:
Aha:
```
C:/msys64/clang64/include/c++/v1/__support/win32/locale_win32.h:76:7: error:
      definition of type 'locale_t' conflicts with typedef of the same name
class locale_t {
      ^
C:/msys64/usr/include/sys/_locale.h:10:28: note: 'locale_t' declared here
typedef struct __locale_t *locale_t;
```

It's mixing mingw and msys headers somehow...

mati865:
Maybe it's the same bug as LLVM had (it's fixed already) and we have to explicitly ignore `/usr`?
https://github.com/msys2/MINGW-packages/blob/063edb5d1f9d9d97584ac773df9a842f2567b944/mingw-w64-clang/PKGBUILD#L294

revelator:
that might be it indeed

jeremyd2019:
That might be the first one, but the second one (where apparently clang doesn't dllexport everything it tries to dllimport when switching which decoration is used on a templatized class) is probably stickier.

raedrizqie:
@jeremyd2019 what is the status of this PR? did you get any further?

jeremyd2019:
I haven't looked at this since then.

raedrizqie:
i managed to compile it successfully for clang64 (locally), but it failed on github actions.. i added 2 more patches to your patches: https://github.com/raedrizqie/MINGW-packages/commit/55c3fdb7c68bec0ca7cecd3ad41f91b694ec06a9

maybe you can have a look..

it failed with these error:

referenced by libbfd.a:
>ld.lld: error: undefined symbol: stpcpy
ld.lld: error: undefined symbol: fcntl
ld.lld: error: undefined symbol: __errno
ld.lld: error: undefined symbol: libintl_dgettext
ld.lld: error: undefined symbol: __getreent
.. and many more

jeremyd2019:
Thanks, that seems to have been the missing piece here.

PR #9438, opened 2021-08-21
build box2d static libs

fundies:
was broken/removed in last commit


mingw-w64-box2d/PKGBUILD
```diff
@@ -34,10 +34,28 @@ build() {
     ../${_realname}-${pkgver}/
 
   make
+  
+  mkdir -p "${srcdir}/build-${CARCH}-static"
+  cd "${srcdir}/build-${CARCH}-static"
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  ${MINGW_PREFIX}/bin/cmake \
+    -G"MSYS Makefiles" \
```
1480c1, line 43: Can you see if `-GNinja` works as well? Would need to add ninja to the builod deps

mingw-w64-box2d/PKGBUILD
```diff
@@ -34,10 +34,28 @@ build() {
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DBOX2D_BUILD_DOCS=OFF \
+    -DBOX2D_BUILD_TESTBED=OFF \
+    -DBOX2D_BUILD_UNIT_TESTS=OFF \
+    ../${_realname}-${pkgver}/
+
+  make
```
1480c1, line 52: could also switch to `cmake --build .`

mingw-w64-box2d/PKGBUILD
fundies: I'm not on windows, the recent update @podsvirov pushed to update removed the static lib and broke our CI. I haven't tested this and I just grabbed the artifacts from the CI to confirm the static lib was present again.

podsvirov: In the build log, in the diff section, you can see that some  `*.js` files from the documentation are missing.

Installing a static build after a shared build overwrites the shared CMake target - it makes it impossible to use the shared library in CMake projects.

lazka:
I went ahead and made the suggested changes. Can anyone confirm that this works?

mingw-w64-box2d/PKGBUILD
fundies: I could care less about the docs. They're online and don't effect builds. Please don't break my builds by randomly removing static libs to update. Every time I blink someone here breaks a package from incorrectly doing an update and it's getting really tiresome. The shared target is not overwritten as both libox2d.dll.a and libbox2d.a are now there.

fundies:
> I went ahead and made the suggested changes. Can anyone confirm that this works?
Both both libox2d.dll.a and libbox2d.a are now there in the artifact. That's good enough for me.

PR #9444, opened 2021-08-22
update nextpnr, iverilog, yosys and scopehal-apps

umarcor:
A regular bump to pick latest enhancements.

Note that the failure of scopehal-apps on MINGW32 is expected.


lazka:
The scopehal build seems to fail for ucrt64 here: https://github.com/azonenberg/scopehal-apps/blob/41c1fbad9192fcd9aadeffe761c3202171f2ed0a/src/glscopeclient/CMakeLists.txt#L156

Is ucrt not supported?

lazka:
I've dropped the ucrt64 build for now

Biswa96:
I shall send some patches to upstream. Let see how it goes.

PR #9442, opened 2021-08-22
ngspice: update to 35

umarcor:
A new release was published.

/cc @holvo

PR #9447, opened 2021-08-23
ngspice: Install shared libraries with make

Biswa96:
@umarcor As discussed here https://github.com/msys2/MINGW-packages/pull/9442


umarcor (approved):
LGTM!

PR #9449, opened 2021-08-23
qwt: Fix qwt.prf file.

### LINKED ISSUE #9448: Config file broke in qwt-qt5 renaming
When the old "qwt" was renamed to qwt-qt5, one library renaming spot was missed:
This is in mingw64/share/qt5/mkspecs/features/qwt.prf:
Last line needs to be:
`qwtAddLibrary($${QWT_INSTALL_LIBS}, qwt-qt5)`

Consequence:
`CONFIG += qwt` no longer works from *.pro files (because it passes an incorrect library name to linker).

How should this be fixed?

1) Should I sumit a PR with a modified `005-rename-qwt-to-remove-qt5-qt6-conflict.patch` that also does this rename?
2) Should I submit a PR with an additional patch file where this step is done?
3) Something else (?)

MehdiChinoune:
Fixes #9448

### DIFF
--- a/mingw-w64-qwt/PKGBUILD
+++ b/mingw-w64-qwt/PKGBUILD
@@ -43,6 +43,32 @@ prepare() {
   patch -p1 -i ${srcdir}/005-rename-qwt-to-remove-qt5-qt6-conflict.patch
 }
 
+build() {
+  # qwt-qt5
+  [[ -d ${srcdir}/build-${MSYSTEM}-qt5 ]] && rm -rf ${srcdir}/build-${MSYSTEM}-qt5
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
@radioflash Could you download the package from https://github.com/msys2/MINGW-packages/actions/runs/1159416577 (Artifacts) and see if it solves your problem.

radioflash:
This definitely fixes it:
I manually deleted the critical files, made sure it broke my build, then installed mingw-w64-x86_64-qwt-qt5-6.2.0-5-any.pkg.tar.zst with `pacman -U`, which made the build work again.

Many thanks!

PR #9454, opened 2021-08-24
[new-package] qt5-doc

MehdiChinoune:


Biswa96:
Compiled from single tarball instead of submodule?

MehdiChinoune:
> Compiled from single tarball instead of submodule?
This is the way archinux build it.
https://github.com/archlinux/svntogit-packages/blob/master/qt5-doc/trunk/PKGBUILD

It is compilcated, see:
https://invent.kde.org/qt/qt/qtdoc/-/blob/kde/5.15/README

lazka:
nice!

PR #9092, opened 2021-07-05
rust: fix build on clang32

jeremyd2019:
For #9091.  The patch to psm was accepted upstream and released as 0.1.14.

This doesn't actually fix it, but it gets much further, to stage0 rustbook.
```
Building stage0 tool rustbook (i686-pc-windows-gnu)
  running: "\\\\?\\C:\\_\\mingw-w64-rust\\src\\i686\\build\\i686-pc-windows-gnu\\stage0\\bin\\cargo.exe" "build" "--target" "i686-pc-windows-gnu" "-Zbinary-dep-depinfo" "-j" "2" "--release" "--frozen" "--color" "always" "--manifest-path" "C:\\_\\mingw-w64-rust\\src\\rustc-1.53.0-src\\src/tools/rustbook\\Cargo.toml" "--message-format" "json-render-diagnostics"
```
...
```
  error: failed to run custom build command for `html5ever v0.25.1`
  
  Caused by:
    process didn't exit successfully: `C:\_\mingw-w64-rust\src\i686\build\i686-pc-windows-gnu\stage0-bootstrap-tools\release\build\html5ever-7b859b6107fbb733\build-script-build` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
    --- stdout
    cargo:rerun-if-changed=C:\_\mingw-w64-rust\src\rustc-1.53.0-src\vendor\html5ever\src/tree_builder/rules.rs
  
    --- stderr
    fatal runtime error: failed to initiate panic, error 5
  warning: build failed, waiting for other jobs to finish...
  error: build failed
  command did not execute successfully: "\\\\?\\C:\\_\\mingw-w64-rust\\src\\i686\\build\\i686-pc-windows-gnu\\stage0\\bin\\cargo.exe" "build" "--target" "i686-pc-windows-gnu" "-Zbinary-dep-depinfo" "-j" "2" "--release" "--frozen" "--color" "always" "--manifest-path" "C:\\_\\mingw-w64-rust\\src\\rustc-1.53.0-src\\src/tools/rustbook\\Cargo.toml" "--message-format" "json-render-diagnostics"
  expected success, got: exit code: 101
  Traceback (most recent call last):
    File "../rustc-1.53.0-src/x.py", line 27, in <module>
      bootstrap.main()
    File "C:/_/mingw-w64-rust/src/rustc-1.53.0-src/src/bootstrap/bootstrap.py", line 1189, in main
      bootstrap(help_triggered)
[...]

Issue #9091, opened 2021-07-04
rust: fails on clang32


jeremyd2019:
```
    = note: lld-link: error: undefined symbol: @rust_psm_on_stack@16
            >>> referenced by libpsm-a103c78323d6f2a6.rlib(x86_windows_gnu.o)
```

Oddly, 
https://github.com/rust-lang/stacker/blob/master/psm/src/arch/x86_windows_gnu.s#L63-L69
> ```s
> .def @rust_psm_on_stack@16
> .scl 2
> .type 32
> .endef
> .globl @rust_psm_on_stack@16
> .p2align 4
> @rust_psm_on_stack@20:
> ```

What's up with that `@20` ?  Maybe that's the cause?

jeremyd2019:
I tried patching this, but the error still occurs.

jeremyd2019:
very strange, the .o and .a still have the `@20` symbol

jeremyd2019:
@mati865 ?  I don't even know where 'upstream' is for that file to report or send a pull request, nor do I know why patching the `@20` out of the source file and I still end up having binaries with `@20` in them.

jeremyd2019:
I've discovered `--enable-vendor` - might have a chance then patching the file in vendor/.  Had to update cc from 1.0.65 to 1.0.67 cause the vendor stuff only had 1.0.60 or 1.0.67.

jeremyd2019:
rust-lang/stacker#54

jeremyd2019:
This seems to be working.  Will try to make a PR (draft cause it's pretty clear that rust doesn't want you to patch their vendored code)

jeremyd2019:
Eventually got this error:
```
error: failed to run custom build command for `html5ever v0.25.1`

Caused by:
  process didn't exit successfully: `T:\mingw-w64-rust\src\i686\build\i686-pc-windows-gnu\stage0-bootstrap-tools\release\build\html5ever-7b859b6107fbb733\build-script-build` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
  --- stdout
  cargo:rerun-if-changed=T:\mingw-w64-rust\src\rustc-1.53.0-src\vendor\html5ever\src/tree_builder/rules.rs

  --- stderr
  fatal runtime error: failed to initiate panic, error 5
```

jeremyd2019:
> [rust-lang/stacker#54](https://github.com/rust-lang/stacker/pull/54)

[...]

PR #9458, opened 2021-08-25
librsvg2.40: rust-free version

jeremyd2019:
Would much rather have a working rust for clang32 (and eventually clangarm64), but if that's not possible, package the last C version of librsvg to unblock things.  Had to resurrect libcroco for this too ☹️ .

#9091, #9092, #9439...


lazka:
Can we limit this to clang32/clangarm64 ?

PR #9467, opened 2021-08-26
curl: Update to 7.78.0

lazka:


zmughal:
`libmetalink` should also be removed for `mingw-w64-wget` <https://github.com/msys2/MINGW-packages/blob/e66953810186067f8059cbf431b4521fc0d42553/mingw-w64-wget/PKGBUILD#L55>.

Currently seeing:
```sh
# ldd /mingw64/bin/wget | grep metalink
        libmetalink-3.dll => not found
```
which breaks wget.

PR #9491, opened 2021-08-29
xmake: 2.5.7

waruqi:


Biswa96:
If you see the [diff] section in CI log here you can see some files are removed after new version. Is that alright?

waruqi:
> If you see the [diff] section in CI log here you can see some files are removed after new version. Is that alright?

what diif? I do not found it.

Biswa96:
See this https://github.com/msys2/MINGW-packages/pull/9491/checks#step:9:660

waruqi:
> See this https://github.com/msys2/MINGW-packages/pull/9491/checks#step:9:660

yeah, I removed them.

PR #9494, opened 2021-08-29
octave: Fix detection of octave binary when installing Octave packages.

mmuetzel:
As reported in #9310, installing an Octave package with `pkg install ...` currently fails.
The proposed changes should be fixing the issue (see `0003-pkg-octave-binary.patch`).

Patching that file of the tarball triggers the build system to re-build the documentation. That doesn't work currently (at least for me, see msys2/msys2-texlive#13).
The additional changes to the build rule and patch `0002-mk-doc-cache-path.patch` should work around that issue.


zklhp:
@mmuetzel Thank you for fixing it. Now I can install packages.

PR #9498, opened 2021-08-30
db: fix clang build

jeremyd2019:
* patch to avoid using name atomic_init
  * from https://github.com/bitcoin/bitcoin/blob/master/depends/patches/bdb/clang_cxx_11.patch
* patch to hide macro "store" from libc++ headers
* hacking around autotools madness


jeremyd2019:
I was originally going to look at updating, but I read that there was a license change after 6.0.19, so thought maybe it was being held at this version intentionally for that reason.

PR #9499, opened 2021-08-30
Add neofetch

mapitman:
Neofetch is a command-line system information tool written in bash 3.2
https://github.com/dylanaraps/neofetch

I originally opened a [pull request](https://github.com/msys2/MSYS2-packages/pull/2385) against msys2/MSYS2-packages, but I don't think that is the correct place for it.


mapitman:
Sounds like the other repo _was_ the correct place.

PR #9506, opened 2021-08-31
blosc2: new package

kmilos:


kmilos:
Just noticed import lib is actually missing, will investigate...

PR #9515, opened 2021-09-01
qt6: update to 6.1.3

MehdiChinoune:


mingw-w64-qt6-3d/PKGBUILD
```diff
@@ -2,46 +2,68 @@
...
+
+package_qt6-3d-debug() {
+  depends=("${MINGW_PACKAGE_PREFIX}-qt6-base-debug"
+           "${MINGW_PACKAGE_PREFIX}-${_realname}")
+  groups=("${MINGW_PACKAGE_PREFIX}-qt6-debug")
+
+  cp -rf ${srcdir}/${MSYSTEM}-debug${MINGW_PREFIX} "${pkgdir}"/
+}
+
+# generate wrappers
```
jeremyd2019, line 63: It looks like these lost the template start comment - without it if the template is ever updated and makepkg-template used to update the copies in the pkgbuild files, these copies won't be updated.

MehdiChinoune: I don't know about that, I have just copied from other files.

jeremyd2019: That comes from https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/mingw-w64-splitpkg-wrappers-1.0.template.  You can just copy it from other files, as long as you get the start and end comments, or you can add `# template input; name=mingw-w64-splitpkg-wrappers; version=1.0;` to the file and run `makepkg-template --template-dir ../mingw-w64-PKGBUILD-templates` to have makepkg-template install it for you (I'd just copy it I think).

PR #9516, opened 2021-09-01
Convert wxWidgets based packages to using single local variable to version of wxWidgets

stahta01:
This change will make it easier to change to wxWidgets version 3.2 when it is released sometime this decade. Or to version 3.1.

I also added 'clang32' as a arch if missing.

Tim S.


stahta01:
Removed package OpenSceneGraph because it failed to build

PR #9521, opened 2021-09-02
qt5-debug: finish move to extra debug packages

lazka:


Alexpux:
I think need to do the same for Qt6

lazka:
We never built debug packages for qt6

PR #9523, opened 2021-09-02
Update KF5 to 5.85.0

jtanx:


lazka:
I'll look into the CI error later

jtanx:
I think there's something wrong with the resolution of the build order (it needs to build extra-cmake-modules before attica). Side-note the script that tries to figure that out seems a bit slow.

Biswa96:
I will be updating some packages tier wise. Let see how it goes. BTW, not only updating version, I was also changing
* MSYS Makefiles to Ninja
* kf5 group to prefixed group
* add pgp key to validate tarball

lazka:
please rebase on master

jtanx:
I'm working on a change that will improve the speed of that dependency resolution, as I think >10mins to resolve that is a bit too slow.

Will see if I can submit something this evening or tomorrow.

lazka:
thanks. can I still merge this in the meantime?

jtanx:
Yep, sure - not sure what's going on with that UCRT64 build though

lazka:
hm, true

jtanx:
I built karchive locally and it seemed to work (although I only ever tried after switching to ninja).

Will see what happens...

lazka:
Unrelated to this PR, but re KF5: Do you know of the static variants of those packages are used?

jtanx:
I'm not too sure myself. 

New mingw32 failure seems like extra-cmake-modules was updated on master but it can't find the package.

jtanx:
I think the build order improvements work, it takes a couple of seconds now instead of >10 minutes to compute, so if the build passes, this PR should be ready.

lazka:
If the ucrt error is a timing issue we can still merge and retry until it works.

jtanx:
Yeah, I'm not too sure what's going on with that. One last shot - since it seems like the failures so far have been with make, I've switched the rest of the packages to use ninja.

.ci/ci-get-build-order.py
```diff
@@ -0,0 +1,93 @@
...
+    results = run(shell, "-c", script, env=env).split("\0")[:-1]
+    assert len(props) == len(results), "Length of props matches results"
+
+    info = {}
+    version_re = re.compile(r"[<>=]")
+    for prop, res in zip(props, results):
+        res = [x for x in res.strip().split() if x]
+        if prop in ("depends", "makedepends"):
+            res = [version_re.split(x)[0] for x in res]
+        info[prop] = res
```
jeremyd2019, line 62: I've just spent some time parsing pacman databases, which use the same version format...

```suggestion
    version_re = re.compile(r"[<>=]")
    for prop, res in zip(props, results):
        res = [x for x in res.strip().split() if x]
        if prop in ("depends", "makedepends", "provides"):
            res = [version_re.split(x)[0] for x in res]
        info[prop] = res
```

Provides can also specify a version.  Theoretically, they all could also end in ": description", but I've not seen anything but optdepends use that.  I found attempting to split that off using a regex is slower than using `.rsplit(": ", 1)`, if you did want to be robust against that.

.ci/ci-get-build-order.py
```diff
@@ -0,0 +1,93 @@
...
+
+    deps = sorted(set(info["depends"]) | set(info["makedepends"]))
+    pkg = PackageInfo(deps, False)
+    for alias in set((package, *info["pkgname"], *info["provides"])):
```
jeremyd2019, line 66: Why iterate over a set instead of just a tuple?

.ci/ci-get-build-order.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def get_pkginfo(package, packageset):
+    props = ["depends", "makedepends", "pkgname", "provides"]
+    script = f'set -e && source "{package}/PKGBUILD"\n'
+    for prop in props:
+        script += f"echo \"${{{prop}[@]}}\" && printf '\\0'\n"
+
+    shell = os.environ.get("SHELL", "bash")
+    env = {"MINGW_PACKAGE_PREFIX": "mingw-w64"}
```
jeremyd2019, line 52: Should probably copy os.environ into a new dict and update with this, passing env to subprocess replaces the default behavior of inheriting env from parent.

jeremyd2019: Heck, you could probably just set it in os.environ, I don't think anything else would care...

.ci/ci-get-build-order.py
jtanx: Yeah, it doesn't need to I guess, just to drop duplicates.

jeremyd2019: ok, I see.  it's just setting items in a dict though, so that will squash duplicates already.  Not sure what's faster though, this, looping over tuple, `packageset.update({alias: pkg for alias in (...)})` or `packageset.update(dict.fromkeys((...), pkg)`.  Time to break out `timeit` I guess :wink:

jeremyd2019: fastest for me (on python 3,8) was just taking set( ) out.

jtanx:
Just looking into the ucrt64 failure  a bit more, I think it's some sort of cmake bug in the automoc process. The command it fails on is:

```
D:/a/_temp/msys/msys64/ucrt64/bin/moc.exe -DKDNSSD_DEPRECATED_WARNINGS_SINCE=0x0 -DKDNSSD_DISABLE_DEPRECATED_BEFORE_AND_AT=0x0 -DKF5DNSSD_EXPORTS -DKF_DEPRECATED_WARNINGS_SINCE=0x060000 -DNOMINMAX -DQT_CORE_LIB -DQT_DEPRECATED_WARNINGS_SINCE=0x060000 -DQT_DISABLE_DEPRECATED_BEFORE=0x050f00 -DQT_NETWORK_LIB -DQT_NO_CAST_FROM_BYTEARRAY -DQT_NO_CAST_TO_ASCII -DQT_NO_DEBUG -DQT_NO_FOREACH -DQT_NO_KEYWORDS -DQT_NO_NARROWING_CONVERSIONS_IN_CONNECT -DQT_NO_URL_CAST_FROM_STRING -DQT_USE_QSTRINGBUILDER -DUNICODE -DWIN32_LEAN_AND_MEAN -DWINVER=0x0600 -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_SCL_SECURE_NO_WARNINGS -D_UNICODE -D_USE_MATH_DEFINES -D_WIN32_IE=0x0600 -D_WIN32_WINNT=0x0600 -IC:/_/mingw-w64-kdnssd-qt5/src/build-x86_64-shared/src -IC:/_/mingw-w64-kdnssd-qt5/src/kdnssd-5.85.0/src -IC:/_/mingw-w64-kdnssd-qt5/src/build-x86_64-shared -ID:/a/_temp/msys/msys64/ucrt64/include/QtNetwork -ID:/a/_temp/msys/msys64/ucrt64/include/QtCore -ID:/a/_temp/msys/msys64/ucrt64/share/qt5/mkspecs/win32-g++ -ID:/a/_temp/msys/msys64/ucrt64/include -ID:/a/_temp/msys/msys64/ucrt64/include/c++/10.3.0 -ID:/a/_temp/msys/msys64/ucrtl64/include/c++/10.3.0/x86_64-w64-mingw32 -ID:/a/_temp/msys/msys64/ucrt64/include/c++/10.3.0/backward -ID:/a/_temp/msys/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/include -ID:/a/_temp/msys/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/include-fixed -ID:/a/_temp/msys/msys64/ucrt64/x86_64-w64-mingw32/include --include C:/_/mingw-w64-kdnssd-qt5/src/build-x86_64-shared/src/KF5DNSSD_autogen/moc_predefs.h --output-dep-file -o C:/_/mingw-w64-kdnssd-qt5/src/build-x86_64-shared/src/KF5DNSSD_autogen/include/moc_domainbrowser.cpp C:/_/mingw-w64-kdnssd-qt5/src/kdnssd-5.85.0/src/domainbrowser.h
```

Comparing this to a local build which works, the difference is here:

```
-IC:/msys64n/ucrtl64/include/c++/10.3.0/x86_64-w64-mingw32
```

Note the `l` in `urct64l` - for some reason that path is being mangled (unless someone knows what ucrt64l is?). Fixing that path makes it work. I suspect somewhere in cmake when it's generating those paths, it's doing something it shouldn't, causing UB which manifests as such.

.ci/ci-get-build-order.py
jtanx: > Provides can also specify a version

What does this format look like?

jtanx: Ah, I see an example on https://wiki.archlinux.org/title/PKGBUILD#provides

jeremyd2019: yeah.  They're all (provides, depends, optdepends, ...) parsed through the same function in libalpm (library behind pacman), effectively rsplit(': ', 1) to get optional description, followed by taking the first part from that split and splitting on the first of '(<=|>=|<|>|=)', first of that is the name and second is the version.

jtanx: I ended up just passing everything through the regex `[<>=:]` to strip out both in one step.

jeremyd2019: That'll work as long as you only care about the name, not in splitting the version from the description (which you don't, so :+1:

jeremyd2019:
hehe, mingw python vs msys bash line endings ftw 😜

jeremyd2019:
I ran my script, and it appears none of the symbols in question are actually imported by anything in the repo.  So I guess I'll make a PR with my patch for clang, probably to be merged after this one so as not to interfere with this.

jeremyd2019:
hmm I guess I take that back, just made patch and built clang64 and mingw64.. compared mingw64 exports to existing package, and only a couple of libstdc++-related things showed up, not anything related to those classes.  So my forthcoming PR is even less concerning than I thought.

jtanx:
mingw32/64 pass, unless a fix is found for that automoc/cmakeissue for ucrt64 I think this pr is as ready as it'll be.

jtanx:
So I've been looking into that ucrt failure, I've narrowed it down a bit more to something in the CMake compiler detection step.

From CMakeOutput.log:

```
2021-09-05T06:34:02.0495390Z Detecting CXX compiler ABI info compiled with the following output:
2021-09-05T06:34:02.0497848Z Change Dir: C:/_/mingw-w64-kdbusaddons-qt5/src/build-x86_64-shared/CMakeFiles/CMakeTmp
2021-09-05T06:34:02.0498529Z 
2021-09-05T06:34:02.0499536Z Run Build Command(s):D:/a/_temp/msys/msys64/ucrt64/bin/ninja.exe cmTC_cc458 && [1/2] Building CXX object CMakeFiles/cmTC_cc458.dir/CMakeCXXCompilerABI.cpp.obj
....

2021-09-05T06:34:02.0531392Z GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
2021-09-05T06:34:02.0531884Z 
2021-09-05T06:34:02.0532334Z ignoring nonexistent directory "/ucrt64/include"
2021-09-05T06:34:02.0532715Z 
2021-09-05T06:34:02.0533332Z GNU assemibgnoring duplicate directory "D:/a/_temp/msys/msys64/ucrt64/x86_64-w64-mingw32/include"
...

021-09-05T06:34:02.0535051Z #include <...> search starts here:
2021-09-05T06:34:02.0535365Z 
2021-09-05T06:34:02.0536022Z  D:/a/_temp/msys/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../include/c++/10.3.0
2021-09-05T06:34:02.0536441Z 
2021-09-05T06:34:02.0537118Z  D:/a/_temp/msys/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../include/c++/10.3.0/x86l_64-w64-mingw32
2021-09-05T06:34:02.0537859Z 
2021-09-05T06:34:02.0538442Z  D:/a/_temp/msys/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../include/c++/10.3.0/backward
```


Inspecting closely, the output is mangled:

* GNU assemibgnoring duplicate
* D:/a/_temp/msys/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../include/c++/10.3.0/**x86l_64-w64-mingw32**

I'm not sure if it's something like stdout and stderr being mixed together.

PR #9527, opened 2021-09-03
Build/package some llvm components separately.

MehdiChinoune:


MehdiChinoune:
requires #9528 to be merged first.

mingw-w64-clang/PKGBUILD
```diff
@@ -351,7 +267,6 @@ build() {
     # TODO: is it safe to do this on clangprefix too?  I remember running into
     # issues trying it (which is why this is in the "else" case)
     libcxx_config+=(-DCMAKE_AR="${srcdir}/build-${MSYSTEM}/bin/llvm-ar.exe"
-      -DCMAKE_ASM_COMPILER="${srcdir}/build-${MSYSTEM}/bin/clang.exe"
```
jeremyd2019, line 354: ?

mingw-w64-clang/PKGBUILD
```diff
@@ -388,8 +299,6 @@ build() {
     -DLLVM_INCLUDE_EXAMPLES=OFF \
     -DLLVM_INSTALL_UTILS=ON \
     -DLLVM_LINK_LLVM_DYLIB=ON \
-    -DLLVM_POLLY_LINK_INTO_TOOLS=OFF \
```
jeremyd2019, line 391: This is a casualty of splitting polly?  I don't know what downside this has, @mati865 ?

mingw-w64-lld/PKGBUILD
```diff
@@ -0,0 +1,140 @@
...
+  [[ -d build-${MSYSTEM} ]] && rm -rf build-${MSYSTEM}
+  mkdir -p build-${MSYSTEM} && cd build-${MSYSTEM}
+
+  if (( _clangprefix )); then
+    export CC='clang'
+    export CXX='clang++'
+  else
+    export CC='gcc'
+    export CXX='g++'
+  fi
```
jeremyd2019, line 114: I think this is already done in makepkg_mingw.conf now...

mingw-w64-openmp/PKGBUILD
```diff
@@ -0,0 +1,90 @@
...
+url="https://openmp.llvm.org/"
+arch=(any)
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+license=("custom:Apache 2.0 with LLVM Exception")
+depends=($( (( _clangprefix )) || echo "${MINGW_PACKAGE_PREFIX}-gcc-libs"))
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-clang"
+             $([[ "$CARCH" == "i686" || "$CARCH" == "x86_64" ]] && echo \
+               "${MINGW_PACKAGE_PREFIX}-uasm"))
```
jeremyd2019, line 22:
```suggestion
             "${MINGW_PACKAGE_PREFIX}-uasm")
```
Since the whole package can be avoided on clangarm64

jeremyd2019 (commented):
All pacakges except openmp should have clangarm64 in mingw_arch

mingw-w64-clang/PKGBUILD
MehdiChinoune: I think the only package that requires an ASM compiler is openmp.

mati865:
I was thinking about splitting it up but others though it'd be too big effort to maintain.

This will break clang subsystems since LLVM ABI changes witch every major version. The only proper way to do it is to version LLVM package like `"${MINGW_PACKAGE_PREFIX}-llvm-12"` and have LLD depend on it. LLDB would be broken as well but it won't matter since it's not necessary for updating. That way we could also split Clang from LLVM.
Alternatively LLD could be built together with LLVM.

Also libc++, libc++abi and libunwind could be split.

mingw-w64-clang/PKGBUILD
mati865: I think polly might be not linked anymore at all.

MehdiChinoune:
> I was thinking about splitting it up but others though it'd be too big effort to maintain.
> 
> This will break clang subsystems since LLVM ABI changes witch every major version. The only proper way to do it is to version LLVM package like `"${MINGW_PACKAGE_PREFIX}-llvm-12"` and have LLD depend on it. LLDB would be broken as well but it won't matter since it's not necessary for updating. That way we could also split Clang from LLVM.
> Alternatively LLD could be built together with LLVM.
> 
> Also libc++, libc++abi and libunwind could be split.

I didn't split the other packages because of the clang subsystems, but I forgot that lld depends on llvm and when llvm is updated it will be broken.
I pushed new changes without lld.
Thanks.

mingw-w64-clang/PKGBUILD
MehdiChinoune: This the only where that option is used:
https://github.com/llvm/llvm-project/blob/main/polly/lib/CMakeLists.txt

PR #9528, opened 2021-09-03
clang: Remove unnecessary pkgrel version dependency.

MehdiChinoune:
They all built from the same PKGBUILD, no need to mention the version.


jeremyd2019:
I think it might still be reasonable to mention the $pkgver, especially once split.  Partial updates are unsupported, but sometimes stuff happens and specifying the versions need to be the same would be another level of protection.

MehdiChinoune:
> I think it might still be reasonable to mention the $pkgver, especially once split. Partial updates are unsupported, but sometimes stuff happens and specifying the versions need to be the same would be another level of protection.

done.

PR #9533, opened 2021-09-04
DocBook and w3-mathml2

JPeterMugaas:
1) All targets have their own  configurations that are separate.
2) Rewrote a comment better.
3) minor code formatting
4) build-$(CARCH) replaced with build-${MSYSTEM}


JPeterMugaas:
Something got out of whack so I have to redo my comets.  Sorry.

Biswa96:
You can force push the commits here.

PR #9537, opened 2021-09-05
Fix ci-get-build-order with some PKGBUILDs

jtanx:
Not all PKGBUILDs source cleanly with `set -e` on (should they?):

```
mingw-w64-blender
mingw-w64-clang
mingw-w64-qt5-static
mingw-w64-rust
mingw-w64-zlib
```

That's because they make use of ternaries like

```
 $([[ "$_variant" == "-shared" ]] && echo \
            $([[ "$_system_assimp" == "yes" ]] && echo "${MINGW_PACKAGE_PREFIX}-assimp") \
            $([[ "$_system_doubleconversion" == "yes" ]] && echo "${MINGW_PACKAGE_PREFIX}-double-conversion") \
...
```

Not sure if anyone has any strong thoughts on whether or not that matters.

### DIFF
--- a/.ci/ci-get-build-order.py
+++ b/.ci/ci-get-build-order.py
@@ -47,7 +47,7 @@ def get_pkginfo(package, packageset):
 
     shell = os.environ.get("SHELL", "bash")
     env = {**os.environ, "MINGW_PACKAGE_PREFIX": "mingw-w64"}
-    results = run(shell, "-ce", script, env=env).split("\0")[:-1]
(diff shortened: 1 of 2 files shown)


jeremyd2019:
I thought of mentioning that when I was reviewing, but I thought they were working despite `set -e`.  In most cases I try to remember to put ` || true` on those, or reverse the logic so I can use `|| echo` rather than `&& echo` to avoid that, but they do work with makepkg...

lazka:
yeah, best to allow what makepkg allows

PR #9540, opened 2021-09-05
CI: track package directory in ci-get-build-order.py

jeremyd2019:
If a package being built depends on a split package, ci-get-build-order would return the split package name, not the original directory name it started with, which may be different.  I don't know if this happens in MINGW-packages, but it was the case with curl/libcurl in MSYS2-packages.

Fixes msys2/MSYS2-packages#2634

/cc @jtanx

### DIFF
--- a/.ci/ci-get-build-order.py
+++ b/.ci/ci-get-build-order.py
@@ -11,6 +11,7 @@ import sys
 
 @dataclass
 class PackageInfo:
+    directory: str
     deps: list
     processed: bool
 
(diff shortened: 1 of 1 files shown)


jeremyd2019:
For review purposes, https://github.com/jeremyd2019/MSYS2-packages/actions/runs/1203717452 is a run in MSYS2-packages, and https://github.com/jeremyd2019/MINGW-packages/actions/runs/1203846486 is a run in MINGW-packages.

PR #9541, opened 2021-09-06
CI: merge with MSYS2-packages

jeremyd2019:
Not much going on over there.


jeremyd2019:
Test run: https://github.com/jeremyd2019/MINGW-packages/actions/runs/1204603984

jeremyd2019:
Maybe a version of the DLL check would be useful here (using ntldd instead of ldd).

PR #9543, opened 2021-09-06
Update Packages

naveen521kk:


naveen521kk:
The CI failure is because `pywin32` is not available. (an issue which exists for a long time). Can I merge this PR @lazka?

jeremyd2019:
It appears that autobuild failed python-build due to a circular dependency between python-build and python-tomli that it didn't happen to break "correctly" (it tried to build python-build first, and failed because there was no python-tomli package)

Issue #9406, opened 2021-08-17
TeXLive install error


Berrysoft:
Output:

```
(1/5) Updating fontconfig cache...
(2/5) Probing GDK-Pixbuf loader modules...
(3/5) Updating TeXLive filename database...
>>> Running mktexlsr
>>> Running mtxrun (this might take a while)
/
(4/5) Updating TeXLive format files...
>>> Running fmtutil (this might take a while)
fmtutil [ERROR]: not building aleph due to missing engine: aleph
fmtutil [ERROR]: running `xetex -ini   -jobname=xetex -progname=xetex -etex xetex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `luajittex -ini   -jobname=luajittex -progname=luajittex luatex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `luahbtex -ini   -jobname=luahbtex -progname=luahbtex luatex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `pdftex -ini   -jobname=pdftex -progname=pdftex -translate-file=cp227.tcx *pdfetex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `euptex -ini   -jobname=euptex -progname=euptex *euptex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `eptex -ini   -jobname=eptex -progname=eptex *eptex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `luajithbtex -ini   -jobname=luajithbtex -progname=luajithbtex luatex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `luatex -ini   -jobname=luatex -progname=luatex luatex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
fmtutil [ERROR]: running `pdftex -ini   -jobname=etex -progname=etex -translate-file=cp227.tcx *etex.ini <nul' return status: 1
fmtutil [ERROR]: returning error due to option --strict
[...]

PR #9544, opened 2021-09-06
texlive: fix issue with installation hooks

naveen521kk:
Fixes https://github.com/msys2/msys2-texlive/issues/14
Fixes https://github.com/msys2/MINGW-packages/issues/9406


naveen521kk:
I am merging this as CI would take a long time to pass.

PR #9545, opened 2021-09-06
CI: don't fail if no packages changed

jeremyd2019:
https://github.com/msys2/MSYS2-packages/pull/2635#discussion_r702816246


jeremyd2019:
Test run with packages: https://github.com/jeremyd2019/MINGW-packages/actions/runs/1207087394

PR #9552, opened 2021-09-08
spice-gtk: Enable clang64 build

heljkon:
clang64 build currently fails with
```
[122/141] "clang" @src/libspice-client-glib-2.0-8.dll.rsp
FAILED: src/libspice-client-glib-2.0-8.dll 
"clang" @src/libspice-client-glib-2.0-8.dll.rsp
lld-link: error: C:/Qemu/tools/hyperv/src/msys2/pkg/mingw-w64-spice-gtk/src/spice-gtk-0.39/src/spice-glib-sym-file: unknown file type
```
I've tried to fix it by renaming spice-glib-sym-file to spice-glib-sym-file.def (also in meson.build), but this led to further errors.

As far as I've understood, -export-symbols is optional, because by default all symbols are exported.
The new patch removes the parts from meson.build, which add the -export-symbols to the linking command. 
I've tested locally created mingw64 and clang64 packages by using produced spicy with qemu. Using remote-viewer with qemu in mingw64 was successful, too (remote-viewer's package virt-viewer is the only package which really depends on spice-gtk).

@jeremyd2019 @Biswa96 @lazka 
Please have a look at my naive attempt, because I am not really sure if that approach is OK.

### DIFF
--- a/mingw-w64-spice-gtk/002-export-symbols.patch
+++ b/mingw-w64-spice-gtk/002-export-symbols.patch
@@ -1,0 +1,63 @@
+diff -Naur spice-gtk-0.39.orig/src/meson.build spice-gtk-0.39/src/meson.build
+--- spice-gtk-0.39.orig/src/meson.build	2020-12-01 17:31:58.770751000 +0100
++++ spice-gtk-0.39/src/meson.build	2021-09-19 14:37:59.316849400 +0200
+@@ -186,9 +186,9 @@
+ spice_gtk_version_script = '-Wl,--version-script=@0@'.format(spice_client_glib_syms_path)
+ spice_gtk_has_version_script = compiler.has_link_argument(spice_gtk_version_script)
+ if not spice_gtk_has_version_script
+-  spice_client_glib_syms = files('spice-glib-sym-file')
+-  spice_client_glib_syms_path = meson.current_source_dir() / 'spice-glib-sym-file'
+-  spice_gtk_version_script = ['-export-symbols', spice_client_glib_syms_path]
++  spice_client_glib_syms = files('spice-glib-sym-file.def')
++  spice_client_glib_syms_path = meson.current_source_dir() / 'spice-glib-sym-file.def'
++  spice_gtk_version_script = ['', spice_client_glib_syms_path]
+ endif
+ 
+ # soversion
+@@ -387,9 +387,9 @@
+   #
+   spice_client_gtk_syms = spice_client_glib_syms
+   if not spice_gtk_has_version_script
+-    spice_client_gtk_syms = files('spice-gtk-sym-file')
+-    spice_client_gtk_syms_path = meson.current_source_dir() / 'spice-gtk-sym-file'
+-    spice_gtk_version_script = ['-export-symbols', spice_client_gtk_syms_path]
++    spice_client_gtk_syms = files('spice-gtk-sym-file.def')
++    spice_client_gtk_syms_path = meson.current_source_dir() / 'spice-gtk-sym-file.def'
++    spice_gtk_version_script = ['', spice_client_gtk_syms_path]
+   endif
+ 
+   # soversion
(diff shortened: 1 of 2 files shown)


Biswa96:
There is a simple workaround. Let me try it.

heljkon:
Everything else may be better...

jeremyd2019:
Those sym-files would be pretty easy to transform into def files, just missing a couple lines of boilerplate at the top:
```
LIBRARY <dll-name>
EXPORTS
```

Where `<dll-name>` is the name of the DLL being built.

Biswa96:
> Those sym-files would be pretty easy to transform into def files

That's what I thought about. But facing some issues:

1. The dlls are with version number. One has to change the version number in def file also, right?
2. If I add those boilerplate on top of that file, this is the error:

```
lld-link: error: mingw-w64-spice-gtk/src/spice-gtk-0.39/src/spice-glib-sym-file: unknown file type
```

3. After doing step 2, if I add .def extension to that file and meson.build file, `-export-symbols` appears as `-e xport-symbols` in command line.

Biswa96:
BTW, is `-export-symbols` documented somewhere? My googlefu failed.

jeremyd2019:
I am not familiar with it as a linker option (and apparently neither is LLD, seeing it as a `-e` short option with a parameter).  I think libtool has something like that though.

jeremyd2019:
> 1. The dlls are with version number. One has to change the version number in def file also, right?

MS says that that name is embedded in the import library as the name of the DLL.  Some testing shows that dlltool will override it with the `-D` option, but `-Wl,--out-implib=` uses the value from the def file not the `-o` dll name.

heljkon:
So this is a minimal change compared to repo version:
Nothing changes for existing packages, because these are able to use "--version-script=map-file"

For Clang I converted *sym-file to *.def, added missing necessary symbols and found a way to provide *.def to ld command line.

@Biswa96, @jeremyd2019 
If you think that's OK, I would rebase and create a real PR

jeremyd2019:
I was thinking about how to generate the def file from the sym file in meson.  I don't know if any platform exists that can use the sym file though (I think it's a libtool thing, that got cargo culted over to meson even though meson doesn't use libtool).  If there's not, switching it to a def file in the tree would make more sense (though it might still be handy for them to use configure_file to sub in the so version).

heljkon:
I tried to use the sym-file instead of "--version-script=map-file". I wasn't able to use it for mingw64, but I guess the "generated" def file would do for mingw64 as well.

I thought copying the sym-file and elavating it to a def file would show the actual modification in a better way. But I don't mind if you think adding complete def files to git would be better. So next version in a few days.

jeremyd2019:
No, I'm just thinking out loud (and mostly about a patch for upstream)

jeremyd2019:
It looks like `vs_module_defs` is the way to specify a def file in meson, rather than via `link_args`.  Also ,there's a meson sample def file that doesn't specify the `LIBRARY`, just `EXPORTS` followed by the symbol list. so that might be better.

heljkon:
Good to talk about. I agree. This needs to be addressed to upstream. But since my meson experience is poor, I would drop an issue only...

jeremyd2019:
Something like this?  If my assumption is correct and the `-export-symbols` is a vestigial libtool option, this should be upstream-able.

heljkon:
To make the patch upstreamable, I would propose to include the conversion of spice-glib-sym-file and spice-gtk-sym-file to spice-glib-sym-file.def and spice-gtk-sym-file.def, because they are not referenced by meson.build anymore.
I don't have an idea if the previous meson build on Linux applied "--version-script=map-file" in every case. As far as I think I understood def files are Windows specific. This leads me to the question, if everything is fine for Linux and other OS?

But for MSYS2 I guess we can leave the PR as is. If you agree, I would squash my 2 commits into one and leave yours as is.

heljkon:
As announced I squashed my first commits. The PR is ready for review

PR #9558, opened 2021-09-09
gtk3: Enable static library build

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/9556


Biswa96:
@jeremyd2019 I think that the template for split package command need to be added here.

jeremyd2019:
Yeah.  This is a more primitive version, that doesn't work right with respect to the per-package variables in all cases (notably `--printsrcinfo` IIRC).

PR #9560, opened 2021-09-09
openlibm: update with upstream patches

jeremyd2019:
I had a productive time with upstream, and got openlibm building for MINGW there, and got them an MSYS2 CI that builds and runs tests successfully.  Sync our package with the results of that.


jeremyd2019:
Oops, noticed when reviewing the diff for the pull request that I lost installing the import library to lib instead of bin...

PR #9568, opened 2021-09-11
numpy: update to 1.21.2

MehdiChinoune:


MehdiChinoune:
What is the purpose of pip-checking the reverse-dependencies of a python package?
@naveen521kk

lazka:
so an update doesn't break reverese deps

PR #9580, opened 2021-09-12
gdal: update to 3.3.2

MehdiChinoune:
It already fails for ucrt64 and clang subsystems.


lazka:
This changes the DLL name, so the following packages need to be rebuild:

```
 mingw-w64-postgis
mingw-w64-OpenSceneGraph
mingw-w64-vtk
```

mingw-w64-vtk/PKGBUILD
```diff
@@ -5,7 +5,7 @@ _realname=vtk
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=8.2.0
-pkgrel=8
+pkgrel=9
```
MehdiChinoune, line 8: Maybe we don't need this. see #9578

lazka: yes, thanks. We still need to be sure things rebuild fine in this PR before merging. I'll rebase the other MR before merging after this is merged.

PR #9586, opened 2021-09-13
[new-package] wget2 2.0.0

MehdiChinoune:
No shared libs on clang subsystems!
@jeremyd2019 any idea?

### DIFF
--- a/mingw-w64-wget2/PKGBUILD
+++ b/mingw-w64-wget2/PKGBUILD
@@ -23,11 +23,25 @@ depends=("${MINGW_PACKAGE_PREFIX}-brotli"
 makedepends=("${MINGW_PACKAGE_PREFIX}-doxygen"
              "${MINGW_PACKAGE_PREFIX}-python")
 checkdepends=("${MINGW_PACKAGE_PREFIX}-libmicrohttpd")
-source=("https://ftp.gnu.org/gnu/wget/${_realname}-${pkgver}.tar.lz"{,.sig})
+source=("https://ftp.gnu.org/gnu/wget/${_realname}-${pkgver}.tar.lz"{,.sig}
+        "https://gitlab.com/gnuwget/wget2/-/raw/v2.0.0/contrib/libtool.patch"
+        "configure-ac-use-srcdir.patch"
+        "mingw-dllexport.patch")
 sha256sums=('da95b1477fa4ede1ac712f179c1354d9624fbc91e778d89cf8d2ae46aa15e3d2'
-            'SKIP')
+            'SKIP'
+            '88d385382cabddb5f12b624af4bd0efbe415bd6ce4d44111de58b7947edac1ed'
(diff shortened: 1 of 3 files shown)


Biswa96:
Add `autoreconf -fiv` in prepare function.

MehdiChinoune:
> Add `autoreconf -fiv` in prepare function.

you suggestion make it fails!

mingw-w64-wget2/PKGBUILD
```diff
@@ -0,0 +1,62 @@
...
+
+build() {
+  [[ -d ${srcdir}/build-${MSYSTEM} ]] && rm -rf ${srcdir}/build-${MSYSTEM}
+  mkdir -p ${srcdir}/build-${MSYSTEM} && cd ${srcdir}/build-${MSYSTEM}
+
+  ../${_realname}-${pkgver}/configure \
+    --prefix=${MINGW_PREFIX} \
+    --with-ssl=openssl \
+    --without-libhsts \
+    --without-libidn
```
kmilos, line 45: Could keep things leaner, no real need for this option: from the script it looks libidn2 is given preference like and this will be disabled anyway if libind2 is found...

jeremyd2019:
> > Add `autoreconf -fiv` in prepare function.
> 
> you suggestion make it fails!

Yes, you  need to autoreconf to get updated libtool files that know how to deal with clang.  Unfortunately, not all projects have autotools files that actually work with current autoreconf.

jeremyd2019:
Oh, I see, something different is going on.  They're trying to patch libtool themselves during configure.  I wonder where the patch file went?

jeremyd2019:
I'm working on this.  Stupid libtool

jeremyd2019:
I got their patch applying to libtool, but still getting `realpath replaced`.  I wish LLD would say which object was replacing `realpath`, and which object the first instance came from, would make it easier to figure out what's going on

jeremyd2019:
Yeah, it looks like they're managing to get a crap-ton of gnulib functions exported from libwget-1.dll (and possibly others, didn't look).  Then when linking wget.exe against libwget.dll.a and gnulib, realpath comes from both, and LLD decides to error out on this.  It appears the same excess exporting is happening with GCC/BFD, but it's not throwing an error in this situation.  It *seems* like the no-whole-archive libtool patch they have might have been intended to work around this, but it doesn't seem to in this case.  What they really should do is explicitly export the symbols they intend to export (libtool has a couple of ways to do this).

jeremyd2019:
OK, I think I see... They have dllimport/export support, but it seems only switched on ifdef `_MSC_VER` rather than `_WIN32`

jeremyd2019:
OK, this builds.  It *should* also be correct for the static libs, as libtool defines `DLL_EXPORT` when building for shared but not when building for static.  I did not test it though, and the `.pc` file does not include any provision for static linking (a consumer would have to define `LIBWGET_STATIC` when compiling, and link in all dependencies themselves).

PR #9601, opened 2021-09-15
URL upgrades/updates/fixups

vszakats:
for the most part:
- to use HTTPS where available (for homepages prefer it even it redirects to a HTTP page)
- autodetected via permanent redirects, https-everywhere, ETags or matching content
- some well-known upgrades, like debian
- prefer HTTP over FTP (in the hope it will become HTTPS)
- upgrade two packages from insecure MD5/SHA1 checksums to SHA256


Biswa96:
Just wondering, would it be better to update the links while updating packages one at a time? Can I take some of the fixes from here?

vszakats:
Though its unlikely the tests will ever pass for this one, the current issue is `:: File /var/cache/pacman/pkg/mingw-w64-x86_64-libtiff-4.3.0-6-any.pkg.tar.zst is corrupted (invalid or corrupted package (checksum)).`, which seems unrelated to this PR. `libtiff` was upgraded to HTTPS, but that source tarball is the exact same as the HTTP one and both match the included SHA256.

vszakats:
@Biswa96: It'd be better of course, but in practice it's unlikely somebody will do it that way.

This update brings these packages near the maximum that can be done at the moment HTTPS-wise. HTTPS adoption plateaued ~around the end of 2020, meaning most websites who wanted to, upgraded to HTTPS by then. So this is probably the last larger patch like this (I've done several similar ones in the past here, and have been doing it for Homebrew for ~5 years.)

I think it'd be useful to merge it and I'm fairly confident that the changes are accurate.

lazka:
Thanks!

vszakats:
@lazka Thank you for merging!

Issue #9602, opened 2021-09-15
pkgconf: prefix workaround no longer works


lazka:
```
$ pkgconf.exe --version
1.8.0
$ pkgconf.exe --cflags poppler-qt6
-I/mingw64/include/poppler/qt6 -I/mingw64/include/poppler

$ pkgconf.exe --version
1.7.4
$ pkgconf.exe --cflags poppler-qt6
-IC:/msys64/mingw64/include/poppler/qt6 -IC:/msys64/mingw64/include/poppler
```

The .pc file is wrong, but pkgconf used to detect this and relocate any path that started with prefix anyway. This worked with 1.7.4 and is broken with 1.8.0

ddennedy:
It seems I experience a related problem where a nightly build started failing 2 days ago with a CMake/pkg-config problem:
```
CMake Error in src/modules/avformat/CMakeLists.txt:
  Imported target "PkgConfig::libavformat" includes non-existent path

    "/d/a/shotcut/shotcut/build/Shotcut/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
```
However, it does exist! I added some logging to a build to ls the directory and print the `libavformat.pc` file. See
https://github.com/mltframework/shotcut/runs/3613783256?check_suite_focus=true#step:4:15123

lazka:
bisected to https://github.com/pkgconf/pkgconf/commit/13fe4c8c589be99b12db62b053c38124afecf2d7

lazka:
reported upstream https://github.com/pkgconf/pkgconf/issues/234

lazka:
revert is in the repo. next step is to maybe figure out what the real fix is, or scan the repo for broken packages and fix their .pc files to not need the workaround.

ddennedy:
Thanks. My build is working again with this updated package with the commit reverted.

MehdiChinoune:
likely fixed!

PR #9604, opened 2021-09-16
pkgconf: revert an upstream commit causing a regression

lazka:
See https://github.com/msys2/MINGW-packages/issues/9602


Alexpux:
Maybe also fix pkg-config files too?
No one know what will be broken again but with correct pkg-config files everything will work

lazka:
sure

PR #9607, opened 2021-09-16
libheif: Generate pkgconfig file with autoconf

Biswa96:
Build pkgconfig file using autoconf because
CMake generates it without ${prefix} variable


novomesk:
Thanks,
I tried the `mingw-w64-x86_64-kimageformats-qt5-5.85.0-2-any.pkg.tar.zst` and  `mingw-w64-x86_64-libheif-1.12.0-5-any.pkg.tar.zst` and they work for me.

PR #9616, opened 2021-09-18
libraw: Fix building on clang subsystems

MehdiChinoune:


MehdiChinoune:
--enable-openmp is required for clang compilers.

PR #9622, opened 2021-09-18
libraw: Reenable openmp after mistakenly removed.

MehdiChinoune:


Biswa96:
Kudos to @kmilos for bringing this up ❤️

kmilos:
Thanks, thought the original patch was already merged upstream, so would've preferred sticking with it (-DLIBRAW_FORCE_OPENMP will be the way to the n future versions anyway).

PR #9628, opened 2021-09-20
pulseaudio: Add package

Biswa96:
Thanks to @1480c1 🙇 without whom this would not be possible. Please review and test the package. If the clang version script workaround is correct I can upstream it.

### DIFF
--- a/mingw-w64-pulseaudio/0001-remove-assert-pid_t-mismatch.patch
+++ b/mingw-w64-pulseaudio/0001-remove-assert-pid_t-mismatch.patch
@@ -1,0 +1,20 @@
+diff --git a/src/pulsecore/macro.h b/src/pulsecore/macro.h
+index ec8d31c..4ec68c8 100644
+--- a/src/pulsecore/macro.h
++++ b/src/pulsecore/macro.h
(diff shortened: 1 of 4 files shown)


1480c1:
Code lgtm, going to test once I wake up

1480c1:
binaries are working

on windows from msys2
```bash
mkdir -p /c/pulse
cat > /c/pulse/config.pa <<'EOF'
load-module module-native-protocol-tcp auth-ip-acl=127.0.0.1;192.168.12.0/24
load-module module-esound-protocol-tcp auth-ip-acl=127.0.0.1;192.168.12.0/24
load-module module-waveout sink_name=output source_name=input record=0
EOF
cd /c/pulse
pulseaudio -F config.pa --exit-idle-time=-1
```

From Manjaro or any other system with pacat

```bash
ffmpeg -i audio.webm -vn -ac 2 -ar 44100 -f s16le - | PULSE_SERVER=tcp:192.168.12.136:4713 pacat
```
the `192.168.12.136` is the Windows machine I'm using and I'm using zerotier to route 192.168.12.0/24

Issue #9632, opened 2021-09-21
mingw-w64-x86_64-libuv 1.42 fails on Windows 7


ustsv:
cmake: not found GetHostNameW in WS2_32.dll
[ uv_os_gethostname on Windows 7 x64 #3260 ](https://github.com/libuv/libuv/issues/3260) Windows 7 RIP?

Biswa96:
The [upstream commit][1], which changes gethostname -> GetHostNameW, does also fix an issue if the name contains non-ascii characters. It also seems to fix an issue in NodeJS. So, which one should be prioritize?

1. Windows 7 support.
Or
2. Reintroduce the non-ascii name issue in libuv.

[1]: https://github.com/libuv/libuv/commit/95f88f47f28b4686a6959377c8d2c31daa40bd64

evpobr:
CMake uses libuv, it is critical issue IMHO.

Biswa96:
OK, I have reverted the upstream commit which added GetHostNameW. Please test the packages from GitHub Actions artifacts in the linked pull request.

evpobr:
Works for me, MinGW64, thank you.

lazka:
> CMake uses libuv, it is critical issue IMHO.

cmake is a developer tool and not something you ship to users, so it's not critical at all to not have win7 support there.

Biswa96:
What if the developer is using Windows 7? 😎

evpobr:
I wouldn't say that. In the case of Node.js it was a case of an incorrect host name, not such a frequent case. But the `libuv ' bug is like a disease, all dependent applications and libraries will stop working in Windows 7, whether the developers want it or not.

stanleyxu2005:
> CMake uses libuv, it is critical issue IMHO.

What's alternative library for CMake to use? I'm getting serious with libuv dropping off support for Windows 7 (and tactically dropping Windows 8)

MehdiChinoune:
> What's alternative library for CMake to use?

There is no alternative.

PR #9640, opened 2021-09-22
libuv: Revert an upstream commit for Win7 support

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/9632


Biswa96:
@mati865 Did I make any mistake?

mati865:
> @mati865 Did I make any mistake?

This PR description omits the regression it does introduce. So it looks like clear improvement.

Biswa96:
Thanks for the review 😊

PR #9649, opened 2021-09-23
libbotan: Update to 2.18.1 and enable the shared build

lazka:


lazka:
clang fails with `lld: error: unknown argument: -soname`

PR #9665, opened 2021-09-25
qt-creator: Update to 5.0.1

### LINKED ISSUE #9664: qt-creator: Missing Devel component for plugins development

podsvirov:
Install `Devel` component.

Closes #9664.

~~NOTES~~: The files `*.dll.a` are installed next to the files `*.dll`. Maybe we need to fix this. (Fixed in 7e0f027)

### DIFF
--- a/mingw-w64-qt-creator/PKGBUILD
+++ b/mingw-w64-qt-creator/PKGBUILD
@@ -5,7 +5,7 @@ _realname=qt-creator
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 __pre=
-_base_ver=5.0.0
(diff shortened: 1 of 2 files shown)


Biswa96:
> The files `*.dll.a` are installed next to the files `*.dll`. Maybe we need to fix this.

Generally, it can be fixed by adding `ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}` in cmake file at `install()` step.

podsvirov:
LGTM. What about to merge?

Biswa96:
The include directory change seems weird. It's like everything is dumped in there. See the CI log.

podsvirov:
This is life :man_shrugging:. And so it was conceived. But everything is neatly placed in one subfolder. I'm not going to fix this now.

PR #9679, opened 2021-09-29
Update GHDL, Yosys and nextpnr

umarcor:
A regular update of some EDA tools.
Yosys 0.10 was released and a different versioning logic is used now (hardcoded instead of using `pkgver()`).

### DIFF
--- a/mingw-w64-ghdl/PKGBUILD
+++ b/mingw-w64-ghdl/PKGBUILD
@@ -1,7 +1,7 @@
 _realname=ghdl
 pkgbase="mingw-w64-${_realname}"
 pkgname='__placeholder__'
-pkgver=1.0.0.r740.g95cbcc921
+pkgver=1.0.0.r804.gf11aa2981
 pkgrel=1
 pkgdesc='GHDL: the open-source analyzer, compiler, simulator and (experimental) synthesizer for VHDL'
 arch=('any')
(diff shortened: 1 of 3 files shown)


Biswa96:
> Yosys 0.10 was released and a different versioning logic is used now

There is no version released in upstream. Also the pkgbuild has two commit hashes.

umarcor:
> There is no version released in upstream.

What do you mean? https://github.com/YosysHQ/yosys/releases

> Also the pkgbuild has two commit hashes.

One of the commits belongs to Yosys, the other one is a dependency (ghdl-yosys-plugin).

Biswa96:
> What do you mean? 

Oops! I saw the ghdl-yosys-plugin repository.

> One of the commits belongs to Yosys, the other one is a dependency (ghdl-yosys-plugin).

As yosys has versioned released, you can use the source tarball from released tag.

umarcor:
> > One of the commits belongs to Yosys, the other one is a dependency (ghdl-yosys-plugin).
> 
> As yosys has versioned released, you can use the source tarball from released tag.

Yes. However, it took them several years to bump from 0.9 to 0.10. Hence, if we used the tarball, we would be likely stuck for several months/years. Hence, we will keep using git. However, instead of picking arbitrary commits, we will use the "Bump version" commits (https://github.com/YosysHQ/yosys/commits/master). In this case, I used https://github.com/YosysHQ/yosys/commit/7a7df9a3b4996b17bb774377483b15de49aa3d9b, which is `YOSYS_VER := 0.10+1`. That's why I used `0.10.r1` in the recipe.

PR #9685, opened 2021-09-30
qt6: update to 6.2.0

MehdiChinoune:


MehdiChinoune:
I have seen that ucrt failure occured randomly, There is no way to reproduce it.

Biswa96:
How about merging this then see which one fails in msys2-autobuild?

Alexpux:
@MehdiChinoune qt 6.2.0 have a lot more modules now

MehdiChinoune:
> @MehdiChinoune qt 6.2.0 have a lot more modules now

What do you see missing?

Alexpux:
@MehdiChinoune sorry, seems added modules are included in other modules mostly

PR #9699, opened 2021-10-02
fftw: enable avx-128-fma/avx2/avx512

MehdiChinoune:
> you can compile with--enable-avx and the code will still run on a CPU without AVX support

http://fftw.org/fftw3_doc/Installation-on-Unix.html#Installation-on-Unix


MehdiChinoune:
Could someone with commit rights start commiting all of these stuck pull-requests.

PR #9700, opened 2021-10-02
ncurses: Remove pthread dependency

raedrizqie:
Remove pthread dependency


lazka:
can you add an inline comment explaining the reason for your change?

raedrizqie:
nanosleep is only defined in pthread library.

Python modules such as _curses, _curses_panel and readline will pull libwinpthread-1.dll silently because of this.

lazka:
sure, I meant add a comment in the PKBUILD above the change

Biswa96:
Wondering if it is possible to move nanosleep to mingw-w64-crt? @lhmouse

lhmouse:
We can't *move* it, otherwise existent applications that import it from winpthreads DLL will fail to start. If we want it to be available from the CRT anyhow, we must duplicate it.

lazka:
ncurses has a windows specific fallback, so at least for this case I don't see the value.

PR #9702, opened 2021-10-02
libserialport: add initial package

maggu2810:


maggu2810:
I am just not sure about the version to choose.
As you can see here https://sigrok.org/gitweb/?p=libserialport.git;a=summary the last release version tag is 5 years old but there is some acitivity in the master branch.
I used 0.commitdate.commitnumerofdate
I tried to get some information from upstream if a new release is planed... Wait for a reply.

PR #9708, opened 2021-10-04
added SLiM population genetic simulation software package

rdinnager:
This package installs the forward population genetics simulation software known as SLiM (https://messerlab.org/slim/). This package culminates a big effort to get this software ported to Windows from linux / MacOS. This is my first package submission so hopefully I've done it right. All local testing has passed.


rdinnager:
Hi, thanks for your help cleaning this up. SLiM_pacman.tar.gz comes from the developer's website. I forgot to mention before that this package is in collaboration with and with the full permission of the software's main developer @bhaller. It is 3.7 because the developer would like this package to be available with the official 3.7 release so they made this archive available early so that we could test and get the MSYS2 package up. I hope that answers you questions. Cheers!

PR #9710, opened 2021-10-04
libusb: fix dll exports on CLANG

### LINKED ISSUE #9626: aravis: build failure on clang32
I tried building this locally (with `MSYS=winjitdebug`) and saw this

```
---------------------------
Aravis-0.8.exe - Entry Point Not Found
---------------------------
The procedure entry point libusb_alloc_transfer could not be located in the dynamic link library C:\msys64\home\adminuser\MINGW-packages\mingw-w64-aravis\src\build-i686-w64-mingw32\src\libaravis-0.8-0.dll. 
---------------------------
OK   
---------------------------
```

The next thing would be to figure out why it thinks that export is from libaravis and not libusb

### LINKED ISSUE #9706: libusb broken exports
not sure if this only happens with clang but libusb seems to have broken exports.
mingw-w64-virt-viewer is broken because of this as it complains about these.

jeremyd2019:
llvm's dlltool/lld do not support --add-stdcall-alias, but libusb already provided a .def file that said exactly what they wanted to be
exported from the DLL (and in fact regenerated their import lib from it).  Convince libtool to actually use that def file.

Also, remove no longer needed switch of platforms for dlltool.

Fixes #9706, fixes #9626 (which means aravis builds for CLANG32 after this fix)

### DIFF
--- a/mingw-w64-libusb/PKGBUILD
+++ b/mingw-w64-libusb/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libusb
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.0.24
-pkgrel=3
+pkgrel=4
 pkgdesc="Library that provides generic access to USB devices (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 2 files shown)


revelator:
exports are still broken for clang32 so unless you intend to deprecate it for this arch then the patch is still needed.

PR #9716, opened 2021-10-05
qt6-quick3d: update to 6.2.0

MehdiChinoune:
see: https://bugreports.qt.io/browse/QTBUG-96339

qt6-quick3d couldn't be built by cmake >= 3.20 (on non-clang subsystems), I have decided to add a new package mingw-w64-cmake-319 for cmake 3.19.8 just to build qt6-quick3d.


lazka:
Shouldn't us building it under `C:/_` in CI work around the issue by reducing the path lengths?

lazka:
maybe this helps: https://github.com/msys2/setup-msys2/pull/163

PR #9719, opened 2021-10-05
[new-package] gcc11

MehdiChinoune:


stahta01:
> I'm thinking of just updating mingw64 to v11 and keeping mingw32 at v10. Any thoughts on that?

I think just not building Ada with mingw32 might be a better option.

Likely Ada would build under either of the two below 32 bit environments.
Anyone, think creating ucrt32 using SJLJ is a good idea? 
Or, switching mingw32 to SJLJ instead of dwarf?

Tim S.

lazka:
> I think just not building Ada with mingw32 might be a better option.

There is a chance that ada not building is just a symptom, so it might break even more things outside.

32bit is on its way out (the timeline is debatable), I don't think it's time for large experiments there..

mati865:
@MehdiChinoune x86_64 uses SEH and its exception handling is entirely different from i686 which uses Dwarf.

stahta01:
> > I think just not building Ada with mingw32 might be a better option.
> 
> There is a chance that ada not building is just a symptom, so it might break even more things outside.
> 
> 32bit is on its way out (the timeline is debatable), I don't think it's time for large experiments there.
> 
> disclaimer: I have zero experience gcc internals and dwarf vs SJLJ, this is just from the POV as a package maintainer.

Ada does its own exception system that needs to interface with Dwarf.
The problem could be in Ada or Dwarf.

What do you think about the GCC in the MINGW64 environment being an multilib version where the 32 bit libs uses SJLJ based  exception?

Tim S.

PR #9721, opened 2021-10-05
pywin32: update to 301

raedrizqie:


Biswa96:
If any patch file or any hunk in a patch file is from a upstream commit or pull request a comment can be added in pkgbuild file.

PR #9724, opened 2021-10-06
GEGL: Update to 0.4.32

lb90:


lazka:
Should we disable the docs for now? Or is that a sign of some general breakage?

lb90:
Yeah, I think that we can upgrade by disabling docs. Certain operations may cause crashes, but:

* They have to be quite complicated
* Probably 0.4.30 has the same issue, it simply doesn't trigger the crash while building docs

So in the end it's still quite usable. Linux distros upgraded anyway: https://gitlab.gnome.org/GNOME/gegl/-/issues/294#note_1288241

PR #9731, opened 2021-10-07
glibmm: Edit m4 full path reference

Biswa96:


mingw-w64-glibmm/PKGBUILD
```diff
@@ -68,6 +68,9 @@ package() {
 
   DESTDIR="${pkgdir}" ${MINGW_PREFIX}/bin/meson.exe install
 
+  local _BINDIR=$(cygpath -wm /usr/bin)
+  sed -s "s|${_BINDIR}|/usr/bin|g" -i "${pkgdir}${MINGW_PREFIX}"/lib/glibmm-2.4/proc/gmmproc
```
lazka, line 72: Should the shebang of that file also be switched to /usr/bin/perl?

Biswa96: Yup.

Biswa96: Feel free to take it over.

lazka: ok :)

PR #9734, opened 2021-10-08
xpra 4.2.3

totaam:
* bump version
* update minor path changes
* remove unbundled html5 client which now lives here: https://github.com/Xpra-org/xpra-html5


Biswa96:
1. There is a build failure in clang for `-Werror`.
2. You can also add GPG key for further source validation, [like this][1].

[1]: https://github.com/archlinux/svntogit-community/blob/packages/xpra/trunk/PKGBUILD

totaam:
> There is a build failure in clang for -Werror.

Sorry, where?
Just above I could see: `All checks have passed`

> You can also add GPG key for further source validation, like this.

Done.

PR #9745, opened 2021-10-09
Postgresql: Update to 14.0

larskanis:


larskanis:
The `libpq.dll` and others keep backward ABI compatibility. They therefore don't contain a ABI version number in the filename. But `libpq.dll` of PostgreSQL-14 has some new functions (`PQpipelineStatus` and some related). These new functions could be used by dependent libraries, but so far there is none of [the dependencies](https://packages.msys2.org/package/mingw-w64-x86_64-postgresql?repo=mingw64) that makes use of it.

lazka:
ok, thanks

lazka:
Please bump the pkgrel of Postgis in this PR then.

larskanis:
`pkgrel` bump is added and CI build of both packages look good to me.

lazka:
thanks

PR #9746, opened 2021-10-09
libgit2: add variant that uses WinHTTP

carrotIndustries:
This has a couple of advantages over OpenSSL:
 - one less dependency
 - uses the windows certificate store, no need to include a certificate bundle


1480c1:
perhaps it might become simpler to copy and paste between the two variants if you had a `-openssl` and `-winhttp` variant and make it so the `-openssl` was the one installed by default, but both could fulfil `mingw-w64-x86_64-libgit2`?

PR #9749, opened 2021-10-10
[NFC] Remove "mingw-w64-" from packages folder names.

MehdiChinoune:


lazka:
Currently the directory name matches pkgbase. Wouldn't that be more confusing?

MehdiChinoune:
> Currently the directory name matches pkgbase. Wouldn't that be more confusing?

Confusing to whom?

jeremyd2019:
> Currently the directory name matches pkgbase.

I wish.  But there are some exceptions (don't remember off the top of my head what they are).
```python
        if not os.path.exists(os.path.join(base, "PKGBUILD")):
            with urlopen("https://packages.msys2.org/api/search?query=%s&qtype=pkg" % base) as f:
                x = json.load(f)
            base = x["results"]["exact"]["source_url"].rsplit("/", 1)[1]
        if not os.path.exists(os.path.join(base, "PKGBUILD")):
            print('%s PKGBUILD not found!' % (base,), file=sys.stderr)
            continue
```

PR #9759, opened 2021-10-12
pywin32: Update to b302

raedrizqie:
-Disable clang build
-Disable adsi module coz adsiid stub is incomplete
-Link to static gcc, stdc++ and winpthread to avoid dll import error issue


Alexpux:
Python itself with its other modules linked with shared GCC line and winpthreads, so I don't see benefit of linking this module statically

PR #9780, opened 2021-10-16
precice: bump pkgrel

raedrizqie:
- disable clang (for now)
- petsc don't have clang build because it depends on fortran lib, so we have to disable clang for now


Biswa96:
Instead of disabling clang build, is it possible to disable Fortran dependency in this package?

PR #9782, opened 2021-10-16
flang: new package split from clang

jeremyd2019:
flang takes too much memory to build with default jobs, so split it off into its own package, with logic to reduce parallel jobs based on memory.

It fails to build on 64-bit GCC, but succeeds for CLANG64, CLANG32, and MINGW32.  I don't know if there's some workaround possible for GCC, or if mingw_arch should just not include GCC subsystems.

Also, I've never tried to do anything with the resulting binaries, only getting them to build.

### DIFF
--- a/mingw-w64-flang/0001-cast-cxx11-narrowing.patch
+++ b/mingw-w64-flang/0001-cast-cxx11-narrowing.patch
@@ -1,0 +1,42 @@
+--- a/runtime/misc-intrinsic.cpp	2021-10-02 17:25:54.576794400 -0700
++++ b/runtime/misc-intrinsic.cpp	2021-10-02 17:27:11.031517900 -0700
+@@ -47,7 +47,7 @@
+         "TRANSFER: could not allocate memory for result; STAT=%d", stat);
+   }
+   char *to{result.OffsetElement<char>()};
+-  std::size_t resultBytes{size * elementBytes};
++  std::size_t resultBytes{static_cast<size_t>(size * elementBytes)};
+   const std::size_t sourceElementBytes{source.ElementBytes()};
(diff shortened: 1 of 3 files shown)


jeremyd2019:
> Since flang depends on clang, why not building with it on all environments.

Good idea - worth a try at least

jeremyd2019:
Well it builds for all subsystems now.  Anyone know how to try to use it for something?

raedrizqie:
> ```
> $ cat main.f90
> ! main.f90
> program test
> 
> write(*,*) "Hello World"
> 
> end program
> 
> $ flang-new.exe main.f90
> error: code-generation is not available yet
> 
> $ f18 main.f90
> 
> $ ./a.exe
>  Hello World
> ```
> 
> (Works in both MINGW64 and MINGW32)

nice!

raedrizqie:
> > > So, we can't compile older package without first patching it to recognize new flang compiler?
> > > Meaning flang still can't be used right?
> > 
> > 
> > Assuming this comment really had nothing to do with #9791... note the message `error: code-generation is not available yet`. This means that flang cannot yet turn fortran source into machine code. The test that worked merely parses and then "unparses" fortran code, and then passes that "unparsed" fortran code to `gfortran`. Since CLANG64 etc don't _have_ `gfortran`, flang won't really do anything useful there yet, as far as getting fortran packages built.
> 
> There is an unupstreamed fork which is able to build binaries https://github.com/flang-compiler/f18-llvm-project/tree/fir-dev. Note that the new flang is not recognizable by CMake https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6323

Thanks for the info. Glad that they are working on this.

PR #9787, opened 2021-10-17
python2: fix build

raedrizqie:


raedrizqie:
error when generating header..

`
  eval $PYTHON_FOR_BUILD ../../Tools/scripts/h2py.py -i "'(u_long)'" $REGENHEADER
  python.exe ../../Tools/scripts/h2py.py -i '(u_long)' =D:/a/_temp/msys64/clang64/bin/include/stddef.h
    File "../../Tools/scripts/h2py.py", line 181, in <module>
      main()
    File "../../Tools/scripts/h2py.py", line 81, in main
      fp = open(filename, 'r')
  IOError: [Errno 22] invalid mode ('r') or filename: '=D:/a/_temp/msys64/clang64/bin/include/stddef.h'
  make: *** [Makefile:1201: ../Python-2.7.18/Lib/plat-win32] Error 1
      Aborting...`

raedrizqie:
> python 2 is dead, so looking into this isn't worth it imo.

agreed.. but might as well remove it from the repo :)

lazka:
* I've started adding some deps for pidgin master recently, which is gtk3 etc.. not sure how ready that is though
* cegui I don't know..
* gimp still requires py2 etc and they depend on msys2 for that, so we can just skip clangXX in the dependency chain

PR #9792, opened 2021-10-18
petsc: disable clang for now

raedrizqie:


Biswa96:
> but some new packages will not be uploaded because of these blocking packages.

Generally, not. If dependent packages are built in mingw64, mingw32 and ucrt64 then those will be uploaded, no matter if those failed in clang64/32. There are many packages can not be built in clang64/32 as of writing.

raedrizqie:
Ok. Should I cancel this PR?

Biswa96:
Probably yes.

PR #9800, opened 2021-10-18
gdal: fix UCRT build

### LINKED ISSUE #9617: [ucrt] Couldn't build gdal on UCRT64 due to a broken at_quick_exit.

jeremyd2019:
Don't define `__MSVCRT_VERSION__` to `0x0700` - that's the default for MINGW*, and is lower than 0x0E00 for UCRT, turning off some functions in the headers (like quick_exit and at_quick_exit).  Fixes #9617


jeremyd2019:
It builds UCRT64 now, and would build CLANG32/64 if not for #9804

PR #9811, opened 2021-10-19
New package: python-inflection

raedrizqie:


Biswa96:
Please add your name or alias as maintainer. See any other PKGBUILD for example or [this wiki][1] may help.

[1]: https://wiki.archlinux.org/title/Arch_package_guidelines

PR #9817, opened 2021-10-19
New package: python-yapf

raedrizqie:


lazka:
What's your goal with all these python packages?

raedrizqie:
ok.. thanks :)

PR #9830, opened 2021-10-21
pywin32: really fix clang build

raedrizqie:
- remove rsync from makedepends
- fix clang builds
- fix adsi extension
- preparing for jupyter updates

### DIFF
--- a/mingw-w64-python-pywin32/PKGBUILD
+++ b/mingw-w64-python-pywin32/PKGBUILD
@@ -25,7 +25,7 @@ source=(https://github.com/mhammond/${_realname}/archive/b${pkgver}.zip
         010-workaround-broken-add-dll-dir.patch)
 sha256sums=('33547aa989573dc615bf69c9cc589ef8aefe7ec3858009faf1c26d59f73e4a30'
             '9e1e4d28b1fe89ac6e189b46d047d84619f00ba282fa3b85b8ff2ac3ec54963b'
-            '514f435553e54e94f15e01039746e579a4dabf7c4636fc1c6d12792b8e868287'
+            'db43a7fb22f836cac413d7c3c386bbfb470528991369530418dda8c47d3bf03d'
             '4ed4f5a99d9be5ef6869ba50e40cca8f977f053624ec85fee8bcba41889096cf'
             '82f7745c4eef1e7c3262f6de71655967090c981c542a2e3b0544fda1649be68b'
             'a364eb52c9e0098516bb08ca5abb687fff4ee1c2e6d4a53dcbe91e60a6820879'


mingw-w64-python-pywin32/PKGBUILD
```diff
@@ -26,44 +26,50 @@ source=("https://github.com/mhammond/${_realname}/archive/b${pkgver}.zip"
...
             '82f7745c4eef1e7c3262f6de71655967090c981c542a2e3b0544fda1649be68b'
             'a364eb52c9e0098516bb08ca5abb687fff4ee1c2e6d4a53dcbe91e60a6820879'
             'a04e4c8ee86606188c55349c4be755c72a9f03830d6f24828aba6bfa251e030d'
             '0168b0ea893cd252201dbde6baad6cfa4b382bc76ba65657606b6c76ce78262f')
 
+case ${MSYSTEM} in
+  CLANG*)
+    makedepends=(
+      binutils
+      ${MINGW_PACKAGE_PREFIX}-python-setuptools)
```
jeremyd2019, line 41:
```suggestion
    # work around lack of windmc in llvm
    makedepends+=("binutils")
```

mingw-w64-python-pywin32/PKGBUILD
```diff
@@ -3,21 +3,21 @@
...
 pkgdesc='Python for Windows Extensions (mingw-w64)'
 license=('PSF')
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 pkgver=302
-pkgrel=1
+pkgrel=2
+depends=(
+    ${MINGW_PACKAGE_PREFIX}-python)
```
jeremyd2019, line 17:
```suggestion
depends=("${MINGW_PACKAGE_PREFIX}-python")
```

mingw-w64-python-pywin32/PKGBUILD
```diff
@@ -3,21 +3,21 @@
...
 makedepends=(
-    ${MINGW_PACKAGE_PREFIX}-python
-    ${MINGW_PACKAGE_PREFIX}-python-setuptools
-    rsync)
-source=("https://github.com/mhammond/${_realname}/archive/b${pkgver}.zip"
+    ${MINGW_PACKAGE_PREFIX}-python-setuptools)
```
jeremyd2019, line 19:
```suggestion
makedepends=("${MINGW_PACKAGE_PREFIX}-python-setuptools")
```

mingw-w64-python-pywin32/002-com-extensions-fix.patch
```diff
@@ -334,6 +329,23 @@ diff -Naur pywin32-b302-orig/com/win32comext/adsi/src/PyIDirectorySearch.i pywin
...
+ extern PyObject* PyIADs_getattro(PyObject *ob, PyObject *obname);
+ extern PyObject* PyIADsUser_getattro(PyObject *ob, PyObject *obname);
+ 
++#ifndef _MSC_VER
++const IID IID_IADsSearch        = {0xC69F7780, 0x4008, 0x11D0, 0xB9,0x4C, 0x00,0xC0,0x4F,0xD8,0xD5,0xA8};
++const GUID CLSID_ADsDSOObject   = {0x549365D0, 0xEC26, 0x11CF, 0x83,0x10, 0x00,0xAA,0x00,0xB5,0x05,0xDB};
++const GUID DBPROPSET_ADSISEARCH = {0xCFCFC928, 0x9AA2, 0x11D0, 0xA7,0x9A, 0x00,0xC0,0x4F,0xD8,0xD5,0xA8};
++const GUID DBGUID_LDAPDialect   = {0xEFF65380, 0x9C98, 0x11CF, 0xB9,0x63, 0x00,0xAA,0x00,0x44,0x77,0x3D};
++const GUID LIBID_ADs            = {0x97D25DB0, 0x0363, 0x11CF, 0xAB,0xC4, 0x02,0x60,0x8C,0x9E,0x75,0x53};
++#endif
```
jeremyd2019, line 345: Maybe just wait for the next mingw-w64 update?

raedrizqie: ok to me

mingw-w64-python-pywin32/PKGBUILD
raedrizqie: It won't allow me to apply this patch.

jeremyd2019: I did the ` ```suggestion ` manually because there were some deleted lines in between that were interfering.

mingw-w64-python-pywin32/PKGBUILD
```diff
@@ -3,21 +3,20 @@
...
 pkgdesc='Python for Windows Extensions (mingw-w64)'
 license=('PSF')
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 pkgver=302
-pkgrel=1
+pkgrel=2
+depends=("${MINGW_PACKAGE_PREFIX}-python")
 makedepends=(
```
jeremyd2019, line 17:
```suggestion
makedepends=("${MINGW_PACKAGE_PREFIX}-python-setuptools")
```

mingw-w64-python-pywin32/PKGBUILD
```diff
@@ -3,21 +3,20 @@
...
-    ${MINGW_PACKAGE_PREFIX}-python
-    ${MINGW_PACKAGE_PREFIX}-python-setuptools
-    rsync)
-source=("https://github.com/mhammond/${_realname}/archive/b${pkgver}.zip"
+    ${MINGW_PACKAGE_PREFIX}-python-setuptools)
```
jeremyd2019, line 18:
```suggestion
```

jeremyd2019:
I sent https://sourceforge.net/p/mingw-w64/mailman/message/37371156/ upstream for libntquery.a,  but ifilter also needs to be patched because `__in` SAL annotation isn't defined in mingw-w64 (apparently it conflicts with something in libstdc++).  I changed it it `_In_` (which is also in the SAL header) in my local directory, and ifilter builds with the libntquery.a from that patch.

Bits would be more interesting.  It seems that they are just using an enum defined from bits_2_5.h, while mingw-w64 only goes to bits_2_0.h.  But, just ifdefing that enum out then fails at linking due to no `-lBits`.  In the Windows SDK that contains the GUIDs (IID,CLSID,LIBID) from bits.

It seems that bits*.h in Windows SDK come from IDL files, but currently it doesn't seem that they do in mingw-w64.  Wine has them though: https://github.com/wine-mirror/wine/blob/e909986e6ea5ecd49b2b847f321ad89b2ae4f6f1/include/bits2_5.idl  I don't know if it's just a matter of importing/copying them from wine or what.

I figure that's all for a future PR though, once mingw-w64 update is built I can re-run the checks here to confirm adsiid is working, and then this could be merged (though it may need to wait for mingw-w64 update to hit the repo, don't think there's an explicit dependency  between this and mingw-w64 so the update may not be pulled in by autobuild from its staging area).

jeremyd2019:
> Nice! Those _In, __in, et al can safely be defined to none

Even better, the mingw-w64 header is not broken, like the Vista SDK one apparently was 😛 
```diff
--- pywin32-b302/com/win32comext/ifilter/src/stdafx.h.orig      2021-10-22 17:41
:56.227210500 -0700
+++ pywin32-b302/com/win32comext/ifilter/src/stdafx.h   2021-10-22 17:44:57.1090
84600 -0700
@@ -13,7 +13,9 @@
 #include <Filter.h>
 #include <Filterr.h>

+#ifndef __MINGW32__
 #define MISSING_PROPSTG
+#endif
 #ifdef MISSING_PROPSTG
 // Ack - NTQuery.h is failing with the Vista SDK - pull in what we need
 // Problem is missing propstg.h, and all the work-arounds are uglier than
```

PR #9840, opened 2021-10-22
texlive-bin: fix extractbb.exe and related

naveen521kk:
Fixes https://github.com/msys2/msys2-texlive/issues/18


naveen521kk:
It works tested locally, merging.

PR #9884, opened 2021-10-26
split some llvm components

MehdiChinoune:
lldb, openmp, polly, mlir, libc++, libc++abi and libunwind


jeremyd2019:
Did you decide not to pursue merging clang-analyzer?  I am strongly leaning toward at least undoing my change of moving clang-check and clang-extdef-mapping to clang-analyzer (part of 94c7653) because they are being checked for when using clang's cmake modules.

jeremyd2019:
Other than that, this should wait until clangarm64 13.0.0-2 is uploaded, so that it definitely will use that instead of deleting it and using -1, potentially having the pkgrel dependency cause issues.

MehdiChinoune:
I think we should remove 'clang-analyzer' from `lldb` make-dependencies to make the issue is resolved.

LABPBS:
While you are at it rebuilding this theres this comment by @revelator which may be relevant(?) https://github.com/msys2/MINGW-packages/discussions/7589#discussioncomment-1427915

jeremyd2019:
This is interesting: https://lists.llvm.org/pipermail/llvm-dev/2021-October/153238.html hopefully they'll have some sort of archive to download for it, rather than having to download all of llvm-project just to get the cmake files for the runtimes.

PR #9901, opened 2021-10-27
Update portmidi with more recent portmedia release

cme:
Portmidi is no longer maintained as a standalone package,
but is maintained as part of the larger portmedia release.


lazka:
semi related, while browsing upstream and arch I noticed that upstream is moving to github:

* context: https://github.com/mixxxdj/portmidi/issues/8
* (WIP?) github repo: https://github.com/rbdannenberg/portmidi

cme:
> semi related, while browsing upstream and Arch I noticed that upstream is moving to github:

Oh, fantastic! Great find. I spent a while trying to work out what the current state of any 'official' development might be, but didn't find this. With PortAudio already on GitHub now as a separate project, having PortMidi be the same will be great.

PR #9907, opened 2021-10-28
asio: Update to 1.20.0

podsvirov:
There is compilation failure for examples `src/examples/cpp14/parallel_group/` because preprocessor macro `ASIO_HAS_POSIX_STREAM_DESCRIPTOR` defined, but why?...


podsvirov:
Replased by #10009

PR #9918, opened 2021-10-29
mscgen: fix build

raedrizqie:


mingw-w64-mscgen/PKGBUILD
```diff
@@ -1,56 +1,59 @@
...
             '62216cf5e6d36c9787eeadbf51817481a41933dbcda151f630a162f7e45b780d'
+            '89b4b9af5b2a8c383980c73c11a75007c912379d0e4ead11d93c58270eaf71ab'
             '1f81fcbb6e644003c84168f7f78de751edfc6572c71591eea4e80807c643c096')
 
 prepare() {
   cd $srcdir/${_realname}-${pkgver}
-  patch -p2 -i ${srcdir}/0.20-default-font.patch
   patch -p1 -i ${srcdir}/0.20-escape.patch
   patch -p1 -i ${srcdir}/0.20-language.patch
+  patch -p1 -i ${srcdir}/0.20-mingw-fix.patch
```
lazka, line 31: could you add a comment on what this fixes and how?

raedrizqie: it conflicts with `getenv_s` in stdlib_s.h

this is the easiest way i think :p

lazka: ah, you renamed it.. please add a comment

mati865: Putting it inside ifdef would be equally easy assuming the declaration is the same.

mingw-w64-mscgen/0.20-mingw-fix.patch
```diff
@@ -0,0 +1,66 @@
...
+ #else
+diff -urN mscgen-0.20-orig/src/main.c mscgen-0.20/src/main.c
+--- mscgen-0.20-orig/src/main.c	2011-03-05 18:51:50 +0800
++++ mscgen-0.20/src/main.c	2021-10-30 00:35:15 +0800
+@@ -1702,7 +1702,7 @@
+     }
+ 
+     /* Parse input, either from a file, or stdin */
+-    if(gInputFilePresent && !strcmp(gInputFile, "-") == 0)
++    if(gInputFilePresent && strcmp(gInputFile, "-") != 0)
```
lazka, line 39: why this change? that shouldn't make a difference

mingw-w64-mscgen/PKGBUILD
raedrizqie: they are not the same

mingw-w64-mscgen/0.20-mingw-fix.patch
raedrizqie: yup.. but clang is very fussy about it.. and the warning is treated as error IIRC

lazka: it's just a warning here: ` warning: logical not is only applied to the left hand side of this comparison [-Wlogical-not-parentheses]`

raedrizqie: ah ok.. will update it soonish..

mingw-w64-mscgen/PKGBUILD
lazka: for completeness, here is the error message: `error: conflicting types for 'getenv_s'`

PR #9931, opened 2021-10-30
graphicsmagick: Use library as relocation reference.

mmuetzel:
When linking a binary to graphicsmagick, it fails to execute correctly unless it is installed in the mingw bin directory.

Steps to reproduce:
[magick.zip](https://github.com/msys2/MINGW-packages/files/7447452/magick.zip)
Compile the example with:
```
g++ -I/mingw64/include/GraphicsMagick -L/mingw64/lib -omagick.exe magick.cc -lGraphicsMagick++ -lGraphicsMagick
```

Expected result:
Executing the program should succeed in any directory.

Actual result:
If the executable is located in any directory other than `/mingw64/bin` it fails with the following error:
```
$ ./magick.exe
terminate called after throwing an instance of 'Magick::ErrorOption'
  what():  Magick: Coder not found (GIF) reported by ../GraphicsMagick-1.3.36/Magick++/lib/Exception.cpp:435 (throwExceptionExplicit)
Magick: abort due to signal 22 (SIGABRT) "Abort"...
```

That is probably caused by the fact that the relocation of the module and filter directories is done relative to the executable.
It would probably be reasonable to do the relocation relative to the graphicsmagick library.

The proposed changes assume that library, modules, and filters reside in paths that are relatively fixed to each other. That allows executables in arbitrary locations to link to the library.

The change causes a divergence between the `pathtools.c` used by graphicsmagick and the one in `mingw-w64-pathtools`.
Is this a showstopper?


lazka:
hm, the goal was to keep those files in sync...

mmuetzel:
That's what I suspected.

Would it be ok to add a new function `single_path_relocation_lib` (or another name you suggest) that would be very similar to `single_path_relocation` but would use the `dll_path` instead of the `exe_path` for relocation?
That function could be added to the upstream `pathtools.c` so they can stay in sync.

lazka:
Sounds good

mmuetzel:
For "symmetry" reasons, I figured there should be `get_relocated_path_list_lib`, `single_path_relocation_lib`, and `pathlist_relocation_lib`. Is that ok?

Hope I didn't miss one of the copies of the pathtools.

Is it ok to not bump `pkgrel` for these packages? After all, nothing should have changed for them. (Apart from graphicsmagick.)

mmuetzel:
I thought I could sneak these changes into the other packages without triggering an immediate rebuild. But it looks like this is not how it works.
Tried again bumping `pkgrel` of all affected packages this time.

mmuetzel:
The checks of the latest attempt didn't pass. Tbh, I don't know how to interpret its output.
Any insight much appreciated.

mmuetzel:
The last check failed with:
```
  /mingw64/bin/import.exe exists in both 'mingw-w64-x86_64-imagemagick' and 'mingw-w64-x86_64-libosmscout'
```

I don't know how this could be caused by the changes in this PR. I'd guess that conflict already existed before.
Should/can these two packages be marked as conflicting with each other?

mmuetzel:
Is there anything I could do to help get this merged?

My main motivation for this change is to be able to run Octave from its build tree without the need to install it again and again when testing changes. But I'd guess any software that links to graphicsmagick that isn't installed in the bin directory would be affected.

mmuetzel:
Would this PR have a higher chance of getting merged if I limited its scope to graphicsmagick and pathtools and ignored the other pathtools dependers for now?

lazka:
No, it's fine. I just have to find some time/motivation to look into it

mmuetzel:
I don't understand the check error:
```
Pip Output: mingw-w64-x86_64-p11-kit-0.24.0-2-any.pkg.tar.zst
  File "D:\a\MINGW-packages\MINGW-packages\.ci\ci-check.py", line 248, in <module>
    main()
  File "D:\a\MINGW-packages\MINGW-packages\.ci\ci-check.py", line 199, in main
    run_pip_check(pkgname)
  File "D:\a\MINGW-packages\MINGW-packages\.ci\ci-check.py", line 156, in run_pip_check
    p = subprocess.run(
  File "C:/msys64/mingw64/lib/python3.9/subprocess.py", line 505, in run
  File "C:/msys64/mingw64/lib/python3.9/subprocess.py", line 951, in __init__
  File "C:/msys64/mingw64/lib/python3.9/subprocess.py", line 1420, in _execute_child
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

What does it mean?

mmuetzel:
Thanks for clarifying.
I don't understand how the changes for this PR would cause any less .dll or .pyd files to be installed though...

lazka:
It's likely just exposing an already broken package (or missing dependency)

mmuetzel:
Re-building ffmpeg on CLANG32 failed with:
```
 In file included from C:/_/mingw-w64-ffmpeg/src/ffmpeg-4.4.1/libavcodec/amfenc.c:35:
  In file included from C:/_/mingw-w64-ffmpeg/src/ffmpeg-4.4.1/libavcodec/amfenc.h:22:
  In file included from D:/M/msys64/clang32/include/AMF/core/Factory.h:37:
  D:/M/msys64/clang32/include/AMF/core/Platform.h:76:20: error: typedef redefinition with different types ('int' vs 'LONG' (aka 'long'))
  typedef signed int HRESULT;
                     ^
  D:/M/msys64/clang32/i686-w64-mingw32/include/winnt.h:419:16: note: previous definition is here
  CC	libavcodec/amfenc_h264.o
    typedef LONG HRESULT;
                 ^
  1 error generated.
  make: *** [/C/_/mingw-w64-ffmpeg/src/ffmpeg-4.4.1/ffbuild/common.mak:67: libavcodec/amfenc.o] Error 1
```

Doesn't look related to this change though.

mmuetzel:
I understand that this isn't very high priority and time for review is limited. But to give some background why I'd like to have this merged:
In the CI for Octave on Windows (MSYS2), we currently need to install Octave before we can start running the test suite.
On the one hand that adds about 3-5 minutes to the total time for the CI run. On the other hand, that limits to the tests that can be run when Octave is already installed. Some tests require that the test suite is run in the build tree (which currently doesn't work due to the relocation issue with graphicsmagick).

mmuetzel:
Some packages - like `firebird2-git` - never built with the toolchain in the `CLANG64` environment. This is a fatal error in the CI now.
Should `'clang64'` be removed from `mingw_arch` for those packages so that the checks can run for packages that would be compiled later?

raedrizqie:
> Some packages - like `firebird2-git` - never built with the toolchain in the `CLANG64` environment. This is a fatal error in the CI now. Should `'clang64'` be removed from `mingw_arch` for those packages so that the checks can run for packages that would be compiled later?

i got firebird4 to compile on clang, but it always fail on *.msg / *.fdb generation..

mmuetzel:
I think I got an idea why the checks are failing for this PR:
The last check that succeeds is for `mingw-w64-x86_64-openssl`. After that check, a list of packages are uninstalled. That list includes `'mingw-w64-x86_64-python'`. Since the  `pacman` command for uninstalling includes `-cns`, all packages that depend on the removed ones are also uninstalled: https://github.com/msys2/MINGW-packages/blob/master/.ci/ci-check.py#L110
That includes `mingw-w64-x86_64-python-pip`.

In the next check, the script fails when it tries to invoke `pip check`.

Should `mingw-w64-x86_64-python-pip` be included in the install step for each checked package?

mmuetzel:
> i got firebird4 to compile on clang, but it always fail on *.msg / *.fdb generation..

@raedrizqie: Not sure if this is relevant to that crash: It looks like you dropped patch `0024-Relocate-paths...`. That's probably still needed.
Anyway, that is getting quite off-topic for this PR. Should that topic be moved to a WIP PR?

raedrizqie:
> > i got firebird4 to compile on clang, but it always fail on *.msg / *.fdb generation..
> 
> @raedrizqie: Not sure if this is relevant to that crash: It looks like you dropped patch `0024-Relocate-paths...`. That's probably still needed. Anyway, that is getting quite off-topic for this PR. Should that topic be moved to a WIP PR?

yeah.. i was just trying to compile without that patch and see if it works.. maybe your updated patch should be used.. anyway, sorry for this off-topic chat :)

lazka:
sorry for the delay

PR #9932, opened 2021-10-30
add: openblas int64 version

3rav:


lazka:
It looks like all major files have a 64bit suffix with this, so I wonder if it is parallel installable?

lazka:
hm, so from what I can see this just builds the same thing with libraries and functions suffixed? Why is this needed?

mati865:
Since this has
```sh
replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-git")
```
Why don't just update base package?

mmuetzel:
IIUC, the main difference from this build rule to the "base" build rule of OpenBLAS is that it is build with `-DBINARY=64 -DINTERFACE64=1` for the x86_64 target.
That means that the Fortran integers that are used to index matrices will be 64 bits wide instead of 32 bits. That would allow to use the OpenBLAS linear algebra functions with even larger matrices.
While this might be interesting for niche applications, not many distributions are building with these options and they are probably not very well tested.
More importantly, many of the linear algebra rely on having the same Fortran indexing size to interface correctly. That means that if OpenBLAS is switched to using 64bit Fortran integers, also other libraries (like LAPACK, BLAS, ARPACK, SuiteSparse, QRUPDATE, SUNDIALS, ...) need to make that transition at the same time.
Additionally, not all projects that are using OpenBLAS functions (or fwiw functions from the other affected libraries) might be ready to work with 64bit Fortran integers.

My two cents: If you consider providing OpenBLAS with 64bit Fortran integers, please don't replace the "base" OpenBLAS with that option. At least, not for the time being...

lazka:
@3rav any context/information on what you need this for/what your goal is appreciated :)

3rav:
I need CaluliX for large fem problems (>0.5 milion nodes) for this I need arpack and pastix solver witch (-DPASTIX_INT64=ON), pastix requires 64 int: openblas, scotch

https://github.com/msys2/MINGW-packages/discussions/9757

mati865:
My point is this package replaces base package so nobody would be able to install normal openblas.

mmuetzel:
I agree that it should be possible for the user to chose which "flavor" of OpenBLAS to install. The "base package" should still use 32-bit indexing.
Ideally, it would be possible to install both flavors at the same time and carefully set compilation flags in user projects to chose which "flavor" is used.
Afaict, that would already be possible apart from the files installed to the `licenses` folder.

Could the entire "`_realname`" logic just be removed from this package? Would that avoid the issue @mati865 is pointing out?
I'm still pretty new to the MSYS2 ecosystem. So, please correct me if that is wrong.

Imho, a "better" description for the new package would probably be: 
```
pkgdesc="An optimized BLAS library based on GotoBLAS2 1.13 BSD, providing optimized blas, lapack, and cblas with 64-bit indexing (mingw-w64)"
```

lazka:
ok, so this needs to be merged into the openblas PKGBUILD to avoid all those duplication. The question is if the openblas package then just ships both variants or if we create a split package.

okhlybov:
A sudden change to 64-bit indices is a risky move as it affects the majority of the math packages - linear solvers, partitioners etc. not to mention the MS-MPI which can't be recompiled and thus will probably stay with 32-bit indices.
In addition, this will force the diverge with still supported i686 architecture where wide indices make no sense at all.

I'd introduce a new experimental lineup of the 64-bit clean packages with a distinctive suffix while leaving the existing packages intact.

lazka:
I meant building it twice in the same PKGBUILD, since the resulting files don't conflict. Or will that still lead to problems?

okhlybov:
You mean having two versions of the binaries within a single package?
This will explode the package sizes (for no reason?).
Also I'm afraid this will make two versions of SDK files (headers, at least) as well as the int width is often hardcoded in configuration.

lazka:
As you can see here, the files are named differently:

```
2021-10-30T21:08:06.0312529Z +/mingw64/
2021-10-30T21:08:06.0314544Z +/mingw64/bin/
2021-10-30T21:08:06.0321684Z +/mingw64/bin/libopenblas_64.dll
2021-10-30T21:08:06.0327806Z +/mingw64/include/
2021-10-30T21:08:06.0329297Z +/mingw64/include/openblas64/
2021-10-30T21:08:06.0329956Z +/mingw64/include/openblas64/cblas.h
2021-10-30T21:08:06.0330558Z +/mingw64/include/openblas64/f77blas.h
2021-10-30T21:08:06.0331174Z +/mingw64/include/openblas64/lapack.h
2021-10-30T21:08:06.0332826Z +/mingw64/include/openblas64/lapacke.h
2021-10-30T21:08:06.0333491Z +/mingw64/include/openblas64/lapacke_config.h
2021-10-30T21:08:06.0335195Z +/mingw64/include/openblas64/lapacke_example_aux.h
2021-10-30T21:08:06.0336177Z +/mingw64/include/openblas64/lapacke_mangling.h
2021-10-30T21:08:06.0336814Z +/mingw64/include/openblas64/lapacke_utils.h
2021-10-30T21:08:06.0337434Z +/mingw64/include/openblas64/openblas_config.h
2021-10-30T21:08:06.0338026Z +/mingw64/include/openblas64/openblas64/
2021-10-30T21:08:06.0338694Z +/mingw64/include/openblas64/openblas64/lapacke_mangling.h
2021-10-30T21:08:06.0339248Z +/mingw64/lib/
2021-10-30T21:08:06.0339691Z +/mingw64/lib/libopenblas_64.a
2021-10-30T21:08:06.0340205Z +/mingw64/lib/libopenblas_64.dll.a
2021-10-30T21:08:06.0340729Z +/mingw64/lib/pkgconfig/
2021-10-30T21:08:06.0341728Z +/mingw64/lib/pkgconfig/openblas64.pc
2021-10-30T21:08:06.0342289Z +/mingw64/share/
2021-10-30T21:08:06.0342739Z +/mingw64/share/cmake/
2021-10-30T21:08:06.0343241Z +/mingw64/share/cmake/OpenBLAS64/
2021-10-30T21:08:06.0343875Z +/mingw64/share/cmake/OpenBLAS64/OpenBLAS64Config.cmake
2021-10-30T21:08:06.0344712Z +/mingw64/share/cmake/OpenBLAS64/OpenBLAS64ConfigVersion.cmake
2021-10-30T21:08:06.0346767Z +/mingw64/share/cmake/OpenBLAS64/OpenBLAS64Targets.cmake
2021-10-30T21:08:06.0347965Z +/mingw64/share/cmake/OpenBLAS64/OpenBLAS64Targets-release.cmake
2021-10-30T21:08:06.0349063Z +/mingw64/share/licenses/
2021-10-30T21:08:06.0349755Z +/mingw64/share/licenses/OpenBLAS/
2021-10-30T21:08:06.0350327Z +/mingw64/share/licenses/OpenBLAS/LICENSE
2021-10-30T21:08:06.0350955Z +/mingw64/share/licenses/OpenBLAS/LICENSE-lapack
````

as for package size, a split package would reduce that.

**Comment by okhlybov** [comment]
Great. Now we'll be having plain 64 builds alongside the 64-64 builds :)
Jokes aside, wouldn't something like `idx64` suffix be more meaningful to make things more explicit?

**Comment by lazka** [comment]
>  Jokes aside, wouldn't something like `idx64` suffix be more meaningful to make things more explicit?

The naming comes from upstream, wouldn't changing it make problems?

**Comment by okhlybov** [comment]
> The naming comes from upstream, wouldn't changing it make problems?

AFAIK for most packages this suffix is user-defined.

**Comment by mmuetzel** [comment]
Using a split package would probably cover most use cases, imho.

Also using the default naming convention from upstream would probably make it easier for downstream packages to possibly detect the variant they prefer automatically in their build rules...

**Comment by mmuetzel** [comment]
I'm currently working on merging this into a split openblas package with 32-bit and 64-bit indexing, respectively.
How do I update this pull request? Or should I just open a new PR when the changes are ready?

**Comment by mmuetzel** [comment]
I tried to create a split package for OpenBLAS here:
https://github.com/mmuetzel/MINGW-packages/blob/openblas-idx64/mingw-w64-openblas/PKGBUILD

However, packaging for the 32-bit targets (`mingw32` and `clang32`) fails because it doesn't build the 64-bit indexing version in that case.

How should that be handled correctly?

**Comment by jeremyd2019** [comment]
Something like 
```bash
pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
         $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas-64"))
```
?

PR #9933, opened 2021-10-31
openxr: update to 1.0.20

MehdiChinoune:


MehdiChinoune:
I tried to build it from https://github.com/KhronosGroup/OpenXR-SDK-Source (same as arch) but it failed on 32 bits environments.

Biswa96:
Hand over to compiler experts @jeremyd2019

jeremyd2019:
I think both will accept no LIBRARY line at all.

MehdiChinoune:
> I think both will accept no LIBRARY line at all.

Take this package, don't apply the second patch '002-libname-def-file.patch ', and see what it gives you with gcc.

jeremyd2019:
What I'm saying is
```diff
diff --git a/mingw-w64-openxr/002-libname-def-file.patch b/mingw-w64-openxr/002-libname-def-file.patch
index 1886fe932..4a9b4f3a0 100644
--- a/mingw-w64-openxr/002-libname-def-file.patch
+++ b/mingw-w64-openxr/002-libname-def-file.patch
@@ -1,11 +1,10 @@
 --- a/src/loader/openxr-loader.def
 +++ a/src/loader/openxr-loader.def
-@@ -7,7 +7,7 @@
+@@ -7,7 +7,6 @@
  ;
  ;;;;  End Copyright Notice ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

 -LIBRARY
-+LIBRARY libopenxr_loader.dll
  EXPORTS
      xrCreateInstance
      xrDestroyInstance
```

MehdiChinoune:
> What I'm saying is
> 
> ```diff
> diff --git a/mingw-w64-openxr/002-libname-def-file.patch b/mingw-w64-openxr/002-libname-def-file.patch
> index 1886fe932..4a9b4f3a0 100644
> --- a/mingw-w64-openxr/002-libname-def-file.patch
> +++ b/mingw-w64-openxr/002-libname-def-file.patch
> @@ -1,11 +1,10 @@
>  --- a/src/loader/openxr-loader.def
>  +++ a/src/loader/openxr-loader.def
> -@@ -7,7 +7,7 @@
> +@@ -7,7 +7,6 @@
>   ;
>   ;;;;  End Copyright Notice ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
> 
>  -LIBRARY
> -+LIBRARY libopenxr_loader.dll
>   EXPORTS
>       xrCreateInstance
>       xrDestroyInstance
> ```
> 
> That works for both GCC and Clang

I see, but my question was about why does clang accept `LIBRARY` alone while gcc rejects it it is not appended by libname?

jeremyd2019:
🤷

PR #9938, opened 2021-10-31
Provide install files for the new environments (ucrt64 and clang64/32)

MehdiChinoune:


Biswa96:
For the python packages, instead of install script, use the PKGBUILD.setup-py-python-pkg template to remove shebang.

MehdiChinoune:
> For the python packages, instead of install script, use the PKGBUILD.setup-py-python-pkg template to remove shebang.

OK, I have removed the changes related to python packages

MehdiChinoune:
Remoe ocaml as it fails to build with gcc11

MehdiChinoune:
mingw32 build is hanging because of blender.

lazka:
if blender fails we can try reverting the ninja change..

PR #9965, opened 2021-11-02
update sip to 6.4.0 and pyqt-builder to 1.12.2

MehdiChinoune:


Biswa96:
pyqt5 & pyqt6 also has newer versions.

MehdiChinoune:
> pyqt5 & pyqt6 also has newer versions.

Thanks, I know. I tried to update pyqt6 to 6.2.1, but it failed to build as It needs a newer sip I think.

PR #9966, opened 2021-11-02
gmsh: update to 4.8.4

MehdiChinoune:


ImperatorS79:
It would be great if msys could consider using static linking for opencascade (-DENABLE_OCC_STATIC=1) with gmsh.

The official build uses it, and that would avoid to install 700MB+ dependencies just to use gmsh...

BTW @MehdiChinoune how did you fix the FLTK error with MSG type ?

MehdiChinoune:
> It would be great if msys could consider using static linking for opencascade (-DENABLE_OCC_STATIC=1) with gmsh.
> 
> The official build uses it, and that would avoid installing 700MB+ dependencies just to use gmsh...

I think in the mingw-w64 platform It is preferred to use dynamic linking over static one.

ImperatorS79:
I know that but let's be honest, who would install 700MB dependencies to use gmsh when you can download the official build which is "standalone" ?

Even I who initially added gmsh here does not use it through msys because of that...

MehdiChinoune:
> I know that but let's be honest, who would install 700MB dependencies to use gmsh when you can download the official build which is "standalone" ?
> 
> Even I who initially added gmsh here does not use it through msys because of that...

Well, The first thing mingw-w64 fellow the Arch Linux Philosophy of packaging to not split packages into runtime and development That increases the size because you extra files for development, The second thing is that other libraries insist on building the static library (opencascade for example) which increases the size a lot.
We could fix the second by banning building static and shared libraries for large projects, and instead separate the static library into its own package (like qt5-static and qt6-static)

ImperatorS79:
Seperating the opencascade package in two packages seems to be an interesting idea.

lazka:
I'm fine with splitting out static libs if they are way too large (like we do for qt5/6)

Issue #9804, opened 2021-10-19
xerces-c: does not export xercesc_3_2::DefaultHandler vtable with clang


jeremyd2019:
clang now fails with
```
ld.lld: error: undefined symbol: __declspec(dllimport) vtable for xercesc_3_2::DefaultHandler
>>> referenced by ogr\\ogrsf_frmts\\o\\.libs\\gmlhandler.o:(GMLXercesHandler::GMLXercesHandler(GMLReader*))
```

_Originally posted by @jeremyd2019 in https://github.com/msys2/MINGW-packages/issues/9800#issuecomment-946223510_

Reproducer: Sample from xerces-c
[SAX2Count.zip](https://github.com/msys2/MINGW-packages/files/7370009/SAX2Count.zip)

## CLANG32
```
$ g++ -DDLL_EXPORT -o SAX2Count *.cpp -lxerces-c
ld.lld: error: undefined symbol: __declspec(dllimport) vtable for xercesc_3_2::DefaultHandler
>>> referenced by D:/t/msys32/tmp/SAX2CountHandlers-e33623.o:(xercesc_3_2::DefaultHandler::DefaultHandler())
>>> referenced by D:/t/msys32/tmp/SAX2CountHandlers-e33623.o:(xercesc_3_2::DefaultHandler::DefaultHandler())
>>> referenced by D:/t/msys32/tmp/SAX2CountHandlers-e33623.o:(xercesc_3_2::DefaultHandler::DefaultHandler())
>>> referenced 3 more times
g++: error: linker command failed with exit code 1 (use -v to see invocation)
```

## MINGW32
```
$ g++ -DDLL_EXPORT -o SAX2Count *.cpp -lxerces-c
```

I tried to make a simple reproducer but it exported the vtable as expected.

jeremyd2019:
https://bugs.llvm.org/show_bug.cgi?id=31826 added my testcase to that, but couldn't dupe the original testcase from that bug.

revelator:
also ran into this one though i got it fixed, let me have a look at what i did to get it running.

jeremyd2019:
[...]

PR #9981, opened 2021-11-03
xerces-c: fix pc file

raedrizqie:


Biswa96:
IMO, a patch file would be better instead of sed.

lazka:
What does this fix for example?

raedrizqie:
> What does this fix for example?

preparing for gdal & maybe cegui.. gdal only assume xerces libs as `-lxerces -lpthread` but it actually depends on `icu` too.

lazka:
hm, weird that this only happens with clang and only on Windows...

raedrizqie:
> hm, weird that this only happens with clang and only on Windows...

i will test linking gdal with xerces-c without icu.. if it compiles, then this PR can be closed

MehdiChinoune:
The issue with xerces on clang is in #9804

PR #9962, opened 2021-11-02
add package: mingw-w64-zxing-cpp

cocomeow:

PR #9982, opened 2021-11-03
add package: mingw-w64-zxing-cpp

cocomeow:
squash merge commit: https://github.com/msys2/MINGW-packages/pull/9962


lazka:
Thanks!

PR #9992, opened 2021-11-04
vapoursynth: update to R57

MehdiChinoune:
The plugins were moved to their own repos:
see: https://github.com/vapoursynth/vapoursynth/commit/67fa8c5


lazka:
> The plugins were moved to their own repos:

does this mean we should package them?

MehdiChinoune:
> > The plugins were moved to their own repos:
> 
> does this mean we should package them?

I don't know!. I don't use it.

lazka:
:)

@ZachBacon any thoughts on this, since you added it initially

ZachBacon:
> :)
> 
> @ZachBacon any thoughts on this, since you added it initially

I mean sure, though I don't remember adding it initially xD (poor memory is both a blessing and a curse)

lazka:
ok, thanks, let's merge and see then

PR #9998, opened 2021-11-04
workaround for AVX misalignment issue for pass-by-value arguments

jgillis:
Fix for https://github.com/msys2/MSYS2-packages/issues/1209


lazka:
Newer source for the debian patch: https://salsa.debian.org/mingw-w64-team/gcc-mingw-w64/-/blob/2ee26ff4f275384aa6173ada884929db4d2f1bbf/debian/patches/vmov-alignment.patch

Issue #9827, opened 2021-10-20
Build errors with boost headers with clang since 13


LABPBS:
```
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/ip/tcp.hpp:19:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/basic_socket_acceptor.hpp:19:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/any_io_executor.hpp:22:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution.hpp:18:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/allocator.hpp:20:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/executor.hpp:20:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/execute.hpp:20:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/detail/as_invocable.hpp:20:
D:/a/_temp/msys64/clang64/include/boost/asio/detail/memory.hpp:88:20: error: reference to unresolved using declaration
  void* ptr = std::aligned_alloc(align, size);
                     ^
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/ip/tcp.hpp:19:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/basic_socket_acceptor.hpp:19:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/any_io_executor.hpp:22:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution.hpp:18:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/allocator.hpp:20:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/executor.hpp:20:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/execute.hpp:20:
In file included from D:/a/_temp/msys64/clang64/include/boost/asio/execution/detail/as_invocable.hpp:20:
D:/a/_temp/msys64/clang64/include/boost/asio/detail/memory.hpp:88:15: error: excess elements in scalar initializer
  void* ptr = std::aligned_alloc(align, size);
              ^                       ~~~~~~
```
_Originally posted by @LABPBS in https://github.com/msys2/MINGW-packages/pull/9823#issuecomment-947841188_

Biswa96:
[...]

PR #10009, opened 2021-11-05
asio: Update to 1.21.0

nightlark:
There's a bug fix in this version for an error detecting the availability of `std::aligned_alloc` when using libc++ that affects the clang64 and clang32 builds.


jeremyd2019:
Nice!  We might want to grab that patch (chriskohlhoff/asio@92aacba) and apply it to Boost's copy too, for #9827

oscarfv:
This updates ASIO to a development snapshot instead to the stable release. In exchange for fixing an easily patchable compile error one ends using a potentially broken library. IMHO this is not a good idea and it is not a good precedent either.

PR #10012, opened 2021-11-06
pywin32: fix problematic patch

jeremyd2019:
Remove infinite recursion from ThisAsIID method.  Disable a workaround
for broken header in Vista SDK that is not broken in MINGW-w64.

/cc @raedrizqie

### DIFF
--- a/mingw-w64-python-pywin32/002-com-extensions-fix.patch
+++ b/mingw-w64-python-pywin32/002-com-extensions-fix.patch
@@ -203,15 +203,6 @@ diff -Naur pywin32-b302-orig/com/win32com/src/include/PythonCOMServer.h pywin32-
                                                   REFIID iid)                                                     \
      {                                                                                                            \
          if (ppResult == NULL)                                                                                    \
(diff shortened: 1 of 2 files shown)


raedrizqie:
i think we should not mark reverse dependencies as fails in packaging check

jeremyd2019:
As an update, @Biswa96 has been working on getting BITS up to snuff in mingw-w64, and I tested that today.  It crashed due to the infinite recursion.  After taking that hunk out of the patch, and switching to use \ insteand of / in the local path passed to BITS, the BITS test from pywin32 worked.

The patch for libntquery.a that I made is now in upstream mingw-w64, but the packages here have not been updated yet.  The ifilter test in pywin32 needed to be updated, apparently due to python 2 vs 3 things, but did ultimately work.

Biswa96:
The required bits headers are now in upstream
* headers https://sourceforge.net/p/mingw-w64/mailman/message/37380444/
* crt https://sourceforge.net/p/mingw-w64/mailman/message/37380445/

PR #10017, opened 2021-11-06
[new-package] oneTBB 2021.4.0

MehdiChinoune:
Which name should we choose for it?
~intel-oneapi-tbb : used by Arch community~
oneTBB : the official project name

~I marked it to provide intel-tbb on clang* as some packages has updated to build with both the intel-tbb and the new one.~
I will try to build the intel-tbb dependants against the new one and see how it works.

Edit: It turns out the most intel-tbb dependants are not updated blender, embree, vtk ..etc and others do not require TBB at all (ogre3d and rocksdb)

### DIFF
--- a/mingw-w64-onetbb/001-fix-build-with-mingw.patch
+++ b/mingw-w64-onetbb/001-fix-build-with-mingw.patch
@@ -1,0 +1,78 @@
+--- a/src/tbb/allocator.cpp
++++ b/src/tbb/allocator.cpp
+@@ -27,11 +27,11 @@
+ 
+ #include <cstdlib>
+ 
+-#if _WIN32 || _WIN64
+-#include <Windows.h>
++#ifdef _WIN32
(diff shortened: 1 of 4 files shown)


lazka:
What about naming it like in the AUR: https://aur.archlinux.org/packages/intel-oneapi-tbb/ ?

MehdiChinoune:
> What about naming it like in the AUR: https://aur.archlinux.org/packages/intel-oneapi-tbb/ ?

In AUR they repackage the official installer for Arch which depends on other Intel OneAPI Tools.
In our case we are packaging OneTBB which is developed by Intel but supports other architectures like ARM64 and NVIDIA-GPUs.

lazka:
hm, ok, can we at least lowercase it?

MehdiChinoune:
> hm, ok, can we at least lowercase it?

OK, done.

PR #10019, opened 2021-11-06
libgphoto2: fix clang build

raedrizqie:


lazka:
This removes docs, translations and apparently usb support (the dll is removed). Is that on purpose?

raedrizqie:
> This removes docs, translations and apparently usb support (the dll is removed). Is that on purpose?

libusb was disabled anyway; `--with-libusb=no` (or is it not?)
there is a note in PKGBUILD: `#Note: --with-libusb=no is required for Windows 10 to work.`

lazka:
No idea :/ the diff suggests something changed though:

```diff
  --- /dev/fd/63	2021-11-06 12:31:58.000000000 +0000
  +++ /dev/fd/62	2021-11-06 12:31:58.000000000 +0000
  @@ -77,8 +77,6 @@
   /mingw64/lib/libgphoto2_port/0.12.0/disk.dll.a
   /mingw64/lib/libgphoto2_port/0.12.0/ptpip.dll
   /mingw64/lib/libgphoto2_port/0.12.0/ptpip.dll.a
  -/mingw64/lib/libgphoto2_port/0.12.0/usb1.dll
  -/mingw64/lib/libgphoto2_port/0.12.0/usb1.dll.a
   /mingw64/lib/pkgconfig/
   /mingw64/lib/pkgconfig/libgphoto2.pc
   /mingw64/lib/pkgconfig/libgphoto2_port.pc
```

raedrizqie:
hmm.. it built successfully on my fork https://github.com/raedrizqie/MINGW-packages/actions/runs/1507278189

raedrizqie:
this may be related to #10020

mstorsjo:
I’m curious - if there’s an explicit `-rpath` in the build files, that doesn’t cause issues when built with gcc/binutils - what is it that causes the flag not to be passed through in those cases? Because in #10020 we concluded that neither ld.bfd not ld.lld support rpath - so it’s something else that gets rid of it (or cause it not to be added at all?). So ideally we’d make it pick the same code paths for clang/ld.lld too, instead of teaching ld.lld to ignore an option that ld.bfd errors out on.

raedrizqie:
i think its better to define `hardcode_libdir_flag_spec` and `hardcode_libdir_flag_spec_CXX` to nothing, as it doesnt add anything on linking stage

PR #10020, opened 2021-11-06
libtool: fix rpath linking in clang

raedrizqie:
clang will fail while linking because of the space after rpath. substitute it with a comma

`-Wl,-rpath -Wl,${libdir}` <- this will fail
`-Wl,-rpath ${libdir}` <- this will fail

`-Wl,-rpath,${libdir}` <- ok
`-rpath ${libdir}` <- ok

### DIFF
--- a/mingw-w64-libtool/0015-Fix-rpath-linking-in-clang.patch
+++ b/mingw-w64-libtool/0015-Fix-rpath-linking-in-clang.patch
@@ -1,0 +1,39 @@
+diff -urN libtool-2.4.6-orig/m4/libtool.m4 libtool-2.4.6/m4/libtool.m4
+--- libtool-2.4.6-orig/m4/libtool.m4	2021-11-06 19:36:48 +0800
++++ libtool-2.4.6/m4/libtool.m4	2021-11-06 19:43:07 +0800
+@@ -5035,7 +5035,7 @@
+     # are reset later if shared libraries are not supported. Putting them
+     # here allows them to be overridden if necessary.
+     runpath_var=LD_RUN_PATH
+-    _LT_TAGVAR(hardcode_libdir_flag_spec, $1)='$wl-rpath $wl$libdir'
++    _LT_TAGVAR(hardcode_libdir_flag_spec, $1)='$wl-rpath,$libdir'
+     _LT_TAGVAR(export_dynamic_flag_spec, $1)='$wl--export-dynamic'
+     # ancient GNU ld didn't support --whole-archive et. al.
+     if $LD --help 2>&1 | $GREP 'no-whole-archive' > /dev/null; then
+@@ -5305,7 +5305,7 @@
+ 	  # DT_RUNPATH tag from executables and libraries.  But doing so
+ 	  # requires that you compile everything twice, which is a pain.
+ 	  if $LD --help 2>&1 | $GREP ': supported targets:.* elf' > /dev/null; then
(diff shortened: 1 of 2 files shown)


Biswa96:
Wondering. Is rpath required in mingw environment?

raedrizqie:
> Wondering. Is rpath required in mingw environment?

it doesnt really work i think.. just there for compatibility maybe

jeremyd2019:
In case you don't know, please also apply any patches to [MSYS2-packages/libtool](https://github.com/msys2/MSYS2-packages/tree/master/libtool) to keep in sync with this package.  But just opening one PR at first for review makes sense.

jeremyd2019:
@mati865 @mstorsjo is this a candidate to patch LLD's commandline parsing?

mati865:
I'm puzzled about how does it work. LLD in MinGW mode doesn't accept `rpath`: https://github.com/llvm/llvm-project/blob/d7b669b3a30345cfcdb2fde2af6f48aa4b94845d/lld/MinGW/Options.td
I think LLD should ignore it for the compatibility with BFD.

jeremyd2019:
That's what I kind of remember too

mstorsjo:
> I'm puzzled about how does it work. LLD in MinGW mode doesn't accept `rpath`: https://github.com/llvm/llvm-project/blob/d7b669b3a30345cfcdb2fde2af6f48aa4b94845d/lld/MinGW/Options.td

Indeed, and I can't reproduce the described cases where some forms of `-rpath` do succeed but not all.

In a clang64 msys2 environment:
```
# clang hello.c -o hello.exe -Wl,-rpath -Wl,foobar
lld: error: unknown argument: -rpath

# clang hello.c -o hello.exe -Wl,-rpath,foobar
lld: error: unknown argument: -rpath
```

In a mingw64 environment, both commands succeed.

> I think LLD should ignore it for the compatibility with BFD.

That'd be fine with me.

jeremyd2019:
I *think* this was related to trying to get gdal to build.  But, the only problem I had with gdal was linking to xerces-c, #9804 (aka https://bugs.llvm.org/show_bug.cgi?id=31826).  Just taking away the dllimport on the entirely-inline class allowed it to build and link fine (but I'm not sure if that would have negative impacts on the gcc build).

raedrizqie:
> I _think_ this was related to trying to get gdal to build. But, the only problem I had with gdal was linking to xerces-c, #9804 (aka https://bugs.llvm.org/show_bug.cgi?id=31826). Just taking away the dllimport on the entirely-inline class allowed it to build and link fine (but I'm not sure if that would have negative impacts on the gcc build).

gdal doesnt have this problem.. only packages that invokes `autoreconf -fiv` and using libtool to in linking stage will fail.

raedrizqie:
> > I _think_ this was related to trying to get gdal to build. But, the only problem I had with gdal was linking to xerces-c, #9804 (aka https://bugs.llvm.org/show_bug.cgi?id=31826). Just taking away the dllimport on the entirely-inline class allowed it to build and link fine (but I'm not sure if that would have negative impacts on the gcc build).
> 
> gdal doesnt have this problem.. only packages that invokes `autoreconf -fiv` and using libtool in linking stage will fail.

lazka:
libgphoto2 (in current master) has the same problem: https://github.com/msys2/MINGW-packages/pull/10019#issuecomment-981055500

Issue #10010, opened 2021-11-05
libavif 0.9.3 update-request


novomesk:
Hello,

Please, when you have time, upgrade libavif package to 0.9.3

I am sure that rebuild of `kimageformats-qt5` will be needed afterwards and most probably also for `darktable` and `libgd`

Thanks!

Biswa96:
Would you like to test the new packages from the linked pull request?

novomesk:
I will test (from the PR or new packages) but probably only after weekend.

I'll tell you if there is some problem.

Thanks again.

novomesk:
I tried following two and they work for me:
`mingw-w64-x86_64-libavif-0.9.3-1-any.pkg.tar.zst`
`mingw-w64-x86_64-kimageformats-qt5-5.85.0-3-any.pkg.tar.zst`

PR #10024, opened 2021-11-06
libavif: Update to 0.9.3

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/10010


Biswa96:
@kmilos Review please.

kmilos:
Looks ok, but one thing: not sure we need the pixbuf patch any more, I vaguely remember pushing for the fix upstream...

kmilos:
Yep, this should've done it already, don't know why we need this extra line now? 

https://github.com/AOMediaCodec/libavif/pull/615

PR #10027, opened 2021-11-07
Fix broken urls

MehdiChinoune:


MehdiChinoune:
The CI failure has nothing to do with my changes

PR #10029, opened 2021-11-07
gnuplot: update to 5.4.2

raedrizqie:


podsvirov:
I'm not an active user of `gnuplot`, but I know that it can be integrated with `texlive`, which is already available in the distribution. Someone please work out this functionality, if it is possible.

PR #10033, opened 2021-11-07
Update mingw-w64, and enable remaining COM extensions in pywin32

jeremyd2019:
Update mingw-w64 to get new headers/libs necessary for these pywin32 extensions:
- bits
- ifilter

I have tested both, and both work.


jeremyd2019:
Is this OK to merge or should I split mingw-w64 update from the pywin32 change?

jeremyd2019:
the build of pywin32 didn't pull in the updated headers/crt in autobuild, so the newly-enabled modules were not built.

lazka:
good catch, I'll rebuild after the repo update

PR #10039, opened 2021-11-08
libraw: enable support for more formats

kmilos:


lazka:
Do you really need this? upstream suggests to not enable it for security reasons.

kmilos:
The additional code paths are less tested indeed... There is no burning issue, i.e. no report that a user is missing support for a Sigma camera etc., so we could leave it as is until such moment.

lazka:
ok, I'd rather not then. OK with my change?

kmilos:
Sure

PR #10043, opened 2021-11-09
Add oatpp library

Samega7Cattac:


Biswa96:
The source should point to the specified tag.

Samega7Cattac:
can't find on the docs how to do that

Biswa96:
Here it is https://wiki.archlinux.org/title/Arch_package_guidelines#Package_sources

Biswa96:
O! my bad. The link already points to pkgver. Need to add Ninja in makedepends.

PR #10059, opened 2021-11-11
python-keyring: fix versioning

raedrizqie:
failed on reverse dependency (spyder)

this is good to go..


Biswa96:
I'm not python expert. Just wondering if the export command should be in build function.

raedrizqie:
> I'm not python expert. Just wondering if the export command should be in build function.

either using this export workaround, or we have to add `setuptools-scm` as dependency for versioning, imo

PR #10064, opened 2021-11-11
ogre3d: rebuild for clang

raedrizqie:


jeremyd2019:
Why does it need a rebuild for clang?  OTOH, it needs a rebuild for the other subsystems because it still has python3.8 paths(!)  Must have been missed in the python3.9 giant rebuild.

raedrizqie:
> Why does it need a rebuild for clang? OTOH, it needs a rebuild for the other subsystems because it still has python3.8 paths(!) Must have been missed in the python3.9 giant rebuild.

i was referring to https://packages.msys2.org/queue and saw it failed to compile for clang64 (matbe because of no freeimage lib for clang64 earlier)

PR #10065, opened 2021-11-12
mdloader: fix clang32 build

jeremyd2019:
properly mangle global asm statement.  GCC doesn't care but apparently Clang does.


jeremyd2019:
https://github.com/graphitemaster/incbin/pull/56

PR #10070, opened 2021-11-12
qt6-base : Enable AVX Support

MehdiChinoune:
It was blocked with MINGW/GCC 64bit due to AVX misalignment that should be fixed by #9998


Biswa96:
Will there be any issue with older CPUs?

MehdiChinoune:
> Will there be any issue with older CPUs?

Did you see any with MINGW32 or CLANG*, The AVX support was blocked only with MINGW/GCC 64bits.
It loads the AVX code when the CPU supports it. like OpenBLAS, FFTW, embree ..

PR #10074, opened 2021-11-12
python-astroid: update to 2.8.4

raedrizqie:
reverse dep failure (pylint)


lazka:
> reverse dep failure (pylint)

that seems relevant in this case, no?

raedrizqie:
> > reverse dep failure (pylint)
> 
> that seems relevant in this case, no?

i will update pylint afterwards

lazka:
Please update it in this PR so we can see if all dependencies are correct

raedrizqie:
> Please update it in this PR so we can see if all dependencies are correct

will do.. thx

lazka:
ok, fair enough

PR #10090, opened 2021-11-14
qt5-quick3d: Disable building with system-assimp

MehdiChinoune:
qt5-quick3d does not support the latest assimp 5.1.0


Biswa96:
Can qt6-quick3d be built with latest assimp? If yes then KDE devs should be notified to import the upstream patches to qt5.

mingw-w64-qt5-pkgs/PKGBUILD
```diff
@@ -252,6 +252,7 @@ build() {
     -system-sqlite \
     -system-zlib \
     -no-iconv \
+    -qt-assimp \
```
lazka, line 255: typo?

MehdiChinoune:
> Can qt6-quick3d be built with latest assimp? If yes then KDE devs should be notified to import the upstream patches to qt5.

No, It can't. I have tried to to build it yesterday, it failed with the same error messages.

MehdiChinoune:
I don't know why `configure.bat` doesn't accept options for other than qtbase?

PR #10094, opened 2021-11-15
openblas: Create split package with 32-bit and 64-bit indexing

mmuetzel:
Continuing from #9932.

This adds a version of OpenBLAS with 64-bit indexing in the form of a split package.

### DIFF
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -3,9 +3,10 @@
 
 _realname=OpenBLAS
 pkgbase=mingw-w64-openblas
-pkgname="${MINGW_PACKAGE_PREFIX}-openblas"
+pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
+         $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas64"))
 pkgver=0.3.18
-pkgrel=3
+pkgrel=4
 pkgdesc="An optimized BLAS library based on GotoBLAS2 1.13 BSD, providing optimized blas, lapack, and cblas (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
@@ -31,7 +32,8 @@ source=(${_realname}-${pkgver}.tar.gz::https://github.com/xianyi/OpenBLAS/archiv
         005-fix-clang32-compile-flags.patch
         006-fix-cmake-detect-clang-64-bits.patch
         007-support-building-both-static-shared.patch
-        008-export-only-shared-library.patch)
+        008-export-only-shared-library.patch
+        009-cmake-Set-SUFFIX64-also-for-NOFORTRAN.patch)
 install=${_realname}.install
 sha256sums=('1632c1e8cca62d8bed064b37747e331a1796fc46f688626337362bf0d16aeadb'
             '6a73ee677d8d37509cd08f380b0359ee7fac2595fd64695dd054e838380a6a33'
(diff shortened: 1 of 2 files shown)


Biswa96:
In the upstream pull request, always remember to add a commit message how your change fixes the issue and the before & after differences.

3rav:
May I analogically create (i8) versions of packages: pastix, arpack, scotch?

mmuetzel:
The patch was accepted upstream.

Rebased the PR on the current HEAD.

mingw-w64-openblas/PKGBUILD
```diff
@@ -92,18 +97,64 @@ build() {
     -DDYNAMIC_ARCH=ON \
     -DUSE_THREAD=ON \
     -DNUM_THREADS=64 \
-    -DTARGET=CORE2 \
+    -DTARGET=GENERIC \
```
lazka, line 100: What's this change about?

mingw-w64-openblas/PKGBUILD
```diff
@@ -3,9 +3,10 @@
 
 _realname=OpenBLAS
 pkgbase=mingw-w64-openblas
-pkgname="${MINGW_PACKAGE_PREFIX}-openblas"
+pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
+         $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas-64"))
```
lazka, line 7: What about openblas64 instead, like in Debian and the pkg-config ID?

mingw-w64-openblas/PKGBUILD
mmuetzel: Oops. I wanted to report that in a different PR. But it looks like it sneaked into this one at some point.

According to this https://github.com/xianyi/OpenBLAS/issues/3139, OpenBLAS should be configured with `TARGET=GENERIC` when building with `DYNAMIC_ARCH=ON` to avoid that the common code uses instructions that might not be available on some processors.

Is it ok to leave it in this PR? Or should I remove that change from here.

mingw-w64-openblas/PKGBUILD
mmuetzel: I'll change that.

mingw-w64-openblas/PKGBUILD
lazka: Please move this to another/future PR

lazka: The last time this was discussed was https://github.com/msys2/MINGW-packages/pull/7830

CORE2 means SSE3, which is quite old.. but if there are real problems we can set it to PENTIUM4 which is what Windows 7 requires.

mmuetzel: Unless someone complains, it's probably ok to keep the current `TARGET`.

We had issues with some people not being able to use the OpenBLAS library that we build in MXE Octave. We used `TARGET=PRESCOTT` because we thought that was the oldest 64-bit processor from the list.
Strange thing: It looked like it was dependent on the processor of the *build* system if the library could run on some systems or not, even though we set `TARGET`...
https://savannah.gnu.org/bugs/?61246

mingw-w64-openblas/PKGBUILD
mmuetzel: Done

mmuetzel:
> May I analogically create (i8) versions of packages: pastix, arpack, scotch?

@3rav: I don't have any authority here. But it looks like this approach is acceptable. So, I'd say: Go ahead! 🙂

mati865:
Won't having `replaces` section for both `${MINGW_PACKAGE_PREFIX}-openblas` and `${MINGW_PACKAGE_PREFIX}-openblas64` packages create issues?
On 64-bit systems that would mean `${MINGW_PACKAGE_PREFIX}-openblas` is replaced by `${MINGW_PACKAGE_PREFIX}-openblas64` so the user won't be able to install it.

mmuetzel:
@mati865: Could you please elaborate? `package_openblas64` has `replaces=()`. Doesn't that mean that it doesn't replace any package?

mati865:
@mmuetzel oh, I have missed it being so low there. It's fine in this case.

PR #10097, opened 2021-11-15
rocksdb: update to 6.26.0

raedrizqie:


MehdiChinoune:
Could you remove intel-tbb from dependencies, since it is not.
https://github.com/facebook/rocksdb/blob/main/INSTALL.md

PR #10098, opened 2021-11-15
qemu: Update to 6.2.0

heljkon:
This PR intends to update qemu to 6.2.0.

https://gitlab.com/qemu-project/qemu/-/commit/984099911275cd4b703e0d9c35b37dd52928acdd fixes native msys2 build on qemu, so --cross-prefix is no longer needed to configure.

Because the first build is done by using prerelease sources, checksums will be added on final release.

### DIFF
--- a/mingw-w64-qemu/msys2.examples.tests.sh
+++ b/mingw-w64-qemu/msys2.examples.tests.sh
@@ -1,8 +1,32 @@
 #!/bin/bash
 
 echo
-echo "Qemu examples/tests prepared to be executed with qemu/win32."
-echo "Executed commandlines will be printed to screen."
+echo "Qemu examples & tests."
+echo "----------------------"
+echo "Created to test Msys2 Qemu package, known to work for Cygwin and Linux, too."
+echo "Executed Qemu commandlines will be printed to screen."
+echo
+CONFIGFILE=~/.qemu.$(basename $0)
+CONFIGFILENAME="$(basename $CONFIGFILE)"
+if [ ! -f "$CONFIGFILE" ]
+then
+	DOWNLOADDIR=$(realpath .)
+	echo "Configuring current directory '$DOWNLOADDIR' as download directory."
+	read -p "Continue? (y|[n]) " TEST
+	echo
+	[ "y" == "$TEST" ] || exit
+	touch "$CONFIGFILE" || exit 1
+	touch "$DOWNLOADDIR/$CONFIGFILENAME" || exit 1
+	echo "$DOWNLOADDIR" > "$CONFIGFILE"
+fi
+DOWNLOADDIR=$(cat "$CONFIGFILE")
+echo "Configuration file: '$CONFIGFILE'"
+echo "Download directory: '$DOWNLOADDIR'"
+if ! mkdir -p "$DOWNLOADDIR" || ! touch "$DOWNLOADDIR/$CONFIGFILENAME"
+then
+	echo "Download directory '$DOWNLOADDIR' is not usable"
+	exit 1
+fi
 echo
 echo "On execution each test needs to download, most test only a few 10 MB or less,"
 echo "but several up to some 100MB."
@@ -11,6 +35,9 @@ read -p "Only accept reasonable downloads? ([y]|n) " TEST
 echo
 read -p "Clean after execution (removes all but downloads)? (y|[n]) " TEST
 [ "y" == "$TEST" ] && REMOVEEXECDIR=1
+echo
+read -p "Audio input (microphone) accessible? ([y]|n) " TEST
+[ "n" != "$TEST" ] && MICROPHONE=1
 echo
 echo "Name block of qemu examples to execute."
 echo "Choose year of qemu-advent-calender (2014, 2016, 2018, 2020) or qemu-desktop (DVD)"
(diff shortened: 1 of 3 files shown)


heljkon:
Seems this PR is too old to be seen...

If anyone could have a look?
@lazka @Biswa96 @jeremyd2019

jeremyd2019:
I had 'subscribed' to this PR when it was created because this is a package that I am interested in and use, but apparently GitHub didn't send a notification when it's marked 'ready for review'.  Besides which several packages were getting thrashed recently and I generally try to avoid adding packages in the middle of something like that :wink:

mingw-w64-qemu/msys2.examples.tests.sh
```diff
@@ -1,8 +1,32 @@
...
-echo "Qemu examples/tests prepared to be executed with qemu/win32."
-echo "Executed commandlines will be printed to screen."
+echo "Qemu examples & tests."
+echo "----------------------"
+echo "Created to test Msys2 Qemu package, known to work for Cygwin and Linux, too."
+echo "Executed Qemu commandlines will be printed to screen."
+echo
+CONFIGFILE=~/.qemu.$(basename $0)
+CONFIGFILENAME="$(basename $CONFIGFILE)"
+if [ ! -f "$CONFIGFILE" ]
```
1480c1, line 11: if this is using bash, it would be better to just go ahead and use `[[` instead of `[` as `[[` is generally more resilient than `[`

1480c1: hmm, I see the rest of the file preferring `[`, I guess this is fine then

mingw-w64-qemu/msys2.examples.tests.sh
```diff
@@ -58,10 +85,48 @@ function testImageInDir {
...
+		PATHTAIL="${PATHTAIL:${#PATHHEAD}+1}"
+	done
+	return 1
+}
+
 function showMsys2ArgConvEnv {
 	[ -n "$MSYSTEM" ] && [ -n "$MSYS2_ARG_CONV_EXCL" ]
 }
 
+function toolexec {
```
1480c1, line 115: an alternative I can suggest instead of this function would be wherever this is called such as `toolexec a b c`, change the form to

```bash
(
    set -x
    a b c
)
```

to instead use bash to print out how the command will be evaluated

```bash
$ a=532

$ ( set -x; echo 1 2 $a; )
+ echo 1 2 532
1 2 532

$ a="532 zz"

$ ( set -x; echo 1 2 $a; )
+ echo 1 2 532 zz
1 2 532 zz

$ ( set -x; echo 1 2 "$a"; )
+ echo 1 2 '532 zz'
1 2 532 zz
```

alternatively, you could convert this function to a subshell based function and use the same thing

```bash
$ function toolexec  (
    set -x
    "${@}"
)

$ toolexec echo bv c "$a"
+ echo bv c '532 zz'
bv c 532 zz
```

mingw-w64-qemu/msys2.examples.tests.sh
```diff
@@ -180,19 +249,9 @@ function checkBinary {
...
-					grep spice &> /dev/null ||
-				echo spice-server
-			;;
-		*)
-			which $PACKAGE &> /dev/null ||
-				which ${BINARIES[$PACKAGE]} &> /dev/null ||
-				echo "$PACKAGE"
-			;;
-	esac
+	which $PACKAGE &> /dev/null ||
```
1480c1, line 252: does $PACKAGE need to not be quoted? I personally prefer always quoting unless there's a reason to not, such as argument expansion etc

mingw-w64-qemu/msys2.examples.tests.sh
```diff
@@ -323,13 +382,18 @@ function accel {
...
 
-	local BINDIR="$(dirname $(which qemu-system-x86_64))"
-	local QI_FW="/usr/share/qemu/firmware"
+	local BINDIR="$(dirname $(qWhich qemu-system-x86_64))"
+	local COMMON_FW="$BINDIR/../share/qemu"
+	local MSYS_FW_OLD="$BINDIR/../lib/qemu"
+	local DIST="$BINDIR"
+	local QI_FWDSC="/usr/share/qemu/firmware"
+	local CYG64_DIST="/cygdrive/c/Program Files/qemu"
+	local CYG32_DIST="/cygdrive/c/Program Files (x86)/qemu"
```
1480c1, line 391: this path doesn't exist for msys2, it would just be `/c/` instead of `/cygdrive/c/`, unless I'm missing something else

1480c1: Hmm, it's used for something I'm not entirely sure about below, ignore this for now

mingw-w64-qemu/msys2.examples.tests.sh
```diff
@@ -474,12 +545,42 @@ executeSpicyForBackgroundQemu
...
+toolexec qemu-img rebase -b '' -f qcow2 $TESTDIR/iso.qcow2
+toolexec rm $TESTDIR/rebase.qcow2
+toolexec qemu-img compare -f raw $LIVE_IMAGE_FILE -F qcow2 $TESTDIR/iso.qcow2
+toolexec qemu-img check $TESTDIR/iso.qcow2
+toolexec qemu-img snapshot -a REBASED $TESTDIR/iso.qcow2
+toolexec qemu-img compare -f raw $LIVE_IMAGE_FILE -F qcow2 $TESTDIR/iso.qcow2
+toolexec qemu-img snapshot -d REBASED $TESTDIR/iso.qcow2
+toolexec qemu-img snapshot -l $TESTDIR/iso.qcow2
+removeDir $TESTDIR
+}
```
1480c1, line 576: Any particular reason for the lack of indentation?

mingw-w64-qemu/msys2.examples.tests.sh
heljkon: The expected values for $PACKAGE are known. But in general you are right and after reviewing my code, I found other variables which should be quoted.

Will be done in next update!

mingw-w64-qemu/msys2.examples.tests.sh
heljkon: The first version of the script was created to test only on mingw shell. 

Now the script is executable on Cygwin (and Linux), too. These paths are necessary to support all test cases with Cygwin and Qemu binary distribution.

mingw-w64-qemu/msys2.examples.tests.sh
heljkon: Hm, well, the lines are long enough. On the one hand I omitted the indentation for the test cases.
And if I remember correctly, the functions around the test cases were added later and I didn't want a complete file changeset.

So: no sensible reason. I'll consider changing it...

mingw-w64-qemu/msys2.examples.tests.sh
heljkon: Will test the output using 'set -x' in the next update.

mingw-w64-qemu/msys2.examples.tests.sh
heljkon: Creating a '{', '[', ']', '}', '\' on a german keyboard causes cramps (AltGr-7, AltGr-8, ...) in the right hand.
I resist it if I can.

PR #10109, opened 2021-11-16
gdal: update to 3.4.0 + fix clang build

raedrizqie:
hmm.. i dont know why the compiling took too long


raedrizqie:
mine compiled successfully here: https://github.com/raedrizqie/MINGW-packages/actions/runs/1468121900

mingw-w64-gdal/006-fix-compile-without-libtool.patch
```diff
@@ -0,0 +1,142 @@
...
+ $(GDAL_LIB):	$(GDAL_OBJ) GDALmake.opt
+ 	rm -f libgdal.a
+ 	$(AR) r $(GDAL_LIB) $(GDAL_OBJ)
+ 	$(RANLIB) $(GDAL_LIB)
+ 
+-$(GDAL_SLIB):	$(GDAL_OBJ) $(GDAL_LIB)
+-	$(LD_SHARED) $(GDAL_SLIB_SONAME) $(GDAL_OBJ) $(GDAL_LIBS) $(LDFLAGS) $(LIBS) \
+-		-o $(GDAL_SLIB)
++$(GDAL_SLIB):	$(GDAL_LIB) $(GDAL_BIN_OBJ) $(GDAL_OBJ)
++	for f in $(GDAL_OBJ); do $(LD_SHARED) $$f $(GDAL_BIN_OBJ) $(GDAL_LIB) $(LDFLAGS) $(LIBS) -Wl,--out-implib,$(GDAL_SLIB_LINK) -o $(GDAL_SLIB) ; done
```
MehdiChinoune, line 92: I don't know why did you add this loop?
but this step took about 3 hours of buildtime (executed two times)

MehdiChinoune:
Building gdal with clang is already fixed, We should wait for xerces to be fixed before enable it for gdal, without those tricks that introduce some anomalities.

mingw-w64-gdal/006-fix-compile-without-libtool.patch
raedrizqie: i had to work around the error while compiling `command line is too long` (something like that) can you help me fix this?

raedrizqie:
> Building gdal with clang is already fixed, We should wait for xerces to be fixed before enable it for gdal, without those tricks that introduce some anomalities.

ok then

PR #10117, opened 2021-11-18
python-imageio: update to 2.10.5

kmilos:


kmilos:
Actually, looks like 2.11 is on the way, will drop this... They're really knocking them out lately ;)

Biswa96:
You can also force push the new update here.

PR #10118, opened 2021-11-18
Mark "imagemagick" and "libosmscout" as conflicting packages.

mmuetzel:
They both install "bin/import.exe". I guess they should be marked as conflicting with each other.

See also PR #9931.


lazka:
one conflict on libosmscout should be enough

mmuetzel:
I'm not sure I understood correctly.
I updated the PR to only change the `PKGBUILD` of `libosmscout`. Is that what you asked for?

PR #10119, opened 2021-11-18
field3d: fix clang build

raedrizqie:


mingw-w64-field3d/006-mingw-fixes.patch
```diff
@@ -0,0 +1,60 @@
...
+ #  else
+ #    ifdef FIELD3D_EXPORT
+-#      define FIELD3D_API __declspec(dllexport)
++#      define FIELD3D_API __attribute__((visibility("default")))
+ #    else
+-#      define FIELD3D_API __declspec(dllimport)
++#      define FIELD3D_API
+ #    endif
+ #  endif
+ #else
```
jeremyd2019, line 27: I haven't looked at this in context, but it doesn't seem right to be switching *away* from dllexport/dllimport to visibility attributes.

raedrizqie: gdal also uses visibility instead of dllexport..

Alexpux: For mingw we need use dllimport/dllexport

jeremyd2019:
sorry, I lost track of this, until I saw #10253 and it reminded me

jeremyd2019:
I'm going to try "hijacking" this and merging with #10253

mingw-w64-field3d/006-clang-fixes.patch
```diff
@@ -0,0 +1,66 @@
...
++  FieldMapping::Ptr readFieldMapping(const OgIGroup &mappingGroup)
+   {
+     ClassFactory &factory = ClassFactory::singleton();
+     
+@@ -220,7 +220,7 @@
+ 
+   //! This function creates a FieldIO instance based on field->className()
+   //! which then writes the field data in layerGroup location
+-  FIELD3D_API bool writeField(OgOGroup &layerGroup, FieldBase::Ptr field)
++  bool writeField(OgOGroup &layerGroup, FieldBase::Ptr field)
```
jeremyd2019, line 63: I'm a little concerned about removing export from these functions.  Are they still declared as exported in the header/declaration, or were these removed to shut up a warning/error that they were *not* exported from the header/declaration?  If the latter, it probably would make more sense to add the export to the header/declaration.

raedrizqie: its because of how they implemented `FIELD3D_API` it was defined to nothing when importing, instead of dllimport

jeremyd2019: OK, I just had a look, and these methods are not even in a header, so should not need to be exported

PR #10127, opened 2021-11-19
Qt5/6-3d: Enable SSE2

MehdiChinoune:
Before someone start complaining, It is just a hint. Look at the build log to be sure.
SSE2 is not enabled for 32 bits environments.


jeremyd2019:
We've already decided that SSE2 can be required even for 32-bit (`-march=pentium4` in makepkg_mingw.conf)

PR #10130, opened 2021-11-19
python-spyder: dependencies update

raedrizqie:


Biswa96:
In my opinion, a patch file with all those modfications would be better.

raedrizqie:
> In my opinion, a patch file with all those modfications would be better.

ok.. i will update it

mingw-w64-python-spyder/spyder-requirements-fix.patch
```diff
@@ -0,0 +1,37 @@
+--- a/setup.py	2021-09-16 09:18:30 +0800
++++ b/setup.py	2021-11-20 17:41:22 +0800
+@@ -211,7 +211,7 @@
+     'ipython>=7.6.0',
+     'jedi>=0.17.2,<0.19.0',
+     'jsonschema>=3.2.0',
+-    'keyring>=17.0.0',
++    'keyring',
```
lazka, line 8: I hope fixed here: 283d66bb008c5dc73a9b

mingw-w64-python-spyder/spyder-requirements-fix.patch
```diff
@@ -0,0 +1,37 @@
...
+-    'pyqt5<5.13',
+-    'pyqtwebengine<5.13',
++    'pyqt5>=5.13',
++    #'pyqtwebengine<5.13',
+     'python-lsp-server[all]>=1.2.2,<1.3.0',
+     'pyxdg>=0.26;platform_system=="Linux"',
+     'pyzmq>=17',
+     'qdarkstyle==3.0.2',
+-    'qstylizer>=0.1.10',
++    'qstylizer',
```
lazka, line 29: fixed in #10135

mingw-w64-python-spyder/spyder-requirements-fix.patch
raedrizqie: i think i tried before.. using setuptools-scm.. but still failed.. hopefully your fix will work.

maybe we have to remove the `.github` dir first

mingw-w64-python-spyder/spyder-requirements-fix.patch
```diff
@@ -0,0 +1,37 @@
...
+     'qtawesome>=1.0.2',
+     'qtconsole>=5.1.0',
+     'qtpy>=1.5.0',
+-    'rtree>=0.9.7',
++    #'rtree>=0.9.7',
```
lazka, line 34: adder here: #10138

mingw-w64-python-spyder/spyder-requirements-fix.patch
lazka: hm, ok, I'll check

lazka: yes, it's fixed now

PR #10140, opened 2021-11-20
lis: update sha256sums and enable complex

MehdiChinoune:
They reuploaded the same file on 18/11!


Biswa96:
Just asking. Do you change SHA checksums manually instead of using `updpkgsums` command?

MehdiChinoune:
> Just asking. Do you change SHA checksums manually instead of using `updpkgsums` command?

Yes, I never use `updpkgsums`

lazka:
> They reuploaded the same file on 18/11!

They are doing this all the time :/ https://github.com/anishida/lis/issues/29

lazka:
> Yes, I never use `updpkgsums`

You should give it a try :)

PR #10141, opened 2021-11-20
assimp: update to 5.1.1

MehdiChinoune:
the compatibility with qt6/qt5-quick3d should be restored now.
https://github.com/assimp/assimp/pull/4184

I will test to rebuild qt5-quick3d and qt6-quick3d with it to be sure.


MehdiChinoune:
Unfortunately, still not compatible with qt5-quick3d/qt6-quick3d.

MehdiChinoune:
Reopen after backporting patch from https://github.com/assimp/assimp/pull/4203

Issue #10151, opened 2021-11-21
cmake fails to interpret arpack-ng's paths


juanjosegarciaripoll:
Create a project with two files. The first one is `CMakeLists.txt`
````
cmake_minimum_required(VERSION 3.14)
project(test)
find_package(arpack-ng)
add_library(test "foo.c")
target_link_libraries(test PUBLIC ARPACK::ARPACK)
````
The second one is a trivial C file
````
#include <arpack.h>
int main()
{
  return 0;
}
````
Now try to build with
````
cmake -B build -S . -G "MinGW Makefiles"
cmake --build build
````
It will fail because the file arpack-ng-config.cmake file has a hardcoded path that cmake does not recognize
````
$ cmake -B build -S . -G "MinGW Makefiles"
-- Configuring done
CMake Error in CMakeLists.txt:
  Imported target "ARPACK::ARPACK" includes non-existent path

    "/mingw64/include/arpack"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.



-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
````

Biswa96:
Hi @juanjosegarciaripoll 👋 Thank you for reporting this issue and providing a minimal code to reproduce this issue. I have sent a pull request to fix this issue. Would you like to test the package in the CI build artifacts from the above linked pull request?

juanjosegarciaripoll:
I verified it solves the issue. Thank!

PR #10160, opened 2021-11-22
arpack: Fix cmake file to relocate prefix

Biswa96:
Changes:
  * Enable ICB for arpack header files.
  * Enable check function.
  * Add patch file to relocate prefix.
  * Use sed to replace MINGE_PREFIX with _IMPORT_PREFIX.

Fixes https://github.com/msys2/MINGW-packages/issues/10151


juanjosegarciaripoll:
The fix is working. Thanks!

PR #10190, opened 2021-11-27
new package: python-jupyter-nbviewer

raedrizqie:


mingw-w64-python-jupyter-nbviewer/PKGBUILD
```diff
@@ -0,0 +1,55 @@
...
+  cp -r "${_realname}-${pkgver}" "python-build-${MSYSTEM}"
+}
+
+build() {
+  msg "Python build for ${MSYSTEM}"
+  cd "${srcdir}/python-build-${MSYSTEM}"
+
+  # Disable check for non-windows modules
+  sed -i "s|newrelic!=2.80.0.60|newrelic!=2.80.0.60; sys.platform!='win32'|g" requirements.txt
+  sed -i "s|pylibmc|pylibmc; sys.platform!='win32'|g" requirements.txt
```
naveen521kk, line 43: Is this really needed? Because it already has the condition to exclude those modules on windows.

raedrizqie: it fails the check without it..

naveen521kk: oh, I see. I guess rather than adding `sys.platform` checks it would be better to remove it from requirements.txt?

raedrizqie: ok

mingw-w64-python-jupyter-nbviewer/PKGBUILD
```diff
@@ -0,0 +1,55 @@
+# Contributor: Raed Rizqie <raed.rizqie@gmail.com>
+
+_realname=nbviewer
+pkgbase=mingw-w64-python-${_realname}
```
lazka, line 4:
```suggestion
pkgbase=mingw-w64-python-jupyter-${_realname}
```

PR #10192, opened 2021-11-27
poco: clang update fix

raedrizqie:


mingw-w64-poco/007-mingw-wide-character-startup-module.patch
```diff
@@ -4,7 +4,7 @@
  	DEFINE_SYMBOL Util_EXPORTS
  )
  
-+if(MINGW)
++if(MINGW OR CLANG)
```
lazka, line 7: That shouldn't change anything

mingw-w64-poco/011-cmake-mingw.patch
```diff
@@ -15,7 +15,7 @@
  			endforeach()
  		endif(X64)
  	endif()
-+	if(MINGW)
++	if(MINGW OR CLANG)
```
lazka, line 18: same

raedrizqie: i dont know, but without this, cmake couldnt find windmc.exe

lazka:
It seems to build fine in #10194. Can I merge that instead?

raedrizqie:
> It seems to build fine in #10194. Can I merge that instead?

of course..

PR #10203, opened 2021-11-28
mingw-w64-mesa/clang prefixes: Enable ZSTD compression

pal1000:
Also don't bother installing GCC on clang prefixes


Biswa96:
Just wondering. ZSTD compression for what?

pal1000:
shader cache.

PR #10209, opened 2021-11-29
petsc: Fix build on clang64 (no mpi)

MehdiChinoune:


lazka:
@okhlybov

okhlybov:
What's the problem with MPI? According to integration tests, the Clang's `msmpi` package is intact.

The only thing thats presumably missing is a FORTRAN support.

MehdiChinoune:
I was the one who brought both `openblas` and `msmpi` to CLANG*, so I know what I am doing.
But since I am not a user of petsc it is not a big deal if this PR doesn't get merged.
Good luck packaging it with openblas and MPI support!

1480c1:
> so I know what I am doing

The issue from what I can tell is that we don't know what you are doing as both the description of the pr and the commit message is empty, thus the questions

MehdiChinoune:
> > so I know what I am doing
> 
> The issue from what I can tell is that we don't know what you are doing as both the description of the pr and the commit message is empty, thus the questions

Then these questions should be rephrased in another way. Not in a way that I am messing with things without knowing what I am doing.

MehdiChinoune:
> > so I know what I am doing
> 
> The issue from what I can tell is that we don't know what you are doing as both the description of the pr and the commit message is empty, thus the questions

Also the description is clear 'Fix build on clang64' what should I say more. '(no mpi)' I couldn't make it build with MPI for clang64.

Honestly, I couldn't say more.

LABPBS:
To the extent of the knowledge lurking around this repo provides me since clang toolchain doesnt have a fortran compiler and thus openblas has no lapack Mehdi is building it with an alternative implementation from #10199
no idea about mpi

1480c1:
> Then these questions should be rephrased in another way. Not in a way that I am messing with things without knowing what I am doing.



> Also the description is clear 'Fix build on clang64' what should I say more. '(no mpi)' I couldn't make it build with MPI for clang64.

There are many reasons why someone would disable something for a package, reasons such as configure error, compile error, runtime error, significant feature/speed difference, etc. My interpretation of the original question was more towards "What's the error with MPI, if there is one?" rather than questioning your ability or knowledge and the question added about openblas in the edited version would be similar to the previous one.

As this is open source, in the future, there probably will be people who will look at the commit history of a PKGBUILD and wonder what cause a change such as disabling MPI, and even if they look at this PR, there's no context other than `(no mpi)`, so the only way they would be able to tell would be to get your changes, and undo the ones specific to MPI and build, but most people, including myself, aren't willing to pull every PKGBUILD and edit it just to find out why something is lacking or missing. Plus if that information was provided at the time it was disabled, it would help with issues that might have been caused by packages or compilers at that specific time and would make it easier to verify if those issues are fixed in the future.

1480c1:
When dealing with people on GitHub and GitLab etc, it might be best to assume that people are simply asking questions because they don't know rather than an attack on your expertise, simply because they might not have the same expertise as you.

https://en.wikipedia.org/wiki/Hanlon%27s_razor (but not as harsh on assuming the other is stupid, just unfamiliar or unaware)

PR #10221, opened 2021-12-01
mlpack: add cereal, remove cotire

raedrizqie:
cereal is listed as dependency on https://github.com/mlpack/mlpack

>  Armadillo      >= 9.800
  Boost (math_c99, spirit) >= 1.58.0
  CMake          >= 3.6
  ensmallen      >= 2.10.0
  cereal         >= 1.1.2

adding cotire as makedepends will conflicts with cmake. i think mlpack uses its internal cotire anyway..

### DIFF
--- a/mingw-w64-mlpack/PKGBUILD
+++ b/mingw-w64-mlpack/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mlpack
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.4.2
-pkgrel=1
+pkgrel=2
 pkgdesc='An intuitive, fast, scalable C++ machine learning library (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


Biswa96:
An explanation of the change as a commit message would be appreciated.

jeremyd2019:
> adding cotire as makedepends will conflicts with cmake.

The cotire package depends on a specific version of cmake, as it installs into a directory with the cmake version number in it.  Cmake was updated since cotire package was.

MehdiChinoune:
> adding cotire as makedepends will conflicts with cmake.

I don't see any conflict, cotire just needs to be rebuilt against the new cmake 3.22

>i think mlpack uses its internal cotire anyway..

We try always to link packages against the system libraries.
- to reduce size
- System libraries are in general adapted to work better on MinGW

raedrizqie:
ok.. i will try updating cotire first

MehdiChinoune:
> cereal is listed as dependency on https://github.com/mlpack/mlpack

It is not, This is for master branch. It was added after 3.1.1 release.
see:
https://github.com/mlpack/mlpack/tree/mlpack-3.1.1

MehdiChinoune:
This PR don't fix anything, or add something new. I think It should be closed.

raedrizqie:
ok

PR #10231, opened 2021-12-02
GHA/CI: some changes

MehdiChinoune:
- Move to windows-2022. I think we still have it installed differently than autobuild repo. This change will speedup the action a little bit ~1minute.
- Add clang32 and clangarm64 repo only when needing them.
- Clean up runner only with CLANGARM64 (I could not find a way to detect a self-hosted runner).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,11 +15,11 @@ jobs:
       fail-fast: false
       matrix:
         include: [
-          { msystem: MINGW64, toolchain: mingw-w64-x86_64-toolchain, can-fail: false, runner: windows-latest },
-          { msystem: MINGW32, toolchain: mingw-w64-i686-toolchain, can-fail: false, runner: windows-latest },
-          { msystem: UCRT64, toolchain: mingw-w64-ucrt-x86_64-toolchain, can-fail: false, runner: windows-latest },
-          { msystem: CLANG64, toolchain: mingw-w64-clang-x86_64-toolchain, can-fail: true, runner: windows-latest },
-          { msystem: CLANG32, toolchain: '', can-fail: true, runner: windows-latest },
+          { msystem: MINGW64, toolchain: mingw-w64-x86_64-toolchain, can-fail: false, runner: windows-2022 },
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -94,7 +95,7 @@ jobs:
           path: C:/_/artifacts/*.pkg.tar.*
 
       - name: "Clean up runner"
-        if: ${{ always() }}
+        if: ${{ matrix.msystem == 'CLANGARM64' }}
```
jeremyd2019, line 98: I think this still needs always() condition in there so it runs on failed or canceled too.  Maybe
```suggestion
        if: ${{ always() && matrix.msystem == 'CLANGARM64' }}
```

.github/workflows/main.yml
```diff
@@ -46,15 +46,12 @@ jobs:
...
-          release: ${{ runner.arch != 'ARM64' }}
-          location: 'D:\M'
+          release: false
 
       - name: Add staging repo
         shell: msys2 {0}
         run: |
           cp /etc/pacman.conf /etc/pacman.conf.bak
-          grep -qF '[clang32]' /etc/pacman.conf || sed -i '1s|^|[clang32]\nInclude = /etc/pacman.d/mirrorlist.mingw\n|' /etc/pacman.conf
-          grep -qF '[clangarm64]' /etc/pacman.conf || sed -i '1s|^|[clangarm64]\nInclude = /etc/pacman.d/mirrorlist.mingw\n|' /etc/pacman.conf
```
jeremyd2019, line 57: -1.  I would prefer those to stay, I don't expect they add too much to the process...

MehdiChinoune: I have just moved them.

jeremyd2019: I know, into a block that isn't always run.  I would prefer that they stay in this unconditional block

MehdiChinoune: Why?

MehdiChinoune: Why?

jeremyd2019: #8762 mostly, but having recently read through the dependency handling code in autobuild/msys2-web those cross-repo dependencies probably wouldn't work there either.

PR #10234, opened 2021-12-02
freetype: update to 2.11.1

MehdiChinoune:


MehdiChinoune:
I

naveen521kk:
> I have found many meson-build packages default to "plain" buildtype

I guess that's what the distro's (like us) have to use?

mingw-w64-freetype/PKGBUILD
```diff
@@ -24,16 +24,14 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
...
 fi
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
-             "${MINGW_PACKAGE_PREFIX}-meson"
-             "${MINGW_PACKAGE_PREFIX}-ninja"
-             "${MINGW_PACKAGE_PREFIX}-pkg-config")
+makedepends=("${MINGW_PACKAGE_PREFIX}-meson"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
 options=('strip' '!libtool' 'staticlibs')
-source=(https://download.savannah.gnu.org/releases/freetype/freetype-${pkgver}.tar.xz{,.sig}
+source=(https://download-mirror.savannah.gnu.org/releases/freetype//freetype-${pkgver}.tar.xz{,.sig}
```
Biswa96, line 30: typo `//` 🤔

Biswa96:
If there is no issue I would suggest not to change them. I can not remember the exact packages but I have seen some where the meson.build applies some extra compiler & linker flags based on release mode. For example, `-fstack-xyz`.

MehdiChinoune:
@Biswa96 Please stop commenting on my PRs or issues.

MehdiChinoune:
> > I have found many meson-build packages default to "plain" buildtype
> 
> I guess that's what the distro's (like us) have to use?

Is that's the case, then we should change the ones with release build-type to plain

lazka:
I only know of glib/gtk which depending on the build type add some extra build flags. I'm fine with letting upstream decide what a "release" build is, even if meson docs recommend something different, so this seems fine with me.

lazka:
@MehdiChinoune if you think someone is de-railing your PR or if you think something is off topic please ask them to create a new issue/discussion instead, or leave it up to the PR reviewer. If you think that a discussion is somehow blocking a merge decision feel free to notify me or anyone else to give feedback/make a decision.

Either way, it would be unwise to ignore valuable input from other experienced community members.

PR #10236, opened 2021-12-02
mingw-w64-python-pyproject2setuppy -

JPeterMugaas:
new package
mingw-w64-python-tomli
  rewrote to use pyproject2setuppy
mingw-w64-python-tomli-w
  new package


Biswa96:
The `mingw_arch` array is required to select $MSYSTEM before building packages.

jeremyd2019:
Where was this yesterday when I was fighting the circular dependency between python-build and python-tomli? :wink:

naveen521kk:
Thanks, can you make a corresponding [PKGBUILD.template](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-PKGBUILD-templates), so that one can refer to it in the future.

naveen521kk (approved):
Looks good thanks

PR #10239, opened 2021-12-03
rust: update to 1.57.0

MehdiChinoune:


lazka:
did some basic tests, seems to work

thanks

PR #10250, opened 2021-12-03
python-nuitka: Update to 0.6.18

podsvirov:


podsvirov:
On my fork `Actions` page I get a error:

```
Invalid workflow file : .github/workflows/main.yml#L4
You have an error in your yaml syntax on line 4
```

Something wrong with `github.ref` context variable?

Cc: @lazka

MehdiChinoune:
For some reason GHA stopped accepting any syntax related to `concurrency` https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency.
Its documentation still there, so assume it is still supported.
I think we should wait until tomorrow morning (Today evening US--AU) and see what's going on.

podsvirov:
@MehdiChinoune, thanks for your reply. Looks like GHA is up again.

PR #10253, opened 2021-12-04
field3d: Fix to build against the latest OpenEXR

MehdiChinoune:
https://github.com/imageworks/Field3D/issues/101


jeremyd2019:
Could we figure out some way to combine with #10119 and take care of clang build at the same time?

MehdiChinoune:
> Could we figure out some way to combine with #10119 and take care of clang build at the same time?

Is it accepted? If yes, then I could.

PR #10254, opened 2021-12-04
[new-package] tbb 2021.4.0

MehdiChinoune:
Arch Linux package : https://archlinux.org/packages/testing/x86_64/tbb/

### DIFF
--- a/mingw-w64-tbb/001-fix-build-with-mingw.patch
+++ b/mingw-w64-tbb/001-fix-build-with-mingw.patch
@@ -1,0 +1,78 @@
+--- a/src/tbb/allocator.cpp
++++ b/src/tbb/allocator.cpp
+@@ -27,11 +27,11 @@
+ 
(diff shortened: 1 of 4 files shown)


MehdiChinoune:
I need it to update openimageio to 2.3.10.0 as It fails to link with the intel-tbb.

lazka:
hm, this replaces intel-tbb but doesn't provide the same DLLs so will break all reverse deps. Should we remove the "replaces" for now?

MehdiChinoune:
> hm, this replaces intel-tbb but doesn't provide the same DLLs so will break all reverse deps. Should we remove the "replaces" for now?

I think some of those DLLs could provide the old one, I will try to get a solution in the coming days.

MehdiChinoune:
For now, you could remove 'replaces'.

lazka:
ok

PR #10265, opened 2021-12-04
opencv: update to 4.5.4

MehdiChinoune:
Before someone complains about AVX* stuff. Those options are enabled by default https://github.com/opencv/opencv/blob/4.5.4/cmake/OpenCVCompilerOptimizations.cmake#L310, I have just added that option to disable AVX512_SKX for CLANG64, As it failed for some reason.

### DIFF
--- a/mingw-w64-ceres-solver/PKGBUILD
+++ b/mingw-w64-ceres-solver/PKGBUILD
@@ -4,14 +4,14 @@ _realname=ceres-solver
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.0.0
-pkgrel=2
+pkgrel=3
 pkgdesc="A large scale non-linear optimization library (mingw-w64)"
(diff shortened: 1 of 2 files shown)


MehdiChinoune:
dll names changed -> Rebuild of many large packages!

lazka:
do we need https://github.com/opencv/opencv/pull/20920 ?

MehdiChinoune:
> do we need [opencv/opencv#20920](https://github.com/opencv/opencv/pull/20920) ?

I don't know, but I will backport it anyway.

MehdiChinoune:
ceres-solver needs a rebuild, opencv is complaining:
```
Failed to find Ceres - Found Eigen dependency, but the version of Eigen found (3.4.0) does not exactly match the version of Eigen Ceres was compiled with (3.3.9). This can cause subtle bugs by triggering violations of the One Definition Rule. See the Wikipedia article http://en.wikipedia.org/wiki/One_Definition_Rule for more details
```

PR #10267, opened 2021-12-05
ruby: Update to 3.0.3

Biswa96:


Biswa96:
@jeremyd2019 Does the mruby patch look correct?

jeremyd2019:
No.  At the least the quote should be before the `$`, but that regex is tricky and I don't think that would be sufficient.

mingw-w64-mruby/proper-line-statements.patch
```diff
@@ -1,11 +1,11 @@
...
-     File.write(t.name, content)
-   end
+--- a/mrbgems/mruby-compiler/mrbgem.rake
++++ b/mrbgems/mruby-compiler/mrbgem.rake
+@@ -29,7 +29,7 @@
  
+   def replace_line_directive(path)
+     content = File.read(path).gsub(%r{
+-      ^\#line\s+\d+\s+"\K.*$ |                  # #line directive
++      ^\#line\s+\d+\s+"\K.*$" |                 # #line directive
```
jeremyd2019, line 8: > @jeremyd2019 Does the mruby patch look correct?

No. At the least the quote should be before the $, but that regex is tricky and I don't think that would be sufficient.

The `\K` resets the start of the match, so you would still wind up just replacing everything after it, which includes the `"` whether or not it is explicitly matched.  The least invasive fix I came up with was to append the closing quote to the replacement, but another option would be to rewrite the regex to use capture groups instead, but I wasn't interested in exploring the ruby syntax to do that.

Biswa96: Let me discuss this in upstream repository.

Biswa96:
I've reported the issue here https://github.com/mruby/mruby/issues/5591. Then attached the patch here. Please review this pull request.

jeremyd2019:
Yep, the upstream fix seems right. positive lookahead, doesn't add the quote to the match, so the replace leaves it alone.

PR #10281, opened 2021-12-06
Libgsf: add read support for zip64 property

salowenh:


Biswa96:
If this patch file is a commit or merge request from upstream please add the link as a comment above the patch command. It will help to track the patch file in future release.

salowenh:
> If this patch file is a commit or merge request from upstream please add the link as a comment above the patch command. It will help to track the patch file in future release.

added

PR #10285, opened 2021-12-07
ncurses 6.3 - update to l newer version, update PGP key, and use sepa…

JPeterMugaas:
…rate paths for all builds.

### DIFF
--- a/mingw-w64-ncurses/PKGBUILD
+++ b/mingw-w64-ncurses/PKGBUILD
@@ -64,7 +64,9 @@ build() {
     --enable-sp-funcs \
     --enable-term-driver \
     --enable-interop \
-    --enable-widec
+    --enable-widec \
+    --enable-pc-files \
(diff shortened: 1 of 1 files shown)


lazka:
The clang builds lost their pkg-config files with this.

JPeterMugaas:
I don't see the problem here because those *.pc files are built when I do "MINGW_ARCH=CLANG64 makepkg-mingw -f" here.  Am I missing something?

Biswa96:
Try to add `--enable-pc-files` option with configure command.

lazka:
That somehow seems to have fixed the problem. Thanks!

naveen521kk:
Hmm, I think this made our python fail to build the `curses` module https://github.com/msys2-contrib/cpython-mingw/runs/4492454357?check_suite_focus=true#step:5:1428.

PR #10287, opened 2021-12-07
new package: python-pywinpty 1.1.6

raedrizqie:


mingw-w64-python-pywinpty/pywinpty-1.1.6-mingw-fix.patch
```diff
@@ -0,0 +1,87 @@
+diff -urN pywinpty-1.1.6-orig/build.rs pywinpty-1.1.6/build.rs
+--- pywinpty-1.1.6-orig/build.rs	2021-11-16 03:47:15.000000000 +0800
```
lazka, line 2: Why is this all needed?

raedrizqie: had to disable conpty detection check because it will pass and set both `ENABLE_CONPTY=1` and `ENABLE_WINPTY=1`

lazka: what problem does that cause?

raedrizqie: build failure because it tries to link with conpty lib which is not present.. so, we have to use the winpty fallback

raedrizqie: @lazka can we merge this? I want to update python-terminado and tornado.. thx

Biswa96: you can add the commit here to update that package. It'll also help to detect any build issue before building in autobuild.

PR #10299, opened 2021-12-09
arpack: switch to cmake (ninja) and create split package with 32-bit and 64-bit indexing

3rav:
This is base on: https://github.com/msys2/MINGW-packages/pull/10094

### DIFF
--- a/mingw-w64-arpack/0002-cmake-set-SUFFIX64.patch
+++ b/mingw-w64-arpack/0002-cmake-set-SUFFIX64.patch
@@ -1,0 +1,71 @@
+diff -urN arpack-ng-3.8.0_orig/CMakeLists.txt arpack-ng-3.8.0/CMakeLists.txt
+--- arpack-ng-3.8.0_orig/CMakeLists.txt	2020-12-07 11:40:45.000000000 +0100
++++ arpack-ng-3.8.0/CMakeLists.txt	2021-12-09 09:55:07.736791800 +0100
+@@ -709,6 +709,10 @@
+ configure_file(PARPACK/SRC/MPI/parpack.pc.in "${PROJECT_BINARY_DIR}/PARPACK/SRC/MPI/parpack${LIBSUFFIX}.pc" @ONLY)
+ configure_file(EXAMPLES/MATRIX_MARKET/arpackSolver.pc.in "${PROJECT_BINARY_DIR}/EXAMPLES/MATRIX_MARKET/arpackSolver.pc" @ONLY)
+ 
++if (INTERFACE64)
++  set(SUFFIX64 64)
++  set(SUFFIX64_UNDERSCORE _64)
++endif()
+ 
+ install(TARGETS arpack
+     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+@@ -727,41 +731,41 @@
+ endif ()
+ 
+ if(ICB)
+-  install(FILES ICB/arpack.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack)
+-  install(FILES ICB/arpack.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack)
++  install(FILES ICB/arpack.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack${SUFFIX64})
++  install(FILES ICB/arpack.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack${SUFFIX64})
+   if (MPI)
+-    install(FILES ICB/parpack.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack)
+-    install(FILES ICB/parpack.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack)
++    install(FILES ICB/parpack.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack${SUFFIX64})
++    install(FILES ICB/parpack.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack${SUFFIX64})
+   endif()
+   if (ICBEXMM)
+-    install(FILES EXAMPLES/MATRIX_MARKET/arpackSolver.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack)
++    install(FILES EXAMPLES/MATRIX_MARKET/arpackSolver.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/arpack${SUFFIX64})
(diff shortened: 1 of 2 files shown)


Biswa96:
In my opinion, separate commits will be easier to review. For example, one for cmake switch and another for split packages.

mingw-w64-arpack/PKGBUILD
```diff
@@ -1,67 +1,152 @@
 # $Id$
-# Maintainer (ArchLinux): Alexander Rødseth <rodseth@gmail.com>
+# Maintainer (ArchLinux): Alexander Rodseth <rodseth@gmail.com>
```
mmuetzel, line 2: Why did you change their spelling?

mingw-w64-arpack/PKGBUILD
```diff
@@ -1,67 +1,152 @@
...
 url='https://forge.scilab.org/index.php/p/arpack-ng/'
 license=('BSD')
-depends=("${MINGW_PACKAGE_PREFIX}-gcc-libgfortran"
+depends=($([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-gcc-libgfortran")
          "${MINGW_PACKAGE_PREFIX}-openblas")
-makedepends=("${MINGW_PACKAGE_PREFIX}-gcc-fortran")
-provides=("${MINGW_PACKAGE_PREFIX}-arpack-ng")
+if [ "$CARCH" = "x86_64" ]; then
+	depends+=("${MINGW_PACKAGE_PREFIX}-openblas64")
+fi
```
mmuetzel, line 22: Would the same syntax you used for conditionally depending on libgfortran work here, too?

mingw-w64-arpack/PKGBUILD
```diff
@@ -1,67 +1,152 @@
...
+makedepends=($([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-gcc-libgfortran")
```
mmuetzel, line 23: Is this a typo? Did you want to write `gcc-fortran`?
Is it necessary to make-depend on packages that are already in the `depends` listing?

mingw-w64-arpack/PKGBUILD
```diff
@@ -1,67 +1,152 @@
...
 pkgver=3.8.0
-pkgrel=2
+pkgrel=3
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 pkgdesc="Fortran77 subroutines designed to solve large scale eigenvalue problems (mingw-w64)"
 url='https://forge.scilab.org/index.php/p/arpack-ng/'
 license=('BSD')
-depends=("${MINGW_PACKAGE_PREFIX}-gcc-libgfortran"
+depends=($([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-gcc-libgfortran")
```
mmuetzel, line 18: `arpack` needs a Fortran compiler to be built.
Would it be better to let the CI fail early because of the missing dependency instead of making it conditional?

mingw-w64-arpack/0002-cmake-set-SUFFIX64.patch
```diff
@@ -0,0 +1,71 @@
+diff -urN arpack-ng-3.8.0_orig/CMakeLists.txt arpack-ng-3.8.0/CMakeLists.txt
+--- arpack-ng-3.8.0_orig/CMakeLists.txt	2020-12-07 11:40:45.000000000 +0100
++++ arpack-ng-3.8.0/CMakeLists.txt	2021-12-09 09:55:07.736791800 +0100
+@@ -709,6 +709,10 @@
+ configure_file(PARPACK/SRC/MPI/parpack.pc.in "${PROJECT_BINARY_DIR}/PARPACK/SRC/MPI/parpack${LIBSUFFIX}.pc" @ONLY)
+ configure_file(EXAMPLES/MATRIX_MARKET/arpackSolver.pc.in "${PROJECT_BINARY_DIR}/EXAMPLES/MATRIX_MARKET/arpackSolver.pc" @ONLY)
+ 
++if (INTERFACE64)
++  set(SUFFIX64 64)
++  set(SUFFIX64_UNDERSCORE _64)
```
mmuetzel, line 10: Is `SUFFIX64_UNDERSCORE` needed?

mmuetzel (commented):
This isn't a complete review. Just some comments to parts of the changes.

mingw-w64-arpack/PKGBUILD
3rav: Changed accidentally, probably the text editor did not handle special characters "ø"

mingw-w64-arpack/PKGBUILD
3rav: Yes, it is a typo, should be "${MINGW_PACKAGE_PREFIX}-gcc-fortran"

mingw-w64-arpack/0002-cmake-set-SUFFIX64.patch
3rav: You are right `SUFFIX64_UNDERSCORE` is redundant

mingw-w64-arpack/PKGBUILD
3rav: You are right it is better making it condition, but what's the best way to do it?

mingw-w64-arpack/PKGBUILD
3rav: as far as I know the libgfortran is available in 32 bit unlike openblas64 (only 64 bit)

mmuetzel: I meant writing something like
```
$([[ ${CARCH} == x86_64 ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas64"
```
Would that work directly in the `depends` listing as well?

mingw-w64-arpack/PKGBUILD
mmuetzel: Would it make sense to remove the condition to let the CI fail early? (See also the other comment.)

mmuetzel:
@3rav: The last few changes didn't trigger the CI for some reason. But in case the library name is still different from before, don't forget to bump the `pkgrel` of all dependent packages to trigger a rebuild. Afaict, that is only `armadillo` and `octave` currently.

Biswa96:
> The last few changes didn't trigger the CI for some reason

Because there is a merge conflict. In the last commit, autotools was added in makedepends.

mmuetzel:
@3rav: You'll need to rebase your changes.
At that point it might be easiest to `git reset` your branch (make a copy of your changed files first!). Then, fetch and merge from upstream. After that, commit the latest version of your changes and force push the updated branch.
But I only have basic git knowledge. And other approaches might be more efficient.

PR #10309, opened 2021-12-10
flatbuffers: Update to 2.0.5

Biswa96:


Biswa96:
I have some doubts about the 0002 patch file. Can anyone provide any info about the "py" command? Without the 0002 patch I get this type of errror:

```
py scripts/generate_code.py --flatc build_dir/flatc.exe
'py' is not recognized as an internal or external command, operable program or batch file.
```

Are the patch files correct? I shall send those to upstream first.

naveen521kk:
> Can anyone provide any info about the "py" command?

https://docs.python.org/3/using/windows.html#from-the-command-line If you run that on CI then MSVC python installed would be used instead of one installed by Pacman.

raedrizqie:
> > Can anyone provide any info about the "py" command?
> 
> https://docs.python.org/3/using/windows.html#from-the-command-line If you run that on CI then MSVC python installed would be used instead of one installed by Pacman.

yes.. it uses the registry to find the default python

Biswa96:
Thanks for the info. I thought that it somehow comes from find_package() call. Any opinion about those new patches?

naveen521kk:
> Any opinion about those new patches?

Lgtm.

Biswa96:
Well then. I have sent those patches to upstream. Then I added the links of those pull request here.

PR #10322, opened 2021-12-12
libmicrohttpd: fix compilation with clang64

neheb:
llvm-dlltool does not support the -e flag.

### DIFF
--- a/mingw-w64-libmicrohttpd/PKGBUILD
+++ b/mingw-w64-libmicrohttpd/PKGBUILD
@@ -7,10 +7,10 @@ _realname=libmicrohttpd
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.9.72
-pkgrel=3
+pkgrel=4
 pkgdesc="GNU libmicrohttpd is a small C library that is supposed to make it easy to run an HTTP server as part of another application (mingw-w64)"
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I think it could avoid dlltool altogether, and just use `-Wl,--out-implib` instead of output-def.

neheb:
huh. did it even succeed?

neheb:
> 

not sure what you mean.

neheb:
This is now fixed.

dennisameling (approved):
First had to build `libtool` myself as it was only [added an hour ago](https://github.com/msys2/MINGW-packages/commit/9c8d2c7100e13e87a15407ad166329a80ecfea95), but can confirm that building this package is working on my arm64 device and all tests have passed 👍🏼 haven't checked the contents of the added patches though

jeremyd2019:
please rebase to pick up changes resulting from #10338

neheb:
Rebased.

mingw-w64-libmicrohttpd/PKGBUILD
```diff
@@ -10,22 +10,31 @@ pkgver=0.9.72
 pkgrel=3
```
jeremyd2019, line 10: I was afraid of that, pkgrel needs to be bumped again.

neheb: Done.

PR #10325, opened 2021-12-14
Build ceres-solver in 2 variants (with SuiteSparse -> GPL; without -> LGPL)

eudoxos:
Tentative for discussion. See https://github.com/msys2/MINGW-packages/issues/9240 for context.

### DIFF
--- a/mingw-w64-ceres-solver/PKGBUILD
+++ b/mingw-w64-ceres-solver/PKGBUILD
@@ -2,7 +2,10 @@
 
 _realname=ceres-solver
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgname=(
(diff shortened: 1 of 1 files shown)


eudoxos:
The builds are failing because the CI tries to install both packages produced simultaneously (they are marked as conflicting) and there is an expected file conflict.

MehdiChinoune:
> The builds are failing because the CI tries to install both packages produced simultaneously (they are marked as conflicting) and there is an expected file conflict.

Add an empty .ci-sequential file (like https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-curl/.ci-sequential)

eudoxos:
> Add an empty .ci-sequential file (like https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-curl/.ci-sequential)

Thanks, that worked :)

PR #10331, opened 2021-12-14
slim-simulator: update mingw_arch (eliminate 32-bit). Also update source sha256sums.

rdinnager:
The source got updated for slim-simulator so I'd like to trigger a rebuild (BTW is there a way to do this automatically without a PR?). I've updated the sha256sums. Also, the developer has informed me that 32-bit builds are not supported so I've removed them from the mingw_arch vector.

### DIFF
--- a/mingw-w64-slim-simulator/PKGBUILD
+++ b/mingw-w64-slim-simulator/PKGBUILD
@@ -2,17 +2,17 @@ _realname=slim-simulator
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.7
-pkgrel=1
+pkgrel=2
 pkgdesc="SLiM forward simulation software package for population genetics (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
 url="https://messerlab.org/slim/"
 license=("GPL")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-cmake")
 depends=("${MINGW_PACKAGE_PREFIX}-qt5-base")
 source=("http://benhaller.com/slim/SLiM_pacman.tar.gz" )
-sha256sums=("003f095120595b2a206fc2d7a451e2189a56639ae7b85a070b76c4ce13def391")
+sha256sums=("5f9c83709666dc0d367b6d8f04e0c42faa1e1b0e0ea549ae6124046f076754ff")
 
 build() {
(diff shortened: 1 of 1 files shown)


Biswa96:
> is there a way to do this automatically without a PR?

Wondering how? 🤔 

> the developer has informed me that 32-bit builds are not supported

Any source of that statement?

rdinnager:
> > the developer has informed me that 32-bit builds are not supported
> 
> Was this mentioned in any release notes?

I don't think so, but when attempting to compile using a 32-bit arch this error is produced:

```
error: static assertion failed: SLiM must be built for 64-bit, not 32-bit.
```

Also see conversation here: https://github.com/MesserLab/SLiM/issues/66#issuecomment-958095940

I'm sure the primary developer @bhaller will also be happy confirm.

bhaller:
Hi @Biswa96 and @rdinnager.  Yes, I'm the primary developer of SLiM, and can confirm that 32-bit builds are not supported.  It's not in any release notes yet because we've never had a release; this build will be our first release (which I hope can happen today if possible).  But the release notes that I have written up to use when the release does happen say this:

> Windows is now supported as a native platform (i.e., without using the WSL), including SLiMgui, thanks to work by Russell Dinnage.  There is an installer for Windows, using pacman.  Only 64-bit Windows is supported.  See chapter 2 of the manual for installer/build instructions.

It would be great if this could happen ASAP, thanks.  And as @rdinnager asked, is there a way to streamline this process for future releases, or does it always require a PR and an approval process?

Biswa96:
Thanks for the confirming that. I was just asking because there are many users who use 32 bit programs.

> does it always require a PR and an approval process?

The pull request has to be reviewed by a human. The package building is done by CI.

bhaller:
> Thanks for the confirming that. I was just asking because there are many users who use 32 bit programs.

Yes; SLiM is scientific simulation software, and thus requires 64-bit to have sufficient elbow room to play.  :->

> > does it always require a PR and an approval process?
> 
> The pull request has to be reviewed by a human. The package building is done by CI.

I see, thanks.

bhaller:
Hi @Biswa96.  Is this PR still waiting for approval, or is it building?  How do I tell what stage things are at?  Sorry for my cluelessness, I'm new to this process and not sure how it all unfolds.  :->  Thanks!

rdinnager:
> Hi @Biswa96. Is this PR still waiting for approval, or is it building? How do I tell what stage things are at? Sorry for my cluelessness, I'm new to this process and not sure how it all unfolds. :-> Thanks!

Hi @bhaller. The PR needs to be merged before building for real will begin (and then probably not immediately, depends on a number of factors I think, packages are built in 'batches' on some sort of schedule). So it is still waiting for approval.

bhaller:
> Hi @bhaller. The PR needs to be merged before building for real will begin (and then probably not immediately, depends on a number of factors I think, packages are built in 'batches' on some sort of schedule). So it is still waiting for approval.

I see.  OK, thanks.

PR #10333, opened 2021-12-14
freetype: enable for clangarm64

dennisameling:
Needs https://github.com/msys2/MINGW-packages/pull/10332


neheb:
As you have such a machine, can you test https://github.com/msys2/MINGW-packages/pull/10322 ?

jeremyd2019:
If you are going to enable freetype, it would probably also be wise to enable harfbuzz, and then once both are available we can turn harfbuzz support on in freetype.

PR #10334, opened 2021-12-14
nasm: enable for clangarm64

dennisameling:


jeremyd2019:
Why?  AFAIK nasm is just an x86 assembler, not arm/arm64.  Looking to cross-compile from arm host to x86 target?

dennisameling:
Good catch. It was a `makedepends` dependency for [`libjpeg-turbo`](https://github.com/msys2/MINGW-packages/pull/10335). When I remove it as a dependency there, `libjpeg-turbo` still builds fine on my arm64 device - can we somehow make the `makedepends` conditional to only be needed for building x86 targets?

MehdiChinoune:
> Good catch. It was a `makedepends` dependency for [`libjpeg-turbo`](https://github.com/msys2/MINGW-packages/pull/10335). When I remove it as a dependency there, `libjpeg-turbo` still builds fine on my arm64 device - can we somehow make the `makedepends` conditional to only be needed for building x86 targets?

`$([[ ${MSYSTEM} != CLANGARM64 ]] && echo "${MINGW_PACKAGE_PREFIX}-nasm")`

Biswa96:
Will enabling arm64 in nasm create any issue? It may be useful to cross compile stuff, as Jeremy said. ArchLinux ARM has that package https://archlinuxarm.org/packages/aarch64/nasm

jeremyd2019:
> `$([[ ${MSYSTEM} != CLANGARM64 ]] && echo "${MINGW_PACKAGE_PREFIX}-nasm")` or `$([[ ${CARCH} != aarch64 ]] && echo "${MINGW_PACKAGE_PREFIX}-nasm")`

I try to write them so that they always evaluate 'true', because sometimes they're evaluated with `set -e`, so reverse the equality and use `||` instead of `&&`.  I tend to use
`$( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || echo "${MINGW_PACKAGE_PREFIX}-nasm" )` but that's based on things in earlier stages of the clang prefixes being disabled for *all* `*-clang-*`.



> Will enabling arm64 in nasm create any issue? It may be useful to cross compile stuff, as Jeremy said. ArchLinux ARM has that package https://archlinuxarm.org/packages/aarch64/nasm

It should be fine.  I'm just trying to keep number of packages to a minimum until we figure out a more sustainable solution.  Also, to reduce confusion that nasm (or uasm, which was a similar situation for mingw-w64-clang) actually does anything useful  for aarch64 packages.

jeremyd2019:
I'm inclined to leave this one, and use its absence on arm64 to detect packages that need their makedepends tweaked.

PR #10335, opened 2021-12-14
libjpeg-turbo: enable for clangarm64

dennisameling:
Needs #10334 


jeremyd2019:
Maybe something like 7c6f5be544ba1640fc8f06a75af941e59e7dfed4?

PR #10338, opened 2021-12-15
libtool: remove and add standalone libltdl instead

lazka:


lazka:
I don't really know if libtool somehow depends on libltdl, or of it is just included fro libtool users. Let's give it a try..

mingw-w64-libltdl/PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=2.4.6
+pkgrel=21
+pkgdesc="A system independent dlopen wrapper for GNU libtool (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+url="https://www.gnu.org/software/libtool"
+license=("GPL")
+makedepends=("${MINGW_PACKAGE_PREFIX}-autotools" "${MINGW_PACKAGE_PREFIX}-gcc")
```
jeremyd2019, line 15: maybe for transitional this should also depend on msys libtool?  Or do you think it would be better to 'catch' those packages and add `${MINGW_PACKAGE_PREFIX}-autotools` to their makedepends as they fail (once libtool is removed from base-devel)?

lazka: libtool is currently in base-devel, so this shouldn't be a problem.

lazka:
These are the packages currently linking against it:

```
mingw-w64-sox
mingw-w64-libgphoto2
mingw-w64-unixodbc
mingw-w64-xmlsec
mingw-w64-graphicsmagick
mingw-w64-hwloc
mingw-w64-imagemagick
mingw-w64-pulseaudio
mingw-w64-graphviz
mingw-w64-openldap
```

lazka:
I've removed all references to libtool in e581d143b78451

mingw-w64-libltdl/PKGBUILD
lazka: (I should read to the end...) yeah autotools needs to be added everywhere, then base-devel can be cleaned up.

lazka:
all packages rebuilt, mingw libtool is gone

PR #10347, opened 2021-12-16
PDCurses: Update from PDCursesMod 4.1.0 to 4.3.1

GitMensch:
* removed unnecessary patch 003-fix-_bool-was-not-declared.patch
* integrated patch 004 into 001 while updating
* removed install of (now) internal-only header acs_defs.h
* create pdcurses.h during package with the necessary arguments to lead to the correct definitions

additional:
* prepared for building more ports - and added wincon as first one (header and link library are identical)
* don't install test executables (which in the meantime are not build by default any more), but generate them in check for manual tests (may be packaged as pdcurses-demos later)


GitMensch:
For some reasons the patch seems to be in the wrong format. Would be nice if someone could take the time to inspect.

Also "for some reason" the clang builds do the same as the other ones and therefore also fail on the same step - but don't abort like the others?

raedrizqie:
`LIBCURSES` was defined 2 times with different name

GitMensch:
> `LIBCURSES` was defined 2 times with different name

Sure - those are within the conditionals (as it was in the previous patch).
Or did you spot the place where the patchfile is wrong?

raedrizqie:
here:

 ```
ifeq ($(DLL),Y)
 	CFLAGS += -DPDC_DLL_BUILD
 	LIBEXE = $(CC)
-	LIBFLAGS = -Wl,--out-implib,pdcurses.a -shared -o
+	LIBCURSES = $(LIBPREFIX)pdcurses.dll
+	LIBARCHIVE = $(LIBCURSES).a
+	LIBFLAGS = -Wl,--out-implib,$(LIBARCHIVE) -shared -o
 	LIBCURSES = pdcurses.dll
 	LIBDEPS = $(LIBOBJS) $(PDCOBJS)
 	LIBSADDED = -lwinmm
```

GitMensch:
@raedrizqie thanks for the note, I finally have seen what you have clearly said. I've missed that because my question was about the failed patching, which likely will still be identical.

PR #10349, opened 2021-12-17
libvips: update to 8.12.1

neheb:
Fixed lld compilation by using autoreconf. Also added openmp as it's
needed to build. Added other clang platforms.


jeremyd2019:
Please don't add clangarm64 for this.  The dependencies are not there yet.

jeremyd2019:
> Can't exec "gtkdocize": No such file or directory

You probably need to makedepend on `${MINGW_PACKAGE_PREFIX}-gtk-doc` (or sometimes msys gtk-doc works too if it's only needed for autoreconf)

neheb:
Done.

jeremyd2019:
I was suspicious, but I confirmed nothing linked to libomp.dll, at least in clang32 so I guess makedepends is right for that.  I don't think that openmp should be necessary for non-clang, but I also don't think it necessarily matters enough to add it only for clang.

neheb:
yeah seems to be some clang requirement. libvips CI passes -fopenmp: https://github.com/libvips/libvips/blob/master/.github/workflows/ci.yml#L107

PR #10355, opened 2021-12-17
librsvg: Update to version 2.52.5

mmuetzel:
Update version and remove patches that have been applied upstream.


lazka:
I'll look into removing the last patch

PR #10370, opened 2021-12-19
SUNDIALS: Update to 6.0.0.

gharveymn:
Note that this release removes the F77 interface, therefore the manual usage of `ar`, and the usage of the first patch have both been removed.

The new patch is just to fix the broken compilation of objects relevant to SuperLU.

This is a binary incompatible release, so maybe wait until Octave has updated with SUNDIALS 6.0.0 support to merge this (@mmuetzel maybe give this a ping when ready?).


mingw-w64-sundials/PKGBUILD
```diff
@@ -66,7 +69,6 @@ build() {
     -D BUILD_ARKODE=ON \
     -D BUILD_CVODE=ON \
     -D BUILD_CVODES=ON \
-    -D BUILD_FORTRAN77_INTERFACE=ON \
     -D BUILD_FORTRAN_MODULE_INTERFACE=OFF \
```
MehdiChinoune, line 72:
```suggestion
    -D BUILD_FORTRAN_MODULE_INTERFACE=ON \
```
Since FORTRAN77 is dropped, Fortran 2003 should be enabled

gharveymn: Unfortunately we won't be able to do that because their Fortran 2003 interface requires 64-bit indexing which we cannot enable because the dependencies are built with 32-bit indexing.

MehdiChinoune: > Unfortunately we won't be able to do that because their Fortran 2003 interface requires 64-bit indexing which we cannot enable because the dependencies are built with 32-bit indexing.

I have taken a look at their Fortran 2003 Interfaces, They are using `integer(C_INT)` in their declarations and this shouldn't be a problem at all.
I couldn't find where do they use 64-bit indexing, Could you give me source location (in https://github.com/LLNL/sundials)

MehdiChinoune: I see now https://github.com/LLNL/sundials/blob/master/src/ida/fmod/fida_mod.f90#L829

gharveymn: Right, and then we're also explicitly blocked by the [build system](https://github.com/LLNL/sundials/blob/master/cmake/SundialsBuildOptionsPre.cmake#L161) as well as a [critical source file](https://github.com/LLNL/sundials/blob/c0097c0bfd56e8f94dbb588e0894fdceacf6f737/src/sundials/fmod/fsundials_types_mod.c#L216). So it doesn't seem like they plan on supporting it any time soon.

mmuetzel:
Cross-link to the related upstream report for SUNDIALS 6.0 in Octave:
https://savannah.gnu.org/bugs/?61701

mingw-w64-sundials/PKGBUILD
```diff
@@ -28,28 +28,31 @@ makedepends=('make'
 optdepends=("${MINGW_PACKAGE_PREFIX}-python: for running examples")
 conflicts=("${MINGW_PACKAGE_PREFIX}-cvode")
 source=("${_realname}-${pkgver}.tar.gz"::https://github.com/LLNL/sundials/releases/download/v${pkgver}/${_realname}-${pkgver}.tar.gz
-        '0001-sundials-link-errors.patch'
+#       '0001-sundials-link-errors.patch'
```
lazka, line 31: Why is that patch lept around?

lazka:
(converted to a draft until the Octave issue is solved)

mmuetzel:
@lazka: Thanks for accepting PR #10414.
That means Octave is no longer blocking an update of SUNDIALS 6. Updating it now would require a rebuild of Octave (which should automatically de-activate support for SUNDIALS in Octave for now).
Support for SUNDIALS 6 was added for Octave 7 which is currently planned to be released at some time in early 2022.

mingw-w64-sundials/PKGBUILD
gharveymn: Ah, sorry. I thought I saw obsolete patches being commented out rather than removed in another package, but maybe I'm remembering an Arch package. It has been removed.

gharveymn:
The aforementioned patch for Octave works with this, so this package should be ready. @lazka 

Octave will require a rebuild.

mmuetzel:
Dependent packages will require a rebuild because the SUNDIALS' API has changed. That's only Octave atm (like you already wrote).
@gharveymn: Could you please bump its `pkgrel`?

mmuetzel:
Sorry, I wasn't clearer before. You'll need to bump the `pkgrel` in the `PKGBUILD` file of Octave. Please, keep it at `1` for SUNDIALS.

gharveymn:
I'm an idiot. Thanks for your patience haha.

PR #10374, opened 2021-12-20
mingw-w64-libmicrohttpd: update to 0.9.74

Karlson2k:
* Removed patches (integrated upstream)
* Enabled full testing (curl was fixed many versions ago)
* Upstream defaults to Vista+, no need for CPPFLAGS

### DIFF
--- a/mingw-w64-libmicrohttpd/PKGBUILD
+++ b/mingw-w64-libmicrohttpd/PKGBUILD
@@ -6,37 +6,25 @@
 _realname=libmicrohttpd
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.9.72
-pkgrel=4
+pkgver=0.9.74
+pkgrel=1
 pkgdesc="GNU libmicrohttpd is a small C library that is supposed to make it easy to run an HTTP server as part of another application (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.gnu.org/software/libmicrohttpd"
 license=("LGPL")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
-             "${MINGW_PACKAGE_PREFIX}-autotools"
              "${MINGW_PACKAGE_PREFIX}-pkg-config")
 checkdepends=("${MINGW_PACKAGE_PREFIX}-curl")
 depends=("${MINGW_PACKAGE_PREFIX}-gnutls")
-source=("https://ftp.gnu.org/gnu/libmicrohttpd/${_realname}-${pkgver}.tar.gz"{,.sig}
-        "010-as.patch::https://git.gnunet.org/libmicrohttpd.git/patch/?id=21be8eccb987b6d5fe05d3827ff97c23b8e630cc"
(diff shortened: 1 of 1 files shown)


mingw-w64-libmicrohttpd/PKGBUILD
```diff
@@ -6,37 +6,25 @@
...
-pkgrel=4
+pkgver=0.9.74
+pkgrel=1
 pkgdesc="GNU libmicrohttpd is a small C library that is supposed to make it easy to run an HTTP server as part of another application (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.gnu.org/software/libmicrohttpd"
 license=("LGPL")
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
-             "${MINGW_PACKAGE_PREFIX}-autotools"
```
Biswa96, line 17: I think we need autotools for building this package.

Karlson2k: No configure/automake/other autotools files are rebuilt during the build, no need for installed autotools.
Bash + sed/grep/etc is enough.

Karlson2k: All CI builds succeed. ;)

Karlson2k: The package needs `bash`, `make`, and `pkg-config`. 
`pkg-config` is already listed. Should I add `make`?

jeremyd2019: Keep in mind also that the version of libtool here is patched to properly work with clang, and some packages need to be `autoreconf`ed to get that patched libtool even if no autoconf-related files are patched in the package itself.  However, a quick glance at the diff shows it doesn't seem to have stopped building a DLL, so it's probably not necessary on this package.

lazka: @Karlson2k for simplicity I created a dummy autotools package which just installs everything that _could_ be needed, so you don't have to care about it. It also allows us to remove autotools related packages from base-devel.

There are no plans to remove "make" from base-devel, so not needed right now.

In this case it might be fine (as long as CI is happy) but for future packages just leave it as is please.

Karlson2k: I'm the author of the upstream of this package. So my updates of this `PKGBUILD` are in sync with required tools. :)
I took some of the locally used libtool patches and applied them before making the tarball, that's why additional patching/autoreconf are not needed.

It is very inconvenient that local libtool patches were not sent to libtool upstream. Until patches are integrated upstream, the amount of maintenance work for MinGW will be increased over the time.

lazka: Lame excuses:

1) There hasn't been a release in 7 years
2) I tried creating a savannah account multiple times during the last year but the account creation always fails to send me a confirmation email.

:)

Karlson2k: There is no need for savannah account. ;)

https://www.gnu.org/software/libtool/contribute.html#patches

Sometimes it could be hard to discuss something with libtool authors, but simple patches like support for grep patterns for llvm-dlltool output, should be accepted easily.

Karlson2k:
Thanks!

PR #10379, opened 2021-12-20
libthai: update to 0.1.29

Adsun701:
This updates libthai to 0.1.29.


Biswa96:
If the 0001-fix-autoreconf.patch file is not required it can be removed. The patch file seem to be applied in upstream.

Adsun701:
@Biswa96 fixed

PR #10388, opened 2021-12-21
openblas: More consistent condition for 64-bit idx

mmuetzel:
Use same condition for calling the actual build rule that is used for defining the package name for the 64-bit indexing version of OpenBLAS.

No rebuild is required currently. So, I didn't bump the `pkgrel`. I hope that is ok.


MehdiChinoune:
You could upgrade OpenBLAS to 0.3.19

PR #10397, opened 2021-12-22
libusb-win32: fix compilation with lld

neheb:


neheb:
ping @fauxpark

fauxpark:
https://github.com/msys2/MINGW-packages/pull/6679

lazka:
@fauxpark do you agree?

fauxpark:
No, this will break things. See the description on the PR above where I added this package.

lazka:
Could you go into a bit more detail re what it would break? I'm not familiar with either package sadly.

fauxpark:
Anyone who is used to the official release of dfu-programmer will be expecting to need the libusb-win32 driver installed on any device they wish to communicate with. If this change goes through, they will need to switch the driver to WinUSB, and would no longer be able to use the official release and the MSYS2 version interchangeably as the two driver APIs are incompatible. libusb-win32 is a windows port of libusb 0.1, while WinUSB is compatible with libusb 1.0.

It would certainly be a good idea to drop libusb-win32, but personally I feel that just isn't possible without breaking users' expectations, until dfu-programmer itself switches to building for libusb 1.0.

neheb:
OTOH dfu-programmer seems quite dead upstream. There's also libusb-compat as well. Does that not work?

fauxpark:
`libusb-compat` is just an interface for libusb 0.1 applications to work with libusb 1.0 - you still need the WinUSB driver installed on the device.

neheb:
Decided to fix compilation with lld instead. Simple enough change.

PR #10400, opened 2021-12-22
OpenBLAS: Update to version 0.3.19

mmuetzel:
Update version. Remove patches that have been applied or that have been resolved differently upstream.

The remaining patches mainly touch the autotools build system. We are using cmake to build OpenBLAS. Should those patches be removed (if possible)?


lazka:
> The remaining patches mainly touch the autotools build system. We are using cmake to build OpenBLAS. Should those patches be removed (if possible)?

Yeah, sounds good.

PR #10408, opened 2021-12-23
grpc: update to 1.43.0

Biswa96:
* Rebase abseil patches.
* Update abseil version to 20211102.0.
* Remove automake and autoconf from makedepends.


Biswa96:
The newly created grpc package contains may files from abseil-cpp project. I thought about creating separate package for abseil-cpp but grpc fails to link with it (at least in my setup). Suggestions are welcomed. Please review.

PR #10411, opened 2021-12-24
gmic: update to 3.0.0

kmilos:


Biswa96:
Just questions. 1. How does a checksum of a commit change? 2. Is it possible to use ninja instead of make?

kmilos:
> 1. How does a checksum of a commit change?

Good point, no idea... MSYS runtime/curl changed CRLF handling? GitHub increased the width of short commit hash for this repo? I used `updpkgsums` to download and update.

> 2. Is it possible to use ninja instead of make?

Nope, seems the last man page generation step does not like being run from `cmd.exe` by Ninja and should be fixed upstream:

```
[5/5] Generating man/gmic.1
FAILED: man/gmic.1 C:/msys64/home/kmilos/MINGW-packages/mingw-w64-gmic/src/build-mingw64/man/gmic.1
cmd.exe /C "cd /D C:\msys64\home\kmilos\MINGW-packages\mingw-w64-gmic\src\build-mingw64 && LD_LIBRARY_PATH=C:\msys64\home\kmilos\MINGW-packages\mingw-w64-gmic\src\build-mingw64 C:/msys64/home/kmilos/MINGW-packages/mingw-w64-gmic/src/build-mingw64/gmic C:/msys64/home/kmilos/MINGW-packages/mingw-w64-gmic/src/gmic-2.9.9/src/gmic_stdlib.gmic it C:/msys64/home/kmilos/MINGW-packages/mingw-w64-gmic/src/gmic-2.9.9/src/gmic_stdlib.gmic reference man > C:/msys64/home/kmilos/MINGW-packages/mingw-w64-gmic/src/build-mingw64/man/gmic.1"
'LD_LIBRARY_PATH' is not recognized as an internal or external command,
operable program or batch file.
```

Biswa96:
There is a newer version 3.0.0.

kmilos:
> There is a newer version 3.0.0

Indeed, but I'm not particularly inclined (i.e. time is an issue) to go down the rabbit hole to build and test for the new "NN" features, so wanted to have the last known stable 2.9.x in the repo...

raedrizqie:
to fix clang build, try adding this in prepare:
`sed -e "s| -fno-ipa-sra||g" -i CMakeLists.txt`

and in build:
`LDFLAGS+=" -fopenmp"`

if those still fails, try updating to v3.0.0

kmilos:
> and in build:
> `LDFLAGS+=" -fopenmp"`

This should be handled by CMake's `find_package(OpenMP)`, no?

The `-fno-ipa-sra` is still present in 3.0.0, so will have patch CMakeLists.txt

kmilos:
Argh, it would help if I actually staged the changes... 🤦

kmilos:
Btw, verified locally that it links against libgomp-1.dll/libomp.dll without the additional flag.

PR #10413, opened 2021-12-24
OpenBLAS: Remove patches for autotools build system

mmuetzel:
I didn't bump the pkgrel because the build rules and sources that we use to build OpenBLAS haven't actually changed.
Please, let me know if it should be bumped anyway.


lazka:
In general a pkgrel bump doesn't hurt.. everything is automated anyway.

PR #10415, opened 2021-12-25
silc-toolkit: fix compilation with clang

neheb:
Add clang32 support where this was tested.


lazka:
I wonder why it was added: https://github.com/silc/silc/commit/97ca3ffe0ce65ac0c5fa3274284825537e996c78

Maybe badly written assembler?

PR #10416, opened 2021-12-25
clucene: fix compilation with clang

neheb:
Changed CXX_STANDARD to 98 as 11 requires stricter list initialization.

Switch to build with Ninja. Faster.

Added pthread patch for UCRT.

Added clang32 where this was tested.

### DIFF
--- a/mingw-w64-clucene/010-pthread.patch
+++ b/mingw-w64-clucene/010-pthread.patch
@@ -1,0 +1,10 @@
+--- a/src/shared/CLucene/LuceneThreads.h
++++ b/src/shared/CLucene/LuceneThreads.h
+@@ -7,6 +7,7 @@
+ #ifndef _LuceneThreads_h
+ #define  _LuceneThreads_h
+ 
++#include <pthread.h>
+ 
+ CL_NS_DEF(util)
+ class CLuceneThreadIdCompare;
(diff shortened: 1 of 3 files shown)


neheb:
What a weird error.

raedrizqie:
you can fix it by something like this:

```
--- clucene-core-2.3.3.4-orig/src/shared/cmake/MacroCheckGccVisibility.cmake
+++ clucene-core-2.3.3.4/src/shared/cmake/MacroCheckGccVisibility.cmake
@@ -26,7 +26,7 @@
    
       # gcc on mac just reports: "gcc (GCC) 3.3 20030304 ..." without the patch level, handle this here:
       if (NOT _gcc_version)
-        string (REGEX REPLACE ".*\\(GCC\\).* ([34]\\.[0-9]) .*" "\\1.0" _gcc_version "${_gcc_version_info}")
+        string (REGEX REPLACE ".*\\(Rev[0-9], Built by MSYS2 project\\).* ([0-9]+\\.[0-9]\\.[0-9]) .*" "\\1.0" _gcc_version "${_gcc_version_info}")
       endif (NOT _gcc_version)
    endif (NOT _gcc_version)
    
```

Biswa96:
Or, disable the check entirely.

neheb:
added patch. builds with mingw now.

neheb:
ping

PR #10420, opened 2021-12-27
Add wb32-dfu-updater_cli

wb-Joy:
Add wb32-dfu-updater_cli.


wb-Joy:
image
It should be redownloaded.

PR #10424, opened 2021-12-27
gtk2 + gtk3 (and deps): enable for clangarm64

dennisameling:
This enables both gtk2 and gtk3 for clangarm64. Note that some packages like `gtk2` and `adwaita-icon-theme` take a long time to build as their builds rely heavily on things like `sh`, which run under x64 emulation in MSYS2 on Windows on ARM.

The update to `mingw-64` is necessary because `libopengl32.a` was added a few days ago in https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/745eaf74-7f4f-e89f-cd7f-f7bbfed2664%40martin.st/#msg37405470 🚀 

Will look into `graphviz` and `gtk4` after this PR has been merged and all packages so far have been published.


mingw-w64-json-glib/PKGBUILD
```diff
@@ -33,7 +33,6 @@ build() {
   meson \
     --prefix="${MINGW_PREFIX}" \
     --buildtype plain \
-    -Ddocs=true \
```
dennisameling, line 36: Got `../json-glib-1.6.6/meson.build:1:0: ERROR: Unknown options: "docs"` so I took out the option completely

Biswa96:
If you are looking for opengl library in System32 it has been added recently in mingw-w64. An update in mingw-w64-crt package will fix it. But there is some confusion about implementation of opengl in Windows 10 vs 11, more here https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/745eaf74-7f4f-e89f-cd7f-f7bbfed2664%40martin.st/#msg37405470

Biswa96:
If crt update is required, I would suggest to update all the mingw-w64 packages (headers, tools etc.). See any previous update for example.

dennisameling:
> If crt update is required, I would suggest to update all the mingw-w64 packages (headers, tools etc.). See any previous update for example.

After the CRT update it works indeed, thanks for the hint! Will update the other packages as well 👍🏼

jeremyd2019:
> If you are looking for opengl library in System32 it has been added recently in mingw-w64. An update in mingw-w64-crt package will fix it. But there is some confusion about implementation of opengl in Windows 10 vs 11, more here

See also https://github.com/mstorsjo/llvm-mingw/issues/209 - maybe you could add what your machine has for OpenGL to that thread.

dennisameling:
I have a working `gtk3` now 🎉 Just need to build `adwaita-icon-theme` and `graphviz` but the first is taking ages to build. Will just start the build tomorrow morning and let it do its thing. Exciting progress!

image

mingw-w64-gtk3/PKGBUILD
```diff
@@ -18,8 +18,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
              "${MINGW_PACKAGE_PREFIX}-sassc"
              "${MINGW_PACKAGE_PREFIX}-libxslt")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
-         $( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || \
-              echo "${MINGW_PACKAGE_PREFIX}-adwaita-icon-theme" )
+         "${MINGW_PACKAGE_PREFIX}-adwaita-icon-theme"
```
jeremyd2019, line 21: You're not going to be able to remove these workarounds just yet, until the packages are built.

mingw-w64-shared-mime-info/PKGBUILD
```diff
@@ -8,7 +8,7 @@ pkgver=2.1
 pkgrel=1
```
jeremyd2019, line 8: Bump pkgrel in packages that change beyond just adding to mingw_arch

mingw-w64-vala/PKGBUILD
```diff
@@ -7,14 +7,13 @@ pkgver=0.54.5
...
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://wiki.gnome.org/Projects/Vala"
 license=("LGPL")
 makedepends=("help2man"
              "${MINGW_PACKAGE_PREFIX}-libxslt"
              "${MINGW_PACKAGE_PREFIX}-autotools"
-              $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || \
-              echo "${MINGW_PACKAGE_PREFIX}-vala")) # Remove this workaround after vala available
+             "${MINGW_PACKAGE_PREFIX}-vala"
```
jeremyd2019, line 16: You're not going to be able to remove these workarounds just yet, until the packages are built.

mingw-w64-json-glib/PKGBUILD
```diff
@@ -6,7 +6,7 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.6.6
 pkgrel=1
```
jeremyd2019, line 7: Bump pkgrel in packages that change beyond just adding to mingw_arch

mingw-w64-vala/PKGBUILD
jeremyd2019: Also, you lost a necessary closing parenthesis here.

jeremyd2019:
It looks like mingw-w64-jbigkit needs to makedepend on libtool now.

jeremyd2019:
This might need to be split, at least the update of mingw-w64.  There aren't generally explicit dependencies on these packages, so the things that try to properly order dependencies will not know that the updated mingw-w64 packages are needed.

Biswa96:
I assume adding clangarm64 does not require pkgrel bump.

dennisameling:
@Biswa96 the MINGW64, MINGW32 and UCRT64 CI jobs are failing due to corrupted packages in the cache. Clang is looking all good. I'm only bumping the pkgrels so that the corrupted packages force a rebuild - that seems to work. If you're comfortable merging this PR without fixing those, I'm good with that. Was just trying to get CI green.

mingw-w64-librsvg-2.40/PKGBUILD
```diff
@@ -19,8 +19,8 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
             "${MINGW_PACKAGE_PREFIX}-vala"
             "setconf")
 depends=("${MINGW_PACKAGE_PREFIX}-gdk-pixbuf2"
-         "${MINGW_PACKAGE_PREFIX}-pango"
-         "${MINGW_PACKAGE_PREFIX}-libcroco")
+         $( [[ ${MINGW_PACKAGE_PREFIX} != *-clang-aarch64* && ${MINGW_PACKAGE_PREFIX} != *-clang-i686* ]] || echo "${MINGW_PACKAGE_PREFIX}-libcroco" )
```
dennisameling, line 22: I think it's weird that CI tries to build `librsvg-2.40` on `MINGW64` in the first place as it's only intended for `clang32` and `clangarm64`. Hopefully this conditional will make CI pass

dennisameling: @jeremyd2019 CI is still trying to build `mingw-w64-librsvg-2.40` for `MINGW64`. Any ideas?

jeremyd2019: CI has never been taught about mingw_arch.  Don't worry about it.  I am testing clangarm64.  You don't need to worry about those corrupted packages errors in this case either.

jeremyd2019:
Please reset/force-push back to 1dd8ba7

jeremyd2019:
https://github.com/msys2-arm/MINGW-packages/runs/4670206373?check_suite_focus=true

1. jbigkit  `gcc: error: the clang compiler does not support '-march=x86-64'`
2. gtk3: 
  ```
[807/1557] Generating gtk/gtkresources_h with a custom command
  FAILED: gtk/gtkresources.h 
  "C:\msys64\clangarm64\bin/glib-compile-resources.EXE" "gtk/gtk.gresources.xml" "--sourcedir" "C:/_/mingw-w64-gtk3/src/build-aarch64/gtk" "--sourcedir" "C:/_/mingw-w64-gtk3/src/gtk+-3.24.31/gtk" "--sourcedir" "gtk/theme/Adwaita" "--sourcedir" "gtk/theme/HighContrast" "--sourcedir" "../gtk+-3.24.31/gtk" "--c-name" "_gtk" "--internal" "--generate" "--target" "gtk/gtkresources.h" "--manual-register"
  Assertion failed: ((((rwlock_t *)*rwl)->valid == LIFE_RWLOCK) && (((rwlock_t *)*rwl)->busy > 0)), file ../mingw-w64/mingw-w64-libraries/winpthreads/src/rwlock.c, line 40
  gtk/gtk.gresources.xml: Child process exited with code 1.
   ```

jeremyd2019:
I got that same assertion failure again with gtkresources_c this time

jeremyd2019:
It succeeded on my raspberry pi, going to try again with 'staging' packages (mingw-w64).

jeremyd2019:
That succeeded too.  Maybe there's a race or something that only likes to show up on a 60-core VM :grin:

jeremyd2019:
Now I am hitting a hang in g-ir-scanner, in librsvg-2.40 and then again in gtk2.  It appears to be the same hang (in cygwin) that I've been seeing sometimes with pacman validating sync dbs, that results in debuggers not being able to get the "context" for the main thread, so I cannot see what it's up to.

jeremyd2019:
g-ir-scanner issue seems to be better after a reboot.  🤷

PR #10425, opened 2021-12-27
devcon-git: update and fix compilation

neheb:
binutils is needed to make.

### DIFF
--- a/mingw-w64-devcon-git/PKGBUILD
+++ b/mingw-w64-devcon-git/PKGBUILD
@@ -7,14 +7,16 @@ pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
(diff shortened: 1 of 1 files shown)


Biswa96:
This need a specific commit hash. See mingw-w64-headers-git as example.

neheb:
why not just use codeload?

Biswa96:
What is codeload and how it is used here?

1480c1:
I think codeload was the zip file?

https://stackoverflow.com/questions/60188254/how-is-codeload-github-com-different-to-api-github-com

Biswa96:
> why not just use codeload?

If it downloads latest default/main branch it will create a different binary/executable/package when upstream changes, instead of the specific commit (f0adcda0) in this pull request.

neheb:
URL would be

```
https://codeload.github.com/microsoft/Windows-driver-samples/tar.gz/f0adcda012820b1cd44a8b3a1953baf478029738
```

for current master. replace the last section with any commit id.

Biswa96:
Yeah, something like that. But I have rarely seen that type of URL format 😄

Biswa96:
Generally, these guidelines are followed https://wiki.archlinux.org/title/VCS_package_guidelines

neheb:
done.

PR #10426, opened 2021-12-28
hclient-git: update

neheb:


neheb:
I'm curious. Wouldn't it make more sense to use codeload here?

Biswa96:
This need a specific commit hash. See mingw-w64-headers-git as example.

neheb:
done.

PR #10439, opened 2021-12-29
libgit2-glib: update to 1.0.0.1

albfan:

PR #10440, opened 2021-12-29
gitg: Updated to 41

albfan:
Depends on #10439


Biswa96:
Just FYI, if a package depends on another you can add those changes in one pull request with multiple commits.

albfan:
@Biswa96 I can rework (actually that's how I test locally) I though it can make review harder.

Changing it

albfan:
libgit2-glib 1.0.0.1 update added

PR #10447, opened 2021-12-30
mimalloc: new package

3rav:


mingw-w64-mimalloc/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+
+  ${MINGW_PREFIX}/bin/cmake --build .
+}
+
+package() {
+  cd "${srcdir}/build-${MSYSTEM}"
+  DESTDIR=${pkgdir} cmake --build . --target install
+  install -Dm644 ${srcdir}/${_realname}-${pkgver}/LICENSE ${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/LICENSE 
+
+  #Temporary fix: The DLL gets installed to ../lib/libmimalloc.dll instead of ../bin/libmimalloc.dll
```
Biswa96, line 40: This may be fixed by changing the install directory in CMakeLists.txt file.

mingw-w64-mimalloc/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+  mkdir -p ${pkgdir}${MINGW_PREFIX}/bin/mimalloc-2.0
```
lazka, line 41: do you really want "bin/mimalloc-2.0/libmimalloc.dll"? It's not in PATH that way

3rav: That's not my whim, see default install locatio ../lib/mimalloc-2.0
default location

OK, I will change this:

`option(MI_INSTALL_TOPLEVEL  "Install directly into $CMAKE_INSTALL_PREFIX instead of PREFIX/lib/mimalloc-version" OFF)`

PR #10449, opened 2021-12-30
EDA tools: ghdl, yosys, prjtrellis, nextpnr and openFPGALoader

umarcor:
This is a regular update of some EDA tools.

Although not critical, I saw it's complaining about the license of openFPGALoader: `message: mingw-w64-openFPGALoader: unformatted license ‘Apache-2.0’`. What is the proper string for that case?

### DIFF
--- a/mingw-w64-ghdl/PKGBUILD
+++ b/mingw-w64-ghdl/PKGBUILD
@@ -1,7 +1,7 @@
 _realname=ghdl
 pkgbase="mingw-w64-${_realname}"
 pkgname='__placeholder__'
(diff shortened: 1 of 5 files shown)


lazka:
> Although not critical, I saw it's complaining about the license of openFPGALoader: `message: mingw-w64-openFPGALoader: unformatted license ‘Apache-2.0’`. What is the proper string for that case?

That's fine, Arch has a very limited set of "official" license strings. I think we should move to SPDX instead, but I need to write a proposal first.

lazka:
I wrote something re licenses here: https://github.com/msys2/MSYS2-packages/issues/2776

PR #10452, opened 2021-12-31
GTK4: update to 4.6.0

lb90:


lazka:
python-docutils for rst2man and gi-docgen for gi-docgen

lb90:
Thanks, @lazka!

PR #10456, opened 2022-01-01
New package additions - prerequisite for some gnome software

JPeterMugaas:
1) mingw-w64-python-pygments - upgrade to latest version, added documentation, fixed path build path.
2) mingw-w64-python-gi-docgen - new package\
3) mingw-w64-python-typogriffy - new package
4) mingw-w64-python-wcap-contrast-ratio - new package.


naveen521kk:
> mingw-w64-python-gi-docgen - new package\

It's already in repo: https://packages.msys2.org/base/mingw-w64-gi-docgen I guess both are the same.

JPeterMugaas:
Mine is slightly different because it also has documentation.  For some reason, sphinx doesn't seem to work during builds making reference to the docutils package in the error message.  It works fine on my computer but I don't know why.

mingw-w64-python-gi-docgen/PKGBUILD
```diff
@@ -0,0 +1,76 @@
...
+      --root="${pkgdir}" --optimize=1 --skip-build
+
+  ## (OPTIONAL) This entire section should be removed
+  ## if the package does NOT instal anything in the ${MINGW_PREFIX}/bin directory.
+  for _f in "${pkgdir}${MINGW_PREFIX}"/bin/*-script.py; do
+    # Remove shebang line
+    sed -e '1 { s/^#!.*$// }' -i "${_f}"
+  done
+  # setup.py puts this in the wrong place
+  mkdir -p "${pkgdir}${MINGW_PREFIX}/lib/pkgconfig"
```
lazka, line 69: What's wrong about "share/pkgconfig" ?

mingw-w64-python-gi-docgen/PKGBUILD
```diff
@@ -0,0 +1,76 @@
+# Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
+
+_realname=gi-docgen
```
lazka, line 3: already exist, please drop, or adjust the original one

lazka:
> For some reason, sphinx doesn't seem to work during builds making reference to the docutils package in the error message

See #10457

mingw-w64-python-gi-docgen/PKGBUILD
JPeterMugaas: I thought that the .pc files were supposed to be in lib/pkgconfig as the standard.

ZachBacon: pkg-config looks both in lib/pkgconfig and share/pkgconfig

JPeterMugaas: I removed that unneeded workaround.

JPeterMugaas:
It looks like another bug was uncovered in failed QA test.

Pip Output: mingw-w64-x86_64-python-pygments-2.11.1-1-any.pkg.tar.zst
Here is what it failed.
manim 0.12.0 requires isosurfaces, which is not installed.

That is NOT related to this package but is a separate issue.

PR #10461, opened 2022-01-02
gettext: update to 0.21

lazka:


lazka:
I'm somewhat torn re the libintl relocation patch.

* It no longer applies and libintl now allows wchar_t paths, so it needs to be rwritten
* Relocating a user provided dynamic path is usually not handled by the library and needs to be done by the library user, which is also the argument of upstream, and I agree. It's hacky...

Toromambo:
my new assignment

PR #10475, opened 2022-01-05
libxml2: Include compiled python bindings that ship with package.

JPeterMugaas:


lazka:
While correct in theory, this will lead to file conflicts on install since Python creates those files at runtime :/

LABPBS:
That can be corrected with ``pacman -Syu --overwrite /${MINGW_PREFIX}/lib/python3.9/site-packages/__pycache__/\*``
reference: https://archlinux.org/news/libxml22912-6-update-may-require-manual-intervention/

lazka:
Yeah, I'm worried users will still be confused.

As an alternative we could look through the repo for similar missing cases and handle it when we upgrade to Python 3.10, when these file names change anyway.

lazka:
I've noted this here for the 3.10 update: https://github.com/msys2/MINGW-packages/issues/10515

Closing this until then.

PR #10477, opened 2022-01-05
helics: update to 3.1.1

raedrizqie:


Biswa96:
Instead of sed commands, a patch file may be better option.

neheb:
how strange that alligned_alloc is not supported. It's part of C++17.

raedrizqie:
> how strange that alligned_alloc is not supported. It's part of C++17.

it is guarded with ``_LIBCPP_USING_IF_EXISTS``
>D:/M/msys64/clang64/include/c++/v1/cstdlib:158:1: note: using declaration annotated with 'using_if_exists' here
  using ::aligned_alloc _LIBCPP_USING_IF_EXISTS;

is there any way to come around this?

jeremyd2019:
https://github.com/chriskohlhoff/asio/issues/922

PR #10478, opened 2022-01-05
Add mingw-w64-c-toolchain package

lazka:
This meta package depends on all packages that are needed to build a executable
with $CC during packaging.

Packages can makedepend on this to get a C compiler.

### DIFF
--- a/mingw-w64-c-toolchain/PKGBUILD
+++ b/mingw-w64-c-toolchain/PKGBUILD
@@ -1,0 +1,24 @@
+# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
+_realname=c-toolchain
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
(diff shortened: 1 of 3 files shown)


Biswa96:
Is it required to add this meta package in every packages as makedepends?

lazka:
> Is it required to add this meta package in every packages as makedepends?

That's the idea. Then we could remove the requirement for installing "mingw-w64-x86_64-toolchain".

Biswa96:
That will be a huge change 😟 toolchain and base-devel are used in many docs and CI scripts.

lazka:
They can stay around, it's only for our CI and autobuild.. and only once enough packages have it.

lazka:
c-toolchain is

```
Total Download Size:    46.85 MiB
Total Installed Size:  393.42 MiB
```

vs the current toolchain group:

```
Total Download Size:    151.34 MiB
Total Installed Size:  1000.17 MiB
```

lazka:
handled via b845c4cc40a43c60 instead now.. similar result.

PR #10480, opened 2022-01-05
pupnp: update to 1.4.12

neheb:
Switch to CMake compilation. Faster and less buggy.

Added testing support.

Remove outdated patches.


Biswa96:
pupnp installs DLLs in lib folder instead of bin folder.

lazka:
the static libs have a s suffix still.. not sure if that is a problem

Biswa96:
There was a similar situation with brotli's static libs. Alexey added a workaround by renaming it 💡

neheb:
At this point it's probably better to replace the patch with a sed script...

neheb:
Updated.

lazka:
thx

Biswa96:
The 010-msvc.patch and 020-fix-install-dir.patch are in upstream now with two mingw64 and mingw32 builds in CI.

PR #10481, opened 2022-01-06
gimp: update to 2.10.30

raedrizqie:


lazka:
Why the ffmpeg change?

raedrizqie:
> Why the ffmpeg change?

gegl depends on ffmpeg dll at runtime (one of it is ff_save.dll), not just buildtime.. gimp fails to build because of this..

lazka:
ok, thanks. I looked like it would maybe survive the plugin not loading, I guess not then :)

PR #10484, opened 2022-01-06
GTK3: update to 3.24.31+44+g0e09bb75b5

lb90:
Contains an important [fix](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4327) for the upcoming Inkscape release and a [crash fix](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4326) for Horizon EDA.

I have compared this with the previous package and everything looks right to me 🙂

### DIFF
--- a/mingw-w64-gtk3/PKGBUILD
+++ b/mingw-w64-gtk3/PKGBUILD
@@ -3,14 +3,16 @@
 _realname=gtk3
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}" "${MINGW_PACKAGE_PREFIX}-gtk-update-icon-cache")
-pkgver=3.24.31
+pkgver=3.24.31+44+g0e09bb75b5
 pkgrel=1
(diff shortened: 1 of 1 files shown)


lazka:
should we encode the revision in the version, like here: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-emacs-pdf-tools-server/PKGBUILD#L6 ?

lb90:
Yep 👍

lb90:
Should I bump pkgrel?

lb90:
I guess not: https://github.com/archlinux/svntogit-packages/commits/packages/gtk3/trunk

lb90:
Ok, here it is! Let me know if there's any other issue. Thanks!

Biswa96:
> Should I bump pkgrel to 2 or not?

Let's recite the holy ArchLinux wiki[^1]. Quote: "When a new version comes out, the release count resets to 1."

Yes, you're right. 

[^1]: https://wiki.archlinux.org/title/Arch_package_guidelines

PR #10485, opened 2022-01-06
gdal: update to 3.4.1

raedrizqie:


mmuetzel:
It looks like GDAL 3.4.1 was released a few days ago:
http://download.osgeo.org/gdal/

Could you update to that version?

Biswa96:
p.s. I have added `-std=c++17` in gdal package for poppler. If gdal version 3.4.0 or higher enables c++17 by default feel free to remove that CXXFLAGS override.

PR #10496, opened 2022-01-07
mingw-w64-mesa: Check driconf feature requirements

pal1000:
Ref: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12158

Also ${MINGW_PACKAGE_PREFIX}-pkg-config is part of ${MINGW_PACKAGE_PREFIX}-toolchain so don't bother installing it.


jeremyd2019:
> Also ${MINGW_PACKAGE_PREFIX}-pkg-config is part of ${MINGW_PACKAGE_PREFIX}-toolchain so don't bother installing it.

@lazka how does this fit with your plans for makedepends/toolchains?

lazka:
yes, toolchain is going away in CI

raedrizqie:
> yes, toolchain is going away in CI

can we just use msys's pkgconf instead? ..and add pkgconf in msys base?

lazka:
> can we just use msys's pkgconf instead? ..and add pkgconf in msys base?

That's not compatible. We could make meson depend on pkg-config, since one could argue that it's not very useful without it.

PR #10505, opened 2022-01-08
EDA: fix nextpnr runtime deps; update nextpnr and iverilog

umarcor:
nextpnr now requires openmp at runtime. This PR updates the recipe. By the way, nextpnr and iverilog are bumped.


umarcor:
@lazka, I believe this is ready to merge now.

PR #10513, opened 2022-01-09
Update typing extension

naveen521kk:


naveen521kk:
I followed [archlinux here](https://github.com/archlinux/svntogit-community/blob/packages/python-typing_extensions/trunk/PKGBUILD) but the tool `python-install` doesn't support `--prefix`/`--root` arguments so changed to `pip` as a temporary workaround. See https://github.com/FFY00/python-install/issues/8

Also, `pyproject2setuppy` doesn't work as expected for this package (not sure why), that's why using `build` + `pip` to build the package here.

lazka:
I don't think wheels support any kind of prefix/root. They just get extracted as is.

naveen521kk:
> I don't think wheels support any kind of prefix/root. They just get extracted as is.
> I wonder what pip does in that case. I'd guess it just gets ignored.

I see. I think pip uses them to find the directory to which they are extracted.

lazka:
Yeah, true.

mingw-w64-python-install/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+
+build() {
+  msg "Python build for ${CARCH}"  
+  cd "${srcdir}/python-build-${CARCH}"
+  ${MINGW_PREFIX}/bin/python setup.py build
+}
+
+package() {
+  cd "${srcdir}/python-build-${CARCH}"
+  MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
```
lazka, line 37:
```suggestion
  cd "${srcdir}/python-build-${CARCH}"
  PYTHONOPTIMIZE=1 \
  MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
```

mingw-w64-python-typing_extensions/PKGBUILD
```diff
@@ -26,14 +30,19 @@ prepare() {
 build() {
   msg "Python build for ${CARCH}"
   cd "${srcdir}/python-build-${CARCH}"
-  ${MINGW_PREFIX}/bin/python setup.py build
+  ${MINGW_PREFIX}/bin/python -m build --wheel --skip-dependency-check --no-isolation
 }
 
 package() {
   cd "${srcdir}/python-build-${CARCH}"
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
```
lazka, line 38:
```suggestion
  cd "${srcdir}/python-build-${CARCH}"
  PYTHONOPTIMIZE=1 \
  MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
```

naveen521kk: What does this do? Just curious.

lazka: to generate the same set of pyc files that we had previously

lazka: oh, that now generates .opt-1.pyc, but no .pyc anymore, which is worse. sorry :/

lazka: let's go with the initial case then.. it's unlikely that users will notice.

PR #10517, opened 2022-01-09
hexchat: Add packages

Biswa96:
Please review. @ZachBacon

### DIFF
--- a/mingw-w64-hexchat/0001-meson-add-winpthread.patch
+++ b/mingw-w64-hexchat/0001-meson-add-winpthread.patch
@@ -1,0 +1,14 @@
+--- a/plugins/python/meson.build
++++ b/plugins/python/meson.build
+@@ -23,8 +23,10 @@
+   install_dir: join_paths(get_option('libdir'), 'hexchat/python')
+ )
+ 
++wpthread_dep = cc.find_library('winpthread')
++
+ shared_module('python', python3_source,
+-  dependencies: [hexchat_plugin_dep, python_dep],
++  dependencies: [hexchat_plugin_dep, python_dep, wpthread_dep],
+   install: true,
+   install_dir: plugindir,
+   name_prefix: '',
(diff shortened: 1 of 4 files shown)


ZachBacon:
The next step is to FHS how it loads plugins, by default it will not load plugins from the level hexchat.exe is in, and then try to load it from %APPDATA%/hexchat. I think the relevant code for plugin loading is in cfgfiles.c  under src/common.

Biswa96:
There are many things to be fixed. This is just beginning. Did you test any package from artifact? I'm little concerned about dependencies.

ZachBacon:
> There are many things to be fixed. This is just beginning. Did you test any package from artifact? I'm little concerned about dependencies.

Not yet, I literally just woke up. Haven't made my way to the computer yet, that's just my first glance view.

ZachBacon:
@Biswa96 you'll have to forgive me, how do I download the artifact?

ZachBacon:
never mind, found out how.

Biswa96:
No problem. From this link https://github.com/msys2/MINGW-packages/actions/runs/1674803415 download the ZIP file > extract .zst > install it with `pacman -U` command.

ZachBacon:
> No problem. From this link https://github.com/msys2/MINGW-packages/actions/runs/1674803415 download the ZIP file > extract .zst > install it with `pacman -U` command.

as I stated, I found out how xD

ZachBacon:
Anyways first glance, you disable dbus. But included dbus-glib as a dependency. Dbus-glib can be removed if dbus is disabled, and cfgfiles.c will need to be patched as well for the plugin location loading.

ZachBacon:
actually I kinda think it's in plugin.c now. It's been a little bit since I looked at the plugin situation

ZachBacon:
Also I the symbols aren't used in some of the plugins, like python No hexchat_plugin_init symbol; is this really a HexChat plugin?

PR #10521, opened 2022-01-10
expat - upgrade to 2.4.2, remove unneeded patch.

JPeterMugaas:
Verified that "make check" tests work.


hartwork (commented):
Hi @JPeterMugaas I'm surprised the patch is no longer needed. How come?

hartwork (commented):
PS: @JPeterMugaas there is still a reference to (and the SHA256 sum of) `002-fix-tests.patch` in the `PKGBUILD` file, it seems.

mingw-w64-expat/PKGBUILD
```diff
@@ -15,7 +15,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cc" "${MINGW_PACKAGE_PREFIX}-autotools")
 options=('strip' 'staticlibs')
 source=("https://github.com/libexpat/libexpat/releases/download/R_${pkgver//./_}/expat-${pkgver}.tar.xz"
         002-fix-tests.patch)
```
1480c1, line 17: It would be necessary to remove this and the sha256sum line

hartwork (commented):
PPS: @JPeterMugaas can I see the effective value of `--host=${MINGW_CHOST}` in the build logs anywhere?

JPeterMugaas:
Working on this stuff now.  The current version does require the patch that basically prevents an attempt to run wine in Windows.  I did not see run.in.sh in the expat main dir in the trunk so that will have to be addressed later when they release a new version.

hartwork:
> Working on this stuff now.

:+1: 

> The current version does require the patch that basically prevents an attempt to run wine in Windows.

Understood, yes.

> I did not see run.in.sh in the expat main dir in the trunk

It is up here: https://github.com/libexpat/libexpat/blob/master/expat/run.sh.in

> so that will have to be addressed later when they release a new version.

I'm involved with libexpat upstream and will be the one releasing the next version (with security fixes) in a few days. So there is a bit of time left to fix things that need fixing, upstream.

JPeterMugaas:
Then the best fix for this situation is to somehow prevent wine from being executed in a native Windows environment.  That's what the patch attempts to do.  My guess is that the tests were assumed to always be a bash shell in a Unix-like environment and merely cross-compiling for Windows

hartwork:
> Then the best fix for this situation is to somehow prevent wine from being executed in a native Windows environment. That's what the patch attempts to do.

Yes!

> My guess is that the tests were assumed to always be a bash shell in a Unix-like environment and merely cross-compiling for Windows

The tests are compiled C binaries. Wrapper `run.sh(.in)`was introduced in commit https://github.com/libexpat/libexpat/commit/49693bb768c431f4861c24fdf6cef7fab3c6351e so that the test suite could be executed on a GNU/Linux host while using the MinGW compiler system on the same host, yes.

JPeterMugaas:
I upset the apple cart because this machine I'm using has Windows 11 on it and not a Unix-like operating system.  I am running bash in Windows.   There's actually quite a number of ways to do this such as:

1) MSYS under Windows - this is what I do
2) Cygwin under Windows
3) Windows Subsystem for Linux 

and who knows what else.

One quick and dirty way of testing would be:

 # MSYSTEM=MINGW32 uname -o
Msys

but Cygwin and WSL probably return something else.

mingw-w64-expat/PKGBUILD
JPeterMugaas: Later when the issue we were discussing is actually fixed in a release version.  Until that happens, things stand as they are.  We' can only cross that bridge when we come to it.

hartwork: @JPeterMugaas using your analogy, do you think https://github.com/libexpat/libexpat/pull/541 will take us to that bridge?

JPeterMugaas: It will.  You still would have to deal with WSL2.  But that bridge can be crossed later because that is unusual.  You have a custom-made Linux kernel running under Windows that runs Windows natively or so I am told.  I don't know first hand.  

I would have to test everything though but I will wait for release.

JPeterMugaas:
@hartwork  The MINGW_CHOST values are like this:

MINGW_ARCH=mingw64  --- MINGW_CHOST=x86_64-w64-mingw32
MINGW_ARCH=mingw32 --- MINGW_CHOST=i686-w64-mingw32
MINGW_ARCH=clang64 --- MINGW_CHOST=x86_64-w64-mingw32
MINGW_ARCH=ucrt64 --- MINGW_CHOST=x86_64-w64-mingw32

mingw-w64-expat/PKGBUILD
hartwork: > It will.

Excellent, I'll go ahead merge that then.

PR #10527, opened 2022-01-10
mesa: libsystre is a required runtime dependency when built with regex

pal1000:
I can't believe nobody complained yet as this has been a problem since 67a20110.
mesa-libsystre-mandatory

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=21.3.3
-pkgrel=2
+pkgrel=3
 pkgdesc="Open-source implementation of the OpenGL specification (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


mingw-w64-mesa/PKGBUILD
```diff
@@ -20,10 +20,10 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-llvm"
              "${MINGW_PACKAGE_PREFIX}-libsystre")
```
jeremyd2019, line 20: You can take it out of makedepends too now that it is in depends.

pal1000: That wouldn't be right because libsystre is actually an [optional make dependency that is used automatically when available](https://gitlab.freedesktop.org/search?search=dep_regex&group_id=1155&project_id=176&scope=&search_code=true&snippets=false&repository_ref=mesa-21.3.3&nav_source=navbar). When not available it's not a runtime dependency either.

jeremyd2019: what I am saying is that when building the package, the union of depends and makedepends are installed before building, so it wouldn't make a practical difference whether it is in makedepends or not once it is in depends.

pal1000: Done as suggested.  Your comment applies to LLVM and vulkan-loader dependencies as well.

PR #10529, opened 2022-01-11
gettext: do include the libintl_*wprintf() functions

dscho:
These functions are declared in `libintl.h`, but not included in `libintl-8.dll`. All programs that use gettext and also call `swprintf()` won't build. See e.g. [this build failure](https://github.com/git-for-windows/git-sdk-64/runs/4756702638?check_suite_focus=true#step:8:672).

Let's fix that.

Fun fact: in v0.19.8.1, those functions were included in the library, but not declared in the header.
In v0.21 the were _not_ included in the library, but they _were_ declared in the header.

So we're actually fixing a long-standing bug here.


dscho:
I submitted the bug fix to the `gettext` project directly, too: https://savannah.gnu.org/bugs/index.php?61794. Let's see what they say ;-)

PR #10530, opened 2022-01-11
bc: update to 1.07.1

1480c1:
Introduces 3 new patches

- 0001-use-rand-on-windows.patch replaces the usage of posix's `random()` with simple `rand()`
- 0002-fix-fix-libmath_h.patch prepends a shell script that lacks a shebang or .sh suffix with sh
- 0003-use-binary-io.patch fixes an issue where due to printf being in TEXT mode, the lines are trailed with `\r\n` instead of `\n` breaking ed along with an issue where due to the buffering, running
    ```bash
    bc -l
    warranty
    ```
    doesn't print anything until you exited the program, unless you used winpty

I have no clue how it will interact with emacs as I don't know how emacs interacts with bc


1480c1:
Also, I have no clue where the actual version controlled sources are for bc as I have searched for svn and git and cvs repos and check Savannah etc. but cannot find it, so I don't know where to upstream these exactly

Biswa96:
I have sent an email to the maintainer asking for the link.

1480c1:
Thanks

PR #10532, opened 2022-01-11
GTK3: update to 3.24.31+56+g7e9935e381

lb90:
Fixes https://gitlab.gnome.org/GNOME/gtk/-/issues/4610


lb90:
I plan to update the package about once a month, but this fix alone is pretty important 🙂

PR #10538, opened 2022-01-12
Meson: update to 0.61.0

lb90:


lb90:
Patches still apply:

```
$ MINGW_ARCH=mingw64 makepkg-mingw
==>  MINGW_ARCH: mingw64
  -> Building mingw64...
==> Making package: mingw-w64-meson 0.61.0-1 (Wed Jan 12 14:21:01 2022)
==> Retrieving sources...
  -> Found meson-0.61.0.tar.gz
  -> Found color-term.patch
  -> Found 0002-Default-to-sys.prefix-as-the-default-prefix.patch
  -> Found 0004-fix-gtk-doc.patch
  -> Found install-man.patch
    meson-0.61.0.tar.gz ... Passed
    color-term.patch ... Passed
    0002-Default-to-sys.prefix-as-the-default-prefix.patch ... Passed
    0004-fix-gtk-doc.patch ... Passed
    install-man.patch ... Passed
==> Extracting sources...
  -> Extracting meson-0.61.0.tar.gz with bsdtar
==> Starting prepare()...
patching file mesonbuild/mlog.py
Hunk #1 succeeded at 37 (offset 6 lines).
patching file mesonbuild/mesonlib/universal.py
Hunk #1 succeeded at 957 (offset 92 lines).
patching file mesonbuild/scripts/gtkdochelper.py
patching file setup.py
Hunk #1 succeeded at 23 with fuzz 2 (offset -13 lines).
==> Starting build()...
```

lb90:
Out of curiosity, were the patches proposed for upstreaming?

lazka:
Rule of thumb, always wait for 0.x.1 with meson :)

* 0002-Default-to-sys.prefix-as-the-default-prefix.patch: very msys2 specific, hard to say what the condition should be
* 0004-fix-gtk-doc.patch: the first part is because we patch gtk-doc to provide .exe instead of .py, so this depends on another patch. The later part I don't know. seems clang specific which doesn't ignore rpath.
* color-term.patch: depends on a patch we have in Python, so it's wrong with official python
* install-man.patch: maybe could be upstreamed, but what conditions should be used? maybe if there is a mingw python...

lazka:
> Rule of thumb, always wait for 0.x.1 with meson :)

Unless you need a specific feature that is only available in the new version, then we can talk of course.

lb90:
> Rule of thumb, always wait for 0.x.1 with meson :)

Yeah, makes sense indeed 🙂 let's wait for the next point release

lazka:
and not to give the wrong impression: some patches could/should probably be upstreamed, but they need more work/thinking.

eli-schwartz:
> Rule of thumb, always wait for 0.x.1 with meson :)

We have managed to break the gnome module in a variety of ways this time, so yeah, waiting might be prudent. There are a bunch of PRs tagged for .1 already.

PR #10542, opened 2022-01-13
imagemagick: dependencies update

raedrizqie:


Biswa96:
A little explanation would be helpful. Also the diff would be clearer if the indentation in depends and makedepends arrays are kept as previous.

raedrizqie:
the reason for this update are:
- clashes between $makedepends/$depends with $optdepends (duplicate entry)
- some packages that depends on imagemagick will fail to detect imagemagick because of missing deps
- most deps should be autodetected by configure (no need to force them in configure) except for rsvg & gslib, because they were disabled by default
- the indentation changes are just for easier auto sorting

mmuetzel:
Just in case I misunderstood: Emacs doesn't detect imagemagick unless the optional dependencies are installed? Are those dependencies also mandatory for imagemagick as a stand-alone package? If they aren't, should emacs depend on those packages instead?

raedrizqie:
> Just in case I misunderstood: Emacs doesn't detect imagemagick unless the optional dependencies are installed? Are those dependencies also mandatory for imagemagick as a stand-alone package? If they aren't, should emacs depend on those packages instead?

its not emacs' fault.. its just that when emacs trying to pull imagemagick as dependencies, the required deps for imagemagick didn't get pulled because (maybe) CI thinks those are just some makedepends and not depends..

this usually happens on clang builds

mmuetzel:
Sorry. It's still not clear to me.
Are these packages required for a stand-alone imagemagick (i.e. without emacs)? Does imagemagick work (as a stand-alone program) when those packages aren't installed on runtime?
If emacs requires that imagemagick provides some of those (otherwise optional) features, should it depend on those optional dependencies instead?

PR #10545, opened 2022-01-13
Add linthesia

albfan:
http://linthesia.sourceforge.net/

music playing


albfan:
Needs alsa (which I suppose is not an option) so probably this needs to use openal or similar?

lazka:
What's the status of this? alsa seems a bit of a dealbreaker, no?

albfan:
No progres on a replacement on Windows. Do we hace other MIDI player to compare?

lazka:
No idea, sry

PR #10548, opened 2022-01-14
ghdl: update to 1.0.0.r964.g70f3c162b

umarcor:
I had to add `*-gcc-ada` as makedepends, because gnatmake would not be found otherwise. I thought that both makedepends and depends were installed before executing `build()`. Did the behaviour change?


MehdiChinoune:
>I thought that both makedepends and depends were installed before executing `build()`.

When depends() is global (not inside another function).

> Did the behaviour change?

previously CI install `toolchain` (which `gcc-ada` is part of) before any build. now It is required to mention the compiler in `makedepends`

PR #10561, opened 2022-01-16
totem-pl-parser: fix clang build

raedrizqie:


Biswa96:
Though the workaround works, it has to be modified in meson.build. If you can wait I would like to propose the change in upstream.

raedrizqie:
ok.. no problem

Biswa96:
Would you like to add a patch file from this merge request https://gitlab.gnome.org/GNOME/totem-pl-parser/-/merge_requests/46 ? It may solve the clang build.

raedrizqie:
> Would you like to add a patch file from this merge request https://gitlab.gnome.org/GNOME/totem-pl-parser/-/merge_requests/46 ? It may solve the clang build.

its ok.. you can generate the patch and submit. thank you

Issue #10547, opened 2022-01-14
[ecppack] Mingw-w64 runtime failure: 32 bit pseudo relocation (...) out of range


umarcor:
Binary `ecppack.exe` from [mingw-w64-prjtrellis](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-prjtrellis/PKGBUILD) is crashing on MINGW64 and UCRT64:

```
# ecppack
Mingw-w64 runtime failure:
32 bit pseudo relocation at 00007FF6E40F66A1 out of range, targeting 00007FFAB9EEDA10, yielding the value 00000003D5DF736B.
```

I'm confused about the 32 bit relocation issue, because both MINGW64 and UCRT64 are expected to be 64 bit environments.

Ref: stnolting/neorv32-setups#11

Biswa96:
The program is crashing in do_pseudo_reloc function, before main function is called. The error output comes from mingw-w64-crt, here https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-crt/crt/pseudo-reloc.c

umarcor:
Yes. It was working for months, and it was broken during the last month.

- A successful run last month: https://github.com/stnolting/neorv32/actions/runs/1561230285
- Builds were broken because nextpnr was missing a dependency: https://github.com/stnolting/neorv32/actions/runs/1589974764
- That was fixed in #10505.
- Since then, ecppack is crashing: https://github.com/stnolting/neorv32/actions/runs/1665427560 (see jobs OrangeCrab or ULX3S)

Biswa96:
If I remove `-DCMAKE_C_COMPILER=clang` and `-DCMAKE_CXX_COMPILER=clang++` options from PKGBUILD file the program works fine in ucrt.

mstorsjo:
To debug this, I'd need to pinpoint exactly which symbol reference, where, in what generated code, is causing this. And that's a bit nontrivial.

To do that, I would need an unstripped exe that is showing this issue, ideally together with all the object files that are input to linking it, plus the corresponding error message (with the fault addresses that are printed in the message for that exact binary).

[...]

PR #10573, opened 2022-01-18
prjtrellis: don't force clang

umarcor:
According to #10547, and tested by several users, using GCC instead of clang does fix the issues with an early check introduced in mingw-w64-crt. This PR changes prjtrellis not to use clang on the default where GCC is the default.

/cc @Biswa96 @stnolting @sylefeb


Biswa96:
Just curious. Was there any reason to use clang in mingw or ucrt environment where gcc is deafult?

umarcor:
No particular reason other than IIRC the default guidelines upstream using clang. I changed the description now to use the default on each environment. I guess the issue (#10547) will remain on clang32 and clang64 only.

jeremyd2019:
the issue was in the combination of clang with BFD ld.  clang+lld should be fine, and is what clang* envs use.

Issue #10577, opened 2022-01-19
xml2-config/xslt-config return unix paths instead of Windows paths


lazka:
python-lxml uses them for example (if this is fixed, we need to make sure lxml still builds fine)

Biswa96:
Like `C:/msys64/ucrt64` instead of `/ucrt64`?

PR #10580, opened 2022-01-19
libxml2: Fix xml2-config to return Windows style path

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/10577


Biswa96:
Does python-lxml require a rebuild?

lazka:
Now that a look at lxml I wonder if it can't just use pkg-config :)

PR #10592, opened 2022-01-21
mingw-w64-python3.10: new package

naveen521kk:
From https://github.com/msys2-contrib/cpython-mingw/tree/mingw-v3.10.0, needs testing


naveen521kk:
Just tested works on a Windows 7 VM.

PR #10601, opened 2022-01-23
f2cblaslapack: Install shared libraries.

mmuetzel:
Also provide package for environments with gcc toolchain.


MehdiChinoune:
f2cblaslapack is just a temporary solution to provide LAPACK for Clang* environments until Flang mature enough. Why build it for others where we have gfortran?

mmuetzel:
You are right. A user will only ever need one single implementation of BLAS/LAPACK. Yet, there are many different implementations available.
I also guess that these libraries won't get widely used (because they are old, probably slow, and maybe buggy). But it might still be useful to have those in more environments. Be it only for debugging possible issues in the clang environments (e.g., by cross-checking if the same issue occurs with f2c BLAS/LAPACK with a GCC toolchain...).

mmuetzel:
@MehdiChinoune: Is this ok to get merged in your opinion?

MehdiChinoune:
> @MehdiChinoune: Is this ok to get merged in your opinion?

I don't think my opinion matters, but I am still against packaging it for non-clang envs.

mmuetzel:
It's not important to me one way or the other.
Updated to no longer package for GCC environments.

MehdiChinoune:
@mmuetzel I think you could backport https://github.com/xianyi/OpenBLAS/commit/d2b5fbf80f02539243cca20b496b0358d2829420 to build OpenBLAS without LAPACK properly.

mmuetzel:
@MehdiChinoune: You are listed as the maintainer of this package. Do you think this PR is ok to be merged in it's current version?

MehdiChinoune:
> @MehdiChinoune: You are listed as the maintainer of this package. Do you think this PR is ok to be merged in its current version?

Not "maintainer", contributor.

mmuetzel:
> You should ask MSYS2 members to approve your changes

I'm interpreting this as you being ok with this change.

I don't know who would be the "correct" member to ask.
@lazka: Ok to merge this?

mingw-w64-f2cblaslapack/PKGBUILD
```diff
@@ -4,32 +4,72 @@ _realname=f2cblaslapack
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=3.4.2.q4
-pkgrel=1
+pkgrel=2
 pkgdesc="f2c BLAS/LAPACK (mingw-w64)"
-provides=("${MINGW_PACKAGE_PREFIX}-lapack")
+provides=($([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] && echo "${MINGW_PACKAGE_PREFIX}-lapack"))
```
lazka, line 9: Does this mean we should expand `mingw_arch` to include gcc environments?

mingw-w64-f2cblaslapack/PKGBUILD
```diff
@@ -4,32 +4,72 @@ _realname=f2cblaslapack
...
 url='https://www.petsc.org/' # Part of petsc
-license=('2-clause BSD')
+license=('spdx:BSD-2-Clause')
 source=("http://ftp.mcs.anl.gov/pub/petsc/externalpackages/${_realname}-${pkgver}.tar.gz"
-        001-no-petsc-conf.patch)
+        001-no-petsc-conf.patch
+        libf2cblas-i686.def
+        libf2cblas-x86_64.def
+        libf2clapack-i686.def
+        libf2clapack-x86_64.def)
```
lazka, line 19: Where are these files coming from? Would be nice to have an explanation in prepare() or something

mingw-w64-f2cblaslapack/PKGBUILD
mmuetzel: @MehdiChinoune preferred to not distribute this package for gcc environments. He has a point in that "normal" users should prefer either the reference implementations or an optimized version of BLAS/LAPACK (i.e., OpenBLAS).
I still think it might be useful in some (edge) cases to have that package for the gcc environments, too. So I'd like if the `PKGBUILD` could still be used to build for those environments locally.
In that case, it shouldn't replace the reference LAPACK for gcc environments.

mingw-w64-f2cblaslapack/PKGBUILD
mmuetzel: I used GCC's `dlltool` to create the base for these files. (Still needed some post-processing like line endings, formatting, ...)
Unfortunately, I didn't get `llvm-dlltool` to produce something similar. But the linkage of those symbols is the same in both environments. So, the files produces in the MINGW64 and MINGW32 environments can be used in the CLANG64 and CLANG32 environments, respectively.
I'll try and summarize that with a few commands in prepare().

mingw-w64-f2cblaslapack/PKGBUILD
lazka: ok, let's wait until someone needs it then

mingw-w64-f2cblaslapack/PKGBUILD
lazka: thanks!

mmuetzel: done.

I hope that is enough to reproduce those files if we ever needed to.

PR #9173, opened 2021-07-19
portaudio: Don't link with uuid; update; port to cmake (msys2#8805)

mmuetzel:
This change should be fixing #8805.

I don't know if this is the expected process to propose a change. Please, let me know if I should do something different instead.

PR #10604, opened 2022-01-23
PortAudio 19.7.0-2 C++ bindings package

volkanorhan:
Adds portaudiocpp package (C++ bindings of portaudio-2.0 library)


Biswa96:
There is already a [portaudio package][1]. Is not this the same project? If yes, is it possible to add the changes in the existing package.

[1]: https://packages.msys2.org/base/mingw-w64-portaudio

mingw-w64-portaudiocpp/PKGBUILD
```diff
@@ -0,0 +1,68 @@
+# Maintainer: Volkan Orhan <volkanorhn@gmail.com>
+
+_realname=portaudiocpp
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+epoch=1
```
MehdiChinoune, line 6: Why?

mingw-w64-portaudiocpp/PKGBUILD
```diff
@@ -0,0 +1,68 @@
...
+pkgver=19.7.0
+_tarver=190700_20210406
+pkgrel=2
```
MehdiChinoune, line 9:
```suggestion
pkgrel=1
```

mingw-w64-portaudiocpp/PKGBUILD
volkanorhan: Because portaudio package doesn't contain C++ bindings

volkanorhan:
> There is already a [portaudio package](https://packages.msys2.org/base/mingw-w64-portaudio). Is not this the same project? If yes, is it possible to add the changes in the existing package.

portaudio package doesn't contain C++ bindings. This package builds only C++ bindings of the portaudio. Of course this can be build together with portaudio but this is a different library than portaudio. On Linux and Homebrew on macOS, portaudiocpp is a different package.

mingw-w64-portaudiocpp/PKGBUILD
MehdiChinoune: I am asking about `epoch=1`, you should read about PKGBUILD syntax.

volkanorhan: OK, I remowed that line since it is not necessary. Thanks Mehdi.

Biswa96:
> On Linux, portaudiocpp is a different package, too.

Generally ArchLinux package naming scheme is followed here. Their portaudio [PKGBUILD][1] file uses `--enable-cxx` option which adds /usr/lib/libportaudiocpp.so file.

Upstream does not have cmake file in cpp bindings. IMO, the cmake file added here should be contributed to upstream first.

[1]: https://github.com/archlinux/svntogit-packages/blob/packages/portaudio/trunk/PKGBUILD

volkanorhan:
> > On Linux, portaudiocpp is a different package, too.
> 
> Generally ArchLinux package naming scheme is followed here. Their portaudio [PKGBUILD](https://github.com/archlinux/svntogit-packages/blob/packages/portaudio/trunk/PKGBUILD) file uses `--enable-cxx` option which adds /usr/lib/libportaudiocpp.so file.
> 
> Upstream does not have cmake file in cpp bindings. IMO, the cmake file added here should be contributed to upstream first.

There is a similar cmake file to be merged on mainstream, and I contributed some enhancements, too, but it will take too much time to have it on stable release version.
Anyway, I am closing this merge request, and I will make portaudio package to use autotools instead of cmake, because stable version doesn't have cmake file for C++ bindings.

Biswa96:
Portaudio was using cmake since this https://github.com/msys2/MINGW-packages/pull/9173.

Biswa96:
Portaudio was using cmake since this https://github.com/msys2/MINGW-packages/pull/9173. Would you like to wait a little please? Let me test this to existing portaudio package.

volkanorhan:
> wait

OK, I am waiting. Thanks @Biswa96

PR #10606, opened 2022-01-24
cx_Freeze 6.10

marcelotduarte:


lazka:
```
cx-freeze 6.10 requires lief, which is not installed.
cx-freeze 6.10 requires wheel, which is not installed.
```

looks like lief is missing an .egg-info

marcelotduarte:
PR #10607 for lief
Updated the depends (wheel)

lazka:
thanks

PR #10626, opened 2022-01-29
i2pd: fix PKGBUILD, update patches

r4sas:
About patches:
* Removed `0001-i2pd-dont-use-pthread_setname.patch ` - it is not really needed, setting of threads names is working and can be used for debugging with gdb on Windows as well
* Changed `0001-i2pd-endian.patch` - Add proposed code instead removing current. If @orignal cofirms that this code is correct, I'll add it to upstream (add: added)

About PKGBUILD:
* Removed `0001-i2pd-dont-use-pthread_setname.patch` patch and changed `make` flags:
  * usage of USE_AESNI is not required because AES-NI availability checked on runtime
  * `USE_MESHNET` make i2pd assume that client always have open ports and direct connetion to other peers. That is incorrect, because in that case router will report incorrect data about availability and didn't uses peer test to do hole punch in NAT covered networks


LABPBS:
Looks good

I barely made a packaging attempt based on my experience building, its great to get feedback from upstream

PR #10634, opened 2022-01-31
Move crt & headers to common target prefix

### LINKED ISSUE #10630: [mingw-w64-headers] why use ${MINGW_PREFIX}/${MINGW_CHOST} prefix ?
Hi,

On Fedora, mingw-headers is installed under the regular target sysroot prefix (for ex `/usr/x86_64-w64-mingw32/sys-root/mingw/`) with the rest of the packages and libraries.

Why is it using a special prefix in msys2?

(fwiw, I am doing an experiment trying to replace the mingw-* packages with a msys2 installation on Linux and bwrap environment, it seems to be doable. has anyone tried that before? that would avoid the proliferation of mingw* packaging and duplication of effort in various distributions, and we could all use msys2 eventually :)

elmarco:
For historical reasons, headers were installed under
${MINGW_PREFIX}/${MINGW_CHOST}, but there is no reason to use a
sub-directory of the target prefix.
    
This align with other mingw target directory structures in Fedora,
Debian ...

Fixes #10630


lazka:
Don't gcc and clang also reference this?

MehdiChinoune:
According to my simple search, there are about 20 packages that should be rebuilt.
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-apr-util/PKGBUILD#L47
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-arm-none-eabi-cc/PKGBUILD#L56
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-assimp/PKGBUILD#L49
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-assimp/PKGBUILD#L59
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-binutils/PKGBUILD#L129
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-cegui/PKGBUILD#L103
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ghostscript/PKGBUILD#L175
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gl2ps/PKGBUILD#L39
....

elmarco:
> hm, don't gcc and clang also reference this?

${MINGW_PREFIX}/${MINGW_CHOST}/ is part of default include & library search path. But the file names used in specs (or elsewhere?) are not absolute, so we can move them to the common target prefix.

elmarco:
> According to my simple search, there are about 20 packages that should be rebuilt. 

Thanks! They don't have to be rebuilt right away, but they may need to be updated. I see some cases where the path shouldn't be hardcoded, but rely on compiler default search path instead.

MehdiChinoune:
> > According to my simple search, there are about 20 packages that should be rebuilt.
> 
> Thanks! They don't have to be rebuilt right away, but they may need to be updated. I see some cases where the path shouldn't be hardcoded, but rely on compiler default search path instead.

No, They should be rebuilt to make sure that this change doesn't break anything.

lazka:
I've asked @Alexpux if he remembers why things are the way they are right now.

elmarco:
> > Thanks! They don't have to be rebuilt right away, but they may need to be updated. I see some cases where the path shouldn't be hardcoded, but rely on compiler default search path instead.
> 
> No, They should be rebuilt to make sure that this change doesn't break anything.

That's a distribution policy, it's not strictly required for end users. Many distributions accept a temporary FTBFS state (fail to build from scratch). This happend every now and then because a new compiler or library dependency regression.

Is there a way to trigger a mass-rebuild of all the packages for testing?

mstorsjo:
> This align with other mingw target directory structures in Fedora, Debian ...

This argument I don't quite understand. When not cross compiling, it's ok to assume that files in `<toolchain>/{include,lib}` are for the current default target, but when cross compiling, the default would be to have those files in `<toolchain>/<cross-triple>/{include,lib}` - just like on Debian, `/usr/bin/x86_64-w64-mingw32-gcc` uses files in `/usr/x86_64-w64-mingw32/{include,lib}` (which aren't mixed in under `/usr/include` etc).

For msys2, I guess this change would be ok, as all toolchains there are mainly non-cross, but I don't see the argument about likeness with how cross compilers are set up in at least Debian (I'm not familiar with Fedora).

elmarco:
> > This align with other mingw target directory structures in Fedora, Debian ...
> 
> This argument I don't quite understand. When not cross compiling, it's ok to assume that files in `<toolchain>/{include,lib}` are for the current default target, but when cross compiling, the default would be to have those files in `<toolchain>/<cross-triple>/{include,lib}` - just like on Debian, `/usr/bin/x86_64-w64-mingw32-gcc` uses files in `/usr/x86_64-w64-mingw32/{include,lib}` (which aren't mixed in under `/usr/include` etc).

On msys2, ${MINGW_PREFIX} is the equivalent of (for ex) Debian `/usr/bin/x86_64-w64-mingw32-gcc`, right?

On Debian, Fedora etc, mingw crt & headers, and various libraries are all installed under the same prefix.

There is no need for an extra sub-directory ${MINGW_PREFIX}/${MINGW_CHOST}.

(having the same directory structure allows to install msys2/mingw packages on Linux and link `/usr/bin/x86_64-w64-mingw32-gcc` -> `/mingw64` (under a user namespace), and you don't need to package all the mingw-* libraries to cross-compile on your Linux distro anymore. Well, that's in theory, I am doing some tests to check that)

mstorsjo:
> On msys2, ${MINGW_PREFIX} is the equivalent of (for ex) Debian `/usr/bin/x86_64-w64-mingw32-gcc`, right?

I presume you mean the equivalent of `/usr/x86_64-w64-mingw32`?

> There is no need for an extra sub-directory ${MINGW_PREFIX}/${MINGW_CHOST}.
> 
> (having the same directory structure allows to install msys2/mingw packages on Linux and link `/usr/bin/x86_64-w64-mingw32-gcc` -> `/mingw64` (under a user namespace), and you don't need to package all the mingw-* libraries to cross-compile on your Linux distro anymore. Well, that's in theory, I am doing some tests to check that)

Ok, sure, that usecase makes sense. And yes, there's a maybe unusal split in msys2 today with the toolchain's files being in `/mingw64/x86_64-w64-mingw32` while third party packages are installed directly in `/mingw64`. Avoiding that can probably useful.

lazka:
Some feedback on mingw-w64 IRC:

```
 09:47 <Biswa96> Is this right thing to do? https://github.com/msys2/MINGW-packages/pull/10634
09:53 <@jon_y> Biswa96: yeah --prefix == sysroot is fine
09:57 <lazka> I wonder if there could be any filename conflicts with random packages
09:57 <lazka> I guess I could write a script to check
09:57 <@jon_y> or packages installing different headers due to bitness difference
09:58 <wbs> yeah in principle it should be fine, but with 2000 packages in msys2, there's certainly something that can break due to excessive assumptions somewhere
```

ok, let's give this a try then.

* There are more packages using that prefix, binutils and winpthreads, etc
* I'd like to check if there are any potential file conflicts with other packages in /lib and /include, just to be sure

elmarco:
> ok, let's give this a try then.
> 
>     * There are more packages using that prefix, binutils and winpthreads, etc
> 
>     * I'd like to check if there are any potential file conflicts with other packages in /lib and /include, just to be sure

thanks, I am not experienced with pacman, but let me know if I can help.

lazka:
Here is a list of file conflicts:

```
mingw-w64-x86_64-freetds /mingw64/include/odbcss.h
mingw-w64-x86_64-glbinding /mingw64/include/KHR/khrplatform.h
mingw-w64-x86_64-hdf4 /mingw64/include/error.h
mingw-w64-x86_64-libgdiplus /mingw64/lib/libgdiplus.a
mingw-w64-x86_64-mesa /mingw64/include/KHR/khrplatform.h
mingw-w64-x86_64-pupnp /mingw64/lib/libupnp.a
mingw-w64-x86_64-unixodbc /mingw64/include/odbcinst.h
mingw-w64-x86_64-unixodbc /mingw64/include/sql.h
mingw-w64-x86_64-unixodbc /mingw64/include/sqlext.h
mingw-w64-x86_64-unixodbc /mingw64/include/sqltypes.h
mingw-w64-x86_64-unixodbc /mingw64/include/sqlucode.h
```

elmarco:
> Here is a list of file conflicts created by the move:
> If anyone has any idea how to handle these, go ahead.

Ok, some ideas below. I suppose ideally you would want to fix all before moving the crt & headers.

> mingw-w64-x86_64-freetds /mingw64/include/odbcss.h

Hi @freddy77 ! the file name conflicts with a windows system header. can it be renamed or moved upstream?

> mingw-w64-x86_64-glbinding /mingw64/include/KHR/khrplatform.h
> mingw-w64-x86_64-mesa /mingw64/include/KHR/khrplatform.h

Existing conflict.

They look like the same file, we should ask upstream to use the system one.

> mingw-w64-x86_64-hdf4 /mingw64/include/error.h

The file doesn't seem to be included by the rest of the headers. Furthermore the symbols are not exported by the dll.

> mingw-w64-x86_64-libgdiplus /mingw64/lib/libgdiplus.a
> mingw-w64-x86_64-pupnp /mingw64/lib/libupnp.a

Those 2, perhaps msys2 doesn't have to ship the static library?

For pupnp, the library name doesn't follow package name. Maybe they can consider a rename for future releases..

> mingw-w64-x86_64-unixodbc /mingw64/include/odbcinst.h
> mingw-w64-x86_64-unixodbc /mingw64/include/sql.h
> mingw-w64-x86_64-unixodbc /mingw64/include/sqlext.h
> mingw-w64-x86_64-unixodbc /mingw64/include/sqltypes.h
> mingw-w64-x86_64-unixodbc /mingw64/include/sqlucode.h

Given that  they are explicitly consistent with the MS headers, they should use the system headers when available.

MehdiChinoune:
> Ok, some ideas below

See #10670

fziglio:
> > Here is a list of file conflicts created by the move:
> > If anyone has any idea how to handle these, go ahead.
> 
> Ok, some ideas below. I suppose ideally you would want to fix all before moving the crt & headers.
> 
> > mingw-w64-x86_64-freetds /mingw64/include/odbcss.h
> 
> Hi @freddy77 ! the file name conflicts with a windows system header. can it be renamed or moved upstream?
> 

They both provides the same header so I think renaming it would make no sense. The idea of this header was providing it for systems not having it. But if the system provides it (that is Windows target) using the system one should be the same (they should be compatibles). So I think the perfect solution to fix the issue would be for FreeTDS not installing the header if already provided by the system. Not sure 100% if they are really compatibles perfectly. In particular some functions exported by Windows ODBC driver manager are not available by Unix ODBC driver managers so I use some static inline definition in this file to override them, but this should not be necessary on Windows targets (the FreeTDS ODBC driver exports the needed functions for the DM).

elmarco:
For mesa, see https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/14019

elmarco:
glbinding header conflict was fixed in v3.3 https://github.com/cginternals/glbinding/issues/299

MehdiChinoune:
> glbinding header conflict was fixed in v3.3 [cginternals/glbinding#299](https://github.com/cginternals/glbinding/issues/299)

The latest release is 3.1.0, there is no mention of a new release anywhere!
When was it released?

elmarco:
> > glbinding header conflict was fixed in v3.3 [cginternals/glbinding#299](https://github.com/cginternals/glbinding/issues/299)
> 
> The latest release is 3.1.0, there is no mention of a new release anywhere! When was it released?

Oh, it's a branch actually, apparently forked from master a few weeks ago: https://github.com/cginternals/glbinding/commits/v3.3

@scheibel did you release 3.2.0 and 3.3.0 ?

elmarco:
For hdf4:
https://github.com/HDFGroup/hdf4/pull/17

I can make a msys2 backport if that helps.

elmarco:
For unixodbc: I tried to compile with the system headers, but it fails. I suggest we use `--includedir=$include/unixodbc`, there is already a .pc file.

scheibel:
> @scheibel did you release 3.2.0 and 3.3.0 ?

We haven't officially released a version `3.2.x` or `3.3.x`, although we absolutely should do so. You can think of those two branches as release candidates.
Are there any time constraints I should be aware of?
Unfortunately, I have a lot going on right now that is getting in the way of a clean release.
Would a tag, pre-release, or an unofficial release be of any help?

elmarco:
Many thanks @lazka ! What do you do of the remaining conflicts?

lazka:
According to my scripts there aren't any left now.

lazka:
There is some potential fallout from this:

* https://github.com/msys2/MINGW-packages/issues/10761
* https://github.com/OSGeo/gdal/issues/5317#issuecomment-1042632989

Though no clue right now.

revelator:
python2 build also broken, looks for system headers in old path and refuses to pick them up from current.
guess we need a patch ?.

PR #10645, opened 2022-02-02
Add python-installer package

lazka:
Untested, just made it build

### DIFF
--- a/mingw-w64-python-installer/PKGBUILD
+++ b/mingw-w64-python-installer/PKGBUILD
@@ -1,0 +1,85 @@
+# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
+_realname=installer
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=0.4.0
+pkgrel=1
+pkgdesc="A low-level library for installing from a Python wheel distribution (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+url='https://github.com/pradyunsg/installer'
+license=('spdx:MIT')
+depends=("${MINGW_PACKAGE_PREFIX}-python")
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-python-flit"
+  "${MINGW_PACKAGE_PREFIX}-cc"
+  "unzip"
+)
+_commit="32836e96e14d85db1f864ab5ff46ceb84b4a7673"
+_launcher_commit="8c455825fc1308d34a1a47238a9eb6d56f732142"
+source=("${_realname}-$pkgver"::"git+https://github.com/pradyunsg/installer.git#commit=${_commit}"
+        "simple_launcher"::"git+https://bitbucket.org/vinay.sajip/simple_launcher.git#commit=${_launcher_commit}")
+sha256sums=('SKIP'
+            'SKIP')
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+
+}
+
+build() {
+  cd "${srcdir}/simple_launcher"
+  windres --input launcher.rc --output resources.res --output-format=coff
+
+  local _gui_cflags="$CFLAGS -mwindows"
+  local _cli_cflags="$CFLAGS"
+  local _gui_lflags="-lshlwapi"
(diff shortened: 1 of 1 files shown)


mingw-w64-python-installer/PKGBUILD
```diff
@@ -0,0 +1,85 @@
...
+
+}
+
+build() {
+  cd "${srcdir}/simple_launcher"
+  windres --input launcher.rc --output resources.res --output-format=coff
+
+  local _gui_cflags="$CFLAGS"
+  local _cli_cflags="$CFLAGS"
+  local _gui_lflags="-lshlwapi"
```
jeremyd2019, line 37: shouldn't there be an `-mwindows` in there somewhere? and should it have `-municode`?

lazka: ah, right, thanks

jeremyd2019: building the same launcher is done in mingw-w64-python-distlib I think

lazka: that's where I copied it from initially. the source is not the same (but the code might be)

lazka:
The next steps are

* Bootstrap flit manually: https://flit.readthedocs.io/en/latest/bootstrap.html
* Port packages using pyproject2setuppy to build+installer
* Drop pyproject2setuppy

jeremyd2019:
Flit is already packaged, why does it need to be bootstrapped again?

lazka:
it uses the now abandoned pyproject2setuppy

jeremyd2019:
It looks like this package has a mechanism that will work without patching to map arches to proper launchers.

One concern I had with these launchers vs the setuptools ones is that they apparently combine the launcher exe and the script, instead of putting a `-script.py` file next to the launcher, but still do shebang processing.  This will make trying to `sed` the shebang after the fact more fraught.  It might be necessary to patch the shebang-generating code instead.  I think the launcher supports finding the executable in the shebang on the path instead of an absolute path, don't remember if it had a mode to look in the same directory as the launcher though.  It seemed to have a bunch of different defines that could be tweaked for different functionality.

jeremyd2019:
Most of that probably also applies to distlib, but I haven't noticed any package actually using the launchers from there yet.

lazka:
> Update: it does support shebang relative to the launcher out-of-the-box with default defines. For that to work, shebang must start with literal string `<launcher_dir>\`

good find, then we maybe just have to patch this package

I found another issue which is blocking us using this: https://github.com/pradyunsg/installer/issues/98

pradyunsg:
> Most of that probably also applies to distlib, but I haven't noticed any package actually using the launchers from there yet.

Really? pip does, whenever it builds via a wheel.

lazka:
We don't use pip for packaging, we might though in the future.

pradyunsg:
You’ll find this blog from a setuptools maintainer useful to read: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

lazka:
Thanks, yes, we are trying. We have one package so far: https://github.com/msys2/MINGW-packages/blob/82ee778df82fc2dc2cb7a8411a7f84ae93407858/mingw-w64-python-typing_extensions/PKGBUILD#L39

Let's merge this and and add some hacks for the wrong prefix in the respective PKGBUILDs. After we dropped pyproject2setuppy we can create a packaging template using pip for the rest.

lazka:
One downside of the new launcher is that stripping breaks it and we need to use "!strip" everywhere.

jeremyd2019:
I guess that makes sense, as it just concatenates the script contents onto the exe.  If somebody were really bugged by this it would probably be possible to embed the script in a resource, I have done that in the past with a dynamic 'payload' on a .exe and it's pretty easy with Windows APIs, and then the exe format knows it's there so tools shouldn't discard it.

lazka:
I guess adding resources would require some help from the toolchain? python-installer uses this just via the built launchers and pure Python..

jeremyd2019:
https://docs.microsoft.com/en-us/windows/win32/menurc/adding-deleting-and-replacing-resources so ctypes or pywin32 (https://mhammond.github.io/pywin32/win32api__UpdateResource_meth.html) could do it.  but probably not worth it.

jeremyd2019:
Just read the README for simple launcher, and it seems it uses python support for running a zip archive, so putting it in a resource probably wouldn't work.

lazka:
Ah, right.

PR #10651, opened 2022-02-03
llvm: update to 13.0.1 - Part1

MehdiChinoune:


mingw-w64-clang/PKGBUILD
```diff
@@ -196,7 +196,7 @@ build() {
 
   if (( _clangprefix )); then
     # A bit hacky but it works
-    local _clang_links="clang++;clang-cpp;as;c++;cc;cpp;g++;gcc;${MINGW_CHOST}-cc;${MINGW_CHOST}-c++;${MINGW_CHOST}-clang;${MINGW_CHOST}-clang++;${MINGW_CHOST}-g++;${MINGW_CHOST}-gcc"
+    local _clang_links="clang++;clang-cpp;as;c++;cc;cpp;${MINGW_CHOST}-cc;${MINGW_CHOST}-c++;${MINGW_CHOST}-clang;${MINGW_CHOST}-clang++"
```
MehdiChinoune, line 199: @jeremyd2019 Do you agree with this change

mingw-w64-clang/PKGBUILD
```diff
@@ -286,7 +286,6 @@ package_clang() {
              "${MINGW_PACKAGE_PREFIX}-winpthreads" \
              || echo "${MINGW_PACKAGE_PREFIX}-gcc"))
   provides=($( (( _clangprefix )) && echo \
-             "${MINGW_PACKAGE_PREFIX}-gcc" \
```
MehdiChinoune, line 289: same here.
All packages now depends on `${MINGW_PACKAGE_PREFIX}-cc`, so I think making clang provides gcc is no longer needed.

lazka: +1

mingw-w64-clang/PKGBUILD
jeremyd2019: I'm not sure about this one, I remember a lot of things assuming the C compiler was called gcc (or $CHOST-gcc).  But now that makepkg_mingw.conf sets CC/CXX env variables maybe that's less of an issue?

MehdiChinoune: Many users (including me) were confused about the default C/C++ compiler when they first used CLANG64.

lazka: Maybe we can have a "clang-as-gcc" split package that just contains the copies, which we can makedepend on if needed?

jeremyd2019: I don't think that's worthwhile.  If a package needs to be modified anyway, it would be better to patch it to respect $CC/$CXX imho

mingw-w64-clang/PKGBUILD
jeremyd2019: There are still 470 unique package bases that have packages that depend or makedepend on `${MINGW_PACKAGE_PREFIX}-gcc` in the clang64.db.

MehdiChinoune: Because they weren't rebuilt.

jeremyd2019: OK, looking at srcinfo.json cache instead, I see 44 PKGBUILDs that depend or makedepend on `${MINGW_PACKAGE_PREFIX}-gcc`.

lazka: Looks like I missed some. We need to get rid of them, and rebuild everything in the DB runtime depending on gcc.

MehdiChinoune: You should reported them earlier, when @lazka was replacing dependencies from gcc to cc.

jeremyd2019: > Looks like I missed some. We need to get rid of them, and rebuild everything in the DB runtime depending on gcc.

ok, will open new issue listing them at least

lazka: I'll have a look tomorrow

mingw-w64-clang/PKGBUILD
lazka: ok, let's see how things turn out.

amyspark: Apologies for hijacking, so far these packages have broken builds locally:

- Boost (via bootstrap.bat),
- cargo-c (specifically winapi-x86_64-pc-windows-gnu, can't find the linker)

Issue #10649, opened 2022-02-03
gcc-libs in clang envs does not pull in openmp


kmilos:
`gcc-libs` is provided by `libc++` in clang environments, but it does not also pull in `openmp` (`libomp`); in GCC toolchain environments OTOH `gcc-libs` does also contain `libgomp`...

This is related to https://github.com/msys2/MINGW-packages/pull/9884

The [workaround for now is to also explicitly require `openmp` for those environments](https://github.com/msys2/MINGW-packages/blob/5fc83fe2dd234957cc0de37db2d9c309104eb70c/mingw-w64-gmic/PKGBUILD#L17-L18), but it would be convenient to have the same behavior for all packages that already require only `gcc-libs` so the recipes don't have to be updated? I guess it would be quite a job to figure out [which of these dependants](https://packages.msys2.org/package/mingw-w64-x86_64-gcc-libs?repo=mingw64) actually need the explicit `openmp` addition for clang.

Alternatively, this can also be handled by some new pseudo-package like it was done w/ `cc` recently...

MehdiChinoune:
This is not related to #9884 at all.
The reason why those builds fail, is that CI was installing `${MINGW_PACKAGE_PREFIX}-toolchain` (which contains openmp on CLANG envs) before building.
@lazka has removed that and add explicit make-dependency on `${MINGW_PACKAGE_PREFIX}-cc`.

I think the solution is to add a package called `${MINGW_PACKAGE_PREFIX}-omp` provided by `gcc-libs` on GCC envs and `openmp` on CLANG envs.

lazka:
hm, something like "cc-libs" which depends on all those things?

> I guess it would be quite a job to figure out [which of these dependants](https://packages.msys2.org/package/mingw-w64-x86_64-gcc-libs?repo=mingw64) actually need the explicit `openmp` addition for clang.

Do you mean if it would have been installed it would have been used?

kmilos:
> Do you mean if openmp would have been installed during the build it would have been used?

Probably, because that package supplies both the header and the library for -clang- envs

lazka:
here is a list of packages actually linking against libgomp (with gcc): https://github.com/lazka/package-grokker/runs/5057107592?check_suite_focus=true#step:4:9

kmilos:
Thanks for that. It's not the hugest list, so not inconceivable to do by explicitly for now...

kmilos:
[...]

PR #10660, opened 2022-02-04
mlpack: enable openmp for clang builds

kmilos:
Testing hypothesis in #10649 that OpenMP will automatically be enabled if the dependency is supplied.

### DIFF
--- a/mingw-w64-mlpack/PKGBUILD
+++ b/mingw-w64-mlpack/PKGBUILD
@@ -11,6 +11,7 @@ mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url='https://mlpack.org/'
 license=('BSD')
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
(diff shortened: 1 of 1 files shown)


lazka:
I've created #10662 for the build error

kmilos:
Yep, in #10662 and -cc only instead of toolchain, one sees e.g. in the CLANG64 log

```
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND)
```

W/ the openmp addition here:

```
-- Found OpenMP_C: -fopenmp=libomp (found version "5.0") 
-- Found OpenMP_CXX: -fopenmp=libomp (found version "5.0") 
-- Found OpenMP: TRUE (found version "5.0") 
```

PR #10665, opened 2022-02-04
arrow: Update to 7.0.0

kou:


lazka:
Does [groonga](https://packages.msys2.org/base/mingw-w64-groonga) need a rebuild?

kou:
Ah, sorry. I forgot it.
I'll add a commit.

kou:
Pushed.

PR #10666, opened 2022-02-04
Add new abseil-cpp PKGBUILD

michallukowski:
Description form [abseil.io](https://abseil.io/): "Abseil is an open source collection of C++ libraries drawn from the most fundamental pieces of Google’s internal codebase. "

PKGBUILD based on [archlinux](https://github.com/archlinux/svntogit-community/blob/packages/abseil-cpp/trunk/PKGBUILD) repository.


lazka:
There are some problems building this as a shard library on clang. Would you be OK if it was a static lib for now?

michallukowski:
Sure.

PR #10668, opened 2022-02-05
[do not merge] Packages for OpenModelica

jdpipe:
I have produced packages for the OpenModelica compiler and Modelica Standard Library that seem to work for me. I would like to see if they're compatible with MSYS2 yet or not. One issue is that the build process uses a JDK at one point, and I'm not sure yet how to do that on the MSYS2 servers. Let's see...


lazka:
Anything we can help with here?

PR #10670, opened 2022-02-05
Fix conflicts with mingw-w64

MehdiChinoune:


mingw-w64-pupnp/PKGBUILD
```diff
@@ -48,6 +48,7 @@ build() {
     -G"Ninja" \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
     -DCMAKE_BUILD_TYPE=Release \
+    -DSTATIC_POSTFIX="-static" \
```
Biswa96, line 51: This may create issue with static linking. pkgconfig file has `-lupnp`.

MehdiChinoune:
About pupnp should we just rename the static library or should we disable it entirely. with the first option, pkgconf could pick the mingw-w64 upnp for static linking. The only package depending on pupnp is vlc and I think it links to the shared library.

mingw-w64-mesa/0008-win32-The-KHR-headers-should-install-to-mesa-KHR-s.patch
```diff
@@ -0,0 +1,11 @@
+--- a/include/meson.build
++++ b/include/meson.build
+@@ -31,7 +31,7 @@
+ 
+ if not with_glvnd
+   if with_gles1 or with_gles2 or with_opengl or with_egl
+-    install_headers('KHR/khrplatform.h', subdir : 'KHR')
++    install_headers('KHR/khrplatform.h', subdir : 'mesa/KHR')
```
lazka, line 8: Shouldn't we just delete it as well after building? No one will find it there, no?

lazka:
I'm fine with disabling that static build for now.

MehdiChinoune:
Unfortunately There is not an option to disable pupnp static library, even if we do pkgconf won't have any info about the availability of the static library and will link the mingw-w64 one.

MehdiChinoune:
Is this PR accepted or not?

Issue #10637, opened 2022-02-01
SVT-AV1 support of ffmpeg


wszqkzqk:
`SVT-AV1` is a new `av1` encoder with high performance. And now can add it to ffmpeg. May add `${MINGW_PACKAGE_PREFIX}-svt-av1` to the `dependence` and `--enable-libsvtav1` to the `common_config`.

Biswa96:
Hi @wszqkzqk, Thank you for reporting the issue. I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/1782084610

Biswa96:
I have added a pull request to add zimg dependency with ffmpeg. The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/1800246850

wszqkzqk:
> I have added a pull request to add zimg dependency with ffmpeg. The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/1800246850

But when I tried to compile, it showed this error:
![``NAHH9X@$X653PFB961> I have added a pull request to add zimg dependency with ffmpeg. The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/1800246850

But when I tried to compile, it showed this error:
5](https://user-images.githubusercontent.com/12961135/152677660-09b376ad-325d-4c76-a0ca-cd041d4df9e0.png)

Biswa96:
I am unaware of your build environment and how you set that up. The new ffmpeg package has been uploaded to pacman repository. You can download it using `pacman -Syu` command.

wszqkzqk:
> I am unaware of your build environment and how you set that up. The new ffmpeg package has been uploaded to pacman repository. You can download it using `pacman -Syu` command.

I just tries to run `makepkg-mingw -si`……

wszqkzqk:
> I am unaware of your build environment and how you set that up. The new ffmpeg package has been uploaded to pacman repository. You can download it using `pacman -Syu` command.

It seems that there is a mistake about relative path.

PR #10671, opened 2022-02-05
ffmpeg: Add zimg dependency

Biswa96:
As requested here https://github.com/msys2/MINGW-packages/issues/10637

### DIFF
--- a/mingw-w64-ffmpeg/PKGBUILD
+++ b/mingw-w64-ffmpeg/PKGBUILD
@@ -7,7 +7,7 @@ _realname=ffmpeg
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=4.4.1
-pkgrel=3
+pkgrel=4
 pkgdesc="Complete and free Internet live audio and video broadcasting solution (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mingw-w64-ffmpeg/PKGBUILD
```diff
@@ -136,7 +138,12 @@ build() {
     --disable-doc
   )
 
-  if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]]; then
+  if [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]]; then
+    common_config+=(
+      --cc=clang
+      --cxx=clang++
+    )
+  else
```
lazka, line 146: We do set CC and CXX now, so in theory you could pass `--cc=$CC -cxx=$CXX` unconditionally

Biswa96: OK, I will add that. Accidentally, that was my another choice. But then I choose to set the options explicitly to make it future proof. Things are changing a lot now-a-days 😅

lazka: rolling ~~distro~~breakage

Biswa96: Done.

PR #10679, opened 2022-02-06
Fix compiler warning about invalid date format with %F.

JPeterMugaas:


Biswa96:
mingw_arch and makedepends were removed accidentally.

mingw-w64-scdoc/parsedate.patch
```diff
@@ -0,0 +1,14 @@
+--- scdoc-1.11.2/src/main.c.orig	2022-02-06 02:16:47.709677200 -0500
++++ scdoc-1.11.2/src/main.c	2022-02-06 02:22:05.844377900 -0500
+@@ -110,7 +110,11 @@ static void parse_preamble(struct parser
+ 		date_time = time(NULL);
+ 	}
+ 	struct tm *date_tm = gmtime(&date_time);
++#ifdef __MINGW32__
```
lazka, line 7:
```suggestion
+#if defined(__MINGW32__) && !defined(_UCRT)
```

JPeterMugaas: I just tried it and got this:

src/main.c: In function 'parse_preamble':
src/main.c:116:40: error: unknown conversion type character 'F' in format [-Werror=format=]
  116 |         strftime(date, sizeof(date), "%F", date_tm);
      |                                        ^
cc1.exe: all warnings being treated as errors

lazka: ok, I guess something is broken there. feel free to revert, sry.

PR #10680, opened 2022-02-06
Remove pyproject2setuppy

lazka:
Depends on #10685


lazka:
Next step is to remove the remaining dep cycle, which should be possible now if we split flit/flit-core.

PR #10685, opened 2022-02-06
python-installer: various fixes

lazka:
* always use python.exe in the launcher dir
* pass all build flags to cc
* set _CONSOLE macro when building the console version
* use installer for itself to install

### DIFF
--- a/mingw-w64-python-installer/001-launcher-relative.patch
+++ b/mingw-w64-python-installer/001-launcher-relative.patch
@@ -1,0 +1,11 @@
+--- installer-0.4.0/src/installer/__main__.py.orig	2022-02-06 15:44:54.255000200 +0100
++++ installer-0.4.0/src/installer/__main__.py	2022-02-06 15:44:57.255156900 +0100
+@@ -90,7 +90,7 @@
+     with installer.sources.WheelFile.open(args.wheel) as source:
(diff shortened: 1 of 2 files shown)


mingw-w64-python-installer/001-launcher-relative.patch
```diff
@@ -0,0 +1,11 @@
+--- installer-0.4.0/src/installer/__main__.py.orig	2022-02-06 15:44:54.255000200 +0100
++++ installer-0.4.0/src/installer/__main__.py	2022-02-06 15:44:57.255156900 +0100
+@@ -90,7 +90,7 @@
+     with installer.sources.WheelFile.open(args.wheel) as source:
+         destination = installer.destinations.SchemeDictionaryDestination(
+             get_scheme_dict(source.distribution, prefix=args.prefix),
+-            sys.executable,
++            r'<launcher_dir>\python.exe',
```
jeremyd2019, line 8: I was thinking `r'<launcher_dir>\' + os.path.basename(sys.executable),` in case somebody was using python3 or something.  Is that a valid concern?

lazka: I can't really think of a scenario where this would be a problem. I'd rather have the launchers all behave the same if possible (the zip and append part makes this not very transparent)

PR #10689, opened 2022-02-07
FFmpeg: Add rav1e dependency

wszqkzqk:
FFmpeg: Add rav1e dependency


Biswa96:
[rav1e][1] package is not available for clang32. It should be added conditionally when MSYSTEM is not clang32 (or other similar checks).

[1]: https://packages.msys2.org/base/mingw-w64-rav1e

wszqkzqk:
> [rav1e](https://packages.msys2.org/base/mingw-w64-rav1e) package is not available for clang32. It should be added conditionally when MSYSTEM is not clang32 (or other similar checks).

ok, I fixed it.

PR #10692, opened 2022-02-07
xa65: Fix build with clang

Biswa96:
Remove predefined CC variable for clang /cc @drfiemost

### DIFF
--- a/mingw-w64-xa65/0001-make-fix-predefined-variables.patch
+++ b/mingw-w64-xa65/0001-make-fix-predefined-variables.patch
@@ -1,0 +1,27 @@
+--- a/Makefile
++++ b/Makefile
+@@ -1,11 +1,11 @@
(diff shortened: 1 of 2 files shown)


mingw-w64-xa65/0001-make-fix-predefined-variables.patch
```diff
@@ -0,0 +1,27 @@
+--- a/Makefile
++++ b/Makefile
+@@ -1,11 +1,10 @@
+ # Unix gcc or DOS go32 cross-compiling gcc
+-#
+-CC = gcc
+-LD = gcc
++# Assuming CC is predefined
```
jeremyd2019, line 8: maybe `CC ?= gcc` in case it's not set?

Biswa96: 👍

PR #10693, opened 2022-02-07
freetds: update to 1.3.7

MehdiChinoune:


lazka:
And removes the conflict for https://github.com/msys2/MINGW-packages/pull/10634#issuecomment-1030608049

no .pc file, but also no users, so let's see if someone complains.

PR #10694, opened 2022-02-07
Update: librest -> 0.9.0

BFamzz:
Update librest to the latest release 0.9.0

### DIFF
--- a/mingw-w64-librest/PKGBUILD
+++ b/mingw-w64-librest/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=rest
 pkgbase=mingw-w64-lib${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-lib${_realname}"
-pkgver=0.8.1
+pkgver=0.9.0
 pkgrel=1
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


mingw-w64-librest/0001-fix-meson-build-subdir.patch
```diff
@@ -0,0 +1,16 @@
+--- a/meson.build
++++ a/meson.build
+@@ -74,12 +74,7 @@
+ # Subdirectories
+ subdir('rest')
+ subdir('rest-extras')
+-if get_option('tests')
```
lazka, line 7: If you want to disabled them you can pass `-Dtests=false -Dexamples=false` to meson

mingw-w64-librest/PKGBUILD
```diff
@@ -3,62 +3,53 @@
...
-  make
+  ${MINGW_PREFIX}/bin/ninja.exe
 }
 
 package() {
   cd "${srcdir}/build-${MINGW_CHOST}"
-  make DESTDIR=${pkgdir} install
+  cd "${srcdir}/build-${MINGW_CHOST}"
 
-  install -Dm644 "${srcdir}/${_realname}-${pkgver}/COPYING" "${pkgdir}${MINGW_PREFIX}/share/licenses/lib${_realname}/COPYING"
```
lazka, line 63: why was this removed?

mingw-w64-librest/0001-fix-meson-build-subdir.patch
BFamzz: Thanks. I really struggled with this. I pored through the meson documentation and tried the different options that get_options() could return.

I'll push an update shortly.

mingw-w64-librest/PKGBUILD
BFamzz: It was an error borne out of the meson get_option() frustration.

I'll push a fix shortly.

Thanks.

mingw-w64-librest/0001-fix-meson-build-subdir.patch
lazka: The meson docs can be quite confusing, yeah :)

lazka:
thanks

PR #10714, opened 2022-02-09
enable a few more clangarm64 packages

kmilos:
Includes pugixml update to 1.12, lcms2 dependency cleanup, and adding missing licenses.

### DIFF
--- a/mingw-w64-exiv2/PKGBUILD
+++ b/mingw-w64-exiv2/PKGBUILD
@@ -9,7 +9,7 @@ pkgver=0.27.5
 pkgrel=1
 pkgdesc="Exif/IPTC/Xmp C++ metadata library and tools (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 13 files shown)


Biswa96:
Try to use `${MSYSTEM}` in build folder, it's trending now 😊

Crystal013TKFM:
If you prefer lowercase, you can write `${MSYSTEM,,}`
On the contrary, `${MINGW_ARCH^^}` will convert it to uppercase

kmilos:
> If you prefer lowercase, you can write `${MSYSTEM,,}`
> On the contrary, `${MINGW_ARCH^^}` will convert it to uppercase

This what the `makepkg-mingw` script does already, both ways.

jeremyd2019:
MINGW_ARCH is allowed to be a space-separated list of arches to build.  MSYSTEM is set to each for each individual build.

kmilos:
Ah, thanks for that insight! Will change it then.

PR #10720, opened 2022-02-09
binutils: update to 2.38

MehdiChinoune:


lazka:
hm, can't we do this in two steps?

MehdiChinoune:
> hm, can't we do this in two steps?

What steps? Do you mean one to upgrade binutils and other to enable it for CLANG*?

lazka:
yes :/

MehdiChinoune:
> yes :/

Sorry, No. I know the fate of the second step If I do.

mati865:
> > yes :/
> 
> Sorry, No. I know the fate of the second step If I do.

I won't stop you from adding GCC/Binutils to `CLANG*` subsystems, just trying to save you from wasting your time since produced tools won't be very useful.
`windmc` from Binutils could be exception though (assuming it's output works with llvm-windres), right now it's either not found or found in `/usr/bin`.

PR #10742, opened 2022-02-13
rename python-pyprojecttosetuptools to build-installer to be more acc…

JPeterMugaas:
…urate.


JPeterMugaas:
I was concerned that the template was a misnomer given that the module it originally used was removed recently.  And yes, we do need a template for a "best practices" whatever those are.

lazka:
I've linked it here: https://github.com/msys2/msys2.github.io/commit/deb616232f31b9d4

PR #10743, opened 2022-02-13
nsis: update to 3.08

raedrizqie:


mingw-w64-nsis/004-crossmingw-tools-fix.patch
```diff
@@ -0,0 +1,26 @@
+diff -urN nsis-3.08-src-orig/SCons/Tools/crossmingw.py nsis-3.08-src/SCons/Tools/crossmingw.py
+--- nsis-3.08-src-orig/SCons/Tools/crossmingw.py	2019-10-07 00:15:14.000000000 +0800
++++ nsis-3.08-src/SCons/Tools/crossmingw.py	2022-02-12 19:27:07.410761600 +0800
+@@ -167,17 +167,17 @@
+     env['SHLINKCOM']   = shlib_action
+     env.Append(SHLIBEMITTER = [shlib_emitter])
+     env['LINK'] = mingw_prefix + 'g++'
```
jeremyd2019, line 7: Maybe this patch can change to use cc/c++ instead of gcc/g++, and avoid the really nasty hack of copying clang/clang++ in build()?

raedrizqie: i tried changing to $CC and $CXX but it still searching for gcc and g++ ...

jeremyd2019:
I haven't looked at this beyond getting it to build without creating gcc/g++ aliases for clang

mingw-w64-nsis/PKGBUILD
```diff
@@ -73,39 +71,53 @@ _build() {
...
+  cd "${srcdir}/build-${MSYSTEM}"
   scons \
       TARGET_ARCH="${target_arch}" \
-      PREFIX="${MINGW_PREFIX}" \
       VERSION="${pkgver}" \
       VER_MAJOR="${major_version}" \
       VER_MINOR="${minor_version}" \
       VER_REVISION="${revision_version}" \
       VER_BUILD="${build_version}" \
+      TOOLSET=gcc,mingw \
```
MehdiChinoune, line 98: Is there a 'clang' option?

raedrizqie: no, afaik

PR #10747, opened 2022-02-14
ag: fix compilation with clang

neheb:
getpagesize() is needed.


lazka:
Any idea why that is only a problem with clang?

Biswa96:
In mingw and ucrt, the function is replaced with something like this `long long getpagesize() { return 0x1000 }`. Not sure where it comes from.

neheb:
Wonder if that's a bug or if pagesise really doesn't differ between platforms.

neheb:
anything else misisng here?

lazka:
seems like GCC somehow injects it (??)

fine I guess

PR #10750, opened 2022-02-14
gspell: Update to 1.9.1

BFamzz:


podsvirov:
Hello @BFamzz! Please rename PR to `gspell: Update to 1.9.1`

podsvirov:
And you too! (As you see, we have many packages :-)

lazka:
1.9.x is an unstable release, so I'm not sure if we should package it. -> https://gitlab.gnome.org/GNOME/gspell/-/blob/master/README

```

gspell follows the GNOME release schedule and uses the same versioning scheme:

    https://developer.gnome.org/programming-guidelines/stable/versioning.html.en

For example the 1.7.x versions are unstable, and the 1.8.x versions are stable.
```

lazka:
So unless there is a specific reason why this update is needed I'm going to close it.

BFamzz:
Hi @lazka. Apologies for late response. This seems to be the last release for the project. The project has been [frozen](https://gitlab.gnome.org/GNOME/gspell/-/issues/17). Although there was a revert freeze [commit](https://gitlab.gnome.org/GNOME/gspell/-/commit/16cdbdbda171d169cbc58452c222d5458ad5e0f0). 

Also, from the [NEWS](https://gitlab.gnome.org/GNOME/gspell/-/blob/5300d93dae00998801ef1be48da71dac4e2ebd1b/NEWS), unstable versions are explicitly state and the changes made in 1.9.1 looks minimal to be unstable (I may be wrong).

Please feel free to close the request if required.

Thanks.

PR #10751, opened 2022-02-14
slim-simulator: Update to version 3.7.1

rdinnager:
This just updates `slim-simulator` to the newest version. Didn't require any changes except for the sha256sums of course.

### DIFF
--- a/mingw-w64-slim-simulator/PKGBUILD
+++ b/mingw-w64-slim-simulator/PKGBUILD
@@ -1,8 +1,8 @@
 _realname=slim-simulator
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.7
-pkgrel=2
+pkgver=3.7.1
+pkgrel=1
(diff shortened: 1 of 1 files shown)


neheb:
```
  C:/_/mingw-w64-slim-simulator/src/SLiM_pacman/eidos/eidos_globals.cpp:87:30: error: static assertion failed: SLiM must be built for 64-bit, not 32-bit.
     87 | static_assert(sizeof(char *) == 8, "SLiM must be built for 64-bit, not 32-bit.");
        |               ~~~~~~~~~~~~~~~^~~~
```

looks like a CI bug.

Biswa96:
> looks like a CI bug.

Nope. mingw32 is not present mingw_arch. So, the package is not built for 32 bit. CI just runs anyway. LGTM.

neheb:
> > looks like a CI bug.
> 
> Nope. mingw32 is not present mingw_arch. So, the package is not built for 32 bit. CI just runs anyway. LGTM.

that's exactly what I meant.

Biswa96:
Understand. That can be ignored for this case.

PR #10773, opened 2022-02-16
Improve Flashrom build

obskyr:
Three changes have been made to the `PKGBUILD` for Flashrom:
* A `versioninfo.inc` is now created. Previously when running `flashrom --version`, it would report `flashrom unknown; on Windows 10.0` – with this change, it now correctly reports `flashrom v1.2; on Windows 10.0`.
* As `patch` doesn't come with MSYS2, it has been added to `makedepends` so that it doesn't need to be installed manually. `patch` is part of `base-devel`, however, so let me know if `base-devel` packages are exempt from `PKGBUILD` and I'll make the necessary change.
* Bumped `pkgrel` to 3.


obskyr:
Updated to bump `pkgrel`.

Biswa96:
The proper solution would be to use upstream provided source tarball. Let me suggest the fix.

PR #10774, opened 2022-02-17
flashrom: Fix version reporting with upstream source tarball

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/pull/10773

### DIFF
--- a/mingw-w64-flashrom/PKGBUILD
+++ b/mingw-w64-flashrom/PKGBUILD
@@ -4,11 +4,11 @@ _realname=flashrom
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.2
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


Biswa96:
@obskyr Does this look good? The upstream source tarball already has versioninfo.inc. So, we don't have to generate it.

obskyr:
This certainly fixes `versioninfo.inc`, yes! One question, however: should `makedepends` not include `"patch"`? When I attempted to build on a clean-ish MSYS2 installation, it complained that there was no such command as `patch`, and I had to install it manually with `pacman`.

Biswa96:
> should makedepends not include "patch"?

It is not required to mention in PKGBUILD file. To build a package, one has to install base-devel package group which contains the patch program and others. It's similar as build-essentials in Debian like systems. More details can be found here https://www.msys2.org/wiki/Creating-Packages/#build-software

obskyr:
Alright; very good to know – thank you for the information! This looks like a great fix, then!

PR #10780, opened 2022-02-17
scotch: fix clang build

raedrizqie:


mingw-w64-scotch/0001-makefile.inc.patch
```diff
@@ -10,12 +10,12 @@ diff -urN scotch_6.0.8.orig/src/Makefile.inc scotch_6.0.8/src/Makefile.inc
 +AR		= ar
 +ARFLAGS	= -ruv
 +CAT		= cat
-+CCS		= gcc
-+CCP		= mpicc
-+CCD		= gcc
++CCS		= $(CC)
++CCP		= $(CC)
```
Biswa96, line 14: Should CCP be assigned with `$(CC)`? CCP is set with mpicc.

raedrizqie: mirroring `src/Make.inc/Makefile.inc.i686_pc_mingw32`, CCP is set to gcc

mingw-w64-scotch/0001-makefile.inc.patch
```diff
@@ -10,12 +10,12 @@ diff -urN scotch_6.0.8.orig/src/Makefile.inc scotch_6.0.8/src/Makefile.inc
...
++CCD		= $(CC)
 +CFLAGS		+= -DCOMMON_RANDOM_FIXED_SEED -DSCOTCH_RENAME -Drestrict=__restrict -DCOMMON_PTHREAD -DSCOTCH_PTHREAD -DCOMMON_FILE_COMPRESS_GZ
 +CLIBFLAGS	=
-+LDFLAGS	= -lm -lz -pthread
++LDFLAGS	= -lmsmpi -lz -lpthread
```
mati865, line 18: `-pthread` does much more than simply linking the library and often is the right thing to do, more info at: https://stackoverflow.com/a/23251828/5940067

raedrizqie: ok, done

mingw-w64-scotch/0001-makefile.inc.patch
mmuetzel: After #10799, `mpicc` should be working again for the clang environments.

mingw-w64-scotch/0001-makefile.inc.patch
```diff
@@ -10,12 +10,12 @@ diff -urN scotch_6.0.8.orig/src/Makefile.inc scotch_6.0.8/src/Makefile.inc
...
-+CCS		= gcc
-+CCP		= mpicc
-+CCD		= gcc
++CCS		= $(CC)
++CCP		= $(CC)
++CCD		= $(CC)
 +CFLAGS		+= -DCOMMON_RANDOM_FIXED_SEED -DSCOTCH_RENAME -Drestrict=__restrict -DCOMMON_PTHREAD -DSCOTCH_PTHREAD -DCOMMON_FILE_COMPRESS_GZ
 +CLIBFLAGS	=
-+LDFLAGS	= -lm -lz -pthread
++LDFLAGS	= -lmsmpi -lz -pthread
```
mmuetzel, line 18: This change might not be necessary if you revert to `mpicc`.

mingw-w64-scotch/PKGBUILD
```diff
@@ -34,10 +35,12 @@ prepare() {
 
 build() {
   cd "${srcdir}/${_realname}-v${pkgver}/src"
-  make -j1 scotch ptscotch esmumps ptesmumps
+  CC=${CC} make -j1 scotch ptscotch esmumps ptesmumps
```
mmuetzel, line 38: Should we set `CCS=${CC} CCD=${CC}` instead? That might make the new patch unnecessary.

mingw-w64-scotch/PKGBUILD
```diff
@@ -34,10 +35,12 @@ prepare() {
...
   cd ../lib
-  gcc -shared -o lib${_realname}.dll -Wl,--enable-auto-import -Wl,--export-all-symbols -Wl,--output-def,lib${_realname}.def -Wl,--out-implib,lib${_realname}.dll.a -Wl,--whole-archive lib${_realname}.a lib${_realname}err.a -Wl,--no-whole-archive
-  mpicc -shared -o libpt${_realname}.dll -Wl,--enable-auto-import -Wl,--export-all-symbols -Wl,--output-def,libpt${_realname}.def -Wl,--out-implib,libpt${_realname}.dll.a -Wl,--whole-archive libpt${_realname}.a libpt${_realname}err.a -Wl,--no-whole-archive -L. -l${_realname}
+  ${CC} -shared -o lib${_realname}.dll -Wl,--export-all-symbols,-no-undefined \
+    -Wl,--out-implib,lib${_realname}.dll.a lib${_realname}.a lib${_realname}err.a -pthread
+  ${CC} -shared -o libpt${_realname}.dll -Wl,--export-all-symbols,-no-undefined \
```
mmuetzel, line 42: After #10799, mpicc should be working again for the clang environments.

raedrizqie: done with all the proposed changes

mingw-w64-scotch/PKGBUILD
```diff
@@ -5,24 +5,25 @@ pkgbase=mingw-w64-${_realname}
...
-		"0001-makefile.inc.patch"
-		"0002-pipe-fix.patch"
-		"0003-thread-fix.patch"
-		"0004-dummysizes-regex.patch"
-)
-sha256sums=('4E54F056199E6C23D46581D448FCFE2285987E5554A0AA527F7931684EF2809E'
-            '2c55b1e99a8d40c976e7147aa9102db2d1c489da1b760e526b424bf5ffd12716'
+        "0001-makefile.inc.patch"
+        "0002-pipe-fix.patch"
+        #"0003-thread-fix.patch"
```
lazka, line 21: Do we know why this is no longer needed?

raedrizqie: err.. why did we need it in the first place anyway?

lazka: @okhlybov do you know why https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-scotch/0003-thread-fix.patch was added/required?

okhlybov: That was a quick hack to make pt-scotch compile. That part of code is auxiliary, so fixing the thread number to 1 wasn't considered a big deal.

lazka: What's pt-scotch? So should we keep this?

okhlybov: PT-scotch should be preserved for at least for MUMPS and (optionally) for PETSc. Some other not yet ported solvers make use of it as well.

lazka: I meant should we keep this patch? :)

okhlybov: Scotch probably won't compile without this patch so yes, keep it, at least for a while.

lazka: hm, but it's compiling here in CI without it.

okhlybov: Then something has changed since this patch was introduced. If it builds now then the patch can be omitted.

raedrizqie: ok.. thx for the clarification

mingw-w64-scotch/PKGBUILD
```diff
@@ -36,8 +37,10 @@ build() {
   cd "${srcdir}/${_realname}-v${pkgver}/src"
   make -j1 scotch ptscotch esmumps ptesmumps
   cd ../lib
-  gcc -shared -o lib${_realname}.dll -Wl,--enable-auto-import -Wl,--export-all-symbols -Wl,--output-def,lib${_realname}.def -Wl,--out-implib,lib${_realname}.dll.a -Wl,--whole-archive lib${_realname}.a lib${_realname}err.a -Wl,--no-whole-archive
-  mpicc -shared -o libpt${_realname}.dll -Wl,--enable-auto-import -Wl,--export-all-symbols -Wl,--output-def,libpt${_realname}.def -Wl,--out-implib,libpt${_realname}.dll.a -Wl,--whole-archive libpt${_realname}.a libpt${_realname}err.a -Wl,--no-whole-archive -L. -l${_realname}
+  ${CC} -shared -o lib${_realname}.dll -Wl,--export-all-symbols,-no-undefined \
+    -Wl,--out-implib,lib${_realname}.dll.a lib${_realname}.a lib${_realname}err.a
```
mmuetzel, line 41: Without `-Wl,--whole-archive` and without `.def` file no symbols will be exported from these shared libraries.

raedrizqie: ok.. thanks

lazka:
Merged #10823 which includes this.

PR #10784, opened 2022-02-17
librsvg: update to 2.52.6

mmuetzel:
Also fix building with clang.

### DIFF
--- a/mingw-w64-librsvg/PKGBUILD
+++ b/mingw-w64-librsvg/PKGBUILD
@@ -3,13 +3,13 @@
 _realname=librsvg
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.52.5
-pkgrel=2
+pkgver=2.52.6
+pkgrel=1
 pkgdesc="SVG rendering library (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 url="https://wiki.gnome.org/Projects/LibRsvg"
-license=("LGPL")
+license=("spdx:LGPL-2.1-or-later")
 makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
              "${MINGW_PACKAGE_PREFIX}-gobject-introspection"
              "${MINGW_PACKAGE_PREFIX}-vala"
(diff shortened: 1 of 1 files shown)


mingw-w64-librsvg/PKGBUILD
```diff
@@ -57,11 +57,11 @@ build() {
     --enable-vala \
     --enable-silent-rules
 
-  make
+  make RUSTFLAGS="-C linker=${CC}"
```
lazka, line 60:
```suggestion
  make
```

mingw-w64-librsvg/PKGBUILD
```diff
@@ -57,11 +57,11 @@ build() {
...
 }
 
 package() {
   cd "${srcdir}/build-${MSYSTEM}"
-  make DESTDIR="${pkgdir}" install
+  make install DESTDIR="${pkgdir}" RUSTFLAGS="-C linker=${CC}"
```
lazka, line 65:
```suggestion
  make DESTDIR="${pkgdir}" install
```

mingw-w64-librsvg/PKGBUILD
mmuetzel: Done

lazka:
Builds fine locally -> ship it

dustwanderer:
I am not sure whether I have understood you. I tried executing doublecmd.exe from a bash shell, or from Windows PowerShell. The error message did not appear in either case. However, If I execute doublecmd.exe from "Windows Command Prompt (CMD.exe)", or from "Windows File Explorer", I get the error message

Biswa96:
I have tried that also and doublecmd.exe executes without any issue. Here is the commands:

```cmd
set PATH=C:\msys64\ucrt64\bin;%PATH%
doublecmd.exe
```

dustwanderer:
Your commands provided the solution. I moved C:\msys64\ucrt64\bin to the top of my PATH stack, and the error disappeared.

It turns out that my PATH had one version of librsvg-2-2.dll, but two versions of libxml2-2.dll. So, I was loading an obsolete libxml2-2.dll instead of the updated libxml2-2.dll in C:\msys64\ucrt64\bin\.

I removed the obsolete libxml2-2.dll, and rearranged my PATH to its original configuration.

Thank you very much for your help.

PR #10787, opened 2022-02-18
ntldd-git: use $CC instead of gcc when building

jeremyd2019:
now that clang does not provide a gcc alias


mati865:
Probably worth trying to upstream this patch.

PR #10803, opened 2022-02-19
avrdude: update to 6.4

fauxpark:


fauxpark:
```
    avrdude-6.4.tar.gz ... FAILED (unknown public key 7E9EADC3030D34EB)
```

Anything I can do to fix this? I swear I've encountered this before, but not sure if/how it was resolved.

fauxpark:
I figured that, but it wasn't working for me locally. I assume that's because CI does the additional step of receiving the keys first.

Can/should the others be removed?

jeremyd2019:
Usually a project will publish a list of keys that can be expected to sign their releases.  I'm not finding that anywhere for avrdude.

fauxpark:
It seems they've moved from Savannah to GitHub recently, I've filed an issue there.

lazka:
That key is used in Arch in multiple places, so I assume it's OK, https://github.com/search?q=56628323218C669FF578705C7E9EADC3030D34EB&type=code

lazka:
(CI currently calls `../.ci/fetch-validpgpkeys.sh` in a package directory to fetch the referenced keys)

PR #10806, opened 2022-02-19
qt6: update to 6.2.3 - Part2

MehdiChinoune:


MehdiChinoune:
UCRT64 build failed because qt6-declarative<6.2.3, should I rebuild them again after qt6-declarative be uploaded to staging?

PR #10809, opened 2022-02-19
mpfr: Enable building shared libraries

### LINKED ISSUE #10808: Need MPFR shared libraries
I am trying to build an MS Windows binary for Normaliz  (https://github.com/Normaliz/Normaliz) with all optional packages that are available under Linux and MacOS. I have succeeded with one exception, namely e-antic (https://github.com/flatsurf/e-antic). It insists on shared libraries upstream, including MPFR. 

Unfortunately my attempts to build a dll for MPFR have failed because configure thinks that I don't have shared libraries for GMP. They exist, and without them the C compiler would not even work.

It did not help to install MPFR by pacman, since shared libraries are build neither. What can I do?

MehdiChinoune:
fixes #10808

### DIFF
--- a/mingw-w64-gmp/PKGBUILD
+++ b/mingw-w64-gmp/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gmp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=6.2.1
-pkgrel=3
+pkgrel=4
 pkgdesc="A free library for arbitrary precision arithmetic (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
@@ -36,8 +36,11 @@ build() {
       extra_config+=(--disable-assembly)
       ;;
   esac
+
   # Build static version
-  mkdir -p "${srcdir}/static-${MINGW_CHOST}" && cd "${srcdir}/static-${MINGW_CHOST}"
+  [[ -d "${srcdir}/static-${MSYSTEM}" ]] && rm -rf "${srcdir}/static-${MSYSTEM}"
+  mkdir -p "${srcdir}/static-${MSYSTEM}" && cd "${srcdir}/static-${MSYSTEM}"
+  CFLAGS+=" -Wno-ignored-attributes -Wno-unused-command-line-argument" \
   ../${_realname}-${pkgver}/configure \
     --build=${MINGW_CHOST} \
     --prefix=${MINGW_PREFIX} \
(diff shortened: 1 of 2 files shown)


lazka:
This adds a runtime dep on all reverse dependencies when they get rebuilt, so we would need to add mpfr there.

MehdiChinoune:
You could add it now without bumping pkgrel, just in case

MehdiChinoune:
So, we went from enabling static libraries to favoring them and disabling the shared ones.
Good progress.

lazka:
I'm confused, what are you referring to?

MehdiChinoune:
The only package where mpfr is not in `depends` is `eigen3` which is just headers library.

MehdiChinoune:
Ping!

MehdiChinoune:
> (This adds a runtime dep on all reverse dependencies when they get rebuilt, so we would need to add mpfr there)

It could be detected now with checking-runtime-dependencies step.

lazka:
What about gcc?

MehdiChinoune:
> What about gcc?

About what?
`mpfr` depends only on `gmp` which doesn't depend on any package => No cycle dependency.

MehdiChinoune:
Ping!

jeremyd2019:
If you install the 'shared' headers for GMP, you cannot use the static GMP library (due to apis being defined dllimport).  With the 'static' headers, you *can* still use the shared GMP library (by virtue of the import thunks)

jeremyd2019:
Oh, I see GMP also has exported data.  Those would require runtime pseudo-relocs for using the shared lib with the static headers, and would probably run into 32-bit range issues on 64-bit with a high imagebase.  I don't know what can be done about that though, given:
>    There's no attempt to support GMP built both static and DLL.  Doing so
   would mean applications would have to tell us which of the two is going
   to be used when linking, and that seems very tedious and error prone if
   using GMP by hand, and equally tedious from a package since autoconf and
   automake don't give much help.

jeremyd2019:
> Oh, I see GMP also has exported data. If something used those, it would require runtime pseudo-relocs for using the shared lib with the static headers, and would probably run into 32-bit range issues on 64-bit with a high imagebase.

Nevermind, refptrs are involved in those that allows a full 64-bit relocation to be used.  So static headers should work fine with shared GMP, but not vice-versa

MehdiChinoune:
Either we fix gmp shared library or disable it, It couldn't be linked to if the header files are from the static build (As @jeremyd2019 explained above)

jeremyd2019:
That's the opposite of what I explained above.  You *can* link to the shared gmp library with the headers from the static build.  You *cannot* link to the static gmp library with the headers from the shared build.

As for mpfr, it attempts to decide how *it* is being built based on the gmp headers, rather than from its own configure script.  This is arguably wrong, except for the fact that the upstream developers already disclaimed all support for having one set of headers for both static and shared, so it's "by design".

PR #10810, opened 2022-02-19
FFmpeg: Add libssh, fribidi and vid.stab dependency

wszqkzqk:


lazka:
Just wondering, are you enabling these because you need them? Or just because you can?

wszqkzqk:
> Just wondering, are you enabling these because you need them? Or just because you can?
No, but archlinux added these to build full featured ffmpeg. And redundant features don't cost much, but it’s much harder to rebuild when to use these infrequently used functions.

wszqkzqk:
> Just wondering, are you enabling these because you need them? Or just because you can?

No, but archlinux added these to build full featured ffmpeg. And redundant features don't cost much, but it’s much harder to rebuild when to use these infrequently used functions.

PR #10813, opened 2022-02-20
pyqt6: update to 6.2.3

MehdiChinoune:


MehdiChinoune:
```sip-build: 'D:/M/msys64/ucrt64/bin/qmake6.exe -query' failed returning 3221225781```
I can't reproduce this failure on my machine.
Could someone test `qmake6 -query` and see if it works or not.

PR #10815, opened 2022-02-20
qt6-base: add runtime dependency on libb2

MehdiChinoune:
qmake6 failed to run (#10813) because of this new dependency.


MehdiChinoune:
I was wondering how didn't we encounter that failure before?
libb2 was added just recently as a dependency of libarchive which is a dependency of cmake.

lazka:
likely as you said, cmake pulls in libb2 at build time, qt6 uses it without us knowing it.

PR #10818, opened 2022-02-20
libsoup3: Add package for gst-plugins-good

Biswa96:
Please review.


lazka:
lgtm

We have to be careful where to include it though, libsoup2 and 3 can't be loaded in the same process I think.

lazka:
> We have to be careful where to include it though, libsoup2 and 3 can't be loaded in the same process I think.

ah, they worked around that in gstreamer at least: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/1044

We have to be careful with other libraries though.

PR #10821, opened 2022-02-20
[CI] Check runtime dependencies

MehdiChinoune:
This an initial work to check runtime dependencies at buildtime.
It checks runtime dependencies of all the `*.dll` and `*.exe` installed in `${MINGW_PREFIX}/bin`

Questions:
- [ ] Should we check other `*.dll` and `*.exe` files outside of `bin` directory?
I think most of them are just extensions/plugins and their dependencies are mostly listed under `optdepends`.
- [x] Should we check python modules `*.pyd`?

Known issues:
- [x] If `ntldd` could not find all the runtime dependencies it fails immediately, I tried to silent it but I don't have enough experience with shell/bash.
- [ ] Some of the `*.dll` and `*.exe` are optional (depends on optional packages), we could ignore them if we could fix the first issue.
- [ ] Split-packages: How to know if one package is using the runtime dependencies pulled by another independent non-conflicting package.

the qt-svg commit is just a test, I will remove it before marking this PR as ready.

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -90,6 +90,13 @@ for package in "${packages[@]}"; do
                     ntldd -R ${binary} | grep -v "ext-ms\|api-ms\|WINDOWS\|Windows\|HvsiFileTrust\|wpaxholder"
                 done
             fi
+            declare -a py_modules=($(pacman -Ql $pkgname | sed -e 's|^[^ ]* ||' | grep -E ${MINGW_PREFIX}/lib/python[0-9]\.[0-9]+/site-packages/.+\.pyd$))
+            if [ "${#py_modules[@]}" -ne 0 ]; then
+                for pyd in ${py_modules[@]}; do
+                    echo "${pyd}:"
+                    ntldd -R ${pyd} | grep -v "ext-ms\|api-ms\|WINDOWS\|Windows\|HvsiFileTrust\|wpaxholder"
+                done
+            fi
             echo "::endgroup::"
 
             echo "::group::[uninstall] ${pkgname}"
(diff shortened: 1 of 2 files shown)


lazka:
> * Should we check other `*.dll` and `*.exe` files outside of `bin` directory?
>   I think most of them are just extensions/plugins and their dependencies are mostly listed under `optdepends`.

I'd say ideally yes, but for that we need to install optdepends somehow, right?

> * Should we check python modules `*.pyd`?

why not, they are just DLLs after all.

MehdiChinoune:
> > * Should we check other `*.dll` and `*.exe` files outside of `bin` directory?
> >   I think most of them are just extensions/plugins and their dependencies are mostly listed under `optdepends`.
> 
> I'd say ideally yes, but for that we need to install optdepends somehow, right?
> 
> > * Should we check python modules `*.pyd`?
> 
> why not, they are just DLLs after all.

they could be added later if someone figures out an efficient way to do it.

I think these changes are good enough as a first step.

lazka:
ok, we can try I guess :)

MehdiChinoune:
Ready

MehdiChinoune:
I have added checking `*.pyd` files runtime dependencies (only the ones installed under `${MINGW_PREFIX}/lib/python[0-9]\.[0-9]+/site-packages` folder).

MehdiChinoune:
ping!

lazka:
> * Split-packages: How to know if one package is using the runtime dependencies pulled by another independent non-conflicting package.

That could be fixed by explicitly duplicating the dependencies, which is imo more correct anyway.

MehdiChinoune:
Thanks for merging!

PR #10823, opened 2022-02-20
scotch: Use split package for 32-bit and 64-bit indexing versions.

mmuetzel:
The two versions of the package got out of sync. Use a split package to make it easier to update both versions at the same time.

This also includes the changes from #10780.


mmuetzel:
Hmmm...
It looks like the CI doesn't like if split packages conflict with themselves.
The rules for `mingw-w64-scotch` and `mingw-w64-scotch-int64` were almost identical. So I figured that would be a prime candidate for a split package.
Feel free to close if this isn't a use case for a split package.

MehdiChinoune:
Just put an empty `.ci-sequential` in mingw-w64-scotch directory.

mmuetzel:
> Just put a file named `.ci-sequential` in mingw-w64-scotch directory.

Thanks! That helped.

mmuetzel:
Oops. Something is wrong with those libraries. They don't export anything...

mmuetzel:
I should have checked without `-D`. But the symbols were indeed not exported. (I guess that's also an issue with #10780.)
Should be fixed with the latest change.

PR #10555, opened 2022-01-16
rust: Update to 1.58.0

filnet:

PR #10824, opened 2022-02-20
clang: add coff fix section name encoding patch

filnet:
The patch fixes stripping of large static libraries.
The stripping issue was impacting LLVM itself.
LLVM had at least one invalid static library after stripping.

As the current LLVM is affected by the issue, we need to do two builds:
- a first build to produce a fixed LLVM
- and a second one to produce a properly stripped package (using the 1st build).

see https://github.com/msys2/MINGW-packages/pull/10555#issuecomment-1046088716


mati865:
We have 2 options to proceed with this:
- temporarily disable stripping for CLANG* subsystems
- rebuild LLVM twice

jeremyd2019:
Should we prefer backporting the full fix, now that that has 'landed'?

filnet:
My tests show that even with the patch the rust build fails. So no urgency on my side unless we want to reproduce it in CI..

BTW, is it possible to download packages built by CI ?

jeremyd2019:
> BTW, is it possible to download packages built by CI ?

Yes, go to the 'checks' tab and there should be an 'Artifacts' drop-down near the top right.

filnet:
I tested the patched LLVM with rust and it looks good.

filnet:
> Should we prefer backporting the full fix, now that that has 'landed'?

I had a quick look and the full fix is on top of major refactoring.

jeremyd2019:
The backport to the 14.x branch applied cleanly (I pushed it here, but am continuing to test locally where it builds faster).

jeremyd2019:
local build of 2x mingw-w64-clang and 1x mingw-w64-rust 1.58.1 succeeded with upstream patches from 14.x branch.

I also made a new split package on mingw-w64-clang that contains the gcc aliases (mingw-w64-clang-*-gcc-compat).  This inherits the '-toolchain' group from the top-level of the PKGBUILD so hopefully users who want to use tools that assume gcc will be able to continue to use them, and packages whose logic is tied to the name 'gcc' like perl can makedepend on it explicitly.

jeremyd2019:
CI run from my fork (before the latest rebase): https://github.com/jeremyd2019/MINGW-packages/actions/runs/1890450317

filnet:
Should I create a new PR for the 2nd build ?

lazka:
(I don't need one at least)

filnet:
> (I don't need one at least)

Sorry, I did not understand.

lazka:
sry, I meant I don't need a PR to continue this.

PR #10827, opened 2022-02-20
scotch: update to version 7.0.1

mmuetzel:
If stripping is enabled, some of the shared libraries don't export symbols. I don't know the reason.
Disabling stripping fixes that issue.


mati865:
Which subsystems are affected by strip issue?

mmuetzel:
In MINGW64 for the 32-bit indexing version, `libscotch.dll` exports symbols, `libptscotch.dll` doesn't. For the 64-bit indexing version, it is the other way round.

PR #10838, opened 2022-02-21
python-colour-science: update to 0.4.0

kmilos:
Also sync w/ latest template

### DIFF
--- a/mingw-w64-python-colour-science/PKGBUILD
+++ b/mingw-w64-python-colour-science/PKGBUILD
@@ -2,49 +2,52 @@
 # Contributor: Miloš Komarčević <miloskomarcevic@aim.com>
 
 _realname=colour-science
-_pypkgname=colour
 pkgbase=mingw-w64-python-${_realname}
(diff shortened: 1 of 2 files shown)


lazka:
https://github.com/colour-science/colour/issues/945 thanks

We can update setuptools if needed, just needs some patching.

kmilos:
Looks like there is a patch upstream already, will wait for that...

kmilos:
Or I could also try switching to `python-build` [like Arch did](https://aur.archlinux.org/cgit/aur.git/commit/?h=python-colour-science&id=e39f2987a27d78a5a3cf19e7101d3bb743021f59)...

naveen521kk:
Slightly editing the pyproject.toml should do the trick without poetry being installed (instead using poetry-core) as mentioned in https://python-poetry.org/docs/pyproject/#poetry-and-pep-517

kmilos:
@naveen521kk Indeed, thanks!

naveen521kk:
Also, I think you can make that change upstream.

kmilos:
Yeah, it seems it recommended anyway, I let upstream know.

Issue #10843, opened 2022-02-22
OpenOCD broken by newest libusb (mingw-w64-x86_64-openocd-0.11.0-1 + mingw-w64-x86_64-libusb-1.0.25-2)


radioflash:
A recent update broke OpenOCD for me.
I narrowed it down to a segfault in libusb-- replacing libusb-1.0.dll (from the "mingw-w64-x86_64-libusb-1.0.25-2"-package) with an older version of the file circumvents the problem for me, no other change is necessary (reinstalling libusb with `pacman -S mingw-w64-x86_64-libusb` also breaks things again).

This is an insiduous problem because it results in a segfault ONLY when actually connecting to the debugger.
Just running openocd.exe without really doing anything still works fine:

> C:\msys64\mingw64\bin>openocd.exe --version
> Open On-Chip Debugger 0.11.0
> Licensed under GNU GPL v2
> For bug reports, read
>         http://openocd.org/doc/doxygen/bugs.html

This also still works when the debugger is not connected:

> $ openocd -f openocd-chip-1.cfg -c "targets stm32f7x.cpu"
> Open On-Chip Debugger 0.11.0
> Licensed under GNU GPL v2
> For bug reports, read
>         http://openocd.org/doc/doxygen/bugs.html
> Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
> Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
> DEPRECATED! use 'adapter speed' not 'adapter_khz'
> adapter speed: 8000 kHz
> 
> Info : Listening on port 6666 for tcl connections
> Info : Listening on port 4444 for telnet connections
> Info : clock speed 8000 kHz
> Error: open failed

But when the debugger is connected, then it segfaults:

> (gdb) r
> Starting program: C:\msys64\mingw64\bin\openocd.exe -f openocd-chip-1.cfg -c "targets stm32f7x.cpu"
> [New Thread 8592.0x3004]
> [New Thread 8592.0x4658]
> [New Thread 8592.0x4d80]
> Open On-Chip Debugger 0.11.0
> Licensed under GNU GPL v2
> For bug reports, read
>         http://openocd.org/doc/doxygen/bugs.html
> Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
> Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
[...]

PR #10848, opened 2022-02-23
Fix OpenOCD to work with libusb >= v1.0.25

radioflash:
By integrating a patch from OpenOCD upstream git repo.
The next OpenOCD release (probably v0.12.0) is going
to obviate this patch.

Only the OpenOCD stlink backend is affected by this,
and libusb versions 1.0.24 and older would also not
trigger the segfault.


radioflash:
This should fix issue #10843

Biswa96:
Thanks for working on this fix 🤗

PR #10851, opened 2022-02-24

kmilos:


jeremyd2019:
the arm builder is going to be busy for a while doing two builds of clang, so this will have to wait probably 24 hours

kmilos:
Thanks for the info. There is no rush.

jeremyd2019:
ci run https://github.com/msys2-arm/MINGW-packages/runs/5327957034?check_suite_focus=true

PR #10853, opened 2022-02-24
FLAC: update to 1.3.4

drfiemost:


lazka:
I already tried, and the static build is broken somehow.

Biswa96:
Do you mean at runtime or the build is broken? I've tried with cmake and it builds fine. But need to rebuild all the depenencies sue to DLL change.

lazka:
The build was broken (shared worked, static complained about libssp things)

Maybe I missed something. cmake + name change is fine too.

drfiemost:
I don't see any useful error message in the log...
The fun thing is that the changelog says "build system: Fixes for MinGW and MSVC"
So, would it be better to switch to cmake?

Biswa96:
> So, would it be better to switch to cmake?

Yes, cmake is preferred over autotools because cmake is faster in mingw environment. Also cmake is easier to use and tinker with but opinion may vary 😊 If you want I can try to send the change in a pull request.

PR #10857, opened 2022-02-25
qt-installer-framework: update to 4.3.0

MehdiChinoune:


lazka:
Built the installer with this, and looks good. Thanks!

MehdiChinoune:
> Built the installer with this, and looks good. Thanks!

Could you the one from CLANG64, I think It's broken.

MehdiChinoune:
> Built the installer with this, and looks good. Thanks!

Could you the one from CLANG64, I think It's broken.

Issue #9670, opened 2021-09-27
Add libjxl support?


Jehan:
GIMP recently added JPEG XL support in its dev version, which uses [libjxl](https://github.com/libjxl/libjxl).

We'd welcome if msys2 provided a package for libjxl. 😉

novomesk:
Please wait with the request. I will provide some more information with my experiences with libjxl under MSYS2.

Jehan:
For the record, @novomesk is the GIMP contributor who implemented the support. Please follow their lead/opinion on this matter rather than mine. 😉

Biswa96:
The libjxl project needs some pampering. It finds dependencies as git modules instead of packages.

novomesk:
libjxl uses bundled dependencies, so after unpacking `libjxl-0.5.tar.gz`, `./deps.sh` must be used to download the dependencies or everything should be downloaded with **--recursive** option.

`git clone --depth 1 https://github.com/libjxl/libjxl.git --recursive`

Prior building libjxl it is good to have following packages:
```
mingw64/mingw-w64-x86_64-highway
mingw64/mingw-w64-x86_64-brotli
```

Suggested configuration:
`cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/mingw64 -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DBUILD_TESTING=OFF -DJPEGXL_ENABLE_MANPAGES=OFF -DJPEGXL_ENABLE_EXAMPLES=OFF -DJPEGXL_ENABLE_PLUGINS=ON -DJPEGXL_ENABLE_SJPEG=OFF -DJPEGXL_ENABLE_SKCMS=ON -DJPEGXL_FORCE_SYSTEM_BROTLI=ON -DJPEGXL_FORCE_SYSTEM_HWY=ON -DCMAKE_CXX_FLAGS="-DHWY_COMPILE_ONLY_SCALAR" ../libjxl-0.5/`

I strongly recommend **Ninja** generator, _Unix Makefiles_ doesn't always work.

libjxl can work with lcms2 or skcms color management systems, but it works better with skcms which is bundled dependency: `-DJPEGXL_ENABLE_SKCMS=ON`

`-DJPEGXL_FORCE_SYSTEM_BROTLI=ON -DJPEGXL_FORCE_SYSTEM_HWY=ON` - it will use highway and brotli packages from MSYS2 (instead of bundled dependencies).

[...]

PR #10859, opened 2022-02-25
libjxl: Add package

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/9670

### DIFF
--- a/mingw-w64-libjxl/PKGBUILD
+++ b/mingw-w64-libjxl/PKGBUILD
@@ -1,0 +1,101 @@
+# Maintainer: Biswapriyo Nath <nathbappai@gmail.com>
+
+_realname=libjxl
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.6.1
+pkgrel=1
+pkgdesc='JPEG XL image format reference implementation (mingw-w64)'
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+url="https://jpeg.org/jpegxl/"
+license=('BSD')
+depends=("${MINGW_PACKAGE_PREFIX}-brotli"
+         #"${MINGW_PACKAGE_PREFIX}-highway"
+         "${MINGW_PACKAGE_PREFIX}-gcc-libs"
+         "${MINGW_PACKAGE_PREFIX}-giflib"
+         "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo"
+         "${MINGW_PACKAGE_PREFIX}-libpng"
+         "${MINGW_PACKAGE_PREFIX}-libwinpthread")
+makedepends=("git"
+             "${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
+# There is no guarantee libjxl will follow stable highway release
+# Also libjxl installs the bundled highway binaries
+conflicts=("${MINGW_PACKAGE_PREFIX}-highway")
+source=("git+https://github.com/libjxl/libjxl.git#tag=v${pkgver}"
+        'git+https://github.com/lvandeve/lodepng.git'
+        'git+https://github.com/webmproject/sjpeg.git'
+        'git+https://skia.googlesource.com/skcms.git'
+        'git+https://github.com/google/highway.git')
+sha256sums=('SKIP'
+            'SKIP'
+            'SKIP'
+            'SKIP'
+            'SKIP')
+
+prepare() {
+  cd "${_realname}"
+  git submodule init
+  git config --local submodule.third_party/lodepng.url "${srcdir}/lodepng"
+  git config --local submodule.third_party/sjpeg.url "${srcdir}/sjpeg"
+  git config --local submodule.third_party/skcms.url "${srcdir}/skcms"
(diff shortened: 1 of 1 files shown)


lazka:
I wonder if the package should contain all the `libhwy` stuff or if that's just supposed to be internal.

lazka:
Maybe we should package https://archlinux.org/packages/community/x86_64/highway/ as well first and use that?

novomesk:
You can enable `-DJPEGXL_ENABLE_TOOLS=ON` to build **cjxl** and **djxl** command line utilities (compressor and de-compressor to/from JXL format).

In GIMP we did not build them because in GIMP we need only libraries.

However, what is very important for MSYS2 gcc build is the `-DHWY_COMPILE_ONLY_SCALAR` because otherwise library may crash on AVX2 processors during encoding/decoding.

kmilos:
We already package [highway](https://packages.msys2.org/base/mingw-w64-highway) in anticipation of libjxl 😉 

Though I'd personally wait for upcoming 0.7 when they finally debundle all the internal stuff...

Biswa96:
> You can enable -DJPEGXL_ENABLE_TOOLS=ON to build cjxl and djxl command line utilities 

For the tools, libjxl needs openexr which itself needs a new dependency imath.

kmilos:
> For the tools, libjxl needs openexr which itself needs a new dependency imath.

It doesn't work with the available OpenEXR 2.5, it must be >=3.0?

novomesk:
openexr support is optional. I suggest to start with `-DJPEGXL_ENABLE_OPENEXR=OFF`

Biswa96:
I have enabled tools, disabled openexr (need some work) and added `-DHWY_COMPILE_ONLY_SCALAR`. Please test the new packages.

novomesk:
I will try to test tomorrow or on Monday.

mingw-w64-libjxl/PKGBUILD
```diff
@@ -0,0 +1,99 @@
...
+  if check_option "debug" "n"; then
+    _extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    _extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  [[ -d "${srcdir}/build-${MSYSTEM}" ]] && rm -rf "${srcdir}/build-${MSYSTEM}"
+  mkdir -p "${srcdir}/build-${MSYSTEM}" && cd "${srcdir}/build-${MSYSTEM}"
+
+  CPPFLAGS+=" -DHWY_COMPILE_ONLY_SCALAR" \
```
novomesk, line 67: This CPPFLAGS have no effect because library was built with AVX2 support anyway. See the following output, there is AVX2 and SSE4 mentioned beside SCALAR:

```
$ djxl.exe input.jxl test-output.png
JPEG XL decoder v0.6.1 53d08bbe [AVX2,SSE4,Scalar]
Read 8772 compressed bytes.
Decoded to pixels.
1024 x 1024, 11.11 MP/s [11.11, 11.11], 1 reps, 4 threads.
Allocations: 440 (max bytes in use: 5.611977E+07)
```

I think with CXXFLAGS it will do what expected.

But what is interesting that it is not crashing now. It used to crash in November 2021. Are you aware of some AVX2 related fixes/patches/upgrades in mingw used by MSYS2?

These are the JXL files which used to crash during decoding on AVX2-enabled processors (you can test using `djxl.exe`), but it was not crashing on older CPUs:

https://github.com/libjxl/conformance/blob/master/testcases/alpha_premultiplied/input.jxl

https://github.com/novomesk/qt-jpegxl-image-plugin/blob/main/testfiles/jpegxl-logo.jxl

Biswa96: Some AVX related things happened with mingw-w64-gcc package. But that was a year ago. So, the `-DHWY_COMPILE_ONLY_SCALAR` workaround is not required anymore, right?

novomesk: Well, it seems that the workaround is not needed now.

But I had same impression in the past too and then the problem re-appeared.

I would still continue to use it. Better safe than sorry.

Biswa96: The constant is checked in processor as `#if defined(HWY_COMPILE_ONLY_SCALAR)` in highway source code. Are you sure that it should be in CXXFLAGS instead of CPPFLAGS?

novomesk: I think it doesn't work with CPPFLAGS because of **AVX2,SSE4,Scalar** message.

That's why I suggest to try CXXFLAGS

Biswa96: Thanks. I have changed it to CXXFLAGS. cjxl and djxl now shows only scalar.

Biswa96:
Let me know if this is ready to merge 😊

kmilos:
Is there perhaps a CMake option for this instead? As CMake doesn't understand CPPFLAGS, you typically need to define both CXXFLAGS _and_ CFLAGS if it is a mixed C/C++ project, so a CMake option is preferable...

Biswa96:
OK, I shall add that with CLFAGS also. AFAIK, [CXXFLAGS] and [CFLAGS] are stored by cmake automatically.

[CXXFLAGS]: https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html
[CFLAGS]: https://cmake.org/cmake/help/latest/envvar/CFLAGS.html

novomesk:
For me it is good as it is now.

It makes no harm to add CFLAGS too but it works also in this form.

Biswa96:
Alright. Done.

PR #10861, opened 2022-02-25
svt-av1: update to 0.9.1

kmilos:


kmilos:
The MINGW32 failure is expected, as this is supposed to be a x86_64 only package.

Issue #10396, opened 2021-12-22
openexr and openimageio need separate imath package


revelator:
missing dependency iMath formerly part of openexr needs to be built seperatly now -> https://github.com/AcademySoftwareFoundation/Imath.
fails to find openjpeg2 upstream issue.

Biswa96:
I tried to build IMath in CI. But CLANG64 fails with error ld.lld: error: too many exported symbols (max 65535).

revelator:
looks like it will need a def with clang then.

revelator:
also something to consider namely building it as a static library.
the limit will be tough to work around as it is not imposed by the build utils but is a hard windows limit, maybe use -Wl,--needed with the linker ?.

revelator:
hmm another bug cropped up the git version needs either drand48 or rand_s, had to hack in the drand48 from openbsd.
openvdb also blows up on the new tbb and despite getiing told to use the mingw python it forces use of the /usr/bin/python.exe instead.

revelator:
small fix for imath -> put this in half_perf_test.cpp

```
// neither drand48 nor s_rand exist for mingw-w64 tagets so emulate it.
#ifdef __MINGW32__
#define drand48() (((double)rand())/((double)RAND_MAX))
#endif
```

got it to build now, i still had the older python3.8 which for some reason is none to happy with the msys2 version being higher, had to update all python packages.

kmilos:
> small fix for imath -> put this in half_perf_test.cpp

Seems they [reimplemented drand48() already which should be preferred](https://github.com/AcademySoftwareFoundation/Imath/blob/304288542b6d3d7b4555539a0a1e2dd9da571c53/src/Imath/ImathRandom.cpp)? Have you tried replacing w/ Imath::drand48() instead?

I guess this could also be [reported upstream](https://github.com/AcademySoftwareFoundation/Imath/issues/new/choose)?

Biswa96:
I have already sent a pull request to upstream. It's pending. They have some weird legal mess.

Biswa96:
openexr can be built with imath. But all the dependent packages have to be rebuilt and those are so huge for my tiny old box.

Biswa96:
Last query before adding the package. openexr has many patches which add version number with DLL names. Should I continue those patching? Or keep the default name without version number set by build system?

Biswa96:
Any opinion about the DLL name? ☝️

PR #10865, opened 2022-02-26
imath: Add package for openexr

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/10396

Please review @revelator @kmilos

### DIFF
--- a/mingw-w64-imath/0001-test-fix-build.patch
+++ b/mingw-w64-imath/0001-test-fix-build.patch
@@ -1,0 +1,38 @@
+--- a/src/ImathTest/half_perf_test.cpp
++++ b/src/ImathTest/half_perf_test.cpp
+@@ -3,7 +3,7 @@
+ // Copyright Contributors to the OpenEXR Project.
+ //
+ 
+-#ifdef _MSC_VER
++#ifdef _WIN32
+ #    define _CRT_RAND_S
+ #endif
+ 
+@@ -14,7 +14,7 @@
+ #include <stdlib.h>
+ #include <limits.h>
+ #include <string.h>
+-#ifdef _MSC_VER
++#ifdef _WIN32
+ #    include <windows.h>
+ #else
+ #    include <time.h>
+@@ -267,7 +267,7 @@
+ int64_t
+ get_ticks (void)
+ {
+-#ifdef _MSC_VER
++#ifdef _WIN32
+     static uint64_t scale = 0;
+     if (scale == 0)
+     {
+@@ -381,7 +381,7 @@
+             perf_test_half_to_float (floats, halfs, numentries);
+ 
+             // test float -> half with real-world values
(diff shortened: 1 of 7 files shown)


kmilos:
Thanks for moving forward w/ this!

LGTM, but do we need to handle conflicts w/ legacy `ilmbase` from OpenEXR 2.5 (the Half library etc.)? Any chance we can bump OpenEXR at the same time? Would love to help, but short on time lately 😞

kmilos:
Scratch that, just saw you did it already!!! 👍

Biswa96:
> but do we need to handle conflicts w/ legacy ilmbase from OpenEXR 2.5 (the Half library etc.)?

Yea, I am about to ask that. I can not figure out how to solve the conflicts. The files from ilmbase are scattered between imath and openexr. If I tried to add provides/conflicts/replaces in both packages CI failed with error.

There is another thing that I can not figure out. The clang build fails with a C++ error. Jeremy provided me some information about that same error with kirigami-qt5 package. But the rabbit hole is too dark for me.

kmilos:
I guess we make all packages that were depending on ilmbase depend on OpenEXR on to be safe? Or we start dissecting what exactly they need...

Biswa96:
Yes, dependent packages have to be rebuilt. But replacing the ilmbase with openexr in _depends_ list would not solve the file conflicts at install time, right?

jeremyd2019:
> There is another thing that I can not figure out. The clang build fails with a C++ error. Jeremy provided me some information about that same error with kirigami-qt5 package. But the rabbit hole is too dark for me.

For the record, this was llvm/llvm-project#37457

kmilos:
And I see now you already build all the Python bindings, but might be an idea to package them separately as `pyimath` and conflict/replace `pyilmbase` as well?

kmilos:
Actually, might be ok to start including the Python bindings in the main package like https://packages.msys2.org/base/mingw-w64-openimageio (and have the optional python dependency), but I guess we still need the conflicts/replaces w/ the old separate Python bindings package...

kmilos:
FYI, 3.1.5 is available, can drop some patches...

revelator:
sorry been away  a while due to problems with some screws in my back getting loose :S, changes look ok to me so far but let me see if anything crops up as i notice numerous changes lately.

Biswa96:
Need to fix two things - clang build and provides/replaces/conflicts entries. I have not figured those out yet. Suggestions/fixes are welcomed.

kmilos:
> FYI, 3.1.5 is available, can drop some patches...

OpenEXR is bumped to 3.1.5 now as well.

> provides/replaces/conflicts entries

How about just

`replaces=("${MINGW_PACKAGE_PREFIX}-ilmbase" "${MINGW_PACKAGE_PREFIX}-pyilmbase")`

for `imath`? Should be it, no?

PR #10868, opened 2022-02-26
Add `jj`, an experimental DVCS

dscho:
[`jj`](https://github.com/martinvonz/jj) (also known as Jujutsu) currently comes with two backends, one that uses Git's native storage, making it interoperable with Git.

### DIFF
--- a/mingw-w64-jj/PKGBUILD
+++ b/mingw-w64-jj/PKGBUILD
@@ -1,0 +1,62 @@
+# Maintainer: Johannes Schindelin <johannes.schindelin@gmx.de>
+
+_realname=jj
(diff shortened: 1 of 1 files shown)


jeremyd2019:
As a practical matter, due to the implementation of autobuild, packages really should be pinned to a commit, for a couple of reasons
1. to make sure every variant package built is from the same commit, in case a new commit is pushed while autobuild is processing the different environments.
2. there is no trigger for autobuild to rebuild a package other than a git commit with an increase in version, so the published packages would be effectively pinned anyway until the PKGBUILD is updated.

dscho:
Fair enough, I adjusted it.

Issue #10869, opened 2022-02-26
SDL2-mixer broken static libs


hessenfarmer:
With the latest build the static linking of SDL2-mixer package leads to undefined references to FLAC like 

undefined reference to `__imp_FLAC__stream_decoder_new

see https://ci.appveyor.com/project/widelands-dev/widelands/builds/42716917/job/2suiogasvac7qhyb for example

Biswa96:
Hi @hessenfarmer, Thank you for reporting the issue. I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/1905227196

Recently, I have changed the flac package build system to cmake. This kind of issue is expected with cmake and autotools.

PR #10870, opened 2022-02-27
sdl2_mixer: Fix static library and linking with FLAC

Biswa96:
dllimport attribute is not added if FLAC__NO_DLL is defined.

Fixes https://github.com/msys2/MINGW-packages/issues/10869

### DIFF
--- a/mingw-w64-SDL2_mixer/PKGBUILD
+++ b/mingw-w64-SDL2_mixer/PKGBUILD
@@ -4,7 +4,7 @@ _realname=SDL2_mixer
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.0.4
-pkgrel=6
+pkgrel=7
 pkgdesc="A simple multi-channel audio mixer (Version 2) (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


hessenfarmer:
> Now FLAC symbols are not present in static library. Here is the diff:
> 
> ```diff
> -                 U __imp_FLAC__stream_decoder_delete
> -                 U __imp_FLAC__stream_decoder_finish
> -                 U __imp_FLAC__stream_decoder_flush
> -                 U __imp_FLAC__stream_decoder_get_state
> -                 U __imp_FLAC__stream_decoder_init_stream
> -                 U __imp_FLAC__stream_decoder_new
> -                 U __imp_FLAC__stream_decoder_process_single
> -                 U __imp_FLAC__stream_decoder_process_until_end_of_metadata
> -                 U __imp_FLAC__stream_decoder_process_until_end_of_stream
> -                 U __imp_FLAC__stream_decoder_seek_absolute
>                   U __imp_midiOutPrepareHeader
>                   U __imp_midiOutSetVolume
>                   U __imp_midiOutUnprepareHeader
> ```

does this mean the package has different dependencies now depending on whether its static lib or its dynamic lib is used?

Biswa96:
> does this mean the package has different dependencies now depending on whether its static lib or its dynamic lib is used?

No, no. Dependencies are same.

PR #10893, opened 2022-03-02
asciidoctor: Update to 2.0.17

lazka:


lazka:
this just made it clear that ruby is broken with clang32. closing for now.

PR #10898, opened 2022-03-03
ruby: patch ucrtbase pattern for 32-bit on Win11/Server2022.

### LINKED ISSUE #10896: ruby: broken in CLANG32 on Windows 11/Server 2022
yeah, ruby is broken there, just running it gives the same error. Downgrading it doesn't help.

_Originally posted by @lazka in https://github.com/msys2/MINGW-packages/issues/10878#issuecomment-1057397880_

Results of investigation:
https://github.com/ruby/ruby/blob/v3_0_3/win32/win32.c#L2591 😱

wow, yeah, `ucrtbase!_isatty` disassembly is majorly different on win11 
I think the code would actually still work, except they no longer use `pop ebp` as the last instruction before `ret`, but rather `leave`
so the pattern no longer finds the end of the function
but given the code is majorly changed, there's a chance the internal struct has changed too
but this is not limited to us, either.  32-bit upstream ruby should be broken on Windows 11 too.  Maybe there's not much use of that?

jeremyd2019:
Fixes #10896.  For now.  At least in the case I saw and tested.  This is some messed up code.


lazka:
Thanks!

PR #10899, opened 2022-03-03
portmidi: upgrade to 2.0.3

kmilos:
Switching to new repository and codebase: https://github.com/PortMidi/portmidi


Biswa96:
Should those test executables be installed? If yes, can the patch file be upstreamed?

kmilos:
They were in the previous package (directly done inside `package()`), so I thought I'd keep them... Indeed, not installed by default by upstream, but e.g. OpenSUSE distributes them as well (w/ `portmidi-` prefix though)...

I'm ok with skipping them as well...

lazka:
I think it's fine to drop them.

kmilos:
Done, vanilla it is.

PR #10913, opened 2022-03-05
cmake: update to 3.22.3 (build with cmake)

MehdiChinoune:


lazka:
Interesting. Why do you think we should do this?

MehdiChinoune:
> Interesting. Why do you think we should do this?

Just for a cleaner configuration (better detection of system libs and Fortran compiler) and better output.

lazka:
OK, thanks. Let's give it a try then.

PR #10927, opened 2022-03-06
Adding mingw-w64-rrdtool

jeroen:


mingw-w64-rrdtool/PKGBUILD
```diff
@@ -0,0 +1,48 @@
+_realname=rrdtool
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=1.7.2
+pkgrel=1
+pkgdesc="Round Robin Database (mingw-w64)"
+arch=('any')
+url='https://oss.oetiker.ch/rrdtool/index.en.html'
+license=('GPL2')
+makedepends=("autoconf" "automake" "libtool")
```
Biswa96, line 10:
```suggestion
makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
                          "${MINGW_PACKAGE_PREFIX}-cc")
```

Please try with this.

mingw-w64-rrdtool/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+    --disable-tcl \
+    --disable-python \
+    --disable-docs \
+    --disable-nls \
+    --disable-perl \
+    --disable-mmap \
+    --disable-ruby \
+    --disable-rrdcached
+
+  make CFLAGS='-D__USE_MINGW_ANSI_STDIO=1'
```
Biswa96, line 42: `__USE_MINGW_ANSI_STDIO` is already defined in CPPFLAGS in /etc/makepkg_mingw.conf file.

mingw-w64-rrdtool/PKGBUILD
```diff
@@ -0,0 +1,48 @@
+_realname=rrdtool
```
Biswa96, line 1: Add maintainer at the first line.

jeroen:
Thanks! All fixed now

Biswa96:
One thing looks different to me. The executables are prefixed with target triplet like x86_64-w64-mingw32-rrdtool.exe. Shouldn't it be only rrdtool.exe? If I remove `--host` and `--target` options from configure command the executables are not prefixed ones.

@lazka Is it required to have `--host` and `--target` options in general? Every PKGBUILD has those.

lazka:
In theory it shouldn't change anything since we set the default here: https://github.com/msys2/MSYS2-packages/blob/master/filesystem/main-config.site

Not sure

Biswa96:
OK. I have seen this type of prefixed executable name in libaacs project. I have sent an email in their mailing list here https://mailman.videolan.org/private/libaacs-devel/2022-March/000716.html. One has to subscribe to see it. Let me check if there is any issue in autotools files.

Biswa96:
I have added a commit to fix the prefixed binaries by removing host and target options, here https://github.com/msys2/MINGW-packages/commit/32e9ec26185d5a970c95d0f44e815940f1c20014. Interestingly, two new binaries appear:

```diff
 ucrt64/
 ucrt64/bin/
-ucrt64/bin/x86_64-w64-mingw32-rrdcgi.exe
-ucrt64/bin/x86_64-w64-mingw32-rrdtool.exe
-ucrt64/bin/x86_64-w64-mingw32-rrdupdate.exe
+ucrt64/bin/rrdcgi.exe
+ucrt64/bin/rrdcreate.exe
+ucrt64/bin/rrdinfo.exe
+ucrt64/bin/rrdtool.exe
+ucrt64/bin/rrdupdate.exe
 ucrt64/include/
 ucrt64/include/rrd.h
 ucrt64/include/rrd_client.h
```

PR #10929, opened 2022-03-06
pyside2: update to 5.15.3

MehdiChinoune:


Biswa96:
Just noticed this. shiboken package may have hardcoded path of clang++.exe file. Here is the output of grep:

```

PR #10930, opened 2022-03-06
seacas: update to 2022-03-04

3rav:


Biswa96:
The CMake files in seacas package contain full path reference. Like these:

```
$ grep -ir msys64 mingw64
mingw64/lib/cmake/SEACAS/SEACASConfig.cmake:set(SEACAS_CXX_COMPILER "D:/M/msys64/mingw64/bin/c++.exe")
mingw64/lib/cmake/SEACAS/SEACASConfig.cmake:set(SEACAS_C_COMPILER "D:/M/msys64/mingw64/bin/cc.exe")
mingw64/lib/cmake/SEACAS/SEACASConfig.cmake:set(SEACAS_LINKER D:/M/msys64/mingw64/bin/ld.exe)
...
...
```

Is this possible to use `sed` to remove/replace that full path? For example, see mingw-w64-libwebsockets has a `for` loop with `sed` command in package function. Make sure it does not change anything incorrectly.

Biswa96:
I have added the required commands to remove full path reference in cmake files.

PR #10934, opened 2022-03-07
xmake: 2.6.4

waruqi:


Biswa96:
The clang build fails because the clang package now does not provide the gcc aliases. The xmake build system can check CC and CXX variable to get the compiler name. Those variables are set in /etc/makepkg-mingw.conf file.

waruqi:
> The clang build fails because the clang package now does not provide the gcc aliases. The xmake build system can check CC and CXX variable to get the compiler name. Those variables are set in /etc/makepkg-mingw.conf file.

Why is the previous version OK, clang is the new build environment?

waruqi:
I can fix them internally, but will probably have to wait until version 2.6.5 to try again.

Biswa96:
> Why is the previous version OK

Because of the same reason. Previously, the clang package provided gcc aliases. In other words, /clang64/bin/gcc.exe existed but not now.

> but will probably have to wait until version 2.6.5 to try again.

No problem. As cmake packages [exist in clang environment][1], only those will have lower versions that mingw & ucrt ones.

[1]: https://packages.msys2.org/base/mingw-w64-xmake

MehdiChinoune:
You could fix it by replacing `${MINGW_PACKAGE_PREFIX}-cc` by `${MINGW_PACKAGE_PREFIX}-gcc` in `makedepends()`

waruqi:
> You could fix it by replacing `${MINGW_PACKAGE_PREFIX}-cc` by `${MINGW_PACKAGE_PREFIX}-gcc` in `makedepends()`

Oh, I will try it. thanks!

PR #10953, opened 2022-03-10
Fix runtime  dependencies (13/n)

MehdiChinoune:


MehdiChinoune:
@naveen521kk could you fix the ci-check.py script to show error output from pacman instead of just `failed to install`.

naveen521kk:
> @naveen521kk could you fix the ci-check.py script to show error output from pacman instead of just `failed to install`.

It does show? https://github.com/msys2/MINGW-packages/runs/5492123237?check_suite_focus=true#step:8:264

MehdiChinoune:
> > @naveen521kk could you fix the ci-check.py script to show error output from pacman instead of just `failed to install`.
> 
> It does show? https://github.com/msys2/MINGW-packages/runs/5492123237?check_suite_focus=true#step:8:264

The failure because of conflicting reverse dependencies should be fixed.

mmuetzel:
> The failure because of conflicting reverse dependencies should be fixed.

It skipped that test afaict. IIUC, the checks are failing because of this:
```
Pip Output: mingw-w64-x86_64-python-ldap-3.2.0-3-any.pkg.tar.zst
  Pip Check Failed for mingw-w64-x86_64-python-ldap
Here is what it failed.
python-ldap 3.2.0 requires pyasn1, which is not installed.
python-ldap 3.2.0 requires pyasn1-modules, which is not installed.
```

PR #10955, opened 2022-03-10
slim-simulator: Remove check for Qt version.

### LINKED ISSUE #10952: Run-time Qt version 5.15.3 does not match compile-time Qt version 5.15.2
I'm trying to run an app on windows (slimGUI.exe, mingw-w64-x86_64-slim-simulator) and Qt is one of the dependancies. I followed the same download protocol today as I have on previous machines but now with this download when I try to run the gui I get the following message on the console,

"Run-time Qt version 5.15.3 does not match compile-time Qt version 5.15.2"

Is it possible for the Qt version on Msys2 to be updated to version 5.15.3?

mmuetzel:
The check for the exact Qt version doesn't seem to be necessary. Usually, the Qt5 libraries are API compatible between minor updates.
If they are not, we'd need to rebuild all dependencies of Qt5 anyway (including slim-simulator).

Fixes #10952.

### DIFF
--- a/mingw-w64-slim-simulator/PKGBUILD
+++ b/mingw-w64-slim-simulator/PKGBUILD
@@ -2,38 +2,57 @@ _realname=slim-simulator
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.7.1
-pkgrel=1
+pkgrel=2
 pkgdesc="SLiM forward simulation software package for population genetics (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64')
 url="https://messerlab.org/slim/"
-license=("GPL")
-makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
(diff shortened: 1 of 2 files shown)


mingw-w64-slim-simulator/PKGBUILD
```diff
@@ -2,38 +2,57 @@ _realname=slim-simulator
...
-pkgrel=1
+pkgrel=2
 pkgdesc="SLiM forward simulation software package for population genetics (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64')
 url="https://messerlab.org/slim/"
-license=("GPL")
-makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+license=("spdx:GPL-3.0-or-later")
+makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
```
Biswa96, line 11: Generally, autotools is not required with cmake. Are you sure about the requirement?

mmuetzel: Yeah. The way how gnulib is built as part of the project requires that autotools are also installed.

PR #10960, opened 2022-03-11
unrar: Add package

Biswa96:
Please review @ZachBacon

### DIFF
--- a/mingw-w64-unrar/0001-fix-makefile.patch
+++ b/mingw-w64-unrar/0001-fix-makefile.patch
@@ -1,0 +1,86 @@
+diff --git a/makefile b/makefile
+index ace156c..ac42fb7 100644
+--- a/makefile
++++ b/makefile
+@@ -11,6 +11,16 @@ AR=ar
+ LDFLAGS=-pthread
+ DESTDIR=/usr
+ 
++ifeq ($(OS), Windows_NT)
++	EXEEXT=.exe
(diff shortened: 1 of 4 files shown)


Biswa96:
I have sent the patches to WinRAR developers and they agree to add some of the fixes. Here is the reply from email.

<details><summary>Click here to see the text</summary>

```
Hello,

Thank you.

I'll replace spaces with tab character in makefile.

Also I'll replace 0 with NULL in 4th ConnectServer parameter.
Even though Microsoft uses 0 here in their example:
https://docs.microsoft.com/en-us/windows/win32/wmisdk/example--getting-wmi-data-from-the-local-computer
you are right that it should be NULL.

If I just remove all #pragma comment, there will be a link failure.
I would need to add these libs to makefile and UnRAR.vcxproj,
which isn't always convenient.

Regarding WINVER and _WIN32_WINNT, if they are already defined to
values different than expected by unrar, it might be better to
redefine them and see a warning than use a value which wasn't tested
with unrar source code.

Regarding makefile, the amount and complexity of changes seems to be
a little more than I expected, so I am not sure about it yet.

Probably I'll include a part of this patch.
Typically developers use MSVC to build unrar in Windows, especially
taking into account that Microsoft provides its free Community edition.

Thank you again.

Eugene
```

</details>

Biswa96:
Thanks for review.

PR #10965, opened 2022-03-11
ci-check: Re-install pip if it was removed.

mmuetzel:
When the `check` job is run with multiple packages, `pip` might be uninstalled when one step is cleaned up with `pacman --remove -cns XXX`. See e.g., https://github.com/msys2/MINGW-packages/pull/9931#issuecomment-1065110743.

IIUC, this change re-installs `pip` when the next package is checked.
I'm not sure if this is the correct place for that. Comments much appreciated.

### DIFF
--- a/.ci/ci-check.py
+++ b/.ci/ci-check.py
@@ -83,6 +83,8 @@ def install_package(pkg: typing.Union[typing.List[str], Path]) -> typing.Generat
             "--refresh",
             "--sync",
             "--noconfirm",
+            "--needed",
+            "mingw-w64-x86_64-python-pip",
             get_pkg_name(pkg),
         ]
         uninstall_command = [
@@ -100,6 +102,8 @@ def install_package(pkg: typing.Union[typing.List[str], Path]) -> typing.Generat
                 "--sync",
                 "--refresh",
                 "--noconfirm",
+                "--needed",
+                "mingw-w64-x86_64-python-pip",
                 *pkg,
             ]
             uninstall_command = [


.ci/ci-check.py
```diff
@@ -100,6 +101,7 @@ def install_package(pkg: typing.Union[typing.List[str], Path]) -> typing.Generat
                 "--sync",
                 "--refresh",
                 "--noconfirm",
+                "mingw-w64-x86_64-python-pip",
```
MehdiChinoune, line 104:
```suggestion
                "--needed",
                "mingw-w64-x86_64-python-pip",
```
Skip reinstalling

.ci/ci-check.py
```diff
@@ -83,6 +83,7 @@ def install_package(pkg: typing.Union[typing.List[str], Path]) -> typing.Generat
             "--refresh",
             "--sync",
             "--noconfirm",
+            "mingw-w64-x86_64-python-pip",
```
MehdiChinoune, line 86:
```suggestion
            "--needed",
            "mingw-w64-x86_64-python-pip",
```

MehdiChinoune:
I think  a better solution is to just install `${MINGW_PACKAGE_PREFIX}-python-pip` before running `./.ci/ci-check.py`, so it won't be removed because It was explicitly installed.

jeremyd2019:
> I think a better solution is to just install `${MINGW_PACKAGE_PREFIX}-python-pip` before running `./.ci/ci-check.py`, so it won't be removed because It was explicitly installed.

Unless one of the packages being removed is a dependency of pip.

.ci/ci-check.py
naveen521kk, line 86: How about adding `--needed` so that `pip` is only installed when it's actually not installed?
```suggestion
            "--noconfirm",
            "--needed",
            "mingw-w64-x86_64-python-pip",
```

.ci/ci-check.py
naveen521kk, line 104: Why should we uninstall `mingw-w64-x86_64-python-pip` here?

.ci/ci-check.py
mmuetzel: Done

.ci/ci-check.py
mmuetzel: IIUC, `install_package` can be called with the first argument being either a `typing.List[str]` or a `Path`. I was thinking that is might be safest to make sure `python-pip` re-installs in both cases. That might be over-cautious though.
Should it be removed from one of the two branches?

naveen521kk: Oh wait, I just noticed the if condition. I thought you were uninstalling `pip` along with other packages. Sorry, makes sense now.

naveen521kk (approved):
lgtm

PR #10967, opened 2022-03-12
Update Packages

naveen521kk:


MehdiChinoune:
numba doesn't support (work with) numpy>1.22.

naveen521kk:
I'll drop that commit then, thanks.

mingw-w64-python-ipython/PKGBUILD
```diff
@@ -24,6 +24,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-python-backcall"
          "${MINGW_PACKAGE_PREFIX}-python-prompt_toolkit"
          "${MINGW_PACKAGE_PREFIX}-python-pygments"
          "${MINGW_PACKAGE_PREFIX}-python-setuptools"
+         "${MINGW_PACKAGE_PREFIX}-python-stack-data"
```
MehdiChinoune, line 27: There is no such package!

naveen521kk: I noticed that I should create one.

PR #10972, opened 2022-03-13
Add Biber package dependencies (Part 1)

naveen521kk:
a wip first planning to add all the Perl packages it needs https://github.com/archlinux/svntogit-community/blob/packages/biber/trunk/PKGBUILD#L12-L40 and then the package itself. Most of these PKGBUILDs here are autogenerated using https://github.com/naveen521kk/msys2-perl-package-creator 

(I give up on this, I'll try it later maybe)


naveen521kk:
Done till [`perl-datetime-calendar-julian`](https://github.com/archlinux/svntogit-community/blob/packages/biber/trunk/PKGBUILD#L20), I think it's simply getting bigger and bigger for doing it all at once, I think let this be the part one.

PR #10973, opened 2022-03-13
scipy: update to 1.8.0

MehdiChinoune:
For some reason it accepts `config_fc ...` when building but rejects it when installing?


MehdiChinoune:
fixed.

PR #10981, opened 2022-03-14
python-scikit-learn: update to 1.0.2

MehdiChinoune:


Biswa96:
Just asking. Should we remove clang* msystems from packages which depend on python-scipy? Many packages are in pending page due to python-scipy. Also, can clang32 be added in python-threadpoolctl?

PR #10983, opened 2022-03-14
[CI] Install split packages separately.

MehdiChinoune:
Good:
- de-duplicate code
- No more need for `.ci-sequential` trick for conflicting split packages.
- Check their runtime dependencies independently.

Bad:
- Take more time because of repeated intall/uninstall


lazka:
nice, thanks :)

lazka:
hm, good question. maybe we need to add them all to the temp repo first?

MehdiChinoune:
the diff is now not correct because it compares its metadata/filelist with the ones from database, which now is the same because it was added to the database before installing it!

@lazka I think you should revert my changes here and the one trying to fix it.

PR #10984, opened 2022-03-14
enable some more clangarm64 packages

kmilos:

PR #10985, opened 2022-03-14
[CI] Fix check runtime dependencies failure

MehdiChinoune:


MehdiChinoune:
@lazka
Fixes #10984 failure.

PR #10975, opened 2022-03-13
imagemagick: Update to version 7.1.0-27

mmuetzel:
Some of the changes in the existing patches seem to have been applied upstream in the meantime. I've removed those parts from "[002-7.0.4.1-build-fixes.patch](https://github.com/msys2/MINGW-packages/compare/master...mmuetzel:imagemagick?expand=1#diff-1126c9848550d528f344c1daa44bb1333024604b83c0557554b7e6ec9244ad56)".

PR #10992, opened 2022-03-14
zbar: Update to version 0.23.90

mmuetzel:
Patches are no longer needed. These issues have been addressed upstream.

Should be unblocking #10975.


MehdiChinoune:
I propose to amend it to #10975, so zbar will be built against the new imagemagick.

PR #10996, opened 2022-03-15
pidgin: update to 2.14.8

jeremyd2019:
I haven't tested this or even tried running it, just got it to build locally.

### DIFF
--- a/mingw-w64-pidgin/001-autotools-and-fhs.patch
+++ b/mingw-w64-pidgin/001-autotools-and-fhs.patch
@@ -1,7 +1,6 @@
-diff -aurN 000/configure.ac 001/configure.ac
---- 000/configure.ac
-+++ 001/configure.ac
-@@ -120,10 +120,44 @@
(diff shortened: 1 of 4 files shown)


jeremyd2019:
I just noticed that it no longer recognizes `--with-python` and configure now looks for `--with-python3` or tries to find a `python3` on the path... so maybe it doesn't need python2 anymore?

lazka:
lgtm (also tested with IRC at least). I doubt this has any users, so I wouldn't worry too much.

ZachBacon:
> lgtm (also tested with IRC at least). I doubt this has any users, so I wouldn't worry too much.

Pretty much what I tested to be honest. Later I'll test some pidgin/libpurple plugins that are listed in archlinux's repo and recompile them as an additional test.

jeremyd2019:
> lgtm (also tested with IRC at least). I doubt this has any users, so I wouldn't worry too much.

OK, that's all I was going to do, but have been busy the last couple of days.

PR #11000, opened 2022-03-16
qemu: Update to 7.0.0

heljkon:
Another qemu update cycle starting with RC0.
To enable compilation in clang, a small patch is necessary to get rid of
```
../qemu-7.0.0-rc0/qga/vss-win32/install.cpp:49:24: error: cannot initialize a variable of type 'char *' with an rvalue of type 'const char *'
    char *msg = NULL, *nul = strchr(text, '(');
                       ^     ~~~~~~~~~~~~~~~~~
```
Will provide the patch to upstream ASAP

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -1,12 +1,17 @@
 _realname=qemu
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-_base_ver="6.2.0"
+pkgname=(
+  "${MINGW_PACKAGE_PREFIX}-qemu"
+  "${MINGW_PACKAGE_PREFIX}-qemu-common"
+  "${MINGW_PACKAGE_PREFIX}-qemu-guest-agent"
+  "${MINGW_PACKAGE_PREFIX}-qemu-image-util"
+)
+_base_ver="7.0.0"
 # QEMU Versioning of RC-SourcePackage and RC-Version differs
 # e.g. qemu-6.1.0-rc0.tar.xz contains 6.0.90, qemu-6.1.0-rc1.tar.xz contains 6.0.91
 # Unset _rc_no to create Release-Build
 #_rc_no="4"
-_rc_base_ver="6.1.9${_rc_no}"
+_rc_base_ver="6.2.9${_rc_no}"
 _rc_file_ver="rc${_rc_no}"
 _tarname=$( [ -z "${_rc_no}" ] && echo ${_realname}-${_base_ver} || echo ${_realname}-${_base_ver}-${_rc_file_ver} )
 pkgver=$( [ -z "${_rc_no}" ] && echo ${_base_ver} || echo ${_rc_base_ver} )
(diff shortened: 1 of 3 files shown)


heljkon:
Many fixes in msys2.examples.tests.sh were done based on review by @1480c1 in https://github.com/msys2/MINGW-packages/pull/10098

mingw-w64-qemu/msys2.examples.tests.sh
```diff
@@ -323,13 +323,13 @@ function perform {
...
 function extractReadme {
-	local FILE=$1
+	local FILE="$1"
 	[ -f "$FILE" ] || return 0
 	local TXT=0
 	local LINE
 	local EOF="EOF"
 	# Print the first HERE-Document to screen
-	cat $FILE | while read LINE
+	cat "$FILE" | while read LINE
```
1480c1, line 332: This can probably be changed to use `< "$FILE"` instead of cat |, please search while read file indirect for the exact syntax if needed

jeremyd2019: As I recall there are odd side-effects of piping into a while like that, beyond just the general optimization of avoiding forking a cat process when it's not needed.

heljkon: Thanks for pointing to this bad practice. I'll correct in the next commit.

BTW: I already experienced these side-effects in other scripts: Variables tend to forget their updated values after while-loop. 
I guess many of my scripts still use this easy-to-write pattern and wait to be fixed...

heljkon:
QEMU Guest Agent has some problems which need to be fixed upstream. I solved these temporarily.

Furthermore I added tests and documentation for QEMU Guest Agent focusing Windows.

Finally I came to the conclusion to create a dedicated qemu-guest-agent package for VM Guests. qemu package still consists of everything inclusive qemu-ga and needs nearly 1GB w/o deps. qemu-ga standalone needs 1MB w/o deps.

heljkon:
New binary package qemu-image-util was created for the same reasons as qemu-guest-agent. It is standalone very small and useful.

mingw-w64-qemu/PKGBUILD
```diff
@@ -108,7 +149,17 @@ check() {
...
+package_qemu() {
+  optdepends=(
+    "${MINGW_PACKAGE_PREFIX}-gtk-vnc: Contains gvncviewer. For connections to Qemu VNC server"
+    "${MINGW_PACKAGE_PREFIX}-spice-gtk: Contains spicy. For connections to Qemu SPICE server"
+    "${MINGW_PACKAGE_PREFIX}-virt-viewer: Contains remote-viewer. For connections to Qemu SPICE or VNC server"
+  )
+  provides=(
+    "${MINGW_PACKAGE_PREFIX}-qemu-guest-agent"
+    "${MINGW_PACKAGE_PREFIX}-qemu-image-util"
+  )
```
jeremyd2019, line 161: IMHO, it would be more 'normal' to not include the same files in multiple packages, and instead depend here (though I don't think qemu necessarily depends on qemu-guest-agent, one is for the host and the other is for the guest, right).

Also, there is a general reluctance to split packages in MINGW-packages.  I don't know where the bar for that is, and if this meets it...

heljkon: You are right. The reason for the copying package is to calm this reluctance. This way the package stays whole, but if you don't want 1GB extra installed for a small tool, you can use it this way.

BTW: Even ArchLinux splits the qemu package.
Generally I would like to prepare smaller packages. I don't really need to decide which way to go, but I would prefer to go in this direction.

@lazka @Biswa96 
Which could be an acceptable split strategy for you?

Anybody else I should poke for opionion?

lazka: (I don't know qemu enough to see what makes sense, but I'm fine with splitting in general if it helps you in any way)

heljkon: As suggested by @jeremyd2019 and after OK from @lazka the split was refactored to use only depends relations.

heljkon:
@jeremyd2019 @lazka @1480c1 @Biswa96 

qemu 7.0.0 is ready to merge

Main features:
- package is split into qemu, qemu-guest-agent, qemu-image-util and qemu-common using only depends relations
- qemu-guest-agent is functional and tested now (with exception of qga-vss, which requires Windows Server tools for test)

PR #11001, opened 2022-03-16
cegui: try to build with Python 3

lazka:
Untested, let's see what happens.


lazka:
oh, this package hasn't been built since 2018...

lazka:
no release in years, maybe it should be dropped

lazka:
Dropped in 0cbc5ff1254d7fd0f8eb67691b87a9e

PR #11014, opened 2022-03-17
gnutls: Update to 3.7.4

lazka:


lazka:
DLL name change, needs checking

jeremyd2019:
Don't see any package using that DLL.

PR #11027, opened 2022-03-18
zbar: Enable pthread.

mmuetzel:
This better matches how upstream is configuring in their CI:
https://github.com/mchehab/zbar/blob/master/.github/workflows/ci.yml#L228


Biswa96:
Does it use native Win32 thread APIs without that option or with another option?

mmuetzel:
@Biswa96: Good point!

The help string for the configure flag `--disable-pthread` says: "omit support for threaded applications".
It is disabled by default on Windows. I thought that meant that the application was single threaded with the current configuration.

But looking at their code, `pthread` seems to be never used on Windows no matter the configure flags. They *always* use native Windows threads afaict. So IIUC, this PR wouldn't change anything.

Closing PR.

PR #11031, opened 2022-03-18
SDL2 clangarm64

driver1998:
Enables SDL2, SDL2_gfx, SDL2_image, SDL2_mixer, SDL2_net, SDL2_ttf on clangarm64.


mingw-w64-SDL2_gfx/PKGBUILD
```diff
@@ -25,11 +25,18 @@ build() {
   export lt_cv_deplibs_check_method='pass_all'
   [[ -d "${srcdir}/build-${MINGW_CHOST}" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}"
   mkdir -p "${srcdir}/build-${MINGW_CHOST}"
+
+  local extra_config=""
+  if [[ ${MINGW_CHOST} == "aarch64-w64-mingw32" ]]; then
+    extra_config+="--disable-mmx"
+  fi
```
jeremyd2019, line 32: Not a big deal, but I would do something like
```suggestion
  case "${CARCH}" in
    i686|x86_64)
      ;;
    *)
      extra_config+="--disable-mmx"
      ;;
  esac
```

mingw-w64-nasm/PKGBUILD
```diff
@@ -7,7 +7,7 @@ pkgver=2.15.05
 pkgrel=1
 pkgdesc="An 80x86 assembler designed for portability and modularity (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
```
jeremyd2019, line 10: I've been kind of holding nasm and similar tools that only target x86 out intentionally, as a sign that a package might need more effort to port to arm64.  This may not be a good argument, because this is the second time it has been enabled in a PR.

It looks like in this PR the package is flac.  Since the makedepends are outside the area that GitHub allows me to make a 'suggestion', we did something like
```diff
-             "${MINGW_PACKAGE_PREFIX}-nasm"
+              $( [[ "${CARCH}" != "i686" \
+                   && "${CARCH}" != "x86_64" ]] \
+                || echo "${MINGW_PACKAGE_PREFIX}-nasm" )
```
in libjpeg-turbo

driver1998: Cool, I always wonder how to conditionally add/remove dependencies.

mingw-w64-SDL2_gfx/PKGBUILD
driver1998: OK, will do.

mingw-w64-nasm/PKGBUILD
driver1998: Updated based on your suggestion.

mingw-w64-SDL2_gfx/PKGBUILD
```diff
@@ -7,7 +7,7 @@ pkgver=1.0.4
 pkgrel=1
```
jeremyd2019, line 7:
```suggestion
pkgrel=2
```

mingw-w64-flac/PKGBUILD
```diff
@@ -7,7 +7,7 @@ pkgver=1.3.4
 pkgrel=1
```
jeremyd2019, line 7:
```suggestion
pkgrel=2
```

jeremyd2019:
ARM64 CI run passed:
https://github.com/msys2-arm/MINGW-packages/runs/5607832167?check_suite_focus=true

mingw-w64-nasm/PKGBUILD
Biswa96: Is this condition really required? The cmake file already checks for nasm program and disables assembly if nasm is absent.

jeremyd2019: This condition doesn't include nasm in makedepends on non-x86

Biswa96: Including nasm would not fail build in aarch64 (non-x86), would it? 🤔

PR #11037, opened 2022-03-19
cairo: Update to 1.17.6

lazka:
0001-fix-font-names-in-pdf-output.patch and 0031-remove-glyph-bounds-check.patch
are included in the new release

0027-win32-print-fix-unbounded-surface-assertion.patch was fixed some other way, see
https://gitlab.freedesktop.org/cairo/cairo/-/issues/131#note_1304830

xml enabled to match the previous build, not sure if we should disable it.


lazka:
This leads to errors in the pycairo test suite. Needs investigating

lazka:
disabled the new font backend for now and filed bugs upstream

PR #11042, opened 2022-03-20
doxygen: enable for clangarm64

jeremyd2019:
Disable wizard, to avoid needing qt5 just yet.

### DIFF
--- a/mingw-w64-doxygen/PKGBUILD
+++ b/mingw-w64-doxygen/PKGBUILD
@@ -4,10 +4,10 @@ _realname=doxygen
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.9.3
-pkgrel=1
+pkgrel=2
 pkgdesc="A documentation system for C++, C, Java, IDL and PHP (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
For Qt5/Qt6 there are few remaining packages:
- assimp
- double-conversion
- dbus
- md4c
- pcre2
- xmlstarlet
- libmariadbclient

jeremyd2019:
my script comes up with mingw-w64-postgresql too (makedepends).

MehdiChinoune:
> my script comes up with mingw-w64-postgresql too (makedepends).

It could be disabled. like we did before postgresql was available for clang*.

driver1998:
> For Qt5/Qt6 there are few remaining packages:
>
> assimp
> minizip
> double-conversion
> dbus
> md4c
> pcre2
> xmlstarlet
> libmariadbclient

The following compiles for me locally:
- minizip-git
- double-conversion
- dbus (tests hang the build process, have to skip)
- md4c
- libmariadbclient

assimp seems to require D3DX9 which is not available for ARM, let's see what can do about it.

MehdiChinoune:
> assimp seems to require D3DX9 which is not available for ARM, let's see what can do about it.

It requires it only for assimp_view tool which it could be disabled on aarch64.
https://github.com/assimp/assimp/blob/v5.2.2/tools/assimp_view/CMakeLists.txt#L42

Biswa96:
It is possible to move the required import libraries to arm crt, like this https://github.com/mingw-w64/mingw-w64/commit/456b8d78be1e418d28d5c8ad26debecf92050a81

driver1998:
It's not the import libraries, those D3DX dlls does not exist on ARM. For ARM32, there is some in the Windows 8 HLK for some reason, but nothing on ARM64.

PR #11048, opened 2022-03-20
OpenCV: update to 4.5.5

MehdiChinoune:


MehdiChinoune:
package-grokker shows only these packages need to be rebuilt:
```
mingw-w64-gst-plugins-bad
mingw-w64-gmic
mingw-w64-openimageio
mingw-w64-itk
```
https://github.com/MehdiChinoune/package-grokker/runs/5633776681?check_suite_focus=true

MehdiChinoune:
I am waiting for #11058 to enable vtk dependency on clang environments.

lazka:
> package-grokker shows only these packages need to be rebuilt:

btw, package-grokker now has a mode where you just have to pass a PR number: https://github.com/lazka/package-grokker/runs/5667298516?check_suite_focus=true

MehdiChinoune:
`gmic` failed to build, but the failure is not related to OpenCV.

MehdiChinoune:
It should be ready now.

PR #11050, opened 2022-03-20
OpenBLAS: Build C_LAPACK in CLANG* environments.

mmuetzel:
https://github.com/xianyi/OpenBLAS/pull/3539 used f2c on a current version of reference LAPACK with manual post-processing to add a version of netlib LAPACK that can be compiled without a Fortran compiler.
This PR adds those changes for the CLANG* environments.

The PR is pretty large and it is still marked as WIP. But it seems to be compiling without major problems in the CLANG64 and CLANG32 environments.

I tested in CLANG64 by compiling igraph linking to this patched version of OpenBLAS (instead of f2cblaslapack). It compiles correctly and passes all tests.

Looking forward, we could transition the packages that currently link to the BLAS/LAPACK libraries from f2cblaslapack to this modified version of OpenBLAS.

As an even longer term goal, would it be possible to establish some kind of mechanism similar to Ubuntu's "alternatives" that would allow switching between different BLAS/LAPACK implementations on the user side?

### DIFF
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -6,12 +6,12 @@ pkgbase=mingw-w64-openblas
 pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
          $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas64"))
 pkgver=0.3.20
-pkgrel=2
+pkgrel=3
 pkgdesc="An optimized BLAS library based on GotoBLAS2 1.13 BSD, providing optimized blas, lapack, and cblas (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


Biswa96:
Instead of copying the whole patch file, it can be grabbed from the PR's link `https://github.com/xianyi/OpenBLAS/pull/3539.patch`. Also it may be better to wait for the PR to be merged.

mmuetzel:
> Instead of copying the whole patch file, it can be grabbed from the PR's link `https://github.com/xianyi/OpenBLAS/pull/3539.patch`.

It looks like some files have CRLF line endings. (At least, the patch has them for some files.)
For some reason, all(?) files have LF line endings when the tarball is unpacked. So, the original patch doesn't apply.

> Also it may be better to wait for the PR to be merged.

I agree. Given the size of this patch, it's probably a good idea to wait at least until it has landed.

mmuetzel:
The change has landed upstream. Pulling it from there.

PR #11051, opened 2022-03-20
Enable more clangarm64 packages

driver1998:
This enables quite a few packages on clangarm64, notably:

- aria2c
- putty
- ghostscript
- gtk-vnc
- dbus
- rtmidi
- other deps used by qt5-pkgs


mingw-w64-postgresql/postgresql-14.0-use-mingw-setjmp-on-clang-arm64.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/src/include/c.h b/src/include/c.h
+index c8ede08..e445524 100644
+--- a/src/include/c.h
++++ b/src/include/c.h
+@@ -1340,17 +1340,18 @@ extern unsigned long long strtoull(const char *str, char **endptr, int base);
+  * so on that toolchain we cheat and use gcc's builtins.
+  */
+ #ifdef WIN32
+-#ifdef __MINGW64__
++#if defined(__MINGW64__) && (defined(__i386__) || defined(__x86_64__))
```
jeremyd2019, line 10: oh dear.  I guess this code doesn't get exercised much.  We've had issues with code trying to use `__builtin_setjmp` on x86 as well in other projects.  Idea (probably needs to be tested):
```suggestion
+#if defined(__MINGW64__) && !defined(_UCRT)
```

jeremyd2019: That's not really germaine to this PR though.

jeremyd2019:
Successful ARM64 CI run:
https://github.com/msys2-arm/MINGW-packages/runs/5619999547?check_suite_focus=true

But, I saw something using `gcc` fly by, so I kind of expect something still needs patching for proper building with clang (my ARM64 runners pre-install -toolchain for performance, but 'normal' runners do not, so gcc-compat will not be present).  So I am also waiting for the other jobs in that run to finish to be sure (since the runs here failed due to network issues).

jeremyd2019:
```
 /C/_/mingw-w64-libmariadbclient/PKGBUILD: line 85: /clang32/bin/gendef: No such file or directory
      Aborting...
```

that would be in `${MINGW_PACKAGE_PREFIX}-tools-git`.

Issue #11005, opened 2022-03-17
opencv: rebuild fails on clang64


jeremyd2019:
due to a problem of float casting in eigen3

```
In file included from E:/projets/MINGW-packages/mingw-w64-opencv/src/build-x86_64/modules/core/matmul.avx512_skx.cpp:2:
In file included from E:/projets/MINGW-packages/mingw-w64-opencv/src/opencv-4.5.5/modules/core/src/precomp.hpp:59:
In file included from E:/projets/MINGW-packages/mingw-w64-opencv/src/opencv-4.5.5/modules/core/include/opencv2/core/private.hpp:70:
In file included from C:/msys64/clang64/include/eigen3/Eigen/Core:188:
C:/msys64/clang64/include/eigen3/Eigen/src/Core/arch/AVX512/PacketMath.h:1044:10: error: cannot initialize return object of type 'float' with an rvalue of type 'typename unpacket_traits<__attribute__((__vector_size__(4 * sizeof(float)))) float>::type' (aka '__attribute__((__vector_size__(4 * sizeof(float)))) float')
  return pfirst(pmul(res, _mm_permute_ps(res, _MM_SHUFFLE(0, 0, 0, 1))));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/clang64/include/eigen3/Eigen/src/Core/arch/AVX512/PacketMath.h:1064:10: error: cannot initialize return object of type 'float' with an rvalue of type 'typename unpacket_traits<__attribute__((__vector_size__(4 * sizeof(float)))) float>::type' (aka '__attribute__((__vector_size__(4 * sizeof(float)))) float')
  return pfirst(_mm_min_ps(res, _mm_permute_ps(res, _MM_SHUFFLE(0, 0, 0, 1))));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/clang64/include/eigen3/Eigen/src/Core/arch/AVX512/PacketMath.h:1083:10: error: cannot initialize return object of type 'float' with an rvalue of type 'typename unpacket_traits<__attribute__((__vector_size__(4 * sizeof(float)))) float>::type' (aka '__attribute__((__vector_size__(4 * sizeof(float)))) float')
[...]

PR #11052, opened 2022-03-20
Update OpenCV to 4.5.5 release

ugur-kurnaz:
updated to OpenCV 4.5.5, it builds and works fine on mingw64/ucrt64 (not staging). clang64 doesn't seem to work on my computer utf-8 characters are strangely added to file names and clang++ build fail.

this pull request is cleaner, some deps have been removed because I think they were too havy (even with some cyclique deps on opencv) and not really of any use (if someone needs them, they simply can rebuild the package to fit their needs).

python biding fails if any of depencies are not installed (even optional), thus I simply removed deps that are not relevant, and added python-numpy as dep, so that one can do opencv mock more easily.

there is an opened issue #11005,  for building on clang64.

I'm not sure of license concern, please someone with  better knowledge to have a look.


ugur-kurnaz:
already another pull request #11048 on going.

PR #11054, opened 2022-03-20
wxPython (and deps): enable for clangarm64

jeremyd2019:
also fix some of the packages to build with clang (rather than a gcc alias for clang)


mati865 (approved):

jeremyd2019:
Successful ARM64 CI run:
https://github.com/msys2-arm/MINGW-packages/runs/5619775594?check_suite_focus=true

PR #11055, opened 2022-03-21

kmilos:
Hm, I guess I could upgrade dav1d to just released 1.0.0 while at it, but probably need to bump dependencies due to soname version change...


kmilos:
Hm, I guess I could upgrade dav1d to just released 1.0.0 while at it, but probably need to bump dependencies due to soname version change...

kmilos:
Now VLC needs https://code.videolan.org/videolan/vlc/-/commit/dbf45cea2a8abdfbef897b8a71f3eb782bb1b712

mingw-w64-libraw/PKGBUILD
```diff
@@ -8,14 +8,16 @@ provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
...
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.libraw.org/"
 license=("LGPL")
-makedepends=("${MINGW_PACKAGE_PREFIX}-autotools" "${MINGW_PACKAGE_PREFIX}-cc")
+makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
+             "${MINGW_PACKAGE_PREFIX}-cc")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
+         $([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-openmp")
```
jeremyd2019, line 20: Last I checked, Clang's openmp was not supported on Windows ARM64 yet.  There is no mingw-w64-clang-aarch64-openmp as a result

driver1998: But it looks like this will disable openmp for clang32 and clang64 as well.

kmilos: Thanks, will revert this change for now. And it doesn't link on clang32/64 anyway w/ autotools (need to specify `-fopenmp=libomp` somewhere)...

Biswa96:
Is it possible to split this change in two commits - one for the clangarm64 addition and another one for dav1d update?

mingw-w64-libraw/PKGBUILD
jeremyd2019: yes.  libtool thinks it knows better than the compiler, passes -nostdlib to the link, and tries to specify all of the necessary libs.  Unfortunately, it doesn't know about libomp.

driver1998:
Looks like not everything in this PR is enabled for clangarm64, ffmpeg doesn't.

Is this expected?

jeremyd2019:
I think some were just rebuilds for the dav1d update.  Hence @Biswa96's request to split this into two PRs.

kmilos:
Yeah, I think it's reasonable to split, the rabbit hole is getting deeper...

mingw-w64-libraw/PKGBUILD
kmilos: And I guess the [`ax_openmp`](http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_openmp.m4) macro is somewhat dated and doesn't include it in the candidate list like CMake does... It looks like defining `OPENMP_CFLAGS=-fopenmp=libomp` should do it, but I'll leave it for the impending LibRaw upgrade...

kmilos:
Should I also add dav1d 0.9.2 for clangarm64 now, or we wait for bulk update to 1.0.0? ([need to wait until they figure out all the VLC changes...](https://code.videolan.org/videolan/vlc/-/merge_requests/1618))

driver1998:
If it can be built without much fixing then sure.

mingw-w64-dav1d/PKGBUILD
```diff
@@ -4,46 +4,50 @@ _realname=dav1d
...
-license=('BSD-2-Clause')
+license=('spdx:BSD-2-Clause')
 makedepends=("${MINGW_PACKAGE_PREFIX}-pkg-config"
-             "${MINGW_PACKAGE_PREFIX}-nasm"
+             $( [[ "${CARCH}" != "i686" \
+                   && "${CARCH}" != "x86_64" ]] \
+                || echo "${MINGW_PACKAGE_PREFIX}-nasm" )
              "${MINGW_PACKAGE_PREFIX}-cc"
-             "${MINGW_PACKAGE_PREFIX}-meson")
-depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
```
jeremyd2019, line 17: What's the current best practice, should packages depend on gcc-libs?

Biswa96:
Interesting! aom can be cross compiled for aarch64 with llvm-mingw. Wondering if `s/_MSC_VER/_WIN32` with arm_cpudetect.c can solve it.

jeremyd2019:
vlc seems to set `-DCONFIG_RUNTIME_CPU_DETECT=0` for arm or aarch64 on Windows

kmilos:
Yes, so is the aom toolchain definition:

https://aomedia.googlesource.com/aom/+/refs/tags/v3.3.0/build/cmake/toolchains/arm64-mingw-gcc.cmake

mingw-w64-dav1d/PKGBUILD
kmilos: AFAIK we only list what we actually link against?

jeremyd2019: gcc-libs contains libgcc and libstdc++ on gcc, and libc++ and libunwind on clang.  These are almost always implicitly linked into binaries.

jeremyd2019: But, I see the new "runtime dependencies" CI output shows these are not linked to dlls from the package, on either clang or gcc.  Must be statically linked to the runtime.  Nevermind

kmilos: Not for C code. There is a runtime linking check as part of CI now, dav1d binaries do not to link to anything apart from the CRT.

kmilos: E.g. for aom there is clear dynamic linkong to those, while there is nothing for dav1d (no C++ exception handling etc I guess):

```
  [MSYS2 CI] mingw-w64-aom: Runtime dependencies for mingw-w64-clang-x86_64-aom
  
  /clang64/bin/aomdec.exe:
  	libaom.dll => D:\M\msys64\clang64\bin\libaom.dll (0x000001ec07660000)
  	 libwinpthread-1.dll => D:\M\msys64\clang64\bin\libwinpthread-1.dll (0x000001ec073a0000)
  	libc++.dll => D:\M\msys64\clang64\bin\libc++.dll (0x000001ec07660000)
  	libunwind.dll => D:\M\msys64\clang64\bin\libunwind.dll (0x000001ec073a0000)
  /clang64/bin/aomenc.exe:
  	libaom.dll => D:\M\msys64\clang64\bin\libaom.dll (0x0000020cd0090000)
  	 libwinpthread-1.dll => D:\M\msys64\clang64\bin\libwinpthread-1.dll (0x0000020cd0b30000)
  	libc++.dll => D:\M\msys64\clang64\bin\libc++.dll (0x0000020cd0090000)
  	libunwind.dll => D:\M\msys64\clang64\bin\libunwind.dll (0x0000020ccfea0000)
  /clang64/bin/libaom.dll:
  	libwinpthread-1.dll => D:\M\msys64\clang64\bin\libwinpthread-1.dll (0x000002272f9c0000)
```

jeremyd2019:
Successful ARM64 CI run:
https://github.com/msys2-arm/MINGW-packages/runs/5666051637?check_suite_focus=true

PR #11056, opened 2022-03-21
New package: rclone

okhlybov:
This creates base rclone executable package as well as its embeddable build in form of dynamic library.


lazka:
thanks

PR #11062, opened 2022-03-22
go: update to 1.18

jeremyd2019:


jeremyd2019:
Apparently the issue that the PKGBUILD was trying to work around is gone now, seems to be building fine in a much more direct method.

PR #11064, opened 2022-03-22
Qt6 clangarm64, part 1

driver1998:


jeremyd2019:
ARM64 CI run passed:
https://github.com/msys2-arm/MINGW-packages/runs/5650363052?check_suite_focus=true

PR #11067, opened 2022-03-22
pango: split docs to a separate package

Biswa96:
Please review.


Biswa96:
I would like to follow these steps in future with other _heavy documented_ packages.

PR #11073, opened 2022-03-23
openvdb: update to 9.0.0

MehdiChinoune:


lazka:
Do we need to rebuild openimageio?

MehdiChinoune:
> Do we need to rebuild openimageio? Does it even link to it... hmm.

Why rebuilding?

lazka:
openimageio depends on openvdb and this is a major version update

lazka:
But it looks like it doesn't actually link against it, so something is wrong in openimageio maybe.

lgtm then

MehdiChinoune:
openimageio will be rebuilt with OpenCV.

PR #11074, opened 2022-03-23
Enable qt5-pkgs, and some other packages on clangarm64

driver1998:


jeremyd2019:
Successful ARM64 CI run:
https://github.com/msys2-arm/MINGW-packages/runs/5666748021?check_suite_focus=true

PR #11069, opened 2022-03-22
enable go packages for clang32 and clangarm64

jeremyd2019:
Except for dnslookup, which fails with
```
C:\Users\runneradmin\go\pkg\mod\github.com\lucas-clemente\quic-go@v0.24.0\internal\qtls\go118.go:6:13: cannot use "quic-go doesn't build on Go 1.18 yet." (untyped string constant) as int value in variable declaration
```

and butler, which never successfully built on CLANG64.

Also, it appears that go packages now makedepend on `git`.  There was a reference to an option in the error that may be able to disable that, but it wasn't clear where that option would go (GOFLAGS? arg to `go build` on package that builds that way?)
```
  go: missing Git command. See https://golang.org/s/gogetcmd
  error obtaining VCS status: exec: "git": executable file not found in %PATH%
  	Use -buildvcs=false to disable VCS stamping.
```

PR #11077, opened 2022-03-23
rclone: depend on cc not gcc, makedepend on git

jeremyd2019:
enable for clang32

unfortunately clangarm64 failed with
```
# github.com/rclone/rclone/librclone
.debug_frame+1MB: addpesym: could not map .debug_frame+1MB symbol with non .text or .rdata or .data section
.debug_frame+2MB: addpesym: could not map .debug_frame+2MB symbol with non .text or .rdata or .data section
```
so not enabling that.  (same error as https://github.com/msys2/MINGW-packages/pull/11069#issuecomment-1075774548)


lazka:
oops, thanks

okhlybov:
Huh, my bad :-\
Should've investigated the cause of clang32 build failure.

jeremyd2019:
Go package was updated to 1.18 and built for clang32 only about the same time as your PR, so I think it was a 'mid-air collision' situation

PR #11078, opened 2022-03-23
libvips: update to 8.12.2

jcupitt:
Plus:

- update project URL
- add support for HEIC
- add support for AVIF
- add support for jpeg2000
- don't build the old vips7 API (saves a few kb)
- remove giflib (libvips now has its own GIF reader)

### DIFF
--- a/mingw-w64-libvips/PKGBUILD
+++ b/mingw-w64-libvips/PKGBUILD
@@ -3,28 +3,31 @@
 _realname=libvips
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=8.12.1
-pkgrel=2
+pkgver=8.12.2
+pkgrel=1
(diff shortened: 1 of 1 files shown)


tknerr:
Thanks @jcupitt for updating the libvips version to 8.12.2. This should hopefully fix the issue we had in https://github.com/libvips/ruby-vips/issues/331.

I found the binaries for this PR build are published here:
https://github.com/msys2/MINGW-packages/actions/runs/2031078489

Does anyone know if we can directly install these published artifacts via `pacman` for testing purposes?

Biswa96:
I have some doubts about the dependency changes. According to the CI output, libvips does not link with aom, libde265 etc. libraries. Are you sure about those changes?

jcupitt:
Hi @Biswa96, thank you for looking at this.

libheif uses those libraries for HEIC and AVIF encode and decode, but you're right, they are already included in the dependencies on that package. I'll remove them from libvips.

PR #11083, opened 2022-03-24
Update Open62541 to version 1.2.4

Samega7Cattac:


Biswa96:
The clang64 failure can be temporarily workaround by removing `-Werror` from CMakeLists.txt file or checking out this commit https://github.com/open62541/open62541/commit/38ff3518b673d0acdb1f7add70c64a003d63bc05 and disable UA_FORCE_WERROR option.

The `pkgver` looks a bit different. Generally, we pin a commit has instead of branch name.

Samega7Cattac:
> The clang64 failure can be temporarily workaround by removing -Werror from CMakeLists.txt file or checking out this commit https://github.com/open62541/open62541/commit/38ff3518b673d0acdb1f7add70c64a003d63bc05 and disable UA_FORCE_WERROR option.

I opened https://github.com/open62541/open62541/issues/5035 for that

> The pkgver looks a bit different. Generally, we pin a commit hash instead of branch name.

In case they backport some fix, they are currently RC for 1.3 but they didn't release a stable version yet.

Btw, is the rest of the script correct? It has some differences from the AUR package (officially maintained but out-of-date)

Biswa96:
Only the pkgver looks incorrect to me, the starting 'v' and the underscores. But I am not sure about it.

Samega7Cattac:
that `pkgver` was updated automatically

jeremyd2019:
Because you changed the pattern in the pkgver function to no longer strip the leading `v` or change the underscores into dots.

Samega7Cattac:
There's a new release to this lib, I will update the PKGBUILD later and do a new PR for it

PR #11084, opened 2022-03-24
baobab: update to 42.0, remove upstreamed patch

rautesamtr:
Removed the upstreamed patch for windows icon in gtk3. Even though this
patch is upstreamed, the fix does not work anymore for gtk4 and the 42
release.


rautesamtr:
"Open Folder" also does not work anymore even with the patch applied, it results in ```Failed to dup() in child process (Bad file descriptor)```

lazka:
hm, it fails to start here with "gdk_gl_context_make_current() failed"

rautesamtr:
Strange, i just did a clean 32bit & 64bit build in a new msys2 install and both just starts fine without any erros.
I also don't get the `gdk_gl_context_make_current() failed` with gtk4-widget-factory.

I am on Windows 11 21H2 22000.556 if that matters.

lazka:
With all the GPU accel it might have to do with my setup (and Gtk not handling all cases or something).

I'm also on Win 11 with an AMD GPU.

lazka:
thanks for re-testing, let's merge this then..

PR #11086, opened 2022-03-24
include-what-you-use: new package

kmilos:


1480c1:
This would be nice

mingw-w64-include-what-you-use/PKGBUILD
```diff
@@ -0,0 +1,53 @@
+# Maintainer: Alexey Pavlov <alexpux@gmail.com>
```
MehdiChinoune, line 1: Is this your name?

mingw-w64-include-what-you-use/PKGBUILD
```diff
@@ -0,0 +1,53 @@
...
+  if check_option "debug" "n"; then
+    extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]]; then
+    CC=clang
+    CXX=clang++
+  fi
```
MehdiChinoune, line 35: those are exported by makepkg-mingw by default!

mingw-w64-include-what-you-use/PKGBUILD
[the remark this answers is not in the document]
kmilos: Thanks, copy/paste gotcha...

PR #11088, opened 2022-03-24
qt6 clangarm64: everything except doc and static

driver1998:


jeremyd2019:
ARM64 CI build with spurious error:
https://github.com/msys2-arm/MINGW-packages/runs/5801097121?check_suite_focus=true

It got to installing websocket, I think all that was left was webconnection, so I'm inclined to say this is OK

PR #11089, opened 2022-03-24
mingw-w64-gegl Prefer pkgconf

lygstate:
Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>


lygstate:
> pkg-config is already provided by pkgconf. What are you trying to fix?

Not fixes anything,just because pkg-config and pkgconf can not coexist.

lazka:
Fine I guess, but this doesn't change anything, pkgconf provides pkg-config, so both can exist.

PR #11094, opened 2022-03-25
Update Qt creator to 7.0.0

Lunar-YZ:


Biswa96:
Would you like to explain the changes please?

Lunar-YZ:
Update Qt creator form 6.0.2 to 7.0.0
when i build the package, tar not worked, so i chagne to 7z to extract files
patchs also not have error, so i ignore patchs operation

lazka:
> when i build the package on my pc, tar not worked, so i chagne to 7z to extract files

What was the error?

> patchs also have error, so i ignore patchs operation

That's not a goo reason to remove patches. What makes you think that you can remove them?

shithappens2016:
> patchs also have error, so i ignore patchs operation

No offense  😂 but this is hilarious.

Lunar-YZ:
> > when i build the package on my pc, tar not worked, so i chagne to 7z to extract files
> 
> What was the error?
> 
> > patchs also have error, so i ignore patchs operation
> 
> That's not a good reason to remove patches. What makes you think that you can remove them?

tar not runing, so i change to 7z to extract files, see images
图片
图片


with patchs, i got this error, so i ignore patch operation, then compile passed
图片

mmuetzel:
> should i just ignore "patch -p1 -i "${srcdir}"/qt-creator-6.0.2-qmldesigner-fix-32bit.patch"? when i ingore this,, patch operation work well.

Ignoring a patch just because it doesn't apply in a newer version is probably never the best way to go about it. Instead it would be necessary to determine *why* it no longer applies:
* It might be because the patch itself or something similar has been applied upstream. In this case, the patch might no longer be necessary and it could be dropped. But you'd need to check if this really is the case.
* It might also be that the code around where the patch originally applied has changed upstream because of *unrelated* reasons. In that case, it might be necessary to adapt the patch so that it can still apply in the newer version. For small changes (e.g., only whitespace changes in the related section of the code), you might be able to adapt the patch manually. Alternatively, you could make the necessary changes in the newer tarball and use a tool like `diff` to create an updated patch.

In any case, you'll need to check *why* it no longer applies.

HTH

lazka:
Updated via #11133 instead.

PR #11100, opened 2022-03-26
adios2: enable for clang32

MehdiChinoune:


jeremyd2019:
you beat me to it - this was the only package enabled by my script in the latest run (other than openshadinglanguage, which failed to rebuild)

PR #11102, opened 2022-03-27
sqlite3: split docs to a separate package

Biswa96:
This reduces package size in half.

Please review. @podsvirov


podsvirov:
I'm sorry for the late reply. I just remembered the [conversation](https://github.com/msys2/MINGW-packages/pull/6336#issuecomment-607752496) about the splitting of documentation into a separate package.

PR #11104, opened 2022-03-27
ffmpeg clangarm64

driver1998:
Disabled stuff:
- rav1e (rust)
- dlfcn (upstream does not support Windows ARM64)
- libmfx (Intel GPU)
- amf (AMD GPU)
- nvenc (NVIDIA GPU)

librsvg uses the 2.40 package, changes clang32 to use it as well


mingw-w64-speexdsp/0001-use-spx-internal-int-types.patch
```diff
@@ -0,0 +1,105 @@
...
+--- a/libspeexdsp/resample_neon.h
++++ b/libspeexdsp/resample_neon.h
+@@ -38,8 +38,8 @@
+ 
+ #ifdef FIXED_POINT
+ #if defined(__aarch64__)
+-static inline int32_t saturate_32bit_to_16bit(int32_t a) {
+-    int32_t ret;
++static inline spx_int32_t saturate_32bit_to_16bit(spx_int32_t a) {
++    spx_int32_t ret;
```
lazka, line 12: What does this change? Shouldn't that be the same type anyway?

driver1998: The file that includes this header, `resample.c`, does not include `stdint.h`, and expects to use these `spx_*` types from `speexdsp_types.h`.

The original code failed with undefined identifiers as a result.

stdint types are not used everywhere except this file, check the x86 sibling `resample_sse.h` for example.

lazka: imo resample_neon.h should just include stdint.h then.

driver1998: Yeah that would do as well.

lazka: thanks!

mati865:
Looks like libvpx requires pthread?

driver1998:
libvpx already depends on libwinpthread-git, why it does not link?

driver1998:
Looks like libwinpthread-git only contains the dll file, but not lib files...

driver1998:
Turns out it is a clang/gcc problem? The configure script forces to use exactly `gcc` to do linker tests if a gcc-like toolchain is specified. So pthread tests failed without `gcc`, and `-lpthread` does not get added to the final LDFLAGS.

mingw-w64-libvpx/0006-do-not-link-with-gcc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/build/make/configure.sh b/build/make/configure.sh
+index 1f56a0a..ad53456 100644
+--- a/build/make/configure.sh
++++ b/build/make/configure.sh
+@@ -700,7 +700,7 @@ setup_gnu_toolchain() {
+   CC=${CC:-gcc}
+   CXX=${CXX:-g++}
+   AR=${AR:-ar}
+-  LD=${LD:-${link_with_cc:-ld}}
++  LD=${LD:-cc}
```
driver1998, line 10: It should be ld, but LDFLAG got a `-pipe` argument from somewhere, and lld by itself does not understand that and errors out. Use cc works fine though.

driver1998: link_with_cc is hardcoded to be gcc in a number of places.

mati865:
> Turns out it is a clang/gcc problem? The configure script forces to use exactly `gcc` to do linker tests if a gcc-like toolchain is specified. So pthread tests failed without `gcc`, and `-lpthread` does not get added to the final LDFLAGS.

Clang doesn't link pthread by the default but GCC does.

jeremyd2019:
> Looks like libwinpthread-git only contains the dll file, but not lib files...

winpthread is one of the rare examples of a split runtime/development package in MINGW-packages.  libwinpthread-git is what you need at runtime, you need winpthreads-git also at build time (so that should probably be added to makedepends rather than changing libwinpthread in depends).

mingw-w64-libvpx/0006-do-not-link-with-gcc.patch
jeremyd2019: could you avoid patching by setting `LD` when calling configure in the PKGBUILD to `${CC}` (or `${CXX}` in case it needs to link some c++)?

mingw-w64-lame/PKGBUILD
```diff
@@ -52,9 +54,11 @@ build() {
     --build=${MINGW_CHOST} \
     --host=${MINGW_CHOST} \
     --target=${MINGW_CHOST} \
-    --enable-nasm \
     --enable-shared \
-    --enable-static
+    --enable-static \
+    $( [[ "${CARCH}" == "i686" \
+      || "${CARCH}" == "x86_64" ]] \
+      && echo "--enable-nasm" || echo "")
```
jeremyd2019, line 61: I see you were bit by `set -e`.  You could use `|| true` since you don't need any output, or reverse the logic and use `|| echo "--enable-nasm")` similar to the usage in makedepends (ditto for the other `|| echo ""` usages)

mingw-w64-libvpx/PKGBUILD
```diff
@@ -4,23 +4,29 @@ _realname=libvpx
...
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 license=('BSD')
-depends=("${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
-makedepends=("${MINGW_PACKAGE_PREFIX}-yasm" "${MINGW_PACKAGE_PREFIX}-autotools" "${MINGW_PACKAGE_PREFIX}-gcc")
+depends=("${MINGW_PACKAGE_PREFIX}-winpthreads")
+makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
+             "${MINGW_PACKAGE_PREFIX}-cc"
+              $( [[ "${CARCH}" != "i686" \
+                && "${CARCH}" != "x86_64" ]] \
+                || echo "${MINGW_PACKAGE_PREFIX}-yasm" ))
```
jeremyd2019, line 17: something like
```suggestion
depends=("${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
             "${MINGW_PACKAGE_PREFIX}-cc"
             "${MINGW_PACKAGE_PREFIX}-winpthreads-git"
              $( [[ "${CARCH}" != "i686" \
                && "${CARCH}" != "x86_64" ]] \
                || echo "${MINGW_PACKAGE_PREFIX}-yasm" ))
```

mingw-w64-libvpx/0006-do-not-link-with-gcc.patch
driver1998: Updated as your suggestion, seems to work.

mingw-w64-ffmpeg/PKGBUILD
```diff
@@ -164,10 +162,18 @@ build() {
     )
   fi
 
-  if [[ "${MINGW_PACKAGE_PREFIX}" != *clang-i686* ]]; then
+  if [[ "${MINGW_PACKAGE_PREFIX}" != *clang-i686* && 
+        "${MINGW_PACKAGE_PREFIX}" != *clang-aarch64* ]]; then
     common_config+=(
        --enable-librav1e
-       --enable-librsvg
```
jeremyd2019, line 170: Should `--enable-librsvg` have moved into the giant common_config list above?

mingw-w64-libvpx/PKGBUILD
```diff
@@ -44,29 +49,36 @@ prepare() {
 
   apply_patch_with_msg 0001-enable-shared-on.mingw.patch
   apply_patch_with_msg 0003-cross-with-native-binutils.mingw.patch
-  apply_patch_with_msg 0005-fix-exports.mingw.patch
+  apply_patch_with_msg 0005-fix-exports.mingw.patch
```
jeremyd2019, line 52:
```suggestion
  apply_patch_with_msg 0005-fix-exports.mingw.patch
```

mingw-w64-ffmpeg/PKGBUILD
driver1998: Thanks, added now.

jeremyd2019:
Successful ARM64 CI run (on my raspberry pi so forgive the 8 hour runtime :grin:)
https://github.com/msys2-arm/MINGW-packages/runs/5730588389?check_suite_focus=true

Issue #11115, opened 2022-03-28
aalib package


Jehan:
In GIMP, we use the library `aalib` for ascii-art export. Here is our build rules for Windows: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/build/windows/gitlab-ci/build-deps-msys2.sh#L54

As you can see, there are 2 patches, also we update manually some `config.*` files because the ones in the tarball were too old and not recognizing some modern platforms, which also means we need to regenerate the autotools files afterwards.

It's explained in this commit message: https://gitlab.gnome.org/GNOME/gimp/-/commit/62ae32123a12492926bfcc2563ee780567b46f4b

Now we haven't had any problem with it and it's not like it's taken much time in our build. But if MSYS2 wanted to integrate this as a package, it would still be easier for us. :-)

Biswa96:
Hi @Jehan, Thank you for reporting the issue. I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/2054503927

Jehan:
Sure. Could you just explain me how to install these? I see I can download a `MINGW64-packages.zip` (which I guess is what we want). Is that it? Where should I unzip it in our windows CI? Or can pacman install from this?

jeremyd2019:
inside MINGW-packages.zip are pacman packages (actually just one in this case).  after unzipping you could do something like `pacman -U *.pkg.tar.*`

lazka:
(if you're only testing via CI, we also just merge it and you can test once it's in the repo)

Jehan:
@lazka Yes I am only testing through CI, then check the log to see it were correctly detected, and would finally test the CI-generated installer in a Windows VM.

I'm not doing a full install of however-this-all-MSYS2-on-Windows thing work, and a build of GIMP in a Windows VM, for sure. 😜

Anyway I can try to `pacman` in a feature branch in the CI tomorrow, either from this zip file, or from the main package if you push it (which obviously would be simpler). Just tell me. :-)

lazka:
@Jehan -> https://packages.msys2.org/base/mingw-w64-aalib

Jehan:
CI running! I'll comment with result in a few hours, I guess. :-)

Biswa96:
> CI running! I'll comment with result in a few hours

[...]

PR #11116, opened 2022-03-28
aalib: Add package for gimp

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/11115


jeremyd2019:
Is there still an active upstream?  I wonder if it would be worth trying to get the first patch changes related to malloc.h in a better state (the whole point of `AC_CHECK_HEADERS` is that the header may not be present on all systems, it defines a preprocessor macro if it is, so really the include of malloc.h should be in an ifdef rather than removed).  If upstream isn't active, probably not worth the effort because the patch works fine.

Biswa96:
> Is there still an active upstream?

Last Update: [2013-05-02](https://sourceforge.net/p/aa-project/activity) 😁

jeremyd2019:
OK then, I guess nobody is left there to complain if we break pre-C89 compatibility.

Biswa96:
For future reference. The patches were removed from GIMP in this commit https://gitlab.gnome.org/GNOME/gimp/-/commit/1349168d7766685433624a050ab7d7ed7a8c35d4

PR #11123, opened 2022-03-29
[CI] Add a post-build job to grok packages.

MehdiChinoune:


jeremyd2019:
That's been my plan, but the scripts need a little more attention before I'd feel comfortable turning them loose on arbitrary pull requests: jeremyd2019/package-grokker#10.

Even then, a PR for something like gcc (so gcc-libs changed) would result in almost all packages being downloaded, which may not be a great thing.  I was trying to avoid this in early versions of the script by implementing a 'local mirror' option so somebody who already had a mirror of the repo could just run against that instead of downloading, but that option has not gotten much exercise because running it in GHA is just too handy.

These concerns not necessarily show-stoppers, but should be considered and accepted before merging.

MehdiChinoune:
> Even then, a PR for something like gcc (so gcc-libs changed) would result in almost all packages being downloaded, which may not be a great thing.

gcc-libs are backward compatible, They don't remove symbols or change libs names. Even if it happened the worst thing is that the job will timeout after 6 hours of running (It happened to me with intel-tbb libs)

jeremyd2019:
> Even if it happened the worst thing is that the job will timeout after 6 hours of running (It happened to me with intel-tbb libs)

The worst thing is that the job will consume a bunch of mirror bandwidth for no good reason, which is what I'm trying to avoid (especially when a package is "known" to be backward compatible like gcc-libs).

MehdiChinoune:
> > Even if it happened the worst thing is that the job will timeout after 6 hours of running (It happened to me with intel-tbb libs)
> 
> The worst thing is that the job will consume a bunch of mirror bandwidth for no good reason, which is what I'm trying to avoid (especially when a package is "known" to be backward compatible like gcc-libs).

Isn't package-grokker supposed to run only when dll name changed or symbols removed, then we have not to worry about gcc-libs at all!

jeremyd2019:
> Isn't package-grokker supposed to run only when dll name changed or symbols removed, then we have not to worry about gcc-libs at all!

Yes, I noticed that it was downloading bunches of packages when it didn't need to and added an `if` to avoid it doing that if no dlls or symbols were removed (in a commit added just 5 days ago).

gcc-libs is just an example.  If they absolutely, positively never remove a dll or symbol export it should never be a problem.  A concrete example that actually happened: in your llvm 14 pull request, a few symbols were removed from libc++.dll  If run against a CLANG MSYSTEM instead of MINGW64, that would have the same effect (though arguably, it should, because if somebody is importing those symbols we need to know about it).

I'm not arguing that this is something that needs changing or anything, just that it be kept in mind as a potential foot-gun that could cause issues for a mirror.  More important IMO is jeremyd2019/package-grokker#10.

.github/workflows/main.yml
```diff
@@ -131,3 +131,33 @@ jobs:
...
+          pip install -r requirements.txt
+          deactivate
+      - name: Grok packages
+        if: steps.artifacts.outcome == 'success'
+        run: |
+          . venv/bin/activate
+          for pkg in artifacts/*.pkg.tar.*; do
+            pkgname=$(basename "$pkg" | rev | cut -d- -f4- | rev)
+            python -u package-update-check.py -e MINGW64 -p "${pkgname}" @PKG@ "file://$(pwd)/${pkg}"
+          done
```
jeremyd2019, line 163: I think I've got the package-grokker code in a better state for this now.  Invocation changed because now it can process a directory of artifacts rather than needing to be called in a loop.

```suggestion
          python -u package-update-check.py -v -e mingw64 artifact_dir ./artifacts
```

jeremyd2019:
I spent a little bit of time on GitHub's 'reusable' mechanisms.  TL;DR I'm not sold on any of them being a great option here.

The most straightforward mechanism for this would seem to have been a composite action.  These automatically get a checkout of their repo so they can run scripts, and they run within a job so the caller could be responsible for getting the artifacts downloaded.  The only downside of that was I wasn't as happy with how the output formatting looked.  Example code https://github.com/jeremyd2019/package-grokker/compare/composite and test run https://github.com/jeremyd2019/package-grokker/runs/5804216689?check_suite_focus=true

I then tried a reusable workflow.  These do not get a checkout of the contents of their repo, but run as their own job and format the same way as any other job does.  Because the workflow includes download-artifact, the 'reusable workflow' couldn't actually be reused for the workflow in package-grokker (because that uses a different action to download artifacts from a different repository). Reusable workflow: https://github.com/jeremyd2019/package-grokker/blob/main/.github/workflows/grok-artifacts.yml, Example code: https://github.com/jeremyd2019/MINGW-packages/commit/0793317887f936b7c9da484357bac5456cc68d70 and test run https://github.com/jeremyd2019/MINGW-packages/runs/5804634712?check_suite_focus=true

jeremyd2019:
I think I figured out how to get the composite action output formatting to where I like it.  The main problem is that github doesn't seem to respect names on composite action steps.

.github/workflows/main.yml
```diff
@@ -131,3 +131,30 @@ jobs:
...
+        run: |
+          python -m venv --system-site-packages venv
+          . venv/bin/activate
+          pip install -r requirements.txt
+          deactivate
+      - name: Grok packages
+        if: steps.artifacts.outcome == 'success'
+        run: |
+          . venv/bin/activate
+          python -u package-update-check.py -v -e mingw64 artifact_dir ./artifacts
```
jeremyd2019, line 160:
```suggestion
      - uses: actions/download-artifact@v3
        id: artifacts
        continue-on-error: true
        with:
          name: MINGW64-packages
          path: artifacts
      - name: Grok packages
        uses: jeremyd2019/package-grokker/grok-artifacts@main
        if: steps.artifacts.outcome == 'success'
        with:
          repo: mingw64
          path: artifacts
```

MehdiChinoune:
```
jeremyd2019/package-grokker/grok-artifacts@main is not allowed to be used in msys2/MINGW-packages.
Actions in this workflow must be: within a repository owned by msys2, created by GitHub or match the following:
msys2/setup-msys2@*, peaceiris/actions-gh-pages@*.
```

jeremyd2019:
looks like somebody locked down permissions :grin:.  It could be added to the whitelist in settings, or I wouldn't be opposed to transferring package-grokker to msys2 if that's preferred.  Or just forget using a composite action or reusable workflow.

MehdiChinoune:
It runs on my fork!

lazka:
>  It could be added to the whitelist in setting

I've added it to the whitelist

jeremyd2019 (approved):

jeremyd2019:
hearing no objections, let's give this a try!

PR #11125, opened 2022-03-29
hiredis: Use the same name for static and shared libs

orgads:


Biswa96:
Please add the upstream pull request link as a comment. It will help us to track it in future.

orgads:
Done

PR #11137, opened 2022-03-31
keystone: 0.9.2 (new package)

raehik:


raehik:
Any way I can test `clangarm64` (without an applicable local system)? I imagine it won't build, but I was curious to try.

Biswa96:
It is possible to cross-compile a project for clangarm64 with [mstorsjo/llvm-mingw](https://github.com/mstorsjo/llvm-mingw) and run it in win-arm in qemu. But the procedure is somewhat complicated at first. If you need to check anything I can try this project in my setup. Just provide the commands and expected output.

raehik:
@Biswa96 Thanks, I'd appreciate it - easy project to build (PR worked first time). If you can use `makepkg` in win-arm, then you should be able to build the PKGBUILD here. Otherwise, it's a fairly regular CMake project, though one part will fail to build due to a non-MSYS Windows fix breaking things (see patches). Build deps are just cmake and a C++ compiler (PKGBUILD uses Ninja and needs Python too).

Biswa96:
There is a warning in build output:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    LLVM_BUILD_32_BITS
    MAKE_OSX_ARCHITECTURES
```

Are those required for this project?

raehik:
Nope, they're just from the install script that I used to form my cmake invocation. They can be removed if it's preferred.

Biswa96:
I have cross compiled this project with llvm-mingw toolchain for win aarch64 and it works in my qemu setup. But I have not done any actual test with the program. Ready to merge?

Biswa96:
The python module is not built/installed in the package. This ArchLinux's [PKGBUILD][1] can be followed to create separate keystone python package.

[1]: https://github.com/archlinux/svntogit-community/blob/packages/keystone/trunk/PKGBUILD

mingw-w64-keystone/0001-cmake-install-dir.patch
```diff
@@ -0,0 +1,25 @@
...
++++ b/llvm/keystone/CMakeLists.txt
+@@ -43,12 +43,18 @@ add_library(keystone
+   EVMMapping.cpp
+ )
+ 
+-install(TARGETS keystone DESTINATION lib${LLVM_LIBDIR_SUFFIX})
++install(TARGETS keystone
++    ARCHIVE DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
++    LIBRARY DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
++    RUNTIME DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}"
```
MehdiChinoune, line 13:
```suggestion
+    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
```
I think relative path are better, and that's what cmake projects use

mingw-w64-keystone/0001-cmake-install-dir.patch
```diff
@@ -0,0 +1,25 @@
...
++)
+ 
+ set_target_properties(
+   keystone
+   PROPERTIES
+-  archive_output_directory ${CMAKE_CURRENT_BINARY_DIR}/lib
++  ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib
++  LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib
++  RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin
```
MehdiChinoune, line 22: This is not needed!

mingw-w64-keystone/0002-kstool-win-msys2.patch
```diff
@@ -0,0 +1,23 @@
...
+--- a/kstool/kstool.cpp
++++ b/kstool/kstool.cpp
+@@ -6,18 +6,11 @@
+ #include <inttypes.h>
+ #include <errno.h>
+ #include <limits.h>
+-#if !defined(WIN32) && !defined(WIN64) && !defined(_WIN32) && !defined(_WIN64)
+ #include <unistd.h>
+-#else
+-#endif
```
MehdiChinoune, line 12:
```suggestion
-#if !defined(WIN32) && !defined(WIN64) && !defined(_WIN32) && !defined(_WIN64)
+#ifndef _MSC_VER
 #include <unistd.h>
-#else
 #endif
```
This is not how to fix issues!

mingw-w64-keystone/0002-kstool-win-msys2.patch
```diff
@@ -0,0 +1,23 @@
...
+ #include <fcntl.h>
+ 
+ #include <keystone/keystone.h>
+ 
+-#if defined(WIN32) || defined(WIN64) || defined(_WIN32) || defined(_WIN64)
+-#include "getopt.h"
+-#endif
+-
```
MehdiChinoune, line 20:
```suggestion
-#if defined(WIN32) || defined(WIN64) || defined(_WIN32) || defined(_WIN64)
+#ifdef _MSC_VER
 #include "getopt.h"
 #endif
 
```
same as above

mingw-w64-keystone/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  cmake \
+    -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=ON \
+    -DLLVM_TARGETS_TO_BUILD=all \
+    -DBUILD_LIBS_ONLY=0 \
+    -DLLVM_BUILD_32_BITS=0 \
+    -DMAKE_OSX_ARCHITECTURES= \
```
MehdiChinoune, line 45:
```suggestion
    -DBUILD_LIBS_ONLY=OFF \
```

mingw-w64-keystone/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+
+_realname=keystone
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.9.2
+pkgrel=1
+pkgdesc="Lightweight multi-platform, multi-architecture assembler framework"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+depends=()
```
MehdiChinoune, line 11:
```suggestion
depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
```
Take a look at `runtime-dependencies` https://github.com/msys2/MINGW-packages/pull/11137/checks#step:9:1631

raehik: Thanks, I didn't catch that!

mingw-w64-keystone/0001-cmake-install-dir.patch
raehik: OK. That fits my experience.

raehik:
> The python module is not built/installed in the package. This ArchLinux's [PKGBUILD](https://github.com/archlinux/svntogit-community/blob/packages/keystone/trunk/PKGBUILD) can be followed to create separate keystone python package.

Yep. I don't mind about the Python bindings, I'd let them be provided by another package.

I'm not sure if you were querying, but we do still need Python if we're not building the Python bindings. It's required as part of building LLVM (the llvm-build util is a Python script). (This is also to placate me because I had idly added it to deps without checking where the error came from.)

mingw-w64-keystone/0002-kstool-win-msys2.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/kstool/kstool.cpp b/kstool/kstool.cpp
+index 0d5ff45..0ebefb4 100644
+--- a/kstool/kstool.cpp
++++ b/kstool/kstool.cpp
+@@ -6,18 +6,11 @@
```
MehdiChinoune, line 5:
```suggestion
@@ -6,18 +6,17 @@
```
Don't forget to update the chack256sums

raehik: Regenerated the patches and all is well!

mingw-w64-keystone/0001-cmake-install-dir.patch
raehik: Thanks, removed (was copied from another pkg where DLLs were placed correctly)

raehik:
I can see libwinpthread in the runtime deps -- @MehdiChinoune as you did with other pkgs I'll add `libwinpthread-git` to deps explicitly

raehik:
Along with the review changes, I fixed some warnings related to pkgdesc and license. If there are no more fixes to make, I think this is ready to merge.

MehdiChinoune:
> I can see libwinpthread in the runtime deps -- @MehdiChinoune as you did with other pkgs I'll add `libwinpthread-git` to deps explicitly

where?
Please add only the libraries not found by ntldd. libwinpthread-git is a dependency of gcc-libs on MINGW/UCRT envs.

raehik:
> where? Please add only the libraries not found by ntldd. libwinpthread-git is a dependency of gcc-libs on MINGW/UCRT envs.

Oh you're right, Clang64 doesn't have it. Reverted.

PR #11138, opened 2022-03-31
hdf5: update to 1.13.1

MehdiChinoune:
I think the first patch should be dropped, It doesn't fix anything.
What's the purpose of adding `-0` to dll name?.
(Although, We have to rebuild all dependent packages!)


lazka:
> I think the first patch should be dropped, It doesn't fix anything.
> What's the purpose of adding `-0` to dll name?.
> (Although, We have to rebuild all dependent packages!)

Feel free to drop it +  rebuilds.

Not sure why it's there.. non-cmake packages tend to communicate their ABI version that way, but cmake packages rarely bother.. Not worth patching it imo.

MehdiChinoune:
@lazka Please revert this, since it is still in staging.
It turns out that releases with odd minor versions are experimental!
https://www.hdfgroup.org/2021/12/hdf5-1-13-0-introducing-experimental-releases/

lazka:
oh, thanks, will do.

MehdiChinoune:
Will you remove it from staging databases?

lazka:
It will do that automatically, will take a few minutes..

PR #11141, opened 2022-03-31
leptonica: update to 1.82.0

kmilos:


Biswa96:
Potential candidate for clangarm64.

```
$ file src/bin/src/.libs/liblept-5.dll
src/bin/src/.libs/liblept-5.dll: PE32+ executable (DLL) (GUI) Aarch64, for MS Windows
````

kmilos:
I thought so, but then don't know how to handle the checkdepends which is not available for clangarm64...

MehdiChinoune:
Isn't possible to build with cmake?

Biswa96:
Need some changes for cmake build.

kmilos:
Yeah, I already saw a bunch of CMake related changes upstream, best to leave it until they release the next version...

PR #11142, opened 2022-03-31
openslide: 3.4.1 (new-package)

MehdiChinoune:


MehdiChinoune:
```
  *** Warning: linker path does not have real file for library -lz.
  *** I have the capability to make that library automatically link in when
  *** you link to this library.  But I can only do this if you have a
  *** shared version of the library, which you do not appear to have
  *** because I did check the linker path looking for a file starting
  *** with libz and none of the candidates passed a file format test
  *** using a file magic. Last file checked: D:/M/msys64/clang64/lib/libz.dll.a
  *** The inter-library dependencies that have been dropped here will be
  *** automatically added whenever a program is linked with this library
  *** or is declared to -dlopen it.
  
  *** Since this library must not contain undefined symbols,
  *** because either the platform does not support them or
  *** it was explicitly requested with -no-undefined,
  *** libtool will only create a static version of it.
```
Could someone explain this!

Biswa96:
Try to add `-autoreconf -fiv` in prepare function. If it does not work try with adding `-no-undefined` to the LDFLAGS in configure.ac or Makefile.am function.

MehdiChinoune:
`autoreconf -fiv` fixed the issue.

PR #11150, opened 2022-04-01
openturns: Don't apply the upstream patch and fix *cmake files instead

MehdiChinoune:
VTK 9.1 couldn't use OpenTURNS with that patch applied.
https://gitlab.kitware.com/vtk/vtk/-/issues/18500


MehdiChinoune:
Please merge this, so I could upgrade VTK

PR #11156, opened 2022-04-02
mingw-w64-netcdf: bump build number to force rebuild against reverted HDF5

rouault:
mingw-w64-netcdf 4.8.1-3 was built on March 30, against hdf5 1.13.1.
However hdf5 was reverted to 1.12.1 on April 1st (cf
https://github.com/msys2/MINGW-packages/commits/master/mingw-w64-hdf5).
Consequently netcdf must be rebuilt against this older hdf5, otherwise
runtime issues occur, such as:

```
$ ncgen -k 4 -o out2.nc out.txt
Warning! ***HDF5 library version mismatched error***
The HDF5 header files used to compile this application do not match
the version used by the HDF5 library to which this application is linked.
Data corruption or segmentation faults may occur if the application continues.
This can happen when an application was compiled by one version of HDF5 but
linked with a different version of static or shared HDF5 library.
You should recompile the application or check your shared library related
settings such as 'LD_LIBRARY_PATH'.
You can, at your own risk, disable this warning by setting the environment
variable 'HDF5_DISABLE_VERSION_CHECK' to a value of '1'.
Setting it to 2 or higher will suppress the warning messages totally.
Headers are 1.13.1, library is 1.12.1
```
(I'm not familiar with the way packages are rebuilt with this distribution, so I assumed bumping the rev number was the correct thing to do)


lazka:
oops, sry

lazka:
fixed now

PR #11177, opened 2022-04-03
python-setuptools: update to 61.3.1

MehdiChinoune:


lazka:
We can't update setuptools, anything >=60 is broken with our Python e

PR #11202, opened 2022-04-03
python-labelme: update to 5.0.1

MehdiChinoune:
`matplotlib` is already at `3.3.4>=3.3`

### DIFF
--- a/mingw-w64-python-labelme/PKGBUILD
+++ b/mingw-w64-python-labelme/PKGBUILD
@@ -3,33 +3,43 @@
 _realname=labelme
 pkgbase=mingw-w64-python-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
-pkgver=4.5.7
-pkgrel=3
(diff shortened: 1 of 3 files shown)


MehdiChinoune:
```
matplotlib 3.5.1 requires packaging, which is not installed.
labelme 5.0.1 requires natsort, which is not installed.
labelme 5.0.1 has requirement matplotlib<3.3, but you have matplotlib 3.5.1.
```
Why didn't find that `matplotlib 3.5.1 requires packaging` in #11199 @naveen521kk

MehdiChinoune:
Another thing, I think failing bacause a python module requires a lower version of another module that has already been updated should be disabled, because we could do nothing about it.
If the new version doesn't support that version of the other package, an old version won't either.

The check should be restricted to the cases where the updated package requires a newer version of one of its dependencies, or one of its reverse-dependencies requires a lower version of it (updated package).
Also only the updated packages should be checked, Why should we deal with unmet dependencies of a package that we didn't touch?

PR #11205, opened 2022-04-04
CI: run `pip check` with & without reverse-dependencies installed

MehdiChinoune:
@naveen521kk

### DIFF
--- a/.ci/ci-check.py
+++ b/.ci/ci-check.py
@@ -197,9 +197,11 @@ def main() -> None:
     for pkgloc in ARTIFACTS_LOCATION.glob("*.pkg.tar.*"):
         with install_package(pkgloc):
             pkgname = get_pkg_name(pkgloc)
+            with gha_group(f"Pip Output (without reverse dependencies): {pkgloc.name}"):
+                run_pip_check(pkgname)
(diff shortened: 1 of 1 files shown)


.ci/ci-check.py
```diff
@@ -197,6 +197,8 @@ def main() -> None:
     for pkgloc in ARTIFACTS_LOCATION.glob("*.pkg.tar.*"):
         with install_package(pkgloc):
             pkgname = get_pkg_name(pkgloc)
+            with gha_group(f"Pip Output: {pkgloc.name}"):
+                run_pip_check(pkgname)
             rdeps = get_rdeps(pkgname)
             with install_package(rdeps):
                 with gha_group(f"Pip Output: {pkgloc.name}"):
```
MehdiChinoune, line 204: I don't know what we should print in output?
Should we print `Pip Output (without reverse dependencies):` and `Pip Output (with reverse dependencies):`

naveen521kk: +1 for printing so, `Pip Output (without reverse dependencies):` and `Pip Output (with reverse dependencies):`

naveen521kk (approved):
lgtm, thanks!

PR #11206, opened 2022-04-04

kmilos:


jeremyd2019:
Successful ARM64 CI run:
https://github.com/msys2-arm/MINGW-packages/runs/5890691212?check_suite_focus=true

PR #11210, opened 2022-04-05
protobuf: update to 3.20.0

MehdiChinoune:


jeremyd2019:
```
Removed DLLs/Symbols
  {b'libprotobuf-lite.dll': {None,
```

That's interesting, I don't recall seeing a `None` symbol show up before

MehdiChinoune:
ready

PR #11237, opened 2022-04-05
QScintilla: update to version 2.13.2

mmuetzel:
The build system of QScintilla has changed notably between versions. It looks like all of the patches and most of the other hacks in the PKGBUILD rules are no longer needed to get it to build correctly.

One `.dll.debug` file no longer builds. I don't know what that was used for or how to build it with the new version.
Additionally, the name of the `.pyd` file in the Python package has changed. I don't know if that means that dependers need to be rebuilt?

Other than that, the updated library seems to work just like the previous one for me.


jeremyd2019:
> I don't know if that means that dependers need to be rebuilt?

It just so happens we recently added a new job to CI to try to answer that question in an automated fashion.

jeremyd2019:
https://github.com/msys2/MINGW-packages/runs/5839029505?check_suite_focus=true#step:3:44 - job saw the removed .pyd file and found no dependent package with binaries that link to it.

lazka:
.pyd files are never linked against, but as long as the first part of the pyd is named the same there is no difference for users.

PR #11242, opened 2022-04-05
jsoncpp: update to 1.9.5

MehdiChinoune:


lazka:
ABI break, needs rebuilds

MehdiChinoune:
- Building CMake with CMake is not reliable
- Again with CMake picking libglut32 over freeglut on 32bit environments, First was partio, Second was mathgl and now vrpn.
  As I said before Either get rid of libglut32 or we patch CMake to ignore it.

mingw-w64-cmake/PKGBUILD
```diff
@@ -99,10 +102,11 @@ build() {
     MSYSTEM=MINGW FC=${MINGW_PREFIX}/bin/gfortran.exe \
       "${srcdir}"/${_realname}-${pkgver}/configure  \
       --prefix=${MINGW_PREFIX}                      \
+      --datadir=/share/cmake                        \
+      --docdir=/share/doc/cmake                     \
+      --mandir=/share/man                           \
```
jeremyd2019, line 107: I'm concerned with these looking like 'absolute paths' now.  Were they not working as 'relative' paths (at least the ones that already existed below)?

MehdiChinoune: I have just copied it from ArchLinux.

jeremyd2019: would probably have been ok in a configure script, but my concern was that some native Windows executable would get them at some point, and the msys2 arg conversion would try to turn them into absolute Windows paths.  As long as the 'relative' form works, no reason to take the risk.

MehdiChinoune:
ready

MehdiChinoune:
ready

PR #11272, opened 2022-04-05
python-hunter: update to 3.3.3

MehdiChinoune:


Biswa96:
Need to remove shebang line.

```
ucrt64/bin/hunter-trace-script.py:#!D:/M/msys64/ucrt64/bin/python.exe
```

Biswa96:
This package installs bundled python-colorama files. Should it use the mingw packaged one?

MehdiChinoune:
It is not possible: https://github.com/ionelmc/python-hunter/blob/v3.3.3/src/hunter/util.py#L16

PR #11277, opened 2022-04-06
mesa: rebuild for llvm/clang 14

jeremyd2019:


lygstate:
What's the          "${MINGW_PACKAGE_PREFIX}-libsystre"
means?
I've seen no depends in mesa

Biswa96:
Here is the package details https://packages.msys2.org/base/mingw-w64-libsystre

PR #11278, opened 2022-04-06
qt6-static: rebuild for llvm/clang 14 (and build fixes)

jeremyd2019:
prune some makedepends that were only there due to messed up llvm cmake files, which have been cleaned up by building llvm projects separately.

turn on system harfbuzz.  It's already being pulled in via freetype, apparently, because I get
```
 ld.lld: error: duplicate symbol: hb_font_funcs_create
 >>> defined at libQt6BundledHarfbuzz.a(hb-font.cc.obj)
 >>> defined at libharfbuzz.a(hb-font.cc.obj)
```
add -lusp10 to HARFBUZZ_DEPENDENCIES.  This is a system DLL import library, was getting a couple of undefined symbols when linking.

### DIFF
--- a/mingw-w64-qt6-static/PKGBUILD
+++ b/mingw-w64-qt6-static/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}-static
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-static")
 _qtver=6.2.1
 pkgver=${_qtver/-/}
-pkgrel=1
+pkgrel=2
 arch=(any)
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I see that qt6-static was switched to bundled harfbuzz in e5f1c3c456668c2dbc2cf0f2e77375bf26d7645a.  I'm not sure of the reason for that change, though, so I don't know if my switching it back to using system harfbuzz is going to break something else.  Notably, qt6-static still uses system freetype, which was built against system harfbuzz, so it seems likely that swapping harfbuzz out for the probably different bundled version may cause issues beyond the duplicate symbol errors I saw from it trying to link in both versions of harfbuzz.

lazka:
filename length limit?

lazka:
2 builds are done, the rest will take a while

PR #11288, opened 2022-04-06
python-keyring: update to 23.5.0

MehdiChinoune:


Biswa96:
Need to remove shebang line.

```
ucrt64/bin/keyring-script.py:#!D:/M/msys64/ucrt64/bin/python.exe
```

PR #11316, opened 2022-04-06
wxWidgets3.1: update to 3.1.6

driver1998:


jeremyd2019:
https://github.com/msys2/MINGW-packages/runs/5846454144?check_suite_focus=true#step:3:60
Needs rebuilds:
mingw-w64-wxPython
mingw-w64-kicad

stahta01:
I could not get mingw-w64-wxPython to build yesterday because of [Replace DoGetBorderSize() with GetWindowBorderSize()](https://github.com/wxWidgets/wxWidgets/pull/2445) and I could not think of a simple C++ fix. In C, I would do a define; but, I am still not a great C++ programming and while I know define is not a good C++ thing. No idea what to use instead.

Tim S.

jeremyd2019:
Damn.
I applied the DoGetBorderSize -> WindowGetBorderSize change to wxPython's bundled wxWidgets (which it runs doxygen against as a first step to autogenerating wrappers for all functions).
It then failed because there is special handling for virtual methods on wxWindow.  I fixed that 'tweak'.
It then failed because there was a wxBitmap -> wxBitmapBundle change in wxHeaderColumn.  I applied *that* change to the bundled wxWidgets, and *that* failed because the sip wrapper claims not to know about wxBitmapBundle.  And sure enough, that class does not seem to be wrapped.

I'd say we should wait for an update to wxPython, but I've noticed a distinctly slow pace of updates over the last couple of years from them.

(there's a bug in the git snapshot of wxWidgets bundled with the latest release of wxPython, that causes transparency in bitmaps to be drawn as black on Windows.  I am currently stuck on python 3.8 for my wxPython projects, because that is the last version of python for which upstream published a wheel that doesn't have that bug).

jeremyd2019:
I consider wxPython is a blocker for updating to wxWidgets to 3.1.6.  I may keep trying to plug away at it, but maybe upstream will figure it out before I do.  If you need an arm manifest fix in the short term, you'd probably be better off backporting it to 3.1.5 as you did for wxWidgets 3.0

driver1998:
> I consider wxPython is a blocker for updating to wxWidgets to 3.1.6.  I may keep trying to plug away at it, but maybe upstream will figure it out before I do.  If you need an arm manifest fix in the short term, you'd probably be better off backporting it to 3.1.5 as you did for wxWidgets 3.0

Currently I only need 3.0 (for PCem), I'll send a backport when I hit it on 3.1.5.

PR #11345, opened 2022-04-06
python-can: update to 4.0.0

MehdiChinoune:


mingw-w64-python-can/001-remove-windows-curses-dependency.patch
```diff
@@ -0,0 +1,10 @@
+--- a/setup.py
++++ b/setup.py
+@@ -85,7 +85,6 @@
+     install_requires=[
+         "setuptools",
+         "wrapt~=1.10",
+-        'windows-curses;platform_system=="Windows" and platform_python_implementation=="CPython"',
```
Biswa96, line 7: Shouldn't windows-curses package be added?

MehdiChinoune: They don't offer any archive to build the, neither on pypi nor in their github repository.
The only way to build it is to clone the repo at specific commit => need extra efforts!

PR #11356, opened 2022-04-06
winsparkle: update to 0.7.0

MehdiChinoune:
It fails to build with WxWidgets 3.0

PR #11371, opened 2022-04-07
gnuplot: Switch back to wxWidgets version 3.0

stahta01:
I was checking what uses wxWidgets 3.1 and found this program failed to find wx3.1 with this build message
`configure: WARNING: wxWidgets can't be found. You can try --with-wx=DIR to give the right path to wx-config. The wxWidgets terminal will not be compiled.`

I changed it back to wxWidgets version 3.0 and it runs a graph file I found online using wxt or wxWidgets terminal

Tim S.


MehdiChinoune:
The issue is that WxWidgets3.1 executables were suffixed by `-3.1`, so during configuration most programs search for `wx-config` while WxWidgets3.1 has `wx-config-3.1`.
Maybe we should make WxWidgets3.1 the default?

stahta01:
It would make more sense to wait till wxWidgets 3.2 is released; 3.2 is a production version and 3.1 is a development version.

Note: If we can get most of the wxWidgets packages using 3.1; then maybe going to 3.1 would make sense.

wxPython will not build using wxWidgets 3.1.6 because of [Replace DoGetBorderSize() with GetWindowBorderSize()](https://github.com/wxWidgets/wxWidgets/pull/2445)

Tim S.

stahta01:
If the devs want, I can start work on making the wxWidgets3.1 package provide both wx-config and wx-config-3.1 and the wxWidgets provide wx-config-3.0 without providing wx-config and see what it breaks. Is this wanted?

I think waiting till wxPython builds would be best before upgrading to wxWidgets 3.1.6 that was just released.

Tim S.

MehdiChinoune:
On #11356 , It fails with WxWidgets 3.0

stahta01:
I am now trying to build using wxWidgets 3.1

stahta01:
I have combined my minor changes; it fails to build with wxWidgets 3.1; but, does build with 3.0.

PR #11378, opened 2022-04-07
adios2: update to 2.8.0

MehdiChinoune:


Biswa96:
Is it possible to use this commit https://github.com/ornladios/ADIOS2/commit/cefccf4e8b80b9c663c370f955766624779c77b4 to fix the type conflicting issue? The patch need some modification to apply in 2.8.0 tag.

MehdiChinoune:
> Is it possible to use this commit [ornladios/ADIOS2@cefccf4](https://github.com/ornladios/ADIOS2/commit/cefccf4e8b80b9c663c370f955766624779c77b4) to fix the type conflicting issue? The patch need some modification to apply in 2.8.0 tag. Also please add the link as a comment for future reference.

It couldn't, I tried it before making my patch.

PR #11200, opened 2022-04-03
python-sphinx_rtd_theme: update to 1.0.0

MehdiChinoune:
I could do nothing about `docutils` dependency that has been already updated to 0.18.1

PR #11379, opened 2022-04-07
python-sphinx: add new dependency python-importlib-metadata

heljkon:
Starting with Sphinx 4.4.0 a dependency 'importlib-metadata' was introduced:	
https://github.com/sphinx-doc/sphinx/commit/6c5c66bbb19cb2dcfdc25da97d4d839eec139db7

The lack of this dependency caused the lack of documentation in #11000 

@Biswa96 Do you agree?


Biswa96:
Looks good to me. According to https://github.com/sphinx-doc/sphinx/blob/v4.5.0/setup.py

```
"importlib-metadata>=4.4; python_version < '3.10'"
```

MehdiChinoune:
This is part of #11200

PR #11385, opened 2022-04-07
rust: Update to 1.60.0

lazka:


jeremyd2019:
It should (hopefully) be safe to enable the 'strip' option now...

lazka:
The clang64 build seems to work fine with stripping now.

PR #11392, opened 2022-04-07
mingw-w64: update to 10.0.0

jeremyd2019:
I would have updated the pkgver to 10.0.0 even, since this is the commit
officially tagged as v10.0.0, but this package is updated against git
commits so often that tweaking the pkgver function for one version
didn't seem worth the effort.

Take this opportunity to switch pkgver to using git describe, as doing
so when the more significant version numbers don't change would make the
new version compare lower.

The --long argument and sed pattern for git describe comes from the
CMake-git PKGBUILD template.

### DIFF
--- a/mingw-w64-headers-git/PKGBUILD
+++ b/mingw-w64-headers-git/PKGBUILD
@@ -4,9 +4,9 @@ _realname=headers
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgdesc="MinGW-w64 headers for Windows"
-pkgver=9.0.0.6454.b4445ee52
+pkgver=10.0.0.r0.gaa08f56da
 pkgrel=1
-_commit='b4445ee520dd5041b051fb30ffee00970f50a118'
+_commit='aa08f56da559016f10336dddca85d59f9bdc9e02'
 arch=('any')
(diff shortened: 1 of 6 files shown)


mingw-w64-headers-git/PKGBUILD
```diff
@@ -25,7 +25,7 @@ sha256sums=('SKIP'
 
 pkgver() {
   cd "${srcdir}/mingw-w64"
-  printf "9.0.0.%s.%s" "$(git rev-list --count $_commit)" "$(git rev-parse --short $_commit)"
+  printf "10.0.0.%s.%s" "$(git rev-list --count $_commit)" "$(git rev-parse --short $_commit)"
```
lazka, line 28:
```suggestion
  printf "10.0.0.r%s.%s" "$(git rev-list --count $_commit)" "$(git rev-parse --short $_commit)"
```

Could we maybe add a "r" while at it?

jeremyd2019:
I'm probably late to the party here, but it just occurred to me that with the 'r' in there, it would be 'safe' to use the version parts from the _mingw_mac.h header.  Those were updated to 11.0.0 in the commit after the one I pinned here (which was tagged as v10.0.0), but 11.0.0.r9999 is still less than 11.0.0.

jeremyd2019:
Or better yet, git describe!

lazka:
Other packages also do "last version + commits", so imo using the next version is confusing.

PR #11393, opened 2022-04-07
msgpack-c: update to 4.0.0

MehdiChinoune:


lazka:
This drops the library. msgpack-c != msgpack-cxx I think

MehdiChinoune:
It looks like, they split the package into c and cxx. Should I Add a new package `msgpack-cxx` for the cxx part.

lazka:
> It looks like, they split the package into c and cxx.

ah, I see.

> Should I Add a new package `msgpack-cxx` for the cxx part.

as you wish.

PR #11403, opened 2022-04-08
mxml: update to 3.3

MehdiChinoune:
On CLANG64:
```
D:/M/msys64/clang64/bin/strip.exe: error: 'clang64/lib/libmxml.a': The end of the file was unexpectedly encountered
```

Issues with stripping?


jeremyd2019:
I've seen that message on other .a files from time to time, that's not something new in LLVM 14.  I never investigated it though.

MehdiChinoune:
> I've seen that message on other .a files from time to time, that's not something new in LLVM 14. I never investigated it though.

I have seen them before, but I thought that they were fixed with LLVM 14

mati865:
It's different issue than the one with DLLs.

PR #11404, opened 2022-04-08
discount: update to 2.2.7b

MehdiChinoune:


mingw-w64-discount/PKGBUILD
```diff
@@ -3,7 +3,7 @@
 _realname=discount
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.2.7
+pkgver=2.2.7b
```
jeremyd2019, line 6: https://github.com/msys2/MINGW-packages/pull/11411#discussion_r846329044

2.2.7b compares less than 2.2.7.  If you really want 2.2.7b, one option would be to write pkgver as 2.2.7.b (and then deal with a `_realver` type var like `_realname`)

PR #11407, opened 2022-04-08
libmaxminddb: update to 1.6.0

MehdiChinoune:


Biswa96:
pkgconfig and man pages are not installed with cmake build. Let me ask that in upstream.

Biswa96:
Would you like to import this commit here for pkgcofig support https://github.com/maxmind/libmaxminddb/commit/0033122b942342d3c76a9c11e96991bc6c469a0b ?

MehdiChinoune:
> Would you like to import this commit here for pkgcofig support [maxmind/libmaxminddb@0033122](https://github.com/maxmind/libmaxminddb/commit/0033122b942342d3c76a9c11e96991bc6c469a0b) ?

That commit should be fixed, the project doesn't use `GNUInstallDirs` which means `${CMAKE_INSTALL_LIBDIR}` is empty, the pkgconfig file end up installed in `/pkgconfig`?

MehdiChinoune:
`GNUInstallDirs` was added after 1.6.0 https://github.com/maxmind/libmaxminddb/commit/4def8ec319630ab4bec076de29f8e29c36f657e8

Biswa96:
Understand. Then it should be imported also, shouldn't it? Instead of adding the patch, the raw patch link can be added in source array.

PR #11410, opened 2022-04-08
libconfig: update to 1.7.3

MehdiChinoune:


mingw-w64-libconfig/001-mingw-fixes.patch
```diff
@@ -12,7 +12,7 @@
  
  #define fileno _fileno
  
--#if _MSC_VER < 1800
+-#if _MSC_VER <= 1800
 +#if defined(_MSC_VER) && _MSC_VER < 1800
```
mmuetzel, line 16: To keep it in sync with the upstream change, should that be:
``` diff
-+#if defined(_MSC_VER) && _MSC_VER < 1800
++#if defined(_MSC_VER) && _MSC_VER <= 1800
```

PR #11413, opened 2022-04-08
octave: update to 7.1.0

mmuetzel:
Remove patches that are no longer needed.


mmuetzel:
We've seen in Octave's CI on msys2/MINGW32 that `configure` sometimes gets stuck indefinitely while exercising the Fortran compiler (`gfortran`). In that case, rebuilding without any changes helps most of the time. (We added a timeout to the configure step to unblock the CI in case that happens.)
I'm guessing that is what is happening with the CI here currently.
This is pretty random and doesn't happen very often. I wasn't able to reproduce it locally. Lacking any steps to reproduce, I didn't report that yet. But it would be nice if that could be avoided somehow...

PR #11415, opened 2022-04-08
mlt: update to 7.6.0

MehdiChinoune:


mingw-w64-mlt/003-fix-detect-cpu-arch.patch
```diff
@@ -0,0 +1,15 @@
...
+@@ -131,10 +131,10 @@
+       list(APPEND MLT_COMPILE_OPTIONS "-mmmx;-msse;-msse2")
+     endif()
+   endif()
+-  if(CMAKE_SYSTEM_PROCESSOR MATCHES "i686" OR (WIN32 AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86"))
++  if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+     set(CPU_X86_32 ON)
+   endif()
+-  if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|AMD64")
++  if(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
Biswa96, line 12: Does not this also mean arm64, ppc64 etc?

MehdiChinoune: Could you stop speculating and look at the source code instead!

jeremyd2019: Let me provide the context:
https://github.com/mltframework/mlt/blob/576b4c7ab210c8edd1cbe04ad5b4d5f0b6c88cb6/CMakeLists.txt#L124-L140

```
# MSVC cl doesn't support GNU inline assembly (but MSVC-compatible clang-cl does)
if(CMAKE_C_COMPILER_ID MATCHES "GNU|Clang")
  if(CMAKE_SYSTEM_PROCESSOR MATCHES "i686|x86|x86_64|AMD64")
    set(CPU_MMX ON)
    set(CPU_SSE ON)
    set(CPU_SSE2 ON)
    if(NOT MSVC) # also NOT clang-cl
      list(APPEND MLT_COMPILE_OPTIONS "-mmmx;-msse;-msse2")
    endif()
  endif()
  if(CMAKE_SYSTEM_PROCESSOR MATCHES "i686" OR (WIN32 AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86"))
    set(CPU_X86_32 ON)
  endif()
  if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|AMD64")
    set(CPU_X86_64 ON)
  endif()
endif()
```

So yes, it does seem that this change is incorrect when considering arm32 and arm64 would also have CMAKE_SIZEOF_VOID_P being 4 or 8 respectively.

What is this patch trying to solve?  Is `CMAKE_SYSTEM_PROCESSOR` something different?

mingw-w64-mlt/PKGBUILD
```diff
@@ -4,58 +4,66 @@
...
-    -DARCH=${MSYSTEM_CARCH}  \
-    -DPython3_EXECUTABLE=${MINGW_PREFIX}/bin/python \
+    ${MINGW_PREFIX}/bin/cmake \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX}  \
+    -G"Ninja"  \
+    -DPython3_EXECUTABLE=${MINGW_PREFIX}/bin/python \
+    -DWINDOWS_DEPLOY=OFF \
+    -DSWIG_PYTHON=ON \
+    -DMOD_RTAUDIO=$([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] && echo "ON" || echo "OFF") \
+    -DMOD_GDK=$([[ ${MINGW_PACKAGE_PREFIX} == "x86_64" ]] && echo "ON" || echo "OFF") \
```
jeremyd2019, line 59: `${MINGW_PACKAGE_PREFIX}` should never `== "x86_64"`

mingw-w64-mlt/003-fix-detect-cpu-arch.patch
jeremyd2019: I see, you've since moved those `CMAKE_SIZEOF_VOID_P` checks up into the previous `if` block checking `CMAKE_SYSTEM_PROCESSOR`.

mingw-w64-mlt/PKGBUILD
MehdiChinoune: My fault, I meant `CARCH`

PR #11418, opened 2022-04-08
quazip: update to 1.3 & split into qt5/qt6

MehdiChinoune:


mingw-w64-quazip/PKGBUILD
```diff
@@ -2,48 +2,78 @@
...
 _realname=quazip
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.9.1
-pkgrel=2
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-qt5"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-qt6")
+pkgver=1.2
+pkgrel=1
 pkgdesc="C++ wrapper for the Gilles Vollant's ZIP/UNZIP C package (mingw-w64)"
```
Biswa96, line 9: This pkgdesc variable may not require now because it is being defined in package function.

MehdiChinoune:
tulip couldn't find quazip and sip, and uses its own. Should I remove them from `depends` and `makedepends`.

PR #11423, opened 2022-04-08
nanodbc: update to 2.14.0

MehdiChinoune:


lazka:
that package hasn't been touched in years, and the removed patch is missing any information why it is needed, so I guess this is fine.

PR #11426, opened 2022-04-09
libimagequant: update to 4.0.0

MehdiChinoune:
Since It builds with rust => No clang32 or clangarm64.


jeremyd2019:
I wish projects would have a little consideration for those of us who can't manage to get rust to work properly, rather than pulling the rug out from under us and the other packages that already took them as a dependency when they could be built.

Are we going to have to go the librsvg route and package an old version for the rust-free prefixes?

MehdiChinoune:
python-pyopenssl 22.0.0, It requires python-cryptography>=36.0 which requires rust to build.
Many other Python packages depend on one of them

jeremyd2019:
Luckily it looks like @mati865 has been busy and might be making progress with rust upstream ❤️

lazka:
Should we wait a bit with this then?

PR #11430, opened 2022-04-09
qt6: update to 6.3.0 - Part1

MehdiChinoune:


lazka:
I've bumped some packages in e9fd3c43a80c6b because of package-grokker.. just in case.

MehdiChinoune:
> I've bumped some packages in [e9fd3c4](https://github.com/msys2/MINGW-packages/commit/e9fd3c43a80c6b1b21fb4d580cfd8b1086174bcc) because of package-grokker.. just in case.

Please don't bump qt6 components like qt6-quick3d, It will be updated later and It will probably fail

lazka:
It will at least block the queue and prevent me from updating the repo.

We/you can always revert if needed.

PR #11434, opened 2022-04-09
lsp-dsp-lib: update to 1.0.1

MehdiChinoune:
I tried to fix it on CLANG64, but it uses `partial linking (-r)` to compile source code files first and then linking them to a shared and static libraries. I could find any equivalent for lld.


lazka:
hm, so this has no header files anymore?

PR #11449, opened 2022-04-10
qt5-static clangarm64

driver1998:
Depends on https://github.com/msys2/MINGW-packages/pull/11450.


MehdiChinoune:
I think you separate them, as each of them takes hours to build, so the jobs will definitely fail.

jeremyd2019:
Also would be a good opportunity to rebase when splitting - xpm-nox has already been enabled for clangarm64 recently.  It's not a conflict, attempting to change it to what it already is on master, just seemed odd in the diff.

driver1998:
OK, split now.

driver1998:
Not sure about what happened to the MINGW failures, the new patch is the same as dynamic qt5.

MehdiChinoune:
> Not sure about what happened to the MINGW failures, the new patch is the same as dynamic qt5.

It requires some (4) patches to make it build with gcc-11 (https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt%405.rb#L72)

driver1998:
Will wait until https://github.com/msys2/MINGW-packages/pull/11469 merges. That upgrades to 5.15.3 and adds those 4 patches for GCC 11.

driver1998:
Should be ready to merge now. Although I know actually building it in autobuild might be challenging...

MehdiChinoune:
> Should be ready to merge now. Although I know actually building it in autobuild might be challenging...

clamarm64 buildbot has no time limit. Because it is a self-hosted one (not hosted by GitHub).

lazka:
@driver1998 do you have actual use for this? It's quite heavy... and will be hard to debug if it fails, so I'm a bit worried tbh.

driver1998:
I am thinking about bringing up 86Box auto build in their repo (they use static Qt on releases), but that would still require a cross toolchain.

driver1998:
Close in favor if #15145.

PR #11450, opened 2022-04-11

driver1998:
For qt5-static and qt6-static.


jeremyd2019:
Successful ARM64 CI run: https://github.com/msys2-arm/MINGW-packages/runs/6072790180?check_suite_focus=true

PR #11456, opened 2022-04-11
python-kiwisolver: update to 1.4.1

MehdiChinoune:
The latest versions of cppy and kiwisolver are 1.2.1 and 1.4.2, but they got built as unknown.


Biswa96:
> but they got built as unknown

Just curious. Would you like to explain that please?

PR #11475, opened 2022-04-12
pyqt6: Update to 6.3.0dev

lazka:


MehdiChinoune:
PyQt6 6.3.0 will be released soon
https://www.riverbankcomputing.com/pipermail/pyqt/2022-April/044552.html

MehdiChinoune:
You could download the tar files from https://www.riverbankcomputing.com/pypi/simple/pyqt6/

lazka:
> You could download the tar files from https://www.riverbankcomputing.com/pypi/simple/pyqt6/

thanks, yes, I used that file for creating the diff. I wasn't sure if that file is going to stay there, that's why.

PR #11478, opened 2022-04-13
libnotify: update to 0.7.9, enable clangarm64

kmilos:


jeremyd2019:
Successful ARM64 CI run: https://github.com/msys2-arm/MINGW-packages/runs/6086661869?check_suite_focus=true

package-grokker showed no removed symbols or dlls.

PR #11480, opened 2022-04-13
libsystre: Copy the dll as libgnurx

orgads:
Useful when cross-compiling with mxe, which links against libgnurx.

### DIFF
--- a/mingw-w64-libsystre/PKGBUILD
+++ b/mingw-w64-libsystre/PKGBUILD
@@ -7,7 +7,7 @@ provides=("${MINGW_PACKAGE_PREFIX}-libgnurx")
 conflicts=("${MINGW_PACKAGE_PREFIX}-libgnurx")
 replaces=("${MINGW_PACKAGE_PREFIX}-libgnurx")
 pkgver=1.0.1
-pkgrel=4
+pkgrel=5
 pkgdesc="Wrapper library around TRE that provides POSIX API (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
@@ -42,4 +42,6 @@ package() {
   cd build-${MINGW_CHOST}
   make DESTDIR="${pkgdir}" install
   install -D -m644 "${srcdir}/systre-${pkgver}/COPYING" "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/LICENSE"
+  cd "${pkgdir}${MINGW_PREFIX}/bin"
+  ln -sf libsystre-0.dll libgnurx-0.dll
 }


mingw-w64-libsystre/PKGBUILD
```diff
@@ -42,4 +42,5 @@ package() {
   cd build-${MINGW_CHOST}
   make DESTDIR="${pkgdir}" install
   install -D -m644 "${srcdir}/systre-${pkgver}/COPYING" "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/LICENSE"
+  cp "${pkgdir}${MINGW_PREFIX}/bin/libsystre-0.dll" "${pkgdir}${MINGW_PREFIX}/bin/libgnurx-0.dll"
```
1480c1, line 45: would a .a copy not be necessary?

orgads: It's already done by make.
```make
install-exec-hook:
        ln -sf libsystre.dll.a $(DESTDIR)$(libdir)/libgnurx.dll.a
        ln -sf libsystre.dll.a $(DESTDIR)$(libdir)/libregex.dll.a
        ln -sf libsystre.a $(DESTDIR)$(libdir)/libgnurx.a
        ln -sf libsystre.a $(DESTDIR)$(libdir)/libregex.a
```

But this gives me the idea to use `ln -sf` instead of copying :)

Done.

Biswa96: Is it required to copy the DLL while the import library already points to that systre DLL?

orgads: Like the commit message says, it is needed when compiling with mxe and running on msys2 env.

Biswa96: Can you share the error message or any output at build time using mingw libsystre package?

orgads: The error is on runtime, not during build. It cannot find the dll.

Biswa96:
> The error is on runtime, not during build. It cannot find the dll.

Wondering how the program links with libgnurx-0.dll. Here is a sample output:

```c
#include <regex.h>

int main() {
    regfree(NULL);
}
```

Compiled with `gcc test.c -lgnurx` and the dependencies:

```
$ ntldd a.exe
        libsystre-0.dll => F:\msys64\ucrt64\bin\libsystre-0.dll (0x000001bfe0650000)
        KERNEL32.dll => C:\Windows\SYSTEM32\KERNEL32.dll (0x000001bfe0930000)
```

orgads:
Because you're compiling in msys2. I'm trying to fix the case when building on mxe.

Biswa96:
I am compiling in mingw environment.

```
$ which gcc
/ucrt64/bin/gcc
```

jeremyd2019:
I think the issue is that some *other* build environment is being used, not mingw or msys2, and that one made the alternate decision to name the dll something else rather than just aliasing via the import library like upstream seems to have intended based on what their makefile already does

orgads:
Ok, there's a way to reproduce it on msys2 environment.

```
$ pacman -S mingw64/mingw-w64-x86_64-gcc mingw64/mingw-w64-x86_64-libgnurx
$ gcc test.c -lgnurx
$ ldd a.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ffb33bd0000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ffb33930000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ffb33660000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7ffb33210000)
        libgnurx-0.dll => /mingw64/bin/libgnurx-0.dll (0x66400000)
$ pacman -S mingw64/mingw-w64-x86_64-libsystre
resolving dependencies...
:: mingw-w64-x86_64-libsystre and mingw-w64-x86_64-libgnurx are in conflict. Remove mingw-w64-x86_64-libgnurx? [Y/n]
...
$ ./a.exe
a.exe: error while loading shared libraries: libgnurx-0.dll: cannot open shared object file: No such file or directory
```

Biswa96:
That's a different package https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-libgnurx and libsystre conflicts with libgnurx package. Also libsystre replaces libgnurx.

orgads:
That's exactly my point. The API and behavior are the same, these packages are interchangable. So my change enables running a package that was built with libgnurx with libsystre.

jeremyd2019:
In order to be able to swap out libraries like this at runtime rather than at build time, they would have to be ABI compatible not just API compatible, and that can be a hard thing to guarantee (even a new version of the same sources frequently break ABI compatibility, it would be shocking if two different codebases had some sort of agreement between them to not break ABI compatibility with each other)

orgads:
If I understand correctly, the entire purpose of `libsystre` is to wrap `libtre` and provide a replacement for `libgnurx`. I actually can't find it anywhere except mingw. Introduced in 008a5b08fe32c30df72a60efe2645939a1e37e3d with a tarball.

I've been using it for at least a year as a drop-in replacement for `libgnurx` (linking with `libgnurx` on mxe, and running on mingw), and it works just fine. It has API and ABI compatibility, and this is extremely unlikely to change. If `libgnurx` will change its ABI, then `libsystre` will need to catch up.

jeremyd2019:
Quick look for ABI incompatibility:
libgnurx:
https://github.com/coapp-packages/libgnurx/blob/19dcad9f1037a81a1766efd24e58cbf7d41be1c3/regex.h#L343-L407

libtre:
https://github.com/laurikari/tre/blob/6092368aabdd0dbb0fbceb2766a37b98e0ff6911/include/tre/tre.h#L67-L70

The caller of regcomp has to provide a pointer to a regex_t, which means they need to have allocated it, knowing how big it was supposed to be.  As you can see, they are vastly different sizes between the implementations.  Just because libgnurx's struct happens bigger, swapping in libsystre at runtime doesn't crash immediately on calling regcomp.  Swapping the other way would though.

orgads:
You should compare with libsystre, not libtre.

jeremyd2019:
I did, libsystre in includes `<tre/tre.h>` for its definition of regex_t

orgads:
Swapping the other way won't work anyway, because of the additional libtre dependency. Notice that I didn't change the dll name, only created a link that allows one-way substitution.

orgads:
Never mind. It won't work if the caller actually reads the internal members (not sure if this is a typical usage, but ok).

Issue #4773, opened 2018-12-13
Boost 1.69-1 -x32/-x64 library file name suffix


stambaughw:
I just installed version 1.69.0-1 of boost which broke building KiCad.  After a bit of digging I found that all of the boost libraries have either -x32 or -x64 appended to the filename which breaks CMake's FindBoost macro.  I'm guessing this would also break any autotools find boost macro as well.  What is the purpose of the -x32/-x64 suffix?  Version 1.68.0-2 of boost did not have this suffix.

oscarfv:
This breaks build scripts in general, but nothing changed in the PKGBUILD to cause this problem, so it comes from upstream.

Also, the new version adds more than 30 MB to the unpacked size. This comes from upstream too.

josch:
I have the same problem. If I read https://github.com/Kitware/CMake/commits/master/Modules/FindBoost.cmake right, then support for boost 1.69.0 was only added in cmake 3.13.

Alexpux:
@josch  @oscarfv I think we need revert changes in boost to generate old style names for libraries as it will break many packages. igging into boost commits....

oscarfv:
I looked at the boost build system and -x32/-x64 are present in library names since 1.66, at least for VC++ builds (search "-x32" in https://www.boost.org/doc/libs/1_66_0/boost/config/auto_link.hpp) Somehow it didn't apply to mingw until now.

Alexpux:
@oscarfv I think I found where irt changed. Rebuilding now boost to see if I fix it. We build tagged layout, not versioned this is why we dont have architecture in names before 1.69. Now it changed for tagged layout too

Alexpux:
Library names will be fixed in 1.69.0-2

stambaughw:
Thanks for the quick fix!

PR #11481, opened 2022-04-13
boost: Replace name-preserving patch with a proper flag

orgads:
This is an upstream convention for quite long.

Reverts fff2fc0d53aa95b85cee0c785e56159b0565ea72.


orgads:
@Alexpux @lazka

orgads:
I upgraded boost on mxe (with the new names) [3 months ago](https://github.com/mxe/mxe/pull/2698), and nobody complained so far :)

orgads:
hmmm... Now I see that the linked `a` files have the suffix, but the actual dll doesn't. Please wait before approving...

lazka:
How will things find those libraries at build time if their name changes?

orgads:
I'd expect applications that build for Windows to have a proper lookup. How do they work with default upstream boost build?

Anyway, I removed the patch, but found a better way to preserve the old names, by passing the layout version to b2.

orgads:
Actually I think we can consider using `--layout=system`, which is the default on linux. The Windows default (versioned/tagged) seems to target MSVC, when all the DLLs are installed in a single directory. This is not the case for MSYS2/MinGW.

What do you say?

lazka:
Given issues like #4773 and boost not advertising the naming I'd be cautious at least. I'm afraid I don't know enough about this to give any input.

Your current change, replacing the patch with the layout flag, looks good to me.

orgads:
Rebased

lazka:
> Actually I think we can consider using `--layout=system`, which is the default on linux. The Windows default (versioned/tagged) seems to target MSVC, when all the DLLs are installed in a single directory. This is not the case for MSYS2/MinGW.

(In theory this sounds good to me)

orgads:
> > Actually I think we can consider using `--layout=system`, which is the default on linux. The Windows default (versioned/tagged) seems to target MSVC, when all the DLLs are installed in a single directory. This is not the case for MSYS2/MinGW.
> 
> (In theory this sounds good to me)

But then again, we will probably face build issues with dependent applications.

Notice that most applications that use boost, use it as header-only. The runtime libraries are only needed for a small subset of boost modules.

PR #11483, opened 2022-04-13
Add Duktape

That-Human-Being:
I was in need of this [tool](duktape.org) for a project and found it was not in the MSYS2 repos. I started with the [Arch Linux](https://archlinux.org/packages/community/x86_64/duktape/) package as a base, then went on to make sure it built on the different MSYSTEMs (with the exception of CLANGARM64, I couldn't get it to work there). I am not able to maintain this package, but want it to be available for anyone else needing to use Duktape.

### DIFF
--- a/mingw-w64-duktape/PKGBUILD
+++ b/mingw-w64-duktape/PKGBUILD
@@ -1,0 +1,50 @@
+_realname=duktape
+pkgbase=mingw-w64-${_pkgname}
+pkgname="${MINGW_PACKAGE_PREFIX}-duktape"
+pkgver=2.7.0
+_dirname=${_realname}-${pkgver}
+pkgrel=1
+url="http://duktape.org/"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+pkgdesc="Embeddable Javascript engine"
+license=("MIT")
+depends=("${MINGW_PACKAGE_PREFIX}-crt-git")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc")
+options=(staticlibs !strip !buildflags)
+source=(
+	"http://duktape.org/duktape-${pkgver}.tar.xz"
+	"duktape.pc"
+	)
+sha256sums=('90f8d2fa8b5567c6899830ddef2c03f3c27960b11aca222fa17aa7ac613c2890'
+            '5a888e877d3ae2c752aba75a88587c22a4ef96b428a53f7b9b435799c758a55a')
+
+_architectures="${MINGW_PACKAGE_PREFIX}"
+
+build() {
+  for _arch in ${_architectures}; do
+    mkdir -p "${srcdir}/${pkgname}-${pkgver}-build-${_arch}"
+    pushd "${srcdir}/${_dirname}" > /dev/null
+    gcc src/duktape.c -std=c99 -O2 -c -o duktape.o
+    ar rcs libduktape.a duktape.o
+    pushd "../${pkgname}-${pkgver}-build-${_arch}" > /dev/null
+    mv "${srcdir}/${_dirname}/libduktape.a" "libduktape.a"
+    rm "${srcdir}/${_dirname}/duktape.o"
+    popd > /dev/null
+    popd > /dev/null
+  done
+}
+
+package() {
+  for _arch in ${_architectures}; do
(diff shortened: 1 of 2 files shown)


mingw-w64-duktape/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+sha256sums=('90f8d2fa8b5567c6899830ddef2c03f3c27960b11aca222fa17aa7ac613c2890'
+            '5a888e877d3ae2c752aba75a88587c22a4ef96b428a53f7b9b435799c758a55a')
+
+_architectures="${MINGW_PACKAGE_PREFIX}"
+
+build() {
+  for _arch in ${_architectures}; do
+    mkdir -p "${srcdir}/${pkgname}-${pkgver}-build-${_arch}"
+    pushd "${srcdir}/${_dirname}" > /dev/null
+    gcc src/duktape.c -std=c99 -O2 -c -o duktape.o
```
1480c1, line 28:
```suggestion
    cc src/duktape.c -std=c99 -O2 -c -o duktape.o
```

pref usage of `cc` over `gcc` or `clang` or alternatively,

```bash
    ${CC:-cc} src/duktape.c -std=c99 -O2 -c -o duktape.o
```

mingw-w64-duktape/duktape.pc
```diff
@@ -0,0 +1,10 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+Name: duktape
+Description: Embeddable Javascript engine
+Version: 2.7.0
+Libs: -L${libdir} -lduktape
+Cflags: -I${includedir}
```
1480c1, line 10: newline at end of file

mingw-w64-duktape/duktape.pc
```diff
@@ -0,0 +1,10 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+Name: duktape
```
1480c1, line 6: Was going to comment on the addition of what seems like an unofficial pkg-config file as it's not present in their 2.6 branch (is there in their 3 branch), but it seems this is semi-standard

https://github.com/archlinux/svntogit-community/blob/packages/duktape/trunk/duktape.pc
https://salsa.debian.org/debian-iot-team/duktape/-/blob/master/duktape.pc.in
https://alpine.pkgs.org/3.15/alpine-community-x86_64/duktape-dev-2.6.0-r0.apk.html
https://centos.pkgs.org/8/epel-x86_64/duktape-devel-2.2.0-5.el8.x86_64.rpm.html

and so on

mingw-w64-duktape/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+	"http://duktape.org/duktape-${pkgver}.tar.xz"
+	"duktape.pc"
+	)
+sha256sums=('90f8d2fa8b5567c6899830ddef2c03f3c27960b11aca222fa17aa7ac613c2890'
+            '5a888e877d3ae2c752aba75a88587c22a4ef96b428a53f7b9b435799c758a55a')
+
+_architectures="${MINGW_PACKAGE_PREFIX}"
+
+build() {
+  for _arch in ${_architectures}; do
```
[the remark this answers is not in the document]
That-Human-Being: I don't know, that came from the Arch Linux package

mingw-w64-duktape/PKGBUILD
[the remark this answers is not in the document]
That-Human-Being: I didn't use them because `make` kept giving me errors but this worked

That-Human-Being: Ok, I really only made this for MinGW64 and then just tested to see if it built for the other environments

mingw-w64-duktape/duktape.pc
That-Human-Being: That was quite literally taken straight from Arch's repo as I couldn't get it to build otherwise

Biswa96:
There are some difference between ArchLinux and msys2's mingw packaging style. If you don't mind can I rewrite this in another pull request?

That-Human-Being:
Yeah sure, I just threw this together to meet my needs. If you want to make it to better fit Msys2s style that's fine

PR #11487, opened 2022-04-13
podofo: update to 0.9.7

c72578:
- Update `fix-export.patch`
- Remove `remove-broken-command.patch`
  Not needed anymore, fixed upstream:
  https://sourceforge.net/p/podofo/code/1942/
- Add patch: `003-podofo-0.9.7-localtime_r.patch`
  Fixes:
```console
    src/podofo-0.9.7/test/unit/DateTest.cpp:143:5: error: 'localtime_r'
    was not declared in this scope; did you mean 'localtime_s'?
```


c72578:
@lazka Could you please check, if `003-podofo-0.9.7-localtime_r.patch` is the recommended way to fix the `localtime_r` error?
An alternative suggestion [1] would be to:
`#define _POSIX_THREAD_SAFE_FUNCTIONS`

Which also leads to `localtime_s` [2] ...
```c
#ifdef _POSIX_THREAD_SAFE_FUNCTIONS
__forceinline struct tm *__CRTDECL localtime_r(const time_t *_Time, struct tm *_Tm) {
  return localtime_s(_Tm, _Time) ? NULL : _Tm;
}
```

[1] https://stackoverflow.com/questions/18551409/localtime-r-support-on-mingw
[2] https://github.com/mirror/mingw-w64/blob/2cc74a530ed8b11fd10cf7ecb5216c6950a88760/mingw-w64-headers/crt/time.h#L300

lazka:
lgtm, though you could just use `#ifdef _WIN32`

c72578:
@lazka Thanks for the feedback.
I have updated `003-podofo-0.9.7-localtime_r.patch` to just use `#ifdef _WIN32`

PR #11488, opened 2022-04-14
add new: FrontISTR

3rav:


mmuetzel:
That should probably be a `PKGBUILD` script inside a folder named `mingw-w64-frontistr`...

PR #11494, opened 2022-04-14
cmake: update to 3.23.1

MehdiChinoune:


mingw-w64-cmake/PKGBUILD
```diff
@@ -45,15 +45,17 @@ source=("https://github.com/Kitware/CMake/releases/download/v${pkgver}/${_realna
         "0002-Do-not-install-Qt-bundle-in-cmake-gui.patch"
         "0004-Output-line-numbers-in-callstacks.patch"
         "0005-Default-to-ninja-generator.patch"
-        "0006-Fix-find-wxwidgets.patch"
-        "https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7105.patch")
-sha256sums=('5ab0a12f702f44013be7e19534cd9094d65cc9fe7b2cd0f8c9e5318e0fe4ac82'
+        "https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7105.patch"
+        "0006-fix-find-glut.patch"
```
1480c1, line 49: For this one, why not just use the .patch url as well?

MehdiChinoune: It couldn't be applied properly! (file changed before my patch)

MehdiChinoune: BTW, Isn't time to remove mask from your avatar!
Just joking!

1480c1: > BTW, Isn't time to remove mask from your avatar! Just joking!

too lazy to change it everywhere else I have it set, plus I'm not sure if I even have the original files without the mask

1480c1: > It couldn't be applied properly! (file changed before my patch)

Ah, I see, makes sense then

PR #11496, opened 2022-04-14
[WIP] wxPython4.0: Add package

stahta01:
This is a work in progress; and is not ready for merging.

I still need to build KICAD to see if it works.

Tim S.


stahta01:
Decided that reverting wxPython back to 4.0.7 was a bad idea. wxWidgets 3.1 has a lot of high DPI improvements that would be lost in KiCAD and I think removing wxPython support till wxPython works with wxWidgets 3.1.6 would be better.

jeremyd2019:
It still might not be a bad idea to package wxPython4.0, if wxWidgets 3.1/wxPython 4.1 are going to be at odds from time to time a 'stable' version might be a good idea.

PR #11505, opened 2022-04-16
boost: update to 1.79.0

MehdiChinoune:


MehdiChinoune:
@jeremyd2019 Did package-grokker list all the dependent libraries or just the ones using the deleted symbols?

jeremyd2019:
On Sat, 16 Apr 2022, مهدي شينون (Mehdi Chinoune) wrote:

> @jeremyd2019 Did package-grokker list all the dependent libraries or
> just the ones using the deleted symbols?

It will list all exe/dll/pyd that attempt to import deleted symbols, or
that attempt to import a deleted exe/dll/pyd.

PR #11512, opened 2022-04-18
kicad: Update to 5.1.12

stahta01:
Finished the change to using opencascade like KiCAD 6.0.4 does.
The clang64 version is missing some features because opencascade does not build for clang64.

I decided to revert the wxWidgets version back to 3.0.x; This required the creation of an wxPython4.0 package.

Tim S.


stahta01:
I have added an untested option KICAD_WIN32_DPI_AWARE to the cmake file; it should work; but, since I have no high DPI monitor I have no way to tell if it works. I will test the packages made without this option set to see if the little I know about kicad work. Then, I will verify that the KICAD_WIN32_DPI_AWARE builds.

Tim S.

stahta01:
This is ready to merge; I think I have it worker better than 5.1.10 did.
I also, got clang64 to build.

Tim S.

stahta01:
This package is ready to merge

stahta01:
The clang64 failed because of a download error; this package builds on my local PC.

Tim S.

stahta01:
Rebase with newer master to show that clang64 builds.

PR #11516, opened 2022-04-18
lapack: update to 3.10.1

mmuetzel:
Also correct the license in the package description.

The checks for CBLAS are failing. This is not a regression from this update but already happened (at least) with 3.10.0.
The other tests ("proper" BLAS, LAPACK) are passing.

Tbh, I don't understand the error messages (something about a missing file).

Is it ok to leave the checks in but ignore its result for the time being?


Biswa96:
In my setup, 97% tests passed, 4 tests failed out of 121 tests for static build. The result is way more than better for any mingw packages 😄

PR #11519, opened 2022-04-19
Fix for missing -l options in grpc libraries

KartikShrivastava:
This is a small fix to properly link `advapi32`, `dbghelp` and `bcrypt` libraries used in grpc package. Currently grpc generates these libraries within quotes `""` rather than a `-l` option. Which makes linking to these libraries errorneous.


Biswa96:
Thank you for the contribution. I think this should be changed at the time of generation of pkgconfig files. Let us check the underlying issue first.

KartikShrivastava:
Hi @Biswa96, thanks for your response.

I tried to look for underlying issue. And I think the issue is probably with this `$<JOIN:${ABSL_CC_LIB_LINKOPTS}, >` here [grpc-1.45.2/third_party/abseil-cpp/CMAKE/AbseilHelpers.cmake](https://github.com/abseil/abseil-cpp/blob/215105818dfde3174fe799600bb0f3cae233d0bf/CMake/AbseilHelpers.cmake#L182).

And the affected pkgconfig files are the only ones which adds libraries using `${ABSL_CC_LIB_LINKOPTS}` that too within quotes `""` as opposed to followed by `-l` option. Just for records the effected files are `absl_base.pc`, `absl_symbolize.pc` and recently added `absl_random_internal_seed_material.pc` with `"advapi32"`, `"dbghelp"` and `"bcrypt"` libraries respectively.

MehdiChinoune:
abseil-cpp was un-bundled from grpc, so this PR couldn't be applied

PR #11521, opened 2022-04-19
x264: Add Cflags.private in pkgconfig file

Biswa96:
Please review. @1480c1

### DIFF
--- a/mingw-w64-x264/0003-pkgconfig-add-Cflags-private.patch
+++ b/mingw-w64-x264/0003-pkgconfig-add-Cflags-private.patch
@@ -1,0 +1,10 @@
+--- a/configure
++++ b/configure
+@@ -1611,6 +1611,7 @@ Version: $(grep POINTVER < x264_config.h | sed -e 's/.* "//; s/".*//; s/ .*//')
+ Libs: -L\${libdir} -lx264 $([ "$shared" = "yes" ] || echo $libpthread $libm $libdl)
+ Libs.private: $([ "$shared" = "yes" ] && echo $libpthread $libm $libdl)
+ Cflags: -I\${includedir} $([ "$shared" = "yes" ] && echo "-DX264_API_IMPORTS")
++Cflags.private: -UX264_API_IMPORTS
+ EOF
+ 
+ filters="crop select_every"
--- a/mingw-w64-x264/PKGBUILD
+++ b/mingw-w64-x264/PKGBUILD
@@ -7,7 +7,7 @@ provides=("${MINGW_PACKAGE_PREFIX}-${_realname}-git")
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}-git")
 replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}-git")
 pkgver=0.161.r3039.544c61f
-pkgrel=3
+pkgrel=4
 pkgdesc="Library for encoding H264/AVC video streams (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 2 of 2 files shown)


1480c1:
Although this seems like it would work around the auto inclusion of `-DX264_API_IMPORTS` in cflags regardless of `--static` from pkgconf, it doesn't seem like `X264_STATIC` is anywhere but here for now.

You might want to add another patch for the configure file to add a `Cflags.private: -DX264_STATIC` line to the pkgconf file along side your current one

Biswa96:
> it doesn't seem like X264_STATIC is anywhere but here for now

I asked this in upstream but they suggest to use `X264_API_IMPORTS` and `X264_API_EXPORTS`. Wouldn't it be easier to define only `X264_STATIC` instead of undefining those two?

1480c1:
Yeah, that's why I suggested adding it to the pkg-config file

1480c1:
lgtm I think, I am wondering if it might be good to also add `$([ "$shared" = "yes" ] && echo "-UX264_API_IMPORTS")` to the .private line as well to make sure it's not defined in the end?

Biswa96:
Here is my logic. If `X264_STATIC` is defined, `X264_DLL_IMPORT` is set to nothing. Then if `X264_API_IMPORTS` is defined `X264_API` will be set to nothing. So, `X264_API_IMPORTS` does not do anything. Right?

1480c1:
Yeah, that's sound logic, just wondered if it would be worth adding an undef just in case there happens to be something else, such as user's code, decided to use the macro to check if it's a dll or something

Biswa96:
Then `X264_STATIC` is not required anymore. The following would also work, doesn't it?

```
Cflags.private: -UX264_API_IMPORTS
```

1480c1:
> Then `X264_STATIC` is not required anymore. The following would also work, doesn't it?
> 
> ```
> Cflags.private: -UX264_API_IMPORTS
> ```

Yeah, that seems like it would work as well

Biswa96:
> I doubt that -U in Cflags.private will work

Would you like to explain that please?

msink:
Well - `Cflags.private` goes to compiler command line, and command line handled **before** .h files, didnt it?
So `-UX264_API_IMPORTS` in command line doed nothing as `X264_API_IMPORTS` is already not defined

msink:
maybe this is even better a little:
```diff
--- a/configure
+++ b/configure
@@ -1636,6 +1636,7 @@ Version: $(grep POINTVER < x264_config.h | sed -e 's/.* "//; s/".*//; s/ .*//')
 Libs: -L$libdir -lx264 $([ "$shared" = "yes" ] || echo $libpthread $libm $libdl)
 Libs.private: $([ "$shared" = "yes" ] && echo $libpthread $libm $libdl)
 Cflags: -I$includedir $([ "$shared" = "yes" ] && echo "-DX264_API_IMPORTS")
+Cflags.private: -DX264_API
 EOF
 
 filters="crop select_every"
--- a/x264.h
+++ b/x264.h
@@ -65,6 +65,7 @@ extern "C" {
  * will need to define X264_API_IMPORTS before including this header.
  * This clause does not apply to MinGW, similar development environments, or non
  * Windows platforms. */
+#ifndef X264_API
 #ifdef X264_API_IMPORTS
 #   define X264_API X264_DLL_IMPORT
 #else
@@ -74,6 +75,7 @@ extern "C" {
 #       define X264_API
 #   endif
 #endif
+#endif
 
 /* x264_t:
  *      opaque handler for encoder */
```

msink:
Well, seems that I totally messed up `X264_DLL_IMPORT` with `X264_API_IMPORTS` ... sorry.

Biswa96:
With `Cflags.private: -UX264_API_IMPORTS` the pkgconfig output is like this:

```
$ pkgconf -cflags -static x264
-IF:/msys64/ucrt64/include -DX264_API_IMPORTS -UX264_API_IMPORTS
```

I have compiled ffmpeg with that change and the libavcodec.a static lib does not have __imp_x264* symbols. I shall try to ask this in upstream first.

Biswa96:
"There is no such thing as "Cflags.private" in standard pkg-config" 🤦‍♂️ 

https://mailman.videolan.org/pipermail/x264-devel/2022-April/012940.html

msink:
Well, something weird goes here.
Problem in #11538 is just opposite - `libavformat.a` references functions with ``__imp_` prefix, while functions in `libx264.a` do not have prefixes.
Maybe life will be easier it add prefix to `libx264.a` functions?

1480c1:
Yeah, those functions with the `__imp_` prefixes exist because `X264_API_IMPORTS` is set, libx264 doesn't have them because they weren't built with any dllimport or export

1480c1 (approved):

PR #11529, opened 2022-04-20
graphicsmagick: increase to 16b quantum depth

kmilos:
Q8 is really not that useful - ImageMagick has been at Q16 by default for a while, and most distros have been packaging GM as Q16 for a while too...

@mmuetzel @podsvirov Please check if any action is needed for the dependencies?


mmuetzel:
Afaict, Octave doesn't use any of the symbols that changed signature. It's probably not necessary to rebuild it.

Biswa96:
The CI shows to rebuild these two packages.

```
mingw-w64-pdf2djvu
mingw-w64-octave
```

kmilos:
To be honest, I also couldn't see what would change in the ABI from just looking at the GM source...

mmuetzel:
> To be honest, I also couldn't see what would change in the ABI from just looking at the GM source...

The package grokker found some removed symbols:
https://github.com/msys2/MINGW-packages/runs/6095169357?check_suite_focus=true#step:3:77

All of those have at least one (8-bit) `unsigned char` argument (`h` in the mangled name). That is probably replaced by some 16-bit integer type with the changes here.
Octave doesn't use any of those symbols afaict. But it wouldn't hurt to bump its pkgrel anyway...

lazka:
octave build is hanging for mingw32 in CI. I might have to revert this temporarily.

mmuetzel:
That is happening occasionally also in the CI of Octave with the MINGW32 environment of MSYS2. If it is occurring, it is always hanging at one of the tests for a working Fortran compiler. 
There is nothing special about these tests afaict. It's only happening for MINGW32 and only occasionally. I haven't actually counted, but my feeling is that it hangs in about 1 out of 10 build attempts. It never hangs when it builds the "actual" sources. I've only ever seen it hanging during `configure` at one of those tests.
I've not yet been able to reproduce that hang locally. (But maybe it would have been possible to reproduce if I had tried harder.) I've only ever seen it on the CI with GitHub hosted runners.
Maybe, it is completely random. Or maybe it depends on some hardware detail of the selected runner. Or maybe it is something else...
In any case, most of the time just re-triggering the build without any changes "fixes" it.

lazka:
Ah, thanks. I now remember you mentioning it somewhere already. Let's re-try a few times then :)

mmuetzel:
Is it possible to have a build step time out earlier to unblock the system in case this is happening?
Maybe something along the following lines in the PKGBUILD script for Octave:
```
timeout 30m ./configure [...]
```

The subsequent `make` step would fail (with the missing `Makefile`) without any further logic.
Is there some logic that would lead to a rebuild attempt of the package in the next round?

lazka:
> Is there some logic that would lead to a rebuild attempt of the package in the next round?

Not automatically, I can trigger a rebuild via a GitHub actions.

I didn't manage to get it working in CI, so I built it locally this time.

PR #11533, opened 2022-04-20
armadillo: update to 11.0.1

mmuetzel:
Build with SuperLU.
Correct license in description.
Use LAPACK in OpenBLAS also for CLANG* environments.


mmuetzel:
The patch was accepted upstream. I added a comment with a link to the commit.

PR #11534, opened 2022-04-20
ttf-paratype: New package

podsvirov:


Biswa96:
Are these fonts free (freedom) and open source?

podsvirov:
Yes. I found similar packages for [ArchLinux](https://aur.archlinux.org/packages/ttf-paratype) and [Debian](https://packages.debian.org/unstable/fonts/fonts-paratype).

PR #11537, opened 2022-04-21
octave: Add timeout to configure step.

mmuetzel:
This is only to unblock the build system in a case something similar to what happened in #11529 should occur again in the future. Nothing changed in the build rules that would actually affect the resulting binaries. No need to bump the pkgrel for this probably.


mmuetzel:
It looks like MINGW32 got stuck again. Let's see if the timeout actually works.

The build seems to hang more often here than in Octave's CI. I wonder if that is just coincidence or if there is a reason for that...

lazka:
Thanks

PR #11542, opened 2022-04-22
gcc: Update to 11.3.0

lazka:
(ignore the package-grokker error)


lazka:
@jeremyd2019 does the second commit make sense? It seems to work here, but not sure I considered all things.

jeremyd2019:
> @jeremyd2019 does the second commit make sense? It seems to work here, but not sure I considered all things.

Seems to make sense.

Issue #10269, opened 2021-12-05
openmp static link library does not exist.


mizar:
`mingw-w64-clang-w86_64-openmp`, `mingw-w64-x86_64-openmp`, ... static link library does not exist.

```
$ pacman -Ql mingw-w64-clang-x86_64-openmp
mingw-w64-clang-x86_64-openmp /clang64/
mingw-w64-clang-x86_64-openmp /clang64/bin/
mingw-w64-clang-x86_64-openmp /clang64/bin/libomp.dll
mingw-w64-clang-x86_64-openmp /clang64/include/
mingw-w64-clang-x86_64-openmp /clang64/include/omp.h
mingw-w64-clang-x86_64-openmp /clang64/lib/
mingw-w64-clang-x86_64-openmp /clang64/lib/libomp.dll.a
mingw-w64-clang-x86_64-openmp /clang64/share/
mingw-w64-clang-x86_64-openmp /clang64/share/licenses/
mingw-w64-clang-x86_64-openmp /clang64/share/licenses/openmp/
mingw-w64-clang-x86_64-openmp /clang64/share/licenses/openmp/LICENSE.TXT
```

```
$ pacman -Ql mingw-w64-x86_64-openmp
mingw-w64-x86_64-openmp /mingw64/
mingw-w64-x86_64-openmp /mingw64/bin/
mingw-w64-x86_64-openmp /mingw64/bin/libomp.dll
mingw-w64-x86_64-openmp /mingw64/include/
mingw-w64-x86_64-openmp /mingw64/include/omp.h
mingw-w64-x86_64-openmp /mingw64/include/omp_lib.h
mingw-w64-x86_64-openmp /mingw64/include/omp_lib.mod
mingw-w64-x86_64-openmp /mingw64/include/omp_lib_kinds.mod
mingw-w64-x86_64-openmp /mingw64/lib/
mingw-w64-x86_64-openmp /mingw64/lib/libomp.dll.a
mingw-w64-x86_64-openmp /mingw64/share/
mingw-w64-x86_64-openmp /mingw64/share/licenses/
mingw-w64-x86_64-openmp /mingw64/share/licenses/openmp/
mingw-w64-x86_64-openmp /mingw64/share/licenses/openmp/LICENSE.TXT
```

test code:
```bash
LANG=C bash <<EOS
echo ":: MSYSTEM=$MSYSTEM";
if [ "$MSYSTEM" = "MINGW64" ]; then
[...]

PR #11546, opened 2022-04-22
openmp: add hacks for compiling both shared and static builds

1480c1:
Fixes https://github.com/msys2/MINGW-packages/issues/10269

The first hunk could probably be upstreamed? The main usage for it is if CPPFLAGS is set with spaces inside, e.g. `-DFLAG=1 -DOTHER_FLAG=2` as its current line would result in `-DFLAG="1 -DOTHER_FLAG=2"`

For the second and third, I'm still not sure why static openmp is disabled on Windows, so I have no clue if they can be upstreamed or not

### DIFF
--- a/mingw-w64-openmp/PKGBUILD
+++ b/mingw-w64-openmp/PKGBUILD
@@ -8,7 +8,7 @@ _realname=openmp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=14.0.0
-pkgrel=1
+pkgrel=2
 pkgdesc="LLVM OpenMP Library (mingw-w64)"
 url="https://openmp.llvm.org/"
 arch=(any)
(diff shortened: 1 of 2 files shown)


mingw-w64-openmp/PKGBUILD
```diff
@@ -72,22 +74,27 @@ build() {
...
+      -B "build-${MSYSTEM}-${shared_libs}"
+    ${MINGW_PREFIX}/bin/cmake --build "build-${MSYSTEM}-${shared_libs}"
+  done
+  unset shared_libs
 }
 
 package() {
-  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --install build-${MSYSTEM}
+  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --install "${srcdir}/build-${MSYSTEM}-shared"
+  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --install "${srcdir}/build-${MSYSTEM}-static"
```
MehdiChinoune, line 97: Shared should be installed last

PR #11547, opened 2022-04-22
mingw-w64-tools-git: patch for mingw-enabled bin_to_includedir

heljkon:
qemu uses widl via meson with expectation that widl knows its resource directories.

qemu references oaidl.idl, ocidl.idl and stdole2.tlb which reside in $MINGW_PACKAGE_PREFIX/include. $MINGW_PACKAGE_PREFIX/include is neither include nor lib search directory for widl

Because stdole2.tlb needs to be found as well --with-widl-includedir is not sufficient. 
There seems to be no way to transform both parameters -I $MINGW_PACKAGE_PREFIX/include and -L $MINGW_PACKAGE_PREFIX/include to configure options.

This results in the applied patch - which I produced without any recent C/C++ experience. Please keep this in mind while reviewing!

The idea of the patch is: add a well-known directory to the search path, but don't remove any existing (maybe broken) directories from the search path.

### DIFF
--- a/mingw-w64-tools-git/PKGBUILD
+++ b/mingw-w64-tools-git/PKGBUILD
@@ -7,7 +7,7 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=10.0.0.r0.gaa08f56da
-pkgrel=1
+pkgrel=2
 _commit='aa08f56da559016f10336dddca85d59f9bdc9e02'
 pkgdesc="MinGW-w64 tools"
 arch=('any')
@@ -18,8 +18,14 @@ groups=("${MINGW_PACKAGE_PREFIX}-toolchain")
 makedepends=("git" "${MINGW_PACKAGE_PREFIX}-cc" "${MINGW_PACKAGE_PREFIX}-libmangle" "${MINGW_PACKAGE_PREFIX}-autotools")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
 options=('staticlibs' '!emptydirs')
-source=("mingw-w64"::"git+https://git.code.sf.net/p/mingw-w64/mingw-w64#commit=$_commit")
-sha256sums=('SKIP')
+source=(
+  "mingw-w64"::"git+https://git.code.sf.net/p/mingw-w64/mingw-w64#commit=$_commit"
+  bin_to_includedir.patch
+)
+sha256sums=(
+  'SKIP'
+  '07c772f93adfe783c7b3fb9a1490f12f585d29d4a97c680d476c74929af20506'
+)
 _tools="gendef genlib genidl genpeimg widl" # genstubdll
 
 pkgver() {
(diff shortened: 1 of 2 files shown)


lazka:
oh, do we maybe need to set `with-widl-includedir`? It currently defaults to `/ucrt64/x86_64-w64-mingw32/include` ...

heljkon:
> oh, do we maybe need to set `with-widl-includedir` when building tools-git? It currently defaults to `/ucrt64/x86_64-w64-mingw32/include` ...

As I stated in PR-Message:

> Because stdole2.tlb needs to be found as well --with-widl-includedir is not sufficient as configure parameter

Using --with-widl-includedir doesn't solve the problems

lazka:
> Using --with-widl-includedir doesn't solve the problems

I'm afraid I don't know enough about this, and you'd better ask mingw-w64 devs for what should be done.

From what I see `--with-widl-includedir` results in the same calls you added with your patch.

heljkon:
> I'm afraid I don't know enough about this, and you'd better ask mingw-w64 devs for what should be done.

OK, I understand my patch seems to make changes too severe...

heljkon:
Hi @kostyanf14  

> So I think we should patch the `configure` file.

Patching configure does not help!

Without --with-widl-includedir in configure make calls:
`gcc ... -DINCLUDEDIR=\""/mingw64/x86_64-w64-mingw32/include"\" -DBIN_TO_INCLUDEDIR=\""../x86_64-w64-mingw32/include"\" -DBIN_TO_DLLDIR=\""../x86_64-w64-mingw32/include"\" -DDLLDIR="\"/mingw64/lib\"" ...`

Result:
```
$ widl -t qga-vss.idl -o qga-vss.tlb
qga-vss.idl:1: error: Unable to open include file oaidl.idl
```

Using --with-widl-includedir="/mingw64/include" in configure results in make calling:
`gcc ... -DINCLUDEDIR=\""/mingw64/include"\" -DBIN_TO_INCLUDEDIR=\""../include"\" -DBIN_TO_DLLDIR=\""../include"\" -DDLLDIR="\"/mingw64/lib\"" ...
`
Now BIN_TO_INCLUDEDIR is set in a way widl.c compiled on Windows can make use of it in line 742:
` wpp_add_include_path(strmake("%s%s/%s", sysroot, exe_path, BIN_TO_INCLUDEDIR));`

But there is no equivalent to make use of BIN_TO_DLLDIR in widl.c compiled on Windows.
 
Result: 
```
$ widl -t qga-vss.idl -o qga-vss.tlb
error: cannot find stdole2.tlb
```

After additionally replacing -DDLLDIR="\"/mingw64/lib\"" with -DDLLDIR=\""/mingw64/include""\ in Makefile.in and Makefile.am, make calls:
`gcc ... -DINCLUDEDIR=\""/mingw64/include"\" -DBIN_TO_INCLUDEDIR=\""../include"\" -DBIN_TO_DLLDIR=\""../include"\" -DDLLDIR=\""/mingw64/include"\" ...`

Even though line 673 makes use of DLLDIR, it is an absolute unix path (/mingw64/lib or /mingw64/include) in compiled Mingw64 code does not address a file.
` static const char *default_dirs[] = { DLLDIR, "/usr/lib/wine", "/usr/local/lib/wine" };`

Result: 
```
$ widl -t qga-vss.idl -o qga-vss.tlb
error: cannot find stdole2.tlb
```

In my opinion widl.c needs to be patched in the way this PR does.

As @lazka already said: This needs to be addressed to upstream..

Because now my analysis is complete, I guess I will try it in the near futuer.

PR #11549, opened 2022-04-22
openmp: swap static and shared order

1480c1:
Installs shared last

### DIFF
--- a/mingw-w64-openmp/PKGBUILD
+++ b/mingw-w64-openmp/PKGBUILD
@@ -8,7 +8,7 @@ _realname=openmp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=14.0.0
-pkgrel=2
+pkgrel=3
 pkgdesc="LLVM OpenMP Library (mingw-w64)"
 url="https://openmp.llvm.org/"
 arch=(any)
@@ -74,7 +74,7 @@ build() {
       ;;
   esac
 
-  for shared_libs in shared static; do
+  for shared_libs in static shared; do
     MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
     ${MINGW_PREFIX}/bin/cmake \
       -GNinja \
@@ -93,8 +93,8 @@ build() {
 }
 
 package() {
-  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --install "${srcdir}/build-${MSYSTEM}-shared"
   DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --install "${srcdir}/build-${MSYSTEM}-static"
+  DESTDIR=${pkgdir} ${MINGW_PREFIX}/bin/cmake --install "${srcdir}/build-${MSYSTEM}-shared"
 
   install -d "$pkgdir${MINGW_PREFIX}"/share/licenses/${_realname}
   install -Dm644 $_pkgfn/LICENSE* -t "$pkgdir${MINGW_PREFIX}"/share/licenses/${_realname}


Biswa96:
++pkgrel

1480c1:
doh

mati865:
Why?
There can be dllimports in shared headers. If thats the case then shared headers will work only with shared library.
Static headers on the other hand work with both static and shared libraries.

1480c1:
Personally, prefer static over shared, but this was recommended by @MehdiChinoune, although it seems like it was retracted?

mati865:
Hmm, I don't think static library is going to work right now:
`src/build-CLANG64-shared/runtime/src/omp.h`:
```C
#   if defined(_WIN32)
#       define __KAI_KMPC_CONVENTION __cdecl
#       ifndef __KMP_IMP
#           define __KMP_IMP __declspec(dllimport)
#       endif
#   else
#       define __KAI_KMPC_CONVENTION
#       ifndef __KMP_IMP
#           define __KMP_IMP
#       endif
#   endif
```

1480c1:
Hmm, not sure what would be the best option here. Could it be as simple as defining `__KMP_IMP` as empty? But looking at it, it seems they are only used for variable names which are replaced by an enum on non-WIN32 system (due to cl.exe and icl.exe not supporting 64-bit enums?)

Do you know if clang and gcc on Windows properly support 64-bit enums? Could I just tighten their macro checks from `_WIN32` to
```diff
diff -Naur openmp-14.0.0.src/runtime/src/include/omp.h.var src/openmp-14.0.0.src/runtime/src/include/omp.h.var
--- openmp-14.0.0.src/runtime/src/include/omp.h.var	2022-03-14 05:44:55.000000000 -0400
+++ openmp-14.0.0.src/runtime/src/include/omp.h.var	2022-04-22 20:33:24.149447300 -0400
@@ -356,7 +356,7 @@
         omp_uintptr_t value;
     } omp_alloctrait_t;
 
-#   if defined(_WIN32)
+#   if defined(_MSC_VER) || defined(__INTEL_COMPILER)
     // On Windows cl and icl do not support 64-bit enum, let's use integer then.
     typedef omp_uintptr_t omp_allocator_handle_t;
     extern __KMP_IMP omp_allocator_handle_t const omp_null_allocator;

```


Would that probably reasonably work?

1480c1:
Compiling with that diff at least works and nm -CAg --defined-only | grep imp shows no `__imp_` prefixed symbols, so I think it works?

lazka:
> Why?

I propagated that rule some time ago because some autotools/meson packages were writing pkg-config files that would break shared linking. If that's different here feel free to ignore.

mati865:
> > Why?
> 
> I proposed that rule some time ago because some autotools/meson packages were writing pkg-config files that would break shared linking. If that's different here feel free to ignore.

FWIW https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libc%2B%2B/PKGBUILD installs static over shared for the reason I mentioned earlier here.

> Do you know if clang and gcc on Windows properly support 64-bit enums?

I have no idea but I'd expect them to support it.

> ```diff
> +#   if defined(_MSC_VER) || defined(__INTEL_COMPILER)
> ```

AFAIK Intel identifies itself as MSVC, similarly how mingw-w64 pretends to be GCC.

> Compiling with that diff at least works and nm -CAg --defined-only | grep imp shows no `__imp_` prefixed symbols, so I think it works?

You could try changing other ifdefs to use 64 bit enums and see if static openmp works but it's not a big deal.

mati865:
FWIW libc++ in LLVM 15 will no longer need to be installed in specific order. Even more, static+shared can be built at same time: https://reviews.llvm.org/D125924

> I compared the two build directories and there is no difference, so the order doesn't matter.

No opposition from me then.

PR #11560, opened 2022-04-25
protobuf: enable clangarm64

kmilos:


kmilos:
Don't know why Clang linker is failing of the bat for protobuf-c...

Also, need to depend on packaged absel for protobuf now.

kmilos:
protobuf-c needed -DNDEBUG to match how protobuf itself was built (destructor defined in the header file rather than exported)

kmilos:
I give up... osgearth tries to build also a debug package, but that can't work now because of above mentioned -DNDEBUG requirement for the protobuf header...

Biswa96:
The arrow package needed a rebuild for aws-sdk-cpp update. So, I have added the patches to fix arrow package with llvm 14. Those can be skipped here. Also, would it be better to only add clangarm64 in this PR? Then update protobuf and dependencies in another day.

Speaking of arrow, would you like to check if clangarm64 can be enabled for these packages - arrow, aws-sdk-cpp and tiledb? Those packages are too big to be cross compiled in my PC.

kmilos:
Ok, giving reducing a scope a go...

I don't have an easy way to check other large packages for clangarm64 either ATM :(

PR #11579, opened 2022-04-28
nsis: update to 3.08

BFamzz:


mmuetzel:
See also #10743 which compiled correctly.

PR #6907, opened 2020-09-01
gcc: specify LDFLAGS_FOR_TARGET and --with-boot-ldflags, disable dynamicbase for compiler

jeremyd2019:
Related to #6899 (sharing the first commit with it), but with a less-hacky solution to disabling dynamicbase for the compiler itself while leaving it enabled for the runtime libraries.  Note that this depends on the binutils from #6906 for the --disable-dynamicbase flag (draft until that's live so CI can build successfully).

Results can be divided into 3 sets:
### Binaries that respect LDFLAGS or --with-boot-ldflags (LDFLAGS applies to stage1, --with-boot-ldflags to later stages):
./mingw-w64-i686-gcc/mingw32/bin/c++.exe
./mingw-w64-i686-gcc/mingw32/bin/cc.exe
./mingw-w64-i686-gcc/mingw32/bin/cpp.exe
./mingw-w64-i686-gcc/mingw32/bin/g++.exe
./mingw-w64-i686-gcc/mingw32/bin/gcc-ar.exe
./mingw-w64-i686-gcc/mingw32/bin/gcc-nm.exe
./mingw-w64-i686-gcc/mingw32/bin/gcc-ranlib.exe
./mingw-w64-i686-gcc/mingw32/bin/gcc.exe
./mingw-w64-i686-gcc/mingw32/bin/gcov-tool.exe
./mingw-w64-i686-gcc/mingw32/bin/gcov.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-c++.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-g++.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-gcc-10.2.0.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-gcc-ar.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-gcc-nm.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-gcc-ranlib.exe
./mingw-w64-i686-gcc/mingw32/bin/i686-w64-mingw32-gcc.exe
./mingw-w64-i686-gcc/mingw32/lib/gcc/i686-w64-mingw32/10.2.0/cc1.exe
./mingw-w64-i686-gcc/mingw32/lib/gcc/i686-w64-mingw32/10.2.0/cc1plus.exe
[...]

PR #11582, opened 2022-04-29
Update patches to GCC 12

lhmouse:
Since GCC 12 a lot of .c files have been renamed to .cc ones. This commit updates such patches. Plain patches have also been re-generated in mbox format for future revision. At the moment these patches apply to the `releases/gcc-12` branch.

~~Note: UCRT64 build is failing.~~


lhmouse:
Please review carefully, especially [0010-Fix-using-large-PCH.patch](https://github.com/msys2/MINGW-packages/pull/11582/files#diff-39be93710643535ef093e23a04cf3bb3295afdc62c6e4eacfdd05019e13c94a0).

MehdiChinoune:
I hope we won't wait months for `Ada on MINGW32' issue to be fixed.

lazka:
> Please review carefully, especially [0010-Fix-using-large-PCH.patch](https://github.com/msys2/MINGW-packages/pull/11582/files#diff-39be93710643535ef093e23a04cf3bb3295afdc62c6e4eacfdd05019e13c94a0).

I can't see any obvious errors at least.

@lhmouse any idea re mingw32 failing?

lhmouse:
Does it have something to do with 7e59389dd7fe748b38356f0fc78f5edeb933aa96 and 0c60660b0cbb485fa29ea09a229cb368e2d01bae?

lazka:
Unclear, ada always fails in this unhelpful way, last time when 11.1.0 was released..

lazka:
It also fails the same without any patches applied.
I reported it upstream: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105507

MehdiChinoune:
To those defending keeping Ada for 32bit. Last year after GCC was updated to 11.2.0 all Ada libraries were renamed from *-10.dll to *-11.dll, but dependent libraries were not rebuilt (I detected that months ago when I fixed runtime dependencies). Those libraries should be broken, but we didn't see any report about that => NO ONE IS USING ADA!

oscarfv:
Some gcc versions ago (8?) after many months with the release blocked because of Ada, when it was decided to drop it from the build, one or two users showed up and thanks to that the problem was eventually solved.

Maybe there are no MSYS2/Ada users, maybe there are some who don't update often... In any case providing Ada is nice.

However, having the release blocked because of Ada-i686 makes little sense, because it is very dubious that nowadays there is demand for Ada on that platform.

bustercopley:
Thanks, @lhmouse
But PCH seems not to be working on this branch. Creating the PCH appears to succeed, but using it fails (error code 1, no error message).

To reproduce:
* create "precompiled.hpp" (can be empty, or anything valid)
* create "main.cpp" with some non-trivial content, e.g., `#include <utility>`
* `g++ -c -o precompiled.hpp.gch -x c++-header precompiled.hpp`
* `g++ -c -o main.o main.cpp -include precompiled.hpp`

Without any `0010-Fix-using-large-PCH.patch` at all, this error doesn't happen in this test case. (But of course, it would fail if the PCH ended up too large.)

lhmouse:
@bustercopley I couldn't reproduce this issue. Would you please check whether [my personal toolchains](https://gcc-mcf.lhmouse.com/) work? Thanks.

```text
lh_mouse@lhmouse-pc ~/Desktop $ cat precompiled.hpp
#include <stdio.h>
lh_mouse@lhmouse-pc ~/Desktop $ cat main.cpp
int
main(void)
  {
    ::puts("meow");
  }
lh_mouse@lhmouse-pc ~/Desktop $ g++ -c -o precompiled.hpp.gch -x c++-header precompiled.hpp && echo $?
0
lh_mouse@lhmouse-pc ~/Desktop $ g++ -o main.exe main.cpp -include precompiled.hpp -Winvalid-pch && echo $?
0
lh_mouse@lhmouse-pc ~/Desktop $ ./main.exe
meow
lh_mouse@lhmouse-pc ~/Desktop $
```

bustercopley:
> @bustercopley I couldn't reproduce this issue. Would you please check whether [my personal toolchains](https://gcc-mcf.lhmouse.com/) work? Thanks.

@lhmouse, your recipe succeeds on both toolchains (my build of this branch / your x64 msvcrt toolchain), but here's a recipe that fails on both. (The difference is I added some content `#include <utility>` to main.cpp.)

```
$ cat precompiled.hpp
#include <stdio.h>
$ cat main.cpp
#include <utility>
int
main(void)
  {
    ::puts("meow");
  }
$ PATH=/mingw64/bin
$ g++ -c -o precompiled.hpp.gch -x c++-header precompiled.hpp ; echo $?
0
$ g++ -c -o main.o main.cpp -include precompiled.hpp -Winvalid-pch ; echo $?
1
$ PATH=/c/mingw-w64-gcc-mcf_20220506_12.1.1_x64-msvcrt_51082ba7326b8d5830b0f6ea25d6577bf9ffd7cc/mingw64/bin
$ g++ -c -o precompiled.hpp.gch -x c++-header precompiled.hpp ; echo $?
0
$ g++ -c -o main.o main.cpp -include precompiled.hpp -Winvalid-pch ; echo $?
1
```

lhmouse:
Looks like that patch should be refactored somehow.

lhmouse:
Reference:  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=14940

lhmouse:
@bustercopley The second hunk for 'gcc/ggc-common.c' seems no longer necessary; please test the new build. See <https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=e4641191287ca613529d78a906afe4f029c1c3cd>.

lhmouse:
New builds have been uploaded to <https://gcc-mcf.lhmouse.com/>

I tested this precompiled header, which resulted in a 313MiB .gch file, and observed no errors in using it:

```c++
#include <functional>
#include <algorithm>
#include <string>
#include <vector>
#include <list>
#include <deque>
#include <map>
#include <set>
#include <queue>
#include <stack>
#include <bitset>
#include <iterator>
#include <thread>
#include <mutex>
#include <chrono>
#include <numeric>
#include <memory>
#include <type_traits>
#include <istream>
#include <ostream>
#include <exception>
#include <random>
#include <valarray>
#include <future>
#include <regex>
#include <utility>

#include <boost/regex.hpp>
#include <boost/date_time.hpp>
#include <boost/circular_buffer.hpp>
#include <boost/filesystem.hpp>
#include <boost/function.hpp>
#include <boost/multi_index_container.hpp>
#include <boost/static_string.hpp>
#include <boost/container/vector.hpp>
#include <boost/container/list.hpp>
#include <boost/container/deque.hpp>
#include <boost/container/map.hpp>
#include <boost/container/set.hpp>
#include <boost/range/algorithm.hpp>
#include <boost/multi_index/random_access_index.hpp>
#include <boost/multi_index/ordered_index.hpp>
#include <boost/multi_index/hashed_index.hpp>
#include <boost/thread/thread.hpp>
#include <boost/thread/mutex.hpp>
#include <boost/thread/condition_variable.hpp>
```

bustercopley:
> @bustercopley The second hunk for 'gcc/ggc-common.c' seems no longer necessary; please test the new build. See https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=e4641191287ca613529d78a906afe4f029c1c3cd.

Thanks, works for me too.

bustercopley:
> Added a new patch as suggested by Jakub Jelinek in https://gcc.gnu.org/pipermail/gcc-patches/2022-May/594556.html.

The PCH relocation does seem to work now. (I rebuilt the gcc packages from this branch, installed them, then set the DYNAMIC_BASE (i.e., ASLR) flag on the compiler binaries under /mingw64/lib/gcc/x86_64-w64-mingw32/12.1.0/, and with those binaries, I can create and use a PCH successfully.)

(See also #6907)

lazka:
```
76f9c7f44fffb0b03266730b137313fe79f1c99e is the first bad commit
commit 76f9c7f44fffb0b03266730b137313fe79f1c99e
Author: Bob Duff <duff@adacore.com>
Date:   Wed Feb 3 05:31:16 2021 -0500

    [Ada] Variable-sized node types

 167 files changed, 18305 insertions(+), 43812 deletions(-)
```

lazka:
> reported upstream...

also added a stack trace there now: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105507#c2

lazka:
The mingw32 problem should be fixed once #11693 is done

lhmouse:
The UCRT64 build starts failing again. Log attached.
[mingw-w64-gcc-12.1.1+20220520-1-x86_64-build.log.gz](https://github.com/msys2/MINGW-packages/files/8737612/mingw-w64-gcc-12.1.1%2B20220520-1-x86_64-build.log.gz)

lazka:
I think this is good to go now. Any objections?

lhmouse:
Do we start building packages with GCC 12, or provide a separate _gcc-12_ package? It's the first in the 12-series release and may contain serious bugs.

lazka:
The package is not really in the shape to be parallel installable, but we can certainly make it an alternative. I'll have a look.

MehdiChinoune:
> Do we start building packages with GCC 12, or provide a separate _gcc-12_ package? It's the first in the 12-series release and may contain serious bugs.

OpenSUSE tmbleweed rebuit all of its packages against gcc-12 (I think Fedora did the same but I am not sure)
Both Fedora and Arch use gcc-12 as their default compiler.

mati865:
We have issues about GCC 11 crashing anyway, can it be much worse?

lazka:
> OpenSUSE tmbleweed rebuit all of its packages against gcc-12 (I think Fedora did the same but I am not sure)
> Both Fedora and Arch use gcc-12 as their default compiler.

good point. let's give it a try.

PR #11587, opened 2022-04-30
add rkdeveloptool for flashing to rockchip emmc

mightymos:
Add package based on this patch:
https://github.com/rockchip-linux/rkdeveloptool/pull/56

### DIFF
--- a/mingw-w64-rkdeveloptool/001-fixes-for-windows.patch
+++ b/mingw-w64-rkdeveloptool/001-fixes-for-windows.patch
@@ -1,0 +1,141 @@
+diff --git a/DefineHeader.h b/DefineHeader.h
+index cea0f02..321265c 100644
+--- a/DefineHeader.h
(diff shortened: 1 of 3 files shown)


Biswa96:
Please follow this PKGBUILD template https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.autotools-tarball.sh. Required changes as of now:

1. Append `(mingw-w64)` in pkgdesc.
2. License should be in SPDX format. e.g. here `license=('spdx:GPL-2.0-or-later')`
3. It is not required to mention automake, autoconf and make separately. Mention `${MINGW_PACKAGE_PREFIX}-autotools` instead.
4. Add proper options with configure command.

And many more.

Biswa96:
The previous pkgver variable and function are required because this package is built from git repository instead of a specific tag.

PR #11591, opened 2022-05-01
Update to GDB 12.1

GitMensch:


GitMensch:
a simple test debugging with the resulting binary worked fine

lazka:
thanks

GitMensch:
Thanks for providing MSYS2!

Side note: the really outstanding feature of GDB12 is that you can finally create new mi commands from a python extension.

PR #11592, opened 2022-05-02
python-imageio: update to 2.19.0

kmilos:


mingw-w64-python-imageio/PKGBUILD
```diff
@@ -14,10 +14,11 @@ license=('spdx:BSD-2-Clause')
 depends=("${MINGW_PACKAGE_PREFIX}-python-numpy"
          "${MINGW_PACKAGE_PREFIX}-python-pillow")
 optdepends=("${MINGW_PACKAGE_PREFIX}-freeimage"
+            $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || echo "${MINGW_PACKAGE_PREFIX}-opencv")
```
Biswa96, line 17: This aarch64 check may not require. optdepends are not checked at build time. It might be better if short descriptions are added with each optional dependencies. See https://wiki.archlinux.org/title/PKGBUILD#optdepends

kmilos: Done, thanks

PR #11598, opened 2022-05-04
libavif: update to 0.10.1, enable clangarm64

kmilos:


Biswa96:
Just wondering if it is possible enable arm for libheif and libgd also.

kmilos:
Yeah, might try...

kmilos:
Turns out x265 is already available, but [libde265 requires quite a bit of patching to get to work on aarch64](https://github.com/strukturag/libde265/issues/308), so disabled it for now...

kmilos:
At the same time, there seem to be aarch64 Linux packages around, so maybe it just works? https://archlinuxarm.org/packages/aarch64/libde265

Biswa96:
Yeah, I've cross-compiled libde265 with llvm-mingw, not sure if it works.

kmilos:
Ok, let's give https://github.com/msys2/MINGW-packages/pull/11599 a go separately

kmilos:
I've seen these random 'Permission denied' failures locally as well...

PR #11599, opened 2022-05-04
libde265: enable for clangarm64

kmilos:


kmilos:
Rolling into https://github.com/msys2/MINGW-packages/pull/11598 after all...

PR #11604, opened 2022-05-05
sundials: upgrade to 6.2.0

gharveymn:
No issues with the build.

PR #11607, opened 2022-05-06
sundials: remove mpi logging

gharveymn:
This fixes the issue raised [here](https://github.com/msys2/MINGW-packages/pull/11604#issuecomment-1118296278) by @mmuetzel. SUNDIALS added logging with MPI and turned it on by default. This just turns it off.


Biswa96:
Need to bump pkgrel.

gharveymn:
Whoops. There you go.

mmuetzel:
According to the CI, the dependency on that library is still there:
https://github.com/msys2/MINGW-packages/runs/6317227016?check_suite_focus=true#step:9:4799

gharveymn:
Yeah MPI has been a dependency since before 6.2.0, but only for libraries with specific support for MPI, PETSc, and SuperLU-Dist. `msmpi` provided by msys2 acts as an import library, but does not distribute the dll, so the "not found" behavior is to be expected for these libraries. Libraries like `libsundials_ida` aren't affected so Octave should be good to go as before.

Biswa96:
> msmpi provided by msys2 acts as an import library

But the msmpi library is provided as libmsmpi.a static library. If it were an import library it would be libmsmpi.dll.a. The file list of msmpi package can be found here https://packages.msys2.org/package/mingw-w64-x86_64-msmpi?repo=mingw64

gharveymn:
It actually behaves like an import library. Try the following:

```c
#include <mpi.h>
int main (void) 
{ 
  MPI_Comm_disconnect(NULL);
  return 0; 
}
```
```commandline
> gcc test.c -lmsmpi
```

Result:

```commandline
> ntdll a.exe
        KERNEL32.dll => C:\WINDOWS\SYSTEM32\KERNEL32.dll (0x00000207dee60000)
        msmpi.dll => C:\WINDOWS\SYSTEM32\msmpi.dll (0x00000207df020000)
        msvcrt.dll => C:\WINDOWS\SYSTEM32\msvcrt.dll (0x00000207df3b0000)
```

mmuetzel:
@gharveymn: Thank you for clarifying.
Given your example and looking at the build rule for msmpi, I agree with @Biswa96 that it would be better if that library were renamed to `libmsmpi.dll.a`. That would probably have caused less confusion on my end.
But that is independent of the issue here. And it shouldn't hold back this fix IMHO.

Biswa96:
Is msmpi sdk from microsoft compiled from https://github.com/Microsoft/Microsoft-MPI repository? Should it be added in our mingw msmpi package? @okhlybov

PR #11612, opened 2022-05-06
wt: rebuild

lazka:
since it failed in autobuild


lazka:
no idea. since the package has no dependants I'm going to disable clang32 for now.

PR #11623, opened 2022-05-07
Fixes issue mentioned in https://github.com/KhronosGroup/SPIRV-Tools/…

lygstate:
…pull/4794

Used to building mesa with OpenCL on Direct3D 12 support


Biswa96:
I think we should wait for upstream to merge the changes. vulkan & spirv things are highly sensitive 😊

Biswa96:
No reply in that upstream pull request. Let merge this, shall we?

lygstate:
OK, lets update it.

PR #11631, opened 2022-05-08
wxWidgets3.1: Update to version 3.1.6

stahta01:
And, I enabled the building of samples when check is ran.
Also, added patch to fix clang64 building of samples.

Edit: This will break wxPython; but, I added wxPython4.0 that depends on wxWidgets 3.0 a few days ago.
This was to support KiCAD and any third party needs for wxPython. The wxPython has not had movement for a few months and fixing it will take a person who knows python or at least sip.

Edit: In the last week, the wxPython project has release an snapshot build.

Tim S.


mingw-w64-wxWidgets3.1/009-wxWidgets-3.1.5-Fix-clang-sample-build.patch
```diff
@@ -0,0 +1,26 @@
...
+--- a/samples/sample.rc
++++ b/samples/sample.rc
+@@ -23,7 +23,11 @@ sample ICON "sample.ico"
+ // set this to 1 if you don't want to use manifest resource (manifest resource
+ // is needed to enable visual styles on Windows XP - see docs/msw/winxp.txt
+ // for more information)
++#if !defined(__clang__)
+ #define wxUSE_NO_MANIFEST 0
++#else
++#define wxUSE_NO_MANIFEST 1
```
driver1998, line 21: What is the exact error? I would like to fix it instead of hiding it.

stahta01: The macro magic fails to work and there is a error because of spaces in the filename of the manifest.
It only happens when samples or anything else using the rc file.

Tim S.

stahta01: The patch is just needed to build the samples locally; it does not effect the library package files.
But, a real fix is needed to the wx.rc file in wx/msw folder that is where the macro commands are that fail to work.

stahta01: I believe the exact error is 
`llvm-rc: Error in 24 statement (ID 1): 
Is a directory`

ZachBacon: Oddly enough, I get those from a CI on my own projects that use wxwidgets, BUT not in local builds.

jeremyd2019: llvm/llvm-project#51286?

driver1998: > The patch is just needed to build the samples locally; it does not effect the library package files. But, a real fix is needed to the wx.rc file in wx/msw folder that is where the macro commands are that fail to work.

Yeah we would need to fix that, otherwise everybody using the built-in manifest in wxWidgets would be affected.

stahta01: It only happens with clang compiler.

stahta01: The problem seems to be in "llvm-windres" or "llvm-rc"; so, it is not likely an wxWidgets problem.

When the 3 and 4 command line params are in double quotes the command should combine the two; but, it does not happed. So, "llvm-windres" at least is not working like GCC mingw windres. This is my guess based on a small amount of testing implies.

Tim S.

mingw-w64-wxWidgets3.1/001-wxWidgets-3.0.2-relocate-prefix-in-bin-wx-config.patch
```diff
@@ -4,7 +4,7 @@
  
  # Determine the base directories we require.
  prefix=${input_option_prefix-${this_prefix:-@prefix@}}
-+if [ "x${MSYSTEM}" = "xMINGW32" ] || [ "x${MSYSTEM}" = "xMINGW64" ]; then
++if [ "x${MSYSTEM}" = "xMINGW32" ] || [ "x${MSYSTEM}" = "xMINGW64" ] || [ "x${MSYSTEM}" = "xUCRT64" ] || [ "x${MSYSTEM}" = "xCLANG64" ] || [ "x${MSYSTEM}" = "xCLANG32" ] || [ "x${MSYSTEM}" = "xCLANGARM64" ]; then
```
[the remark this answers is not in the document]
Biswa96: I would like to consider that "unnecessary" 😊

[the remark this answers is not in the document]
stahta01: FYI: It is possible; but, hard to use the wx-config outside of the MSys2 environment; I try testing it once and it works; but, you need to pass the prefix value to the script. No idea if removing the guard make it easier or impossible to do or have no effect.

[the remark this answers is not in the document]
Biswa96: Outside of msys2 environment, cygpath would not exist. So, the prefix value has to be passed explicitly, right?

stahta01:
Summary of the current changes

wxWidgets3.1: Update to version 3.1.6; and,

Removed MSYSTEM guard in "wx-config.in".
Redo manifest filename logic in "wx/msw/wx.rc".
Add replaces "wxmsw3.1-git" package.
Add "enable-graphics-d2d" to help support Code::Blocks IDE.

Tim S.

stahta01:
Merged changes into single commit in #11691 

Tim S.

PR #11632, opened 2022-05-08
ca-certificate: update to 20211016 and explicitly request .exe p11-kit

1480c1:


Biswa96:
Is it possible to add a commit message why this change is need and why the patch is removed?

1480c1:
Changed commit message to
```none
    ca-certificate: update to 20211016 and explicitly request .exe p11-kit
    
    blacklist-dst-root-ca-x3.patch has been added upstream
    
    using .exe explicitly allows for usage through archlinux's pacman through
    wine. If not, then programs using libcurl such as git will fail
    due to the .crt files being empty
    
    Signed-off-by: Christopher Degawa <ccom@randomderp.com>
```

PR #11633, opened 2022-05-09
xmlcatalog: explicitly request .exe

1480c1:
fixes an issue where the .INSTALL file wouldn't run properly on archlinux


Biswa96:
Doesn't docbook-xml also need a pkgrel bump?

1480c1:
bumped docbook-xml, also clarified why in commit message

PR #11635, opened 2022-05-09
python-dotty-dict: new package

fauxpark:
Patches are temporary, due to https://github.com/pawelzny/dotty_dict/issues/69 - the changes have been merged, but no new release yet. I figured this was better than submitting one or both of this and [qmk-dotty-dict](https://pypi.org/project/qmk-dotty-dict/), which was created to solve this problem on MSYS2.


Biswa96:
Is that change required for this package? or does python itself have the issue? Found something https://stackoverflow.com/questions/45855297

fauxpark:
I'm not 100% sure. Adding `encoding='utf-8'` to the readme `open()` probably does fix *that* error, but I wonder if the `setup()` call it is being passed to will have anything to say.

fauxpark:
Tested and working in a fresh VM with Japanese language set 👍 seems like I had to change the locale for non-Unicode programs for `locale.getpreferredencoding()` to finally give me `cp932` instead of `cp1252`.

PR #11640, opened 2022-05-10
clang: Update to 14.0.3

lazka:
because why not


jeremyd2019:
should do libc++ also?  There was a bug fixed in 14.0.1 that I remember we were discussing, but don't remember what happened with that

lazka:
as for fixes, I only remember someone linking https://github.com/mstorsjo/llvm-mingw/issues/274

Issue #11649, opened 2022-05-12
advancecomp ported as msys2 package request


Craw93L:
I need someone to port advanecomp to works as msys2 package
Some games are using advpng from it (its PNG Compression Utility)
I just started to use msys2 and I don't know how to do it
https://github.com/amadvance/advancecomp/releases/download/v2.3/advancecomp-2.3.tar.gz (source)
https://github.com/amadvance/advancecomp/releases/download/v2.3/advancecomp-2.3-windows-x64.zip (binary x64)
https://github.com/amadvance/advancecomp/releases/download/v2.3/advancecomp-2.3-windows-x86.zip (binary x86)

lazka:
https://aur.archlinux.org/packages/advancecomp

Biswa96:
Hi @Craw93L, I have added a pull request to add the package. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/2311920283. Please follow these steps:

1. Scroll down and download the correct artifact. It'll be a zip file.
2. Extract the package (.tar.zst) from zip file.
3. Install it with `pacman -U the-package-name.tar.zst` command.

PR #11650, opened 2022-05-12
advancecomp: Add package

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/11649


Biswa96:
ucrt environments are disable due to build failure. Some "portable" wrapper functions access struct _iobuf members which are not defined in ucrt.

PR #11652, opened 2022-05-12
add glib2 as runtime depends to aravis utils

eudoxos:
Reported here:
https://aravis-project.discourse.group/t/dll-missing-error-when-trying-to-open-arv-camera-test-0-8/450


Biswa96:

It is OK to add this dependency. Just wondering. gobject-introspection-runtime depends on glib2. Doesn't it pull glib2 while installing aravis?

eudoxos:
> It is OK to add this dependency. Just wondering. gobject-introspection-runtime depends on glib2. Doesn't it pull glib2 while installing aravis?
I just see the g-i-r→glib2 dependency. In that case I am closing the PR as it is redundant. As I am the original packager, I though I screwed up ando verreacted based on one forum report. If the problem comes again, I will reopen (and bump the pkgrel). Thanks for your clarifications.

PR #11659, opened 2022-05-14
fd: new package

Freed-Wu:
Refer archlinux's PKGBUILD.

### DIFF
--- a/mingw-w64-fd/PKGBUILD
+++ b/mingw-w64-fd/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: Wu Zhenyu <wuzy01@qq.com>
+_realname=fd
+
+pkgbase=mingw-w64-$_realname
+pkgname=$MINGW_PACKAGE_PREFIX-$_realname
+pkgver=8.3.2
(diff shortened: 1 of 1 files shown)


Biswa96:
Please follow this [template PKGBUILD][1] file and remove those unnecessary variables and comments.

[1]: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.cargo-tarball.sh

Biswa96:
Did you test the package with lto enabled? I am not sure if lto should be enabled in mingw packages.

Freed-Wu:
<https://github.com/sharkdp/fd/blob/master/Cargo.toml#L78>

```
[profile.release]
lto = true
```

Did you mean it?

Biswa96:
I mean adding this command `export CFLAGS+=" -ffat-lto-objects"`.

Freed-Wu:
Now lto has been removed

Freed-Wu:
Now it can work.

Freed-Wu:
Are there any problem about this package?

PR #11662, opened 2022-05-15
gdu: new package

Freed-Wu:
`go test` seem cannot pass in windows, but the binary file compiled is same as 
the release of the author.


Freed-Wu:
See <https://github.com/dundee/gdu/issues/144>, the test failure has many reason, such as `/dev/null` not exist, the `/home/XXX` not exist. So I think the `check()` should be deleted to convince it can be compiled. and since now the mintty seems not support interaction mode of gdu.

lazka:
you can get interactive mode via `winpty gdu`

PR #11679, opened 2022-05-17
rav1e: fix .pc file

kmilos:
Currently says `prefix=/usr/local`


lazka:
prefix is ignored anyway, but why not :)

PR #11685, opened 2022-05-18
precice: update

3rav:
Update PKGBUILD

Update PKGBUILD

Update PKGBUILD

### DIFF
--- a/mingw-w64-precice/PKGBUILD
+++ b/mingw-w64-precice/PKGBUILD
@@ -1,4 +1,4 @@
-# Maintainer: Rafal Brzegowy <rafal.brzegowy@yahoo.com>
+# Maintainer: Rafał Brzegowy <rafal.brzegowy@yahoo.com>
 # Contributor: Raed Rizqie <raed.rizqie@gmail.com>
 
 # petsc don't have clang build because it depends on fortran lib,
@@ -7,8 +7,8 @@
 _realname=precice
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=2.3.0
-pkgrel=6
+pkgver=2.4.0
+pkgrel=1
 pkgdesc='A Coupling Library for Partitioned Multi-Physics Simulations on Massively Parallel Systems (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


Biswa96:
The binprecice executable is installed as symbolic link ([ci log][1]). Shouldn't it be installed as a copy of precice-tool.exe? symbolic link creation requires admin privilege if developer mode is not enabled, right?

[1]: https://github.com/msys2/MINGW-packages/runs/6494750829?check_suite_focus=true#step:9:1457

lazka:
yeah, that needs to be fixed somehow. installing the package will fail without symlink support.

fsimonis:
Hi, this issue is tricky to fix as there is no copy_on_error option of the CMake CLI version of the `create_symlink` command.

We will release one more minor version before version 3. So, this symlink will not stick around for much longer.

I recommend to simply replace it by a `copy` or `copy_if_different`.

```diff
   add_custom_target(
     binprecice
     ALL
-    COMMAND ${CMAKE_COMMAND} -E create_symlink precice-tools binprecice
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different precice-tools binprecice
     DEPENDS precice-tools
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     BYPRODUCTS binprecice
     )
```

mingw-w64-precice/PKGBUILD
```diff
@@ -26,7 +26,12 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-petsc-build"
 license=('LGPL3')
 url='https://www.precice.org'
 source=(https://github.com/precice/precice/archive/v${pkgver}.tar.gz)
-sha256sums=('57BAB08E8B986F5FAA364689D470940DBD9C138E5CFA7B861793E7DB56B89DA3')
+sha256sums=('762E603FBCAA96C4FB0B378B7CB6789D09DA0CF6193325603E5EEB13E4C7601C')
+
+prepare() {
+  cd "${_realname}-${pkgver}"
+  sed -i 's/BYPRODUCTS binprecice/#BYPRODUCTS binprecice/' CMakeLists.txt
```
fsimonis, line 33: This should work for packages until version 3

```suggestion
  sed -i 's/E create_symlink/E copy_if_different/' CMakeLists.txt
```

mingw-w64-precice/PKGBUILD
```diff
@@ -26,7 +26,14 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-petsc-build"
...
+  #Temporary fix: https://github.com/msys2/MINGW-packages/pull/11685#issuecomment-1200834235
+  sed -i 's/E create_symlink/E copy_if_different/' CMakeLists.txt
```
mmuetzel, line 35: You might need to include the exe extension here:
``` diff
-  sed -i 's/E create_symlink/E copy_if_different/' CMakeLists.txt
+  sed -i 's/E create_symlink precice-tools binprecice/E copy_if_different precice-tools${CMAKE_EXECUTABLE_SUFFIX} binprecice${CMAKE_EXECUTABLE_SUFFIX}/' CMakeLists.txt
```

And also rebase on a current head of the master branch.

Issue #11539, opened 2022-04-21
static `libavcodec.a` links against dynamic `libopenjp2.dll.a`


msink:
static `libavcodec.a` links against dynamic libopenjp2.dll.a:

```
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x1ca): undefined reference to `__imp_opj_create_decompress'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x1e9): undefined reference to `__imp_opj_set_error_handler'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x204): undefined reference to `__imp_opj_set_warning_handler'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x21f): undefined reference to `__imp_opj_set_info_handler'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x247): undefined reference to `__imp_opj_setup_decoder'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x252): undefined reference to `__imp_opj_stream_default_create'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x26e): undefined reference to `__imp_opj_stream_set_read_function'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x27e): undefined reference to `__imp_opj_stream_set_skip_function'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x28e): undefined reference to `__imp_opj_stream_set_seek_function'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x29f): undefined reference to `__imp_opj_stream_set_user_data'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x2ac): undefined reference to `__imp_opj_stream_set_user_data_length'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x2bd): undefined reference to `__imp_opj_read_header'
...\lib\libavcodec.a(libopenjpegdec.o):libopenjpegdec:(.text+0x3d7): undefined reference to `__imp_opj_image_destroy'
[...]

Issue #11538, opened 2022-04-21
static `libavcodec.a` links against dynamic `libx264.dll.a`


msink:
static `libavcodec.a` links against dynamic `libx264.dll.a`:

```
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x50): undefined reference to `__imp_x264_picture_init'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x12f): undefined reference to `__imp_x264_encoder_encode'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x2f6): undefined reference to `__imp_x264_encoder_delayed_frames'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x46f): undefined reference to `__imp_x264_encoder_reconfig'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x528): undefined reference to `__imp_x264_encoder_reconfig'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x5c1): undefined reference to `__imp_x264_encoder_reconfig'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x63d): undefined reference to `__imp_x264_encoder_reconfig'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x957): undefined reference to `__imp_x264_encoder_reconfig'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text+0x9db): more undefined references to `__imp_x264_encoder_reconfig' follow
...\lib\libavcodec.a(libx264.o):libx264.c:(.text.unlikely+0x30): undefined reference to `__imp_x264_param_cleanup'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text.unlikely+0x42): undefined reference to `__imp_x264_encoder_close'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text.unlikely+0x75): undefined reference to `__imp_x264_param_parse'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text.unlikely+0x151): undefined reference to `__imp_x264_param_default'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text.unlikely+0x18b): undefined reference to `__imp_x264_param_default_preset'
...\lib\libavcodec.a(libx264.o):libx264.c:(.text.unlikely+0x369): undefined reference to `__imp_x264_param_parse'
[...]

Issue #11515, opened 2022-04-18
Static linking with `libopus.a` and `libcelt0.a`


msink:
If both `libopus.a` and `libcelt0.a` have to be linked in application - linker error:

```
...\lib\libopus.a(celt_encoder.o):(.text+0x830): multiple definition of `celt_encoder_get_size'; ...\lib\libcelt0.a(celt.o):(.text+0x4bb0): first defined here
...\lib\libopus.a(celt_encoder.o):(.text+0x6940): multiple definition of `celt_encoder_init'; ...\lib\libcelt0.a(celt.o):(.text+0x4d50): first defined here
...\lib\libopus.a(mdct.o):(.text+0x0): multiple definition of `clt_mdct_init'; ...\lib\libcelt0.a(mdct.o):(.text+0x0): first defined here
...\lib\libopus.a(mdct.o):(.text+0x180): multiple definition of `clt_mdct_clear'; ...\lib\libcelt0.a(mdct.o):(.text+0x110): first defined here
...\lib\libopus.a(pitch.o):(.text+0x2c0): multiple definition of `pitch_downsample'; ...\lib\libcelt0.a(pitch.o):(.text+0x170): first defined here
...\lib\libopus.a(pitch.o):(.text+0x620): multiple definition of `pitch_search'; ...\lib\libcelt0.a(pitch.o):(.text+0x3c0): first defined here
...\lib\libopus.a(quant_bands.o):(.text+0x5b0): multiple definition of `quant_coarse_energy'; ...\lib\libcelt0.a(quant_bands.o):(.text+0x500): first defined here
...\lib\libopus.a(quant_bands.o):(.text+0xc10): multiple definition of `quant_fine_energy'; ...\lib\libcelt0.a(quant_bands.o):(.text+0xa20): first defined here
...\lib\libopus.a(quant_bands.o):(.text+0xe50): multiple definition of `quant_energy_finalise'; ...\lib\libcelt0.a(quant_bands.o):(.text+0xbf0): first defined here
...\lib\libopus.a(quant_bands.o):(.text+0x1090): multiple definition of `unquant_coarse_energy'; ...\lib\libcelt0.a(quant_bands.o):(.text+0xdd0): first defined here
...\lib\libopus.a(quant_bands.o):(.text+0x1370): multiple definition of `unquant_fine_energy'; ...\lib\libcelt0.a(quant_bands.o):(.text+0x1070): first defined here
[...]

Issue #11514, opened 2022-04-18
static `libavformat.a` links against dynamic `libxml2.dll.a`


msink:
static `libavformat.a` links against dynamic `libxml2.dll.a`:
```
libavformat.a(dashdec.o):(.text+0xb15): undefined reference to `__imp_xmlGetProp'
libavformat.a(dashdec.o):(.text+0xb1c): undefined reference to `__imp_xmlFree'
libavformat.a(dashdec.o):(.text+0xc73): undefined reference to `__imp_xmlGetProp'
libavformat.a(dashdec.o):(.text+0xc82): undefined reference to `__imp_xmlFree'
libavformat.a(dashdec.o):(.text+0x1582): undefined reference to `__imp_xmlNodeGetContent'
libavformat.a(dashdec.o):(.text+0x15b6): undefined reference to `__imp_xmlFree'
libavformat.a(dashdec.o):(.text+0x15ca): undefined reference to `__imp_xmlNodeGetContent'
libavformat.a(dashdec.o):(.text+0x15ff): undefined reference to `__imp_xmlFree'
libavformat.a(dashdec.o):(.text+0x1612): undefined reference to `__imp_xmlNodeGetContent'
libavformat.a(dashdec.o):(.text+0x1647): undefined reference to `__imp_xmlFree'
libavformat.a(dashdec.o):(.text+0x165a): undefined reference to `__imp_xmlNodeGetContent'
libavformat.a(dashdec.o):(.text+0x168f): undefined reference to `__imp_xmlFree'
libavformat.a(dashdec.o):(.text+0x21e4): undefined reference to `__imp_xmlCheckVersion'
libavformat.a(dashdec.o):(.text+0x2204): undefined reference to `__imp_xmlReadMemory'
libavformat.a(dashdec.o):(.text+0x2212): undefined reference to `__imp_xmlDocGetRootElement'
libavformat.a(dashdec.o):(.text+0x226e): undefined reference to `__imp_xmlFreeDoc'
libavformat.a(dashdec.o):(.text+0x2274): undefined reference to `__imp_xmlCleanupParser'
libavformat.a(dashdec.o):(.text+0x227f): undefined reference to `__imp_xmlFreeNode'
libavformat.a(dashdec.o):(.text+0x237b): undefined reference to `__imp_xmlGetProp'
libavformat.a(dashdec.o):(.text+0x23be): undefined reference to `__imp_xmlFree'
[...]

PR #11689, opened 2022-05-19
ffmpeg: Add pkgconfig static flag for static library

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/11514
Fixes https://github.com/msys2/MINGW-packages/issues/11515
Fixes https://github.com/msys2/MINGW-packages/issues/11538
Fixes https://github.com/msys2/MINGW-packages/issues/11539


msink:
Ok, but when creating that issues for each dependency, I assumed they will be closed when "need some changes also" will be done for them individually :)

Biswa96:
Would you like to test the package from CI artifacts please? I have tested with your provided tutorial07.c in the main issue. This change seems to solve all the linked issues in my setup.

msink:
At least in MINGW64 mode CI artifact fixes all 4 issues above.

PR #11691, opened 2022-05-19
wxWidgets3.1: Update to version 3.1.6; and,

stahta01:
Removed MSYSTEM guard in "wx-config.in",
the other option of adding all the other environments 
like CLANG64 was advised against in the prior PR.
Redo manifest filename logic in "wx/msw/wx.rc".
The prior manifest logic failed to work under Clang64;
When building samples during check.
Add replaces "wxmsw3.1-git" package.
Add "enable-graphics-d2d" to help support Code::Blocks IDE.
Enable check code to build DLL samples.


lazka:
thanks!

PR #11696, opened 2022-05-20
python-qmk: new package

fauxpark:


fauxpark:
Hm, pip is unhappy about the setup.cfg specifying `qmk-dotty-dict` instead of `dotty-dict`. The former should no longer be needed as per #11635. This may need to be held off until we can cut a new release. I'll also see about properly fixing dotty-dict itself.

PR #11699, opened 2022-05-22
opencascade: update to 7.6.2

mmuetzel:
Also fix building with clang.
The patch for building with clang is pretty large. But it can be generated semi-automatically. I tried to outline the steps to re-create the patch in its description in case this should ever be necessary.
I was considering proposing the change upstream. But their process for contribution is pretty convoluted. If there is anyone who can already contribute upstream, feel free to propose these changes. I don't claim any copyright to them.

The API of OpenCascade changed notably between versions, and dependencies need to be updated as well. One of these dependencies (kicad) is pretty large, too. Compilation takes pretty long and the CI is likely to time out. (I hope that the PR can still be merged even in this case.)

While working on this, I noticed that `doxygen` in the CLANG32 environment doesn't seem to work currently. I de-activated building the documentation of OpenCascade in that environment for now to work-around the issue. It can probably be re-activated once `doxygen` is fixed.


Biswa96:
> The patch for building with clang is pretty large.

I think, it is better not to have a clang port instead of maintaining huge patches.

mmuetzel:
> I think, it is better not to have a clang port instead of maintaining huge patches.

In general, I'd agree. But in this specific case, the changes are trivial and they can be generated (almost entirely) automatically (were it not for class templates which I struggled to detect with a simple sed replacement). If those could be sorted out somehow, the patch could be replaced by a single command.
On the other hand, applying a patch is probably faster than running that command (if we had one) on a large code base...

stahta01:
I suggest setting "_build_support_libraries" to no to stop the build timeouts.

I added that so the user do not need to download the support libs that only changes with the first release.

Tim S.

mmuetzel:
The size of the tarballs for the support libraries differs significantly from the one that was previously packaged.
It would probably be a good idea to update those (at least once) for the 7.6.x release.

mmuetzel:
The gmsh server hang up on the MINGW64 builder:
```
 ==> Retrieving sources...
    -> Downloading gmsh-gmsh_4_10_2.tar.gz...
    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                   Dload  Upload   Total   Spent    Left  Speed
  
    0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
    0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
    0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
    0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
    0 27.5M    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
    0 27.5M    0 40960    0     0   6909      0  1:09:39  0:00:05  1:09:34  6910
    0 27.5M    0 40960    0     0   6275      0  1:16:41  0:00:06  1:16:35  7416
    0 27.5M    0 47261    0     0   6531      0  1:13:41  0:00:07  1:13:34 10947
    0 27.5M    0 63261    0     0   7556      0  1:03:41  0:00:08  1:03:33 14209
    0 27.5M    0 95261    0     0  10484      0  0:45:54  0:00:09  0:45:45 21131
    0 27.5M    0  104k    0     0  11594      0  0:41:30  0:00:09  0:41:21 20115
  curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104
  ==> ERROR: Failure while downloading https://gitlab.onelab.info/gmsh/gmsh/-/archive/gmsh_4_10_2/gmsh-gmsh_4_10_2.tar.gz
      Aborting...
```

The others finished within the 6 hours limit now.

Is there an option to re-run only the failed job while retaining the others?

mmuetzel:
Unlucky selection of runners again: The UCRT64 runner timed out this time.

But all of the environments built successfully at some point now (albeit never in the same run). Afaict, all dependencies are being rebuilt anyway with this PR. So, I hope it is ok that the package grokker never ran.

lazka:
OK to merge?

mmuetzel:
Imho, it is. 👍

stahta01:
Oops, I figured they would stay till newer ones replaced them.

Tim S.

mmuetzel:
I opened a PR that re-enables building the packages with the support libraries here: #11751

PR #11718, opened 2022-05-24
libgphoto2: enable curl

kmilos:


kmilos:
Thanks. How did it work for clang32 before?

lazka:
unclear. it crashes in doxygen somewhere, but I couldn't reliably reproduce and I got impatient :)

kmilos:
Fair enough...

lazka:
doxygen issue is now tracked here: #11787

PR #11722, opened 2022-05-25
cmake: Remove hardcoded gcc/g++ from MinGW/MSYS Makefiles

### LINKED ISSUE #11710: cmake couldn't generate "MSYS Makfiles" or "MinGW Makfiles" project on CLANG64
Reported upstream https://gitlab.kitware.com/cmake/cmake/-/issues/23542

related issue: #10800

MehdiChinoune:
Fixes #11710


MehdiChinoune:
> Does that mean it will look for clang too now? or just fail? What if clang and gcc are on PATH?
> 
> If it's gcc -> clang -> fail, then this seems fine to me.

It will work the same as Ninja:
cc->gcc->cl->bcc->xlc->icx->clang
https://gitlab.kitware.com/cmake/cmake/-/blob/v3.23.1/Modules/CMakeDetermineCCompiler.cmake#L62
c++->g++->cl->bcc->xlC->icpx->icx->clang++
https://gitlab.kitware.com/cmake/cmake/-/blob/v3.23.1/Modules/CMakeDetermineCXXCompiler.cmake#L61

lazka:
ok, thanks :)

PR #11724, opened 2022-05-25
mingw-w64-tools: set --with-widl-includedir to the default include dir

lazka:
It defaults to /ucrt64/x86_64-w64-mingw32/include otherwise, which no
longer exists.

See #11547


heljkon:
As I tried to explain in #11547 :

If *.tlb files are referenced in source *.idl, compilation won't succeed without setting "-L" to /ucrt64/include on cli as well

lazka:
Would you still consider it an "improvement" at least ? :)

heljkon:
It's definitly an improvement, because it sets the correct "-I" default.

lazka:
ok, thanks, let's do this for starters then.

PR #11728, opened 2022-05-26
openexr: WIP update to 3.1.5 New PR

Familyfirst4:
PR TO VALIDATE Ruby


Biswa96:
openexr v3 can not be built for clang msystems. Is there any difference in this PR?

PR #11734, opened 2022-05-27
gcc: fix libgomp crashing in various places

### LINKED ISSUE #11729: Conditional OpenMP directive fails w/ GCC 12
See https://github.com/msys2/MINGW-packages/issues/11730 for my minimal reproducer and https://github.com/darktable-org/darktable/issues/11889

This was previously ok w/ GCC 11.3 and is ok on Linux GCC 12.1

lazka:
It has started to mix _aligned_malloc() with normal free() which
isn't supported on Windows and crashes.

This adds some temporary fixes. Upstream might fix this differently though,
we'll see.

Fixes #11729


lazka:
fast tracking this to get a fix out

PR #11737, opened 2022-05-28
junitparser: add new package

1480c1:
Used in conjunction with google-test and other unit test frameworks that output in junit form

### DIFF
--- a/mingw-w64-junitparser/PKGBUILD
+++ b/mingw-w64-junitparser/PKGBUILD
@@ -1,0 +1,41 @@
+# Contributor: Christopher Degawa <ccom@randomderp.com>
+
+_realname=junitparser
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=2.5.0
+pkgrel=1
+pkgdesc="Pythonic JUnit/xUnit Result XML Parser (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+url="https://github.com/weiwei/junitparser"
+license=('spdx:Apache-2.0')
+depends=("${MINGW_PACKAGE_PREFIX}-python"
+         "${MINGW_PACKAGE_PREFIX}-python-future")
+makedepends=("${MINGW_PACKAGE_PREFIX}-python-setuptools")
+source=("${_realname}-${pkgver}.tar.gz::${url}/archive/refs/tags/${pkgver}.tar.gz")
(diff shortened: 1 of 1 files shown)


Biswa96:
python-future does not have arm package but it may be enabled here.

1480c1:
> python-future does not have arm package but it may be enabled here.

Interesting, any particular reason why? Is there a source incompatibility with it, or was it just not yet built for it?

Biswa96:
python-future is just not yet built for clangarm64. clangarm64 is not added for all packages like things were done in ucrt64, clang64, clang32 cases.

mingw-w64-junitparser/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+
+package() {
+  cd "${srcdir}/python-build-${MSYSTEM}"
+  MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
+    ${MINGW_PREFIX}/bin/python setup.py install --prefix=${MINGW_PREFIX} --root="${pkgdir}" -O1 --skip-build
+
+  PYTHON_EXECUTABLE="${MINGW_PREFIX}/bin/python"
+  for _f in "${pkgdir}${MINGW_PREFIX}"/bin/*-script.py; do
+    sed -e "1 { s%^#!.*$%#!${PYTHON_EXECUTABLE}%}" -i "${_f}"
+  done
```
lazka, line 39: Any reason you are not following the template here?

1480c1: none particularly, I just pulled from another package that ran setup.py, initially deleted this for loop since it stripped the py file of the shebang, then readded it but used the python executable stuff after makepkg complained about references to my root dir.

any particular file/template I should be pulling from?

1480c1: changed it to the version that removes the shebang since I realized that the scripts are primarily to be called from the executable prepending python to them

1480c1:
Seems there may need to be extra work done on this since I'm getting
```bash
$ junitparser merge --glob 'tmp/[0-9]+.xml' report.xml
failed to create process (\mingw64\bin\python "C:\msys64\mingw64\bin\junitparser-script.py" "merge" "--glob" "tmp/[0-9]+.xml" "report.xml").
```
I'm not entirely sure where to look for when trying to fix this one though

1480c1:
Well, at least it matches the print inside mingw-w64-python-setuptools/0002-Allow-usr-bin-env-in-script.patch, so that might be a start

1480c1:
> Well, at least it matches the print inside mingw-w64-python-setuptools/0002-Allow-usr-bin-env-in-script.patch, so that might be a start

seems I might have been wrong and was working with an older binary

PR #11739, opened 2022-05-28
gnutls: Update to 3.7.6

lazka:


lazka:
This release was broken, reverted until we know more.

PR #11748, opened 2022-05-30
libtiff: update to 4.4.0

kmilos:


kmilos:
Something changed in pkgconf behaviour? It now outputs `-I/jxrlib` for the freeimage patch instead of the full path...

Biswa96:
I get linking error while building freeimage with this new libtiff.

```
ld.exe: ./Source/Metadata/XTIFF.o:XTIFF.cpp:(.text+0xdfd): undefined reference to `_TIFFDataSize'
```

ArchLinux removed libtiff from freeimage dependency years ago in this commit https://github.com/archlinux/svntogit-community/commit/1b3e4087d6015def80fa3c161ef44b4c9cfe0af8. Linked with this bug report https://bugs.archlinux.org/task/58998

Biswa96:
Is it possible to add a patch to re-add `_TIFFDataSize` function which was removed in commit https://gitlab.com/libtiff/libtiff/-/commit/11f3f279608ae9e68f014717393197f430f9be58 ? But the patch will be here forever just for one dependency 😰

kmilos:
Yeah, not ideal, so this would seem to work around it instead: https://sourceforge.net/p/freeimage/discussion/36109/thread/2018fdc6e7/

I also wish the libtiff team bumped the soversion as this is clearly an ABI change :/

We'll also run into other problem w/ freeimage soon - opejpeg already throwing deprecation warnings, it probably won't build ootb w/ openexr >= 3.x, etc. Don't know if their trunk is in a better shape and switch tracking that at some point?

kmilos:
Should be good to go now. Still don't know why 'pkg-config' failed, but works now w/ `pkgconf` instead.

PR #11750, opened 2022-05-30
binutils: Move include and libs back to native directories

orgads:
They were moved to a subdirectory in 047d1d662f (binutils: resolve
conflicts with headers and libs with gdb, Dec 2013). Since then, gdb and
binutils repositories were merged, and gdb does not install these headers
and libs at all, so there is no conflict.

Linux systems have bfd.h, libbfd.a and libiberty.a in include and lib, and
not in a subdirectory.

Also, placing them in a subdirectory requires the users to specify the lib
directory, and have non-portable #include statements (binutils/bfd.h on
MinGW, else bfd.h).

Keep the original locations as symbolic links to maintain backwards
compatibility.

### DIFF
--- a/mingw-w64-binutils/PKGBUILD
+++ b/mingw-w64-binutils/PKGBUILD
@@ -6,7 +6,7 @@ _realname=binutils
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.38
-pkgrel=2
+pkgrel=3
 pkgdesc="A set of programs to assemble and manipulate binary and object files (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mati865:
The change sounds good but we don't support symlinks so `ln -s` will create copies.

orgads:
Yes, I've noticed that in the build artifacts :/

But I did see in some other packages that `ln -s` is used (`git grep 'ln -s' -- '*/PKGBUILD'`).

Do you still prefer to change it to `cp`?

orgads:
I personally don't really care. Not many applications use these headers and libraries anyway.

orgads:
Other opinions?

lazka:
If anyone ends up here because it broke things: please let us know :)

PR #11751, opened 2022-05-30
kicad: re-enable building support libraries

mmuetzel:
Also, build in clang32 environment.

See: #11699

### DIFF
--- a/mingw-w64-kicad/PKGBUILD
+++ b/mingw-w64-kicad/PKGBUILD
@@ -5,7 +5,7 @@
 # This results in only the core KiCAD package being built and reduces
 # the build time and band width usage because the support libraries
 # do not change very often and are very large.
-_build_support_libraries=no
+_build_support_libraries=yes
 # meta wants to download the other packages which waste time during
 # development; but, needs to be yes after development is over.
 _build_meta=yes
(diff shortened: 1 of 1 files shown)


stahta01:
I am working on a KICAD package change that will allow clang32 to build. So, the PR likely will not be needed.
But, not certain how soon it will be ready. It will use ninja.

Tim S.

mmuetzel:
The main reason for this PR is that the version that is currently in staging removes all support libraries. I figured that if we need a package rebuild anyway, I might as well activate clang32 now.

The license is still wrong here. And no license file is installed with the package. I'll wait until the CI has finished (to see if there are any more issues) before updating that.

stahta01:
If you need to do another PR can you change 
`depends=(${MINGW_PACKAGE_PREFIX}-${_realname})`
to
`optdepends=(${MINGW_PACKAGE_PREFIX}-${_realname})`

That was what I see ArchLinux did.

Tim S.

mmuetzel:
> If you need to do another PR can you change `depends=(${MINGW_PACKAGE_PREFIX}-${_realname})` to `optdepends=(${MINGW_PACKAGE_PREFIX}-${_realname})`
> 
> That was what I see ArchLinux did.

I assume that is for the packaging of the support libraries? Do you know what the reasoning for that is? IIUC, those support libraries are only useful when KiCAD is installed. So, it would probably be nice if pacman installed kicad when a user installs one of the support libraries.
Or does that change help to avoid having to rebuild those (very large) support libraries at every minor update?

mmuetzel:
It looks like ArchLinux moved the support libraries to a separate package rule:
https://github.com/archlinux/svntogit-community/blob/packages/kicad/trunk/PKGBUILD
and
https://github.com/archlinux/svntogit-community/blob/packages/kicad-library/trunk/PKGBUILD

That way they can probably avoid to rebuild them on each minor update of the "main" kicad package.

mmuetzel:
Force pushed with the correct license tags.

stahta01:
I can but, I think it would not be accepted because more packages seem to be against the current trend.
But, I will try it.

Tim S.

mmuetzel:
LGTM. All environments built. I installed the clang32 artifact, and the GUI opens and seems to work correctly.

mmuetzel:
> I can but, I think it would not be accepted because more packages seem to be against the current trend.

Which current trend are you referring to?
The current organization of this build rule is imho comparable to - let's say - if Python would pack the build rules for all packages into the build rule for the main python package. An update to Python might require a rebuild of dependent packages. But they are still in separate build rules (that would require a bump in that case).
The comparison isn't completely fair admittedly in that Python packages are usually less closely bundled to Python itself. But imho it's still valid.

mmuetzel:
I opened PR #11756 that splits the core application and the support libraries into two separate `PKGBUILD`.

Issue #11562, opened 2022-04-25
Debundle abseil-cpp from gprc


kmilos:
Seems there is a conflict now ATM (both install abseil-cpp objects), and looks like protobuf on main will soon depend on abseil-cpp as well...

There might be an opportunity to debundle other 3rd party libs as well (e.g. xxhash for makedepends only etc.) - benchmark is also a candidate, but it itself bundles gtest as well...

PR #11758, opened 2022-06-01
grpc: debundle abseil

kmilos:
Resolves https://github.com/msys2/MINGW-packages/issues/11562


kmilos:
Converting to draft, as there are indeed linking errors to abseil-cpp...

kmilos:
These are indeed not exported from `ucrt64/bin/libabsl_strings.dll`...

Maybe because we built abseil w/ C++17 and `std` versions are used instead?

coryan:
FWIW,  https://github.com/abseil/abseil-cpp/issues/696 may be a good read as it suggests possible workarounds.  Note that while `-DABSL_PROPAGATE_CXX_STD=ON` sounds promising, it [only requires C++11](https://github.com/abseil/abseil-cpp/issues/696#issuecomment-1042302941). Since this PR always compiles Abseil with C++17, that suggests `include/absl/options.h` should be changed with something like:

```
sed -i 's/^#define ABSL_OPTION_USE_\(.*\) 2/#define ABSL_OPTION_USE_\1 1/' "absl/base/options.h"
```

kmilos:
@coryan I have actually read it. AFAICT the conclusion was that you'd need to build any clients of Abseil w/ the same configuration without fiddling w/ the source.

coryan:
> @coryan I have actually read it. [AFAICT the conclusion was that you'd need to build any clients of Abseil w/ the same configuration without fiddling w/ the source](https://github.com/abseil/abseil-cpp/issues/696#issuecomment-636949929).

That works, it just fails with fairly incomprehensible error messages when folks do not  follow the rules.  Consider an application developer using these packages, and trying to use it with code compiled with `-std=c++11`.  Their code will compile, and then fail to link.  With the suggested change the code will fail to compile.

In other words, I am not saying that your current PR is wrong, it is a big improvement over the current state.  I am saying it could provide less confusing errors when people do the "Wrong Thing":tm:

kmilos:
Ok, let's make it somewhat *-proof then, hope this looks good now.

PR #11760, opened 2022-06-01
glpk: update to 5.0

mmuetzel:
According to package grokker, one symbol was removed with the update. So, rebuilding dependencies, too.

I took the opportunity to build python-cvxopt with OpenBLAS also for the CLANG environments. That might result in better performance on some operations.


MehdiChinoune:
> According to package grokker, one symbol was removed with the update. So, rebuilding dependencies, too.

package-grokker list the packages that should be rebuilt, removed symbols doesn't imply rebuild dependent packages.

lazka:
In theory the ABI could still change for existing symbols (say int -> size_t of some parameter or something), so it depends.

In this case the NEWS file says nothing changed, so should be good.

PR #11768, opened 2022-06-02
update python-numpy to 1.22.4 and python-numba to 0.55.2

MehdiChinoune:


mmuetzel:
Afaict, OpenBLAS is currently not built for clangarm64. It should probably be installed conditionally in the build rule for numpy.

MehdiChinoune:
> Afaict, OpenBLAS is currently not built for clangarm64. It should probably be installed conditionally in the build rule for numpy.

You are right, It was my fault when rebasing.

Issue #10515, opened 2022-01-09
Python 3.10 Update/Transition


lazka:
Note: Unrelated package updates are currently paused to focus on the Python rebuilds.

Just an issue to collect things for when the time comes.

* [x] https://github.com/msys2-contrib/cpython-mingw/issues/52
* [x] Some packages are missing .pyc files. We should look for them, and if they are versioned we can manually compile them without breaking updates. Example: https://github.com/msys2/MINGW-packages/pull/10475, python-jsonschema
* [x] rebase .pyc PRs and merge them
* [x] Rebuild everything
* [x] disable bootstrap for wheel/tomli, rebuild clang

MehdiChinoune:
Last year, Upgrading to Python 3.9 was delayed until June-2021 because of the huge job of rebasing patches and testing.
Then a new repository - https://github.com/msys2-contrib/cpython-mingw - was created to not make it happen again (That's my understanding).
Was that goal achieved? I don't think so.
I think that was because of the approach chosen, waiting until final release and rebase/chery-pick patches. That work should started with alpha release.
For example with Qt6 I start test packaging it with beta releases.
The same apply to other essential/huge projects like LLVM, GCC. We should start testing them with beta releases (Maybe even better by forking them and setup a scheduled-GHA to rebase our patches)

lazka:
Packages missing bytecode files:

```
mingw-w64-adios2
mingw-w64-d-feet
mingw-w64-gedit
mingw-w64-geographiclib
mingw-w64-gexiv2
mingw-w64-gitg
mingw-w64-graphviz
mingw-w64-gst-editing-services
mingw-w64-gst-python
mingw-w64-kicad
mingw-w64-libbotan
mingw-w64-libcaca
mingw-w64-libftdi
mingw-w64-libgit2-glib
mingw-w64-libproxy
mingw-w64-libsbml
mingw-w64-libvoikko
mingw-w64-libxml2
mingw-w64-mlt
mingw-w64-nlopt
mingw-w64-openturns
mingw-w64-partio
mingw-w64-plplot
mingw-w64-pyside2-qt5
mingw-w64-pyside6
mingw-w64-python-entrypoints
mingw-w64-pygobject
mingw-w64-python-ipykernel
mingw-w64-python-ipython
mingw-w64-python-jsonschema
mingw-w64-python-jupyter_client
mingw-w64-python-jupyter_core
mingw-w64-python-maturin
mingw-w64-python-mistune
[...]

PR #11773, opened 2022-06-03
python-qmk: fix launcher crash

fauxpark:
Attempting to run `qmk` results in `Fatal error in launcher: U`. The appended Python script had the CI cygpath in its hashbang, but for some reason `makepkg-mingw` didn't pick up on it. I referenced python-jsonschema which also has an .exe launcher.

### DIFF
--- a/mingw-w64-python-qmk/PKGBUILD
+++ b/mingw-w64-python-qmk/PKGBUILD
@@ -47,7 +47,7 @@ package() {
   cd "${srcdir}/python-build-${MSYSTEM}"
 
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
-    "${MINGW_PREFIX}"/bin/python -m installer --no-compile-bytecode --prefix=${MINGW_PREFIX} \
+    "${MINGW_PREFIX}"/bin/python -m installer --prefix=${MINGW_PREFIX} \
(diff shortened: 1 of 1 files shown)


mingw-w64-python-qmk/PKGBUILD
```diff
@@ -47,13 +47,8 @@ package() {
...
 
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
-    "${MINGW_PREFIX}"/bin/python -m pip install dist/*.whl \
-      --compile \
-      --no-deps \
-      --no-warn-script-location \
-      --ignore-installed \
-      --prefix="${MINGW_PREFIX}" \
-      --root="${pkgdir}"
+    "${MINGW_PREFIX}"/bin/python -m installer --no-compile-bytecode --prefix=${MINGW_PREFIX} \
```
lazka, line 50: Please remove `--no-compile-bytecode`

fauxpark: `python-jsonschema` also has this. Should it be removed there too?

lazka: No, removing for a published packages will break updates since the user will have generated those .pyc files locally and pacman will error out due to file conflicts.

I've already noted jsonschema in #10515

PR #11778, opened 2022-06-04
Standarize GNU AVR toolchain PKGBUILDs

mati865:
Do we need to bump pkgrel to produce missing `ucrt64` build of `mingw-w64-avr-gcc`?
It build fine on my machine.

Disabled `mingw-w64-avr-gcc` for Clang because it fork-bombed my machine by recursively spawning tens of thousands compiler processes during: `checking whether compiler driver understands Ada...`


mmuetzel:
Need to add `"${MINGW_PACKAGE_PREFIX}-readline"` (and possibly more?) to the dependencies of `mingw-w64-avr-gdb`?
See also the list of dependencies of [`mingw-w64-gdb`](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gdb/PKGBUILD#L16).

PR #11792, opened 2022-06-05
scc: new package

Freed-Wu:


mingw-w64-scc/PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: Wu Zhenyu <wuzy01@qq.com>
+
+realname=scc
+pkgbase=mingw-w64-$realname
+pkgname=$MINGW_PACKAGE_PREFIX-$realname
+pkgver=3.0.0
+pkgrel=2
```
naveen521kk, line 7: Start with pkgrel=1
```suggestion
pkgrel=1
```

Freed-Wu:
Thanks!

mingw-w64-scc/PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: Wu Zhenyu <wuzy01@qq.com>
+
+realname=scc
```
naveen521kk, line 3: It's better to start this variable with an underscore (`_`) and change its reference. Since Archlinux doesn't define this variable it's better to mark it private as it's done on the templates.
```suggestion
_realname=scc
```

naveen521kk (commented):
Also, does the resulting binary work? I'm wondering because this is a `go` package but the strip option isn't disabled. Previously, `strip` would create binaries which don't properly work on `go` packages, for example, 
https://github.com/msys2/MINGW-packages/blob/ae86ece965ac45004cbb76085af396b4e8db878e/mingw-w64-github-cli/PKGBUILD#L16

PR #11793, opened 2022-06-05
scrcpy: new package

Freed-Wu:
Refer <https://github.com/Genymobile/scrcpy/blob/master/BUILD.md>

### DIFF
--- a/mingw-w64-scrcpy/PKGBUILD
+++ b/mingw-w64-scrcpy/PKGBUILD
@@ -1,38 +1,48 @@
 # Maintainer: Wu Zhenyu <wuzy01@qq.com>
 
-realname=scrcpy
-pkgbase=mingw-w64-$realname
-pkgname=$MINGW_PACKAGE_PREFIX-$realname
+_realname=scrcpy
+pkgbase="mingw-w64-${_realname}"
+pkgname="$MINGW_PACKAGE_PREFIX-${_realname}"
 pkgver=1.23
 pkgrel=1
 pkgdesc='Display and control your Android device (mingw-w64)'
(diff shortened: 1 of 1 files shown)


Freed-Wu:
chanage b2sums to sha256sums

Biswa96:
I'm not sure if this is right way to provide a package. Two signifact parts are/can not provided as mingw package. The adb client program and scrcpy server written in Java.

Freed-Wu:
I refer the arch's PKGBUILD, maybe they use a pre-built scrcpy-server not build it because of convenience. About `adb`, prehaps it can be built by mingw, although I have not tried.

MehdiChinoune:
It couldn't be used without adb!

Biswa96:
I have compiled adb with mingw gcc and clang partially. See this project here https://github.com/Biswa96/android-tools-win. The only catch is that DLLs require atlmfc library which can be only compiled with msvc. Also the library is license restricted. So, one has to reimplement it in mingw-w64 without looking at the actual code.

MehdiChinoune:
A package that's not usable at all shouldn't be added at first place!
I still wondering why was it merged?

driver1998:
> I have compiled adb with mingw gcc and clang partially. See this project here https://github.com/Biswa96/android-tools-win. The only catch is that DLLs require atlmfc library which can be only compiled with msvc. Also the library is license restricted. So, one has to reimplement it in mingw-w64 without looking at the actual code.

I would rather update adb to not use ATL at this point... Not sure what it is used for, cppwinrt and wil are sutiable alternatives if you just need some COM wrappers.

PR #11794, opened 2022-06-05
msmpi: fix static linking issue due to missing plain .a library

okhlybov:


Biswa96:
Is it possible to compile the static library from msmpi source repository? If yes I can try it to include in this package.

okhlybov:
I think it is not possible since MS itself does not ship static version with the MS-MPI SDK.

mmuetzel:
Is this change correct? That library isn't a "real" static library. IIUC, it *is* an import library. Linking to it with `-static` will still require the *shared* MPI library on runtime.
Instead of "lying" to the user that they successfully linked to a static library, it would be more "honest" to error out in that case imho.

mmuetzel:
Couldn't you link with something like `-static -lother_static_lib1 -shared -lmsmpi -lother_shared_lib -static -lother_static_lib2`? IIUC, that would be the "correct" way to link with msmpi while linking statically with some other libraries.

mmuetzel:
Build error of PETSc in staging:
https://github.com/msys2/msys2-autobuild/runs/6750781686?check_suite_focus=true#step:10:21204
```
  TESTING: checkMPIDistro from config.packages.MPI(/c/M/mingw-w64-petsc/src/build-MINGW64/petsc-3.17.2/config/BuildSystem/config/packages/MPI.py:660)
  TESTING: configureMPI2 from config.packages.MPI(/c/M/mingw-w64-petsc/src/build-MINGW64/petsc-3.17.2/config/BuildSystem/config/packages/MPI.py:381)
  *******************************************************************************
           UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
  -------------------------------------------------------------------------------
  PETSc requires some of the MPI-2.0 (1997), MPI-2.1 (2008) functions - they are not available with the specified MPI library
  *******************************************************************************
  
      Aborting...
```

okhlybov:
> Couldn't you link with something like `-static -lother_static_lib1 -shared -lmsmpi -lother_shared_lib -static -lother_static_lib2`? IIUC, that would be the "correct" way to link with msmpi while linking statically with some other libraries.

This requires intricate knowledge about *libmsmpi* being the dynamic only which might confuse build systems.
This is less of a problem when using the `mpi??` drivers, though.

okhlybov:
> Build error of PETSc in staging: https://github.com/msys2/msys2-autobuild/runs/6750781686?check_suite_focus=true#step:10:21204
> 
> ```
>   TESTING: checkMPIDistro from config.packages.MPI(/c/M/mingw-w64-petsc/src/build-MINGW64/petsc-3.17.2/config/BuildSystem/config/packages/MPI.py:660)
>   TESTING: configureMPI2 from config.packages.MPI(/c/M/mingw-w64-petsc/src/build-MINGW64/petsc-3.17.2/config/BuildSystem/config/packages/MPI.py:381)
>   *******************************************************************************
>            UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
>   -------------------------------------------------------------------------------
>   PETSc requires some of the MPI-2.0 (1997), MPI-2.1 (2008) functions - they are not available with the specified MPI library
>   *******************************************************************************
>   
>   ==> ERROR: A failure occurred in build().
>       Aborting...
> ```

Confirmed.

It is the thin archive that is the cause of this failure. Revering to plain copying solves the problem.

This is strange as thin archive feature used to be reliable.

lazka:
reverted in b54c1be56ab4e299c btw

PR #11796, opened 2022-06-05
petsc: version bump

okhlybov:


MehdiChinoune:
What's the problem with MPI? According to integration tests, the Clang's msmpi package is intact.
Also, there is a Clang's openblas package. Does it work?
The only thing thats presumably missing is a FORTRAN support.

PR #11803, opened 2022-06-07
petsc: build system unification fixes

okhlybov:


mingw-w64-petsc/0001-mpi-detection-override.patch
```diff
@@ -9,6 +10,15 @@
      self.includes           = ['mpi.h']
      liblist_mpich         = [['fmpich2.lib','fmpich2g.lib','fmpich2s.lib','mpi.lib'],
                               ['fmpich2.lib','fmpich2g.lib','mpi.lib'],['fmpich2.lib','mpich2.lib'],
+@@ -383,7 +383,7 @@
+     oldFlags = self.compilers.CPPFLAGS
+     oldLibs  = self.compilers.LIBS
+     self.compilers.CPPFLAGS += ' '+self.headers.toString(self.include)
+-    self.compilers.LIBS = self.libraries.toString(self.lib)+' '+self.compilers.LIBS
++    self.compilers.LIBS = self.libraries.toString(self.lib)+' '+self.compilers.LIBS+' -lmsmpi'
```
mmuetzel, line 18: It probably doesn't matter because this change is for the clang environments that use `lld`. But it might be more logical to move the library next to the ones proper to this package:
``` diff
-    self.compilers.LIBS = self.libraries.toString(self.lib)+' '+self.compilers.LIBS
+    self.compilers.LIBS = self.libraries.toString(self.lib)+' -lmsmpi '+self.compilers.LIBS
 ```

Does that also work?

File: mingw-w64-petsc/0001-mpi-detection-override.patch [hunk 2]
```diff
@@ -18,3 +28,25 @@
...
      if self.checkLink('#include <mpi.h>\n', 'int provided; if (MPI_Init_thread(0,0,MPI_THREAD_FUNNELED,&provided)) return 0;'): # MPI-2.1
        self.addDefine('HAVE_MPI_INIT_THREAD',1)
+@@ -603,8 +605,8 @@
+     self.log.write('Checking for header mpi.h\n')
+     # check if MPI_Finalize from c++ exists
+     self.log.write('Checking for C++ MPI_Finalize()\n')
+-    if not self.libraries.check(self.lib, 'MPI_Finalize', prototype = '#include <mpi.h>', call = 'int ierr;\nierr = MPI_Finalize();', cxxMangle = 1):
+-      raise RuntimeError('C++ error! MPI_Finalize() could not be located!')
++    #if not self.libraries.check(self.lib, 'MPI_Finalize', prototype = '#include <mpi.h>', call = 'int ierr;\nierr = MPI_Finalize();', cxxMangle = 1):
++    #  raise RuntimeError('C++ error! MPI_Finalize() could not be located!')
```
mmuetzel, line 38: Why do these tests fail (this and the next one that you removed)?
Do you have access to the logs with the result of these checks?

mingw-w64-petsc/PKGBUILD
```diff
@@ -114,19 +112,24 @@ _petsc() {
   desc=
   case $1 in
     ?m?)
-      opts+=" --with-mpi=1 --with-mpi-compilers=1 --with-pthread=0 --with-openmp=0 --with-metis=1 --with-parmetis=1"
+	  if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]]; then
+		opts+=" --with-mpi-compilers=1"
+	  else
+		opts+=" --with-mpi-compilers=0"
```
mmuetzel, line 118: Are the mpi-compilers not working correctly in the clang*-environments?
If so, why are they failing?

okhlybov:
These are the empirical hacks to get through the PETSc build system. Perhaps not pretty yet working whereas the PETSc's defaults are not.

After all, we can sustain some hardcoding since we know the environment we are building for.

As for the clang's `mpi**` drivers there some arcane interaction between the compiler and the PETSc build system that is yet to be  recognized. More specifically, the GCC's `mpicc` is accepted while the Clang's is not with multiple complaints about preprocessing test failures etc. even though both `mpicc` are compiled from exactly the same source.

mmuetzel:
Sorry. The Fortran modules for the MPI variant didn't compile with that change.
See, e.g., here for a possible changeset that seems to work:
https://github.com/mmuetzel/MINGW-packages/commit/465a22ad5b94f09eca469f08199cad1432fb8bbd

mmuetzel:
The latest changes look good to me. 👍

okhlybov:
Yep. Just need to make a run of the integration tests, though.

mmuetzel:
When this is done, you should probably check if this breaks the clang builds of SLEPc.

mingw-w64-petsc/PKGBUILD
```diff
@@ -90,44 +90,44 @@ prepare() {
...
 
 _petsc() {
+  desc=
   opts="--with-single-library=1 --disable-shared --with-windows-graphics=0 --with-x=0 --with-hwloc=1 --with-openblas=1 --with-openblas-dir=$MINGW_PREFIX"
   pc="hwloc openblas"
+  iflags=
   cflags="$CFLAGS"
   cxxflags="$CXXFLAGS"
-  fflags="$CFLAGS"
+  fflags="$CFLAGS -fallow-invalid-boz -fallow-argument-mismatch -I$MINGW_PREFIX/include"
```
MehdiChinoune, line 105: Are those flags required here?
I don't think so.
These flags are required with legacy Fortran code to get it compiled with gfortran>=11
Some projects used them as a workaround before fixing their code.
PETSc doesn't require them.

okhlybov:
> When this is done, you should probably check if this breaks the clang builds of SLEPc.

The local Clang builds of SLEPc are OK, but this needs to be re-verified with the CI builds of PETSc.

mingw-w64-petsc/PKGBUILD
okhlybov: These flags are required to get past the PETSc's MPI configuration otherwise the Python code has to be fixed as it was proposed initially.

PR #11805, opened 2022-06-08
Update lcov to 1.16

GitMensch:
restoring it, now using MINGW_PREFIX in perl interpreter

### DIFF
--- a/mingw-w64-lcov/001-mingw-lcov1.12-handle-equals-signs.patch
+++ b/mingw-w64-lcov/001-mingw-lcov1.12-handle-equals-signs.patch
@@ -1,0 +1,23 @@
+Description: Handle "=====" as another form of zero.
+ gcov prints "=====" instead of "######" when an unexecuted line is
+ "reachable only by exceptional paths such as C++ exception handlers."
+ This should be handled the same as "######" for our purposes.
+Author: Zack Weinberg <zackw@panix.com>
+Last-Update: 2013-02-01
+
+Index: lcov-1.12/bin/geninfo
+===================================================================
+--- lcov-1.12.orig/bin/geninfo
++++ lcov-1.12/bin/geninfo
+@@ -1771,8 +1771,9 @@ sub read_gcov_file($)
+ 				$number = (split(" ",substr($_, 0, 16)))[0];
+ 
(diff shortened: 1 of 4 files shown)


GitMensch:
It is a bit bad that we need one package for each of MINGW64,MINGW32,MSYS2,... - not because there is anything specific in this package (but the perl shebang - but that could be replaced with "just perl"), but because of the port specific perl dependency.

GitMensch:
Sight, rechecked with resulting version - the MinGW perl does not work without a patch:

~~~
lcov --capture --directory . --output-file coverage.info
Capturing coverage data from .
The command "D:/msys64/mingw64/bin/geninfo" is not found.
~~~

So we need to patch lcov to execute either "perl" or "bash" instead of the geninfo "command" (as its shebang is not understood by the Win32 native environment).

GitMensch:
Tests working now, to make this work lcov had to be adjusted, which was published upstream with https://github.com/linux-test-project/lcov/pull/155 and then integrated into this PR.

lazka:
Needs a rebase I guess?

re patch: puh, I'm not very keen on maintaining that patch downstream, but I guess we can always drop it again and fall back to msys lcov.

GitMensch:
No rebase, but a "special merge" on the command line. I wanted to preserve the history so explicit branched the old version before delete.

I actually have hope that this patch will be included in the next lcov version and therefore not necessary then.

lazka:
Does this mean you expect me to merge it manually? :)

GitMensch:
I actually do not know how to tell git "yes, drop this deletion" and somehow I can't do it via browser.
If no one has an idea how this should be solved I can of course also only add the files "as new", but that felt wrong.

GitMensch:
Thank you for taking care of this!

lazka:
just wondering, have you considered https://gcovr.com/en/stable/index.html as an alternative?

GitMensch:
I wasn't aware of gcovr - is there a package for it already?
Note: as lcov is integrated into exiting packages (it looks like the command line options of gcovr are not compatible so just swapping the "the to execute" wouldn't be enough) it would still be reasonable to have a working package in MinGW for lcov :-)

lazka:
ok, makes sense.

PR #11809, opened 2022-06-09
msmpi: static linking fix

okhlybov:
A follow-up to #11794.

1. Thin archives tend to be fragile.
   - Library stripping destroys thin archives.
   - LLVM's AR seems to produce broken (empty) thin archives.

2. Fixing the `mpi**` drivers is not an option as *-Bdynamic* is a command line -wise option which affects all libraries that follow it, either explicitly specified or implicitly added by the linker. This effectively defeats production of statically linked MPI binaries with `mpi** -static` without severe drivers modifications.

3. No way to use *-Bdynamic* in .pc files for the same reason as in [2].

~~The bottom line: the best option is simply to have `.a` and `.dll.a` archives with the same contents and to let compiler pick the appropriate version based on the supplied options.~~

**UPD**

The `-l:` solution works for both GCC and Clang.

### DIFF
--- a/mingw-w64-msmpi/PKGBUILD
+++ b/mingw-w64-msmpi/PKGBUILD
@@ -4,7 +4,7 @@ _realname=msmpi
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=10.1.1
-pkgrel=8
+pkgrel=9
 pkgdesc="Microsoft MPI SDK (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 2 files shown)


mmuetzel:
I haven't tried that. But would it be possible to add `-l:libmsmpi.dll.a` to the linker flags in the .pc file as an alternative?

okhlybov:
> I haven't tried that. But would it be possible to add `-l:libmsmpi.dll.a` to the linker flags in the .pc file (and the compiler wrappers) as an alternative?

Is `-l: ` a valid option? This could be the solution.

mmuetzel:
> Is `-l: ` a valid option? This could be the solution.

It's documented for `ld`:
https://linux.die.net/man/1/ld
> -l namespec
--library=namespec
Add the archive or object file specified by namespec to the list of files to link. This option may be used any number of times. If namespec is of the form :filename, ld will search the library path for a file called filename, otherwise it will search the library path for a file called libnamespec.a.

I don't know if it works for `lld`, too. But it might because in general they aim to be command line compatible IIUC...

mmuetzel:
LGTM 👍

PR #11810, opened 2022-06-09
gmsh: update to 4.10.3

mmuetzel:


Biswa96:
I'm not sure about that. The website of the project suggests to download the source code from this link `https://gmsh.info/src/gmsh-4.10.3-source.tgz`. And it would not require the `_pkgver2` variable.

mmuetzel:
Thanks for checking. I "just" updated the version and checked whether the updated build was still working.
I didn't notice that there is an alternative or newer official source for the tarball.

Updated with the source from their website.

PR #11813, opened 2022-06-09
graphviz: drop gtk2 dependency

lazka:
We want to get rid of gtk2 eventually


lazka:
If someone needs it please complain and we can add it back as a split plugin/package so it doesn't get pulled in by default at least.

PR #11825, opened 2022-06-11
Python: Use `-O3` level optimization

wszqkzqk:
Python now is compiled with `-O2` level optimization, which makes it obviously slower than official python compiled by MSVC with the max optimization level.

For a simple test:
``` python
from timeit import default_timer as timer
import os

def plus():
    a = 0
    for x in range(100000000):
        a += 1

def minus():
    a = 100000000
    for x in range(100000000):
        a -= 1

def multiply():
    a = 1
    for x in range(100000000):
        a *= 1.00000003

def divide():
    a = 1000000000
    for x in range(100000000):
        a /= 1.00000003

if __name__ == '__main__':
    s = timer()
    plus()
    e = timer()
    print('Plus:',round(100000000/(e-s)/1000000,2),'M/s')
    s = timer()
    minus()
    e = timer()
    print('Minus:',round(100000000/(e-s)/1000000,2),'M/s')
    s = timer()
    multiply()
    e = timer()
    print('Multiply:',round(100000000/(e-s)/1000000,2),'M/s')
    s = timer()
    divide()
    e = timer()
    print('Divide:',round(100000000/(e-s)/1000000,2),'M/s')
    
    input("\nPlease press \"Enter\" key to exit . . .")
```

Official python built with MSVC:
``` powershell
Plus: 39.06 M/s
Minus: 36.23 M/s
Multiply: 36.65 M/s
Divide: 38.08 M/s

Please press "Enter" key to exit . . .
```

Python in MSYS2(UCRT64) with `-O2` level optimization:
``` powershell
Plus: 32.37 M/s
Minus: 32.86 M/s
Multiply: 32.84 M/s
Divide: 32.52 M/s

Please press "Enter" key to exit . . .
```

Python in MSYS2(UCRT64) with `-O3` level optimization:
``` powershell
Plus: 38.3 M/s
Minus: 37.98 M/s
Multiply: 41.9 M/s
Divide: 38.33 M/s

Please press "Enter" key to exit . . .
```

And **archlinux** also used [**`-O3`** level optimization](https://github.com/archlinux/svntogit-packages/blob/dff680229ec36d8a4a2fe75fb225c8ae94142806/trunk/PKGBUILD#L43), so we may change to `-O3` optimization for performance.

### DIFF
--- a/mingw-w64-python/PKGBUILD
+++ b/mingw-w64-python/PKGBUILD
@@ -355,7 +355,7 @@ build() {
     _extra_config+=("--enable-optimizations")
     # PGO should be done with -O3
     CFLAGS="${CFLAGS/-O2/-O3}"
-    CFLAGS+=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang* ]] || echo " -ffat-lto-objects")
+    # CFLAGS+=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang* ]] || echo " -ffat-lto-objects")
     # FIXME: https://github.com/msys2-contrib/cpython-mingw/issues/10
     # _extra_config+=("--with-lto")
   else
--- a/mingw-w64-python/PKGBUILD
+++ b/mingw-w64-python/PKGBUILD
@@ -353,8 +353,10 @@ build() {
   declare -a _extra_config
   if check_option "debug" "n"; then
     _extra_config+=("--enable-optimizations")
+
     # PGO should be done with -O3
-    CFLAGS="${CFLAGS/-O2/-O3}"
+    # CFLAGS="${CFLAGS/-O2/-O3}"
+    CFLAGS+=" -O3"
     # CFLAGS+=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang* ]] || echo " -ffat-lto-objects")
     # FIXME: https://github.com/msys2-contrib/cpython-mingw/issues/10
     # _extra_config+=("--with-lto")
--- a/mingw-w64-python/PKGBUILD
+++ b/mingw-w64-python/PKGBUILD
@@ -353,11 +353,8 @@ build() {
   declare -a _extra_config
   if check_option "debug" "n"; then
     _extra_config+=("--enable-optimizations")
-
-    # PGO should be done with -O3
-    # CFLAGS="${CFLAGS/-O2/-O3}"
+    # Upstream defaults to -O3, so we can do too
     CFLAGS+=" -O3"
-    # CFLAGS+=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang* ]] || echo " -ffat-lto-objects")
     # FIXME: https://github.com/msys2-contrib/cpython-mingw/issues/10
     # _extra_config+=("--with-lto")
   else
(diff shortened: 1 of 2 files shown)


mingw-w64-python/PKGBUILD
```diff
@@ -353,6 +353,9 @@ build() {
   declare -a _extra_config
   if check_option "debug" "n"; then
     _extra_config+=("--enable-optimizations")
+    # PGO should be done with -O3
+    CFLAGS="${CFLAGS/-O2/-O3}"
```
lazka, line 357: just appending -O3 should work too

mingw-w64-python/PKGBUILD
```diff
@@ -353,6 +353,9 @@ build() {
...
+    CFLAGS+=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang* ]] || echo " -ffat-lto-objects")
```
lazka, line 358: Does that have an effect since we don't do lto right now?

wszqkzqk: I've not found any problems caused by adding CFLAG `-ffat-lto-objects`. Maybe my tests are not enough.

MehdiChinoune: LTO could be enabled by adding `lto` to `options`, something like `$([[ "${MSYSTEM}" != "CLANG64" ]] || echo 'lto')`

wszqkzqk: > LTO could be enabled by adding `lto` to `options`, something like `$([[ "${MSYSTEM}" != "CLANG64" ]] || echo 'lto')`

I just followed what archlinux did to avoid unknown problems.

wszqkzqk: > Does that have an effect since we don't do lto right now?

It seems that performance drop after removing CFLAG `-ffat-lto-objects` is slight.

`-ffat-lto-objects` removed:
``` powershell
Plus: 37.84 M/s
Minus: 37.86 M/s
Multiply: 41.82 M/s
Divide: 36.76 M/s

Please press "Enter" key to exit . . **.**
```
`-ffat-lto-objects` added:
``` powershell
Plus: 38.3 M/s
Minus: 37.98 M/s
Multiply: 41.9 M/s
Divide: 38.33 M/s

Please press "Enter" key to exit . . .
```

mingw-w64-python/PKGBUILD
wszqkzqk: > just appending -O3 should work too

But is it confusing?😖

wszqkzqk: > just appending -O3 should work too
> 
> sounds good though..

OK, I changed it.

naveen521kk:
I think this PR should wait until the Python 3.10 update rebuild is done(see https://packages.msys2.org/queue). Also, lazka is unavailable atm and will soon review this PR.

mingw-w64-python/PKGBUILD
lazka: looks like it uses it by default anyway, so need to touch this: https://github.com/python/cpython/blob/71354adff07f8beba8374767532bb9da34546e66/configure.ac#L1830

PR #11829, opened 2022-06-11
binutils: add prefixed executables

metab0t:
Fixed https://github.com/msys2/MSYS2-packages/issues/2595 by copying executables and add prefix.


lazka:
As I said in the linked issue I'm not convinced that we need this since we don't provide a cross compiler. Unless there is another reason I'm not getting (??)

PR #11831, opened 2022-06-11
wxWidgets3.1: Add new wx/filedlgcustomize.h to the list of files installed

stahta01:
My update to wxPython is complaining about missing wx/filedlgcustomize.h header.
And, wxWidgets git just added a fix that I am working on testing; still a few hours before test is finished.
The patch file is the git commit that is supposed to add the header.

Tim S.


Biswa96:
If the patch is from upstream the raw patch from that git commit can be applied instead of copying it here.

stahta01:
> If the patch is from upstream the raw patch from that git commit can be applied instead of copying it here.

I have no idea what you mean. Tim S.

Biswa96:
Just saying. It's optional. Can you add the commit link as a comment above the patch command? It'll help to track the patch in future.

stahta01:
I did what you suggested and used the git commit as a diff; I did not really see any advantage or any disadvantage.

Tim S.

stahta01:
See also #11835 it includes wxPython package update.

PR #11836, opened 2022-06-12
python-llvmlite: update to 0.38.1

mmuetzel:
Use (modified) upstream patch to fix compilation with LLVM 14.

The modifications to the upstream patch are for style changes in the repository to make it apply on the release tarball.


Biswa96:
The patch file seems to be too large. Can we wait for next release?

lazka:
We need this to get the package to build at all (assuming the patch works).

In the long run we might need some old llvm package for this like in Arch Linux.

PR #11837, opened 2022-06-13
prjtrellis: rebuild

lazka:


mmuetzel:
I'm seeing the same `cc1plus.exe: out of memory allocating 65536 bytes` error when trying to build in the mingw32 environment locally.
Not sure what is causing this. (Maybe extensive use of templates?)
Even less of an idea how to fix it.

The only package that requires this is `nextpnr`.
Should we just stop building it for MINGW32?

mmuetzel:
See #11849 for a possible workaround.

PR #11838, opened 2022-06-13
opencolorio: rebuild

lazka:


mmuetzel:
It looks like it is missing `-lOpenImageIO_Util` in the linker command.

lazka:
See #11846

PR #11839, opened 2022-06-13
precice: rebuild

lazka:
```
2022-06-12T16:25:47.6180968Z -- Could NOT find MPI_CXX (missing: MPI_:libmsmpi_LIBRARY MPI_-l:libmsmpi_LIBRARY MPI_CXX_WORKS) 
2022-06-12T16:25:47.6184715Z CMake Error at D:/a/msys64/mingw64/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
2022-06-12T16:25:47.6185591Z   Could NOT find MPI (missing: MPI_CXX_FOUND)
2022-06-12T16:25:47.6186245Z Call Stack (most recent call first):
2022-06-12T16:25:47.6186966Z   D:/a/msys64/mingw64/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
2022-06-12T16:25:47.6187771Z   D:/a/msys64/mingw64/share/cmake/Modules/FindMPI.cmake:1830 (find_package_handle_standard_args)
2022-06-12T16:25:47.6188552Z   CMakeLists.txt:193 (find_package)
```

PR #11844, opened 2022-06-13
cmake: fix FindMPI module

mmuetzel:
`cmake`'s `FindMPI` module doesn't correctly handle linker flags starting with `-l:`.
The proposed change could be used until that is properly fixed upstream:
https://gitlab.kitware.com/cmake/cmake/-/issues/23620

See also #11839.


mmuetzel:
These changes are now part of PR #11845.

PR #11845, opened 2022-06-13
precice: fix building with gcc 12

mmuetzel:
File streams are no longer implicitly defined when including `<istream>`. Include the correct header instead.

~Depends on #11844 for MPI.~ #11844 is included in this PR now.

Fixes #11839.


Biswa96:
As this depends on MPI fix, that commit can be added here. We can verify the build from the resultant CI logs.

mmuetzel:
Error 503. Server down? Or too many requests?

mmuetzel:
Seems to be building correctly now.

PR #11846, opened 2022-06-13
opencolorio: fix linking issues

mmuetzel:
See #11838.


mmuetzel:
I noticed a little late that openimageio is conditioned out as a dependency for clang environments. That's probably no longer necessary.
Should we remove that condition now? Or better wait until the Python update has settled down?

mmuetzel:
Removed the condition for the OpenImageIO dependency.

PR #11840, opened 2022-06-13
partio: rebuild

lazka:
```
2022-06-12T04:42:57.7796269Z FAILED: src/tools/partview.exe 
2022-06-12T04:42:57.7806546Z cmd.exe /C "cd . && D:\a\msys64\mingw32\bin\g++.exe -march=pentium4 -mtune=generic -O2 -pipe -O3 -DNDEBUG -pipe -Wl,--no-seh    -LD:/a/msys64/mingw32/lib -lfreeglut src/tools/CMakeFiles/partview.dir/partview.cpp.obj -o src\tools\partview.exe -Wl,--out-implib,src\tools\libpartview.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -LC:/M/mingw-w64-partio/src/build-MINGW32/src/lib src/lib/libpartio.dll.a  D:/a/msys64/mingw32/lib/libz.dll.a  -lglu32  -lopengl32  D:/a/msys64/mingw32/lib/libz.dll.a  -lfreeglut  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
2022-06-12T04:42:57.7808424Z D:/a/msys64/mingw32/lib/gcc/i686-w64-mingw32/12.1.0/../../../../i686-w64-mingw32/bin/ld.exe: src/tools/CMakeFiles/partview.dir/partview.cpp.obj:partview.cpp:(.text+0x3b2): undefined reference to `glutTimerFunc'
2022-06-12T04:42:57.7809378Z D:/a/msys64/mingw32/lib/gcc/i686-w64-mingw32/12.1.0/../../../../i686-w64-mingw32/bin/ld.exe: src/tools/CMakeFiles/partview.dir/partview.cpp.obj:partview.cpp:(.text+0x3b7): undefined reference to `glutPostRedisplay'
2022-06-12T04:42:57.7810301Z D:/a/msys64/mingw32/lib/gcc/i686-w64-mingw32/12.1.0/../../../../i686-w64-mingw32/bin/ld.exe: src/tools/CMakeFiles/partview.dir/partview.cpp.obj:partview.cpp:(.text+0x3c3): undefined reference to `glutGet'
[...]

PR #11847, opened 2022-06-13
freeglut: fix static flags in .pc file

mmuetzel:
Should be fixing #11840.


lazka:
Thanks!

pal1000:
@mmuetzel can you do something like this for mingw-w64-libsystre regex.pc. See #8598.

PR #11849, opened 2022-06-13
prjtrellis: no longer build Python library for mingw32

mmuetzel:
Build from release tarball.

The memory footprint of `cc1plus` is too large during compilation of `PyTrellis.cpp`. It fails with `cc1plus.exe: out of memory allocating 65536 bytes`. Maybe something in GCC (or Boost?) changed since this was last compiled that makes it require too much memory now. When building in a MINGW64 environment, the memory footprint peaks at about 4.1 GiB. IIUC, the maximum memory footprint for one 32-bit process on Windows is 2 GiB. If the 32-bit executable would require approx. half as much memory as the 64-bit one, that would still be too much (and it looks like it *is* too much).

See also #11837.

@umarcor: Is this change ok?
Does it make sense to skip building the Python library on mingw32 and continue building the remainder? Is that still useful?

### DIFF
--- a/mingw-w64-prjtrellis/PKGBUILD
+++ b/mingw-w64-prjtrellis/PKGBUILD
@@ -3,13 +3,13 @@
 _realname=prjtrellis
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.r1033.3ae21cf
-pkgrel=2
+pkgver=1.2.1
+pkgrel=1
 pkgdesc="Documenting the Lattice ECP5 bit-stream format (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url="https://github.com/YosysHQ/prjtrellis"
-license=('MIT')
+license=('spdx:MIT')
 groups=("${MINGW_PACKAGE_PREFIX}-eda")
 depends=(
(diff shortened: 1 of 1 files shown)


umarcor:
Package prjtrellis is not available on MINGW32 (it never was), precisely because of the memory issues when running the build process on an environment with a limited memory space. I did try to cheat by building the databases on MINGW64, regardless of the build being on/for MINGW32. Unfortunately, I could not get it to work: https://github.com/msys2/MINGW-packages/pull/7568.

Therefore, it is ok to keep not building for MINGW32. However, it seems that this PR includes multiple unrelated but also interesting enhancements. For instance, using Ninja or installing through cmake instead of make. Could you please split those to separated commits and open a PR in [msys2/MINGW-packages](https://github.com/msys2/MINGW-packages)?

mmuetzel:
> Package prjtrellis is not available on MINGW32 (it never was), precisely because of the memory issues when running the build process on an environment with a limited memory space.

I'm not sure I understand. The last time prjtrellis was built, it seems to have built successfully on MINGW32. At least, that package is available in the repository:
https://packages.msys2.org/package/mingw-w64-i686-prjtrellis?repo=mingw32

> Therefore, it is ok to keep not building for MINGW32.

To clarify: This PR doesn't propose to stop building prjtrellis entirely for MINGW32. It just proposes to stop building the failing Python module. The remainder of the package seems to still build fine for MINGW32.

> Could you please split those to separated commits?

Done.
This will hopefully make it a bit easier to review these changes.

PR #11850, opened 2022-06-13
llvm: provide gcov on CLANG*

MehdiChinoune:
lcov expects gcov to be in PATH.


lazka:
Because I looked it up: https://llvm.org/docs/CommandGuide/llvm-cov.html#description

"If the program is invoked with a base name of gcov, it will behave as if the llvm-cov gcov command were called. Otherwise, a command should be provided."

@MehdiChinoune I'm wondering if this should be part of gcc-compat (?)

MehdiChinoune:
> @MehdiChinoune I'm wondering if this should be part of gcc-compat (?)

I prefer not to.

lazka:
I guess one could argue that it's an officially supported mode of llvm, unlike the gcc aliases -> fair enough.

PR #11855, opened 2022-06-15
gdal: update to 3.5.0

mmuetzel:
Although this is just a minor version bump, the build rules change significantly. Their build system is transitioning from autotools to cmake. The cmake toolchain is the recommended one now.
I didn't figure out how to build static and shared libraries in one go. So, I opted to build them separately (like done in many other build rules).
Bumping the pkgrel of some dependent packages is still pending. But imho it is better to wait for #11853 and #11854 before doing that. (To avoid having to fix issues with the current versions that would be obsolete once/if these other PRs land.) 
Leaving this as a draft for now.


mmuetzel:
Linking osgearth failed because the linker flags for GLEW are missing. See #11892 for a potential fix.

mmuetzel:
Some runners didn't finish in the 6 hours limit. But I think this is good to go now.

PR #11863, opened 2022-06-17
meson build test

lazka:
See #11864


lazka:
Learning from this; `PYTHONVERBOSE=1` makes the error go away :(

PR #11871, opened 2022-06-18
qt6-multimedia: Enable gstreamer

MehdiChinoune:


Biswa96:
Is it possible to separate gstreamer plugin as different package? gstreamer has many dependencies.

MehdiChinoune:
> Is it possible to separate gstreamer plugin as different package? gstreamer has many dependencies.

There is no plugin for gstreamer (you should see one in file-diff if any), QtMultimedia is linked directly to gstreamer. see: https://github.com/archlinux/svntogit-packages/blob/packages/qt6-multimedia/trunk/PKGBUILD

They have separated gstreamer as a plugin, but that's for 6.4.0

PR #11873, opened 2022-06-18
[WIP] Kicad packages upgrade to 6.0.x

stahta01:
Work in Progress DO NOT MERGE

Upgrade KiCAD packages to 6.0.5. Now, 6.0.6.

Plan is to wait till wxWidgets 3.2.0 is released and wxPython works with wxWidgets 3.2.0.

Know problems include:
1. Old Intel GPU driver bug is tripped by manifest file this results in 3dview not working.
2. Under UCRT64, sub-section "Plugin and Content manager" locks up on launch.
3. UCRT64 is now failing to build on CI server; the problem is now intermittent.

Tim S.


stahta01:
Todo list:
1. See if sip version 6 package can build KiCAD.

mingw-w64-kicad-library/PKGBUILD
```diff
@@ -7,26 +7,26 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-footprints"
...
+pkgrel=1
 pkgdesc="Support libraries for KiCad (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url='https://www.kicad.org/'
-license=("spdx:AGPL-3.0-or-later")
+license=("spdx:GPL-3.0-or-later")
 groups=("${MINGW_PACKAGE_PREFIX}-eda")
-depends=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+optdepends=("${MINGW_PACKAGE_PREFIX}-${_realname}")
```
mmuetzel, line 18: Why is this an *optional* dependency? Is there any meaningful use of the libraries without the "main" application?

stahta01: The size of the support libraries is very large as in about 500 mega bytes I made it an optdepends to avoid removing it by mistake; and, I hope to be about to only have one copy of the support libraries and multiple copies of the application at a future time.

Tim S.

mmuetzel: Ah ok. Got it. 👍

stahta01: By the time I am ready to have this merged I might change back to depends; but, it wastes about 10 minutes each time I have to re-install the package on my external USB hard drive.

stahta01:
Working on Updating wxWidgets to 3.2.0 RC

PR #11878, opened 2022-06-19
Minor Changes on opencv

3togo:
After setting  non-free algorithms" ON, some minor changes are made.


mingw-w64-opencv/0001-mingw-w64-cmake.patch
```diff
@@ -38,7 +38,9 @@ Subject: [PATCH 1/8] mingw-w64-cmake
   # OpenCV cmake options
   # ----------------------------------------------------------------------------
  @@ -211,14 +197,14 @@ OCV_OPTION(OPENCV_ENABLE_NONFREE "Enable non-free algorithms" OFF)
-  
+ -OCV_OPTION(OPENCV_ENABLE_NONFREE "Enable non-free algorithms" OFF)
+ +OCV_OPTION(OPENCV_ENABLE_NONFREE "Enable non-free algorithms" ON)
```
1480c1, line 42: 1. this would be better added to the PKGBUILD file
2. I would **not** recommend accepting this change if `non-free` means cannot distribute without potential legal issues

mingw-w64-opencv/0104-rgbd-module-missing-include.patch
```diff
@@ -19,6 +19,17 @@ index 002be7b..45e55ff 100644
  #endif
  
  // GL Extention definitions missing from standard Win32 gl.h
+@@ -153,7 +154,8 @@
+                           params.raycast_step_factor)),
+     pyrPoints(), pyrNormals(), warpfield()
+ {
+-#ifdef HAVE_OPENGL
++//#ifdef HAVE_OPENGL
++#ifdef HAVE_GLEW
```
1480c1, line 28: If this is really correct, it may be better to see about upstreaming, or at least checking with upstream

1480c1 (commented):
needs pkgrel update

mingw-w64-opencv/runme.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
```
1480c1, line 1: what's the point of this file? Does this need to be here?

mingw-w64-opencv/0001-mingw-w64-cmake.patch
lazka: yeah, please add why this isn't enabled by default, and why we still should.

mingw-w64-opencv/0104-rgbd-module-missing-include.patch
mmuetzel: Might be a workaround for the issue addressed in #11892. (But I haven't actually checked if it is.)

PR #11886, opened 2022-06-19
libdeflate: fix dynamic linking using pkg-config

mmuetzel:
The header file of libdeflate contains the following comment:
```
/*
 * On Windows, if you want to link to the DLL version of libdeflate, then
 * #define LIBDEFLATE_DLL.  Note that the calling convention is "stdcall".
 */
```

This PR adds a patch for the .pc file that reflects that requirement.
IIUC, the `-U` flag should undo the `-D` flag when linking statically.

This fixes a build error with GDAL.

### DIFF
--- a/mingw-w64-libdeflate/002-pkg-config.patch
+++ b/mingw-w64-libdeflate/002-pkg-config.patch
@@ -1,0 +1,10 @@
+diff -urN libdeflate-1.12/libdeflate.pc.in.orig libdeflate-1.12/libdeflate.pc.in
+--- libdeflate-1.12/libdeflate.pc.in.orig	2022-06-12 21:52:59.000000000 +0200
++++ libdeflate-1.12/libdeflate.pc.in	2022-06-19 15:39:22.766071700 +0200
+@@ -7,4 +7,5 @@
+ Description: Fast implementation of DEFLATE, zlib, and gzip
+ Version: @VERSION@
+ Libs: -L${libdir} -ldeflate
+-Cflags: -I${includedir}
++Cflags: -I${includedir} -DLIBDEFLATE_DLL
++Cflags.private: -ULIBDEFLATE_DLL
--- a/mingw-w64-libdeflate/PKGBUILD
+++ b/mingw-w64-libdeflate/PKGBUILD
@@ -4,21 +4,24 @@ _realname=libdeflate
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.12
-pkgrel=1
+pkgrel=2
 pkgdesc="Heavily optimized library for DEFLATE/zlib/gzip compression and decompression (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url='https://github.com/ebiggers/libdeflate'
-license=('MIT')
+license=('spdx:MIT')
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc")
 source=(${_realname}-${pkgver}.tar.gz::"${url}/archive/v${pkgver}.tar.gz"
-        "001-libdeflate-makefile-mingw-fix.patch")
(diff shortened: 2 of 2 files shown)


Biswa96:
Should this require rebuild of dependent packages?

lazka:
hm, looks good in general.

This will enable __stdcall for 32bit, I wonder we should patch that out, see https://github.com/ebiggers/libdeflate/pull/178

mmuetzel:
Hm...
I don't think so. If the packages have been built fine before the recent update of `libdeflate`, they should still be linking correctly to the updated library (iiuc).
There is only an issue when trying to build with the updated header and link statically when using the flags from the (un-patched) .pc file.

mmuetzel:
> This will enable __stdcall for 32bit, I wonder we should patch that out, see [ebiggers/libdeflate#178](https://github.com/ebiggers/libdeflate/pull/178)

IIUC, `libdeflate` *always*(?) used `__stdcall` on 32-bit Windows before. Given that they switched to `cdecl` upstream, we might as well cherry-pick that now. That would make make the change to the .pc file no longer necessary iiuc.
But we'd need to rebuild dependencies if we do that now. (However, we'd need to do that at some point in the future anyway given the upstream change...)

lazka:
Is LIBDEFLATE_DLL set somewhere in the build process? I'm worried that users will see stdcall while the library is built for cdecl.

mmuetzel:
> Is LIBDEFLATE_DLL set somewhere in the build process?

Yes, it is set when building the objects for the shared library.

lazka:
> Given that they switched to `cdecl` upstream, we might as well cherry-pick that now

I don't see any commits after the last release.

> Yes, it is set when building the objects for the shared library.

ok, thanks, looks good then.

lazka:
I'll rebuild things for good measure anyway.

mmuetzel:
It looks like `mingw-w64-python-imagecodecs` isn't picking up that flag:
```
  clang -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -march=pentium4 -mtune=generic -O2 -pipe -O3 -march=pentium4 -mtune=generic -O2 -pipe -O3 -march=pentium4 -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 -DWIN32=1 -Iimagecodecs -ID:/a/msys64/clang32/include/python3.10 -ID:/a/msys64/clang32/lib/python3.10/site-packages/numpy/core/include -c imagecodecs/_deflate.c -o build/temp.mingw_i686_clang-3.10/imagecodecs/_deflate.o
```

lazka:
Should we patch python-imagecodecs or drop __stdcall?

mmuetzel:
> Should we patch python-imagecodecs or patch out __stdcall?

I don't know. Fixing `python-imagecodecs` would probably be easy.
On the other hand, there is no real reason why libdeflate is using __stdcall (at least afaict). And just getting rid of it might be "most future proof". Maybe upstream would even accept a patch?

mmuetzel:
See #11893 for a potential fix for python-imagecodec.

lazka:
> Edit: Maybe there _is_ a good reason for it to use __stdcall:
> [ebiggers/libdeflate@5f38309](https://github.com/ebiggers/libdeflate/commit/5f3830983c255faf72edc81746f52f9bed65c493)

everyhing is using cdecl, and VB6 is kinda dead, so imo that's not a very good reason.

> See #11893 for a potential fix for python-imagecodecs.

thanks! let's do that and wait and see then.. if there are too many problems we can patch it out and rebuild.

lazka:
Upstream has changed it now after asking: https://github.com/ebiggers/libdeflate/pull/207

I'm going to backport that and do another round of pkgrel bumps.

kmilos:
Think we also need to define BUILDING_LIBDEFLATE when building the DLL which is not done in the Makefile AFAICT... But then that potentially screws up the static build, so we need to run two builds?

kmilos:
Actually, scratch that, it's defined internally: https://github.com/ebiggers/libdeflate/blob/master/lib/lib_common.h

PR #11890, opened 2022-06-19
osgworks-git: support relocation

mmuetzel:
Also use ninja generator in build rule.

This prevent errors caused by the hardcoded paths in the .cmake file when re-building dependent packages (e.g., `osgbullet-git`) locally.


mmuetzel:
The current errors are because some filesystem layout details are leaked into the generated .cmake files (e.g., `C:/building/msys64/mingw64/[...]`).
That no longer happens with this change. And dependent packages seem to build fine again.

PR #11892, opened 2022-06-19
cmake: fix FindGLEW module

mmuetzel:
This avoids an issue with some packages (e.g., `mingw-w64-osgearth`) because some of the variables the documentation promises (e.g., `GLEW_LIBRARIES`) aren't set:
https://cmake.org/cmake/help/latest/module/FindGLEW.html

See also this report upstream:
https://gitlab.kitware.com/cmake/cmake/-/issues/19662#note_665760

And this related workaround:
https://github.com/microsoft/vcpkg/blob/master/ports/glew/vcpkg-cmake-wrapper.cmake


mmuetzel:
This change allows to rebuild `osgearth`.

See also the currently failing pkgrel bump in #11855.

PR #11893, opened 2022-06-19
python-imagecodecs: use correct calling convention for libdeflate

mmuetzel:
Also remove an older patch that wasn't applied any more.


Biswa96:
@kmilos Would you like to review this change please?

lazka:
lgtm

lazka:
> MAybe we should patch the .pc file as well, no?

do you mean #11886 ?

PR #11894, opened 2022-06-20
windows-default-manifest: Drop Vista and support long Path

MehdiChinoune:


MehdiChinoune:
Does someone know a specific issue/bug related to long path where we could test if it is fixed by this change or not?

mati865:
Rust's build (with `check` enabled) very easily fails with too long path issue. I can try it in the evening.

lazka:
Everything depending on MAX_PATH filenames will result in buffer overflows.. so I don't think we can do it.

MehdiChinoune:
binutils fails to be rebuilt with this change!

MehdiChinoune:
> Everything depending on MAX_PATH filenames (and using the functions mentioned at the bottom of https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd) will result in buffer overflows.. so I don't think we can do it.

It will fail in both cases when hitting that MAX_PATH limit.
 
> As a start we could document how downstreams can port and enable it maybe?

What's the difference between being enabled in MSYS2 or enabled in downstream project?.

lazka:
> It will fail in both cases when hitting that MAX_PATH limit.

but one is an expected error, the other a buffer overflow resulting in a crashes and security vulnerabilities.

MehdiChinoune:
> > It will fail in both cases when hitting that MAX_PATH limit.
> 
> but one is an expected error, the other a buffer overflow resulting in a crashes and security vulnerabilities.

Yes, I know. Enabling it in downstream doesn't change that.

MehdiChinoune:
I have tried to build qt6-quick3d without assimp to see if this change fixes a bug I hit before (https://bugreports.qt.io/browse/QTBUG-96339), after rebuilding binutils of course (windres).
Well still fails.
Maybe the next step is to try after changing MAX_PATH in mingw-w64.
https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/limits.h#L20
https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/include/minwindef.h#L33
https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/ddk/include/ddk/usbstorioctl.h#L4
https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/include/shtypes.idl#L102

mati865:
Thank you for the ping and sorry the late reply.
Something has changed in Rust's build system that made it generate shorter file names so I couldn't reproduce it even with unmodified binutils. But as you have already noticed rebuilding binutils with this manifest is not enough.

> Maybe the next step is to try after changing MAX_PATH in mingw-w64.

That will break a lot of software.

PR #11903, opened 2022-06-21
bullet: update to 3.24

mmuetzel:
This also fixes building in the clang* environments.

Some of the existing patches address issues that have been fixed upstream in the meantime. Those (parts of) the patches are removed by this PR.

On Windows, all symbols must be defined when linking. That requires some changes to the cmake build rules.

Bumping the pkgrel of dependent packages (only `osgbullet-git`) is still pending. That package currently doesn't build (see #11890 and #11892).


mmuetzel:
Thanks for pushing! 👍

Realizing now (a bit late) that I forgot to add ninja to the makedepends of bullet. It still built fine in the CI though.

Is ninja installed by default on the runners now and no longer needs to be added explicitly?

lazka:
I'll add it

Issue #11912, opened 2022-06-22
`cmocka` static linking


msink:
```
$ ls /mingw64/lib | grep cmocka
libcmocka-static.a
libcmocka.dll.a

$ pkg-config --static --libs cmocka
-lcmocka
```

here should be either 
- `libcmocka.a` installed instead of `libcmocka-static.a`, or
- `-lcmocka-static` in `pkg-config` output

Biswa96:
Hi @msink, Thank you for reporting the issue. I have added a pull request to fix it. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/2542797703

msink:
Thanks, now in simple scenarions it works, but I failed to force `meson` to use static version:
https://github.com/libui-ng/libui-ng/pull/116#issuecomment-1164071725

Biswa96:
For the record, I have sent the _non-controversial_ part of the patch in upstream MR here https://gitlab.com/cmocka/cmocka/-/merge_requests/48

PR #11914, opened 2022-06-22
cmocka: Fix static library and pkgconfig file

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/11912

### DIFF
--- a/mingw-w64-cmocka/004-fix-pkgconfig-files.patch
+++ b/mingw-w64-cmocka/004-fix-pkgconfig-files.patch
@@ -1,0 +1,27 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -60,7 +60,7 @@
+ endif ()
+ 
+ # pkg-config file
+-configure_file(cmocka.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/cmocka.pc)
++configure_file(cmocka.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/cmocka.pc @ONLY)
+ install(
+   FILES
+     ${CMAKE_CURRENT_BINARY_DIR}/cmocka.pc
+--- a/cmocka.pc.cmake
++++ b/cmocka.pc.cmake
+@@ -1,5 +1,9 @@
+-Name: ${PROJECT_NAME}
++prefix=@CMAKE_INSTALL_PREFIX@
++libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
++includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
++
++Name: @PROJECT_NAME@
+ Description: The cmocka unit testing library
+-Version: ${PROJECT_VERSION}
+-Libs: -L${CMAKE_INSTALL_FULL_LIBDIR} -lcmocka
+-Cflags: -I${CMAKE_INSTALL_FULL_INCLUDEDIR}
++Version: @PROJECT_VERSION@
++Libs: -L${libdir} -lcmocka
++Cflags: -I${includedir}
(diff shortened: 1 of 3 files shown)


mingw-w64-cmocka/004-fix-pkgconfig-files.patch
```diff
@@ -0,0 +1,27 @@
...
+ install(
+   FILES
+     ${CMAKE_CURRENT_BINARY_DIR}/cmocka.pc
+--- a/cmocka.pc.cmake
++++ b/cmocka.pc.cmake
+@@ -1,5 +1,9 @@
+-Name: ${PROJECT_NAME}
++prefix=@CMAKE_INSTALL_PREFIX@
++libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
++includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```
eli-schwartz, line 18: This is something that can and should be upstreamed, because even if upstream has unfortunate opinions about the output name (why do they use cmocka-static? Maybe they don't care, but maybe they are very attached to the name. Who knows)...

... it's unquestionably correct to fix the pkg-config file to correctly use prefix/libdir etc.

However, it is also not so simple: in autotools you use `libdir=@libdir@` and autotools internally knows how to make that use `${prefix}`, and in Meson the pkg-config file is correctly generated for you, but in cmake it's all wild-west stuff. And CMAKE_INSTALL_LIBDIR might be defined as `/usr/lib` or `/mingw64/lib` instead of `lib`, so you end up with `libdir=${prefix}/usr/lib` and that expands to `libdir=/usr/usr/lib`.

Yay!

Of course this works if you know that people always define `-DCMAKE_INSTALL_PREFIX=lib` or use the default value.

But, it isn't upstreamable. For an upstreamable solution you need JoinPaths.cmake from https://github.com/jtojnar/cmake-snips/

(Isn't cmake fun?)

Biswa96: Agree. There are some opinions about `CMAKE_INSTALL_FULL_INCLUDEDIR` vs. `CMAKE_INSTALL_INCLUDEDIR`. But some parts of the patch can be upstreamed.

eli-schwartz: They don't have that suffix except by inconsistent convention. Meson's rationale has this source code comment:

```
# By default a static library is named libfoo.a even on Windows because
# MSVC does not have a consistent convention for what static libraries
# are called. The MSVC CRT uses libfoo.lib syntax but nothing else uses
# it and GCC only looks for static libraries called foo.lib and
# libfoo.a. However, we cannot use foo.lib because that's the same as
# the import library. Using libfoo.a is ok because people using MSVC
# always pass the library filename while linking anyway.
```

More generally, it doesn't explain why, even if the unwise choice of `foo-static` is used, that is used regardless of OS or compiler toolchain... leading to broken behavior on all non-Windows systems, such as BSD, Linux, or macOS, and also... broken behavior on MSYS2. :)

MehdiChinoune: > ```
> # By default a static library is named libfoo.a even on Windows because
> # MSVC does not have a consistent convention for what static libraries
> # are called. The MSVC CRT uses libfoo.lib syntax but nothing else uses
> # it and GCC only looks for static libraries called foo.lib and
> # libfoo.a. However, we cannot use foo.lib because that's the same as
> # the import library. Using libfoo.a is ok because people using MSVC
> # always pass the library filename while linking anyway.
> ```

That's meson build-system choice, cmake and qmake build-system chose `.lib` as the default suffix for both import and static libraries on Windows/MSVC.

Biswa96: Static library was removed in upstream https://gitlab.com/cmocka/cmocka/-/merge_requests/13. I have asked about it. Let see.

PR #11920, opened 2022-06-23
netcdf: fix segfault if file cannot be opened

### LINKED ISSUE #11918: netcdf: segmentation fault when creating a file in a non-existing directory
This is a regression due to the update to 4.9.0 done in https://github.com/msys2/MINGW-packages/pull/11879. Previous 4.8.1-4 package errors out cleanly. Issue found on the GDAL autotest continuous integration (cf https://github.com/OSGeo/gdal/issues/5970)

```
$ nccopy --version
[...]
netCDF library version 4.9.0 of Jun 21 2022 18:22:22 $
$ nccopy byte_no_cf.nc "\non_existing_path\out.nc"
Error:NCpathcvt: stat=-36 (NetCDF: Invalid argument)
Segmentation fault
```
where byte_no_cf.nc is in the attached zip (but true for any netCDF file). 
[byte_no_cf.nc.zip](https://github.com/msys2/MINGW-packages/files/8967630/byte_no_cf.nc.zip)

vs
```
$ nccopy --version
[...]
netCDF library version 4.8.1 of Apr  2 2022 18:28:59 $
$ nccopy byte_no_cf.nc "\non_existing_path\out.nc"
No such file or directory
Location: file ; line 2016
```

mmuetzel:
It looks like `errno` isn't set when file cannot be opened. Set it manually instead.
Fixes #11918.

Also fix some linker flags that get messed up if the detected debug library is the same as the release library.

### DIFF
--- a/mingw-w64-netcdf/0001-no-debug-libraries.patch
+++ b/mingw-w64-netcdf/0001-no-debug-libraries.patch
@@ -1,0 +1,48 @@
+diff -urN netcdf-c-4.9.0/cmake/modules/FindZip.cmake.orig netcdf-c-4.9.0/cmake/modules/FindZip.cmake
+--- netcdf-c-4.9.0/cmake/modules/FindZip.cmake.orig	2022-06-23 17:20:15.443959500 +0200
++++ netcdf-c-4.9.0/cmake/modules/FindZip.cmake	2022-06-23 17:20:24.156687000 +0200
+@@ -39,7 +39,7 @@
+                PATHS ${Zip_LIBRARY_DIRS} NO_DEFAULT_PATH)
+ 
+   SET(Zip_LIBRARIES )
+-  IF(Zip_DEBUG_LIBRARY AND Zip_RELEASE_LIBRARY)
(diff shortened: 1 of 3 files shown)


lazka:
thanks

hm, it looks like it only doesn't set errno if converting/parsing the path fails: https://github.com/Unidata/netcdf-c/blob/ac5c2f55c1b523cb2fec1942614643bd34957403/libdispatch/dpathmgr.c#L431 so 

so this is kinda hacky imo, but better than crashing.

mjwoods:
Hi @mmuetzel, the path conversion code in netcdf has recently been updated for mingw, so it would be good to get this bug 
fixed upstream. Has the issue been reported?

mjwoods:
Ah yes, now I see that upstream has been dealt with in #11918 . I should mention that the netcdf developers have recently put in a solid effort fixing problems for mingw (with some contributions from me), and they added mingw to their CI tests a few months ago. It would benefit the mingw project if future bugs are reported upstream as a matter of routine.

mmuetzel:
OT: I'm not sure if that "convert from ANSI to UTF-8" logic is a good idea. It basically means that it isn't possible to open files on Windows that have names that contain characters that can't be represented in the system's locale.
Instead, it might be better to require the interface to be UTF-8 on all platforms from the start. (Or at least make that an option...)

PR #11924, opened 2022-06-24
pastix: build without fortran support under clang

lazka:


lazka:
it still looks for fortran all over the place...

PR #11925, opened 2022-06-24
open62541: Update to 1.2.5, try to fix clang

lazka:


mmuetzel:
The MINGW64 runner failed for some reason while the others succeeded. The error message is kind of odd:
https://github.com/msys2/msys2-autobuild/runs/7039796356?check_suite_focus=true?check_suite_focus=true#step:10:24975
```
  /usr/bin/python: can't open file '/c/M/mingw-w64-open62541/src/build-MINGW64/C:/M/mingw-w64-open62541/src/open62541/tools/generate_nodeid_header.py': [Errno 2] No such file or directory
```

It looks like it appended two paths (where it maybe should have constructed a relative path?). Or is this a missing space?

naveen521kk:
Maybe mingw python should be added as a makedep?

PR #11929, opened 2022-06-24
aubio: Add package

Biswa96:


not-matt:
Big

Biswa96:
The clang builds fail due to pthread requirement. I would like examine if the requirement is for aubio or fftw. Then add winpthread library in the pkgconfig file of that package.

lazka:
in the repo now -> https://packages.msys2.org/base/mingw-w64-aubio

PR #11938, opened 2022-06-26
cmake: fix regression in FindGLEW

### LINKED ISSUE #11932: Bug in FindGLEW module
## Describe the issue

When using the GLEW CMake package, the following error is triggered:
```
CMake Error at D:/a/_temp/msys64/mingw64/share/cmake/Modules/FindGLEW.cmake:84 (select_library_configurations):
  Unknown CMake command "select_library_configurations".
```

I would guess this problem can probably be attributed to the recent commit [1a060cdaa2059734d7cf91038b5bf6d71ef6aace](https://github.com/msys2/MINGW-packages/commit/1a060cdaa2059734d7cf91038b5bf6d71ef6aace) by @mmuetzel, as this commit adds a `select_library_configurations` call.

## Steps to Reproduce the Problem

1. Install mingw-w64-cmake and mingw-w64-x86_64-glew.
2. Use `find_package(GLEW REQUIRED)` in the CMakeLists.txt file of a project.
3. Get the following output when running `cmake` for the project:
```
CMake Error at D:/a/_temp/msys64/mingw64/share/cmake/Modules/FindGLEW.cmake:84 (select_library_configurations):
  Unknown CMake command "select_library_configurations".
```

## Additional Context: Operating System, Screenshots

- OS: Windows 10 (tested both on a "real" Windows 10 PC and a GitHub Actions server)

mmuetzel:
Also use patches as applied upstream (as far as possible).

This fixes a regression caused by 1a060cdaa2059734d7cf91038b5bf6d71ef6aace.

Fixes #11932.

Sorry, I missed that with the previous change. The include must have happened somewhere before when I tried it with some "real-world" build rules. The test suite didn't catch it either unfortunately...


lazka:
thanks for the quick fix!

PR #11940, opened 2022-06-27
New package: liblcf

AntumDeluge:
Library to handle RPG Maker 2000/2003 and EasyRPG game data

- Homepage: https://easyrpg.org/
- Upstream: https://github.com/EasyRPG/liblcf
- Current version: 0.7.0

I don't think it is an issue, but I get the following output when building:
```
==> Tidying install...
  -> Removing libtool files...
  -> Purging unwanted files...
  -> Stripping unneeded symbols from binaries and libraries...
...\mingw64\bin\objdump.exe: mingw64/include/lcf/config.h: file format not recognized
...\mingw64\bin\objdump.exe: mingw64/include/lcf/context.h: file format not recognized
...
```

For some reason, `objdump` wants to scan `.h` files.

### DIFF
--- a/mingw-w64-liblcf/PKGBUILD
+++ b/mingw-w64-liblcf/PKGBUILD
@@ -1,0 +1,66 @@
+# Maintainer: Jordan Irwin <antumdeluge@gmail.com>
+
+_realname=liblcf
+pkgbase="mingw-w64-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.7.0
+pkgrel=1
+pkgdesc="Library to handle RPG Maker 2000/2003 and EasyRPG game data (mingw-w64)"
+arch=("any")
+mingw_arch=("mingw32" "mingw64" "clang32" "clang64" "ucrt64")
+url="https://easyrpg.org/"
+license=("MIT")
+source=("${_realname}-${pkgver}.tar.gz::https://github.com/EasyRPG/liblcf/archive/refs/tags/${pkgver}.tar.gz")
+sha256sums=("a98dca6396f28b1902ae299e66ad40e0ff7fb5052269d6e619d06aaae4d2aeb2")
+depends=("${MINGW_PACKAGE_PREFIX}-expat"
+         "${MINGW_PACKAGE_PREFIX}-icu")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-pkgconf"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-cmake>=3.10")
+
+
+build() {
+  declare -a extra_config
+  if check_option "debug" "n"; then
+    extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  [[ -d "${srcdir}/build-${MSYSTEM}-shared" ]] && rm -rf "${srcdir}/build-${MSYSTEM}-shared"
+  mkdir -p "${srcdir}/build-${MSYSTEM}-shared" && cd "${srcdir}/build-${MSYSTEM}-shared"
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  ${MINGW_PREFIX}/bin/cmake.exe \
+    -G "Ninja" \
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DBUILD_SHARED_LIBS=ON \
+    "${extra_config[@]}" \
+    ../${_realname}-${pkgver}
+
+  "${MINGW_PREFIX}/bin/cmake.exe" --build ./
+
+  [[ -d "${srcdir}/build-${MSYSTEM}-static" ]] && rm -rf "${srcdir}/build-${MSYSTEM}-static"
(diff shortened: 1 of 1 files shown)


Biswa96:
1. Some variables seem to be unnecessary e.g. _upstream, dir_shared, dir_static. Those can be use as is.
2. It is not required to mention bsdtar, expat, icu in makedepends because the later two are in depends list and bsdtar is used implicitly.
3. It is not required to mention _build_shared and _build_static functions separately. Both can be done in one build function.
4. The static lib install step can be done with `cmake --install .` command. See other PKGBUILD files.
5. If possible use ninja instead of make.
6. Use `MSYSTEM` variable instead of `MINGW_CHOST` for build directory.

AntumDeluge:
@Biswa96 I will go ahead & make the changes. But I have questions about the following:

> 4. The static lib install step can be done with `cmake --install .` command. See other PKGBUILD files.

Isn't that just redundant & installs a bunch of the same files over again? Also, won't that potentially overwrite CMake configurations for shared libs?

AntumDeluge:
Okay. So I need to make sure to install static libs first.

AntumDeluge:
Done. Let me know if there is anything else I need to change.

mingw-w64-liblcf/PKGBUILD
```diff
@@ -0,0 +1,64 @@
+# Maintainer: Jordan Irwin <antumdeluge@gmail.com>
+
+_realname=liblcf
+pkgbase="mingw-w64-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.7.0
+pkgrel=1
+pkgdesc="Library to handle RPG Maker 2000/2003 and EasyRPG game data"
```
mmuetzel, line 8: The package description for mingw packages should end with ` (mingw-w64)`.
See the template:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball.sh

mingw-w64-liblcf/PKGBUILD
```diff
@@ -0,0 +1,64 @@
...
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  ${MINGW_PREFIX}/bin/cmake.exe \
+    -G "Ninja" \
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DBUILD_SHARED_LIBS=ON \
+    "${extra_config[@]}" \
+    ../${_realname}-${pkgver}
+
+  ninja
```
mmuetzel, line 42: Consider this change (to make it easier should there ever be the need to switch generators):
``` diff
-  ninja
+  "${MINGW_PREFIX}"/bin/cmake.exe --build .
```

mingw-w64-liblcf/PKGBUILD
```diff
@@ -0,0 +1,64 @@
...
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  ${MINGW_PREFIX}/bin/cmake.exe \
+    -G "Ninja" \
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DBUILD_SHARED_LIBS=OFF \
+    "${extra_config[@]}" \
+    ../${_realname}-${pkgver}
+
+  ninja
```
mmuetzel, line 55: Same as for the shared library.

mingw-w64-liblcf/PKGBUILD
```diff
@@ -0,0 +1,64 @@
...
+}
+
+package() {
+  cd "${srcdir}/build-${MSYSTEM}-static"
+  DESTDIR="${pkgdir}" ninja install
```
mmuetzel, line 60: Consider replacing this with:
``` diff
-  DESTDIR="${pkgdir}" ninja install
+  DESTDIR="${pkgdir}" "${MINGW_PREFIX}"/bin/cmake.exe --install .
```

mingw-w64-liblcf/PKGBUILD
```diff
@@ -0,0 +1,64 @@
...
+
+  cd "${srcdir}/build-${MSYSTEM}-shared"
+  DESTDIR="${pkgdir}" ninja install
```
mmuetzel, line 63: same

mmuetzel (commented):
Just some nit and minor changes.

mmuetzel:
Ianal, but I think the MIT license requires including it in the distribution. Maybe, it should be installed similarly to the command close to the end of the `PKGBUILD` template.

mingw-w64-liblcf/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+pkgbase="mingw-w64-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.7.0
+pkgrel=1
+pkgdesc="Library to handle RPG Maker 2000/2003 and EasyRPG game data (mingw-w64)"
+arch=("any")
+mingw_arch=("mingw32" "mingw64" "clang32" "clang64" "ucrt64")
+url="https://easyrpg.org/"
+license=("MIT")
+source=("${_realname}-${pkgver}.tar.xz::https://github.com/EasyRPG/liblcf/archive/refs/tags/${pkgver}.tar.gz")
```
mmuetzel, line 13: That seems to be working in the CI. But still, it seems a little odd to change the file extension from `.tar.gz` to `.tar.xz`.

AntumDeluge:
Done.

mingw-w64-liblcf/PKGBUILD
AntumDeluge: Oops.

AntumDeluge: Fixed.

PR #11941, opened 2022-06-27
More clang arm64 builds

hmartinez82:
Towards building GIMP for Windows on ARM


Biswa96:
Did you make sure that all of the packages build for arm architecture?

hmartinez82:
@Biswa96 yes. I ran the builds locally in a Surface Pro X, even ran the unit tests for packages that provide it :)

Biswa96:
The gpg check of mpfr package works fine here. `"${MINGW_PACKAGE_PREFIX}-cc"` has to be added in makedepends of mpfr package. BTW, as these are small changes all commits can be put into one.

hmartinez82:
Ok, I'll squash the commits

hmartinez82:
I meant PGP , not GPG :)

Biswa96:
Typo. That PGP check succeed in CI. See the output

```
      mpfr-4.1.0.tar.xz ... Passed
```

lazka:
re pgp, you can run `../.ci/fetch-validpgpkeys.sh` in a package dir to get the right keys. makepkg normally fetches them automatically, but only if the signature contains the full fingerprint.

hmartinez82:
@kmilos  https://github.com/msys2/MINGW-packages/pull/12526 😀

kmilos:
Amazing work, thank you very much!

And `gmic` was only an optional dependency (with a huge chain behind it indeed!), darktable would've worked without it with some minor features disabled...

PR #11947, opened 2022-06-27
Remove libgnomecanvas and libart

lazka:
Both unmaintained for 11+ years and no users


lazka:
If anyone still needs them let us know and we can bring them back.

lazka:
Restored in #12054

gtk2 it on the way out though, and eventually we will drop them again.

PR #11029, opened 2022-03-18
add recipe for swi-prolog (mingw32, mingw64, ucrt64)

mgondan:
This recipe is for a mingw32/mingw64/ucrt64 of the SWI-Prolog programming language. I made a second PR which is now based on a PKGBUILD template, as suggested in our previous conversation.

A few issues will be solved in future versions:

* a little patch that is needed to avoid overwriting dlls that are already installed on the MinGW system (e.g., zlib1.dll). I have already made the respective PR to SWI-Prologs development version.
* some of the installed dlls match the source folders. I have reported this potential problem to the SWI-Prolog developers.
* a few compiler warnings

Best regards,

Matthias

PR #11949, opened 2022-06-27
swi-prolog: update to 8.4.3

mgondan:


Biswa96:
The files are installed in different paths. Please check the CI log https://github.com/msys2/MINGW-packages/runs/7081687344?check_suite_focus=true#step:9:2120

mgondan:
Yes, this is intended. In the old version, quite a few module-specific dlls with rather generic names were copied into common places such as e.g. /mingw64/bin, these have moved elsewhere.

See https://github.com/msys2/MINGW-packages/pull/11029#issuecomment-1076839966

mingw-w64-swi-prolog/PKGBUILD
```diff
@@ -44,8 +50,9 @@ build() {
 
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
   ${MINGW_PREFIX}/bin/cmake \
-    -G"Ninja" \
+    -G Ninja \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -DMSYS2=1 \
```
mmuetzel, line 55: The name of this parameter is a bit misleading. This rule doesn't build for MSYS2 but for native Windows.

mmuetzel (commented):
I haven't tested any of this. Just some comments.

mingw-w64-swi-prolog/PKGBUILD
```diff
@@ -3,16 +3,13 @@
...
 pkgrel=1
 pkgdesc="Prolog environment (mingw-w64)"
 arch=(any)
-mingw_arch=('mingw64' 'mingw32' 'ucrt64')
+mingw_arch=('mingw64' 'mingw32' 'ucrt64' 'clang64' 'clang32')
 url="https://www.swi-prolog.org/"
 license=("FreeBSD")
-depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
-         "${MINGW_PACKAGE_PREFIX}-libwinpthread-git"
-         "${MINGW_PACKAGE_PREFIX}-zlib")
```
mmuetzel, line 15: Do the parts of swi-prolog that used these dependencies no longer work?

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
```diff
@@ -0,0 +1,52 @@
...
+@@ -670,16 +670,17 @@
+	    FILES ${CMAKE_CURRENT_SOURCE_DIR}/swipl.ico
+	    DESTINATION ${SWIPL_INSTALL_PREFIX})
+
+-else(WIN32)
++endif()
+
+-if(NOT BUILD_MACOS_BUNDLE)
++if((NOT BUILD_MACOS_BUNDLE AND NOT WIN32) OR MSYS2)
+ # Create symbolic link from public installation dir to executables
```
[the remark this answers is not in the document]
mgondan: You're right, and, in fact, cmake's ilink places a normal copy of swipl.exe, swipl-ld.exe and libswipl.dll into the destination folder (e.g., /mingw64/bin). It works, but it's not pretty, and it's on my todo list. 

Of course, any good hint is welcome. I've seen a promising solution in python, with a python.exe in /mingw64/bin as well as something related in /mingw64/lib/python3.9/venv/scripts/nt, but I am unsure what is the specific difference between the two.

mingw-w64-swi-prolog/PKGBUILD
mgondan: I see. The parameter is an interim solution, as long as the Windows version of SWI-Prolog relies on a different compiler chain. Hope it is ok for now, it will disappear in future releases.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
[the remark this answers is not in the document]
mmuetzel: If cmake's `ilink` already makes copies, I guess that is ok.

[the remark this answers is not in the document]
mmuetzel: But why do you need this change?
Without it, the condition is true.
With it, the condition is still true...

mingw-w64-swi-prolog/PKGBUILD
mmuetzel: Looking at the other patches, you could probably get rid of that interim solution already.
The only point where it is "really" used at the moment is in your `03-unix-style-directory-structure.patch`. IIUC, the existing condition uses `WIN32` as a synonym for "this is a MSVC compatible compiler toolchain". But the build rule should probably use `MSVC` as the condition instead (which is more specific and should give the correct result on MinGW and with MSVC).

mgondan: Ah, I see your point. No, the "normal" Windows version of SWI-Prolog is also build with a MinGW-compiler (crosscompiled from some Linux flavor). Conditions based on MSVC are not of any help here.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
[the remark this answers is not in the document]
mgondan: The condition evaluates to true for Linux and Windows/MSYS2, but not for MacOS. Unsure if I miss something.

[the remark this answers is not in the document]
mmuetzel: Exactly. But it does that before and after your change.

mingw-w64-swi-prolog/02-stack-size-for-clang.patch
```diff
@@ -0,0 +1,11 @@
+--- swipl-8.4.3/cmake/CStack.cmake	2022-03-19 09:27:42.673709000 +0100
++++ swipl-8.4.3/cmake/CStack.cmake	2022-03-19 09:34:03.447846000 +0100
+@@ -5,7 +5,7 @@
+ 			  LINK_FLAGS "-Wl,--stack,${limit}")
+   else()
+     set_target_properties(${target} PROPERTIES
+-			  LINK_FLAGS /STACK:${limit})
++			  LINK_FLAGS "-Wl,--stack,${limit}")
```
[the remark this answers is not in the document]
mmuetzel: This branch was clearly meant to be used for a MSVC compatible toolchain. Using `MSVC` as the condition would be more logical.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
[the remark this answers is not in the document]
mgondan: Ah, ok, now I got it. Of course, you're right. 

For what it's worth: The patch looks like this to mimic a pull request that I made upstream in swi-prolog (the "not-msys-version"). If the maintainer of swi-prolog compiles swipl, MSYS2 is not set. Admittedly, this was not obvious.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
```diff
@@ -0,0 +1,52 @@
+--- swipl-8.4.3/CMakeLists.txt.old	2022-06-17 10:48:26 +0000
++++ swipl-8.4.3/CMakeLists.txt	2022-06-17 10:49:01 +0000
+@@ -104,7 +104,7 @@
+ ################
+ # Installation directories
+ 
+-if(WIN32)
++if(WIN32 AND NOT MSYS2)
```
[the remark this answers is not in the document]
mmuetzel: Looking at this part in a bit more detail: Does this work correctly? I've seen in other projects that executables/libraries installed in an "arch" directory couldn't be found on Windows...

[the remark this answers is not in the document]
mgondan: The executables (basically, swipl.exe and swipl-ld.exe) are "ilinked" (i.e., copied) into /mingw64/bin. The dlls were in that folder, too, in the last version. It's quite a few, and they have rather generic names. And, in fact, your colleagues made a comment along the lines of "good enough for a start". Now, I changed the directory structure and moved the dlls into the same place as they would be on a Linux computer. When registering a module in swipl, win_add_dll_directory is called to add the folder of the respective dll to the search path.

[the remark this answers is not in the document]
mmuetzel: Got it. Thanks for clarifying. 👍

(PS: I'm just a contributor like you. So, I guess we are all colleagues. 🙂)

mingw-w64-swi-prolog/PKGBUILD
mmuetzel: This is not meant to stall anything. Just out of interest:
Could you please give a hint how and why the "normal" Windows version is different from this one?

mingw-w64-swi-prolog/02-stack-size-for-clang.patch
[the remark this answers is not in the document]
mgondan: Done as suggested.

mingw-w64-swi-prolog/02-stack-size-for-clang.patch
```diff
@@ -0,0 +1,17 @@
+--- swipl-8.4.3/cmake/CStack.cmake.old  2022-06-28 13:46:34.320820200 +0200
++++ swipl-8.4.3/cmake/CStack.cmake      2022-06-28 14:56:54.511545700 +0200
+@@ -1,11 +1,9 @@
+ function(target_c_stack target limit)
+ if(WIN32)
+-  if(CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
```
mmuetzel, line 6: It looks like this patch isn't based on the release tarball. But on some intermediate state after applying a previous patch from this PR...

mingw-w64-swi-prolog/PKGBUILD
mmuetzel: Thanks for clairifying. I think I got it.
Maybe, they could also use `GNUInstallDirs` to select which parts should be installed where below the set prefix...

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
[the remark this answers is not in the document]
mmuetzel: I see you marked this as resolved. But IIUC, you wanted to replace the (confusing) condition by this simpler one:
``` diff
-if((NOT BUILD_MACOS_BUNDLE AND NOT WIN32) OR MSYS2)
+if(NOT BUILD_MACOS_BUNDLE)
```

mgondan:
I saw your thumbs up after my explanation, so I mistakenly understood it as approved. Will do the fix.

I submitted two more PRs upstream in SWI-Prolog and I am currently waiting for their feedback, therefore the delay.

Best wishes,

Matthias

Am 30.06.2022 um 08:02 schrieb Markus Mützel ***@***.******@***.***>>:


@mmuetzel commented on this pull request.

________________________________

In mingw-w64-swi-prolog/03-unix-style-directory-structure.patch<https://github.com/msys2/MINGW-packages/pull/11949#discussion_r910638030>:

> ++endif()
+
+ set(SWIPL_CMAKE_NAMESPACE swipl::)
+ set(SWIPL_INSTALL_CMAKE_CONFIG_DIR lib/cmake/swipl)
+--- swipl-8.4.3/src/CMakeLists.txt.old 2022-06-17 12:01:22 +0000
++++ swipl-8.4.3/src/CMakeLists.txt     2022-06-17 12:00:52 +0000
+@@ -670,16 +670,17 @@
+           FILES ${CMAKE_CURRENT_SOURCE_DIR}/swipl.ico
+           DESTINATION ${SWIPL_INSTALL_PREFIX})
+
+-else(WIN32)
++endif()
+
+-if(NOT BUILD_MACOS_BUNDLE)
++if((NOT BUILD_MACOS_BUNDLE AND NOT WIN32) OR MSYS2)
+ # Create symbolic link from public installation dir to executables


I see you marked this as resolved. But IIUC, you wanted to replace the (confusing) condition by this simpler one:

-if((NOT BUILD_MACOS_BUNDLE AND NOT WIN32) OR MSYS2)
+if(NOT BUILD_MACOS_BUNDLE)

mmuetzel:
> I saw your thumbs up after my explanation, so I mistakenly understood it as approved. Will do the fix.

Sorry for the miscommunication. I intended to thumbs up to this:
> I'll simplify it.

lazka:
How does one start it now that there is nothing in PATH?

mgondan:
swipl.exe is in e.g. /mingw64/bin. Does this work for you?

lazka:
there is a "swipl" file without an extension:

```
$ swipl
SWI-Prolog: [FATAL ERROR: at Sun Jul 03 19:45:57 2022
        Could not find system resources]
```

mgondan:
I'll have to check this. Can you help me a bit?

/mingw64/bin/swipl works from msys2 bash, correct?

lazka:
> /mingw64/bin/swipl works from msys2 bash, correct?

yes, but as state above, it fails with "FATAL ERROR"

mgondan:
Sorry, pressed the button too fast, see amendment above.

The fatal error needs to be addressed, I'll check this next week.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
```diff
@@ -0,0 +1,51 @@
...
+-else(WIN32)
++endif()
+
+ if(NOT BUILD_MACOS_BUNDLE)
+ # Create symbolic link from public installation dir to executables
+ install(DIRECTORY DESTINATION bin)
+ ilink(${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/swipl
+       ${CMAKE_INSTALL_PREFIX}/bin/swipl)
+ ilink(${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/swipl-ld
+       ${CMAKE_INSTALL_PREFIX}/bin/swipl-ld)
```
mmuetzel, line 45: Maybe consider suffixing these with `${CMAKE_EXECUTABLE_SUFFIX}`.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
```diff
@@ -0,0 +1,51 @@
...
++ilink(${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/libswipl.dll
++      ${CMAKE_INSTALL_PREFIX}/bin/libswipl.dll)
```
mmuetzel, line 47: Maybe consider suffixing with `${CMAKE_SHARED_LIBRARY_SUFFIX}` instead of hardcoding `.dll`. (At least when upstreaming. But probably better also here.)

mgondan: Done!

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
mgondan: Done!

mmuetzel:
`/mingw64/bin/swipl.exe` still fails with:
```
$ swipl.exe
SWI-Prolog: [FATAL ERROR: at Mon Jul 04 14:02:10 2022
        Could not find system resources]
```

(Now probably also with other shells.)

mgondan:
> > /mingw64/bin/swipl works from msys2 bash, correct?
> 
> yes, but as stated above, it fails with "FATAL ERROR"

Ok, getting there. SWI-Prolog wants to access its home directory, which is not /mingw64, but /mingw64/lib/swipl. Makes perfect sense, it was just not visible on my computer. There are several ways to tell the program where to find its home (command-line arguments, environment variable, a little file swipl.home in the parent folder. The last one would be an option, i.e., put a file named "swipl.home" into /mingw64, with the content "lib/swipl". Is this a way to go?

mmuetzel:
How does swipl know where to look for that file? Could it be patched to look at the correct location in the first place? I didn't see the file swipl.home installed directly at the prefix in other (Linux) distributions.

mgondan:
It's the thing with the symbolic link. "Normally" (i.e., under linux), the executable of swipl is installed at /usr/lib/swipl/bin (or something like this), and swipl.home is found in /usr/lib/swipl. The symbolic link in /usr/bin/swipl points to /usr/lib/swipl/bin/swipl, and everything works as intended.

Under Windows, symbolic links are difficult/not possible, therefore, swipl.exe is copied from /mingw64/lib/swipl/bin/amd64-windows/swipl.exe to /mingw64/bin/swipl.exe. Now, it does not find swipl.home at its parent directory. So I suggest to copy swipl.home to /mingw64, as well.

Regarding a patch: In linux, you may have seen another swipl.home in /usr/lib/swipl/bin, pointing to "..", the parent directory. My impression is, though, that this file is currently ignored. I could ask the developer of swi-prolog to fix this. As a consequence, there'd be a possibility to copy a file swipl.home with content "../lib/swipl" to /mingw64/bin. This avoids polluting the "root" folder. What do you think. It would take a bit of time I guess, but there's no hurry.

lazka:
Another hack would be a bash script that just forwards all arguments instead of a symlink.. downside that it only works in bash.

mgondan:
> Another hack would be a bash script that just forwards all arguments instead of a symlink.. downside that it only works in bash.

I like the idea. Pushed the repective fix.

Please note, however, that there's still something broken, namely, swipl does currently not find the dlls. This only came to my attention after lazka reported

```
$ swipl.exe
SWI-Prolog: [FATAL ERROR: at Mon Jul 04 14:02:10 2022
        Could not find system resources]
```

Therefore, please do not yet merge. Should I close this pull request in the meantime?

lazka:
whatever you prefer.

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
```diff
@@ -43,12 +43,12 @@
...
-+ilink(${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/swipl-ld${CMAKE_EXECUTABLE_SUFFIX}
-+      ${CMAKE_INSTALL_PREFIX}/bin/swipl-ld${CMAKE_EXECUTABLE_SUFFIX})
-+ilink(${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/libswipl${CMAKE_SHARED_LIBRARY_SUFFIX}
-+      ${CMAKE_INSTALL_PREFIX}/bin/libswipl${CMAKE_SHARED_LIBRARY_SUFFIX})
++file(WRITE ${CMAKE_CURRENT_SOURCE_DIR}/swipl "#!/usr/bin/bash\n${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/swipl $@\n")
++file(WRITE ${CMAKE_CURRENT_SOURCE_DIR}/swipl-ld "#!/usr/bin/bash\n${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/swipl-ld $@\n")
++ilink(${CMAKE_CURRENT_SOURCE_DIR}/swipl
++      ${CMAKE_INSTALL_PREFIX}/bin/swipl)
++ilink(${CMAKE_CURRENT_SOURCE_DIR}/swipl-ld
++      ${CMAKE_INSTALL_PREFIX}/bin/swipl-ld)
```
mmuetzel, line 51: This looks a little odd. Why create the file in the source tree and create a symlink to it in the install tree? The latter only happens to create a copy on Windows (and I guess that might change at some time).
Why don't you create the script directly in the install tree?

mgondan:
This hack is needed to get the bash script to ${CMAKE_INSTALL_PREFIX}/bin. Expansion of this variable is blocked in PKGBUILD (and that is probably needed). For some reasons, the ilink directive still expands CMAKE_INSTALL_PREFIX to e.g. /mingw64, and I make use of it here. If I use file(WRITE with CMAKE_INSTALL_PREFIX, the file isn't written anywhere.

Any hint to a better solution is appreciated.

mmuetzel:
I tried to reproduce the issue locally. But the configuration step failed for me.
It turns out that there is a target name collision if Qt happens to be installed when building this package. To avoid that name collision (and build the GUI by default), I used this change:
[0001-swi-prolog-build-GUI.zip](https://github.com/msys2/MINGW-packages/files/9127390/0001-swi-prolog-build-GUI.zip)

With that, the build completed. (It might be missing some makedepends though?)

To avoid using `ilink`, I used the following additional change:
[0002-swi-prolog-install-shell-scripts.zip](https://github.com/msys2/MINGW-packages/files/9127394/0002-swi-prolog-install-shell-scripts.zip)

With that, I can successfully start `swipl` and `swipl-win` from a bash.
Does this work for you, too?

Additionally, I'm not sure if the binaries are still distributable with a BSD license. Afaict, it links to gmp which will probably require to distribute the whole thing as LGPL. (But ianal.)

mingw-w64-swi-prolog/03-unix-style-directory-structure.patch
mgondan: I have changed "ilink" to "install", according to your suggestion below.

mgondan:
> Additionally, I'm not sure if the binaries are still distributable with a BSD license. Afaict, it links to gmp which will probably require to distribute the whole thing as LGPL. (But ianal.)

I just used the swipl license. By the way, recent versions of GMP are now LGPL, not GPL. In my understanding, the LGPL is "non-viral", such that proprietary projects can dynamically link to it (https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License#Differences_from_the_GPL).

mmuetzel:
LGTM.
If you are also happy with it, you could probably convert this from a draft to "ready for review".

Minor possible change: Maybe Qt could be made an optional dependency for the GUI. The remainder is still pretty useful. And iiuc, it doesn't need that (rather large) dependency.
IIUC, that would require moving `"${MINGW_PACKAGE_PREFIX}-qt6-base"` from `depends` to `optdepends` and `makedepends`. See, e.g., https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-cmake/PKGBUILD#L41

mmuetzel:
I downloaded and installed the latest MINGW64 build artifact. `check_installation.` looks ok to me.
Does it also work for you now?

mgondan:
Yes! I have reopened it for review. Thank you very much for all your help and constructive comments.

PR #11961, opened 2022-06-29
[WIP] wxpython: Fix build error

stahta01:
Something in the build environment changed and it broke package build.

Edit rewrote waf configure and build to fix build error.

Tim S.

### DIFF
--- a/mingw-w64-wxPython/PKGBUILD
+++ b/mingw-w64-wxPython/PKGBUILD
@@ -6,7 +6,7 @@ _wx_basever=3.1
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=4.1.2a1.dev5434+7d45ee6a
-pkgrel=2
+pkgrel=3
 pkgdesc="A wxWidgets GUI toolkit for Python (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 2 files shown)


MehdiChinoune:
We have fixed shebang so it no longer requires MSYS python to be installed.
You should fix building wxPython with MINGW python instead of using the MSYS one.

stahta01:
I have made it slightly farther into the waf configure step; but, still a lot to do.

Tim S.

mingw-w64-wxPython/PKGBUILD
```diff
@@ -18,22 +18,25 @@ depends=("${MINGW_PACKAGE_PREFIX}-python"
          "${MINGW_PACKAGE_PREFIX}-python-pillow"
          "${MINGW_PACKAGE_PREFIX}-python-six"
          "${MINGW_PACKAGE_PREFIX}-wxmsw${_wx_basever}")
-makedepends=("${MINGW_PACKAGE_PREFIX}-doxygen"
+makedepends=('patch'
```
MehdiChinoune, line 21: `patch` is part of base-devel, no need to list under `makedepends`

mingw-w64-wxPython/PKGBUILD
```diff
@@ -71,17 +78,26 @@ build() {
       --no_allmo --no_magic --regenerate_sysconfig \
         dox touch etg sip
 
+  echo "MSYSTEM := $MSYSTEM"
+
   local _jobs=${MAKEFLAGS:--j1}
   MSYS2_ARG_CONV_EXCL="--prefix=;--install-scripts=;--install-platlib=" \
-  CC_NAME=${CC} CXX_NAME=${CXX} \
+  CC=${MINGW_PREFIX}/bin/cc.exe \
+  CXX=${MINGW_PREFIX}/bin/${CXX}.exe \
```
MehdiChinoune, line 86:
```suggestion
  CXX=${MINGW_PREFIX}/bin/c++.exe \
```

mingw-w64-wxPython/PKGBUILD
stahta01: Since, you moved the directions so many times I have no idea what packages should normally be installed!

Tim S.

MehdiChinoune: https://www.msys2.org/dev/new-package/#building-the-package

PR #11967, opened 2022-06-30
osgearth: update to 3.3

mmuetzel:
The release tarball doesn't contain the git submodules. Checkout the tagged git commit instead.

Other than that, just a minor cleanup. And install the license.


mmuetzel:
Wrong calling convention for some OpenGL(?) functions on 32-bit...

PR #11979, opened 2022-07-01
lua: Update to 5.4.4

osch:
Upate Lua to version 5.4.4


Biswa96:
Thank you for the contribution. I have added the required changes to fix the clang builds. It is not related to the lua update but only specific to the build environment.

PR #11981, opened 2022-07-02
Added clangarm64 support to fftw

hmartinez82:


Biswa96:
Is the patch in upstream?

hmartinez82:
@Biswa96  no it's not. I would've updated the version instead if it already was :)

PR #11982, opened 2022-07-02
wxpython: Fix build error; by

stahta01:
rewriting waf configure and build.
Also, add local copy of waflib so it can be edited.

Edit: No longer using local copy of waflib; instead, added/edited patch files.

### DIFF
--- a/mingw-w64-wxPython/PKGBUILD
+++ b/mingw-w64-wxPython/PKGBUILD
@@ -6,7 +6,7 @@ _wx_basever=3.1
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=4.1.2a1.dev5434+7d45ee6a
-pkgrel=2
+pkgrel=3
 pkgdesc="A wxWidgets GUI toolkit for Python (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
@@ -24,16 +24,18 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-doxygen"
              "${MINGW_PACKAGE_PREFIX}-waf"
              "${MINGW_PACKAGE_PREFIX}-cc"
              # for waf
-             "python" "python-setuptools")
+             "${MINGW_PACKAGE_PREFIX}-python-setuptools")
 options=('strip' 'staticlibs' 'buildflags')
 source=(#"https://files.pythonhosted.org/packages/b0/4d/80d65c37ee60a479d338d27a2895fb15bbba27a3e6bb5b6d72bb28246e99/${_realname}-${pkgver}.tar.gz"
         "https://wxpython.org/Phoenix/snapshot-builds/${_realname}-${pkgver}.tar.gz"
         '201-Remove-obsolete-entries-from-Doxyfile.patch'
-        '301-Revert-some-of-python-sip-changes.patch')
+        '301-Revert-some-of-python-sip-changes.patch'
+        '501-MSys2-fixes-for-waf-configure.patch')
 noextract=("${_realname}-${pkgver}.tar.gz")
 sha256sums=('9ab0b6a42d6bb6eacd45ba70f0e9940af56e1711596374c72b4a9110761e36db'
             '97b5c9dd353f2c570b7a940585868c1127d10a96a3442cd96706359410494aa8'
-            '84f836f22f4bbe3f0644040188997fe82ad8db2d4ece6ae081f99d405aff7aa5')
+            '84f836f22f4bbe3f0644040188997fe82ad8db2d4ece6ae081f99d405aff7aa5'
+            '35f2a3ac7c693306f81322221fd2c132baf235c74f1ea2f873cdf8e5a01453d5')
 
 prepare() {
   plain "Extracting ${_realname}-${pkgver}.tar.gz due to symlink(s) without pre-existing target(s)"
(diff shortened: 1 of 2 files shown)


stahta01:
Per https://status.gitlab.com/ gitlab is doing an upgrade; so, this PR is failing.

mingw-w64-wxPython/PKGBUILD
```diff
@@ -49,7 +61,14 @@ prepare() {
...
 
-  # backport wxPython git changes 4xx
+  # MSys2 patches need to fix build problems
+  patch -Np1 -i ${srcdir}/501-MSys2-fixes-for-waf-configure.patch
+
+  sed -i "s|isMsys2 = False|isMsys2 = True|" wscript
+  sed -i "s|COMPILER = 'msvc'|COMPILER = 'mingw32'|" buildtools/config.py
+  # Work around to fix error on trying to find MSVC compiler on CI Server
+  sed -i "s|'msvc', 'gcc',|'gcc',|" waflib/Tools/compiler_c.py
+  sed -i "s|'msvc', 'g++',|'g++',|" waflib/Tools/compiler_cxx.py
```
Biswa96, line 71: Would it be better to do this in a patch file?

Biswa96:
Is it possible to customize the mingw-w64-waf package and then use it here? Also any plan to upstream the changes?

mingw-w64-wxPython/PKGBUILD
```diff
@@ -73,15 +92,22 @@ build() {
...
+  CC_NAME="$( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] && echo "clang" || echo "gcc" )" \
+  CXX_NAME=${CC_NAME} \
+  LDFLAGS="${LDFLAGS} $(${MINGW_PREFIX}/bin/python-config --ldflags)" \
   PYTHON_CONFIG="${MINGW_PREFIX}/bin/python-config" \
-    "${MINGW_PREFIX}/bin/waf" \
+    waf \
       --prefix="${MINGW_PREFIX}" --python="${MINGW_PREFIX}/bin/python.exe" \
-      --check-cxx-compiler=${CXX} --check-c-compiler=${CC} --color=yes --jobs=${_jobs#-j} \
-      --wx_config=wx-config-${_wx_basever} \
+      --check-cxx-compiler="$( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] && echo "clang++" || echo "g++" )" \
```
Biswa96, line 103: Can `CXX` variable be used here?

mingw-w64-wxPython/PKGBUILD
stahta01: Maybe; but, not if I hope to have the patch used by upstream.
And, the waflib files should not need changed; but, it was the only way to get the build to work on the CI server.
Right now just getting it to build is my goal. The last time I got it to build all I got was complaints about using /usr/bin/python. At least this time the feedback is more reasonable; me, I think waf is broken and that is why I need to edit the waflib files. But, a patch for buildtools/config.py makes some sense; but, it seems to be impossible to test other version of waflib right now, I hope an earlier or later version might not need changed.

Tim S.

mingw-w64-wxPython/PKGBUILD
stahta01: Please read the lines above where I had to change CXX to the complete path; I am still working to confirm what changes are needed; I spent over 40 hours getting this to build because the CI server has MSVC detection problems. Does the CI server have MSVC on it? Or any hint why MSVC detection fails?

stahta01:
I think the correct thing is to do all the changes needed in wscript; but, I have not yet figured out what changes are correct to do there.

stahta01:
What about my new idea of four patch files?
1. msys2 patch file like I already have; but, slightly smaller.
2. buildtools/config.py patch file instead of sed command.
3. gcc patch file for wscript to hardcode gcc compiler
4. clang patch file for wscript to hardcode clang compiler
The prepare would apply either gcc or clang patch file based on "${MINGW_PACKAGE_PREFIX} == *-clang-*".

Does the selection of patches based on "${MINGW_PACKAGE_PREFIX}" meet the design rules?

Tim S.

mingw-w64-wxPython/PKGBUILD
mati865: Can we use `cc` and `c++` for better compatibility with Clang?

stahta01:
Will code like below be okay in prepare() ?

```
    patch -Np1 -i ${srcdir}/503-Use-CLANG-compiler.patch
  else
    patch -Np1 -i ${srcdir}/502-Use-GCC-compiler.patch
  fi
```
Tim S.

mingw-w64-wxPython/PKGBUILD
stahta01: Please read post below asking about patches in prepare being different based on clang build or not.

stahta01:
I have hopes of figuring out how to use waf enough to condense the three 5xx patches into a single patch and passing an option to the waf command.

stahta01:
FYI: Based on wxPython git repo there likely will not be an 4.1.2 release because the repo now shows work on 4.2.0 Alpha  release.

MehdiChinoune:
Could you report an issue upstream about defaulting to MSVC.

stahta01:
> Could you report an issue upstream about defaulting to MSVC.

Probably not worth the effort without at least a partial patch.
But, I will review my single 5xx patch and see if it can be made ready to apply to git master with it still defaulting to MSVC compiler and maybe they can figure out a way to add waf option to trip it.

Tim S.

stahta01:
To all: Does anyone know why checking for MSVC ( on the CI Server) resulted in an error when done by waf configure?
Or, does anyone know if MSVC is on the CI server?
I guessed it was a permission issue.

Tim S.

PR #11985, opened 2022-07-02
WIP nss: update to 3.80

lb90:
(to be merged after the updated nspr hits the official repos)


mmuetzel:
The clang builds failed with:
```
  make[3]: gcc: No such file or directory
  make[3]: *** [../../coreconf/rules.mk:292: WINNT5.0_x86_64_gcc_64_OPT.OBJ/arena.o] Error 127
  make[2]: *** [../coreconf/rules.mk:44: base] Error 2
```

Maybe a hardcoded `gcc` that could be replaced by a `${CC}`?

Likely here in the `PKGBUILD`:
```
  make -j1 -C nss OS_TARGET=WINNT OS_RELEASE=5.0 XP_WIN=1 \
    NSINSTALL="${srcdir}"/build-${CARCH}/nsinstall NS_USE_GCC=1 ${conf64} \
    NSS_USE_SYSTEM_SQLITE=1 USE_SYSTEM_ZLIB=1 ZLIB_LIBS=${MINGW_PREFIX}/lib/libz.dll.a \
    CC=gcc RC=windres RANLIB=ranlib
```

Can `CC=gcc` just be dropped from the `make` flags?

lb90:
Hi @mmuetzel! Yes, it's the hardcoded CC=gcc that makes the clang build fail. It's already fixed locally, I'll push un update to this PR once the updated nspr is available from the repos. Thanks! 🙂

mmuetzel:
> It's already fixed locally

Perfect. 👍

> I'll push un update to this PR once the updated nspr is available from the repos.

Afaict, it is already out of staging and in the repos.

lb90:
Note for reviewers:

* `NS_USE_GCC=1 CC="cc"` tells the Makefile to use the default compiler (either `gcc` or `clang`), but consider it as gcc-compatible. That way the Makefile can use more advanced features of the toolchain.
* `-Wno-unused-command-line-argument` has been added to extra cflags because the Makefile passes `-mwindows` to every compiler invocation. This argument only has an effect when building EXEs, so `clang` outputs the following warning: `cc: warning: argument unused during compilation: '-mwindows' [-Wunused-command-line-argument]`

mmuetzel:
That seems to be working. 👍
Afaict, `CC` is already set in the environment by `makepkg-mingw`. Is it possible to drop overriding it in the `make` flags. E.g., in case the `cc` alias will be dropped or renamed at some point in the future. (Not that this is likely to happen. But still...)

lb90:
I tried that at first, but `make` ended up invoking `gcc`. That's probably due to `NS_USE_GCC=1`. After that I tried commenting out those [three lines](https://github.com/nss-dev/nss/blob/master/Makefile#L91), but didn't make any difference

lb90:
Does `ifdef CC` evaluate to true if CC is an environment variable?

mmuetzel:
I see. `CC=cc` will probably also be fine for a *very* long time.

> Does `ifdef CC` evaluate to true if CC is an environment variable?

IIUC, probably not...
https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html

Something like `ifneq ("x${CC}", "x")` might do that. But your current rule looks better imho.

lb90:
Ok, here it is: https://github.com/nss-dev/nss/blob/master/coreconf/WIN32.mk

lb90:
Green lights! ✔️ 

Is it ok for you to patch WIN32.mk? Let me know, otherwise I can change the PR, no problem!

PR #11986, opened 2022-07-02
python-setuptools: update to 62.6.0

naveen521kk:
also default to stdlib distutils as the embedded one doesn't support mingw


lazka:
lgtm

PR #12004, opened 2022-07-05
zint: Add package

mvzlb:


mingw-w64-zint/PKGBUILD
```diff
@@ -0,0 +1,64 @@
...
+
+_realname=zint
+pkgbase=mingw-w64-${_realname}
+pkgname=(
+  "${MINGW_PACKAGE_PREFIX}-zint"
+  "${MINGW_PACKAGE_PREFIX}-zint-qt"
+)
+pkgver=2.10.0
+pkgrel=1
+pkgdesc='Barcode encoding library supporting over 50 symbologies'
```
mmuetzel, line 14: The `pkgdesc` of MINGW packages should end with ` (mingw-w64)`. See also the templates, e.g.:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball.sh

mingw-w64-zint/PKGBUILD
```diff
@@ -0,0 +1,65 @@
...
+
+  DESTDIR="${pkgdir}" cmake --install build
+
+  echo "MINGW_PREFIX: ${MINGW_PREFIX}"
+  echo "pkgdir: ${pkgdir}"
+  rm "${pkgdir}${MINGW_PREFIX}/bin/zint-qt.exe"
+}
+
+package_zint-qt() {
+  pkgdesc='Zint Barcode Studio GUI'
```
mmuetzel, line 46: The `pkgdesc` of MINGW packages should end with ` (mingw-w64)`.

mingw-w64-zint/PKGBUILD
```diff
@@ -0,0 +1,65 @@
...
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    ${MINGW_PREFIX}/bin/cmake -S ${_realname}-${pkgver}-src -B build -G Ninja -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX}
+
+  ${MINGW_PREFIX}/bin/cmake --build build
+}
+
+package_zint() {
+  depends=("${MINGW_PACKAGE_PREFIX}-libpng")
+
+  DESTDIR="${pkgdir}" cmake --install build
```
mmuetzel, line 38: Consider using `cmake` including the respective prefix.

mmuetzel:
Ianal, but I think the GPL requires the license to be included in the distribution. Even if it isn't, it doesn't hurt much to be on the safe side.
See, e.g., the command close to the end of the `PKGBUILD` template for how the license could be installed.

The `.cmake` files are installed at an interesting location (e.g., `/mingw64/share/apps/cmake/modules/FindZint.cmake`). Are they found at that location for dependent projects? (That's an honest question because I don't know where cmake will be looking for such files.)
Many other projects install these files, e.g., at `/mingw64/lib/cmake`...

mvzlb:
> The `.cmake` files are installed at an interesting location (e.g., `/mingw64/share/apps/cmake/modules/FindZint.cmake`). Are they found at that location for dependent projects? (That's an honest question because I don't know where cmake will be looking for such files.)

No, not with the default modules path. `share/apps/cmake/modules/` seems to be a KDE4 thing: https://cmake.org/cmake/help/latest/module/FindKDE4.html

mmuetzel:
The PKGBUILD and patch lgtm.
I tested the build artifacts for mingw64. They also seem to be working correctly here.

LGTM. 🚀

mvzlb:
> The PKGBUILD and patch lgtm. I tested the build artifacts for mingw64. They also seem to be working correctly here.
> 
> LGTM. rocket

Thanks!

PR #12007, opened 2022-07-06
[WIP] wxpython 4.2.0a1 and wxWidgets 3.2.0 related changes

stahta01:
Major changes include:
1. Renamed mingw-w64-wxWidgets to mingw-w64-wxWidgets3.0
   Removed wx-config and other items from mingw-w64-wxWidgets3.0 package
2. Renamed mingw-w64-wxWidgets3.1 to mingw-w64-wxWidgets
   a. Add split package wxconfig
   b. Add split package msw_cb_headers
   c. Upgrade wxMSW to 3.2.0
3. Updated wxPython to 4.2.0a1 and wxWidgets 3.2.0

Note: The  PR #12011 needs applied before this one.


stahta01:
> I don't agree with swapping out wxwidgets package with 3.1, currently 3.1 series is considered still as a development package and even if it's considered stable enough, shouldn't supplant the current stable until an actual new stable is released.

True; but you do understand what draft and work in progress mean, right?

Tim S.

stahta01:
Any opinion on when to change this PR from draft?
The day of the wxWidgets release and a week after the release both sounds good to me.

Tim S.

ZachBacon:
> I don't have a complete picture of this transition. But I slightly agree with @ZachBacon. 
> IIUC, wxWidgets 3.2 breaks API compatibility to wxWidgets 3.0. If that is true (and please correct me if I'm wrong), we should probably give dependent packages some time to adapt to the new API before defaulting to it. I don't know what would be a good time frame for that. Maybe a couple of months after the release of wxWidgets 3.2? Half a year? Longer?

From the release notes, it looks like it's still Abi compatible with 3.0 

"We recommend using this version for any new wxWidgets applications. Existing applications using wxWidgets 3.0 or 3.1 shouldn't require any special effort to upgrade to this version too, so please try upgrading them to it."

https://github.com/wxWidgets/wxWidgets/releases/tag/v3.2.0

mmuetzel:
Thanks for the hint. 👍
It's still not entirely clear to me though. IIUC, they only promise ABI compatibility for the 3.2.x release series.
The following sentence might mean that 3.0.x and 3.2.x are still API compatible. So, rebuilding all dependencies might be enough. (And they wouldn't need to change their code.) Is that correct?
That might make the transition a bit easier...

stahta01:
> I don't have a complete picture of this transition. But I slightly agree with @ZachBacon. IIUC, wxWidgets 3.2 breaks API compatibility to wxWidgets 3.0. If that is true (and please correct me if I'm wrong), we should probably give dependent packages some time to adapt to the new API before defaulting to it. I don't know what would be a good time frame for that. Maybe a couple of months after the release of wxWidgets 3.2? Half a year? Longer?

The packages in this repo should all stay at 3.0; that is what my other PR did.
And, both wxWidgets3.0 and wxWidgets (3.2) can be installed at the same time; I have been thinking about this change for over two years. The other PR was already applied see #12011

Tim S.

stahta01:
List of changes to help third party conversion to wxWidgets versions 3.0 and 3.2.
Note: All packages using wxWidgets3.1 will need work. And, packages using static wxWidgets 3.0 will need work.

wxWidgets3.0 package changes.
Make into split package
1. wxmsw3.0
2. wxconfig; that by default finds wxmsw3.0 dll.
3. wxWidgets package that results in wxmsw3.0 and wxconfig being installed.
Review of package wxmsw3.2 to confirm it has no issues with meta wxWidgets being installed.
Note: the wxconfig from the two packages, 3.0 and 3.2. will conflict with each other.

I just got paid programming job; so, it could be a few weeks before I can do above.

Tim S.

MehdiChinoune:
https://wxwidgets.org/news/2022/07/wxwidgets-3.2.0-final-release/

I think this PR should be split into two, one for wxWidgets upgrade and the other for wxPython

stahta01:
I am working on another PR for the wxWidgets 3.0 and 3.2 packages; it is being written to allow third party packages to build using the new packages with minor changes for most of them.

Tim S.

stahta01:
Closing this till after I get my new wxWidgets 3.2.0 split package to build.

PR #12011, opened 2022-07-06
Change packages to use wxWidgets 3.0.x

stahta01:
These changes make packages use wxWidgets 3.0.x instead of using the version returned by wx-config.
This is needed to prevent these four packages from using wxWidgets 3.2.0 when PR #12007 is applied.

Tim S.


lazka:
thanks!

PR #12025, opened 2022-07-10
[WIP] Switch to wxWidgets 3.2.0 as the standard production wxWidgets package.

stahta01:
Edit: wxWidgets 3.0 is changed to a split package and wxWidgets 3.2 is also a split package with this PR.

I am not used to doing split or meta packages so please review that.
And, I have not used groups names very much.
Also, please give feedback on package descriptions.
I added usage_notes.txt for people who use these packages.
I have done my best to make it easy for third party users to continue
to use the wxWidgets 3.0 package with only minor changes.

Tim S.


stahta01:
Since, no feed back closing

PR #12038, opened 2022-07-14
[WIP] wxWidgets: convert to split packages wxmsw, wxconfig, bakefile, local…

stahta01:
…e, and wxrc.

Also, add wxWidgets meta package that depends on wxmsw, wxconfig,
locale, and wxrc.


stahta01:
Decided to close while I finish wxWidgets 3.2 package

Issue #11553, opened 2022-04-24
libc++.a missing symbols (new and delete)


shithappens2016:
@Biswa96
@lazka
@ImperatorS79
@MehdiChinoune
@jeremyd2019

```C++
#include <string>
#include <iostream>

int main() {
  auto test = new std::string("test");
  std::cout << *test << std::endl;
  delete test;
  return 0;
}
```

```bash
/mingw64/bin/clang++ test.cc -stdlib=libc++ -static -fuse-ld=lld
```

```bash
ld.lld: error: undefined symbol: operator new(unsigned long long)
>>> referenced by /tmp/main-f1fa00.o:(main)
>>> referenced by libc++.a(string.cpp.obj):(std::__1::to_string(int))
>>> referenced by libc++.a(string.cpp.obj):(std::__1::to_string(long))
>>> referenced 156 more times

ld.lld: error: undefined symbol: operator delete(void*)
>>> referenced by /tmp/main-f1fa00.o:(main)
>>> referenced by /tmp/main-f1fa00.o:(main)
>>> referenced by libc++.a(exception.cpp.obj):(std::nested_exception::~nested_exception())
>>> referenced 1191 more times
```

LABPBS:
cant reproduce with clang64 toolchain

```sh
/clang64/bin/clang++ test.cc -static
./a.exe
test
```

its *strongly* adviced to use clang toolchain if you want to compile with clang *specially* if you also want to link with lld *and* use libc++ those missing symbols are from libc++abi.a (the clang equivalent of gcc's libsupc++.a)

more info

https://www.msys2.org/docs/environments/#gcc-vs-llvmclang
https://github.com/mstorsjo/llvm-mingw/

mati865:
It's Binutils `strip` bug.

shithappens2016:
@LABPBS
Thanks for the info, but I hope this can be fixed.

shithappens2016:
@mati865
Is this related to the last clang upgrade（#11087）

str4d:
> @mati865 Is this related to the last clang upgrade（#11087）

@shithappens2016 I don't believe so; I encountered this missing symbols problem with 13.0.1-1 as well (and had to [use 13.0.0-3 instead](https://github.com/zcash/zcash/commit/fdb5709e7ebfdc617902fb05ecb090a9e766ebe7)).

PR #12042, opened 2022-07-14
Disable `strip` for libc++

mati865:
Since nobody wanted to work on manual stripping with `llvm-strip` lets just disable it. Size increase is only 0.75 MiB (all packages combined) and it might be also helpful for debugging.

Fixes https://github.com/msys2/MINGW-packages/issues/11553


Biswa96:
A comment above the `options=()` might be helpful.

lazka:
I added a regression test here: https://github.com/msys2/msys2-tests/pull/7

lazka:
I added a regression test here: https://github.com/msys2/msys2-tests/pull/7

PR #12043, opened 2022-07-15
SDL2_mixer: build with default config flags

fabiangreffrath:


fabiangreffrath:
This will enable support for playing Ogg files using stb_vorbis,
support for playing FLAC files using dr_flac and support for
playing MP3 files using dr_mp3.

Dependencies on external libraries are reduced accordingly.

Biswa96:
May I ask why this change is required? Some GNU/Linux distributions do include those dependencies.
* Arch: https://archlinux.org/packages/extra/x86_64/sdl2_mixer/
* Ubuntu: https://packages.ubuntu.com/jammy/libsdl2-mixer-2.0-0

fabiangreffrath:
SDL upstream has decided to switch to using embedded micro-libraries for the most populr formats by default instead of relying on third-party libraries. This change was introduced in the most-recent 2.6 release, I guess the Linux distributions have not adjusted their packaging yet. For us this means we can redistribute SDL-related libraries with our game engines without pulling a massive dependency chain behind us.

fabiangreffrath:
Also, Linux distributions tend to avoid embedded libraries in favor of shared libraries to prevent mass-rebuilds of affected packages in case of security issues. Furthermore, SDL is mostly targeted at games and thus graphical desktop environments which usually have most audio format support libraries installed by means of e.g. media players, so the costs of pulling in these dependencies by SDL packages are negligible.

Biswa96:
> Linux distributions tend to avoid embedded libraries in favor of shared libraries to prevent mass-rebuilds of affected packages in case of security issues.

That looks a valid reason to me. But opinions may vary. Let see what others think.

fabiangreffrath:
Can we just get rid of the extra library dependencies, please?

MehdiChinoune:
> Also, Linux distributions tend to avoid embedded libraries in favor of shared libraries

We do the same here.
> to prevent mass-rebuilds of affected packages in case of security issues.

and for less package size.

lazka:
I've asked upstream: https://github.com/libsdl-org/SDL_mixer/issues/427

fabiangreffrath:
> I've asked upstream: [libsdl-org/SDL_mixer#427](https://github.com/libsdl-org/SDL_mixer/issues/427)

> I would recommend building with all codecs, using the built-in decoders as much as possible to minimize system dependencies.

fabiangreffrath:
So, another three weeks have passed. What else is needed?

fabiangreffrath:
Okay, so this is two months old now. You wanted to ask upstream, you got your reply from upstream recommending to go with the built-in codecs. What else is missing?

fabiangreffrath:
Please do either apply this or close it with an explanation, but don't let people wait for months. Thanks!

fabiangreffrath:
Looks like the built-in codecs get preference anyway, unless a very specific set of configure flags is passed, rendering the current set of configure flags entirely useless -- so they could just as well get removed altogether (which is what this PR suggests):

```sh
$ objdump -x /mingw32/bin/SDL2_mixer.dll | grep "DLL Name"
        DLL Name: libgcc_s_dw2-1.dll
        DLL Name: KERNEL32.dll
        DLL Name: libmpg123-0.dll
        DLL Name: msvcrt.dll
        DLL Name: libopusfile-0.dll
        DLL Name: SDL2.dll
        DLL Name: WINMM.DLL
```

PR #12047, opened 2022-07-15
openshadinglanguage: fix cmake config files

mmuetzel:
There are two issues with the cmake config files of OSL:
1. One of the files is moved after it is installed. That leaves the cmake files pointing to the wrong locations.
2. A `sed` rule is used to replace the literal prefix of the build system with `${_IMPORT_PREFIX}`. But that variable isn't set in one of the .cmake files.

The file that is moved is a "module" library. IIUC, those are libraries that aren't meant to be linked to. But their symbols should be loaded on runtime. It looks like `cmake` installs those in the `lib` directory (not `bin`). Maybe, it does that for a reason. Instead of trying to change where `cmake` installs "module" libraries (which would probably also lead to consistent .cmake files), this change removes moving that library after it is installed.

It also adds a patch that fixes the other .cmake file.


lazka:
Are we sure it can find `osl.imageio.dll` at runtime?

mmuetzel:
IIUC, OpenImageIO needs the environment variable `OIIO_LIBRARY_PATH` to be set to the path with the plugins. The only documentation about it that I can find advices for it to be set to the "`lib` directory where OpenImageIO is installed": https://github.com/OpenImageIO/oiio#building-and-installation

So, I guess it doesn't matter where we install it as long as the user sets the respective path in the environment. We might be even closer to the upstream documentations. Since the plugin library isn't thought to be used at load time anyway, it might be even better to not install it somewhere that appears in `PATH`.

lazka:
ok, thanks for investigating. Let's see

PR #12055, opened 2022-07-17
opencolorio: update to 2.1.2

mmuetzel:
This is a major version update. Most (or all) of the existing patches aren't needed any more.

This has a dependency cycle with OpenImageIO. Is there a way to tell the CI to build OpenColorIO first?


mmuetzel:
The package-grokker complains about openimageio. That is probably because the CI rebuilt it before the updated opencolorio was built.
If I build it (on MINGW64) locally in the correct order (opencolorio first, then openimageio), openimageio links to the new dll with the version suffix. And it builds correctly with the updated version.

mmuetzel:
The autobuild runners likely need some indication to build opencolorio first to break the cycle.

Issue #11864, opened 2022-06-17
Meson errors with Python 3.10


lazka:
It randomly fails when calling into meson scripts:

1) mesa:

> FAILED: src/util/u_unfilled_gen.c 
  "D:/M/msys64/mingw64/bin/meson" "--internal" "exe" "--capture" "src/util/u_unfilled_gen.c" "--" "D:\M\msys64\mingw64\bin/python3.EXE" "../../src/util/indices/u_unfilled_gen.py"

>FAILED: src/gallium/targets/osmesa/osmesa.dll.p/osmesa.dll.symbols 
"D:/M/msys64/mingw64/bin/meson" "--internal" "symbolextractor" "C:/_/mingw-w64-mesa/src/mesa-22.1.2/build/windows-x86_64" src/gallium/targets/osmesa/osmesa.dll "src/gallium/targets/osmesa/osmesa.dll.a" src/gallium/targets/osmesa/osmesa.dll.p/osmesa.dll.symbols 

2) gobject-introspection

> FAILED: girepository/libgirepository-1.0-1.dll.p/libgirepository-1.0-1.dll.symbols 
"D:/a/msys64/clang64/bin/meson" "--internal" "symbolextractor" "C:/M/mingw-w64-gobject-introspection/src/build-x86_64-w64-mingw32" girepository/libgirepository-1.0-1.dll "girepository/libgirepository-1.0.dll.a" girepository/libgirepository-1.0-1.dll.p/libgirepository-1.0-1.dll.symbols 

There was one more packages during the Python rebuild that failed that way (with "symbolextractor" I think), but I didn't write it down sadly.

eli-schwartz:
It's completely bizarre that the symbolextractor fails *without* emitting an error message of any sort. It's reminiscent of https://github.com/pypa/pip/issues/10875 (was a bug in the launcher *.exe files pip uses on Windows).

lazka:
thanks, good point. I can try porting meson from setuptools to builder/installer so it gets a newer launcher.

naveen521kk:
>  (was a bug in the launcher *.exe files pip/distlib uses on Windows, reproducible in CI but not running interactively).

Hmm, meson was built with setuptools launcher which shouldn't have this issue afaik.

lazka:
same error with the python-installer launcher (#11865)

eli-schwartz:
[...]

PR #12184, opened 2022-07-18
babl: update to 0.1.92

mmuetzel:
Also fix segmentation fault on i686.


Biswa96:
Packages with meson can not be built in CI now. See the pinned issue.

mmuetzel:
This is the intermittent issue #11864. Could someone with the required permissions please restart the MINGW64 runner?

Biswa96:
Yes, the meson issue can not be reproduced every time. We are getting lucky here. But this may fail in autobuild step and one may have to restart CI multiple times.

PR #12187, opened 2022-07-19
[RFC] wxwidgets3.2: Add package version 3.2.0

stahta01:
I decided that renaming wxWidgets3.1 was not a good idea; because other people might be using it for production work. Adding this package should allow the users to decide when to switch wxWidgets versions.

Please give feedback on this package; once this package is ready; I plan to change wxWidget 3.0 and 3.1 into split packages.

Edit: I still need to test that this package works; plan is to build KiCAD 6.0.6 using it.
Edit2: Found and fixed the first error building wxPython 4.2.0a1; once that builds will work on building KiCAD
Edit3: The CI server refuses to build KiCAD; so, I am trying to build CodeLite 16.1.0; but, it segfaults when ran.
It looks to be wxWidgets 3.2.0 issue; so, working on determining if it is an CodeLite or wx issue before continuing. 
Edit4: It was a known CodeLite issue; when upgrading CodeLite version and CodeLite program segfaults before display is shown delete the AppData codelite settings folder is often needed.
Edit5: Codelite builds and runs okay after I remove wxcrafter it seems to have problems with wxWidgets 3.2.0

Tim S.


stahta01:
I am now going to work on making wxWidgets 3.0 and 3.1 into split packages.

stahta01:
I give up! It a waste of my time and energy to try to get this package to be approved.
What I want is not possible with the way Arch Linux did it and no one here seems to care!

PR #12205, opened 2022-07-24
harfbuzz: update to 5.0.1

MehdiChinoune:


mingw-w64-harfbuzz/PKGBUILD
```diff
@@ -4,7 +4,7 @@
 _realname=harfbuzz
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=4.4.1
+pkgver=5.0.1
```
naveen521kk, line 7: Since this is a major update shouldn't dependencies be rebuilt?

MehdiChinoune: > Since this is a major update shouldn't dependencies be rebuilt?

Package grokker said No.

MehdiChinoune: > Since this is a major update shouldn't dependencies be rebuilt?

Package grokker said No.

PR #12221, opened 2022-07-25
python-sphinx-basic-ng: New package

podsvirov:


Biswa96:
Add `options=('!strip')`

podsvirov:
> Add `options=('!strip')`

I'm having trouble with this option. Should we add if there are no launchers in the package (like in #12208)?

Biswa96:
Agree. I was thinking about future proofing 😄

PR #12223, opened 2022-07-25
glfw: Add documentation

podsvirov:


Biswa96:
Is it possible move the docs to a separate package like Arch does https://archlinux.org/packages/community/x86_64/glfw-doc/ ?

podsvirov:
@Biswa96, please see [there](https://github.com/msys2/MINGW-packages/pull/6336#issuecomment-607908235).

Biswa96:
OK, system overridden 🤖

PR #12194, opened 2022-07-21
wxWidgets: Convert to split package

stahta01:

PR #12224, opened 2022-07-25
wxwidgets3.2: Add package version 3.2.0

stahta01:


stahta01:
It would be best if PR #12194 was applied before this package; but, either can be applied first.

Tim S.

stahta01:
I am trying to figure out how to get CodeLite IDE to build and run; currently wxCrafter plugin crashes on startup.
So, I am testing various wx3.2.0 builds

stahta01:
Gave up on fixing CodeLite wxCrafter plugin crash see #12271

stahta01:
I am testing the wxGTK split packages and found an error. That is why I added back the WIP

stahta01:
I was able to build the Code::Blocks IDE code CB project linking against wxGTK/Win after a moderate amount of editing.
And, it even ran and displayed without error; did not try use it after it was displayed.

stahta01:
Any new feedback on this package?

lazka:
I'll have a look once I've cleaned up the 3.0 dependencies

lazka:
What about renaming "base-libs" to "common-libs", or "common" to "base", so there is some symmetry?

stahta01:
> What about renaming "base-libs" to "common-libs", or "common" to "base", so there is some symmetry?
> 
> I've pushed a small fix: [92433ff](https://github.com/msys2/MINGW-packages/commit/92433ff74c98dec0279f2315894830f71052c1a1)

wxWidgets uses wxBase; but, Arch Linux use common. I can see common-libs as being more inline with Arch Linux.
So, that will be the way I think the others will most accept; so, I will change it that way if that is what the msys2 team.
Tim S.

lazka:
You don't have to change it, it's just something that came to mind :)

asmwarrior:
Hi, @stahta01 since wxWidgets 3.2.1 has released, can you take time to update msys2's package to wx 3.2.1? Thanks.

stahta01:
Okay, I have already updated the package; but, I have not really tested it.

Tim S.

PR #12261, opened 2022-07-26
[new-package] python-hatch-vcs 0.2.0

MehdiChinoune:


MehdiChinoune:
This package is required by other python packages like `jsonschema` and `apipkg`

PR #12262, opened 2022-07-27
ninja: use $TERM to decide when to use fancy output

oscarfv:
... such as one-line progress messages, color, etc.

### DIFF
--- a/mingw-w64-ninja/PKGBUILD
+++ b/mingw-w64-ninja/PKGBUILD
@@ -4,7 +4,7 @@ _realname=ninja
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.11.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Ninja is a small build system with a focus on speed (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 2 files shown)


oscarfv:
What's preventing the merge of this PR?

lazka:
We'll enable conpty in the near future, which makes this obsolete, but can't hurt I guess.

lazka:
This breaks ninja, I have to revert it.

```
# CMakeLists.txt
project(foo)
message(STATUS ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
```

`CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES` is empty with the patched ninja version.

oscarfv:
> We'll enable conpty in the near future, which makes this obsolete, but can't hurt I guess.

This benefits any application that launches ninja and has terminal emulation capabilities.

> This breaks ninja, I have to revert it.
> 
> ```
> # CMakeLists.txt
> project(foo)
> message(STATUS ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
> # cmake .
> ```
> 
> `CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES` is empty with the patched ninja version.
> 
> cmake logs indicate that cmakes somehow gets ninja debug output and tries to parse it as gcc output, which fails.

Thanks for the analysis. Under mintty $TERM is set to `xterm`. This is what ninja sees when it is invoked by cmake.

I'm afraid that it would be necessary to implement an explicit option to enable ninja's fancy output on Windows, because there is no general mechanism to decide if the output is going to a terminal (real or emulated) or not.

Sorry for the inconvenience.

PR #12266, opened 2022-07-27
starpu: fix and enable opencl

3rav:


mingw-w64-starpu/patch.patch
```diff
@@ -0,0 +1,35 @@
...
++			      cl_event *        /* event */
++			      ) CL_API_SUFFIX__VERSION_1_2;
++
++extern CL_API_ENTRY cl_int CL_API_CALL
++soclEnqueueBarrierWithWaitList(cl_command_queue /* command_queue */,
++			       cl_uint           /* num_events_in_wait_list */,
++			       const cl_event *  /* event_wait_list */,
++			       cl_event *        /* event */
++			       ) CL_API_SUFFIX__VERSION_1_2;
+
```
MehdiChinoune, line 33: beside identation, I don't see any difference.
Please, when adding a patch keep it minimal. Don't add your formatting (spaces and identations), so we could see the difference.

3rav: I received this patch file from the developers after reporting a build problem for mingw32

MehdiChinoune: O.K
They have added `CL_API_CALL`

3rav: https://gitlab.inria.fr/starpu/starpu/-/commit/3e1de7001c67c25faaffb9c03d7edc87a118429f

mingw-w64-starpu/PKGBUILD
```diff
@@ -1,32 +1,44 @@
...
+pkgrel=4
 pkgdesc='StarPU is a task programming library for hybrid architectures (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 depends=("${MINGW_PACKAGE_PREFIX}-fftw"
          "${MINGW_PACKAGE_PREFIX}-libssp"
          "${MINGW_PACKAGE_PREFIX}-hwloc"
-         "${MINGW_PACKAGE_PREFIX}-libwinpthread-git")
+         "${MINGW_PACKAGE_PREFIX}-libwinpthread-git"
+         "${MINGW_PACKAGE_PREFIX}-opencl-headers"
```
lazka, line 15: should this be in makedepends maybe?

mmuetzel: I guess  what @lazka meant was that the *headers* are probably just a build time dependency. You'll probably still need the library in run time...

mmuetzel:
Since OpenCL is disabled for 32-bit targets in the build rule, the dependency on OpenCL could probably be added conditionally.

PR #12271, opened 2022-07-27
[WIP] [RFC] codelite: Upgrade to version 16.2.0 and use wx3.2. and standalone wxCrafter

stahta01:
Disable building of wxCrafter plugin because it crashes on startup And, add wxCrafter stand alone app.

Fixing the wxCrafter plugin crash requires some with more knowledge or experience than I have.
The wxCrafter  stand alone app does not crash; so, likely needs an wxWidgets debugging experience and CodeLite plugin system knowledgeable person to fix that problem.

Tim S.


stahta01:
Is adding the wxCrafter exe a good enough work around for not have the wxCrafter plugin?

Tim S.

stahta01:
Someone else did PR; so, let them worry about these issues.

PR #12275, opened 2022-07-27
libgit2: update to 1.5.0

MehdiChinoune:


MehdiChinoune:
Need help with llvm-rc

Biswa96:
> Need help with llvm-rc

The removed hunk in 0001 patch file is required. The changed file should be src/libgit2/CMakeLists.txt instead of src/CMakeLists.txt.

PR #12276, opened 2022-07-28
suitesparse: Add support for CLANGARM64

hmartinez82:


mingw-w64-suitesparse/PKGBUILD
```diff
@@ -32,6 +32,7 @@ sha256sums=('5fb0064a3398111976f30c5908a8c0b40df44c6dd8f0cc4bfa7b9e45d8c647de'
             'e0faab259d5c474c6af51c9b471f645810a308a1c83976275b1969e81e7d8390'
             '8628100f41800e0e4238a8f949fb99e8667e0b58208130606881612d7172835e'
             '1cf6cd948e5178cbcb16bb0d3c97829311250dedda1e9dcbe0c85dd6bdc4461e')
+openmp_flag=$([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] || echo "-fopenmp")
```
Biswa96, line 35: This global variable should be prefixed with an underscore.

Biswa96:
It failed to build in autobuild. The aarch64 version of those dependencies are missing. See the log https://packages.msys2.org/queue

hmartinez82:
@Biswa96 I'll investigate, because it's true, `2022-07-28T06:24:01.9750464Z error: target not found: mingw-w64-clang-aarch64-openmp`  openmp does not exist for arm64, but the package recipe doesn't try to use it when using clang

PR #12278, opened 2022-07-28
python-markdown-it-py: update to 2.1.0

MehdiChinoune:


Biswa96:
Is it possible to add clangarm64 also?

MehdiChinoune:
you could add it later if you want (without bumping `pkgrel`)

PR #12280, opened 2022-07-28
suitesparse: Fix CLANGARM64 detection logic

hmartinez82:


Biswa96:
Need a pkgrel bump for autobuild.

hmartinez82:
@lazka yes, I actually found an error with the original intent of that line. The comments, (including the comments in SuiteSparse.mk) is to not link OpenMP when using Clang.  Now, this is probably from the old days, OpenMP got much better now with the recent LLVM releases. Do you prefer that I only disable OpenMP for AArch64?

lazka:
> Do you prefer that I only disable OpenMP for AArch64?

yes please

hmartinez82:
@lazka done :)

mingw-w64-suitesparse/PKGBUILD
```diff
@@ -18,7 +18,7 @@ source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/S
         "0005-suitesparse-5.8.1-skip-building-Mongoose-GraphBLAS.patch"
         "0006-suitesparse-5.8.1-fix-mp-link-order.patch")
 depends=("${MINGW_PACKAGE_PREFIX}-openblas"
-         $([[ ${MINGW_PACKAGE_PREFIX} != *-clang-*86* ]] || echo "${MINGW_PACKAGE_PREFIX}-openmp")
+         $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || echo "${MINGW_PACKAGE_PREFIX}-openmp")
```
lazka, line 21: gcc doesn't need it, so this shouldn't be needed.

hmartinez82: @lazka so I got it all wrong. Can you confirm this is what we want?
MINGW - Do not use OpenMP
CLANG32/64 - Use OpenMP
CLANGARM64 - Do not use OpenMP

I can't leave `$([[ ${MINGW_PACKAGE_PREFIX} != *-clang-*86* ]] || echo "${MINGW_PACKAGE_PREFIX}-openmp")` as it was originally because I'm trying to disable OpenMP for CLANGARM64

kmilos: Why not for GCC, should be enabled AFAIK? Just that there is no "openmp" package for GCC environments, OpenMP is part of gcc-libs which should be added (unconditionally). So you want to install openmp only for clang32 and clang64 for now, i.e. this was correct before this change. See e.g. the graphicsmagick recipe...

The flag below is ok (enable for everything except clangarm64).

mingw-w64-suitesparse/PKGBUILD
```diff
@@ -32,6 +32,7 @@ sha256sums=('5fb0064a3398111976f30c5908a8c0b40df44c6dd8f0cc4bfa7b9e45d8c647de'
             'e0faab259d5c474c6af51c9b471f645810a308a1c83976275b1969e81e7d8390'
             '8628100f41800e0e4238a8f949fb99e8667e0b58208130606881612d7172835e'
             '1cf6cd948e5178cbcb16bb0d3c97829311250dedda1e9dcbe0c85dd6bdc4461e')
+_openmp_flag=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || echo "-fopenmp")
```
mmuetzel, line 35: It looks like this flag is not used in this build rule.
Did you forget to use it?

mingw-w64-suitesparse/PKGBUILD
mmuetzel: I agree. The original condition looks correct to me.

mingw-w64-suitesparse/PKGBUILD
hmartinez82: @mmuetzel No. See this line: https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/master/SuiteSparse_config/SuiteSparse_config.mk#L120

If I don't explicitly tell it to not use openmp, then it will default to enable. This needs to stay. The  `depends` can revert to original. I'll fix it and attempt a build locally.

mmuetzel: I didn't mean to imply that this condition is wrong. But you'll probably need to *use* this flag somewhere - probably in the `make` command...

hmartinez82: That's what was throwing me off. Because of that line in [SuiteSparse_config.mk#L120](https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/master/SuiteSparse_config/SuiteSparse_config.mk#L120) I thought that even the gcc builds were using openmp, even when the package recipe was telling it not to.

hmartinez82:
Finally:

GCC: `gcc -DLONGBLAS='long long' -D__USE_MINGW_ANSI_STDIO=1  -O3 -fexceptions -fPIC  -I../Include -I../../SuiteSparse_config ...` (Not using OpenMP)
Clang(x86/x64): `clang -DLONGBLAS='long long' -D__USE_MINGW_ANSI_STDIO=1  -O3 -fexceptions -fPIC -fopenmp -I../../SuiteSparse_config ...` (Using OpenMP)
ClangARM64: `clang -DLONGBLAS='long long' -D__USE_MINGW_ANSI_STDIO=1  -O3 -fexceptions -fPIC  -I../../SuiteSparse_config ...` (Not using OpenMP)

mingw-w64-suitesparse/PKGBUILD
hmartinez82: @mmuetzel I've updated the PR. It's used in the configure script.

mmuetzel:
That's still wrong. GCC should use OpenMP. You just don't need to install it separately.

mingw-w64-suitesparse/PKGBUILD
```diff
@@ -32,6 +32,7 @@ sha256sums=('5fb0064a3398111976f30c5908a8c0b40df44c6dd8f0cc4bfa7b9e45d8c647de'
             'e0faab259d5c474c6af51c9b471f645810a308a1c83976275b1969e81e7d8390'
             '8628100f41800e0e4238a8f949fb99e8667e0b58208130606881612d7172835e'
             '1cf6cd948e5178cbcb16bb0d3c97829311250dedda1e9dcbe0c85dd6bdc4461e')
+_openmp_flag=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* || ${MINGW_PACKAGE_PREFIX} != *-clang-*86* ]] || echo "-fopenmp")
```
mmuetzel, line 35: The condition that you had here before looked correct to me.

hmartinez82: > That's still wrong. GCC should use OpenMP. You just don't need to install it separately.

Omg. I didn't know. **This explains** why I've been so lost!

hmartinez82: Got it working now on GCC too 
`gcc -DLONGBLAS='long long' -D__USE_MINGW_ANSI_STDIO=1  -O3 -fexceptions -fPIC -fopenmp -I../Include -I../../SuiteSparse_config`

mmuetzel:
Maybe, add a comment where you define `_openmp_flag` to explain that it can be removed when/if OpenMP is available for CLANGARM64.

lazka:
If anyone has a name for a openmp meta package that depends on openmp for clang and gcc(-libs) for gcc -> let me know :)

kmilos:
> If anyone has a name for a openmp meta package that depends on openmp for clang and gcc(-libs) for gcc -> let me know :)

The canonical 'libgomp' for GCC and 'libomp' for Clang?

lazka:
> Scratch that, we do need only the one that'll pull in both... Switch the Clang one to `libomp` (to match `libc++`, `libunwind` scheme) and re-purpose `openmp` as meta?

Sounds good

PR #12285, opened 2022-07-28
lerc: update to 4.0.0

kmilos:


kmilos:
Argh, libjxl is not yet available for clangarm64, will have to patch this up...

Biswa96:
It's OK, many packages disable some dependency for clangarm64. e.g. ffmpeg, spice-gtk etc.

kmilos:
I know it's ok, it's just extra work I wasn't prepared for and can't do this instant - it's not just having the conditional in the deps list, have to patch the imagecodecs source as well as it doesn't to automatic detection...

Biswa96:
It may be better to update imagecodecs in a separate pull request.

kmilos:
This version depends on the Lerc version being bumped, so kinda have to merge them together...

kmilos:
Let's just wait for the next libjxl release, we haven't had that feature enabled yet anyway...

PR #12289, opened 2022-07-29
qemu: Update to 7.1.0

heljkon:


heljkon:
Update to qemu 7.1.0 with only small adjustments

@lazka @jeremyd2019  @1480c1 @Biswa96
If you would like to have a look...

mingw-w64-qemu/PKGBUILD
```diff
@@ -117,6 +117,10 @@ prepare() {
   [[ -d "${srcdir}"/${_tarname} ]] && rm -rf "${srcdir}"/${_tarname}
   # tar cannot create links here, therefore hide the failure
   tar -xf "${srcdir}"/${_tarname}.tar.xz -C "${srcdir}" || true
+  # Adjust VERSIONINFO resource (see https://docs.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource)
+  local _vrc="${srcdir}"/${_tarname}/version.rc
+  # Add missing required ProductVersion
+  sed -i "s%\(      VALUE \"FileVersion\", QEMU_VERSION\)%\1\n      VALUE \"ProductVersion\", QEMU_VERSION%" "$_vrc"
```
Biswa96, line 123: Using a patch file might be better to understand and maintain. Also that patch should be sent to upstream.

heljkon: OK, how about removing these lines for now? I'll have a look later.

Biswa96: As you wish.

PR #12291, opened 2022-07-29
nspr: Add CLANGARM64

hmartinez82:


lazka:
thanks!

hmartinez82:
@lazka I saw that you tried to build NSS already. That's my next attempt. Right now, nss is failing with a wrong define in ìntrin.h` it's pointing to x86_64 intrin instead of arm64 :-/  I'll investigate

Biswa96:
@hmartinez82 shouldn't `--enable-64bit` option be also enabled for aarch64? Source https://archlinuxarm.org/packages/armv7h/nspr/files/PKGBUILD

hmartinez82:
@Biswa96 I thought so too originally, but it's automatically done internally for aarch64 at https://hg.mozilla.org/projects/nspr/file/tip/configure#l7065

PR #12299, opened 2022-07-29
libb64: new package

fauxpark:


Biswa96:
The package only has a static library. Is that expected?

fauxpark:
I've switched out the static lib for `bin/libb64-0.dll`. Not sure if the name is correct or if I need a `.dll.a` in lib as well.

Biswa96:
I am not sure if shared library is required, just asking. If yes then an import lib is required for linking.

fauxpark:
I'm not 100% sure how to build the import lib. Do you know if this is correct?
```
gendef libb64.dll
dlltool -d libb64.def -D libb64.dll -l libb64.dll.a
```

fauxpark:
Okay, I think this is good to go now, the two example programs compile and run as expected (took me a while to remember `-lb64` goes at the end...).

Biswa96:
Thanks. I have compiled transmission with this package and compilation went well.

PR #12314, opened 2022-07-30
python-tornado: update to 6.2.0

MehdiChinoune:
I disabled CLANG32 in `python-pywinpty` and `python-terminado` because of the unavailability of `rust`.

### DIFF
--- a/mingw-w64-python-pywinpty/PKGBUILD
+++ b/mingw-w64-python-pywinpty/PKGBUILD
@@ -1,0 +1,50 @@
+# Contributor: Mehdi Chinoune <mehdi.chinoune@hotmail.com>
+
+_realname=pywinpty
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=2.0.6
+pkgrel=1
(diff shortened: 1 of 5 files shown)


mingw-w64-python-pywinpty/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+  cp -r "${_realname}-${pkgver}" "python-build-${MSYSTEM}" && cd "python-build-${MSYSTEM}"
+
+  ${MINGW_PREFIX}/bin/python -m build --wheel --skip-dependency-check --no-isolation
+}
+
+check() {
+  cd "${srcdir}/python-build-${MSYSTEM}"
+
+# The test command will usually depend upon what is contained in the tox.ini file
+# or in the [testenv:py] section of the pyproject.toml file.
```
lazka, line 38: what's this about btw? is this referring to tox?

MehdiChinoune: from template

lazka: ah I see, that is targeted at the template user, but confusing. I'll remove it then

PR #12330, opened 2022-07-31
libnatpmp: new package

fauxpark:
I hope this is not too controversial...

I could not for the life of me get the makefile to play ball, so I ended up doing things by hand here. If this library were using autotools or CMake or something, perhaps it may have gone better, but I don't know.

Running `natpmpc` produces what appears to be normal output, which gives me hope for the Transmission package - I've been trying to bring it closer to Arch with its split lib/cli/gtk/qt packages.


Biswa96:
Does the source come from this repository https://github.com/miniupnp/libnatpmp ? If yes, we can upstream the required changes and build the package from git repository.

fauxpark:
Promising... it uses CMake but doesn't build the shared library or do any installation...

PR #12376, opened 2022-07-31
poppler: Update to 22.07 and enable test

hmartinez82:


Biswa96:
The poppler dll name is changed after the update. Dependent packages need to be rebuilt.

```diff
  -/mingw64/bin/libpoppler-117.dll
  +/mingw64/bin/libpoppler-122.dll
```

hmartinez82:
@Biswa96 How should I proceed. Do I need to wait for this to be merged then update the deps or should I update the deps and push the changes in this PR ?

hmartinez82:
All I need is to bump the `pkgrel`, correct?

Biswa96:
All the "required by" packages [here](https://packages.msys2.org/package/mingw-w64-x86_64-poppler?repo=mingw64) need to be rebuilt. 

> All I need to do is to bump the pkgrel

That is the ideal situation. But sometimes poppler update breaks ABI/API in various packages. Those need to be checked.

lazka:
```
mingw-w64-poppler-qt6
mingw-w64-pdf2djvu
mingw-w64-emacs-pdf-tools-server
mingw-w64-gdcm
mingw-w64-inkscape
mingw-w64-gdal
```

to be exact

hmartinez82:
> All the "required by" packages [here](https://packages.msys2.org/package/mingw-w64-x86_64-poppler?repo=mingw64) need to be rebuilt.
> 
> > All I need to do is to bump the pkgrel
> 
> That is the ideal situation. But sometimes poppler update breaks ABI/API in various packages. Those need to be checked.

Ugh you are right, except for the qt6 one, three other that I tried already are not cleanly compiling. I got pdf2djvu patched, working on the emacs server now

lazka:
Note, you only need to rebuild them for environments that are already in the repo, so no arm64. Just let CI handle it.

hmartinez82:
@lazka 
I've updated the lists packages above, except Inkscape. It will take more work
For Inkscape. I already looked at their source code. Version 1.2.1 is already poppler 22.07 compatible. So
- Should I apply a patch to the current Inkscape version, or
- Deal with the pain of updating Inkspace to 1.2.1?

On another subject, how do we handle bsdtar failing to extract symlinks? Is there a way to tell it to extract the linked files as copies instead of trying to creating a symlink?

Biswa96:
I have tried to build inkscape v1.2.1 but it fails to build in CI. inkscape now has a patched version of cairo which fails in CI. The src/3rdparty/CMakeLists.txt file is full of hacks. I wonder if we can apply the patch to mingw cairo package instead. Then build inkscape without that external cairo project.

hmartinez82:
@Biswa96 yes, Inkspace is the only one that I still need to patch to work with the new Poppler

Biswa96:
Opps, missed gtk-doc, thanks to printf debugging 🤩

hmartinez82:
Finally. and PDF import working \o/ 
image

hmartinez82:
Finally \o/
And using the Poppler/Cairo import
image

hmartinez82:
@lazka @Biswa96 Among these repos:
```
mingw-w64-poppler-qt6
mingw-w64-pdf2djvu
mingw-w64-emacs-pdf-tools-server
mingw-w64-gdcm
mingw-w64-inkscape
mingw-w64-gdal
mingw-w64-gtk-doc
```
Some of them are building without any changes. Do I still need to just bump their `pkgrel` to force a new build?

Biswa96:
Yes, because the dll name is changed and probably api/abi with it.

Biswa96:
Inkscape has been update to v1.2.1 which does not require the poppler patch anymore. Would you like to rebase the commits here? Also, I think this can be wrapped up in two commits - poppler update and rebuilds.

hmartinez82:
@Biswa96 will do later today👍🏼

hmartinez82:
New PR on https://github.com/msys2/MINGW-packages/pull/12467

lazka:
(note that you can just force-push to existing PRs)

hmartinez82:
Not at all 😊

hmartinez82:
> @hmartinez82 I reused your screenshot here: https://twitter.com/msys2org/status/1556716408305852417 I hope that's OK.

I just hope they don't think it's a screenshot of Inkscape running on Windows on Arm. We're not there yet 😅

lazka:
oh, I assumed it was :/

I've delete that tweet then

PR #12383, opened 2022-08-01
aws-c-io: update to 0.12.0

MehdiChinoune:
Fails to build on CLANG64/32:
```
  C:/_/mingw-w64-aws-c-io/src/aws-c-io-0.12.0/source/windows/iocp/socket.c:1222:5: error: expected expression
      int win_error = GetLastError(); /* logging may reset error, so cache it */
      ^
```
https://github.com/awslabs/aws-c-io/blob/v0.12.0/source/windows/iocp/socket.c#L1220

It seems like `GetLastError` was defined somewhere to an empty string or something.


Biswa96:
Please try with this patch https://github.com/awslabs/aws-c-io/pull/503

MehdiChinoune:
> Please try with this patch [awslabs/aws-c-io#503](https://github.com/awslabs/aws-c-io/pull/503)

Maybe we should also report a bug to LLVM/Clang about not accepting that code.

Biswa96:
I think this is a C/C++ rule and clang/llvm is right. According to https://en.cppreference.com/w/cpp/language/goto

> If transfer of control enters the scope of any automatic variables (e.g. by jumping forward over a declaration statement), the program is ill-formed (cannot be compiled), unless all variables whose scope is entered have

Also https://stackoverflow.com/questions/8384388/variable-declaration-after-goto-label

mstorsjo:
See https://github.com/llvm/llvm-project/issues/55821 and https://github.com/llvm/llvm-project/issues/56040

MehdiChinoune:
It succeeded on UCRT64 on my side!

Biswa96:
> It succeeded on UCRT64 on my side!

Yes, ucrt64 fails with this package randomly. The issue can not be reproduced every time.

PR #12384, opened 2022-08-01
python-tifffile: update to 2022.7.31

kmilos:


kmilos:
`scikit-image 0.18.1 requires scipy, which is not installed.`

This doesn't seem to be true, `scipy` is installed previously.

lazka:
hm, could be a scipy issue, I'll have a look

PR #12385, opened 2022-08-01
Added packages for Conan and dependencies

jhol:
[Conan](https://conan.io/) is a popular package manager for C/C++ applications written in Python. This PR adds support for Conan and its dependencies.


lazka:
Should we package an older version of node-semver, or patch out the requirement?

PR #12386, opened 2022-08-01
scipy: install egg-info

lazka:
The meson build seems to be missing it, so work around it letting
setuptools handle it.

### DIFF
--- a/mingw-w64-python-scipy/PKGBUILD
+++ b/mingw-w64-python-scipy/PKGBUILD
@@ -8,7 +8,7 @@ provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
 pkgver=1.9.0
-pkgrel=2
+pkgrel=3
 pkgdesc="SciPy is open-source software for mathematics, science, and engineering (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


eli-schwartz:
The `PKG-INFO` file from the published sdist is created by the `meson-python` build backend with the exact same content as the `*.dist-info/METADATA` file would be. The other file provided as wheel metadata is `WHEEL` but I don't think that's actually necessary for other projects to depend on it (and it basically matches the `PKG-INFO` file provided by egg-info, which doesn't have `WHEEL` anyway...)

lazka:
I see, thanks. I'll try the meson-python build first, and fall back to that if it doesn't work out

eli-schwartz:
One possible issue is that the meson-python build backend doesn't currently support passing options through to `meson setup`.

lazka:
Ok, that's not ideal then. I wonder if they shouldn't provide an option to write install package metadata in meson, so the meson install is self contained.

eli-schwartz:
This is my plan!

The issue is that doing so requires reading pyproject.toml to get the metadata, which means we need a toml module, and Meson doesn't do dependencies outside the stdlib.

Meanwhile, the Python ecosystem has declared toml to be the standard, and arguments about how terrible a format it is aside, Python has toml as core infrastructure for years and years now, but only got around to adding a stdlib module to read it in python 3.11. Which isn't even released yet.

lazka:
Ah, right, I remember the discussion on IRC.

lazka:
And as I said on IRC; just vendor it ;) https://github.com/pypa/flit/tree/main/flit_core/flit_core/vendor

(I know, the rules...)

PR #12396, opened 2022-08-01
unbound: build static library

selsta:
Is there any reason this is disabled? I've looked into the `git history` and it was disabled since the day this package got added.


Biswa96:
> Is there any reason this is disabled?

Most of the time, packages are built with shared library only by default. Static libraries are added when requested. Also some issues appear when static libraries are added. For example, mishandling of dllimport attribute, proper cflags.private in pkgconfig file, library name etc.

I did not find any issue with this static library. Did you test the static library to compile any other project? Would you also bump the pkgrel value?

selsta:
> Did you test the static library to compile any other project?

I don't have any Windows setup myself, I just saw an issue on our issue tracker and this seemed the easiest way to solve this instead of explaining our Windows users how to manually compile static unbound.

> Would you also bump the pkgrel value?

Done.

PR #12398, opened 2022-08-01
spyder: More dependencies

podsvirov:
Because I can't launch spyder without this dependencies


lazka:
It's questionable if we should continue supporting it: https://github.com/python-lsp/python-lsp-server/pull/232#issuecomment-1171949774

Unless we find a way to package python things will all deps vendored somehow.

podsvirov:
I can't launch spyder without this dependencies and #12427

PR #12412, opened 2022-08-02
Added unifdef

jhol:
The `unifdef` utility selectively processes conditional C preprocessor `#if` and `#ifdef` directives. It removes from a file both the directives and the additional text that they delimit, while otherwise leaving the file alone.


Biswa96:
Required `makedepends=("${MINGW_PACKAGE_PREFIX}-cc")`

jhol:
> Required `makedepends=("${MINGW_PACKAGE_PREFIX}-cc")`

Fixed now!

PR #12416, opened 2022-08-03
ogre3d: update to 13.4.3

MehdiChinoune:


mmuetzel:
Maybe, it would be possible to avoid the out-of-memory issue on MINGW32 by skipping to build the Python module for that environment. Similar to #11849.

MehdiChinoune:
> Maybe, it would be possible to avoid the out-of-memory issue on MINGW32 by skipping to build the Python module for that environment. Similar to #11849.

It succeeds on my side.
The ultimate solution is to drop 32biy entirely. Unfortunately instead of dropping MINGW32 we got CLANG32. What a joke!

mingw-w64-ogre3d/014-install-pkhconfig-and-fix-plugin-extension.patch
```diff
@@ -0,0 +1,17 @@
+--- a/CMake/ConfigureBuild.cmake
```
mmuetzel, line 1: Nit: Typo in file name. `pkhconfig` -> `pkgconfig`

MehdiChinoune:
As always, Not a single word if it is accepted, rejected or need some modifications.

PR #12426, opened 2022-08-03
binutils: backport fix for "local symbol `0' has no section" errors

lazka:
This usually got triggered when linking meson subprojects.

https://sourceware.org/bugzilla/show_bug.cgi?id=29389

@lb90 it's not clear to me if this depends on other patches? git was happy to merge without conflicts at least.


lb90:
Hi @lazka, seems to me that it's good to go 👍

Thanks!

PR #12452, opened 2022-08-05
codelite: rebuild against new wxw packages

lazka:


stahta01:
I saw this package failure and I am working to fix it by adding an wxWidgets 3.0 STL build; the error was likely triggered by the GCC compiler version upgrade. But, the error can likely be fixed by patching CodeLite to use an STL version of wxWidgets 3.0 like it does with 3.1. Likely be a lot of codelite patches needed.

lazka:
There is an upstream fix: https://github.com/eranif/codelite/commit/1d57763c8a19b3 but it doesn't apply cleanly, so I gave up for now.

stahta01:
I will try to manually apply the fix and see if it works by itself.

stahta01:
See PR #12459

PR #12460, opened 2022-08-06
inkscape: update to 1.2.1

Biswa96:
* Remove some upstream patches.
* Add build dependencies for bundled cairo project.
* Extract source tarball manually due to presence of symlinks.


lazka:
Does it run for you? It doesn't here, without any errors.

lazka:
It's missing symbols in cairo.

Biswa96:
I have used the artifacts from this CI and Inkscape is working without any issue. Do I have to test some specific options?

lazka:
No, just `inkscape.exe`.

Biswa96:
Yup, it's working. I have tested ucrt64, btw.

lazka:
hm, but how I wonder. CI also shows it's linking against system cairo

Biswa96:
The bundled cairo was added in this commit https://gitlab.com/inkscape/inkscape/-/commit/2e3e3d6ab1e0ea7417dced92c0b377a5c76bb19a. It is disabled for unix systems and can be disabled using `WITH_INTERNAL_CAIRO` option explicitly. Should it be disabled to be safe?

lazka:
yes please!

PR #12463, opened 2022-08-06
bin2header: update to 0.3.0

AntumDeluge:
Release info: https://github.com/AntumDeluge/bin2header/releases/tag/v0.3.0


Biswa96:
I think quote changes are not required. Those should be interpreted as strings and do not require further shell expansion[^1].

[^1]: https://www.gnu.org/software/bash/manual/bash.html#Single-Quotes

AntumDeluge:
> I think quote changes are not required. Those should be interpreted as strings and do not require further shell expansion.

Okay, makes sense. I only changed them so they would all match. But I can see why it is useful to have mixed in a shell script. I will keep that in mind for future updates.

PR #12465, opened 2022-08-06
inkscape: Disable internal cairo dependency

Biswa96:
Inkscape links with system cairo package by default. But the bundled
cairo adds APIs which are not present in system one. This may cause
unexpected crashes.


lazka:
works again now, thanks!

PR #12483, opened 2022-08-08
openblas: update to 0.3.21

MehdiChinoune:


mmuetzel:
I installed the MINGW64 build artifact from this PR and tested a few calculations in Octave. The updated version seems to give the same results as the prior version.
LGTM.

PR #12506, opened 2022-08-09
libffi: Update to 3.4.2

lazka:
Rebuilds:

```
mingw-w64-glib2
mingw-w64-python-cffi
mingw-w64-lua-lgi
mingw-w64-pygobject2
mingw-w64-pygobject
mingw-w64-clang
mingw-w64-gobject-introspection
mingw-w64-p11-kit
mingw-w64-python2
mingw-w64-python
```


lazka:
libLLVM-14.dll depends on libffi-7.dll, so we likely need to ship both so we can transition

PR #12521, opened 2022-08-10
crt: Add DirectX libs to aarch64

### LINKED ISSUE #12516: DirectX libraries not present in CLANGARM64's crt
The Direct3D libs are present, but not DirectX. (That's what I inferred from the names).

E.g.: `libd3dx10.a` is missing from  `/clangarm64/lib`

hmartinez82:
Fixes #12516 , but this should be fixed upstream. There's no code compilation involved. Its just adding new .def files to create .a files from. 
cc @Biswa96 How can I get in touch with the upstream maintainers?


hmartinez82:
Fixed #12516 , but this should be fixed upstream. There's no code compilation involved. Its just adding new .def files to create .a files from. 
cc @Biswa96 How can I get in touch with the upstream maintainers?

MehdiChinoune:
@mstorsjo

driver1998:
Are they actually exist? I was under the impression of D3DX being unsupported on ARM. At least D3DX9 does not exist.

hmartinez82:
@driver1998 It does, but they don't match the filenames from x64, that's why I removed it and only kept dx11. I'll revisit this later with the proper file names for dx9 . I have to learn if it's possible to create a `A.a` file that points to `B.dll` instead of `A.dll`

hmartinez82:
Closing since this was only needed to build Ogre3D with DirectX 9 support. But Ogre3D, on ARM64, is configured to only build the DirectX 11 renderer which does not require the Legacy DirectX SDK files.

PR #12520, opened 2022-08-10
Ogre3D deps: Enable aarch64 

hmartinez82:
Enable aarch64 builds for tinyxml, zziplib, glsl-optimizer and hlsl2glsl

PR #12522, opened 2022-08-10
ogre3d: Add aarch64 support

hmartinez82:
Requires #12520 

image

### DIFF
--- a/mingw-w64-ogre3d/PKGBUILD
+++ b/mingw-w64-ogre3d/PKGBUILD
@@ -102,6 +102,7 @@ build() {
     -DOGRE_BUILD_SAMPLES=ON \
     -DOGRE_INSTALL_SAMPLES=ON \
     -DOGRE_CONFIG_FILESYSTEM_UNICODE=ON \
+    $([[ ${MSYSTEM_CARCH} == aarch64 ]] && echo "-DOGRE_BUILD_RENDERSYSTEM_D3D9=OFF") \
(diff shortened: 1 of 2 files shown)


hmartinez82:
This can go ahead since it doesn't depend on the legacy directx files anymore.

mingw-w64-ogre3d/PKGBUILD
```diff
@@ -99,6 +102,7 @@ build() {
     -DOGRE_BUILD_SAMPLES=ON \
     -DOGRE_INSTALL_SAMPLES=ON \
     -DOGRE_CONFIG_FILESYSTEM_UNICODE=ON \
+    $([[ ${MSYSTEM_CARCH} == aarch64 ]] && echo "-DOGRE_BUILD_RENDERSYSTEM_D3D9=OFF") \
```
Biswa96, line 105: May I ask why this was disabled?

hmartinez82: Windows on Arm doesn't support DirectX 9 natively. It has a redirection to dx 12 using a completely dll filename. Also mingw-w64 doesn't provide that import libraries for these different files,  the  d3dx*.dll ones.

hmartinez82: The DirectX 11 renderer works perfectly :)

Biswa96: If those **native arm64** version of d3dx*.dll files exist or provided by any MS package then those import libraries can be added in mingw-w64.

PR #12528, opened 2022-08-10
libffi: drop libffi-7.dll

lazka:
This was only here to rebuild clang with itself, so now we can drop it.


mmuetzel:
The package grokker lists `mingw-w64-p11-kit`. Does it need to be rebuilt? Or is that no longer necessary with the recent version `8.4.3-3`?

lazka:
I just pushed 0.24.1-3 to the repo

PR #12530, opened 2022-08-11
inkscape: Add aarch64 support 🏁

hmartinez82:
Requires #12529  

I had to manually create `libmscms.a` to be able to finish compiling so work is required in the `crt-git` package. This time, unlike the DirectX files, `mscms.dll` does exist in Windows on ARM. 

@lazka You may have to undelete that tweet soon 😊

image

### DIFF
--- a/mingw-w64-inkscape/PKGBUILD
+++ b/mingw-w64-inkscape/PKGBUILD
@@ -6,11 +6,11 @@ _realname=inkscape
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.2.1
-pkgrel=3
+pkgrel=4
 _pkg_suffix=2022-07-14_9c6d41e410
 pkgdesc="Vector graphics editor using the SVG file format (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://inkscape.sourceforge.io/"
 license=("GPL" "LGPL")
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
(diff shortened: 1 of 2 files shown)


mingw-w64-inkscape/PKGBUILD
```diff
@@ -100,8 +110,14 @@ build() {
     -DCMAKE_AR=${MINGW_PREFIX}/bin/ar \
     -DWITH_IMAGE_MAGICK=OFF \
     -DWITH_INTERNAL_CAIRO=OFF \
+    $([[ ${MSYSTEM_CARCH} == aarch64 ]] && echo "-DWITH_OPENMP=OFF") \
     ../${_realname}-${pkgver}
 
+  # Generate missing static library for linking mscms.dll in aarch64
+  if [[ ${MSYSTEM_CARCH} == aarch64 ]]; then
+    dlltool -m arm64 -k --as=as --output-lib ./lib/libmscms.a --input-def ../${_realname}-${pkgver}/mscms.def
```
mmuetzel, line 118: Is this an import library (rather than a "proper" static library)? In that case, the more "natural" name for it might be `libmscms.dll.a`.

hmartinez82: @mmuetzel It is. I was using the name that the crt-git package is using. This is temporary just for inkscape. Once they provide this import library in mingw-64 I'll remove this patch

mmuetzel: Thanks for pointing that out. Afaict, that is also "just" an import library. So, this names seems good for "symmetry".

lazka:
@Biswa96 do you happen to know how we could upstream that `libmscms.a` change to mingw-w64? Otherwise I can try :)

hmartinez82:
I've created this there. I'll give it a try once I'm back home next week. https://sourceforge.net/p/mingw-w64/feature-requests/105/

Biswa96:
Patch sent upstream https://sourceforge.net/p/mingw-w64/mailman/message/37693325/

hmartinez82:
@Biswa96 How long does it usually take for changes in the upstream to get merged then make into MSYS2? 
Also. Is it okay to use the same import lib for all archs? For instance, the libarm32 one has way less exported methods than the libarm64 import library

Biswa96:
Please wait for a day or two. The maintainers are also busy with their work.

lazka:
(Once it's in mingw-w64 we'll sync it right away)

Biswa96:
The patch has been added upstream in this commit https://sourceforge.net/p/mingw-w64/mingw-w64/ci/54fd3075b9c534ba4e3832899ec039fbca64162e/

lazka:
> (Once it's in mingw-w64 we'll sync it right away)

-> #12583

hmartinez82:
Good news. No need to libmscms patch anymore after https://github.com/msys2/MINGW-packages/pull/12583

hmartinez82:
@lazka @Biswa96  This is ready for merging?

lazka:
> @lazka You may have to undelete that tweet soon 😊

@hmartinez82 I add some new ones now :)

PR #12536, opened 2022-08-12
Backport patches to support symbol visibility

mati865:
In preparation for enabling all targets in LLVM 15 this must be merged first.

Tested with multi stage bootstrap: built CLANG64 version of this PGBUILD, installed resulting packages, built this PKGBUILD again (so this step built LLVM with patched LLVM), installed resulting packages.
Then I rebuilt libc++, mlir and flang, they do not crash so seems good to go.


mati865:
Oh and LLVM 15 bootstrapped with itself with all targets enabled is at ~35,3k exports (~65,5k is the limit).

PR #12537, opened 2022-08-12
sundials: upgrade to 6.3.0

gharveymn:
Builds just fine.

@mmuetzel Tested building Octave 8.0.0 from around January. I don't have the time to work out the kinks to test on a newer version, but that older version had no issues.

### DIFF
--- a/mingw-w64-sundials/PKGBUILD
+++ b/mingw-w64-sundials/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=sundials
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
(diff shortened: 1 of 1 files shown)


mmuetzel:
Not directly related. But maybe, this is an opportune moment to change that:
https://computation.llnl.gov/projects/sundials (no longer?) seems to work. This link is working for me:
https://computing.llnl.gov/projects/sundials

mmuetzel:
I downloaded and installed the build artifact for MINGW64 from this PR. With it, the tests for the functions that use features from this library still pass in Octave:
```
octave:2> test ode15i
PASSES 48 out of 48 tests
octave:3> test ode15s
PASSES 40 out of 40 tests
```
LGTM.

PR #12546, opened 2022-08-12
clang: rebuild

lazka:
see #12536


mati865:
It won't change anything as previous PR only prepared ground for the future.
https://github.com/llvm/llvm-project/commit/2c2fb0c7375061147711cd21396d79faad7dfdfb is the commit that will reduce symbols count when building with Clang+LLD or Clang+patched Binutils.

lazka:
Ah, I see, thanks!

PR #12550, opened 2022-08-13
llvm: Build all targets

MehdiChinoune:


lazka:
@mati865 should we try this?

mati865:
If you think there is benefit from it go ahead.

@MehdiChinoune have you tested if mlir still builds against LLVM with symbols visibility patch? If not I can do it later (it takes about 20 mins to finish makepkg on Clang PKGBUILD for me when using existing sources).

MehdiChinoune:
> @MehdiChinoune have you tested if mlir still builds against LLVM with symbols visibility patch?

No, I haven't.

mati865 (approved):
Rebuild all LLVM packages and they still do work.

MehdiChinoune:
I have enabled all targets on all environments.

lazka:
From what I understood, clang needs to be built with clang instead of gcc under mingw32/mingw64/ucrt64 to exclude those symbols. But let's see :)

mati865:
> From what I understood, clang needs to be built with clang instead of gcc

Yeah, it's totally true.

lazka:
mingw64:

```console
$ ntldd --list-exports /mingw64/bin/libLLVM-14.dll | wc -l
64128
$ ntldd --list-exports ./libLLVM-14.dll | wc -l
33978
```

mingw-w64-clang/PKGBUILD
```diff
@@ -375,7 +365,8 @@ package_llvm() {
   depends=("${MINGW_PACKAGE_PREFIX}-libffi"
            "${MINGW_PACKAGE_PREFIX}-gcc-libs"
            "${MINGW_PACKAGE_PREFIX}-zlib"
-           "${MINGW_PACKAGE_PREFIX}-libxml2")
+           "${MINGW_PACKAGE_PREFIX}-libxml2"
+           $([[ "$_compiler" == "clang" ]] && echo "${MINGW_PACKAGE_PREFIX}-libc++"))
```
mati865, line 369: I believe we should still use libstdc++ and libgcc in GNU subsystems.

MehdiChinoune: It links to `libgcc_s` but not `libstdc++`.
It wasn't my choice, for some reason clang chose `libc++` over `libstdc++`.
I have added `libc++` as a dependency of `llvm` because `runtime-dependencies` showed missing `libc++` on all libraries and executables.

mati865: It's this line https://github.com/msys2/MINGW-packages/blob/60579931f962cc840c3c7a62ef82d52e846725a6/mingw-w64-clang/PKGBUILD#L206

MehdiChinoune: It was you who added those options in
https://github.com/msys2/MINGW-packages/commit/014716a5f7c076fffa1ed20119470948cd1cac20

mingw-w64-clang/PKGBUILD
```diff
@@ -230,18 +232,6 @@ build() {
       ;;
   esac
 
-  # List of targets to build in CMake style
-  # For more info see: https://github.com/msys2/MINGW-packages/discussions/9765
-  local _targets="AArch64;AMDGPU;WebAssembly;X86"
```
lazka, line 235: If we'd keep the limited list with` if [ "${_compiler}" == "gcc" ]; then` then we could bootstrap with gcc, if the need arises.

MehdiChinoune: done

lazka:
lgtm, thanks, let's get that in once the icu transition is done

MehdiChinoune:
icu transiotion is done.

oscarfv:
Renaming the patches seems unnecessary to me. Are we going to mass rename the patches every time one is dropped?

oscarfv:
> mingw64:
> 
> ```
> $ ntldd --list-exports /mingw64/bin/libLLVM-14.dll | wc -l
> 64128
> $ ntldd --list-exports ./libLLVM-14.dll | wc -l
> 33978
> ```
> 
> edit: I've rebuilt mingw64 clang locally with this, and all looks good.

This is going to cause problems very soon.

mati865:
> > mingw64:
> > ```
> > $ ntldd --list-exports /mingw64/bin/libLLVM-14.dll | wc -l
> > 64128
> > $ ntldd --list-exports ./libLLVM-14.dll | wc -l
> > 33978
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > edit: I've rebuilt mingw64 clang locally with this, and all looks good.
> 
> This is going to cause problems very soon.

What do you mean?

lazka:
> Renaming the patches seems unnecessary to me. Are we going to mass rename the patches every time one is dropped?

yeah, should probably be avoided, as it makes tracking patches via git harder than it should be. though github now supports "--follow" in the web ui.

lazka:
ucrt64 failed with 

```
2022-08-15T22:43:35.5507444Z FAILED: tools/clang/lib/Tooling/ASTNodeAPI.json C:/M/mingw-w64-clang/src/build-UCRT64/tools/clang/lib/Tooling/ASTNodeAPI.json 
2022-08-15T22:43:35.5646537Z cmd.exe /C "cd /D C:\M\mingw-w64-clang\src\build-UCRT64\tools\clang\lib\Tooling && C:\M\mingw-w64-clang\src\build-UCRT64\bin\clang-ast-dump.exe --skip-processing=0 -I C:/M/mingw-w64-clang/src/build-UCRT64/lib/clang/14.0.6/include -I C:/M/mingw-w64-clang/src/clang/include -I C:/M/mingw-w64-clang/src/build-UCRT64/tools/clang/include -I C:/M/mingw-w64-clang/src/build-UCRT64/include -I C:/M/mingw-w64-clang/src/llvm/include -I D:/a/msys64/ucrt64/include/c++/12.1.0 -I D:/a/msys64/ucrt64/include/c++/12.1.0/x86_64-w64-mingw32 -I D:/a/msys64/ucrt64/include/c++/12.1.0/backward -I D:/a/msys64/ucrt64/lib/clang/14.0.6/include -I D:/a/msys64/ucrt64/include --json-output-path C:/M/mingw-w64-clang/src/build-UCRT64/tools/clang/lib/Tooling/ASTNodeAPI.json"
2022-08-15T22:43:35.5899723Z Mingw-w64 runtime failure:
2022-08-15T22:43:37.6445002Z 32 bit pseudo relocation at 00007FFFFA738CB6 out of range, targeting 00007FF821660A60, yielding the value FFFFFFF826F27DA6.
```

I'm going to re-try..

oscarfv:
> > > mingw64:
> > > ```
> > > $ ntldd --list-exports /mingw64/bin/libLLVM-14.dll | wc -l
> > > 64128
> > > $ ntldd --list-exports ./libLLVM-14.dll | wc -l
> > > 33978
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > edit: I've rebuilt mingw64 clang locally with this, and all looks good.
> > 
> > 
> > This is going to cause problems very soon.
> 
> What do you mean?

That's very close to the maximum 2^16 limit of exported symbols for Windows dlls.

oscarfv:
> > > mingw64:
> > > ```
> > > $ ntldd --list-exports /mingw64/bin/libLLVM-14.dll | wc -l
> > > 64128
> > > $ ntldd --list-exports ./libLLVM-14.dll | wc -l
> > > 33978
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > edit: I've rebuilt mingw64 clang locally with this, and all looks good.
> > 
> > 
> > This is going to cause problems very soon.
> 
> What do you mean?

Wait... those numbers are not what I thought. So that's the effect of the "symbol visibility" patch? A net decrease of 30k+ exported symbols? That's great!

mati865:
Exactly, depending on the subsystem we were ~100-300 symbols from the limit with only select targets enabled.
With symbols visibility patch and all targets enabled we are now 30k symbols away from it. As a bonus it's now possible to build debug + assertions shared libraries which might be really useful for debugging.

lazka:
The following failed in the integration tests: https://github.com/msys2/msys2-tests/blob/main/toolchain/custom/test.sh

```
Mingw-w64 runtime failure:
32 bit pseudo relocation at 00007FF706D1F9ED out of range, targeting 00007FFBB1B10A60, yielding the value 00000004AADF106F.
clang++: error: linker command failed due to signal (use -v to see invocation)
```

I think I'm going to revert the MINGW env clang back to build with gcc and limited targets, and leave that for later.

jeremyd2019:
> The following failed in the integration tests: https://github.com/msys2/msys2-tests/blob/main/toolchain/custom/test.sh
> 
> ```
> Mingw-w64 runtime failure:
> 32 bit pseudo relocation at 00007FF706D1F9ED out of range, targeting 00007FFBB1B10A60, yielding the value 00000004AADF106F.
> clang++: error: linker command failed due to signal (use -v to see invocation)
> Error: Process completed with exit code 1.
> ```

@mstorsjo Any thoughts about this ☝️?  This same sort of error showed up when trying to build UCRT64 clang (one of the .exes that are built and run during the build failed with it).

lazka:
Could be a similar issue to https://github.com/msys2/MINGW-packages/pull/12042 ?

mstorsjo:
> Could be a similar issue to #12042 (binutils strip breaking clang compiled things) ? I guess I could test locally without stripping when I have some time.

Stripping shouldn't be causing these issues.

So you've switched the llvm/clang build, in the ucrt64 environment, from building with gcc, to clang (+ the symbol visibility patches) and re-enabled all targets - and then we hit this issue? Can we isolate which part of these changes is causing the issue? I.e. build with clang instead of gcc, but without new symbol visibility patches and without adding more targets. Not that I think that it really would give that much extra info - I guess it's from building with clang instead of gcc.

Another thing which might be worth trying, is to see if the same thing happens if linking with lld instead of ld.bfd. If that resolves the issue, there's something different in the clang generated code that lld handles differently. If it doesn't, it's probably the clang code generation when targeting libstdc++ which is causing the issue. (I presume that the clang used here is configured correctly wrt `-femulated-tls` for working with libstdc++.)


The main issue that the error is telling you about, is that there's some symbol, which was referenced in the compiler generated code as if it was within the same DLL/EXE (i.e. with a 32 bit relative reference), but which turned out to get autoimported from a different DLL (which was loaded too far away in the virtual memory). Apparently gcc and clang generate some symbol references differently. We need to find out and pinpoint which one, and why.

We pretty much need to resolve all those addresses to what they mean (which can be tricky with address space layout randomization). In the error message, `32 bit pseudo relocation at 00007FF706D1F9ED out of range, targeting 00007FFBB1B10A60, yielding the value 00000004AADF106F`, the first address `00007FF706D1F9ED` is an address in our executable (the exe itself or e.g. libLLVM). This would be a location somewhere in the middle of some function (normally).

Various ways of figuring this out:
- If we'd have debug info, we might be able to addr2line on it and get a pretty good reference to the source location
- We can make a link map of the executable (pass e.g. `-Map out.txt` to the linker) and use that to resolve the address back to which object file it originated from.

The object file which contained this piece of code is the important bit that we need to find; we'd need to see the code (`objdump -d -r`) for the area that contains this problematic symbol reference - plus that we need to know what symbol it references (and where it's loaded from in the link).

(We can also use the other address, `00007FFBB1B10A60`, to find out which symbol in the other DLLs this is about - but we pretty much need to find the referencing object file too.)

mati865:
> Can we isolate which part of these changes is causing the issue? I.e. build with clang instead of gcc, but without new symbol visibility patches and without adding more targets. Not that I think that it really would give that much extra info - I guess it's from building with clang instead of gcc.

I've tested in MINGW64 enviroment with Clang patched to handle symbol visibility but LLVM's CMake was patched to disable visibility annotations for MinGW, X86 as the only target and then linking with LLD.
None of these helped as LLD tests were failing with the same relocation error. I'll try with libc++ later today (and I'm expecting this to help).

> I presume that the clang used here is configured correctly wrt -femulated-tls for working with libstdc++

Yup, we patch Clang from GGC based environments to use emulated TLS by the default.

> If we'd have debug info, we might be able to addr2line on it and get a pretty good reference to the source location

I did release with debug info build but due to lack of required skills or ASLR couldn't get anything meaningful.

mati865:
Yup, it must be symbol coming from libstdc++. I've linked using libc++ with both LLD and ld.bfd and it did not reproduce within LLD testsuite.

So to recap: the problem exists only when building LLVM with Clang+libstdc++, there seem to be no other variables here.

mstorsjo:
I've had a look now, and the issue is how Clang generates accesses to emutls symbols within libstdc++.

A bit of background - how I debugged it. To ease debugging, I added a bunch of extra printouts in the mingw-w64-crt pseudo relocs code. Unfortunately I think they'd bring in too many extra symbol dependencies to be enabled by default though...:
```diff
diff --git a/mingw-w64-crt/crt/pseudo-reloc.c b/mingw-w64-crt/crt/pseudo-reloc.c
index d4589ca66..27858711f 100644
--- a/mingw-w64-crt/crt/pseudo-reloc.c
+++ b/mingw-w64-crt/crt/pseudo-reloc.c
@@ -16,6 +16,7 @@
 */

 #include <windows.h>
+#include <psapi.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
@@ -80,6 +81,21 @@ typedef struct {
   DWORD version;
 } runtime_pseudo_reloc_v2;

+static void dump_maps(void) {
+  HANDLE process = GetCurrentProcess();
+  HMODULE mods[1024];
+  DWORD needed;
+
+  if (!EnumProcessModules(process, mods, sizeof(mods), &needed))
+    return;
+  for (unsigned i = 0; i < (needed / sizeof(HMODULE)); i++) {
+    char name[1000];
+    if (GetModuleFileNameEx(process, mods[i], name, sizeof(name))) {
+      fprintf(stderr, "%p: %s\n", mods[i], name);
+    }
+  }
+}
+
 static void ATTRIBUTE_NORETURN
 __report_error (const char *msg, ...)
 {
@@ -154,6 +170,7 @@ __report_error (const char *msg, ...)
   vfprintf (stderr, msg, argp);
 #endif
   va_end (argp);
+  dump_maps();
   abort ();
 #endif
 }
@@ -454,9 +471,9 @@ do_pseudo_reloc (void * start, void * end, void * base)
           ptrdiff_t max_unsigned = (1LL << bits) - 1;
           ptrdiff_t min_signed = UINTPTR_MAX << (bits - 1);
           if (reldata > max_unsigned || reldata < min_signed)
-           __report_error ("%d bit pseudo relocation at %p out of range, "
+           __report_error ("%d bit pseudo relocation at %p out of range (in module at %p), "
                             "targeting %p, yielding the value %p.\n",
-                            bits, reloc_target, addr_imp, reldata);
+                            bits, reloc_target, &__MINGW_LSYMBOL(_image_base__), addr_imp, reldata);
         }

       /* Write the new relocation value back to *reloc_target */
```

I rebuilt a new `libmingw32.a` with this and placed it in the uct64 sysroot, to be used when compiling new binaries. I then relinked the lld executable which showed the issue.

With that in place, I got the following debug output:
```
32 bit pseudo relocation at 00007FF64EAAE4DD out of range (in module at 00007FF64E7F0000), targeting 00007FFBA2E60A60, yielding the value 00000005543B257F.
00007FF64E7F0000: C:\code\llvm-project\llvm\build-ucrt64-clang\bin\lld.exe
00007FFBC1740000: C:\Windows\SYSTEM32\ntdll.dll
00007FFBBFC10000: C:\Windows\System32\KERNEL32.DLL
00007FFBBE4A0000: C:\Windows\System32\KERNELBASE.dll
00007FFBBE140000: C:\Windows\System32\ucrtbase.dll
00007FFBA3400000: C:\msys64\ucrt64\bin\libgcc_s_seh-1.dll
00007FFBAC330000: C:\msys64\ucrt64\bin\libwinpthread-1.dll
00007FFBAB5E0000: C:\msys64\ucrt64\bin\zlib1.dll
00007FFBA2D40000: C:\msys64\ucrt64\bin\libstdc++-6.dll
00007FFB929A0000: C:\code\llvm-project\llvm\build-ucrt64-clang\bin\libLLVM-15.dll
00007FFBC0050000: C:\Windows\System32\ADVAPI32.dll
00007FFBC1600000: C:\Windows\System32\msvcrt.dll
00007FFBBECD0000: C:\Windows\System32\sechost.dll
00007FFBBEA40000: C:\Windows\System32\RPCRT4.dll
00007FFBBF0D0000: C:\Windows\System32\ole32.dll
00007FFBBF8E0000: C:\Windows\System32\combase.dll
00007FFBBE2D0000: C:\Windows\System32\bcryptPrimitives.dll
00007FFBBF020000: C:\Windows\System32\GDI32.dll
00007FFBBE770000: C:\Windows\System32\gdi32full.dll
00007FFBBE350000: C:\Windows\System32\msvcp_win.dll
00007FFBBFEB0000: C:\Windows\System32\USER32.dll
00007FFBBE240000: C:\Windows\System32\win32u.dll
00007FFBC0100000: C:\Windows\System32\SHELL32.dll
00007FFBBE260000: C:\Windows\System32\cfgmgr32.dll
00007FFBBEB70000: C:\Windows\System32\shcore.dll
00007FFBBD9F0000: C:\Windows\System32\windows.storage.dll
00007FFBBD7C0000: C:\Windows\System32\profapi.dll
00007FFBBD760000: C:\Windows\System32\powrprof.dll
00007FFBBEFC0000: C:\Windows\System32\shlwapi.dll
00007FFBBD740000: C:\Windows\System32\kernel.appcore.dll
00007FFBBE2B0000: C:\Windows\System32\cryptsp.dll
00007FFBA2BF0000: C:\msys64\ucrt64\bin\libxml2-2.dll
00007FFBBEDE0000: C:\Windows\System32\WS2_32.dll
00007FFBA2770000: C:\msys64\ucrt64\bin\libzstd.dll
00007FFBA2650000: C:\msys64\ucrt64\bin\libiconv-2.dll
00007FFBA33D0000: C:\msys64\ucrt64\bin\liblzma-5.dll
00007FFBBFCD0000: C:\Windows\System32\IMM32.DLL

```
With this info, we know that the offending relocation is at `00007FF64EAAE4DD` in a module at `00007FF64E7F0000`. The module at that address is `lld.exe` itself, and by subtracting the base, we get the offset `0x2BE4DD` in lld.exe.

Similarly, we notice that the target of the relocation is `00007FFBA2E60A60`, which seems to be in `libstdc++-6.dll` at the offset `00007FFBA2D40000`. By subtracting that base, we find that the problematic symbol is at offset `0x120A60` in that DLL. By running `llvm-readobj --coff-exports /ucrt64/bin/libstdc++-6.dll` and searching for that offset, we find:
```
Export {
  Ordinal: 6707
  Name: __emutls_v._ZSt15__once_callable
  RVA: 0x120A60
}
```

By rerunning the link command that produced lld.exe, but adding `-Wl,-Map,map` we get a map file for that executable, where we can look for which object file might contain the offset `0x2BE4DD`. In this case, it's this:
```
.text$_ZNSt13__future_base17_Async_state_implINSt6thread8_InvokerISt5tupleIJZN3lld4coffL19createFutureForFileENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEE4$_12EEEESt4pairISt10unique_ptrIN4llvm12MemoryBufferESt14default_deleteISI_EESt10error_codeEED2Ev
                0x00000001402be410       0x8f lib/liblldCOFF.a(Driver.cpp.obj)
```

The offset here has the default image base of `0x140...` which makes it a bit harder to find. In practice I actually ran the link command with `-fuse-ld=lld` since I find lld's map output a bit easier to read. In the case of lld, the offset to find was `0x305BD` instead:
```
00030580 000000af    16         Driver.cpp.obj:(.text$_ZNSt13__future_base21_Async_state_commonV217_M_complete_asyncEv)
00030580 00000000     0                 std::__future_base::_Async_state_commonV2::_M_complete_async()
```

In any case, Driver.cpp.obj is the one that has got the problematic relocation. By rerunning the compilation command for Driver.cpp.obj, but switching out the `-c -o ....obj` to `-S -o -`, I find this bit in the output:
```
        leaq    __emutls_v._ZSt15__once_callable(%rip), %rcx
        callq   __emutls_get_address
```
When rerunning the same compilation command with GCC instead of Clang, the corresponding output contains this:
```
        movq    .refptr.__emutls_v._ZSt15__once_callable(%rip), %rcx
        call    __emutls_get_address
```

So that's it - for these cases, Clang should generate the access to the `__emutls_v._ZSt15__once_callable` variable (which ends up imported from a different DLL) with the usual `.refptr` indirection, so that it is safe when automatically dllimported.

mstorsjo:
And I thought this felt familiar - a similar issue was discussed before in https://github.com/msys2/MINGW-packages/issues/8706, with a fix in https://reviews.llvm.org/D102970. Now I wonder why that didn't help for this one...

mati865:
There might be some regression: https://github.com/msys2/MINGW-packages/issues/1677#issuecomment-1225728958
I have it on my backlog to check if it really regressed next week.

mstorsjo:
> There might be some regression: [#1677 (comment)](https://github.com/msys2/MINGW-packages/issues/1677#issuecomment-1225728958)
> I have it on my backlog to check if it really regressed next week.

That issue sounds like something entirely unrelated.

I think I see what the issue is. The fix in https://reviews.llvm.org/D102970 doesn't seem to apply when building with the msys2 clang that defaults to emulated tls, but if I manually add `-femulated-tls` to it, it ends up working. So apparently there's some different place in Clang which needs a tweak in a second place too, for the default-enabled emulated tls.

mstorsjo:
This seems like it should fix the issue:
```diff
diff --git a/clang/lib/CodeGen/CodeGenModule.cpp b/clang/lib/CodeGen/CodeGenModule.cpp
index 74c14fed6575..44ffd22b921f 100644
--- a/clang/lib/CodeGen/CodeGenModule.cpp
+++ b/clang/lib/CodeGen/CodeGenModule.cpp
@@ -1143,7 +1143,7 @@ static bool shouldAssumeDSOLocal(const CodeGenModule &CGM,
     // such variables can't be marked as DSO local. (Native TLS variables
     // can't be dllimported at all, though.)
     if (GV->isDeclarationForLinker() && isa<llvm::GlobalVariable>(GV) &&
-        (!GV->isThreadLocal() || CGM.getCodeGenOpts().EmulatedTLS))
+        (!GV->isThreadLocal() || CGM.useEmulatedTLS()))
       return false;
   }
 
diff --git a/clang/lib/CodeGen/CodeGenModule.h b/clang/lib/CodeGen/CodeGenModule.h
index f57afdca4942..b10c392dc1bb 100644
--- a/clang/lib/CodeGen/CodeGenModule.h
+++ b/clang/lib/CodeGen/CodeGenModule.h
@@ -746,6 +746,12 @@ public:
 
   const TargetCodeGenInfo &getTargetCodeGenInfo();
 
+  bool useEmulatedTLS() const {
+    if (CodeGenOpts.ExplicitEmulatedTLS)
+      return CodeGenOpts.EmulatedTLS;
+    return getTriple().hasDefaultEmulatedTLS();
+  }
+
   CodeGenTypes &getTypes() { return Types; }
 
   CodeGenVTables &getVTables() { return VTables; }
```

Without this, the fix for https://github.com/msys2/MINGW-packages/issues/8706 / https://reviews.llvm.org/D102970 never had any effect in the default-emulated-tls clang in msys2.

Let me know if this seems to work, and I can try to take it upstream for discussion (I'm not sure if this is the exact right place for the utility function etc).

lazka:
I can confirm that the above patch fixes https://github.com/msys2/msys2-tests/pull/9 for me. I've created #12852

mati865:
For the completeness I have rebuilt UCRT64 Clang with Clang+libstdc++ and LLD testsuite gave no warnings about relocations so this issue is indeed fixed.
There is one more issue with LLD tests where they hang on exit with UCRT (both UCRT64 and CLANG64 envs) but that's another thing on todo...

jeremyd2019:
> There is one more issue with LLD tests where they hang on exit with UCRT (both UCRT64 and CLANG64 envs) but that's another thing on todo...

This is because when LLD is running tests it disables the quick_exit path, resulting in global destructors being run from the 'DllMain'-like context, *after* all other threads have already been killed by the system.  So if it ends up waiting on a resource some other thread held, that other thread will never be releasing it...

mati865:
It doesn't hang with MSVCRT though.

lazka:
This is now enabled for all variants again (#12866), and all tests pass.

Thanks everyone!

PR #12562, opened 2022-08-14
qt5-static: update to 5.15.5

tsl0922:


lazka:
thanks

Biswa96:
Also make sure the static libs are not importing from shared libraries of any dependencies (except system DLLs). e.g. `nm libqt5*.a | grep __imp`

MehdiChinoune:
qt5-static is now at 5.15.7

Issue #12569, opened 2022-08-15
missing `rust.description` for mingw-w64-rust


Zapeth:
Normally this would be set in the `config.toml` when compiling the toolchain ([reference](https://github.com/rust-lang/rust/blob/4916e2b9e6ef8cee6f9c6abb75bd01ba9dc07e5c/config.toml.example#L521-L524)), but if I understand the compilation steps correctly it should also be possible via `--release-description` parameter when calling `configure` ([reference](https://github.com/rust-lang/rust/blob/4916e2b9e6ef8cee6f9c6abb75bd01ba9dc07e5c/src/bootstrap/configure.py#L152)).

I guess it would make most sense to set this the same as the package name, ie `mingw-w64-x86_64-rust` or `mingw-w64-clang-x86_64-rust`?

filnet:
I am working on it.

PR #12570, opened 2022-08-15
rust: Add release description

filnet:
fixes https://github.com/msys2/MINGW-packages/issues/12569

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -8,7 +8,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-rust-docs")
 pkgver=1.63.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Systems programming language focused on safety, speed and concurrency (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
@@ -114,6 +114,7 @@ build() {
     --host=$OSTYPE \
     --target=$OSTYPE \
     --release-channel=stable \
+    --release-description="Rev${pkgrel}, Built by MSYS2 project" \
     --enable-ninja \
     --enable-extended \
     --disable-llvm-static-stdcpp \


1480c1:
It might be better to match this closer to the output of `gcc --version` that we have with the whole `Build by MSYS2 project`, unless this component is something that's expected to be grepable or parsed in some way.

I do see that `rustc --version` (managed via `rustup` as far as I remember) on my linux box shows
```bash
rustc --version
rustc 1.61.0 (fe5b13d68 2022-05-18)
```

but I don't think the commit hash really matters for us since we are using a tarball rather than a checkout

Zapeth:
> but I don't think the commit hash really matters for us since we are using a tarball rather than a checkout

rustc currently just includes basic version information when built via tarball (ie no commit hash and no commit date), but there is a pending PR that addresses this -> https://github.com/rust-lang/rust/pull/100557

filnet:
Tell me what you want. I have no opinion.

filnet:
The description can be any string.
The string can contain these variables `${MINGW_PACKAGE_PREFIX}` `${_realname}` `${pkgver}` `${pkgrel}`.

1480c1:
Perhaps something like,
```bash
rustc --version
rustc 1.62.1 (2022-08-17 Built by MSYS2 project)
```
or exposing which subsystem it was built for along with date and ownership
```bash
rustc --version
rustc 1.62.1 (mingw64 Rev${pkgrel} 2022-08-17 Built by MSYS2 project)
```
or somehow (I don't think moving the version to the end is possible) to bring it closer to `gcc.exe --version` while retaining the date (don't know if we want to if we are aiming for reproducible builds)
```bash
rustc --version
rustc (Rev${pkgrel}, Built by MSYS2 project on 2022-08-17) 1.62.1
```

Just some ideas on how to put some things together, but I would say it's based on how easy it is to put them there, we probably want to convey the fact that this build is specifically from msys2 rather than an official build from the Rust Language organization for Windows so that way upstream devs can except that it might not be "pristine" and could have additional patches on top of it.

filnet:
I can do exactly like gcc:

```
$ gcc --version
gcc.exe (Rev3, Built by MSYS2 project) 12.1.0
$ gcc --version

filnet:
Do we really want the date ? `gcc` does not have it.

Zapeth:
I don't think date would be that useful, especially since that information should be somewhat tied to `${pkgrel}` and could be looked up if necessary.

As for possibly including subsystem information, are there any patches that are dedicated to a specific subsystem? If yes then it might be worth including, though I guess the subsystem information could also be derived from the executable itself.

But these are just my two cents, I mainly brought up this issue to point out the fact that the release description is currently empty, so I don't have any particular preference what that description should look like,

mati865:
I don't know if reproducible builds are a thing for us but many distros stopped inserting dates because it's not easily reproducible.

lazka:
> I don't know if reproducible builds are a thing for us but many distros stopped inserting dates because it's not easily reproducible.

The reproducible builds way would be https://reproducible-builds.org/docs/source-date-epoch/#bash--posix-shell

lazka:
Let's merge the minimal version, until some has an actual use case for more information.

PR #12575, opened 2022-08-15
libgd: update to 2.3.3

MehdiChinoune:


lazka:
I've reverted this as it broke mingw32/clang32 builds. To reproduce, running "dot -c" failed because it failed to find symbols in libgd.

MehdiChinoune:
https://github.com/msys2/MINGW-packages/runs/7846657595 didn't show anything.

lazka:
yes, I don't know yet what the problem was

PR #12559, opened 2022-08-14
Split wxsvg Into Multiple Packages Built Against Different wxWidgets Versions

AntumDeluge:
- replaces package `wxsvg` with `wxsvg-wx3.0`
- adds package `wxsvg-wx3.1`
- adds package `wxsvg-wx3.2`

PR #12578, opened 2022-08-16
Build wxSVG against wxWidgets 3.2

AntumDeluge:
- updates wxSVG to be built against wxWidgets 3.2
- changes package name to `wxsvg-wx3.2`

There are conflicts between wxWidgets packages, so multiple wxSVG packages cannot be built from a single `PKGBUILD` script. See: #12559

**Edit:** Until the wxWidgets version conflicts can be fixed, wxSVG can only be built against one version, or there would have to be unique `PKGBUILD` scripts for each version. See: [wxsvg_multi_pkgbuild](https://github.com/AntumDeluge/MINGW-packages/tree/wxsvg_multi_pkgbuild)

### DIFF
--- a/mingw-w64-wxSVG/PKGBUILD
+++ b/mingw-w64-wxSVG/PKGBUILD
@@ -1,16 +1,19 @@
 # Maintainer: Jordan Irwin <antumdeluge@gmail.com>
 
 _realname=wxsvg
-_wx_basever=3.0
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+_wx_basever=3.2
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-wx${_wx_basever}"
 pkgbase="mingw-w64-${_realname}"
 pkgver=1.5.23
-pkgrel=2
+pkgrel=1
 pkgdesc="A C++ library to create, manipulate and render SVG files (mingw-w64)"
 arch=(any)
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url="https://wxsvg.sourceforge.io/"
 license=("custom:wxWindows")
+provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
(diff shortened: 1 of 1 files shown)


mingw-w64-wxSVG/PKGBUILD
```diff
@@ -1,16 +1,18 @@
...
 pkgver=1.5.23
-pkgrel=2
+pkgrel=1
 pkgdesc="A C++ library to create, manipulate and render SVG files (mingw-w64)"
 arch=(any)
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url="https://wxsvg.sourceforge.io/"
 license=("custom:wxWindows")
+provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
```
lazka, line 15:
```suggestion
conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}")
```

I think we should move users over, no?

AntumDeluge: Oh, the [documentation I read](https://wiki.archlinux.org/title/PKGBUILD#replaces) said to use `provides` & `conflicts` if just uploading an alternative version of a package. But I guess I misunderstand the use of `replaces`.

AntumDeluge: Done.

lazka: I means all existing users of wxsvg will be automatically moved over to this package, and the old one will be uninstalled. It's somewhat questionable since they may not be compatible, but we don't have any users in MSYS2 itself, so hard to say.

The alternative is that user will stay on the old package and when they try to install wxsvg-wx3.2 it will complain about the conflict and let them switch.

AntumDeluge:
Ready for review again.

lazka:
thanks, waiting for the build queue to settle first.

AntumDeluge:
Please excuse my impatience. I know there have been a lot of updates recently. Just curious if there is an ETA for this?

PR #12579, opened 2022-08-16
jsoncpp: update to 1.9.5

MehdiChinoune:


mingw-w64-vrpn/PKGBUILD
```diff
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=7.34
```
1480c1, line 6: It seems there's a 7.35 https://github.com/vrpn/vrpn/releases and the commit from the above patch is in there https://github.com/vrpn/vrpn/commit/6944d138f29ca3c44cea76376e3db5ca5185c9eb, have you tried updating to that yet?

mingw-w64-vrpn/PKGBUILD
```diff
@@ -17,24 +17,27 @@ depends=("${MINGW_PACKAGE_PREFIX}-hidapi"
          "${MINGW_PACKAGE_PREFIX}-python"
          "${MINGW_PACKAGE_PREFIX}-freeglut"
          "${MINGW_PACKAGE_PREFIX}-swig")
-makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
-             "${MINGW_PACKAGE_PREFIX}-cc")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
```
1480c1, line 22: I think we can rely on ninja being a dependency of cmake, but up to you

mingw-w64-cmake/PKGBUILD
```diff
@@ -1,13 +1,13 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 # Contributor: Ray Donnelly <mingw.android@gmail.com>
 
-_bootstrap=0
+_bootstrap=1
```
lazka, line 4: what's this about?

lazka: ah, because of the dll name change

MehdiChinoune: Yes, cmake depends on jsoncpp. It fails to run against the new jsoncpp => It couldn't build by itself.

PR #12590, opened 2022-08-16
goxel: Update to 0.11.0

podsvirov:


Biswa96:
The updated package builds fine in my PC and without any of those compiler warnings.

podsvirov:
@lazka, can you check what wrong?

lazka:
I'm afraid I also can't reproduce locally.

Biswa96:
Well then, you could build and update the package manually like Alexpux did those days.

lazka:
> Well then, you could build and update the package manually like Alexpux did those days.

ok

lazka:
For future reference: the clang64 builds also fails here with a clean msys2 env with a minimal set of packages. I don't know what packages makes the difference.

PR #12591, opened 2022-08-17
openvr, bullet: Enable aarch64 builds

hmartinez82:
image


hmartinez82:
o.O I didn't touch that file.

hmartinez82:
@lazka  Is there precedent for that? The patch that failed, which is downloaded from https://patch-diff.githubusercontent.com/raw/ValveSoftware/openvr/pull/1524.patch, had a different sha256 checksum this time.

lazka:
@hmartinez82 PRs can be force pushed, so patches having a PR as source can change.

PR #12593, opened 2022-08-17
ogre3d: Fix samples plugin not exporting symbols to be loaded by SampleBrowser

hmartinez82:
image

### DIFF
--- a/mingw-w64-ogre3d/015-export-symbols-samples-plugin.patch
+++ b/mingw-w64-ogre3d/015-export-symbols-samples-plugin.patch
@@ -1,0 +1,12 @@
+diff -bur ogre-13.4.4-c/Samples/Common/include/SamplePlugin.h ogre-13.4.4/Samples/Common/include/SamplePlugin.h
+--- ogre-13.4.4-c/Samples/Common/include/SamplePlugin.h	2022-08-06 14:44:00.000000000 -0600
++++ ogre-13.4.4/Samples/Common/include/SamplePlugin.h	2022-08-16 22:43:44.863604600 -0600
+@@ -35,7 +35,7 @@
+ #if defined( OGRE_STATIC_LIB )
+ #  define _OgreSampleExport
+ #else
+-#  if (OGRE_PLATFORM == OGRE_PLATFORM_WIN32 || OGRE_PLATFORM == OGRE_PLATFORM_WINRT) && !defined(__MINGW32__)
++#  if (OGRE_PLATFORM == OGRE_PLATFORM_WIN32 || OGRE_PLATFORM == OGRE_PLATFORM_WINRT)
+ #    define _OgreSampleExport __declspec(dllexport)
+ #  elif defined ( OGRE_GCC_VISIBILITY )
+ #   define _OgreSampleExport  __attribute__ ((visibility("default")))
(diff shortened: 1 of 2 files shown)


mmuetzel:
While you are at it, you could also remove `${MINGW_PACKAGE_PREFIX}-pugixml` from `depends`. (See @MehdiChinoune's comment in the other commit.)

hmartinez82:
@mmuetzel Ok, should I keep it in makedepends and optdepends?

MehdiChinoune:
> @mmuetzel Ok, should I keep it in makedepends and optdepends?

Yes, and Please add `XML Converter` in front of `pugixml` in `optdepends`
```
            "${MINGW_PACKAGE_PREFIX}-pugixml: XML converter"
```

mmuetzel:
The libraries are exporting a lot less symbols now. (But I guess that was the point of this change.)
The package grokker suggests to rebuild the following packages:
```
mingw-w64-ogre3d
mingw-w64-ogitor-git
mingw-w64-opencv
```

I downloaded and installed the MINGW64 artifact from this PR. But it doesn't seem to be working for me. If I try to run `SampleBrowser.exe`, I see the following pop up window:
image

Am I doing something wrong?

hmartinez82:
@MehdiChinoune I need your input here. In the last version of Ogre3D, those dlls (the renderers) were in the ${MINGW_PREFIX}/bin folder, correct?

hmartinez82:
Could this line `-DOGRE_PLUGINS_PATH=lib/OGRE \` be the culprit?

lazka:
oh, I guess I merged too early, sry. what should I do?

lazka:
> The package grokker suggests to rebuild the following packages:

that could be because it's comparing to 1.12.6 in the repo

mmuetzel:
I haven't tested with the package in staging before this PR. So, this might have been an issue already before. Merging this PR might not have made anything worse than it was anyway.

hmartinez82:
@mmuetzelI @lazka and @MehdiChinoune I am able to repro , manually copying those Renderer dlls to the bin folder solves the issue.

The screenshot is running the sample browser from the build bin folder, that's why I didn't notice this issue. Should I remove the `-DOGRE_PLUGINS_PATH=lib/OGRE` line and give it a try?

mmuetzel:
> Should I remove the `-DOGRE_PLUGINS_PATH=lib/OGRE` line and give it a try?

Either that, or check why it looks for the plugins in the bin folder even though it was configured with that flag. (Your proposal might be the easier one of those options though.)

MehdiChinoune:
> @mmuetzelI @lazka and @MehdiChinoune I am able to repro , manually copying those Renderer dlls to the bin folder solves the issue.
> 
> The screenshot is running the sample browser from the build bin folder, that's why I didn't notice this issue. Should I remove the `-DOGRE_PLUGINS_PATH=lib/OGRE` line and give it a try?

You will end up with import libraries installed in bin.

I think we should create a patch to install runtime in bin and library/archive in lib.

MehdiChinoune:
see: #12595

mmuetzel:
Still failed to build in staging. This time with:
https://github.com/msys2-arm/msys2-autobuild/runs/7875552919?check_suite_focus=true#step:10:14172
```
  ==> Installing missing dependencies...
  resolving dependencies...
  warning: cannot resolve "mingw-w64-clang-aarch64-openvr", a dependency of "mingw-w64-clang-aarch64-bullet"
  :: The following package cannot be upgraded due to unresolvable dependencies:
        mingw-w64-clang-aarch64-bullet
  
  :: Do you want to skip the above package for this upgrade? [Y/n] 
```

What does that mean? `mingw-w64-clang-aarch64-openvr` *is* in the repos afaict:
https://repo.msys2.org/mingw/clangarm64/mingw-w64-clang-aarch64-opencv-4.5.5-4-any.pkg.tar.zst

Maybe, some kind of dependency cycle that needs to be broken differently?

mmuetzel:
Edit: Maybe fixed by something like https://github.com/msys2/msys2-autobuild/pull/59 ?

PR #12598, opened 2022-08-17
libwebp: switch to CMake

### LINKED ISSUE #12594: FR: Switch libwebp to CMake
More and more CMake-based projects rely on `find_package(WebP)` that's currently not possible on MSYS2.

Probably nice to haves:
* Adding SOVERSION to shared lib output name to avoid rebuilding the huge deps list
* Moving config files from "share/WebP/cmake" to the more usual "lib/cmake/WebP"

kmilos:
Resolves #12594

### DIFF
--- a/mingw-w64-libwebp/0001-mingw-cmake-output.patch
+++ b/mingw-w64-libwebp/0001-mingw-cmake-output.patch
@@ -1,0 +1,31 @@
+--- libwebp-1.2.4/CMakeLists.txt.orig	2022-08-06 01:49:26.000000000 +0200
++++ libwebp-1.2.4/CMakeLists.txt	2022-08-17 14:54:32.963092000 +0200
+@@ -138,7 +138,7 @@
+ 
+ function(configure_pkg_config FILE)
+   configure_file("${CMAKE_CURRENT_SOURCE_DIR}/${FILE}.in"
+-                 "${CMAKE_CURRENT_BINARY_DIR}/${FILE}")
++                 "${CMAKE_CURRENT_BINARY_DIR}/${FILE}" @ONLY)
+ 
+   if(HAVE_MATH_LIBRARY)
+     # MSVC doesn't have libm
+@@ -395,7 +395,9 @@
+     PROPERTIES VERSION
+                ${LT_CURRENT_MINUS_AGE}.${LT_AGE}.${LT_REVISION}
+                SOVERSION
+-               ${LT_CURRENT_MINUS_AGE})
++               ${LT_CURRENT_MINUS_AGE}
++               RUNTIME_OUTPUT_NAME
++               ${NAME_IN_MAKEFILE}-${LT_CURRENT_MINUS_AGE})
+ endmacro()
+ set_version(Makefile.am webp webp)
+ set_version(Makefile.am webpdecoder webpdecoder)
+@@ -723,7 +725,7 @@
+         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+-set(ConfigPackageLocation ${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}/cmake/)
++set(ConfigPackageLocation ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/)
+ install(EXPORT ${PROJECT_NAME}Targets
+         NAMESPACE ${PROJECT_NAME}::
+         DESTINATION ${ConfigPackageLocation})
--- a/mingw-w64-libwebp/PKGBUILD
+++ b/mingw-w64-libwebp/PKGBUILD
@@ -5,7 +5,7 @@ _realname=libwebp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.2.4
-pkgrel=1
+pkgrel=2
 pkgdesc="A library to encode and decode images in WebP format (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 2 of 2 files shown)


Biswa96:
One shared library got extra `lib` prefix in its name.

mingw-w64-libwebp/0001-mingw-cmake-output.patch
```diff
@@ -0,0 +1,31 @@
...
+   if(HAVE_MATH_LIBRARY)
+     # MSVC doesn't have libm
+@@ -395,7 +395,9 @@
+     PROPERTIES VERSION
+                ${LT_CURRENT_MINUS_AGE}.${LT_AGE}.${LT_REVISION}
+                SOVERSION
+-               ${LT_CURRENT_MINUS_AGE})
++               ${LT_CURRENT_MINUS_AGE}
++               RUNTIME_OUTPUT_NAME
++               ${TARGET_NAME}-${LT_CURRENT_MINUS_AGE})
```
Biswa96, line 19: Should DLLs name be changed from default ones?

kmilos: This is to keep DLL names identical to autotools ones (the default there includes the soversion in the name), so this becomes a drop-in replacement. Otherwise, need to rebuild a whole bunch of dependent packages...

Biswa96: There were some instances where DLL names were changed due to autotools → cmake conversion. Then dependent packages were rebuilt. The downside is that the patch has be maintained for lifetime.

mingw-w64-libwebp/0001-mingw-cmake-output.patch
```diff
@@ -0,0 +1,31 @@
...
+   if(HAVE_MATH_LIBRARY)
+     # MSVC doesn't have libm
+@@ -395,7 +395,9 @@
+     PROPERTIES VERSION
+                ${LT_CURRENT_MINUS_AGE}.${LT_AGE}.${LT_REVISION}
+                SOVERSION
+-               ${LT_CURRENT_MINUS_AGE})
++               ${LT_CURRENT_MINUS_AGE}
++               RUNTIME_OUTPUT_NAME
++               ${NAME_IN_MAKEFILE}-${LT_CURRENT_MINUS_AGE})
```
1480c1, line 19: Would this be a Windows specific change, or is this general? I can work on getting this upstreamed on the reasoning of keeping autotools and cmake closer in sync, but I am not sure if autotools on linux needs this

kmilos: This is indeed Windows (MINGW?) specific - AFAIK setting the SOVERSION property will automatically add that version suffix for Unix so libs (so autotools and cmake should be producing identical artefacts there already), but not so for Windows DLLs...

1480c1: Okay, after this PR is finallized and merged, I'll work on getting a `WIN32` specific version of this hunk for upstream, I'll also try to submit the `@ONLY` part since that's pretty standard either way

kmilos: If one wants to match up CMake and autotools output on MINGW (we don't care about MSVC DLLs as there is no autotools build there), there are several options:

https://fossies.org/linux/uriparser/CMakeLists.txt
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/8162eddf041e0be26f5c671bb6528723c55fed9d/sharedlib/CMakeLists.txt#L65

I'll let the MSYS2 maintainers decide if this is desired behaviour or not...

lazka: Whatever results in fewer patches for us.

I don't mind rebuilds due to one time changes. DLL name changes on every minor release without any ABI break would be annoying though.

mingw-w64-libwebp/0001-mingw-cmake-output.patch
```diff
@@ -0,0 +1,31 @@
...
+ endmacro()
+ set_version(Makefile.am webp webp)
+ set_version(Makefile.am webpdecoder webpdecoder)
+@@ -723,7 +725,7 @@
+         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+-set(ConfigPackageLocation ${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}/cmake/)
++set(ConfigPackageLocation ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/)
```
1480c1, line 28: Sent upstream https://chromium-review.googlesource.com/c/webm/libwebp/+/3876624

mingw-w64-libwebp/0001-mingw-cmake-output.patch
kmilos: @1480c1 How about upstreaming this one as well?

1480c1: https://chromium-review.googlesource.com/c/webm/libwebp/+/3877044

kmilos: @1480c1 This seems to be default behaviour on Cygwin/MSYS already, so it's just MINGW we're talking about. An even easier patch might be to add enable this setting on MINGW like Cygwin/MSYS do instead of trying to set the output name directly: https://gitlab.kitware.com/cmake/cmake/-/blob/83686df1d63e7d2c922de19489ea2bab42e98a4a/Modules/Platform/CYGWIN.cmake#L62

kmilos: Actually, this patch might not be needed at all: I found that simply setting ` -DCMAKE_SHARED_LIBRARY_NAME_WITH_VERSION=1` at configure time has the same effect. @1480c1

1480c1: > Actually, this patch might not be needed at all: I found that simply setting `-DCMAKE_SHARED_LIBRARY_NAME_WITH_VERSION=1` at configure time has the same effect. @1480c1

https://chromium-review.googlesource.com/c/webm/libwebp/+/3877044/comments/12923fd7_884ff122

@jzern did note that, personal issue would be that it would be project specific or rely on the user to set that, meaning one person compile for `MINGW` might not have the same effect as another compiling, if they do not know about that cmake variable

PR #12606, opened 2022-08-17
s2geometry: New package

podsvirov:


Biswa96:
Patches name should not contain version name. Other than that the package looks good to me 👌

podsvirov:
Please see Patch files section in https://www.msys2.org/wiki/Creating-Packages/

Biswa96:
I mean the version number (here 0.10.0), not the leading patch numbers.

podsvirov:
> Please see Patch files section in https://www.msys2.org/wiki/Creating-Packages/

Please see target description

Issue #10864, opened 2022-02-26
svt-av1 failed with ffmpeg


wszqkzqk:
```shell
Stream mapping:
  Stream #0:0 -> #0:0 (hevc (native) -> av1 (libsvtav1))
  Stream #0:1 -> #0:1 (aac (native) -> aac (native))
Press [q] to stop, [?] for help
Svt[info]: -------------------------------------------
Svt[info]: SVT [version]:       SVT-AV1 Encoder Lib v0.9.1
Svt[info]: SVT [build]  :       GCC 11.2.0       64 bit
Svt[info]: -------------------------------------------
[1]    1396 segmentation fault
```
And then it just exit with out encoding.

wszqkzqk:
And downgrade `svt-av1` to 0.9.0 fixed this.

wszqkzqk:
OK, it works now.

PR #12608, opened 2022-08-18
SVT-AV1: update to v1.2.1

wszqkzqk:


kmilos:
Might want to bump libavif as well...

MehdiChinoune:
1- minor versions supposed to be compatible, there is no reason to rebuild Both `ffmpeg` and `libavif`.
2- Both `ffmpeg` and `libavif` were rebuilt before `svt-av1` in CI, They were rebuilt against the svt-av1 1.1.0

wszqkzqk:
> 1- minor versions supposed to be compatible, there is no reason to rebuild Both `ffmpeg` and `libavif`. 2- Both `ffmpeg` and `libavif` were rebuilt before `svt-av1` in CI, They were rebuilt against the svt-av1 1.1.0

But ffmpeg cannot work with higher version of `svt-av1`.

MehdiChinoune:
> But ffmpeg cannot work with higher version of `svt-av1`.

Is it a known issue (link).

wszqkzqk:
> > But ffmpeg cannot work with higher version of `svt-av1`.
> 
> Is it a known issue (link).

Yes, [here is it](https://github.com/msys2/MINGW-packages/issues/10864)

wszqkzqk:
> 1- minor versions supposed to be compatible, there is no reason to rebuild Both `ffmpeg` and `libavif`. 2- Both `ffmpeg` and `libavif` were rebuilt before `svt-av1` in CI, They were rebuilt against the svt-av1 1.1.0

Isn't the package built after the rebuilt of its dependency?

MehdiChinoune:
> Yes, [here is it](https://github.com/msys2/MINGW-packages/issues/10864)

That was before 1.0.0, If it still hepenning then your report an issue to `svt-av1` about incompatibility of their minor releases.
Could you please verify if `ffmpeg` still works after updating `svt-av1` and rebuild it if required.

wszqkzqk:
> > Yes, [here is it](https://github.com/msys2/MINGW-packages/issues/10864)
> 
> That was before 1.0.0, If it still hapenning then you should report an issue to `svt-av1` about incompatibility of their minor releases. Could you please verify if `ffmpeg` still works after updating `svt-av1` and rebuild it if required.

I have verified `svt-av1` v1.2.1 fails with the ffmpeg built with `svt-av1` v1.1.0.

```
Press [q] to stop, [?] for help
Svt[info]: -------------------------------------------
Svt[info]: SVT [version]:       SVT-AV1 Encoder Lib v1.2.0
Svt[info]: SVT [build]  :       GCC 12.1.0       64 bit
Svt[info]: -------------------------------------------
```
The encoding process was blocked after the information above showed and the output file contains nothing.

MehdiChinoune:
The issue should be reported upstream, either to `svt-av1` or `ffmpeg`.

1480c1:
Fields were added at the end of the configuration struct between 1.1.0 and 1.2.1, which unfortunately means that the struct in FFmpeg that holds a statically allocated copy of the struct would not have enough size to hold it, it would require changing FFmpeg's libsvtav1.c to change it to a dynamically allocated one or add some other padding between it and the other fields in their struct to prevent potential issues when a struct is extended

lazka:
@MehdiChinoune please consider unblocking people -> https://github.com/settings/blocked_users

PR #12655, opened 2022-08-18
vulkan-loader: update to 1.3.224

MehdiChinoune:


lazka:
for some reason this failed in autobuild

MehdiChinoune:
> for some reason this failed in autobuild

Vulkan-loader 1.3.224 is not compatible with Vulkan-headers 1.3.225

lazka:
I've reverted both for now to unblock autobuild. I hope that's ok.

MehdiChinoune:
> I've reverted both for now to unblock autobuild. I hope that's ok.

Please re-revert vulkan-headers, see #12698

lazka:
ok, done

PR #12685, opened 2022-08-18
python-hjson: update to 3.1.0

MehdiChinoune:


lazka:
this adds a setuptools script and needs the shebang line fixed

PR #12696, opened 2022-08-19
update python-pyopenssl to 22.0.0 and python-cryptography to 37.0.4

MehdiChinoune:
python-cryptography now requires rust which is not available for clangarm64 and clang32


lazka:
There are 5 packages depending on this form clangarm64, which means we should create a separate (outdated) package for that, similar to https://packages.msys2.org/base/mingw-w64-librsvg-2.40

I can look into that if wanted.

lazka:
I guess the same applies to pyOpenSSL, puh, this is getting tricky..

MehdiChinoune:
I think we could merge this now.

lazka:
thanks!

PR #12704, opened 2022-08-19
python-numba: update to 0.56.0

MehdiChinoune:
numba 0.56.0 requires llvmlite>=0.39 which itself doesn't support any LLVM>11!


lazka:
> numba 0.56.0 requires llvmlite>=0.39 which itself doesn't support any LLVM>11!

yes, it's a pain...

PR #12711, opened 2022-08-20
openldap: update to 2.6.3

MehdiChinoune:


Biswa96:
Is it possible to disable slapd (i.e. add `--disable-slapd` option) for all MSYSTEMs? `sys/uio.h` was added in this commit in upstream https://git.openldap.org/openldap/openldap/-/commit/9f4de680e302d6045ca778b2a93dc21adf952296.

PR #12607, opened 2022-08-18
gegl: Support aarch64 build

hmartinez82:
@Biswa96 @Alexpux I'm not sure if ruby and/or openmp are still deps (see https://www.gegl.org/build.html). I was able to build in aarch64 without either and just disabling the dep on lua. 
Should we remove dep on openmp and ruby?
There's not mention of openmp nor ruby on the meson configuration output, even in x64 builds:
```
gegl 0.4.38

  Directories
    prefix          : /clang64
    libdir          : lib

  GEGL docs
    Reference       : False
    Docs            : False

  Optional features
    Build workshop  : False
    Introspection   : True
    Vala support    : True

  Optional build utilities
    asciidoc        : True
    dot             : False
    pygobject       : False
    source-highlight: False
    w3m             : False

  Optional dependencies
    avlibs          : True
    Cairo           : True
    GDKPixbuf       : True
    gexiv2          : True
    Jasper          : True
    lcms            : True
    libnsgif        : True
    libraw          : True
    Luajit          : True
    maxflow         : True
    mrg             : False
    Pango           : True
    pangocairo      : True
    poly2tri-c      : True
    poppler         : True
    OpenEXR         : True
    rsvg            : True
    SDL1            : False
    SDL2            : True
    spiro           : True
    TIFF            : True
    umfpack         : True
    V4L             : False
    V4L2            : False
    webp            : True

  Subprojects
    libnsgif        : YES
    poly2tri-c      : YES

  User defined options
    auto_features   : enabled
    buildtype       : plain
    prefix          : /clang64
    docs            : false
    introspection   : true
    libv4l          : disabled
    libv4l2         : disabled
    mrg             : disabled
    pygobject       : disabled
    sdl1            : disabled
```

PR #12731, opened 2022-08-20
gegl: Depend on libheif directly instead of transitively via vala,graphviz,libgd

hmartinez82:
See discussion in [#12607 ](https://github.com/msys2/MINGW-packages/pull/12607#issuecomment-1221391113)

### DIFF
--- a/mingw-w64-gegl/PKGBUILD
+++ b/mingw-w64-gegl/PKGBUILD
@@ -5,10 +5,10 @@ _realname=gegl
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.4.38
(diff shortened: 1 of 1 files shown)


Biswa96:
I am not sure if it is correct to add libheif in gegl dependencies. libheif is not mentioned even a single time in gegl repository.

hmartinez82:
@Biswa96 should we add the libheif dep to jasper then? https://github.com/jasper-software/jasper/search?q=heif

Biswa96:
libheif is optional in jasper. I think I found the issue. jasper's pkgconfig file has libheif but libheif is not present at build time. We may need to fix the 002-pkg-config-static.patch.

hmartinez82:
@Biswa96 can you tackle that change? :)

PR #12732, opened 2022-08-20
libdsm: update to 0.4.3

MehdiChinoune:


Biswa96:
This may break vlc because the DLL name has changed.

PR #12748, opened 2022-08-21
jasper: Fix pkgconfig file

Biswa96:
* Remove libheif from Requires.private field because it is not present as a dependency.
* Add prefix field to fix path relocation.
* Remove gcc_s from Libs.private because it is for gcc only.


Biswa96:
@hmartinez82 Would you like to try to build gegl with this change? I have also fixes other issues including the unnecessary libheif in pkgconfig.

hmartinez82:
@Biswa96 It built in my Surface Pro X :)

PR #12753, opened 2022-08-21
libdsm: Temporarily enable versioned DLL

Biswa96:


Biswa96:
Please review this pull request. This is a temporary workaround due to VLC can not be re-built right now. The libdsm DLL name has been changed in latest version update and libdsm_plugin.dll of vlc package depends on it.

MehdiChinoune:
Maybe you should fix building vlc instead.
protobuf also showed vlc in package-grocker https://github.com/msys2/MINGW-packages/runs/7861009282
`dav1d` shouldn't be updated in first place, so maybe creating a temporary `dav1d0.9` would be an option!

Biswa96:
There is a patch in upstream for dav1d. Let me check if that works.

MehdiChinoune:
> There is a patch in upstream for dav1d. Let me check if that works.

see #12755

Biswa96:
Thanks for the fix.

PR #12773, opened 2022-08-22
libjxl, hiredix, ccache: Enable aarch64 builds

hmartinez82:
After this, the last piece for GIMP on aarch64 is graphviz, which right now is failing to compile because ghostscript on aarch64 is broken (it builds, but the ps2pdf tool for instance is not working, only on aarch64 and I don't know why 😑)


Biswa96:
May I ask what error were being fixed by the mingw-w64-libjxl/0004-aarch64.patch file? From the changes, it seems that the fix is for all CPU architectures.

hmartinez82:
@Biswa96 it's, not. It's just for aarch64. It's hard to see that with the few lines in the patch, that that whole section is only enabled if ARM64 is detected. The issue at hand is that, originally, once ARM64 was detected then the code assumes it has to be Linux:
https://github.com/libjxl/libjxl/blob/v0.6.1/tools/cpu/cpu.cc#L25

PR #12774, opened 2022-08-23
slim-simulator: update to version 4.0

rdinnager:
Just updating to latest version (v4.0). No changes needed except the sha256sums.


rdinnager:
I keep on having issues with clang, especially with gnulib. I'd like to just drop support for this architecture, since it is not generally needed for the users of this software. It is only easy for me to test on mingw64 locally anyway, and generally I haven't had any issues with ucrt, so might as well leave that in?

lazka:
sure, since it's a leaf package and you are the maintainer dropping clang support is fine.

PR #12776, opened 2022-08-23
transmission: split package into cli,gtk,qt

fauxpark:
Aligning more closely with the Arch package: https://archlinux.org/packages/extra/x86_64/transmission/
The existing package only builds the GTK client.

For some reason the 3.00 release tarball does not completely match up with the 3.00 source tarball (excluding the usual generated files etc): https://github.com/transmission/transmission/discussions/3585
Not sure what to make of that, but it appears to compile fine with the missing files patched in.


lazka:
Fine with me in general (the gtk/qt split etc), but:

* Arch added libtransmission for some external app we don't have (https://bugs.archlinux.org/task/72015), and also it's not a shared library, so seems a bit pointless.
* Upstream dropped autotools 1.5 years ago (https://github.com/transmission/transmission/pull/1465), so it doesn't seem wise to switch to it now.

Given the that upstream hasn't released in a while I'm fine with packaging from git, if that makes things easier somehow.

LABPBS:
about a year ago i was thinking about updating the transmission package and adding the daemon, that was before i fully settled for qbittorrent as my client (which in the last few releases landed some fixes to build qbittorrent-nox (the web UI daemon) on windows

FWIW heres my last custom build of transmission (the commit is about the last one before the giant ongoing c++ rewrite, and the other patch is one that was laying around in the PRs and landed after the c++ rewrite) https://github.com/LABPBS/MINGW-packages/commit/71a357a03dccb9fe4f08b102216891ee53d5cdeb 

leaving that there cause i may not commit to make a good PR for that but hopefully it can help here

fauxpark:
Okay, I've removed libtransmission and gone back to CMake. All the same files appear to be generated, except for the web UI assets in transmission-cli: https://archlinux.org/packages/extra/x86_64/transmission-cli/
This is because there is no CMakeLists in the `web` folder, only Makefile.am. Not sure what to do about this...

And for future reference, the patch here (`skip-broken-cmake-rc.patch`) appears to be due to the GTK3 CFLAGS:
```
$ pkg-config --cflags gtk+-3.0
-IC:/msys64/mingw64/include/gtk-3.0 -IC:/msys64/mingw64/include/pango-1.0 -IC:/msys64/mingw64/include -IC:/msys64/mingw64/include/glib-2.0 -IC:/msys64/mingw64/lib/glib-2.0/include -IC:/msys64/mingw64/include/harfbuzz -IC:/msys64/mingw64/include/freetype2 -IC:/msys64/mingw64/include/libpng16 -IC:/msys64/mingw64/include/fribidi -IC:/msys64/mingw64/include/cairo -IC:/msys64/mingw64/include/lzo -IC:/msys64/mingw64/include/pixman-1 -IC:/msys64/mingw64/include/gdk-pixbuf-2.0 -IC:/msys64/mingw64/include/atk-1.0 -mms-bitfields
```
These are passed to `windres`, which does not know what `-m` means. Is this something that could/should be fixed in the GTK package?

LABPBS:
> All the same files appear to be generated, except for the web UI assets This is because there is no CMakeLists in the `web` folder, only Makefile.am. Not sure what to do about this...

dont think we could package the web ui files, however transmission allows loading custom web uis (and i used one when i used it) so it should be fine without packaging them, im not an actual transmission user but if i wanted to use the daemon/web ui i would fetch a custom one. for reference it looks in the env var ``TRANSMISSION_WEB_HOME`` or in specific windows dirs according to https://github.com/transmission/transmission/blob/3.00/libtransmission/platform.c#L525

Biswa96:
We can workaround those requirements. Is this pull request ready to merge?

fauxpark:
Yep, ready to go.

> for reference it looks in the env var TRANSMISSION_WEB_HOME or in specific windows dirs

I did see that, and I wonder if we should patch this to use the Unix behaviour with `XDG_DATA_DIRS`? Or is it okay for MSYS2 packages to "reach" outside of the MSYS2 environment like this?

Biswa96:
> These are passed to windres, which does not know what -m means. Is this something that could/should be fixed in the GTK package?

The `-mms-bitfields` flag comes from glib-2.0.pc pkgconfig file. And glib adds that for Windows platform only.

```meson
if host_system == 'windows' and cc.get_id() != 'msvc' and cc.get_id() != 'clang-cl'
  # Ensure MSVC-compatible struct packing convention is used when
  # compiling for Win32 with gcc. It is used for the whole project and exposed
  # in glib-2.0.pc.
  win32_cflags = ['-mms-bitfields']
```

fauxpark:
So I guess the solution is to remove it somehow before passing the cflags to windres?

Biswa96:
I did not find any way to remove a specific cflags for a specific file in cmake. One way is to remove `-mms-bitfields` for whole transmission-gtk target. But that may break ABI with glib and other, I guess.

lazka:
We can likely remove it from glib since ms-bitfields is the default since 2012 (GCC 4.7): https://github.com/gcc-mirror/gcc/commit/a6c467624ade35128

PR #12777, opened 2022-08-23
libbotan: Update to 2.19.2

jankoerner:
I just updated botan from 2.18.2 to 2.19.2 which was released on 2022-06-03


lazka:
thanks!

PR #12792, opened 2022-08-25
kde-frameworks: update to 5.97.0 - Part1

MehdiChinoune:
kcore-addons fails on 32bit envs. Something related to `__stdcall`.


Biswa96:
The issue seems to be in mingw-w64 headers. Let me reproduce the issue in my system. Thanks for updating these KDE packages.

Biswa96:
I have sent a patch to upstream mingw-w64 https://sourceforge.net/p/mingw-w64/mailman/message/37698415/. It may fix the issue in kcoreaddons-qt5 package. But I can not build all the packages using mingw32 in my old rusty box 😢

Biswa96:
ucrt64 build failed with kcompletion-qt5 package. I have built it in my PC without any issue. I think this can be fixed in autobuild by rebuilding.

MehdiChinoune:
> ucrt64 build failed with kcompletion-qt5 package. I have built it in my PC without any issue. I think this can be fixed in autobuild by rebuilding.

The previous time it failed because of the other random `#include "stdlib.h"`.

PR #12793, opened 2022-08-25
SDL2: update to 2.24.0

madebr:
- Remove patch fixing sdl2.pc/sdl2-config relocation
- Remove patch fixing link order
- Update patch providing sdl2-config mixed windows path
- Remove patch fixing joystick input with mingw-w64
- Build and package libSDL2_test.a
- CMake installs the license itself

tested by building using `makepkg-mingw -sCLf` for mingw-w64-x86_64

/cc @Alexpux


Biswa96:
Thank you for the contribution. Just asking.

* Are you sure that the linking order patch is not required anymore? There are some differences in pkgconfig and sdl2-config files.
* Would you like to also update [other SDL2 packages](https://packages.msys2.org/search?q=SDL2)? That may ensure if other packages are linking correctly with this update.

madebr:
> Thank you for the contribution. Just asking.
> 
>     * Are you sure that the linking order patch is not required anymore? There are some differences in pkgconfig and sdl2-config files.

The issue with link order comes from the linker not retaining the `main` symbol from the `libSDL2main.a` static library. Upstream now handles this [in a different way](https://github.com/libsdl-org/SDL/blob/8c9beb0c873f6ca5efbd88f1ad2648bfc793b2ac/CMakeLists.txt#L3036-L3042).
This is tested on ci by building a gui and cli gui application using the installed cmake config scripts, sdl2.pc file and sdl2-config scripts.

So I'm quite confident SDL2 upstream has fixed this.

To be ultimately sure, the SDL2 package contains a cmake/sdl2.pc/sdl2-config self-test.
I suppose msys packaging conventions don't allow for doing (short) self tests?


>     * Would you like to also update [other SDL2 packages](https://packages.msys2.org/search?q=SDL2)? That may ensure if other packages are linking correctly with this update.

I'll take a stab at it.

madebr:
About updating the other `SDL2_*` packages for testing this one.
It looks like these updates are already pending (except for SDL2_net)

And FYI, the SDL2_* libraries won't test the link order as these don't link to SDL2main and don't contain a main function.

Biswa96:
Thanks for the explanation.

Issue #12854, opened 2022-08-27
File "command line", line 1: Error: Unbound module Stdlib - OCaml version 4.11.1


demosthenesk:
### Description

I have just installed msys2 ocaml packages :
mingw32/mingw-w64-i686-ocaml 4.11.1-2 [installed]
mingw32/mingw-w64-i686-ocaml-compiler-libs 4.11.1-2 [installed]
mingw64/mingw-w64-x86_64-ocaml 4.11.1-2 [installed]
mingw64/mingw-w64-x86_64-ocaml-compiler-libs 4.11.1-2 [installed]

when i try to run from mingw64 or mingw32 the ocaml command i get the following error:
$ ocaml
        OCaml version 4.11.1

File "command line", line 1:
Error: Unbound module Stdlib

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Microsoft Windows [Version 10.0.22000.856]

### MINGW environments affected

- [X] MINGW64
- [X] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Expected behavior

to run ocaml without error

### Actual behavior

$ ocaml
        OCaml version 4.11.1

File "command line", line 1:
Error: Unbound module Stdlib

### Repro steps

1. Run MSYS2 MINGW32 or MSYS2 MINGW64
2. Run command ocaml
3. Get the error

Biswa96:
Thank you for reporting the issue. The mingw ocaml package is full of workarounds. I am thinking of updating the package and upstream the necessary changes. If things go well I will provide some news here.

MehdiChinoune:
It looks like It is something related to permissions.
https://discuss.ocaml.org/t/unbound-module-stdllib/5133/14

After changing the permission of `/mingw64/lib/ocaml/stdlib.cmi` to `755`, `ocaml` runs but I couldn't get out (CTRL+C and CTRL+D don't work!).

Biswa96:
The issue seems to be related to path relocation. The workaround is to add the full ocaml standard library path with `CAMLLIB` or `OCAMLLIB` environment variable. This works in my case

```
$ CAMLLIB=/f/msys64/ucrt64/lib/ocaml ocaml
```

[...]

PR #12858, opened 2022-08-27
ocaml: update to 4.14.0

MehdiChinoune:


MehdiChinoune:
Unfortunately It doesn't fix #12854

Biswa96:
Is it possible to merge this update? The update itself does not cause any issue with some little examples from https://ocaml.org/docs/first-hour.

MehdiChinoune:
see: #13382

PR #12859, opened 2022-08-27
lokalize: Update to 22.08.0

podsvirov:


Biswa96:
Is it possible not to change the patch file name? Same patch file is easier to track in git history.

podsvirov:
> Is it possible not to change the patch file name? Same patch file is easier to track in git history.

Ok. Done.

Issue #10620, opened 2022-01-27
Impossible to compile okular with msys2.


jointri:
There isn't an error with the cmake making the makefile, but one issue that arises at one point be it 30% or 99% is that kf.doctools.core: Error: Could not find kdoctools catalogs, which would result in terminating the compilation. I even did the prefix.sh script like I was supposed to, so this seems to be a msys2 issue. Is there any kdoctools calalogs package? If not, can it please be added. It's the missing piece to be able to compile okular from source because the cmake was satisfied with dependencies, but not the makefile it generated.

Biswa96:
Did you install the [kdoctools-qt5][1] package?

[1]: https://packages.msys2.org/base/mingw-w64-kdoctools-qt5

MehdiChinoune:
I think this is an upstream bug, Their cmake setup should detect kdoctools before building.

jointri:
> Did you install the [kdoctools-qt5](https://packages.msys2.org/base/mingw-w64-kdoctools-qt5) package?

Yes. I had this installed among one of my first tries with compiling okular.



> I think this is an upstream bug, Their cmake setup should detect kdoctools before building.

Are you sure?

MehdiChinoune:
> Are you sure?

If the failure was due the absence of a package, then yes.
The role of CMake is to generate projects (Makefiles, ninja-files, VS ...) and check dependencies.

jointri:
> > Are you sure?
> 
> If the failure was due the absence of a package, then yes. The role of CMake is to generate projects (Makefiles, ninja-files, VS ...) and check dependencies.

The reason why I'm skeptical is that the KDE buildbot is able to generate nightly releases for windows without issues. I wonder why I cannot do that with my msys2 on windows.

MehdiChinoune:
Another possibility is that kdoctools has an issue on MSYS2/MINGW, for that we need a minimized reproducer.

jointri:
> Another possibility is that kdoctools has an issue on MSYS2/MINGW, for that we need a minimized reproducer.

I got a theory. We should have kf5 and other KDE libraries under /usr as well because I think kdoctools is looking for a hardcoded /usr/share/so on/and/so forth. I think I'll try a fresh profile and see how the results turn out moving kf5 from /mingw64 to /usr.

MehdiChinoune:
> > Another possibility is that kdoctools has an issue on MSYS2/MINGW, for that we need a minimized reproducer.
> 
[...]

PR #12865, opened 2022-08-28
kde-frameworks: update to 5.97.0 - Part3

MehdiChinoune:
`kdoctools` fails to run.
see #10620


Biswa96:
Another path relocation issue with kdoctools. Let me check some code.

MehdiChinoune:
Maybe this:
https://github.com/KDE/kdoctools/blob/v5.97.0/config-kdoctools.h.cmake
used by:
https://github.com/KDE/kdoctools/blob/v5.97.0/src/meinproc.cpp#L2
https://github.com/KDE/kdoctools/blob/v5.97.0/src/xslt.cpp#L10

Biswa96:
Not only that. Every DTD file contains full path from CI. Those need to be handled somehow. Maybe a install file.

PR #12866, opened 2022-08-28
clang: try to build with clang and full targets again

lazka:


mati865:
The dependencies are wrong, compiler-rt and libc++ are used only in clangprefix but right now they are listed when building with clang.

PR #12869, opened 2022-08-29
kdoctools-qt5: Fix path relocation for data directory

Biswa96:


MehdiChinoune:
I have installed the artifacts locally, and It seems It works.

Biswa96:
Thank you for testing it. Do you know any other package which has workaround related to kdoctools?

@podsvirov lokalize package will have docs with this fix.

podsvirov:
> lokalize package will have docs with this fix.

Good!

PR #12871, opened 2022-08-29
OpenEXR: Update to 3.1.5

MehdiChinoune:
I have fixed building OpenEXR libraries on CLANG environments, but building tools still fails because of some missing imported symbols.


MehdiChinoune:
Beside missing tools on CLANG envs, All the dependents packages build or fixed to build against OpenEXR/Imath 3.1.5. Except Devil which is an unmaintained project (last activity from 2017) and ogre3d which doesn't want to support OpenEXR 3 (https://github.com/OGRECave/ogre/issues/2179).

I am going to merge this PR in the next few hours, if there is no objection.

lazka:
Does this new version include ilmbase/pyilmbase? If yes maybe we should add conflicts/replaces so updates are smooth?

MehdiChinoune:
> Does this new version include ilmbase/pyilmbase? If yes maybe we should add conflicts/replaces so updates are smooth?

Ilmbase/pyilmbase are no longer offered!
Parts of Ilmbase went to Imath and others went to OpenEXR.

I have marked `imath` to conflict with `ilmbase` because It contains some same files.

https://github.com/AcademySoftwareFoundation/Imath/blob/main/docs/PortingGuide2-3.md

lazka:
I see, thanks. I think we should make OpenEXR conflict/replace with ilmbase/pyilmbase, so those packages get removed on update for all users. Otherwise they have to manually remove them to get rid of the conflict.

MehdiChinoune:
I am OK with conflict/replace, but not with `provide`.

lazka:
sounds good.

lazka:
Done in b03874b931c7936

MehdiChinoune:
> Done in b03874b931c7936

That shouldn't be done until everything got built in msys2-autobuild. 'opencolorio' now fails to build.

lazka:
> That shouldn't be done until everything got built in msys2-autobuild. 'opencolorio' now fails to build.

I just had to find the "right" build order :)

PR #12872, opened 2022-08-29
openblas: Use less aggressive optimization with GCC-12.

### LINKED ISSUE #12857: Segmentation fault with mingw-w64-x86_64-openblas package
### Description

I encountered a segmentation fault when I developed a c++ program using LAPACK zheev driver included in the openblas package (`-lopenblas` option). I'm not an expert of c++ but at least the same program ran correctly with LAPACK with reference BLAS (`-llapack -lblas` options with mingw-w64-x86_64-lapack package).

### Verification

### Windows Version

Microsoft Windows [Version 10.0.22000.856]

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Expected behavior

The compiled program runs without error and finishes silently.

### Actual behavior

The compiled program echoes "Segmentation fault" and finished with a non-zero exit code.

### Repro steps

I have used *Windows sandbox* to try the following steps in a clean environment. However, the segmentation fault should be reproduced even without using the sandbox.

1. Install MSYS2 from https://www.msys2.org/
2. Run `pacman -Syu` twice
3. Run `pacman -S mingw-w64-x86_64-gcc mingw-w64-x86_64-openblas`
4. Run the following
    ```
    mkdir test_openblas
    cd test_openblas
    nano test_openblas.cpp
    g++ test_openblas.cpp -o a.out -lopenblas
    ./a.out
    ```

The content of the *test_openblas.cpp* is shown below.
```cpp
#include <complex>
#include <vector>

using complex = std::complex<double>;

extern "C" {
  void zheev_(const char& JOBZ, const char& UPLO, const int& N, complex* A,
              const int& LDA, double* W, complex* WORK, const int& LWORK,
              double* RWORK, int& INFO);
};


int main() {
    using std::vector;
    constexpr int n = 256;
    vector<complex> a(n*n);
    int lda = n;
    double w[n];
    int lwork;
    complex work_temp[1];
    double rwork[3*n - 2];
    int info;
    zheev_('V', 'U', n, a.data(), lda, w, work_temp, -1, rwork, info);
    lwork = (int) work_temp[0].real();
    auto work = new complex[lwork];
    zheev_('V', 'U', n, a.data(), lda, w, work, lwork, rwork, info); // Segmentation fault
    return info;
}
```

A strange thing I found is that it runs correctly when `n = 255`.

_No response_

mmuetzel:
See also: https://github.com/xianyi/OpenBLAS/issues/3740

Fixes #12857

### DIFF
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -69,7 +69,8 @@ _build_openblas() {
     else
       # GCC 12 enables `-ftree-vectorize` at optimization level `-O2`.
       # Library built with that optimization might segfault for some functions.
-      # See also: https://github.com/xianyi/OpenBLAS/issues/3740
+      # See https://github.com/msys2/MINGW-packages/issues/12857 and
+      # also see https://github.com/xianyi/OpenBLAS/issues/3740
       _optimization_opts=("-DCOMMON_OPT=-O2 -fno-tree-vectorize"
                           "-DCCOMMON_OPT=-O2 -fno-tree-vectorize"
                           "-DFCOMMON_OPT=-O2 -fno-tree-vectorize")
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-openblas
 pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
          $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas64"))
 pkgver=0.3.21
-pkgrel=1
+pkgrel=2
 pkgdesc="An optimized BLAS library based on GotoBLAS2 1.13 BSD, providing optimized blas, lapack, and cblas (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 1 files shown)


mingw-w64-openblas/PKGBUILD
```diff
@@ -59,13 +59,22 @@ _build_openblas() {
...
+    # See: https://wiki.archlinux.org/title/CMake_package_guidelines#CMake_can_automatically_override_the_default_compiler_optimization_flag
+    _build_type+=("None")
   else
     _build_type+=("Debug")
   fi
 
+  declare -a _optimization_opts
   if [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]]; then
     _c_lapack_opt="-DC_LAPACK=ON"
+  else
```
LABPBS, line 71: in this case its being built with no optimization flags for clang environments

mmuetzel: Good point!

I changed the logic to only use the build type `"None"` for non-clang (i.e., gcc subsystems). That's necessary or cmake would override the `-O2` with `-O3` for `"Release"`.

IIUC, ArchLinux builds all packages with `-O2`. Should MSYS2 be doing the same (for all subsystems incl. clang?)?

mmuetzel:
IIUC, it was an unexpected side effect that OpenBLAS is compiled with optimization level `-O3` instead of `-O2` when the build rules switched from `make` to `cmake`.
Given that, I updated the PR again to revert to building with `-O2` for all environments.

mmuetzel:
I installed the MINGW64 build artifact from this PR. With it, the example program from #12857 no longer crashes.

Issue #2673, opened 2017-07-06
Inappropriate vulkan library name


wirx6:
I noticed that libraries using dynamic loading of vulkan runtime are referring to `vulkan-1.dll` instead of `libvulkan-1.dll`.
https://github.com/glfw/glfw/blob/master/src/vulkan.c#L53
https://github.com/bkaradzic/bgfx/blob/master/src/renderer_vk.cpp#L740
https://github.com/renelindsay/Vulkan-WSIWindow/blob/master/WSIWindow/VulkanWrapper/vulkan_wrapper.cpp#L29
We have two ways of fixing this and I don't know what is right.
- strip `lib` prefix from vulkan dlls
- patch names in all libraries

Biswa96:
Would it be a possible workaround if libvulkan-1.dll just copied to vulkan-1.dll? Both DLLs will be present and program will use what it wants.

lygstate:
> Would it be a possible workaround if libvulkan-1.dll just copied to vulkan-1.dll? Both DLLs will be present and program will use what it wants.

short-turn it's OK, longturn, remove libvulkan-1.dll

Biswa96:
> remove libvulkan-1.dll

Then we have to check if any programs expects that exact DLL file. That is somewhat tedious but possible to do.

MehdiChinoune:
> Then we have to check if any programs expects that exact DLL file. That is somewhat tedious but possible to do.
Use package-grokker

MehdiChinoune:
Package-grokker list these packages:
```
mingw-w64-gst-plugins-bad
mingw-w64-magnum
mingw-w64-mesa
mingw-w64-libplacebo
mingw-w64-ffmpeg
mingw-w64-mpv
```

Biswa96:
@lygstate I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/2951500265

lygstate:
> @lygstate I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/2951500265

I think it's may still lack something, suppose we will remove `libvulkan-1.dll` in future,
what's the `libvulkan.dll.a` should be?
And what's the `vulkan.pc`?

Biswa96:
[...]

PR #12875, opened 2022-08-29
vulkan-loader: Workaround dll name by copying to vulkan-1.dll

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/2673


lygstate:
vulkan-1.dll are present in the package.

PR #12881, opened 2022-08-30
WIP: boost: fix build on msys64/mingw32

orgads:
Incomplete, since CFLAGS and CXXFLAGS still have -march=x86_64

### DIFF
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -8,7 +8,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.80.0
 _boostver=${pkgver//./_}
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


Biswa96:
What does this fix?

orgads:
Building in mingw32 environment on msys64 setup. It currently passes `address-model=64`, and the compilation fails.

Biswa96:
It does not look so from the pkgbuild file. Here $CARCH is changed to $MSYSTEM_CARCH which is same as previous.

MehdiChinoune:
Were you building boost with makepkg instead of makepkg-mingw.

MehdiChinoune:
> Building in mingw32 environment on msys64 setup. It currently passes `address-model=64`, and the compilation fails.

It builds in CI.

orgads:
> Were you building boost with makepkg instead of makepkg-mingw.

Ahhh I wasn't aware of that. Thanks :)

Biswa96:
Here is the documentation which explains the packaging procedure https://www.msys2.org/dev/new-package/

PR #12891, opened 2022-09-01
glib2: disable explicit `-mms-bitfields` in pkgconfig cflags

fauxpark:
As noted here: https://github.com/msys2/MINGW-packages/pull/12776#issuecomment-1231211803
It seems this flag is not needed as it has been the default for many versions now. As such the patch in Transmission can be rendered unnecessary.


lazka:
Thanks!

Biswa96:
@lb90 Would you like to take a look if this patch file can be added in upstream?

lb90:
Sure! :)

PR #12892, opened 2022-09-01
UnixOdbc: add CJK language support

taozuhong:


Biswa96:

mingw-w64-unixodbc/PKGBUILD
```diff
@@ -36,6 +36,8 @@ build() {
     --build=${MINGW_CHOST} \
     --host=${MINGW_CHOST} \
     --target=${MINGW_CHOST} \
+    --enable-iconv=yes \
+    --with-iconv-char-enc=GB18030 \
```
mmuetzel, line 40: Not sure if I understand correctly: Would that mean that the encoding *needs* to be GB18030? What if a user prefers a different encoding?
Is enabling iconv alone not sufficient for your use case? Does leaving the default (`auto-search`, see https://github.com/lurcher/unixODBC/blob/master/configure.ac#L215) work?

taozuhong: hardcode with GB18030 for making the default CJK language support, the lib should get locale, or fall back to the default language.

mmuetzel: I'm not sure this is the case.

If I follow the code correctly, this configure switch ends up as `#define ASCII_ENCODING "GB18030"` in `config.h`.
That means that that encoding is used unconditonally:
https://github.com/lurcher/unixODBC/blob/master/DriverManager/__info.c#L521
```C
    if ( strcmp( ASCII_ENCODING, "auto-search" ) == 0 && strcmp( connection -> unicode_string, "auto-search" ) == 0 )
[...]
    else
    {
        strcpy( ascii, ASCII_ENCODING );
        strcpy( unicode, connection -> unicode_string );
    }
```

Only if `ASCII_ENCODING` is set to its default value `"auto-search"` there is any kind of encoding detection.
In theory, the code checks for the system's locale encoding first in that case. Alas, there is currently no code that handles Windows for detecting the locale encoding:
https://github.com/lurcher/unixODBC/blob/master/DriverManager/__info.c#L510-L515
```C
#if defined( HAVE_NL_LANGINFO ) && defined(HAVE_LANGINFO_CODESET)
    /* 
     * Try with current locale settings first 
     */
    asc[ 0 ] = nl_langinfo(CODESET);
#endif
```

`nl_langinfo` is POSIX-only.

PR #12895 adds code to detect the system's locale encoding on Windows. 
Does that work for you?

taozuhong: Ah, I'm using built-in ODBC component on Windows, UnixOdbc library for *nix system;

I'll try it with my app, thank you for your great work.

mmuetzel: Thank you for offering to test. Please, let us know about your test results (good or bad) with a comment in #12895.

PR #12898, opened 2022-09-02
opencv: python bindings depends on qt5-base and vtk

marcelotduarte:
In a cleaner msys2 installation, with `mingw-w64-x86_64-python-opencv 4.6.0-3` installed.
Run:
`python -c "import cv2"`
I get:
ImportError: DLL load failed while importing cv2: N▒o foi poss▒vel encontrar o m▒dulo especificado

After installing qt-base and vtk it works!

### DIFF
--- a/mingw-w64-opencv/PKGBUILD
+++ b/mingw-w64-opencv/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-python-${_realname}")
 pkgver=4.6.0
-pkgrel=3
+pkgrel=4
 pkgdesc="Open Source Computer Vision Library (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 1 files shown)


Biswa96:
Curious. ntldd does not show the dependencies for the python module. It only depends on python besides opencv.

```
$ ntldd /ucrt64/lib/python3.10/site-packages/cv2.cp310-mingw_x86_64_ucrt.pyd | grep -v opencv
        libgcc_s_seh-1.dll => F:\msys64\ucrt64\bin\libgcc_s_seh-1.dll (0x000001e3ebbe0000)
        KERNEL32.dll => C:\Windows\SYSTEM32\KERNEL32.dll (0x000001e3ec640000)
        libstdc++-6.dll => F:\msys64\ucrt64\bin\libstdc++-6.dll (0x000001e3ec640000)
        libpython3.10.dll => F:\msys64\ucrt64\bin\libpython3.10.dll (0x000001e3ec640000)
```

Does it pull the vtk and qt5 from opencv itself? If yes those dependencies are already mentioned optional in opencv package.

marcelotduarte:
I used pepper to check the dependencies and found the same as you, but reading the PKGBUILD I verified that dependencies are not optional for python-opencv.

Biswa96:
Should those dependencies set as optional for python-opencv package also? Would removing `unset optdepends` command be the right solution?

marcelotduarte:
With ntldd you do not see any reference to numpy, but python-opencv depends on it.
The 2 questions I cannot answer. @lazka 
For me qt5-base is clearly a dependency for opencv. vtk can be turned off, but it is on.

Biswa96:
OK, let it be decided later 😊

PR #12904, opened 2022-09-02
transmission: remove patch to skip RC file for GTK client

fauxpark:
See #12891


Biswa96:
It seems that we traded one patch between glib and transmission packages 😃

fauxpark:
Maybe that one could be eliminated too if someone asks nicely on the GLib GitLab.

PR #12962, opened 2022-09-04
openblas: limit scope of `no-tree-vectorize` optimization

mmuetzel:
Use patch from upstream to limit the scope of `no-tree-vectorize` optimization to only the files that seem to be affected by the optimization issue in gcc-12 reported in #12857.

### DIFF
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-openblas
 pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
          $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas64"))
 pkgver=0.3.21
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


Biswa96:
`-O2` is set in CFLAGS and CXXFLAGS by makepkg. Is it required to mention explicitly?

mmuetzel:
> `-O2` is set in CFLAGS and CXXFLAGS by makepkg. Is it required to mention explicitly?

Oh. I wasn't aware of that.
Let me try later (today or tomorrow).

lazka:
I personally am fine with trusting upstream to select a proper optimization level for release builds.

mmuetzel:
I updated the PR to revert to the build type `Release` following @lazka's hint.
In a local test, the segmentation fault reported in https://github.com/msys2/MINGW-packages/issues/12857 is still fixed with this.

mmuetzel:
I downloaded and installed the MINGW64 build artifact of this PR. With it, the segmentation fault reported in https://github.com/msys2/MINGW-packages/issues/12857 is still fixed.

PR #12970, opened 2022-09-05
chafa: Add package

Freed-Wu:
Fix 12968


Biswa96:
I will be adding this. This package need some patching and more dependencies than that.

PR #12972, opened 2022-09-05
libheif: update patches

kmilos:
The second one is waiting to be merged upstream.


lazka:
Is the `@includedir@/@PACKAGE@` part still needed? In upstream I see things like https://github.com/strukturag/libheif/blob/1d36c10ade9b2acfc2dcc4ca8b6176a9789c902c/examples/heif_enc.cc#L41 which seems to suggest not.

kmilos:
Quite possibly not, indeed. I'll check the dependent packages just in case.

lazka:
Feel free to trigger some rebuilds here, if you want to let CI test it

kmilos:
Looks like only ImageMagick needs this still: https://github.com/ImageMagick/ImageMagick/blob/8ba9e67d34fb860fc4e02d752885bbe8a8e17070/coders/heic.c#L77-L78

So we either patch IM and remove here (could be an option, even report it to them upstream), or leave as is...

lazka:
> So we either patch IM and remove here (could be an option, even report it to them upstream), or leave as is...

patching IM sounds good to me.

> Edit: [ImageMagick/ImageMagick#5538](https://github.com/ImageMagick/ImageMagick/pull/5538)

nice, thanks

kmilos:
This is good for merge then? The current IM binaries don't care, so we can patch it for the next big update if needed.

PR #12975, opened 2022-09-05
ninja: use $TERM to decide when to use fancy output, take 2

oscarfv:
... for one-line progress messages on terminal emulators without using hacks such as winpty.

Do no activate color, because ANSI escape sequences breaks CMake.

### DIFF
--- a/mingw-w64-ninja/PKGBUILD
+++ b/mingw-w64-ninja/PKGBUILD
@@ -4,7 +4,7 @@ _realname=ninja
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.11.1
-pkgrel=1
+pkgrel=2
 pkgdesc="Ninja is a small build system with a focus on speed (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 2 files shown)


oscarfv:
Ping.

lazka:
We have conpty now enabled by default. Is this still needed?

oscarfv:
Just checked it, nice! That solves the problem for the MSYS2 console on modern enough Windows versions. If those conditions are not met, nothing changes, though.

I think the one-line progress meter is a significant usability feature, not just a cosmetic thingy. I've been using it on Emacs for the last months and it saves a lot of scrolling up/down for chasing the relevant compiler diagnostics (which are usually those that are first shown.)

OTOH I'm aware that you might have reservations about having this as a local patch. I'm not 100% happy about that situation either, but upstream did not cooperate with finding what would be an acceptable solution.

Anyways, after the conpty change I would like to rework the patch to give priority to the existence of a console over the $TERM environment variable, that would ensure that ninja works the same way as today under the MSYS2 console with conpty support.

lazka:
> Just checked it, nice! That solves the problem for the MSYS2 console on modern enough Windows versions. If those conditions are not met, nothing changes, though.

fyi, that's 97% of our active user base -> https://github.com/msys2/MSYS2-packages/discussions/2832#discussioncomment-3718506

oscarfv:
> fyi, that's ~97% of our active user base -> [msys2/MSYS2-packages#2832 (comment)](https://github.com/msys2/MSYS2-packages/discussions/2832#discussioncomment-3718506)

Yes, any problem related to out-of-date Windows will solve with time. However, that's just the part that relates to MSYS2 console. Running child processes such as Ninja from other host applications is also a reality, and ConPTY [has limitations](https://github.com/microsoft/terminal/issues/9174) that makes supporting it on some of those host applications (such as Emacs) unrealistic.

lazka:
Ah, ok, I didn't get the "inside emacs" part.

GitMensch:
@oscarfv can you please rebase?

PR #12976, opened 2022-09-06
tbb: Fix OUTPUT_NAME

MehdiChinoune:


Biswa96:
Both cmake and pkgconfig file has `-ltbb12` and this patch changes the import library name from libtbb12.dll.a to libtbb.dll.a.

MehdiChinoune:
> Both cmake and pkgconfig file has `-ltbb12` and this patch changes the import library name from libtbb12.dll.a to libtbb.dll.a.

It changes only the runtime library.

MehdiChinoune:
> Both cmake and pkgconfig file has `-ltbb12` and this patch changes the import library name from libtbb12.dll.a to libtbb.dll.a.

For pkgconfig I will take a look, but I don't see `-ltbb12` in cmake files anywhere?!

Biswa96:
> but I don't see -ltbb12 in cmake files anywhere?!

Oops, I mean TBBTargets-release.cmake file has libtbb12.dll.a. Also, can you explain the reason for this change? Should we modify the import lib name?

MehdiChinoune:
> Oops, I mean TBBTargets-release.cmake file has libtbb12.dll.a

No, It doesn't

> Also, can you explain the reason for this change? Should we modify the import lib name?

When I tried to build `openvdb` against `tbb` It failed to find `tbb` because It expects the library name to be the same as the component (tbb -> libtbb) which is true on non-WIN32.

Biswa96:
> No, It doesn't

I mean in the current package, without this change.

> When I tried to build openvdb against tbb It failed to find tbb

Shouldn't openvdb need a fix for that? I think we should not modify upstream's decision for library name. Otherwise we would end up like the vulkan loader situation.

MehdiChinoune:
> Shouldn't openvdb need a fix for that? I think we should not modify upstream's decision for library name. Otherwise we would end up like the vulkan loader situation.

For the moment that's the easiest change, If you have any other idea then go ahead.

Biswa96:
OK, let me check.

MehdiChinoune:
https://github.com/oneapi-src/oneTBB/pull/895

MehdiChinoune:
Well, I think fixing openvdb is very simple, just add `CONFIG` to `find_package`

Biswa96:
> Well, I think fixing openvdb is very simple, just add `CONFIG` to `find_package`

Awesome 😍 That would be better solution. I have tried the following change in openvdb package and cmake detects it without any error.

```diff
--- a/openvdb/openvdb/CMakeLists.txt
+++ b/openvdb/openvdb/CMakeLists.txt
@@ -127,9 +127,9 @@
   endif()
 endif()
 
-find_package(TBB ${MINIMUM_TBB_VERSION} REQUIRED COMPONENTS tbb)
+find_package(TBB ${MINIMUM_TBB_VERSION} REQUIRED COMPONENTS tbb CONFIG)
 if(OPENVDB_FUTURE_DEPRECATION AND FUTURE_MINIMUM_TBB_VERSION)
-  if(${Tbb_VERSION} VERSION_LESS FUTURE_MINIMUM_TBB_VERSION)
+  if(${TBB_VERSION} VERSION_LESS FUTURE_MINIMUM_TBB_VERSION)
     message(DEPRECATION "Support for TBB versions < ${FUTURE_MINIMUM_TBB_VERSION} "
       "is deprecated and will be removed.")
   endif()
```

PR #12978, opened 2022-09-06
wxPython: rebuild against split wxwidgets packages

lazka:
will likely fail until staging is ready

edit: failed because upstream removed the tarball... I guess we need to  update to 4.2.0


lazka:
Updating to 4.2.0 requires patch refreshes, so leaving this for later

PR #12981, opened 2022-09-07
Move from intel-tbb to tbb - Part1

MehdiChinoune:


mingw-w64-opencv/PKGBUILD
```diff
@@ -13,7 +13,7 @@ url="https://opencv.org/"
 license=("BSD")
 depends=(#"${MINGW_PACKAGE_PREFIX}-gst-plugins-base
          "${MINGW_PACKAGE_PREFIX}-ceres-solver"
-         "${MINGW_PACKAGE_PREFIX}-ffmpeg"
+         #"${MINGW_PACKAGE_PREFIX}-ffmpeg"
```
1480c1, line 16: Does this have a specific reason that could be documented alongside this comment? I'm assuming something failed to build?

MehdiChinoune: I am going to re-enable it after we complete the transition.

1480c1: Ah, I see, thanks for the clarification, cyclic dependencies are always fun (sarcasm)

PR #12982, opened 2022-09-07
vulkan-loader: Fix DLL and import library name

Biswa96:
Really Fixes https://github.com/msys2/MINGW-packages/issues/2673


Biswa96:
I do not know why ffplay.exe was not built. Can anyone help please?

mmuetzel:
Compiling `ffplay.exe` is probably disabled because of this in the config.log:
```
pkg-config --exists --print-errors sdl2 >= 2.0.1 sdl2 < 2.1.0
Package dependency requirement 'sdl2 < 2.1.0' could not be satisfied.
Package 'sdl2' has version '2.24.0', required version is '< 2.1.0'
```

https://github.com/FFmpeg/FFmpeg/commit/e5163b1d34381a3319214a902ef1df923dd2eeba might help.

Biswa96:
Thank you for the help 🙇 ffplay.exe is now built.

PR #12984, opened 2022-09-07
tbb: Fix building on clangarm64

MehdiChinoune:
@hmartinez82


mingw-w64-tbb/004-support-mingw-aarch64.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/tbbmalloc/TypeDefinitions.h
++++ b/src/tbbmalloc/TypeDefinitions.h
+@@ -25,7 +25,7 @@
+ #       define __ARCH_ipf 1
+ #   elif defined(_M_IX86)||defined(__i386__) // the latter for MinGW support
+ #       define __ARCH_x86_32 1
+-#   elif defined(_M_ARM) || defined(_M_ARM64)
++#   elif defined(_M_ARM) || defined(_M_ARM64) || defined(__aarch64__) // the latter for MinGW support
```
Biswa96, line 8: Theoretically, this patch should not be required. mingw-w64 defines `_M_ARM64` if `__aarch64__` is defined. See https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/_mingw_mac.h

jeremyd2019: Probably hasn't included anything that ultimately includes _mingw_mac.h at that point.  I've run into that once or twice when checking for `_UCRT`

MehdiChinoune: Then, why does it show this https://github.com/oneapi-src/oneTBB/blob/v2021.5.0/src/tbbmalloc/TypeDefinitions.h#L31

Biswa96: OK, Jeremy is right. For example, the following code fails in x86_64

```c
#ifdef _M_X64
#warning "right"
#else
#error "wrong"
#endif

#include <stdio.h>

int main(){}
```

But if `stdio.h` is included at first things are fine.

lazka:
should we just try in autobuild?

MehdiChinoune:
> should we just try in autobuild?

Maybe1

Issue #12967, opened 2022-09-04
Add [quassel-core]


SonaliBendre:
### Package name

quassel-core

### Brief description of package

Next-generation distributed IRC client - core only

### URL for package's homepage

https://archlinux.org/packages/community/x86_64/quassel-core/

### MINGW environments where you need the package

- [X] MINGW64
- [X] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Biswa96:
It is already available, see https://packages.msys2.org/base/mingw-w64-quassel

SonaliBendre:
> It is already available, see https://packages.msys2.org/base/mingw-w64-quassel

No dude, thats quassel-mono+client+core. I Want quassel-core only so need to build with this flags 
```
-DCMAKE_BUILD_TYPE="Release" \
	-DCMAKE_INSTALL_PREFIX=/usr \
	-DWANT_CORE=ON \
	-DWANT_MONO=OFF \
	-DWANT_QTCLIENT=OFF \
	-DWITH_KDE=OFF
```

Biswa96:
Understood. You want split packages. Let me see how it goes.

MehdiChinoune:
> No dude, thats quassel-mono+client+core. I Want quassel-core only so need to build with this flags

Could elaborate why? Do you encounter any issue when using the whole `quassel` package or you just want a package with less size.

SonaliBendre:
> > No dude, thats quassel-mono+client+core. I Want quassel-core only so need to build with this flags
> 
> Could elaborate why? Do you encounter any issue when using the whole `quassel` package or you just want a package with less size.

I want a package with less size. I don't need the gui client because I use android app. I only need quassel-core which is cli only

SonaliBendre:
Can you package it also? https://aur.archlinux.org/packages/smuxi-server-git
Only smuxi-server not smuxi-gnome-frontend

MehdiChinoune:
> > > No dude, thats quassel-mono+client+core. I Want quassel-core only so need to build with this flags
> > 
> > 
> > Could elaborate why? Do you encounter any issue when using the whole `quassel` package or you just want a package with less size.
> 
> I want a package with less size. I don't need the gui client because I use android app. I only need quassel-core which is cli only

I don't think this is a valid reason to create a new package.
We all need some packages with less size. Personally I asked before to minimize packages size by disabling static libraries or split them.
[...]

PR #12990, opened 2022-09-08
quassel: Add separate quassel-core package

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/12967

### DIFF
--- a/mingw-w64-quassel/PKGBUILD
+++ b/mingw-w64-quassel/PKGBUILD
@@ -7,21 +7,22 @@ _realname=quassel
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=0.14.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Next-generation distributed IRC client (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url="https://quassel-irc.org/"
-license=('GPL')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
I don't see any new separate package?

MehdiChinoune:
Sorry, Isn't quassel-core contained in quassel?.
What's the problem this PR is trying to solve?. Is there an issue with `quassel`?
If we follow user requests we will end up like debian/redhat packaging system (runtime + plugins + development + debug + docs).
I was expecting two split packages where `quassel` depends on `quassel-core`. This approach just wastes buildtime for no valid reason.

Biswa96:
> Isn't quassel-core contained in quassel?

Yes, that's what I first thought of. But then I follow what ArchLinux does. I think this approach is more flexible for adding any further configuration (like Arch does) in future.

mingw-w64-quassel/PKGBUILD
```diff
@@ -36,9 +38,12 @@ prepare() {
 }
 
 build() {
-  cd "${srcdir}"
-  [[ -d build-${CARCH} ]] && rm -rf build-${CARCH}
-  mkdir build-${CARCH} && cd "${srcdir}"/build-${CARCH}
```
mmuetzel, line 41: Would it make sense to continue making sure that the build directory is clean?

Biswa96: See https://github.com/msys2/MINGW-packages/commit/10700daef1e47f6f842b4e987c0f2baebd59e6af

PR #12991, opened 2022-09-08
lokalize: Update to 22.08.1

podsvirov:
```shell
    lokalize-22.08.1.tar.xz ... FAILED (unknown public key BB463350D6EF31EF)
```
@Biswa96 can you solve this?


podsvirov:
Thanks!

PR #12993, opened 2022-09-08
ispc 1.18.0 + update openimagedenoise to 1.4.3

MehdiChinoune:


Biswa96:
I would suggest to report the issue in upstream, not sure if it is ispc or odin.

PR #13089, opened 2022-09-13
Test CI configuration

lb90:


Biswa96:
You can test CI in your own fork. Just enable GitHub Actions in the repository settings.

lb90:
Ah, great! Then I'm going to clode this PR. Sorry for the noise! :)

jeremyd2019:
A trick I picked up a while back is that you can use [tmate](https://github.com/mxschmitt/action-tmate) to poke things interactively, if it becomes too onerous to keep committing single things to try.  Example: https://github.com/jeremyd2019/tmate-test/blob/main/.github/workflows/main.yml

PR #13095, opened 2022-09-14
Imagemagick: Enable clangarm64

driver1998:
Also fixes ghostscript on clangarm64.


Biswa96:
> Also fixes ghostscript on clangarm64.

Is that patch from upstream?

driver1998:
> > Also fixes ghostscript on clangarm64.
> 
> Is that patch from upstream?

No.

PR #13103, opened 2022-09-15
igraph: remove unneeded dependencies

szhorvat:
This PR removed some unneeded dependencies:

  - `suitesparse` is not a dependency since igraph 0.10
  - I am not sure why `zlib` was there as it is only an indirect dependency through libxml2
  - Python is only needed for building the docs, which is not done here

I would also like to suggest removing the `-DIGRAPH_WARNINGS_AS_ERRORS=OFF` and instead reporting any failures to upstream.


lazka:
thanks, please bump pkgrel

> I would also like to suggest removing the `-DIGRAPH_WARNINGS_AS_ERRORS=OFF` and instead reporting any failures to upstream.

This is usually only done in upstream CI systems. New compiler versions will result in new warnings which would break our build.

szhorvat:
> thanks, please bump pkgrel

I'm sorry, I'm not sure what that means. Could you perhaps take it from here? I have not contributed to msys2 before, and in fact this was meant to be an issue, but then it seemed easier to just show the changes through a PR. Note that these changes do not affect the binaries that are produced.

szhorvat:
Thank you!

PR #13107, opened 2022-09-16
Drop qt5 meta-package

MehdiChinoune:
It has no user now.


lazka:
There might be some fallout of users installing it in CI etc... but I can write a NEWS entry about it.

MehdiChinoune:
`pacman -S ${MINGW_PACKAGE_PREFIX}-qt5` will install qt5 group for them which contains all the Qt5 packages. the same as `toolchain`.

lazka:
ah, I see, thanks

MehdiChinoune:
https://github.com/search?q=mingw-w64-x86_64-qt5&type=code

PR #13109, opened 2022-09-16
llvm: fix system libs for llvm-config

### LINKED ISSUE #13108: llvm-config returns build time path for libzstd
```console
$ llvm-config.exe --link-static --system-libs
-lpsapi -lshell32 -lole32 -luuid -ladvapi32 -lpthread -lz D:/a/msys64/ucrt64/bin/libzstd.dll -lxml2
```

This breaks meson tests: https://github.com/mesonbuild/meson/pull/10816#issuecomment-1248725385

With 14.0.6 the output was:

```console
$ llvm-config.exe --link-static --system-libs
-lpsapi -lshell32 -lole32 -luuid -ladvapi32 -lpthread -lz -lxml2
```

The source of that value is the `LLVM_SYSTEM_LIBS` macro

lazka:
For some reason it added an absolute path there instead of the library. Fix it with my limited cmake knowledge..

Fixes #13108


lazka:
I'm wondering if this is also broken on Linux, the missing get_library_name() seems like an oversight at least.

PR #13120, opened 2022-09-18
Update libpanel to 1.0.0

taozuhong:


Biswa96:
Let me update this in a separate commit. Some packaging rules were not followed.

PR #13127, opened 2022-09-18
CI: Run prepare step first.

MehdiChinoune:
This will make CI fails earlier if patches couldn't be applied or checksums are wrong. It helps with automatic updates


MehdiChinoune:
This won't work if `prepare` requires extra-tools like `git`.

PR #13176, opened 2022-09-19
python-matplotlib: update to 3.6.0

MehdiChinoune:


MehdiChinoune:
`labelme` requires `matplotlib<3.3` because of `PyInstaller`, I don't think It is something could block this PR.

PR #13184, opened 2022-09-19
[new-package] highlight 4.3

legends2k:
Didn't use the [latest (stable) version 4.2](https://gitlab.com/saalen/highlight/-/releases)'s `.zip`; went for commit SHA instead since no releases were made since [my upstream MinGW patches](https://gitlab.com/saalen/highlight/-/merge_requests/136).

### DIFF
--- a/mingw-w64-highlight/PKGBUILD
+++ b/mingw-w64-highlight/PKGBUILD
@@ -1,0 +1,25 @@
+_realname=highlight
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=4.2
+pkgrel=1
+url="http://www.andre-simon.de/doku/highlight/highlight.html"
+pkgdesc="Fast and flexible source code highlighter (CLI version)"
+license=('GPL')
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+makedepends=("${MINGW_PACKAGE_PREFIX}-lua")
+_commit='80c5735e1f2a525c4f85e5f220c6b3c2bb26c063'
+source=("https://gitlab.com/saalen/highlight/-/archive/${_commit}/highlight-${_commit}.tar.gz")
+sha256sums=('6f9ef2fe18f8bbf97c9a5cdff5b3b061f017d7ce00cb9bb34adf0b9b6c0df6fa')
+options=('strip')
+
+build() {
+  cd "$srcdir/${_realname}-${_commit}/"
+  make
+}
+
+package() {
+  cd "$srcdir/${_realname}-${_commit}/"
+  DESTDIR="${pkgdir}" make install
+}
--- a/mingw-w64-highlight/PKGBUILD
+++ b/mingw-w64-highlight/PKGBUILD
@@ -13,13 +13,16 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-boost")
 source=("https://gitlab.com/saalen/highlight/-/archive/v${pkgver}/highlight-v${pkgver}.tar.bz2"
-        "001-fix-installation-prefix-and-destdir.patch")
+        "001-fix-installation-prefix-and-destdir.patch"
+        "002-fix-data-dir-stat.patch")
 sha256sums=('3377562676ce1aadd97648aa0c4e4e85d26db190a090df49587ae8d2471ca851'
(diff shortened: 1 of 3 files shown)


mmuetzel:
The clang environments are failing to build this package. E.g., for CLANG64:
```
  ./core/lspclient.cpp:364:55: error: use of undeclared identifier 'getpid'
          params["processId"] =  picojson::value((float)getpid());
                                                        ^
  1 error generated.
  make[1]: *** [makefile:198: lspclient.o] Error 1
```
Maybe it helps to `#include <process.h>` in that file? Maybe guarded with `#ifdef WIN32` (like it is used in the remainder of that file)?

legends2k:
> Maybe it helps to #include <process.h> in that file? Maybe guarded with #ifdef WIN32 (like it is used in the remainder of that file)?

Thanks for the suggestion; it worked.  Not sure why GCC alllowed it while Clang didn't; perhaps Clang is stricter; or perhaps GCC's headers internally include `unistd.h` or `process.h`.

I think the more apt header is `unistd.h`.  I've raised an upstream pull request.  Once it's in will update this one.

mingw-w64-highlight/PKGBUILD
```diff
@@ -0,0 +1,27 @@
...
+url="http://www.andre-simon.de/doku/highlight/highlight.html"
+pkgdesc="Fast and flexible source code highlighter (CLI version)"
+license=('GPL')
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-lua"
+             "${MINGW_PACKAGE_PREFIX}-boost")
+_commit='80c5735e1f2a525c4f85e5f220c6b3c2bb26c063'
+source=("https://gitlab.com/saalen/highlight/-/archive/${_commit}/highlight-${_commit}.tar.gz")
```
Biswa96, line 15: As this is a tagged release, I think a specific tarball can be used. Please test the following

```suggestion
source=("https://gitlab.com/saalen/highlight/-/archive/v${pkgver}/highlight-v${pkgver}.tar.bz2")
```

mmuetzel:
> I think the more apt header is `unistd.h`. I've raised an [upstream pull request](https://gitlab.com/saalen/highlight/-/merge_requests/138). Once it's in will update this one.

Glad it worked for you.

Afaict, the `unistd.h` header is an extension of MinGW. It isn't available on MSVC. (Please, correct me if that is wrong.)
`#include <process.h>` should make that function declaration available on MinGW and MSVC. (I don't know if upstream cares for MSVC though...)

mingw-w64-highlight/PKGBUILD
legends2k: @Biswa96 Tagged release doesn't build on MSYS2.  Upstream hasn't made a release yet since my upstream fixes.  Another option to still go with a released version (as opposed to a commit id) would be have the same fixes a patches downstream.  Do you recommend that?

Biswa96: You can use the tagged source tarball and patch the source code with the upstream or your patches.

legends2k:
@Biswa96 Done!  I've used version 4.3 which contains all my upstream fixes :)

mingw-w64-highlight/PKGBUILD
legends2k: @Biswa96 Done!  I've used version 4.3 which contains all my upstream fixes :)

mingw-w64-highlight/PKGBUILD
```diff
@@ -0,0 +1,26 @@
...
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-lua"
+             "${MINGW_PACKAGE_PREFIX}-boost")
+source=("https://gitlab.com/saalen/highlight/-/archive/v${pkgver}/highlight-v${pkgver}.tar.bz2")
+sha256sums=('3377562676ce1aadd97648aa0c4e4e85d26db190a090df49587ae8d2471ca851')
+options=('strip')
+
+build() {
+  cd "$srcdir/${_realname}-v${pkgver}/"
+  make
```
MehdiChinoune, line 20: No file was installed!
```
  [MSYS2 CI] mingw-w64-highlight: File listing diff for mingw-w64-x86_64-highlight
  
  warning: database file for 'ci' does not exist (use '-Fy' to download)
```
I think you should change it to:
```suggestion
  make PREFIX=${MINGW_PREFIX}
```

legends2k:
@MehdiChinoune I missed to setting `PREFIX` and `DESTDIR` earlier, thanks for fixing it!

There were a couple of more issues even after all the above fixes.  They're fixed now.  I've tested them.

@Biswa96 What else do I need for this to be merged?

PR #13194, opened 2022-09-20
python-poetry: update to 1.2.1

MehdiChinoune:


MehdiChinoune:
`petry-plugin-export` is commented in `poetry` PKGBUILD because of the cycle-dependency. I will uncomment it after this PR got merged.

PR #13196, opened 2022-09-20
python-cryptography: update to 38.0.1

MehdiChinoune:
Fails to build on CLANG64, a similar issue occured before with GNU ld https://github.com/retep998/winapi-rs/issues/539, but I don't know/understand how It was solved.
I have encountered the same issue when updating starship.

@mati865

### DIFF
--- a/mingw-w64-python-cryptography/001-disable-abi3.patch
+++ b/mingw-w64-python-cryptography/001-disable-abi3.patch
@@ -1,7 +1,7 @@
 --- a/setup.py
 +++ b/setup.py
-@@ -45,12 +45,6 @@
-                 "_rust",
+@@ -48,12 +48,6 @@
+                 "cryptography.hazmat.bindings._rust",
                  "src/rust/Cargo.toml",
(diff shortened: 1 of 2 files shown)


mati865:
Try defining `WINAPI_NO_BUNDLED_LIBRARIES=1`.

MehdiChinoune:
> Try defining `WINAPI_NO_BUNDLED_LIBRARIES=1`.

only on CLANG64 or everywhere?

MehdiChinoune:
It works!

mati865:
> > Try defining `WINAPI_NO_BUNDLED_LIBRARIES=1`.
> 
> only on CLANG64 or everywhere?

Its required only when linking with LLD but should never hurt since winapi provides very old libraries.

I have added new triples for LLVM based subystems: https://doc.rust-lang.org/rustc/platform-support/pc-windows-gnullvm.html but we need to wait a bit until update for `windows-rs` crate gets propagated across the ecosystem or some crates will fail to build on CLANG64.

PR #13200, opened 2022-09-21
imagemagick: enable JPEG XL

kmilos:


kmilos:
Looks like GM requires (unreleased) libjxl 0.7.0...

Biswa96:
Should not libjxl mentioned as optional dependencies? The new DLL file in the package seems to be a plugin.

kmilos:
You're probably right - but that goes for most of the other delegates as well then?

Anyway, let's hold off on this until #13204 is done...

kmilos:
Dropped in favour of #13206

Issue #13204, opened 2022-09-21
[libjxl] update to 0.7.0


novomesk:
### Package name

libjxl

### New version number

0.7.0

### Other information that may be useful

Hello,

this is the new version we waited for a long time:
https://github.com/libjxl/libjxl/releases/tag/v0.7.0

This version works with system installed `highway` 1.0.1 package. You can remove the conflict and use `-DJPEGXL_FORCE_SYSTEM_HWY=ON` configuration.

Do you remember, that we had to use the workaround with `-DHWY_COMPILE_ONLY_SCALAR` to avoid crashes in AVX2 generated code? It worked but there was performance penalty.
I tried to build libjxl on MSYS2 without this workaround and I was unable to reproduce the crashes. I believe it could be because of newer gcc in MSYS2 and/or improvements in the highway library.

There is a GIMP2 plugin in libjxl repo. Meanwhile, we have also limited JXL plug-in also in [GIMP 2.10.32](https://www.gimp.org/news/2022/06/14/gimp-2-10-32-released/#jpeg-xl) (not only in GIMP 2.99.x like before).

When GIMP2 is a dependency of libjxl, GIMP plugin from libjxl repo is built.

When libjxl is a dependency of GIMP, JXL plug-in in official GIMP is built.

In the future, the GIMP plug-in will be probably removed from libjxl repo and further development will be done on GIMP side only.

When you want to disable the plug-in in libjxl repo, use `-DJPEGXL_ENABLE_PLUGIN_GIMP210=OFF`

You can also use `-DJPEGXL_ENABLE_PLUGIN_MIME=OFF` because `image/jxl` definition is already added to shared-mime-info in MSYS2.

There is some work with reviewing those old patches, some of them are maybe no longer needed.

### Are you willing to submit a PR?

Perhaps @kmilos as he is more skilled in PKGBUILD area

kmilos:
Thanks for the update @novomesk, about time. ;)

I guess it could be a while before all the MSYS2 dependant packages catch up w/ the new API, but I'll try to tweak the build script locally in the meantime...

novomesk:
Old 0.6.x API can be still used with 0.7.0 it just triggers deprecated warnings during compilation. I think there won't be big troubles.

kmilos:
Can you please upstream this first one?

```diff
--- a/plugins/gdk-pixbuf/CMakeLists.txt
+++ b/plugins/gdk-pixbuf/CMakeLists.txt
[...]

PR #13206, opened 2022-09-21
libjxl: update to 0.7.0

kmilos:
Resolves https://github.com/msys2/MINGW-packages/issues/13204

Deps bump to follow once this is reviewed.


mingw-w64-libjxl/PKGBUILD
```diff
@@ -87,25 +58,26 @@ build() {
...
     "${_extra_config[@]}" \
     -DJPEGXL_ENABLE_PLUGINS=ON \
+    -DJPEGXL_ENABLE_PLUGIN_GIMP210=OFF \
+    -DJPEGXL_ENABLE_PLUGIN_MIME=OFF \
     -DBUILD_TESTING=OFF \
     -DJPEGXL_WARNINGS_AS_ERRORS=OFF \
     -DJPEGXL_ENABLE_BENCHMARK=OFF \
     -DJPEGXL_ENABLE_EXAMPLES=OFF \
     -DJPEGXL_ENABLE_MANPAGES=OFF \
+    -DJPEGXL_ENABLE_SKCMS=OFF \
```
novomesk, line 74: `pixbufloader-jxl.c` uses SKCMS internally so we will not avoid building it anyway.

What about using `-DJPEGXL_ENABLE_SKCMS=ON` so we don't have mixture of lcms2/skcms. What do you think?

kmilos: I prefer the plugin implemented the choice so we can go all lcms2 ;)

novomesk:
I think `-DJPEGXL_ENABLE_SJPEG=OFF` helps to avoid sjpeg completely.

mingw-w64-libjxl/PKGBUILD
novomesk: For me doesn't matter too much when both choices work correctly.

kmilos:
I'm also 50/50 about (not) shipping the MIME info... Although present in `shared-mime-info`, I guess there are non-GTK apps that could still benefit from this? We ship it in `libavif` and `libheif`...

novomesk:
Without shared-mime-info it doesn't make sense. Duplicate definition is not necessary.

kmilos:
> Without shared-mime-info it doesn't make sense. Duplicate definition is not necessary.

TODO then to remove from `libavif` and `libheif`...

novomesk:
I know libheif adds their own XML files but libavif doesn't.

kmilos:
I also tried enabling the man pages, but still no dice... Reported upstream though: https://github.com/libjxl/libjxl/issues/1772

novomesk:
Until resolved, we can continue to live without manpages.

Biswa96:
Wait, found an workaround. Would you like to try this patch?

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -442,8 +442,8 @@ if (ASCIIDOC_PY_FOUND)
     # does not recognize it.
     add_custom_command(
       OUTPUT "${PAGE}.1"
-      COMMAND "${ASCIIDOC_PY}"
-      ARGS ${ASCIIDOC}
+      COMMAND ${ASCIIDOC}
+      ARGS
         --format manpage --destination-dir="${CMAKE_CURRENT_BINARY_DIR}"
         "${CMAKE_CURRENT_SOURCE_DIR}/doc/man/${PAGE}.txt"
       MAIN_DEPENDENCY "${CMAKE_CURRENT_SOURCE_DIR}/doc/man/${PAGE}.txt")
```

Make sure to enable `JPEGXL_ENABLE_MANPAGES` option in cmake. All credit goes to https://stackoverflow.com/a/59635064

kmilos:
> Wait, found an workaround.

Yeah, I tried something similar (equivalent setting on line 410), and reported in the upstream issue, thanks.

kmilos:
Next, this new `cjxl_hdr` looks pretty large, and seems statically linked, don't know why...

novomesk:
When I tried older snapshot I got 93.7 MB for cjpeg_hdr.exe

On my Gentoo Linux I have only:
`-rwxr-xr-x 1 root root 2724016 sep 14 19:08 /usr/bin/cjpeg_hdr`

kmilos:
Yeah, it's around 3 MB here as well... (about the same as the libjxl library itself)

I don't see any linking in tools/CMakeLists.txt, it really does look like it is source only.

So I think this is good to go now unless there are other objections.

The grokker didn't complain, does it really mean no bumping of deps is necessary?

novomesk:
I wish that next-gen `gimp3` and `kimageformats-qt5` are rebuilt.

And that libjxl is added as dependency for old 2.10.32 `gimp`

kmilos:
Any other feedback? Shall I bump the dependants next?

Biswa96:
Would you like to rebase the changes to latest commit of this repository? The kimageformats-qt5 looks different, no clangarm64.

kmilos:
Great... I guess Clang 15 is now more strict about even more stuff:

```
 ../../gimp-2.10.32/libgimp/gimpprogressbar.c:192:12: error: incompatible pointer to integer conversion returning 'HGDIOBJ' (aka 'void *') from a function with result type 'guint32' (aka 'unsigned int') [-Wint-conversion]
      return GDK_WINDOW_HWND (gtk_widget_get_window (GTK_WIDGET (window)));
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

@novomesk Can you please provide a patch for Gimp 2.10.x?

Biswa96:
Workaround `CFLAGS+=" -Wno-int-conversion" CXXFLAGS+=" -Wno-int-conversion"`

kmilos:
Looks like it's the same for Gimp 2.99.x 😞

kmilos:
At least the Gimp build seems to be passing now... That'll be fun for the arm runner 😮

MehdiChinoune:
Shouldn't this PR only for updating jxl, Enabling jxl should be done separately in other PRs.

kmilos:
I'll split this indeed, the build times are very long...

MehdiChinoune:
It looks like python-imagecodecs should be patched to build against blosc2>=2.3.0 https://github.com/msys2/MINGW-packages/actions/runs/2924965192/jobs/4665294829

MehdiChinoune:
If someone familiar with cython could add `#define BLOSC1_COMPAT 1` to https://github.com/cgohlke/imagecodecs/blob/master/imagecodecs/_blosc2.pyx

kmilos:
I'll probably just turn off that error temporarily as well (a similar one fails GraphicsMagick too).

Let's see how upstream patches iz: https://github.com/cgohlke/imagecodecs/issues/47

MehdiChinoune:
I am going to patch blosc2 instead

kmilos:
Let's be pragmatic and just turn the error of until the upstream fix.

Btw, look like a lot of projects are being bitten by these two issues since Clang 15 came out:

https://discourse.llvm.org/t/configure-script-breakage-with-the-new-werror-implicit-function-declaration/65213/3

kmilos:
It further looks like these were downgraded to warnings in 15.0.1 and I rebased this after @lazka updated it, but `python-imagecodecs` must use `-Wall` somewhere...

MehdiChinoune:
building python-imagecodecs should be fixed now.

kmilos:
Thanks!

kmilos:
We might have to disable manpages on arm64, a2x/xmllint fails there for some reason...

Biswa96:
libjxl build was failed due to error in path relocation of libxml2.

kmilos:
Great, I saw the libxml2 update and the successful build now, thanks all!

PR #13260, opened 2022-09-25
libxml2: Fix relocation

MehdiChinoune:


MehdiChinoune:
libjxl could be built now, so I assume It's OK.

mingw-w64-libxml2/0031-apply-msvc-relocation.patch
```diff
@@ -0,0 +1,55 @@
...
+ 	catalogs = (const char *) getenv("XML_CATALOG_FILES");
+ 	if (catalogs == NULL)
+-#if defined(_WIN32) && defined(_MSC_VER)
++#if defined(_WIN32)
+     {
+ 		void* hmodule;
++#ifdef _MSC_VER
+ 		hmodule = GetModuleHandleA("libxml2.dll");
++#else
++		hmodule = GetModuleHandleA("libxml2-LIBXML_MAJOR_VERSION.dll");
```
jeremyd2019, line 23: This works?  I would have expected something like
```c
"libxml2-" LIBXML_MAJOR_VERSION ".dll"
```

or may need some macro magic to stringify LIBXML_MAJOR_VERSION.

MehdiChinoune: You are right about stringifying as I have encountered some issues with them before, but I am not sure if `LIBXML_MAJOR_VERSION` get replaced by `2` or not.

lazka: I've hardcoded it for now, see 4b5373a3b542a77a00

jeremyd2019: It's been a while, but I seem to remember there's a `GetModuleHandle` variant that will get the `HMODULE` for a function pointer.  It wasn't always available, but I think it was XP and up...  Yeah, https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandleexa with `GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS` (and probably `GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT` for the same refcount behavior as `GetModuleHandleA`)

Issue #8399, opened 2021-04-17
Hangs with OpenBLAS from 32-bit MinGW


szhorvat:
We see consistent hangs when using `mingw-w64-i686-openblas` and running the test suite of [igraph](https://github.com/igraph/igraph/) on GitHub Actions.  See here for more details:

https://groups.google.com/g/openblas-users/c/iRRfgOw8pKk/m/KfSs3njWAQAJ

I am reporting this here as well because it is possible that something is wrong with the 32-bit OpenBLAS that is included in MSYS2. For example, I notice that MSYS2's OpenBLAS is not built using USE_OPENMP=1, while all Linux distros I checked use this option. It is likely that at least some dependencies of OpenBLAS in MSYS2 are using OpenMP. While I have no evidence that the hang we see is related to this option, I thought this was useful to note. See also: https://github.com/xianyi/OpenBLAS/wiki/faq#multi-threaded

Currently, the reason why [igraph in MSYS2 MinGW](https://packages.msys2.org/package/mingw-w64-x86_64-igraph) uses vendored dependencies (which limit performance) is these hangs. (Ping @jannick0 who contributed igraph to MSYS2)

lazka:
I have no idea, but I've created #8401

lazka:
It built fine at least. We still need someone to say that this is an OK change to make though.

jannick0:
> See also: https://github.com/xianyi/OpenBLAS/wiki/faq#multi-threaded

@szhorvat Wouldn't it be more efficient for universal use of the library to apply `openblas_set_num_threads(1)` at `igraph` runtime instead of hard-coding restrictive configs of the library build time here? 

> Currently, the reason why [igraph in MSYS2 MinGW](https://packages.msys2.org/package/mingw-w64-x86_64-igraph) uses vendored dependencies (which limit performance) is these hangs. (Ping @jannick0 who contributed igraph to MSYS2)

@szhorvat Well, not exactly. The reason for using internal libraries is to reduce dependencies which would potentially require `igraph` re-builds. Up and until now there was no complaint about performance issues, but I am open, of course. [Off-topic: The more pressing issue is to have a simple and stable procedure in place to build shared and static igraph libraries (with consistent header files working for both the shared and static library.)]

szhorvat:
> @szhorvat Wouldn't it be more efficient for universal use of the library to apply `openblas_set_num_threads(1)` at `igraph` runtime instead of hard-coding restrictive configs of the library build time here?

[...]

PR #13267, opened 2022-09-26
igraph: disable OpenMP support

szhorvat:
This PR disables OpenMP support in igraph.

igraph in MSYS2 is currently set up to link to OpenBLAS, both directly and indirectly (through ARPACK). OpenBLAS conflicts with OpenMP, unless it is built specifically to be compatible. Otherwise there is a risk of a hang. See #8399 for details.

To avoid this issue in igraph, we have two choices:

 1. Disable OpenMP
 2. Avoid linking to OpenBLAS (use bundled BLAS or reference BLAS). Since ARPACK in MSYS2 links to OpenBLAS, this means that we must also avoid ARPACK, and use the ARPACK that is bundled with igraph. This is an old version that is less reliable.

The first choice has a smaller impact (fewer igraph functions affected, smaller performance impact, smaller reliability impact), so this is what I recommend.

----

If #8399 is addressed at some point in the future, this change can be reverted. The change I proposed in #8399 has wide impact, so I expect that it will take some time until a decision is reached about it. Until then, it would be nice to have a reliable igraph in MSYS2 that does not hang.

### DIFF
--- a/mingw-w64-igraph/PKGBUILD
+++ b/mingw-w64-igraph/PKGBUILD
@@ -4,7 +4,7 @@ _realname=igraph
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=0.10.1
-pkgrel=2
+pkgrel=3
 pkgdesc="Library for the analysis of networks (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
@@ -19,8 +19,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-glpk"
          "${MINGW_PACKAGE_PREFIX}-libxml2"
          "${MINGW_PACKAGE_PREFIX}-openblas"
          $([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] && echo \
-           "${MINGW_PACKAGE_PREFIX}-arpack" || echo \
-           "${MINGW_PACKAGE_PREFIX}-openmp"))
+           "${MINGW_PACKAGE_PREFIX}-arpack"))
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja")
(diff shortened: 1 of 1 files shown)


szhorvat:
It is the nature of race conditions that the bug may or may not manifest itself depending on circumstances outside of one's control. For example, I cannot reproduce the hang with MinGW32 locally, but I see it consistently on GitHub Actions. Just because I have not yet observed it on 64-bit does not mean that there is no risk of hang.

As stated in #8399, the OpenBLAS docs are clear unless OpenBLAS was compiled with `USE_OPENMP=1` (which it isn't in MSYS2), it should not be used together with OpenMP. https://github.com/xianyi/OpenBLAS/wiki/faq#multi-threaded

mmuetzel:
I'm not sure what the impact is on performance. But would this (from the FAQ you linked) also work?
> Call `openblas_set_num_threads(1)` in the application on runtime.

szhorvat:
> But would this (from the FAQ you linked) also work?

If you are suggesting to patch igraph, that is not possible. This is an OpenBLAS-specific call. igraph does not know what BLAS implementation it is using. Some systems (like Fedora) even support dynamically switchable BLAS.

Also, just to avoid misunderstandings: igraph is not incompatible with OpenBLAS. It is OpenMP that is incompatible with a specific configuration of OpenBLAS, which MSYS2 happens to be using. I _assume_ that this is why some Linux distros configure OpenBLAS to be compatible with OpenMP ([Arch](https://github.com/archlinux/svntogit-community/blob/packages/openblas/trunk/PKGBUILD#L23)) and some others provide multiple variants ([Debian](https://packages.debian.org/source/sid/openblas)).

> I'm not sure what the impact is on performance.

I am a member of the igraph team. In my opinion, from a performance consideration, disabling OpenMP for igraph is much better than disabling multithreading in OpenBLAS.

----

In the end, my goal with this PR is to ensure that MSYS2 users get a reliable build of igraph.

mati865:
We also could give OpenMP another shot and keep it only for subsystems where is does work but I have no strong opinion here.

MehdiChinoune:
You know the rule `innocent until proven guilty`.

MehdiChinoune:
I see you re-enabled OpenMP in https://github.com/igraph/igraph/commit/418d61432d68a627dccb357983a82f39a1186ce4
Is it fixed now?

lazka:
See #13708 for giving openmp a try

lazka:
Closing this for #13708 for now. Let's see...

szhorvat:
OpenMP support has already been disabled for 32-bit. Now that the problem is solved, I suggest re-enabling it.

https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-igraph/PKGBUILD#L41

lazka:
Sounds good

szhorvat:
@lazka Will you do it or do you need a PR from me?

lazka:
I can do it

PR #13272, opened 2022-09-26
libuv: update to 1.44.2

MehdiChinoune:
The commit couldn't be reversed anymore.

~Moving to CMake buildsystem requires a bootstrap pf CMake as libuv is one of its dependencies.~ (Fixed)


Biswa96:
Is it required to use cmake to build this package? The bootstrap conflicts can be skipped with autotools, right?

MehdiChinoune:
> Is it required to use cmake to build this package?

No, It could be built by both CMake and autotools.
> The bootstrap conflicts can be skipped with autotools, right?

autotools with CMake or libuv?
I have fixed the library name to avoid rebuilding CMake.

Biswa96:
I mean, let build libuv with autotools. In that way, we can avoid the dependency cycle.

MehdiChinoune:
> I mean, let build libuv with autotools. In that way, we can avoid the dependency cycle.

The CMake build is faster and provide cmake-config files.

Biswa96:
OK, understood.

MehdiChinoune:
Is there any objection?

PR #13284, opened 2022-09-29
flang: update to 15.0.1

jeremyd2019:
Looks like CLANG64 still fails in CI, so will probably require manual building.  Might need to wait for @lazka before merging to avoid breakage in the queue.


jeremyd2019:
If it might actually be useful to someone, I should be able to build flang manually for clangarm64 on the raspberry pi (since that has more RAM).  If not I won't bother

Issue #3444, opened 2018-02-25
mingw-w64-imagemagick: configure.xml file not located in the correct place


paulalanthompson:
After installing mingw-w64-imagemagick with no errors, a run-time error was encountered:

convert.exe: unable to find configure file 'configure.xml' etc etc

The file was found in 
/mingw64/lib/ImageMagick-7.0.6/config-Q16HDRI

After copying the file to 
/mingw64/etc/ImageMagick-7
the error went away, and specific scripts which use IM now run correctly. 

This is an misconfiguration issue in the install utility.

jho101:
see also here: https://github.com/Alexpux/MINGW-packages/issues/2995

jeroen:
Configure with `--enable-zero-configuration --without-modules` to prevent runtime dependencies on external xml or dll files.

PR #13287, opened 2022-09-30
imagemagick: Fix path relocation

Biswa96:
* Use unix-like directory separator to prevent escaping character.
* Relocate paths with respect to DLL instead of calling executable. This helps to run executable from any directory.

Fixes https://github.com/msys2/MINGW-packages/issues/3444 @kmilos

### DIFF
--- a/mingw-w64-imagemagick/001-7.0.4.1-relocate.patch
+++ b/mingw-w64-imagemagick/001-7.0.4.1-relocate.patch
@@ -48,25 +48,25 @@ diff -aur ImageMagick-7.0.1-3/MagickCore/configure.c.orig ImageMagick-7.0.1-3/Ma
  #if defined(MAGICKCORE_INSTALLED_SUPPORT)
  #if defined(MAGICKCORE_SHARE_PATH)
 -  (void) AppendValueToLinkedList(paths,ConstantString(MAGICKCORE_SHARE_PATH));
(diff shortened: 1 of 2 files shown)


Biswa96:
@kmilos Would you like to test the packages please?

kmilos:
Thanks for the patched package, but still a no-go unfortunately:

1. `magick -list format` is still empty, although `magick -list configure` does now find the config file
2. I can no longer build dt, now getting a bunch of `undefined reference to `__imp_NewMagi
ckWand'`; this is unrelated to this change though, as I also see it w/ the recently merged 7.1.0.39-5 update...

kmilos:
I have now managed to work around the dt cmake related issue and built successfully, and can confirm this now does load images using the MagickWand API. I have no clue why `magick -list format` is still empty though...

Biswa96:
> magick -list format is still empty

Yes, this PR is just focused on path relocation. If everything was fixed that will be a blessing 👼 

> I have now managed to work around the dt cmake related issue and built successfully

If you think that cmake issue is somehow related to msys2's mingw packages please create an issue.

kmilos:
Thanks.

> If you think that cmake issue is somehow related to msys2's mingw packages please create an issue.

I don't think so. Not 100% clear on why it worked before, but it is obvious now the dt cmake scripts needs to be more robust...

PR #13289, opened 2022-09-30
The gst-plugins-bad wrongly pulls in GPL dependencies, despite saying it is LGPL licensed

bassani0:
The (A)GPL licensed plugins in gst-plugins-bad, according to: https://github.com/freedesktop/gstreamer

    dts (DTS audio decoder plugin)
    faad (Free AAC audio decoder plugin)
    iqa (Image quality assessment plugin based on dssim-c)
    mpeg2enc (MPEG-2 video encoder plugin)
    mplex (audio/video multiplexer plugin)
    ofa (Open Fingerprint Architecture library plugin)
    resindvd (Resin DVD playback plugin)
    x265 (HEVC/H.265 video encoder plugin)

At the moment it wrongly pulls in these two:

faad2
x265 

I'm also not sure why it specifically disabled certain GPL libs, but sets gpl=enabled? I'm guessing the lib-specific disablings are a holdover from before the gpl flag was added.

### DIFF
--- a/mingw-w64-gst-plugins-bad/PKGBUILD
+++ b/mingw-w64-gst-plugins-bad/PKGBUILD
@@ -27,7 +27,6 @@ makedepends=(
   "${MINGW_PACKAGE_PREFIX}-cairo"
   "${MINGW_PACKAGE_PREFIX}-chromaprint"
   "${MINGW_PACKAGE_PREFIX}-curl"
-  "${MINGW_PACKAGE_PREFIX}-faad2"
   "${MINGW_PACKAGE_PREFIX}-faac"
   "${MINGW_PACKAGE_PREFIX}-fdk-aac"
   "${MINGW_PACKAGE_PREFIX}-fluidsynth"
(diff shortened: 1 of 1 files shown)


Biswa96:
The same dependencies and options are also present in ArchLinux and Debian.
* ArchLinux: https://github.com/archlinux/svntogit-packages/blob/packages/gstreamer/trunk/PKGBUILD
* Debian: https://salsa.debian.org/gstreamer-team/gst-plugins-bad1.0/-/blob/master/debian/control

bassani0:
> The same dependencies and options are also present in ArchLinux and Debian.
> 
>     * ArchLinux: https://github.com/archlinux/svntogit-packages/blob/packages/gstreamer/trunk/PKGBUILD
> 
>     * Debian: https://salsa.debian.org/gstreamer-team/gst-plugins-bad1.0/-/blob/master/debian/control

Actually, I'm starting to think this was accident. If you look at the source file, you'll see that it bizarrely explicitly disables libde265:

        -Dlibde265=disabled \

despite that library being LGPL. I think the maintainer(s) confused x265 with de265, and thought the former was LGPL and the latter, GPL, when it's the other way around.

Biswa96:
I am not familiar with the intricacies of software licenses. Others may help in that area. The package may loose some functionality which some users may rely on.

lazka:
They are both LGPL, am I missing something?

* https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/faad/gstfaad.c
* https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/x265/gstx265enc.c

DioEgizio:
x265 itself is definitely GPL though

lazka:
Ok, ideally the license is for the content of the package and not any dependencies. There are definitely errors in our license metadata, so fixes there are welcome.

lazka:
I don't see any remaining issues here, so closing.

PR #13301, opened 2022-10-01
protobuf: update to 21.7

MehdiChinoune:


podsvirov:
I try to update, but get the error:
```shell
installing mingw-w64-i686-protobuf (21.7-1) breaks dependency 'mingw-w64-i686-protobuf=21.6' required by mingw-w64-i686-python-protobuf
```
Looks like, we need update `python-protobuf` too.

PR #13357, opened 2022-10-01
llvm: Enable Exception Handling and RTTI on CLANG envs

### LINKED ISSUE #11630: LLVM doesn't support for RTTI,
When building mesa with clover suupport.

```
DirectX-Headers| Build targets in project: 1
DirectX-Headers| Subproject DirectX-Headers finished.

Program glslangValidator found: YES (C:\CI-Tools\msys64\mingw64\bin/glslangValidator.EXE)

../../meson.build:946:4: ERROR: Problem encountered: The Clover OpenCL state tracker requires rtti

A full log can be found at C:/work/xemu/xemu-opengl/mesa/build/windows-mingw64/meson-logs/meson-log.txt

```
```
WARNING: Optional LLVM module all-targets was not found
Run-time dependency LLVM (modules: amdgpu, bitreader, bitwriter, core, coroutines, coverage, executionengine, frontendopenmp, instcombine, instrumentation, ipo, irreader, libdriver, linker, lto, mcdisassembler, mcjit, native, objcarcopts, option, profiledata, 
scalaropts, target, transformutils) found: YES 14.0.0

../../meson.build:1775:8: ERROR: Problem encountered: LLVM was built without RTTI, so Mesa must also disable RTTI. Use an LLVM built with LLVM_ENABLE_RTTI or add cpp_rtti=false.

A full log can be found at C:/work/xemu/xemu-opengl/mesa/build/windows-mingw64/meson-logs/meson-log.txt
```

MehdiChinoune:
Fixes #11630

### DIFF
--- a/mingw-w64-clang/PKGBUILD
+++ b/mingw-w64-clang/PKGBUILD
@@ -26,7 +26,7 @@ _version=15.0.1
 _rc=""
 _tag=llvmorg-${_version}${_rc}
 pkgver=${_version}${_rc/-/}
-pkgrel=3
+pkgrel=4
 pkgdesc="C language family frontend for LLVM (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
@@ -229,6 +229,8 @@ build() {
       -DLLVM_ENABLE_LIBCXX=ON
       -DCLANG_DEFAULT_LINKER=lld
       -DLLVM_ENABLE_LLD=ON
+      -DLLVM_ENABLE_EH=ON
+      -DLLVM_ENABLE_RTTI=ON
       -DCLANG_LINKS_TO_CREATE="${_clang_links}"
       -DLLVM_INSTALL_BINUTILS_SYMLINKS=ON)
   fi


lazka:
Why is this needed/wanted?

MehdiChinoune:
> Why is this needed/wanted?

#11630

lazka:
And why is `LLVM_ENABLE_EH` needed?

MehdiChinoune:
> thanks, and why is `LLVM_ENABLE_EH` needed?

I read once @mati865 talking about to enabling it after the exported symbols were reduced. I don't remember where.

pal1000:
@lazka @MehdiChinoune, it's no longer necessary to have LLVM built with RTTI ever since [landed](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/17055)

mati865:
> > thanks, and why is `LLVM_ENABLE_EH` needed?
> 
> I read once @mati865 talking about to enabling it after the exported symbols were reduced. I don't remember where.

I don't recall it. Maybe it was `LLVM_ENABLE_ASSERTIONS`?

MehdiChinoune:
> I don't recall it. Maybe it was `LLVM_ENABLE_ASSERTIONS`?

Yes, That's it (https://github.com/msys2/MINGW-packages/pull/12550#issuecomment-1216214972)

mati865:
Lets split it into three parts:
- RTTI - just go for it
- EH - I believe it will slow down LLVM (slightly but still) for no benefit
- assertions - it's useful for making sure the compiler works and would even reject some code that would miscompile but also would significantly slow down the compiler and reject some valid code

MehdiChinoune:
> Lets split it into three parts:
> 
> * RTTI - just go for it

Everywhere or just CLANG envs.

mati865:
IMO we should enable it everywhere, it should not cause any issues now that we are not symbol count limited.

PR #13358, opened 2022-10-01
gr: enable qt6

MehdiChinoune:


MehdiChinoune:
automoc starts failing in 32bit envs. I suspect that a package residing in staging is responsible for that, but Which one?

MehdiChinoune:
Reverting to build `libuv` with `autotools` doesn't fix the issue, but downgrading does.

PR #13360, opened 2022-10-02
Add `Cflags.private` to fribidi.pc

shithappens2016:
Switch to autotools due to meson's pkgconfg module do not support `Cflags.private`.

Add `Cflags.private: -DFRIBIDI_LIB_STATIC` to `fribidi.pc` should fix `libavfilter.a` link against `libfribidi.dll`.

fribidi

#8824

### DIFF
--- a/mingw-w64-fribidi/PKGBUILD
+++ b/mingw-w64-fribidi/PKGBUILD
@@ -4,7 +4,7 @@ _realname=fribidi
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.0.12
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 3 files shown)


eli-schwartz:
If you're going to have to patch this with an msys2-specific patch anyway, why not just add it to the meson installed pkg-config file after the fact?

shithappens2016:
@eli-schwartz
I'm concerned that modifying the compiled result may be not the way.

Biswa96:
> I'm concerned that modifying the compiled result may be not the way.

Can you explain what will go wrong if the pkgconfig file is modified at install step?

shithappens2016:
@Biswa96 
Each upgrade may result in a change in compilation results. I don't want to compile it every time to get fribidi.pc before making a patch, it's a waste of time.

lazka:
yes, please try without switching build systems. We are trying to get rid of autotools where possible

PR #13367, opened 2022-10-02
qt6-doc: update to 6.4.0

MehdiChinoune:

PR #13368, opened 2022-10-02
CI: unset VCPKG_ROOT

MehdiChinoune:
Recently GHA started defining VCPKG_ROOT environment variable (https://github.com/actions/runner-images/pull/6192).
So unset it to not let it autodetected and interfere with our build jobs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -89,6 +89,7 @@ jobs:
             . shell ${MSYSTEM,,}
             set -e
           fi
+          unset VCPKG_ROOT
           pacman -S --needed --noconfirm ${MINGW_PACKAGE_PREFIX}-ntldd
           MINGW_ARCH=${{ matrix.msystem }} ./.ci/ci-build.sh
 


Biswa96:
The build log of qt6-doc has many vcpkg related variables. Should not all of the be unset? 💭

lazka:
Just curious, is there some package that failed because of this?

MehdiChinoune:
> Just curious, is there some package that failed because of this?

I think PkgConfig couldn't be found because of it (#13367).
vcpkg has its own CMAKE_TOOLCHAIN_FILE that defines many CMake variables and functions, which could interfere with our default CMake settings.

lazka:
I see, thanks

MehdiChinoune:
> The build log of qt6-doc has many vcpkg related variables. Should not all of the be unset? 💭

The main reason is that vcpkg get autodetected in https://invent.kde.org/qt/qt/qtbase/-/blob/6.4.0/cmake/QtAutoDetect.cmake#L533 which rely on the environment variable `VCPKG_ROOT` to be defined in https://invent.kde.org/qt/qt/qtbase/-/blob/6.4.0/cmake/QtAutoDetect.cmake#L159

lazka:
Also done in autobuild: https://github.com/msys2/msys2-autobuild/commit/c9dd9afe5e0ca2a35fbfd01bae0143b1fba1d9ab

MehdiChinoune:
actions/runners#6376

NicKolai48 (approved):

PR #13378, opened 2022-10-03
Update OpenSSL package to 3.0.5

ls5302:


ls5302:
> Did you check if all the dependent packages are compatible with this newer openssl version? All the dependent packages need to be rebuilt and checked. This list of dependent packages can be found here https://packages.msys2.org/package/mingw-w64-x86_64-openssl?repo=mingw64

lazka:
> @Biswa96 good point, I had not taken than into consideration. There are a lot of packages that depend on OpenSSL.
> 
> Now I'm wondering if I can build a versioned package which doesn't nuke the existing one, but allows the latest release of OpenSSL to be used when required.

That's certainly possible. Arch Linux has this: https://archlinux.org/packages/core/x86_64/openssl-1.1/

> I am aware that OpenSSL reaches EOL in 2023 so this will need to be done at some point.

I just did the upgrade on a smaller scale for the cygwin package: https://github.com/msys2/MSYS2-packages/issues/3013 There haven't been too many issues. But we have lots more packages here :/

MehdiChinoune:
You can't just drop patches like that.

MehdiChinoune:
superseded by #15041

PR #13398, opened 2022-10-06
update mingw-w64

MehdiChinoune:
https://github.com/mingw-w64/mingw-w64/compare/ad45df377..ee5758234

Should we enable `Control Flow Guard` on CLANG environments?


MehdiChinoune:
I will merge it as it is.
We could enable CFG later.

mati865:
> Should we enable `Control Flow Guard` on CLANG environments?

Not yet, we need LLVM 16 (spring 2023) for it.

We should however stop building GCC's libssp since it's already provided by mingw-w64, instead we need to create empty libs: https://github.com/mstorsjo/llvm-mingw/commit/bb3abab2f7b6ecd4f2eb9d4d583ae72c8f898235#diff-0e0768067f48884d3692a25eda52eb11aca3b2ca79870f8da0d1ad7f54c5c289R39-R44
And stop providing DLL: https://github.com/mstorsjo/llvm-mingw/commit/9fa4a6e17c0180a32c4a0e808a44f601c34ba73c

lazka:
re cfg, sounds good to me

PR #13405, opened 2022-10-07
mingw-w64-crt: create dummy ssp libs

MehdiChinoune:


mingw-w64-crt-git/PKGBUILD
```diff
@@ -4,14 +4,17 @@ _realname=crt
...
+pkgrel=2
 _commit='ee5758234f77a6b573e415b26a1d387c46717f6b'
 pkgdesc='MinGW-w64 CRT for Windows'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url='https://mingw-w64.sourceforge.io/'
 license=('custom')
 groups=("${MINGW_PACKAGE_PREFIX}-toolchain")
+conflicts=("${MINGW_PACKAGE_PREFIX}-libssp")
+replaces=("${MINGW_PACKAGE_PREFIX}-libssp")
```
mmuetzel, line 16: Will this cause issues with packages that currently link to `libssp-0.dll`?
Do you plan to recompile them (so they no longer link to that library) when this is merged?

MehdiChinoune: There is no other option, We have to make it replace `libssp`, otherwise we will end up with unresolvable conflict. Almost all packages depends on `crt`, so the ones depending on `libssp` will certainly have a conflict between `crt` and `libssp`. I am talking here about updating packages.
Of course we should recompile packages that depend on `libssp`.

MehdiChinoune:
package-grokker shows that these packages should be rebuilt:
```
mingw-w64-spice
mingw-w64-cjson
mingw-w64-virt-viewer
mingw-w64-mkvtoolnix
mingw-w64-libotr
mingw-w64-opus
mingw-w64-libvirt
mingw-w64-libdsm
mingw-w64-v8
mingw-w64-mxml
mingw-w64-gtk-vnc
mingw-w64-sed
mingw-w64-starpu
mingw-w64-libressl
mingw-w64-flac
mingw-w64-libgusb
mingw-w64-tor
mingw-w64-xapian-core
mingw-w64-gcab
mingw-w64-embree
mingw-w64-vlc
mingw-w64-inkscape
```

mingw-w64-crt-git/PKGBUILD
mmuetzel: Thanks!
Does that only apply to packages for the CLANG* environments?
Should `gcc-libs` stop packaging `libssp-0.dll`, too?

MehdiChinoune: > Should `gcc-libs` stop packaging `libssp-0.dll`, too?

Yes, I am going to open PR about that, once building gcc finishes on my side.

mmuetzel: I guess that is motivated by this upstream change:
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/10394c9a966f8e93e9e2f09677dab273a0f6c00c/

I might be misunderstanding. But the commit message sounds to me like the mingw implementation is only thought as a replacement for mingw environments that "don't otherwise use GCC's runtime libraries
(such as LLVM based ones)". Aren't those only the CLANG* environments?

MehdiChinoune: Do you see any condition in that commit.
> "this also allows relying **entirely** on libmingwex"


> "This **allows** toolchains that don't otherwise use GCC's runtime libraries
(**such as** LLVM based ones)"

a GCC runtime built without `libssp` should belong to those toolchains.

lazka:
Couldn't we just remove the libssp.a/libssp.dll.a from gcc/libssp, so there is no conflict, and then rebuild things afterwards? Or do you think making the full transition in one go is easier?

MehdiChinoune:
> Couldn't we just remove the libssp.a/libssp.dll.a from gcc/libssp, so there is no conflict, and then rebuild things afterwards when we want? Or do you think making the full transition in one go is easier?

That's what I was going to do.

mingw-w64-libssp/PKGBUILD
```diff
@@ -42,16 +42,14 @@ prepare() {
...
+  make libssp-0.dll -f ../libssp/Makefile
 }
 
 package() {
-  cd "${srcdir}/build-${CARCH}"
+  cd "${srcdir}/build-${MSYSTEM}"
   mkdir -p "${pkgdir}${MINGW_PREFIX}"/{bin,lib}
-  cp -f libssp.a "${pkgdir}${MINGW_PREFIX}/lib"
-  cp -f libssp_nonshared.a "${pkgdir}${MINGW_PREFIX}/lib"
   cp -f libssp.dll.a "${pkgdir}${MINGW_PREFIX}/lib"
```
lazka, line 53:
```suggestion

```

should this be removed as well? so things link to the new dummy static lib?

mingw-w64-crt-git/PKGBUILD
mmuetzel: Thanks for clarifying. 👍

MehdiChinoune:
OK to merge?

MehdiChinoune:
My plan is:
1- Merge this PR to provide libssp and libssp_noshared on CLANG*
2- Open PRs to remove explicit dependency on libssp (do not merge) and remove those workarounds of adding `CFLAGS+=" -fstack-protector"` and `LDFLAGS+=" -lssp"`. 
3- merge #13140
4- Wait until gcc is available from staging and rebase PRs => merge.

MehdiChinoune:
#13410 should be merged before this one to avoid conflict between crt and gcc.

lazka:
I will try to test building some libssp packages with this and #13410, to see if it actually works as advertised.

lazka:
yeah, some users might still reference it though.. but not sure how realistic that is

PR #13410, opened 2022-10-07
gcc: disable building libssp

MehdiChinoune:


lazka:
going for #13414 instead

Issue #13401, opened 2022-10-06
Remove libssp / replace with dummy lib


lazka:
As pointed out here: https://github.com/msys2/MINGW-packages/pull/13398#issuecomment-1270214458

jeremyd2019:
There probably needs to be some sort of transition for the DLL, as existing binaries would still link to it.

Biswa96:
Wouldn't only import lib and static lib be replaced with dummy files? Then there will be no issue as DLL are present, right?

MehdiChinoune:
Packages to be rebuilt

- [ ] mingw-w64-spice
- [ ] mingw-w64-cjson
- [ ] mingw-w64-virt-viewer
- [ ] mingw-w64-mkvtoolnix
- [ ] mingw-w64-libotr
- [ ] mingw-w64-opus
- [ ] mingw-w64-libvirt
- [ ] mingw-w64-libdsm
- [ ] mingw-w64-v8
- [ ] mingw-w64-mxml
- [ ] mingw-w64-gtk-vnc
- [ ] mingw-w64-sed
- [ ] mingw-w64-starpu
- [ ] mingw-w64-libressl
- [ ] mingw-w64-flac
- [ ] mingw-w64-libgusb
- [ ] mingw-w64-tor
- [ ] mingw-w64-xapian-core
- [ ] mingw-w64-gcab
- [ ] mingw-w64-embree
- [ ] mingw-w64-vlc
- [ ] mingw-w64-inkscape

lazka:
I've added a NEWS entry: https://www.msys2.org/news/#2022-10-10-libssp-is-no-longer-required feedback welcome.

MehdiChinoune:
all packages were rebuilt.

MehdiChinoune:
Remaining:
#13527 
#13528

MehdiChinoune:
Shouldn't this issue be closed?

lazka:
Yes, thanks for all the package updates.

Followup here https://github.com/msys2/MSYS2-packages/pull/3222 as suggested here https://github.com/msys2/MINGW-packages/issues/5803#issuecomment-1273777694

PR #13414, opened 2022-10-07
gcc: remove libssp static and import lib

lazka:
but still provide the .DLL.

In combination with #13405 this will make all rebuilds link the libssp stuff against the crt while keeping existing packages already linking against the .DLL working until we rebuild them.

This is an alternative to #13410


lazka:
@MehdiChinoune sorry for the back and forth. Does this make sense to you together with #13401?

MehdiChinoune:
> @MehdiChinoune sorry for the back and forth. Does this make sense to you together with #13405? (edit: linked wrong PR initially)

LGTM.

Issue #13390, opened 2022-10-04
[ucrt64|mingw64] pkgconf 1.9.3-1 issue with static library ordering


1480c1:
### Description

Installing opusfile and running `pkgconf --static --libs opusurl` provides incorrect ordering of libraries

```bash
cddeg@ccom-laptop UCRT64 /c/Users/cddeg
$ pkgconf.exe --version
1.8.0

cddeg@ccom-laptop UCRT64 /c/Users/cddeg
$ pkgconf --static --libs opusurl
-LC:/msys64/ucrt64/lib -lopusurl -lopusfile -LC:/msys64/ucrt64/lib -logg -lopus -lm -lssp -LC:/msys64/ucrt64/lib -lssl -LC:/msys64/ucrt64/lib -lws2_32 -lgdi32 -lcrypt32 -lcrypto -lws2_32 -lgdi32 -lcrypt32

cddeg@ccom-laptop UCRT64 /c/Users/cddeg
$ pacman -Su
:: Starting core system upgrade...
 there is nothing to do
:: Starting full system upgrade...
resolving dependencies...
looking for conflicting packages...

Packages (1) mingw-w64-ucrt-x86_64-pkgconf-1.9.3-1

Total Installed Size:  0.49 MiB
Net Upgrade Size:      0.01 MiB

:: Proceed with installation? [Y/n] y
(1/1) checking keys in keyring                                           [#######################################] 100%
(1/1) checking package integrity                                         [#######################################] 100%
(1/1) loading package files                                              [#######################################] 100%
(1/1) checking for file conflicts                                        [#######################################] 100%
(1/1) checking available disk space                                      [#######################################] 100%
:: Processing package changes...
(1/1) upgrading mingw-w64-ucrt-x86_64-pkgconf                            [#######################################] 100%

cddeg@ccom-laptop UCRT64 /c/Users/cddeg
$ pkgconf.exe --version
1.9.3

cddeg@ccom-laptop UCRT64 /c/Users/cddeg
$ pkgconf --static --libs opusurl
-LC:/msys64/ucrt64/lib -lssl -LC:/msys64/ucrt64/lib -lcrypto -lws2_32 -lgdi32 -lcrypt32 -lopusurl -lopusfile -LC:/msys64/ucrt64/lib -logg -lopus -lm -lssp
```

for reference, on a linux install
```bash
$ pkgconf --static --libs opusurl
[...]

PR #13417, opened 2022-10-07
Revert pkgconf

lazka:
There have been regressions/behaviour changes resulting in build errors.
Let's revert until we figure out more.

See https://github.com/msys2/MINGW-packages/issues/13390


lazka:
If anyone wants to reproduce the issues found after this:

```
pacman -U https://mirror.msys2.org/mingw/ucrt64/mingw-w64-ucrt-x86_64-pkgconf-1.9.3-1-any.pkg.tar.zst
```

PR #13426, opened 2022-10-08
libvirt: remove dependency on libssp

MehdiChinoune:
~Succeeds on CLANG64 and UCRT64 and Fails elsewhere.
xsltproc fails to run both the one from MSYS and MINGW32|MINGW64|CLANG32.~


MehdiChinoune:
Building docs fails.
I am going to merge it as it is.
If someone needs docs then he is welcome to investigate xslt random failure and fix it.

PR #13444, opened 2022-10-08
arm-none-eabi-gcc: update to 12.2.0

MehdiChinoune:
times out!

### DIFF
--- a/mingw-w64-arm-none-eabi-gcc/PKGBUILD
+++ b/mingw-w64-arm-none-eabi-gcc/PKGBUILD
@@ -5,23 +5,24 @@
 _realname=gcc
 _target=arm-none-eabi
 
-_enable_ada=yes
+_enable_ada=no
 
 pkgbase=mingw-w64-${_target}-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_target}-${_realname}")
-pkgver=10.1.0
-pkgrel=2
+pkgver=12.2.0
+pkgrel=1
 pkgdesc='GNU Tools for ARM Embedded Processors - GCC (mingw-w64)'
 arch=('any')
-mingw_arch=('mingw32' 'mingw64')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
 license=('GPL')
 url='https://www.gnu.org/software/gcc/gcc.html'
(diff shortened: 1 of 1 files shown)


lazka:
ok if I drop clang64 again? It's really annoying to build locally since it takes ages :(

MehdiChinoune:
> ok if I drop clang64 again? It's really annoying to build locally since it takes ages :( so I want to limit the number of builds

Up to you, It's not something I need or someone asked for.

MehdiChinoune:
Why not dropping MINGW32 instead?

lazka:
ok, fine with me

pfeerick:
Is this PR the reason there is no longer a `mingw-w64-i686-arm-none-eabi-gcc` package in the MSYS packages repo? Trying to figure out why our github builds have just started breaking when `pacman` is trying to grab dependencies for a 32bit build.

lazka:
> Is this PR and [52a0f4e](https://github.com/msys2/MINGW-packages/commit/52a0f4e32b3d5f22defaa4b71c8bea90789e7a96) the reason there is no longer a `mingw-w64-i686-arm-none-eabi-gcc` package in the MSYS packages repo? Trying to figure out why our github builds have started breaking when `pacman` is trying to grab dependencies for a 32bit build.

yes. do you need it? I can add it back then.

Biswa96:
Is it required to build a project with 32bit cross compiler? You are already cross compiling for arm-eabi.

MehdiChinoune:
Your project has two workflow files https://github.com/EdgeTX/edgetx/blob/main/.github/workflows/win-cpn-32.yml and https://github.com/EdgeTX/edgetx/blob/main/.github/workflows/win_cpn-64.yml
Why the 32bit installs `arm-none-abi-gcc` while the 64bit doesn't!

pfeerick:
>  Why the 32bit installs `arm-none-abi-gcc` while the 64bit doesn't!

Good question... the person who modified the win64 build to do a win32 build must have thought (incorrectly) that it was needed. 

> yes. do you really need it? I can add it back then.

No, was just looking for confirmation of the assumption before looking deeper after a nights sleep :) Removing this dependency (or rather, non-dependency) will resolve the issue.

Thank you both! :)

Issue #11777, opened 2022-06-04
Imagemagick: `-list format` doesn't output anything


webbedspace:
Here's the output when I run `magick -version`:
```
% magick -version
Version: ImageMagick 7.1.0-35 Q16-HDRI x86_64 20062 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules OpenMP(4.5)
Delegates (built-in): bzlib cairo djvu fftw flif fontconfig freetype gslib gvc heic jbig jng jp2 jpeg lcms lqr ltdl lzma openexr pangocairo png ps raqm raw rsvg tiff webp wmf xml zlib
Compiler: gcc (12.1)
```
Here's the output when I run `magick -list format`
```
% magick -list format

%
```
Most other `-list` commands work correctly (haven't checked all of them)

Biswa96:
Want to try the suggestion here https://github.com/ImageMagick/ImageMagick/discussions/5219 ?

kmilos:
[Moving the config file as described here](https://github.com/msys2/MINGW-packages/issues/3444) does solve the `magick -list configure` problem, but not the format/coders access...

> Want to try ...

Unfortunately can't find the time to go down that rabbit hole now.

Biswa96:
> but not the format/coders access...

Can you provide any hint about that issue? Like any error output and what will be the expected result.

Biswa96:
The search path of configure.xml file is missing a slash.

PR #13499, opened 2022-10-09
imagemagick: Fix list format command

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/11777


Biswa96:
Would you like to test the build artifact's when it will be finished? Alos please make sure the changes didn't create any other issue with Darktable.

kmilos:
Indeed, `magick -list format` is fixed, and I can still load using an IM coder from darktable, thanks!

PR #13521, opened 2022-10-10
riscv64-unknown-elf-gcc: enable clang64

MehdiChinoune:


mati865:
Maybe one day we can make this PKGBUILD actually use Clang on Clang subsystems 😁

MehdiChinoune:
> Maybe one day we can make this PKGBUILD actually use Clang on Clang subsystems 😁

What do you mean?, Do you mean building cross-clang on CLANG envs and cross-gcc on GCC envs?

mati865:
We don't need to build Clang again since it can target anything it supports (GCC has to be build for 1 specific target).
Probably only building compiler-rt would be necessary.

PR #13528, opened 2022-10-10
gcc: disable libssp

MehdiChinoune:


dscho:
Hmm. This might have made sense from the point of view that everything inside MSYS2 is built and packaged by MSYS2. But what about custom-built software? (Or for that matter, Git for Windows?)

It would probably have made more sense to keep a working `libssp-0.dll` in place, at least for a while, maybe with a `DllMain()` that warns about that library going away. As it is, there were a scant 3 days grace period for downstream users to adapt and plan ahead, 3 days during which some people were on vacation (me, for example, until I got pulled back ☹️).

jeremyd2019:
https://github.com/msys2/MINGW-packages/issues/13401#issuecomment-1270470685 I guess I wasn't explicit that included 3rd party binaries, not just those in packages.

lazka:
Sorry for the inconvenience, reverted now (in git)

dscho:
It's not a big deal, but a good opportunity to keep in mind how far-reaching seemingly innocuous changes can be.

lazka:
It's been 7 months now, and I checked that both git-sdk-64 and git-sdk-32 no longer link anything against libssp. So I think we can drop it now.

lazka:
PR for removal: #17080

lazka:
removed now

dscho:
> It's been 7 months now, and I checked that both git-sdk-64 and git-sdk-32 no longer link anything against libssp. So I think we can drop it now.

Thank you for your diligence. Indeed, we do not need this anymore. FWIW we worked around it very easily, via git-for-windows/git-sdk-64@706bd79657552828851df823edf8a189461bcd7d

PR #13530, opened 2022-10-11
mcfgthread: Update to 1.2, and optimize it for size

lhmouse:


mingw-w64-mcfgthread/PKGBUILD
```diff
@@ -33,6 +33,8 @@ build() {
   [[ -d "${srcdir}/build-${MINGW_CHOST}" ]] && rm -rf "${srcdir}/build-${MINGW_CHOST}"
   mkdir -p "${srcdir}/build-${MINGW_CHOST}" && cd "${srcdir}/build-${MINGW_CHOST}"
 
+  export CFLAGS='-Os -g'
```
Biswa96, line 36: Should `debug` option be checked for these compiler flags?

lhmouse: Well I am not familiar with it.

In principle, debug symbols exist only in static libraries; they will be stripped from DLLs even when `-g` is given.

But indeed, `=` is incorrect here. I will update it.

Biswa96: makepkg applies `-O2` option. Is `-Os` required in this package?

lhmouse: No, optimizing for size is not required, but recommended. The default (optimizing for speed) makes the DLL larger by ~1600 bytes for no benefit I think.

Biswa96: Thanks for the explanation.

lhmouse:
Thank you.

Issue #13539, opened 2022-10-12
[libbdplus] Build failure after libgpg-error update


Biswa96:
### Description

mingw-w64-libbdplus fails to build with libgpg-error version 1.46. Here is the output:

```
checking for gpg-error-config... no
configure: error: gpg-error not found on system
```

In case of libbdplus, the project was not updated for newer libgpg-error changes. Here is the changes of files between libgpg-error version 1.45 and 1.46.

```diff
--- /dev/fd/63	2022-10-09 15:06:30.000000000 +0000
+++ /dev/fd/62	2022-10-09 15:06:30.000000000 +0000
@@ -1,7 +1,6 @@
 /mingw64/
 /mingw64/bin/
 /mingw64/bin/gpg-error.exe
-/mingw64/bin/gpg-error-config
 /mingw64/bin/gpgrt-config
 /mingw64/bin/libgpg-error-0.dll
 /mingw64/bin/yat2m.exe
@@ -96,6 +95,3 @@
 /mingw64/share/locale/zh_TW/
 /mingw64/share/locale/zh_TW/LC_MESSAGES/
 /mingw64/share/locale/zh_TW/LC_MESSAGES/libgpg-error.mo
-/mingw64/share/man/
-/mingw64/share/man/man1/
-/mingw64/share/man/man1/gpgrt-config.1.gz
```

As you can see, the `gpg-error-config` file was removed. Instead `gpgrt-config` should be used. So, what should be the proper way to fix the issue? Some proposal:
1. Copy gpgrt-config to a new gpg-error-config file to make it backward compatible. Though I am not sure if those give equivalent output.
2. Patch all old packages (here libbdplus) to use gpgrt-config file instead of gpg-error-config file.
3. Patch all old packages (here libbdplus) to check for both gpgrt-config and gpg-error-config file (like libksba).
4. Just abandon old packages. Zero maintenance cost.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Microsoft Windows [Version 10.0.19044.1645]

### MINGW environments affected

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [X] CLANGARM64

### Expected behavior

libbdplus builds without any issue.

### Actual behavior

libbdplus fails to build.

### Repro steps

Build mingw-w64-libbdplus with makepkg-mingw.

[...]

PR #13540, opened 2022-10-12
libgpg-error: Install gpg-error-config script explicitly

Biswa96:
This is required for old packages like libbdplus.
Fixes https://github.com/msys2/MINGW-packages/issues/13539


Biswa96:
Why does it always nothing but sandyriver mirror only fails 🤦‍♂️ https://github.com/msys2/MINGW-packages/actions/runs/3237315421/jobs/5304226630#step:6:327

PR #13546, opened 2022-10-13
libyuv: update to 1844.r2374.f9fda6e7

kmilos:
In preparation for upcoming libavif 0.11.0


Biswa96:
Should this be merged now? Or does libavif need to be rebuilt?

kmilos:
Yep, libavif and all of its deps need to be rebuild after libyuv.

Biswa96:
Would you like to rebuild those here or in different PR?

kmilos:
Sure, I can do so here...

kmilos:
Looks like there's a few KDE related components to bump, so I'll probably leave that at 5.98

Biswa96:
Yeah, you can exclude the kimageformats. I shall try to update KDE packages after this PR.

novomesk:
FYI: when you try to upgrade kimageformats to 5.99, it will not build with the new libavif 0.11.0
Patching of kimageformats will be needed.

It is a result of unfortunate situation / timing. I changed to code in kimageformats to reflect changes in libavif git and kimageformats 5.99 was released. Meanwhile libavif reverted the changes and the released kimageformats 5.99 no longer works with that.

It is fixed in kimageformats git but next release will be approximately November 12, 2022.

Biswa96:
> Patching of kimageformats will be needed.

Agree https://github.com/archlinux/svntogit-packages/commit/7b4dbf81760382afaf1b858d6ef568acb1b99f3f

kmilos:
I just bumped kimageformats 5.98 instead so this can be dealt with separately.

novomesk:
> Agree [archlinux/svntogit-packages@7b4dbf8](https://github.com/archlinux/svntogit-packages/commit/7b4dbf81760382afaf1b858d6ef568acb1b99f3f)

Those patches referenced there are focusing on another problem.

Biswa96:
> when you try to upgrade kimageformats to 5.99, it will not build with the new libavif 0.11.0

So, the better choice would be to wait for next KDE updates, right?

novomesk:
For my point of view, it is OK to wait till 5.100

kmilos:
I give up, something weird is up w/  CI now...

PR #13553, opened 2022-10-14
igraph: update to 0.10.2

### LINKED ISSUE #8399: Hangs with OpenBLAS from 32-bit MinGW

MehdiChinoune:
Fixes #8399

### DIFF
--- a/mingw-w64-igraph/PKGBUILD
+++ b/mingw-w64-igraph/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=igraph
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=0.10.1
-pkgrel=2
+pkgver=0.10.2
+pkgrel=1
 pkgdesc="Library for the analysis of networks (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


szhorvat:
Is there a particular concern about this version?

PR #13604, opened 2022-10-16
go: update to 1.19

aryan9600:


mingw-w64-go/PKGBUILD
```diff
@@ -4,7 +4,7 @@
 _realname=go
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=1.18
+pkgver=1.19
 pkgrel=2
```
Biswa96, line 8: pkgrel should be reset to 1 when pkgver is updated.

aryan9600: i might be misunderstanding the purpose of `pkgrel`. Is it not indicate the patch version of the upstream project?

Biswa96: > pkgrel is usually a positive integer number that allows to differentiate between consecutive builds of the same version of a package.[^1]

[^1]: https://wiki.archlinux.org/title/PKGBUILD#pkgrel

PR #13613, opened 2022-10-17
[new-package] nvc 1.8-devel-4efbf71

amb5l:
The NVC VHDL Compiler and Simulator.

### DIFF
--- a/mingw-w64-nvc/PKGBUILD
+++ b/mingw-w64-nvc/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=nvc
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-_version="1.7.0.r209"
-_commit="4ef679f1"
+_version="1.7.0.r289"
+_commit="134941af"
 pkgver="${_version}.g${_commit}"
 pkgrel=1
 pkgdesc="VHDL compiler and simulator (mingw-w64)"
--- a/mingw-w64-nvc/PKGBUILD
+++ b/mingw-w64-nvc/PKGBUILD
@@ -1,0 +1,68 @@
+# Maintainer: Adam Barnes <ambarnes@gmail.com>
+
+_realname=nvc
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+_version="1.7.0.r209"
+_commit="4ef679f1"
+pkgver="${_version}.g${_commit}"
+pkgrel=1
+pkgdesc="VHDL compiler and simulator (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
+url='https://www.nickg.me.uk/nvc/'
+license=('GPL3')
+groups=("${MINGW_PACKAGE_PREFIX}-eda")
+depends=(
+  "${MINGW_PACKAGE_PREFIX}-llvm" # required by NVC for compiling
+  "${MINGW_PACKAGE_PREFIX}-cc"   # required by NVC for linking
+)
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-autotools"
+  "git"
+  "${MINGW_PACKAGE_PREFIX}-ncurses"
+  "${MINGW_PACKAGE_PREFIX}-libffi"
+  "${MINGW_PACKAGE_PREFIX}-check"
+  "${MINGW_PACKAGE_PREFIX}-pkg-config"
+)
+source=("${_realname}-${pkgver}::git+https://github.com/nickg/nvc.git#commit=${_commit}")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+  git describe --long | sed 's/^r\(.*\)/\1/; s/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+
+  ./autogen.sh
+}
+
+build() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+  mkdir -p "${srcdir}/build-${MSYSTEM}" && cd "${srcdir}/build-${MSYSTEM}"
+
+  ../"${_realname}-${pkgver}"/configure \
+    --prefix="${MINGW_PREFIX}" \
+    --build="${MINGW_CHOST}" \
+    --host="${MINGW_CHOST}" \
+    --target="${MINGW_CHOST}" \
+    --enable-static \
+    --enable-shared \
+    --program-prefix=
+
+  make
+}
+
+check() {
+  cd "${srcdir}/build-${MSYSTEM}"
+
+  make check
+}
+
+package() {
+  cd "${srcdir}/build-${MSYSTEM}"
+
+  make install DESTDIR="${pkgdir}"
+}


mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,38 @@
+pkgname=${MINGW_PACKAGE_PREFIX}-nvc
+_version="1.8_devel"
+_commit="4efbf71"
+pkgver="${_version}_${_commit}"
+pkgrel=1
+pkgdesc="VHDL compiler and simulator"
+arch=('x86_64')
+mingw_arch=('mingw64')
```
1480c1, line 8: is this only available for mingw64?

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+  "base-devel"
+  "git"
+  "autoconf"
+  "automake-wrapper"
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+  "${MINGW_PACKAGE_PREFIX}-llvm"
+  "${MINGW_PACKAGE_PREFIX}-ncurses"
+  "${MINGW_PACKAGE_PREFIX}-libffi"
+  "${MINGW_PACKAGE_PREFIX}-check"
+  "${MINGW_PACKAGE_PREFIX}-pkg-config"
```
1480c1, line 22: are all of these really `depends` and not just `makedepends`?  I think you only need to list the package in one or the other. As far as I understand, `depends` and `makedepends` are installed when building the package, but only `depends` is installed when the end user runs `pacman -S`

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+)
+makedepends=( "${depends[@]}" )
+provides=("${MINGW_PACKAGE_PREFIX}-nvc")
```
1480c1, line 25: Don't think this provides is necessary, since there isn't a category of packages that could provide nvc other than just this one

mingw-w64-nvc/PKGBUILD
amb5l: Agreed. I will remove `depends`.

1480c1: Now, if some of them are required as dlls for loading the binary, then they should be kept, else moved to makedepends if they are needed at build time

amb5l: You're right - there may be some runtime dependancies. I will carry out some testing.

mingw-w64-nvc/PKGBUILD
amb5l: I will do some testing with other subsystems to see if I can add some.

mingw-w64-nvc/PKGBUILD
amb5l: Agreed.

MehdiChinoune:
Please take a look at https://www.msys2.org/dev/new-package/, and Please fellow the template in https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.autotools-tarball.sh

mingw-w64-nvc/PKGBUILD
amb5l: NVC does not support 32 bit subsystems. I've added ucrt64. clang64 does not work at the moment.

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
+_realname=nvc
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+_version="1.8_devel"
+_commit="4efbf71"
+pkgver="${_version}_${_commit}"
+pkgrel=1
+pkgdesc="VHDL compiler and simulator"
```
1480c1, line 7: I think convention is to add `(mingw-w64)` at the end of the pkgdesc

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64')
+url='https://github.com/nickg/nvc'
+license=('GPL3')
+groups=("${MINGW_PACKAGE_PREFIX}-eda")
+depends=(
+  "${MINGW_PACKAGE_PREFIX}-llvm"
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+)
```
1480c1, line 16: I think this is normally gcc-libs or something, unless it requires llvm and gcc at runtime?

mingw-w64-nvc/PKGBUILD
amb5l: Thanks, I'll add this.

mingw-w64-nvc/PKGBUILD
amb5l: I believe that it does actually need these at runtime - NVC compiles VHDL to LLVM IR. For example it fails without libLLVM-15.dll.a, which is part of llvm.

1480c1: I see, so it's a compiler frontend essentially. Does it need gcc then? That description leaves me with the impression that it only needs llvm

amb5l: It uses gcc for elaboration (like linking) I believe.

1480c1: Ah, interesting, so it doesn't stick to using llvm's toolchain all the way though. Could you potentially add that info as a comment next to the individual packages? It may help if someone looks at this file in the future without the context of this PR

mingw-w64-nvc/PKGBUILD
1480c1: Do you happen to have a log of what errored out for clang64?

amb5l: I copied the first couple of errors to NVC's author:
```
  CC       src/vhpi/vhpi-model.o
../src/jit/jit-irgen.c:113:20: warning: unused function 'jit_value_is_addr' [-Wunused-function]
static inline bool jit_value_is_addr(jit_value_t value)
                   ^
../src/util.c:841:7: error: expected expression
      void (*fn)(void *) = atomic_load(&ctrl_c_fn);
      ^
```

mingw-w64-nvc/PKGBUILD
amb5l: I'll ask NVC's author Nick Gasson for some help getting this right and will add these as requested.

nickg: It uses GCC as linker only, to link the object files generated by LLVM. It should be possible to use clang as a linker too (pass `--use-system-cc=clang` to `configure`), but I've never tested that on Windows.

nickg: On Linux it's actually possible to just call `ld` directly and not use the gcc/clang frontend, but I don't think that works on Windows as there's a lot of extra flags/libraries that need to be passed to `ld` that only gcc/clang know about.

1480c1: would `--use-system-cc=cc` not work (I think for clang64, it has `cc` -> `clang` and for the others, it has `cc` -> `gcc`)

amb5l: I will test this.

1480c1: seems that flag is not present in my clone of configure after running autogen

mingw-w64-nvc/PKGBUILD
1480c1:
```diff
diff --git a/src/util.c b/src/util.c
index 51dc9163..d99137f1 100644
--- a/src/util.c
+++ b/src/util.c
@@ -836,9 +836,10 @@ void register_signal_handlers(void)
 #ifdef __MINGW32__
 static BOOL win32_ctrl_c_handler(DWORD fdwCtrlType)
 {
+   void (*fn)(void *);
    switch (fdwCtrlType) {
    case CTRL_C_EVENT:
-      void (*fn)(void *) = atomic_load(&ctrl_c_fn);
+      fn = atomic_load(&ctrl_c_fn);
       if (fn != NULL)
          (*fn)(ctrl_c_arg);
       return TRUE;

```
seems this is enough to make clang compile, don't know exactly why though atm, haven't investigated

amb5l: Thanks for this, I've added a patch file.

mingw-w64-nvc/PKGBUILD
amb5l: Thanks @nickg, it's building for clang64 on my machine with the patch from @1480c1.

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
+_realname=nvc
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
```
MehdiChinoune, line 2:
```suggestion
_realname=nvc
pkgbase=mingw-w64-${_realname}
pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
```

amb5l: Agreed. This will be in my next push.

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+_version="1.8_devel"
+_commit="4efbf71"
+pkgver="${_version}_${_commit}"
```
MehdiChinoune, line 5: Please use pkgver in form `1.8.rNNN.g${_commit}`, see other packages as an example.

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+groups=("${MINGW_PACKAGE_PREFIX}-eda")
+depends=(
+  "${MINGW_PACKAGE_PREFIX}-llvm"
+  "${MINGW_PACKAGE_PREFIX}-gcc"
+)
+makedepends=(
+  "base-devel"
+  "git"
+  "autoconf"
+  "automake-wrapper"
```
MehdiChinoune, line 21:
```suggestion
  "${MINGW_PACKAGE_PREFIX}-autotools"
  "git"
```
I told you to fellow the template!

mingw-w64-nvc/PKGBUILD
amb5l: Thanks for figuring that out. I've hardwired the NVC commit ID into the PKGBUILD and this PR's subject, the patch is working fine, so I won't bump it just now.

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+  "${MINGW_PACKAGE_PREFIX}-ncurses"
+  "${MINGW_PACKAGE_PREFIX}-libffi"
+  "${MINGW_PACKAGE_PREFIX}-check"
+  "${MINGW_PACKAGE_PREFIX}-pkg-config"
+)
+source=("${_realname}::git+${url}.git#commit=${_commit}")
+sha512sums=('SKIP')
+build() {
+  cd ${srcdir}/${_realname}
+  ./autogen.sh
```
MehdiChinoune, line 31:
```suggestion
prepare() {
  cd ${srcdir}/${_realname}
  ./autogen.sh
}
build() {
```
autogen should be part of prepare.

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+  mkdir build && cd build
+  ../configure --prefix=${MINGW_PREFIX}
```
MehdiChinoune, line 33: Please fellow template!

mingw-w64-nvc/PKGBUILD
amb5l: I'll go with 1.8.0.r001.g${_commit} - a 3 digit version like GHDL.

mingw-w64-nvc/PKGBUILD
amb5l: You did, I thought these dependencies were examples not requirements. I will change this.

mingw-w64-nvc/PKGBUILD
amb5l: Agreed.

mingw-w64-nvc/PKGBUILD
amb5l: Lesson learned!

MehdiChinoune:
Please add pkgver function.
https://wiki.archlinux.org/title/VCS_package_guidelines

mingw-w64-nvc/PKGBUILD
```diff
@@ -0,0 +1,68 @@
+# Maintainer: Adam Barnes <ambarnes@gmail.com>
+
+_realname=nvc
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+_version="r1.7.0.r183"
```
MehdiChinoune, line 6: Please adjust pkgver function to remove the first `r` from version

amb5l: Sorry that was a typo. Done.

amb5l:
Please pause this PR for now, testing of the built packages has revealed some problems. I will push a fix shortly.

amb5l:
We now have a version that passes the important tests. This PR is a bit of a mess - 10+ commits - would you prefer me to close and try again, or submit a final commit to this PR?

mmuetzel:
You could also squash your commits and force push to your branch to condense the commits in this PR.

amb5l:
The [latest run](https://github.com/msys2/MINGW-packages/actions/runs/3355231905) looks very strange, as if the wrong PKGBUILD was used. I will have another go at squashing commits.

mmuetzel:
It doesn't look like squashing the commits in this PR worked. One way to manually squash your commits is:
* Locally strip all your changes from this PR with `git reset --soft bf014c6c5f4b9caae8ca99073f32b9550b44ca9b~1`. That way all files will stay the same. But the local branch is back to where you started.
* Optionally, fast-forward to a current head.
* Commit your changes (in one single changeset or whatever number of incremental steps you prefer).
* Force push the new state of your local branch to your remote branch.

Those steps worked for me in most of the cases.

amb5l:
@mmuetzel thank you for your assistance. My fork now has a single commit.

amb5l:
Hi all - please let me know if anything else is required for you to accept this PR.

PR #13617, opened 2022-10-19
Levenshtein: new package

wlph17:


mingw-w64-python-python-Levenshtein/PKGBUILD
```diff
@@ -0,0 +1,40 @@
+# Contributor: Liu Pak Hin <wl@m-labs.hk>
+
+_realname=python-Levenshtein
```
Biswa96, line 3: package name should be in lowercase.

Biswa96:
The directory has duplicate 'python' name.

PR #13618, opened 2022-10-19
aws-c-auth: rebuild

kou:
Because this depends on aws-c-common.


MehdiChinoune:
Why?
Dud aws-c-common break ABI with the latest update?

kou:
Yes.
URI related codes were moved to aws-c-common from aws-c-io:
* https://github.com/awslabs/aws-c-common/commit/be1224267cd1b875c50b202cce013cb06b0be458
* https://github.com/awslabs/aws-c-io/commit/550e3192b9d1c7ac865d3f71273bfd7729460900

MehdiChinoune:
package-grokker (https://github.com/msys2/MINGW-packages/actions/runs/3255290437/jobs/5344618607#step:3:64) showed:
```
mingw-w64-aws-c-http
mingw-w64-aws-c-mqtt
mingw-w64-aws-c-s3
mingw-w64-aws-c-auth
mingw-w64-aws-crt-cpp
```

kou:
Does it show that these packages don't need to be rebuilt?

lazka:
> I'm sorry but I don't know how to interpret the result...
> Could you explain it?

It shows the packages that should have been rebuilt.

Though it's unusual for a package to require rebuilds for a patch update.

kou:
Ah, thanks. I understand. Then we may need to rebuild mingw-w64-aws-c-http and mingw-w64-aws-c-mqtt too. But they don't have any problem on my environment. They may not use removed APIs.

AWS SDK don't use semantic versioning. So we can't infer ABI compatibility from version number.

lazka:
ok, thanks. I triggered a rebuild anyway, just to be safe.

PR #13651, opened 2022-10-20
libde265: update to 1.0.9

kmilos:


kmilos:
```
  ../../../libde265-1.0.9/libde265/x86/sse-motion.cc:746:29: error: always_inline function '_mm_packus_epi32' requires target feature 'sse4.1', but would be inlined into function 'ff_hevc_weighted_pred_avg_8_sse4' that is compiled without support for 'sse4.1'
                         r0 = _mm_packus_epi32(r0, r2);
                              ^
```

Presumably related to recent -march=core2 changes?

lhmouse:
The offending function in 'sse-emotion.c' should be excluded from compilation, or an appropriate `target` attribute has to be applied, because Core 2 has no SSE4 at all. Please provide a patch, and report this issue upstream.

lhmouse:
Please try adding `__attribute__((__target__("sse4.1"))` in front of `ff_hevc_weighted_pred_avg_8_sse4`.

Looks like a blind check [here](https://github.com/strukturag/libde265/blob/7d5aeb5f11531de33f5b7ae0e768ffc50da4facb/configure.ac#L207), not necessarily related to the `-march=` option.

kmilos:
> Looks like a blind check [here](https://github.com/strukturag/libde265/blob/7d5aeb5f11531de33f5b7ae0e768ffc50da4facb/configure.ac#L207), not necessarily related to the `-march=` option.

FWIW, there was no change to `x86/sse-motion.cc` code since 1.0.8 (nor anything SSE related in the configure script), and it built just fine before the -march=core2 change, no?

kmilos:
In the passing x86 case I see `-march=pentium4 -msse4.1`, but for x86_64 I don't see `-msse4.1` added despite the compiler being ok.

lhmouse:
Examined Ci log, and tried building locally. The issue is caused by ` -mno-ssse3` which disables SSE4.1 as well. libde265 has `-msse4.1` in 'x86/Makefile.am' but it's in front of user-defined `CXXFLAGS` and gets overriden. @lazka If `core` is not an option due to SSSE3 we can switch to `nocona`:

- **prescott** Improved version of Intel Pentium 4 CPU with MMX, SSE, SSE2 and SSE3 instruction set support.
- **nocona** Improved version of Intel Pentium 4 CPU with 64-bit extensions, MMX, SSE, SSE2 and SSE3 instruction set support.
- **core2** Intel Core 2 CPU with 64-bit extensions, MMX, SSE, SSE2, SSE3 and SSSE3 instruction set support

lazka:
hm, ok. so that means any "-mno-*" are basically not allowed in CFLAGS. thanks, I'll have another look.

jeremyd2019:
The GCC page I was looking at doesn't list CX16 on nocona, while it does for core2: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html.  CMPXCHG16B was the whole reason for the exercise, wasn't it?

lazka:
> The GCC page I was looking at doesn't list CX16 on nocona, while it does for core2: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html. CMPXCHG16B was the whole reason for the exercise, wasn't it?

the GCC page is wrong in this case: https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/gcc/common/config/i386/i386-common.c#L1742

lhmouse:
> The GCC page I was looking at doesn't list CX16 on nocona, while it does for core2: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html.  CMPXCHG16B was the whole reason for the exercise, wasn't it?

```
lh_mouse@lhmouse-dev ~ $ echo 'typedef int __attribute__((__mode__(__TI__))) T;  T myxch(T* ptr) { return __sync_lock_test_and_set(ptr, 42);  }' | gcc -x c - -O2 -S -o - -march=nocona | fgrep cmpxchg
        lock cmpxchg16b (%rdi)
lh_mouse@lhmouse-dev ~ $
```

lazka:
I've created https://github.com/msys2/MSYS2-packages/pull/3239

jeremyd2019:
Never trust the docs 😁

lazka:
> Never trust the docs 😁

filed upstream https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107337

kmilos:
Thanks everyone!

lazka:
> filed upstream https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107337 (I didn't bother to go the ML route.. let's see..)

meh, didn't work

PR #13666, opened 2022-10-21
llvm: Update to 15.0.3

lazka:
openblas fix was backported: https://github.com/llvm/llvm-project/commit/61fa70903191f

"llvm-config.exe --link-static --system-libs" looks good now without the patch, likely due to https://github.com/llvm/llvm-project/commit/b4840279846e1eea44

and https://github.com/llvm/llvm-project/commit/f3c5289e7846 likely means we can rebuild openblas now with full optimizations


lazka:
confirmed that llvm-config still works

PR #13672, opened 2022-10-21
update llvm packages to 15.0.3

jeremyd2019:
(except for flang, which takes too long).

We'll see if including mlir in this PR is too much for CI or not...


jeremyd2019:
🟢 https://github.com/msys2-arm/MINGW-packages/actions/runs/3300964591/jobs/5445974479

PR #13673, opened 2022-10-22
flang: update to 15.0.3

jeremyd2019:


mati865:
Funny how Clang managed to build it in GNU subsystems but failed in LLVM one...

PR #13687, opened 2022-10-24
CI: Move from deprecated set-output

MehdiChinoune:


naveen521kk (approved):
lgtm

naveen521kk:
Actually, it would be better to write it directly to `$GITHUB_OUTPUT` inside that script but this should also work I think.

MehdiChinoune:
> Actually, it would be better to write it directly to `$GITHUB_OUTPUT` inside that script but this should also work I think.

$GITHUB_OUTPUT is an environment file not an environment variable.

naveen521kk:
> $GITHUB_OUTPUT is an environment file not an environment variable.

Hmm, wondering what an "environment file" is. I thought `$GITHUB_OUTPUT` is an environment variable containing a path to a file which we can write to for setting the output.

naveen521kk:
By the way, `$GITHUB_OUTPUT` is an environment variable containing a path to a file, see https://github.com/msys2/MINGW-packages/pull/13710/.

MehdiChinoune:
> By the way, `$GITHUB_OUTPUT` is an environment variable containing a path to a file, see #13710.

Yes, That what I meant. You couldn't set it as `GITHUB_OUTPUT=`.

PR #13688, opened 2022-10-24
mcfgthread: Add `!buildflags` in options

lhmouse:
Recently, `-D_FORTIFY_SOURCE=2` has been added into the default CFLAGS. As thislibrary has to be built before the CRT, where those overflow checking routines have not been available, it is necessary to build without default flags.


lhmouse:
Thank you!

PR #13690, opened 2022-10-24
svt-av1: update to 1.3.0

kmilos:


Biswa96:
Does this require any packages to be rebuilt?

kmilos:
Hm, possibly, I just saw the ABI version bump here: https://gitlab.com/AOMediaCodec/SVT-AV1/-/compare/v1.2.1...v1.3.0?from_project_id=24327400&page=10&straight=false#fea033795c54df35effffa170a7b6e52541bfeea

lazka:
> Edit: Seems to be only one addition to the public struct, so might be safe to ignore?

that seems like an ABI break to me.

kmilos:
Ok, bumping deps...

PR #13693, opened 2022-10-24
Amtk: update and dep cleanup

swilmet:
Amtk: update and dep cleanup.


mingw-w64-amtk/PKGBUILD
```diff
@@ -3,7 +3,7 @@
 _realname=amtk
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=5.5.1
+pkgver=5.5.2
 pkgrel=2
```
Biswa96, line 7: `pkgrel` need to be reset to 1 in case of version update.

lazka:
Ah I missed that

PR #13696, opened 2022-10-25
mfem: update to 4.5

3rav:


Biswa96:
I just noticed that the DLL path was not fixed in upstream. So, I added the removed patch from here.

Issue #13698, opened 2022-10-25
kitemmodels qml plugin installed to the incorrect path


BLumia:
### Description

The qml plugin provided by `mingw-w64-x86_64-kitemmodels-qt5` installed to the incorrect path and file name.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Microsoft Windows [Version 10.0.22000.1098]

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Expected behavior

The QML plugin should be installed to:

-  `/mingw64/share/qt5/qml/org/kde/kitemmodels`
 
instead of:

- `/mingw64/lib/qml/org/kde/kitemmodels/`

Also, `libitemmodelsplugin.dll` should remove the `lib` prefix, as `itemmodelsplugin.dll`.

### Actual behavior

The qml plugin was installed to `/mingw64/lib/qml/org/kde/kitemmodels/` which will cause it cannot be found by apps that rely on it:
`module "org.kde.kitemmodels" is not installed`

The `lib` prefix of the dll will cause the plugin could not be loaded:
`module "org.kde.kitemmodels" plugin "itemmodelsplugin" not found` if we move the plugin dir to the correct place

### Repro steps

Just see the file list of this package: https://packages.msys2.org/package/mingw-w64-x86_64-kitemmodels-qt5

If you want to reproduce the issue manually, you can clone https://github.com/KDE/kirigami-gallery , build it and try to run it. Building can be successed, the error messages provided above will be shown if you try to run it.

### Are you willing to submit a PR?

Maybe yes if I know how to fix it correctly :)

BLumia:
> Just curious. How did you find that the patch should be in share directory?

I have a little experience developing a QML plugin for Linux and MSYS2 is uses very similar folder structure :)

A patch has been submitted, I only wrote some PKGBUILD for Arch Linux before and never wrote one for MSYS2 so please let me know if anything is wrong. btw the patch didn't fix the `lib` prefix issue, but it might be an upstream issue as well. I might take some time to investigate it tomorrow.

Biswa96:
I can reproduce the issue. With the 'lib' prefixed plugin DLL, I get the following error while configuring kirigami-gallery.

```
[...]

PR #13699, opened 2022-10-25
fix qml plugin from kitemmodels installed to incorrect dir

BLumia:
Resolve https://github.com/msys2/MINGW-packages/issues/13698

### DIFF
--- a/mingw-w64-kitemmodels-qt5/PKGBUILD
+++ b/mingw-w64-kitemmodels-qt5/PKGBUILD
@@ -21,13 +21,19 @@ else
             "${MINGW_PACKAGE_PREFIX}-qt5-declarative")
 fi
 groups=("${MINGW_PACKAGE_PREFIX}-kf5")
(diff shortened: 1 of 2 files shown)


Biswa96:

Biswa96:
Is this ready to merge?

Biswa96:
I did not see your merge request in upstream. Should we wait for that patch to be merged? We can import the patch from upstream in PKGBUILD. That may be better than my lazy workaround.

BLumia:
> I did not see your merge request in upstream.

https://invent.kde.org/frameworks/kitemmodels/-/merge_requests/49/diffs

> Should we wait for that patch to be merged? We can import the patch from upstream in PKGBUILD. That may be better than my lazy workaround.

Personally I think your patch is perfect for the current case, since we only care about MSYS2 package build, your patch just done the fix and didn't touch any other part of the project (won't affact existing tests, file generation and resource processing, etc.). If my upstreamed patch get merged then we can drop the current patch when we upgrade the package to that version. My 2cents :)

PR #13700, opened 2022-10-25
gtk4: bump to 4.8.2 and pick g-i 1.74 fixes

elmarco:


Biswa96:
Is it also possible to update gtk4? version 4.8.2 is reported as latest one https://packages.msys2.org/outofdate. Then we would not need pkgrel bump.

PR #13702, opened 2022-10-26
curl: Update to 7.86.0

lazka:


lazka:
https://github.com/curl/curl/pull/9521#issuecomment-1291810398

PR #13703, opened 2022-10-26
nghttp2: Update to 1.50.0

lazka:
the tools seem to depend on unix APIs, so disable them


lazka:
curl needs a rebuild when this gets merged

dscho:
It looks as if the `deflatehd` and `inflatehd` tools [are no longer being built](https://github.com/msys2/MINGW-packages/actions/runs/3330447184/jobs/5508945880#step:10:1536). This apparently removes the last users of jansson and jemalloc (see [here](https://github.com/git-for-windows/git-sdk-32/actions/runs/3342715301/jobs/5535217675#step:5:9) and [here](https://github.com/git-for-windows/git-sdk-64/actions/runs/3342704814/jobs/5535196728#step:5:9)).

@lazka maybe we should drop them from the `depends` array in `mingw-w64-nghttp2/PKGBUILD`?

lazka:
> It looks as if the `deflatehd` and `inflatehd` tools [are no longer being built](https://github.com/msys2/MINGW-packages/actions/runs/3330447184/jobs/5508945880#step:10:1536). This apparently removes the last users of jansson and jemalloc (see [here](https://github.com/git-for-windows/git-sdk-32/actions/runs/3342715301/jobs/5535217675#step:5:9) and [here](https://github.com/git-for-windows/git-sdk-64/actions/runs/3342704814/jobs/5535196728#step:5:9)).

yeah, I got lots of unix only compile errors with them, so it seemed like they are not Windows compatible. tbh I didn't look much into it.

thanks for the suggestion, noted!

dscho:
FWIW I was under the impression that those executables were used by nghttp2, but a quick `grep` on the binaries did not turn up any user...

lazka:
Removed the deps in https://github.com/msys2/MINGW-packages/pull/13789 and ported to cmake while at it.

PR #13707, opened 2022-10-27
Python influxdb: new package

wlph17:


Biswa96:
Is this package same as this repository https://github.com/influxdata/influxdb-python ? It has some dependencies.

wlph17:
yes, oh right

PR #13708, opened 2022-10-27
openblas: build with openmp support

lazka:
To match what Arch does.

### DIFF
--- a/mingw-w64-openblas/PKGBUILD
+++ b/mingw-w64-openblas/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-openblas
 pkgname=("${MINGW_PACKAGE_PREFIX}-openblas"
          $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-openblas64"))
 pkgver=0.3.21
-pkgrel=5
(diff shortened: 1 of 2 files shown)


lazka:
@szhorvat

szhorvat:
Yes, this is what I was suggesting. Based on my reading of https://github.com/xianyi/OpenBLAS/wiki/faq#how-can-i-use-openblas-in-multi-threaded-applications, if any other packages that depend on OpenBLAS use OpenMP, then this is necessary.

lazka:
thanks. I've also opened https://github.com/xianyi/OpenBLAS/issues/3798

jeremyd2019:
Openmp is not built for clangarm64.  Last I heard it was unsupported on windows/arm64

lazka:
Ah, missed that, thx

carlkl:
I can't find a reference for it, but AFAIK openblas used by scipy shouldn't be configured with OpenMP support.

lazka:
I get the same `13 failed, 37771 passed` for the scipy test suite with both openblas configs at least.

szhorvat:
I confirm that the hangs with igraph went away after this change.

lazka:
thanks for testing

PR #13711, opened 2022-10-27
kf5: update to 5.99.0 - Part1

MehdiChinoune:


Biswa96:
There was a discussion about waiting for next KDE update due to conflict with libavif. Should we wait for that?

MehdiChinoune:
> There was a discussion about waiting for next KDE update due to conflict with libavif. Should we wait for that?

Conflict of what?

MehdiChinoune:
We could backport the patch.

PR #13790, opened 2022-10-29
libtomcrypt libtommath: fix building

raedrizqie:


lazka:
This drops the .pc files. Is that a problem? Otherwise looks good

raedrizqie:
i think its ok without the .pc files.. the linkage is straight forward

raedrizqie:
thanks!

PR #13791, opened 2022-10-29
CI: test building with a larger pagefile

lazka:
It defaults to 1.8GB, bump it a bit so we can build
flang under clang64 which otherwise hits OOM

Use C: since we have 95GB free there and only 12GB on D:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,6 +41,14 @@ jobs:
           .\procdump64.exe -accepteula -ma -i "${{ runner.temp }}/_dumps"
           .\procdump.exe -accepteula -ma -i "${{ runner.temp }}/_dumps"
 
+      - name: Configure Pagefile
+        if: ${{ matrix.msystem != 'CLANGARM64' }}
+        uses: al-cheb/configure-pagefile-action@86589fd789a4de3e62ba628dda2cb10027b66d67
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -41,6 +41,14 @@ jobs:
           .\procdump64.exe -accepteula -ma -i "${{ runner.temp }}/_dumps"
           .\procdump.exe -accepteula -ma -i "${{ runner.temp }}/_dumps"
 
+      - name: Configure Pagefile
+        if: ${{ matrix.msystem != 'CLANGARM64' }}
+        uses: al-cheb/configure-pagefile-action@v1.3
+        with:
+          minimum-size: 8GB
+          maximum-size: 16GB
+          disk-root: "C:"
```
jeremyd2019, line 50: If there's enough room, I think `D:` is more performant, would be a better choice for pagefile.

lazka: We install MSYS2 and all packages there, and it only has 12GB, so I'm not sure

lazka:
Changed minimum from 8 to 4, to not take away space if we build qt-static or something which afair quite large.

pinned the action to a specific commit to avoid supply chain issues.

lazka:
Same for autobuild: https://github.com/msys2/msys2-autobuild/pull/66

PR #13800, opened 2022-10-30
onevpl: Add package

Biswa96:
Please review this package.

### DIFF
--- a/mingw-w64-onevpl/0002-cmake-install-vars-sh.patch
+++ b/mingw-w64-onevpl/0002-cmake-install-vars-sh.patch
@@ -1,0 +1,11 @@
+--- a/env/CMakeLists.txt
++++ b/env/CMakeLists.txt
+@@ -16,7 +16,7 @@
+ file(TO_NATIVE_PATH "${CMAKE_INSTALL_LIBDIR}" _install_libdir)
+ file(TO_NATIVE_PATH "${CMAKE_INSTALL_BINDIR}" _install_bindir)
+ 
+-if(WIN32)
++if(MSVC)
+   install(
+     PROGRAMS "cmd/vars.bat"
+     DESTINATION "${ONEAPI_INSTALL_ENVDIR}"
--- a/mingw-w64-onevpl/0002-cmake-install-vars-sh.patch
+++ b/mingw-w64-onevpl/0002-cmake-install-vars-sh.patch
@@ -1,11 +1,11 @@
 --- a/env/CMakeLists.txt
 +++ b/env/CMakeLists.txt
-@@ -16,7 +16,7 @@
- file(TO_NATIVE_PATH "${CMAKE_INSTALL_LIBDIR}" _install_libdir)
- file(TO_NATIVE_PATH "${CMAKE_INSTALL_BINDIR}" _install_bindir)
- 
--if(WIN32)
-+if(MSVC)
-   install(
-     PROGRAMS "cmd/vars.bat"
+@@ -33,7 +33,7 @@
+     PROGRAMS "${CMAKE_CURRENT_BINARY_DIR}/${ENV_SCRIPT_NAME}"
      DESTINATION "${ONEAPI_INSTALL_ENVDIR}"
+     COMPONENT dev)
+-else()
++
+   file(TO_CMAKE_PATH "${env_rel_prefix}" env_rel_prefix)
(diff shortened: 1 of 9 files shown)


mingw-w64-onevpl/0002-cmake-install-vars-sh.patch
```diff
@@ -0,0 +1,11 @@
+--- a/env/CMakeLists.txt
++++ b/env/CMakeLists.txt
+@@ -16,7 +16,7 @@
+ file(TO_NATIVE_PATH "${CMAKE_INSTALL_LIBDIR}" _install_libdir)
+ file(TO_NATIVE_PATH "${CMAKE_INSTALL_BINDIR}" _install_bindir)
+ 
+-if(WIN32)
++if(MSVC)
```
1480c1, line 8: I wonder if it would be better to install both vars.bat and vars.sh in case someone wanted to use their environment variables outside bash?

Biswa96: I am not sure if that would work with different install directories. But yeah, I can try to install both files.

Biswa96:
Question. It seems that the binaries oneapi-src/oneVPL repository requires other libraries also. Should other onevpl packages be added like AUR https://aur.archlinux.org/packages?O=0&K=onevpl ?

1480c1:
Based on https://github.com/oneapi-src/oneVPL-cpu#onevpl-architecture, seems we should 

> To use oneVPL for video processing you need to install at least one implementation. Here is a list of current implementations.
>    [oneVPL-cpu](https://github.com/oneapi-src/oneVPL-cpu) for use on CPU
>    [oneVPL-intel-gpu](https://github.com/oneapi-src/oneVPL-intel-gpu) for use on Intel Xe graphics and newer
 >   [Media SDK](https://github.com/Intel-Media-SDK/MediaSDK) for use on legacy Intel graphic

So we need to at least package one of the above, and I don't know if libmfx counts for the last one

Biswa96:
Now that take us into a rabbit hole. onevl-cpu searches for svt-av1, svt-hevc, dav1d and ffmpeg. So, there will be a circular dependency with onevpl-cpu and ffmpeg.

Also opevpl-cpu searches for those library in a horrible way. The repository is full of assumptions and workaround, see https://github.com/oneapi-src/oneVPL-cpu/blob/master/cpu/ext/ffmpeg-codecs/CMakeLists.txt

1480c1:
Hmm

All I can say is, I don't want to touch that with a 500km pole

Biswa96:
I have built the onevpl-cpu with minimal changes. Let me ask about those in upstream.

mingw-w64-onevpl/0002-cmake-install-vars-sh.patch
Biswa96: I have changed the patch file to install both shell and batch script.

Biswa96:
openvpl-cpu package has been added. The repository bundles its own ffmpeg, x264, dav1d, svt-av1 and svt-hevc with specific patch & build options. So, it would be appropriate to use those bundled dependencies instead of mingw packages.

@1480c1 Please review the new packages and check if those work in m-ab-s.

1480c1:
> @1480c1 Please review the new packages and check if those work in m-ab-s.

I will test them once I get enough time

1480c1:
@Biswa96 I tested compilation and running with vp9_qsv on decoding and encoding, and it seemed to be working

image

According to https://trac.ffmpeg.org/wiki/Hardware/QuickSync and https://www.intel.com/content/www/us/en/products/sku/213796/intel-xeon-w11955m-processor-24m-cache-2-60-ghz/specifications.html, I should be able to do av1 decode, but was unable to. Maybe something to report upstream. I was also unable to tell if av1 encode worked. I do not think the -cpu backend was actually working.

Biswa96:
Can you provide the steps to test the packages? I have older CPU. I am not sure if the packages will work in my system but I can try.

1480c1:
barebones
```bash
./ffmpeg -y -i input.mkv -c:v vp9_qsv -c:a copy out.webm
./ffmpeg -y -c:v av1_qsv -i input.mkv -c:v vp9_qsv -c:a copy out.webm
```
stuff like that

Biswa96:
Both commands fails in my pc. But the ffmpeg can find and link with libvpl. These packages will help to start the ground work.

PR #13807, opened 2022-10-31
Create a virtual package for OpenMP - Part1

MehdiChinoune:
Also provides a virtual package for Fortran Compiler "fc".

I have chosen `omp` because `openmp` is already the name of LLVM OpenMP.
Should we create a separate package `libgomp` or leave it as it is and make `gcc-libs` provides `omp`.


lazka:
> Also provides a virtual package for Fortran Compiler "fc".

lgtm
 
> I have chosen `omp` because `openmp` is already the name of LLVM OpenMP. Should we create a separate package `libgomp` or leave it as it is and make `gcc-libs` provides `omp`.

"omp" seems fine to me. The C pragmas are named "omp" and the env vars start with "OMP_", so this seems like a good match, thanks.

lazka (approved):

PR #13810, opened 2022-10-31
nghttp2: restore the old DLL name

lazka:
it's easy enough


dscho:
> it's easy enough

... but every time the DLL name is changed, I have to do unplanned follow-up work... it's fine to change it, of course, but can we please leave it now? And if it is changed to the unversioned file name again, leave it there?

lazka:
Someone complained on discord that the name change broke git4windows, so I changed it back here :/

dscho:
I had fixed Git for Windows. Now I have to fix it again.

dscho:
Besides, those people who complain about an _MSYS2_ update breaking Git for Windows: please do not lend them your ear. It is [documented _very well_](https://github.com/git-for-windows/git/wiki/Install-inside-MSYS2-proper) that they are using an unsupported scenario and are free to do that as long as they don't complain about breakages. Gosh, they're such a nuisance.

lazka:
While we are here, do you have any opinions on soversion in DLL name vs none, in general?

dscho:
> I'll try to keep that in mind.

Thank you. They really do not contribute anything but more work for me, it seems.

> do you have any opinions on soversion in DLL name vs none, in general?

@lazka I sort of do: for properly SemVer'ed projects, it might make sense to keep the major version encoded in the DLL name, but for other projects less so.

But more important than that question in my mind is what disruption is caused by changing the status quo. I do not have anything against ripping out the version name from `libnghttp2.dll`, as long as it is done once. That's a one-time cost, that's manageable. I also have nothing against keeping the major version encoded in the name, that's a once-per-major-version-bump cost. Also manageable, if a bit more involved (due to the need to rebuild all dependencees).

lazka:
Rebuilds are needed on ABI breaks/major bumps anyway, so the "cost" shouldn't change with different naming (if you ignore switching build systems, like we did here)

I'll try to propose this upstream.. let's see what they say.

PR #13811, opened 2022-10-31
glew: Add doc and improve glew.pc

podsvirov:


mingw-w64-glew/PKGBUILD
```diff
@@ -54,6 +54,13 @@ package () {
     -e "s|@requireslib@||g" \
     ${srcdir}/${_realname}-${pkgver}/glew.pc.in \
     > ${pkgdir}${MINGW_PREFIX}/lib/pkgconfig/glew.pc
+  # For building a static library or executable on Windows.
+  # For more info see https://glew.sourceforge.net/install.html.
+  echo "Cflags.private: -DGLEW_BUILD=GLEW_STATIC" \
```
lazka, line 59: This doesn't look correct: https://github.com/nigels-com/glew/blob/80e061f85d2f92d8721959b1508accef50bb8e1c/auto/src/wglew_head.h#L19 they are two different macros

podsvirov: > This doesn't look correct: https://github.com/nigels-com/glew/blob/80e061f85d2f92d8721959b1508accef50bb8e1c/auto/src/wglew_head.h#L19 They are two different macros

@lazka, your right, just `-DGLEW_STATIC`. Fixed.

PR #13824, opened 2022-11-01
lcms2: update to 2.14

kmilos:


kmilos:
Hm, not sure what's going on here w/ CLang since the last release - `-fvisibility=hidden` is also present for GCC but it works out there?!

lazka:
hm, looks like clang defaults to not exporting anything while gcc exports all. and the .def file isn't used.

kmilos:
But `-fvisibility=hidden` is defined for GCC builds as well?

kmilos:
I don't think we ever defined `CMS_BUILD_DLL` so MINGW32/CLANG32 shouldn't be using stdcall anyway.

Also, the .pc file doesn't seem to add -DCMS_DLL either, so all MSYS2 dependencies should've used the default visibility rules and calling convention when linking to lcms2 anyway up to now...

mmuetzel:
> But `-fvisibility=hidden` is defined for GCC builds as well?

Afaict, `-fvisibility=hidden` doesn't do anything for GCC targeting Windows. It's only a pretty recent addition to Clang that `visibility` attributes (and `-fvisibility=hidden` I guess) are taken into account for Windows targets.
GCC only supports the `dllexport`/`dllimport` attributes when it comes to "visibility" of symbols. If there no `dllexport` attributes in the sources it compiles, it exports everything.

kmilos:
Thanks @mmuetzel, makes sense.

mmuetzel:
> I did look at release notes for Clang 14 and 15 but couldn't find a reference...

That's because it wasn't released yet. 🙂
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-clang/0009-Use-hidden-visibility-when-building-for-MinGW-with-Clang.patch

lazka:
> Afaict, `-fvisibility=hidden` doesn't do anything for GCC targeting Windows. It's only a pretty recent addition to Clang that `visibility` attributes (and `-fvisibility=hidden` I guess) are taken into account for Windows targets.

Oh, I somehow missed that this applies non-clang-internal code as well. That's a bit annoying...

kmilos:
@lazka So how do we proceed on this PR?

Issue #13799, opened 2022-10-30
Add kirigami-addons-qt5


BLumia:
### Package name

kirigami-addons-qt5

### Brief description of package

Add-ons for the Kirigami framework.

### URL for package's homepage

https://invent.kde.org/libraries/kirigami-addons

### Provide a basic test case to validate the package's functionality.

Use `qml` to run any qml file under `tests/` in the source tree.

### MINGW environments where you need the package

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

Maybe? I have a working PKGBUILD currently but I'm not sure if I did it correctly.

BLumia:
The PKGBUILD I currently use (modified from the PKGBUILD file for `kirigami2-qt5`):

```bash
# Maintainer (MSYS2): Michael Hansen <zrax0111@gmail.com>
# Contributor: Zach Bacon <zachbacon@vba-m.com>

_variant=-${KF5_VARIANT:-shared}
source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
_kde_f5_init_package "${_variant}" "kirigami-addons"
pkgver=0.5
pkgrel=1
arch=('any')
mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
pkgdesc="Add-ons for the Kirigami framework (mingw-w64)"
license=('GPL' 'LGPL')
makedepends=("${MINGW_PACKAGE_PREFIX}-extra-cmake-modules>=${pkgver}"
             "${MINGW_PACKAGE_PREFIX}-ninja"
             "${MINGW_PACKAGE_PREFIX}-ki18n-qt5"
             "${MINGW_PACKAGE_PREFIX}-cc")
depends=()
if [ "${_variant}" = "-static" ]; then
  _kde_f5_add_depends "${_variant}" "${MINGW_PACKAGE_PREFIX}-qt5${_namesuff}"
else
  makedepends+=("${MINGW_PACKAGE_PREFIX}-qt5-tools")
  depends+=("${MINGW_PACKAGE_PREFIX}-qt5-base"
            "${MINGW_PACKAGE_PREFIX}-qt5-declarative"
            "${MINGW_PACKAGE_PREFIX}-qt5-quickcontrols2"
[...]

PR #13833, opened 2022-11-02
kirigami-addons: New package

BLumia:
Resolve https://github.com/msys2/MINGW-packages/issues/13799

BTW, should I replace the "Maintainer" comment header to myself? If the answer is yes, then what's the policy to update the package? Do I need to follow any extra setup (e.g. email to notify outdated?) to maintain this package?

### DIFF
--- a/mingw-w64-kirigami-addons-qt5/PKGBUILD
+++ b/mingw-w64-kirigami-addons-qt5/PKGBUILD
@@ -1,0 +1,59 @@
+# Maintainer (MSYS2): Gary Wang <wzc782970009@gmail.com>
+# Contributor: Zach Bacon <zachbacon@vba-m.com>
+
+_variant=-${KF5_VARIANT:-shared}
+source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
+_kde_f5_init_package "${_variant}" "kirigami-addons"
+pkgver=0.5
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+pkgdesc="Add-ons for the Kirigami framework (mingw-w64)"
+license=('GPL' 'LGPL')
+makedepends=("${MINGW_PACKAGE_PREFIX}-extra-cmake-modules>=${pkgver}"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-ki18n-qt5"
+             "${MINGW_PACKAGE_PREFIX}-cc")
+depends=(${MINGW_PACKAGE_PREFIX}-kirigami2-qt5)
+if [ "${_variant}" = "-static" ]; then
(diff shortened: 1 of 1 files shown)


Biswa96:
You can replace the maintainer tag to yourself. There is no hardcore policy.

mingw-w64-kirigami-addons-qt5/PKGBUILD
```diff
@@ -0,0 +1,62 @@
...
+  _kde_f5_add_depends "${_variant}" "${MINGW_PACKAGE_PREFIX}-qt5${_namesuff}"
+else
+  makedepends+=("${MINGW_PACKAGE_PREFIX}-qt5-tools")
+  depends+=("${MINGW_PACKAGE_PREFIX}-qt5-base"
+            "${MINGW_PACKAGE_PREFIX}-qt5-declarative"
+            "${MINGW_PACKAGE_PREFIX}-qt5-quickcontrols2"
+            "${MINGW_PACKAGE_PREFIX}-qt5-svg")
+fi
+groups=("${MINGW_PACKAGE_PREFIX}-kf5")
+source=("https://download.kde.org/stable//${_basename}/${_basename}-${pkgver}.tar.xz"{,.sig})
```
Biswa96, line 28:
```suggestion
source=("https://download.kde.org/stable/${_basename}/${_basename}-${pkgver}.tar.xz"{,.sig})
```

BLumia: Patch updated :)

Biswa96:
@lazka Is it possible to save and use the PGP keys in the package folder like ArchLinux does?

lazka:
Sounds good to me. I don't know how that works though (does makepkg handle it? or do we need to expand https://github.com/msys2/MINGW-packages/blob/master/.ci/fetch-validpgpkeys.sh to import files from a local directory?)

BLumia:
I'm actually taking [kitemmodels' PKGBUILD](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-kitemmodels-qt5/PKGBUILD) as a reference and see that file is trying to validate the PGP key. Not sure why that one is successed but this one failed. Nether kitemmodels' nor this one can be validate on my machine, tho.

Should I remove the PGP validation part from this patch?

Biswa96:
> Should I remove the PGP validation part from this patch?

Yes, please. You can just remove the `.sig` part from the `source` variable and then `updpkgsums`.

BLumia:
Patch updated :)

Biswa96:
kirigami-addons needs kirigami2, see https://github.com/archlinux/svntogit-packages/blob/packages/kirigami-addons/trunk/PKGBUILD

BLumia:
Sorry I should pay more attention to CI result before :joy: Should be fixed now.

btw, is there a way to test if a PKGBUILD is missing required `depends` locally without uninstalling the packages in my current working environment?

Biswa96:
As far as I know, `ntldd` can show the dependent DLLs but one has to figure out the dependencies manually. `ntldd` output is somewhat verbose.

PR #13834, opened 2022-11-02
make[critical]: Fix bad timestamps on mingw32

orgads:
Fixes a critical regression between 4.3 and 4.4.


Biswa96:
Did you asked upstream about this patch? If yes, can you add a comment with url of that upstream commit or thread? It'd help to track the patch in future.

lazka:
lgtm

orgads:
https://lists.gnu.org/archive/html/bug-make/2022-11/msg00017.html

orgads:
I'll add it as a comment in PKGBUILD.

orgads:
Done

lazka:
in the repo now, thanks

PR #13918, opened 2022-11-04
highlight: Make program relocatable

legends2k:
#13184 missed making the program relocatable.

MSYS2 can be installed anywhere by a user so having a fixed build prefix won’t
work.  Highlight has config files, language syntax, plugins and themes spread
across `$PREFIX/share`, `$PREFIX/etc`.  With this fix, highlight looks for these
directories relative to its own executable’s directory.


legends2k:
@Biswa96 This has been in _wait_ state for very long without any issues.  Would you be kind enough to inspect and merge this if there're no issues?

Biswa96:
Please either rebase the branch or open a new pull request.

legends2k:
Rebased; please check now.

PR #13919, opened 2022-11-04
Levenshtein: add dependence

wlph17:
it need rapidfuzz to run


Biswa96:
It would be better if this commit is added in the python-rapidfuzz PR. CI can build both at one run.

wlph17:
ok

PR #13920, opened 2022-11-04
New package: Influxdb

wlph17:


wlph17:
I think this is ready to merge, thank you.

wlph17:
I think this is ready to merge, thank you.

lazka:
Why the old/unsupported version?

wlph17:
image
It is because when I build influxdbV2 it runs into this problem, I don't know why running a bash script is not possible is GNUmakefile

wlph17:
I think what it does is change pkg-config to a custom "pkg-config.sh", however when make call "pkg-config.sh", it returns %1 is not a valid Win32 application. And I tried some solutions but they don't work.

wlph17:
It could also because some dependencies are missing (https://github.com/influxdata/influxdb/issues/22986), but I included all that msys2 have.

PR #13923, opened 2022-11-04
mcfgthread: Upgrade to 1.3

lhmouse:
This is the required version for building GCC 13 with Objective-C enabled. GCC 13 will support `--enable-threads=mcf` once it is released. Support for LLVM's libc++ has not been tested.

The return value of `std::recursive_mutex::try_lock()` has been fixed.

Reference: https://github.com/lhmouse/mcfgthread/releases/tag/v1.3-ga.1


lhmouse:
Thank you.

PR #13928, opened 2022-11-05
7zip: fix clangarm64 build

jeremyd2019:
Add a temporary patch to change ARM64 target attributes to a format clang understands (clang should support the GCC format as of version 16, see https://github.com/msys2/MINGW-packages/pull/13894#issuecomment-1304319971).

Add another `-Wno-error` option, rather than trying to patch out `force_align_arg_pointer` attribute for non-x86.

### DIFF
--- a/mingw-w64-7zip/clang-arm-target-fixes.patch
+++ b/mingw-w64-7zip/clang-arm-target-fixes.patch
@@ -1,0 +1,54 @@
+--- C/7zCrc.c	2021-04-01 18:00:00.000000000 +0000
++++ C/7zCrc.c	2022-11-05 05:16:33.444743500 +0000
+@@ -86,7 +86,7 @@
+       #if !defined(__ARM_FEATURE_CRC32)
+         #define __ARM_FEATURE_CRC32 1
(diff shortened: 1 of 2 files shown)


mingw-w64-7zip/clang-arm-target-fixes.patch
```diff
@@ -0,0 +1,54 @@
...
+     #define ATTRIB_SHA __attribute__((__target__("fpu=crypto-neon-fp-armv8")))
+   #endif
+@@ -247,6 +247,10 @@
+ #if defined(_MSC_VER) && defined(MY_CPU_ARM64)
+ #include <arm64_neon.h>
+ #else
++
++#define __ARM_FEATURE_AES
++#define __ARM_FEATURE_SHA2
++#define __ARM_FEATURE_CRYPTO
```
jeremyd2019, line 30: Oops, these defines (or probably defining them to `1` instead of empty) probably need to stick around beyond clang 16.  If that turns out to be the case, they should probably be proposed upstream (like the similar define in 7zCrc.c ☝️ )

PR #13932, opened 2022-11-05
octave: update to version 7.3.0

mmuetzel:
PCRE isn't actually a new dependency. But it looks like it was pulled in implicitly before. But that no longer seems to be happening.


MehdiChinoune:
Is there a way to use pcre2 instead of pcre?

mmuetzel:
> Is there a way to use pcre2 instead of pcre?

There probably is. But it hasn't been implemented yet.
See this upstream bug report:
https://savannah.gnu.org/bugs/index.php?61542

MehdiChinoune:
Fedora is going to drop pcre soon.
https://fedoraproject.org/wiki/PcreDeprecation
https://fedoraproject.org/wiki/PcreRetirement

mmuetzel:
> Fedora is going to drop pcre soon.
> https://fedoraproject.org/wiki/PcreDeprecation
> https://fedoraproject.org/wiki/PcreRetirement

Yeah. I know. 😓
Transitioning would probably not be that hard. We'd just need someone to actually implement it...

mmuetzel:
Fwiw, support for PCRE2 has been added recently in Octave:
https://hg.savannah.gnu.org/hgweb/octave/rev/d1165473e4b0

It will be part of Octave 8 when it will be released probably at some point early next year.

MehdiChinoune:
> Fwiw, support for PCRE2 has been added recently in Octave: https://hg.savannah.gnu.org/hgweb/octave/rev/d1165473e4b0
> 
> It will be part of Octave 8 when it will be released probably at some point early next year.

Could it backported?

mmuetzel:
> Could it backported?

Maybe. I'd rather not without the necessary testing.
Is it important?

MehdiChinoune:
> Is it important?

Just to know if we could drop pcre earlier.

PR #13941, opened 2022-11-06
pyqt5: disable building tools on 32bit envs

MehdiChinoune:
For some unknown reason sip hit an assertion when generating binding for pylupdate.

### DIFF
--- a/mingw-w64-pyqt5/PKGBUILD
+++ b/mingw-w64-pyqt5/PKGBUILD
@@ -9,7 +9,7 @@ conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}"
 replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}"
           "${MINGW_PACKAGE_PREFIX}-${_realname}-common")
 pkgver=5.15.7
-pkgrel=2
+pkgrel=3
 pkgdesc="Qt5 bindings for Python (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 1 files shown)


Biswa96:
The assertion comes from this code in sip-6.7.4/code_generator/heap.c

```c
h = malloc(n);
assert(h != NULL);
```

In which cases malloc fails in 32bit environment?

lazka:
hm, I wonder if we should compile the setuptools launcher with " -Wl,--large-address-aware", so sip can allocate more memory.

mati865:
FYI Rust has been building everything wth large address for years: https://github.com/rust-lang/rust/commit/6b130e3dd9547e233dca2cfd72a5968891672d9c

jeremyd2019:
I thought the setuptools launcher launches rather than embeds python, so it really shouldn't be using very much memory itself.  But I also thought large address aware was in the default CFLAGS for 32-bit? (or is that built with `!buildflags`)

lazka:
> I thought the setuptools launcher launches rather than embeds python, so it really shouldn't be using very much memory itself.

ah, right, so python itself.

> But I also thought large address aware was in the default CFLAGS for 32-bit? (or is that built with `!buildflags`)

that's also an option.. I would mind giving that a try

jeremyd2019:
For a quick test you could use `peflags` or `genpeimg` to set the large address aware flag on an existing executable, rather than waiting for a whole rebuild of it.

lazka:
> For a quick test you could use `peflags` (from `rebase`) or `genpeimg` (from `mingw-w64-tools`) to set the large address aware flag on an existing executable, rather than waiting for a whole rebuild of it.

thanks. I can confirm this fixes things.

jeremyd2019:
Can this PR be closed now as the failure it was trying to work around has been fixed?  Or is the removal of `--concatenate 2` option also significant?

MehdiChinoune:
> Or is the removal of `--concatenate 2` option also significant?

It is not.

lazka:
> > But I also thought large address aware was in the default CFLAGS for 32-bit? (or is that built with `!buildflags`)
> 
> that's also an option.. I would mind giving that a try (I'd assume code dealing with 64bit ASLR can also deal with that?)

I've created https://github.com/msys2/MSYS2-packages/issues/3283 for this

PR #13942, opened 2022-11-06
sundials: upgrade to 6.4.1

gharveymn:
### Notes

- Removed the petsc-build makedepend since SUNDIALS can now build with the normal library. This required a small patch to be able to find our renamed PETSc pkg-config files. Related config file patching has also been removed due to the aforementioned removal. 
There was another route where I could have continued to use petsc-build, however that package's pkg-config files are broken, so this seemed easier (however, I will still open another PR with a fix for that issue).
- It seems like most of the patches have been upstreamed in some way or another. They have therefore been removed.
- Removed `ENABLE_CALIPER` and `EXAMPLES_ENABLE_CUDA` since they were emitting warnings and are `OFF` by default anyway.

CC @mmuetzel


mmuetzel:
I installed the build artifact for MINGW64 from this PR and re-ran the test suite of Octave with it.
There are no new failing tests.

The changes look good to me. 👍

PR #13952, opened 2022-11-08
libclc: update to 15.0.4

jeremyd2019:
This package was out of sync with the other llvm-project packages.  This brings it back up to date, and makes the format of the PKGBUILD a little closer.

I don't use libclc, and don't have a way to know if it works or not... It builds (but with a lot of warnings!)


Biswa96:
> I don't use libclc, and don't have a way to know if it works or not... It builds

Me too 😔 

> but with a lot of warnings!

As this is part of llvm project, I think those should be fixed by llvm project also.

Biswa96 (approved):

PR #13955, opened 2022-11-08
python-lsp-server: Reallow newlest depends

podsvirov:
For example, version of the `autopep8` module currently is `2.0.0`.


Biswa96:
Is it possible not to change the patch file name? It would be easier to review changes.

podsvirov:
> Is it possible not to change the patch file name? It would be easier to review changes.

Ok. Done.

Please, review again.

MehdiChinoune:
Those dependencies don't get checked at runtime, Why changing them?

podsvirov:
> Those dependencies don't get checked at runtime, Why changing them?

But in my experience, those dependencies checked at runtime.

Firstly merge #13956, then try to launch `spyder`.

raedrizqie:
> But in my experience, those dependencies checked at runtime.

yes

PR #13956, opened 2022-11-08
python-spyder: Restore shebang line

podsvirov:
To launch `spyder` the #13955  will be useful too.


podsvirov:
After #13938, shebang line contains temporary path that looks like:

```shell
#!D:/a/msys64/${MSYS_ARCH}/bin/pythonw.exe
```

MehdiChinoune:
I think the recommended way is to remove the shebang line.

podsvirov:
@lazka, thanks for fix and merge!

PR #14030, opened 2022-11-10
qt-installer-framework: update to 4.5.0

MehdiChinoune:


lazka:
thanks, basics seem to work. For some reason it skips the "Installation Folder" page initially, not sure why yet.

lazka:
found a fix, so all good :)

PR #14036, opened 2022-11-11
Build Rust in Clang32 subsystem

mati865:
Fixes https://github.com/msys2/MINGW-packages/issues/9091

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -12,10 +12,10 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-rust-docs")
 pkgver=1.65.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Systems programming language focused on safety, speed and concurrency (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
 url="https://www.rust-lang.org/"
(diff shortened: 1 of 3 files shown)


mati865:
With this fix we can drop workarounds like https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-cryptography3/PKGBUILD

mingw-w64-rust/PKGBUILD
```diff
@@ -145,6 +151,9 @@ build() {
   if [ "${_bootstrapping}" = "no" ]; then
     _rust_conf+=("--local-rust-root=$(cygpath -m ${MINGW_PREFIX})")
   fi
+  if [[ $MINGW_PACKAGE_PREFIX != *-clang-i686 ]]; then
+    _rust_conf+=("--enable-vendor")
```
jeremyd2019, line 155: Took me a second... you are disabling `--enable-vendor` for clang32 because you added a new dependency (for the manifest), which would not be present in the vendored sources?  Is that something that's going upstream, so this is a temporary workaround?

jeremyd2019:
I had a script for enabling things for clang32.  Haven't run it in a long time, but I can remove the explicit blacklisting of rust, and see what happens. (I should probably run it once before and see what gets enabled, and if whatever actually builds for clang32).

jeremyd2019:
Do you plan to create `i686-pc-windows-gnullvm` target now, so it can propagate through the 'ecosystem' and eventually all clang* envs can switch to `*-gnullvm`?

mingw-w64-rust/PKGBUILD
mati865: Yeah, exactly,

mati865:
> Do you plan to create `i686-pc-windows-gnullvm` target now, so it can propagate through the 'ecosystem' and eventually all clang* envs can switch to `*-gnullvm`?

Eventually yes but there are still issues with running Rust's testsuite in Clang32 so it might require more fixes.

PR #14050, opened 2022-11-12
Add conflicts/replaces for old non-rust clang32 packages

lazka:
so users that have then installed get transferred to the new ones


lazka:
First we need to remove the old packages from the DB, so we don't get "replaces" cycles.

PR #14052, opened 2022-11-12
python-pallets-sphinx-themes: New package

podsvirov:


Biswa96:
It'll be better if package name is as same as in ArchLinux https://github.com/archlinux/svntogit-community/blob/packages/python-pallets-sphinx-themes/trunk/PKGBUILD

podsvirov:
Right note, renamed!

PR #14055, opened 2022-11-13
arm-none-eabi-*: enable clang32

jeremyd2019:
I noticed these did not have mingw32 either, so I'm opening a pull request rather than just enabling clang32 like most, in case there was a good reason not to have 32-bit support (it seems riscv64 didn't have any issues building for clang32 either, but my script was able to enable it (I guess it was still set up for bootstrapping))


MehdiChinoune:
see #13444

jeremyd2019:
OK.  It looks like arm-none-eabi-newlib was able to finish in time.  I don't know if -gcc would on its own though.  I don't have any reason for having it built for clang32 though, just noticed that my script wasn't enabling it (probably due to the circular dependency between gcc and newlib).

PR #14074, opened 2022-11-13
python-jinja: Add documentation

podsvirov:
Depends:
- [x] #14064
- [x] #14065


podsvirov:
Docs for [click](https://github.com/msys2/MINGW-packages/pull/14076) now added. What about docs for `jinja`?

PR #14076, opened 2022-11-13
python-click: Add documentation

podsvirov:
Depends:
- [x] #14064
- [x] #14065
- [x] #14075

### DIFF
--- a/mingw-w64-python-click/PKGBUILD
+++ b/mingw-w64-python-click/PKGBUILD
@@ -4,7 +4,7 @@ _realname=click
 pkgbase=mingw-w64-python-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
 pkgver=8.1.3
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


Biswa96:
Just curious. Does anyone read HTML documentation from packages? Most of the python modules have readthedocs website.

podsvirov:
I am reading. :smile: 📖

brisingraerowing:
Offline docs are useful (I use them sometimes), but putting them in a separate package (like is done with SQLite) might be preferable (some people might not want them cluttering up their filesystem).

podsvirov:
> Offline docs are useful (I use them sometimes), but putting them in a separate package (like is done with SQLite) might be preferable (some people might not want them cluttering up their filesystem).

Please see https://github.com/msys2/MINGW-packages/pull/6336#issuecomment-607908235.

lazka:
> Please see [#6336 (comment)](https://github.com/msys2/MINGW-packages/pull/6336#issuecomment-607908235).

We are less strict nowadays, with having more automation in place. It's still easier for everyone involved to not split though.

PR #14077, opened 2022-11-13
enable usage of rav1e for clang32 now that it's available

jeremyd2019:
Unfortunately, rav1e is still not available for clangarm64 because it makedepends on cargo-c, which has crates that it depends on that still depend on versions of other crates that don't yet support the new -gnullvm targets.


jeremyd2019:
```
      libheif-1.13.0.tar.gz ... Passed
      001-pkgconfig-match-autotools.patch ... FAILED
      002-cmake-enable-gdk-pixbuf.patch ... FAILED
```

I should download the source package for libheif, and compare the patches from that to the patches that get downloaded now and see what the changes are (if they're just cosmetic, or substantive)

jeremyd2019:
Cosmetic: just added a digit to the git commit hashes.  Will update checksums

PR #14079, opened 2022-11-14
ldt: New package

rmojab63:


lazka:
hm, with no stars and no website it would be nice have some kind of pitch why you think this should be added and why it is useful to other users.

rmojab63:
Hi. Currently, this is the external part of an R package and I was hoping to distribute it here. This is not a common approach for developing an R package, but I think it has much potential and is more flexible than other approaches, at least for Windows users (Please note that this can be a complex task; e.g., see "Finesse Library Dependencies" section in https://cran.r-project.org/web/packages/Rcpp/vignettes/Rcpp-libraries.pdf).

Of course, this is not all about a specific R package. This library is for statistical analysis and can search and sort large sets of regression models. It is capable of automatic time-series prediction, which is a common need in many applications. I don't want it to be specific to R users (and with a GPL license).

PR #14083, opened 2022-11-15
qemu: Update to 7.2.0

heljkon:


heljkon:
Please have a look!

jeremyd2019:
(I don't have a gitlab account or I'd probably ask there instead)

Did the patch referenced as being queued for 7.2 in https://gitlab.com/qemu-project/qemu/-/issues/513 ever actually get applied?  I looked at their git a while back and didn't see it, but there were a couple more RCs since I did.  I'd probably still stick with the combined bios mode anyway for my existing install for fear of confusing Windows though.

heljkon:
> (I don't have a gitlab account or I'd probably ask there instead)
> 
> Did the patch referenced as being queued for 7.2 in https://gitlab.com/qemu-project/qemu/-/issues/513 ever actually get applied? I looked at their git a while back and didn't see it, but there were a couple more RCs since I did. I'd probably still stick with the combined bios mode anyway for my existing install for fear of confusing Windows though.

It wasn't applied.
There was a discussion about the vars still being readonly and it turned out there was no way to make it work. 
I am not sure if it was forgotten or dropped. I'll ask for it via https://gitlab.com/qemu-project/qemu/-/issues/513

MehdiChinoune:
Could be qemu build against pcre2 instead of pcre.

heljkon:
> Could be qemu build against pcre2 instead of pcre.

I didn't find any issues using pcre2 instead of pcre after performing a few tests without pcre.
I didn't find any direct references to pcre-dlls in QEMU binaries. Both partial builds (locally performed) look identically with and without pcre.

QEMU sources reference both dependencies, but as long as we do not intend to create an installer package for qemu-guest-agent, which references a pcre-dll by name, we can try it.

Would it be sufficient to replace it with QEMU 8.0.0, because I usally perform many tests on the QEMU-RCs before release?

MehdiChinoune:
I see Arch building qemu 7.1.1 against pcre2

heljkon:
> I see Arch building qemu 7.1.1 against pcre2

Yes, I guess everything will be fine with pcre2 instead of pcre. My tests with Msys2/Mingw64 confirm this, too.

If you want to modify it now, fine. Otherwise, I will replace pcre with release of QEMU 8.0.0 in March/April 2023

Issue #3298, opened 2018-01-19
[mingw-w64-make] Is it possible to bring back Guile support?


mbakhterev:
It is much more easier and efficient to automate complex build processes with embedded Guile procedures than with GNU Make's own internal language. Is it possible to enable Guile support again?

irakhlin:
I don't believe that guile currently builds for mingw64 due to a long standing issue with the guile package itself. https://github.com/Alexpux/MINGW-packages/issues/699

Alexpux:
@irakhlin try to build both 32 and 64 yourself

vzh:
Hi there,
There was reported on the guile-user mailing list that the following branch should work:
http://git.savannah.gnu.org/cgit/guile.git/log/?h=wip-mingw-guile-2.2
There are several patches for MinGW there.

Alexpux:
@vzh I will not build package from git, only release versions

vzh:
OK, thank you! I'll ask guile devs if they have any plans on releasing it.

spk121:
@vzh @Alexpux Yeah, that mingw guile git branch is mine, so I suppose it is up to me to see if I can get those patches past the maintainers and into mainline.  I'll see what I can do.

gjanssens:
@spk121 Have you made some progress in mainlining your patches ? It would really be helpful to have a mingw guile package.

spk121:
The upcoming 3.0.6 release should support MinGW out of the box at least at some level. The unthreaded, 32-bit build will definitely work, though it won't pass the test suite 100% due to missing features. We're hoping to get the 64-bit build and the threaded build fixed by 3.0.6 as well, but, it is TBD.

alex-tee:
Guile 3.0.7 is out. is it possible to re-add guile support with this release?

Biswa96:
Are we asking for mingw-w64-guile package here? Or somehow adding guile in make 😕

alex-tee:
I personally just want libguile (so a mingw-w64-guile package), but the guile base package is also stuck at version 2 and could use an update (if possible)

jcalve:
I tried to compile 3.0.7, master, and a branch named "wip-mingw", the furthest I got was on the later with `./configure --enable-mini-gmp --disable-networking` but it still fails after a while...

Biswa96:
[...]

PR #14091, opened 2022-11-15
guile1.8: new package (lightweight guile version, 32 bit only)

h-c-mueller:
A lightweight guile version for MINGW32 and CLANG32. Currently even the only version available for MINGW. Especially useful for linking against libguile. If MSYS2 is also installed, no need to set GUILE_LOAD_PATH, ref. patch 0017. Fixes for localtime() on CLANG32, ref. patch 0018.

### DIFF
--- a/mingw-w64-guile1.8/0003-Fix-the-SRFI-60-copy-bit-documentation.patch
+++ b/mingw-w64-guile1.8/0003-Fix-the-SRFI-60-copy-bit-documentation.patch
@@ -1,0 +1,31 @@
+From 45f900bd1974431bbae64538e0994290d051a5a2 Mon Sep 17 00:00:00 2001
+From: Rob Browning <rlb@defaultvalue.org>
+Date: Sat, 23 Apr 2011 14:57:49 -0500
+Subject: Fix the SRFI 60 copy-bit documentation.
+
+---
+ srfi/srfi-60.c | 4 ++--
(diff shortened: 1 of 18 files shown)


lazka:
Just wondering, what would this be used/useful for?

h-c-mueller:
Simply speaking, because there is a similar open request for it, ref. #3298.

There have been numerous efforts to bring back guile support for `mingw`. But none version after 1.8.8 was able to meet this request. I assume it will still take a while, until we have back guile support with a contemporary version including 64 bit support. Until that, it's appreciated to have at least available what is working with `mingw`. Therefore, I used package name `guile1.8` to not spoil the name `guile` for a later version and related packaging attempts.
Furthermore, guile 1.8 is much more lightweight than e.g. guile 3.0, but sufficient to provide linking to `libguile`.

lazka:
There is someone working on guile mingw support: https://twitter.com/MikeSpike3/status/1673727567080329216

I don't know how promising that is though.

Biswa96:
That Twitter is not shown without logging in.

h-c-mueller:
Is it maybe related to
https://github.com/spk121/guile

Biswapriyo Nath ***@***.***> schrieb am Sa., 1. Juli 2023,
09:14:

Kreijstal:
how hard is this to port to 64bit

PR #14098, opened 2022-11-16
cmake: update to 3.25.0

MehdiChinoune:


raedrizqie:
emacs now exist for clang*

MehdiChinoune:
> emacs now exist for clang*

Yes, I know. That's why I have excluded it only on CLANGARM64.

raedrizqie:
ah.. sorry.. my bad

PR #14099, opened 2022-11-16
netcdf-cxx: add package 4.3.1

jannick0:


jannick0:
URT64 failed due to slow server operation, not because of error in CI build.

mingw-w64-netcdf-cxx/PKGBUILD
```diff
@@ -0,0 +1,47 @@
+#! /usr/bin/bash
```
Biswa96, line 1: PKGBUILD is not used as shell script. This shebang can be removed. Instead, you can add yourself with `#Maintainer:` tag.

mingw-w64-netcdf-cxx/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+url="https://github.com/Unidata/netcdf-cxx4/"
+license=(custom)
+depends=("${MINGW_PACKAGE_PREFIX}-autotools"
+         "${MINGW_PACKAGE_PREFIX}-cc"
+         "${MINGW_PACKAGE_PREFIX}-netcdf>=4.6.0")
+source=("${url}/archive/v${pkgver}/${pkgname}-${pkgver}.tar.gz")
+sha256sums=('e3fe3d2ec06c1c2772555bf1208d220aab5fee186d04bd265219b0bc7a978edc')
+
+prepare() {
+  local v=$(echo ${pkgver} | sed 's/\..*$//')
```
Biswa96, line 18: Instead of `sed` command, try to use pattern substitution with`${pkgver}` variable if possible. For example, `${pkgver//_/-}` replaces underscore with hyphen.

mingw-w64-netcdf-cxx/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+  cd "${pkgname}${v}-${pkgver}"
+
+  autoreconf -vfi
+}
+
+build() {
+  local v=$(echo ${pkgver} | sed 's/\..*$//')
+  mkdir -p "build-${pkgname}-${pkgver}-${MINGW_CHOST}"
```
Biswa96, line 26: For build directory, MSYSTEM is preferred instead of MINGW_CHOST because the former one is unique in different cases.

jannick0:
New PR as requested force-pushed.

Biswa96:
libh5bzip2.dll is installed in $prefix/lib directory. Is that expected behavior? For example, see this portion of the CI log https://github.com/msys2/MINGW-packages/pull/14099/checks#step:11:1028

jannick0:
I know, this is the deliberate autotools configuration related to filter modules, as far as I can tell. I guess that will be amended on various platforms in future releases of the netcdf and/or netcdf-cxx. I would suggest to leave it there for now to not annoy any package user and wait for upstream to get the config of netcdf and netcdf-cxx in sync.

mingw-w64-netcdf-cxx/PKGBUILD
```diff
@@ -0,0 +1,42 @@
+
+pkgname=netcdf-cxx
```
jeremyd2019, line 2: `pkgbase`/`pkgname` are incorrect for mingw package

lazka: thanks, #14211

PR #14146, opened 2022-11-18
mpfr: update to 4.1.1; enable shared build

MehdiChinoune:


lazka:
thanks, I'll give it a try

jeremyd2019:
this package has been a pain in the past, both it and gmp clearly assume that they'll only be built either static or shared on Windows, not both, and that the headers between them are incompatible.

need to make sure that both the static and shared variants can be used successfully (without truncated relocations), and particular attention should be paid to gcc (I remember an entertaining failure that made gcc crash when it tried to parse a double constant).  Also depends may need to be adjusted for gcc (or any other dependent package) if it should start pulling in the dll variant unexpectedly.

lazka:
I tried to build gcc with this, but that is currently broken because of other things: #14163

lazka:
I've built gcc twice with this and all looks good. We also have a shared version of mpfr in msys already, so I hope this is will work out.

We'll need to do rebuilds after this and add runtime deps where needed (in case of gcc it's all packages except gcc-libgfortran and gcc-libs)

MehdiChinoune:
I am planning to also enable shared library for `mpc` which requires this PR.

MehdiChinoune:
`make check` fails to link tests:
```
ld.lld: error: unable to automatically import from __gmpfr_emin with relocation type IMAGE_REL_AMD64_SECREL in libfrtests.a(tests.o)
```

MehdiChinoune:
@jeremyd2019 you said in https://github.com/msys2/MINGW-packages/pull/10809#issuecomment-1079520737 that gmp static headers could work fine with shared gmp, could you explain how to fix relocations.

MehdiChinoune:
mpc tests succeeded.

jeremyd2019:
where do you see those errors?

jeremyd2019:
Oh, those are from mpfr, and are thread-local 😬 (https://github.com/bennoleslie/mpfr/blob/7a2049c5d2326a65b6de6dcecaef2d1363f684f8/src/mpfr-impl.h#L253-L259).  I don't think it is possible to import thread-locals from a dll properly (see https://github.com/bennoleslie/mpfr/blob/7a2049c5d2326a65b6de6dcecaef2d1363f684f8/src/mpfr-thread.h#L35).

@mstorsjo could probably explain more as to why, as I recall this has come up before

MehdiChinoune:
> where do you see those errors?

mpfr's PKGBUILD `check()` function

jeremyd2019:
Thanks for taking the time to provide the details.  One note:
> Overall, direct access to variables from a different DLL is discouraged (as it either requires using proper dllimport annotations, or generating runtime pseudo relocations in the case of mingw)

Doesn't 'direct' auto-import access to variables avoid pseudo-relocation by linking to the `__imp_` import table entry, assuming the compiler generates accesses through refptrs?  So the only case that should actually require pseudo-relocations is accesses offset from the imported address?  Or am I mis-remembering?  (It seems like offset accesses should also work if refptrs are used, because it would get the pointer and then offset from it)

jeremyd2019:
Ah, I think I was thinking of something related to this comment:
https://github.com/bminor/binutils-gdb/blob/c392c0e0ae1b654c3fafc42e2a9ed7cbfcee0769/ld/pe-dll.c#L2844-L2846
>   /* In the original implementation, the pseudo-relocator was only used when
     the addend was not null.  In the new implementation, the OS loader is
     completely bypassed and the pseudo-relocator does the entire work.  */

mstorsjo:
> > If variable is accessed with an offset, I would expect the compiler to generate the refptr stub pointing directly at the symbol and applying the offset in code on top of it, but I'm not sure (do you have a good example nearby, to throw into Compiler Explorer?).
> 
> Made one up with the two cases I could think of: https://gcc.godbolt.org/z/hnc3Kjrqv

Ah, I see. The generated code there seems reasonable - since the compiler knows it's supposed to be dllimported. If the dllimport attribute is replaced with `extern`, it generates code with refptr stubs that reference the raw symbol, and the offsetting is done in code.

> > Don't know how to get it to do the same thing for mingw-w64 gcc though.
> 
> It may be possible to [run a local instance](https://github.com/compiler-explorer/compiler-explorer#running-a-local-instance) with mingw toolchain.

There's really no need for the full Compiler Explorer - it's just a handy way to share and look at the same thing. I just took the example code you provided above, replaced `__declspec(dllimport)` with `extern` and saved in a local file, and ran `x86_64-w64-mingw32-gcc -S -o - test.c  -O2` and saw that GCC handled it in the same way as Clang.

> Ah, I think I was thinking of something related to this comment: https://github.com/bminor/binutils-gdb/blob/c392c0e0ae1b654c3fafc42e2a9ed7cbfcee0769/ld/pe-dll.c#L2844-L2846
> 
> > /* In the original implementation, the pseudo-relocator was only used when
> > the addend was not null.  In the new implementation, the OS loader is
> > completely bypassed and the pseudo-relocator does the entire work.  */
> 
> I probably read something about that original implementation and internalized the idea that "the pseudo-relocator was only used when the addend was not null" (ie, offset is not 0)

Oh, ok, I see. Yeah that original implementation is ancient history, and modern pseudo relocations handle offsets if needed. You can end up with this e.g. with something like this:
```
extern int bar[42];
int *barPtr = &bar[10]; 
```
Anyway, that was a side track - the earlier post above summarizes the impact from various uses of autoimported variables with various tools.

MehdiChinoune:
What should we do about this?

PR #14151, opened 2022-11-18
onevpl-cpu: update to 2023.1.0

MehdiChinoune:


Biswa96:
I assume the issue is related to this change https://github.com/oneapi-src/oneVPL-cpu/commit/09d55c6867950dafb6cbef394a6a0a4d13bd4ef3#diff-8f17961f5265d368963546b1fffdcf52c928b16633ba3819782e6d9609d9524aR304

I do not understand what the reason behind that change. For this packaging scenario, the return value of `is_repo_root()` function can be set to false as a workaround. But the workaround will break if the ffmpeg in the package is built twice.

PR #14156, opened 2022-11-19
suitesparse: update to 6.0.2

MehdiChinoune:
- A Fortran Compiler is now required to build suitesparse (No CLANG32)
- Something wrong with either mpfr or gmp on GCC based envs, I tried to force linking to static gmp but still getting the same linking errors. Maybe #14146 could fix this issue!


MehdiChinoune:
Buidling against #14146 succeeds.

mmuetzel:
Regarding the new dependency on a Fortran compiler, it might help to revert this changeset:
https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/70369da8d4aca602770331cd8f70f2147ee01f4a

IIUC, it only adds demos that aren't actually needed for the libraries.

MehdiChinoune:
> Regarding the new dependency on a Fortran compiler, it might help to revert this changeset: [DrTimothyAldenDavis/SuiteSparse@70369da](https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/70369da8d4aca602770331cd8f70f2147ee01f4a)
> 
> IIUC, it only adds demos that aren't actually needed for the libraries.

Fortran listed in other components not just AMD.

MehdiChinoune:
Building SPEX still fails with GCC based envs.

mmuetzel:
> Building SPEX still fails with GCC based envs because of static mpfr.

Could that be an issue with the `.pc` files?

PR #14163, opened 2022-11-19
gcc: rebuild

lazka:
it seems broken here, so let's try in CI :/


lazka:
`-iprefix C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/` vs `-iprefix C:/msys64/ucrt64/bin/../lib/gccx86_64-w64-mingw32/12.2.0/`

looks like a slash is missing so maybe https://github.com/msys2/msys2-runtime/pull/103 ?

lazka:
broken in CI as well, tracking in #14173

PR #14200, opened 2022-11-21
libidn2: Update to 2.3.4

lazka:


raedrizqie:
try removing libunistring from depends..

lazka:
thanks, I'll give that a try tomorrow :)

raedrizqie:
hmm.. the last time i built it CI.. it compiles fine.. without extra option..

```
Gnulib updated and now libunistring-optional is used.
This allows you to force libidn2 to use internal libunistring with the
following command: ./configure --with-included-libunistring
```

maybe we had to add `--with-included-libunistring` to force internal unistring..

raedrizqie:
you forgot to add it for static build :)

lazka:
no idea, there is an upstream bug: https://gitlab.com/libidn/libidn2/-/issues/113

PR #14205, opened 2022-11-21
Update mingw-w64-pkgconf/PKGBUILD to use version 1.9.3

zlaski:
This updates the mingw-w64-pkgconf project to use the latest pkgconf-1.9.3 release.  A couple of patch files previously created are no longer necessary.


MehdiChinoune:
We couldn't update pkgconf to 1.9.3 see #13390

PR #14210, opened 2022-11-22
netcdf-cxx: Fix PKGBUILD

MehdiChinoune:


lazka:
I fixed just the naming in #14211 in the mean time. I hope that's ok.

PR #14217, opened 2022-11-23

kmilos:


Biswa96:
Some build fails in CI. Please see the log here https://packages.msys2.org/queue

lazka:
Saved logs for future reference: https://gist.githubusercontent.com/lazka/4c05102e3f30b1bffdd8e478cd9073bd/raw/f917afe9507ae56a9f4e73c4849438524a99f61b/openimagedenoise+embree.arm64.log

kmilos:
Thanks for dealing with this, I'll take a look when I have more time. They're both Intel projects, so I was carefully optimistic, although they say Arm is supported - well, at least on the embree page (embree is also available in Homebrew and Debian for arm64, so it should be possible...)

MehdiChinoune:
> although they say Arm is supported - well, at least on the embree page (embree is also available in Homebrew and Debian for arm64, so it should be possible...)

That's true for the latest release, not for the one we are using.

kmilos:
> That's true for the latest release, not for the one we are using.

It looks like `_mm_malloc`/`_mm_free` are redefined even on master: https://github.com/embree/embree/blob/master/common/simd/arm/sse2neon.h

PR #14219, opened 2022-11-23
readline: Fix regression in patch 001

### LINKED ISSUE #14213: readline: Arrow keys in Octave CLI no longer working
### Description

Octave can be started either with a GUI (`octave --gui`) or with a command line interface (`octave`).
When using the arrow keys at the command line interface (CLI), some characters appear instead of moving the cursor (for the left or right arrow key) or browsing the command history (for the up or down arrow key).
It is still working correctly in the GUI.

That is probably related to the update of readline to version 8.2. When I downgrade to readline 8.1 (`pacman -U mingw-w64-x86_64-readline-8.1.002-2-any.pkg.tar.zst`), the CLI is working correctly again.

I thought I tested that when proposing that update. But I must have done something wrong. (Please, apologize for that.)

### Verification

### Windows Version

Microsoft Windows [Version 10.0.22621.819]

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Expected behavior

The arrow keys should move the cursor or browse the command history.

### Actual behavior

Some characters appear at the command prompt when using the arrow keys.

### Repro steps

1. Install Octave with `pacman -S mingw-w64-x86_64-octave`.
2. Start its command line interface with `octave`.
3. Use the arrow keys at the command prompt.

Yes, in principle. But I'm not sure when I'll be able to investigate further.

mmuetzel:
This fixes #14213 for me.

It looks like the issue was caused by part of the upstream patch 001. If I understand its intention correctly, the main goal is to have a valid fallback locale in case detecting it (with `setlocale`) fails. That should still be the case with the proposed change.

Also thank you to @lostbard for his hints in https://savannah.gnu.org/bugs/index.php?63400.

### DIFF
--- a/mingw-w64-readline/0002-event-hook.patch
+++ b/mingw-w64-readline/0002-event-hook.patch
@@ -1,7 +1,7 @@
-diff -ur readline-8.1.3/input.c readline-8.1.4/input.c
---- readline-8.1.3/input.c	2020-12-19 07:56:01.572775578 -0500
-+++ readline-8.1.4/input.c	2020-12-19 07:57:24.608319024 -0500
-@@ -142,6 +142,11 @@
+diff -urN readline-8.2/input.c.orig readline-8.2/input.c
+--- readline-8.2/input.c.orig	2022-04-08 21:43:24.000000000 +0200
++++ readline-8.2/input.c	2022-11-22 16:54:55.099070500 +0100
+@@ -176,6 +176,11 @@
  static unsigned char ibuffer[512];
  static int ibuffer_len = sizeof (ibuffer) - 1;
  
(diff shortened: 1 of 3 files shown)


mmuetzel:
I tested the MINGW64 build artifact from this PR with the CLI and the GUI of Octave. The bug described in #14213 no longer occurs. The arrow keys are working as expected again.

Biswa96:
Should we wait for any upstream commit?

mmuetzel:
I didn't report this upstream (and haven't planned to do so yet). I don't know if they care about Windows...

PR #14220, opened 2022-11-23
clean up Zlib license tags

kmilos:


kmilos:
No idea why this failed, but it shouldn't cause rebuilds anyway...

PR #14259, opened 2022-11-24
cppwinrt: Fix dependency and enable for ucrt64

MehdiChinoune:


Biswa96:
cppwinrt headers can not be used with gcc yet. So, cppwinrt.exe in ucrt64 will not be useful.

MehdiChinoune:
> cppwinrt headers can not be used with gcc yet. So, cppwinrt.exe in ucrt64 will not be useful.

It won't hurt!

PR #14297, opened 2022-11-27
libxml2: add ICU support

MehdiChinoune:


lazka:
I don't think we need this, see https://github.com/msys2/MSYS2-packages/commit/aeb3a4cd367933026b1d6e93b0cdfc8f0a8c81da

lazka:
I don't think we need this, it either uses libiconv or libicu, and the later seems overkill: https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/encoding.c

Or is there another reason?

MehdiChinoune:
> Or is there another reason?

Actually I am trying to add qt6-webengine package. It rejects libxml2/libxslt unless libxml2 was built with ICU, but It's not a big deal.

lazka:
> Actually I am trying to add qt6-webengine package. It rejects libxml2/libxslt unless libxml2 was built with ICU, but It's not a big deal.

huh, weird. fine with me then

MehdiChinoune:
I am thinking about a scenario when we upgrade icu (which happens a lot) -> broken libxml2 -> broken llvm!

lazka:
uh, yeah :/

Issue #13783, opened 2022-10-28
Create a openmp meta/virtual package


lazka:
So we can depend on openmp without any environment specific conditions (except arm64, which doesn't have it). See https://github.com/msys2/MINGW-packages/pull/12280#issuecomment-1198985768

So I don't forget again...

kmilos:
I guess this will obsolete https://github.com/msys2/MINGW-packages/issues/10649

mati865:
I think we should roughly do what https://github.com/msys2/MINGW-packages/issues/10649 suggests and make it installed by the default on CLANG* subsystems. Dunno about the best way to achieve that.

lazka:
PRs in #13807 and #13808

This assumes the "-omp" is in depends and "-cc" im makedepends for all packages.

lazka:
This is now in the repo.

lazka:
PKGBUILDs to clean up:

```
mingw-w64-adol-c
mingw-w64-bsdfprocessor
mingw-w64-cas-offinder
mingw-w64-clustal-omega
mingw-w64-frontistr
mingw-w64-gegl
mingw-w64-gmic
mingw-w64-graphicsmagick
mingw-w64-hypre
mingw-w64-igraph
mingw-w64-inkscape
mingw-w64-libbotan
mingw-w64-libosmscout
mingw-w64-libsoxr
mingw-w64-libsquish
mingw-w64-mypaint
mingw-w64-nextpnr
mingw-w64-ngspice
mingw-w64-onednn
mingw-w64-openblas
mingw-w64-pdf2djvu
mingw-w64-petsc
mingw-w64-python-numba
mingw-w64-scopehal-apps
mingw-w64-suitesparse
mingw-w64-sundials
mingw-w64-superlu_dist
mingw-w64-tesseract-ocr
mingw-w64-tulip
mingw-w64-vid.stab
```

kmilos:
@lazka Ok to start enabling for `clangarm64` as well (could also wait until they get touched)?

lazka:
Sure, I'm not aware of any arm64 issues with openmp yet... But it' also rarely used yet. Only one way to find out :)

lazka:
I've updated the package list above

MehdiChinoune:
a virtual package `mingw-w64-omp` is provided by ${MINGW_PACKAGE_PREFIX}-gcc-libs and ${MINGW_PACKAGE_PREFIX}-llvm-openmp.

PR #14326, opened 2022-11-30
gmic: use omp metapackage

kmilos:
Related to #13783


kmilos:
No idea why gegl build now fails... recent GLib update?

lazka:
yeah, that's a recent glib regression, already fixed upstream afaik

PR #14331, opened 2022-11-30
mingw-w64 headers: drop dx9 patch

lazka:
VLC seems to build fine without it and we should avoid carrying mingw-w64 patches downstream, so remove it.

If this breaks something let us know.


lazka:
@Biswa96 any thoughts on this?

Biswa96:
I do not know why it was added. The author of that patch @robUx4 may know.

lazka:
good point, any context/input appreciated.

But, either way, this seems fine to try out.

PR #14352, opened 2022-12-01
synth: New package

iamwacko:


iamwacko:
Synth is failing to compile and I don't quite know why, so I'm just going to close this. I might open another one in the future.

PR #14374, opened 2022-12-02
install-depends fix: arv-viewer needs videoflip GStreamer plugin, whi…

eudoxos:
…ch is in -good plugins

(-base is still sufficient for building)

### DIFF
--- a/mingw-w64-aravis/PKGBUILD
+++ b/mingw-w64-aravis/PKGBUILD
@@ -7,7 +7,7 @@ pkgname=(
 	"${MINGW_PACKAGE_PREFIX}-${_realname}-gst"
 )
 pkgver=0.8.22
-pkgrel=2
+pkgrel=3
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 pkgdesc="A vision library for genicam based cameras"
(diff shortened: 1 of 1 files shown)


Biswa96:
Can you provide any code which proves the requirement? From the meson.build, I found that the gstreamer related pkgconfig files belong to gst-plugin-base package.

eudoxos:
Right here: https://github.com/AravisProject/aravis/blob/main/viewer/arvviewer.c#L62 it will fail at runtime.

Biswa96:
Is it like a plugin to the actual main program? Shouldn't gst-plugins-good be mentioned as option dependency?

eudoxos:
No, it is not optional. The `*-aravis-gst` package only contains the arv-viewer* binary (and its resources), and the viewer will not run without the -good plugins set. It is simply broken without the -good plugins.

eudoxos:
(I am the original packager, BTW, so I know this firsthand)

Biswa96:
Thanks for the hint. In the runtime dependencies list, only gst-plugins-base DLLs are present https://github.com/msys2/MINGW-packages/actions/runs/3601668822/jobs/6067735084#step:11:1759. So, I just make sure of the requirement.

BTW, bump the pkgrel variable please.

PR #14404, opened 2022-12-03
skia: update to 108.0.5359.95

naveen521kk:
- use gn from the repo, instead of building one
- updated patches


raedrizqie:
try using fmax instead of max here:
```
dst[i] = std::max(dst[i], 10u);
```

naveen521kk:
@raedrizqie Can you check whether https://github.com/msys2/MINGW-packages/pull/14404/files#diff-d98450ec56f6824a4b53de571493a41f73d20f79b6088a6effce0fffa41f4c52 looks good, from [StackOverflow](https://stackoverflow.com/a/38449573/12858827)? (noob C++ dev here)

PR #14405, opened 2022-12-03
octave: change console codepage to UTF-8

mmuetzel:
Unfortunately, I messed up the upstream commit. So, we need a slightly different change.

See also: msys2/MSYS2-packages#3338

### DIFF
--- a/mingw-w64-octave/0001-console-codepage.patch
+++ b/mingw-w64-octave/0001-console-codepage.patch
@@ -1,0 +1,121 @@
+# HG changeset patch
+# User Markus Mützel <markus.muetzel@gmx.de>
+# Date 1670071302 -3600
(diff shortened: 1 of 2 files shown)


mmuetzel:
The package is still working with these changes. Just out of interest: What are those symbols that the package grokker notices to be removed from all executables?
E.g.;
```
Removed DLLs/Symbols
  {b'mkoctfile-7.3.0.exe': {b'__mingw_winmain_hInstance',
                            b'__mingw_winmain_lpCmdLine',
                            b'__mingw_winmain_nShowCmd'},
```

Biswa96:
Probably due to this change in mingw-w64 https://github.com/mingw-w64/mingw-w64/commit/22f6182b803d114f7310b473cbeeb2130768a1f6. See the change in mingw-w64-crt/crt/crtexe.c.

PR #14408, opened 2022-12-03
readline: Support entering non-ASCII characters also with other shells

mmuetzel:
Like @lazka rightfully pointed out in msys2/MSYS2-packages/#3338, the POSIX locale variables (like `LC_CTYPE`) might not be set when the shell startup script from MSYS2 isn't used.
To determine whether readline should be using an UTF-8 locale, it is probably better to query the console codepage instead of using the return value of `setlocale (LC_CTYPE, (char *)NULL);` on Windows. The proposed PR is doing that.

With this change (and PR #14405), it is possible to enter non-ASCII characters at the Octave prompt also when starting the program with a double-click on `octave.exe` in `mingw64/bin` (i.e., when the shell startup script didn't run).

Sorry, this took a couple of attempts. But I hope we are getting there now.

### DIFF
--- a/mingw-w64-readline/0004-locale.patch
+++ b/mingw-w64-readline/0004-locale.patch
@@ -1,10 +1,38 @@
 --- readline-8.2/nls.c.orig	2022-11-22 18:41:48.302144300 +0100
-+++ readline-8.2/nls.c	2022-12-01 19:23:48.168183300 +0100
-@@ -138,11 +138,15 @@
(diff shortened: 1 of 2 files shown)


mmuetzel:
Please, wait a bit before merging this. There is a slight think-o in the current version.

mmuetzel:
Forgot to comment after I force-pushed: Should be fixed now.

PR #14413, opened 2022-12-04
sed: update to 4.9

MehdiChinoune:


raedrizqie:
remove the LDFLAGS before configure and add this sed:

```
sed -e "s|sed/libver.a lib/libsed.a|sed/libver.a lib/libsed.a -lbcrypt|g" -i sed/local.mk
```

MehdiChinoune:
> remove the LDFLAGS before configure and add this sed:
> 
> ```
> sed -e "s|sed/libver.a lib/libsed.a|sed/libver.a lib/libsed.a -lbcrypt|g" -i sed/local.mk
> ```

Thanks.

Biswa96:
I have sent the fix in upstream, see this mail https://lists.gnu.org/archive/html/sed-devel/2022-12/msg00000.html.

Biswa96:
The patch has been added in this commit https://git.savannah.gnu.org/cgit/sed.git/commit/?id=3bb0a12761f5f98348dfa4f3eb31b6fe57bb466f

PR #7430, opened 2020-12-11
Add lib3mf-1 package

ImperatorS79:
Needed by openscad.

PR #14415, opened 2022-12-04
Remove lib3mf-1

MehdiChinoune:


Biswa96:
Why is this package being removed?

MehdiChinoune:
> Why is this package being removed?

Because It's no longer needed (an orphan)

Biswa96:
That does not mean it is not used anywhere.

The upstream project is in still in development. Some branches have new commits in https://github.com/3MFConsortium/lib3mf

ArchLinux mentioned is as a dependency of openscad, see https://archlinux.org/packages/community/x86_64/lib3mf-1/

MehdiChinoune:
> That does not mean it is not used anywhere.
> 
> The upstream project is in still in development. Some branches have new commits in https://github.com/3MFConsortium/lib3mf
> 
> ArchLinux mentioned is as a dependency of openscad, see https://archlinux.org/packages/community/x86_64/lib3mf-1/

We have another package `lib3mf`.

Biswa96:
I am not sure if lib3mf-1 and lib3mf can be used interchangeably. Maybe it's like openssl v1 vs v3.

MehdiChinoune:
> I am not sure if lib3mf-1 and lib3mf can be used interchangeably. Maybe it's like openssl v1 vs v3.

We don't have any package requiring lib3mf-1, Isn't that enough to remove it?

MehdiChinoune:
> The author of the PR #7430 mentioned that lib3mf-1 is required by openscad.

Ours doesn't

Biswa96:
I would suggest to wait for other opinions. @lazka

lazka:
openscad supports lib3mf 2.0 in the release we have: https://github.com/openscad/openscad/blob/master/RELEASE_NOTES.md

In Ubuntu 22.04 openscad is also the only reverse dependency, so that suggests there not many other users of the old library, so I think it is fine to drop it.

Biswa96:
OK, thanks.

PR #14450, opened 2022-12-04
sse2neon: add package

driver1998:
This is a SSE to NEON translation library, therefore clangarm64 only.
Needed by blender.


Biswa96:
1. Does sse2neon require to be a package? It's just one header file.
2. Does blender include sse2neon in source tarball?
3. Does blender require/check any specific version of the sse2neon header?

PR #14452, opened 2022-12-04
update mingw-w64; bump default _WIN32_WINNT to Windows 8.1

MehdiChinoune:


mingw-w64-headers-git/PKGBUILD
```diff
@@ -39,7 +39,7 @@ build() {
   local _default_win32_winnt
   case "${CARCH}" in
     i686|x86_64)
-      _default_win32_winnt=0x601
+      _default_win32_winnt=0x603
```
MehdiChinoune, line 42: Is this OK?

lazka: fine with me

lazka: @Biswa96 ok?

Biswa96: 👌

Biswa96: I do not use < Win10 btw, so biased.

mingw-w64-headers-git/PKGBUILD
```diff
@@ -4,9 +4,9 @@ _realname=headers
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgdesc="MinGW-w64 headers for Windows"
-pkgver=10.0.0.r153.g31e6f8e5a
+pkgver=10.0.0.r157.gd295924f0
 pkgrel=2
```
jeremyd2019, line 8: hah, yeah, I've missed resetting pkgrel on these before :grin:

LukeUsher:
Is there a justification for updating the minver to win8.1?

I ask because we are seeing a lot of complaints that our software has ceased to work on Windows 7; I know it's an EOL operating system but it still makes up a considerable share of the windows userbase.

I see no changes here that specifically *require* win8.1 rather than win7, so it seems arbitrary? 

Would like to keep using mingw-w64 distributions if possible, would rather not resort to have to building our own toolchain to keep supporting a significant portion of our userbase.

LukeUsher:
> > Is there a justification for updating the minver to win8.1?
> 
> Yes, we are dropping support for Windows 7. It also brings us closer to MSVC, which afair defaults to Win10.
> 
> > I ask because we are seeing a lot of complaints that our software has ceased to work on Windows 7; I know it's an EOL operating system but it still makes up a considerable share of the windows userbase.
> 
> What is causing your software to break? The mingw-w64 libraries should still work on Win7 with this.

libcxx's implementation of std:chrono choses a code path that uses GetSystemTimePreciseAsFileTime rather than GetSystemTimeAsFileTime, if _WIN32_WINNT is set to Win8.1 or above. I first opened an issue with llvm, but they informed me that it is up to the compiler distribution to ensure that _WIN32_WINNT  is set to a suitable value at compile time.

The relevant code in libcxx:
```c++
#if (_WIN32_WINNT >= _WIN32_WINNT_WIN8 && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)) || \
    (_WIN32_WINNT >= _WIN32_WINNT_WIN10)
  GetSystemTimePreciseAsFileTime(&ft);
#elif !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
  GetSystemTimeAsFileTime(&ft);
#else
  GetSystemTimeAsFileTimeFunc.fp(&ft);
#endif
```

lazka:
I see, thanks. I'm fine with keeping libcxx win7 compatible for a bit longer.

lazka:
Just wondering, does this mean you are distributing/installing UCRT with your software for win7 users?

LukeUsher:
> Just wondering, does this mean you are distributing/installing UCRT with your software for win7 users?

Just the resulting exe from compilation, libcxx gets statically linked as far as I can tell.

mati865:
There is no such thing as static linking of CRT with mingw-w64 so you still depend on UCRT DLLs.

jeremyd2019:
> > What is causing your software to break? The mingw-w64 libraries should still work on Win7 with this.
> 
> libcxx's implementation of std:chrono choses a code path that uses GetSystemTimePreciseAsFileTime rather than GetSystemTimeAsFileTime if _WIN32_WINNT is set to Win8.1 or above. I first opened an issue with llvm, but they informed me that it is up to the compiler distribution to ensure that _WIN32_WINNT is set to a suitable value at compile time.
> 
> The relevant code in libcxx:

Is that in a header or in a compiled library?  In general, a project should be able to define _WIN32_WINNT to whatever they need, and what's specified here is just a default.  In fact, in the past I have worked with projects that defined _WIN32_WINNT to a higher value than the minimum supported OS, in order to get access to defines/constants for newer OSes, and then was just careful not to use anything from newer than the minimum outside a version check (or using GetProcAddress for function calls).  Having standard library behavior change like that based on the _WIN32_WINNT value would have broken this.

lazka:
> I first opened an issue with llvm, but they informed me that [..]

For more context, this probably refers to https://github.com/llvm/llvm-project/issues/59539

jeremyd2019:
Ouch, that's in the library indeed, so `_WIN32_WINNT` would need to be set to the intended minimum os when builing libc++.  That's not cool.  Projects are supposed to be able to override that at compile time to what they want it to be.

mati865:
> Ouch, that's in the library indeed, so `_WIN32_WINNT` would need to be set to the intended minimum os when builing libc++. 

I think it's not really possible, there are code paths inside mingw-w64 CRT and winpthreads that call different functions  depending on how `_WIN32_WINNT` was set during the build.

jeremyd2019:
I thought mingw-w64 overrode `_WIN32_WINNT` to `0x0f00` during its build, so that all functions/constants would be available.

mati865:
Oh, indeed. I was wrong and it's used for winpthreads and winstorecompat only.
For GCC based subsystems this is enough to make targeting earlier systems impossible because even unwind part of `libgcc` links winpthreads.
In my opinion (biased since I develop for Linux only) dropping support for ancient OS would save some of the maintenance cost. This has been announced long time ago, Python has already dropped that support (I don't know if Python packages by MSYS2 even works on Win 7), Rust is also looking to stop supporting Win 7 in the future (no ETA) and more tools are only to come.

jeremyd2019:
> In my opinion (biased since I develop for Linux only) dropping support for ancient OS would save some of the maintenance cost. 

Yes, there needs to be a line drawn somewhere, and everybody just argues over where to draw it.

My only reason to wade into this is that it does not seem to make sense that the 'default' _WIN32_WINNT version that mingw-w64-headers' configure takes should affect libc++ in this way.  My understanding of _WIN32_WINNT is that it is meant for a 'project' to define to communicate to the 'SDK' what version APIs should be defined.  libc++ seems to be using it as a mechanism for the 'SDK' to tell the 'project' which APIs it is allowed to use.  This effectively transforms the default _WIN32_WINNT into a minimum target os version.

(in this case the 'project' is libc++'s dynamic/static library, normally you'd think of the 'project' being the consumer of libc++ but these are built into the library.  If they were in headers/inline code, it could justifiably use _WIN32_WINNT like this with the assumption that the 'project' that consumes libc++ would define it appropriately).

I should probably add this to the upstream llvm bug, maybe it makes some sense to somebody.

mstorsjo:
> My only reason to wade into this is that it does not seem to make sense that the 'default' _WIN32_WINNT version that mingw-w64-headers' configure takes should affect libc++ in this way. My understanding of _WIN32_WINNT is that it is meant for a 'project' to define to communicate to the 'SDK' what version APIs should be defined. libc++ seems to be using it as a mechanism for the 'SDK' to tell the 'project' which APIs it is allowed to use. This effectively transforms the default _WIN32_WINNT into a minimum target os version.

FWIW, libc++ isn't the only piece of code to do this; e.g. `psapi.h` (both in mingw toolchains and in WinSDK) does `#if NTDDI_VERSION >= NTDDI_WIN7` (which in most cases boils down to `_WIN32_WINNT`), and picks between linking against e.g. `EnumProcessModules` (from `psapi.dll`) or `K32EnumProcessModules` (from `kernel32.dll` since Win7 only). I've seen a handful of other cases where projects use this for picking between implementations too - but I don't have other examples ready offhand.

Yes, `_WIN32_WINNT` also affects which APIs/constants are visible at compile time (which is more of a "max SDK known at compile time" instead of "min target version") - but that also works automatically as a "min target version" when used together with projects that have configure checks for the modern APIs that they might use; with `_WIN32_WINNT` set to an older version, the configure checks will fail to use the new APIs, while if it is set to newer version, the new APIs end up used (possibly linked directly against, although it concievably also could be used for enabling code that tries to runtime load the new APIs).

>  If they were in headers/inline code, it could justifiably use _WIN32_WINNT like this with the assumption that the 'project' that consumes libc++ would define it appropriately

Well on the other hand, I would expect that it wouldn't be hard to add a `-D_WIN32_WINNT=0x0601` to the PKGBUILD file for libc++ (adding it to e.g. `CMAKE_CXX_FLAGS` or similar) to explicitly make the libc++ library support older versions. (Setting it to Vista wouldn't help making it compatible with Vista though, since libc++'s win32 thread usage uses `TryAcquireSRWLockExclusive` which only appeared in Win7, even though most similar functions are available since Vista).

mstorsjo:
> FWIW, libc++ isn't the only piece of code to do this; e.g. `psapi.h` (both in mingw toolchains and in WinSDK) does `#if NTDDI_VERSION >= NTDDI_WIN7` (which in most cases boils down to `_WIN32_WINNT`), and picks between linking against e.g. `EnumProcessModules` (from `psapi.dll`) or `K32EnumProcessModules` (from `kernel32.dll` since Win7 only). I've seen a handful of other cases where projects use this for picking between implementations too - but I don't have other examples ready offhand.

Another case with the same logic, in the official WinSDK, is where `ws2tcpip.h` defines IPv6 functions like `getaddrinfo`, and optionally redirects them to inline functions that do `LoadLibrary` + `GetProcAddress`. Here's a snippet from the logic there, from `ws2tcpip.h`:

```
//
// Unless the build environment is explicitly targeting only
// platforms that include built-in getaddrinfo() support, include
// the backwards-compatibility version of the relevant APIs.
//
#if !defined(_WIN32_WINNT) || (_WIN32_WINNT <= 0x0500)
#include <wspiapi.h>
#endif
```
(Then `wspiapi.h` contains inline compat functions, and do `#define getaddrinfo WspiapiGetAddrInfo`, redirecting it to the compat function. The mingw-w64 headers do contain similar things but not the automatic include of `wspiapi.h`.)

So with that prior art, I thought it would be reasonable to use `_WIN32_WINNT` to indicate the lowest required platform version within libc++ too.

oscarfv:
Coming from #14573 

> We started setting default `_WIN32_WINNT` to `0x0603` (Win8.1) in #14452, Maybe that's why.

Just a PR, merge it and that's all? No announcement about the incoming change? No News entry in the website nor in the project's channels?

One thing is desupporting an OS with an arguable justification. Another thing entirely is doing it without proper warning.

Biswa96:
> No announcement about the incoming change? No News entry in the website nor in the project's channels?

There was a new entry in April here https://www.msys2.org/news/#2022-04-06-windows-7-8-support-will-be-dropped-late-2022-or-early-2023

There is a dedicated page here https://www.msys2.org/docs/windows_support/

Also in these reports Windows 7 user percentage is near 1% https://github.com/msys2/MSYS2-packages/discussions/2832

oscarfv:
C'mon, one thing is announcing that Win7 will be deprecated at some fuzzy future point and another is actually deprecating it. Knowing the level of attention that you guys put into this project's management and communication, is as if you are trying to keep a low profile for this change, as if you are feeling guilty ;-)

As for those statistics, one thing is what developers use, another thing is where software runs. My user base has larger than 5% of Win7 users. No problem with me, I'll manage the transition by forcing those users to upgrade and freezing their installs meanwhile. But I need to know that the transition is coming, otherwise a terrible mess could happen.

lazka:
> C'mon, one thing is announcing that Win7 will be desupported at some fuzzy future point and another is actually desupporting it. Knowing the level of attention that you guys put into this project's management and communication, is as if you are trying to keep a low profile for this change, as if you are feeling guilty ;-)

With that many packages it's not clear what "actually desupporting" means. This change doesn't do that at least (well, except for the unforeseen libcxx issue mentioned above)

There are probably already various packages in the repo that require newer Windows or hardware newer then the minimum Windows requirements, since we have no way of testing for this.

I'll write up a short NEWS entry for this change, thanks for the suggestion.

> As for those statistics, one thing is what developers use, another thing is where software runs. My user base has larger than 5% of Win7 users. No problem with me, I'll manage the transition by forcing those users to upgrade and freezing their installs meanwhile. But I need to know that the transition is coming, otherwise a terrible mess could happen.

Yeah, two different types of users.

MehdiChinoune:
> Coming from #14573 
> 
> > We started setting default `_WIN32_WINNT` to `0x0603` (Win8.1) in #14452, Maybe that's why.
> 
> Just a PR, merge it and that's all? No announcement about the incoming change? No News entry in the website nor in the project's channels?
> 
> One thing is desupporting an OS with an arguable justification. Another thing entirely is doing it without proper warning.

We announced about phasing Windows 7 months ago.

jeremyd2019:
Yeah, I was surprised that libc++ stopped working on win7 after this change.  I had thought about suggesting overriding `_WIN32_WINNT` for libc++ build for that reason.  But, there has also been a report on mingw-w64 list about a file in gnulib doing something similar (https://sourceforge.net/p/mingw-w64/mailman/message/37752394/).  So I think this change will actually cause creeping non-support of win7 as packages get rebuilt.

I think the lack of a big scary announcement saying "win7 support is going away NOW" is just due to the fact that nobody seems to have really expected *this* change to have that result.

lazka:
I'm fine with reverting this for now.

lazka:
Any thoughts on the proposal here https://github.com/msys2/MINGW-packages/pull/14736 ? i.e. only bump for ucrt envs

lazka:
libc++ is now rebuilt to target Windows 7 everywhere, and for MINGW32 and MINGW64 we reverted `_WIN32_WINNT` to Windows 7.

loreno-heer:
So if I understand this correctly this means (for now) I can upgrade all msys2/mingw packages and recompile my softwares again on win 7 as long as I don't upgrade anymore after that?

lazka:
In theory, yes, but no guarantees.

helohe:
It would be great if pacman just refuses to update any more packages on win 7 as soon as packages are pushed which break the build there

jeremyd2019:
This would require additional logic in pacman to look at the Windows version (I think I remember somebody looking at that), but perhaps more importantly a priori knowledge of when win7 is broken.  I am guessing nobody in this project still uses win7, nor is there any GHA hosted runner that old.

PR #14456, opened 2022-12-05
blender: bump pkgrel

driver1998:
I forgot to bump pkgrel again after I resolve conflicts (I started with 2 and bumped to 3).


jeremyd2019:
I think you forgot to actually add `'clangarm64'` to `mingw_arch` also.

PR #14477, opened 2022-12-08
swipl: update to version 9.0.2

mgondan:
unit tests are passing now


mgondan:
Patch 1 and 2 are MinGW-specific. Patches 3 and 4 habe been reported, yes, and already committed, so the patches will be removed again in the next version.

lazka:
perfect, thanks!

PR #14478, opened 2022-12-08
cmake: Fix Python_SOABI value

MehdiChinoune:


MehdiChinoune:
I will go ahead and merge it until the issue got fixed upstream

lazka:
OK. https://github.com/msys2-contrib/cpython-mingw/issues/119 for future reference

Issue #13198, opened 2022-09-21
Kate not completely quit


jack6th:
### Description

mingw-w64-i686-kate 22.08.1-1 on Windows 10 21H2 x64   
Kate not completely quit.   
I need end it manually in the task manager.   

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Windows 10 x64 19044.1526

### MINGW environments affected

- [ ] MINGW64
- [X] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Expected behavior

Kate.exe is not in the processes list of task manager.  

### Actual behavior

Kate.exe is still in the processes list of task manager.  

### Repro steps

1. Run kate   
2. Close kate

Biswa96:
Thank you for reporting this issue. After some searching with the error message from Kate, I found this thread https://bugreports.qt.io/browse/QTBUG-84498. It says,

> That is unsolvable. Windows has a major defect in how it shuts down its threads and there's no solution.

Let us check if the issue can be reported to KDE project.

jack6th:
Kate built with https://binary-factory.kde.org does not have this problem.   
But the source code is not available.    
https://binary-factory.kde.org/view/Windows%2064-bit/job/Kate_Release_win64/

Biswa96:
Yeah, I have noticed that. One noticeable difference I found that the "factory" kate does not have dbus libraries and does not launch dbus-session for that.

lb90:
When kate hangs, gdb is not able to get any stacktrace. I have then tried with VS2022 and here's what comes out:

```
ntdll.dll!_NtWaitForSingleObject@12()   Sconosciuto                             Codice non utente. Simboli caricati.
KernelBase.dll!WaitForSingleObjectEx()  Sconosciuto                             Codice non utente. Simboli caricati.
Qt5Core.dll!6311f9ae()                  Sconosciuto                             Codice non utente. Il file binario non è stato compilato con le informazioni di debug.
Qt5Core.dll!6311c2f0()                  Sconosciuto                             Codice non utente. Il file binario non è stato compilato con le informazioni di debug.
Qt5Core.dll!6339e84a()                  Sconosciuto                             Codice non utente. Il file binario non è stato compilato con le informazioni di debug.
[...]

Issue #10158, opened 2021-11-21
KF5 adventures - zombie processes + qt plugin install dir


naksyl:
Windows 11 breaks main menus in Qt applications so I decided to compile breeze - this leads me to some issues. Here is [PKGBUILD](https://github.com/naksyl/MINGW-kde/blob/master/mingw-w64-breeze-qt5/PKGBUILD) if somebody wants check it. 

### KF5 zombie processes
Discovered when playing with breeze, but it is not related. Probably all apps that links to some kf5 lib are affected - cant figure out which one. Even qt apps (designer, assistant) running in breeze style are affected. Steps to reproduce:
```sh
# launch couple of kwrite/kate processes - it doesnt matter if launched from console or gui
for i in {1..3}; do kwrite & done
# or simply run as kwrite - you dont see the prompt after window is closed and CTRL-C isnt working
# close all windows by clicking 'x'
# verify that processes still exists
ps -W | grep kwrite
```
I cannot kill those by bashes kill, cmd taskkill, or powershell Stop-Process only stopping process from GUI task list works. They are listed as background processes, but still eats resources? Correct me if I am wrong.

Following KF5 tutorials [THIS](https://techbase.kde.org/Development/Tutorials/Using_KActions/KF5) is the first affected application - it introduces toolbar through xmlgui - shoud be installed with `ninja install`. I found old issue  #6217 which suggest that dbus is guilty.

### Some qt/kf5 packages install libs in wrong directory
`qtdiag.exe | grep plugin` reports `$PREFIX/share/qt5/plugins` directory but:
```sh
$ ls /mingw64/lib/plugins/*/*.dll
/mingw64/lib/plugins/designer/kcompletion5widgets.dll*
/mingw64/lib/plugins/designer/ktextwidgets5widgets.dll*
/mingw64/lib/plugins/designer/kdewebkit5widgets.dll*
/mingw64/lib/plugins/designer/kwidgetsaddons5widgets.dll*
/mingw64/lib/plugins/designer/kiconthemes5widgets.dll*
/mingw64/lib/plugins/designer/kxmlgui5widgets.dll*
/mingw64/lib/plugins/designer/kitemviews5widgets.dll*
/mingw64/lib/plugins/iconengines/KIconEnginePlugin.dll*
/mingw64/lib/plugins/designer/kplotting5widgets.dll*
```
Those are probably designer plugins (there should be plenty of kf5 widgets but ther are only few) only, but still... Some proposal to fix this one:
[...]

PR #14482, opened 2022-12-08
qt5-base: Fix zombie processes due to qt dbus

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/10158
Fixes https://github.com/msys2/MINGW-packages/issues/13198

/cc @revelator @lb90

Credits goes to @naksyl who suggested the patch in the above mentioned issue 🤗


lb90:
Great, thank you @Biswa96 @naksyl!

revelator:
nice :)

PR #14483, opened 2022-12-09
KF5: Fix plugin directory using template function

Biswa96:
This install plugin directory in prefix/share/qt5/plugins instead of
prefix/lib/plugins. The prefix/share/qt5/plugins path is returned from
'qtpath --plugin-dir' command.

Fixes https://github.com/msys2/MINGW-packages/issues/10158
Similar as https://github.com/msys2/MINGW-packages/pull/4523


Biswa96:
ucrt64 😑

PR #14538, opened 2022-12-10
Add libunibreak

lazka:
New optional dep for libass, so give it a try


Biswa96:
Is it possible to add clangarm64 also? I have cross-compiled it in my system.

```
$ file ./src/.libs/libunibreak-5.dll
./src/.libs/libunibreak-5.dll: PE32+ executable (DLL) (GUI) Aarch64, for MS Windows
```

Otherwise libass may fail to build due to absence of libunibreak in clangarm64.

PR #14550, opened 2022-12-10
mpfr: Update to 4.1.1.p1

lazka:


lazka:
@MehdiChinoune I sadly don't know how to make the shared build work, hopefully we can re-address it in the future.

MehdiChinoune:
> @MehdiChinoune I sadly don't know how to make the shared build work (without many patches), hopefully we can re-address it in the future.

We could, but sadly there is a little interest on it.
There was a claim that static gmp could work in all cases, and then it turns out that's not.
If that's the case, shouldn't shared libraries prioritized over static ones like most distributions/platforms do!

PR #14567, opened 2022-12-12
Favor shared over static for gmp to build shared for mpfr and mpc

MehdiChinoune:
This is my last attempt to build shared libraries for mpfr and mpc.
I have proposed this change before but @jeremyd2019 said that we shloud keep gmp as it is and the issue comes from mpfr and mpc being too strict, but It turns out It wasn't the case and he was wrong about that.
I think building shared libraries should be prioritized over static ones, If there is a conflict. So if someone wants static mpfr and mpc then he should find the solution.

### DIFF
--- a/mingw-w64-gmp/PKGBUILD
+++ b/mingw-w64-gmp/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gmp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=6.2.1
-pkgrel=3
+pkgrel=4
 pkgdesc="A free library for arbitrary precision arithmetic (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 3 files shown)


MehdiChinoune:
I should add that shared mpfr and mpc are required to enable their usage with openturns and also to fix building suitesparse/SPEX on GCC envs.

jeremyd2019:
> I have proposed this change before but @jeremyd2019 said that we shloud keep gmp as it is and the issue comes from mpfr and mpc being too strict, but It turns out It wasn't the case and he was wrong about that. 

This was something that was told to me a couple years ago, but I do believe it is true, still.  gmp seems fine.  I did not investigate mpfr, but it attempting to dllexport/dllimport thread_local variables is a no-no even for MSVC (it prevents enabling both threadsafe and shared in that case, and apparently should for any native TLS environment such as our CLANG*, and building the whole stack shared wouldn't solve this).

I don't know what the best solution here would be.  If mpfr is just exporting those thread_local caches for their tests to be able to access, it might be fine to just ignore them and ship the shared lib anyway.  If they are exported so that macros or inline functions could use them, or worse if they are part of the public API, that might be fatal for some project.

If this isn't the issue that you are referring to, I apologize, I haven't been following too closely lately.

MehdiChinoune:
Seems like no one is against this change!

jeremyd2019:
I'm not really for or against it.  I still think installing the 'static' variant of the headers is more correct for these, based on how they work.  Also, my biggest concern is breaking gcc (these packages are going to have to use @lazka's trick of package an old soname dll on update, rebuild gcc, and then rebuild these without the old dll).  But it doesn't really affect me, I don't generally mess with gcc :wink:

mmuetzel:
> I should add that shared mpfr and mpc are required to enable their usage with openturns and also to fix building suitesparse/SPEX on GCC envs.

It doesn't seem to be required for suitesparse/SPEX at least.

MehdiChinoune:
Any objection?

PR #14574, opened 2022-12-13
suitesparse: build SPEX with GCC

mmuetzel:
Use upstream patch to fix linking the SPEX library with gcc.
Also, remove the Fortran compiler as a build dependency.
Remove `metis` from dependencies.
Remove unnecessary `cmake` flags.


mmuetzel:
The pkgrel of dependent packages have already been bumped as part of #14156. So, it's probably not necessary to bump them again.

mmuetzel:
I re-triggered the CI after the dependent packages moved from staging to the stable repositories.
According to the package grokker, it's indeed not necessary to rebuild anything again.

PR #14581, opened 2022-12-15
kcoreaddons-qt5: Remove redundant patch file disabling FAM

Biswa96:


Biswa96:
This does not change any actual code. So, pkgrel bump is not required.

Also might be related to removal of gamin in msys2 https://github.com/msys2/MSYS2-packages/commit/aec9a2b0999813d8da7cdac93fb493f0f22b7549

PR #14584, opened 2022-12-15
libtiff: update to 4.5.0

kmilos:
soversion changed, so needs to be followed up by a massive rebuild of deps, and will need help with that


kmilos:
Thanks!

Issue #10725, opened 2022-02-10
Maybe it's time to upgrade ffmpeg to 5.0


wszqkzqk:
[Archlinux have already upgraded ffmpeg to 5.0 in staging repo](https://archlinux.org/packages/staging/x86_64/ffmpeg/), so maybe it's for time for MSYS2 to make plans to upgrade to ffmpeg 5.0……？

Biswa96:
It is in stating repository for a reason 😉

lazka:
https://archlinux.org/todo/ffmpeg-50/

jeroen:
FYI, I maintain some bindings for ffmpeg libav and there were quite a few breaking changes in ffmpeg-5. Mainly just headers that were previously included automatically that you now need to `#include` explicitly, but also some API's that are no longer public. So there may be some bumps in the road.

wszqkzqk:
[mingw-w64-x86_64-ffmpeg-5.0-4-any.pkg.tar.zst](https://github.com/wszqkzqk/mingwpkgcloudbuild/releases/download/mingw64/mingw-w64-x86_64-ffmpeg-5.0-4-any.pkg.tar.zst)
I‘ve built the ffmpeg 5.0 package, but I have not tested it.

ZachBacon:
> [mingw-w64-x86_64-ffmpeg-5.0-4-any.pkg.tar.zst](https://github.com/wszqkzqk/mingwpkgcloudbuild/releases/download/mingw64/mingw-w64-x86_64-ffmpeg-5.0-4-any.pkg.tar.zst)
> I‘ve built the ffmpeg 5.0 package, but I have not tested it.

The main problem is software that rely on ffmpeg needs time to adapt. If anything it would need go through a bunch of testing and a bunch of rebuilds of the software that depend on it. 

I personally vote for a separate package until we can be sure it can be made default.

lazka:
Arch has updated now, so feel free to try (and look for patches in Arch if needed)

Biswa96:
ArchLinux updated the ffmpeg package to v5 as a separate package. The older V4 is packaged as ffmpeg4.4 packge which is linked with other projects like vlc.

lazka:
ah, thanks, I missed that

lmorand:
For reference, I've tried building every package that depends on ffmpeg and this was the result (a check mark means it built successfuly using ffmpeg 5.1.2, but no functionality was tested):

[...]

PR #14591, opened 2022-12-16
ffmpeg: update to 5.1.2

lmorand:


Biswa96:
This may not be possible as-is

1. Either all the dependent packages need to be rebuilt for ffmpeg v5.
2. Or ffmpeg v4 and v5 packages should have clear separation so that both can be installed side-by-side.

See this discussion https://github.com/msys2/MINGW-packages/issues/10725

lmorand:
I see; I'll close this then. Is there anything specific I could do to help with this? I'm not very knowledgeable in this whole ecosystem, but I'm willing to try.

Biswa96:
You can work any of the points that I have mentioned. Not all programs have moved to ffmpeg v5. So, separating ffmpeg v4 and v5 may be the easier way to start. Feel free to discuss or share any idea 😊

PR #14592, opened 2022-12-16
rust: update to 1.66.0

filnet:


mati865:
Upstream bug: https://github.com/rust-lang/rust/issues/105696

filnet:
After removing git we now have:
```thread 'main' panicked at 'missing miri', install.rs:205:14```

**Comment by filnet** [comment]
Somewhat related: https://github.com/rust-lang/rust/commit/6cfa7ef2ba1a4781605697c78901c9c10353c58f ?

**Comment by mati865** [comment]
Sorry, forgot to cross-link miri issue: https://github.com/rust-lang/rust/issues/105816
Your patch should solve the problem for us.

**Comment by filnet** [comment]
Looks like my hack did the trick but hacks are pilling up...

Dumb question: rust supports mingw, but is this mingw or another one ? I never fully grasped the history of cygwin/msys/mingw.

**Comment by mati865** [comment]
If we translate our subsystems to Rust's support then:
- MINGW32 and MINGW64 are tier 1 (distibuted and tested for each commit),
- UCRT64 is unsupported but happens to work so far,
- CLANG* is Tier 3 so it's not distributed or tested.

Git issue is related to the fact we have Cygwin git but Rust only tests using native Git for Windows. Miri issue I think affects all or most of the platforms.

**Comment by mati865** [comment]
@jeremyd2019 do you think we want to keep bootstrapping hacks (AArch64 would require updating to even work)?
I think we could set `_bootstrapping=no` for all targets (or at least all CLANG* subsystems) and drop the hacks eventually in separate PR.

**Comment by Biswa96** [comment]
> rust supports mingw, but is this this mingw or another one ?

I was just wondering, should there be a page for rust to document the differences and issues like [python][1] ?

[1]: https://github.com/msys2/msys2.github.io/blob/source/web/docs/python.md

**Comment by jeremyd2019** [comment]
I haven't been paying very close attention to msys2 for a bit...

> @jeremyd2019 do you think we want to keep bootstrapping hacks (AArch64 would require updating to even work)? I think we could set `_bootstrapping=no` for all targets (or at least all CLANG* subsystems) and drop the hacks eventually in separate PR.

My only concern would be that the 'institutional knowledge' of how to bootstrap might get lost.  It would be in the git history, but would whoever might need to try in the future think to look there?  Hopefully, upstream will eventually get their bootstraps in order and the necessary hacks will be fewer in future

**Comment by mati865** [comment]
> Hopefully, upstream will eventually get their bootstraps in order and the necessary hacks will be fewer in future

They would "only" need to provide prebuilt binaries which was so far blocked on dependencies that didn't build on with `gnullvm` targets and on free builders count. I hope to find some time in January to look into 32-bit target to see why some tests are failing and ask them if they have capacity to build more targets.

**Comment by lazka** [comment]
merge?

**Comment by mati865** [comment]
It builds so why not?

**Comment by lazka** [comment]
I'm never sure with aarch64 :)

**Comment by jeremyd2019** [comment]
```
    = note: clang: warning: argument unused during compilation: '-nolibc' [-Wunused-command-line-argument]
            clang: warning: argument unused during compilation: '-no-pie' [-Wunused-command-line-argument]
            lld: error: unable to find library -lwindows
            
  
  error: could not compile `bootstrap` due to previous error
```

Was the patch force-updating the vendored crates lost somehow?

**Comment by mati865** [comment]
Do you have full log? I can look into it tomorrow.
BTW vendoring patch will be unnecessary with Rust 1.67 in another 6 weeks.

**Comment by jeremyd2019** [comment]
Just https://github.com/msys2-arm/msys2-autobuild/actions/runs/3760773308/jobs/6391858457

**Comment by lazka** [comment]
ucrt64 failed in autobuild:

```
2022-12-23T16:15:12.0013038Z    Compiling bootstrap v0.0.0 (C:\M\mingw-w64-rust\src\rustc-1.66.0-src\src\bootstrap)
2022-12-23T16:15:12.0123392Z    Compiling itoa v1.0.2
2022-12-23T16:15:12.5104614Z    Compiling sha2 v0.10.2
2022-12-23T16:15:13.1097246Z    Compiling getopts v0.2.21
2022-12-23T16:15:13.8870914Z    Compiling tar v0.4.38
2022-12-23T16:15:14.2742849Z    Compiling fd-lock v3.0.6
2022-12-23T16:15:14.4147607Z    Compiling ignore v0.4.18
2022-12-23T16:15:15.6225766Z    Compiling xz2 v0.1.6
2022-12-23T16:15:15.9885470Z    Compiling toml v0.5.9
2022-12-23T16:15:16.7363179Z    Compiling cmake v0.1.48
2022-12-23T16:15:17.3358695Z    Compiling opener v0.5.0
2022-12-23T16:15:17.6133228Z    Compiling hex v0.4.3
2022-12-23T16:15:42.5721253Z     Finished dev [unoptimized] target(s) in 1m 08s
2022-12-23T16:15:42.6521632Z thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }', config.rs:863:74
2022-12-23T16:15:42.6522421Z stack backtrace:
2022-12-23T16:15:42.7482619Z    0: rust_begin_unwind
2022-12-23T16:15:42.7488160Z              at /rustc/897e37553bba8b42751c67658967889d11ecd120/library\std\src/panicking.rs:584:5
2022-12-23T16:15:42.7490004Z    1: core::panicking::panic_fmt
2022-12-23T16:15:42.7490798Z              at /rustc/897e37553bba8b42751c67658967889d11ecd120/library\core\src/panicking.rs:142:14
2022-12-23T16:15:42.7491879Z    2: core::result::unwrap_failed
2022-12-23T16:15:42.7497480Z              at /rustc/897e37553bba8b42751c67658967889d11ecd120/library\core\src/result.rs:1785:5
2022-12-23T16:15:42.7498823Z    3: core::result::Result<T,E>::unwrap
2022-12-23T16:15:42.7499476Z              at /rustc/897e37553bba8b42751c67658967889d11ecd120\library\core\src/result.rs:1107:23
2022-12-23T16:15:42.7500015Z    4: bootstrap::config::Config::parse
2022-12-23T16:15:42.7500972Z              at C:\M\mingw-w64-rust\src\rustc-1.66.0-src\src\bootstrap\config.rs:863:28
2022-12-23T16:15:42.7501490Z    5: bootstrap::main
2022-12-23T16:15:42.7502117Z              at C:\M\mingw-w64-rust\src\rustc-1.66.0-src\src\bootstrap\bin\main.rs:14:18
2022-12-23T16:15:42.7503073Z    6: core::ops::function::FnOnce::call_once
2022-12-23T16:15:42.7503643Z              at /rustc/897e37553bba8b42751c67658967889d11ecd120\library\core\src\ops/function.rs:248:5
2022-12-23T16:15:42.7504253Z note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
2022-12-23T16:15:42.7537375Z Build completed unsuccessfully in 0:01:31
```

mati865:
Was `git` package installed there?

lazka:
Yes, the ucrt64 builder installs the `VCS` group for building source packages. I could look into trying to avoid that.

mati865:
Alternatively we could disable git detection in Rust's build, it doesn't like UNIX style paths served by emulated Git.

> ```
>     = note: clang: warning: argument unused during compilation: '-nolibc' [-Wunused-command-line-argument]
>             clang: warning: argument unused during compilation: '-no-pie' [-Wunused-command-line-argument]
>             lld: error: unable to find library -lwindows
>             clang: error: linker command failed with exit code 1 (use -v to see invocation)
>             
>   
>   error: could not compile `bootstrap` due to previous error
> ```
> 
> Was the patch force-updating the vendored crates lost somehow? Or does it need to be updated?

Looks like it haven't used vendored dependencies.

lazka:
> Alternatively we could disable git detection in Rust's build, it doesn't like UNIX style paths served by emulated Git.

I wonder if `export GIT_DIR=/dev/null` would work

mati865:
Build process calls `git rev-parse --show-toplevel` so yeah, should do the trick.

jeremyd2019:
> > ucrt64 failed in autobuild (two times):
> 
> Was `git` package installed there?

Might be a problem on arm too, my runner has some more packages installed than the clean-from-scratch runners, and I think git is there...

lazka:
> I wonder if `export GIT_DIR=/dev/null` would work

It did.

jeremyd2019:
@mati865 let me know if you need me to test any patches on arm64.  I think the issue would happen to any `-gnullvm` target though, so it'd probably be faster (both in terms of compute power and turn-around) for you to test against x86_64

mati865:
On the initial test it seems to work but I need to do some digging. Do you have `CARGO_HOME` defined by any chance?

jeremyd2019:
No special env vars like that defined.  But there is a `.cargo` directory in the Windows user profile, it's likely to have cruft from prior runs in it.  I can try deleting that and seeing if that helps.

lazka:
same error as before in the new run

mati865:
Couldn't reproduce but noticed that `.cargo` dir inside Rust source dir disapperars. I'll have ony a phone until tomorrow.
Maybe you could do build with something like `stat <path>/.cargo || echo absent` put all over the place inside `build()` to discover when it goes missing and verify if it's copied to build directory?

lazka:
Reverted for now, to unblock the build queue.

jeremyd2019:
> Couldn't reproduce but noticed that `.cargo` dir inside Rust source dir disapperars. I'll have ony a phone until tomorrow. Maybe you could do build with something like `stat <path>/.cargo || echo absent` put all over the place inside `build()` to discover when it goes missing and verify if it's copied to build directory?

There is only one command after copying `.cargo`, running `x.py install`.

jeremyd2019:
`.cargo` exists both before and after x.py install fails.

jeremyd2019:
`--enable-vendor` is missing from config.toml.  Maybe `${_rust_conf}` needs to be `"${_rust_conf[@]}"` to configure?  Trying that now

jeremyd2019:
that seems to have been it.  I put in #14777 that reverts the revert and fixes the usage of `_rust_conf` array

mati865:
Oh, nice catch.

PR #14593, opened 2022-12-16
angleproject: update to 2.1.2c60b86e3

raedrizqie:


Biswa96:
Just curious. Instead of checking environment variables, would it be same if `is_mingw` and `is_clang` flag are passed to `--args=""` option of gn?

raedrizqie:
yes.. its the same and less patching :)

btw.. d3d11 would be supported if some of the missing headers are present.. maybe you could add those missing headers upstream?

Biswa96:
> maybe you could add those missing headers upstream?

Yeah, noticed that. Those headers are imported from wine and wine is now in feature freeze mode. I may not be able to upstream all of the required types. If you look at the [AUR's mingw angleproject][1] the author used his [own repository][2] to dx headers. Wondering if those can be used in meantime.

[1]: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mingw-w64-angleproject
[2]: https://github.com/Martchus/additional-mingw-header.git

raedrizqie:
i see.. ok thanks

Biswa96:
> d3d11 would be supported if some of the missing headers are present.. 

Bad news. It requires winrt headers which itself takes eons to upstream.

```
src/libANGLE/renderer/d3d/d3d11/converged/CompositorNativeWindow11.h:18:10: fatal error: windows.ui.composition.h: No such file or directory
```

For the test build, I have enabled these

```
    angle_enable_d3d11=true
    angle_enable_d3d9=true
    angle_enable_gl=true
    angle_enable_gl_desktop_frontend=true
    angle_enable_essl=true
    angle_enable_glsl=true
    angle_enable_hlsl=true
```

> ...the author used his [own repository](https://github.com/Martchus/additional-mingw-header.git) to dx headers. Wondering if those can be used in meantime.

Answering my own question. Those headers would not work. The mingw-w64-angleproject in AUR is ancient. It is tracking commit from 02 Dec 2016 https://chromium.googlesource.com/angle/angle/+/9f09037

raedrizqie:
well, at least we have d3d9 :)

alvinhochun:
This struck me as odd because in Krita we build a somewhat recent version of ANGLE (f2280c0c5 from 2022-05-03) with mingw-w64 using the CMake build borrowed from vcpkg, instead of the quite dated version that comes with Qt 5.12, and we rely heavily on its d3d11 support. Then I looked into it and see that `Windows.UI.Composition` is only used in `CompositorNativeWindow11` and the CMake build explicitly excludes this file on mingw and also do not define `ANGLE_ENABLE_D3D11_COMPOSITOR_NATIVE_WINDOW`, which explains why it works for us with no issues.

mingw-w64-angleproject/003-angle-src-fixes.patch
```diff
@@ -0,0 +1,50 @@
...
+--- a/src/common/mathutil.h
++++ b/src/common/mathutil.h
+@@ -1027,8 +1027,8 @@ inline uint32_t BitfieldReverse(uint32_t value)
+ }
+ 
+ // Count the 1 bits.
+-#if defined(_MSC_VER) && !defined(__clang__)
+-#    if defined(_M_IX86) || defined(_M_X64)
++#if defined(_WIN32) && !defined(__clang__)
++#    if defined(_M_IX86) || defined(_M_X64) || defined(__X86__) || defined(__x86_64__)
```
Biswa96, line 12: Is `__X86__` valid here? Shouldn't it be `__i386__` or `__i686__` ?

raedrizqie: yeah.. which one is better? i386 or i686?

Biswa96: I am not sure about their difference in modern days. Though `__i386__` is used widely in mingw-w64. @mati865 or @jeremyd2019 may provide correct answer.

PR #14638, opened 2022-12-18
cmake: Enable SOVERSION in shared library names

MehdiChinoune:


Biswa96:
Wouldn't this enable SOVERSION in shared library names which do not have SOVERSION at present? Should this option be set at configure time instead of by-default?

MehdiChinoune:
> Wouldn't this enable SOVERSION in shared library names which do not have SOVERSION at present?

Yes, It does
>Should this option be set at configure time instead of by-default?

I am waiting for feedback from other contributors.

lazka:
hm, I'm a bit worried of potential fallout for users depending on the naming somehow... hard to say :/

Biswa96:
> potential fallout for users depending on the naming somehow...

Yeah, I have seen some projects use DLLs names with wildcard to create an installer. For example, Inkscape has this https://gitlab.com/inkscape/inkscape/-/blob/master/CMakeScripts/InstallMSYS2.cmake

So, I think SOVERSION should be enabled in individual PKGBUILD which already has versioned DLLs.

MehdiChinoune:
> Yeah, I have seen some projects use DLLs names with wildcard to create an installer. For example, Inkscape has this https://gitlab.com/inkscape/inkscape/-/blob/master/CMakeScripts/InstallMSYS2.cmake

That's not an issue, Projects like Inkscape maintains that list and fix it when It's necessary (as they should).
See the history of that file.
https://gitlab.com/inkscape/inkscape/-/commit/8f51041b3a38fc5a46418faa23bc7a83e513b803
https://gitlab.com/inkscape/inkscape/-/commit/c0d40a2e9872afd7a91a4a952b1c59398aba18d1
https://gitlab.com/inkscape/inkscape/-/commit/4cb4922157ddef128709c8998e87b8c422c93f39
https://gitlab.com/inkscape/inkscape/-/commit/411cb25859fb8c250c8196dc4a02ce18e88393f4
https://gitlab.com/inkscape/inkscape/-/commit/c6a6fd43b51a67fba2d6526c29efa915d3c35d5b

lazka:
Would there be some middle ground, like

* if there is an ABI break which requires a rebuild we enable CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION  for that package
* and we can also add it to our PKGBUILD templates by default

?

Issue #14637, opened 2022-12-18
[7zip] 7z.dll missing?


prgwiz:
### Description / Steps to reproduce the issue

E:\msys64\mingw64\bin>7z a "c:\swap\test" "c:\swap\uttpd2-main\*"

7-Zip 22.01 (x64) : Copyright (c) 1999-2022 Igor Pavlov : 2022-07-15

Codec Load Error: E:\msys64\mingw64\bin\7z.dll : The specified module could not be found.

System ERROR:
Not implemented

### Expected behavior

Would expect 7z to archive the files in the dir.

### Actual behavior

Either the 7z.exe does not need a 7z.dll or is missing from the package.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-22621

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Biswa96:
Would you like to test the packages from https://github.com/msys2/MINGW-packages/actions/runs/3723571555 ?

prgwiz:
Tested and works.  Thank you.

PR #14639, opened 2022-12-18
7zip: Build 7z.dll

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/14637

### DIFF
--- a/mingw-w64-7zip/PKGBUILD
+++ b/mingw-w64-7zip/PKGBUILD
@@ -4,7 +4,7 @@ _realname=7zip
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=22.01
-pkgrel=2
+pkgrel=3
 pkgdesc="A file archiver with a high compression ratio (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
@@ -20,7 +20,7 @@ sha256sums=('393098730c70042392af808917e765945dc2437dee7aae3cfcc4966eb920fbc5'
 prepare () {
   case ${MSYSTEM} in
     CLANG*)
-      sed -i 's/-Werror/& -Wno-error=missing-exception-spec -Wno-error=unused-but-set-variable -Wno-error=unknown-attributes/' */*.mak */*/*.mak
+      sed -i 's/-Werror/ -Wno-missing-exception-spec -Wno-unused-but-set-variable -Wno-unknown-attributes/' */*.mak */*/*.mak
       # This patch can be dropped once we're using Clang 16 or later
       patch -p0 -i "${srcdir}/clang-arm-target-fixes.patch"
       ;;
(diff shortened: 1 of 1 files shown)


Biswa96:
@dscho Would you like to review this pull request please?

mingw-w64-7zip/PKGBUILD
```diff
@@ -30,13 +30,15 @@ prepare () {
...
 
+  make -C CPP/7zip/Bundles/Format7zF -f makefile.gcc
   make -C CPP/7zip/UI/Console -f makefile.gcc
   make -C CPP/7zip/Bundles/SFXCon -f makefile.gcc
 }
 
 package() {
   cd "${srcdir}"
 
+  install -D -m755 CPP/7zip/Bundles/Format7zF/_o/7z.dll "${pkgdir}"${MINGW_PREFIX}/bin/7z.dll
```
dscho, line 41: Is this the only `.dll` that now needs to be copied?

dscho (commented):
Thank you for fixing this!

mingw-w64-7zip/PKGBUILD
Biswa96: That's the only DLL compiled in that directory.

Biswa96:
I am not sure how to fix the arm64 build. May I ask for help please? @jeremyd2019

mingw-w64-7zip/PKGBUILD
dscho: I just wonder how future-proof it is to hard-code the path to the currently-only DLL file.

Biswa96: I did not fully understand the question. Could you rephrase the question please? A diff suggestion may help.

BTW, I am not good at making things stable 👶

jeremyd2019:
```
../../../../C/Sha1Opt.c:313:49: warning: call to undeclared function 'vsha1h_u32'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```
looks familiar.  Probably needs to be patched to include some arm intrinsic header, to use a form of target attribute that clang accepts, and to `#define` something to convince the arm header to actually include the required intrinsics even though the instructions aren't enabled globally.

jeremyd2019:
Ah, see the patch for Sha256Opt.c in https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-7zip/clang-arm-target-fixes.patch and try to apply similar to Sha1Opt.c?

mingw-w64-7zip/PKGBUILD
dscho:
```suggestion
  install -D -m755 $(find CPP/7zip/Bundles/ -name \*.dll) "${pkgdir}"${MINGW_PREFIX}/bin/
```

Biswa96: It did not change anything in the package though. Should I add that for any other reason?

dscho: > It did not change anything in the package though.

Of course not. That would be a bug.

> Should I add that for any other reason?

Certainly. The `find` command is robust against future changes in 7-Zip. Hard-coding file names is always fragile.

Biswa96: > The find command is robust against future changes in 7-Zip.

According to that point, there is a possibility that Format7zF/_o/7z.dll path will be changed. Would not there a similar chance of having multiple DLL files in Format7zF/_o/ folder?

dscho: Let's not jump from one extreme (totally fragile code) to the other extreme (totally over-engineered yet robust code).

PR #14659, opened 2022-12-19
7zip: fix arm64 build

jeremyd2019:
Addresses https://github.com/msys2/MINGW-packages/pull/14639#issuecomment-1357848517

### DIFF
--- a/mingw-w64-7zip/PKGBUILD
+++ b/mingw-w64-7zip/PKGBUILD
@@ -4,7 +4,7 @@ _realname=7zip
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=22.01
-pkgrel=3
+pkgrel=4
 pkgdesc="A file archiver with a high compression ratio (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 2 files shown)


Biswa96 (approved):
[I have no idea what I am reading](https://youtu.be/rR4n-0KYeKQ?t=45) :)

jeremyd2019:
Clang doesn't currently handle the `__attribute__((__target__("+something")))` format, but does the right thing without the `+` (IIRC, this is supposed to be fixed in clang 16, but don't quote me on that).

Those preprocessor defines are meant to be defined by the compiler when the corresponding instruction sets are enabled at compile time, and arm_neon.h uses them to decide which intrinsics to define.  But, since this code is only enabling the instruction sets on a per-function basis, the compiler did not define the corresponding defines, so we have to do it manually to convince arm_neon.h to define the functions we need.

Biswa96:
> Those preprocessor defines are meant to be defined by the compiler

Just curious. Is it possible to add an compiler flag which define those macros implicitly? For example, in x86_64

```
$ gcc -mavx -dM -E - < /dev/null | grep -i avx
#define __AVX__ 1
```

jeremyd2019:
> > Those preprocessor defines are meant to be defined by the compiler
> 
> Just curious. Is it possible to add an compiler flag which define those macros implicitly? For example, in x86_64
> 
> ```
> $ gcc -mavx -dM -E - < /dev/null | grep -i avx
> #define __AVX__ 1
> ```

That's what happens.  But this code is not using a command-line flag to enable the instructions for the whole compilation unit, they are only enabling it on a function-by-function basis with the target attribute on the functions.

PR #14662, opened 2022-12-20
mcfgthread: Update to 1.4-ga.1

lhmouse:
1. Use `build-${MSYSTEM}` as build directories, like other packages.
2. Remove the useless `--target` for `configure`.
3. Add C++ tests.

For the record: Some i686 test programs on my Windows 10 machine fail because the result of `QueryPerformanceCounter()` occasionally goes backwards. It is of course unexpected but I don't know whether it's a Windows bug or normal behavior. All the x86_64 tests pass without such issues.


lhmouse:
Thank you.

Issue #14658, opened 2022-12-19
SuiteSparse 6 crashes on 32-bit


mmuetzel:
### Description / Steps to reproduce the issue

1. Install `octave` for MINGW32: `pacman -S mingw-w64-i686-octave`.
2. Start Octave with `octave` from the MINGW32 shell.
3. Execute, e.g., the following commands:
```
a = sparse (5, 6);
a(3,1) = 0.8;
a(2,2) = 1.4;
a(1,6) = -0.5;
r = qr (a);
```

### Expected behavior

The QR decomposition of `a` should be returned.

### Actual behavior

Octave crashes with the following output:
```
$ octave
GNU Octave, version 7.3.0
Copyright (C) 1993-2022 The Octave Project Developers.
This is free software; see the source code for copying conditions.
There is ABSOLUTELY NO WARRANTY; not even for MERCHANTABILITY or
FITNESS FOR A PARTICULAR PURPOSE.  For details, type 'warranty'.

Octave was configured for "i686-w64-mingw32".

Additional information about Octave is available at https://www.octave.org.

Please contribute if you find this software useful.
For more information, visit https://www.octave.org/get-involved.html

Read https://www.octave.org/bugs.html to learn how to submit bug reports.
For information about changes from previous versions, type 'news'.

octave:1> a = sparse (5, 6);
octave:2> a(3,1) = 0.8;
octave:3> a(2,2) = 1.4;
octave:4> a(1,6) = -0.5;
octave:5> r = qr (a);
fatal: caught signal Segmentation fault -- stopping myself...
```

I compiled SuiteSparse with debug flags and without stripping symbols. With that version, I see the following top of the backtrace when catching the segfault with `gdb`:
```
Thread 7 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 4692.0x7938]
0x56672a86 in cholmod_l_postorder (Parent=Parent@entry=0x1e1f0d20, n=<optimized out>, n@entry=0, Weight=Weight@entry=0x0, Post=Post@entry=0x1e1f0d20, Common=Common@entry=0x1e17ef80) at D:/mingw-w64-suitesparse/src/build-MINGW32/CHOLMOD/Cholesky/cholmod_postorder.c:202
warning: Source file is more recent than executable.
202                 p = Parent [j] ;
(gdb) bt
#0  0x56672a86 in cholmod_l_postorder (Parent=Parent@entry=0x1e1f0d20, n=<optimized out>, n@entry=0,
[...]

PR #14663, opened 2022-12-20
Add package for SuiteSparse 5

mmuetzel:
This addresses #14658.

Most of the libraries in SuiteSparse have an API that can be categorized into two groups: One narrow indexing API and one wide indexing API. In version 5 and version 6, the narrow indexing API uses 32-bit integers. In version 5, the wide indexing API used 64-bit integers on 64-bit platforms and 32-bit integers on 32-bit platforms. In version 6, the wide indexing API uses 64-bit independent of the platform. That cannot work for 32-bit platforms. Indices must not be larger than `sizeof(void *)`. That means half of the API is broken on 32-bit platforms for those libraries.

Some libraries (e.g., SPQR) have only the wide indexing API. That means, they are completely broken on 32-bit platforms.

This PR adds a package that is basically the version of SuiteSparse as it was packaged before the update to version 6.
It currently reverts to SuiteSparse 5 only for 32-bit Octave. A couple of other packages probably also need to be rebuilt with the older version of SuiteSparse (likely `ceres-solver` and `python-cvxopt`). The other dependers don't need to be *rebuilt*. But they should probably still *depend* on SuiteSparse 5 for their 32-bit packages.

I'm not sure if the approach in this PR is the right way to handle this situation.
I'd be happy for any feedback.

### DIFF
--- a/mingw-w64-suitesparse5/PKGBUILD
+++ b/mingw-w64-suitesparse5/PKGBUILD
@@ -17,7 +17,7 @@ source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/S
         "0004-mingw-w64-install-static-libs.patch"
         "0005-suitesparse-5.8.1-skip-building-Mongoose-GraphBLAS.patch"
         "0006-suitesparse-5.8.1-fix-mp-link-order.patch")
-provides=("${MINGW_PACKAGE_PREFIX}-suitesparse")
+provides=("${MINGW_PACKAGE_PREFIX}-suitesparse=${pkgver}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-suitesparse")
 replaces=("${MINGW_PACKAGE_PREFIX}-suitesparse")
 depends=("${MINGW_PACKAGE_PREFIX}-openblas"
--- a/mingw-w64-suitesparse5/PKGBUILD
+++ b/mingw-w64-suitesparse5/PKGBUILD
@@ -19,7 +19,6 @@ source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/S
         "0006-suitesparse-5.8.1-fix-mp-link-order.patch")
 provides=("${MINGW_PACKAGE_PREFIX}-suitesparse=${pkgver}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-suitesparse")
-replaces=("${MINGW_PACKAGE_PREFIX}-suitesparse")
 depends=("${MINGW_PACKAGE_PREFIX}-openblas"
          "${MINGW_PACKAGE_PREFIX}-omp"
          "${MINGW_PACKAGE_PREFIX}-mpfr"
--- a/mingw-w64-suitesparse5/PKGBUILD
+++ b/mingw-w64-suitesparse5/PKGBUILD
@@ -9,7 +9,7 @@ pkgdesc='A suite of sparse matrix algorithms (mingw-w64)'
 url="https://faculty.cse.tamu.edu/davis/suitesparse.html"
 license=('spdx:LGPL-3.0-or-later AND BSD-3-Clause AND LGPL-2.1-or-later AND GPL-2.0-or-later AND LGPL-2.0-or-later AND Apache-2.0 AND GPL-3.0-or-later')
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+mingw_arch=('mingw32' 'clang32')
 source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/v${pkgver}.tar.gz"
         "0001-mingw-w64-Use-a-not-lib-as-AR_TARGET-extension.patch"
         "0002-mingw-w64-Set-SO_OPTS--shared-move-dlls-create-import-libs.patch"
--- a/mingw-w64-suitesparse5/PKGBUILD
+++ b/mingw-w64-suitesparse5/PKGBUILD
@@ -1,0 +1,113 @@
+# Maintainer: Ray Donnelly <mingw.android@gmail.com>
+
+_realname=suitesparse5
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-suitesparse5")
+pkgver=5.13.0
+pkgrel=1
+pkgdesc='A suite of sparse matrix algorithms (mingw-w64)'
+url="https://faculty.cse.tamu.edu/davis/suitesparse.html"
+license=('spdx:LGPL-3.0-or-later AND BSD-3-Clause AND LGPL-2.1-or-later AND GPL-2.0-or-later AND LGPL-2.0-or-later AND Apache-2.0 AND GPL-3.0-or-later')
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/v${pkgver}.tar.gz"
+        "0001-mingw-w64-Use-a-not-lib-as-AR_TARGET-extension.patch"
+        "0002-mingw-w64-Set-SO_OPTS--shared-move-dlls-create-import-libs.patch"
+        "0003-mingw-w64-no-CUDA.patch"
+        "0004-mingw-w64-install-static-libs.patch"
+        "0005-suitesparse-5.8.1-skip-building-Mongoose-GraphBLAS.patch"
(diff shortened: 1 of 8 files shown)


mmuetzel:
@MehdiChinoune: Could you please comment with your thoughts on this (because you pushed the update to SuiteSparse 6)?

mingw-w64-suitesparse5/PKGBUILD
```diff
@@ -0,0 +1,113 @@
...
+source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/v${pkgver}.tar.gz"
+        "0001-mingw-w64-Use-a-not-lib-as-AR_TARGET-extension.patch"
+        "0002-mingw-w64-Set-SO_OPTS--shared-move-dlls-create-import-libs.patch"
+        "0003-mingw-w64-no-CUDA.patch"
+        "0004-mingw-w64-install-static-libs.patch"
+        "0005-suitesparse-5.8.1-skip-building-Mongoose-GraphBLAS.patch"
+        "0006-suitesparse-5.8.1-fix-mp-link-order.patch")
+provides=("${MINGW_PACKAGE_PREFIX}-suitesparse")
+conflicts=("${MINGW_PACKAGE_PREFIX}-suitesparse")
+replaces=("${MINGW_PACKAGE_PREFIX}-suitesparse")
```
lazka, line 22: replaces results in suitesparse getting replaced by this on upgrades, I'd guess we don't want that

mingw-w64-suitesparse5/PKGBUILD
```diff
@@ -0,0 +1,113 @@
...
+_realname=suitesparse5
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-suitesparse5")
+pkgver=5.13.0
+pkgrel=1
+pkgdesc='A suite of sparse matrix algorithms (mingw-w64)'
+url="https://faculty.cse.tamu.edu/davis/suitesparse.html"
+license=('spdx:LGPL-3.0-or-later AND BSD-3-Clause AND LGPL-2.1-or-later AND GPL-2.0-or-later AND LGPL-2.0-or-later AND Apache-2.0 AND GPL-3.0-or-later')
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
```
lazka, line 12: Should we limit it to 32bit?

mingw-w64-suitesparse5/PKGBUILD
```diff
@@ -0,0 +1,113 @@
...
+source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/v${pkgver}.tar.gz"
+        "0001-mingw-w64-Use-a-not-lib-as-AR_TARGET-extension.patch"
+        "0002-mingw-w64-Set-SO_OPTS--shared-move-dlls-create-import-libs.patch"
+        "0003-mingw-w64-no-CUDA.patch"
+        "0004-mingw-w64-install-static-libs.patch"
+        "0005-suitesparse-5.8.1-skip-building-Mongoose-GraphBLAS.patch"
+        "0006-suitesparse-5.8.1-fix-mp-link-order.patch")
+provides=("${MINGW_PACKAGE_PREFIX}-suitesparse")
```
lazka, line 20:
```suggestion
provides=("${MINGW_PACKAGE_PREFIX}-suitesparse=${pkgver}")
```

mingw-w64-suitesparse5/PKGBUILD
mmuetzel: I'm not sure I understand what that entails. We likely want to remove SuiteSparse 6 when installing any package that depends on SuiteSparse 5.
Would that still happen without the `replaces`?

lazka: replaces will happen at `pacman -Su` time, for every user, not when you install suitesparse. provides/conflicts will ask the user at install time

mingw-w64-suitesparse5/PKGBUILD
mmuetzel: Whatever you prefer. The 64-bit versions still work. But they are probably not strictly necessary.

mingw-w64-suitesparse5/PKGBUILD
mmuetzel: Done

mingw-w64-suitesparse5/PKGBUILD
lazka: I'd prefer fewer packages, we can always expand it if needed.

mmuetzel: Done

mmuetzel:
Is the condition I used in the `PKGBUILD` for octave ok? Should we use the same condition in all packages that depend on `suitesparse` currently?

lazka:
> Is the condition I used in the `PKGBUILD` for octave ok? Should we use the same condition in all packages that depend on `suitesparse` currently?

If you think they are broken too, then yes. But we can also do this in a separate step/PR once octave is happy again.

lazka:
Just wondering, are there any plans to drop Windows 32bit support in octave?

mmuetzel:
> Just wondering, are there any plans to drop Windows 32bit support in octave?

We are running into more and more issues with the 32-bit version. There hasn't been a decision yet. But it's not unlikely that support will be dropped at some point in the not so far future. Octave 9 will likely no longer support Windows 32-bit (unless someone motivated will step in and maintain it).
I was still hoping Octave 8 would run on 32-bit Windows. But, I guess we'll see if it does...

mmuetzel:
> If you think they are broken too, then yes.

The package grokker noticed that `ceres-solver` and `python-cvxopt` needed to be rebuilt for SuiteSparse 6. So, they are likely using the same API that broke for 32-bit applications with the update to SuiteSparse 6.
The other packages probably don't need to be rebuilt. If I understand the `provides` logic correctly, they would also be happy with `suitesparse5` being installed even if they depend on `suitesparse`. Is that right?

lazka:
> The other packages probably don't need to be rebuilt. If I understand the `provides` logic correctly, they would also be happy with `suitesparse5` being installed even if they depend on `suitesparse`. Is that right?

yes, if they are running is another question. With the provides being versioned they could switch to "suitesparse>=6" for that to no longer work.

mmuetzel:
> yes, if they are running/working is another question. With the provides being versioned they could switch to "suitesparse>=6" for that to no longer work.

I'd guess they use the part of the API that didn't change with the update. So, there is probably no need to restrict to newer versions. Adding the version condition would probably also cause issues if users would like to install packages that *need*  SuiteSparse 5 together with package that really don't care which version is there.
If we care about that all packages are actually built with the version of SuiteSparse with which they will be used eventually, we should probably prefer switching all 32-bit packages to `suitesparse5`. (But that is probably not mandatory.)

lazka:
Note that if they use any new symbols then they will fail to link at runtime, even if they never use them normally.

mmuetzel:
Another thing to keep in mind: SuiteSparse is a collection of libraries. Some of those libraries that didn't build with SuiteSparse 5 do build with SuiteSparse 6.
Some dependers might check on compile time which of those libraries are present and working (e.g., Octave does). Depending on that test results, features might be enabled if the "normal" `suitesparse` package is installed that won't work with the `suitesparse5` package.

To be on the safe side, we could rebuild "everything". Or we could wait until issues are reported...
Or as a third option: We could already change the `PKGBUILD` but not bump the pkgrel yet.

mmuetzel:
I installed the build artifacts for MINGW32 from this PR. Using them, the example in #14658 no longer crashes Octave.

mmuetzel:
Is there a way to have the packages built with the `suitesparse5` package on 32-bit, but at the same time allow (some of) them to run with the `suitesparse` package (which is currently at version 6.x)?
Assuming this is possible somehow: Would that be a reasonable approach to allow packages that require `suitesparse5` to be installed side-by-side with packages that would also work with `suitesparse`?

lazka:
> Is there a way to have the packages built with the `suitesparse5` package on 32-bit, but at the same time allow (some of) them to run with the `suitesparse` package (which is currently at version 6.x)?

If the versions are parallel installable, sure. But it doesn't look like they are since there is no version in any of the filenames.

>Assuming this is possible somehow: Would that be a reasonable approach to allow packages that require `suitesparse5` to be installed side-by-side with packages that would also work with `suitesparse`?
> 
> Edit: Would adding `suitesparse5` to `makedepends` while keeping `suitesparse` in depends do that for those packages?

I assume they are not ABI compatible, no?

mmuetzel:
> If the versions are parallel installable, sure. But it doesn't look like they are since there is no version in any of the filenames.

They are not parallel installable. But they should (mostly) work interchangeably.

> I assume they are not ABI compatible, no?

Half of the interface should be ABI compatible between versions. The other half (that is broken in version 6) is not ABI compatible.
Programs that only use functions from the half that didn't break should be working with either version of SuiteSparse.

lazka:
That all seems very brittle to me, I'd rather have things run against the things they build.

PR #14667, opened 2022-12-21
angleproject: enable d3d11

### LINKED ISSUE #14665: Header conflict between angleproject and mingw-w64-headers package

Seems mingw-w64-headers include /KHR/khrplatform.h where angleproject also includes that in that package before

### Expected behavior

file conflict resolution

### Actual behavior

File conflicts since both headers are included in each package.

### Verification

### Windows Version

MINGW64_NT-10.0-22621

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

_No response_

raedrizqie:
Fixes #14665 

- enable d3d11
- removes khr header
- fix license type

### DIFF
--- a/mingw-w64-angleproject/PKGBUILD
+++ b/mingw-w64-angleproject/PKGBUILD
@@ -5,13 +5,13 @@
 _realname=angleproject
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.1.2c60b86e356
-pkgrel=1
+pkgver=2.1.2c60b86e
+pkgrel=2
 pkgdesc='A conformant OpenGL ES implementation for Windows, Mac, Linux, iOS and Android (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
 url='https://chromium.googlesource.com/angle/angle'
-license=('spdx:LGPL-2.0-or-later')
+license=('spdx:BSD-3-Clause')
 depends=("${MINGW_PACKAGE_PREFIX}-jsoncpp"
          "${MINGW_PACKAGE_PREFIX}-libjpeg-turbo"
          "${MINGW_PACKAGE_PREFIX}-libpng"
@@ -32,6 +32,7 @@ source=("${_realname}::git+https://chromium.googlesource.com/angle/angle.git#com
         001-add-mingw-toolchain.patch
         002-buildflags-fixes.patch
         003-angle-src-fixes.patch
+        004-add-missing-d3d11-macro.patch
         jpeg.gn
         jsoncpp.gn
         png.gn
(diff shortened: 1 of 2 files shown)


mingw-w64-angleproject/PKGBUILD
```diff
@@ -24,14 +24,15 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-python"
              "git")
 _commit=("2c60b86e35652689435be5c943fbd2f7fc2d2fd7")
-options=('!buildflags')
+options=('!strip' '!buildflags')
```
lazka, line 27: why is this needed?

raedrizqie: all those exe and dll (reside in exe.unstripped & lib.unstripped dir) are stripped, before they were copied to out\Release-${MSYSTEM}

..and stripping static archives (*.a) on clang give warnings..

lazka: The install size increased from 37mb to 240mb though. Is that still expected?

raedrizqie: ah..ok.. i'll revert it back

Biswa96:
The symbol in the added patch has been _semi-upstreamed_ in wine https://gitlab.winehq.org/wine/wine/-/commit/d059dd1cd0770e9c63da33562f6bcba4ef52846b commit. mingw-w64 will be synced probably next year's wine 8.0 release.

raedrizqie:
May I request a library stub to be upstreamed? it is msxml2 or maybe msxml6? :)

Biswa96:
I will absolutely do that. Can you provide any linking error output?

raedrizqie:
its unrelated to angle package.. just want it for atl/mfc. right now i had to link it with msxml2.lib

Biswa96:
> just want it for atl/mfc. 

How? ATL/MFC is not possible with mingw-w64 yet. Did you port it to mingw-w64? One has to implement those without looking at the actual code by Microsoft. Or the implementation has to be different like old unix vs gnu tools.

raedrizqie:
> How? ATL/MFC is not possible with mingw-w64 yet. Did you port it to mingw-w64? 

yes..

> One has to implement those without looking at the actual code by Microsoft. Or the implementation has to be different like old unix vs gnu tools.

I'm using mfc v10 (the last version that comes with vs makefiles). Only Clang can compiles it because of `if_exists`, `if_not_exists` statements and some other things I can't remember right now..

Of course, the resulting mfc100u.dll fails at runtime though.. memory heap error or something like that..

Biswa96:
Can libuuid.a from mingw-w64 used in place of msxml2/6 ? I see some definitions in msxml-uuid.c.

raedrizqie:
linking with mingw's uuid and msxml2.lib just gives warnings of end directives being corrupt or sime sort.. but the linking was successfull..

Biswa96:
Understood. Let me create msxml2/6 libraries in my fork for your testing. ⌛

PR #14668, opened 2022-12-21
check msys2 tessdata location

mugiseyebrows:
Tesseract looks for `tessdata` in executable directory, but on msys2 data is located in `dirname(executable)/../share/tessdata` (`${MSYSTEM}/share/tessdata`), as a result if TESSDATA_PREFIX is not set appropirately, tesseract fails to perform ocr saying "Error opening data file".
This little patch simplifies using tesseract right out of msys2 tree.

### DIFF
--- a/mingw-w64-tesseract-ocr/0001-check-msys2-tessdata-location.patch
+++ b/mingw-w64-tesseract-ocr/0001-check-msys2-tessdata-location.patch
@@ -1,0 +1,32 @@
+From f346fcdd7cac201b8620cbfc8f329d0199baa44b Mon Sep 17 00:00:00 2001
+From: Doronin Stanislav <mugisbrows@gmail.com>
+Date: Wed, 21 Dec 2022 00:46:16 +0300
+Subject: [PATCH] check msys2 tessdata location
+
+---
+ src/ccutil/ccutil.cpp | 9 +++++++++
+ 1 file changed, 9 insertions(+)
+
+diff --git a/src/ccutil/ccutil.cpp b/src/ccutil/ccutil.cpp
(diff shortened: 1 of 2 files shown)


mingw-w64-tesseract-ocr/0001-check-msys2-tessdata-location.patch
```diff
@@ -0,0 +1,32 @@
...
+---
+ src/ccutil/ccutil.cpp | 9 +++++++++
+ 1 file changed, 9 insertions(+)
+
+diff --git a/src/ccutil/ccutil.cpp b/src/ccutil/ccutil.cpp
+index aad8ef6..18ade02 100644
+--- a/src/ccutil/ccutil.cpp
++++ b/src/ccutil/ccutil.cpp
+@@ -69,6 +69,15 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
+           datadir = subdir;
```
lazka, line 15: hm, shouldn't it be enough to change the following to `../share/tessdata`? https://github.com/tesseract-ocr/tesseract/blob/080da83cc51c4ef8b324a7e03146fe0bd7e0944b/src/ccutil/ccutil.cpp#L67

mugiseyebrows: I assumed that someone might want to deploy tesseract by storing executable and dependencies and tessdata in one directory, it would be good to preserve original behavour as well (as fallback).

lazka: I see, fair enough :+1:

PR #14669, opened 2022-12-21
PKGBUILD-templates: Mention setup option with meson

Biswa96:
Please review and feel free to merge.


Biswa96:
I think it is not required to create a commit which will add `setup` option for all PKGBUILDs. That option can be added when those packages are updated or modified.

PR #14671, opened 2022-12-22
osslsigncode: add clangarm64 target

dennisameling:
This is needed by `git-for-windows` to sign executables. I built it locally with `MINGW_ARCH=clangarm64 makepkg-mingw -sLf` and the binary works as expected:

```
$ osslsigncode --version
osslsigncode 2.5, using:
        OpenSSL 1.1.1s  1 Nov 2022 (Library: OpenSSL 1.1.1s  1 Nov 2022)
        libcurl/7.87.0 OpenSSL/1.1.1s (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.3 libpsl/0.21.1 (+libidn2/2.3.1) libssh2/1.10.0 nghttp2/1.51.0
```

CC @dscho


Biswa96:
FYI, it is not required to set `MINGW_ARCH` explicitly if you are compiling native binary. `MINGW_ARCH` is set to the value of `MSYSTEM` implicitly in makepkg-mingw.

PR #14717, opened 2022-12-23
itk: update to 5.3.0

MehdiChinoune:


Biswa96:
Just wondering. Is it official that itk project does not support 32 bit Windows platform?

MehdiChinoune:
> Just wondering. Is it official that itk project does not support 32 bit Windows platform?

I don't know.

lazka:
Does it not build any more for 32bit?

MehdiChinoune:
> Does it not build any more for 32bit? Dropping it is fine, some reasoning as to why would still be helpful, in case someone asks in the future.

The patch couldn't be applied as It requires updating, Giving that it is an orphan package for medical imaging I don't think 32bit is required.

lazka:
> The patch couldn't be applied as It requires updating, Giving that it is an orphan package for medical imaging I don't think 32bit is required.

OK, thanks for the info

(I guess you mean "leaf package")

MehdiChinoune:
> (I guess you mean "leaf package")

Yes.

lazka:
Here is the upstream issue, if someone wants to fix it upstream to resurrect it: https://github.com/InsightSoftwareConsortium/ITK/issues/2529

revelator:
[mingw-w64-itk.zip](https://github.com/msys2/MINGW-packages/files/10824864/mingw-w64-itk.zip)

this fixes the 32 bit compilation issues.

PR #14719, opened 2022-12-23
angleproject: version bump

raedrizqie:
- update the version to trigger CI rebuild
- rebase some patches

### DIFF
--- a/mingw-w64-angleproject/002-buildflags-fixes.patch
+++ b/mingw-w64-angleproject/002-buildflags-fixes.patch
@@ -1,5 +1,5 @@
 diff --git a/BUILD.gn b/BUILD.gn
-index 3a37b6af9..734cff987 100644
+index 6e965691e..6911fb7c6 100644
 --- a/BUILD.gn
 +++ b/BUILD.gn
 @@ -14,7 +14,10 @@ if (angle_has_build) {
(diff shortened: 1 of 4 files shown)


Biswa96:
The last digit in pkgver may not increment with newer commit. We need some kind of integer which is incremented with every commit. For example, see `git rev-list --count HEAD` part in mingw-w64-headers-git package.

Biswa96:
Now according to vercmp, this is a downgrade, right?

```
$ vercmp "2.1.2c60b86e" "2.1.r20194.6c41793f"
1
```

raedrizqie:
hmm.. so it's better without rev-list then?

Biswa96:
The `.2c60b86e` and `.6c41793f` parts are coming from git commit hash which can be random. It's just coincident that here .6 > .2 but it may not be true all the time.

jeremyd2019:
also version components that start with letters sort less than numbers, so the fact that it's `.r999` is less than `.2xxx`.

raedrizqie:
so how do fix this? remove the package and add it back?

jeremyd2019:
epoch, or just live with it being a downgrade

raedrizqie:
ok.. so it will be downgrade for this version only.. and its ok.. but will CI rebuild the package?

jeremyd2019:
I think so.  @lazka?

lazka:
I wont build older versions, but I can remove the package temporarily if wanted. I don't have anything against epoch.

raedrizqie:
ok.. please remove this temporarily.. thank you.. sorry for the mess

PR #14720, opened 2022-12-23
ceres-solver, python-cvxopt: rebuild with suitesparse5 on 32-bit

mmuetzel:
Rebuild the packages with `suitesparse5` on 32-bit that (probably) use the broken API of `suitesparse`.

That leaves the other dependers of `suitesparse` that should probably also be rebuilt with `suitesparse5` on 32-bit to allow installing them in parallel with the packages that require `suitesparse5`.
See also the discussion in #14663.


mmuetzel:
With respect to the remaining packages: They don't really need to be rebuilt. They have been built before the update to SuiteSparse 6.
Would it be an option to remove `mingw32` and `clang32` from the `mingw_arch` of the `suitesparse` package? Would that pull `suitesparse5` on rebuilds of those dependers? (Is that covered by the `provides` in `suitesparse5`?)

PR #14768, opened 2022-12-25
Remove m2r

MehdiChinoune:
unmaintained, couldn't be built anymore https://github.com/miyakogi/m2r/issues/66 and not required by any other package.


Biswa96:
If required this fork can be built instead https://github.com/crossnox/m2r2

PR #14772, opened 2022-12-26
libepoxy: enable angle backend

raedrizqie:


lazka:
Just wondering, do you have any project in mind that would use this?

raedrizqie:
Just trying to add EGL support in qemu, as @ZachBacon said on discord..

Biswa96:
Should angleproject be mentioned as optional dependency with libepoxy? It seems that libepoxy detects EGL or GLESv2 at runtime.

raedrizqie:
> Should angleproject be mentioned as optional dependency with libepoxy? It seems that libepoxy detects EGL or GLESv2 at runtime.

although it just do dynamic load of libEGL, gtk expect this entrypoint at runtime: `epoxy_eglBindApi` from libepoxy-0.dll

so, either add support to egl for epoxy or don't..

Biswa96:
> gtk expect this entrypoint at runtime

Can you provide the reference of that in gtk code?

raedrizqie:
replace libepoxy-0.dll with the old one (or one that compiled without egl enabled).. you will get that runtime error

lazka:
There is some fallout from this, in case anyone wants to chime in: https://github.com/msys2/MINGW-packages/pull/16357

PR #14777, opened 2022-12-27
rust: fix clangarm64 build

jeremyd2019:
The bash array _rust_conf was being used as a 'scalar' (for lack of a
better term), resulting in only the first item being interpolated,
losing the `--enable-vendor` option necessary for a -gnullvm target to
build due to necessary patches to the vendored sources.  Currently, only
clangarm64 uses a gnullvm target, so it was the only one affected.

Addresses https://github.com/msys2/MINGW-packages/pull/14592#issuecomment-1363332277


lazka:
(I missed the review request, oops)

jeremyd2019:
Just did that to make sure he saw it, not important.

mati865:
Sorry, I've only got home this morning 😓
Looks good, great catch!

PR #14779, opened 2022-12-27
darktable: unbundle libraw

kmilos:
The recently released and included LibRaw 0.21.0 is practically the same as the bundled (approx. RC1) version.


Biswa96:
Is this reported to upstream or an expected behavior? Shouldn't there be some option like DARKTABLE_USE_SYSTEM_LIBRAW or similar?

kmilos:
Yes, I created a PR with such an option on dt master: https://github.com/darktable-org/darktable/pull/13030

dt had to bundle the master tree of LibRaw as 0.20.2 was too old and there was some functionality needed from the dev snapshots/master. This is no longer true, as they are now in sync (until next time I guess as LibRaw stable releases seem to be very infrequent...).

PR #14781, opened 2022-12-27
libinih: new package

kmilos:


Biswa96:
Does the shared library name look good? I mean the 'lib' prefix and `-0` versioned suffix.

kmilos:
Sure, why not? Those are the defaults on MinGW, no (lib prefix and SOVERSION suffix)?

Biswa96:
Just to be confirm.

PR #14782, opened 2022-12-27
editorconfig-qtcreator: Fix makedepends()

MehdiChinoune:


MehdiChinoune:
seems like `editorconfig-qtcreator` is no longer buildable with qt-creator because It's not possible to mix qt5 libs (dependencies of `syntax-highlighting-qt5`) with qt6 ones

podsvirov:
You're right.  We are waiting for KF6 and KF6 support in this plugin.

Biswa96:
> We are waiting for KF6 and KF6 support in this plugin.

It's in the distant future. A gist of KF6 progress can be found here https://iskdeusingqt6.org/

PR #14783, opened 2022-12-27
perl: Add clangarm64 targets for several packages

stephenjust:


Biswa96:
There is no perl aarch64 package https://packages.msys2.org/base/mingw-w64-perl

stephenjust:
Whoops, forgot to submit the PR for -perl. See #14784

Biswa96:
Did you check if it is possible to build perl in clangarm64 environment? @jeremyd2019

stephenjust:
Yes, I did build perl in this environment. The basics work, tests mostly pass/skip. There were a couple test failures, but core functionality is working.

jeremyd2019:
I sort of remember getting perl working early on, and then discovering I didn't actually need it for the 'core' packages I was trying to get going, so I didn't actually enable it (at that point I was only building a very limited few packages, nothing that wasn't absolutely necessary).

stephenjust:
The packages modified here all build with no problem (once perl is built/installed). Not every perl module is so easy though. perl-win32-api (not modified here) has assembly language components that haven't been ported to ARM64 yet.

Issue #14785, opened 2022-12-27
boost update to include CMake configuration files


slavaandrejev:
### Package name

boost

### Other information that may be useful

Boost provides CMake configuration files. These files are the ultimate truth about how Boost libraries are to be compiled. It's better to rely on them, then on CMake provided `FindBoost.cmake` module. The latter can become stale, or it can contain bugs, which will make users dependent on the ability of CMake authors to fix it. Boost provided configuration files are always up-to-date and tested with the version we are installing.

I understand there was issue #5233 in the past, if it is relevant today, maybe we can perform a more thorough investigation of the issue. I am going to try to build 3DTK with Boost's CMake configuration files.

slavaandrejev:
I checked 3DTK, it seems to me they use their own script to find Boost. I also installed boost from the source code using
```
b2 -j14 --toolset=gcc --prefix=$MSYSTEM_PREFIX link=shared,static release install
```
and tried a simple test project with the following `CMakeLists.txt`:
```
cmake_minimum_required(VERSION 3.16)

project(test)

find_package(Boost CONFIG
    REQUIRED program_options
    REQUIRED iostreams
    REQUIRED filesystem
)

add_executable(test
    test.cpp
)

target_link_libraries(test
    Boost::program_options
    Boost::iostreams
    Boost::filesystem
)
```
Everything works like clock, I don't see any problems. Any objections to adding CMake configuration files for boost back?

lazka:
Sure, we can give it a try

slavaandrejev:
Let me prepare a PR.

slavaandrejev:
I am curious about all the patches inside `PKGBUILD`. In my experiment, everything was built from a single command, no patches, no `bootstrap.sh`, no custom `.jam` files. I see from the patches, the biggest effort was about Python and ICU. I'll check if it works without all the patches. It would be nice not to have them because they have to be updated with every Boost release.

slavaandrejev:
I think we can close it now.

PR #14790, opened 2022-12-29
Add back CMake configuration files

slavaandrejev:
Boost provides CMake configuration files. These files are the ultimate truth about how Boost libraries are to be compiled. It's better to rely on them, then on CMake provided FindBoost.cmake module. The latter can become stale, or it can contain bugs, which will make users dependent on the ability of CMake authors to fix it. Boost provided configuration files are always up-to-date and tested with the version we are installing.

### DIFF
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -8,7 +8,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.81.0
 _boostver=${pkgver//./_}
-pkgrel=1
+pkgrel=2
 pkgdesc="Free peer-reviewed portable C++ source libraries (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
@@ -25,7 +25,8 @@ optdepends=("${MINGW_PACKAGE_PREFIX}-python: For Boost.Python"
 makedepends=("${MINGW_PACKAGE_PREFIX}-python"
              "${MINGW_PACKAGE_PREFIX}-python-numpy"
              "${MINGW_PACKAGE_PREFIX}-cc"
-             "lndir")
+             "lndir"
+             "patch")
 options=('strip' 'buildflags' 'staticlibs')
 source=(https://boostorg.jfrog.io/artifactory/main/release/${pkgver}/source/boost_${_boostver}.tar.bz2
         boost-1.63.0-python-test-PyImport_AppendInittab.patch
@@ -112,7 +113,6 @@ setb2args() {
     --debug-configuration \
     --prefix=${pkgdir}${MINGW_PREFIX} \
     --layout=tagged-1.66 \
-    --no-cmake-config \
     -sHAVE_ICU=1 \
     -sICU_PATH=${MINGW_PREFIX} \
     -sICU_ICUUC_NAME=icuuc \
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -25,8 +25,7 @@ optdepends=("${MINGW_PACKAGE_PREFIX}-python: For Boost.Python"
 makedepends=("${MINGW_PACKAGE_PREFIX}-python"
              "${MINGW_PACKAGE_PREFIX}-python-numpy"
              "${MINGW_PACKAGE_PREFIX}-cc"
-             "lndir"
-             "patch")
+             "lndir")
 options=('strip' 'buildflags' 'staticlibs')
 source=(https://boostorg.jfrog.io/artifactory/main/release/${pkgver}/source/boost_${_boostver}.tar.bz2
         boost-1.63.0-python-test-PyImport_AppendInittab.patch


slavaandrejev:
This PR addresses #14785.

lazka:
Thanks, I'll try to rebuild some things with this to make sure it works.

mingw-w64-boost/PKGBUILD
```diff
@@ -25,7 +25,8 @@ optdepends=("${MINGW_PACKAGE_PREFIX}-python: For Boost.Python"
 makedepends=("${MINGW_PACKAGE_PREFIX}-python"
              "${MINGW_PACKAGE_PREFIX}-python-numpy"
              "${MINGW_PACKAGE_PREFIX}-cc"
-             "lndir")
+             "lndir"
+             "patch")
```
MehdiChinoune, line 29: patch is part (dependency) of base-devel.

intractabilis: > patch is part (dependency) of base-devel.

I am not certain what you are suggesting. Can you explain?

Here was my reason. Without this line, if the patch package is not installed, the build fails with
```
/somewhere/MINGW-packages/mingw-w64-boost/PKGBUILD: line 54: patch: command not found
    Aborting...
```
instead of
```
==> Missing dependencies:
  -> patch
==> ERROR: Could not resolve all dependencies.
```
The latter is much more clear what a user should do.

slavaandrejev: > patch is part (dependency) of base-devel.

I am not certain what you are suggesting. Can you explain?

Here was my reason. Without this line, if the patch package is not installed, the build fails with
```
/somewhere/MINGW-packages/mingw-w64-boost/PKGBUILD: line 54: patch: command not found
    Aborting...
```
instead of
```
==> Missing dependencies:
  -> patch
==> ERROR: Could not resolve all dependencies.
```
The latter is much more clear what a user should do.

MehdiChinoune: `base-devel` is assumed to be installed when making packages. remove that `patch` from `makedepends()`

MehdiChinoune:
building precice fails after this change, @slavaandrejev could you take a look and fix the issue.

slavaandrejev:
@MehdiChinoune

Is an undefined reference to `__imp_boost::filesystem::detail::path_traits::convert(char const*, char const*, std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >&, std::codecvt<wchar_t, char, _Mbstatet> const*)` is what you see? The library provides `boost::filesystem::detail::path_traits::convert(char const*, char const*, std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >&, std::codecvt<wchar_t, char, _Mbstatet> const*)`, not `__imp_boost::...`. The error looks legit. My guess would be precice configured its make files in a way that the boost headers think the dynamic linking will be used, but the linking step is configured to use static boost libraries.

MehdiChinoune:
I don't think so: https://github.com/precice/precice/blob/v2.5.0/CMakeLists.txt#L277

slavaandrejev:
> I don't think so: https://github.com/precice/precice/blob/v2.5.0/CMakeLists.txt#L277

Why you don't think so? It's precisely what I said: `BOOST_ALL_DYN_LINK` is set, but the linking step is trying to link the static libraries.

MehdiChinoune:
> > I don't think so: https://github.com/precice/precice/blob/v2.5.0/CMakeLists.txt#L277
> 
> Why don't you think so? It's precisely what I said: `BOOST_ALL_DYN_LINK` is set, but the linking step is trying to link the static libraries.

Because It's using BoostConfig instead of CMake module, something wrong with BoostConfig

MehdiChinoune:
I think BoostConfig default to use static libraries on WIN32, should be fixed to use Shared ones
libboost_filesystem-variant-mt-shared.cmake:
```cmake

if(DEFINED Boost_USE_STATIC_LIBS)
  if(Boost_USE_STATIC_LIBS)
    _BOOST_SKIPPED("libboost_filesystem-mt.dll.a" "shared, Boost_USE_STATIC_LIBS=${Boost_USE_STATIC_LIBS}")
    return()
  endif()
else()
  if(WIN32 AND NOT _BOOST_SINGLE_VARIANT)
    _BOOST_SKIPPED("libboost_filesystem-mt.dll.a" "shared, default on Windows is static, set Boost_USE_STATIC_LIBS=OFF to override")
    return()
  endif()
endif()
```
That `WIN32` should be changed to `MINGW`.

slavaandrejev:
> Because It's using BoostConfig instead of CMake module, something wrong with BoostConfig

If two bugs annulled each other, it doesn't make them not bugs. CMake's configuration files provided by the authors is the most reliable way of including a package into your project. If you specified `BOOST_ALL_DYN_LINK`, you _must_ link to the dynamic libraries.

MehdiChinoune:
> If two bugs annulled each other, it doesn't make them not bugs. CMake's configuration files provided by the authors is the most reliable way of including a package into your project. If you specified `BOOST_ALL_DYN_LINK`, you _must_ link to the dynamic libraries.

Yes, Why boostConfig is choosing static libraries instead of shared ones.
You should fix that if you want to use BoostConfig instead of Cmake modules, as I said above your change make packages link to static libraries instead of shared ones.

MehdiChinoune:
I found it
https://github.com/boostorg/boost_install/blob/boost-1.81.0/boost-install.jam#L231
https://github.com/boostorg/boost_install/blob/boost-1.81.0/boost-install.jam#L249

slavaandrejev:
> your change make packages link to static libraries instead of shared ones

If this were true, they wouldn't have linked, like in the case of precice.

slavaandrejev:
> I found it
> https://github.com/boostorg/boost_install/blob/boost-1.81.0/boost-install.jam#L231
> https://github.com/boostorg/boost_install/blob/boost-1.81.0/boost-install.jam#L249

This just prints a diagnostic message if you set `Boost_VERBOSE`.

slavaandrejev:
> I think BoostConfig default to use static libraries on WIN32, should be fixed to use Shared ones

Why? Any reason besides it's your personal preference? It's a documented feature:
```
# Boost_USE_STATIC_LIBS:    When ON, uses static Boost libraries; when OFF,
#                           uses shared Boost libraries; when not set, uses
#                           static on Windows, shared otherwise. 
```

MehdiChinoune:
> > I found it
> > https://github.com/boostorg/boost_install/blob/boost-1.81.0/boost-install.jam#L231
> > https://github.com/boostorg/boost_install/blob/boost-1.81.0/boost-install.jam#L249
> 
> This just prints a diagnostic message if you set `Boost_VERBOSE`.

There is a `return()` after it, can't you see!

MehdiChinoune:
> > I think BoostConfig default to use static libraries on WIN32, should be fixed to use Shared ones
> 
> Why? Any reason besides it's your personal preference? It's a documented feature:
> 
> ```
> # Boost_USE_STATIC_LIBS:    When ON, uses static Boost libraries; when OFF,
> #                           uses shared Boost libraries; when not set, uses
> #                           static on Windows, shared otherwise. 
> ```

Linking to shared libraries is preferred over static libraries for many reasons:
- No need to rebuild reverse-dependencies to propagate a fix especially when It's about security.
- Less size.

slavaandrejev:
> Linking to shared libraries is preferred over static libraries for many reasons:

I know the pros and cons of using shared libraries. Changing a documented default behavior, however, asks for troubles. You need more arguments than general truisms and one library author, who wasn't aware of the documented behavior.

MehdiChinoune:
> > Linking to shared libraries is preferred over static libraries for many reasons:
> 
> I know the pros and cons of using shared libraries. Changing a documented default behavior, however, asks for troubles. You need more arguments than general truisms and one library author, who wasn't aware of the documented behavior.

We could change that documentation too.
So, well let's back to using the CMake module. This save many troubles!
Please stop pushing you personal preference on the project.
The packages were linking to shared boost for a long time. That was the default and no one complained about it on MINGW.
Boost chose to default to static on WIN32 because most users use it to deploy their apps independently of package managers which is very different on MSYS2 platform.

I will try to fix default boost behavior.

slavaandrejev:
> So, well let's back to using the CMake module.

It's a categorical **NO**. Relying on the author's provided CMake configuration files is the most reliable way of working with packages.

> This save many troubles!

So far, I see only a case of one library, whose author wasn't aware of the default documented behavior. Moreover, it's an easy fix for this library.

slavaandrejev:
> The packages were linking to shared boost for a long time.

Again, if they were trying to link to static now instead, they wouldn't have compiled as in the case of precice.

> Please stop pushing you personal preference on the project.

It's not my preference. It's a preference of the Boost project. We can speak with them and see their reasoning.

MehdiChinoune:
You need more time to understand how things work on MSYS2.

slavaandrejev:
> You need more time to understand how things work on MSYS2.

Oh, wow. Instead of constructive arguments, let's mansplain things.

MehdiChinoune:
You don't seem to listen at all, I have presented my arguments but you insist on following the upstream default even It's not something good for MSYS2!

slavaandrejev:
@pdimov

Hi Peter. Would you mind explaining why the Boost project has chosen to link Boost libraries statically on Windows? Do you think it's a reasonable default for MINGW? If not, how would you suggest changing this default to dynamic?

MehdiChinoune:
> @pdimov
> 
> Hi Peter. Would you mind explaining why the Boost project has chosen to link Boost libraries statically on Windows? Do you think it's a reasonable default for MINGW? If not, how would you suggest changing this default to dynamic linking?

Should be:
Do you think it's a reasonable default for MSYS2/MINGW platform which is very similar to Linux packaging system.

slavaandrejev:
> You don't seem to listen at all, I have presented my arguments but you insist on following the upstream default even It's not something good for MSYS2!

I am listening. And I calmly, without changing to ad hominem, explained to you where your argumentation is weak. Your choice of continuing the discussion is telling me to shut up because I don't understand how things work here. Well, I guess, I don't.

slavaandrejev:
> Should be: Do you think it's a reasonable default for MSYS2/MINGW platform which is very similar to Linux packaging system.

I would really appreciate, if you don't correct me, and let Peter express his opinion, without steering it your way.

MehdiChinoune:
> > Should be: Do you think it's a reasonable default for MSYS2/MINGW platform which is very similar to Linux packaging system.
> 
> I would really appreciate, if you don't correct me, and let Peter express his opinion, without steering it your way.

No, because saying MINGW is very different than MSYS2/MINGW.
Many people don't use MINGW through MSYS2, where they build dependencies statically and link to them for easy deployment. Which is very different than MSYS2.

slavaandrejev:
> No, because...

Ask your own questions. Don't correct mine according to what you think they should be. Deal?

pdimov:
> Hi Peter. Would you mind explaining why the Boost project has chosen to link Boost libraries statically on Windows?

The Boost default when you build with `b2` without specifying `--build-type=complete` or `link=shared` on Windows is (and has always been) to build static libraries.

https://github.com/boostorg/boost/blob/5d3f39d6270f107267f398130d97b8508166058f/boostcpp.jam#L322-L324

On non-Windows, the default is to build both static and shared.

https://github.com/boostorg/boost/blob/5d3f39d6270f107267f398130d97b8508166058f/boostcpp.jam#L313-L314

So BoostConfig defaults to that.

> Do you think it's a reasonable default for MINGW? If not, how would you suggest changing this default to dynamic linking?

We (upstream) have no opinion on what the default needs to be in your case. You could patch the config files (or boost-install.jam) to change `WIN32` to `MSVC` so that the static default only applies to non-GCC compilers.

I can in principle make this change in Boost proper, but then what `b2` builds by default under GCC/Windows won't work from CMake. That's probably not what we want.

slavaandrejev:
Thank you, Peter, for a quick and detailed answer! If this is the case, I think we may try to change the default here to dynamic linking. I agree that you shouldn't change the Boost proper, it should be done via a patch.

@MehdiChinoune Do you agree with the proposed by Peter method of changing `WIN32` to `MSVC`?

slavaandrejev:
It seems like MehdiChinoune proceeded with the changes without further discussion.

PR #14792, opened 2022-12-29
sox: do not link with libssp, fix shared build on clang

driver1998:
libtool complains about not finding a shared version of libssp. libssp.a is a stub now anyways, let's remove it.


lazka:
would `export lt_cv_deplibs_check_method='pass_all'` work too here? That's what we do in other packages afair

driver1998:
Let's try that, would certainly prefer less patching.

driver1998:
Works fine on clang64 locally, let's see how autobuild goes.

PR #14793, opened 2022-12-29
opencv: update to 4.7.0

MehdiChinoune:


MehdiChinoune:
building `gst-plugins-bad` fails:
```
Dependency opencv4 found: NO found 4.7.0 but need: '>= 4.0.0', '< 4.7.0' (cached)
```

Biswa96:
> building gst-plugins-bad fails

It may be better not to update opencv than to break existing packages. Even gst-plugins-bad git repository have that same opencv version requirement, see https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/opencv/meson.build

Also opencv APIs may differ in newer versions. Upstream may change opencv version requirement in future. For example, this happen in previous 4.6 version https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2557

MehdiChinoune:
> It may be better not to update opencv than to break existing packages. Even gst-plugins-bad git repository have that same opencv version requirement, see https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/opencv/meson.build

Who said I want to break existing packages, gst-plugins-bad mention the tested/supported versions that doesn't mean newer won't work.
 
> Also opencv APIs may differ in newer versions.

Glad you told me, I didn't know about that.

> Upstream may change opencv version requirement in future. For example, this happen in previous 4.6 version https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2557

They have just increased the version from `<4.6.0` to `<4.7.0`, and introduced a macro for opencv version, they did nothing to fix any issue.

Issue #14803, opened 2022-12-30
File conflict between angleproject with mesa and opencl-headers


mmuetzel:
### Description / Steps to reproduce the issue

Running `pacman -Syu` fails with the following file conflicts for me:
```
Fehler: Der Vorgang konnte nicht durchgeführt werden (In Konflikt stehende Dateien)
mingw-w64-x86_64-angleproject: /mingw64/bin/libEGL.dll existiert im Dateisystem (gehört zu mingw-w64-x86_64-mesa)
mingw-w64-x86_64-angleproject: /mingw64/bin/libGLESv1_CM.dll existiert im Dateisystem (gehört zu mingw-w64-x86_64-mesa)
mingw-w64-x86_64-angleproject: /mingw64/bin/libGLESv2.dll existiert im Dateisystem (gehört zu mingw-w64-x86_64-mesa)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_d3d10.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_d3d11.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_dx9_media_sharing.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_dx9_media_sharing_intel.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_egl.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_ext.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
mingw-w64-x86_64-angleproject: /mingw64/include/CL/cl_ext_intel.h existiert im Dateisystem (gehört zu mingw-w64-x86_64-opencl-headers)
[...]

PR #14804, opened 2022-12-30
angleproject: don't package OpenCL headers

mmuetzel:
This fixes part of #14803 by not packaging the OpenCL headers.

It's not clear at the moment how to resolve the conflict with the GLES and EGL libraries and headers from the mesa package. So, I just made the current conflict explicit.
It might be possible to remove that conflict again if it can be resolved somehow.

### DIFF
--- a/mingw-w64-angleproject/PKGBUILD
+++ b/mingw-w64-angleproject/PKGBUILD
@@ -6,7 +6,7 @@ _realname=angleproject
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.1.r20194.6c41793f
-pkgrel=1
+pkgrel=2
 pkgdesc='A conformant OpenGL ES implementation for Windows, Mac, Linux, iOS and Android (mingw-w64)'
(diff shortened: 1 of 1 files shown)


mingw-w64-angleproject/PKGBUILD
```diff
@@ -128,6 +129,7 @@ build() {
     angle_enable_abseil=false
     angle_enable_metal=false
     angle_enable_d3d11_compositor_native_window=false
+    angle_enable_cl=false
```
Biswa96, line 132: Is this option required to resolve the conflict? Would not this disable the opencl features in angle?

mmuetzel: If I understand correctly, `angle_enable_cl` is false by default. I just figured, it would make sense to disable it explicitly if we stop distributing the corresponding headers.

Biswa96: Headers are not installed by gn or ninja. If it is 'false' by default this change may not have any effect.

mmuetzel:
I removed the commit that added the conflict with mesa (assuming that #14848 will also be merged).

PR #14805, opened 2022-12-30
octave: force downgrade to SuiteSparse 5 for 32-bit

mmuetzel:
Revert back to forcing a downgrade to SuiteSparse 5. Octave uses parts of the API that are broken on 32-bit SuiteSparse 6.

### DIFF
--- a/mingw-w64-octave/PKGBUILD
+++ b/mingw-w64-octave/PKGBUILD
@@ -22,7 +22,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-libsndfile"
          "${MINGW_PACKAGE_PREFIX}-pcre"
          "${MINGW_PACKAGE_PREFIX}-qhull"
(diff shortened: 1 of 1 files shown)


mingw-w64-octave/PKGBUILD
```diff
@@ -22,7 +22,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-libsndfile"
          "${MINGW_PACKAGE_PREFIX}-pcre"
          "${MINGW_PACKAGE_PREFIX}-qhull"
-         "${MINGW_PACKAGE_PREFIX}-suitesparse"
+         $([[ ${MSYSTEM} == *32 ]] && echo "${MINGW_PACKAGE_PREFIX}-suitesparse5" || echo "${MINGW_PACKAGE_PREFIX}-suitesparse")
```
MehdiChinoune, line 25:
```suggestion
         $([[ ${CARCH} == i686 ]] && echo "${MINGW_PACKAGE_PREFIX}-suitesparse5" || echo "${MINGW_PACKAGE_PREFIX}-suitesparse")
```

mmuetzel: Thanks. 👍
Committed as you suggested.

PR #14841, opened 2022-12-30
Allow SFML PKGBUILD to build on clangarm64

ZachBacon:
Signed-off-by: Zach Bacon <zachbacon@vba-m.com>


Biswa96:
Enabling clangarm64 does not require pkgrel bump. Have you built this package in clangarm64?

ZachBacon:
I have indeed, as a packager I tend to build before pushing changes. And yeah the habit of a bump in rel is a habit I have with solus. General rule of thumb there was if you touch it, then bump it.

Biswa96:
Bumping pkgrel would not create any issue here but it results in additional update for users. pkgrel should be incremented only when fixes or addition features are added in PKGBUILD[^1]. Here adding clangarm64 does not add or change new features in packages of other MSYSTEMs.

[^1]: https://wiki.archlinux.org/title/PKGBUILD

ZachBacon:
Yes I'm aware, I'm just saying it was a force of habit that made me do it without realizing it till much after the fact.

Biswa96:
So, would you like to revert the last commit please?

ZachBacon:
That's the thing, I am not well versed in reverting commits.

Biswa96:
It's OK. I have added your commit in https://github.com/msys2/MINGW-packages/commit/131b9762af3fc6d4dd9c20df16795d622dca3be1. Thank you.

git commit can be reverted with `git reset --hard HEAD~1` command and then force push.

PR #14844, opened 2022-12-31
minimize build time by passing `--default-library=both` to meson

MehdiChinoune:


lazka:
In theory this results in broken static libraries.

MehdiChinoune:
There are many libraries built with that option, Should we split their building to static+shared.

PR #14848, opened 2022-12-31
 mesa: Update to 22.3.2

pal1000:
- Disable GLES and EGL as MSYS2 provides angleproject (#14803), a far superior implementation;
- Disable shared glapi as it's not needed without GLES and EGL;
- Stop building unit tests since they aren't packaged.


Biswa96:
> Disable GLES and EGL 

Would this cause any issue if user depends on those libraries/features? Splitting packages is also another option to mitigate the file conflict issue.

pal1000:
> Would this cause any issue if user depends on those libraries/features?

Everything that depends on GLES and EGL will have to use angleproject.

PR #14849, opened 2022-12-31
kicad: update to 6.0.10

mmuetzel:
This is based upon the changes from #11873 by @stahta01.

I did some exploratory testing with a locally built version in MINGW64. And it seems to be working for me.


mmuetzel:
I installed the MINGW64 build artifacts from this PR and it seems to be working fine.

PR #14863, opened 2023-01-01
python-ipywidgets: update to 8.0.4

MehdiChinoune:
Could someone with a good Python knowledge take a look and see what's going on!


raedrizqie:
https://github.com/jupyter/jupyter-packaging/blob/192ea82f27d7b3ce35089bd2ef535dab5a681435/jupyter_packaging/setupbase.py#L976

This one may be the culprit?
`return "(?ms){res}\\Z".format(res=res)`

PR #14881, opened 2023-01-01
gspell: add --enable-installed-tests

swilmet:
The gspell unit tests need to be installed in order to run successfully.

Otherwise they don't see the hunspell dictionaries (only the en_US one is required for the unit tests).


lazka:
huh, the latest file update in staging fails to detect .tar.xz

PR #14889, opened 2023-01-02
shotcut: Fix timeline view with adding qt5-quickcontrols

Biswa96:


Biswa96:
All credit goes to @driver1998 who found the issue and suggested the fix 👏

PR #14897, opened 2023-01-04
mkvtoolnix: update to 73.0.0

MehdiChinoune:
I have disabled 32bit envs because I am going to drop 32bit from most of Qt6 packages.


Biswa96:
> I am going to drop 32bit from most of Qt6 packages.

A proper explanation and a news entry may be helpful to notify users.

PR #14945, opened 2023-01-05
opencascade: update to 7.7.0

mmuetzel:
This version is not API compatible to previous versions and reverse dependencies will probably need to be rebuilt.
Let's wait for the package grokker.


mmuetzel:
The package grokker determined gmsh, OpenSceneGraph, and kicad that need to be rebuild.
All of these packages are rather large projects. And the CI will likely time out.

But I tried building those packages with the MINGW64 build artifact from this PR. And it worked locally for me.

mmuetzel:
At least one environment (MINGW32) finished to successfully compile all packages before the timeout was reached. (But OpenSceneGraph was missing a description and I added one now.)

Probably good to go now imho. But maybe we should wait until the Qt updates are done.

PR #14954, opened 2023-01-06
gmp: Do not declare functions as `dllimport`

lhmouse:
GNU toolchains do not require `dllimport` when calling functions from DLLs. Having `dllimport` by default causes static linking to fail, observed when building GDB.

(cherry picked from commit 91d231a51056c2c429e55ae1546794b93089bb64)

### DIFF
--- a/mingw-w64-gmp/PKGBUILD
+++ b/mingw-w64-gmp/PKGBUILD
@@ -4,16 +4,18 @@ _realname=gmp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=6.2.1
-pkgrel=4
+pkgrel=5
 pkgdesc="A free library for arbitrary precision arithmetic (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://gmplib.org/"
 license=('LGPL3' 'GPL')
 makedepends=("${MINGW_PACKAGE_PREFIX}-autotools" "${MINGW_PACKAGE_PREFIX}-cc")
-source=(https://gmplib.org/download/gmp-${pkgver}/${_realname}-${pkgver}.tar.xz{,.sig})
+source=(https://gmplib.org/download/gmp-${pkgver}/${_realname}-${pkgver}.tar.xz{,.sig}
+        do-not-use-dllimport.diff)
 sha256sums=('fd4829912cddd12f84181c3451cc752be224643e87fac497b69edddadc49b4f2'
-            'SKIP')
+            'SKIP'
+            '385ab704f82c47f3aecc9141f43c96e7b8de2bf0e654dc457ce0f1a039db2c68')
 validpgpkeys=('343C2FF0FBEE5EC2EDBEF399F3599FF828C67298') # Niels Möller <nisse@lysator.liu.se>"
 
 prepare() {
@@ -21,6 +23,7 @@ prepare() {
   [[ -d ../stash ]] && rm -rf ../stash
   mkdir ../stash
   cp config.{guess,sub} ../stash
+  patch -p2 -i ${srcdir}/do-not-use-dllimport.diff
   autoreconf -fiv
   cp -f ../stash/config.{guess,sub} .
 }
--- a/mingw-w64-gmp/do-not-use-dllimport.diff
+++ b/mingw-w64-gmp/do-not-use-dllimport.diff
@@ -1,0 +1,13 @@
+diff --git a/gmp-6.2.1.orig/gmp-h.in b/gmp-6.2.1/gmp-h.in
+index 3d449d427..942fab4b1 100644
+--- a/gmp-6.2.1.orig/gmp-h.in
++++ b/gmp-6.2.1/gmp-h.in
+@@ -101,7 +101,7 @@ see https://www.gnu.org/licenses/.  */
+ 
+ #if defined (__GNUC__)
+ #define __GMP_DECLSPEC_EXPORT  __declspec(__dllexport__)
+-#define __GMP_DECLSPEC_IMPORT  __declspec(__dllimport__)
++#define __GMP_DECLSPEC_IMPORT
+ #endif
+ #if defined (_MSC_VER) || defined (__BORLANDC__)
+ #define __GMP_DECLSPEC_EXPORT  __declspec(dllexport)


Biswa96:
> cherry picked from commit 91d231a51056c2c429e55ae1546794b93089bb64

From which repository?

lhmouse:
> > cherry picked from commit 91d231a51056c2c429e55ae1546794b93089bb64
> 
> From which repository?

Oh I did that from my local branch. I can remove that line.

lhmouse:
The commit message has been updated. GDB built without issues so far.

jeremyd2019:
> I still think installing the 'static' variant of the headers is more correct for these, based on how they work.

From https://github.com/msys2/MINGW-packages/pull/14567#issuecomment-1356855852

lhmouse:
> > I still think installing the 'static' variant of the headers is more correct for these, based on how they work.
> 
> From [#14567 (comment)](https://github.com/msys2/MINGW-packages/pull/14567#issuecomment-1356855852)

That's not what happens in practice. People will get surprised if linking against the static library requires a different header, and in addition, GDB configure.ac needs to patched for this new header as well.

jeremyd2019:
I meant the previous state of gmp: install both the shared and static libraries, but only the static headers.  This seemed to work pretty well, but some gmp-related configure scripts (mpc, mpfr) refused to build shared libraries if gmp headers indicated it was built static.  I don't remember if the final PR that got merged patched this out and got both static and shared libraries of mpc and mpfr, or if it switched them to shared only.  There were several iterations

lhmouse:
> I meant the previous state of gmp: install both the shared and static libraries, but only the static headers. This seemed to work pretty well, but some gmp-related configure scripts (mpc, mpfr) refused to build shared libraries if gmp headers indicated it was built static. I don't remember if the final PR that got merged patched this out and got both static and shared libraries of mpc and mpfr, or if it switched them to shared only. There were several iterations

How do static and shared headers differ? AFAICT the only difference is `dllimport`, which can be removed from functions (if thankfully there is no dllimport'd variable) so there will be no difference at all.

Biswa96:
> How do static and shared headers differ? 

Here is the diff

```diff
--- static-UCRT64/install/ucrt64/include/gmp.h
+++ shared-UCRT64/install/ucrt64/include/gmp.h
@@ -58,7 +58,7 @@
 /* Instantiated by configure. */
 #if ! defined (__GMP_WITHIN_CONFIGURE)
 #define _LONG_LONG_LIMB 1
-#define __GMP_LIBGMP_DLL  0
+#define __GMP_LIBGMP_DLL  1
 #endif
```

lhmouse:
> > How do static and shared headers differ?
> 
> Here is the diff
> 
> ```diff
> --- static-UCRT64/install/ucrt64/include/gmp.h
> +++ shared-UCRT64/install/ucrt64/include/gmp.h
> @@ -58,7 +58,7 @@
>  /* Instantiated by configure. */
>  #if ! defined (__GMP_WITHIN_CONFIGURE)
>  #define _LONG_LONG_LIMB 1
> -#define __GMP_LIBGMP_DLL  0
> +#define __GMP_LIBGMP_DLL  1
>  #endif
> ```


I suspect it only affects `__GMP_DECLSPEC` and `__GMP_DECLSPEC_XX`, by defining them to `__GMP_DECLSPEC_IMPORT` for shared, and empty for static?

Then this PR makes some sense, because `__GMP_DECLSPEC_IMPORT` can be empty. Thus, in either case, when not inside GMP itself, `__GMP_DECLSPEC` and `__GMP_DECLSPEC_XX` are always empty, and there is no difference between headers for shared and static libraries.

lazka:
Should I merge?

jeremyd2019:
I believe there are dllimported variables.  examples:
```c
#define mp_bits_per_limb __gmp_bits_per_limb
__GMP_DECLSPEC extern const int mp_bits_per_limb;

#define gmp_errno __gmp_errno
__GMP_DECLSPEC extern int gmp_errno;

#define gmp_version __gmp_version
__GMP_DECLSPEC extern const char * const gmp_version;
```

My understanding is that either refptr magic (just in lld?) or pseudo-relocations would take care of those.

> I suspect it only affects `__GMP_DECLSPEC` and `__GMP_DECLSPEC_XX`, by defining them to `__GMP_DECLSPEC_IMPORT` for shared, and empty for static?

I think so.

Biswa96:
Would separating static and shared libraries & headers in their own packages be a proper solution?

raedrizqie:
Would it ork if we change dllexport to `__attribute__((visibillity("default")))` and dllimport to nothing?

jeremyd2019:
> Would it ork if we change dllexport to `__attribute__((visibillity("default")))` and dllimport to nothing?

No.

There was a patch to clang that allowed hidden visibility to exclude symbols from the linker auto export feature, but this is not supported in gcc, and would not have any effect on the import side of things, which is where the issues are here.


> Would separating static and shared libraries & headers in their own packages be a proper solution?

That's the solution that's kind of implied by how upstream gmp is set up, but we seem to be trying really hard not to do that.

raedrizqie:
Well, better break other packages than fixing gmp, i guess :)

lhmouse:
> > Would separating static and shared libraries & headers in their own packages be a proper solution?
> 
> That's the solution that's kind of implied by how upstream gmp is set up, but we seem to be trying really hard not to do that.

Do GMP developer actually recommend this setup? It looks very unidiomatic, especially for people who are familiar with libraries on Linux.

bitthief:
I was having issues ever since this change compiling static TigerVNC (GnuTLS has a dependency on gmp), this fixed it!

jeremyd2019:
> > > Would separating static and shared libraries & headers in their own packages be a proper solution?
> > 
> > 
> > That's the solution that's kind of implied by how upstream gmp is set up, but we seem to be trying really hard not to do that.
> 
> Do GMP developer actually recommend this setup? It looks very unidiomatic, especially for people who are familiar with libraries on Linux.

Well, MPC and MPFR won't build a shared library against a static GMP (__GMP_LIBGMP_DLL 0), or build a static library against a shared GMP (__GMP_LIBGMP_DLL 1), so the only way to build both static and shared MPC/MPFR would be to split them somehow, or patch out that check and do it anyway (with whatever side-effects that may cause).  See also #14146

lhmouse:
> Well, MPC and MPFR won't build a shared library against a static GMP (__GMP_LIBGMP_DLL 0), or build a static library against a shared GMP (__GMP_LIBGMP_DLL 1), so the only way to build both static and shared MPC & MPFR would be to split static/shared gmp somehow, or patch out that check and do it anyway (with whatever side-effects that may cause). See also #14146 (Note that the issues with an exported thread local variable now exist in the repo regardless, due to mpfr now being shared only, IIRC)

I am curious how this is done on Linux. I suspect it does not apply to Linux, because DLL is a Windows thing?

As long as it's only functions that are imported, there should be no issue (import libraries i.e. `.dll.a` files provide thunks as conventional functions which can be used as such). It _will_ be an issue if, for example, if static MPC references a variable from shared GMP. We have workarounds for this in mingw-w64: It is possible to provide `__MINGW_IMP_SYMBOL`s in static libraries, so if someone else links against this static library, but expects a `dllimport`'d symbol, they can find one. In this case, sources of GMP should be patched for dummy `__imp_` symbols, and a 'shared' header will suffice.

mmuetzel:
I tried to rebuild gdb. It fails with the following in the config.log:
```
configure:10473: checking for libgmp
configure:10493: gcc -o conftest.exe -march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -I/mingw64/include/ncurses     -D__USE_MINGW_ANSI_STDIO=1 -I/mingw64/include/ncurses -I/mingw64/include  -pipe -Wl,--stack,12582912 conftest.c -lncursesw -lm -ldl  /mingw64/lib/libgmp.a >&5
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\msys64\tmp\ccx3YvEM.o:conftest.c:(.text.startup+0x22): undefined reference to `__imp___gmpz_init'
configure:10493: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME ""
| #define PACKAGE_TARNAME ""
| #define PACKAGE_VERSION ""
| #define PACKAGE_STRING ""
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define __EXTENSIONS__ 1
| #define _ALL_SOURCE 1
| #define _GNU_SOURCE 1
| #define _POSIX_PTHREAD_SEMANTICS 1
| #define _TANDEM_SOURCE 1
| #define HAVE_DLFCN_H 1
| #define HAVE_WINDOWS_H 1
| #define _FILE_OFFSET_BITS 64
| #define HAVE_CXX11 1
| #define ENABLE_NLS 1
| #define PACKAGE "gdb"
| #define DEBUGDIR "/mingw64/lib/debug"
| #define DEBUGDIR_RELOCATABLE 1
| #define BINDIR "/mingw64/bin"
| #define GDB_DATADIR "/mingw64/share/gdb"
| #define GDB_DATADIR_RELOCATABLE 1
| #define AUTO_LOAD_DIR "$debugdir:$datadir/auto-load"
| #define AUTO_LOAD_SAFE_PATH "$debugdir:$datadir/auto-load"
| #define DEFAULT_BFD_ARCH bfd_i386_arch
| #define DEFAULT_BFD_VEC x86_64_pe_vec
| #define HAVE_ETEXT 1
| #define PKGVERSION "(GDB) "
| #define REPORT_BUGS_TO "<https://www.gnu.org/software/gdb/bugs/>"
| #define HAVE_LIBM 1
| #define HAVE_ICONV 1
| #define ICONV_CONST 
| #define HAVE_READLINE_READLINE_H 1
| #define SIZEOF_UNSIGNED_LONG_LONG 8
| #define SIZEOF_UNSIGNED_LONG 4
| #define SIZEOF_UNSIGNED___INT128 16
| #define JIT_READER_DIR "/mingw64/lib/gdb"
| #define JIT_READER_DIR_RELOCATABLE 1
| #define HAVE_LIBEXPAT 1
| #define HAVE_XML_STOPPARSER 1
| /* end confdefs.h.  */
| #include <gmp.h>
| int
| main ()
| {
| mpz_t n;
|                        mpz_init (n);
|   ;
|   return 0;
| }
configure:10503: result: no
configure:10527: error: GMP is missing or unusable
```

Is this the same issue?

jeremyd2019:
What's the status of this?  Is this sufficient, or given that there *are* variables exported in addition to functions will there need to be more invasive patching?

jeremyd2019:
Interesting bit from mpfr.h:
```c
/* Support for WINDOWS Dll:
   Check if we are inside a MPFR build, and if so export the functions.
   Otherwise does the same thing as GMP */
#if defined(__MPFR_WITHIN_MPFR) && __GMP_LIBGMP_DLL
# define __MPFR_DECLSPEC __GMP_DECLSPEC_EXPORT
#else
# ifndef __GMP_DECLSPEC
#  define __GMP_DECLSPEC
# endif
# define __MPFR_DECLSPEC __GMP_DECLSPEC
#endif
```

jeremyd2019:
It looks like mpfr has refactored in such a way that they no longer attempt to export thread local variables, at least in their git version.  I don't know what if any released version that may be in.

lhmouse:
Why do MPFR and MPC check for `__GMP_LIBGMP_DLL` instead of their own `DLL_EXPORT`?

raedrizqie:
> Why do MPFR and MPC check for `__GMP_LIBGMP_DLL` instead of their own `DLL_EXPORT`?

because of this in gmp..
```
 #if ! defined (__GMP_WITHIN_CONFIGURE)
 #define _LONG_LONG_LIMB 1
-#define __GMP_LIBGMP_DLL  0
+#define __GMP_LIBGMP_DLL  1
 #endif
```

like you said, it should just use `DLL_EXPORT` instead.. we could have both shared and static libs packaged together

jeremyd2019:
> Why do MPFR and MPC check for `__GMP_LIBGMP_DLL` instead of their own `DLL_EXPORT`?

Because they're silly and are making the assumption that they'll always be configured the same as gmp, hence my earier observation that it seems upstream expects static and shared to be distributed separately on Windows.

lhmouse:
This looks created by someone with inferior knowledge about Windows DLLs, and there is no general guideline about how to do so. Maybe we should file a bug report to them, to request for a change.

Basically the guideline on how to create a Windows DLL with GNU toolchains is as follows:

1. For library _MYLIB_, functions from _MYLIB_ shall be specified `dllexport` only when building the DLL of _MYLIB_.
    * The first approach is to define a macro in `CPPFLAGS` of _MYLIB_, so after a header is included, it is possible to tell whether it is used inside _MYLIB_ or outside of it:
        ```c
        /* mylib.h */
        #if defined BUILDING_MYLIB && defined DLL_EXPORT
        #  define MYLIB_EXPORT  __attribute__((__dllexport__))  /* when building the DLL  */
        #else
        #  define MYLIB_EXPORT  /* when building the archive, or included by user code  */
        #endif
        MYLIB_EXPORT int mylib_some_function(void);
        ```
    * The second is to define `MYLIB_EXPORT` in a private header that will not be installed anyway (I personally recommend this):
        ```c
        /* mylib.h  */
        #ifndef MYLIB_EXPORT
        #  define MYLIB_EXPORT     /* when included by user code  */
        #endif
        ```
        and in a private header that is include by every source file:
        ```c
        #ifdef DLL_EXPORT
        #  define MYLIB_EXPORT  __attribute__((__dllexport__))  /* when building the DLL  */
        #else
        #  define MYLIB_EXPORT  /* when building the archive  */
        #endif
        #include "mylib.h"
        ```
2. It is not recommended to export variables from DLLs.
    * This means that instead of
        ```c
        /* mylib.h  */ 
        extern int mylib_some_variable;
        ```
        it is recommended to do
        ```c
        /* mylib.h  */ 
        extern __attribute__((__const__)) int* mylib_some_variable_address(void);
        #define mylib_some_variable  (* mylib_some_variable_address())
        ```
    * If a variable is still preferred to a function because it has less overhead, this hack may be helpful:
        ```c
        /* mylib.h */
        #ifndef MYLIB_EXPORT
        #  define MYLIB_EXPORT     /* when included by user code  */
        #endif
        #if defined BUILDING_MYLIB && defined DLL_EXPORT
        #  define MYLIB_EXPORT_VAR  __attribute__((__dllexport__))  /* when building the DLL  */
        #else
        #  define MYLIB_EXPORT_VAR  __attribute__((__dllimport__))  /* when building the archive, or included by user code  */
        #endif
        extern MYLIB_EXPORT_VAR int mylib_some_variable;
        ```
        and define a dummy import symbol for it:
        ```c
        int mylib_some_variable = 0;
        int __MINGW_IMP_SYMBOL(mylib_some_variable) =
            &mylib_some_variable;  /* for linking statically, and by the DLL itself  */
        ```

MehdiChinoune:
https://gmplib.org/manual/Notes-for-Particular-Systems
https://gmplib.org/list-archives/gmp-discuss/2004-May/thread.html#1155

lhmouse:
> https://gmplib.org/manual/Notes-for-Particular-Systems

On that page: 'Static and DLL libraries can’t both be built, since certain export directives in gmp.h must be different.'

^ This is exactly what the macro `DLL_EXPORT`, defined by libtool, is used for.

> https://gmplib.org/list-archives/gmp-discuss/2004-May/thread.html#1155

This is a very old thread and I think some hindrance might no longer exist.

mcendu:
What is the current status?

lhmouse:
This PR only fixes GDB. From comments by jeremyd2019, there seems to have to be more work against GMP and MPFR.

jeremyd2019:
I think this is an improvement to the status quo, but there may need to be more work (especially if we want to avoid pseudo-relocations)

mmuetzel:
I downloaded and and installed the MINGW64 build artifacts from this PR. I can confirm that that fixes building `gdb`. IIUC, some `dllimport`ed variables are still not handled correctly with the changes from this PR.

Would it be ok to merge the changes from this PR and keep an issue open to track the changes needed for the `dllimport`ed variables?

mmuetzel:
There haven't been any negative reactions in the last couple of days. And the previous comments have been positive about this being a partial fix.
Merging this PR.

PR #14974, opened 2023-01-07
wwrando: update to 1.10.0

MehdiChinoune:


Biswa96:
The updated package now contains git related files.

MehdiChinoune:
> The updated package now contains git related files.

and vscode related files
What should I do about them, I left them because I considered them harmless

PR #14306, opened 2022-11-28
enable clangarm64 for more python packages

kmilos:

PR #14993, opened 2023-01-07
pywin32 and Jupyter: Enable clangarm64

driver1998:
Patches for pywin32 is based on the one from @Biswa96.


Biswa96:
> Patches for pywin32 is based on...

Did you mean this patch https://github.com/msys2/MINGW-packages/pull/14306#issuecomment-1354175822 ?

The clang64 build fails here https://github.com/msys2/MINGW-packages/actions/runs/3863716806/jobs/6588993779#step:11:4313. long -> uintptr_t may help, not sure though.

PR #15000, opened 2023-01-08
libheif: update to 1.14.2

kmilos:
Removed the pc file workaround, it's been dealt with upstream already.


Biswa96:
> it's been dealt with upstream already.

Can you share the commit here? Just for the record.

kmilos:
Sure, here goes: https://github.com/strukturag/libheif/pull/719

Biswa96:
Thank you for upstreaming it.

PR #15003, opened 2023-01-08
python-pywin32: Remove unused hunks from patches

Biswa96:
* The added symbols are upstreamed in mingw-w64.
* IAsyncOperation was renamed to IDataObjectAsyncCapability in pywin32 upstream.


Biswa96:
@raedrizqie Any comment about this change?

raedrizqie:
Nice work.. all ok to me

PR #15004, opened 2023-01-09
angle: Enable clangarm64

Biswa96:
* Remove patch for adding D3D11_CLIP_OR_CULL_DISTANCE_COUNT symbol.
* Define ANGLE_PLATFORM_WINDOWS instead of _WIN32.
* Add arm64 target for mingw toolchain gn file.
* Use -msse3 flag for x86 and x86_64 only.

### DIFF
--- a/mingw-w64-angleproject/001-add-mingw-toolchain.patch
+++ b/mingw-w64-angleproject/001-add-mingw-toolchain.patch
@@ -800,12 +800,14 @@ index a5075393b..bc8f4e2c6 100644
    if (current_cpu == "x86") {
      asmflags = [
        # When /safeseh is specified, the linker will only produce an image if it
-@@ -214,6 +215,15 @@ config("compiler") {
+@@ -214,6 +215,17 @@ config("compiler") {
      # the source file is a no-op.
      "/ignore:4221",
    ]
 +  } else {
-+    cflags = [ "-msse3" ]
++    if (current_cpu == "x86" || current_cpu == "x64") {
++      cflags = [ "-msse3" ]
++    }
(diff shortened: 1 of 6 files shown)


Biswa96:
@jeremyd2019 Would you like to check if angle and libepoxy build with this change? I have only cross compiled it.

jeremyd2019:
```
==> Starting build()...
[7888:0109/195639.563:FATAL:args.cc(379)] Check failed: false. OS architecture not handled. ()
    Aborting...
```

Biswa96:
I have added the arm64 platform in gn. But the ninja command goes in an endless loop. After Ctrl+C, it shows some python message which I am not familiar with.

jeremyd2019:
> I have added the arm64 platform in gn. But the ninja command goes in an endless loop. After Ctrl+C, it shows some python message which I am not familiar with.

Seemed to work for me: https://github.com/msys2-arm/MINGW-packages/actions/runs/3889880322/jobs/6638775933  :shrug:

Biswa96:
Thank you for checking that. Let me rebase the changes.

jeremyd2019:
Note I also removed the special case for clangarm64 not depending on angleproject in libepoxy on that branch, I don't know if you want to do that in this PR too...

Biswa96:
> Note I also removed the special case for clangarm64 not depending on angleproject in libepoxy...

Yeah, I also wanted to do that. Hence I asked previously "if angle and libepoxy build with this change".

Biswa96:
Though it is expected, the `GDK_GL=gles gtk3-demo --run=glarea` command shows no GL implementation in Windows ARM in Rpi4.

Issue #8105, opened 2021-03-10
Multiple definitions of symbol PC when statically linking readline and termcap into a program


intvsteve:
This used to work when using the package `mingw-w64-i686-readline-7.0.005-1-any.pkg.tar.xz`.

Now, when upgrading to the latest (`mingw-w64-i686-readline-8.0.004-2-any.pkg.tar.zst`), linking both of these libraries results in an error from `ld`:

```
D:/msys64/mingw32/lib\libtermcap.a(termcap.o):(.bss+0x8): multiple definition of `PC'; D:/msys64/mingw32/lib\libreadline.a(terminal.o):(.bss+0x5c): first defined here
collect2.exe: error: ld returned 1 exit status
```

Biswa96:
Can you provide a simple code to reproduce your issue?

intvsteve:
The code I'm building does not appear to link to curses.

The code I'm building is the jzintv emulator. I have access to a slightly more recent version, but this aspect has not changed and the code [here](http://spatula-city.org/~im14u2c/intv/dl/jzintv-20200712-src.zip) will likely reproduce the issue as well.

Some caveats
- use the 32-bit msys environment as only the working build for Windows is currently the 32-bit version
- you will need SDL1 and / or SDL2 on the system, depending on which build you invoke (issue is same for both)
- example build: `make -f Makefile.w32_sdl2` in the `src` directory

I think most of what is needed for the 32-bit build can be installed via:
>pacman -S base-devel
pacman -S gcc
pacman -S mingw-w64-i686-gcc
pacman -S mingw-w64-i686-SDL
pacman -S mingw-w64-i686-SDL2
pacman -S mingw-w64-i686-readline

I had to reconstruct my msys2 when `pacman` stopped working completely. However, I could downgrade to the 32-bit version of the `mingw-w64-i686-readline` package, as mentioned, and get things to link. I initially filed this issue against the readline package.

intvsteve:
@Biswa96 Unfortunately, the sample code I provided is not "simple". I have not created a "toy" program to reproduce.

revelator:
hmm could have happened if say one of the libraries you link to include say ncurses as we only have a static version of that it might have exported the PC symbol to said library but it is hard to say.

can you try to build it with CFLAGS+=" -fcommon" CXXFLAGS+=" -fcommon" as i noticed this causes some packages that have previously linked without this error to suddenly throw a lot of multiple defines

intvsteve:
[...]

PR #15012, opened 2023-01-10
readline: define NEED_EXTERN_PC for static linking

1480c1:
tells readline to use the PC variable from libtermcap instead of defining its own. This fixes the multiple definition error when attempting to statically link readline

Fixes https://github.com/msys2/MINGW-packages/issues/8105

I have not tested shared linking, but have tested with compiling mujs' master branch

### DIFF
--- a/mingw-w64-readline/PKGBUILD
+++ b/mingw-w64-readline/PKGBUILD
@@ -6,7 +6,7 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 _basever=8.2
 _patchlevel=001
 pkgver=${_basever}.${_patchlevel}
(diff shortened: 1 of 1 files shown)


raedrizqie:
Finally a fix for this.. Nice!
I think *.pc files also need to define this?

mmuetzel:
> I think *.pc files also need to define this?

I don't think so.

mmuetzel:
Looks good to me.
I tried to rebuild gdb with the build artifact from this PR (as an example that links dynamically to readline). But that is failing.
I don't think the build error is related to the change here though. More likely, it is the same issue that is discussed in #14954.

PR #15018, opened 2023-01-10
Adding GW build

kcleal:


kcleal:
Hi @Biswa96,

Sorry for the repeated edits. I managed to build it on my system with the following recipe:

     pacman -Syu
    pacman -Sy base-devel
    pacman -Sy openssh
    pacman -Sy unzip
    pacman -Sy gcc
    pacman -Sy make
    pacman -Sy git
    
    pacman -S mingw-w64-ucrt-x86_64-{gcc,make,freeglut,glew,libpng,ncurses,wget}
    pacman -S mingw-w64-ucrt-x86_64-{htslib,glfw,skia,fontconfig,dlfcn}
    
    git clone https://github.com/kcleal/gw.git && cd gw
    LDFLAGS+="-L/ucrt64/lib -L/ucrt64/lib/skia" CPPFLAGS+="-I/ucrt64/include -I/ucrt64/include/skia -I/ucrt64/include/ncurses" make

Just trying to adapt this. Would really appreciate some help if you have any time!
Thanks

Biswa96:
There are many issues I found. Here are some of them.

1. Remove msys2 specific paths in Makefile, like this `LDFLAGS += -L/ucrt64/lib -L/home/nopel/gw_build/ucrt64/lib`
2. Add ncurses include directory in Makefile, like this `CPPFLAGS += $(shell pkgconf -cflags skia) $(shell ncursesw6-config --cflags)` . It's CFLAGS, by the way.
3. How to test? I got segfault in ucrt64

```
$ ./gw.exe
Select reference genome number from list F:\test\gw/.gw.ini:
:h

Segmentation fault
```

4. From the above message, it seems that the full path is not correct. It probably needs `std::filesystem::operator/` like thing.
5. Need some info about the expected full path of `.gw.ini` file.

kcleal:
Thanks for looking at this. 
I fixed 1 and 2.
`gw.exe --help` should show the options. There is some test data in `test` - this will open a new window: `gw.exe test/test.fa -b test/test.bam -r chr1:10000-11000`, or this will generate a .png image  `gw.exe test/test.fa -b test/test.bam -r chr1:10000-11000 --outdir . --no-show`
The .gw.ini file should live in the home directory or `home/.config`  - the home directory is defined for windows like:

      # if defined(_WIN32)
          const char *homedrive_c = std::getenv("HOMEDRIVE");
	      const char *homepath_c = std::getenv("HOMEPATH");
          std::string homedrive(homedrive_c ? homedrive_c : "");
	      std::string homepath(homepath_c ? homepath_c : "");
	      std::string home = homedrive + homepath;

Not sure if this is going to work under msys2? 

Do you need the updates to be put in a .tar.gz release to test again? Thanks for the help again

kcleal:
Thanks for going over this.
I have addressed points 1, 2 and 4 with a commit. Shall I create a new release tag with the new commit?

The .gw.ini is searched for in `/home/.gw.ini`, `/home/.config/.gw.ini` or in the same location as gw.exe (in that order). Im not sure if I am defining the home directory in a msys2 compatible way - do you have any advice here? Currently home is defined like this:

    # if defined(_WIN32)
        const char *homedrive_c = std::getenv("HOMEDRIVE");
	    const char *homepath_c = std::getenv("HOMEPATH");
        std::string homedrive(homedrive_c ? homedrive_c : "");
	    std::string homepath(homepath_c ? homepath_c : "");
	    std::string home = homedrive + homepath;



You can test using `./gw.exe test/test.fa -n -b test/test.bam -r chr1 > test.png`

Biswa96:
Thanks for the update. Now I get compiler error like the following any many more.

```
src/themes.cpp:298:30: error: no match for 'operator=' (operand types are 'std::string' {aka 'std::__cxx11::basic_string<char>'} and 'std::filesystem::__cxx11::path')
  298 |             path = homedir / gwini;
      |                              ^~~~~
```

In previous comment, I just provided those points as example.

I think it's OK to use Windows's side HOMEPATH. GitHub Desktop's git also uses `.gitconfig` in that directory. But Windows HOMEPATH is not same as msys2's $HOME.

Biswa96:
Another.

```
src/themes.cpp: In member function 'void Themes::IniOptions::readIni()':
src/themes.cpp:318:32: error: no matching function for call to 'mINI::INIFile::INIFile(std::filesystem::__cxx11::path&)'
  318 |         mINI::INIFile file(path);
      |                                ^
```

If you want I can try to add a CI script to automate build/test steps in msys2/mingw.

kcleal:
Thank you for your patience! I believe the string/filesystem problem should be fixed now. You are welcome to contribute a CI script if possible, that would be amazing.

kcleal:
Looks like that is during cleanup. Perhaps I need an explicit check for nullptr for the bug on Line 102 in plot_manager.cpp. Ive changed `if (window) { ` to `if (window != nullptr) {` which might fix that? I dont know about the mingw64 problem

Biswa96:
I have compiled to program from latest git commit. But it still segfault with your given example command and the generated test.png is not valid also.

kcleal:
Ok I have added a new test that should work without needing a gpu context:

`gw.exe test/test.fa -v test/test.vcf -b test/test.bam -o testout --no-show -t 1`

Biswa96:
Should not the segfault be fixed?

kcleal:
Are you testing on a server by any chance? The main use case for GW is local desktop machines.

kcleal:
Hmm im not sure why its not working. I have managed to get a demo working on two Windows systems so far. Im stumped

image (1)

kcleal:
The Makefile has -O3 switch, maybe change this to debug might but useful?

kcleal:
Hi @Biswa96,
I think I fixed the segfault in the latest commit. If you have time, would you be able to confirm if its working... Thanks!

Biswa96:
The segfault got fixed but the generated png file is somewhat corrupt.

```
$ ./gw.exe test/test.fa -n -b test/test.bam -r chr1 > test.png
$ file test.png
test.png: data
```

BTW, the `-ldl` in Makefile is not required in mingw.

kcleal:
Ok I realised that writing the file to stdout is not going to work under msys2, the file gets mangled. Ive added an error message if a user attempts this on windows. The correct way to test is `./gw.exe test/test.fa -n -b test/test.bam -r chr1  -o testout`

Biswa96:
I now get a valid png file.

kcleal:
Any ideas why the build is still failing? Some weird error message from clang64:

       D:/M/msys64/clang64/include/_mingw.h:435:8: error: division by zero in preprocessor expression
      #elif (__USE_MINGW_ANSI_STDIO + 0) != 0 || (1 - __USE_MINGW_ANSI_STDIO - 1) == 2
             ^~~~~~~~~~~~~~~~~~~~~~
      <command line>:3:35: note: expanded from here
      #define __USE_MINGW_ANSI_STDIO 1-I/include
                                       ~^~~~~~~~

UCRT64 has something similar:

       <command-line>: error: division by zero in #if

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+source=(https://github.com/kcleal/${_realname}/archive/refs/tags/v${pkgver}.tar.gz)
+sha256sums=('9084f95c7e28146aedbeae591bbbcefe429556313e584f2776c5c8472c749253')
+
+prepare() {
+  cd ${_realname}-${pkgver}
+}
+
+build() {
+    cd ${_realname}-${pkgver}
+    LDFLAGS+="-L${PREFIX}/lib -L${PREFIX}/lib/skia" CPPFLAGS+="-I${PREFIX}/include -I${PREFIX}/include/skia -I${PREFIX}/include/ncurses" make
```
mmuetzel, line 33: When appending to environment variables, start with a space `" "`. That way, you'll avoid that subsequent arguments concatenate into a single one with the first in this list.

Biswa96: This environment variables addition may not be required. The pkgconfig commands in Makefile already include those.

kcleal: Thanks! Ill add this in

mmuetzel: Additionally, I'm not sure if `${PREFIX}` exists as an environment variable (without setting it manually). Maybe, you meant `${MINGW_PREFIX}`?
But that might be mute if setting those variables isn't needed anyway.

mmuetzel:
Something seems to be off with the CI. Some runners didn't build anything with `[MSYS2 CI] SUCCESS: No changes in package recipes.` in the logs...
@MehdiChinoune also noticed something similar in a different PR earlier today.

mingw-w64-gw/PKGBUILD
mmuetzel: > Any idea why this message just happened?

The following linker command is missing the ${EXEEXT}:
```
  g++ -g  ./src/drawing.o  ./src/glfw_keys.o  ./src/hts_funcs.o  ./src/main.o  ./src/parser.o  ./src/plot_controls.o  ./src/plot_manager.o  ./src/segments.o  ./src/term_out.o  ./src/themes.o  ./src/utils.o -pipe -lskia -lm -ljpeg -lpng -lsvg -lhts -lfontconfig -lpthread -lskia -lgdi32 -lopengl32 -lole32 -loleaut32 -luuid -lfontsub -luser32 -lusp10 -lexpat -lpng16 -lz -lwebpmux -lwebpdecoder -lwebpdemux -lwebp -ljpeg -lharfbuzz -licuin -licuuc -licudt  -lharfbuzz-subset -lglfw3 -o gw
```

It should end in `-o gw.exe` on Windows.

kcleal: When building on a local windows machine, it added the .exe automatically. Ill try adding this to the recipie:

    mv "${pkgdir}${MINGW_PREFIX}/gw" "${pkgdir}${MINGW_PREFIX}/bin/gw.exe"

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+    cd ${_realname}-${pkgver}
+    make
+}
+
+package() {
+  ls
+  cd "${srcdir}"
+  mv "${pkgdir}${MINGW_PREFIX}/gw" "${pkgdir}${MINGW_PREFIX}/bin/gw.exe"
+  _exename=gw
+  cp -n "${pkgdir}${MINGW_PREFIX}/.gw.ini" ~
```
mmuetzel, line 41: You probably wouldn't want to copy this to the home directory of the build system.
Afaict, that won't do anything for the package.

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+
+build() {
+    cd ${_realname}-${pkgver}
+    make
+}
+
+package() {
+  ls
+  cd "${srcdir}"
+  mv "${pkgdir}${MINGW_PREFIX}/gw" "${pkgdir}${MINGW_PREFIX}/bin/gw.exe"
```
mmuetzel, line 39: Are you sure anything is already installed in the package directory at that point?
Wouldn't you want to copy/move that file from the build directory?

kcleal: Actually the ini file can be copied to the same location as the exe. The mv command fails with the above, and with mv gw.exe. Im confused about whats going on here. The gw binary looks like it was built from the make output, but how do I copy it to the correct mingw bin folder?

mmuetzel:
``` suggestion
  cp gw.exe "${pkgdir}${MINGW_PREFIX}/bin"
```

mmuetzel:
Let's see how that fares. Anyway, it might be worth using `install` instead. I.e., something along the lines of this:
``` bash
  install -D -m755 gw.exe "${pkgdir}${MINGW_PREFIX}/bin"
  install -D -m644 .gw.ini "${pkgdir}${MINGW_PREFIX}/bin"
```

mmuetzel:
Oops. That should probably have been:
``` bash
  install -D -m755 gw.exe "${pkgdir}${MINGW_PREFIX}/bin/gw.exe"
  install -D -m644 .gw.ini "${pkgdir}${MINGW_PREFIX}/bin/.gw.ini"
```

kcleal:
Yuuus

mmuetzel:
Some libraries that the executable links to weren't installed. E.g., for `CLANG64`:
```
  	glfw3.dll => not found
  	libjpeg-8.dll => not found
  	libwebp-7.dll => not found
  	libwebpmux-3.dll => not found
  	libwebpdecoder-3.dll => not found
  	libwebpdemux-2.dll => not found
```
Maybe (some of) the `makedepends` should actually be `depends`?

mmuetzel:
> Maybe (some of) the `makedepends` should actually be `depends`?

That'll probably be `"${MINGW_PACKAGE_PREFIX}-libwebp"` and `"${MINGW_PACKAGE_PREFIX}-glfw"`.

Biswa96:
> That'll probably be `"${MINGW_PACKAGE_PREFIX}-libwebp"` and `"${MINGW_PACKAGE_PREFIX}-glfw"`.

FYI. The libraries are required for skia. Due to unstable API in skia, its libraries are provided as static libraries.

mmuetzel:
> FYI. The libraries are required for skia.

So, should the dependency be added to the skia package instead?
Is it the same for glfw? It is currently a makedepends of this package.

I also noticed that the hts library was installed without its winpthreads dependency on CLANG64.

Biswa96:
> should the dependency be added to the skia package instead?

I have a little bit doubt there. The skia libraries are makedepends and not installed for users. 🤷‍♂️

mmuetzel:
> The skia libraries are makedepends and not installed for users. 🤷‍♂️

So, all makedependencies of skia (including this package) need to depend on `libwebp` for those shared libraries?

Biswa96:
> all makedependencies of skia (including this package) need to depend on 

Probably not all. Whenever I am in doubt about dependencies, I grab the packages from ntldd output, here `ntldd gw.exe`.

mmuetzel:
> Probably not all. Whenever I am in doubt about dependencies, I grab the packages from ntldd output, here ntldd gw.exe.

IIUC, that is essentially what the CI also does.
So, I'd guess for this package, we'd need to add the dependency on `webp`.

I'll open a PR to add the winpthread dependency to htslib.

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,43 @@
+# Maintainer: kez cleal <clealk@cardiff.ac.uk>
+
+_realname=gw
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.4.3
+pkgrel=1
+pkgdesc="A genome browser for bam/cram and vcf files"
```
mmuetzel, line 8: The `pkgdesc` of native Windows packages should end in " (mingw-w64)".

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+arch=(any)
+mingw_arch=('ucrt64' 'clang64')
+url="https://github.com/kcleal/gw"
+license=('MIT')
```
mmuetzel, line 12: Use an SPDX identifier to make the license field in the package overview clickable
``` suggestion
license=('spdx:MIT')
```

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+
+build() {
+    cd ${_realname}-${pkgver}
+    make
+}
+
+package() {
+  ls
+  cd "${srcdir}/${_realname}-${pkgver}"
+  ls -lah
```
mmuetzel, line 40: Please remove the `ls` commands from here.

mingw-w64-gw/PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: kez cleal <clealk@cardiff.ac.uk>
+
+_realname=gw
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.4.3
+pkgrel=1
+pkgdesc="A genome browser for bam/cram and vcf files (mingw-w64)"
+arch=(any)
+mingw_arch=('ucrt64' 'clang64')
```
mmuetzel, line 10: I didn't follow from the beginning. So, please excuse me if this was answered already:
Does the executable build in other environments, too?

kcleal:
It needs 64 environment, not sure if it will build on the other 64 environments

mingw-w64-gw/PKGBUILD
mmuetzel: > It needs 64 environment, not sure if it will build on the other 64 environments

Just for a test:
``` suggestion
mingw_arch=('mingw64' 'ucrt64' 'clang64')
```

kcleal:
> Now the example command fails with segfault. Here is the backtrace
> 
> ```
> (gdb) bt
> #0  0x00007ffe11718af3 in ?? () from F:\msys64\ucrt64\bin\glfw3.dll
> #1  0x00007ff7fc46ee94 in Manager::GwPlot::~GwPlot (this=0x3d5fdfab10, __in_chrg=<optimized out>) at src/plot_manager.cpp:102
> #2  0x00007ff7fcaf6b66 in main (argc=7, argv=0x20b180e09c0) at src/main.cpp:689
> ```
> 
> Also the mingw64 build fails with `undefined reference to '__imp__dupenv_s'` error. Let me add a CI script in gw repo. Back to drawing board...

Actual this was from @Biswa96

kcleal:
Oh yea, this:
   
    C:\_\mingw-w64-gw\src\gw-0.4.3/./include/glob.h:160: undefined reference to `__imp__dupenv_s'

kcleal:
Just wondering what the process if for getting this merged? Would someone be able to review and merge this build? Thanks

PR #15024, opened 2023-01-11
capstone: enable clangarm64

heljkon:
Try for clangarm64. Seems to be the last missing dependency for qemu clangarm64


Biswa96:
Adding mingw_arch does not require pkgrel bump generally.

heljkon:
OK, thanks for this information. Should I reset the pkgrel bump?

BTW: Since there was no clangarm64 check, how can this be tested? I don't own an arm64 machine, so this is only a try...

Biswa96:
> Should I reset the pkgrel bump?

Probably, because there is no change besides mingw_arch.

> I don't own an arm64 machine

We all have an aarch64 machine, don't we? 😉 Though most of them running a _weird_ version of Linux.

> how can this be tested?

For my case, I cross compile packages in x86_64 machine and run it in Windows ARM in Raspberry Pi 4. Others have better hardware. Without arm hardware, Windows ARM can be used in qemu 🐌

Biswa96:
> Has this to be done manually?

IIRC, @jeremyd2019 can test a PR in semi-automatic way.

The program can be cross compiled for Windows ARM without any issue. Do you need any specific test?

jeremyd2019:
I thought about having qemu on arm64, but AFAIK whpx is part of the x86 section of qemu code, so probably wouldn't *just work* on Windows on ARM64.  If the only option is tcg, may as well run tcg on an x86_64 box that is much more common.

I will try this PR though, wouldn't hurt to have the binary.

mingw-w64-capstone/PKGBUILD
```diff
@@ -10,7 +10,7 @@ pkgrel=3
 pkgdesc='Lightweight multi-platform, multi-architecture disassembly framework'
 url='https://www.capstone-engine.org/index.html'
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32', 'clangarm64')
```
jeremyd2019, line 13:
```suggestion
mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
```

No commas here

jeremyd2019:
https://github.com/msys2-arm/MINGW-packages/actions/runs/3895710764/jobs/6651362448 - worked

mingw-w64-capstone/PKGBUILD
heljkon: awkward fault, thanks

heljkon:
> I thought about having qemu on arm64, but AFAIK whpx is part of the x86 section of qemu code, so probably wouldn't _just work_ on Windows on ARM64. If the only option is tcg, may as well run tcg on an x86_64 box that is much more common.

I'd agree if accelerated virtualization was the only application. QEMU contains other tools as well, which may be interesting on Windows on ARM64, too. QEMU Agent, Qemu Image and of course emulation of other systems.

PR #15027, opened 2023-01-11
crt: Remove full path reference

Biswa96:


lazka:
What does this fix? (i.e. why is crt special here?)

Biswa96:
When there is any full path reference in a executable I use `grep -air msys64 pkg` to look into binaries. But it also shows the path from crt libraries. With this change, the full paths from crt are removed and the binaries will contain only full path from their code only.

I was using this method in my setup to fix path relocation issues, thought it may help others also. This does not fix any issue directly.

lazka:
ok, I guess.

I wonder if there is something we should do for all packages?

lazka (commented):
ok, I guess.

I wonder if there is something we should do for all packages?

Biswa96:
I am not sure if this will be useful for general use cases. Feel free to close.

lazka:
If it helps you find relocation issues, then it's fine with me.

lhmouse:
Do those paths exist only in symbols? Removing them might make debugging CRT code difficult, but it's also not so easy for MSVCRT or UCRT, so maybe it's fine.

Biswa96:
> Do those paths exist only in symbols? 

Full paths exists in .o and .a files and it comes from `__FILE__` macro in crt code. There may be some trick in autotools to resolve that path to current directory only. But I can not recall it right now.

Biswa96:
I am not sure if this will create any issue. So, I shall use this patch for my personal testing. Feel free to use or modify this change.

PR #15029, opened 2023-01-12
gettext: update to 0.21.1

mmuetzel:
Most patches have been applied upstream.

I tried to update the Fedora patch to their current one:
https://src.fedoraproject.org/rpms/gettext/blob/7aca11f7d43d10c47fbeb4efd16af2884b34f9c7/f/gettext-0.21.1-disable-libtextstyle.patch
However, it doesn't build with it.

`libtextstyle` seems to build fine for me locally in MINGW64.
However, there are duplicate symbol errors with symbols defined in the ncurses library (e.g., `tparm`) when linking using `lld` in the CLANG* environments. (I guess those are weak symbols on POSIX?) ld.bfd seems to ignore those duplicate symbols. But it's not clear to me which symbols it is eventually using. The ones from the object files from gettext or the one from the (static) ncurses library?
Anyway, disabling curses avoids those duplicate symbol errors.


mmuetzel:
Fwiw, from the DEPENDENCIES file:
> * GNU ncurses (preferred)
  or libtermcap (discouraged) or a curses library (legacy).
>  + Highly recommended.
    Needed for styling of terminal output (libtextstyle and the --color
    option of the 'msgcat' program).

lazka:
Do we know what it uses ncurses for?

mmuetzel:
IIUC, it's used for colored output of terminal applications.
Without it the output would be monochrome.

mmuetzel:
The main use-case of `gettext` is to provide localized text strings for applications. Not having colored terminal output for the executables included in this package is likely not an issue for dependent packages.
Also: We didn't build libtextstyle at all previously. Having a version with "reduced functionality" might not be worse than having no version of the library at all.

mmuetzel:
> From what I know libtextstyle is disabled in all distros because of it containing an vulnerable CSS parser that no one is fixing. Since there is no distro shipping this, no one will miss it.

Ah, thanks. I wasn't aware of that.
I'll try to have a look again at why it was failing to build with the patch from Fedora.

lazka:
Arch also has a patch here: https://github.com/archlinux/svntogit-packages/tree/packages/gettext/trunk

mmuetzel:
It needed some Windows specific workarounds/hacks. But it was building for me locally now without libtextstyle.

lazka:
thanks!

PR #15030, opened 2023-01-12
libiconv: add CP65001 as alias for UTF-8

### LINKED ISSUE #14947: [gdb][libiconv] Fails to preview strings on UTF-8 locales

GDB fails to display strings on UTF-8 locales, that is, when "Beta: Use Unicode UTF-8 for worldwide language support" is checked on the host.

Requirements for reproducing:
- No `.gdbinit`.

Steps to reproduce:

1. Write a variant of Hello World as follows:
```c
#include <stdio.h>

int hello(char *str)
{
    return printf("%s\n", str);
}

int main(void)
{
    hello("Hello World!");
}
```

2. Compile with `cc -O0`.
3. Run `gdb a.exe` (or whatever you named in `-o).
4. Set a breakpoint at the line 5, then run.

This is the result of autodetection from GDB:

```
(gdb) show target-charset
The target character set is "auto; currently CP65001".
(gdb) show host-charset
The host character set is "auto; currently CP65001".
```

### Expected behavior

gdb should output `"Hello World!"` for `str` without any configuration.

### Actual behavior

In place of `"Hello World!"`, gdb outputs `<error reading variable: Converting character sets: Invalid argument.>` for the parameter `str`.

### Verification

### Windows Version

MINGW64_NT-10.0-22621

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Maybe

mcendu:
- [x] Depends on #15628


mcendu:
not sure how to sign binaries...

MehdiChinoune:
bump pkgrel `pkgrel=2`

alvinhochun:
Have you reported the issue to or discussed with upstream (both libiconv and binutils)? This is probably something that should be fixed upstream so all other mingw-w64 toolchains can benefit from the fix.

mcendu:
> Have you reported the issue to or discussed with upstream (both libiconv and binutils)? This is probably something that should be fixed upstream so all other mingw-w64 toolchains can benefit from the fix.

Have not, yet.

lazka:
hm, why is this a 4000+ line patch?

mcendu:
The following is the only actual change made:

```diff
diff -urN a/lib/encodings.def b/lib/encodings.def
--- a/lib/encodings.def	2021-06-06 17:50:48.000000000 +0800
+++ b/lib/encodings.def	2023-01-12 21:17:32.598803000 +0800
@@ -59,7 +59,7 @@

 DEFENCODING(( "UTF-8",                  /* IANA, RFC 2279 */
             /*"UTF8",                      JDK 1.1 */
-            /*"CP65001",                   Windows */
+              "CP65001",                /* Windows */
             ),
             utf8,
             { utf8_mbtowc, NULL },        { utf8_wctomb, NULL })
```

The rest is due to automatically generated files.

lazka:
I see, thanks. Can we move the re-generation of those files into the PKGBUILD recipe?

mcendu:
If we modify the PKGBUILD recipe to regenerate everything (by calling `make -f Makefile.devel totally-clean` followed by `make -f Makefile.devel all`), we also have to remove modifications to the autogenerated files from previous patches &ndash; namely `fix-pointer-buf.patch` &ndash; as without doing so patching would fail. Is a nice idea, though.

mcendu:
Caveat: the original `fix-pointer-buf.patch` patched the `encoding-*.h` files manually, which possibly explains why things weren't upstreamed.

Should I append to `fix-pointer-buf.patch`, or should I create another patch to for changes applied to `genaliases*.c`?

mcendu:
Upstream `Makefile.devel` is assuming that the system compiler is always GCC; this is not the case here.

lazka:
I've extracted the regen part of this to #15628

What's the status of the rest?

mcendu:
looks ok to me

mcendu:
`iconv` now seems to accept CP65001 as valid input. Did not fix the bug mentioned without rebuilding gdb.

mmuetzel:
Instead of merging the branches, could you please rebase on a current head? The PR doesn't apply any more.

mmuetzel:
Please, bump the pkgrel for this change.

mmuetzel:
An updated version of gmp is in staging.
Could you please bump the pkgrel of `gdb` as part of this PR and check if this fixes the issue for you?

mcendu:
Fixed.

Should I do it in a separate pull request?

mmuetzel:
> Should I do it in a separate pull request?

No, you can bump the pkgrel of `gdb` as part of this PR.

mmuetzel:
I installed the MINGW64 packages from this PR. And the example by @lhmouse with CJK characters from https://github.com/msys2/MINGW-packages/issues/14947#issuecomment-1374397251 (still) works if I manually set the console codepage to UTF-8 with `chcp 65001`. 
That was the same with the packages before this change for me.
This is with the system locale set to a single-byte ANSI codepage (cp850 ~ latin-1). I.e., I don't use the experimental "global" UTF-8 locale.
No regressions with this change here afaict.

If it fixes the issue on a system that uses the experimental "global" UTF-8 locale for you, I guess it is ok to merge this.

mcendu:
As I can tell from the logs, GitHub Actions uninstalled the modified libiconv from this PR before building gdb, meaning that gdb built by GitHub Actions for this PR is essentially identical to the previous version.

Maybe the gdb bump should be in a separate PR after all…

在 2023年2月17日，22:21，Markus Mützel ***@***.***> 写道：

﻿

I installed the MINGW64 packages from this PR. And the example by @lhmouse<https://github.com/lhmouse> with CJK characters from #14947 (comment)<https://github.com/msys2/MINGW-packages/issues/14947#issuecomment-1374397251> (still) works if I manually set the console codepage to UTF-8 with chcp 65001.
That was the same with the packages before this change for me.
This is with the system locale set to a single-byte ANSI codepage (cp850 ~ latin-1). I.e., I don't use the experimental "global" UTF-8 locale.
No regressions with this change here afaict.

If it fixes the issue on a system that uses the experimental "global" UTF-8 locale for you, I guess it is ok to merge this.

mcendu:
I would say that everything up to a1d8658fbbed1cc8aa68801a5c497ce0cc06f677 is OK for merge now. GDB version bump should or should not be split off, depending on how msys2 staging works.

mmuetzel:
Afaict, the `iconv` packages have been built *before* the `gdb` package in the CI. (That can sometimes go wrong if there are circular dependencies. But that's not an issue with the changes from this PR.)
I'm not sure why you think this has something to do with the packages in the staging repository. Afaict, the only package that might be related (`gmp`) is already out of staging and in the "main" repositories.

Tbh, I don't understand why you propose to revert bumping the pkgrel of `iconv`. There are clearly changes to the `iconv` package that require its pkgrel to be bumped.

Did you download and install the build artifacts from this PR and they don't work for you? If that is the case, why do you think it would be different if that commit were reverted?

mcendu:
> Afaict, the `iconv` packages have been built _before_ the `gdb` package in the CI. (That can sometimes go wrong if there are circular dependencies. But that's not an issue with the changes from this PR.) I'm not sure why you think this has something to do with the packages in the staging repository. Afaict, the only package that might be related (`gmp`) is already out of staging and in the "main" repositories.
> 
> Tbh, I don't understand why you propose to revert bumping the pkgrel of `iconv`. There are clearly changes to the `iconv` package that require its pkgrel to be bumped.
> 
> Did you download and install the build artifacts from this PR and they don't work for you? If that is the case, why do you think it would be different if that commit were reverted?

I didn't meant to revert changes to `iconv`. I just meant that the `gdb` artifacts build here won't work due to something CI-related.

mcendu:
CI log for reference:
```
[install] mingw-w64-x86_64-libiconv
...
[uninstall] mingw-w64-x86_64-libiconv
...
[build] mingw-w64-gdb
...
```

mmuetzel:
> I didn't meant to revert pkgrel bumping of `iconv`. I just meant that the `gdb` artifacts built here won't work due to something CI-related.

Did you actually test that? Or is this just a claim by you without any substance to it?

See the build log for gdb in the CI:
https://github.com/msys2/MINGW-packages/actions/runs/4202975032/jobs/7291789781#step:11:1553
>   mingw-w64-x86_64-libiconv-1.17-3-any downloading...

That *is* the version that was just built immediately before in the CI.

One more time: Did you actually download and install the build artifacts from this PR and they don't fix the issue for you?

mcendu:
It works. Sorry for the drama.

PR #15033, opened 2023-01-13
Build HDF5 1.14.0 for mingw-w64

mkitti:
This adapts the mingw-w64-hdf5 script for version 1.14.0 of the HDF5 C library.

Note that a 1.12.3 release is also expected in the near future.

Patch hdf5-fix-pkgconfig.patch was [already applied upstream](https://github.com/HDFGroup/hdf5/commit/dcccc355261e305a1d877a798d7fd46556b3cbce).


MehdiChinoune:
Could you pass `-DHDF5_INSTALL_MODULE_DIR=include/mod` to cmake to fix Fortran modules installation.

mkitti:
Sure. How does that work? Do we still need to copy the mod files?

MehdiChinoune:
> Sure. How does that work? Do we still need to copy the mod files?

That's how Fortran works!

mkitti:
Does using that option replace what we are doing with this copy command?

```
    cp ${srcdir}/build-${MSYSTEM}/mod/static/*.mod ${srcdir}/build-${MSYSTEM}/bin/static/Release
```

mmuetzel:
ArchLinux seems to install Fortran .mod files directly in the `include` folder. E.g.: https://archlinux.org/packages/community/x86_64/netcdf-fortran/files/
or: https://archlinux.org/packages/community/x86_64/hdf5/files

Should we do the same? I.e., `-DHDF5_INSTALL_MODULE_DIR=include`?

> Does using that option replace what we are doing with this copy command?
> 
> ```
>     cp ${srcdir}/build-${MSYSTEM}/mod/static/*.mod ${srcdir}/build-${MSYSTEM}/bin/static/Release
> ```

Yes, these copy command would no longer be needed, I'd guess.

mkitti:
I think we should probably still deploy the mod files to `${srcdir}/build-${MSYSTEM}/bin/static/Release` here purely based on precedent.

mkitti:
Do we have some other precedent in MSYS2/MINGW where Fortran users would expect .mod files to be in the include folder?

mmuetzel:
> Do we have some other precedent in MSYS2/MINGW where Fortran users would expect .mod files to be in the include folder?

E.g., `flang` installed the .mod files in `include/flang`:
https://packages.msys2.org/package/mingw-w64-clang-x86_64-flang?repo=clang64

IIUC, it is up to the project if they prefer installing their header directly in `include` or in `include/${name}`. I'd guess the same can be applied for Fortran modules.

mmuetzel:
> I think we should probably still deploy the mod files to `${srcdir}/build-${MSYSTEM}/bin/static/Release` here purely based on precedent.

Are there actually any .mod files that are copied by that command? Does it fail to install the .mod files if those commands are removed?
Maybe, that copy command was only needed in older versions?

MehdiChinoune:
```diff
  @@ -167,64 +175,11 @@
   /mingw64/include/H5VLpassthru.h
   /mingw64/include/H5VLpublic.h
   /mingw64/include/h5z.mod
  +/mingw64/include/H5Zdevelop.h
   /mingw64/include/H5Zpublic.h
   /mingw64/include/hdf5.h
   /mingw64/include/hdf5.mod
   /mingw64/include/hdf5_hl.h
  -/mingw64/include/shared/
  -/mingw64/include/shared/h5_gen.mod
  -/mingw64/include/shared/h5a.mod
  -/mingw64/include/shared/h5d.mod
  -/mingw64/include/shared/h5ds.mod
  -/mingw64/include/shared/h5e.mod
  -/mingw64/include/shared/h5f.mod
  -/mingw64/include/shared/h5fortkit.mod
  -/mingw64/include/shared/h5fortran_types.mod
  -/mingw64/include/shared/h5g.mod
  -/mingw64/include/shared/h5global.mod
  -/mingw64/include/shared/h5i.mod
  -/mingw64/include/shared/h5im.mod
  -/mingw64/include/shared/h5l.mod
  -/mingw64/include/shared/h5lib.mod
  -/mingw64/include/shared/h5lt.mod
  -/mingw64/include/shared/h5lt_const.mod
  -/mingw64/include/shared/h5o.mod
  -/mingw64/include/shared/h5p.mod
  -/mingw64/include/shared/h5r.mod
  -/mingw64/include/shared/h5s.mod
  -/mingw64/include/shared/h5t.mod
  -/mingw64/include/shared/h5tb.mod
  -/mingw64/include/shared/h5tb_const.mod
  -/mingw64/include/shared/h5vl.mod
  -/mingw64/include/shared/h5z.mod
  -/mingw64/include/shared/hdf5.mod
  -/mingw64/include/static/
  -/mingw64/include/static/h5_gen.mod
  -/mingw64/include/static/h5a.mod
  -/mingw64/include/static/h5d.mod
  -/mingw64/include/static/h5ds.mod
  -/mingw64/include/static/h5e.mod
  -/mingw64/include/static/h5f.mod
  -/mingw64/include/static/h5fortkit.mod
  -/mingw64/include/static/h5fortran_types.mod
  -/mingw64/include/static/h5g.mod
  -/mingw64/include/static/h5global.mod
  -/mingw64/include/static/h5i.mod
  -/mingw64/include/static/h5im.mod
  -/mingw64/include/static/h5l.mod
  -/mingw64/include/static/h5lib.mod
  -/mingw64/include/static/h5lt.mod
  -/mingw64/include/static/h5lt_const.mod
  -/mingw64/include/static/h5o.mod
  -/mingw64/include/static/h5p.mod
  -/mingw64/include/static/h5r.mod
  -/mingw64/include/static/h5s.mod
  -/mingw64/include/static/h5t.mod
  -/mingw64/include/static/h5tb.mod
  -/mingw64/include/static/h5tb_const.mod
  -/mingw64/include/static/h5vl.mod
  -/mingw64/include/static/h5z.mod
  -/mingw64/include/static/hdf5.mod
   /mingw64/lib/
   /mingw64/lib/cmake/
   /mingw64/lib/cmake/hdf5-config.cmake
  @@ -257,6 +212,61 @@
   /mingw64/lib/pkgconfig/hdf5_hl.pc
   /mingw64/lib/pkgconfig/hdf5_hl_cpp.pc
   /mingw64/lib/pkgconfig/hdf5_hl_fortran.pc
  +/mingw64/mod/
  +/mingw64/mod/shared/
  +/mingw64/mod/shared/h5_gen.mod
  +/mingw64/mod/shared/h5a.mod
  +/mingw64/mod/shared/h5d.mod
  +/mingw64/mod/shared/h5ds.mod
  +/mingw64/mod/shared/h5e.mod
  +/mingw64/mod/shared/h5f.mod
  +/mingw64/mod/shared/h5fortkit.mod
  +/mingw64/mod/shared/h5fortran_types.mod
  +/mingw64/mod/shared/h5g.mod
  +/mingw64/mod/shared/h5global.mod
  +/mingw64/mod/shared/h5i.mod
  +/mingw64/mod/shared/h5im.mod
  +/mingw64/mod/shared/h5l.mod
  +/mingw64/mod/shared/h5lib.mod
  +/mingw64/mod/shared/h5lt.mod
  +/mingw64/mod/shared/h5lt_const.mod
  +/mingw64/mod/shared/h5o.mod
  +/mingw64/mod/shared/h5p.mod
  +/mingw64/mod/shared/h5r.mod
  +/mingw64/mod/shared/h5s.mod
  +/mingw64/mod/shared/h5t.mod
  +/mingw64/mod/shared/h5tb.mod
  +/mingw64/mod/shared/h5tb_const.mod
  +/mingw64/mod/shared/h5vl.mod
  +/mingw64/mod/shared/h5z.mod
  +/mingw64/mod/shared/hdf5.mod
  +/mingw64/mod/static/
  +/mingw64/mod/static/h5_gen.mod
  +/mingw64/mod/static/h5a.mod
  +/mingw64/mod/static/h5d.mod
  +/mingw64/mod/static/h5ds.mod
  +/mingw64/mod/static/h5e.mod
  +/mingw64/mod/static/h5f.mod
  +/mingw64/mod/static/h5fortkit.mod
  +/mingw64/mod/static/h5fortran_types.mod
  +/mingw64/mod/static/h5g.mod
  +/mingw64/mod/static/h5global.mod
  +/mingw64/mod/static/h5i.mod
  +/mingw64/mod/static/h5im.mod
  +/mingw64/mod/static/h5l.mod
  +/mingw64/mod/static/h5lib.mod
  +/mingw64/mod/static/h5lt.mod
  +/mingw64/mod/static/h5lt_const.mod
  +/mingw64/mod/static/h5o.mod
  +/mingw64/mod/static/h5p.mod
  +/mingw64/mod/static/h5r.mod
  +/mingw64/mod/static/h5s.mod
  +/mingw64/mod/static/h5t.mod
  +/mingw64/mod/static/h5tb.mod
  +/mingw64/mod/static/h5tb_const.mod
  +/mingw64/mod/static/h5vl.mod
  +/mingw64/mod/static/h5z.mod
  +/mingw64/mod/static/hdf5.mod
   /mingw64/share/
   /mingw64/share/doc/
   /mingw64/share/doc/hdf5/
```
The proposed change is not about those files copied to bin directory, It's about the files above.
They were installed under `include` now they are installed in `mod`.

mmuetzel:
> The proposed change is not about those files copied to bin directory, It's about the files above.
> They were installed under `include` now they are installed in `mod`.

Correct. That would be fixed by setting `-DHDF5_INSTALL_MODULE_DIR=include`.

Continuing, @mkitti started to look at other parts of the build rules. Those might be worth checking now, too, imho.

mkitti:
I used the new option and cleaned up the patch files.

mingw-w64-hdf5/hdf5-proper-library-names-mingw.patch
```diff
@@ -1,20 +1,16 @@
---- hdf5-1.12.0/config/cmake_ext_mod/HDFMacros.cmake.orig	2020-04-26 10:08:44.715403300 +0300
-+++ hdf5-1.12.0/config/cmake_ext_mod/HDFMacros.cmake	2020-04-26 10:34:09.889478600 +0300
-@@ -125,6 +130,8 @@
+--- a/config/cmake/HDFMacros.cmake	2023-01-13 04:34:01.443034400 -0500
++++ b/config/cmake/HDFMacros.cmake	2023-01-13 04:36:14.419475400 -0500
+@@ -126,10 +126,12 @@
        OUTPUT_NAME_RELEASE        ${LIB_RELEASE_NAME}
        OUTPUT_NAME_MINSIZEREL     ${LIB_RELEASE_NAME}
        OUTPUT_NAME_RELWITHDEBINFO ${LIB_RELEASE_NAME}
 +      RUNTIME_OUTPUT_NAME        ${LIB_RELEASE_NAME}-0
```
kmilos, line 7: These two lines can probably be removed: `RUNTIME_OUTPUT_NAME` can be replaced w/ `-DCMAKE_SHARED_LIBRARY_NAME_WITH_VERSION=ON` in PKGBUILD, and default `ARCHIVE_OUTPUT_NAME` should be correct already?

MehdiChinoune: HDF5 doesn't define SOVERSION for MINGW https://github.com/HDFGroup/hdf5/blob/hdf5-1_14_0/config/cmake/HDFMacros.cmake#L169, It requires patching

kmilos: That's a bit weird... So either we bite the bullet, use the same SOVERSION as other platforms and rebuild, or leave things as is...

mkitti: Could we make both 1.14.0 and 1.12.3 available? Are they mutually exclusive.

1.12.3 is the end of the road for 1.12.x by the way, so I think we would want to move to 1.14 immediately.

image

mkitti:
I think we may be trying to do too much here. Could some of this be deferred to a `pkgrel=2`?

mingw-w64-hdf5/hdf5-proper-library-names-mingw.patch
mmuetzel: > Btw, the SOVERSION did change between 1.12.2 and 1.14.0, so keeping this artificially at `-0` makes no sense at all. A rebuild is needed anyway.

The package grokker only pointed out that `mingw-w64-python-h5py` would need to be rebuilt. (Maybe, that is the only package that uses a function for which the SOVERSION was bumped upstream?)

mmuetzel (changes requested):

mingw-w64-hdf5/PKGBUILD
```diff
@@ -93,12 +88,8 @@ build() {
...
   cd "${srcdir}/build-${MSYSTEM}"
 
-  if [[ "${MINGW_PACKAGE_PREFIX}" != *-clang-* ]]; then
-    mkdir -p ${srcdir}/build-${MSYSTEM}/bin/static/Release/
-    mkdir -p ${srcdir}/build-${MSYSTEM}/bin/shared/Release/
-    cp ${srcdir}/build-${MSYSTEM}/bin/static/*.mod ${srcdir}/build-${MSYSTEM}/bin/static/Release
-    cp ${srcdir}/build-${MSYSTEM}/bin/shared/*.mod ${srcdir}/build-${MSYSTEM}/bin/shared/Release
-  fi
+  # Previously we moved mod files to bin/[static|shared]/Release
+  # mod file are now in include
```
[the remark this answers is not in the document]
mkitti: Removed in  168f22c

mmuetzel:
According to the package grokker results, you'll still need to bump the pkgrel of `mingw-w64-python-h5py`.

mkitti:
> According to the package grokker results, you'll still need to bump the pkgrel of `mingw-w64-python-h5py`.

Sorry, I misunderstood. I've bumped it in https://github.com/msys2/MINGW-packages/pull/15033/commits/0907e257647ec6158d0254037bf82c797261a21c

mmuetzel (approved):

kmilos:
Please take this opportunity to remove the nonsense -0 versioning.

mmuetzel:
> Please take this opportunity to remove the nonsense -0 versioning.

I agree that this could probably be removed (unless there are some dependent packages that *need* the library to have a SOVERSION). But I also agree with @mkitti that it might be better to defer this to a subsequent PR because it would require to rebuild all of the reverse dependencies.

mkitti:
They are asking about the so version upstream as well:
https://forum.hdfgroup.org/t/msys2-mingw-w64-hdf5-1-14-0-build-scripts-send-patches-upstream/10693/2?u=kittisopikulm

I'll try dropping the `-0` later today. If it gets complicated, then maybe we defer?

mkitti:
We're going to need a new h5py release with https://github.com/h5py/h5py/pull/2194 for hdf5 1.14.0 compatibility

mkitti:
h5py upstream looks for `-[0-9]`

https://github.com/h5py/h5py/blob/01be8231e9e143793ae09050610f794b590c8bac/setup_configure.py#L248-L253

mkitti:
Maybe we should bump to `libhdf5-1.dll`?

mkitti:
Here's the PR that added the search for `-0` in upstream h5py. It was for mingw compat:
https://github.com/h5py/h5py/pull/2105

cygwin uses `cyghdf5-200.dll` apparently?

mkitti:
Building https://github.com/h5py/h5py/commit/01be8231e9e143793ae09050610f794b590c8bac with this patch works...

```
diff --git a/setup_configure.py b/setup_configure.py
index a36aa1fd..d0b67860 100644
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -249,8 +249,8 @@ class HDF5LibWrapper:
                 default_path = 'hdf5.dll'
                 regexp = re.compile(r'^hdf5.dll')
             else:
-                default_path = 'libhdf5-0.dll'
-                regexp = re.compile(r'^libhdf5-[0-9].dll')
+                default_path = 'libhdf5.dll'
+                regexp = re.compile(r'^libhdf5.dll')
             if sys.version_info >= (3, 8):
                 # To overcome "difficulty" loading the library on windows
                 # https://bugs.python.org/issue42114
```

It does look like we're going to need a h5py release though to get the h5py package to work with HDF5 1.14.

mkitti:
Without h5py master, I get the following error:

```
$ python
>>> import h5py
D:/msys64/mingw64/lib/python3.10/site-packages/numpy/core/getlimits.py:493: UserWarning: Signature b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\xcc\xfb\xbf\x
1e\xb2*\x00\x00\x00' for <class 'numpy.float128'> does not match any known type: falling back to type probe function.
This warnings indicates broken support for the dtype!
  machar = _get_machar(dtype)

# Gap inserted by me for clarity

  File "<stdin>", line 1, in <module>
  File "D:/msys64/mingw64/lib/python3.10/site-packages/h5py/__init__.py", line 56, in <module>
    from . import h5a, h5d, h5ds, h5f, h5fd, h5g, h5r, h5s, h5t, h5p, h5z, h5pl
  File "h5py/h5fd.pyx", line 220, in init h5py.h5fd
RuntimeError: Wrong file driver version # (wrong file driver version #)
```

The fix for this is https://github.com/h5py/h5py/pull/2187 which is now merged to h5py master branch.

Do we cherry pick this, pull h5py master, or wait for h5py to release a new version before proceeding?

mkitti:
I cherry picked https://github.com/h5py/h5py/pull/2194

We can now `import h5py`

```
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
D:/msys64/mingw64/lib/python3.10/site-packages/numpy/core/getlimits.py:493: UserWarning: Signature b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\xcc\xfb\xbf\x
1eS\x88\x00\x00\x00' for <class 'numpy.float128'> does not match any known type: falling back to type probe function.
This warnings indicates broken support for the dtype!
  machar = _get_machar(dtype)
>>> 
```

mmuetzel:
If you prefer to remove the SOVERSION suffix now, you'll need to rebuild *all* dependers as part of this PR. And check that the respective configuration steps are still able to detect the library without the suffix.

MehdiChinoune:
> check that the respective configuration steps are still able to detect the library without the suffix.

configuration steps search for import libraries which do not have soversion

mmuetzel:
> > check that the respective configuration steps are still able to detect the library without the suffix.
> 
> configuration steps search for import libraries which do not have soversion

Apparently, at least h5py is looking for the .dll.

mingw-w64-python-h5py/PKGBUILD
```diff
@@ -23,16 +23,20 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cython"
              "${MINGW_PACKAGE_PREFIX}-pkg-config")
 source=(https://github.com/h5py/h5py/releases/download/${pkgver}/h5py-${pkgver}.tar.gz
         001-mingw-python.patch
+	002-Merge-pull-request-2194-from-h5py-libhdf5-1.14.patch
```
mmuetzel, line 26: Please, use spaces instead of tab for indentation.

MehdiChinoune:
Please keep the `-0` or patch it to use SOVERSION.

mingw-w64-python-h5py/PKGBUILD
mmuetzel: When cherry-picking a patch from upstream, it is preferable to download the patch directly instead of checking in a copy here:
``` suggestion
        002-Merge-pull-request-2194-from-h5py-libhdf5-1.14.patch::https://github.com/h5py/h5py/commit/3dc3d7c089acaa4532f43ad2283e65a440ec55ac.patch
```
That will also make it clearer where this patch came from
That is, in case it applies as-is. If it does, don't forget to update the checksums if necessary and remove the patch from the repository.

mkitti: I literally `git cherry-picked` it on the 3.7.0 tag since I did not know if the the patch would apply directly. Let me see if I can generate the patch directly from the commit and have it apply.

mkitti: I'm not sure how to download the patch directly because it is a merge.

mkitti: > Please, use spaces instead of tab for indentation.

Fixed in [0461af9](https://github.com/msys2/MINGW-packages/pull/15033/commits/0461af9eedb1fe81e4eea76d6ec3fb354975bc26)

MehdiChinoune:
Could you please stash all these 16 commits into one and then force-push it.

mkitti:
You mean squash?

MehdiChinoune:
> You mean squash?

Yes, sorry!

mkitti:
Commits squashed

mingw-w64-python-h5py/PKGBUILD
mmuetzel: > I'm not sure how to download the patch directly because it is a merge.

Does the suggested change not download a patch that suffices. On a superficial look, the only difference to the merge is changes to their CI (which we don't need). Did I miss something?

mkitti: I added the annotation in [76568a0](https://github.com/msys2/MINGW-packages/pull/15033/commits/76568a0d6b5ef666a46f819fb5029e74cdb3dfb4)

mmuetzel: With that changed, please delete the checked in copy of the patch (both locally and from the repository) and update with `updpkgsums`.

mmuetzel:
Could you please squash the commits and force push again?
Other than that, this looks good to me now. 👍

mkitti:
@mmuetzel I squashed again to a single commit. Are we good to merge?

mmuetzel:
The auto-generated commit message from the squashed commit is pretty long but doesn't contain much useful information. I don't know how important "good" commit messages are for this project... 🤷‍♂️
(It might also have been a good idea to keep the changes for HDF5 and its dependency in two separate commits. Sorry, for not pointing that out earlier.)
But the changes themselves look good to me. 👍

mmuetzel:
`lib/libhdf5.settings` seems to leak a path from the build system:
```
               Module Directory: C:/_/mingw-w64-hdf5/src/build-MINGW64/mod
```

mkitti:
> `lib/libhdf5.settings` seems to leak a path from the build system:

How should we fix that?

mmuetzel:
> > `lib/libhdf5.settings` seems to leak a path from the build system:
> 
> How should we fix that?

Does this diff help?
``` diff
diff --git a/CMakeInstallation.cmake b/CMakeInstallation.cmake
index f643afcfb0..ec86fb02bb 100644
--- a/CMakeInstallation.cmake
+++ b/CMakeInstallation.cmake
@@ -126,6 +126,7 @@ if (H5_WORDS_BIGENDIAN)
 else ()
   set (BYTESEX little-endian)
 endif ()
+set (fmoddir ${CMAKE_INSTALL_PREFIX}/${HDF5_INSTALL_MODULE_DIR})
 configure_file (
     ${HDF_RESOURCES_DIR}/libhdf5.settings.cmake.in
     ${HDF5_SRC_BINARY_DIR}/libhdf5.settings ESCAPE_QUOTES @ONLY
```

mkitti:
I think there might be a way to detect an absolute path via cmake:
https://cmake.org/cmake/help/latest/command/if.html

mmuetzel:
> I think there might be a way to detect an absolute path

They don't seem to care about it for the pkgconfig files. So, maybe they won't care for it in this case either...

Does it build with that change? Does it help to omit leaking the directory from the build system?

mkitti:
I'm checking it out now. My suspicion is that we may need to do this earlier in the configuration process, but we'll see.

mmuetzel:
> My suspicion is that we may need to do this earlier in the configuration process, but we'll see.

Why? They are setting another substitution for that file immediately in the line before that...

mkitti:
I think we want to affect this line:
https://github.com/HDFGroup/hdf5/blob/509fe962d42b3f02d356d2668b04712df06daf7f/config/cmake/libhdf5.settings.cmake.in#L53

That means we need to modify `CMAKE_Fortran_MODULE_DIRECTORY` or change that to something else.
https://github.com/HDFGroup/hdf5/blob/509fe962d42b3f02d356d2668b04712df06daf7f/config/cmake/HDFMacros.cmake#L451-L452

mmuetzel:
https://github.com/HDFGroup/hdf5/blob/hdf5_1_14_0/src/libhdf5.settings.in#L55

Not sure which commit you are looking at.


Did the proposed psych apply to the release tarball? Does it build with that patch? Did it revolve the issue?

mkitti:
That libhdf5.settings gets overwritten by cmake I think. The relevant one here is here:

https://github.com/HDFGroup/hdf5/blob/0553fb7ac7f03a919c91cdbbd5648b8aadbe05af/config/cmake/libhdf5.settings.cmake.in#L53

mkitti:
The issue is resolved in
https://github.com/msys2/MINGW-packages/pull/15033/commits/106e9f58165907980f09fc6241750bd3d3040212#diff-0434b77fe796721c24f8931ad070599a3b19d26b1dc782eb6395f9926f0a20d2

mmuetzel:
Ah. They have different templates for their Autotools and cmake build systems. I didn't catch that while perusing their repo.

mmuetzel:
Something went wrong during the recent commits and the PR no longer applies on the repository.
It might be best to reset your local branch to the original HEAD with `git reset --soft efdc8b5d43493c48133f516c6808bcf1856f95dc~1`, rebase on a current head, and then apply the changes again in two commits: 1. update the hdf5 package, 2. update the h5py package.

mkitti:
OK, I rebased in two commits.

mkitti:
Looks good?

mmuetzel:
Probably a final thought: Since the license of HDF5 is a liberal but non-default license, it might make sense to bundle it in the package.
E.g., use a command similar to this one at the end of the `package()` function:
``` bash
  install -Dm644 "${srcdir}/${_realname}-${_ver}${patch}/COPYING" "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/COPYING"
```

mkitti:
There is a license file currently packaged at `mingw64/share/doc/hdf5/COPYING`. Do you want to remove it from that location and move it to `mingw64/share/license/hdf5/COPYING` ?

mmuetzel:
> There is a license file currently packaged at `mingw64/share/doc/hdf5/COPYING`. Do you want to remove it from that location and move it to `mingw64/share/license/hdf5/COPYING` ?

In that case, nevermind. Leave it as is. 👍

mmuetzel:
I installed the MINGW64 build artifact from this PR and tried to re-run the test suite of Octave.
It is failing with the following error:
```
  libinterp\octave-value\ov-oncleanup.cc-tst ..................... pass    1/1
  libinterp\octave-value\ov-range.cc-tst .........................Warning! ***HDF5 library version mismatched error***
The HDF5 header files used to compile this application do not match
the version used by the HDF5 library to which this application is linked.
Data corruption or segmentation faults may occur if the application continues.
This can happen when an application was compiled by one version of HDF5 but
linked with a different version of static or shared HDF5 library.
You should recompile the application or check your shared library related
settings such as 'LD_LIBRARY_PATH'.
You can, at your own risk, disable this warning by setting the environment
variable 'HDF5_DISABLE_VERSION_CHECK' to a value of '1'.
Setting it to 2 or higher will suppress the warning messages totally.
Headers are 1.12.2, library is 1.14.0
        SUMMARY OF THE HDF5 CONFIGURATION
        =================================
[...]
```

It is likely that some (or all?) of the dependent packages need to be rebuilt even without the SOVERSION bump.

For a list of packages that likely need a pkgrel bump, see e.g. the `Required by` section here:
https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-hdf5?repo=ucrt64

mmuetzel:
The PR doesn't apply again. That might be because of the merge commit. Could you please try to strip that (and maybe rebase your changes instead)?

Additionally, `adios2` fails to compile with the following error:
```
  C:/_/mingw-w64-adios2/src/ADIOS2-2.8.3/source/h5vol/H5VolUtil.c:59:48: error: unknown type name 'usigned'; did you mean 'signed'?
     59 | void gUtilConvert(hsize_t *fromH5, size_t *to, usigned int ndims)
        |                                                ^~~~~~~
        |                                                signed
```
That looks like a typo in one of our patches to me:
https://github.com/msys2/MINGW-packages/blob/0f343fd86704354130a900ac5f1a09db9099739d/mingw-w64-adios2/001-no-unit-on-mingw.patch#L8
That should probably `unsigned` (no `usigned`). I wonder why that didn't fail before.

mkitti:
Let's get this working first, and then I'll rebase. Do you want the pkgrel bumps all in one commit or multiple?

mmuetzel:
> Let's get this working first, and then I'll rebase.

Whatever you prefer. But keep in mind that stripping a merge commit might be getting more and more complicated if you push further changes on top of it.

> Do you want the pkgrel bumps all in one commit or multiple?

Whatever works best for you. I personally often separate commits if they are more than "just" a pkgrel bump.

mingw-w64-adios2/PKGBUILD
```diff
@@ -31,7 +31,7 @@ source=(https://github.com/ornladios/ADIOS2/archive/v${pkgver}/${_realname}-${pk
         001-no-unit-on-mingw.patch
         002-link-bpls-with-shlwapi.patch
         003-no-regex_h-with-clang.patch)
-sha256sums=('4906ab1899721c41dd918dddb039ba2848a1fb0cf84f3a563a1179b9d6ee0d9f'
+sha256sums=('c497df0848747292fe246bbf98b97e4b3e1b7ad180bef1be2ab718102d6db902'
```
mmuetzel, line 34: Are you sure you updated the correct checksum? The changed patch is the second file in the `source` list. You changed the first checksum...

mkitti: Fixed

mmuetzel: Btw, you could use `updpkgsums` to have a script update the checksums for you.

mkitti: Yes, but I'm not on a Windows-based computer at the moment nor am I using Arch Linux.

mmuetzel:
Please, wait a second with further pushes. I opened #15223 which will hopefully "future proof" the `uint` replacement for adios2.

mkitti:
Sorry. I did not see your comment until just now. I rebased without a rebuild for adios2 incorporated to see if there are any other challenges ahead.

mmuetzel:
> I rebased without a rebuild for adios2 incorporated to see if there are any other challenges ahead.

Good idea. 👍 Let's see how that fares.

In the meantime, #15223 is merged. When the current CI run has finished, please rebase again and bump the pkgrel of the updated adios2 package.

mmuetzel:
The pkgrel of Octave also needs to get bumped. (That might have gone lost when you rebased.)

But that might have been a "lucky error" in the current test run because it takes pretty long to compile. If we are lucky, the CI will finish without it before the 6(?) hours timeout.

mmuetzel:
CI timed out. But at least until that point, it didn't fail...

I asked on Discord if it would be ok to merge anyway (after the pkgrel of Octave and ADIOS2 is bumped again).

mmuetzel:
I rebased, bumped the pkgrel of the remaining packages and merged.

@mkitti: Thank you very much!

MehdiChinoune:
We missed the opportunity to enable SOVERSION for HDF5.

mmuetzel:
> We missed the opportunity to enable SOVERSION for HDF5.

We can still do that in another PR.
IIRC, you were ok with either keeping `-0` or enabling SOVERSION.

MehdiChinoune:
> IIRC, you were ok with either keeping `-0` or enabling SOVERSION.

Keeping `-0` to avoid rebuilding all dependent packages, but you rebuild all of them after all.

mmuetzel:
Sorry. That wasn't clear to me. I'd thought this was because some packages seemed to look for a hard-coded `-0` at the end of the library name...

mmuetzel:
But I agree: The way how the library refuses to work if the version in the public headers doesn't match the library essentially means that it **requires** a SOVERSION.

mkitti:
h5py seems to allow a range of single digit numbers for the SOVERSION.  Frankly, the patch applied to h5py was intended specifically for msys2. It seems they would be happy to accommodate whatever scheme is determined here.

Before coming up with our own scheme, I also suggest researching what happens on other platforms. I can communicate with The HDF Group directly regarding their opinion on the matter as well.

lazka:
vigra failed to build, as it doesn't find boost python (despiteit being there) and the later on fails when fixing the cmake files.

* Arch builds it from git + some patches, which might be something to look at: https://github.com/archlinux/svntogit-community/tree/packages/vigra/trunk
* The easiest "fix" would be to disable its Python support, not sure if anyone would notice.

PR #15041, opened 2023-01-14
openssl: update to 3.0.7

MehdiChinoune:


MehdiChinoune:
It's not possible to rebuild all those packages in this PR, so I am going to merge this PR first, wait until It is available on staging repo then I will push rebuilding packages gradually.

mingw-w64-python2/PKGBUILD
```diff
@@ -18,7 +18,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-bzip2"
          "${MINGW_PACKAGE_PREFIX}-libffi"
          "${MINGW_PACKAGE_PREFIX}-ncurses"
-         "${MINGW_PACKAGE_PREFIX}-openssl"
+         #"${MINGW_PACKAGE_PREFIX}-openssl"
```
lazka, line 21: puh... python without ssl module

MehdiChinoune: It detects it as openssl 0.0.0 and doesn't use at all and builds its own.
I think python2 only needed for gimp, so It's not a big deal I guess.

lazka: yeah, even if it builds its own (not sure how it would) it would be broken due to missing relocation support. and python2-pip is broken too.

I'll try https://github.com/fedora-python/cpython/commit/6c79f8bc416bee20df786e28f5bbc5417bdbe8f6 maybe

MehdiChinoune: I have revert python2 changes, I am going to open a separate PR for it.
OK to merge.

MehdiChinoune:
Could someone test it on clangarm64 @jeremyd2019 @driver1998 @Biswa96

mingw-w64-openssl/PKGBUILD
```diff
@@ -81,22 +74,16 @@ build() {
     x86_64)
       _mingw=mingw64
       ;;
-    armv7)
-      _mingw="mingwarm32 no-asm"
-      ;;
     aarch64)
       _mingw="mingwarm64 no-asm"
```
jeremyd2019, line 78: I think when I looked into openssl that mstorsjo had contributed fixes to the asm so it would work, but it was in openssl 3.  Might want to look into that again, see if we can enable asm now.

jeremyd2019: can't find it now.  maybe it was gnutls.  will try without no-asm anyway, see what happens

jeremyd2019: @mstorsjo hate to bug you on this, was openssl asm something you looked into, or am I mis-remembering?

jeremyd2019: https://github.com/openssl/openssl/commit/b863e1e4c69068e4166bdfbbf9f04bb07991dd40 is interesting...

MehdiChinoune: Does it build as it is or not.
We could look at asm later

mstorsjo: I think I did hack openssl to builds for arm64 at some point, but it probably was without assembly, and I think I didn't try to upstream it. But if there's something upstream for clang-cl based builds, that's should probably be fairly easy to adapt for mingw configurations too.

jeremyd2019: > Does it build as it is or not.

Sorry, I am doing other things today, and was firing off builds trying different things for ASM.  This should be my last attempt (I'm pretty sure that above-mentioned commit would have to be backported, there was no tag associated with it), then I will try this PR as-is.

jeremyd2019:
Why did you close/delete this, do you still want me to try it on arm64?

MehdiChinoune:
I have pushed directly to master branch.

mingw-w64-openssl/PKGBUILD
jeremyd2019: yeah, that patch is apparently not preset in 3.0.7 at least, will try as-is (with no-asm)

jeremyd2019:
Well, you should get an answer quicker that way :grin:  I only have the QC710 for autobuild and a raspberry pi 4 8GB for testing things, and the raspberry pi is much slower (but it *can* build flang, because it has 8GB rather than the 4GB of the QC710).

MehdiChinoune:
Maybe 2 or 3 other QC710 should be acquired for MINGW-packages and parallel msys2-autobuild!

mstorsjo:
> Maybe 2 or 3 other QC710 should be acquired for MINGW-packages and parallel msys2-autobuild!

For buying new HW, you really should go for the MS DevKit 2023 instead, which is supposedly quite performant - contrary to what I've heard about the QC710.

jeremyd2019:
Yes, I am in discussion with someone from Microsoft about getting one contributed, maybe.  Really the biggest issue seems to be with licensing - there is currently only "client" Windows on ARM, but the "client" license doesn't let you provide a service with it (like a github runner) or run it on shared tenancy cloud hardware (which shot down the idea of getting some Azure usage sponsored and using their Windows ARM64 support).  There is such a thing as a client license with shared tenancy allowed, but it's only available as a subscription to enterprises, not to mere mortals, so we couldn't figure out how to get that (not that that would have helped the 'providing a service' restriction).

jeremyd2019:
My run on the raspberry pi shows as failed, but when I looked at the logs it said all packages built successfully, so I don't know what failed, but it wasn't anything with the packages themselves.

Biswa96:
There was a temporary failure with mingw-w64-dcmtk, see https://github.com/msys2-arm/msys2-autobuild/actions/runs/3922445287/jobs/6705385420

PR #15042, opened 2023-01-14
boost: Extract names of ICU libraries from a pkg-config configuration file

slavaandrejev:
Names of libraries should be configured either using pkg-config or CMake configuration modules. In the case of ICU, there are pkg-config configuration files provided by the ICU maintainers. This change uses pkg-config to extract names of ICU libraries from the configuration files.


lazka:
Seems a bit brittle to me, is that really worth it?

mingw-w64-boost/PKGBUILD
```diff
@@ -108,6 +108,14 @@ setb2args() {
         local variant=release
     fi
 
+    # Extract names of ICU libraries from the pkg-config configuration file.
+    # Note that the acual order of the libraries in the array is irrelevant.
+    # They are all linked in the same way.
+    local -a icu_libs
+    IFS=" " read -r -a icu_libs <<< "$(pkgconf --libs icu-i18n)"
```
lazka, line 115:
```suggestion
    IFS=" " read -r -a icu_libs <<< "$(pkgconf --libs-only-l icu-i18n)"
```

slavaandrejev:
> Seems a bit brittle to me, is that really worth it? In which case would this help?

This is how it has to be done. We shouldn't “guess” and hard-code the names of the libraries. For example, on my system the names are: icui18n, icuuc, icudata. ICU libraries in boost has always been a huge pane in the rear. You can see it from the last change: after decades of existence, boost is still fixing stuff related to ICU. The reason is simple: boost ignores the maintainer provided configuration files. This change will help keeping things right for the future.

lazka:
yeah, but `pkgconf --libs` could also start outputting dependencies with some future release, breaking this code, and the order is not defined, which is what I mean with brittle.

slavaandrejev:
> the order is not defined

The order is irrelevant, that's another strange thing about what boost is doing. There are three libraries you have to link to, period. There is no need to write separate code for the three.

> could also start outputting dependencies with some future release

It's still better than it was. I am open to suggestions how to improve it.

slavaandrejev:
If there will be extra dependencies, it will break the existing boost build system for sure. We won't make things worse than they are. But this change improves them.

slavaandrejev:
@lazka Hi Christoph. What's the fate of this PR going to be?

lazka:
tbh, I still don't see the point in adding code that extracts names that wont change in the future.

But if you insist I'll merge it.

slavaandrejev:
> tbh, I still don't see the point in adding code that extracts names that wont change in the future.

It's a design principle of having less assumptions. We've learned it a hard way that assumptions lead to failures. The classic nowadays example is the maiden flight of Ariane 5 in 1996. A simple assumption that some number will never exceed the maximum value caused 300 million euros.

Yes, if you don't mind, please merge it.

lazka:
You are adding undocumented assumptions and removing documented ones, so I don't get that argument.

Anyway, fine.

PR #15031, opened 2023-01-12
mesa: Update to 22.3.3

pal1000:

PR #15044, opened 2023-01-14
mesa: enable clangarm64 support

driver1998:
OpenGL: 
d3d12 and llvmpipe working

Vulkan: 
dzn assert fails in the Qualcomm Direct3D 12 driver, works fine if you force it to use WARP.
```
WARNING: dzn is not a conformant Vulkan implementation, testing use only.
WARNING: dzn is not a conformant Vulkan implementation, testing use only.
Device 0 : Microsoft Direct3D12 (Qualcomm(R) Adreno(TM) 680 GPU)
Device 1 : Microsoft Direct3D12 (Microsoft Basic Render Driver)
FATAL ERROR: (QIC != QIC_TEX && QIC != QIC_LDST) && "TEX LDST instruction still needs rematerialization"
Assertion failed: (vk_res) == VK_SUCCESS, file Cube.cpp, line 791
```
```
(lldb) target create "a.exe"
(rrent executable set to 'D:\Documents\Desktop\Vulkan-Cube-master\a.exe' (aarch64).
(lldb) r
(lldb) Process 6764 launched: 'D:\Documents\Desktop\Vulkan-Cube-master\a.exe' (aarch64)
Process 6764 stopped
* thread #1, stop reason = Exception 0x80000003 encountered at address 0x7ffebe1fd770
    frame #0: 0x00007ffebe1fd774 qcdxarm64xcompiler8180.DLL`compile + 347764
qcdxarm64xcompiler8180.DLL`compile:
->  0x7ffebe1fd774 <+347764>: adrp   x8, 5424
    0x7ffebe1fd778 <+347768>: add    x8, x8, #0xd40
    0x7ffebe1fd77c <+347772>: cmp    x22, #0x0
    0x7ffebe1fd780 <+347776>: csel   x0, x22, x8, ne
(lldb)
```

lavapipe: Crashes on startup, see https://gitlab.freedesktop.org/mesa/mesa/-/issues/8085.

cc @pal1000 

~~This is unrelated to https://github.com/msys2/MINGW-packages/pull/15031, but might want to merge that first.~~


driver1998:
Can somebody restart the CI for me? Looks like it is complaining about not finding CC. It works fine locally.

I think it is ready to merge.
The dzn crash is not something we can act on (it crashes in the Qualcomm driver).
As for the lavapipe crash, I can push the workaround mentioned in that upstream issue if you want.

jeremyd2019:
> Can somebody restart the CI for me? Looks like it is complaining about not finding CC. It works fine locally.

I don't know if it'll do any good, but I will.  I think that at some point cc is no longer assumed to be installed, so maybe some package got missed adding it to makedepends?

RogueZach:
Maybe your device is preinstalled with a compatibility layer, "OpenCL™, OpenGL® and Vulkan® Compatibility Pack” 

I found that uninstalling it helps. Also note that, "dzn is not a conformant Vulkan implementation, testing use only," is the result of having said pack installed. Cheers

driver1998:
@RogueZach The compatibility pack is actually mesa as well, that's why it shows the same message.
Also at the time I sent this PR the compatibility pack does not have Vulkan (dzn) support.

pal1000:
> The compatibility pack is actually mesa as well,

Yes it only contains Mesa drivers maintained by Microsoft plus Microsoft OpenCL ICD.which [is its own separate project](https://github.com/microsoft/OpenCLOn12).

PR #15143, opened 2023-01-17
ca-certificates: Fix installation destination for certificates

MehdiChinoune:


MSP-Greg:
Please consider placing the files in both ssl and etc/ssl.  As stated elsewhere, OpenSSL 1.1.1 is EOL in September of this year...

dscho:
Hmm. This broke Git for Windows. And neither the commit message nor the PR description even so much as attempts to do an adequate job of explaining what on Earth is going on that this needed to change.

I had to follow some random references to figure out that [this change is because Arch does it this way](https://github.com/msys2/MINGW-packages/commit/2f97826e8a8fce0b9a49da7ea2bffbab7ce98eb5#r96397884). Well, given that RubyInstaller and Git for Windows got broken by this, it looks to me as if that was not a very strong justification for that change.

PR #15145, opened 2023-01-17
qt5-static: update to 5.15.8 and enable clangarm64

MehdiChinoune:


MehdiChinoune:
qt5-static need to be rebuilt against OpenSSL 3.0 anyway.

PR #15147, opened 2023-01-17
qt6-static: update to 6.4.2

MehdiChinoune:


MehdiChinoune:
It builds but It fails because of the long destination path.
Someone has to build it manually outside of the repo, somewhere like "D:\_\PKGBUILD"

lazka:
I did it manually with `C:\qt`.

I created https://github.com/msys2/msys2-autobuild/issues/71 so we don't forget

PR #15155, opened 2023-01-19
zxing-cpp: update description and source URL

axxel:
- the library can do way more than decode QRCodes
- the project moved to a new home recently


axxel:
The two build failures seem to be caused by:
```
      zxing-cpp-2.0.0.tar.gz ... FAILED
```
I can't make any sense of that because:
 * how is this related to my changes?
 * why is that very general error not showing up on the the other 3 CI builds?

PR #15157, opened 2023-01-19
libjxl: update to 0.8.0

kmilos:


Biswa96:
Though package-grokker did not report any rebuild, the [v0.8.0 release page](https://github.com/libjxl/libjxl/releases/tag/v0.8.0) mentioned the removed and changed APIs. Should the dependent packages be rebuilt?

kmilos:
True, but those removed calls were very "obscure" and I don't think, and the grokker didn't indeed find any dependent packages using them.

PR #15159, opened 2023-01-19
Add static library generation for libexiv2

shithappens2016:
some how qt6-static related to it, so it's better to have a static library
无标题

### DIFF
--- a/mingw-w64-exiv2/PKGBUILD
+++ b/mingw-w64-exiv2/PKGBUILD
@@ -65,6 +65,7 @@ build() {
     extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
   fi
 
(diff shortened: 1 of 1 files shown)


Biswa96:
qt6-static does not depends on libexiv2. That "Blocked by" simply tells the packager that not all packages are built in the queue list. 

If you want static library that's OK. The static libraries contains symbols from curl library with dllimport attributes. Those need to be fixed with proper macros.

```
$ nm libexiv2.a | grep -i __imp_curl
                 U __imp_curl_easy_cleanup
                 U __imp_curl_easy_getinfo
                 U __imp_curl_easy_init
                 U __imp_curl_easy_perform
                 U __imp_curl_easy_reset
                 U __imp_curl_easy_setopt
                 U __imp_curl_easy_strerror
                 U __imp_curl_version_info
```

PR #15161, opened 2023-01-19
Updated to stable ffmpeg 5.1.2

Maymne:
Removed the 0001 which has been included in base, updating the makefile lines for 0005 where the source no longer matched, updating the SHA256 for 5.1.2, and including the patch to remove Vulkan extensions from ffmpeg source.


lazka:
Related: #10725

Maymne:
Makes sense. The issue I was trying to solve is that ffmpeg 4 has issues with merging some formats properly and that was solved with ffmpeg 5.1.2, but with blender, opencascade, and OpenSceneGraph still have those holding issues blocking the upgrade we can't simplify everything, yet.

PR #15163, opened 2023-01-19
qt6-static: Use `Ninja+Release` on clangarm64

MehdiChinoune:


MehdiChinoune:
@jeremyd2019 Could you try of this one builds on clangarm64.

jeremyd2019:
I could try it on the raspberry pi (so as not to interfere with any autobuilds on the QC710) but I don't know if that would give any useful information.

MehdiChinoune:
`Ninja+Release` builds successfully for MINGW64 and UCRT64 (build+package), so I think It would too on clangarm64.

jeremyd2019:
https://github.com/msys2-arm/MINGW-packages/actions/runs/3962410561/jobs/6789032637 will see what happens.  the pi does have more free space and ram than the qc710, but is slower

mingw-w64-qt6-static/PKGBUILD
```diff
@@ -138,12 +138,17 @@ build() {
...
+  else
+    _extra_config+=('-GNinja' '-DCMAKE_BUILD_TYPE=Release')
+  fi
+
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=;-DCMAKE_CONFIGURATION_TYPES=;-DCMAKE_FIND_LIBRARY_SUFFIXES=" \
   ${MINGW_PREFIX}/bin/cmake \
     -Wno-dev \
     --log-level=STATUS \
-    -G "Ninja Multi-Config" \
-    -DCMAKE_CONFIGURATION_TYPES="Release;Debug" \
```
jeremyd2019, line 146: Just noticed I don't think you are actually using the new `extra_config` variable anywhere.

mingw-w64-qt6-static/PKGBUILD
```diff
@@ -138,12 +138,18 @@ build() {
...
+    ${_extra_config[@]} \
```
jeremyd2019, line 152:
```suggestion
    "${_extra_config[@]}" \
```

mingw-w64-qt6-static/PKGBUILD
```diff
@@ -138,12 +138,18 @@ build() {
   [[ -d B${MSYSTEM} ]] && rm -rf B${MSYSTEM}
   mkdir -p B${MSYSTEM} && cd B${MSYSTEM}
 
+  declare -a _extra_config
+  if [[ ${CARCH} != aarch64 ]]; then
+    _extra_config+=('-G"Ninja Multi-Config"' '-DCMAKE_CONFIGURATION_TYPES="Release;Debug"')
```
jeremyd2019, line 143:
```suggestion
    _extra_config+=('-GNinja Multi-Config' '-DCMAKE_CONFIGURATION_TYPES=Release;Debug')
```

jeremyd2019: I don't know about what cmake wants it to be, might need to be
```suggestion
    _extra_config+=('-G' 'Ninja Multi-Config' '-DCMAKE_CONFIGURATION_TYPES=Release;Debug')
```
To match what it used to be before the array

mingw-w64-qt6-static/PKGBUILD
jeremyd2019: I expect you still need to quote this ☝️ before it will work right

jeremyd2019:
Looks like clangarm64 failed with the same packaging error as clang64 (long paths).  So this package *could* be done for clangarm64, manually and without the debug variant, but is it worth the effort?  Is there anything that having it would unblock for you?

If I had to build this manually, I could probably find some external storage to do it on which would be slower but could provide the debug variant too.

MehdiChinoune:
> but is it worth the effort? Is there anything that having it would unblock for you?

Personally I don't use static libraries at all. It's just about qt-installer-framework could move to Qt6 in the near future (which fails to build againt shared qt5).

jeremyd2019:
> qt-installer-framework could move to Qt6 in the near future

That would result in the removal of 32-bit qt-installer-framework, wouldn't it?  32-bit Windows is still supported until October 2025, and an installer should be one of the *last* things to drop support, at least to show a more friendly error to users that the program can't be installed on their system.  Heck, for a long time installers came with a 16-bit stub that told the hypothetical 16-bit Windows user that the program wouldn't work for them... And all executables come with a DOS 'This program can't be run in DOS mode' stub built in.

MehdiChinoune:
Let's wait until a package requires this package.

PR #15164, opened 2023-01-19
suitesparse: update to 7.0.0

MehdiChinoune:


MehdiChinoune:
I don't undestand, why no changes were detected on 32bit environments?

MehdiChinoune:
> > I don't understand, why no changes were detected on 32bit environments?
> 
> SuiteSparse is broken on 32-bit platforms. We are building the last working version 5 for those.

How about octave?

mmuetzel:
> How about octave?

It will probably not compile with that version of SuiteSparse.
I have a patch for that lying around. But I intended to wait for the release before doing anything more because that particular part of the API changed again for the second beta. I wanted to avoid adapting the code again in case it would change again...

MehdiChinoune:
> > How about octave?
> 
> It will probably not compile with that version of SuiteSparse. I have a patch for that lying around. But I intended to wait for the release before doing anything more because that particular part of the API changed again for the second beta. I wanted to avoid adapting the code again in case the API would change again...

I was talking about "Why the changes were not detecetd on 32bit environments", but they were changes in octave PKGBUILD.

mmuetzel:
Octave doesn't build in CLANG* (Fortran compiler). I don't know why nothing built for MINGW32 though.

It should be possible to build Octave against SuiteSparse 7.0.0 by cherry-picking this commit:
https://hg.savannah.gnu.org/hgweb/octave/rev/aaffac4fbe30

PR #15167, opened 2023-01-20
gmp: enable static and shared libs

### LINKED ISSUE #15123: gmp: can no longer link gmp as a static library

`g++ test.cpp -static -lgmp`

where test.cpp is
`#include <gmp.h>`
`int main()`
`{`
`	mpz_t t; mpz_init(t);`
`	return 0;`
`}`

### Expected behavior

With the previous versions of MSYS2 - gmp, it builds fine.

### Actual behavior

`undefined reference to '__imp___gmpz_init'`

The diff of gmp.h with the previous version is
`> #define __GMP_LIBGMP_DLL  1`
`< #define __GMP_LIBGMP_DLL  0`
This is the reason of the error: __GMP_LIBGMP_DLL must be set to 0.

### Verification

### Windows Version

MSYS_NT-10.0-22621

_No response_

raedrizqie:
Fixes #15123


Biswa96:
I'm thinking to separate shared and static GMP builds in their own packages. This way no patches are required, I think.

raedrizqie:
without patching the gmp header, we will need 2 sets of gmp headers (if we want it to coexist).. that will make us to patch dependant packages to use the non default header, right?

Biswa96:
> we will need 2 sets of gmp headers (if we want it to coexist)..

Yeah, something like. I was imagining like these - 1. gmp with all libraries 2. shared headers 3. static headers. Then add specific headers in makedepends. Maybe I am missing something, not sure.

> that will make us to patch dependant packages to use the non default header, right?

But the dependent packages are already built with shared libraries. Do they need any changes for that above idea?

Biswa96:
> gdb will fail to build on current gmp because its expecting static gmp, mpfr and some other libs..

Understood. gdb could depends on mingw-w64-gmp and makedepends on mingw-w64-gmp-static-headers, would not that fix the build? Just thinking, please correct me.

raedrizqie:
ah.. understood.. fixing it in PKGBUILD..

Biswa96:
I want to avoid patching gmp package. But separating headers and libraries will break things in other projects which uses mingw-w64-gmp. I was waiting for others opinions too.

mmuetzel:
@Biswa96's idea sounds promising to me.
Just an idea that could possibly retain backward compatibility:
1. Rename the package containing the libraries (without headers) to `mingw-w64-gmp-libs`.
2. Make both packages that contain either the static or the shared header depend on `mingw-w64-gmp-libs`. Those could be named `mingw-w64-gmp-headers-static` and `mingw-w64-gmp-headers-shared`.
3. Have one of the two packages that contain the headers provide `mingw-w64-gmp`.

IIUC, that way you could decide which of the two versions (shared or static) would become the default for packages that depend on `mingw-w64-gmp`.

Depend packages that would like to provide shared *and* static libraries might also need similar treatment.

PR #15168, opened 2023-01-20
Add mingw-w64-dtc

elmarco:


elmarco:
For the context & motivation, we are trying to remove dtc submodule from QEMU: https://lists.gnu.org/archive/html/qemu-devel/2023-01/msg04178.html

Biswa96:
Should not the package contain the tools? For comparison, msys2's dtc has .exe files https://packages.msys2.org/package/dtc?repo=msys&variant=x86_64

elmarco:
@Biswa96 there are compilation issues I didn't try to solve. QEMU doesn't make use of the tools afaik, and it could use those provided by msys environment in that case.

Biswa96:
Understood. I can try to explore that area. Thanks.

Biswa96:
The [mingw-w64-dtc](https://packages.msys2.org/base/mingw-w64-dtc) package is in pacman repository now.

I have tested it in a weird way 🤓 

* Get the devicetree files from an Android phone with ADB

```
MSYS2_ARG_CONV_EXCL="*" adb.exe shell tar cf /sdcard/devicetree.tar /sys/firmware/devicetree
MSYS2_ARG_CONV_EXCL="*" adb.exe pull /sdcard/devicetree.tar
tar -xf devicetree.tar
```

* Create a dts file using dtc.exe from mingw-w64-dtc package

```
dtc.exe -I fs sys/firmware/devicetree/base/ -o devicetree.txt
cat devicetree.txt | head -n 11
/dts-v1/;

/ {

        soc {

                cpu-pmu {
                        compatible = "arm,armv8-pmuv3";
                        interrupts = <0x01 0x07 0xff00>;
                };
```

One positive side is that all the executables are compiled with gcc

image

PR #15169, opened 2023-01-20
dtc: Enable tools

Biswa96:


Biswa96:
@elmarco Would you like to review the changes please? Also just curious, is it possible to add a 'tests' option in meson.build in upstream? Tests can be built for Windows platform but requires many changes. Having an meson option to disable tests may help.

elmarco:
@Biswa96 second patch is missing. Yes, I guess we can add an option for upstream to disable tests.

Biswa96:
> second patch is missing.

It's intentional. That patch is downloaded from upstream and not required to save in this repository.

PR #15171, opened 2023-01-20
Disable stripping gi-docgen

janbrummer:
Current version does not work and throws:
Fatal error in launcher: Unable to find an appended archive.

Fix this by disabling strip.


Biswa96:
Thanks for the fix. Can you bump the pkgrel variable please?

janbrummer:
> Thanks for the fix. Can you bump the pkgrel variable please?

Sure. Updated.

PR #15215, opened 2023-01-21
libsquish: Enable OpenMP in arm64 builds

hmartinez82:


Biswa96:
Could you bump the pkgrel please?

Issue #14856, opened 2022-12-31
Add libva


pal1000:
### Package name

libva

### Brief description of package

Open source VA-API implementation

### URL for package's homepage

https://github.com/intel/libva

### Provide a basic test case to validate the package's functionality.

Try building Mesa3D D3D12 VA-API driver. See https://github.com/pal1000/mesa-dist-win/blob/master/buildscript/modules/mesa3d.cmd - search for `va-`.

### MINGW environments where you need the package

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

Maybe, if I have time.

pal1000:
Building libva on Windows uses Meson and it's straightforward:  https://github.com/pal1000/mesa-dist-win/blob/master/buildscript/modules/libva.cmd

PR #15219, opened 2023-01-21
libva: add package

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/14856

### DIFF
--- a/mingw-w64-libva/0001-add-lib-prefix-def.patch
+++ b/mingw-w64-libva/0001-add-lib-prefix-def.patch
@@ -1,0 +1,15 @@
+--- a/va/libva.def
++++ b/va/libva.def
+@@ -1,4 +1,4 @@
+-LIBRARY va
++LIBRARY libva
+ EXPORTS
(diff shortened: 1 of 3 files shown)


Biswa96:
@pal1000 Would you like to review this pull request please?
* Are the DLL and import libraries name correct? Should those be prefixed with 'lib' ? Should those have so version?
* Should I also add another commit to add `-Dgallium-va=enabled` with mesa?

pal1000:
> Are the DLL and import libraries name correct?

Yes, but it might not matter as Mesaa3D only uses libva headers according to [source](https://gitlab.freedesktop.org/search?search=_dep_va_name&nav_source=navbar&project_id=176&group_id=1155&search_code=true&repository_ref=main).

> Should I also add another commit to add -Dgallium-va=enabled

Why not.

Biswa96:
I have added the mesa changes. Looks good?

pal1000:
Yes.

PR #15221, opened 2023-01-22
gimp3: Use git commit instead of tarball when building

hmartinez82:


Biswa96:
May I ask why this is required?

hmartinez82:
@Biswa96 Because the last version that has a tarball is from last year. Since this is heavily under development still I thought that devs would be more interested in having a close to 2.99.15 as possible build

Biswa96:
> I thought that devs would be more interested in having a close to 2.99.15 as possible build

msys2 is not that much bleeding edge, is it? 🤔 According to [the download page](https://download.gimp.org/pub/gimp/v2.99/), version 2.99.14 is just 2 months old.

mmuetzel:
Afaict, builds from release tarballs are preferred over bleeding-edge repository checkouts. That is, unless there are good reasons for not using the latest release...

PR #15231, opened 2023-01-23
vigra: workaround for python-nose issue

mmuetzel:
There seems to be an issue with python-nose during the configuration step of vigra that eventually leads to it being built without numpy:
```
-- Searching for Python numpy: ok
  File "<string>", line 1, in <module>
  File "C:/msys64/mingw64/lib/python3.10/site-packages/nose/__init__.py", line 1, in <module>
    from nose.core import collector, main, run, run_exit, runmodule
  File "C:/msys64/mingw64/lib/python3.10/site-packages/nose/core.py", line 11, in <module>
    from nose.config import Config, all_config_files
  File "C:/msys64/mingw64/lib/python3.10/site-packages/nose/config.py", line 68
    except ConfigParser.Error, exc:
           ^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: multiple exception types must be parenthesized
-- Could NOT find Python nosetests ('import nose' failed)
[...]
--   Vigranumpy dependencies not found (vigranumpy disabled)
```

The packaging step fails because vigranumpy wasn't built.

I don't know enough about the Python ecosystem to actually fix the issue with python-nose. But these changes avoid the error in the packaging step.

### DIFF
--- a/mingw-w64-vigra/PKGBUILD
+++ b/mingw-w64-vigra/PKGBUILD
@@ -5,12 +5,12 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.11.1
 _pkgver=${pkgver//./-}
-pkgrel=9
+pkgrel=10
 pkgdesc="vigra - Generic Programming for Computer Vision (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


lazka:
ah, that's basically what I suggested here as second option: https://github.com/msys2/MINGW-packages/pull/15033#issuecomment-1401098891

I'll have a look tomorrow

lazka:
on a second though, just get this in. No one external has worked on this package during the last five years.

PR #15237, opened 2023-01-24
Add new package "sbc"

randstr:


Biswa96:
The clang builds fail with this error

```
lld: error: unknown argument: --version-script=../sbc-2.0/sbc/sbc.sym
```

Is it possible to use a [DEF file](https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files) to export APIs from DLL.

randstr:
I'll look into that... thanks for the pointer.

randstr:
I removed the --version-script linker argument and the visible DLL symbols are those listed in the sym file. Should be good enough...?

Biswa96:
> ...the visible DLL symbols are those listed in the sym file. Should be good enough...?

Not all symbols are exported. Here is the diff of exported APIs between ucrt64 and clang64.

```diff
--- libsbc-ucrt.def
+++ libsbc-clang.def
@@ -15,9 +15,6 @@
 sbc_init
 sbc_init_a2dp
 sbc_init_msbc
-sbc_init_primitives
-sbc_init_primitives_mmx
-sbc_init_primitives_sse
 sbc_parse
 sbc_reinit
 sbc_reinit_a2dp
```

Biswa96:
It seems that sbc_init_primitives, sbc_init_primitives_mmx and sbc_init_primitives_sse should not be exported by shared library, libsbc-1.dll. Because those functions are not defined with SBC_EXPORT macro. Also those are not exported in .so file in Linux. In case of gcc, probably `-Wl,--export-all-symbols` kicks in and exports all APIs.

Both gcc and clang build have the required APIs, so LGTM.

PR #15238, opened 2023-01-24
Add new package libsmi

randstr:


Biswa96:
The shared library creation was failed with this error message

```
libtool: link: warning: undefined symbols not allowed in x86_64-w64-mingw32 shared libraries
```

Should the library contain shared library? If yes the issue can be fixed my adding `-no-undefined` libtool option in appropriate automake file i.e. Makefile.am.

randstr:
Thanks, I added `-no-undefined` as you suggested.

randstr:
Should I add autoreconf to the PKGBUILD? I think the shared library is enabled now.

Biswa96:
> Should I add autoreconf to the PKGBUILD?

Yes, `autoreconf` command should be added in `prepare()` step. Otherwise, `make` command will re-run automake command at build time like the following.

```
cd ../../libsmi-0.5.0 && /bin/sh /C/_/B/src/libsmi-0.5.0/missing automake-1.14 --foreign lib/Makefile
```

randstr:
Added autoreconf. :crossed_fingers:

Biswa96:
Thank you for the contribution.

Biswa96:
I have checked that both clang64 and ucrt64 DLLs exports same APIs. Also enabled clangarm64 in this commit https://github.com/msys2/MINGW-packages/commit/4600f97ba16232f7f5dfb46dab9717a1e01defea

PR #15239, opened 2023-01-24
hdf5: enable soversion

MehdiChinoune:


mmuetzel:
It looks like the changes added the VERSION - not the SOVERSION - to the library name.
Was that intended?

PR #15243, opened 2023-01-25
gimp3: Fix exe, py and scm script support

hmartinez82:
I haven't found any other example of a completely new file being needed like this. 
What's the preferred way of adding the gimp-script-fu-interpreter.interp file below?

- Add file and just copy it to the package output
- Add a `.patch`  that creates the file in place
- Create the file via code in PKGBUILD


mingw-w64-gimp3/PKGBUILD
```diff
@@ -117,9 +120,14 @@ package() {
   cd "${srcdir}"/build-${MSYSTEM}
   ${MINGW_PREFIX}/bin/meson.exe install --destdir="${pkgdir}"
 
-  sed -e "s|${MINGW_PREFIX}/bin/python3|python3w.exe|g" -i "${pkgdir}${MINGW_PREFIX}"/lib/gimp/2.99/interpreters/pygimp.interp
+  # Adjust the python interpreter path
```
hmartinez82, line 123: The sed substitution from the GIMP 2 recipe didn't work. I think that the .interp file is a bit different in GIMP 3 (or when built with Meson). 
This is the file before the sed call:
```
python=C:\msys64\mingw64\bin/python3.EXE
python3=C:\msys64\mingw64\bin/python3.EXE
/usr/bin/python=C:\msys64\mingw64\bin/python3.EXE
/usr/bin/python3=C:\msys64\mingw64\bin/python3.EXE
:Python:E::py::python3:
```

Biswa96: Instead of editing file at package time, how about fix the code which generates the pygimp.interp file? If you want I can try look into this topic. Can you create an issue about this?

hmartinez82: @Biswa96 Yep. that's where I took the idea with sed originally. :)
The script-fu interp is only created inside the .iss recipe for the Windows installer, so I had to put the file in there.

Who can look at restarting the builds after the network error?

hmartinez82:
Closed this and moved to https://github.com/msys2/MINGW-packages/pull/15251 which includes fixes for Lua plugins too

PR #15247, opened 2023-01-25
hdf5: Use SOVERSION instead of VERSION in library name.

mmuetzel:
This PR would use the SOVERSION instead of the VERSION as a component of the names of the HDF5 libraries. That would make it more similar to the naming conventions used on other platforms.

@MehdiChinoune: Was using the VERSION in the library name intentional? Please, let me know if this PR looks good to you.
In case it does, we'd need to bump the pkgrel of the remaining dependent packages again.


mmuetzel:
According to the package grokker, the following packages would need be be rebuilt (again), too:
```
mingw-w64-matio
mingw-w64-armadillo
mingw-w64-netcdf
mingw-w64-pdal
mingw-w64-cgns
mingw-w64-vigra
mingw-w64-med
mingw-w64-alembic
mingw-w64-mathgl
mingw-w64-itk
mingw-w64-adios2
mingw-w64-netcdf-cxx
mingw-w64-opencv
mingw-w64-seacas
mingw-w64-openturns
mingw-w64-gdal
mingw-w64-octave
mingw-w64-vtk
```

MehdiChinoune (approved):

mmuetzel:
Thank you for your approval.

PR #15251, opened 2023-01-26
gimp3: Fix exe, py, lua and scm script support 

hmartinez82:
~~Needs https://github.com/msys2/MINGW-packages/pull/15244 for lua-lgi in aarch64 and https://github.com/msys2/MINGW-packages/pull/15257 for lua-lgi to work under luajit at all~~
All deps are satisfied
image


Biswa96:
I still get the following errors after running gimp in mingw terminal

```
GIMP-Error: Unable to run plug-in "web-browser.exe"
(F:\msys64\ucrt64\lib\gimp\2.99\plug-ins\web-browser\web-browser.exe)

Failed to execute child process (No such file or directory)
...
GIMP-Error: Unable to run plug-in "spyro-plus.py"
(F:\msys64\ucrt64\lib\gimp\2.99\plug-ins\spyro-plus\spyro-plus.py)

Failed to execute child process (No such file or directory)
```

and many more... Is that expected?

Biswa96:
Just asking. It seems that [luajit](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-luajit/PKGBUILD) is in conflict with lua51. Should lua51 be replaced with luajit?

hmartinez82:
Those errors are expected. There a glib PR to fix it.

For the lua-lgi case a decision must be made . Look at the lua-lgi Issue.

hmartinez82:
@Biswa96 @lazka As you commented, luajit should be used instead of lua51. I got a fix for luajit not loading lua-lgi in 
https://github.com/msys2/MINGW-packages/pull/15257

Also, for CLANG based builds, https://github.com/msys2/MINGW-packages/pull/15253 is also needed for any plugin to load at all.

hmartinez82:
I'll update this PR once the luajit fix is in

hmartinez82:
@Biswa96 @MehdiChinoune I've update the build recipe to depend on luaji instead of lua51 (Except on aarch64 that only has lua51) . But we should merge https://github.com/msys2/MINGW-packages/pull/15257 before merging this so lua-lgi works under luajit.

hmartinez82:
@Biswa96 https://github.com/msys2/MINGW-packages/pull/15257 Has been merged \o/ . This is ready to be merged, also if you can check that the plugins load now with luajit (after https://github.com/msys2/MINGW-packages/pull/15257) is built

hmartinez82:
@lazka The fix to luajit was built and uploaded already. This is ready for the merge :)

mingw-w64-gimp3/0003-mingw-add-ModuleFileName-dir-to-PATH.patch
```diff
@@ -0,0 +1,37 @@
+diff -bur gimp-2.99.14-c/app/main.c gimp-2.99.14/app/main.c
+--- gimp-2.99.14-c/app/main.c	2023-01-24 18:49:00.342793400 -0700
++++ gimp-2.99.14/app/main.c	2023-01-24 18:49:33.030776800 -0700
+@@ -637,6 +637,33 @@
+ 
+     SetDllDirectoryW (w_bin_dir);
+ 
++    /* Add ourself to system path */
```
lazka, line 8: Is there an upstream issue for this, or some comment why this is needed and what this helps with?

hmartinez82: I didn't research that sorry. I just did what was done in GIMP 2 too
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gimp/0003-mingw-add-ModuleFileName-dir-to-PATH.patch

lazka: Ah, I see, thanks. Just curious, are you aware of anything breaking without this patch?

hmartinez82: The .exe plugins fail to find libgimp @lazka, but it could've been something else. I can give it a try without this patch

hmartinez82: @lazka Gone is the patch 😊. I think that I was facing all plugins not working and tried to do what Gimp2 did, but without this patch, the .exe plugins still work and even double clicking gimp.exe from Windows Explorer also correctly loads the .exe plugins

hmartinez82:
Network issue during download 😑

mingw-w64-gimp3/PKGBUILD
```diff
@@ -117,9 +118,21 @@ package() {
...
   ${MINGW_PREFIX}/bin/meson.exe install --destdir="${pkgdir}"
 
-  sed -e "s|${MINGW_PREFIX}/bin/python3|python3w.exe|g" -i "${pkgdir}${MINGW_PREFIX}"/lib/gimp/2.99/interpreters/pygimp.interp
+  # Adjust the python interpreter path
+  sed -e "s|\=.*|\=python3w.exe|g" -i "${pkgdir}${MINGW_PREFIX}"/lib/gimp/2.99/interpreters/pygimp.interp
+  sed -e "s|::python3|::python3w.exe|g" -i "${pkgdir}${MINGW_PREFIX}"/lib/gimp/2.99/interpreters/pygimp.interp
 
-  rm -f "${pkgdir}${MINGW_PREFIX}/lib/gimp/2.0/modules/*.dll.a"
+  # Create the scm scripts interpreter path (Only created by the installer in Windows)
+  cp "${srcdir}"/../gimp-script-fu-interpreter.interp "${pkgdir}${MINGW_PREFIX}"/lib/gimp/2.99/interpreters
```
mmuetzel, line 126: Would `install -Dm644` be better instead of these `cp` commands?

lazka: what's the difference?

mmuetzel: I don't know. If file permissions don't matter, they are probably equivalent. 
It would just make it a bit more consistent with other packaging rules (e.g., how we install the license files...).

lazka: ok :+1:

PR #15253, opened 2023-01-26
glib2: Fix g_spawn_*() in CLANG builds

hmartinez82:
Backports the fix for https://gitlab.gnome.org/GNOME/glib/-/issues/2900


hmartinez82:
This issue was probably also present in UCRT builds

Issue #15245, opened 2023-01-25
lua-lgi: Fails to be found when running under luajit


hmartinez82:
### Description 

When installing lua-lgi after luajit, lua-lgi installs without errors (since luajit provides lua51), but it's then not found by luajit.

What are the options here ?
- Either luajit should also look for modules where lua51 modules are expected, or
- Create a luajit flavor of lua-lgi, for instance luajit-lgi 

### Steps to reproduce the issue
- Install `mingw-w64-x86_64-luajit`
- Install `mingw-w64-x86_64-lua51-lgi`
- Run `luajit -e 'print(require("lgi")._VERSION)'`

### Expected behavior

lua-lgi's version is returned. E.g: 
`0.9.2`

### Actual behavior

The lgi module is not found:
```
C:/msys64/mingw64/bin/luajit.exe: (command line):1: module 'lgi' not found:
        no field package.preload['lgi']
        no file '.\lgi.lua'
        no file 'C:\msys64\mingw64\bin\lua\lgi.lua'
        no file 'C:\msys64\mingw64\bin\lua\lgi\init.lua'
        no file '.\lgi.dll'
        no file 'C:\msys64\mingw64\bin\lgi.dll'
        no file 'C:\msys64\mingw64\bin\loadall.dll'
stack traceback:
        [C]: in function 'require'
        (command line):1: in main chunk
        [C]: at 0x7ff70c7a2100    
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22621

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [x] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

Yes. I can create an PR after a solution is picked up as in the Decription

hmartinez82:
@Biswa96 what should I do? :)

Biswa96:
I am not familiar with lua. The lua51 was being replaced with luajit and hence I just asked about that.

hmartinez82:
I prefer luajit too (I'm a `jit all the things` kind of guy), but lua-lgi is not working when luajit is used in lieu of lua51

PR #15257, opened 2023-01-27
luajit: Fix cc and enhance modules search path with lua5.1 paths

hmartinez82:
Since luajit is meant to be a drop-in replacement for lua5.1 it needs to support searching for modules in the same paths as lua5.1

Fixes https://github.com/msys2/MINGW-packages/issues/15245

```
luajit -e 'print(require("lgi")._VERSION)'
0.9.2
```

### DIFF
--- a/mingw-w64-luajit/003-lua51-modules-paths.patch
+++ b/mingw-w64-luajit/003-lua51-modules-paths.patch
@@ -1,0 +1,28 @@
+diff -bur LuaJIT-2.1.0-beta3-orig/src/luaconf.h LuaJIT-2.1.0-beta3/src/luaconf.h
+--- LuaJIT-2.1.0-beta3-orig/src/luaconf.h	2023-01-26 22:15:03.558193500 -0700
++++ LuaJIT-2.1.0-beta3/src/luaconf.h	2023-01-26 22:15:17.283052500 -0700
+@@ -18,12 +18,20 @@
+ ** In Windows, any exclamation mark ('!') in the path is replaced by the
+ ** path of the directory of the executable file of the current process.
+ */
+-#define LUA_LDIR	"!\\lua\\"
+-#define LUA_CDIR	"!\\"
++
++/* Paths for modules expecting non-jit lua5.1  */
++#define LUA_LDIR	"!\\..\\share\\lua\\5.1\\"
++#define LUA_CDIR	"!\\..\\lib\\lua\\5.1\\"
++
++#define LUAJIT_LDIR	"!\\lua\\"
++#define LUAJIT_CDIR	"!\\"
+ #define LUA_PATH_DEFAULT \
+-  ".\\?.lua;" LUA_LDIR"?.lua;" LUA_LDIR"?\\init.lua;"
++  ".\\?.lua;" LUAJIT_LDIR"?.lua;" LUAJIT_LDIR"?\\init.lua;" \
++              LUA_LDIR"?.lua;"  LUA_LDIR"?\\init.lua;" \
++              LUA_CDIR"?.lua;"  LUA_CDIR"?\\init.lua"
+ #define LUA_CPATH_DEFAULT \
+-  ".\\?.dll;" LUA_CDIR"?.dll;" LUA_CDIR"loadall.dll"
++  ".\\?.dll;" LUAJIT_CDIR"?.dll;" LUA_CDIR"?.dll;" LUAJIT_CDIR"loadall.dll"
++
+ #else
+ /*
+ ** Note to distribution maintainers: do NOT patch the following lines!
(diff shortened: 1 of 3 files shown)


mingw-w64-luajit/004-fix-default-cc.patch
```diff
@@ -0,0 +1,12 @@
+diff -bur LuaJIT-2.1.0-beta3-orig/src/Makefile LuaJIT-2.1.0-beta3/src/Makefile
+--- LuaJIT-2.1.0-beta3-orig/src/Makefile	2023-01-26 22:40:03.639259400 -0700
++++ LuaJIT-2.1.0-beta3/src/Makefile	2023-01-26 22:40:18.635694100 -0700
+@@ -24,7 +24,7 @@
+ # removing the '#' in front of them. Make sure you force a full recompile
+ # with "make clean", followed by "make" if you change any options.
+ #
+-DEFAULT_CC = gcc
++DEFAULT_CC = cc
```
hmartinez82, line 9: Their own Makefile is not doing what their docs say. Even you call make with `HOST_CC=cc` it still gets ignored by this line

hmartinez82:
@kou  @jeremyd2019 luajit on non-Windows was already looking for modules in the right places. I've just updated the Windows build to also look for modules where lua5.1 modules normally reside.

mingw-w64-luajit/003-lua51-modules-paths.patch
```diff
@@ -0,0 +1,28 @@
...
++#define LUA_LDIR	"!\\..\\share\\lua\\5.1\\"
++#define LUA_CDIR	"!\\..\\lib\\lua\\5.1\\"
++
++#define LUAJIT_LDIR	"!\\lua\\"
++#define LUAJIT_CDIR	"!\\"
+ #define LUA_PATH_DEFAULT \
+-  ".\\?.lua;" LUA_LDIR"?.lua;" LUA_LDIR"?\\init.lua;"
++  ".\\?.lua;" LUAJIT_LDIR"?.lua;" LUAJIT_LDIR"?\\init.lua;" \
++              LUA_LDIR"?.lua;"  LUA_LDIR"?\\init.lua;" \
++              LUA_CDIR"?.lua;"  LUA_CDIR"?\\init.lua"
```
kou, line 21: Is this line needed?

hmartinez82: We need it to be cc because there's no gcc in CLANG builds, from my understanding cc becomes gcc in MINGW builds and clang in CLANG builds

That line is the most important part because it's basically what I'm trying to fix 😅
It joins the search paths from luajit and lua into one

mingw-w64-luajit/004-fix-default-cc.patch
hmartinez82: We need it to be cc because there's no gcc in CLANG builds, from my understanding cc becomes gcc in MINGW builds and clang in CLANG builds

mingw-w64-luajit/003-lua51-modules-paths.patch
kou: It seems that it's done by `LUA_CPATH_DEFAULT`.

hmartinez82: @kou Correct, you have to expand the PR to see that that's exactly the constant I'm setting with the line above

hmartinez82: Does  `LUA_CPATH_DEFAULT` define where Lua modules are searched for? I that `LUA_CPATH_DEFAULT` only defined where to find native dynamic libraries.

mingw-w64-luajit/004-fix-default-cc.patch
kou: I see.

Anyway, does `make CC=cc` work without this patch?

mingw-w64-luajit/003-lua51-modules-paths.patch
kou: `L` in `LDIR` means Lua modules. `C` in `CDIR` means C modules.
It's strange that we use `CDIR` in `LUA_PATH_DEFAULT`.

hmartinez82: @kou Yes, naming not great, but I tried to stick as close as possible with the original https://github.com/lua/lua/blob/v5.1.1/luaconf.h#L90

mingw-w64-luajit/004-fix-default-cc.patch
hmartinez82: No. I just tried just in case in CLANG64:
```
Building LuaJIT 2.1.0-beta3
make -C src amalg
make[1]: Entering directory '/d/dev/Github/MINGW-packages/mingw-w64-luajit/src/luajit-2.1.0-beta3/src'
make[1]: gcc: No such file or directory
make[1]: gcc: No such file or directory
```
...

PR #15277, opened 2023-01-27
swi-prolog: update to 9.0.4

mgondan:


mgondan:
The patch has been reported upstream

lazka:
Are all those path changes, like `/mingw64/lib/swipl/library/http/http_exception.pl` -> `/mingw64/library/http/http_exception.pl` on purpose?

mgondan:
No, that doesn't make sense. Thanks for reporting.

PR #15314, opened 2023-01-28
python-llvmlite: install libllvmlite.dll.a to fix numba-jit

MehdiChinoune:


MehdiChinoune:
@ognevnydemon Could you try python-llvmlite package from this PR.

MehdiChinoune:
Sorry, It doesn't fix that issue. numba doesn't support numpy 1.24.

PR #15320, opened 2023-01-28
wireshark: Add package

randstr:
Once Wireshark 4.2 is released (probably around october) the package can be changed to follow the stable branch.

### DIFF
--- a/mingw-w64-wireshark/PKGBUILD
+++ b/mingw-w64-wireshark/PKGBUILD
@@ -1,0 +1,98 @@
+# Maintainer: Wireshark Core Team <wireshark-dev@wireshark.org>
+
+_realname=wireshark
+_commit=e7d5c49fe1dadfd6393539ba1b4d535087d79abe
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=4.1.0rc0.r1576.ge7d5c49fe1
+pkgrel=1
+pkgdesc="Network traffic and protocol analyzer/sniffer (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
+url="https://www.wireshark.org"
+license=('GPL2')
+depends=("${MINGW_PACKAGE_PREFIX}-bcg729"
+         "${MINGW_PACKAGE_PREFIX}-brotli"
+         "${MINGW_PACKAGE_PREFIX}-c-ares"
+         "${MINGW_PACKAGE_PREFIX}-glib2"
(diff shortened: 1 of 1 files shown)


Biswa96:
> ...the package can be changed to follow the stable branch.

Since it will be changed to stable release, is it possible to not use the `-git` suffix in package name?

Just wondering, does this work without the npcap driver? Should there be a message that user need that kernel mode driver?

randstr:
> Since it will be changed to stable release, is it possible to not use the -git suffix in package name?

I was thinking the package would be renamed then but if that is preferable then sure, I can do that.

> Just wondering, does this work without the npcap driver? Should there be a message that user need that kernel mode driver?

It works without npcap driver to read pcap files and other supported capture storage formats. To enable live capture from network interfaces npcap must be installed separately/manually.

randstr:
Oh and the software suggests installing npcap to the user the first time it runs. This message can be dismissed.

mingw-w64-wireshark-git/PKGBUILD
```diff
@@ -0,0 +1,92 @@
...
+         "${MINGW_PACKAGE_PREFIX}-zstd")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-python"
+             "${MINGW_PACKAGE_PREFIX}-qt6-tools"
+             'git')
+provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+source=("${_realname}"::"git+https://gitlab.com/wireshark/wireshark.git")
```
lazka, line 48: Please pin the commit, like here: https://github.com/msys2/MINGW-packages/blob/7c67d9c46554d10276b2017bcfd0290dfe24a1bf/mingw-w64-libserialport/PKGBUILD#L15 so we get somewhat reproducible builds.

lazka: (and remove the "-git" in the name as suggested)

Biswa96:
> I was thinking the package would be renamed...

Yeah, it can be renamed. But we also have to add `provides` and `replaces` options so that update process can replace wireshark-git to wireshark. I was just thinking to skip those options. Just opinion, not so important.

mingw-w64-wireshark-git/PKGBUILD
randstr: done

randstr:
Renamed package, pinned commit and added some extra options with comments.

PR #15358, opened 2023-01-29
vulkan-validation-layers: update to 1.3.240

MehdiChinoune:
Fails on CLANG environments!


Biswa96:
I can try to upstream the changes.

MehdiChinoune:
VkLayer_khronos_validation is a module library, It doesn't have an import library.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/v1.3.240/layers/CMakeLists.txt#L160

PR #15365, opened 2023-01-30
libde265: update to 1.0.10

kmilos:


kmilos:
I **_think_** CMake might be missing some functionality vs the Autotools build, like the visual inspection tool... I'd rather wait until upstream makes CMake default and deprecates Autotools. In any case, plenty of other project seem to rely on Autotools, hence the upstream bug report, and that pc file will need to get patched upstream anyway.

Biswa96:
The includedir has changed like this

```diff
-includedir=${prefix}/include/libde265
+includedir=${prefix}/include
```

Is that expected?

kmilos:
Yes, that has been the upstream default forever. I don't know why it was changed for MSYS2... Only libheif depends on libde265, and it already includes libde265/libde265.h so no rebuild is needed.

PR #15366, opened 2023-01-30
gstreamer: Update to 1.22

lazka:


lazka:
@Biswa96 mediafoundation is missing some APIs again. I've disabled it for now.

mingw-w64-gst-plugins-bad/PKGBUILD
```diff
@@ -155,6 +159,10 @@ build() {
     -Donnx=disabled \
     -Dopenaptx=disabled \
     -Dqroverlay=disabled \
+    -Ddirectshow=disabled \
+    -Dwic=disabled \
+    -Dgtk3=disabled \
```
Biswa96, line 164: Apology for this late question. gtk3 was disabled here but the dependency still has it. Should gtk3 be removed from depends list also?

lazka: good question, sadly I don't remember why I disabled it.

lazka: ah, I think the option got added with that release and is badly named: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/commit/7e18fc1b1f25501a6077fe681ef3ece1c7a1417c

so there might be other things depending on gtk3 besides this

Biswa96: Understood. If I enabled gtk3 the build process shows the following error

```
../gst-plugins-bad-1.22.2/ext/gtk/meson.build:10:18: ERROR: Dependency "gtk+-wayland-3.0" not found, tried pkgconfig and cmake
```

So, it should be disabled for Windows probably.

PR #15370, opened 2023-01-30
cmakerc: New package

podsvirov:


Biswa96:
Just to make sure, Is this package only for this file `$prefix/lib/cmake/cmakerc/cmakerc-config.cmake` ?

podsvirov:
Yes!

PR #15376, opened 2023-01-31
qrupdate: update to 1.1.5

mmuetzel:
Arch Linux also transitioned from the "classic" qrupdate (which is currently unmaintained) to qrupdate-ng:
https://github.com/archlinux/svntogit-community/blob/packages/qrupdate/trunk/PKGBUILD

The upstream build system changed from a Makefile to cmake rules. So, also update the build rules accordingly.

It currently doesn't build with LLVM Flang (because of missing _imag and some other math functions). But I hope the current build rules will allow to easily add `clang64` to the `mingw_arch`s in case it will compile with LLVM Flang 16 (or a later version).

Octave's test suite still passes with the updated package.

### DIFF
--- a/mingw-w64-qrupdate/PKGBUILD
+++ b/mingw-w64-qrupdate/PKGBUILD
@@ -6,34 +6,65 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}-svn")
 replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}-svn")
-pkgver=1.1.2
-pkgrel=2
+pkgver=1.1.5
+pkgrel=1
 pkgdesc="Fortran library for fast updates of QR and Cholesky decompositions (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


mingw-w64-qrupdate/PKGBUILD
```diff
@@ -6,34 +6,65 @@ pkgbase=mingw-w64-${_realname}
...
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}-svn")
 replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}-svn")
-pkgver=1.1.2
-pkgrel=2
+pkgver=1.1.5
+pkgrel=1
 pkgdesc="Fortran library for fast updates of QR and Cholesky decompositions (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
 url="https://sourceforge.net/projects/qrupdate/"
```
kmilos, line 14: Change the URL as well?

mmuetzel: Is there a newer URL for the package?
Arch Linux still uses the old URL.

Maybe: https://gitlab.mpi-magdeburg.mpg.de/koehlerm/qrupdate-ng ? That is more a repository than a website for the package. But it might contain more information than the sourceforge page...

What do you think?

kmilos: Sure

PR #15433, opened 2023-02-01
New package: portsmf

hmartinez82:
That pointer arithmetic is **old**!

### DIFF
--- a/mingw-w64-portsmf/001-portsmf-shared.patch
+++ b/mingw-w64-portsmf/001-portsmf-shared.patch
@@ -1,0 +1,39 @@
+diff -bur portsmf-orig/configure.ac portsmf/configure.ac
+--- portsmf-orig/configure.ac	2023-02-01 02:00:58.257411500 -0700
++++ portsmf/configure.ac	2023-02-01 02:03:22.187179700 -0700
+@@ -23,7 +23,7 @@
+ dnl keep autools files in a subdirectory
+ AC_CONFIG_AUX_DIR(autotools)
+ # which also contains a subdir containing macros
+-AC_CONFIG_MACRO_DIR([autotools/m4])
++dnl AC_CONFIG_MACRO_DIR([autotools/m4])
+ dnl set up automake
+ AM_INIT_AUTOMAKE([-Wall foreign])
+ dnl Audacity policy: don't enable automatic rebuild of configure et al if 
+@@ -49,7 +49,8 @@
+ dnl we will need an "install" program to be available
+ AC_PROG_INSTALL
+ dnl and ranlib for static libraries
+-AC_PROG_RANLIB
++AC_PROG_LIBTOOL
++LT_INIT([win32-dll])
+ 
+ dnl Make the install prefix available to the program so that the pkg-config file
+ dnl can be created correctly
+diff -bur portsmf-orig/Makefile.am portsmf/Makefile.am
+--- portsmf-orig/Makefile.am	2023-02-01 02:00:58.241785900 -0700
++++ portsmf/Makefile.am	2023-02-01 02:03:55.411446900 -0700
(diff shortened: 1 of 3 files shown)


mingw-w64-portsmf/001-portsmf-shared.patch
```diff
@@ -0,0 +1,39 @@
...
+ dnl set up automake
+ AM_INIT_AUTOMAKE([-Wall foreign])
+ dnl Audacity policy: don't enable automatic rebuild of configure et al if 
+@@ -49,7 +49,8 @@
+ dnl we will need an "install" program to be available
+ AC_PROG_INSTALL
+ dnl and ranlib for static libraries
+-AC_PROG_RANLIB
++AC_PROG_LIBTOOL
++LT_INIT([win32-dll])
```
hmartinez82, line 19: My knowledge of autotools is almost none. Is this necessary? I went based on what I read here https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html

Biswa96:
Can the project code be browsed in a website?

hmartinez82:
@Biswa96  yes https://sourceforge.net/p/portmedia/code/234/tree/portsmf/trunk/  :)

mingw-w64-portsmf/002-ptr-arithm-64bits.patch
```diff
@@ -0,0 +1,30 @@
...
+ #if defined(_WIN32)
+ #pragma warning(default: 4311 546)
+ #endif
+@@ -609,7 +609,7 @@
+ #if defined(_WIN32)
+ #pragma warning(disable: 4311 4312)
+ #endif
+-        assert((char *)(((long) (ptr + 7)) & ~7) <= fence);
++        assert((char *)(((long long) (ptr + 7)) & ~7) <= fence);
+ #if defined(_WIN32)
```
Biswa96, line 19: Generally, uintptr_t is used to cast int -> ptr type. Also make sure the program/library works in 64 bit Windows. The wiki page seems too old.

hmartinez82: This library is used by Audacity, and they have a 64 bits build. Of course building it with mingw on 64 bits is another story. I'll let you know once I get Audacity working. This is only a library. They don't even provide proper unit tests :(

hmartinez82:
@Biswa96 I'm unable to answer that.
I went based on portsmf being here https://github.com/audacity/audacity/blob/master/linux/required_libraries.md
and https://archlinux.org/packages/extra/x86_64/portsmf/

mingw-w64-portsmf/002-ptr-arithm-64bits.patch
hmartinez82: @lazka I'll give it a try. I thought that the arch flavor were gospel :)

hmartinez82: @lazka @Biswa96 That repo is not building . I know it belongs to the original author, but he broke his own build 🥲.
In that git repo, the README was renamed to .md from .txt , and now:
```
make: *** No rule to make target 'README.txt', needed by 'all-am'.  Stop.
```

hmartinez82: @lazka no problem. The package is ready for merge. I'm trying to get Audacity I'm only 3 other packages remaining

mingw-w64-portsmf/001-portsmf-shared.patch
lazka: if it helps building it, good enough. I'd guess it works because the library has no dllexport at all and also all visible symbols on Linux, so everything is exported.

lazka:
Followup in #15532

PR #10852, opened 2022-02-24
new packages: QGIS and qtkeychain-qt5

lbartoletti:
Hello,

I'd like to propose a new package: QGIS and a new library required by qgis: qtkeychain-qt5.

> QGIS is the leading Free and Open Source Desktop GIS. It allows you to create, edit, visualise, analyse and publish geospatial information on Windows, Mac OS, Linux, BSD and Android (via the QField app). We also provide an OGC Web Server application, a web browser client and developer libraries. The QGIS project is under very active development by an enthusiastic and engaged developer community with good mechanisms for help via stack exchange, mailing lists and (optionally) through a global network of commercial support providers.

https://qgis.org

PR #15435, opened 2023-02-01
[new-package] qgis 3.28.3

MehdiChinoune:
based on #10852
Fails to build because of some errors about some `dll_export`ed functions declared as `thread_local`.


MehdiChinoune:
Building bindings fails with ninja because of the known issue with add_custom_command/add_custom_target, but building with "MSYS Makefiles" takes about 1.5X time.

PR #15441, opened 2023-02-02
New package: suil

hmartinez82:


hmartinez82:
Interesting. LV2 is getting installed
`installing mingw-w64-clang-x86_64-lv2...`

But then the configure step fails to find it:
```

**Comment by hmartinez82** [comment]
Who can take a look at the network error and restart the build?
```
 installing mingw-w64-ucrt-x86_64-adwaita-icon-theme...
  error: command (/usr/bin/bash /usr/bin/bash -c . /tmp/alpm_xB5IXz/.INSTALL; post_install 43-1 ) failed to execute correctly
  installing mingw-w64-ucrt-x86_64-atk...
```

File: mingw-w64-suil/PKGBUILD [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+  "${MINGW_PACKAGE_PREFIX}-gtk3"
+  "${MINGW_PACKAGE_PREFIX}-lv2"
+  "${MINGW_PACKAGE_PREFIX}-meson"
+  "${MINGW_PACKAGE_PREFIX}-pkgconf"
+  "${MINGW_PACKAGE_PREFIX}-qt5-base" 
+)
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+optdepends=(
+  "${MINGW_PACKAGE_PREFIX}-glib2: Support for wrapping Qt5 in GTK3 and GTK2"
+  "${MINGW_PACKAGE_PREFIX}-gtk2: Support for wrapping Qt5 in GTK2 and GTK2 in Qt5"
```
lazka, line 24: is gtk2 needed? we are trying to not add any more users.

though it seems this will remove all plugins(??)

hmartinez82: @lazka I'll try to build without without tonight and will report back.

hmartinez82: @lazka It works without GTK2 🚀

mingw-w64-suil/PKGBUILD
```diff
@@ -50,6 +48,7 @@ build() {
     -Dcocoa=disabled \
     -Dx11=disabled \
     -Ddocs=disabled \
+    -Dgtk2=disabled \
```
hmartinez82, line 51: If GTK 2 is not explicitly disabled:
```
Run-time dependency gtk+-2.0 found: NO (tried pkgconfig and cmake)

../suil-0.10.18/meson.build:69:0: ERROR: Dependency "gtk+-2.0" not found, tried pkgconfig and cmake

A full log can be found at D:/Dev/Github/MINGW-packages/mingw-w64-suil/src/build-CLANG64/meson-logs/meson-log.txt
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
 ```

mingw-w64-suil/PKGBUILD
lazka: Does it really work? As I above now it's no longer linking to glib/gtk3/qt.

I don't know how to test it.

If this is good enough as a build dependency for other packages, then sure, we can leave it as is.

hmartinez82: @lazka I never used this library. It's just an upstream dep for Audacity. From my understanding suil only needs headers when building the compatibility between GUI frameworks. It seems that when running, it loads the libraries dynamically. 
Would be perhaps saying that we don't need the optdeps entry anymore?

lazka: yeah.

I'd say maybe revert the last commit here and makedepend on gtk2 for now.

I've asked upstream: https://gitlab.com/lv2/suil/-/issues/9

drobilla: The only relevant wrapper in suil here is win_in_gtk2.  This (like the name suggests) lets Gtk2 applications load native Windows (`HWND`) plugin UIs.  It's only useful for Gtk2 applications, and will only be missed at runtime by them.

So, it's reasonable enough to package suil without that wrapper for Windows.  Although the library does almost nothing in that case, that's suil doing its job: isolating this stuff from applications entirely.

PR #15443, opened 2023-02-02
Add gtkglarea

asctime:
[Requested by gcad3d](https://github.com/gcad3d/gcad3d/issues/47) team for Windows 64-bit migration (from MSVC 32-bit)


Biswa96:
GTK 2 reached its end of life in 2020, see https://blog.gtk.org/2020/12/16/gtk-4-0/. New packages with gtk2 will not be added. Also the source tarball is too old from 2014, see https://download.gnome.org/sources/gtkglarea/

Biswa96:
Some source code in gcad3d project has reference to gtk3. Is it possible to use gtk3 instead of gtk2?

asctime:
> GTK 2 reached its end of life in 2020, see https://blog.gtk.org/2020/12/16/gtk-4-0/. New packages with gtk2 will not be added.

Please reconsider. Many of your packages and GIMP and PIDGIN still require GTK2. Drawing a a purely arbitrary line doesn't help anybody. Second, software that works is better than software that doesn't, and I understand MSYS2 repo has quite a bit of that purely to stay "with the latest version". Your decision seems purely arbitrary. I am not asking you to support it (since I am told you can't really give GTK support anyway), just release it.

> Also the source tarball is too old from 2014, see https://download.gnome.org/sources/gtkglarea/

No. Source tarball is already pkgver=2.1.0 please look.

gcad3d:
Why gcad3d needs Gtk2-gtkglarea:
gcad3d is a 3D BREP - CAD software, 700 000 lines of pure C-code, is -
- the only CAD software completely independent from closed software,
- using Gtk. CAD needs many special display-objects -
- annotations, dimensions, symbols, different linetypes, highlighed, dimmed ..
- OpenGL-Displaylist (Legacy-OpenGL) supports this needs much better than VBOs,
- Gtk3 does NOT support Legacy - OpenGL;
- Gtk4 is not yet suitable for the operating systems of many users.
So the only possible way for the next 2 years is to use Gtk2+gtkglarea.

Biswa96:
There was some discussion about removing python2 or gtk2 like old packages. Please wait for @lazka 's decision.

gcad3d:
How about ctags and cscope (the fastest tools for handling really big projects)?

asctime:
@Biswa96 we should probably go with 'stable' and 'experimental' like in Debian, with archived numeric or named binary releases. All of the stuff like gnutls that doesn't really work yet can stay in experimental until proven otherwise and all tests pass. Right now things are a mess.

lazka:
GIMP with gtk3 support will hopefully come out at the end of this year, which is when we are planning to drop gtk2.

If this helps you port things to some non-long-dead libraries then I don't mind it.

Note that if we hit some problems with it in future rebuilds we will ping you and drop it if there is no help.

mingw-w64-gtkglarea/001-mingw-autotools.patch
```diff
@@ -0,0 +1,32 @@
...
++AM_PROG_AR
+ 
+ dnl Initialize libtool
+ LT_PREREQ([2.2.6])
+@@ -53,7 +54,7 @@ dnl  Check for the operating system
+ dnl
+ AC_MSG_CHECKING([for Win32])
+ case "$host" in
+-  *mingw* | pw32* | *cygwin*)
++  *mingw* | pw32* | *cygwin* | MSYS*)
```
lazka, line 18: this looks wrong, there is not MSYS host

asctime: @lazka, there was. Seems to have changed then. Recommend keeping keep it so long it doesn't hurt anything.

lazka: fair enough

asctime (commented):
Does it still build a DLL? native_win32=yes appeared to have that function. I think there was also some influence on OpenGL API via the generated config.h.
Edit: looks like it does..

ckitgui:
Sorry: cscope is OK; my fault was to use the windows-vim-package;
after install msys-vim evythings fine. Thanks to BBROEKER.

PR #15452, opened 2023-02-02
binutils: update to 2.40

MehdiChinoune:


MehdiChinoune:
For some reason autotools/libtool is preferring static libraries over shared ones, this happened with GDB where It picks libgmp.a instead of libgmp.dll.a and here where It picks libintl.a and libiconv.a.

PR #15464, opened 2023-02-03
tiled: New package

podsvirov:
With [Tiled](https://www.mapeditor.org/) we can make and edit levels, that can be read by [pytmx](https://packages.msys2.org/base/mingw-w64-python-pytmx) and rendered by [pygame](https://packages.msys2.org/base/mingw-w64-python-pygame).
And looks like it's will be first package in distro that use [qbs](https://packages.msys2.org/base/mingw-w64-qbs) for make.


Biswa96:
Should we wait for upstream to add that patch? Some CI builds failed in that upstream pull request.

podsvirov:
No - we don't have to wait.

It seems to me that the errors in their CI are not related to my changes.

podsvirov:
The patch is upstream and may be removed in the next package update when a new version of `tiled` is released.

PR #15465, opened 2023-02-03
qgis: Fix building on clang environments.

MehdiChinoune:


mingw-w64-qgis/001-fix-building-with-mingw-w64-clang.patch
```diff
@@ -0,0 +1,118 @@
...
+--- a/src/core/textrenderer/qgsfontmanager.cpp
++++ b/src/core/textrenderer/qgsfontmanager.cpp
+@@ -921,8 +921,8 @@
+ 
+   auto cleanFontFamily = []( const QString & family ) -> QString
+   {
+-    const thread_local QRegularExpression charsToRemove( QStringLiteral( "[^a-z]" ) );
+-    const thread_local QRegularExpression styleNames( QStringLiteral( "(?:normal|regular|light|bold|black|demi|italic|oblique|medium|thin)" ) );
++    const QRegularExpression charsToRemove( QStringLiteral( "[^a-z]" ) );
++    const QRegularExpression styleNames( QStringLiteral( "(?:normal|regular|light|bold|black|demi|italic|oblique|medium|thin)" ) );
```
lazka, line 115: I don't know much C++. What does removing thread_local everywhere change?

MehdiChinoune: According to Clang a function (class member) couldn't be declared thread and dllexported at the same time.

lazka: I see, thanks. So something like https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2492?view=msvc-170&viewFallbackFrom=vs-2019 I guess

MehdiChinoune:
I was waiting for #15466 to be fixed first to merge this, but not a big deal.

lazka:
oops, sorry

Issue #15469, opened 2023-02-04
[binutils] 


DaGiHu:
### Description / Steps to reproduce the issue

Error occurs building a rather complex application linking to ffmpeg, cef, sdl2, ssp libraries.
This error happened overnight, just after updating packages with pacman. I update Msys2 on a almost daily basis.

The code is exactly the same that linked correctly a few days ago, before the pacman update.

### Expected behavior

Linkage expected to finish and .exe being generated.

### Actual behavior

This errors appears and build stops:
C:/Msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: internal error: aborting at ../../binutils-2.40/ld/ldlang.c:527 in compare_section
C:/Msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: please report this bug
collect2.exe: error: ld returned 1 exit status

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Biswa96:
The exact steps to reproduce the issue would be helpful.

DaGiHu:
I'm sorry being so concise. I'm building the application using a IDE. This is its output:

Despite it reports 0 errors, the exe file is not generated.
Thank you.

C:/Msys64/mingw64/bin/mingw32-make.exe -j8 SHELL=cmd.exe -e -f  Makefile
"----------Building project:[ MyApp - Debug ]----------"
mingw32-make[1]: Entering directory 'C:/_DEV_PRJ/dev_cl/MyApp'
[...]

PR #15471, opened 2023-02-05
Revert "binutils: update to 2.40"

rouault:
This reverts commit 594c064fd4150f4fad50e1ec19f86af58e5357fd.

This breaks building a number of applications. Cf https://github.com/msys2/MINGW-packages/issues/15469


rouault:
CC @MehdiChinoune

MehdiChinoune:
I don't think that failing to link against MSVC lib is an acceptable reason to revert binutils to 2.39.

PR #15472, opened 2023-02-05
scummvm: Enable aarch64 builds

hmartinez82:
image


mingw-w64-scummvm/PKGBUILD
```diff
@@ -44,6 +44,7 @@ build() {
     --prefix=${MINGW_PREFIX} \
     --enable-release \
     --enable-optimizations \
+    $([[ ${MSYSTEM_CARCH} == aarch64 ]] || echo "--disable-nasm") \
```
mmuetzel, line 47: IIUC, this disables nasm everywhere but on aarch64.
Looking at the rule for the dependencies, you likely wanted to have that the other way round (disable only on aarch64).

hmartinez82: Yes. My mistake. I'll fix it. It should be `&&`

hmartinez82:
Yes. My mistake
I'll fix it :)

mingw-w64-scummvm/PKGBUILD
hmartinez82: Done

mmuetzel: Since this seems to have built for you with the incorrect condition for this flag, do we need to add it at all?

hmartinez82: @mmuetzel I'm fine with either. I just want to confirm that this will always be the case. I thought that being explicitly about "not building with this feature" was the preferred option.

mmuetzel: I'm not sure what we'd like to express here. Is it:
* Don't use `nasm` even if that package were available in the CLANGARM64 environment?
* Let the configure checks of the package determine whether it can use `nasm`?

In the former case, making that switch explicit would feel like the way to go. In the latter case, removing the switch would be better imho.

hmartinez82: I see, I'll remove it

mingw-w64-scummvm/PKGBUILD
```diff
@@ -4,10 +4,10 @@ _realname=scummvm
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.6.1
-pkgrel=1
+pkgrel=2
```
mmuetzel, line 7: No big deal: But since nothing changes for the existing packages, it is probably not strictly necessary to bump the pkgrel.

PR #15478, opened 2023-02-06
breeze-icons: Fix missing icons

Biswa96:


Biswa96:
@techie-guy 🛎️ 

Good news. Here are some differences before and after this change.

* The upper left column in Kate



* The 'Help' tab in menu bar



As you can see some missing icons came back.

Bad news. Dark mode and light mode icons do not change dynamically. But that issue was present before this change. Before changing to dark mode, run this command

```
cp $MINGW_PREFIX/bin/data/icons/breeze-dark/breeze-icons-dark.rcc $MINGW_PREFIX/bin/data/icontheme.rcc
```

The icontheme.rcc file is copied from light theme in the packaging step. The only missing part is somehow combine icons between breeze-icons.rcc and breeze-icons-dark.rcc. Maybe another day 😌

techie-guy:
Thanks:)

PR #15514, opened 2023-02-06
[RFC] CI: remove saneman

lazka:
I've never found it particular helpful and usually just ignored it.

One could argue that it could be improved to do what we want, but I personally don't want to write that in bash.


lazka:
No one objected on discord.

I'll try to come up with a plan for a new linter some time.

PR #15532, opened 2023-02-07
portsmf: move to new git upstream

lazka:
set version to 0.1 to match the pkgconf file


Biswa96:
Just curious. Is there any relationship between the github and sourceforge projects - like same project or moved to github or different maintainers? I did not find any official announcement or documentation.

lazka:
But good point, I should have made that clearer in the commit.

PR #15540, opened 2023-02-07
binutils: try to work around msvc lib linking regression

lazka:
See #15469


Radagan:
Thank you! I hit this exact error today, was headed to report it as requested, only to find it already fixed.

PR #15576, opened 2023-02-09
python-hatchling: update to 1.13.0

MehdiChinoune:


Rx-ReligionX:
im getting hundreds of emails i did not subscribe to, each from a different
thread. Do you know why, or how to stop it? (yours included amongst them)
I'd ppreciate any help, as I'm a bit tech illiterate.

lazka:
@Rx-ReligionX in the top right corner there is a "Watch" button, maybe you set it by accident.

PR #15629, opened 2023-02-11
wxwidgets3.2: Update to version 3.2.2

stahta01:
Edit: filezilla built okay and it opened and shutdown without error.  Only tested CLANG64 because I needed to verify the removing of patch from wxWidgets to fix CLAN64 rc building was no longer needed.

Tim S.


lazka:
rebuilds in 55455ab

PR #15634, opened 2023-02-11
qdia: New package

podsvirov:


MehdiChinoune:
remaining i986 qt6 packages will be dropped with the next release, please don't add new i686 packages with qt6 dependency.

PR #15635, opened 2023-02-11
qdia: build against Qt5 on i686

MehdiChinoune:


lazka:
thanks!

(I'm also fine with dropping 32bit in general for niche applications)

PR #15641, opened 2023-02-12
wireshark: Enable clangarm64

Biswa96:


Biswa96:
@jeremyd2019 Would you like to check if the packages can be built please?

Biswa96:
Let see.

Issue #10434, opened 2021-12-28
GDB TUI broken


GitMensch:
I have no clue since when this occurs or if this was never working, I guess so, because it is already referenced in #1744...

`gdb ./myconsole.exe -x start` --> works fine
`gdb ./myconsole.exe -tui -ex start` --> results in "During startup program exited with code 0xc0000142."

As also analyzed in the referenced issue by @dimztimz this is related to the console handles. The only thing that looks bad is the following, seeing this issue persists over 5 years later:

> a new feature that is buggy in a very specific case and that will be sorted out in the future

Maybe the future is now :-)

The work around, as mentioned by him still works:

     gdb ./myconsole.exe -tui -ex 'set new-console on' -ex start

If we can't get that to work reliably I suggest we add a patch to always enable new-console when in tui-mode.

Note: if we do it the other way around it _does_ work:

     gdb ./myconsole.exe -ex start -ex 'tui enable'
     
(in this case you "obviously" need to temporarily disable the tui to see the display because that doesn't happen on the same console)

ssbssa:
If ncurses is updated to [the latest patch](https://lists.gnu.org/archive/html/bug-ncurses/2023-02/msg00005.html), then this bug should be fixed (the `CreateConsoleScreenBuffer()` call was the problem).

Biswa96:
Would you like to test the packages from GitHub Actions artifacts of that pull request? The test packages can be downloaded from here https://github.com/msys2/MINGW-packages/actions/runs/4156876469

PR #15642, opened 2023-02-12
ncurses: update to 6.4.20230211

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/10434

### DIFF
--- a/mingw-w64-ncurses/PKGBUILD
+++ b/mingw-w64-ncurses/PKGBUILD
@@ -3,7 +3,9 @@
 _realname=ncurses
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=6.4
+_base_ver=6.4
+_date_rev=20230211
+pkgver=${_base_ver}.${_date_rev}
 pkgrel=1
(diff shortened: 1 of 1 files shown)


lhmouse:
Is using tarballs from `https://invisible-mirror.net/archives/ncurses/current/` adhoc or permanent? For example, if ncurses 6.5 is released someday, do we still fetch tarballs there?

Biswa96:
> if ncurses 6.5 is released someday, do we still fetch tarballs there?

It depends. If the mirror link is official and provides signed tarball it would not be changed. If I recall correctly, there was some issue with that link in GitHub actions. So, in that situation source link will be changed.

Biswa96:
Would you like to test if this change fixes the GDB TUI issue?

lhmouse:
I built and installed that tarball locally (with the same SHA-256, albeit not with the PKGBUILD that you have proposed in this PR), and it did fix the issue for me: https://lists.gnu.org/archive/html/bug-ncurses/2023-02/msg00006.html

The issue can be reproduced with https://lists.gnu.org/archive/html/bug-ncurses/2023-02/msg00003.html.

Biswa96:
Not sure what happened. The version number was correct though.

```
$ vercmp 6.4.20230211-1 6.4-1
1
```

PR #15643, opened 2023-02-12
gcc: Install libssp.dll.a to lib folder

stahta01:
When working Julia master branch, I saw it wanted "libssp.dll.a". So, I added it to GCC package which has the shared libssp DLL from GCC libs split package. The static libssp was from the CRT package; so, I did not add them to this package.

Tim S.

### DIFF
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -24,7 +24,7 @@ pkgver=12.2.0
 #_majorver=${pkgver:0:1}
 #_sourcedir=${_realname}-${_majorver}-${_snapshot}
 _sourcedir=${_realname}-${pkgver}
-pkgrel=10
+pkgrel=11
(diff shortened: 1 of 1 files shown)


Biswa96:
Wasn't it removed in this commit https://github.com/msys2/MINGW-packages/commit/9288f278fe41c08c58f0a26da38a8863d2943321 ?

lazka:
Yeah, the removal was on purpose, and the DLL will also go away eventually.

stahta01:
Is it okay to add the import lib for the DLL? What you posted strongly implied no; but, did not outright say no.

Tim S.

lazka:
It is a no. It would make things link against libssp again which we want to remove.

stahta01:
> It is a no. It would make things link against libssp again which we want to remove.
> 
> Afaik other platforms also don't have a libssp, so hopefully there is another way to fix this in julia?

I will try using the static libssp from the CRT package.

Tim S.

PR #15664, opened 2023-02-13
mapserver: New package

podsvirov:


podsvirov:
Look like it's expected. For more info [see](https://github.com/MapServer/MapServer/tree/rel-8-0-0/mapscript/python#testing).

PR #15679, opened 2023-02-14
kicad: depend on wxWidgets 3.2

mmuetzel:
This PR updates the wxWidgets dependency of Kicad to version 3.2.
To avoid conflicts, wxPython also needs to depend on wxWidgets 3.2 for this to work. The alpha tarball from which wxPython was built the last time is no longer available. So, I updated to the latest stable release wxPython (Phoenix) 4.2.0.

IIUC, these are the last dependers on wxWidgets 3.1. Afaict, that was (only) a development version that we needed intermediately to build some package (Kicad?). @MehdiChinoune already moved all other dependent packages to wxWidgets 3.2 afaict. `wxWidgets3.1` can probably be removed if this PR is merged. (I can also remove it as part of this PR if you prefer.)

`waf` (that is used as a build tool for wxPython) has troubles calling the `wx-config` shell script when using a mingw Python. It works when using the MSYS2 Python (but needs some hints to correctly detect the C and C++ compilers). Additionally, it gets stuck when waiting for a pipe to close. I added a patch with a dirty workaround to just ignore that after a timeout.
I tried to update `waf` to the latest version to check if those issues are resolved. But it turned out to still be the same. I kept the waf update in this PR anyway (although it is not strictly needed - but the patch with the workaround is needed anyway).

### DIFF
--- a/mingw-w64-kicad/PKGBUILD
+++ b/mingw-w64-kicad/PKGBUILD
@@ -5,11 +5,11 @@
 # updating this build rule
 
 _realname=kicad
-_wx_basever=3.1
+_wx_basever=3.2
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=6.0.11
-pkgrel=1
+pkgrel=2
 pkgdesc="Software for the creation of electronic schematic diagrams and printed circuit board artwork (mingw-w64)"
 arch=(any)
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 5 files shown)


mmuetzel:
@stahta01: Could you please take a look on whether these changes look reasonable to you? Would it be ok to remove wxWidgets3.1 after this? Or should we keep that package for some time?

stahta01:
> @stahta01: Could you please take a look on whether these changes look reasonable to you? Would it be ok to remove wxWidgets3.1 after this? Or should we keep that package for some time?

It looked okay to me; but, I thought some complained last time I tried to use MSys2 python.

Tim S.

mmuetzel:
> I thought some complained last time I tried to use MSys2 python.

I didn't need to add MSYS2 Python as a build dependency. It was already installed as a dependency (for waf) before this PR. I guess I only needed to make it use that Python explicitly. (It might have been used implicitly before. Maybe some shebang changed?)

mmuetzel:
I'm going to merge this PR as it currently is.
If someone would like to have a look into how to fix `waf` properly, that can also be done in a follow-up PR.

lazka:
we likely need https://gitlab.com/kicad/code/kicad/-/issues/13887

mmuetzel:
Thanks. I'll try that locally.

PR #15694, opened 2023-02-15
clang: stop using access time for file comparison

### LINKED ISSUE #15695: Clang: Sporadic linker errors happen when a sysroot is used

If you use clang as a cross compiler and give it a new sysroot via the --sysroot this error can occur sporadically. 

We have a very large CMake project. When we build the project in CI, we sometimes get build errors. However, this is highly sporadic and occurs in about 1% of linker actions.

Locally you can reproduce it if you update the sysroot folder access time during the build. (`while true; do touch -a <sysroot> ; done`) 

The comparison of access times was introduced with http://reviews.llvm.org/D18456. I think this happened by mistake.

### Expected behavior

lld should not have sporadic errors and should link independent of the access time.

### Actual behavior

lld sporadically fails to resolve GNU linker scripts.

### Verification

### Windows Version

MSYS_NT-10.0-19044

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

https://github.com/msys2/MINGW-packages/pull/15694

rvogg:
To compare files on windows the access times is used.
This leads to the fact that the lld sets the variable isUnderSysroot to false by mistake with strongly
parallelized compiles. (Since the access time of the sysroot folder can be updated between two stats).
    
This prevents the ScriptParser from correctly resolving the path from a linker script if sysroot is set 
and sporadic linker errors occur.

Fixes #15695


Biswa96:
Would you like to explain the problem in a new issue?

rvogg:
I have created an issue:  https://github.com/msys2/MINGW-packages/issues/15695

rvogg:
What info is still missing, is there anything else I can add? 
How can I help?

mstorsjo:
The patch is on its way upstream in https://reviews.llvm.org/D144172.

PR #15726, opened 2023-02-16
libheif: update to 1.15.1

kmilos:


Biswa96:
The $MIGNW_PREFIX/share/mime/ directory is removed after this update. Is that expected?

kmilos:
Yes. These are already part of shared-mime-info for a while now.

PR #15734, opened 2023-02-17
add package for Lua5.3

neojaw:
For a cross operating system project we needed a Lua 5.3 package (older Ubuntus do not ship Lua 5.4 yet). I don't know what the policy on version specific packages is but I thought it might be useful to some people. If you don't want to integrate the package, I can just close the PR and leave it here for future users to find.

Inspiration taken from the `mingw-w64-lua` and `mingw-w64-lua51` packages.

### DIFF
--- a/mingw-w64-lua53/LICENSE
+++ b/mingw-w64-lua53/LICENSE
@@ -1,0 +1,30 @@
+License
+
+Lua is free software distributed under the terms of the MIT license
+reproduced below; it may be used for any purpose, including commercial
+purposes, at absolutely no cost without having to ask us. The only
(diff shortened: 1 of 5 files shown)


mmuetzel:
The files in `${MINGW_PREFIX}/share/doc/lua` conflict with the (unversioned) `lua` package.
Is that an issue?

neojaw:
Oops, missed that. Should be fixed now.

mmuetzel:
I also don't know what the policy is on versioned packages that don't have any directly depending packages in MSYS2.
Other than that, LGTM.

PR #15762, opened 2023-02-19
kicad: support subrelease field in wxWidgets version number

mmuetzel:
The upstream patch doesn't apply directly because some folder names changed.


mmuetzel:
CI moved past wxWidgets detection in all five build environments.
Assuming the rest will now build successfully.

PR #15765, opened 2023-02-20
[libbs2b] Add support for clangarm64

stephenjust:
Needed to run autoreconf for aarch64 to build.

### DIFF
--- a/mingw-w64-libbs2b/PKGBUILD
+++ b/mingw-w64-libbs2b/PKGBUILD
@@ -4,10 +4,10 @@ _realname=libbs2b
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=3.1.0
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


mingw-w64-libbs2b/PKGBUILD
```diff
@@ -17,6 +17,10 @@ sha256sums=('6aaafd81aae3898ee40148dd1349aab348db9bfae9767d0e66e0b07ddd4b2528')
 
 prepare() {
   cd "$srcdir"/${_realname}-${pkgver}
+
+  sed -i 's/dist-lzma/dist-xz/g' configure.ac
```
Biswa96, line 21: Would you like to explain why this change is required?

stephenjust: Latest autoconf deprecated dist-lzma. I just used the same solution from upstream Arch pkgbuild.

Biswa96: ok, seems good.

PR #15776, opened 2023-02-20
spooles: add new patch, enable clang

3rav:
spooles: enable clang, new patch file from archlinux added because error occurs (clang):

```
transform.c:294:23: error: incompatible pointer to integer conversion passing 'void *' to
      parameter of type 'int' [-Wint-conversion]
temp = IVinit(nfront, NULL) ;
                      ^~~~
```

Patch from: https://aur.archlinux.org/cgit/aur.git/tree/spooles-1.patch?h=spooles

### DIFF
--- a/mingw-w64-spooles/PKGBUILD
+++ b/mingw-w64-spooles/PKGBUILD
@@ -18,6 +18,7 @@ url="https://www.netlib.org/linalg/spooles"
 license=('custom')
 source=("https://www.netlib.org/linalg/${_realname}/${_realname}.${pkgver}.tgz"
         spooles.patch
+        # Patch from (spooles-1): https://aur.archlinux.org/cgit/aur.git/tree/spooles-1.patch?h=spooles
         spooles-1.patch
         LICENSE)
 sha256sums=('a84559a0e987a1e423055ef4fdf3035d55b65bbe4bf915efaa1a35bef7f8c5dd'
--- a/mingw-w64-spooles/PKGBUILD
+++ b/mingw-w64-spooles/PKGBUILD
@@ -6,19 +6,23 @@ _realname=spooles
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.2
-pkgrel=4
+pkgrel=5
 pkgdesc="SParse Object Oriented Linear Equations Solver (mingw-w64)"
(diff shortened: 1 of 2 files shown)


mingw-w64-spooles/PKGBUILD
```diff
@@ -53,5 +58,7 @@ package() {
   # Fix permissions
   cd ${pkgdir}${MINGW_PREFIX}/include/spooles
   chmod -R oug+r *
+  cd ${pkgdir}${MINGW_PREFIX}/lib
+  chmod oug+r *
```
mmuetzel, line 62: IIUC, the files are installed into the `lib` folder using `install -Dm644`. Do we really need to change the permissions after that?

Talking of that: Would it be better to install the files to `include/spooles` using `install -Dm644` (instead of `cp` + `chmod`), too?

mingw-w64-spooles/PKGBUILD
```diff
@@ -28,10 +32,11 @@ prepare() {
   cd "${srcdir}/${_realname}-${pkgver}"
   tar -xzf "${srcdir}/${_realname}.${pkgver}.tgz"
   patch -Np1 -i "${srcdir}"/spooles.patch
+  patch -Np1 -i "${srcdir}"/spooles-1.patch
 }
 
 build() {
-  make lib -C "${srcdir}/${_realname}-${pkgver}"
+  make CC=${CC} CFLAGS="$CFLAGS -Wno-error=format-security" lib -C "${srcdir}/${_realname}-${pkgver}"
```
mmuetzel, line 39: Do we need to set `CC=${CC}`? Isn't that a no-op?

mingw-w64-spooles/PKGBUILD
3rav: My bad

MehdiChinoune:
Could you please better commit messages, Those "Update package"/"Add Package" don't any of what changes were done when viewing repo history.

mmuetzel:
Could you please add a comment in the PKGBUILD file that references the source where you got that patch from?

mmuetzel:
I agree with @MehdiChinoune that the commit messages in this PR are not really helpful.
Personally, I'd prefer if they summarize what was the reason for the change. (Just a few words are sufficient in many cases imho.)

In this particular case, you could squash the commits into one with a commit message like "spooles: enable clang".

PR #15777, opened 2023-02-20
gdb: update to 13.1

orgads:


Biswa96:
There is a similar PR with this topic.

orgads:
Yes, I just found it now :)

Closing.

PR #15782, opened 2023-02-20
kicad: update to 7.0.0

MehdiChinoune:
kicad fails to build on i686, anyone using them is welcome to fix them.


lazka:
Looking at https://downloads.kicad.org/kicad/windows/explore/stable it looks like upstream no longer provides 32bit builds. So I think it's fine do too.

PR #15799, opened 2023-02-21
gdb: fix multiarch package

orgads:
The real executable is in gdb/.libs/gdb.exe, gdb/gdb.exe executes it. Install it properly and copy from the install target.

### DIFF
--- a/mingw-w64-gdb/PKGBUILD
+++ b/mingw-w64-gdb/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-multiarch")
 pkgver=13.1
-pkgrel=1
+pkgrel=2
 pkgdesc="GNU Debugger (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


orgads:
cc @MehdiChinoune @lazka @GitMensch

lazka:
thanks

mingw-w64-gdb/PKGBUILD
```diff
@@ -138,10 +138,15 @@ package_gdb-multiarch() {
...
+  cd ${builddir}
+  make DESTDIR=${builddir}/inst install-gdb install-gdbserver
+
   destdir=${pkgdir}${MINGW_PREFIX}/bin
+  bindir=${builddir}/inst${MINGW_PREFIX}/bin
   mkdir -p $destdir
-  strip -o ${destdir}/gdb-multiarch.exe ${srcdir}/build-${MSYSTEM}-multiarch/gdb/gdb.exe
-  strip -o ${destdir}/gdbserver-multiarch.exe ${srcdir}/build-${MSYSTEM}-multiarch/gdbserver/gdbserver.exe
+  cp ${bindir}/gdb.exe ${destdir}/gdb-multiarch.exe
+  cp ${bindir}/gdbserver.exe ${destdir}/gdbserver-multiarch.exe
```
mmuetzel, line 149: Do we care about file permissions? Maybe, replace `cp` with `install -Dm755`?

GitMensch: agreed

GitMensch (approved):
The general approach is the right one: don't copy fro build dir but from installation directory and DESTDIR is quite useful here.
Use either `install` or - in this case preferably - `cp -p` (as this copies the permissions as the install target has created them, which is what we use in other places).

mingw-w64-gdb/PKGBUILD
orgads: Done

GitMensch:
Thank you, I'm looking forward for GDB 13 in MSYS2; testing the resulting package shows:

* starts fine
* similar `--configuration` as GDB 12
* more architectures supported according to `set architecture [TAB]`
* remote debugging works to different targets

So - all fine!

Note: I'm not sure if highlighting via python-pygments works "as expected" (or at least as it did with previous versions), it would be good if someone else could test that.

PR #15823, opened 2023-02-22
opencascade, ngspice: Add aarch64 builds

hmartinez82:
The two remaining deps for KiCAD
image


hmartinez82:
Who can take a look? There were no changes to the source code. I just tried to build locally in te UCRT env and it builds correctly. There's no way that such a file (math.h) would be missing without breaking tons of builds.

mmuetzel:
That looks like the hickup that sometimes happens on the `UCRT64` CI runners (ISTR something about `-isystem` vs. `-I` in compiler arguments...).
If it is that, it might be some kind of race condition that only happens intermittently. I'll restart the test for that runner...

PR #15848, opened 2023-02-23
vulkan-validation-layers: enable use of robin_hood hashing

filnet:


filnet:
I don't understand why UCRT64 fails.

mmuetzel:
> I don't understand why UCRT64 fails.

Restarted that runner.

PR #15867, opened 2023-02-24
x264: drop ffmpeg support

iBobbyTS:
x264 doesn't require ffmpeg.

### DIFF
--- a/mingw-w64-x264/PKGBUILD
+++ b/mingw-w64-x264/PKGBUILD
@@ -12,7 +12,6 @@ url="https://www.videolan.org/developers/x264.html"
 license=("custom")
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-autotools"
-             "${MINGW_PACKAGE_PREFIX}-ffmpeg"
              "${MINGW_PACKAGE_PREFIX}-l-smash"
              $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-ffms2")
               $( [[ "${CARCH}" != "i686" \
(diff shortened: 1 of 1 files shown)


mingw-w64-x264/PKGBUILD
```diff
@@ -74,7 +73,6 @@ package_x264() {
   # Note: libx264 is stricly not needed, but existing users of the x264 package
   # before the split expect the library to be included
```
lazka, line 74: imo this would break too many things

iBobbyTS:
Adding ffmpeg while user only need x264 will cause too many extra libraries and potential problems. Is there a way for users to disable ffmpeg installation when installing x264 only on their case?

lazka:
(ignore my previous comment, I mixed things up)

x264 currently uses ffmpeg, so the dependency is correct.

There is currently no way besides rebuilding.

lazka:
Some examples in other distors:

* Debian uses ffmpeg: https://packages.debian.org/bookworm/x264
* Arch doesn't: https://archlinux.org/packages/extra/x86_64/x264/

I personally don't know enough about users of x264 to say what feature set of x264 they expect.

filnet:
@iBobbyTS commit log should contain project name. Take a quick look at the commit history : https://github.com/msys2/MINGW-packages/commits/master

Actually, @lazka edited the PR title to match what the commit log should be.

lazka:
If we can get any good reasons why installing ffmpeg is indeed a problem then we can talk about this. If not I'll close this.

iBobbyTS:
When I build ffmpeg with a third party filter, I would success with libaom, libx265, but if I install x264 with pacman, which comes with a huge number of libraries, it would build fail.

lazka:
So ffmepg fails to build for you is ffmpeg is installed already? How does it fail?

lazka:
as I'm still not sure why exactly this would be needed, closing. If you can provide any more info feel free to create a new issue.

PR #15901, opened 2023-02-26
[new-package] krita: Add build recipe for ucrt64, mingw64, clang64 and clangarm64

hmartinez82:
Depends on https://github.com/msys2/MINGW-packages/pull/15911

image

### DIFF
--- a/mingw-w64-krita/PKGBUILD
+++ b/mingw-w64-krita/PKGBUILD
@@ -25,6 +25,7 @@ makedepends=(
   "${MINGW_PACKAGE_PREFIX}-xsimd"
 )
 depends=(
+  "${MINGW_PACKAGE_PREFIX}-angleproject"
   "${MINGW_PACKAGE_PREFIX}-exiv2"
   "${MINGW_PACKAGE_PREFIX}-fftw"
   "${MINGW_PACKAGE_PREFIX}-gsl"
@@ -68,6 +69,8 @@ prepare() {
   patch -Np1 -i "${srcdir}/001-mingw-interface-conflict.patch"
   patch -Np1 -i "${srcdir}/002-exiv2-missing-wpath.patch"
   patch -Np1 -i "${srcdir}/003-xsimd-10.patch"
+
+  # Workaround for https://bugreports.qt.io/browse/QTBUG-111527
   patch -Np1 -i "${srcdir}/004-qt5-qguiapplication.patch"
 }
 
--- a/mingw-w64-krita/PKGBUILD
+++ b/mingw-w64-krita/PKGBUILD
@@ -8,7 +8,7 @@ pkgdesc="A professional FREE and open source painting program"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://krita.org/"
-license=("spdx:GPL3")
+license=("spdx:GPL-3.0-only")
 makedepends=(
   "${MINGW_PACKAGE_PREFIX}-cc"
   "${MINGW_PACKAGE_PREFIX}-boost"
--- a/mingw-w64-krita/PKGBUILD
+++ b/mingw-w64-krita/PKGBUILD
@@ -1,0 +1,97 @@
+
+_realname=krita
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=5.1.5
+pkgrel=1
+pkgdesc="A professional FREE and open source painting program"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url="https://krita.org/"
+license=("spdx:GPL3")
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-cc"
+  "${MINGW_PACKAGE_PREFIX}-boost"
+  "${MINGW_PACKAGE_PREFIX}-eigen3"
+  "${MINGW_PACKAGE_PREFIX}-extra-cmake-modules"
+  "${MINGW_PACKAGE_PREFIX}-kdoctools-qt5"
+  "${MINGW_PACKAGE_PREFIX}-libheif"
+  "${MINGW_PACKAGE_PREFIX}-libjxl"
+  "${MINGW_PACKAGE_PREFIX}-libmypaint"
+  "${MINGW_PACKAGE_PREFIX}-poppler-qt5"
+  "${MINGW_PACKAGE_PREFIX}-python-pyqt5"
+  "${MINGW_PACKAGE_PREFIX}-qt5-tools"
+  "${MINGW_PACKAGE_PREFIX}-sip"
+  "${MINGW_PACKAGE_PREFIX}-xsimd"
+)
+depends=(
+  "${MINGW_PACKAGE_PREFIX}-exiv2"
+  "${MINGW_PACKAGE_PREFIX}-fftw"
+  "${MINGW_PACKAGE_PREFIX}-gsl"
+  "${MINGW_PACKAGE_PREFIX}-icoutils"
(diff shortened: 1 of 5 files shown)


Biswa96:
The $MINGW_PREFIX/installer/, $MINGW_PREFIX/krita-msix/ and $MINGW_PREFIX/MakeinstallerNsis.cmake should not exist in the package. The following diff may help, I am not sure though.

```diff
--- CMakeLists.txt
+++ CMakeLists.txt
@@ -1205,7 +1205,7 @@
 check_function_exists(powf HAVE_POWF)
 configure_file(config-powf.h.cmake ${CMAKE_CURRENT_BINARY_DIR}/config-powf.h)
 
-if(WIN32)
+if(MSVC)
     include(${CMAKE_CURRENT_LIST_DIR}/packaging/windows/installer/ConfigureInstallerNsis.cmake)
     include(${CMAKE_CURRENT_LIST_DIR}/packaging/windows/msix/ConfigureMsixFiles.cmake)
 endif()
```

@alvinhochun may be interested in this.

alvinhochun:
If you ask me, I would really hope that MSYS2 will not carry a Krita package. For our official releases, we maintain a lot of patches in our dependencies, some of which are required for certain core features (e.g. HDR support). I also don't want the team to start getting bug reports for a MSYS2 build of Krita. We already have too many bug reports for the official builds; a new build variant will make things more confusing and increases the workload on the team.

Biswa96:
I understand and agree with your concern. Let wait for @hmartinez82 's comment.

amyspark:
> The following diff may help, I am not sure though.

The diff is actually incorrect, those files are needed if you want to create the NSIS installer (MSVC or MINGW). In this case, we remove them in the packaging script, that step could be implemented in the recipe.

> I also don't want the team to start getting bug reports for a MSYS2 build of Krita.

We've wanted to support Surfaces for a bit, but lack the hardware and dev power; and having a build that tracks rolling release Clang would be 👍 👍 👍 in my opinion.

mingw-w64-krita/PKGBUILD
```diff
@@ -0,0 +1,100 @@
...
+_realname=krita
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=5.1.5
+pkgrel=1
+pkgdesc="A professional FREE and open source painting program"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url="https://krita.org/"
+license=("spdx:GPL3")
```
mmuetzel, line 11: Afaict, that's not a valid spdx identifier.
For a list of valid identifiers see, e.g., https://spdx.org/licenses/

mingw-w64-krita/PKGBUILD
```diff
@@ -0,0 +1,100 @@
...
+  [[ -d "build-${MSYSTEM}" ]] && rm -rf "build-${MSYSTEM}"
+  mkdir -p "${srcdir}/build-${MSYSTEM}"
+  cd "${srcdir}/build-${MSYSTEM}"
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  ${MINGW_PREFIX}/bin/cmake \
+    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+    -GNinja \
+    -DCMAKE_BUILD_TYPE="Release" \
+    -DBUILD_TESTING="OFF" \
```
mmuetzel, line 87: Should this really be a string `"OFF"`? Or do you mean a logical value `OFF`?
In the latter case:
``` suggestion
    -DBUILD_TESTING=OFF \
```

Biswa96:
> The diff is actually incorrect, those files are needed

My apology for the confusion. I did not mean for upstream. To disable packaging, a separate cmake option may be useful.

alvinhochun:
> We've wanted to support Surfaces for a bit, but lack the hardware and dev power; and having a build that tracks rolling release Clang would be 👍 👍 👍 in my opinion.

Well, the x64 build of Krita should already work on ARM64 Surfaces so support shouldn't be an issue (but if it doesn't then we should fix that). It's more about getting native performance. I'm not opposed to keeping the source compatible with Windows ARM64, but for us to really make an official release we need feature parity, that means fully patched deps for ANGLE and HDR render support. Oh and also passing tests. And I'd say we should make our whole build compatible with cross compilation, at least from x64 Windows. But I digress.

Having a MSYS2 ARM64 build is a nice addition (btw try building master instead of 5.1.5), but shipping an easily installable package is a bit worrying. You know how bug reports sometimes go: "I'm using the distro package." "Try the AppImage instead." "Thanks it works!" *That said*, this is probably not a real concern considering normal users won't even think of trying to install Krita from MSYS2 (unlike on Linux where users often try the package manager first). But I would hope to make it clear somehow that only the official Windows release is supported.

mingw-w64-krita/PKGBUILD
alvinhochun: If the quotes are handled by the shell I don't think it matters. All CMake variables are strings anyway, with certain values being treated as true or false inside an if condition.

mmuetzel:
I'd guess that many of the binary packages that are distributed by MSYS2 are not directly supported by upstream. Is there a significant difference for Krita that makes it different from the other packages?

mingw-w64-krita/PKGBUILD
```diff
@@ -0,0 +1,100 @@
...
+_realname=krita
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=5.1.5
+pkgrel=1
+pkgdesc="A professional FREE and open source painting program"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url="https://krita.org/"
+license=("spdx:GPL-3.0")
```
mmuetzel, line 11: That spdx identifier is deprecated.
``` suggestion
license=("spdx:GPL-3.0-only")
```
Or is it `GPL-3.0-or-later`?

mingw-w64-krita/PKGBUILD
hmartinez82: That was the behavior I saw indeed.

amyspark:
> @alvinhochun @Biswa96 should I enable only clangarm64 builds of Krita then?

Feel free to enable the rest, but clangarm64 would definitely be a welcome addition.

>   Angle: It has been built for Windows on ARM for years now. Chrome, Electron, Qt 5 all use Angle and it correctly uses the D3D11 as the backend.

There are additional patches required for HDR to work, see 3rdparty/ext_googleangle for the list. There are also Krita-specific patches that have not been upstreamed which are kept in https://invent.kde.org/szaman/qt5.git .

hmartinez82:
First time depending on e brand new package. Lesson learned: https://github.com/msys2/MINGW-packages/pull/15911

hmartinez82:
@Biswa96 Patch applied

nekoteai:
@hmartinez82 I install it using msys2 and come up with a no response crash when starting program.please check and fix

Biswa96:
How did you install it? krita package is working fine in my system.

```
$ krita --version
krita 5.1.5 (git ce6e5e8)

$ which krita
/ucrt64/bin/krita
```

nekoteai:
@Biswa96 oops I forget to say that its arm64 build.

jeremyd2019:
On a VM, I get no output on --version.  With no arguments, I got a splash screen but then exits.  On the dev kit 2023 I still get no output with --version, but I do get a window and a message about no resource bundles being enabled.

Maybe it's the opengl compatibility store app thing?  The dev kit 2023 came with that preinstalled.  Just a random guess though

hmartinez82:
The opengl compat app is required
The bundled resource message. I got that too. I had to download one from
the Krita website

PR #15902, opened 2023-02-26
gdb: change multiarch dependency to match gdb exact version

orgads:
It uses artifacts from gdb package, so they must match.

### DIFF
--- a/mingw-w64-gdb/PKGBUILD
+++ b/mingw-w64-gdb/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-multiarch")
 pkgver=13.1
-pkgrel=2
+pkgrel=3
 pkgdesc="GNU Debugger (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


mingw-w64-gdb/PKGBUILD
```diff
@@ -136,7 +136,7 @@ package_gdb() {
 
 package_gdb-multiarch() {
   pkgdesc="GNU Debugger (supports all targets)"
-  depends=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+  depends=("${MINGW_PACKAGE_PREFIX}-${_realname}=${pkgver}-${pkgrel}")
```
orgads, line 139: I'm unsure if the release is mandatory. What do you think?

LABPBS: c45c199fe7e3eafe73012c27af92bcba6b4c21ab for reference

lazka:
Shouldn't they match anyway?

orgads:
They should. Question is - if you for example downgrade gdb-multiarch from 13.1-3 to 13.1-1, must you also downgrade gdb (and vice versa)?

lazka:
The same is true for every other package though, if you downgrade ffmpeg lots of things will break. Not saying this is wrong, though, but we don't usually care about partial up/downgrades as we don't support them.

orgads:
I was asking specifically about release-downgrade (13.1-3 -> 13.1-1). For real version change, I think there should be a restriction. I needed to test something yesterday, so I downgraded `gdb-multiarch` to 12.1, but then nothing worked, because the data content is different, and it comes from `gdb` package. For release, I think it shouldn't matter usually.

Removed `pkgrel` now.

lazka:
if it helps you and doesn't cause problems, then fair enough I guess.

orgads:
Thank you :)

PR #15903, opened 2023-02-26
pygame: use correct `pkgver`

wszqkzqk:


lazka:
Just wondering, why is this "correcr" ?

Biswa96:
correct = workaround without epoch

lazka:
Ah, ok

PR #15911, opened 2023-02-27
[new-package] kseexpr: Add package

hmartinez82:


lazka:
This only includes a static library, I wonder if that is on purpose.

hmartinez82:
Yes. I noted that too. It does not have the switches to create a shared lib. It's only statically linked where I've seen it being used

lazka:
Yeah, I see it's special cased on Windows. Maybe we can look at that later.

PR #13821, opened 2022-11-01
Update yosys and ghdl-yosys-plugin

umarcor:

PR #15919, opened 2023-02-28
yosys: Update to 0.30 and fix TCL detection

hmartinez82:
@umarcor Can you please take a look too 👀

### DIFF
--- a/mingw-w64-yosys/001-fix-tcl-flags.patch
+++ b/mingw-w64-yosys/001-fix-tcl-flags.patch
@@ -1,0 +1,33 @@
+diff -bur yosys-yosys-0.27-orig/Makefile yosys-yosys-0.27/Makefile
+--- yosys-yosys-0.27-orig/Makefile	2023-04-14 04:52:22.546510600 -0600
++++ yosys-yosys-0.27/Makefile	2023-04-14 04:52:41.159631100 -0600
+@@ -458,25 +458,12 @@
+ LDLIBS += -lz
+ endif
+ 
+-
+ ifeq ($(ENABLE_TCL),1)
+ TCL_VERSION ?= tcl$(shell bash -c "tclsh <(echo 'puts [info tclversion]')")
+-ifeq ($(OS), $(filter $(OS),FreeBSD OpenBSD NetBSD))
+-# BSDs usually use tcl8.6, but the lib is named "libtcl86"
+-TCL_INCLUDE ?= /usr/local/include/$(TCL_VERSION)
+-TCL_LIBS ?= -l$(subst .,,$(TCL_VERSION))
+-else
+-TCL_INCLUDE ?= /usr/include/$(TCL_VERSION)
+-TCL_LIBS ?= -l$(TCL_VERSION)
+-endif
+-
+-ifeq ($(CONFIG),mxe)
+-CXXFLAGS += -DYOSYS_ENABLE_TCL
+-LDLIBS += -ltcl86 -lwsock32 -lws2_32 -lnetapi32 -lz -luserenv
+-else
+-CXXFLAGS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --cflags tcl || echo -I$(TCL_INCLUDE)) -DYOSYS_ENABLE_TCL
+-LDLIBS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --libs tcl || echo $(TCL_LIBS))
+-endif
++TCL_INCLUDE ?= $(PREFIX)/include/$(TCL_VERSION)
++TCL_LIBS ?= -L$(PREFIX)/lib -ltcl
++CXXFLAGS += -I$(TCL_INCLUDE) -DYOSYS_ENABLE_TCL
++LDLIBS += $(TCL_LIBS)
+ endif
+ 
+ ifeq ($(ENABLE_GCOV),1)
--- a/mingw-w64-yosys/PKGBUILD
+++ b/mingw-w64-yosys/PKGBUILD
@@ -1,7 +1,8 @@
+
 _realname=yosys
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.12.r45
+pkgver=0.30
 pkgrel=1
 pkgdesc="A framework for RTL synthesis tools (mingw-w64)"
 arch=('any')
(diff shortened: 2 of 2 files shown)


mmuetzel:
`MINGW32` failed with errors like the following:
```
  D:/M/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/12.2.0/../../../../i686-w64-mingw32/bin/ld.exe: kernel/driver.o: in function `main':
  C:\_\B\src\yosys-yosys-0.26/kernel/driver.cc:574: undefined reference to `_imp__Tcl_EvalFile'
  D:/M/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/12.2.0/../../../../i686-w64-mingw32/bin/ld.exe: C:\_\B\src\yosys-yosys-0.26/kernel/driver.cc:575: undefined reference to `_imp__Tcl_GetStringResult'
```
Maybe an issue with calling conventions that don't match?

mmuetzel:
`UCRT64` failed with the following error:
https://github.com/msys2/MINGW-packages/actions/runs/4289118302/jobs/7471844561#step:11:2019
```
  D:/M/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find D:/a/msys64/ucrt64/lib/gcc/x86_64-w64-mingw32/12.2.0/adalib/libgnat.a: No such file or directory
  make: *** [passes/techmap/Makefile.inc:58: yosys-filterlib.exe] Error 1
```
That library should already be installed at `/ucrt64/lib/gcc/x86_64-w64-mingw32/12.2.0/adalib/libgnat.a`:
https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-gcc-ada?repo=ucrt64
Is it broken for some reason?

hmartinez82:
@mmuetzel I'm going to try to debug the MINGW32 one. And indeed. I'm seeing that hardcoded wrong path in my local UCRT64 builds too :(

umarcor:
@hmartinez82 see #13821. I will bump it. Please submit other modifications as separated PRs, since they are not related to updating the package.

hmartinez82:
@umarcor I did see that PR, but it was missing other parts in the PKGBUILD file, like installing the `COPYING` file. Do you mean these changes like that?

hmartinez82:
@umarcor Awesome. I'll close this PR and wait until you update the version to 0.26 in https://github.com/msys2/MINGW-packages/pull/13821

I'm also curious about these other dependencies in  https://archlinux.org/packages/community/x86_64/yosys/ . Are they just "nice to have". Should I add them in my next PR?

umarcor:
@hmartinez82 #13821 is updated and CI is running now.

With regard to other dependencies, I don't think graphviz or xdot should be installed by default. If users want to use those features, I believe they can install those packages without rebuilding yosys.

With regard to solvers (cvc4, yices and z3) I think those are runtime dependencies as well, for users doing formal verification (through sby). So, it seems that Arch package is building both yosys and symbiyosys at the same time.

mmuetzel:
> Installing `COPYING` is an enhancement to the recipe, unrelated to the issues with updating the version of the tools.

You could still do that in the same PR (maybe not same commit) to avoid needing to bump the pkgrel just for the addition of the license file.

hmartinez82:
@mmuetzel. I found the issue for UCRT64 https://github.com/msys2/MINGW-packages/issues/16034

hmartinez82:
@mmuetzel You were right, the symbols are a mess:
objdump -t /mingw32/lib/libtcl.dll.a | grep Tcl_EvalFile returns:
``` 
[  6](sec  1)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x00000000 _Tcl_EvalFile
[  7](sec  3)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x00000000 __imp__Tcl_EvalFile
```
$ objdump.exe -t  /mingw64/lib/libtcl.dll.a | grep Tcl_EvalFile returns:
```
[  5](sec  1)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x0000000000000000 Tcl_EvalFile
[  6](sec  3)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x0000000000000000 __imp_Tcl_EvalFile
```
Is that extra underscore the standard behavior? 
And yosys in MINGW32 is expecting:
```
D:\Dev\Github\MINGW-packages\mingw-w64-yosys\src\yosys-yosys-0.26/kernel/driver.cc:574: undefined reference to '_imp__Tcl_EvalFile'
```

mmuetzel:
Those additional underscores are there to mark which calling convention is used when calling those functions across library boarders:
https://learn.microsoft.com/en-us/cpp/build/reference/decorated-names?view=msvc-170

If the calling convention used in the importing library doesn't match the calling convention expected by the exporting library, you might be seeing those `undefined reference to` errors.
I don't know if this is the case here.

mmuetzel:
Those additional underscores are there to mark which calling convention is used when calling those functions across library boarders:
https://learn.microsoft.com/en-us/cpp/build/reference/decorated-names?view=msvc-170

If the calling convention used in the importing library doesn't match the calling convention expected by the exporting library, you might be seeing those `undefined reference to` errors.
I don't know if this is the case here.

hmartinez82:
Ah, I knew about MSVC's name mangling 😊, I just wasn't sure about GCC's.

hmartinez82:
MING32 is probably failing due to https://github.com/msys2/MINGW-packages/issues/16814

hmartinez82:
Added workaround for detecting TCL's version and compiler flags

mingw-w64-yosys/001-fix-tcl-flags.patch
```diff
@@ -0,0 +1,33 @@
...
+-ifeq ($(CONFIG),mxe)
+-CXXFLAGS += -DYOSYS_ENABLE_TCL
+-LDLIBS += -ltcl86 -lwsock32 -lws2_32 -lnetapi32 -lz -luserenv
+-else
+-CXXFLAGS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --cflags tcl || echo -I$(TCL_INCLUDE)) -DYOSYS_ENABLE_TCL
+-LDLIBS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --libs tcl || echo $(TCL_LIBS))
+-endif
++TCL_INCLUDE ?= $(PREFIX)/include/$(TCL_VERSION)
++TCL_LIBS ?= -L$(PREFIX)/lib -ltcl
++CXXFLAGS += $(echo -I$(TCL_INCLUDE)) -DYOSYS_ENABLE_TCL
```
mmuetzel, line 29: Is there a reason to shell out and call `echo`. Wouldn't the following be the same but avoid spawning a new process?
``` suggestion
+CXXFLAGS += -I$(TCL_INCLUDE) -DYOSYS_ENABLE_TCL
```

hmartinez82: @mmuetzel doh, you are right. I even did that for LDLIBS already!

hmartinez82: Fixed :)

hmartinez82:
@Biswa96 I'll give it a try after The Merge

mingw-w64-yosys/PKGBUILD
```diff
@@ -1,51 +1,57 @@
...
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
 license=('spdx:ISC')
 url="http://www.clifford.at/yosys/"
 groups=("${MINGW_PACKAGE_PREFIX}-eda")
-depends=(
-  "${MINGW_PACKAGE_PREFIX}-ghdl"
-  "${MINGW_PACKAGE_PREFIX}-python"
-)
+depends=("${MINGW_PACKAGE_PREFIX}-ghdl"
```
umarcor, line 13: Please, preserve the format, to avoid modifying the lines that are not changed.

mingw-w64-yosys/PKGBUILD
```diff
@@ -1,51 +1,57 @@
+
 _realname=yosys
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.12.r45
+pkgver=0.27
```
umarcor, line 5: Yosys 0.28 was tagged five days ago: https://github.com/YosysHQ/yosys/releases/tag/yosys-0.28

hmartinez82: Updated to 0.30 :)

mingw-w64-yosys/PKGBUILD
```diff
@@ -12,52 +13,63 @@ groups=("${MINGW_PACKAGE_PREFIX}-eda")
...
-  "${_realname}::git+https://github.com/YosysHQ/${_realname}.git#commit=${_commit}"
-  "ghdl-yosys-plugin::git+https://github.com/ghdl/ghdl-yosys-plugin.git#commit=09a32cd1"
-)
-sha256sums=(
-  'SKIP'
-  'SKIP'
-)
+_commit='f7a8284'
+_plugin_commit='5b64ccf'
+source=("https://github.com/YosysHQ/yosys/archive/refs/tags/${_realname}-${pkgver}.tar.gz"
```
umarcor, line 29: Why did you change it to use a tarball when downloading yosys but still use git to retrieve ghdl-yosys-plugin?

mingw-w64-yosys/001-fix-tcl-flags.patch
```diff
@@ -0,0 +1,33 @@
...
+-endif
+-
+-ifeq ($(CONFIG),mxe)
+-CXXFLAGS += -DYOSYS_ENABLE_TCL
+-LDLIBS += -ltcl86 -lwsock32 -lws2_32 -lnetapi32 -lz -luserenv
+-else
+-CXXFLAGS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --cflags tcl || echo -I$(TCL_INCLUDE)) -DYOSYS_ENABLE_TCL
+-LDLIBS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --libs tcl || echo $(TCL_LIBS))
+-endif
++TCL_INCLUDE ?= $(PREFIX)/include/$(TCL_VERSION)
```
umarcor, line 27: Adding the `$(PREFIX)` to https://github.com/YosysHQ/yosys/blob/master/Makefile#L469 seems a sensible contribution to be done upstream.

With regard to TCL_LIBS, it seems that we can set it in the PKGBUILD, without requiring a patch.

mingw-w64-yosys/PKGBUILD
hmartinez82: Because there's a preference for tarball. Best would be if the plugins repo also provided tags/tarballs. But I can move everything back to git

umarcor: It's the same. GitHub provides zipfiles/tarballs for any "ref". For instance: `https://github.com/ghdl/ghdl-yosys-plugin/archive/5b64ccfdeee6c75f70487c1ea153ec3e1fb26cd1.zip`. It works with a shorter hash as well: `https://github.com/ghdl/ghdl-yosys-plugin/archive/5b64ccfd.zip`.

mingw-w64-yosys/PKGBUILD
```diff
@@ -12,52 +13,63 @@ groups=("${MINGW_PACKAGE_PREFIX}-eda")
...
   fi
 
-  if [ "$CARCH" = "x86_64" ]; then
+  if [ "$MSYSTEM_CARCH" = "x86_64" ]; then
     make config-msys2-64
   fi
 
-  mv "${srcdir}/ghdl-yosys-plugin"/src frontends/ghdl
+  mv "${srcdir}/ghdl-yosys-plugin/src" frontends/ghdl
+  local _tclver=$(echo 'set a "[info tclversion]" ; puts $a ; exit' | $MINGW_PREFIX/bin/tclsh)
```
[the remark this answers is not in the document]
hmartinez82: Even if it's for clarity?

mingw-w64-yosys/001-fix-tcl-flags.patch
hmartinez82: The CLANG32 issue was fixed we could use your original PR if we prefer.

hmartinez82:
I noticed that yosys has already been updated to 0.30. Closing this.

PR #15928, opened 2023-02-28
fftw: Add option DENABLE_OPENMP=ON

lock042:
The lib is not built with openmp and can be quite slow.

### DIFF
--- a/mingw-w64-fftw/PKGBUILD
+++ b/mingw-w64-fftw/PKGBUILD
@@ -88,6 +88,7 @@ build() {
       ${_config} \
       --enable-static \
       --enable-shared \
+      --enable-openmp \
       --enable-threads \
       --with-combined-threads \
       --with-our-malloc \
--- a/mingw-w64-fftw/PKGBUILD
+++ b/mingw-w64-fftw/PKGBUILD
@@ -16,7 +16,8 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-autotools"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              $( [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || \
                echo "${MINGW_PACKAGE_PREFIX}-gcc-fortran"))
-depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
+         "${MINGW_PACKAGE_PREFIX}-omp")
 options=('staticlibs' 'strip')
 source=("https://www.fftw.org/${_realname}-${_pkgver}.tar.gz"
         0001-detect-arm64-counter.patch)
--- a/mingw-w64-fftw/PKGBUILD
+++ b/mingw-w64-fftw/PKGBUILD
@@ -60,6 +60,7 @@ build() {
         ${_config_cmake} \
         -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
         -DBUILD_SHARED_LIBS=ON \
+        -DENABLE_OPENMP=ON \
         ../${_realname}-${_pkgver}
 
     # build library using autotools


lock042:
Of course I meant openmp

mmuetzel:
You'll probably need to add `"${MINGW_PACKAGE_PREFIX}-omp"` to the `depends`.

lock042:
@mmuetzel : Thx for the tip. This done now.

mmuetzel:
Looking at the build rules in a bit more detail, it looks like cmake is only used to create the `*Depends.cmake` file. It's not used to build the libraries.

To actually enable building the libraries with OpenMP, you'd probably need to add `--enable-openmp` to the `configure` flags a bit further down.

MehdiChinoune:
It was built with threading enabled which is sufficient to make it run in parallel.

lock042:
@mmuetzel : done it.
@MehdiChinoune : It looks like it is not the case: https://gitlab.com/free-astro/siril/-/issues/1025

mmuetzel:
IIUC, both variants (OpenMP and pthreads) can be built without conflicting with each other:
https://www.fftw.org/fftw3_doc/Usage-of-Multi_002dthreaded-FFTW.html

mmuetzel:
> This is strange, because it looks like it is not the case: https://gitlab.com/free-astro/siril/-/issues/1025

Is that project linked with the necessary libraries?
https://www.fftw.org/fftw3_doc/Usage-of-Multi_002dthreaded-FFTW.html
> First, programs using the parallel complex transforms should be linked with -lfftw3_threads -lfftw3 -lm on Unix, or -lfftw3_omp -lfftw3 -lm if you compiled with OpenMP. You will also need to link with whatever library is responsible for threads on your system (e.g. -lpthread on GNU/Linux) or include whatever compiler flag enables OpenMP (e.g. -fopenmp with gcc).

mmuetzel:
Error:
```
  configure: error: --with-combined-threads incompatible with --enable-openmp
```

Maybe, they can both be installed but not be built at the same time?

mmuetzel:
From the `configure.ac` file:
```
AC_ARG_ENABLE(openmp, [AC_HELP_STRING([--enable-openmp],[use OpenMP directives for parallelism])], enable_openmp=$enableval, enable_openmp=no)

if test "$enable_openmp" = "yes"; then
   AC_DEFINE(HAVE_OPENMP,1,[Define to enable OpenMP])
   AX_OPENMP([], [AC_MSG_ERROR([don't know how to enable OpenMP])])
fi

AC_ARG_ENABLE(threads, [AC_HELP_STRING([--enable-threads],[compile FFTW SMP threads library])], enable_threads=$enableval, enable_threads=no)

if test "$enable_threads" = "yes"; then
   AC_DEFINE(HAVE_THREADS,1,[Define to enable SMP threads])
fi

AC_ARG_WITH(combined-threads, [AC_HELP_STRING([--with-combined-threads],[combine threads into main libfftw3])], with_combined_threads=$withval, with_combined_threads=no)

if test "$with_combined_threads" = yes; then
    if test "$enable_openmp" = "yes"; then
        AC_MSG_ERROR([--with-combined-threads incompatible with --enable-openmp])
    fi
    if test "$enable_threads" != "yes"; then
        AC_MSG_ERROR([--with-combined-threads requires --enable-threads])
    fi
fi
```

The current versions of the fftw libraries probably already include the multithreading functions using pthread.
IIUC, we'd need to link all dependent projects with `fftw3_threads` or `fftw3_omp` if we'd like to build both variants here and like to support downstream users to select either implementation.
If we aren't careful about that, dependent projects might loose multithreading support in FFTW function calls.

mmuetzel:
Also note that "most threads" doesn't need to equal "fastest". See, e.g., this thread:
https://octave.discourse.group/t/fftw-threads-3-give-best-performance-for-28-cores-machine/3121/2

lock042:
@mmuetzel : I agree, but what I note is the difference behaviour between linux and windows. This is why I thought the missing DENABLE_OPENMP was responsible for this

mmuetzel:
I re-ran the example from the Octave thread with the octave from MSYS2 MINGW64. And the results are pretty similar:
image

Notable increase in performance for up to 3 threads. After that, the multithreading overhead seems to negate any potential performance increase by using more threads.

The library itself *is* using multiple threads though afaict...

At still seems to scale much worse if not operating on a power of 2 though.

mmuetzel:
Fwiw, ArchLinux builds both the pthreads and OpenMP versions of FFTW:
https://github.com/archlinux/svntogit-packages/blob/packages/fftw/trunk/PKGBUILD

Maybe, we should try and do the same...

lock042:
@mmuetzel : I think that would be more consistent yes.

mmuetzel:
See #15934.

lock042:
Closing this PR because #15934 covers the needs.

PR #15934, opened 2023-02-28
fftw: build with openmp

mmuetzel:
See also #15928. 

This change will likely require that at least the dependers that use the threading interface of FFTW will be rebuilt.
Let's see what the package grokker thinks...


lock042:
Ok thx for the PR. I think we can close mine.

mingw-w64-fftw/PKGBUILD
```diff
@@ -53,13 +66,19 @@ build() {
       _config_cmake="-DENABLE_QUAD_PRECISION=ON"
     fi
 
+    if [[ ${MINGW_PACKAGE_PREFIX} != *-clang-aarch64* ]]; then
```
Biswa96, line 69: Shouldn't this check CARCH == aarch64 ?

mmuetzel: I mirrored the condition from the part using autotools. But we could change it down there, too. 👍

Biswa96: OK, not so important. I was thinking about gcc with aarch64-w64-mingw32 target in future.

mmuetzel:
The package grokker pointed to itk and octave needing a rebuild. So, I bumped the pkgrel for just those two.

mingw-w64-fftw/PKGBUILD
mmuetzel: Thanks for the hint. Changed that in both places.

mingw-w64-fftw/PKGBUILD
```diff
@@ -87,7 +106,7 @@ build() {
       --enable-static \
       --enable-shared \
       --enable-threads \
-      --with-combined-threads \
```
ajeb78, line 90: From [fftw.org](https://www.fftw.org/fftw3_doc/Installation-on-Unix.html):

```--with-combined-threads: By default, if --enable-threads is used, the threads support is compiled into a separate library that must be linked in addition to the main FFTW library. This is so that users of the serial library do not need to link the system threads libraries. If --with-combined-threads is specified, however, then no separate threads library is created, and threads are included in the main FFTW library. This is mainly useful under Windows, where no system threads library is required and inter-library dependencies are problematic.```

So I think perhaps all the config flags are required: `--with-omp --with-threads --with-combined-threads`

mmuetzel: See https://github.com/msys2/MINGW-packages/pull/15928#issuecomment-1447960110

mmuetzel:
I rebuilt Octave to link to the OpenMP multithreading library of FFTW. More threads still don't necessarily mean faster evaluation. But performance seems to be a bit better overall and it seems to scale better than with the pthreads version (see https://github.com/msys2/MINGW-packages/pull/15928#issuecomment-1447990098)
Number of elements is 256 (power of 2):
image

Number of elements is 251 (prime):
image

mmuetzel:
I downloaded the MINGW64 build artifacts from this PR and installed the fftw and octave packages. Running the test suite in Octave with `__run_test_suite__`, no related tests are failing.

mmuetzel:
Since the OpenMP threading seems to be performing a bit better than POSIX threading at least in one software, I'm going to merge this. This will also give downstream projects a choice to select which multi-threading "mechanism" for FFTW they prefer. And it will align more with the libraries that are installed by most Linux binary distributions.

PR #15978, opened 2023-03-02
python-sphinx: update to 6.1.3

MehdiChinoune:


Biswa96:
Should there be `!strip` option because both python-sphnix and python-breath have .exe files?

PR #15992, opened 2023-03-02
gnutls: Update to 3.8.0

lazka:
guile bindings have been removed


lazka:
the now missing srp stuff was disabled upstream by default: https://gitlab.com/gnutls/gnutls/-/merge_requests/1681

PR #16001, opened 2023-03-03
gst-plugin-bad: Enable aarch64 builds

hmartinez82:


mingw-w64-openh264/PKGBUILD
```diff
@@ -1,6 +1,7 @@
 # Maintainer: Marcin Sielski <marcin.sielski@motorolasolutions.com>
 
 _realname=openh264
+_arch="i686"
```
Biswa96, line 4: Is this global variable required? It could be declared before using in build function as a local variable.

Biswa96: Oh, it is also required in package function.

PR #16012, opened 2023-03-03
openh264: Enable aarch64 builds

hmartinez82:
I fixed the build for ARM64 in https://github.com/msys2/MINGW-packages/pull/16001, but forgot to enable the build 😑


Biswa96:
Done in b4c3b19af03bf38328bff95ec95af2672d580982

PR #16027, opened 2023-03-04
neovim: Add package

Biswa96:
Requested by @techie-guy


Biswa96:
Previous attempt https://github.com/msys2/MINGW-packages/pull/6269 /cc @kassane @Freed-Wu

PR #16032, opened 2023-03-04
LLVM 16

MehdiChinoune:
requested by @jeremyd2019 and @mmuetzel on discord.


ognevnydemon:
will llvmlite and numba work ok with LLVM16?

MehdiChinoune:
> will llvmlite and numba work ok with LLVM16?

No, see https://github.com/numba/llvmlite/issues/897#issuecomment-1433080488

mmuetzel:
Thank you very much for providing these updates.

Is it ok to add some feedback to some of the changes at the current time?

I tried to build with them locally. But I'm not sure if I'm doing it correctly.
I started with `mingw-w64-clang` in the CLANG64 environment. Is that correct?
Building the packages succeeded. But there were some warning during stripping similar to the following:
```
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.asan-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.asan_cxx-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.asan_dynamic_runtime_thunk-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.fuzzer-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.fuzzer_no_main-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.stats-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.ubsan_standalone-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/libclang_rt.ubsan_standalone_cxx-x86_64.a': The end of the file was unexpectedly encountered
C:/msys64/clang64/bin/strip.exe: error: 'clang64/lib/clang/16/lib/windows/liborc_rt-x86_64.a': The end of the file was unexpectedly encountered
```
Is that expected?

After that, I tried to install the packages with the following command:
```
pacman -U mingw-w64-clang-x86_64-clang-16.0.0rc3-1-any.pkg.tar.zst mingw-w64-clang-x86_64-lld-16.0.0rc3-1-any.pkg.tar.zst mingw-w64-clang-x86_64-compiler-rt-16.0.0rc3-1-any.pkg.tar.zst mingw-w64-clang-x86_64-llvm-16.0.0rc3-1-any.pkg.tar.zst
```

But that failed with:
```
Fehler: Der Vorgang konnte nicht durchgeführt werden (In Konflikt stehende Dateien)
/clang64/lib/clang/16/include/fuzzer/FuzzedDataProvider.h existiert in »mingw-w64-clang-x86_64-llvm« und »mingw-w64-clang-x86_64-compiler-rt«
/clang64/lib/clang/16/include/orc/c_api.h existiert in »mingw-w64-clang-x86_64-llvm« und »mingw-w64-clang-x86_64-compiler-rt«
/clang64/lib/clang/16/include/profile/InstrProfData.inc existiert in »mingw-w64-clang-x86_64-llvm« und »mingw-w64-clang-x86_64-compiler-rt«
/clang64/lib/clang/16/include/sanitizer/allocator_interface.h existiert in »mingw-w64-clang-x86_64-llvm« und »mingw-w64-clang-x86_64-compiler-rt«
/clang64/lib/clang/16/include/sanitizer/asan_interface.h existiert in »mingw-w64-clang-x86_64-llvm« und »mingw-w64-clang-x86_64-compiler-rt«
[...]
```
And a lot more similar ones after that.
Sorry for the locale. (I don't know how to change it.) It basically says that there are file conflicts for the packages I was trying to install.

MehdiChinoune:
You could take a look at build-log and see that all the packages (except flang which times-out) build and install successfully without any conflict!
About stripping I don't know!

mingw-w64-clang/PKGBUILD
```diff
@@ -297,6 +262,8 @@ check() {
  # Remove || true once testcase doesn't cause failures.
 #  make check || true
 #  make check-clang || true
+  # TODO: download unittest and put it in thirdparty folder
+  ${MINGW_PREFIX}/bin/cmake.exe DLLVM_INCLUDE_TESTS=ON ../llvm
```
mmuetzel, line 266:
``` suggestion
  ${MINGW_PREFIX}/bin/cmake.exe -DLLVM_INCLUDE_TESTS=ON ../llvm
```

mingw-w64-clang/PKGBUILD
```diff
@@ -297,6 +262,8 @@ check() {
  # Remove || true once testcase doesn't cause failures.
 #  make check || true
 #  make check-clang || true
+  # TODO: download unittest and put it in thirdparty folder
```
mati865, line 265: I think it will fail due to MSYS2 path conversions.

mingw-w64-clang/PKGBUILD
mmuetzel: Fwiw, skipping the check "fixed" the file conflicts....

mingw-w64-flang/PKGBUILD
```diff
@@ -114,14 +95,17 @@ build() {
...
     "${extra_args[@]}" \
-    ../flang-${pkgver}.src
+    ../${_pkgfn}
 
   ${MINGW_PREFIX}/bin/cmake.exe --build .
 }
 
 check() {
   cd "${srcdir}"/build-${MSYSTEM}
+  ${MINGW_PREFIX}/bin/cmake.exe -DFLANG_INCLUDE_TESTS=ON ../${${_pkgfn}}
```
mmuetzel, line 107:
``` suggestion
  ${MINGW_PREFIX}/bin/cmake.exe -DFLANG_INCLUDE_TESTS=ON ../${_pkgfn}
```

mmuetzel:
I was able to build Octave with LLVM 16. It's frequently crashing. But that's a first step. 🎉
See also: https://octave.discourse.group/t/what-causes-a-c-template-to-be-implicitly-instantiated-how-to-build-octave-with-flang/3474/12

MehdiChinoune:
> I was able to build Octave with LLVM 16. It's frequently crashing. But that's a first step. 🎉 See also: https://octave.discourse.group/t/what-causes-a-c-template-to-be-implicitly-instantiated-how-to-build-octave-with-flang/3474/12

I tried, but it fails to configure:
```
checking for sgemm_ in -lopenblas... yes
checking whether LSAME is called correctly from Fortran... yes
checking whether ISAMAX is called correctly from Fortran... yes
checking whether SDOT is called correctly from Fortran... yes
checking whether DDOT is called correctly from Fortran... yes
checking whether CDOTU is called correctly from Fortran... no
checking whether ZDOTU is called correctly from Fortran... no
checking BLAS library integer size... 4
checking for sgemm_ in -lopenblas... yes
checking whether LSAME is called correctly from Fortran... yes
checking whether ISAMAX is called correctly from Fortran... yes
checking whether SDOT is called correctly from Fortran... yes
checking whether DDOT is called correctly from Fortran... yes
checking whether CDOTU is called correctly from Fortran... no
checking whether ZDOTU is called correctly from Fortran... no
checking BLAS library integer size... 4
configure: error: BLAS and LAPACK libraries are required
    Aborting...
```
flang couldn't call complex functions correctly from openblas:
```Fortran
      program main
      complex cdotu,a(1),b(1),w
      external cdotu
      a(1) = cmplx(1e0,1e0)
      b(1) = cmplx(1e0,2e0)
      w = cdotu(1,a,1,b,1)
      print*, "openblas cdotu:", w
      print*, "Fortran:", a(1)*b(1)
      if (w .ne. a(1)*b(1)) stop 1
      end
```
UCRT64:
```
$ gfortran main.f -o main_gf -lopenblas

$ ./main_gf.exe
 openblas cdotu:            (-1.00000000,3.00000000)
 Fortran:            (-1.00000000,3.00000000)
```
CLANG64:
```
$ flang main.f -o main -lopenblas

$ ./main.exe
 openblas cdotu: (3.,1.)
 Fortran: (-1.,3.)
Fortran STOP: code 1
```

mmuetzel:
@MehdiChinoune: Was that with the OpenBLAS as it is currently distributed for CLANG64?
It looks like the complex LAPACK functions in that library sometimes crash. And other times return results that are completely wrong. Maybe, I just got lucky that it didn't crash during the configure tests for me...

Does OpenBLAS compile with LLVM Flang without CLAPACK? In case it does, we could test if this is an error in the f2c translation of reference LAPACK that we are currently building as part of the OpenBLAS library on CLANG64.

MehdiChinoune:
> @MehdiChinoune: Was that with the OpenBLAS as it is currently distributed for CLANG64?

Yes
> Does OpenBLAS compile with LLVM Flang without CLAPACK? In case it does, we could test if this is an error in the f2c translation of reference LAPACK that we are currently building as part of the OpenBLAS library on CLANG64.

Well It doesn't because of some missing OpenMP (omp block) features but I will try to disable them or disable OpenMP completely.

mmuetzel:
It's very likely an issue with OpenBLAS (or how Flang interfaces to C functions).
I built the reference BLAS with LLVM (see #16061). Using that, your example program also works when built with Flang.

MehdiChinoune:
> Edit: It only links to the static BLAS library though. When trying to link to the shared BLAS library, I get the following error:
> ```
> $ flang test-cdotu.f -otest-cdotu -lblas
> ld.lld: error: Fortran::runtime::ExecutionEnvironment::Configure(int, char const**, char const**, EnvironmentDefaultList const*) was replaced
> flang-new: error: linker command failed with exit code 1 (use -v to see invocation)
> ```
> 
> Tbh, I don't know what that error is about...

Same here, tried it last night. I didn't try the static one.

mmuetzel:
I guess the linking of the `libblas` library went wrong somehow? `libblas.dll` is exporting the function from the error message (and a lot of others that are from the Fortran runtime). And the (static) runtime is linked again when trying to create the executable. IIUC, "*** was replaced" means: A symbol that should have been imported from a shared library was replaced by a symbol of the same name from one of the object files that are being linked currently...

I'm not sure if I've seen that runtime functions from the C or C++ runtime are exported from dlls.
Maybe, `--export-all-symbols` needs some more exceptions for the Flang runtime? Or the linker flags for the library are wrong?

MehdiChinoune:
> Maybe, `--export-all-symbols` needs some more exceptions for the Flang runtime? Or the linker flags for the library are wrong?

The flang libraries are all static there is no dllexport/dllimport.

mmuetzel:
That's maybe [this list](https://github.com/llvm/llvm-project/blob/main/lld/COFF/MinGW.cpp#L30-L55) that needs additions with `libFortran_main`, `libFortranRuntime`, and `libFortranDecimal`.

mmuetzel:
> The flang libraries are all static there is no dllexport/dllimport.

Yes, they are static. But since there are no `dllexport` attributes anywhere, all symbols (including those from those static libraries) are exported from the resulting shared library.

MehdiChinoune:
@mstorsjo could you investigate this one.

mmuetzel:
> Can you file a patch at https://reviews.llvm.org for it?

I opened a report here: https://reviews.llvm.org/D145389

MehdiChinoune:
> > Can you file a patch at https://reviews.llvm.org for it?
> 
> I opened a report here: https://reviews.llvm.org/D145389

Applying that patch to lld and rebuild lapack fixes the issue. It links now.

mmuetzel:
> Applying that patch to lld and rebuild lapack fixes the issue. It links now.

Hm. It didn't fix it for me. `nm -C /clang64/bin/libblas.dll | grep "T Fortran::"` still returns lots of symbols even after rebuilding...

MehdiChinoune:
> 

I was talking about linking the example above to blas.
Did you try rebuilding octave with reference blas instead of openblas (`--with-blas=-lblas`).

mmuetzel:
> I was talking about linking the example above to blas.

Linking your example also still fails with the same error.

mmuetzel:
> Did you try rebuilding octave with reference blas instead of openblas (`--with-blas=-lblas`).

I rebuilt Octave linking to reference BLAS and LAPACK. But some operations are still crashing.
E.g.:
```
A = [1+3i, 2+3i; 3-8i, 8+3i]
B = [8+i, 3+i; 4-9i, 3+i]
[v, d] = eig (A, B)
```

The top of the backtrace from that crash:
```
Thread 7 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 17548.0x3450]
0x00007ffd9f388678 in _FortranAzpowi () from C:\msys64\clang64\bin\liblapack.dll
(gdb) bt
#0  0x00007ffd9f388678 in _FortranAzpowi () from C:\msys64\clang64\bin\liblapack.dll
#1  0x00007ffd9f1ba18c in zhgeqz_ () from C:\msys64\clang64\bin\liblapack.dll
#2  0x00007ffd9f14fa97 in zggev_ () from C:\msys64\clang64\bin\liblapack.dll
#3  0x00007ffd2f112a3d in EIG::init (this=0xcd071fb110, a=..., b=..., calc_rev=<optimized out>,
    calc_lev=<optimized out>, force_qz=<optimized out>) at ../liboctave/numeric/EIG.cc:680
#4  0x00007ffd3774340e in EIG::EIG (this=this@entry=0xcd071fb110, a=..., b=..., calc_rev=true,
    calc_lev=<optimized out>, force_qz=<optimized out>) at ../liboctave/numeric/EIG.h:95
#5  0x00007ffd3773ef82 in octave::Feig (args=..., nargout=2) at ../libinterp/corefcn/eig.cc:300
#6  0x00007ffd375fe730 in octave::tree_evaluator::execute_builtin_function (this=0x1de032932d0, builtin_function=...,
    nargout=2, args=...) at ../libinterp/parse-tree/pt-eval.cc:3362
```

MehdiChinoune:
> > Did you try rebuilding octave with reference blas instead of openblas (`--with-blas=-lblas`).
> 
> I rebuilt Octave linking to reference BLAS and LAPACK.

Could you share the patch to make it build, It fails on my side
```
  CXX      libinterp/octave-value/liboctave_value_la-ov-java.lo
In file included from ../octave-7.3.0/libinterp/octave-value/ov-java.cc:48:
In file included from ../octave-7.3.0/libinterp/corefcn/Cell.h:34:
In file included from ../octave-7.3.0/liboctave/array/Array.h:43:
In file included from ../octave-7.3.0/liboctave/util/lo-utils.h:38:
../octave-7.3.0/liboctave/util/quit.h:84:15: error: conflicting types for 'operator=='
  inline bool operator == (const frame_info& a, const frame_info& b)
              ^
D:/Programs/msys64_test/clang64/include/guiddef.h:180:15: note: previous definition is here
__inline bool operator== (REFGUID guidOne, REFGUID guidOther) { return !!IsEqualGUID (guidOne, guidOther); }
           
   ^
```

> But some operations are still crashing. E.g.:
> 
> ```
> A = [1+3i, 2+3i; 3-8i, 8+3i]
> B = [8+i, 3+i; 4-9i, 3+i]
> [v, d] = eig (A, B)
> ```
> 
> The top of the backtrace from that crash:
> 
> ```
> Thread 7 received signal SIGSEGV, Segmentation fault.
> [Switching to Thread 17548.0x3450]
> 0x00007ffd9f388678 in _FortranAzpowi () from C:\msys64\clang64\bin\liblapack.dll
> (gdb) bt
> #0  0x00007ffd9f388678 in _FortranAzpowi () from C:\msys64\clang64\bin\liblapack.dll
> #1  0x00007ffd9f1ba18c in zhgeqz_ () from C:\msys64\clang64\bin\liblapack.dll
> #2  0x00007ffd9f14fa97 in zggev_ () from C:\msys64\clang64\bin\liblapack.dll
> #3  0x00007ffd2f112a3d in EIG::init (this=0xcd071fb110, a=..., b=..., calc_rev=<optimized out>,
>     calc_lev=<optimized out>, force_qz=<optimized out>) at ../liboctave/numeric/EIG.cc:680
> #4  0x00007ffd3774340e in EIG::EIG (this=this@entry=0xcd071fb110, a=..., b=..., calc_rev=true,
>     calc_lev=<optimized out>, force_qz=<optimized out>) at ../liboctave/numeric/EIG.h:95
> #5  0x00007ffd3773ef82 in octave::Feig (args=..., nargout=2) at ../libinterp/corefcn/eig.cc:300
> #6  0x00007ffd375fe730 in octave::tree_evaluator::execute_builtin_function (this=0x1de032932d0, builtin_function=...,
>     nargout=2, args=...) at ../libinterp/parse-tree/pt-eval.cc:3362
> ```

The same thing happened to me with another local app.

mmuetzel:
I'm building from a current head of the default branch of the Mercurial repository (hg id 90d7278afb00).
I made the following local changes: 
[octave-changes-for-flang.zip](https://github.com/msys2/MINGW-packages/files/10902410/octave-changes-for-flang.zip)

I didn't clean them up yet. Some of those changes might not actually be necessary. Or they might be breaking compilation with other compilers or on other platforms...

jeremyd2019:
I was messing with workflows today, and happened to notice that the only llvm project package that you did not update here was mingw-w64-libclc.

jeremyd2019:
I tried building updated libclc in my attempts but got a non-descript error, don't know what went wrong with it.

mmuetzel:
I was trying to build the LAPACK_EXAMPLES from here:
https://people.sc.fsu.edu/~jburkardt/f77_src/lapack_examples/lapack_examples.html

But linking it fails with the following error:
```
$ flang lapack_examples_test.f -llapack -lblas -o lapack_examples_test.exe
ld.lld: error: undefined symbol: clock_gettime
>>> referenced by libFortranRuntime.a(time-intrinsic.cpp.obj):(_FortranACpuTime)
>>> referenced by libFortranRuntime.a(time-intrinsic.cpp.obj):(_FortranASystemClockCount)
flang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Afaict, that function exists in the pthread library. Does that mean that -pthread needs to be added to the runtime library flags of Flang? That looks like a big dependency to me.
There is already some logic to build the runtime without that function. Would it make sense to disable it for MinGW?
Does the following alternative patch to `0004-link-runtime-to-pthread.patch` look reasonable?
``` diff
From 63f4b973b0134614b326900e3ec99948ecb5ca9e Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Markus=20M=C3=BCtzel?= <markus.muetzel@gmx.de>
Date: Wed, 8 Mar 2023 18:36:40 +0100
Subject: [PATCH] [flang] Don't use functions from pthread in runtime on MinGW.

clock_gettime is implemented in the pthread library for MinGW.
That would require that all programs that link libFortranRuntime are linked
to pthread. Instead, don't use the function.
---
 runtime/time-intrinsic.cpp | 7 +++++++
 1 file changed, 7 insertions(+)

diff --git a/runtime/time-intrinsic.cpp b/runtime/time-intrinsic.cpp
index 5f7b50964fe2..e22703e7a78f 100644
--- a/runtime/time-intrinsic.cpp
+++ b/runtime/time-intrinsic.cpp
@@ -68,6 +68,13 @@ template <typename Unused = void> double GetCpuTime(fallback_implementation) {
 #undef CLOCKID
 #endif
 
+#if defined(__MINGW32__)
+// clock_gettime is implemented in the pthread library for MinGW.
+// That would require that all programs that link libFortranRuntime are linked
+// to pthread. Instead, don't use the function.
+#undef CLOCKID
+#endif
+
 #ifdef CLOCKID
 // POSIX implementation using clock_gettime. This is only enabled where
 // clock_gettime is available.
-- 
2.38.0.windows.1
```

jeremyd2019:
> I tried building updated libclc in my attempts but got a non-descript error, don't know what went wrong with it.
> 
> ```
>   [2862/2871] Generating spirv-mesa3d-.spv
>   FAILED: spirv-mesa3d-.spv C:/_/B/src/build-CLANG32/spirv-mesa3d-.spv 
>   cmd.exe /C "cd /D C:\_\B\src\build-CLANG32 && D:\M\msys64\clang32\bin\llvm-spirv.exe --spirv-max-version=1.1 -o spirv-mesa3d-.spv builtins.link.spirv-mesa3d-.bc"
> ```

This also occurs trying to build libclc 15.0.7 with llvm/clang 16.0.0rc3

mati865:
> Afaict, that function exists in the pthread library. Does that mean that -pthread needs to be added to the runtime library flags of Flang?

Definitely not, there should be MSVC path available that doesn't rely on pthread.

MehdiChinoune:
> > Afaict, that function exists in the pthread library. Does that mean that -pthread needs to be added to the runtime library flags of Flang?
> 
> Definitely not, there should be MSVC path available that doesn't rely on pthread.

He was talking about `clock_gettime` which is not available for MSVC. I fixed the compilation with MSVC in https://github.com/llvm/llvm-project/commit/360411957
Looking at the source code I don't think there is an implementation for MSVC, Flang developers don't care that much about MSVC. It was external developers (no paid ones from Nvidia/AMD/ARM/US-Labs ...) who make it build with MSVC.

jeremyd2019:
turns out llvm-spirv.exe comes from [mingw-w64-spirv-llvm-translator](https://packages.msys2.org/base/mingw-w64-spirv-llvm-translator) not from an llvm project, so that explains that failure.  I assume that project will make a release once llvm does (they have branches for different llvm releases, including 16, and their main branch has already moved to 17 to match llvm).

mmuetzel:
> Definitely not, 

I agree, pthread feels like a heavy dependency for a runtime that should be avoided if possible.

> there should be MSVC path available that doesn't rely on pthread.

IIUC, the patch that I was proposing would select the code path that doesn't rely on the function from pthread on MinGW. Thank you to @MehdiChinoune who -- as it turns out -- opened that code path in the first place.

MehdiChinoune:
OK to merge?

mingw-w64-openmp/PKGBUILD
```diff
@@ -20,32 +20,24 @@ license=("custom:Apache 2.0 with LLVM Exception")
...
-makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja"
-             "${MINGW_PACKAGE_PREFIX}-cc"
-             $( (( _clangprefix )) || echo "${MINGW_PACKAGE_PREFIX}-gcc-fortran")
-             $( [[ $CARCH != *86* ]] || echo "${MINGW_PACKAGE_PREFIX}-uasm" ))
-options=('!debug' 'strip')
+             $( [[ $MINGW_PACKAGE_PREFIX == *-clang-* && ${MSYSTEM} != CLANG64 ]] || echo \
+               "${MINGW_PACKAGE_PREFIX}-fc")
```
jeremyd2019, line 27: Is this actually using the fortran compiler on CLANG64, or does something more need to be done to enable it with cmake options or something?  I am inclined to enable flang for clangarm64 for llvm 16, since it seems to do some useful things now, and I can at minimum manually build it, but I noticed this was the only change in openmp for fortran support on CLANG64...

jeremyd2019:
> OK to merge?

Do we need to wait for dependent packages to catch up first (so they have versions that build against llvm 16), or just let them be broken until they do?

mingw-w64-openmp/PKGBUILD
MehdiChinoune: You are right, It requires `-DLIBOMP_FORTRAN_MODULES=ON` to be passed to cmake.

MehdiChinoune:
> Do we need to wait for dependent packages to catch up first (so they have versions that build against llvm 16), or just let them be broken until they do?

- [ ] mingw-w64-include-what-you-use
- [ ] mingw-w64-mesa
- [ ] mingw-w64-spirv-llvm-translator
- [ ] mingw-w64-openshadinglanguage
- [ ] mingw-w64-creduce
- [ ] mingw-w64-qt-creator
- [ ] mingw-w64-qt6-static

mingw-w64-openmp/PKGBUILD
jeremyd2019: Should openmp update then be delayed until after flang is updated?  (that'd make it easier to enable clangarm64 at the same time)

jeremyd2019:
I am particularly interested in mingw-w64-spirv-llvm-translator as that is needed to build mingw-w64-libclc (which is released as part of llvm), and I think both are needed to build mesa.

mingw-w64-openmp/PKGBUILD
MehdiChinoune: the old one (15.0.7) couldn't build the modules.

jeremyd2019:
If you have a branch handy with all of the llvm 16 changes ready to go, I can hopefully try to build everything on arm64 tomorrow.  (If not, I can probably piece something together easily, I still have the -rc4 branch).

MehdiChinoune:
> If you have a branch handy with all of the llvm 16 changes ready to go, I can hopefully try to build everything on arm64 tomorrow. (If not, I can probably piece something together easily, I still have the -rc4 branch).

https://github.com/MehdiChinoune/MINGW-packages/tree/llvm-16-p2

jeremyd2019:
> https://github.com/MehdiChinoune/MINGW-packages/tree/llvm-16-p2

This builds on arm64, with the following changes:
```diff
diff --git a/mingw-w64-flang/PKGBUILD b/mingw-w64-flang/PKGBUILD
index 24c7a08c3..714862db0 100644
--- a/mingw-w64-flang/PKGBUILD
+++ b/mingw-w64-flang/PKGBUILD
@@ -9,7 +9,7 @@ pkgver=${_version}${_rc/-/}
 pkgrel=1
 pkgdesc="Fortran frontend for LLVM (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw64' 'ucrt64' 'clang64')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://flang.llvm.org/"
 license=("custom:Apache 2.0 with LLVM Exception")
 provides=($([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-fc"))
diff --git a/mingw-w64-openmp/PKGBUILD b/mingw-w64-openmp/PKGBUILD
index 90279e6cb..5d410fbb3 100644
--- a/mingw-w64-openmp/PKGBUILD
+++ b/mingw-w64-openmp/PKGBUILD
@@ -23,7 +23,7 @@ depends=($( (( _clangprefix )) || echo "${MINGW_PACKAGE_PREFIX}-gcc-libs"))
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja"
-             $( [[ $MINGW_PACKAGE_PREFIX == *-clang-* && ${MSYSTEM} != CLANG64 ]] || echo \
+             $( [[ $MINGW_PACKAGE_PREFIX == *-clang-* && ${MSYSTEM} != *64 ]] || echo \
                "${MINGW_PACKAGE_PREFIX}-fc")
              $( [[ $CARCH == aarch64 ]] || echo "${MINGW_PACKAGE_PREFIX}-uasm" ))
 _url=https://github.com/llvm/llvm-project/releases/download/${_tag}
@@ -61,7 +61,7 @@ prepare() {
   fi
   apply_patch_with_msg 002-hacks-for-static-linking.patch
   # Fix building Fortran modules with flang
-  sed -e 's|^!DIR|^!!DIR|g' -i runtime/src/include/omp_lib.f90.var
+  sed -e 's|^!DIR|!!DIR|g' -i runtime/src/include/omp_lib.f90.var
 }
 
 build() {
@@ -92,7 +92,7 @@ build() {
       -DCMAKE_BUILD_TYPE=${_build_type} \
       -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
       -DCMAKE_SYSTEM_IGNORE_PATH=/usr/lib \
-      -DLIBOMP_FORTRAN_MODULES="$( [[ (( _clangprefix )) && ${MSYSTEM} != CLANG64 ]] && echo OFF || echo ON )" \
+      -DLIBOMP_FORTRAN_MODULES="$( [[ (( _clangprefix )) && ${MSYSTEM} != *64 ]] && echo OFF || echo ON )" \
       -DLIBOMP_ENABLE_SHARED="$( [[ $shared_libs == static ]] && echo OFF || echo ON )" \
       "${platform_config[@]}" \
       -DPython3_EXECUTABLE=${MINGW_PREFIX}/bin/python \
```

MehdiChinoune:
I don't think we could merge this until OpenShadingLanguage support LLVM-16. It's not a package that we could break, as It's required by many other packages.

oscarfv:
Doesn't pacman support version constraints in dependencies? And transitive application of those constraints?

MehdiChinoune:
> Doesn't pacman support version constraints in dependencies? And transitive application of those constraints?

How about new installation, reinstallation, CI ...
pacman contains only one version of each package, If for example pinned OSL dependency to llvm<16 then:
- The upgrade will be refused if OSL is already installed
- The installation will fail for anyone attempting to install OSL as there is only one version llvm=16.0.0

oscarfv:
The good news here is that the problem faced by OSL is not so difficult as the maintainer thinks: LLVM 16 still supports typed pointers, he just need to enable them.

For LLVM 17 only opaque pointers will be supported.

mati865:
>  Doesn't pacman support version constraints in dependencies? And transitive application of those constraints?

To achieve that we would have to introduce `<prefix>-llvm<version>` package like Arch Linux does.

MehdiChinoune:
> > Doesn't pacman support version constraints in dependencies? And transitive application of those constraints?
> 
> To achieve that we would have to introduce `<prefix>-llvm<version>` package like Arch Linux does.

They did that by installing LLVM libs in different location and then symlinking them with `-version` suffix, something we could not do on Windows. A similar approach were made for Qt6 to avoid conflict with Qt5.

mati865:
No, they install everything at the same locations build build older version with suffixes.

MehdiChinoune:
> No, they install everything at the same locations build build older version with suffixes.

https://github.com/archlinux/svntogit-packages/blob/packages/llvm14/trunk/PKGBUILD#L68
There is no option to suffix programs and libraries with LLVM

oscarfv:
> > No, they install everything at the same locations build build older version with suffixes.
> 
> https://github.com/archlinux/svntogit-packages/blob/packages/llvm14/trunk/PKGBUILD#L68 There is no option to suffix programs and libraries with LLVM

Programs no, but libraries can have a suffix. There is `LLVM_VERSION_SUFFIX`, used in the function `llvm_add_library` in `cmake/modules/AddLLVM.cmake`.

MehdiChinoune:
> 

That's only used for SOVERSION (no static libraries) and appended to LLVM_MAJOR_VERSION, and for output name for only libLLVM (when SONAME passed to llvm_add_library), which already the case.

oscarfv:
> > 
> 
> That's only used for SOVERSION (no static libraries) and appended to LLVM_MAJOR_VERSION, and for output name for only libLLVM (when SONAME passed to llvm_add_library), which already the case.

Changing `llvm_add_library` to use `LLVM_VERSION_SUFFIX` or a new variable on all cases seems quite simple.

If projects like OSL do not update to opaque pointers on the next 6 months, we will face the same problem unless we are capable of making a versioned, libraries-only LLVM-16 package. That or waiting indefinitely for OSL.

MehdiChinoune:
creduce and IWYU both requires Clang==15
IWYU will publish a release soon, creduce will be broken until they add support for LLVM-16.

jeremyd2019:
I tried building updated spirv-llvm-translator on arm64, but it failed:
```
  FAILED: lib/SPIRV/CMakeFiles/LLVMSPIRVLib.dir/LLVMToSPIRVDbgTran.cpp.obj 
  C:\msys64\clangarm64\bin\clang++.exe -D_FILE_OFFSET_BITS=64 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -IC:/_/B/src/SPIRV-LLVM-Translator-16.0.0/include -IC:/msys64/clangarm64/include/spirv/include -IC:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV -IC:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/libSPIRV -IC:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/Mangler -O2 -pipe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -O3 -DNDEBUG -std=gnu++17  -fno-exceptions -MD -MT lib/SPIRV/CMakeFiles/LLVMSPIRVLib.dir/LLVMToSPIRVDbgTran.cpp.obj -MF lib\SPIRV\CMakeFiles\LLVMSPIRVLib.dir\LLVMToSPIRVDbgTran.cpp.obj.d -o lib/SPIRV/CMake...
  In file included from C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/LLVMToSPIRVDbgTran.cpp:38:
  In file included from C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/LLVMToSPIRVDbgTran.h:42:
  In file included from C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/libSPIRV/SPIRVModule.h:44:
  In file included from C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/libSPIRV/SPIRVEntry.h:44:
  C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/libSPIRV/SPIRVEnum.h:455:17: error: use of undeclared identifier 'CapabilityFPGAArgumentInterfacesINTEL'
                 {CapabilityFPGAArgumentInterfacesINTEL});
                  ^
  C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/libSPIRV/SPIRVEnum.h:454:3: error: invalid application of 'sizeof' to an incomplete type 'SPIRVCapabilityKind[]' (aka 'spv::Capability[]')
    ADD_VEC_INIT(DecorationConduitKernelArgumentINTEL,
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  C:/_/B/src/SPIRV-LLVM-Translator-16.0.0/lib/SPIRV/libSPIRV/SPIRVEnum.h:144:32: note: expanded from macro 'ADD_VEC_INIT'
      SPIRVCapVec V(C, C + sizeof(C) / sizeof(C[0]));                            \
                                 ^~~
```

The errors all seemed to be things that ended in `INTEL` being undefined.

Biswa96:
> I tried building updated spirv-llvm-translator on arm64, but it failed

Probably, it needs a specific spirv-headers git commit. spirv-llvm-translator project uses a specific git commit from here https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/spirv-headers-tag.conf.

MehdiChinoune:
I think we could merge this
with (static) LLVM-15 #16480, OSL and IWYU will be fixed.

lazka:
I haven't read all of the above, so it might be already known (or will be fixed by the flang update), but integration tests currently crash in flang:

```
autoreconf-2.71: running: aclocal --force 
#0 0x00007ff6d3306a10 llvm::opt::OptTable::ParseOneArg(llvm::opt::ArgList const&, unsigned int&, unsigned int, unsigned int) const (D:\a\_temp\msys64\clang64\bin\flang-new.exe+0x156a10)
#1 0x00007ff6d3306f7f llvm::opt::OptTable::ParseArgs(llvm::ArrayRef<char const*>, unsigned int&, unsigned int&, unsigned int, unsigned int) const (D:\a\_temp\msys64\clang64\bin\flang-new.exe+0x156f7f)
#2 0x00007ff6d31b17ee main (D:\a\_temp\msys64\clang64\bin\flang-new.exe+0x17ee)
#3 0x00007ff6d31b1316 WinMainCRTStartup (D:\a\_temp\msys64\clang64\bin\flang-new.exe+0x1316)
#4 0x00007ff6d31b1366 mainCRTStartup (D:\a\_temp\msys64\clang64\bin\flang-new.exe+0x1366)
#5 0x00007ff9f5c74de0 (C:\Windows\System32\KERNEL32.DLL+0x14de0)
#6 0x00007ff9f6d7e40b (C:\Windows\SYSTEM32\ntdll.dll+0x7e40b)
./test.sh: line 12:  1388 Segmentation fault      "$FC" hello.f90 -o hello.exe
```

MehdiChinoune:
> I haven't read all of the above, so it might be already known (or will be fixed by the flang update), but integration tests currently crash in flang:

It builds and runs with this version
```
$ flang main.f90 -o main

$ ./main
 Hello, World!
```

lazka:
Another crash when building starpu under clang32:

```
Exception Code: 0xC0000005
  #0 #00x75d99f84 0x75d99f84 (C:\WINDOWS\System32\ucrtbase.dll+ 0x29f84)(C:\WINDOWS\System32\ucrtbase.dll+0x29f84)
 #1
  #10x75d99db0 0x75d99db0  ((C:\WINDOWS\System32\ucrtbase.dll+C:\WINDOWS\System32\ucrtbase.dll+0x29db0)
0x29db0)
 #2 0x75de6d8e (C:\WINDOWS\System32\ucrtbase.dll+0x76d8e)
  #3 #20x75d99c2a  (C:\WINDOWS\System32\ucrtbase.dll+0x75de6d8e 0x29c2a)
 #4( C:\WINDOWS\System32\ucrtbase.dll+0x76d8e)
 0x583fd870 llvm::raw_fd_ostream::write_impl(char const*, unsigned int) (C:\msys64\clang32\bin\libLLVM-16.dll+0x12d870)
 #5 0x75d9fdbd (C:\WINDOWS\System32\ucrtbase.dll+#30x2fdbd)
  #60x75d99c2a  0x583fc2eb( C:\WINDOWS\System32\ucrtbase.dllllvm::raw_ostream::write(char const*, unsigned int)+ (0x29c2a)C:\msys64\clang32\bin\libLLVM-16.dll
+0x12c2eb)
 #7 0x5839d003 llvm::isPrefixedHexStyle(llvm::HexPrintStyle) (C:\msys64\clang32\bin\libLLVM-16.dll+0xcd003)
 #8 0x5839c69a llvm::write_integer(llvm::raw_ostream&, unsigned long long, unsigned int, llvm::IntegerStyle) (C:\msys64\clang32\bin\libLLVM-16.dll+0xcc69a)
 #9 0x5839c66b llvm::write_integer(llvm::raw_ostream&, unsigned long long, unsigned int, llvm::IntegerStyle) (C:\msys64\clang32\bin\libLLVM-16.dll+0xcc66b)
#10 0x00c812c1 WinMainCRTStartup (C:\msys64\clang32\bin\clang.exe+0x12c1)
#11 0x76b77d69 (C:\WINDOWS\System32\KERNEL32.DLL+0x17d69)
#12 0x775eb74b (C:\WINDOWS\SYSTEM32\ntdll.dll+0x6b74b)
#13 0x775eb6cf (C:\WINDOWS\SYSTEM32\ntdll.dll+0x6b6cf)
 #4 0x583fd870 llvm::raw_fd_ostream::write_impl(char const*, unsigned int) (C:\msys64\clang32\bin\libLLVM-16.dll+0x12d870)
```

lazka:
the new flang works nicely again, and the starpu error disappeared once I fixed the code related error. So those are likely no blockers any more.

PR #16040, opened 2023-03-05
octave: link against openblas instead of lapack

MehdiChinoune:
This should make octave compute faster. waiting for some tests from @mmuetzel

### DIFF
--- a/mingw-w64-octave/PKGBUILD
+++ b/mingw-w64-octave/PKGBUILD
@@ -2,7 +2,7 @@ _realname=octave
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
 pkgver=7.3.0
-pkgrel=12
+pkgrel=13
 pkgdesc="GNU Octave: Interactive programming environment for numerical computations (mingw-w64)"
 url="https://www.octave.org"
 license=('spdx:GPL-3.0-or-later')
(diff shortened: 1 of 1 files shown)


mmuetzel:
Thanks. 

I didn't notice that before because it looks like it was linking to both the reference BLAS and OpenBLAS. So, Octave could "see" functions that were specific to OpenBLAS but was still using most of the functions from the reference BLAS library. The output from `version -blas` in Octave was misleading.

Before your change:
```
octave:1> A=randn(2000); B=A; tic, C=A*B; toc
Elapsed time is 4.30051 seconds.
```

After your change:
```
octave:1> A=randn(2000); B=A; tic, C=A*B; toc
Elapsed time is 0.0486281 seconds.
```

Looks good to me!

Unrelated to this PR: In general, it might be nice if we had a mechanism similar to Debian's alternatives which would allow users to switch between reference BLAS/LAPACK (better precision in general) and OpenBLAS (better performance in general) without having to rebuild all dependent libraries. But that is difficult currently because we bundle the BLAS and LAPACK functions in the same OpenBLAS library. IIUC, Debian's OpenBLAS library only provides the BLAS functions. (And afaict, OpenBLAS doesn't improve upon the reference LAPACK functions anyway.)

mmuetzel (approved):

mmuetzel:
Just to leave some additional information for anyone coming back to this in the future:
The performance difference isn't caused by using functions from the reference LAPACK implementation. It's the functions from the reference BLAS implementation that are making the difference.

What was probably causing this confusion is likely fixed by the changes in #16044.

PR #16061, opened 2023-03-06
lapack: Prepare for compilation with LLVM Flang.

mmuetzel:
Not sure if we should really bump the pkgrel already. The existing packages shouldn't be changing.
Please, let me know and I'll revert that part.

### DIFF
--- a/mingw-w64-lapack/PKGBUILD
+++ b/mingw-w64-lapack/PKGBUILD
@@ -4,15 +4,15 @@ _realname=lapack
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.11.0
-pkgrel=1
+pkgrel=2
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
 pkgdesc="Linear Algebra PACKage (mingw-w64)"
-depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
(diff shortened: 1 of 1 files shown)


mingw-w64-lapack/PKGBUILD
```diff
@@ -28,37 +28,42 @@ build() {
...
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
     ${MINGW_PREFIX}/bin/cmake.exe \
       -G"Ninja" \
       -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
-      -DCMAKE_Fortran_COMPILER=${MINGW_PREFIX}/bin/gfortran.exe \
       -DCMAKE_BUILD_TYPE=${_btype} \
       -DBUILD_SHARED_LIBS=ON \
       -DCMAKE_NEED_RESPONSE=ON \
-      -DBUILD_TESTING=ON \
+      -DBUILD_TESTING=OFF \
```
lazka, line 46: just wondering, why?

mmuetzel: It was failing to link the tests with the library.
I'll re-do that now it is [likely fixed](https://reviews.llvm.org/D145389).

lazka: ah, ok

mmuetzel:
Given that LAPACK crashes on CLANG64 for some functions with complex input, I'm not sure if we should package it.
Preparing the build rules might still make sense in case someone would like to build it manually and experiment with it.

lazka:
lgtm

PR #16079, opened 2023-03-06
cargo-generate: New package

podsvirov:


podsvirov:
This package does documentation provided by `mdbook build` comand invoked in `guide` subdirectory from source code tarball.

filnet:
Is it the same guide as the one available onlie at https://cargo-generate.github.io/cargo-generate/templates/index.html ?

podsvirov:
Yes.

filnet:
Thanks for the answers. 

I have to admit that I don't get the point of this new package :)

podsvirov:
The meaning is the same as for any other package in the distribution.  The user simply installs the package and immediately starts using the new functionality, rather than building and installing it from source codes for tens of minutes.

filnet:
Yes but what is the advantage over doing `cargo install cargo-generate` (which seems more straightforward) ?

filnet:
I mean, why would we want to duplicate what is already provided by the rust ecosystem ? Feels like an unnecessary maintenance burden. That is unless I have missed a subtle point.

filnet:
Is the use case CI (to speed it up) ?

lazka:
No strong opinion here..

I see that some distros package it at least: https://repology.org/project/cargo-generate/versions

podsvirov:
What about if we will be packaging too? 🤔

filnet:
I don't have a strong opinion neither.
I do have a question though: does this package need to be updated in lockstep with rust ?

filnet:
I am assuming that my question(s) go unanswered because no one here as an answer.
The best way to get answers it to try it out. So please merge :)

PR #16099, opened 2023-03-08
Add MinGW git.

### LINKED ISSUE #16091: Would you like a MinGW git?
### Package name

mingw-w64-git

It seems that it is a long history that MSYS2 cannot distribute git-for-windows at last. The git-for-windows project finally decided to pack it themselves, and MSYS2 refused the patches from that project because it was too large and complicated.

Nowadays, git-for-windows project, basically and directly, uses the MSYS2 runtime without (or sometimes with a few) patches. Git also writes a new buildsystem alternative with cmake. It is possible to build a MinGW git again with MSYS2.

I have written a `PKGBUILD` for it, and have use it for more than half a year. The repo is here: https://github.com/Berrysoft/msys-git-for-windows/ . It also contains some other git tools, for example, I also build `git-lfs` myself.

This `PKGBUILD` contains only 4 patches. 3 of them are necessary patches to build for 5 MinGW targets (I cannot test clangarm64). One of them is a patch that wasn't accepted by upstream: https://github.com/git-for-windows/git/pull/3869 , which is about shebang parsing. The upstream maintainers thought that they don't need to parsing *that* correctly because after all they pack all MSYS2 tools themselves, but I need this patch to make it work well with my MSYS2 environment.

I'm now asking that if you, the maintainers of this repo, would like a MinGW git. If so, I'll open a PR based on my `PKGBUILD` and without the customized version string.

I have also found that it is possible to build for CLANGARM64 environment. All additional work is to remove the custom allocator, and use the system `malloc/free`. It could be simply done by removing one code file. I cannot check if it really works as expected, but I know it could be built successfully.

https://gitforwindows.org/

``` bash
$ git --version
```

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [X] CLANGARM64

Yes

Berrysoft:
Fixes #16091 

I need some words to explain patch `6-interpr.patch`:

Originally, git-for-windows uses regex to parse shebang, to get the interpreter name. Then it looks up the `PATH` env var and finds the first matched interpreter. It is not what we want, because MSYS2 environment is rather complicated. The perl scripts in git could only be executed by MSYS2 perl, but not MinGW perl. If a user has both perl installed, the original parsing and finding algorithm will find the wrong interpreter, and make `git-send-email` fail to work.

Therefore, after some times of trying, I found that `env.exe` could easily solve this problem. It reads the shebang properly, and could also recognize all kinds of paths, both MSYS2 UNIX path, and Windows path. In addition, if we have MSYS2 installed, we 100% have `env.exe` installed, and it is placed in `PATH` in every MSYS2 environments. Therefore I don't need to parse the shebang at all. Simply setting the "interpreter" as `env.exe`, and all scripts and executables should exactly work.

This patch means that if users want to use `git` outside MSYS2, they need to add both `/{mingw64, ucrt64, clang64}/bin` and `/usr/bin` to `PATH` to make it work. Actually I always do that. Maybe we should document this behavior somewhere.

I also add `provides` to `PKGBUILD`. There are some packages requires `git`, and now they can use this MinGW git.


Biswa96:
All the compiling steps should be in `build()` function.

@dscho would you like to review this pull request?

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+pkgbase="mingw-w64-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=2.39.2.windows.1
+pkgrel=1
+pkgdesc="The fast distributed version control system (mingw-w64)"
+arch=('any')
+url="https://git-for-windows.github.io/"
+license=('GPL2')
+groups=('VCS')
+provides=('git' 'git-core')
```
MehdiChinoune, line 13:
```suggestion
provides=('git' 'git-core')
```
I don't think mingw-w64 could provide an msys2 one.

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+_realname=git
+pkgbase="mingw-w64-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=2.39.2.windows.1
+pkgrel=1
+pkgdesc="The fast distributed version control system (mingw-w64)"
+arch=('any')
+url="https://git-for-windows.github.io/"
+license=('GPL2')
+groups=('VCS')
```
MehdiChinoune, line 12: We don't have such a group. that group belongs to msys2 environment.

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+provides=('git' 'git-core')
+
+options=('strip')
```
MehdiChinoune, line 15:
```suggestion
```
this the default. not needed

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+
+makedepends=("diffutils"
+             "docbook-xsl"
+             "make"
+             "patch"
+             "xmlto"
```
MehdiChinoune, line 21:
```suggestion
makedepends=("${MINGW_PACKAGE_PREFIX}-docbook-xsl"
             "xmlto"
```
other packages are installed by default when installing base-devel.

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+         "perl-MIME-tools"
+         "perl-Net-SMTP-SSL"
+         "perl-TermReadKey")
+
+optdepends=("subversion: git-svn"
+            "${MINGW_PACKAGE_PREFIX}-python: git-p4"
+            "${MINGW_PACKAGE_PREFIX}-tk: gitk & git-gui")
+
+replaces=("${MINGW_PACKAGE_PREFIX}-${_realname}-doc-html"
+          "${MINGW_PACKAGE_PREFIX}-${_realname}-doc-man")
```
MehdiChinoune, line 50: Do these packages exist?

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+source=("https://github.com/git-for-windows/git/archive/refs/tags/v$pkgver.tar.gz")
+
+patches=("1-mingw.patch"
+         "2-version.patch"
+         "4-headless.patch"
+         "6-interpr.patch")
+source+=(${patches[@]})
+
+arm64_patches=("7-arm64.patch")
+source+=(${arm64_patches[@]})
```
MehdiChinoune, line 61:
```suggestion
source=("https://github.com/git-for-windows/git/archive/refs/tags/v$pkgver.tar.gz"
         "1-mingw.patch"
         "2-version.patch"
         "4-headless.patch"
         "6-interpr.patch"
         "7-arm64.patch")
```

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+
+package () {
+    # Git wants to decide itself whether to use ANSI stdio emulation or not
+    CPPFLAGS="$(echo "$CPPFLAGS" |
+        sed 's/-D__USE_MINGW_ANSI_STDIO\(=[0-9]*\)\?//')"
+
+    # run cmake first, it will generate GIT-VERSION-FILE
+    mkdir -p $srcdir/build
+    cd $srcdir/build
+    MSYS2_ARG_CONV_EXCL="-DFALLBACK_RUNTIME_PREFIX=" ${MINGW_PREFIX}/bin/cmake.exe $srcdir/$_realname-$pkgver/contrib/buildsystems \
```
MehdiChinoune, line 94:
```suggestion
    MSYS2_ARG_CONV_EXCL="-DFALLBACK_RUNTIME_PREFIX=" \
    ${MINGW_PREFIX}/bin/cmake.exe \
    $srcdir/$_realname-$pkgver/contrib/buildsystems \
```

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+}
+
+package () {
+    # Git wants to decide itself whether to use ANSI stdio emulation or not
+    CPPFLAGS="$(echo "$CPPFLAGS" |
+        sed 's/-D__USE_MINGW_ANSI_STDIO\(=[0-9]*\)\?//')"
+
+    # run cmake first, it will generate GIT-VERSION-FILE
+    mkdir -p $srcdir/build
+    cd $srcdir/build
```
MehdiChinoune, line 93:
```suggestion
    mkdir -p $srcdir/build-${MSYSTEM}
    cd $srcdir/build-${MSYSTEM}
```

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+    MSYS2_ARG_CONV_EXCL="-DFALLBACK_RUNTIME_PREFIX=" ${MINGW_PREFIX}/bin/cmake.exe $srcdir/$_realname-$pkgver/contrib/buildsystems \
+        -GNinja \
+        -DUSE_VCPKG=off \
+        -DSKIP_DASHED_BUILT_INS=ON \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX="$pkgdir/$MINGW_PREFIX" \
+        -DFALLBACK_RUNTIME_PREFIX="$MINGW_PREFIX"
+    ${MINGW_PREFIX}/bin/ninja.exe install
```
MehdiChinoune, line 101:
```suggestion
    ${MINGW_PREFIX}/bin/cmake --install .
```

mingw-w64-git/PKGBUILD
Berrysoft: Then I need to change several other packages because `git-lfs`, `delta` etc. now depends on MSYS2 git, but they could depend on MinGW git.

mingw-w64-git/PKGBUILD
Berrysoft: OK, but I need the MSYS2 `docbook-xsl` to build docs.

mingw-w64-git/PKGBUILD
Berrysoft: They may exist if an old user uses `git-for-windows` pacman registry to install MinGW git.

Berrysoft:
Some questions:

* Should I change the dependencies of `git-lfs` and `delta` or just write `provides=("git")` here?
* Why CLANG32 in the CI doesn't build this package :)

mingw-w64-git/PKGBUILD
Berrysoft: Should I consider that possibility? There should be many users did that.

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,136 @@
+# Maintainer: Yuyi Wang <Strawberry_Str@hotmail.com>
+
+_realname=git
+pkgbase="mingw-w64-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=2.39.2.windows.1
+pkgrel=1
+pkgdesc="The fast distributed version control system (mingw-w64)"
+arch=('any')
```
lazka, line 9:
```suggestion
arch=('any')
mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
```

wszqkzqk: > 

Also `'clangarm64'`

Berrysoft: Added.

lazka: (#16091 pointed out that there might be some extra patching needed)

Berrysoft: I've added `7-arm64.patch` to disable nedmalloc for arm64.

Berrysoft:
Also note for CLANGARM64: I don't have devices to check if it really works. I just *hope* it would work well.

mingw-w64-git/PKGBUILD
wszqkzqk: > Then I need to change several other packages because `git-lfs`, `delta` etc. now depends on MSYS2 git, but they could depend on MinGW git.

The main problem is that many packages depend on `git` as `makedepends`, so you may need to change all of then:(

mingw-w64-git/PKGBUILD
wszqkzqk: > Some questions:
> 
>     * [ ]  Should I change the dependencies of `git-lfs` and `delta` or just write `provides=("git")` here?
> 
>     * [x]  Why CLANG32 in the CI doesn't build this package :)
> 
>     * [ ]  Should I enable CLANGARM64? It should build successfully but I cannot check if it works well.

It seems that the CI of `clangarm64` will not run here.

wszqkzqk:
> Some questions:
> 
>     * [ ]  Should I change the dependencies of `git-lfs` and `delta` or just write `provides=("git")` here?
> 
>     * [x]  Why CLANG32 in the CI doesn't build this package :)
> 
>     * [ ]  Should I enable CLANGARM64? It should build successfully but I cannot check if it works well.

It seems that the CI of `clangarm64` will not run here.

lazka:
> * Should I change the dependencies of `git-lfs` and `delta` or just write `provides=("git")` here?

I'd say the former. Or just remove the provides and don't change any dependencies until we are more confident it works for all our use cases.

mingw-w64-git/PKGBUILD
lazka: Maybe add a comment above the replaces explaining why they are there.

mingw-w64-git/PKGBUILD
Berrysoft: Well, not all. Those packages that use `git` as a make dependency, should still use `git` because `makepkg` is a bash script.

lazka: In theory there is no reason why they couldn't switch to mingw git.

Berrysoft:
> I'd say the former. Or just remove the provides and don't change any dependencies until we are more confident it works for all our use cases.

I'd like to give users opportunities to choose their prefer `git`. However, it seems that the only way to give choices is to add a `provides` entry... In this case, pacman will ask the user which git they want to install together with `git-lfs`.

mingw-w64-git/PKGBUILD
Berrysoft: Comments added.

wszqkzqk:
> Another solution is to set both `git` and `mingw-w64-git` as optional dependencies for these packages.

But this allows to install none of them.

mingw-w64-git/PKGBUILD
Berrysoft: It's the maintainers' turn to choose which git to use. I don't need to bother them in this PR.

Berrysoft:
> But this allows to install none of them.

Yes. So I choose to add `provides`.

ognevnydemon:
what about to create new package group "mingw-w64-VCS"?

mingw-w64-git/PKGBUILD
wszqkzqk: > It's the maintainers' turn to choose which git to use. I don't need to bother them in this PR.

It seems that msys2 does not have a complex and hierarchical team structure like Arch Linux (which has core team and TU), so the dependencies of most packages are determined by the official.

Berrysoft:
> what about to create new package group "mingw-w64-VCS"?

Well, I'd like to, but it will only contain `mingw-w64-git` :)

ognevnydemon:
> Well, I'd like to, but it will only contain mingw-w64-git :)

so we should make all other packages from VCS to mingw-w64 :D

Berrysoft:
It seems that the code moving from `package()` to `build()` breaks some paths. I'll fix it immediately:)

mingw-w64-git/PKGBUILD
Berrysoft: OK then it's official's turn:) I don't think it is easy to switch msys2 git to mingw git as a make dep. It needs support from makepkg. I would rather not changing them. What would I like to change is the git tools that depends on git only because a user cannot use them directly. A mingw git will work well without large change.

Berrysoft:
I've come up with another reason to set `provides`: there are packages like `git-extras` that only provides shell scripts for git. Both `git` and `mingw-w64-git` could work well with them. There will be some cases that a user would like to install these scripts but don't want another `git` installed. It's true that we can provide another packages called `mingw-w64-git-extras`, but it seems redundant.

mingw-w64-git/PKGBUILD
lazka: Yes, that can all wait..

MehdiChinoune:
Makig mingw-w64-git provides will make it impossible to coinstall it from different environments, you could only have one git.

Berrysoft:
> Makig mingw-w64-git provides will make it impossible to coinstall it from different environments, you could only have one git.

Not exactly. According to arch wiki here: https://wiki.archlinux.org/title/PKGBUILD#provides

> Packages providing the same item can be installed side-by-side, unless at least one of them uses a `conflicts` array.

Therefore you can have several git with same `provides`.

MehdiChinoune:
Let's clarify, when someone uses msys2 environment and want to use git, pacman won't install the msys one if mingw-w64-git from one of the environments is already installed. That's what provides is for.
mingw-w64-git won't work perfectly with msys2 apps/libs.

Berrysoft:
> Let's clarify, when someone uses msys2 environment and want to use git, pacman won't install the msys one if mingw-w64-git from one of the environments is already installed.

Yes, but if the users want the msys2 git, they can install themselves.

> mingw-w64-git won't work perfectly with msys2 apps/libs.

We need mingw git because it works perfectly with native win32 apps/libs. I wonder the definition of "work perfectly". Of course a mingw git cannot recognize POSIX paths. However, as for the git tools, for example, `git-lfs`, `git-extras`, `delta`, it works well the same as the msys2 git. It *provides* full git functionality, but is not a *replacement* of msys2 git. That's why I choose `provides` instead of `conflicts` or `replaces`.

mmuetzel:
If I understand @MehdiChinoune's point correctly, the MSYS2 `git` won't be installed automatically by `makepkg` if any of the "native" `git` packages is already installed. That would be an issue if someone wanted to build a package from `MSYS2-packages` that contains `git` in its `makedepends`. Or it could be an issues on installation of a package from `MSYS2-packages` (with `pacman`) if said package contains `git` in its `depends`. A user could still install the MSYS2 `git` package manually. But the required package might no longer be installed automatically if another package that claims to provide it is already installed.

Berrysoft:
OK, I understand your point, @MehdiChinoune . I'll consider another solution for that.

dscho:
> @dscho would you like to review this pull request?

Quite honestly, no. I would have loved to collaborate on a solution that works _with_ Git for Windows, but this PR seems specifically _not_ to want to collaborate with the Git for Windows project, otherwise I would have heard about this a lot sooner than being at-mentioned somewhere in the middle of this PR. And it's not like this PR couldn't have used the collaboration.

It's all of the little things like using CMake (which should _not_ be used outside of Visual Studio to build Git, instead the standard `make.exe` procedure should be used _directly_) to totally shunning the work on quoting arguments passed to `sh.exe` (most likely regressing on the fixes in git/git@9e9da23c2765050ff30d34540fbab62a1b4e5d01), to _not_ contributing [`4-headless.patch`](https://github.com/msys2/MINGW-packages/pull/16099/files#diff-9b031864cb4b6b415a60792a4a90328bfcfc11526eaa0e50cbd776bc7d0da28b) to the Git for Windows project, to the lack of attention to the Git test suite (which will fail, due to the behavior of `msys-2.0.dll`). Essentially, this PR tries to establish a different build of `mingw-w64-git` instead of joining forces.

I have enough friction in my life and am in no need for more, and this PR seems designed to add to it.

dscho:
> when someone uses msys2 environment and want to use git, pacman won't install the msys one if mingw-w64-git from one of the environments is already installed.

`makepkg` has hard-coded `/usr/bin/git` in it, and for good reason. A standard practice when having `git+https://` style `source`s is to first make a bare clone and then making a full clone in `src/` using `--reference=<bare-clone>`. This `--reference` adds absolute paths to `src/<name>/.git/objects/info/alternates`, which are Unix paths for `/usr/bin/git` that `/mingw64/bin/git` won't understand.

And that's just one of the discrepancies I run into on a regular basis.

Berrysoft:
@dscho I'm sorry if you feel unhappy about my behavior, but this PR contains a patch you have just refused half a year ago. Both git and msys2 are open source project, right? I'm here contributing them, not having frictions with others.

Patch 4 is a small patch that I believe is simply a typo. I contributed Patch 3 and patch 5, which are not in this PR, to the git upstream. If you care about that so much, I'll also contribute path 4 to git-for-windows later.

Cmake is a powerful tool. Although the cmake buildsystem is built for VS, it works well with mingw env, and even with Linux. I'm using cmake here because I'd like to provide builds for all 6 mingw env in msys2.

As for the tests, let's see what will happen next:) If they are really urgent, it is a good chance to port some changes of msys2 runtime from git-for-windows. Up to now, I haven't found any fatal errors.

dscho:
The Git project is quite clear on the fact that the `Makefile` build is the one that is officially supported, with CMake support only added for Visual Studio's benefit. And the CMake support is now explicitly maintained in the Git for Windows, i.e. by me. And I'm telling you that it's better to use `make`. So...

About the patch I rejected, I don't remember, but I wonder why it's better to stuff it into the MSYS2 project than to either convince the Git for Windows reviewers, or improve the patch, or to drop it. There must have been a reason why I rejected it.

Berrysoft:
I wonder how to use make to build for UCRT64, CLANG32 and CLANG64. They are not handled by Makefile.

The PR related to the patch see here: https://github.com/git-for-windows/git/pull/3869 which I've mentioned in #16091 . In short words, git-for-windows project doesn't need this patch because all tools are bundled together, and `cmd/git.exe` will prepend the paths to the environment variable `PATH` every time it executes the real git binary. However, as I've said in the PR and the issue, we need it when installing mingw git in MSYS2 because in that case you cannot control the user which perl to install.

Berrysoft:
I've run the tests on my device. It costs a lot time. They need about 3 quarters with 16 threads. Maybe I shouldn't add them to PKGBUILD :)

I've looked through the log briefly. It seems that some tests failed because of some miracle reasons, for example, the wrong expansion of bash. I cannot check all of them, but that's should be what @dscho have said - due to the behavior of `msys-2.0.dll`.

Anyway, I put the test results below. Most tests passed.

```
95% tests passed, 54 tests failed out of 988

Total Test time (real) = 2725.78 sec

The following tests FAILED:
          2 - t0001-init.sh (Failed)
          4 - t0003-attributes.sh (Failed)
         43 - t0060-path-utils.sh (Failed)
         44 - t0061-run-command.sh (Failed)
         51 - t0068-for-each-repo.sh (Failed)
         66 - t0204-gettext-reencode-sanity.sh (Failed)
        100 - t1091-sparse-checkout-builtin.sh (Failed)
        101 - t1092-sparse-checkout-compatibility.sh (Failed)
        117 - t1402-check-ref-format.sh (Failed)
        138 - t1501-work-tree.sh (Failed)
        141 - t1504-ceiling-dirs.sh (Failed)
        143 - t1506-rev-parse-diagnosis.sh (Failed)
        145 - t1508-at-combinations.sh (Failed)
        150 - t1513-rev-parse-prefix.sh (Failed)
        157 - t1800-hook.sh (Failed)
        190 - t2070-restore.sh (Failed)
        221 - t2501-cwd-empty.sh (Failed)
        223 - t3001-ls-files-others-exclude.sh (Failed)
        240 - t3070-wildmatch.sh (Failed)
        247 - t3200-branch.sh (Failed)
        270 - t3400-rebase.sh (Failed)
        274 - t3404-rebase-interactive.sh (Failed)
        325 - t3701-add-interactive.sh (Timeout)
        327 - t3703-add-magic-pathspec.sh (Failed)
        334 - t3903-stash.sh (Failed)
        336 - t3905-stash-include-untracked.sh (Failed)
        461 - t4202-log.sh (Failed)
        467 - t4208-log-magic-pathspec.sh (Failed)
        470 - t4211-line-log.sh (Failed)
        538 - t5500-fetch-pack.sh (Failed)
        553 - t5516-fetch-push.sh (Failed)
        603 - t5580-unc-paths.sh (Failed)
        607 - t5601-clone.sh (Failed)
        608 - t5602-clone-remote-exec.sh (Failed)
        609 - t5603-clone-dirname.sh (Failed)
        621 - t5615-alternate-env.sh (Failed)
        633 - t5801-remote-helpers.sh (Failed)
        638 - t5813-proto-disable-ssh.sh (Failed)
        659 - t6018-rev-list-glob.sh (Failed)
        678 - t6130-pathspec-noglob.sh (Failed)
        680 - t6132-pathspec-exclude.sh (Failed)
        681 - t6133-pathspec-rev-dwim.sh (Failed)
        760 - t7201-co.sh (Failed)
        763 - t7400-submodule-basic.sh (Failed)
        830 - t7810-grep.sh (Failed)
        837 - t7817-grep-sparse-checkout.sh (Failed)
        838 - t7900-maintenance.sh (Failed)
        853 - t9001-send-email.sh (Failed)
        929 - t9300-fast-import.sh (Failed)
        933 - t9304-fast-import-marks.sh (Failed)
        947 - t9700-perl-git.sh (Failed)
        985 - t9850-shell.sh (Failed)
        987 - t9902-completion.sh (Failed)
        988 - t9903-bash-prompt.sh (Failed)
```

hmartinez82:
I can help with the CLANGARM64 testing

mingw-w64-git/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+build() {
+    # Git wants to decide itself whether to use ANSI stdio emulation or not
+    CPPFLAGS="$(echo "$CPPFLAGS" |
+        sed 's/-D__USE_MINGW_ANSI_STDIO\(=[0-9]*\)\?//')"
+
+    # run cmake first, it will generate GIT-VERSION-FILE
+    mkdir -p $srcdir/build-${MSYSTEM}
+    cd $srcdir/build-${MSYSTEM}
+    MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=;-DFALLBACK_RUNTIME_PREFIX=" \
+        ${MINGW_PREFIX}/bin/cmake.exe \
```
dscho, line 88: Using `cmake` to build Git is still as wrong as when I pointed it out the first time.

dscho: It is a bit unprofessional to call a review ridiculous.

The reason why I pointed out that using `cmake` is wrong here is that the Git project specifically focuses on adding new features to the `Makefile`. _Some_ of those features are then also supported in the CMake file, and the focus in that build process lies squarely with making it work well inside Visual Studio.

Since you most likely want to be able to use all of Git's latest-and-greatest features, and since you don't want to use Visual Studio to compile Git, I see only good reasons to use `make` to build Git in this PR's context.

Insisting on using anything but the paved path is simply introducing unnecessary friction.

dscho: BTW all of these unnecessary distractions cost time, time that would be better spent to prepare the MSYS2 runtime to support Git better, to figure out how to support Git for Windows' `http.sslBackend`, and thinking how to create synergies between the mingw-w64-git work and the Git for Windows project instead of spending that time on antagonizing and on making things intentionally incompatible.

Berrysoft: The `Makefile` doesn't handle all possible `MSYSTEM` values.

It's the third time I've repeated this line. You gave me a lot of reason of *why* I should use Makefile, but I was *not able to* use it. I choose CMake not because I like CMake, but because I haven't other choices.

The `Makefile` in git-for-windows only handles `MINGW32`, `MINGW64`, and `CLANGARM64`. We also need `UCRT64`, `CLANG32` and `CLANG64`. I've tried two times to fix it, but failed all the time. That's why I choose to use CMake.

hmartinez82:
If CLANGARM64 is already supported it probably means that most of headaches for a UCRT based runtime is probably already being addressed, at least in the `main` branch. The work for getting the CLANGARM64 build has been ongoing for over a year. There are many commits, but you can look https://github.com/git-for-windows/git/pull/3916/files and its peer commits associated with https://github.com/git-for-windows/git/issues/3107

Berrysoft:
Well, I'll try again to use Makefile when I'm free.

mingw-w64-git/PKGBUILD
Berrysoft: Yes, I now change to only provide `git-core`. Then it won't break the existing scripts and worktrees.

dscho: But you cause confusion because it is not at all clear why `git` _and_ `git-core` exist.

dscho:
> I've run the tests on my device. It costs a lot time. They need about 3 quarters with 16 threads

If you would have cared to ask, I could have pointed you to the way this is done in Git for Windows: by [parallelizing the test runs](https://github.com/git-for-windows/git/blob/v2.40.0.windows.1/.github/workflows/main.yml#L113-L146). And yes, I agree that this is impractical for the `PKGBUILD`.

Does that mean, however, that you should feel no need at all to run the tests?

> Anyway, I put the test results below. Most tests passed.

"Look, there were many bullets flying around, but it's all fine, most of them passed without doing any harm."

I am quite shocked about this, as you can tell. "Most tests passed"? Is that the entirety of the analysis?

> ```
> 95% tests passed, 54 tests failed out of 988
> ```

There is a reason why Git for Windows' policy is to never accept any PR with failed Checks. Those tests are there _for a reason_, you know?

mingw-w64-git/PKGBUILD
dscho: > The `Makefile` doesn't handle all possible `MSYSTEM` values.

Oh, so the `Makefile` is set in stone?

If you already figured out that the `MSYSTEM` does not handle all the values, how about imitating what others have done and fix those issues?

There is a reason why commits like https://github.com/git-for-windows/git/commit/a19211c0c2e30bb00711d257c321a0ad343f22a7 exist, and that reason is that some contributors actually collaborate instead of causing unnecessary friction.

Berrysoft:
I now know that it is far more complicated to add a MinGW git. Not only because of technical reasons, but also social reasons. I hope there will be some other much more skilled and *professional* contributors trying on this project. I believe they are OK using Makefile instead of CMake, and can fix the official MSYS2 runtime to pass *all* tests.

It's my fault trying to contribute to this project. I wanted to make everyone happy and satisfied, but it seems that the maintainer of git-for-windows isn't happy at all. I'll close this PR because I'm a bit tired about it, and also because I'm not going to cause more *frictions*.

PR #16100, opened 2023-03-08
qemu: enable clangarm64

driver1998:


Biswa96:
Q. Is it required to save those patches in this repository? Those are already downloaded from upstream commit.

driver1998:
I just left those by accident, removed now.

PR #16121, opened 2023-03-09
octave: update to 8.1.0

mmuetzel:
Also remove some upstream patches that are already part of the new version.


MehdiChinoune:
Could backport/prepare changes to fix building with clang/flang.

mmuetzel:
> Could backport/prepare changes to fix building with clang/flang.

I didn't come around to cleaning up and backporting the patches that I have on top of the development branch of Octave (which is already ahead of the release branch). I also didn't come around to checking whether my local changes would break compilation with gcc/gfortran. I'll try to do that (maybe during the weekend).
It would be nice if those additional changes wouldn't block the update.

mmuetzel:
I downloaded and installed the MINGW64 build artifact. There are no new failing tests when running Octave's test suite.

Is it ok to merge as it is?

MehdiChinoune:
I still prefer to backport changes to fix building with llvm toolchain, you could apply them only CLANG environments if you are not sure if they break building with GCC toolchain or not:
```
  patch ...
fi
```

mmuetzel:
> I still prefer to backport changes to fix building with llvm toolchain

Like I already wrote, the earliest I might be able to do that is during the weekend.
And even then, I'd vote against distributing binaries for those environments (because they are crashing for many of the basic operations).

MehdiChinoune:
> And even then, I'd vote against distributing binaries for those environments.

I asked to be able to build octave locally! Is that too much?

mmuetzel:
> I asked to be able to build octave locally! Is that too much?

I answered: Likely earliest on the weekend. I have a day job.

mmuetzel:
I'd prefer to open a draft PR for the changes that are required to build with LLVM. It might require some tinkering to get it right.

Could we merge this PR so I could base that draft PR on the changes here?

MehdiChinoune:
I don't have any objection.

mmuetzel:
I opened PR #16156 for building Octave with LLVM 16.

PR #16125, opened 2023-03-09
verilator: update to 5.012

umarcor:
Verilator 5.008 was released: https://github.com/verilator/verilator-announce/issues/60

### DIFF
--- a/mingw-w64-verilator/PKGBUILD
+++ b/mingw-w64-verilator/PKGBUILD
@@ -7,7 +7,8 @@ pkgver=5.012
 pkgrel=1
 pkgdesc="The fastest free Verilog HDL simulator (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
> LLVM ERROR: out of memory

suggests this release maybe isn't buildable for 32bit anymore?

umarcor:
Since Verilator 5.012 was released (https://github.com/verilator/verilator-announce/issues/62), I updated this PR.

@lazka it seems to work on MINGW32 but not on CLANG32. Maybe CLANG needs more memory than available in GitHub runners?

lazka:
> @lazka it seems to work on MINGW32 but not on CLANG32. Maybe CLANG needs more memory than available in GitHub runners?

the limit would be the 32bit user space address space (3GB afair) and not the runner RAM (8GB)

PR #16153, opened 2023-03-11
universal-ctags: update to 6.0.0, remove git suffix

Biswa96:


raedrizqie:
isnt this the same package as this one?
https://packages.msys2.org/base/mingw-w64-ctags

Biswa96:
Not actually same. The ctags repository is old and not maintained. This universal-ctags is maintained till now and has many features. I guess, there are two ctags packages to provide a choice to users.

I came to know about this package while poking into pulseaudio project. The old ctags does not produce any output but the new universal-ctags does. Maybe there are some differences with regex like rules between the old and new ctags.

raedrizqie:
i see.. i was wondering earlier because the link to arch's package are the same..

PR #16155, opened 2023-03-11
Quazip static

tim-lebedkov:
this builds QuaZip statically for Qt 5


tim-lebedkov:
How long does it normally take to review a pull request?

lazka:
just some general questions, maybe for other devs.

Why only for the qt5 build and not the qt6 one?
Do we want to split the static builds, we don't do that usually?

tim-lebedkov:
I just don't need the Qt 6 build. Somehow there is a difference how QuaZIP is built: statically or not.

PR #16163, opened 2023-03-12
octave: build against pcre2

MehdiChinoune:


mmuetzel:
LGTM.
While you are at it, could you please also remove `ac_cv_search_tputs=-ltermcap` from the configure flags? I missed that in the PR for the version update.

PR #16170, opened 2023-03-12
julia: Update version to 1.9.0 and redesign package

stahta01:
I have finished the package and I downloaded all three copies from CI server and the install and work enough to open and close with no error.

Tim S.


mmuetzel:
I don't know how to use `julia`. But the MINGW64 build artifact from this PR seems to start correctly:
```
$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.0-rc1
 _/ |\__'_|_|_|\__'_|  |
|__/                   |

julia>
```

mmuetzel:
UCRT64 failed with:
```
/bin/sh: line 1: rsync: command not found
  make[1]: *** [/C/_/B/src/build-UCRT64/deps/llvm.mk:277: /C/_/B/src/build-UCRT64/usr-staging/llvm/build_Release.tar] Error 127
  make: *** [Makefile:62: julia-deps] Error 2
      Aborting...
```

Afaict, that is part of the [`rsync`](https://packages.msys2.org/package/rsync?repo=msys&variant=x86_64) package:
```
$ pacman -Fx rsync.exe
msys/rsync 3.2.7-2 (net-utils) [Installiert]
    usr/bin/rsync.exe
```
Is it necessary to add it to the `makedepends`? But then, why isn't it needed for MINGW64 or MINGW32?

mmuetzel:
It's a bit strange that it is installed in its own installation tree with its own set of bundled headers and libraries (which are potentially conflicting if users include both the default and Julia's include or lib folders). That looks a bit like it doesn't really "belong" into MSYS2.
Is there a reason for that?

stahta01:
> It's a bit strange that it is installed in its own installation tree with its own set of bundled headers and libraries (which are potentially conflicting if users include both the default and Julia's include or lib folders). That looks a bit like it doesn't really "belong" into MSYS2. Is there a reason for that?

It a lot of ways I say it does not belong; but, this build method is the preferred Julie build method.
I need to build UCRT64 from source using the non preferred option of even using the System CSL Compiler Support Libs.

Tim S.

stahta01:
> Is it necessary to add it to the `makedepends`? But then, why isn't it needed for MINGW64 or MINGW32?

I forgot about rsync I edited the packaged again likely fail because I decided to use server to see if one of the packages was needed.

If the power that be wish to drop Julia or wish to fork my change more in keeping with MSys2 MinGW 64 methods so be it.
But, someone said okayed last time I used the separate /opt installation. I have learned enough to likely fix it closer to normal method now. But, I would suggest moving the project to development because it is not a package that is easy to maintain if not mainly done using the Julia preferred method. I am thinking the UCRT64 build is at risk of timing out.

Tim S.

stahta01:
> I don't know how to use `julia`. But the MINGW64 build artifact from this PR seems to start correctly:
> 
> ```
> $ julia
>                _
>    _       _ _(_)_     |  Documentation: https://docs.julialang.org
>   (_)     | (_) (_)    |
>    _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
>   | | | | | | |/ _` |  |
>   | | |_| | | | (_| |  |  Version 1.9.0-rc1
>  _/ |\__'_|_|_|\__'_|  |
> |__/                   |
> 
> julia>
> ```

Thank for the test; so, the binary blobs style build works for MINGW64 and likely MINGW32.

Tim S.

mmuetzel:
Imho, the non-default installation prefix is odd, but not as much of a potential issue as the duplicate (potentially conflicting) dependencies at that prefix.
Do I understand correctly, that option 2 would mean that julia in MSYS2 would be built using the existing binary packages as dependencies (instead of building/downloading their own [conflicting?] dependencies)?
If that is true, it would be best imho to go with that option and to just drop MINGW32.

But please remember that this is just my personal opinion given the limited info I have about the issue. If there are good reasons why that is problematic, please let us know.

stahta01:
> Imho, the non-default installation prefix is odd, but not as much of a potential issue as the duplicate (potentially conflicting) dependencies at that prefix. Do I understand correctly, that option 2 would mean that julia in MSYS2 would be built using the existing binary packages as dependencies (instead of building/downloading their own [conflicting?] dependencies)? If that is true, it would be best imho to go with that option and to just drop MINGW32.
> 
> But please remember that this is just my personal opinion given the limited info I have about the issue. If there are good reasons why that is problematic, please let us know.

The run-time problems reported using system packages in standard install locations will likely be ignored by the Julia team till we duplicate them using an non using system package build. We could do that using Cygwin or build an local MINGW64 without System packages in /opt location.

mmuetzel:
What does "system package" mean in this case?

stahta01:
> What does "system package" mean in this case?

An MSys2 MINGW package.

Tim S.

mmuetzel:
I'm sorry. I'm still struggling to get the point. What are the run-time problems you are referring to?

mmuetzel:
If the julia team explicitly does not want that we use the binary packages from MSYS2 for a `julia` package in MSYS2, I think we shouldn't distribute it. There are other tools (like chocolatey) that would probably do a better job at "just installing" that software.
I've always understood MSYS2 as a kind of distribution where all packages (ideally) work together. Julia requiring their separate set of (conflicting) dependencies doesn't really fit MSYS2 imho.

mmuetzel:
I'm not sure I'm reading the current package build rules correctly.
But aren't the rules for UCRT64 already what would basically be a "proper" MSYS2? Just skip copying all the libraries and install at the default prefix.
Would that work?

stahta01:
> I'm not sure I'm reading the current package build rules correctly. But aren't the rules for UCRT64 already what would basically be a "proper" MSYS2 package? Just skip copying all the libraries and install at the default prefix. Would that work?

The UCRT64 is not following Julia rules; they say the will not always look at issues cause by using USE_SYSTEM in the Make.user file. To build UCRT64 I need to use USE_SYSTEM (MSys2 MINGW Packages). I can also do the same with MINGW64.

I can build MINGW64 using either Julia approved methods or standard MSys2 MINGW methods; I can not do both.

Tim S.

stahta01:
I am now getting more Julia check failures; so, I will need to work on at least some of them.
The mpfr test failure at least gives me an idea where it is located.

Tim S.

mmuetzel:
> I can build MINGW64 using either Julia approved methods or standard MSys2 MINGW methods; I can not do both.

I understand that. There are several other packages in MSYS2 that aren't directly supported by upstream. If it becomes too much of a hassle to maintain them as part of MSYS2, we could always just remove them.

Afaict, the last version installs at the default prefix. That is a step in the correct direction imho. However, it still includes a couple of bundled dependencies that conflict with the packages from MSYS2. E.g.: SuiteSparse, PCRE2, zlib, ...

Would it be possible to use the existing binaries instead of bundling (conflicting) versions of those libraries?
If the bundled versions are built with patches to make them work correctly with Julia, would it make sense to downstream those patches to the packages in MSYS2?

stahta01:
> > I can build MINGW64 using either Julia approved methods or standard MSys2 MINGW methods; I can not do both.
> 
> I understand that. There are several other packages in MSYS2 that aren't directly supported by upstream. If it becomes too much of a hassle to maintain them as part of MSYS2, we could always just remove them.
> 
> Afaict, the last version installs at the default prefix. That is a step in the correct direction imho. However, it still includes a couple of bundled dependencies that conflict with the packages from MSYS2. E.g.: SuiteSparse, PCRE2, zlib, ...
> 
> Would it be possible to use the existing binaries instead of bundling (conflicting) versions of those libraries? If the bundled versions are built with patches to make them work correctly with Julia, would it make sense to downstream those patches to the packages in MSYS2?

Maybe, but, I already have check/test failures that need fixed. Till I figure out why MSys2 MINGW mpfr is failing test adding more MSys2 packages make no sense. IIRC, SuiteSparse is the wrong version as in Julia uses an older version.
PCRE2 and ZLIB are likely able to be changed.

Tim S.

stahta01:
I have decided to change directions after two of the GMP patches from Julia failed to fix the MPFR test failure; they did fix the numbers test failure. I am going to use the Julia source build of GMP and MPFR with an edit to rename the DLL to something like "mpfrjl.dll". Likely be a few weeks before I get it done depends on how real life events proceed.

Tim S.

Vonfry:
Is this pr still active?

stahta01:
Does anyone use Julia and are they willing to give feedback?

Tim S.

stahta01:
I have updated the package to 1.9.0 release version and MINGW32, MINGW64, and UCRT64 now builds.
MINGW32 uses non MSys2 standard installation location. And, it is likely the MSys2 team will not like it.
But, the 32 bit version of Julia uses SJLJ and I was afraid it would have weird bugs doing it any other way.

Tim S.

mmuetzel:
I downloaded the MINGW64 build artifact from the latest CI run.
It installs and runs fine afaict. But I don't know Julia. So my testing was very limited.

Also, the way it now installs within the respective prefix makes it feel like it actually "belongs" to MSYS2. 👍

stahta01:
> I downloaded the MINGW64 build artifact from the latest CI run. It installs and runs fine afaict. But I don't know Julia. So my testing was very limited.
> 
> Also, the way it now installs within the respective prefix makes it feel like it actually "belongs" to MSYS2. 👍

Thank you for the feedback. 
I just pushed an commit to disable the HTML doc where it errored out on MINGW64. If it works and I can download all 3 packages I will combine my commits to a single commit and remove the WIP from title.

Tim S.

mmuetzel:
> I just pushed an commit to disable the HTML doc where it errored out on MINGW64.

If I understood the log correctly, it didn't error out. But the CI happened to chose a particularly slow runner and the build timed out. When I restarted the CI and a different (faster) runner was picked, the CI went through without issues.

If having the HTML documentation seems valuable, you might consider reverting the last change.

stahta01:
> > I just pushed an commit to disable the HTML doc where it errored out on MINGW64.
> 
> If I understood the log correctly, it didn't error out. But the CI happened to chose a particularly slow runner and the build timed out. When I restarted the CI and a different (faster) runner was picked, the CI went through without issues.
> 
> If having the HTML documentation seems valuable, you might consider reverting the last change.
> 
> Edit: For the successful CI run before the latest change, see: https://github.com/msys2/MINGW-packages/actions/runs/5100555105

I currently am not building the docs split package and in the past the docs was disabled for a unknown reason.
So, disabling the docs to help the build finish is not an negative to me. Julia built in help will likely still work and that along with the Julia online help should be enough. Also, I am thinking the HTML help was likely broken because of all the warning it gives.

Tim S.

goyalyashpal:
> Changed title to [WIP] while I test it.
\- @ stahta01 at https://github.com/msys2/MINGW-packages/pull/16170#issuecomment-1554820569

tip: github suggests using `Draft: ` title prefix, also, if u use github-web, then there's a button there to mark PRs as draft:

stahta01:
> > Changed title to [WIP] while I test it.
> > - @ stahta01 at [#16170 (comment)](https://github.com/msys2/MINGW-packages/pull/16170#issuecomment-1554820569)
> 
> tip: if u use github-web, then there's a button there to mark PRs as draft. it saves from changing the title which otherwise causes disruption in email notifications
> 
> 

How do you change it back to draft?

I never saw a way to do that; since it takes weeks for a response to happen because there seems to be a small number of people who really care about this package.

Note: There does seem to be more people who are willing to complain about it!

But, only 2 or at most 3 who are willing to help.

Tim S.

stahta01:
Link to [artifacts](https://github.com/msys2/MINGW-packages/actions/runs/5210586445#artifacts)

ognevnydemon:
is there any chance to update package then?

stahta01:
I gave up on this package I waited too long for little return.

PR #16177, opened 2023-03-12
Add new package: bcftools

teepean:
Bcftools is related to htslib and samtools. PKGBUILD copied from mingw-w64-samtools.


teepean (commented):
32-bit builds hang during make check as well.

PR #16178, opened 2023-03-12
i2pd: update to 2.47.0

r4sas:
Note: we have added cmake support, but there is no install target for the configs and certificates, that's why `install` done manually.

Patch `0002-cmake-nonstatic.patch` included directly from repository and can be removed in next release.


Biswa96:
Is the mingw-w64-i2pd/0001-i2pd-makefile.patch still required after porting to cmake?

r4sas:
> Is the mingw-w64-i2pd/0001-i2pd-makefile.patch still required after porting to cmake?

Only if the package requires the ability to build with make as a fallback.

PR #16180, opened 2023-03-13
glib2: update to v2.76.0

wszqkzqk:


lazka:
There is some fallout:

* https://gitlab.gnome.org/GNOME/glib/-/issues/2937
* https://gitlab.gnome.org/GNOME/glib/-/issues/2936

so I think we should wait a bit

PR #16205, opened 2023-03-14
tree-sitter: Update to 0.20.8

podsvirov:


podsvirov:
The [issue](https://github.com/tree-sitter/tree-sitter/issues/2125) about the `v0.20.8` tag adventure.

lazka:
thanks, reverted for now

PR #16207, opened 2023-03-14
rust: set _bootstrapping=no for all architectures

filnet:


filnet:
It's experimental. Please don't merge.

filnet:
The `bootstrapping=no` mode will tell the Rust build to use the local rust.
The rust package must therefore be added to `makedepends`.

We know this mode works for aarch64 and here I am testing it on other architectures.

The `bootstrapping=yes` mode lets the build proceed by default.
It will download and use the latest official release of rust (which does not exist for aarch64?).

The big issue with using the local rust is that it must be the latest.
This means we can't skip a beat and must produce all versions...

filnet:
So it is "eat our own dog food" or "play it safe".
aarch64 being in the former category.

filnet:
Are build time stable/reproducible or random ?

Biswa96:
0013-revert-lib-host-symlink.patch file was not removed.

filnet:
Forgot to push the upgrade to 1.69.0 (was still on the pre-release).

ognevny:
I believe that this PR is non-relevant anymore. starting with v1.91.0 we can fully switch to _bootstrapping=yes in all environments since stage0 will be available for *-gnullvm hosts too

PR #16216, opened 2023-03-15
qemu: update to 8.0.0

heljkon:
starting with first build on 8.0.0-rc0

the patches for clangarm64 are already included in source code of 8.0.0-rc and thus removed from build


heljkon:
@lazka @Biswa96 @jeremyd2019 
Please have a look!

PR #16218, opened 2023-03-15
python-wheel: remove usage of vendored packaging.

MehdiChinoune:


lazka:
hm, why is this needed?

MehdiChinoune:
> 

The usage of vendored packaging was removed before this PR, but python-wheel started using it in more locations so I have just fixed that.
There was a packaging failure this morning of botocore in msys-autobuild.

lazka:
hm, I see. that was e9206bfb3bb91cfb3ff23e0

I myself wouldn't bother unbundling things, but ymmv.

PR #16239, opened 2023-03-16
mold: update to 1.11.0

3rav:


3rav:
> Just wondering, do you really use this package? How?

Not at the moment, but I see a lot of potential for the future.

PR #16240, opened 2023-03-16
trilinos: enable Kokkos

3rav:


Biswa96:
Would you like to explain the requirement of this change?

3rav:
> Would you like to explain the requirement of this change?

The previous version 13.4.1 required the `Trilinos_ENABLE_Kokkos=OFF` flag, the current one doesn't require it to compile successfully, so I removed it.

PR #16243, opened 2023-03-16
qt6: update to 6.4.3 - Part1

MehdiChinoune:


lazka:
I did b3867d950c5db3e, assuming it was missed.

MehdiChinoune:
> I did [b3867d9](https://github.com/msys2/MINGW-packages/commit/b3867d950c5db3e00181a8afc450a546a3077423), assuming it was missed.

I forgot that translations requires tools in `makedepends`

PR #16285, opened 2023-03-18
blender: update to 3.2.2

MehdiChinoune:
I tried to update to 3.4.1, but since 3.3.0 it requires some WRL classes that are not implemented in mingw-w64.

@hmartinez82 Could you please make sure it builds on clangarm64.


lazka:
just for future reference, build error with clangarm64:

```
2023-03-18T13:43:37.2627458Z C:/M/mingw-w64-blender/src/blender-3.2.2/extern/fast_float/fast_float.h:309:17: error: use of undeclared identifier '__umulh'
2023-03-18T13:43:37.2732566Z 
2023-03-18T13:43:37.3046658Z   answer.high = __umulh(a, b);
2023-03-18T13:43:37.3669532Z 
2023-03-18T13:43:37.3670980Z                 ^
2023-03-18T13:43:37.4452448Z 
2023-03-18T13:43:37.4609967Z 1 error generated.
```

lazka:
Maybe this shouldn't exclude clang? https://github.com/blender/blender/blob/b93a66d603008352b52fed0bc2b2820cf428bc37/extern/fast_float/fast_float.h#L158

hmartinez82:
@MehdiChinoune I'll test today :)

MehdiChinoune:
> @MehdiChinoune I'll test today :)

You mean fix, as I have already pushed it, and It failed as mentioned above.

hmartinez82:
According to this https://learn.microsoft.com/en-us/cpp/intrinsics/umulh?view=msvc-170 it should be present for both x64 and ARM64, but intrin.h in clangarm64/include only has the x64 flavor.

MehdiChinoune:
> According to this https://learn.microsoft.com/en-us/cpp/intrinsics/umulh?view=msvc-170 it should be present for both x64 and ARM64, but intrin.h in clangarm64/include only has the x64 flavor.

That should be reported to mingw-w64.

hmartinez82:
Yeah. I noticed that for ARM64, several Microsoft-specific intrinsics are missing. I have a work around for now for Blender. Inline asm it is. PR coming shortly

hmartinez82:
https://github.com/msys2/MINGW-packages/pull/16296

PR #16296, opened 2023-03-18
blender: UMULH intrinsic is missing from MINGW-64 aarch64

hmartinez82:


hmartinez82:
Fun  `ERROR: Failure while downloading https://download.blender.org/source/blender-3.2.2.tar.xz`

PR #16314, opened 2023-03-20
wget: Update to 1.21.3

lazka:
* Move to gnutls like Arch/Debian/Fedora
* Drop the openssl patch, it no longer applies and we use gnutls now
* Drop the taskbar patch, this was added in 2017 an no longer applies. If someone misses it, please report it upstream.
* Drop c-ares, for some reason it wasn't used, and no other distro uses it. Please tell if you are missing it.


lazka:
There was some concern raised that gnutls has issues on Windows, but things work fine here.

If someone ends up here because of TLS issues, let us know and we can look into fixing gnutls and/or switching back to openssl.

PR #16340, opened 2023-03-21
opencascade: Remove unnecessary batch scripts

Biswa96:
The env.bat file get confused with env command in shell. That fails to build qemu.


Biswa96:
This issue was reported here https://github.com/msys2/MINGW-packages/discussions/16304

PR #16357, opened 2023-03-22
libepoxy: add patch

lb90:
Make sure that libepoxy can run in absence of EGL / GLES libraries. While it's true that libepoxy depends on angleproject since commit e19b3fb9bc4a10ca339b1252c1e593d8d11f571f, a few projects have stated their preference for bundling libepoxy without ANGLE: https://gitlab.com/inkscape/inkscape/-/issues/4031#note_1312666347

Regarding that, maybe it's worth switching angleproject to a makedepend only?

**How to reproduce the issue:**

Reproducing the issue is not quite straightforward because libepoxy depends on angleproject ATM. However, you can either:

* Temporarily force-remove angleproject with `pacman -Rdd mingw-w64-ucrt-x86_64-angleproject`, then run `gtk3-demo` and launch the OpenGL demo
* Download the latest development bundle of Inkscape: https://inkscape.org/release/inkscape-dev/?latest=1, open Inkscape and enable OpenGL rendering

Note that the issue is not reproducible on Linux because of a specific GLX workaround (added in https://github.com/anholt/libepoxy/commit/a7c270f8e2b8dcbf9874632e18a0948255de49d3) which avoids calling `epoxy_egl_get_current_gl_context_api` in that specific case (and on Linux you only have GLX or EGL, nothing else)


lb90:
> I don't know enough about all this, but I don't mind reverting the angle dependency if it makes problems.

Updated! angleproject does no harm, but should be optional IMO as it's not really required by libepoxy to be operational.

Let's also ask @raedrizqie 🙂

raedrizqie:
im ok with this fix.. but shouldnt it be in optdepends instead of makedepend?

lb90:
Currently libepoxy needs EGL headers to build with EGL support:

```
In file included from ../libepoxy/include/epoxy/egl.h:46,
                 from ../libepoxy/src/dispatch_common.h:59,
                 from ../libepoxy/src/dispatch_egl.c:28:
include/epoxy/egl_generated.h:11:10: fatal error: EGL/eglplatform.h: No such file or directory
   11 | #include "EGL/eglplatform.h"
```

But probably with can work around that with very little effort. Looks like libepoxy only needs EGL/eglplatform.h during build, which is a very small and self-contained header

lb90:
Here's the content of `$prefix/include/EGL/`:

```
egl.h
eglext.h
eglext_angle.h
eglplatform.h
```

Those headers are maintained upstream by Khronos (apart from `eglext_angle.h` which is specific to ANGLE):

```c
/*
** Copyright 2013-2020 The Khronos Group Inc.
** SPDX-License-Identifier: Apache-2.0
**
** This header is generated from the Khronos EGL XML API Registry.
** The current version of the Registry, generator scripts
** used to make the header, and the header can be found at
**   http://www.khronos.org/registry/egl
**
** Khronos $Git commit SHA1: 6fb1daea15 $ on $Git commit date: 2022-05-25 09:41:13 -0600 $
*/
```

I have also tried building MESA with EGL / GLES support. Here's the content of MINGW-packages\mingw-w64-mesa\pkg\mingw-w64-clang-x86_64-mesa\clang64\include\EGL:

```
egl.h
eglext.h
eglext_angle.h
eglmesaext.h
eglplatform.h
```

Apart from provider-specific headers (eglext_angle.h, eglmesaext.h), the headers are almost identical.

So maybe we should create a `egl-headers` (or `egl-base`) package with things maintained upstream by Khronos, then make ANGLE and MESA depend on that.

lb90:
Here they are: https://github.com/KhronosGroup/EGL-Registry/tree/main/api/EGL

Biswa96:
angle can be added in makedepends **and** optdepends if and only if angle is not required at runtime by libepoxy.

lb90:
It's not required at runtime :) libepoxy runs just fine without it (but consider that this PR is needed to fix a small bug)

PR #16364, opened 2023-03-22
pgmodeler: New package

podsvirov:


mingw-w64-pgmodeler/PKGBUILD
```diff
@@ -0,0 +1,71 @@
...
+_realname=pgmodeler
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=1.0.2
+pkgrel=1
+pkgdesc="PostgreSQL Database Modeler: an open source CASE tool for modeling PostgreSQL databases (mingw-w64)"
+url="https://pgmodeler.io"
+license=('GPL3')
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
```
Biswa96, line 12: 32-bit can be removed.

podsvirov: Is this a question or a suggestion?

I check 32-bit builds and they work. I am using this on a 32-bit computer.

I prefer to keep 32-bit builds if possible.

Biswa96: Suggestion. 32bit qt6 packages are being removed. See more here https://www.msys2.org/news/#2023-01-05-dropping-32bit-support-for-qt-6

podsvirov: Ok. Thanks for note and link to news page.

mingw-w64-pgmodeler/PKGBUILD
```diff
@@ -0,0 +1,71 @@
...
+depends=("${MINGW_PACKAGE_PREFIX}-libxml2"
+         "${MINGW_PACKAGE_PREFIX}-postgresql"
+         "${MINGW_PACKAGE_PREFIX}-qt6-svg")
```
MehdiChinoune, line 15: qt6-svg i686 (and all other i686 qt6 package) will be disabled with 6.5.0 (April/2023).

podsvirov: Should I remove the 32-bit builds now or can I leave it as is for now?

MehdiChinoune: You could build them against qt5-svg
```suggestion
         $([[ ${CARCH} != i686 ]] && echo \
           "${MINGW_PACKAGE_PREFIX}-qt6-svg" || echo \
           "${MINGW_PACKAGE_PREFIX}-qt5-svg"))
```

MehdiChinoune: but you need also to use qmake instead of qmake6
```shell
if [[ ${CARCH} != i686 ]]; then
  QMAKE=qmake6
else
  QMAKE=qmake
fi
```

podsvirov: Unfortunately this is not possible. The project also [requires Qt6](https://github.com/pgmodeler/pgmodeler/blob/v1.0.2/pgmodeler.pri#L16). There is only one thing left - i'm just remove 32-bit builds.

podsvirov: The 32-bit builds removed.

podsvirov:
Thanks to all!

PR #16378, opened 2023-03-23
grpc: update to 1.53.0

MehdiChinoune:
@mstorsjo Could you please investigate this one.

Disabling `disabling-relocation` and enabling `CMAKE_POSITION_INDEPENDENT_CODE` didn't fix building shared grpc++ libs.

### DIFF
--- a/mingw-w64-arrow/PKGBUILD
+++ b/mingw-w64-arrow/PKGBUILD
@@ -4,7 +4,7 @@ _realname=arrow
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=11.0.0
-pkgrel=3
+pkgrel=4
 pkgdesc="Apache Arrow is a cross-language development platform for in-memory data (mingw-w64)"
 arch=(any)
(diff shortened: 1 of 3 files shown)


mstorsjo:
Ok, so except for the massive amounts of warnings about `__declspec(empty_bases)`, I see this error:
```
ld.lld: error: unable to automatically import from _ZN9grpc_core9Timestamp25thread_local_time_source_E with relocation type IMAGE_REL_AMD64_SECREL in CMakeFiles/grpc++.dir/src/core/ext/transport/binder/client/channel_create_impl.cc.obj
```
The `IMAGE_REL_AMD64_SECREL` relocations come from accesses to thread local variables with the Windows Native TLS model (as opposed to emulated TLS as used in the GCC based environments). With Windows Native TLS, it's not possible to directly access a TLS variable in a different DLL (as in, you can't reference it directly by the data symbol name - but if there's an accessor, that works). This isn't supported in MSVC based environments either.

I see that the codebase is somewhat prepared for MSVC, but it probably doesn't support building it as DLLs (as it lacks export/import annotations, and these bits certainly wouldn't have worked across DLL boundaries).

For cases like these, it's kinda straightforward to wrap all access to the TLS variable into a helper function. FWIW, this also generates less complicated code for checking whether the variable has been initialized already, in all the callers. See https://martin.st/temp/grpc.patch.

PR #16424, opened 2023-03-25
adwaita-icon-theme: Update to 44.0

lazka:


lazka:
pngs are missing, something is broken maybe

PR #16429, opened 2023-03-25
libidn: Update to 1.41

lazka:
gnu.org server is down, or at least some mirrors

upstream is aware


Biswa96:
> gnu.org server is down, or at least some mirrors

Working here.

```
$ /usr/bin/curl -I "https://ftp.gnu.org/gnu/libidn/"
HTTP/1.1 200 OK
Date: Sat, 25 Mar 2023 13:11:18 GMT
Server: Apache/2.4.29 (Trisquel_GNU/Linux)
Strict-Transport-Security: max-age=63072000
Content-Security-Policy: default-src 'self'; img-src 'self' https://static.fsf.org https://static.gnu.org https://gnu.org http://static.fsf.org http://static.gnu.org http://gnu.org; object-src 'none'; frame-ancestors 'none'; child-src 'self' https://static.gnu.org http://static.fsf.org http://static.gnu.org http://gnu.org;
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
Content-Type: text/html;charset=UTF-8
```

lazka:
yes, it's also working here sometimes, but not in the US

MehdiChinoune:
gnu.org It's unreachable here since this morning.

PR #16430, opened 2023-03-25
libpaper: Update to 2.1.0

lazka:
move to new upstream like other distros


naveen521kk:
looks like this update is causing https://github.com/msys2/MINGW-packages/issues/17771. need to check what's wrong though

naveen521kk:
seems like libpaper is broken, here's some test code
```c
#include <paper.h>
#include <stdio.h>

double paper_width              = 595.0;
double paper_height             = 842.0;

void
select_paper (const char *paperspec)
{
  const struct paper *pi;

  pi = paperinfo(paperspec);
  printf("pi pointer: %p\n", pi);
  if (pi && papername(pi)) {
    paper_width  = paperpswidth (pi);
    paper_height = paperpsheight(pi);
  } else {
    ERROR("paper `%s' not found\n", paperspec);
    error = 1;
  }
}

int main() {
  select_paper("a4");
  printf("paper_width = %f\n", paper_width);
  printf("paper_height = %f\n", paper_height);
  return 0;
}
```
This should print something like this (from ubunutu):
```
0x7f6a2a70c0c0
paper_width = 595.275591
paper_height = 841.889764
```
but it ended up printing this on mingw (notice it returning NULL)
```
0000000000000000
Unrecognized paper format: a4
paper_width = 595.000000
paper_height = 842.000000
```

lazka:
yeah, looks like this update broke relocation

lazka:
See https://github.com/msys2/MINGW-packages/pull/17773

PR #16446, opened 2023-03-26
nss: Change database backend to sqlite3; remove -mwindows on utilities

asctime:


mingw-w64-nss/PKGBUILD
```diff
@@ -10,6 +10,7 @@ pkgdesc="Mozilla Network Security Services (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS"
+install=nss-${CARCH}.install
```
MehdiChinoune, line 13:
```suggestion
install=nss-${MSYSTEM}.install
```
Please, rename/add the required files and edit accordingly, for example install-UCRT64 with _mingw=ucrt64

asctime: @MehdiChinoune So when we do a .install script now we have to make six copies?

MehdiChinoune: Yes.

asctime:
I am unable to test for the other platforms so please let me know if there is a problem.

mingw-w64-nss/PKGBUILD
MehdiChinoune: Please rename install-i686 and add install-CLANG32, even you couldn't test them.

asctime:
So cloudbuild is still doing 32-bit checks - - and I don't see any checks for CLANGARM64 - - which ones?

MehdiChinoune:
```clang: warning: argument unused during compilation: '-mwindows' [-Wunused-command-line-argument]```
It seems that compile flag still used!

**Comment by asctime** [comment]
My guess is the patch somehow silently failing. That's useful. Anyway, maybe PKGBUILD needs a slightly different format (I just did diff -U 3)? I'll check it in a bit.

**Comment by MehdiChinoune** [comment]
> My guess is the patch somehow silently failing. That's useful. Anyway, maybe PKGBUILD needs a slightly different format (I just did diff -U 3)? I'll check it in a bit.

You have apply the patch in `prepare()`!

**Comment by asctime** [comment]
Yes I see it got dropped when I reverted to make the other changes. It will be added now.

File: mingw-w64-nss/nss-3.83-db_and_tool_fixes.patch
Comment (MehdiChinoune) [inline mingw-w64-nss/nss-3.83-db_and_tool_fixes.patch:1]: Please remove version from patch name.
Also nss is at 3.89, why are you building 3.83?

**Comment by MehdiChinoune** [comment]
Bump pkgrel.
Please, squash your changes to a single commit and then froce-push it.

File: mingw-w64-nss/nss-CLANG64.install [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+  fi
+  if [ -f "${_mingw}/bin/certutil.exe" ]; then
+      ${_mingw}/bin/certutil.exe -N --empty-password -d ${_nssdb}
+      ${_mingw}/bin/modutil.exe -create -force -dbdir ${_nssdb}
+  fi
+}
+
+post_upgrade() {
+  post_install
+}
```
MehdiChinoune, line 21: I think only post_upgrade should be added, new installations don't have those directories.

mingw-w64-nss/PKGBUILD
MehdiChinoune: where is nss-MINGW32.install?

mingw-w64-nss/nss-CLANG64.install
asctime: Yes but they need them. That's one of the points of the entire script, to generate (or re-generate) the nss certdb. But I did make one additional change and added the MINGW32 now as well (although I see other packages like Blender no longer carry it). Once everything looks for you I will re-submit.

asctime: OK it's done. Any objections to close and re-submit?

MehdiChinoune:
Change pkgrel to 2
`pkgrel=2`

lazka:
Stupid question, the Arch Linux package doesn't deal at all with those databases, why do we need to?

mingw-w64-nss/PKGBUILD
```diff
@@ -5,11 +5,12 @@ _realname=nss
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.89
-pkgrel=1
+pkgrel=2
 pkgdesc="Mozilla Network Security Services (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS"
+install=nss-${MSYSTEM}.install
```
lazka, line 13: There is no point in migrating DBs, we would have replaced them on every package update anyway.

asctime: > Stupid question, the Arch Linux package doesn't deal at all with those databases, why do we need to?

We don't need to but creating a system cert db for nss hardly hurts. Its suggested in the Netscape docs as well. Can't be any worse than the secmod.db that didn't do anything XD

asctime: > There is no point in migrating DBs, we would have replaced them on every package update anyway.

Yeah, not if the user decides to actually make use of that db, which is entirely supported in the upstream documentation.

mingw-w64-nss/nss-CLANG64.install
MehdiChinoune: Why closing, squash and force-push!

asctime:
@MehdiChinoune sorry I am not being offered the option to squash this PR it seems (unless they moved the button for it)

MehdiChinoune:
> @MehdiChinoune sorry I am not being offered the option to squash this PR it seems (unless they moved the button for it) EDIT: yeah I don't have "contributor" permissions here mate.

From your side. not from GitHub.

asctime:
> From your side. not from GitHub.

I did the final PR through my own github clone

MehdiChinoune:
> > From your side. not from GitHub.
> 
> I did the final PR through my own github clone. It wasn't direct, I didn't expect so many issues to be honest.

I couldn't even squash your commits from my side because you merged msys2/master into your branch!
Better close, clone MINGW-packages, create a new branch, push and open a PR.

PR #16454, opened 2023-03-26
Add LLVM 14.0.6 and build python-llvmlite and OSL against it.

MehdiChinoune:


MehdiChinoune:
I think another package LLVM-15 should be added for creduce and IWYU.

lazka:
Do I understand correctly that this moves them to statically link to llvm v14?

> I think another package LLVM-15 should be added for creduce and IWYU.

feel free, but if creduce doesn't gain anything from newer llvm then wen might as well try to limit things to one older llvm for everything.

MehdiChinoune:
> Do I understand correctly that this moves them to statically link to llvm v14?

Yes
> feel free, but if creduce doesn't gain anything from newer llvm then wen might as well try to limit things to one older llvm for everything.

creduce could only built against LLVM-15.

MehdiChinoune:
OK to merge?

lazka:
sure

PR #16460, opened 2023-03-26
mesa: Update to 23.0.1

pal1000:


Biswa96:
Just asking. From the build log, it seems that meson is downloading the directx-header at setup phase. Should there be a separate directx-headers package for mesa?

pal1000:
If you want that I have no problem adding a package but I am not interested in working on it.

PR #16461, opened 2023-03-26
blender: build against opensubdiv & enable openpgl on clangarm64

MehdiChinoune:


hmartinez82:
I can confirm that it's building on CLANGARM64

PR #16481, opened 2023-03-28
fix gdl dock object notify missing bugs

RickLeaf:
gdl can't emit dock object notify when change dock object property in glib 2.74 and above, this a patch to fix it


Biswa96:
Thank you for the contribution. Did you report the issue or send the patch in upstream repository?

RickLeaf:
> Thank you for the contribution. Did you report the issue or send the patch in upstream repository?

No, I still try to register gitlib and contact the owner.

lazka:
If it's caused by a glib update please file an issue with glib. Removing G_PARAM_CONSTRUCT seems weird to me.

RickLeaf:
> If it's caused by a glib update please file an issue with glib. Removing G_PARAM_CONSTRUCT seems weird to me, and changes the API behavior.
I'm not sure that is a glib's issue. These nofity must be emited when g_object_set is called, not only at construct stage

PR #16488, opened 2023-03-28
libatomic_ops: update to 7.8.0

MehdiChinoune:
I tried to build it
- buildflags
- !buildflags
- buildflags + "-mcx16"
- !builflags + "-mcx16"
It always fails with GCC x86_64, I have found many issues about it on GH and Bugzilla none of the is fixed!


Biswa96:
Is the build issue solved?

MehdiChinoune:
Yes, I have disabled using GNU Intrinsics with MINGW64 and UCRT64.

PR #16493, opened 2023-03-28
Khronos API headers

lb90:
Add new packages for headers maintained by Khronos:

* https://github.com/KhronosGroup/EGL-Registry
* https://github.com/KhronosGroup/OpenGL-Registry

We already have packages for [vulkan-headers](https://packages.msys2.org/base/mingw-w64-vulkan-headers) and [opencl-headers](https://packages.msys2.org/base/mingw-w64-opencl-headers). This PR adds the new packages `egl-headers` and `gles-headers`. More can be added in the future, if wanted (OpenVG, OpenMax etc.)

On Linux headers are provided mostly by [mesa](https://archlinux.org/packages/extra/x86_64/mesa/) and [libglvnd](https://archlinux.org/packages/extra/x86_64/libglvnd/), however on Windows there are multiple GL driver projects (angleproject, MESA), so it makes sense to separate API / protocol packages and implementation packages (much like Wayland and Wayland compositors)


lazka:
thanks!

Issue #10572, opened 2022-01-18
Doxygen error -- ghostscript "invalid font" probably due to "GenericResourceDir" not pointing to a valid resource directory


lluisalemanypuig:
I installed MSYS2 some time ago, and since the fresh install I've had problems in generating the doxygen documentation of a [C++ project](https://github.com/LAL-project/linear-arrangement-library.git). I [asked Stack Overflow](https://stackoverflow.com/questions/70364526/gpl-ghostscript-doxygen-on-msys2-outputs-unrecoverable-error) about this, I opened [an issue at Ghostscript bug tracker](https://bugs.ghostscript.com/show_bug.cgi?id=704803), and also opened [an issue at doxygen github repo](https://github.com/doxygen/doxygen/issues/9029), but nowhere could I find a definite way to fix this.

As explained in the posts before, I installed the relevant software with:
```
$ pacman -S mingw-w64-x86_64-texlive-font-utils
$ pacman -S mingw-w64-x86_64-texlive-bibtex-extra
$ pacman -S mingw-w64-x86_64-doxygen
```
and just a few minutes ago I did a full system upgrade by running `pacman -Syuu` three times (the 3rd time it did nothing). 

The error seems to have to do with an option of ghostscript being set up in a way that it stops the program from finding certain fonts.
```
*** Warning: GenericResourceDir doesn't point to a valid resource directory.
               the -sGenericResourceDir=... option can be used to set this.

Error: /invalidfont in /findfont
Operand stack:
   boxfont   Times-Roman
Execution stack:
   %interp_exit   .runexec2   --nostringval--   .findfontop   --nostringval--   2   %stopped_push   --nostringval--   .findfontop   .findfontop   false   1   %stopped_push   1990   1   3   %oparray_pop   1989   1   3   %oparray_pop   1988   1   3   %oparray_pop   .findfontop   1977   1   3   %oparray_pop   1833   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   .findfontop   --nostringval--   2   %stopped_push   --nostringval--   1925   2   3   %oparray_pop
Dictionary stack:
   --dict:736/1123(ro)(G)--   --dict:0/20(G)--   --dict:93/200(L)--
Current allocation mode is local
Last OS error: No such file or directory
Current file position is 646
GPL Ghostscript 9.53.3: Unrecoverable error, exit code 1
[...]

PR #16495, opened 2023-03-28
texlive-core: Fix invoking `ghostscript` through tex commands

naveen521kk:
Previously, `TEXLIVE_WINDOWS_EXTERNAL_GS` is not set which caused some issues with ghostscript finding the correct font files.

Fixes https://github.com/msys2/MINGW-packages/issues/10572


mtc-mlx:
This fixes the issue for me. Thank you @naveen521kk!

naveen521kk:
Thank you for testing! @mtc-mlx

PR #16528, opened 2023-03-30
starpu: fix 32bit build with clang v16

lazka:
it rightfully complains the the callback calling convention doesn't match


lazka:
@3rav fyi

upstream git isn't public, so not sure if we can upstream this.

or maybe we should drop 32bit?

3rav:
> @3rav fyi
> 
> upstream git isn't public, so not sure if we can upstream this.
> 
> or maybe we should drop 32bit?
> 
> or just leave it as is for now...

For me drop 32bit

lazka:
> Isn't this one the official repository https://gitlab.inria.fr/starpu/starpu ? Also there is an official mirror https://github.com/starpu-runtime/starpu/

Ah, I missed the mirror, thanks (the former requires you to send an email to someone to request an account)

Biswa96:
Would you like to upstream the changes? Or can I?

lazka:
https://github.com/starpu-runtime/starpu/pull/13

PR #16538, opened 2023-03-30
gdb: add patch

lb90:
Fixes the issue described in https://gitlab.com/inkscape/inkscape/-/issues/4031#note_1231250877. GDB is unable to retrieve any information regarding the callstack of the debugged program. Hence one may think that a stack corruption is going on, but instead it's just a call of a NULL function pointer.

Also fix compilation with clang 16 as suggested by @mmuetzel. See https://discourse.llvm.org/t/clang-16-notice-of-potentially-breaking-changes/65562 for more info

Reference https://sourceware.org/pipermail/gdb-patches/2023-March/198196.html

### DIFF
--- a/mingw-w64-gdb/PKGBUILD
+++ b/mingw-w64-gdb/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-multiarch")
 pkgver=13.1
-pkgrel=3
+pkgrel=4
 pkgdesc="GNU Debugger (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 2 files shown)


lb90:
I can't reproduce the clang64/clang32 build failure locally 🤷‍♂️

mmuetzel:
That might be because clang 16 that is currently in staging is stricter when it comes to what is valid C.
You could probably work around that by adding `-Wno-enum-constexpr-conversion` to `CFLAGS`.

mingw-w64-gdb/PKGBUILD
```diff
@@ -91,6 +97,12 @@ do_build() {
   CFLAGS+=" -I${MINGW_PREFIX}/include/ncurses"
   CXXFLAGS+=" -I${MINGW_PREFIX}/include/ncurses"
 
+  # Fix compilation with clang 16. Consider removing
+  # once fixed upstream
+  if [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]]; then
+    CFLAGS+=" -Wno-enum-constexpr-conversion"
+  fi
```
mmuetzel, line 104: Sorry. I saw that the compilation failed for a .c file and assumed that it was build by clang. But it turns out it is being built by clang++
``` suggestion
  # once fixed upstream
  declare -a _extra_config
    CFLAGS+=" -Wno-enum-constexpr-conversion"
    CXXFLAGS+=" -Wno-enum-constexpr-conversion"
    # Checking for a working Ada compiler takes very long and ultimately fails.
    _extra_config+=("acx_cv_cc_gcc_supports_ada=no")
```

The last part is an independent suggestion. For that to work, we'd need to add `${_extra_config[@]}` to the configure flags. Feel free to skip that part if you prefer. But it speeds up configure by more than 10 minutes for me. (I thought it got stuck completely. But then eventually resumed.)

mmuetzel: Forgot to add: With these changes, it is building for me locally with Clang 16.

lb90: No problem at all, and thank you for all the infos! I'll wait for clang 16 to move out of staging and get back to this

mmuetzel: Just for info: Clang 16 packages are being distributed by MSYS2 since a couple of days now.

PR #16550, opened 2023-03-31
lapack: enable clang64 and clangarm64

MehdiChinoune:


mmuetzel:
Just playing devil's advocate here (because I haven't decided yet for myself what I think might be best):
Initial tests suggested that Flang (the compiler, some runtime, or something else) produces erroneous results for some operations (with complex numbers?). Should we really distribute binaries for which we suspect they don't work correctly?

MehdiChinoune:
> Just playing devil's advocate here (because I haven't decided yet for myself what I think might be best): Initial tests suggested that Flang (the compiler, some runtime, or something else) produces erroneous results for some operations (with complex numbers?). Should we really distribute binaries for which we suspect they don't work correctly?

It's not like we are changing an existing package, We are offering a possibly-broken package which I think is better than nothing. It should help us know more about flang maturity in general and in MINGW .

mmuetzel:
Maybe, we should at least let the users know somewhere that LLVM Flang is still not considered "production ready" (neither by MSYS2 nor by upstream). Users should expect binaries compiled with it (including those distributed by MSYS2) behave unexpectedly.

MehdiChinoune:
LAPACK is the first library that flang developers made sure to be buildable by Flang because It's mostly an F77 code, It's kind of supported.
If something go wrong then we should investigate it and fix for MINGW.

mmuetzel:
Where did you find that LAPACK is tested with LLVM Flang? This board still lists it as "To Do":
https://github.com/flang-compiler/f18-llvm-project/projects/11

MehdiChinoune:
> Where did you find that LAPACK is tested with LLVM Flang? This board still lists it as "To Do": https://github.com/flang-compiler/f18-llvm-project/projects/11

I don't remember exactly!

MehdiChinoune:
> > Where did you find that LAPACK is tested with LLVM Flang? This board still lists it as "To Do": https://github.com/flang-compiler/f18-llvm-project/projects/11
> 
> I don't remember exactly!

I the early days of the project when It was not part of LLVM, there were some discussions about LAPACK, https://github.com/flang-compiler/f18-llvm-project/pull/22. Just search the repo.

mmuetzel:
Having some commits that fix issues that were found specifically with LAPACK doesn't necessarily mean that it is "supported".
Additionally, LAPACK uses Fortran 90 in some parts (e.g., https://github.com/Reference-LAPACK/lapack/blob/master/SRC/clartg.f90). I don't know if LLVM claims to support that yet...

MehdiChinoune:
> Having some commits that fix issues that were found specifically with LAPACK doesn't necessarily mean that it is "supported". Additionally, LAPACK uses Fortran 90 in some parts (e.g., https://github.com/Reference-LAPACK/lapack/blob/master/SRC/clartg.f90). I don't know if LLVM claims to support that yet...

I said *mostly F77*. BTW you could have an F77 code in .f90 file.
Also It's mentioned in their bi-weekly reports that they tested CMake support with LAPACK.
Anyway, supported or not, I am with merging this PR, users will know later (if they don't already know) about its issues. the worst thing is to tell them that Flang compiled libraries are still experimental, Is that difficult!

MehdiChinoune:
If you fellow you logic, we should disable flang all together because It's still experimental!

mmuetzel:
> If you fellow you logic, we should disable flang all together because It's still experimental!

I don't understand how you came to that conclusion. I never wrote that we should disable or enable anything. I just asked a question. Please, try to not assume the worst of people that you are interacting with.

Why is giving users a hint of caution so bad? That could be along the lines of: "Binaries compiled with LLVM Flang are not yet thoroughly tested. Please, report any issues you come across."

MehdiChinoune:
> Why is giving users a hint of caution so bad? That could be along the lines of: "Binaries compiled with LLVM Flang are not yet thoroughly tested. Please, report any issues you come across."

How?, Where? or When?

mmuetzel:
> How?, Where? or When?

Now, those are constructive questions. 👍

Maybe, with a short announcement on the webpage here:
https://www.msys2.org/news/

That could also include one or two sentences announcing that MSYS2 will start to distribute Clang and Flang 16.

MehdiChinoune:
I don't see any issue of deploying packages that weren't existed at all!
We are not RedHat, SUSE or Canonical to worry that much about packages maturity and users feedback. As a rolling-based platform It's expected to have experimental/unstable packages.

PR #16554, opened 2023-03-31
goocanvas: Fix C string with single quotes, and enable aarch64

hmartinez82:
I think that this, 8 years old patch, was probably handling something specific from Glib/Gtk back then.
I've tested current code with the existing patches and it won't build with CLANG64 nor MINGW64 because we get a C string being initialized with single quotes.


lazka:
dlls are missing, so something is broken.

hmartinez82:
Interesting. Something changed? When I first created the PR, all builds were passing. Let me update my master branch and use the staged packages and try again

hmartinez82:
Lol, what an interesting bug
glib-mkenums is being called correctly, but then, there's some comment in the code that is making the whole file to be commented out:

```

/* This file is generated by glib-mkenums, do not modify it. This code is licensed under the same license as the containing project. Note that it links to GLib, so must comply with the LGPL linking clauses. */

#include <glib-object.h>
#include "goocanvas.h"

/* Enumerations from "goocanvasitem.h" *C:/msys64/     <---- this line is making the whole file below to be commented out!
GType
goo_canvas_animate_type_get_type (void)
{
  static GType etype = 0;
 if( etype == 0 ) 
 
```

hmartinez82:
So. This line 
`--fprod "\n/* Enumerations from \"@filename@\" */" \` in Makefile.in
is replacing the */ with *C:/msys64/

But intereting, the .h generationg has
`--fprod "/* Enumerations from \"@filename@\" */\n" \` 

That extra `\n` in the end is making it not convert */ into a path

lazka:
> Interesting. Something changed?

maybe fallout from https://github.com/msys2/msys2-runtime/pull/143

hmartinez82:
Seems likely. I got a workaround already above. 
It would be nice if this scenario here would also be considered.

This explains why it didn't fail when I created the PR

lazka:
the DLLs are still missing though :)

maybe a `export lt_cv_deplibs_check_method='pass_all'` helps?

hmartinez82:
DLL is back now. I don't know what `export lt_cv_deplibs_check_method='pass_all'` did, but it fixed it. What does it do?

PR #16556, opened 2023-03-31
OpenLDAP: Update to 2.6.4 and enable aarch64 builds

hmartinez82:


hmartinez82:
😑 I actually only tested on CLANG64 and it builds without issues
```
...
checking for regex.h... yes
checking for library containing regfree... -lregex
checking for compatible POSIX regex... yes
...
```

Looking into it...

hmartinez82:
Oh, it's already using LLVM 16

mingw-w64-openldap/PKGBUILD
```diff
@@ -71,8 +72,6 @@ build() {
     --with-tls=openssl \
     --with-cyrus-sasl \
     --enable-modules=yes \
-    --enable-hdb=no \
-    --enable-bdb=no \
```
hmartinez82, line 75: `configure: WARNING: unrecognized options: --enable-hdb, --enable-bdb`

PR #16489, opened 2023-03-28
qgis: Enable python plugins

hmartinez82:
image

PR #16563, opened 2023-03-31
qgis: update to 3.30.1

MehdiChinoune:


MehdiChinoune:
> @MehdiChinoune Can you bring the changes from #16489 too? It's missing owlib and python-gdal here :(

Very strange, I have rebased my changes which were changing pkgver and sha256sums on your PR, but you pushed your changes later.

PR #16580, opened 2023-04-01
open62541: Update to 1.3.5 and enable aarch64 builds

hmartinez82:


mingw-w64-open62541/PKGBUILD
```diff
@@ -1,14 +1,14 @@
 _realname=open62541
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.3.3.r0.gff7326882
+pkgver=1.3.5.r0.g0bb2b0142
 pkgrel=1
-_commit="ff73268829359639531ff02905c889f73a77b408"
+_commit="0bb2b0142c4a9a89eae8a6a7a1ea95801ab3582d"
```
lazka, line 6: Do you think we could use the tarball instead of git?

hmartinez82: @lazka I tried, but they don't provide tarballs, only the git repro compressed as-is. And they have submodules which are not included and the mere repo compressed files!
I then went to their website, https://www.open62541.org/download/ and the link there just points to the same files, without the submodules.

lazka: Ah, ok

PR #16601, opened 2023-04-03
python-wmi: new package

jeremyd2019:


jeremyd2019:
I plan to try to test (run check) locally before marking ready for review.

jeremyd2019:
check succeeded on clangarm64.

Biswa96:
Just wondering. Is the project still maintained? In [pypi WMI](https://pypi.org/project/WMI/), the last release was in Apr 28, 2020.

jeremyd2019:
I don't know, I suspect it just doesn't change much, since it is wrapping a Windows API that has been around for 20-some years.  I was actually surprised it was updated as recently as it was...

In any event, it is listed in the requirements.txt of the os-win project, and is installed when pip installing os-win with normal Windows python.

PR #16610, opened 2023-04-03
qt6: update to 6.5.0 - Part1

MehdiChinoune:


mingw-w64-qt6-base/PKGBUILD
```diff
@@ -4,11 +4,11 @@ _realname=qt6-base
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-debug")
-_qtver=6.4.3
+_qtver=6.5.0
 pkgver=${_qtver/-/}
 pkgrel=1
 arch=(any)
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
```
orgads, line 11: Why did you remove mingw32?

ognevnydemon: because qt6 dropped 32 bit support

orgads: In which version? I couldn't find it in the release notes.

ognevnydemon: https://www.qt.io/blog/qt6-development-hosts-and-targets

ognevnydemon: also from MSYS2 news

> Qt project dropped support for Windows version older than Windows 10 from Qt 6, see this official [blog post](https://www.qt.io/blog/qt6-development-hosts-and-targets). It was also added that we will not have 32 bit x86 Windows support available. With this above condition, we too have very few users who are using 32 bit x86 Windows 10. So, we decided to remove 32 bit builds for Qt 6 and their dependencies. The remaining 32 bit x86 packages which depends on Qt will be linked with Qt 5. With this, our Qt 6 packages will be available for all official [Windows platforms](https://doc.qt.io/qt-6/windows.html).

orgads: As long as the build is not broken, why actively remove the package?

Quoting from Oliver's comment on the blogpost:

> The end of 32 bit packages for Windows was perhaps phrased a bit unfortunately by Santtu. The decision no longer to support 32 bit on Windows was done deliberately from the Windows version support. This decision does not mean that we will actively remove support for 32 bit on Windows though. Having a platform as "supported" means more than "it works by chance".

orgads: Is it possible to compile 32-bit target with mingw64? I tried using `-m32`, but I'm getting this error:

```
Fatal error: selected target format 'pe-bigobj-i386' unknown
```

MehdiChinoune: What do you need mingw32 for?

orgads: I have a project that is incompatible with win64, and it uses qmake. All I care about is qmake (that's qt6-base). I really don't mind dropping the rest.

MehdiChinoune: Well, you should upgrade your project to be win64 compatible.

orgads: Easier said than done.

I would love to also convert it from qmake to cmake, but that would take weeks I don't have for that.

Biswa96: Could you share the project link please?

orgads: It's not public.

MehdiChinoune: You know you could build packages by yourself!

orgads: I know, but I prefer to use upstream maintained and up-to-date packages rather than building it myself. Isn't that the whole point of having a package manager?

Is dropping 32-bit packages a common thing in msys2? Do you plan to drop it completely at some point?

MehdiChinoune: > I know, but I prefer to use upstream maintained and up-to-date packages rather than building it myself. Isn't that the whole point of having a package manager?

The maintainers are not willing to maintain packages for practically non-existing users.
There is no real users of 32bit packages, It's just developers (like you) complaining about it which make other developers complain about it and so on.
> 
> Is dropping 32-bit packages a common thing in msys2? Do you plan to drop it completely at some point?

Yes, We agreed before to start dropping 32bit packages if the upstream dropped it which is the case with Qt6.
Qt6 support only Windows-64bit>10, There is no users (~0) with Windows-10+32bit.

orgads: I think that comments on a merged change are not the right place for discussing this decision. I suggest to open a proper issue to discuss this.

The upstream did *not* drop 32bit, and it compiles just fine. They just don't plan to support it in case it will require more effort than they're willing to put into it. As long as it works, I don't see a reason to drop it, unless you completely drop mingw32, which is not the case.

I'm using Win10-64, but I do need 32-bit compilations for some projects.

orgads: - #17447

PR #16630, opened 2023-04-04
new packages python-pymi and python-os-win

jeremyd2019:
pymi is a dependency of os-win, which is a library from openstack for managing hyper-v.

Draft until the patches for mingw-w64 are applied and the packages are updated, not worth carrying a downstream patch over.

### DIFF
--- a/mingw-w64-python-os-win/PKGBUILD
+++ b/mingw-w64-python-os-win/PKGBUILD
@@ -1,0 +1,69 @@
+# Maintainer: Jeremy Drake <github@jdrake.com>
+
+_realname=os-win
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=5.9.0
+pkgrel=1
+pkgdesc="Windows/Hyper-V Python library for OpenStack projects (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+url='https://opendev.org/openstack/os-win'
+license=('spdx:Apache-2.0')
+depends=("${MINGW_PACKAGE_PREFIX}-python-pbr"
+         "${MINGW_PACKAGE_PREFIX}-python-eventlet"
+         "${MINGW_PACKAGE_PREFIX}-python-oslo-concurrency"
+         "${MINGW_PACKAGE_PREFIX}-python-oslo-config"
+         "${MINGW_PACKAGE_PREFIX}-python-oslo-log"
+         "${MINGW_PACKAGE_PREFIX}-python-oslo-utils"
+         "${MINGW_PACKAGE_PREFIX}-python-oslo-i18n"
+         "${MINGW_PACKAGE_PREFIX}-python-pymi"
+         "${MINGW_PACKAGE_PREFIX}-python-wmi")
+makedepends=("${MINGW_PACKAGE_PREFIX}-python-build"
+             "${MINGW_PACKAGE_PREFIX}-python-installer"
+             "${MINGW_PACKAGE_PREFIX}-python-setuptools"
+             "${MINGW_PACKAGE_PREFIX}-python-wheel")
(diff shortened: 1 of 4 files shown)


jeremyd2019:
On clangarm64:
Tests for PyMI fail (and also, fail due to not finding the C module, this is something that got taken care of with `setup.py test` but apparently not with `pytest`).

All but one test pass for os-win:
```
==============================
Failed 1 tests - output below:
==============================

os_win.tests.functional.test_mutex.MutexTestCase.test_already_acquired_mutex
----------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "C:/M\mingw-w64-python-os-win\src\python-build-CLANGARM64/os_win/tests/functional/test_mutex.py", line 56, in test_already_acquired_mutex
    self.assertFalse(self._mutex.acquire(timeout_ms=0))

      File "C:/msys64/clangarm64/lib/python3.10/unittest/case.py", line 681, in assertFalse
    raise self.failureException(msg)

    AssertionError: True is not false



======
Totals
======
Ran: 874 tests in 3.8750 sec.
 - Passed: 873
 - Skipped: 0
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 1
Sum of execute time for each test: 7.3904 sec.
```

jeremyd2019:
https://github.com/msys2-arm/MINGW-packages/actions/runs/4602716650 has CLANGARM64 artifacts too

jeremyd2019:
mingw-w64 has caught up now, ready for review

mingw-w64-python-os-win/PKGBUILD
```diff
@@ -0,0 +1,69 @@
+# Maintainer: Jeremy Drake <github@jdrake.com>
+
+_realname=os-win
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=5.9.0
+pkgrel=1
+pkgdesc="Windows/Hyper-V library for OpenStack projects (mingw-w64)"
```
jeremyd2019, line 8:
```suggestion
pkgdesc="Windows/Hyper-V python library for OpenStack projects (mingw-w64)"
```
should I specify "python" in pkgdesc?  Otherwise the descriptions were copied from the project itself somewhere.

mingw-w64-python-pymi/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+         "${MINGW_PACKAGE_PREFIX}-python-six")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-python-build"
+             "${MINGW_PACKAGE_PREFIX}-python-installer"
+             "${MINGW_PACKAGE_PREFIX}-python-setuptools"
+             "${MINGW_PACKAGE_PREFIX}-python-wheel")
+checkdepends=("${MINGW_PACKAGE_PREFIX}-python-pytest"
+              "${MINGW_PACKAGE_PREFIX}-python-testtools")
+source=("https://pypi.org/packages/source/${_realname::1}/${_realname}/${_realname}-${pkgver}.tar.gz"
+        "fixes.patch")
```
jeremyd2019, line 23: Don't really like `fixes.patch` as a name, but naming is always the hardest part...

mingw-w64-python-os-win/PKGBUILD
Biswa96: Isn't it obvious that the package is a python library from the package name? Though adding it would not cause any confusion 😃

mingw-w64-python-pymi/PKGBUILD
Biswa96: 0001-pymi-fix-mingw-compiler-error.patch

jeremyd2019:
Oh, I confirmed that os-win does indeed use both PyMI *and* wmi (even though PyMI provides a replacement wmi module).  Apparently there were issues on Server 2012, so it's not a very likely path to hit anymore, but it does still exist:
https://github.com/openstack/os-win/blob/318b1052d30baf7a11335399941c211c553b8c7a/os_win/utils/baseutils.py#L105-L116

PR #16635, opened 2023-04-04
oiio: update to 2.4.10.0

kmilos:


kmilos:
The CI failure seems to related to https://github.com/msys2/MINGW-packages/issues/10761? No idea why it happened on UCRT64 but not on MINGW64...

PR #16637, opened 2023-04-04
qt6-static: update to 6.5.0

MehdiChinoune:


MehdiChinoune:
I have disabled `Debug` build type, It won't time-out or saturate the storage (like last time)
If someone want Debug version the he should build it himself like any other package.

PR #16641, opened 2023-04-04
vala: update to v0.56.6

wszqkzqk:


hmartinez82:
LLVM 16 tidbits
```
../../vala-0.56.6/gobject-introspection/grealpath.h:29:19: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
          extern __stdcall GetFullPathNameA(const char*, int, char*, char**);
```

Needs patching or using `-Wno-implicit-int`

Biswa96:
@wszqkzqk Would you like to apply the attached patch file? [0002-vala-fix-function-decl.patch](https://github.com/msys2/MINGW-packages/files/11151458/0002-vala-fix-function-decl.patch)

Biswa96:
The patch was sent to upstream in this merge request https://gitlab.gnome.org/GNOME/vala/-/merge_requests/304

wszqkzqk:
> @wszqkzqk Would you like to apply the attached patch file? [0002-vala-fix-function-decl.patch](https://github.com/msys2/MINGW-packages/files/11151458/0002-vala-fix-function-decl.patch)

OK

PR #16666, opened 2023-04-05
LLVM 16.0.1 - Part1

lazka:


MehdiChinoune:
I have a patch for flang, I want to apply. It has two parts one in clang package and the other in flang package.

lazka:
ok, feel free to close this if you want to take over :)

PR #16674, opened 2023-04-05
update some gnome packages

Biswa96:


Biswa96:
@hmartinez82 Would you like to check if these can be compiled in clangarm64?

hmartinez82:
tepl, gnome-latex, gedit, gedit-plugins? Did I miss anything?

Biswa96:
> tepl, gnome-latex, gedit, gedit-plugins? Did I miss anything?

Nope, but their dependencies may not present in clangarm64 also.

hmartinez82:
Tepl ✅

hmartinez82:
gnome-latex ✅

Biswa96:
Providing the patch which enabled clangarm64 for all these would be sufficient.

hmartinez82:
gedit: ✅ (but requires https://github.com/msys2/MINGW-packages/pull/16690)

hmartinez82:
gedit-plugins: ✅ (but requires https://github.com/msys2/MINGW-packages/pull/16691)

hmartinez82:
Wow, gedit with GTK4 was totally broken before, it's super stable now
image

PR #16711, opened 2023-04-06
python-winsdk: Add package

mimi89999:
I only built it and tested it on `clang64`

### DIFF
--- a/mingw-w64-python-winsdk/PKGBUILD
+++ b/mingw-w64-python-winsdk/PKGBUILD
@@ -1,0 +1,47 @@
+# Maintainer: Michel Le Bihan <michel@lebihan.pl>
+
+_realname=winsdk
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=1.0.0b9
+pkgrel=1
+pkgdesc="Python package with bindings for Windows SDK (mingw-w64)"
+arch=('any')
+mingw_arch=('clang64' 'clang32')
+url='https://github.com/pywinrt/python-winsdk'
+license=('spdx:MIT')
+depends=("${MINGW_PACKAGE_PREFIX}-python")
+makedepends=("${MINGW_PACKAGE_PREFIX}-python-build"
+             "${MINGW_PACKAGE_PREFIX}-python-installer"
+             "${MINGW_PACKAGE_PREFIX}-python-scikit-build"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-cc")
+source=("https://pypi.org/packages/source/${_realname::1}/${_realname}/${_realname}-${pkgver}.tar.gz")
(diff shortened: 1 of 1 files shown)


Biswa96:
Is it possible to upstream the patches first? The mingw-w64-python-winsdk/001-Regenerate-pywinrt-bindings.patch file is huge.

mimi89999:
They are upstream, but we would need to ask @dlech to regenerate the bindings in the repo.

mimi89999:
To avoid having such a big patch file, I could also fork the repo and apply the patch there

Biswa96:
> To avoid having such a big patch file, I could also fork the repo and apply the patch there

The package should be built from the actual repository. If the patch is in that upstream repository I can be downloaded from the commit in `source=()` array.

mimi89999:
> > To avoid having such a big patch file, I could also fork the repo and apply the patch there
> 
> The package should be built from the actual repository. If the patch is in that upstream repository I can be downloaded from the commit in `source=()` array.

I know. All the code in `pywinrt/winsdk/` is generated using https://github.com/pywinrt/pywinrt/commit/64d3e0706bf516334db90d7b2cd1e7827c42b0de, so the patches are technically upstream, but the bindings in the https://github.com/pywinrt/python-winsdk would need to be updated. Upstream will probably only update them on the next `pywinrt` release.

mimi89999:
@Biswa96 Yesterday winsdk 1.0.0b9 was released that includes all the required patches for building in Clang MSYS2. I updated the PR and it doesn't include any patch file now.

mingw-w64-python-winsdk/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-cc")
+source=("https://pypi.org/packages/source/${_realname::1}/${_realname}/${_realname}-${pkgver}.tar.gz")
+sha256sums=('207d35f9ab445521d4f060b5c1f2181660390cd3db27b32961af7438478fb226')
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+
+  # Remove due to the lack of required headers
+  rm pywinrt/winsdk/src/py.Windows.Graphics.Capture.Interop.cpp
```
dlech, line 27: This should no longer be needed.

Biswa96: Would you like to explain the reason please?

dlech: https://github.com/pywinrt/pywinrt/commit/1c7d7118b3025c14e33fc7d01b8afa30c22b0380

dlech: If the header is not found at compile time, the equivalent mingw-compatible interface declaration is provided inline.

Biswa96: I can try to add the windows.graphics.capture.interop.h file in mingw-w64. Need some time though.

mimi89999: Thanks for especially providing a workaround for MSYS2 and sorry for missing this. I will fix it during the next release. Currently I'm still working on getting it to build on MINGW

PR #16714, opened 2023-04-07
qtbinpatcher: remove

MehdiChinoune:
Qt is now relocatable.


lazka:
qtbinpatcher is used in some install files if I'm not mistaken

MehdiChinoune:
> qtbinpatcher is used in some install files if I'm not mistaken

It was used to relocate qt5 binaries/files https://github.com/Fsu0413/QQtPatcher#for-users-of-qt-514-or-later

lazka:
I mean https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-qt5-static/qt5-MINGW32.install

should that be removed?

MehdiChinoune:
I forgot to remove those.
an update to 5.15.9 is ready, but it requires a manual build for at least 3 envs!

PR #16753, opened 2023-04-08
python-axolotl-curve25519: Add package

mimi89999:


mingw-w64-python-python-axolotl-curve25519/PKGBUILD
```diff
@@ -0,0 +1,45 @@
+# Maintainer: Michel Le Bihan <michel@lebihan.pl>
+
+_realname=python-axolotl-curve25519
```
Biswa96, line 3:
```suggestion
_realname=axolotl-curve25519
```

Biswa96:
It seems that the repository is not maintained. The last commit in the upstream repository was [on Apr 14, 2018](https://github.com/tgalal/python-axolotl-curve25519/commit/293f9cd5a42459f7e68748764de8387f37147db2)
. Should it be added as a package here?

mimi89999:
> It seems that the repository is not maintained. The last commit in the upstream repository was [on Apr 14, 2018](https://github.com/tgalal/python-axolotl-curve25519/commit/293f9cd5a42459f7e68748764de8387f37147db2) . Should it be added as a package here?

I saw that. The package is used by `python-axolotl` that is used by Gajim, so I think that it should be packaged.

mingw-w64-python-python-axolotl-curve25519/PKGBUILD
mimi89999: Shouldn't `_realname` always be the name on PyPi?

Biswa96: There is no rule like that. The issue is that the package name has two 'python' words.

mimi89999: OK. Renamed.

PR #16755, opened 2023-04-09
CI: check whether strip is disabled when using python-installer

naveen521kk:
This should avoid something like https://github.com/msys2/MINGW-packages/pull/16750 to occur


.ci/ci-check.py
```diff
@@ -189,12 +202,73 @@ def gha_group(title: str) -> typing.Generator:
...
+    print(f"::error file={filename.relative_to(REPO_ROOT).as_posix()},title={title}::{message}")
+    stdout.flush()
+
+
 def get_pkg_name(pkgloc: Path):
     return "-".join(pkgloc.name.split("-")[:-3])
 
 
+def check_strip_disabled_for_pkg(pkg_name: str) -> bool:
+    pkgbuild_file = REPO_ROOT / pkg_name.replace("-ucrt-x86_64", "") / "PKGBUILD"
```
jeremyd2019, line 215: ugh... what you really want here is to use pkgbase (could be a split package), but even that doesn't work in all cases.  https://github.com/jeremyd2019/clang32ize/blob/bafb362db6c81458ccd5fda8adfafd07328b7bed/pkgbuild.py#L25-L32

.ci/ci-check.py
```diff
@@ -189,12 +202,73 @@ def gha_group(title: str) -> typing.Generator:
...
+    pkg = db.get_pkg(pkg_name)
+    if pkg is None:
+        raise ValueError(f"Unable to find {pkg_name} in the database.")
+    return any(
+        i.startswith("mingw-w64-ucrt-x86_64-python-installer") for i in pkg.makedepends
+    )
+
+
+def check_whether_files_in_bin_folder(pkg_loc: Path) -> bool:
+    with pacdb.ExtTarFile.open(pkg_loc) as pkg:
```
jeremyd2019, line 242: If you open `ci.files.tar.gz` with pacdb, you can check the files list instead of having to open the tar file yourself (sort of like the early optimization in https://github.com/jeremyd2019/package-grokker/blob/c0ed4c7ff56bb2261d91c5bd2d116dbcfb7e9df7/grokkermod.py#L51)

naveen521kk: Will it work? For some reason, this returns an empty list:
```py
>>> import pacdb
>>> a = pacdb.mingw_db_by_name('mingw64')
>>> v = a.get_pkg('mingw-w64-x86_64-python-markdown-it-py')
>>> v.files
[]
```
Maybe will it work for local db's?

.ci/ci-check.py
naveen521kk: ugh, there should be a better way that, can't query the server each time. One thing we have is the git log, maybe we can do something with that? need to thing of better ideas...

.ci/ci-check.py
jeremyd2019: for that, you'd need the optional second parameter to `mingw_db_by_name` to be `'files'`.  For a local db, you'd give the path to the `.files.tar.X` instead of `.db.tar.X`

.ci/ci-check.py
jeremyd2019: luckily it's pretty rare for the PKGBUILD directory not to be the same as the pkgbase, so it's not too bad to query as a fallback, but I did make a suggestion in https://github.com/msys2/MSYS2-packages/issues/3699#issuecomment-1501159074 to make it a requirement.

.ci/ci-check.py
naveen521kk: I see. I guess then it's better to do that instead of manually reading the archive, will change.

.ci/ci-check.py
naveen521kk: does https://github.com/msys2/MINGW-packages/pull/16755/commits/c1b2be2f1930af29eb60b0e2fc99511674f818b1 look good?

.ci/ci-check.py
naveen521kk: I've done it https://github.com/msys2/MINGW-packages/pull/16755/commits/b7fa4b016040dc2b1787e99d1a5533cf8dfea90e

.ci/ci-check.py
```diff
@@ -189,12 +207,79 @@ def gha_group(title: str) -> typing.Generator:
...
+    if pkg is None:
+        raise ValueError(f"Unable to find {pkg_name} in the database.")
+    return any(
+        i.startswith("mingw-w64-ucrt-x86_64-python-installer") for i in pkg.makedepends
+    )
+
+
+def check_whether_files_in_bin_folder(pkg_name: str) -> bool:
+    db = get_pacdb_files_database()
+    return any(i.startswith("ucrt64/bin") for i in db.get_pkg(pkg_name).files)
```
jeremyd2019, line 255:
```suggestion
    return any(i.startswith("ucrt64/bin") and i.endswith(".exe") for i in db.get_pkg(pkg_name).files)
```
?

.ci/ci-check.py
```diff
@@ -189,12 +207,79 @@ def gha_group(title: str) -> typing.Generator:
...
+
+
+def check_whether_using_python_installer(pkg_name: str) -> bool:
+    db = get_pacdb_database()
+    pkg = db.get_pkg(pkg_name)
+    if pkg is None:
+        raise ValueError(f"Unable to find {pkg_name} in the database.")
+    return any(
+        i.startswith("mingw-w64-ucrt-x86_64-python-installer") for i in pkg.makedepends
+    )
```
jeremyd2019, line 250:
```suggestion
    return "mingw-w64-ucrt-x86_64-python-installer" in pkg.makedepends
```

naveen521kk:
after you approve the PR @jeremyd2019, I'll squash commits to make some sense. Also, thanks for the review!

.ci/ci-check.py
jeremyd2019: in case that wasn't clear (or for somebody else reading this), pacdb's `Package.*depends` methods return a dict of package name to `DependsEntry` which has the split out name, "mod" (`<`, `>`, `<=`, etc...), version_str (as in, not a `Version` object, since most callers aren't interested in the version, save the time parsing it), and optional `desc`ription (optdepends uses those).  So you can just check if the package name is in the dict

naveen521kk: Yeah, I would've also used `in`, but the code was initially checking whether any of the strings ends with `python-installer`. Later, I saw that ucrt is used everwhere and ended up with this code.

jeremyd2019 (approved):

.ci/ci-check.py
jeremyd2019: ok, somebody looking at the raw data could also have assumed that it returned the raw list of strings, with optional version constraint and optional description still included, and in that case startswith would have been a reasonable way to search it.

.ci/ci-check.py
```diff
@@ -42,6 +50,16 @@ def get_msys2_root():
...
 
 
+@cache
+def get_pacdb_database():
+    return pacdb.Database("ci", TEMP_REPO_PATH / "ci.db.tar.gz")
+
+
+@cache
+def get_pacdb_files_database():
+    return pacdb.Database("files", TEMP_REPO_PATH / "ci.files.tar.gz")
```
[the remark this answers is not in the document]
naveen521kk: > Only other observation is that you don't need to load both regular and files dbs, all information is present in the files db (it's a superset of the normal db, with files list added).

I see, never knew that, though something would be missing with the files db. I'll change that to also open files db in that case.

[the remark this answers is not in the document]
naveen521kk: I've done https://github.com/msys2/MINGW-packages/pull/16755/commits/c3250d78655b8fcf32fb6af4d263c6b011c219fb

naveen521kk:
This PR is ready to be merged :)

PR #16771, opened 2023-04-12
mcfgthread: Update to v1.5-beta.1

lhmouse:
TLS and per-thread exit callbacks are now functional in threads that have been created by other libraries, e.g. by CreateThread() or pthread_create(). (https://github.com/lhmouse/mcfgthread/issues/81)


lhmouse:
Thank you!

PR #16772, opened 2023-04-12
nspr: Use nsinstall from msys2 package

Biswa96:
This helps to remove all the moztools nsinstall workarounds.
    Also the moztools' nsinstall can not be used without makepkg
    becuase it shows Permission denied error.


Biswa96:
There is an error in CI log https://github.com/msys2/MINGW-packages/actions/runs/4679131456/jobs/8288739482#step:11:42

PR #16819, opened 2023-04-14
imath: enable python bindings for clang builds

kmilos:
`0002-fix-openexr-clang-build.patch` seems no longer necessary for CLang 16

Even before this change, the libPyImath implib is 90MB, no idea what gives... links python lib in statically?


MehdiChinoune:
`0002-fix-openexr-clang-build.patch` was added to fix building OpenEXR, have you try it?

kmilos:
Ah, I thought it was restricted to Imath... I'll bring it back, sorry.

PR #16828, opened 2023-04-16
python-winsdk: Enable building on all mingw arch

mimi89999:
I don't have an arm64 computer with Windows, so I couldn't test clangarm64.


Biswa96:
> I don't have an arm64 computer with Windows, so I couldn't test clangarm64.

I can try to build it in rpi4 .What tests do you need to perform?

mimi89999:
I don't have any tests. The only place where it is used is https://dev.gajim.org/gajim/gajim/-/merge_requests/927

We could also test the notification libraries like https://github.com/DatGuy1/Windows-Toasts

Biswa96:
Would you like to add clangarm64? @hmartinez82 confirmed that it can be compiled for arm64.

mimi89999:
@Biswa96 I updated the PR

PR #16830, opened 2023-04-17
sndfile: enable MPEG Encode/Decode Support

fabiangreffrath:
Uses libmpg123 for decode, liblame for encode.


fabiangreffrath:
> Would you like to bump the pkgrel value please?

Sure, thanks!

Issue #16837, opened 2023-04-18
openxr : missing validation layer


sdegrande:
Could it be possible to build and add the validation layer (https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/src/api_layers/README_core_validation.md) in the openxr package ?
It's quite mandatory when developing and debugging an OpenXR code...

Thanks for your great work - before msys2/mingw64 my life as a multi-platform dev was so hard ! :-)

Biswa96:
Thank you for reporting this issue.

* It seems that the mingw package is using OpenXR-SDK repository. The validation layers are installed if OpenXR-SDK-Source repository is used. Would like to make sure that OpenXR-SDK-Source repository should be used for packaging? 
* Do you have any test to make sure that the validation layer is installed in correct directory? There was an issue with install path in vulkan validation layer package. After using the OpenXR-SDK-Source repository, the validation layers are installed like the following paths

```diff
 ucrt64/
 ucrt64/bin/
+ucrt64/bin/api_layers/
+ucrt64/bin/api_layers/XrApiLayer_api_dump.json
+ucrt64/bin/api_layers/XrApiLayer_core_validation.json
+ucrt64/bin/api_layers/libXrApiLayer_api_dump.dll
+ucrt64/bin/api_layers/libXrApiLayer_core_validation.dll
+ucrt64/bin/hello_xr.exe
 ucrt64/bin/libopenxr_loader.dll
+ucrt64/bin/list_json.exe
+ucrt64/bin/openxr_runtime_list.exe
 ucrt64/include/
 ucrt64/include/openxr/
 ucrt64/include/openxr/openxr.h
```

Biswa96:
I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request?

PR #16841, opened 2023-04-18
openxr: Build api layers

Biswa96:
* Use the OpenXR-SDK-Source repository which has the api_layers directory.
* Add DEF with source list to export proper APIs.

Fixes https://github.com/msys2/MINGW-packages/issues/16837

### DIFF
--- a/mingw-w64-openxr/001-use-def-file-on-mingw-w64.patch
+++ b/mingw-w64-openxr/001-use-def-file-on-mingw-w64.patch
@@ -1,3 +1,21 @@
+--- a/src/api_layers/CMakeLists.txt
++++ b/src/api_layers/CMakeLists.txt
+@@ -63,6 +63,7 @@
+ 
+ add_library(XrApiLayer_api_dump SHARED
+     api_dump.cpp
++    XrApiLayer_api_dump.def
+     ${PROJECT_SOURCE_DIR}/src/common/hex_and_handles.h
+     # target-specific generated files
+     ${GENERATED_OUTPUT}
+@@ -129,6 +130,7 @@
+ 
+ add_library(XrApiLayer_core_validation SHARED
+     core_validation.cpp
++    XrApiLayer_core_validation.def
+     ${PROJECT_SOURCE_DIR}/src/common/hex_and_handles.h
+     ${PROJECT_SOURCE_DIR}/src/common/object_info.cpp
+     ${PROJECT_SOURCE_DIR}/src/common/object_info.h
 --- a/src/loader/CMakeLists.txt
 +++ a/src/loader/CMakeLists.txt
 @@ -7,7 +7,7 @@
--- a/mingw-w64-openxr/002-libname-def-file.patch
+++ b/mingw-w64-openxr/002-libname-def-file.patch
@@ -1,3 +1,25 @@
+--- a/src/api_layers/XrApiLayer_api_dump.def
++++ b/src/api_layers/XrApiLayer_api_dump.def
+@@ -23,7 +23,7 @@
+ ;
+ ;;;;  End Copyright Notice ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+ 
+-LIBRARY XrApiLayer_api_dump
++LIBRARY libXrApiLayer_api_dump.dll
+ EXPORTS
+ xrNegotiateLoaderApiLayerInterface
+ 
+--- a/src/api_layers/XrApiLayer_core_validation.def
++++ b/src/api_layers/XrApiLayer_core_validation.def
+@@ -23,7 +23,7 @@
+ ;
+ ;;;;  End Copyright Notice ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+ 
+-LIBRARY XrApiLayer_core_validation
++LIBRARY libXrApiLayer_core_validation.dll
+ EXPORTS
+ xrNegotiateLoaderApiLayerInterface
+ 
 --- a/src/loader/openxr-loader.def
 +++ a/src/loader/openxr-loader.def
 @@ -7,7 +7,7 @@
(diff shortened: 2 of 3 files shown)


sdegrande:
I downloaded the artifact. The validation layer is still not found.

I looked at the openxr loader spec (https://registry.khronos.org/OpenXR/specs/1.0/loader.html#api-layer-discovery) and the json files are not where they are expected to be (should be in /usr/share).
Moreover, some binaries are now missing (they were in the previous package) : hello_xr, openxr_runtime_list.

I know nothing about package creation using pacman, but it appears that the Arch openxr package contains all what is needed, at the 'right' place. See : https://archlinux.org/packages/community/x86_64/openxr/

Thanks for your work !

Biswa96:
Could you provide an simple example program to test the api_layers file path?

> the json files are not where they are expected to be (should be in /usr/share).

The path may not be same with Linux packages. For example, see vulkan-validation-layers package 
* Arch: /usr/share/vulkan/explicit_layer.d/VkLayer_khronos_validation.json
* mingw: /mingw64/bin/VkLayer_khronos_validation.json

> Moreover, some binaries are now missing (they were in the previous package) : hello_xr, openxr_runtime_list.

I do not see them in older openxr packages. Those fail to build in clang64. Though I can try to workaround the linker error.

sdegrande:
> Could you provide an simple example program to test the api_layers file path?
> 

A simple example program is not enough. The openxr package provides a 'loader' which connect the application to an openxr runtime. So you need to have such a runtime installed (it can be SteamVR or Monado) to be able to run an openxr application.

> > the json files are not where they are expected to be (should be in /usr/share).
> 
> The path may not be same with Linux packages. For example, see vulkan-validation-layers package
> 
>     * Arch: /usr/share/vulkan/explicit_layer.d/VkLayer_khronos_validation.json
> 
>     * mingw: /mingw64/bin/VkLayer_khronos_validation.json
> 

But then it's not "standard compliant"... strange...

> 
> > Moreover, some binaries are now missing (they were in the previous package) : hello_xr, openxr_runtime_list.
> 
> I do not see them in older openxr packages. Those fail to build in clang64. Though I can try to workaround the linker error.

Oh, my mistake then, I thought they were installed... I probably mix up with my other computer (running on Linux).

Allright, I'll build it by myself, and let you know what I find...

sdegrande:
If you want to be coherent with the vulkan package, then you can keep this build as is, I guess... The workaround is simple, it just needs to set a global envvar.

Biswa96:
For Windows platform, the loader seems to search for the api layers in registry `HKEY_LOCAL_MACHINE\SOFTWARE\Khronos\OpenXR\1\ActiveRuntime`[^1]. I do not know if the logic should be modified further for mingw package.

[^1]: https://flightsimulator.zendesk.com/hc/en-us/articles/360018555179-How-to-install-OpenXR-

sdegrande:
> For Windows platform, the loader seems to search for the api layers in registry `HKEY_LOCAL_MACHINE\SOFTWARE\Khronos\OpenXR\1\ActiveRuntime. I do not know if the logic should be modified further for mingw package.

That's what the spec defines for Windows. But msys2/mingw64 purpose is to 'mimic' the Unix concepts, one of them being to use config files rather than a database, as you obviously know. That's what makes msys2/mingw64 so easy to use for multi-platform development, in my sense.

So, in my humble opinion, you should not try to adapt the openxr package to the windows way of doing...

Biswa96:
Agree. I shall look into it later.

sdegrande:
Here are 2 small patches to:
-  install the layers files (json manifest, dll, dll.a) in the 'standard' XDG dirs
- enable the user of the XDG_DATA_DIRS envvar

With this 2 additional patches,  it works the same way than on Linux., and so it is no more needed to use the XR_API_LAYER_PATH envvar to help the loader find the layer manifests.

Note that the 2nd patch is rather 'hackish'. Perhaps the right way would be to have XR_OS_LINUX being defined with mingw, but there could be some unknown side effects...

[0001-xdg_conform_install.zip](https://github.com/msys2/MINGW-packages/files/11285462/0001-xdg_conform_install.zip)
[0002-use_xdg_date_dirs.zip](https://github.com/msys2/MINGW-packages/files/11285470/0002-use_xdg_date_dirs.zip)

sdegrande:
(nb: perhaps the same kind of patches should be applied to the Vulkan packages ?)

sdegrande:
I talked with guys from Collabora/OpenXR.

When build on Windows (even with Mingw64), the current OpenXR-SDK-Source is set to produce an openxr loader meant to be use in applications run "from" Windows (by double-clicking on the application's icon, for instance). And so it follows Windows standards and uses the registry to find the API layer manifests.
Using the loader in an application run "from" Msys2, and so using Unix standards, is not something that was considered.

There's however a central point: the loader has to find the active runtime to interface to it. And OpenXR runtimes for Windows (such as SteamVR) do store their data into the registry... So it's not possible to avoid the use of the registry.

The patch that I proposed is an "in-between", actually. The runtime is searched in the registry and the layers in XDG_DATA_DIRS. That's not a so good solution...

So, it would mean that if you want the openxr package to work "out of the box", then registry is to be filled with the path of the layer directory, during package's installation. 

Is that something that can be done, and in conformance with the Msys2/Mignw64 "packaging rules" ?

sdegrande:
Ah... annoying. It means that a user's action is needed after the installation... without the user knowing that this has to be done...

Biswa96:
> without the user knowing that this has to be done...

It is possible to show a message to add the environment variable or the registry value when user is installing the package.

sdegrande:
Ok. So here is the needed registry key that must be added in order for the loader to find the API layers manifests:
```
[HKEY_LOCAL_MACHINE\SOFTWARE\Khronos\OpenXR\1\ApiLayers\Explicit]
"C:\\msys64\\clang64\\bin\\api_layers\\XrApiLayer_api_dump.json"=dword:00000000
"C:\\msys64\\clang64\\bin\\api_layers\\XrApiLayer_core_validation.json"=dword:00000000
```

If modification of HKEY_LOCAL_MACHINE needs some admin rights, then HKEY_CURRENT_USER can be used instead.

PR #16846, opened 2023-04-18
Add findutils

cyrilarnould:
Patch 001 adds the poll and windows-stat-inodes gnulib modules. The poll module provides the poll functionality that is absent on mingw. windows-stat-inodes ensures the -inum option of find is working properly. The modules have been taken from the same gnulib commit that is linked in the findutils 4.9.0 release (v0.1-5153-g6ef3d78333).

Patch 002 modifies the configure script to allow using the mountlist module on mingw. The code doesn't compile without it but it also doesn't seem broken. The mountlist module also specifies that using the header is fine on unsupported systems.

Patches 003, 004, 006, 007, 010 and 015 have been adapted from Eli Zaretskii's ezwinports:

https://sourceforge.net/projects/ezwinports/files

Patches 005, 009, 013 and 014 disable features that had not been implemented yet in the 4.2.30 findutils release that ezwinports is based on. I also don't know how I'd implement them; they are all based on functionality that is not available on mingw as far as I can tell.

Patches 008 and 011 are fairly straightforward compilation error fixes.

Patch 009 tries to address the differences in how the PATH environment variable is separated on Windows.

Patch 016 fixes linking on clang builds.

Note that not all of the tests pass. I could provide patches for some of them if this is preferable, but for others I think there's no use case on Windows. I keep a write-up here:

https://github.com/cyrilarnould/findutils/blob/2ccb0be2c1c70d791f051d24b9dba73e64bdb446/README-mingw

### DIFF
--- a/mingw-w64-findutils/001-add-gnulib-modules.patch
+++ b/mingw-w64-findutils/001-add-gnulib-modules.patch
@@ -1,0 +1,3359 @@
+
+--- a/gl/lib/Makefile.am
++++ b/gl/lib/Makefile.am
+@@ -102,6 +102,7 @@
+ #  parse-datetime \
+ #  pathmax \
+ #  perror \
++#  poll \
+ #  progname \
+ #  quotearg \
+ #  readlink \
+@@ -145,6 +146,7 @@
+ #  warnings \
+ #  wchar \
+ #  wcwidth \
++#  windows-stat-inodes \
+ #  xalloc \
+ #  xalloc-die \
+ #  xgetcwd \
+@@ -2269,6 +2271,42 @@
+ 
+ ## end   gnulib module pipe-posix
+ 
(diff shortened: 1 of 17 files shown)


brisingraerowing:
There's a Rust reimplementation of Findutils (and Coreutils) by the [UUtils Project](https://github.com/uutils), though Findutils doesn't build with Clang 16 ATM.

cyrilarnould:
I've noticed that `CC="gcc -m64"` will instruct gcc to create a 64 bit binary, which of course makes no sense for a mingw32 build. For lack of a better option I'm using `--disable-year2038`.

@brisingraerowing I'm not sure how this relates to my pull request.

cyrilarnould:
I've just discovered that `git format-patch` is a thing, I want to create new patches accordingly and then try this again.

PR #16853, opened 2023-04-20
GMT: add new package

ZhengXueke114:
GMT is a Generic Mapping Tools and provides C API.

It has been tested on the UCRT64 and works well


mingw-w64-gmt/0001-changes-needed-to-build-with-mingw.patch
```diff
@@ -0,0 +1,47 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index 9acc139..af1276c 100644
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -530,9 +530,9 @@ if (HAVE_DLADDR AND HAVE_LIBDL)
+ 	target_link_libraries (gmtlib dl)
+ endif (HAVE_DLADDR AND HAVE_LIBDL)
+ 
+-if (MSVC AND FFTW3_FOUND)
++if ((MSVC OR MINGW) AND FFTW3_FOUND)
```
kmilos, line 10: Could replace w/ just `if(WIN32 AND FFTW3_FOUND)` as this is platform, and not compiler specific.

Has this been reported upstream?

xiaoke0O: Yes, this patch is from upstream.

https://github.com/GenericMappingTools/gmt/pull/7286

kmilos: Ah, thanks, then we should directly download from the upstream PR instead of maintaining separately...

PR #16854, opened 2023-04-20
opus: Update to 1.4; port to meson

lazka:


lazka:
https://github.com/xiph/opus/pull/267

PR #16847, opened 2023-04-18
[DRAFT] rust: update to pre-release 1.69.0

filnet:
see https://blog.rust-lang.org/inside-rust/2023/04/17/1.69.0-prerelease.html

PR #16855, opened 2023-04-21
clang: fix Any linker error with multiple compilers

filnet:
reverts this change: https://github.com/llvm/llvm-project/commit/95b27b2a1e0e9fe0781ec76015b4016b9d04423d

fixes the Rust build, see https://github.com/msys2/MINGW-packages/pull/16847

### DIFF
--- a/mingw-w64-clang/0005-Fix-Any-linker-error-with-multiple-compilers.patch
+++ b/mingw-w64-clang/0005-Fix-Any-linker-error-with-multiple-compilers.patch
@@ -1,0 +1,11 @@
+--- a/include/llvm/ADT/Any.h
++++ b/include/llvm/ADT/Any.h
+@@ -124,7 +124,7 @@
+   std::unique_ptr<StorageBase> Storage;
+ };
+ 
+-template <typename T> char Any::TypeId<T>::Id = 0;
++template <typename T> char Any::TypeId<T>::Id = 1;
+ 
+ template <typename T>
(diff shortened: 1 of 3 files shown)


filnet:
Please consider merging. New Rust version is dependent on this new version of  LLVM.

lazka:
Can we change it to the new workaround?

filnet:
> Can we change it to the new workaround? Or would you prefer to do that later?

I'll adapt the patch to the new workaround.

filnet:
C'est parti...

mati865:
I think this patch shouldn't be merged for LLVM based subsystems but since it's merged already we can keep probably keep it (unless something regress).

filnet:
> I think this patch shouldn't be merged for LLVM based subsystems but since it's merged already we can keep probably keep it (unless something regress).

Not sure I understand but the patch should land upstream sometimes in the future.

mati865:
Oh, it's straight from the phabricator. Sorry, missed that bit.
So let's just keep an eye on it and see if sticks or gets reverted.

filnet:
It is not yet merged : https://reviews.llvm.org/D148953

filnet:
The fix has been merged upstream.

For sake of completeness, an alternative solution mentioned in the fix discussion:
> One could stand on the point that if objects from different compilers are linked, the user needs to supply -f[no-]zero-initialized-in-bss to make sure they behave the same.

We should keep that in mind should a similar issue popup.

PR #16925, opened 2023-04-22
mapserver: update to 8.0.1

MehdiChinoune:


mingw-w64-mapserver/PKGBUILD
```diff
@@ -73,14 +69,15 @@ build() {
     extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
   fi
 
+  CFLAGS+=" -Wno-implicit-function-declaration" \
```
Biswa96, line 72: Would you like to check if adding the following patch file fixes that compiler error?

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,7 +275,7 @@
 if(WIN32)
   set(REGEX_SOURCES ${REGEX_DIR}/regex.c)
   include_directories(${REGEX_DIR})
-  add_definitions(-DREGEX_MALLOC -DUSE_GENERIC_MS_NINT -DHAVE_STRING_H)
+  add_definitions(-DREGEX_MALLOC -DUSE_GENERIC_MS_NINT -DHAVE_STRING_H -DSTDC_HEADERS)
   add_definitions(-D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE)
   set(REGEX_MALLOC 1)
   set(USE_GENERIC_MS_NINT 1)
```

MehdiChinoune: I prefer cflags over patches, the compiler error occurs only with Clang>=16

Biswa96: OK, understood. I have sent that patch to upstream. The CFLAGS change is not required after adding that patch. Thanks for the update.

PR #16934, opened 2023-04-22
whisper.cpp: Update to 1.3.0

lazka:


mingw-w64-whisper.cpp/PKGBUILD
```diff
@@ -43,7 +43,8 @@ build() {
     extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
   fi
 
-  CFLAGS+=" $(pkgconf.exe --cflags openblas)"
+  CFLAGS+=" $(pkgconf.exe --cflags openblas) "
+  CXXFLAGS+=" -D_WIN32_WINNT=0x603"
```
Biswa96, line 47: How about disabling mingw64 and mingw32? Doesn't adding that CXXFLAGS already break win7 support?

lazka: yeah, I agree, but it seemed easy enough to fix *shrug*

lazka: I've removed mingw32/64

PR #16930, opened 2023-04-22
Add blueprint-compiler

realmazharhussain:
Closes #16929

PR #16939, opened 2023-04-23
Fix: blueprint-compiler directory has no 'mingw-w64-' prefix

realmazharhussain:
Somehow I and everyone didn't notice in #16930 that unlike every other package in this repository, the directory for blueprint-compiler didn't have `mingw-w64-` prefix in its name. So, this PR exists to fix that.


Biswa96:
Thanks.

PR #16945, opened 2023-04-24
cargo-c: update to 0.9.19

MehdiChinoune:
@hmartinez82 Could you please test building it on clangarm64. last time 0.9.18 failed to build.

### DIFF
--- a/mingw-w64-cargo-c/PKGBUILD
+++ b/mingw-w64-cargo-c/PKGBUILD
@@ -28,8 +28,24 @@ prepare() {
 
     patch -p1 -i "${srcdir}"/001-add-lib-prefix-on-windows-gnu.patch
 
-    cargo fetch \
-        --locked
+    cargo vendor \
+        --locked \
+        --versioned-dirs
+
+    mkdir -p .cargo
+    cat >> .cargo/config.toml <<END
+[source.crates-io]
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = "vendor"
+END
+
+
+    sed -i -e 's/"cfg(all(target_arch = \\"\([^\\]\+\)\\", target_env = \\"gnu\\", target_abi = \\"llvm\\", not(windows_raw_dylib)))"/"\1-pc-windows-gnullvm"/g' \
+        "vendor/windows-targets-0.48.0/Cargo.toml"
+    sed -i -e 's/\("Cargo.toml":\)"[^"]\+"/\1"'"$(sha256sum "vendor/windows-targets-0.48.0/Cargo.toml" | cut -d' ' -f1)"'"/' \
+        "vendor/windows-targets-0.48.0/.cargo-checksum.json"
 }
 
 build() {
--- a/mingw-w64-cargo-c/PKGBUILD
+++ b/mingw-w64-cargo-c/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=cargo-c
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.9.17
-pkgrel=3
+pkgver=0.9.19
+pkgrel=1
 pkgdesc='A cargo subcommand to build and install C-ABI compatibile dynamic and static libraries (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang32' 'clang64' 'clangarm64')
@@ -18,8 +18,8 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-rust")
 source=("${_realname}-${pkgver}.tar.gz"::"https://github.com/lu-zero/cargo-c/archive/v${pkgver}.tar.gz"
         "${_realname}-${pkgver}.Cargo.lock"::"https://github.com/lu-zero/cargo-c/releases/download/v${pkgver}/Cargo.lock"
         001-add-lib-prefix-on-windows-gnu.patch)
-sha256sums=('95fcf50505516dfa6a941e666d1388810da9d7a9e1c623c09068faef5d50b3b9'
-            '53a6340faee4ae1c56ca8a61087908b0fccce0352a070de404bcce05d20f369a'
+sha256sums=('c2633ff22e52da9985394f30c8ef5e9abbac4d14c9b79e3690c8e95cf500ab97'
+            '7dc8d68efcda151e0b284654c9ca0a17af4cc8bbf4db88342297b016361ef0d3'
             'b749ffdb37eb6016dca37c41c2d644156bbd4a7d8ffaa7bc18c7813f7af646c0')
 
 prepare() {


jeremyd2019:
@mati865 

[CLANGARM64 verbose-build.log](https://github.com/msys2/MINGW-packages/files/11348861/verbose-build.log)

Notably, I see it compiling v0.42.2 but not v0.48.0 before the error, and there is no `output` that adds the link directory for v0.48.0 like there is for v0.42.2.

mati865:
Can you try to navigate to `windows-targets 0.48` sources (somewhere in `~/.cargo` dir) and manually replace:
```
[target.'cfg(all(target_arch = "aarch64", target_env = "gnu", target_abi = "llvm", not(windows_raw_dylib)))'.dependencies]
windows_aarch64_gnullvm = { path = "../../targets/aarch64_gnullvm", version = "0.48.0" }
```
with:
```
[target.aarch64-pc-windows-gnullvm.dependencies]
windows_aarch64_gnullvm = { path = "../../targets/aarch64_gnullvm", version = "0.48.0" }
```
Or similar if it differs but you should get an idea.

jeremyd2019:
In `Cargo.toml`, was
```
[target."cfg(all(target_arch = \"aarch64\", target_env = \"gnu\", target_abi = \"llvm\", not(windows_raw_dylib)))".dependencies.windows_aarch64_gnullvm]
version = "0.48.0"
```
now
```
[target.aarch64-pc-windows-gnullvm.dependencies.windows_aarch64_gnullvm]
version = "0.48.0"
```

Doesn't seem to have made any difference ☹️

mati865:
`[target.aarch64-pc-windows-gnullvm.dependencies.windows_aarch64_gnullvm]` this doesn't look quite right. There should be no second `windows_aarch64_gnullvm`.

jeremyd2019:
I don't know, Cargo.toml.orig looked closer to what you said, I changed that too but didn't help.  I looked at Cargo.lock, and it just had a list of all dependencies, and aarch64 gnullvm was listed in that.

jeremyd2019:
Let me know if there's anything else I can try to help figure this out.  I don't know much about rust/cargo, but I do have the arm64 setup to be able to test things.

jeremyd2019:
It looks like this also affected mdbook 0.4.29 in https://github.com/msys2-arm/msys2-autobuild/actions/runs/4973744023/jobs/8899790826?check_suite_focus=true

mati865:
Sorry for the radio silence. I've been struggling to reproduce it and feel burnt out.
I still believe I cannot reproduce it because of `target_abi = "llvm"` which is nightly only feature but when cross-compiling I had to use nightly compilers. So I've tried to produce native stable x86_64-pc-windows-gnullvm compiler but encountered crashes inside LLVM (haven't investigated, some optimization passes are unhappy).

The last thing I can think of to try without having access to the reproducer is to:
1. wipe Cargo registry cache (everything inside `$USERPROFILE/.cargo/registry`)
2. do the build, it will fail but Cargo will get deps
3. replace all `Cargo.toml` and `Cargo.toml.orig` (this one shouldn't be necessary though) in all `windows-targets-0.48.*` crates located in directories inside `$USERPROFILE/.cargo/registry/src/*`
4. try building again with clean `target/` dir, it can be wiped with `cargo clean`

jeremyd2019:
I will try to look at this after llvm issues are sorted out...  by "do the build" do you mean `prepare` and `build` steps or just `build`?  keep in mind you are talking to somebody with 0 knowledge of rust and cargo, so the PKGBUILD is my point of reference on how to build these...

mati865:
> by "do the build" do you mean `prepare` and `build` functions or just `build`?

Sorry, I've been thinking about build as `cargo build` so yeah you should skip prepare step when running makepkg second time (`-e`).

jeremyd2019:
No go.  New thought: this data seems to be recorded in binary files in `.cargo/registry/index/github.com-1ecc6299db9ec823/.cache/wi/nd/windows-targets`.  I may try to hack that file a bit, but maybe that's why it's not picking up the changes to Cargo.toml under src?

jeremyd2019:
Yes, after manually editing the targets in that binary index file, this package built successfully (after removing target/ and `makepkg -e`).

mati865:
Oh, that this the new format for crates sources, I thought it wasn't enabled on stable yet and haven't manually modified sources in a long time. Anyway that's fantastic news confirming my assumption.
Tomorrow I'll prepare PR to windows-rs crate and once new version is up all those build issues should be gone,

jeremyd2019:
planning for the next steps, does anybody remember what all will be unblocked by this being fixed?  I only remember this and mdbook (which will now require reverting 11945b4236e596b0ffeb819cd35e672855e95fad)

Biswa96:
Try to build/update every rust packages in clangarm64.

jeremyd2019:
add rav1e to the list of packages broken by this ☹️.  Let me know if/when there's something I should try wrt microsoft/windows-rs#2515

jeremyd2019:
This seems to be sufficient to get it to build.  Probably not advisable though.  (Is there an option to make it not use `$USERPROFILE` but put these files under `$srcdir`?  Might be slightly less nasty)

```diff
diff --git a/mingw-w64-cargo-c/PKGBUILD b/mingw-w64-cargo-c/PKGBUILD
index 3651ac563..6eecd40c0 100644
--- a/mingw-w64-cargo-c/PKGBUILD
+++ b/mingw-w64-cargo-c/PKGBUILD
@@ -30,6 +30,8 @@ prepare() {

     cargo fetch \
         --locked
+    sed -i -e 's/"cfg(all(target_arch = \\"\([^\\]\+\)\\", target_env = \\"gnu\\", target_abi = \\"llvm\\", not(windows_raw_dylib)))"/"\1-pc-windows-gnullvm"/g' \
+        "$USERPROFILE/.cargo/registry/index/github.com-1ecc6299db9ec823/.cache/wi/nd/windows-targets"
 }

 build() {
```

mati865:
I guess doing `cargo vendor` would pull those deps to some dir inside `$srcdir` but I'll have to check.

mati865:
Some POC:
1. This will download sources to "${srcdir}/${_realname}-${pkgver}/vendor", you might want to add -q to skip the output:
```
$ cargo vendor
```
2. Add vendor info to Cargo config:
```
$ echo '[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"' >> .cargo/config.toml
```
3. Test by renaming the dependency to some non-existent:
```
$ sed -i 's/x86_64/broken/g' vendor/windows-targets/Cargo.toml

$ cargo b
error: no matching package named `windows_broken_gnu` found
location searched: registry `crates-io`
required by package `windows-targets v0.48.0`
    ... which satisfies dependency `windows-targets = "^0.48.0"` (locked to 0.48.0) of package `windows-sys v0.48.0`
    ... which satisfies dependency `windows-sys = "^0.48.0"` (locked to 0.48.0) of package `filetime v0.2.21`
    ... which satisfies dependency `filetime = "^0.2.9"` (locked to 0.2.21) of package `cargo v0.70.1`
    ... which satisfies dependency `cargo = "^0.70.0"` (locked to 0.70.1) of package `cargo-c v0.9.19+cargo-0.70 (D:\tmp\cargo-c)`
```

jeremyd2019:
what about `.cargo-checksum.json`?
```
==> Starting build()...
error: the listed checksum of `C:\_\mingw-w64-cargo-c\src\cargo-c-0.9.19\vendor\windows-targets-0.48.0\Cargo.toml` has changed:
expected: 189df185f8620816d1b23dedd0eaadb652099cd9459f6ed25327fe98dda598ec
actual:   5a60839e5b2950d1a68b46fca6db2fa75e8c36761cb69f376c1718bb8de63888

directory sources are not intended to be edited, if modifications are required then it is recommended that `[patch]` is used with a forked copy of the source
```

jeremyd2019:
This seems to work:

```diff
diff --git a/mingw-w64-cargo-c/PKGBUILD b/mingw-w64-cargo-c/PKGBUILD
index 3651ac563..43bb160a3 100644
--- a/mingw-w64-cargo-c/PKGBUILD
+++ b/mingw-w64-cargo-c/PKGBUILD
@@ -28,8 +28,24 @@ prepare() {

     patch -p1 -i "${srcdir}"/001-add-lib-prefix-on-windows-gnu.patch

-    cargo fetch \
-        --locked
+    cargo vendor \
+        --locked \
+        --versioned-dirs
+
+    mkdir -p .cargo
+    cat >> .cargo/config.toml <<END
+[source.crates-io]
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = "vendor"
+END
+
+
+    sed -i -e 's/"cfg(all(target_arch = \\"\([^\\]\+\)\\", target_env = \\"gnu\\", target_abi = \\"llvm\\", not(windows_raw_dylib)))"/"\1-pc-windows-gnullvm"/g' \
+        "vendor/windows-targets-0.48.0/Cargo.toml"
+    sed -i -e 's/\("Cargo.toml":\)"[^"]\+"/\1"'"$(sha256sum "vendor/windows-targets-0.48.0/Cargo.toml" | cut -d' ' -f1)"'"/' \
+        "vendor/windows-targets-0.48.0/.cargo-checksum.json"
 }

 build() {
```

I went with `--versioned-dirs` because I wanted to make it clear which version of `windows-targets` I was messing with.

jeremyd2019:
Successful ARM64 run: https://github.com/msys2-arm/MINGW-packages/actions/runs/5105668092/jobs/9177389129

mati865:
Interesting, it didn't show up for me locally. Maybe because of `--frozen` added to `cargo build`.

jeremyd2019:
I noticed the error about the checksum mismatch mentioned a `[patch]` section, and I figured that would be the recommended way to patch a dependency, but it seems you need a whole fork of the dependency, not just a patch against it.

It might have been more readable to use a patch file instead of these couple of `sed`s, but hopefully a new version of windows-targets will be released soon, and this hack won't be needed very long.

PR #16953, opened 2023-04-25
libplacebo: build static lib in proper way

kasper93:


lazka:
"default-library both" doesn't really work on Windows if the library has dependencies, so I wonder what is proper about this?

kasper93:
> "default-library both" doesn't really work on Windows if the library has dependencies

Why? How do I reproduce? I know it may have issues, but so far people are telling me that it "will not work", while it is working.

> so I wonder what is proper about this?

Not building twice over the same thing. If there is good reason not to do that ever, it would need to be fixed on meson side to disable this feature.

> `grep -nrI default-library | grep both | wc -l`

50. That's how much MINGW packages uses `both` in this repository. I explicitly checked before if it is used.

eli-schwartz:
It probably depends on the software. libplacebo uses dllexport in a macro, so that gets baked into the *.o files. Other projects may use *.def files which apply export mappings at link time. Using the dllexport approach, compiling all object files a single time means the static library uses dllexport. Using the def file approach, only the shared DLL is created with exported symbols.

kasper93:
> It probably depends on the software. libplacebo uses dllexport in a macro, so that gets baked into the *.o files. Other projects may use *.def files which apply export mappings at link time. Using the dllexport approach, compiling all object files a single time means the static library uses dllexport. Using the def file approach, only the shared DLL is created with exported symbols.

I know how it works, that's exactly why I made this patch. Generally current PKGBUILD will not work for next libplacebo version, because we generate different .pc for static/shared/both version. If you overwrite static  .pc with shared one, static library will not linkable. And the other way could work, but not always, because dllimport is required for data imports from dlls.

Biswa96:
> because we generate different .pc for static/shared/both version.

Here is the diff between the pkgconfig file in shared and static build.

```diff
--- UCRT64-shared/meson-private/libplacebo.pc
+++ UCRT64-static/meson-private/libplacebo.pc
@@ -16,7 +16,6 @@
 Name: libplacebo
 Description: Reusable library for GPU-accelerated video/image rendering
 Version: 6.274.0
-Requires.private: shaderc >= 2019.1, spirv-cross-c-shared >= 0.29.0, vulkan, lcms2 >= 2.9, dovi >= 1.6.7
-Libs: -L${libdir} -lplacebo
-Libs.private: -lm -lshlwapi -lversion
-Cflags: -I${includedir}
+Requires: shaderc_combined >= 2019.1, spirv-cross-c-shared >= 0.29.0, vulkan, lcms2 >= 2.9, dovi >= 1.6.7
+Libs: -L${libdir} -lplacebo -lm F:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../lib/libshlwapi.a F:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../lib/libversion.a
+Cflags: -I${includedir} -DPL_STATIC
```

Is it possible to add `-DPL_STATIC` in Cflags.private instead of Cflags?

> If you overwrite static .pc with shared one, static library will not linkable.

One can add `CFLAGS="-DPL_STATIC"` manually for static linking. Though that may break some use cases at first time, shared library is the first choice in mingw packages.

kasper93:
> Is it possible to add -DPL_STATIC in Cflags.private instead of Cflags?

That's kinda a problem. Meson doesn't support `Cflags.private` and while I did hack it, I want to avoid this on the off-chance it breaks. So it is used only for `both` build. If one build specific version it uses `Cflags`.

It is this part https://code.videolan.org/videolan/libplacebo/-/blob/master/src/meson.build#L223-229

```
extra_cflags = []
if get_option('default_library') == 'static'
  extra_cflags = ['-DPL_STATIC']
elif get_option('default_library') == 'both'
  # meson doesn't support Cflags.private, insert it forcefully...
  extra_cflags = ['\nCflags.private:', '-DPL_STATIC']
endif
```

kasper93:
> IIUC, while fixing the .pc file, the changes from this PR would break the symbols that are exported from the static library (which is arguably worse).

Understand correctly what? Could you elaborate, because I don't understand what you are saying. This PR doesn't break anything, neither with old release, nor new one.

Sorry, but I'm tired of people saying that "something" may have not work or other similar things. If you have concrete use-case that would not work I'm happy to fix that after evaluating the given case.

> If you prefer to not have the Cflags.private field by default in the .pc file upstream

I prefer to have it, but it is not supported by Meson and I want to minimize usage of hacks that may break. I would suggest asking Meson to add proper support for that in https://github.com/mesonbuild/meson/issues/10881

> would it be ok to add a special case for Windows (where default-library both can't be used)? Or alternatively (with lower preference), carry a patch here that uses the elif branch unconditionally?

Again, please tell me what configuration you think wouldn't work and I can fix it.

mmuetzel:
> Understand correctly what? Could you elaborate, because I don't understand what you are saying. This PR doesn't break anything, neither with the old release, nor the new one.
> 
> Sorry, but I'm tired of people saying that "something" may have not work or other similar things. If you have concrete use-case that would not work I'm happy to fix that after evaluating the given case

While it is possible on other platforms - e.g., Linux - to link the same object file to either a static or a shared library, that won't work with the `dllexport` mechanism on Windows.
When building a shared library the "standard Windows" way, you add a `dllexport` attribute to functions that should be exported from the `.dll`. When building a static library, you *must not* use a `dllexport` attribute for those functions. Since these attributes affect the symbols in the compiled object files, it is necessary to build the object files for that static library and the object files for the shared library and keep them apart.

> Again, please tell me what configuration you think wouldn't work and I will fix it.

If the object files that were built for a shared library are linked to a static library, this will likely cause problems for projects trying to link to that static library.

kasper93:
> When building a static library, you must not use a dllexport attribute for those functions.

Citation needed. Decorating function with dllexport adds them to export table, but the function itself doesn't change. When building static lib it may be not desirable that they got exported when linking to other shared library, but this is what linux does also when you not use `--exclude-libs`.

Only problems you can have when linking to shared lib and WANT to export something from static one, but if the object is not referenced it wouldn't be exported because linker would not look at that export table, but it is completely different problem and works just fine when using object files directly.

I'd love to be corrected, but so far no one provided any concrete reason, except vague claims. And since it works just fine in all cases, you need to give me more than that.

> If the object files that were built for a shared library are linked to a static library, this will likely cause problems for projects trying to link to that static library.

No, it does work. Like I said, please tell me what exactly will not work and I will fix that.

kasper93:
> Edit: That's essentially what you already wrote.

Yes and I'm aware of this behavior. But this PR is not about symbol conflict, but merely building the library in expected way, so the build system can know what's build and produce proper pkg-config file. Scenario you describing is not limited to Windows and should be handled on case by case basis.

To resolve your case, this is needed https://github.com/mesonbuild/meson/issues/3304 so feel free to move discussion there.

Also for whatever reason the `Cflags.private` issue https://github.com/mesonbuild/meson/issues/10881 has been completely removed.

eli-schwartz:
> Also for whatever reason the `Cflags.private` issue [mesonbuild/meson#10881](https://github.com/mesonbuild/meson/issues/10881) has been completely removed.

This was most certainly not done by the Meson team!!!

I checked in my emails for the notifications from that issue. It was originally filed by github user @shithappens2016, whose profile page now 404s. Funnily enough, the account still exists and in the preview box for this comment I am currently drafting, you can even see in the hover card:
> Committed to this repository
> Started 1 discussion in this repository

I suspect Github of some terrible, horrible, no-good very bad censorship which, for completely unwarranted and unacceptable reasons has decided to delete extremely valuable contributions to the Meson project.

Disgusting and revolting and entirely unethical.

Biswa96:
@kasper93 Could you provide what issue this change fixes? Just want to experiment with it.

kasper93:
> @kasper93 Could you provide what issue this change fixes? Just want to experiment with it.

Wrong libplacebo.pc because it is overwritten with shared only version, not "both" one. So static library is not linkable, because all symbols would be decorated with `dllimport`, because of missing `PL_STATIC`.

I see the update is ongoing in #19148 I can report proper issue with examples after #19148 is merged.

PR #16968, opened 2023-04-26
gcc: update to 13.1.0

MehdiChinoune:
based on @lhmouse fork


lhmouse:
Do you really want to make Intel syntax default? Some packages will require `-masm=att` in `CPPFLAGS` which was default.

MehdiChinoune:
> Do you really want to make Intel syntax default? Some packages will require `-masm=att` in `CPPFLAGS` which was default.

I thought It fixes some issues.

lhmouse:
Mostly looks good to me.

My only question is whether you would like to add `set disassembly-flavor intel` in `gdbinit`. That will have a little impact on user experience, but given almost all tools on Windows speak purely Intel syntax, it's less likely an issue.

lazka:
lgtm

lazka:
MINGW32 ada is broken somewhat:

The following packages fail with `raised SYSTEM.PERFECT_HASH_GENERATORS.TOO_MANY_TRIES : s-pehage.adb:693`

* mingw-w64-xmlada
* mingw-w64-ghdl
* mingw-w64-libgpr
* mingw-w64-gprbuild
* ...

MehdiChinoune:
This is not the first time we face an issue with Ada on i686, Maybe It's time to just disable it.
Who is using Ada? probably <1%
Who does really need i686 <2%
0.01*0.02=0.0002=0.02%
I don't think It worths the effort.

lazka:
Yeah. I'll not let this block other things if possible.

lazka:
I've reduced it down to:

```ada
with GNAT.Spitbol.Patterns;
procedure Broken is
   M : GNAT.Spitbol.Patterns.Match_Result;
   type Header_Symbol is (None, Name, Attr, Conv, Prag);
begin
   null
end Broken;
```

I'll file a bug report with this.

lazka:
Looking for the error there is https://github.com/AdaCore/gnat-llvm/issues/25#issuecomment-1057198363 which indicates the a similar issue we had last time, exception handling being broken.

lazka:
There is more brokenness with mingw32:

1)

* start gdb
* type "bt"

expected: "No stack."
actual: "terminate called after throwing an instance of 'gdb_exception_quit'"

2) cmake is broken:

terminate called after throwing an instance of '(anonymous namespace)::registry_error'
  what():  The system cannot find the file specified.

I'm going to revert this.

lazka:
Taking only "libgcc_s_dw2-1.dll" from gcc 12.2.0 fixes the issue. It also fixes the ada issue.

TheShermanTanker:
> 

:(

Oh well, back to using @lhmouse fork for reference then :P

TheShermanTanker:
```
    gcc-13.1.0.tar.xz ... FAILED (unknown public key 6C35B99309B5FA62)
```

Well...

Biswa96:
`/usr/bin/gpg --recv-keys 6C35B99309B5FA62`

lazka:
> You may try bisecting on my builds https://gcc-mcf.lhmouse.com/. I just checked 20221104 and it was fine.

`mingw-w64-gcc-mcf_20221104_13.0.0_x86-msvcrt` is the last working version, `mingw-w64-gcc-mcf_20221221_13.0.0_x86-msvcrt` is the first broken one

MehdiChinoune:
You could build snapshots from https://gcc.gnu.org/pub/gcc/snapshots/ to narrow it.

lazka:
Just looking at the libgcc history, only these two commits are in that range:

* https://github.com/gcc-mirror/gcc/commit/6e56633daae79f514b0e71f4d9849bcd8d9ce71f
* https://github.com/gcc-mirror/gcc/commit/1c118c9970600117700cc12284587e0238de6bbe

lazka:
https://github.com/gcc-mirror/gcc/commit/1c118c9970600117700cc12284587e0238de6bbe is at fault

lhmouse:
> [gcc-mirror/gcc@1c118c9](https://github.com/gcc-mirror/gcc/commit/1c118c9970600117700cc12284587e0238de6bbe) is at fault

Is this only ever used on `i686-w64-mingw32`, not on Linux or anywhere else?

lazka:
> Is this only ever used on `i686-w64-mingw32`, not on Linux or anywhere else?

I don't know.

Since the patch claims to be just a performance optimization I've reverted it and updated mingw32 as well now.

Lord-Nightmare:
I'm not entirely sure whether this is the best place to report this, but the mingw-w64-x86_64-gcc-13.1.0-1 package is the last version which will successfully compile MAME ( github.com/mamedev/mame ); all later versions which were released (13.1.0-3 and 13.1.0-4) fail with a bizarre state where the makefile fails during compiling of luaengine.o:
```make[2]: *** [frontend.make:774: ../../../../mingw-gcc/obj/x64/Release/src/frontend/mame/luaengine.o] Error 1```
Given the changes here seem to be the only major differences between 13.1.0-1 and 13.1.0-3 (barring the revert and recommit of 13.1.0 stuff) something in that range must be to blame, although I'm not entirely sure what.

**Comment by MehdiChinoune** [comment]
> github.com/mamedev/mame

Could you please, open an issue about it with instructions to How to reproduce the failure.

**Comment by lazka** [comment]
The main thing happening between 1 and 3 is that we enabled stack protector for gcc itself (by passing through our main CFLAGS)

**Comment by MehdiChinoune** [comment]
It occurs in #17092 and #17093.

**Comment by amyspark** [comment]
@lazka I have one based on libmp4v2 source code, it looks like a stack corruption issue in cc1plus (hence why it goes down without a pip, I suppose). I'm in the process of rebuilding the package pre-17094, and will let you know once I have the stacktrace symbolicated.

**Comment by amyspark** [comment]
@lazka Here's a repro of the crash I believe everyone's been experiencing. Basically, cc1plus blows up the stack: https://gist.github.com/amyspark/be93638fc5b5779594dd138aa8995860

**Comment by Lord-Nightmare** [comment]
I can confirm that #17094 fixes the MAME compile crash.

**Comment by nvgeele** [comment]
I can't comment on MAME, but I can confirm that there were issues with the previous versions of the GCC package, that are fixed with #17094. I can't share any code as its confidential, but I also got `Error 1` from `make` with some source files. I `strace`'d GCC during the compilation of those files, and I noticed it crashes due to a stack overflow. So it does seem that it is very much related to the stack protection you mentioned @lazka.

**Comment by lazka** [comment]
@nvgeele it turned out that my change here in 518e217a129eeeb92e509 broke it by setting BOOT_LDFLAGS. While that works on Linux because there is no default, on Windows BOOT_LDFLAGS defaults to extending the stack size to something larger like the default on Linux, so then large stack usage would then stack overflow.

I'll try to add some regression tests for this so this doesn't happen again. Sorry for the breakage.

**Comment by sezero** [comment]
Since this gcc-13 upgrade, we are receiving a `-Warray-bounds` warning in
libxmp x86 builds, but not in x64 builds:
https://github.com/libxmp/libxmp/actions/runs/5674163749/job/15380900650

```
CC src/depackers/mmcmp.lo
In function 'block_unpack_8bit',
    inlined from 'decrunch_mmcmp' at src/depackers/mmcmp.c:441:8:
src/depackers/mmcmp.c:263:35: warning: array subscript 4294967295 is above array bounds of 'const uint32[8]' {aka 'const unsigned int[8]'} [-Warray-bounds=]
  263 |                 if (d >= cmd_8bits[numbits]) {
      |                          ~~~~~~~~~^~~~~~~~~
src/depackers/mmcmp.c: In function 'decrunch_mmcmp':
src/depackers/mmcmp.c:60:21: note: while referencing 'cmd_8bits'
   60 | static const uint32 cmd_8bits[8] = {
      |                     ^~~~~~~~~
In function 'block_unpack_16bit',
    inlined from 'decrunch_mmcmp' at src/depackers/mmcmp.c:435:8:
src/depackers/mmcmp.c:178:35: warning: array subscript 4294967295 is above array bounds of 'const uint32[16]' {aka 'const unsigned int[16]'} [-Warray-bounds=]
  178 |                 if (d >= cmd_16bit[numbits]) {
      |                          ~~~~~~~~~^~~~~~~~~
src/depackers/mmcmp.c: In function 'decrunch_mmcmp':
src/depackers/mmcmp.c:68:21: note: while referencing 'cmd_16bit'
   68 | static const uint32 cmd_16bit[16] = {
      |                     ^~~~~~~~~
```


The source in question is this: https://github.com/libxmp/libxmp/blob/master/src/depackers/mmcmp.c
The warning seems bogus and would result in build failures if one ever
uses `-Werror` in CFLAGS.

Biswa96:
> Since this gcc-13 upgrade, we are receiving a -Warray-bounds warning in libxmp x86 builds, but not in x64 builds

The issue should be reported to gcc, I think.

PR #16969, opened 2023-04-26
luarocks: added windows executable prefix.

Kushal-Chandar:
I need some help to run with powershell when added to path.
image


Biswa96:
luarocks and luarocks-admin are script files and not Windows executable (PE). Adding .exe would not work here. You have to run it in a shell which understand the shebang line and execute the interpreter program e.g. bash, zsh, fish etc.

PR #17023, opened 2023-04-28
flang: use upstreamed version of patch

mmuetzel:
Use the upstreamed version of the patch instead of our preliminary version.
There should be no functional difference.

See also: https://reviews.llvm.org/D149051


MehdiChinoune:
Is there something a user could benefit from this change?
I don't agree on updating packages for no reason.
You not everyone in this world has a fast internet.

mmuetzel:
> Is there something a user could benefit from this change?
> I don't agree on updating packages for no reason.
> You not everyone in this world has a fast internet.

The same could probably be said for the aws-* packages which are updated every few days with probably very minor changes that are probably affecting only a very small portion of its users.
I didn't know that there is a policy in practice that a change needs to be "significant" enough to be accepted.

MehdiChinoune:
> > Is there something a user could benefit from this change?
> > I don't agree on updating packages for no reason.
> > You not everyone in this world has a fast internet.
> 
> The same could probably be said for the aws-* packages which are updated every few days with probably very minor changes that are probably affecting only a very small portion of its users. I didn't know that there is a policy in practice that a change needs to be "significant" enough to be accepted.

Those are updates they contain some additions or fix some issues, You are comparing two different things!
What does this update fix?
It doesn't affect no one == 00 users.

lazka:
I don't have a good pictures of internet speeds around the world, probably something I should give more attention to. On the other hand everything that makes maintenance easier for MSYS2 is good for the project in the long run, also for the users.

I pushed something that hopefully works for everyone.

mmuetzel:
Our preliminary patch will probably still apply with the upstreamed version of the change. But it will no longer be necessary in the next major release (LLVM 17).
As long as we remember that, your change is fine.
The previous change would have the advantage that trying to apply the patch would result in an error for LLVM 17 when it will be released.

Maybe, we could revert to the previous change but not bump the `pkgrel` (since there shouldn't be a difference in the resulting binaries)?

MehdiChinoune:
This change could be part of the update to 16.0.3 or 16.0.4

mmuetzel:
> This change could be part of the update to 16.0.3 or 16.0.4

Has it been cherry-picked to LLVM 16.x?

MehdiChinoune:
> > This change could be part of the update to 16.0.3 or 16.0.4
> 
> Has it been cherry-picked to LLVM 16.x?

I mean wait until 16.0.3 (released hours ago) or 16.0.4 and change file to link.

mmuetzel:
I reverted back to the original change but without bumping the pkgrel.
That way, it can be part of the next minor version of this packages when that update will be done.

PR #17025, opened 2023-04-28
Add python-lmdb package

thomasfire:
Add mingw-w64-python-lmdb package, needed by https://github.com/m-labs/artiq/issues/1743


mingw-w64-python-lmdb/PKGBUILD
```diff
@@ -0,0 +1,62 @@
...
+url="https://github.com/jnwatson/py-lmdb"
+license=('OLDAP-2.8')
+depends=("${MINGW_PACKAGE_PREFIX}-python")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-python-setuptools"
+             "${MINGW_PACKAGE_PREFIX}-python-installer"
+             "${MINGW_PACKAGE_PREFIX}-python-wheel"
+             "${MINGW_PACKAGE_PREFIX}-python-build")
+source=("https://pypi.org/packages/source/${_realname::1}/${_realname}/${_realname}-${pkgver}.tar.gz"
+        "001-py-lmdb-1.4.1-mingw-build.patch" "002-py-lmdb-1.4.1-update-lmdb.patch")
```
lazka, line 20: What's the source of these patches? Will they be applied upstream?

thomasfire: First patch is actually my own build fix, since the original package assumes that everything built on windows is built by MSVC. I'm going to report it as an issue at the https://github.com/jnwatson/py-lmdb . Plus there is a segfault on opening temp files, which needs to be investigated further (I did not, since I had to test all this with virtual machine).

Second patch is reverted since it appeared to be not needed as there is already up-to-date lmdb package.

ognevnydemon:
but clangarm64 doesn't have lmdb package...

lazka:
yeah, I removed it again, thanks

PR #17027, opened 2023-04-28
Trigger rebuild of source-highlight

Geoffrey-A:
Attempt to fix #17026.

Symptom: running source-highlight or source-highlight-settings doesn’t seem to
do anything and returns exit code 127. Running with strace shows a windows
dialog with below text content:

    source-highlight.exe - Entry Point Not Found

    The procedure entry point
    _ZN5boost11regex_errorC1ENS_15regex_constants10error_typeE could not be
    located in the dynamic link library
    C:\msys64\mingw64\bin\libsource-highlight-4.dll.

Issue #17026, opened 2023-04-28
[source-highlight] 3.1.9.4 exits with code 127


Geoffrey-A:
### Description / Steps to reproduce the issue

Looks similar to #7010 and #7928, which seem to have been addressed with 0fb08d3 and 62d191c respectively. 

Symptom: running `source-highlight` or `source-highlight-settings` doesn’t seem to do anything and returns exit code 127. Running with strace shows a windows dialog with below text content:

> source-highlight.exe - Entry Point Not Found
> -
> The procedure entry point _ZN5boost11regex_errorC1ENS_15regex_constants10error_typeE could not be located in the dynamic link library C:\msys64\mingw64\bin\libsource-highlight-4.dll.

Sounds like a rebuild is in order.

### Expected behavior

Binaries to execute without error.

### Actual behavior

Nothing happens.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22621

### MINGW environments affected

- [X] MINGW64
- [X] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

sure

Issue #7928, opened 2021-02-10
[source-highlight] v. 3.1.9-2 - Installs but just quits without any output, when run.


jtm94:
mingw-w64-x86_64-source-highlight 3.1.9-2 appears to be installed correctly but immediately quits when called.  

This seems to be the same as / similar to issue #7010.

PR #17032, opened 2023-04-28
source-highlight: Rebuild for boost update

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/pull/17027
Fixes https://github.com/msys2/MINGW-packages/issues/17026
Fixes https://github.com/msys2/MINGW-packages/issues/7928


Geoffrey-A:
Thanks for merging a fix. It appears that the pacman repository has not been updated yet, but I can see your new build is visible here: https://packages.msys2.org/base/mingw-w64-source-highlight
How can we make it available in the pacman repo?

Biswa96:
Apology for the delay. The package queue has been stuck due to gcc update, see https://packages.msys2.org/queue. You can test the package from CI artifacts in 'checks' tab. Download, extract .zst file, install with `pacman -U` command.

PR #17035, opened 2023-04-28
vulkan-loader: remove custom .def file

### LINKED ISSUE #17029: Vulkan unusable on 32-bit builds

Simple example:
```
#include <stdio.h>
#include <vulkan/vulkan_core.h>

int main(void) {
    printf("%p\n", &vkCreateDisplayPlaneSurfaceKHR);
    return 0;
}
```
`cc c.c -lvulkan-1 && ./a.exe`
It links, but fails to find functions in dll.

64-bit works as expected.

~~It breaks mpv's CI currently~~
~~You can look at the history, it is recent breakage, was working day or two ago.~~

### Expected behavior

Vulkan is functions are resolved during runtime.

### Actual behavior

Vulkan is functions are not resolved during runtime.

### Verification

### Windows Version

MINGW32_NT-10.0-19045

- [ ] MINGW64
- [X] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

_No response_

### LINKED ISSUE #17042: [luajit] Crashes on 32-bit builds

mpv CI started failing few days ago.
See https://github.com/mpv-player/mpv 

Example of crashing build https://github.com/mpv-player/mpv/actions/runs/4836193035/jobs/8619203629

It crashes during loading/executing lua scripts. Wasn't a problem few days ago as we can see on CI history.

Interestingly it is also not caused by recent luajit update, because I reverted this locally to older version and still crashes.

To reproduce:
`./mpv.com --no-config -v -v`

### Expected behavior

No crashes/exceptions

### Actual behavior

Crashes/exceptions

### Verification

### Windows Version

MINGW32_NT-10.0-19045

- [ ] MINGW64
- [X] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

_No response_

lb90:
I still don't know if we should also stop defining dllexport. Tipically one uses either dllexport or .def files, but I have to investigate further on that

Fixes #17029
Fixes #17042


lb90:
We don't need `dllexport` because functions are exported with a `.def` file anyway. Even in case that the .def file doesn't include a few functions, it's best to have a compilation error than make programs which don't work on most systems, IMO.

Note that we may remove the local patch altogether, but it's best to contribute a few changes upstream first. So maybe for the next release :)

lb90:
Let's put this on hold for a moment. Apparently, the custom .def file was added to work around a mingw-w64 gcc issue in the way import libs are generated. Basically, when .def files are used, symbols in import libs have the name specified in the .def file rather than the standard decorated name. That's not the case on other compilers (clang32, msvc on x86).

Let's se if we can sole that first, otherwise we may use the trick in https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/191#issuecomment-1279657967

lb90:
Ok, tentative patch at https://sourceware.org/bugzilla/show_bug.cgi?id=30421.

lb90:
See #17163

lb90:
Now we have to rebuild most packages that depend on vulkan-loader. Should that be done here or in follow-up PRs?

Biswa96:
It'd be better if the dependent packages are rebuilt in this pull request. We can check if things are ok before merging it.

Biswa96:
Could you provide what has been changed in the shared library and import library in before/after fashion? The import library in mingw32 looks same to me.

lb90:
Yes, indeed at a glance they look quite the same! Open libvulkan-1.dll.a in a hex editor (a text editor will do) and search for all occurences of a function name, for example `vkTrimCommandPool`. 

**Matches before:**
```
_vkTrimCommandPool@16
__imp__vkTrimCommandPool@16
vkTrimCommandPool@16
_vkTrimCommandPool@16
__imp__vkTrimCommandPool@16
```

**Matches after:**
```
_vkTrimCommandPool@16
__imp__vkTrimCommandPool@16
vkTrimCommandPool
_vkTrimCommandPool@16
__imp__vkTrimCommandPool@16
```

The change is only in the third match. Basically, the symbols table has decorated symbols (before & after), but the PE imports table contains undecorated names now

Biswa96:
> the PE imports table contains undecorated names now

That undecorated name are exported only if it is mentioned in DEF, right?

lb90:
Right, to have undecorated exports one has to use .def files. Actually there's a trick to achieve the same without .def files by using `/EXPORT`, for example: `#pragma comment(linker, "/EXPORT:foo=_foo@4)`, But it's not used that much (and I believe it's MSVC-only)

* https://learn.microsoft.com/en-us/cpp/build/reference/export-exports-a-function?view=msvc-170#remarks
* https://stackoverflow.com/questions/69072717/dll-export-stdcall-without-def-file-and-decorated-function-names-with-vs

lb90:
Magnum doesn't compile on newer CLang, seems related to https://blog.magnum.graphics/backstage/forward-declaring-stl-container-types/

PR #16941, opened 2023-04-23
Add gst-plugins-rs

realmazharhussain:
Closes #16940.

There is one catch though. The default GitHub Actions VMs/Containers don't have enough storage space to build this package. It requires around 30 GiB free storage space to build (not counting the size of build environment e.g. MSYS).

PR #17037, opened 2023-04-29
Add gst-plugin-gtk4

realmazharhussain:
Adds GTK4 plugin for GStreamer.


lazka:
Can we still name it gst-plugins-rs? So we can try to fix it in the future to include more stuff.

realmazharhussain:
Sorry, there were a few things I forgot.

- [ ] ` (mingw-w64)` postfix in pkgdesc
- [ ] license in the format `spdx:<license-id>`
- [ ] install license file

realmazharhussain:
> Can we still name it gst-plugins-rs? So we can try to fix it in the future to include more stuff.

I think there has been some miscommunication here. When I asked if I should try and bundle all plugins with cargo-c in PR #16941, I meant I could most probably reopen and fix #16941.

I didn't realize it before but I think gst-plugins-rs does not currently support meson completely. It seems like they are still figuring some things out.

realmazharhussain:
#16941 is ready. If it is preferred that all plugins be bundled together, #16941 can be reviewed and merged. If the GTK4 plugin is preferred to be bundled alone, this MR can be reviewed and merged instead.

lazka:
thanks, closing since I merged #16941

PR #17043, opened 2023-04-29
mingw-w64 update to 11.0.1

jeremyd2019:
Because 11.0.1 was tagged on a branch rather than master, it seems the pkgver doesn't get to be 11.0.1.

Made this PR to try to figure out what pkgver would look like in this case


Biswa96:
I hope this will not cause any issue https://sourceforge.net/p/mingw-w64/mailman/message/37836471/ 🤞

jeremyd2019:
I don't have any need to update right now, I just wanted to make sure [my script](https://gist.github.com/jeremyd2019/3700e03dd78831cbf6067432c3eb45e9#file-update-mingw-w64-sh) handled this correctly, and I had *hoped* that the release would be more stable than some random git commit :grin:

PR #17054, opened 2023-05-03
Update fpm PKGBUILD

zoziha:
Update fpm to version 0.8.1 .


Biswa96:
Would you like to explain the requirement of `_bootstrap` option please?

mmuetzel:
Not sure if I understand your last comment correctly:
Does fpm not work at all with this update? Or is it just bootstrapping that wouldn't work with that version? (But we don't *need* to do that because there is already a working fpm in the binary repository.)

Could you please clarify?

PR #17059, opened 2023-05-04
gcc: update to 13.1.0 on x86_64 and keep 12.2.0 on i686

MehdiChinoune:


wszqkzqk:
Usually the package name should be `gcc12` instead of `gcc-12`.

See also [Arch Linux's clang14](https://archlinux.org/packages/extra/x86_64/clang14/)

PR #17060, opened 2023-05-04
libheif: update to 1.16.1

kmilos:
libwebp: provides libsharpyuv

### DIFF
--- a/mingw-w64-libwebp/PKGBUILD
+++ b/mingw-w64-libwebp/PKGBUILD
@@ -5,7 +5,7 @@ _realname=libwebp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.3.0
-pkgrel=2
+pkgrel=3
 pkgdesc="A library to encode and decode images in WebP format (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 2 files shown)


mingw-w64-libwebp/PKGBUILD
```diff
@@ -18,6 +18,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-giflib"
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja")
+provides=("${MINGW_PACKAGE_PREFIX}-libsharpyuv")
```
Biswa96, line 21: Is it required to add this new package name? There wasn't any mingw libsharpyuv package.

kmilos: It's not absolutely required. It's a relatively new component shipped currently as part of libwebp.

libheif (and [libavif](https://github.com/AOMediaCodec/libavif/commit/8c0cf99a919efeae64d19cdaff892bf4dd304880)) depends only on libsharpyuv, and not entire libwebp, so I thought it would be convenient to narrow it down. There is I guess a chance libsharpyuv gets split out from libwebp at some point in the future if there are more clients for it...

Biswa96: > a chance libsharpyuv gets split out from libwebp at some point in the future

Should that libsharpyuv package be added at that time? There is no downside adding that libsharpyuv name now, just asking.

kmilos: > Should that libsharpyuv package be added at that time? There is no downside adding that libsharpyuv name now, just asking.

The only downside is that then we'd need to go through all the libwebp dependents and figure out if they only depend on libsharpyuv or entire libwebp...

I don't really care either way.

Biswa96: OK, understood.

kmilos: Thanks. I'll change libavif shorty then as well.

PR #17061, opened 2023-05-04
libavif: enable rav1e on aarch64, depend on libsharpyuv

kmilos:


kmilos:
Ah, I guess https://github.com/msys2/MINGW-packages/commit/f32eddbfce975a856d459900c29c0780182cbe52#diff-dfecd6df9cf4746bbea31cbe8fc330d952cd6a7938f17d94360e5385bb729376 needs to get to staging first...

Please re-run when appropriate.

Biswa96:
Does the CI result look good? BTW, should gdk-pixbuf2 be added as option dependency like libjxl?

kmilos:
> Does the CI result look good?

A-ok now, thanks!

> BTW, should gdk-pixbuf2 be added as option dependency like libjxl?

I think we do indeed have a bit of a mix of makedepends/optdepends/depends for libjxl, libheif, libavif, webp-gdk-pixbuf and jp2-gdk-pixbuf. I'll create a summary first here when I find the time...

Do you have a suggestion for desired behaviour?

Biswa96:
> Do you have a suggestion for desired behaviour?

I do not have any strong opinion here. Just asking.

kmilos:
Ok, so here's the summary of packages shipping GdkPixbuf plugins

libavif: makedepends
libheif: makedepends
libjxl: mkedepends+optdepends
jp2-pixbuf-loader: depends
webp-pixbuf-loader: depends

I guess it makes sense to add optdepends to both libavif and libheif on their next updates...

PR #17062, opened 2023-05-04
cglm: Fix aarch64 intrin headers

Biswa96:
This fixes the following error

    In file included from cglm-0.9.0/src/../include/cglm/simd/intrin.h:71:
    clangarm64/lib/clang/16/include/arm64intr.h:12:15: fatal error: 'arm64intr.h' file not found
     include_next <arm64intr.h>
                  ^~~~~~~~~~~~~

@podsvirov


podsvirov:
@Biswa96, thanks for fix!

PR #17079, opened 2023-05-06
Ada updates and bring back mingw32

lazka:


mmuetzel:
Not exactly sure what happened to the old side. But gtkada release tarballs might be here now:
https://github.com/AdaCore/gtkada/releases

See also here: https://docs.adacore.com/live/wave/gtkada/html/gtkada_ug/intro.html#

lazka:
thanks, the newer releases fail to build, so I'm just going to leave gtkada as is for now.

lazka:
libadalang for mingw32 fails with "D:/M/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.1.0/../../../../i686-w64-mingw32/bin/ld.exe: error: export ordinal too large: 70731"

I'm going to drop it for mingw32.

PR #17086, opened 2023-05-08
sundials: update to 6.5.1

mmuetzel:
Removed a patch for an issues that was resolved differently upstream.
See: https://github.com/LLNL/sundials/commit/f772af3aa39e98825a0d5b3f6e98331099d1b459

Some of the libraries have version suffices. So, Octave needs to be rebuilt with the new version.


mmuetzel:
I downloaded and installed the MINGW64 build artifacts from this PR. And Octave's test suite is still passing.

PR #17093, opened 2023-05-09
qbs: update to 2.0.1

MehdiChinoune:
GCC also crashes here without any error message (the same as #17092)


lazka:
I guess I'll have to revert that stack protector thing.

Issue #17115, opened 2023-05-12
glib-compile-schemas` failed when source directory contains the non Latin symbols.


dantld:
### Description / Steps to reproduce the issue

1. Open `MINGW` shell.
2. Create any directory with non-Latin symbols.
3.  Start command glib-compile-schemas.exe /c/<ANY_NAME_WITH_NON_LATIN_SYMBOLS>.
4. Get an error, like this one:
```
(glib-compile-schemas.exe:7476): GLib-CRITICAL **: 19:05:16.746: g_dir_open_with_errno: assertion 'wpath != NULL' failed
Error opening directory в??(NULL)в??: No error
```

### Expected behavior

`glib-compile-schemas` must not failed on MS Windows platform under the MINGW64 environment when source directory, passed through the command line, contains the non Latin symbols.

### Actual behavior

`glib-compile-schemas` failed on MS Windows platform under the MINGW64 environment when source directory, passed through the command line, contains the non Latin symbols.

Example below show the problem: calling `glib-compile-schemas` with non-Latin symbols in source directory.
```sh
$ glib-compile-schemas.exe /c/Shared/vista-client-service/Юзер

(glib-compile-schemas.exe:7476): GLib-CRITICAL **: 19:05:16.746: g_dir_open_with_errno: assertion 'wpath != NULL' failed
Error opening directory в??(NULL)в??: No error
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

YES

PR #17116, opened 2023-05-12
DONE: #17115 Fix glib-compile-schemas.

dantld:
Fixed: glib-compile-schemas failed when the source directory has non Latin symbols, under MNGW64 on MS Windows platform.

Root cause: using g_option_context_parse is incorrect under the MS Windows platform due the it expects system codepage instead of UTF-8.

Resolution: using g_option_context_parse_strv when G_OS_WIN32 is defined with g_win32_get_command_line.


lazka:
https://gitlab.gnome.org/GNOME/glib/-/issues/3003

lazka:
Backported in #17352, thanks!

PR #17122, opened 2023-05-13
libplist: update to 2.3.0

MehdiChinoune:


lazka:
It had an ABI break: https://github.com/libimobiledevice/libplist/releases/tag/2.3.0

Maybe we should rebuild things just to be sure?

lazka:
https://github.com/libimobiledevice/libimobiledevice-glue/issues/31

lazka:
https://github.com/archlinux/svntogit-packages/blob/packages/libimobiledevice/trunk/libplist-2.3.0.diff

Issue #17117, opened 2023-05-12
Emacs fails to build on CLANG64


svraka:
### Description / Steps to reproduce the issue

I was checking whether the latest compilers might solve https://github.com/msys2/MINGW-packages/issues/16190. Simply run `makepkg-mingw` to build Emacs for CLANG64 (other environments are unaffected).

### Expected behavior

Emacs can be compiled on CLANG64.

### Actual behavior

The fails with compilation errors:

```
../../emacs-28.2/src/sysdep.c:472:13: error: call to undeclared function 'waitpid'; ISO C99 and later do not support
      implicit function declarations [-Wimplicit-function-declaration]
      pid = waitpid (child, status, options);
            ^
../../emacs-28.2/src/sysdep.c:518:43: error: use of undeclared identifier 'WNOHANG'
  return get_child_status (child, status, WNOHANG | options, 0);
                                          ^
2 errors generated.
make[1]: *** [Makefile:388: sysdep.o] Error 1
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22621

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Biswa96:
Thank you for reporting the issue. I can reproduce it in my system. I shall try to send the patch to upstream and backport it here.

Biswa96:
I have asked about this issue in upstream https://lists.gnu.org/archive/html/help-gnu-emacs/2023-05/msg00097.html

Biswa96:
I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request?

Issue #16190, opened 2023-03-14
mingw-w64-x86_64-emacs-28.2-3-any.pkg.tar.zst has a problem 


kotsu63:
### Description / Steps to reproduce the issue

The package does not properly run emacs after installation

### Expected behavior

Emacs should boot properly. 

### Actual behavior

It does not boot. 

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Windows 10

kotsu63:
I see. In my case, emacs does not reach the welcome display at all. Let me try to identify the cause of this problem. Some original setups I have in my windows PC may be creating the problem. When the emacs contacts ``http://tromey.com/elpa/,'' it gets suck without reaching the welcome display.

kotsu63:
mingw-w64-x86_64-emacs-28.2-2 works totally, while mingw-w64-x86_64-emacs-28.2-3 does not work. In particular, it never reaches the welcome screen. Very mysterious

Biswa96:
I used the 28.2-3 version.

```
$ pacman -Qi $MINGW_PACKAGE_PREFIX-emacs | grep Version
Version         : 28.2-3
```

kotsu63:
Yes, I understand. I will keep thinking about the causes of this problem that may be due to some original setup in my windows pc.

Biswa96:
You could try in a clean install of msys2. Or reinstall emacs and its every dependencies which can be get from `pactree -u $MINGW_PACKAGE_PREFIX-emacs` command.

kotsu63:
Thank you so much. I will try what you suggest here.

kotsu63:
I tried reinstallation of all packages from ``pactree'' as suggested. After that, I install the emacs latest version. However, the emacs status does not change and never reaches the welcome display as it gets stuck with the following screen: /home/fedhomekota/Dropbox/2023-03-14.png

Biswa96:
Why does Emacs wnt to contact that URL? How the issue can be reproduced?

kotsu63:
As of now, I may not want to install MSYS2 again from scratch due to the limitation of my available time

kotsu63:
My emacs init file contains a list of URLs for package management.

kotsu63:
I now delete them for checking some problems

kotsu63:
[...]

PR #17129, opened 2023-05-13
emacs: Fix compiling with clang due to sys/wait.h

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/17117

    Compiling with clang failed due to wait(2) function is not declared in
    emcas sys/wait.h file. The following error is shown in config.log file

    conftest.c:70:3: error: call to undeclared function 'wait'; ISO C99 and
    later do not support implicit function declarations [-Wimplicit-function-declaration]

    This adds the suggestion from https://lists.gnu.org/archive/html/help-gnu-emacs/2023-05/msg00107.html


svraka:
Thanks, now the package seems to back in the state where it builds but https://github.com/msys2/MINGW-packages/issues/16190 still is not solved.

By the way, why did you turn off jit? This has a serious performance impact on Emacs. Once https://github.com/msys2/MINGW-packages/issues/16190 is fixed, it needs to be turned on again.

PR #17130, opened 2023-05-13
hdf5: update to 1.14.1

mmuetzel:
Two of the patches are no longer needed because the respective issues have been fixed upstream.

I'm not exactly sure what's the deal with the `-2` at the end of the name of the tarball.


lazka:
> I'm not exactly sure what's the deal with the `-2` at the end of the name of the tarball.

https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.14/hdf5-1.14.1/src/hdf5-1.14.1-2-RELEASE.txt

> HDF5 1.14.1-2 is a patch release for HDF5 1.14.1. The only change in the patch
release is that Autoconf 2.71 was used to generate the Autotools build files,
which allows building with Intel's oneAPI.

Issue #17038, opened 2023-04-29
Consider bundling GStreamer plugins differently


realmazharhussain:
## The problem

Currently, GStreamer plugins of different categories are bundled together. So, if someone needs all the codecs to be installed they have to install all the plugin bundles `gst-plguins-{base,good,bad,ugly}` which contain the codecs support but also other stuff (significantly more than what that person needs) e.g. GTK3, OpenCV, networking plugins, etc.

For example, I am working on a media player based on LibAdwaita/GTK4 and I need h264, x265, opus, srt (and so on) plugins from the `gst-plugins-bad` package. But when I install `gst-plugins-bad` package, it also installs GTK3 and OpenCV, both of which I don't want.

## Suggested Solution
GStreamer has [merge modules which group plugins by function](https://gstreamer.freedesktop.org/documentation/deploying/windows.html#deploy-only-necessary-packages-using-provided-merge-modules), e.g. codecs, network protocols, playback, etc.

MSYS could try a similar approach. We could
- split each plugin bundle into a separate package for each plugin
- group similar plugins with either meta packages or package groups

For backwards compatibility with any CI/CD scripts/recipes which rely on `gst-plugins-{base,good,bad,ugly}`, we could keep `gst-plugins-{base,good,bad,ugly}` as meta packages.

Biswa96:
Awesome idea. ArchLinux also splits the gtk and opencv plugins in different packages https://archlinux.org/packages/?sort=&q=gst-

Please make sure the package name is same as ArchLinux. Also update the dependencies in the related mingw packages.

realmazharhussain:
> ArchLinux also splits the gtk and opencv plugins in different packages

Just doing this is fine with me but I was hoping we could do more than just that.

If we do everything I suggested, package names won't be the same as Arch but we also won't need to update the dependencies in the related mingw packages.

Biswa96:
I have added a pull request to split the gtk and opencv plugins like ArchLinux does. Would you like to check the packages from GitHub Actions artifacts of that pull request?

realmazharhussain:
I have downloaded the relevant artifacts but I can test for my use case and system only. I'm using UCRT64 MSYSTEM and my app does not need OpenCV and GTK3 plugins.

In other words, I'll only be testing the following packages from the build artifacts

```
mingw-w64-ucrt-x86_64-gst-plugins-bad-1.22.2-2-any.pkg.tar.zst
[...]

PR #17155, opened 2023-05-16
Split gtk and opencv gstreamer plugins into separate packages

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/17038

### DIFF
--- a/mingw-w64-gst-plugins-bad/PKGBUILD
+++ b/mingw-w64-gst-plugins-bad/PKGBUILD
@@ -4,9 +4,11 @@
 
 _realname=gst-plugins-bad
 pkgbase=mingw-w64-${_realname}
-pkgname=("${MINGW_PACKAGE_PREFIX}-gst-plugins-bad-libs" "${MINGW_PACKAGE_PREFIX}-gst-plugins-bad")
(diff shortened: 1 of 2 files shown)


realmazharhussain:
So, I tested UCRT64 packages from the Actions artifacts and found the following issues.

* `gst-plugins-bad` package still depends on `gtk3` package.

And even if it didn't depend on `gtk3`, I still cannot remove GTK3 or OpenCV because of the following dependencies.

* `appstream-glib` -> `gtk3`
* `gst-libav` -> `ffmpeg` -> `frei0r-plugins` -> `opencv`

On Arch Linux, `ffmpeg` does not depend on `frei0r-plugins`, not even optionally. But weirdly enough, even on Arch, `appstream-glib` does depend on `gtk3`. Although, they also ship the official alternative, [`appstream`](https://github.com/ximion/appstream), as well.

Biswa96:
> gst-plugins-bad package still depends on gtk3 package.

I have removed the gtk3 from gst-plugins-bad.

> On Arch Linux, ffmpeg does not depend on frei0r-plugins, not even optionally. 

I shall look into it.

PR #17161, opened 2023-05-16
uvicorn: New package

podsvirov:
Contains #17143


podsvirov:
@lazka, thanks for review and merge!

PR #17163, opened 2023-05-17
binutils: backport patch for x86 import libs

lb90:
Fixes an issue in ld where import libraries are generated with undecorated symbols. See https://sourceware.org/bugzilla/show_bug.cgi?id=30421


lazka:
Awesome :)

Can I use https://github.com/lb90/my-shared-lib/tree/main under https://spdx.org/licenses/ISC.html and add it to our regression test suite?

lazka:
In the repo now

PR #17166, opened 2023-05-18
gst-plugins-rs: Fix some plugins not included

realmazharhussain:
In my previous PR, I did not notice that some plugins (e.g. the GTK4 plugin) were being skipped. This PR is supposed to fix that issue.


Biswa96:
The package contains .def files. Should those be removed?

realmazharhussain:
Probably.

I should create a *new* PR for that, right?

Biswa96:
You can add the commit in this PR as well. Also bump the pkgrel value please.

PR #17175, opened 2023-05-19
helix: update to 23.05

MehdiChinoune:
@mati865 any idea how to pass CFLAGS to cargo?


mati865:
FYI Cargo doesn't care about CFLAGS, it's `cc-rs` that is used in this case to invoke the compiler.
Since in this case the build fails in C++ project you have to use `CXXFLAGS+=" -Wno-implicit-int"`.

PR #17176, opened 2023-05-19
imagemagick: update to 7.1.1.9

kmilos:


Biswa96:
I wonder, should pthread be disabled for imagemagick? In that distribute-cache.c file, the following logic is used.

```c
#if defined(MAGICKCORE_THREAD_SUPPORT)
    // use pthread_create()
#elif defined(MAGICKCORE_WINDOWS_SUPPORT)
    // use CreateThread()
#else
    Not implemented!
#endif
```

So, if we add `--without-threads` option with configure command that should disable the pthread_create call and use the native Windows thread APIs. Is that right?

kmilos:
No idea, there likely might be other stuff going on w/ `MAGICKCORE_WINDOWS_SUPPORT`, no time to look into it now, sorry...

PR #17185, opened 2023-05-20
update some qt5 packages

MehdiChinoune:


mingw-w64-qt5-connectivity/PKGBUILD
```diff
@@ -6,8 +6,8 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-debug")
 _qtver=5.15.9
 pkgver=${_qtver}+kde+r5
```
Biswa96, line 8: Shouldn't pkgver be changed here also?

MehdiChinoune: No, last time I forgot to update the commit hash, the package was in reality at 5.15.8.
You could check file-diff in CI log.

PR #17186, opened 2023-05-20
qt5-static: enable vulkan

MehdiChinoune:
Reduce buildtime by using release build mode instead of release-and-debug which builds release and debug libraries separately hence the long buildtime. We did the same with qt6-static.


MehdiChinoune:
```diff
-Download Size   : 944.15 MiB
-Installed Size  : 5868.89 MiB
+Compressed Size : 154.20 MiB
+Installed Size  : 975.66 MiB
```

MehdiChinoune:
with this change, autobuild could be enabled for qt5-static

Issue #11676, opened 2022-05-16
mingw-w64-grep: No results past 100'000 files


cyrilarnould:
If there are too many files, `grep -r` will at some point simply stop reporting, seemingly if there's more than 100'000 files. I've tested this both in the "MSYS2 MinGW x64" environment as well as Windows cmd. Running Windows10 21H2 build 19044.1706.

I've created a test folder with 200'000 text files: [filetest.tar.gz](https://github.com/msys2/MINGW-packages/files/8703717/filetest.tar.gz) Every file simply contains one line with "hello" and the file number.

Executing `grep -r hello` will report up to file 100'000 and then stop. Executing the same command in the "MSYS2 MSYS" environment instead (i.e. with `/usr/bin/grep`), I get a result from each file.

stevenclarke9:
can you use find and use the exec option to grep each filename for "hello". I use find command first to filter the filetypes to text as grep will also look in the binary files as well.

cyrilarnould:
Regardless of possibilities to work around it, I think this issue should probably be treated as a bug since the mingw-w64-grep behaves differently from the MSYS grep.

Ideally, I'd like to also use `grep` in the Windows command prompt, but `findutils` is not available as a native package and I'd rather abstain from including the `/usr/bin` in my Windows PATH. Also, `grep -Ir` can be used to ignore binary files.

For now, my workaround is to simply use the `/usr/bin/grep` from the MSYS environment instead; it's slower than mingw-w64-grep but on large folders the operation takes quite some time anyway.

cyrilarnould:
I could file the report, but I think it might make more sense for a maintainer or a contributor of the PKGBUILD to do so instead... I would be of little help in resolving the bug, I'm rather inexperienced in this environment and I rarely program in C. I can't even get the PKGBUILD to work.

PR #17190, opened 2023-05-20
grep: update to 3.11

### LINKED ISSUE #11675: mingw-w64-grep: Too many open files
If there are too many subfolders, `grep -r` will at some point fail with the message "too many open files". I've tested this both in the "MSYS2 MinGW x64" environment as well as Windows `cmd`. Running Windows10 21H2 build 19044.1706.

I've created a test folder with 3000 subfolders containing 100 text files each: [foldertest.tar.gz](https://github.com/msys2/MINGW-packages/files/8703492/foldertest.tar.gz) Every file simply contains one line with "hello" and the file number.

`grep -r` fails in the first file of folder 2041. It prints `grep.exe: folder002041/file001.txt: Too many open files` for each file in that folder, then again for each file in `folder002042`. All subsequent folders only get a single error line: `grep.exe: folder002043: Too many open files`

cyrilarnould:
- The patch is not necessary anymore for the build to succeed.
- `gl_cv_have_weak` doesn't need to be set anymore, `HAVE_WEAK_SYMBOLS` is properly set in `config.h`

This update also seems to fix #11675 and #11676. I haven't verified if these issues have been addressed explicitly by upstream though.


lazka:
thanks!

PR #17193, opened 2023-05-20
autoupdate pypi packages

MehdiChinoune:
Needs a new account/(an msys2 bot) with its Personal Access Token stored as a GHA secret. Unless someone among maintainers volunteer and uses his own instead.
GHA could not be triggered by commits and PRs from github-actions bot.


.github/workflows/autoupdate.yml
```diff
@@ -0,0 +1,35 @@
...
+          pacboy: jq:p python-nvchecker:p python-packaging:p
+          update: true
+          release: false
+
+      - name: update
+        shell: msys2 {0}
+        run: |
+          ./update-pypi.sh
+          git config --global core.autocrlf input
+          if [[ `git status --porcelain -- oldver.json` ]]; then
```
lazka, line 31: oldver-pypi?

MehdiChinoune: yes!

lazka:
In theory a "Fine-grained personal access tokens" should be enough. and using

```
git config --global user.email "github-actions[bot]@users.noreply.github.com"
git config --global user.name "github-actions[bot]"
```

makes github mark it as a bot commit.

----

I guess I can try this offline as well?

Some thoughts for potential future changes:

* Does it need to be in this repo? Something opening PRs via a fork would be good enough I think.
* check-pypi.toml could be auto-generated based on information from PKGMETA.yaml (and add pypi names there)

lazka:
and just wondering, what makes pypi a better source then https://packages.msys2.org/outofdate ? Assuming there would be an API etc...

MehdiChinoune:
> In theory a "Fine-grained personal access tokens" should be enough. and using
> 
> ```
> git config --global user.email "github-actions[bot]@users.noreply.github.com"
> git config --global user.name "github-actions[bot]"
> ```
> 
> makes github mark it as a bot commit.

That's the problem GHA won't run if the commits and PRs are from GHA bot.
> 
> I guess I can try this offline as well first?
> 
> Some thoughts for potential future changes:
> 
> * Does it need to be in this repo? Something opening PRs via a fork would be good enough I think.

No, It could be in a fork.

> * check-pypi.toml could be auto-generated based on information from PKGMETA.yaml (and add pypi names there)

Maybe, I think PyPi names are case sensitive, so they should be extracted correctly.

MehdiChinoune:
> and just wondering, what makes pypi a better source then https://packages.msys2.org/outofdate ? Assuming there would be an API etc...

I haven't understand well, but PyPi shows the latest version but msys2-outdated shows the latest version from arch, cygwin or aur.

lazka:
> I haven't understand well, but PyPi shows the latest version but msys2-outdated shows the latest version from arch, cygwin or aur.

I see. The upside is that arch/cygwin have already done some basic testing etc. before we get the "update". So it has downsides and upsides :)

I'll give this a try later, thanks!

lazka:
somewhat related, I copied the package mapping from check-pypi.toml to [PKGMETA.yml](https://github.com/msys2/MINGW-packages/commit/5d9c0d572abe9369842da6a980651434871065da#diff-46226baf7834b68989c5076b41066a8227f9bdc9c1c2c9312a9e9fbe905c6202) and show a link on packages.msys2.org, like https://packages.msys2.org/base/mingw-w64-python-jaraco.functools

lazka:
The pypi date is now used for comparing version numbers with our packages.

Here is a very basic API for fetching all outdated packages: https://packages.msys2.org/api/outofdate

PR #17195, opened 2023-05-20
pgmodeler: Update to 1.0.4

podsvirov:


podsvirov:
Currently done in #17298 and close with reason like in #17282.

PR #17214, opened 2023-05-21
mesa: depend on llvm-libs instead of clang

lazka:
in theory this should be enough


lazka:
It's not enough, it depends on libclang-cpp

PR #17263, opened 2023-05-23
openxr-sdk: remove custom .def file

lb90:
Same as #17035 for openxr-sdk.

Blender is excluded because we don't provide `mingw32` builds


lb90:
BTW, looks like openxr-sdk doesn't really depend on `vulkan-loader`, it should suffice to make-depend on `vulkan-headers`. What do you think?

lb90:
CLANG build failure seems related to a backward-incompatible change in ADIOS2: https://github.com/ornladios/ADIOS2/issues/2127. Don't know about GCC

PR #17323, opened 2023-05-25
gpgme: update to 1.18.0

Biswa96:


Biswa96:
gpgme v1.19 and v1.20 have changed a lot and need some patches for python 3.10. So, these versions are kept to compare if there is any issue with latest versions.

PR #17338, opened 2023-05-27
[New Package]: libilbm

cmyk-student:
This follows up #17329 to finish adding support for GIMP's new [AMIGA IFF image plug-in](https://gitlab.gnome.org/GNOME/gimp/-/merge_requests/897). 

This patch proposes adding the libilbm package to the MSYS2 repository, which is required to build the GIMP IFF plug-in on Windows now that libiff has been accepted.
As the libiff and libilbm repositories are structured the same, the PKGBUILD file is nearly identical to what @Biswa96 assisted me with for libiff (thanks again!)

I tested and was able to build the package with this locally - hopefully there are no new issues!

### DIFF
--- a/mingw-w64-libilbm/PKGBUILD
+++ b/mingw-w64-libilbm/PKGBUILD
@@ -3,14 +3,13 @@
 _realname=libilbm
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.1
-_realver=0.1
+pkgver=0.1.r68.g170c269
 pkgrel=1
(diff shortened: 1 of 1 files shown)


Biswa96:
Both libiff and libilbm are compiled as static libraries. Is that appropriate for GIMP?

cmyk-student:
> Both libiff and libilbm are compiled as static libraries. Is that appropriate for GIMP?

Hi again! I believe so - they will only be used in that plug-in, and it follows the pattern of another library used in the GIMP plug-in for ASCII art, [aalib](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aalib/PKGBUILD#L42).

PR #17342, opened 2023-05-27
emacs: Remove _FORTIFY_SOURCE from build options

### LINKED ISSUE #16413: [emacs] Resizing shell in graphical emacs let's freeze emacs and even Windows is laggy

cyrilarnould:
Redo of #17319. `_FORTIFY_SOURCE` breaks the emacs builds. 

Resolves #16413 #16375 and #16190.

### DIFF
--- a/mingw-w64-emacs/PKGBUILD
+++ b/mingw-w64-emacs/PKGBUILD
@@ -6,7 +6,7 @@ _realname=emacs
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=28.2
-pkgrel=4
+pkgrel=5
 pkgdesc="The extensible, customizable, self-documenting, real-time display editor (mingw-w64)"
 url="https://www.gnu.org/software/${_realname}/"
 license=('spdx:GPL-3.0')
@@ -69,6 +69,8 @@ build() {
 
   # Required for nanosleep with clang
   export LDFLAGS="${LDFLAGS} -lpthread"
+  # -D_FORTIFY_SOURCE breaks build
+  CFLAGS=${CFLAGS//"-Wp,-D_FORTIFY_SOURCE=2"}
 
   ../${_realname}-${pkgver}/configure \
     --prefix="${MINGW_PREFIX}" \


mingw-w64-emacs/PKGBUILD
```diff
@@ -69,6 +69,8 @@ build() {
 
   # Required for nanosleep with clang
   export LDFLAGS="${LDFLAGS} -lpthread"
+  # -D_FORTIFY_SOURCE breaks build
+  CFLAGS=${CFLAGS//"-Wp,-D_FORTIFY_SOURCE=2"}
```
lazka, line 73:
```suggestion
  CFLAGS+=" -U_FORTIFY_SOURCE"
```

would that work too?

cyrilarnould:
Can't test it today, but it does seem like the better solution. Thanks!

cyrilarnould:
For the record, I've now tested adding `-U_FORTIFY_SOURCE` rather than removing `-Wp,-D_FORTIFY_SOURCE=2` and the problems were back.

lazka:
ok, thanks for testing

PR #17346, opened 2023-05-28
gpgme: update to 1.20.0

Biswa96:
* Remove hunks which replace gpgme-config with shell, fixed in https://dev.gnupg.org/rMae9258fbf3b9d434495ef11fc184a91fe7c4ca57
* Check exit code for python disutils due to deprecation message.
* Disable python test due to this [Makefile:645: pubring-stamp] Error 2


Biswa96:
@naveen521kk This seems an exception in the 'check' logic. It fails but the python-gpgme package does not have any executable.

PR #17347, opened 2023-05-28
aarch64 rust fixes

jeremyd2019:
From discussion on #16945.  Not re-enabled is mingw-w64-mdbook-pikchr, due to podsvirov/mdbook-pikchr#1 (which happens on x64 as well as aarch64)


jeremyd2019:
Successful ARM64 CI run: https://github.com/msys2-arm/MINGW-packages/actions/runs/5102422505/jobs/9171974644

PR #17349, opened 2023-05-28
mingw-w64-gdb update to 13.2

GitMensch:
update by version and sha adjustment (possibly more to come if CI doesn't work)


GitMensch:
Thanks for merging. Could you recheck if it now builds also on the previously disabled environment?

Issue #17330, opened 2023-05-26
MKVToolNix does not open without ffmpeg installed


fauxpark:
### Description / Steps to reproduce the issue

The MKVToolNix GUI stopped opening for me recently. After poking around in Process Monitor I discovered it went on a frantic search for some ffmpeg DLLs. `mkvtoolnix-gui` itself does not reference these DLLs, but it does depend on `qt6-multimedia`, which in turn provides an ffmpeg plugin that is apparently now required.

I see there have been some updates to `qt6-multimedia` in the last month or so, but I'm not sure whether ffmpeg should be promoted from an optional dependency there, or if it should be added to `mkvtoolnix-gui`.

### Expected behavior

MKVToolNix should open.

### Actual behavior

MKVToolNix does not open and exits 127.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

fauxpark:
I found this note: https://doc.qt.io/qt-6/qtmultimedia-index.html#target-platform-and-backend-notes
> In Qt 6.5, we're setting [FFmpeg framework](http://ffmpeg.org/) as the default backend on Windows, macOS, Android, and Linux except Yocto distribution. Native backends, namely gstreamer on Linux, AVFoundation on macOS, WMF on Windows, and the MediaCodec framework on Android, were the default in 6.4. These are still available but with limited support.

Which sounds to me like ffmpeg can no longer be an optional dependency.

Biswa96:
Is it possible to split the ffmpeg and gstreamer plugins in separate packages like [ArchLinux does](https://gitlab.archlinux.org/archlinux/packaging/packages/qt6-multimedia/-/blob/main/PKGBUILD)?

fauxpark:
It would be WMF, not gstreamer, but yes.

Biswa96:
Is this fixed after that qt6-multimedia backend split?

fauxpark:
Still waiting on upload for v78, but 73.0.0 does seem to work.

PR #17351, opened 2023-05-29
qt6-multimedia: change ffmpeg to required dependency

fauxpark:
#17330


lazka:
I haven't looked into it, but I confirmed that it starts without ffmpeg if another backend is selected via the env var.. so this looks good to me.

PR #17364, opened 2023-05-29
qt6-multimedia: split out backend packages

fauxpark:
Per https://github.com/msys2/MINGW-packages/issues/17330#issuecomment-1566413954

### DIFF
--- a/mingw-w64-qt6-multimedia/PKGBUILD
+++ b/mingw-w64-qt6-multimedia/PKGBUILD
@@ -56,10 +56,10 @@ build() {
 package_qt6-multimedia() {
   depends=("${MINGW_PACKAGE_PREFIX}-qt6-base"
            "${MINGW_PACKAGE_PREFIX}-qt6-shadertools"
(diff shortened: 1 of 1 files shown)


lazka:
I'm not sure if it's defined which of those two packages gets installed then, and we might end up in the old broken case without ffmpeg.

fauxpark:
pacman on Arch appears to prompt the user when there are multiple candidates available for a dependency:
image

It is only single choice but users can still manually install other providers after the fact. A package can also depend on a specific provider (but I don't think this is necessary for MKVToolNix at least).

As I understand it, the "old broken case" is only due to the fact both backends are currently installed but `ffmpeg` was not. #17351 fixed this, but since the FFmpeg backend is what actually references `ffmpeg` the dependency has been moved accordingly. The WMF backend has no other dependencies as the DLLs it wants are obviously part of Windows.

lazka:
Ahhh, ok, thanks

fauxpark:
Sounds much better, updated. Wasn't sure how much you want to diverge from how Arch does things.

PR #17365, opened 2023-05-29
mdbook-pikchr: Update to 0.1.6

podsvirov:


podsvirov:
Cc: @jeremyd2019

jeremyd2019:
You *should* be able to enable it for clangarm64 now too.  It was disabled because mdbook was failing to build, and then couldn't be re-enabled due to podsvirov/mdbook-pikchr#1.  Now it looks like windows-targets 0.48.0 is being pulled in, so it probably requires the nasty hack from #17347 to be applied here too.

jeremyd2019:
Unexpectedly, it builds as-is on clangarm64, even though it builds windows-targets 0.48.0 (without the corresponding windows_aarch64_gnullvm 0.48.0).

podsvirov:
Thanks for checking and editing.  By the way, `mdbook` can also be updated #17363.

PR #17367, opened 2023-05-30
luajit: disable unwind on 32-bit builds

kasper93:
This fixes crashes on 32-bit build. MSYS explicitly disables SEH for 32-bit builds which breaks the unwinding code in luajit.

Fixes: #17042

### DIFF
--- a/mingw-w64-luajit/PKGBUILD
+++ b/mingw-w64-luajit/PKGBUILD
@@ -7,7 +7,7 @@ pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 # https://github.com/LuaJIT/LuaJIT/issues/665#issuecomment-784452583
 _commit="51fb2f2c3af778f03258fccee9092401ee4a0215"
 pkgver=2.1.0.beta3.r481.g51fb2f2c
-pkgrel=1
+pkgrel=2
 pkgdesc="Just-in-time compiler and drop-in replacement for Lua 5.1 (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


mingw-w64-luajit/PKGBUILD
```diff
@@ -46,11 +46,16 @@ prepare() {
 build() {
   cd "${_realname}"
 
+  XCFLAGS=-DLUAJIT_ENABLE_GC64
+  if [[ "$MSYSTEM" == "MINGW32" || "$MSYSTEM" == "CLANG32" ]]; then
```
Biswa96, line 50: I think checking `${CARCH} == "i686"` would be more appropriate, wouldn't it?

kasper93: I copied conditions form https://github.com/msys2/MSYS2-packages/blob/master/pacman/makepkg_mingw.conf for systems where the `no-seh` flag is added.

Indeed those are only `i686` build, will change it later.

lazka: I don't mind checks being more explicit.

jeremyd2019:
I assume this was talking about the `-Wl,--no-seh` LDFLAG on 32-bit builds.  I included that when we were adding hardening flags, since GCC doesn't use SEH on 32-bit.  You should also be able to override/filter this flag out on a package that does in fact use SEH (luajit?).  The no-seh flag means there are not allowed to be any SEH handlers in the image.  So you can enable it on a per-image basis, not per-process or anything (static linking is a different issue).

kasper93:
Since every (most?) package is compiled with `-Wl,--no-seh` I wanted to keep this behavior for both shared and static library. So they are both the same. Why we override linker flags is other topic.

PR #17369, opened 2023-05-30
pango: add patches

lb90:


lb90:
@lazka this reenables DWrite fonts, but maybe we should wait for cairo 1.18?

lazka:
> Just curious. Is there any alternative symbols or macros for the integers in https://gitlab.gnome.org/GNOME/pango/-/merge_requests/689 ?

`UINT32` is a windows API typedef, `guint16` is a glib typedef, both for historical reasons from pre-c99 times. If that's what you mean.

Biswa96:
I meant instead of doing,

```c
UINT32 max_ranges = (table_size - 4) / (sizeof (guint16) * 2);
```

would it be possible to write something like,

```c
UINT32 max_ranges = (table_size - SOMETHING_HAS_FOUR_VALUE) / (sizeof (guint16) * SOMETHING_HAS_TWO_VALUE);
```

to make the code more readable.

lazka:
ah, ok, yeah, I agree

lb90:
> Does this change the drawing part of pango?

Yes: Pango uses DirectWrite for font enumeration / discovery, but Cairo uses DirectWrite rendering only if the `cairo_font_face_t` is created specifically for DirectWrite (i.e created with `cairo_dwrite_font_face_create_for_dwrite_fontface ()`)

lazka:
The only difference I can see is it no longer using color subpixel rendering, but apparently that's a downside of DirectWrite: https://en.wikipedia.org/wiki/ClearType#ClearType_in_DirectWrite

Feel free to merge.

lb90:
Dropped the patch that enabled DWrite rendering; that really should be enabled after updating Cairo, as important fixes have landed in trunk. For now, better stay with GDI-based rendering

I have talked to Emmanuele Bassi and he said that Cairo 1.18.0 is planned for relase at around October for GNOME 45, and that no more 1.17.x snapshots will be made

Let me know if it's good to go 🙂

lazka:
lgtm

jeroen:
Are these patches still needed for cairo-1.18.0 with pango 1.51?

lazka:
Since they are all merged upstream the'll get removed anyway with the next stable release.

jeroen:
Ah does pango use the even-odd unstable versioning?

lazka:
In theory, yes.. though I remember distros picking up odd versions from time to time due to missing releases.

lb90:
BTW, now that we have Cairo 1.18 we can enable DWrite fonts. I'm going to open a PR next week!

jeroen:
Hmm something else changed in cairo, that requires to use `-DCAIRO_WIN32_STATIC_BUILD" when building static libraries.

lazka:
Yeah, looks like it uses dllexport since like everything else https://gitlab.freedesktop.org/cairo/cairo/-/commit/ba4d5fbd5dad46c84e38ffbb032799a7a9380f42

jeroen:
They should probably add `-DCAIRO_WIN32_STATIC_BUILD` to the `private.cflags` in the `.pc` file...

lazka:
https://github.com/msys2/MINGW-packages/pull/18660

PR #17373, opened 2023-05-30
Support of 32-bit compiler for large files and a few other improvements

zanud:
1. Made functions for operation with large files available to 32-bit compiler
2. 32-bit example64.exe now passes testing
3. Removed zlib1.dll dependency on non-system DLLs
4. Fulfilled the zlib developers' requirements of to export of zlib1.dll
5. Added VERSIONINFO resource to zlib1.dll


mmuetzel:
You'll need to update the checksums of those patches in the `PKGBUILD` file. E.g., run `updpkgsums` in the folder with the `PKGBUILD` file to do that.
Please, also bump the `pkgrel` in the `PKGBUILD` file for this change.

mmuetzel:
The CI fails when trying to link with a lot of undefined symbol errors.
Are you sure you can remove `-Wl,--export-all-symbols` from the linker flags?

zanud:
`-Wl,--export-all-symbols` contradicts to the requirement of the zlib developers: only functions from zlib.h (win32\\zlib.def) must be exported by zlib1.dll. And they introduced ZLIB_DLL flag exactly for this purpose.

On my system building of this DLL finishes flawlessly. And the only difference between my build environment and github's one is: which directory is current during compilation. On my system it is the directory with source files, and on github it is some other one. When I tried to do the same on my system, I recreated this problem.

And now I am trying to understand why this problem happens at all. The `configure` script creates `zconf.h` header file in the current directory. When this is the source files directory, `zconf.h` is included via `#include "zconf.h"`, and compiler is happy. When this is any other directory, `zconf.h` is included via `-include zconf.h` command line parameter, and compiler complains on contradictions between functions definitions in `*.h` and `*.c` files. In both cases contents of `zconf.h` is the same (exactly), the difference is only in method of including.

It is quite easy to change the `configure` script and unify command lines and directory for `zconf.h`, but I prefer not to do this without necessity and find other solution.

mmuetzel:
If you really want to sort out the `dllexport` issue instead of just exporting all symbols, here some hints.
See warnings like the following during compilation:
```
  gcc -march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -DZLIB_DLL -D_LARGEFILE64_SOURCE=1 -include zconf.h -DPIC -c -o objs/crc32.o ../zlib-1.2.13/crc32.c
  ../zlib-1.2.13/adler32.c:63:15: warning: 'adler32_z' redeclared without dllimport attribute: previous dllimport ignored [-Wattributes]
     63 | uLong ZEXPORT adler32_z(adler, buf, len)
        |               ^~~~~~~~~
```

Together with this snippet from `zconf.h`:
```
#  ifdef ZLIB_DLL
#    if defined(WIN32) && (!defined(__BORLANDC__) || (__BORLANDC__ >= 0x500))
#      ifdef ZLIB_INTERNAL
#        define ZEXTERN extern __declspec(dllexport)
#      else
#        define ZEXTERN extern __declspec(dllimport)
#      endif
#    endif
#  endif  /* ZLIB_DLL */
```

During compilation, `ZLIB_INTERNAL` needs to be defined. But I can't see a `-DZLIB_INTERNAL` in the compilation commands. There is only `-DZLIB_DLL`.

Looking into why it is missing, might give a clue at how to fix that.

zanud:
So why the problem does not exist when I run `configure` and `make` in the sources directory? `zconf.h` is the same, and the only differences is its location and compiler command line:
`gcc -O3 -DZLIB_DLL -D_LARGEFILE64_SOURCE=1 -DPIC -c -o objs/adler32.o adler32.c` - when I run `make` from `zlib_src` directory (`zconf.h` is here)
vs
`gcc -O3 -DZLIB_DLL -D_LARGEFILE64_SOURCE=1 -include zconf.h -DPIC -c -o objs/adler32.o ../adler32.c` - when I run `make` from `zlib_src/build` directory (`zconf.h` is here).

mmuetzel:
Many `.c` files contain the following snippet:
```
#define ZLIB_INTERNAL
#include "zlib.h"
```

`zlib.h` includes `zconf.h`. So, if the file is included via the code in the files, `ZLIB_INTERNAL` is defined before the `zconf.h` is included.
If however, `zconf.h` is force-included via a command line switch, it is "as if "#include "file"" appeared as the first line of the primary source file" [1]. So, `ZLIB_INTERNAL` isn't defined before the header is included.

Inserting `-DZLIB_INTERNAL` into the command line (just before) `-include zconf.h` might fix the issue.

[1]: https://linux.die.net/man/1/gcc

zanud:
_If, however, zconf.h is force-included via a command line switch, it is "as if "#include "file"" appeared as the first line of the primary source file" [1](https://linux.die.net/man/1/gcc). So, ZLIB_INTERNAL isn't defined before the header is included._

Thanks. That's it.
(I mainly played with a first reported file (`adler32.c`), and it does not contain `#define ZLIB_INTERNAL` because it is included indirectly - via `#include "zutil.h"`)

_Alternatively (and I've seen that used more often), inserting -I. into the command line instead of -include zconf.h might also do the trick._

This part was written not by me, but by previous developers, and I always prefer to change existent code as little as possible (to avoid breaking something). But if you say that this is widely used, I'll try to use it too. (At least on my system compiling with `gcc -O3 -DZLIB_DLL -D_LARGEFILE64_SOURCE=1 -I. -DPIC ...` produces the same DLL as during my initial testing before PR creation.)

mmuetzel:
According to the package grokker results, `zlib1.dll` no longer exports the following symbols after the changes from here:
```
  {b'zlib1.dll': {b'_dist_code',
                  b'_length_code',
                  b'_tr_align',
                  b'_tr_flush_bits',
                  b'_tr_flush_block',
                  b'_tr_init',
                  b'_tr_stored_block',
                  b'_tr_tally',
                  b'deflate_copyright',
                  b'gz_error',
                  b'inflate_copyright',
                  b'inflate_fast',
                  b'inflate_table',
                  b'z_errmsg',
                  b'zcalloc',
                  b'zcfree'}}
```

Are all of those meant to be internal?

zanud:
Yes. According to `win32\DLL_FAQ.txt`:
> The exported symbols are exclusively defined in the source files "zlib.h" and "zlib.def", found in an official zlib source distribution.

these functions must not be exported by `zlib1.dll`, as no one of them is present in the `win32\zlib.def`.

mingw-w64-zlib/01-zlib-1.2.11-1-buildsys.mingw.patch
```diff
@@ -17,43 +26,48 @@
...
 +        CFLAGS="${CFLAGS} -D_LARGEFILE_SOURCE=1 -D_FILE_OFFSET_BITS=64"
-+        SFLAGS="${CFLAGS}"
++        SFLAGS="${CFLAGS} -DZLIB_DLL"
++        RES_OBJ=' zlib1rc.lo '
 +        LDFLAGS="-L. ${old_ldflags}"
 +        shared_ext=".dll"
 +        SHAREDLIB='cygz.dll'
 +        IMPORTLIB='libz.dll.a'
-+        LDSHARED=${LDSHARED-"$cc -shared -Wl,--export-all-symbols -Wl,--enable-auto-image-base -Wl,--out-implib=${IMPORTLIB}"}
++        LDSHARED=${LDSHARED-"$cc -shared -static-libgcc -Wl,--enable-auto-image-base -Wl,--out-implib=${IMPORTLIB}"}
```
mmuetzel, line 51: Is `-static-libgcc` really necessary? I'm not sure if it is a good idea to link libgcc statically in MinGW.
See also this comment in the build rule of gcc:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gcc/PKGBUILD#L227
```
  # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105507#c3
  # At least with mingw32 + dwarf-2 exceptions there can only be one libgcc in
  # process, or exceptions will no longer work. Since some of the gcc deps are
  # linked dynamically like gmp/zlib/zstd and those pull in libgcc we can't
  # allow libgcc to be linked statically. The default is "-static-libstdc++
  # -static-libgcc" for both, so we drop "-static-libgcc" here:
```
(But that comment might be about something different... 🤷‍♂️)

mmuetzel:
Oops. I tried to push an additional change. But that seems to have messed up your branch.
Trying to clean that up now...

mmuetzel:
Not exactly sure what went wrong. But to recover from that, I applied your changes to a fresh checkout in the form of patches and force pushed to your fork. I hope that is ok.

I then pushed another change that reverts the part to statically link to libgcc.
Please, let me know if you think that is not ok for some reason.

If no one opposes, I'll try to merge this change early next week.

mmuetzel:
The package-grokker failed with an error that I don't understand.

@jeremyd2019: Could you please take a look when you come around to it?

jeremyd2019:
> The package-grokker failed with an error that I don't understand.
> 
> @jeremyd2019: Could you please take a look when you come around to it?

Looks like it was just an http error downloading packages.

Or do you mean all the noisy warnings from pefile? I started printing those out because there was an issue that a warning would have caught, but I'm not entirely sure what the warnings all mean.

mmuetzel:
Thank you for looking into this.
The package grokker now ran three times and failed each time while trying to download a different package.
Are we hitting a rate limit?

zanud:
_Excuse me for a huge delay and for answering questions that are not there..._

`Is -static-libgcc really necessary?`

ZLIB.DLL should be useable with any application produced by any compiler. Is there guaranty that libgcc_s_dw2-1.dll from MinGW 8.0 will be compatible with libgcc_s_dw2-1.dll from MinGW 18.0?
Last year I have a lot of troubles with application that I was developing at that time until discovered that a root of them is mutual incompatibility between Qt DLLs from MSYS2 and QtCreator environments (even though compiler was exactly the same).

```
 I'm not sure if it is a good idea to link libgcc statically in MinGW.
See also this comment in the build rule of gcc:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gcc/PKGBUILD#L227
(But that comment might be about something different... 🤷‍♂️)
```
This is hard question.

1. That discussion was about compiler that uses exceptions internally. Not been GCC developer, I can not tell where that restriction is applicable for `zlib` that is written in a pure C.
2.  I created a test program that uses exceptions and my ZLIB1.DLL. This program worked correctly regardless of whether it  was compiled with the same version of MinGW as DLL, or with a different one. But again, not been GCC developer, I can not be sure that this test means anything.

Anyway, I a chose way (ugly or clever - you decide) to completely get rid off `libgcc` completely. From `libgcc` a whole `zlib` uses only one simplest function, and this function is called from one file only. So I put a code for this function into that file.

Biswa96:
Have you tried to upstream the changes? The diff is not so small.

zanud:
> Have you tried to upstream the changes?

This is a bitter story.
File `adler32.c` contains [only one!] line with `MOD63` expression. `MOD63` is a simplest macro expanded as `%= 65521`. Completely innocent code. The only catch is: its operand is of 64 bit integer type. For 64 bit compiler this is not a problem at all - it uses corresponding hardware operation. But 32 bit compiler has to use some sort of software emulation.
In `gcclib` 64 bit integer `%` operation is implemented as `__moddi3()` function. So I simply copy-pasted it from `gcclib` sources into `adler32.c` (with some optimization for speed and size).

> The diff is not so small.

45 lines (including comments and empty lines) added.
No one existed line is changed.

zanud:
I am an old moron.
`adler32.c` already contains its own software implementation of arithmetic macros, but only for platforms without hardware division operation. So I (almost without thinking) added my own one (for x86 platform). And only yesterday I realized that in situation where one of the operands (divisor) is always the same I do not need so complex universal implementation. So I have dropped everything, and now "%" is calculated by one assembler command.

mmuetzel:
If you really think this is necessary, I'd still prefer if you could split the changes that don't affect the build system (i.e., the change to `adler32.c`) to a separate patch.

zanud:
Sorry. Initially I wrongly understood your:
> split it off to a separate patch

PR #17392, opened 2023-06-01
add documentation.

JPeterMugaas:


Biswa96:
Would it be possible to split the docs in separate packages? Also, please split the commit into two for each packages with package name at first in the commit message.

Biswa96:
It seems that the commands used to split the docs are incorrect. I have added a pull request to split the docs in separate package. See the linked pull request above. Thank you.

I was wondering if you might consider learning how to write more effective commit messages? These guidelines will help https://wiki.winehq.org/Submitting_Patches#The_commit_message

PR #17411, opened 2023-06-02
nodejs: Fix executing error

MehdiChinoune:
node fails with error:
```
Node.js is only supported on Windows 8.1, Windows Server 2012 R2, or higher.
Setting the NODE_SKIP_PLATFORM_CHECK environment variable to 1 skips this
check, but Node.js might not execute correctly. Any issues encountered on
unsupported platforms will not be fixed
```
Fix it by setting _WIN32_WINNT to Windows 8.1

### DIFF
--- a/mingw-w64-nodejs/0001-Skip-check-OS-version.patch
+++ b/mingw-w64-nodejs/0001-Skip-check-OS-version.patch
@@ -1,0 +1,25 @@
+--- a/src/node_main.cc
++++ b/src/node_main.cc
+@@ -32,22 +32,6 @@
+ #define SKIP_CHECK_VALUE "1"
+ 
+ int wmain(int argc, wchar_t* wargv[]) {
+-  // Windows Server 2012 (not R2) is supported until 10/10/2023, so we allow it
+-  // to run in the experimental support tier.
+-  char buf[SKIP_CHECK_SIZE + 1];
+-  if (!IsWindows8Point1OrGreater() &&
+-      !(IsWindowsServer() && IsWindows8OrGreater()) &&
+-      (GetEnvironmentVariableA(SKIP_CHECK_VAR, buf, sizeof(buf)) !=
+-       SKIP_CHECK_SIZE ||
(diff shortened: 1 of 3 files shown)


lazka:
Just wondering, where does this error occur? Since our "hello world" integration tests are green atm.

MehdiChinoune:
> Just wondering, where does this error occur? Since our "hello world" integration tests are green atm and maybe we can improve them.

On CLANG64 environment.

lazka:
huh, indeed, I wonder why it doesn't fail in CI..

lazka:
hm, so this is broken with clang?

```c
#include <windows.h>
#include <VersionHelpers.h>

int main(void)
{
    return IsWindows8Point1OrGreater();
}
```

lazka:
ah, this function depends on the manifest, and we don't have a default manifest with clang envs. https://learn.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindows8point1orgreater

MehdiChinoune:
> huh, indeed, I wonder why it doesn't fail in CI..
> 
> edit: ah the IsWindowsServer check I guess

I have Windows 10 which is supposed to pass but It fails.
MS docs says that IsWindows8Point1OrGreater requires a manifest file (https://learn.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindows8point1orgreater), but doesn't mention that for IsWindowsServer and IsWindows8OrGreater. Maybe that's why It passes in GHA.

lazka:
ah, this function depends on the manifest, and we don't have a default manifest with clang envs. https://learn.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindows8point1orgreater

So the "right" fix would be to apply this somehow: https://github.com/nodejs/node/blob/de4fb11e4f29b91656b831a6fb7619c94b630626/src/res/node.exe.extra.manifest#L8

MehdiChinoune:
Exactly, clang environments do not have a default manifest.

lazka:
(skipping the check is of course a valid "workaround" for now..)

PR #16643, opened 2023-04-04
dxil-redist: new package

lb90:
Contains copies of the dxil.dll redistributable libraries from the Windows SDK.

As dxil.dll is not normally present in system installations, we have to ship it ourselves.

Note that another official distribution channel for dxil.dll is https://github.com/microsoft/DirectXShaderCompiler/releases, but we avoid using that because the distributed modules are dependant upon the Visual C++ runtime, while the modules from the Windows SDK is built with the Visual C++ runtime linked statically.

TODO: figure out a valid SPDX license id?

PR #17421, opened 2023-06-02
mesa: apply patch to avoid crash on missing dxil.dll

dvrogozh:
As of now we miss dxil.dll within msys2 installation. This is being addressed by https://github.com/msys2/MINGW-packages/pull/16643, but creating a package for dxil will take some time due to resolving redistribution concerns.

In a meanwhile this commit applies patch from mesa main branch to avoid a crash on missing dxil.dll. This allows some workflows to actually work, for example, vaon12 media workflows which don't require shader operations.

Issue: https://gitlab.freedesktop.org/mesa/mesa/-/issues/8718
Upstream: https://gitlab.freedesktop.org/mesa/mesa/-/commit/098342ce53f692b719dd3aeb876737ad886c37f0

### DIFF
--- a/mingw-w64-mesa/0001-Add-checks-for-NULL-dxil_validator.patch
+++ b/mingw-w64-mesa/0001-Add-checks-for-NULL-dxil_validator.patch
@@ -1,0 +1,86 @@
+From 3136fe409a0ca0f9609604e3d908ad73ddcd8b1d Mon Sep 17 00:00:00 2001
+From: Luca Bacci <luca.bacci@outlook.com>
+Date: Mon, 27 Mar 2023 17:53:11 +0200
+Subject: [PATCH] Add checks for NULL dxil_validator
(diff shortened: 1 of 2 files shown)


dvrogozh:
@pal1000, @lazka : can you, please, help review?

pal1000:
LGTM.

dvrogozh:
@pal1000 : can you, please, advice whether `pkgrel=1` should be bumped with this kind of changes which add patches, but don't change component version (23.1.1 as of today)?

dvrogozh:
>> @pal1000 : can you, please, advice whether pkgrel=1 should be bumped with this kind of changes which add patches, but don't change component version (23.1.1 as of today)?
>
> yes, it's required, pkgrel is used exactly for such a case :)

fixed, bumped as `pkgrel=2`

PR #17426, opened 2023-06-03
rust ntapi fixes for aarch64

jeremyd2019:
These are all the rust packages I saw that were failing to build due to ntapi-related errors.

bottom: needed windows-targets 0.48.0 fix too.

dust: just needed patch to link ntdll.

delta: just needed the patch to link ntdll.

I'm not sure it's worth patching things like this...  Also, I'd like to get @mati865's thoughts on the ntdll-linking patch before opening a pull request against ntapi.


jeremyd2019:
The update to delta from #17427 removed the need to backport the fix for 64-bit Windows from ntapi, now just the same ntdll linking patch that dust needed (they both used 0.4.0, while bottom used 0.4.1, so the checksums file differed between them)

mati865:
This ntdll patch would be fine for us but not necessarily for the upstream because it would affect `windows-msvc` targets.
It should also check if target name contains or ends with `windows-gnullvm`.

mati865 (approved):

jeremyd2019:
I am curious how ntapi works on msvc targets without this.  winapi has similar logic in their build.rs, but only to determine whether to prefix the library name, not whether to try to link it (which it does for all targets).

PR #17446, opened 2023-06-04
soapysdr: Add version 0.8.1

antoniovazquezblanco:


Biswa96:
Thank you for the contribution. One thing to notice. The cmake files are installed in wrong directory. It should be installed in prefix/lib/cmake/ or prefix/share/cmake. You can compare the file list from CI log and ArchLinux's soapysdr file list https://archlinux.org/packages/extra/x86_64/soapysdr/

mingw-w64-soapysdr/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.8.1
+_gitver=soapy-sdr-$pkgver
+pkgrel=1
+pkgdesc="Vendor and platform neutral SDR support library (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+url="https://github.com/pothosware/SoapySDR"
+license=('Boost')
+depends=()
```
Biswa96, line 13:
```suggestion
depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
         "${MINGW_PACKAGE_PREFIX}-libwinpthread")
```

mingw-w64-soapysdr/PKGBUILD
```diff
@@ -0,0 +1,56 @@
...
+makedepends=("patch"
```
Biswa96, line 14: patch is not required in makedepends. It is already part of base installation. Same rule in ArchLinux also.

antoniovazquezblanco:
Updated with suggested changes. Thanks for the feedback!

Biswa96:
soapysrd is now available in mingw packages https://packages.msys2.org/base/mingw-w64-soapysdr 🎉

antoniovazquezblanco:
Fixes for this package have been merged and next release may not need the patches. :)

PR #17459, opened 2023-06-05
rtl-sdr: Add version 0.8.0

antoniovazquezblanco:
Used the same source code as in Archlinux repos (it seems that has multiple fixes not present on the original source....

See https://gitlab.archlinux.org/archlinux/packaging/packages/rtl-sdr/-/blob/main/PKGBUILD


mingw-w64-rtl-sdr/PKGBUILD
```diff
@@ -0,0 +1,48 @@
+_realname=rtl-sdr
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.8.0
+pkgrel=1
+pkgdesc="Driver for Realtek RTL2832U, allowing general purpose software defined radio (SDR) (mingw-w64)"
+arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
```
Biswa96, line 7:
```suggestion
arch=('any')
mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
```

mingw-w64-rtl-sdr/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+url="https://sdr.osmocom.org/trac/wiki/rtl-sdr"
+license=('GPL')
+depends=("${MINGW_PACKAGE_PREFIX}-libusb>=1.0"
+         "${MINGW_PACKAGE_PREFIX}-gcc-libs"
+         "${MINGW_PACKAGE_PREFIX}-libwinpthread")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-cc")
+options=('staticlibs')
```
Biswa96, line 16: This option is enabled by-default and not required here. Default options can be found in `/etc/makepkg-mingw.conf` file.

mingw-w64-rtl-sdr/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+  if check_option "debug" "n"; then
+    extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    "${MINGW_PREFIX}"/bin/cmake.exe \
+      -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+      -DCMAKE_C_FLAGS="$CFLAGS -ffat-lto-objects" \
```
Biswa96, line 35:
```suggestion
```

Biswa96:
* `__MINGW64__` is for 64 bit only.
* `__MINGW32__` is for both 32 bit and 64 bit.

Though, in this case `#ifdef _MSC_VER` check would be simpler. The logic will be like - include sys/time.h and unistd.h when Visual Studio or MSVC is not used.

Biswa96:
ArchLinux switched to a new upstream repository of this project. Can we do the same?

PR #17469, opened 2023-06-06
new package: libipt

GitMensch:
Adding Intel(R) Processor Trace (Intel PT) Decoder Library; the intended use is adding that to GDB as a follow-up.

### DIFF
--- a/mingw-w64-libipt/PKGBUILD
+++ b/mingw-w64-libipt/PKGBUILD
@@ -1,0 +1,88 @@
+# Maintainer: Bet4 <0xbet4@gmail.com>
+
+_realname=libipt
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgbase=mingw-w64-${_realname}
+pkgver=2.0+164.r903.20230324.7a863ba
+pkgrel=1
+pkgdesc='Intel(R) Processor Trace decoder library (mingw-w64)'
+arch=(
+  'i686'
+  'x86_64'
+)
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+url='https://github.com/intel/libipt'
+license=('BSD')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
+           #  'pandoc' # Only required for building the manpage, we likely don't want that
+             "${MINGW_PACKAGE_PREFIX}-cc"
+             'git')
+_commit='7a863ba'
+source=("${_realname}"::"git+${url}#commit=${_commit}")
+sha256sums=('SKIP')
+
+pkgver() {
+  cd "${srcdir}/${_realname}"
+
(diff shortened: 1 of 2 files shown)


mingw-w64-libipt/PKGBUILD
```diff
@@ -0,0 +1,90 @@
...
+
+  "${MINGW_PREFIX}"/bin/cmake.exe --build . --target test
+}
+
+package() {
+  cd "${srcdir}/build-${MSYSTEM}"
+
+  DESTDIR="${pkgdir}" "${MINGW_PREFIX}"/bin/cmake.exe --install .
+  # work around bad install name
+  mv "${pkgdir}${MINGW_PREFIX}/lib/liblibipt.dll.a" "${pkgdir}${MINGW_PREFIX}/lib/libipt.dll.a"
```
Biswa96, line 86: This should be fixed in cmake file.

mingw-w64-libipt/PKGBUILD
```diff
@@ -0,0 +1,90 @@
+# Maintainer: Bet4 <0xbet4@gmail.com>
+
+_realname=libipt
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgbase=mingw-w64-${_realname}
+pkgver=2.0+164.r903.20230324.7a863ba
```
Biswa96, line 6: Is it possible to format the version like other packages using git repository? e.g. headers-git, crt-git etc.

mingw-w64-libipt/PKGBUILD
GitMensch: I'm  (obviously) no cmake user (just started with the template matching most), so I don't know how to tackle that.
arch uses `make DESTDIR="${pkgdir}" install` her instead - not sure if that's better / will solve anything.

Could you have a look at this or provide a note _how_ to solve that in cmake file (as a patch - I guess there are similar)?

mingw-w64-libipt/PKGBUILD
GitMensch: Just took what arch had, but sure, I'll look into that.

GitMensch: I _think_ I've found the issue:

> fatal: No annotated tags can describe '7a863ba07e2baf8e3a9b584ddf35d96d500d8cca'.
However, there were unannotated tags: try --tags.

The reason is that Intel has not a recent tag: https://github.com/intel/libipt/tags so that commit does not match something.

We have one of the four options:

* use as-is
* use as-is, but manually change the pkgver in a way that pkgver() would not match
* use `--tags` to get `v2.0-164-g7a863ba`, but in this case the version dos not match the checkout
* use the ibipt version from the last tag

Any preference?

mingw-w64-libipt/PKGBUILD
lazka: I've commented on the upstream issue and added a patch

mingw-w64-libipt/PKGBUILD
lazka: I guess we went with option 5, use a commit from the stable branch after the last tag

PR #17476, opened 2023-06-08
gegl: remove exiv2 as direct dep

kmilos:
Depending on `gexiv2` is sufficient.


kmilos:
This was cosmetic preparation for easier bookkeeping - no need to actually patch/rebuild gegl for upcoming exiv2 0.28 API changes.

Looks like more trouble than benefit, so dropping.

PR #17490, opened 2023-06-10
PyOpenGL-accelerate 3.1.7

totaam:
`PyOpenGL` was updated to 3.1.7 in cd65d1a7efabec69153fc511542a8cdf5f87ce5e but this package was left behind at 3.1.6
The two packages should always be updated in tandem - perhaps a note should be added in `PyOpenGL`?

This PR also switches to using releases from pypi, as those exist, unlike the repository on github.

### DIFF
--- a/mingw-w64-python-pyopengl-accelerate/PKGBUILD
+++ b/mingw-w64-python-pyopengl-accelerate/PKGBUILD
@@ -1,10 +1,10 @@
 # Contributor: Mehdi Chinoune <mehdi.chinoune@hotmail.com>
 
 _realname=pyopengl-accelerate
+_pyname=PyOpenGL-accelerate
 pkgbase=mingw-w64-python-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
-pkgver=3.1.6
-_commit=227f9c66976d9f5dadf62b9a97e6beaec84831ca # McFletch does not release bundles on GitHub, therefore we use commit hashes to get specific versions
+pkgver=3.1.7
 pkgrel=1
 pkgdesc="Acceleration code for PyOpenGL (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


mingw-w64-python-pyopengl-accelerate/PKGBUILD
```diff
@@ -15,12 +15,17 @@ depends=("${MINGW_PACKAGE_PREFIX}-python-pyopengl")
 makedepends=("${MINGW_PACKAGE_PREFIX}-python-setuptools"
              "${MINGW_PACKAGE_PREFIX}-cython"
              "${MINGW_PACKAGE_PREFIX}-cc")
-source=("https://github.com/mcfletch/pyopengl/archive/${_commit}.tar.gz")
-sha256sums=('6b556e5aa2fd6517809e1559593038c99d3318d21155be4de3d18acfef093d40')
+source=("https://files.pythonhosted.org/packages/93/09/d08b3d07dbd88258276496a47273778f330f5ccf8390cb21b16b29d660de/${_pyname}-${pkgver}.tar.gz")
```
MehdiChinoune, line 18: Please use a permanent link, like the one in [mingw-w64-PKGBUILD-templates/PKGBUILD.python.sh](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.python.sh)

totaam: @MehdiChinoune I've always used the long URLs you copy and paste directly from pypi and none of them have ever expired.
Where can you copy the other form from?

MehdiChinoune: Sorry, for the confusion, I said permanent as It's valid for any version.
The form I mentioned allows auto-updating by just changing pkkgver and running updpkgsums.

MehdiChinoune, line 18:
```suggestion
source=("https://pypi.org/packages/source/${_pyname::1}/${_pyname}/${_pyname}-${pkgver}.tar.gz")
```

PR #17506, opened 2023-06-12
miniupnpc: split python bindings to separate package

Biswa96:


Biswa96:
@scribam Would you like to review this pull request please?

scribam:
Does python-miniupnpc need to have the LICENSE file as well? Arch seems to install it: https://gitlab.archlinux.org/archlinux/packaging/packages/miniupnpc/-/blob/main/PKGBUILD#L50

Otherwise, LGTM

scribam (approved):

PR #17507, opened 2023-06-12
clang: Update to 16.0.5

lazka:
\+ a backport for an issue affecting meson: https://github.com/mesonbuild/meson/issues/11642#issuecomment-1586524670


jeremyd2019:
I wasn't sure when I was doing the dlltool patches if I should use the github .patch URL or include the patch files as in the past.  I went with including the patch files, mainly because the existing patches are applied in the subdirectory so the patch files need to be generated with `--relative`.

lazka:
thanks for taking over

> I wasn't sure when I was doing the dlltool patches if I should use the github .patch URL or include the patch files as in the past. I went with including the patch files, mainly because the existing patches were applied in the subdirectory so the patch files needed to be generated with `--relative`.

I'll try to match it next time.. I just went with the easier thing :/

PR #17508, opened 2023-06-12
llvm 16.0.5 update (1/3)

jeremyd2019:
backport llvm-dlltool patches

These are necessary to be able to bootstrap rust on CLANG* environments
using the upstream *-pc-windows-gnu binaries.  They now call dlltool
with arguments that up to now were not supported by llvm-dlltool.

Also update README-patches.md


jeremyd2019:
oops, conflict with #17507

PR #17516, opened 2023-06-13
emacs-pdf-tools-server: update to 1.1.0

cyrilarnould:
Two points of note:

- I did not find a way to verify the checksum, as it is not included in the [release](https://github.com/vedang/pdf-tools/releases/tag/v1.1.0) and there doesn't seem to be a way to [view it on Github](https://github.com/orgs/community/discussions/23512). I determined it from my locally downloaded copy. If this causes security concerns, please indicate how I should determine the checksum.
- The binary should be used together with v1.1.0 of the Emacs package as available via [NonGNU ELPA](https://elpa.nongnu.org/nongnu/pdf-tools.html) or [MELPA Stable](https://stable.melpa.org/#/pdf-tools) (the latter of which seems to still be stuck on v1.0.0 though). Discrepancies between the lisp package and the binary can otherwise [cause issues](https://github.com/vedang/pdf-tools/commit/a9c9a12c3ecf2005fa641059368ac8284f507620). I've included this information as a comment in relation to the package version since I didn't find a better place to document this.

### DIFF
--- a/mingw-w64-emacs-pdf-tools-server/PKGBUILD
+++ b/mingw-w64-emacs-pdf-tools-server/PKGBUILD
@@ -3,12 +3,13 @@
 _realname=emacs-pdf-tools-server
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.0
+# Issues may arise if not used with the same release of the pdf-tools emacs package
+pkgver=1.1.0
 pkgrel=1
 pkgdesc="Emacs support library for PDF files"
 source=("${_realname}-${pkgver}.tar.gz::https://github.com/vedang/pdf-tools/archive/refs/tags/v${pkgver}.tar.gz")
 noextract=(${_realname}-${pkgver}.tar.gz)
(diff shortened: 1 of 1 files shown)


cyrilarnould:
For reference, I fixed the clang build using https://github.com/vedang/pdf-tools/pull/134

Biswa96:
The patch file seem to be applied in v1.1.0 tag in upstream repository. How is it required here with that same version?

cyrilarnould:
It's explained in the linked PR above, the patch has been reverted immediately afterwards because it broke compilation on MacOS. It is not applied in v1.1.0.

It wasn't applied in v1.0.0 either in fact, which means building v1.0.0 should fail as well. I've just tested this on my local machine and indeed it does fail, some change in the last year in `mingw-w64-clang` or maybe in the compiler flags must have brought forth this error.

Biswa96:
Do we need that patch? Previously, the build failed due to vasprintf() function was not declared. But vasprintf() is declared in stdio.h, one have to define `_GNU_SOURCE` only.

cyrilarnould:
Whereas most online sources I found suggest only need to define `_GNU_SOURCE`, I found the solution in the following post where a comment mentions they also need to pass the `-D_GNU_SOURCE` flag:

https://stackoverflow.com/questions/61306157/unable-to-compile-program-due-to-function-asprintf

PR #17526, opened 2023-06-14
emacs-pdf-tools-server: update to 1.1.0

svraka:
CLANG builds fail, investigating.


Biswa96:
Duplicate of https://github.com/msys2/MINGW-packages/pull/17516

PR #17533, opened 2023-06-14
findutils: new package

cyrilarnould:
Redo of #16846. 

Most of the patches have been adapted from Eli Zaretskii's [ezwinports](https://sourceforge.net/projects/ezwinports/files), with the following exceptions:

Patch 0001 adds the poll and `windows-stat-inodes` gnulib modules. The `poll` module provides the poll functionality that is absent on mingw. `windows-stat-inodes` ensures the `-inum` option of `find` is working properly. The modules have been taken from the same gnulib commit that is linked in the `findutils` 4.9.0 release (`v0.1-5153-g6ef3d78333`).

Patch 0002 modifies the `configure` script to allow using the `mountlist` module on mingw. The code doesn't compile without it but it also doesn't seem broken. The `mountlist` module also specifies that using the header is fine on unsupported systems.

Patches 0009, 0012, 0013, and 0014 disable features that had not been implemented yet in the 4.2.30 `findutils` release that ezwinports is based on. I also don't know how I'd implement them; they are all based on functionality that is not available on mingw as far as I can tell.

Patch 0008 tries to address the differences in how the `PATH` environment variable is separated on Windows.

Patches 0006 and 0015 are fairly straightforward compilation error fixes, the latter one coming from a future upstream commit.

Patch 0016 fixes linking on clang builds.

Note that not all of the tests pass. I could provide patches for some of them if this is preferable, but for others I think there's no use case on Windows. I keep a write up in the repository which the patches are based off of:  

https://github.com/cyrilarnould/findutils

### DIFF
--- a/mingw-w64-findutils/0001-Add-poll-and-windows-stat-inodes-gnulib-modules.patch
+++ b/mingw-w64-findutils/0001-Add-poll-and-windows-stat-inodes-gnulib-modules.patch
@@ -1,0 +1,3376 @@
+From 756f946ea0c68d5248dd2dd4ab20f33a92ce5c71 Mon Sep 17 00:00:00 2001
+From: Cyril Arnould <cyril.arnould@outlook.com>
+Date: Tue, 13 Jun 2023 16:33:43 +0200
+Subject: [PATCH 01/17] Add poll and windows-stat-inodes gnulib modules
+
+* poll.h is missing in mingw
+* windows-stat-inodes is necessary to get -inum option working
+---
+ bootstrap.conf    | 2 ++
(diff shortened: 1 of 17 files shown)


Kreijstal:
it has unsynced from upstream

cyrilarnould:
Upstream's `v4.10.0` is out, so I've added a [findutils-4.10.0](https://github.com/cyrilarnould/MINGW-packages/tree/findutils-4.10.0) branch to my fork. Some patches had to be modified, most of them are still valid.

I'll leave this PR as is for now unless updating it would be preferred.

PR #17540, opened 2023-06-14
[DO NOT MERGE] rust: revert to bootstrapping=yes for clang and depend on llvm-libs

filnet:


filnet:
Follow up on  https://github.com/msys2/MINGW-packages/pull/17406.

filnet:
```
package_llvm-libs() {
  pkgdesc="Low Level Virtual Machine Runtime Libraries (mingw-w64)"
  depends=("${MINGW_PACKAGE_PREFIX}-libffi"
           "${MINGW_PACKAGE_PREFIX}-libxml2"
           "${MINGW_PACKAGE_PREFIX}-zstd")
[...]
```

So we should remove `libffi` and `zlib` from makedepends ?

ognevnydemon:
> So we should remove `libffi` and `zlib` from makedepends ?

yes, it should be ok

ognevnydemon:
tested this package from CLANG64 artifact. compiled rustpython and my own project successfully

lazka:
> So we should remove libffi and zlib from makedepends ?

split package deps are ignored when building.

jeremyd2019:
> > So we should remove libffi and zlib from makedepends ?
> 
> split package deps are ignored when building.

I think this was comparing the depends of llvm-libs against the makedepends of rust, so they would be considered (as the depends of a makedepends), so I think it comes down to whether rust directly makes use of the dependencies or if they are only needed because llvm-libs needs them.

filnet:
> > > So we should remove libffi and zlib from makedepends ?
> > 
> > 
> > split package deps are ignored when building.
> 
> I think this was comparing the depends of llvm-libs against the makedepends of rust, so they would be considered (as the depends of a makedepends), so I think it comes down to whether rust directly makes use of the dependencies or if they are only needed because llvm-libs needs them.

Yes, I was comparing the dependencies of llvm-libs to the makedepends of Rust.
The idea was to not have in the makedepends things that are provided by llvm-libs.

mingw-w64-rust/PKGBUILD
```diff
@@ -23,7 +23,7 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-libffi"
              "${MINGW_PACKAGE_PREFIX}-libssh2"
              "${MINGW_PACKAGE_PREFIX}-lldb"
-             "${MINGW_PACKAGE_PREFIX}-llvm"
+             "${MINGW_PACKAGE_PREFIX}-llvm-libs"
```
jeremyd2019, line 26: Maybe @mati865 or somebody who knows rust better could weigh in here, but I don't think this change is a good idea.  Reasoning in prior comment ☝️

jeremyd2019 (changes requested):

mingw-w64-rust/PKGBUILD
mati865: This isn't right, build will fail without `llvm-config`.
So why did it succeed? I suppose `lldb` pulled in `clang` which pulled in `llvm`.

jeremyd2019: Oh.  I guess then a case could be made that clang should depend on llvm-libs instead of llvm now (assuming it doesn't need any of that other stuff at runtime), but definitely not changing rust's makedepends.

lazka: We kinda depend on "cc" pulling in a complete toolchain in various places, including cli tools, so I'm not sure that's a good idea

filnet: So the consensus is to not change the makedepends at all.

Issue #17374, opened 2023-05-30
Re-enabling native compilation in Emacs


svraka:
Recently native compilation support was turned off in bfa4238 while merging https://github.com/msys2/MINGW-packages/pull/17129 due to build timeouts. The reason given in the commit message was slow compilation under GCC 13. Based on my testing I don't think that's the case, although the build still fails in https://github.com/svraka/MINGW-packages/tree/emacs-libgccjit.

After reverting bfa4238 the CI gets stuck ([CI summary](https://github.com/svraka/MINGW-packages/actions/runs/5115456102/jobs/9196734675)) at exactly the same step as a couple of weeks ago with @Biswa96's commit ([CI summary](https://github.com/msys2/MINGW-packages/actions/runs/4967459226/jobs/8889468717?pr=17129)), with the log displaying `ELN  ../../emacs-28.2/lisp/x-dnd.eln`.

The build reached this point fairly quickly, in just over 10 minutes in both cases. In my local testing the whole build from a fully up-to-date MSYS2 takes about 10 minutes (the CI would take more, as it needs to install dependencies) with `-j2` (emulating the GitHub runner with two cores).

Given how `make` handles targets in parallel builds, this no guarantee that we are stuck exactly at the same point but it is at least *very* suspicious.

After adding [Telemetry](https://github.com/marketplace/actions/workflow-telemetry) to the CI actions (see [CI summary](https://github.com/svraka/MINGW-packages/actions/runs/5121360287)) I don't see any resource bottlenecks but the CPU graph is pretty strange. CPU usage in the first 20 minutes what I would expect. But later it looks like there's not much going on.

So I added the [debugging with tmate](https://github.com/marketplace/actions/debugging-with-tmate) action to the CI steps and SSHd into the CI machine. `lisp/x-dnd.eln` in the source tree is successfully native compiled, and at this point native compilation of `lisp/files.eln` is the only running target but it's hung with 0% CPU usage (linebreaks added for clarity):

```
C:\_\B\src\build-UCRT64\src\emacs.exe \
    -batch --no-site-file --no-site-lisp \
    --eval "(setq load-prefer-newer t)" \
    -l comp -f byte-compile-refresh-preloaded \
    --eval "(batch-native-compile t)" ../../emacs-28.2/lisp/files.el
```

I can kill this process and the build will successfully finish without native compiling `files.el`.

Can someone try building Emacs with native compilation support on their machine? I have no idea what could be the problem with the CI.

[...]

PR #17549, opened 2023-06-16
emacs: Re-enable native-compilation, disable -foptimize-sibling-calls

cyrilarnould:
For now this is a draft to test whether the build succeeds, I lack the knowledge to set up the CI.

I might have found the optimization that breaks the native-compilation build, `-foptimize-sibling-calls`. See also https://debbugs.gnu.org/cgi/bugreport.cgi?bug=63365.


cyrilarnould:
Resolves #17374

lazka:
ok to merge?

PR #17573, opened 2023-06-19
inkscape: Update to 1.3beta0

lazka:
1.2.x is EOL and no longer builds


MehdiChinoune:
CLANG* environments needs another patch https://gitlab.com/inkscape/inkscape/-/merge_requests/5383

PR #17576, opened 2023-06-19
add libipt to GDB

GitMensch:
note: if linking does not work configure will not abort so needs a manual verification - which I've did:

* earlier the result of `record btrace pt` was
> Intel Processor Trace support was disabled at compile time.
* now, on both an an AMD box an modern I7, the result is
> You can't do that when your target is `native'

Not sure how to go on from that... so asked on the libipt repo https://github.com/intel/libipt/issues/100.


GitMensch:
closed as there is no Intel PT userspace API in the Windows Operating system - and we can't add that either

PR #17586, opened 2023-06-20
svt-av1: use on clangarm64

kmilos:


kmilos:
Dropped ffmpeg changes, requires 32-bit vulkan patching I don't know how to deal with...

lazka:
I'll have a look, probably the same fixes that we needed for ffmpeg4.4 recently

lazka:
https://github.com/msys2/MINGW-packages/pull/17587

Issue #17579, opened 2023-06-19
MuPDF has a very ugly icon


angelog0:
### Description / Steps to reproduce the issue

MuPDF has a very bad/ugly icon compared with the upstream binaries. May you do something about this?

### Expected behavior

.

### Actual behavior

.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Biswa96:
Could you provide any screenshot to compare the icons?

angelog0:
In MSYS2 I have installed the same `gl` version, 

```
$ pacman -Ss mupdf | grep mingw64
mingw64/mingw-w64-x86_64-mupdf-mupdf-gl 1.22.1-2 [installato]
```
This upstream:
![mupdf-upstream](https://github.com/msys2/MINGW-packages/assets/35931460/72cbeba0-fd77-4bcd-a2cc-01234b03095d)

and this from MSYS2:
![mupdf-MSYS2-MINGW64](https://github.com/msys2/MINGW-packages/assets/35931460/254e2a1e-c468-4a1b-8dcf-50dc41b2fd41)

angelog0:
I asked this because, usually, I put a link to Start Menu and having that icon in Start is a bit annoying...

Biswa96:
Hi @angelog0, Thank you for reporting the issue. I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request?

PR #17588, opened 2023-06-20
mupdf: update to 1.22.2

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/17579


Biswa96:
The patch was added in this upstream commit https://git.ghostscript.com/?p=mupdf.git;a=commit;h=a4e6c3ba091f73565bc76b1a1931e6513eed2098

PR #17591, opened 2023-06-20
Add sdl12-compat

lazka:
try to build it as a normal library without -nostdlib -static-libgcc stuff, which breaks for example stack protector. I might be missing something though.

Also no "replaces" for now, until we are sure it can replace sdl1


Biswa96:
Some related discussion https://discord.com/channels/792780131906617355/1059806620353318993/1100325471410978846

lazka:
thanks. any objections to this hackery then?

Biswa96:
No idea.

lazka:
Let's give it a try then. At least the examples run fine here.

PR #17598, opened 2023-06-21
exiv2: Update to 0.28.0

lazka:
new libinih dependency

EXIV2_ENABLE_WIN_UNICODE was removed in https://github.com/Exiv2/exiv2/pull/2090


lazka:
From what I see this has ABI/API breaks for all users and removed all the unicode APIs.. so I'm not sure what to do.

Biswa96:
Similar PR was pending.

PR #17602, opened 2023-06-21
darktable: update to 4.4.0

kmilos:
w/ some deps cleanup


Biswa96:
Should this be merged after the exiv2 update?

kmilos:
Nah, go for it.

There's still exiv2 dependants to be patched upstream, and .0 has some bugs/regressions, so we might as well wait for .1...

PR #17603, opened 2023-06-22
python-numba: update to 0.57.1

MehdiChinoune:


ognevnydemon:
it seems that numba is broked again by numpy. will test today

PR #17615, opened 2023-06-23
libipt: update to 2.0.6

GitMensch:
also allows dropping the import-lib-naming.patch which is now included


mingw-w64-libipt/PKGBUILD
```diff
@@ -12,24 +12,22 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
...
 
 pkgver() {
   cd "${srcdir}/${_realname}"
 
   git describe --long --tags --abbrev=7 | sed 's/^v//;s/\([^-]*-g\)/r\1/;s/-/./g'
 }
 
-prepare() {
-  cd "${srcdir}/${_realname}"
+# prepare() {
```
Biswa96, line 25: The comments can be removed.

GitMensch: done

PR #17641, opened 2023-06-25
python-soundfile: Update to 0.12.1

lazka:


lazka:
not much we can do re numba, so merging

Issue #17564, opened 2023-06-18
lua-lpeg doesn't build if luajit is installed instead of lua51


FractalU:
### Description / Steps to reproduce the issue

The lua-lpeg package doesn't build if luajit is installed instead of lua51.

A linker error is thrown:
```
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpvm.o:lpvm.c:(.text+0x9e): undefined reference to `lua_newuserdatauv'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpvm.o:lpvm.c:(.text+0x150): undefined reference to `lua_newuserdatauv'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpvm.o:lpvm.c:(.text+0x4b8): undefined reference to `lua_rotate'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpcap.o:lpcap.c:(.text+0x35e): undefined reference to `lua_rotate'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpcap.o:lpcap.c:(.text+0x48b): undefined reference to `lua_callk'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpcap.o:lpcap.c:(.text+0x637): undefined reference to `lua_rotate'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 lpcap.o:lpcap.c:(.text+0x65c): undefined reference to `lua_callk'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
[...]

PR #17650, opened 2023-06-26
lua-lpeg: derive include path from package config and fix CFLAGS in makefile patch

FractalU:
Fixes https://github.com/msys2/MINGW-packages/issues/17564


mingw-w64-lua-lpeg/lpeg-0.12-makefile.patch
```diff
@@ -9,12 +9,15 @@
...
  # -Wunreachable-code \
  
  
 -CFLAGS = $(CWARNS) $(COPT) -std=c99 -I$(LUADIR) -fPIC
-+CFLAGS += $(CWARNS) $(COPT) -std=c99 -I$(LUADIR) -fPIC
++ifdef DLLFLAGS
++	HOST_CFLAGS := $(CFLAGS)
++	CFLAGS += $(CWARNS) $(COPT) -std=c99 -I$(LUADIR) -fPIC
++endif
  CC = gcc
```
mmuetzel, line 21: To fix the build error in the clang environments, you might need to additionally change this like so:
``` suggestion
 CC ?= gcc
```

mmuetzel: > Or, `make CC=${CC}` may work.

That might be the better choice because it doesn't require additional patching. 👍

FractalU: These solutions don't work. It always defaults back to gcc. `${CC}` from PKGBUILD should be `clang` when compiling from clang64 or clang32, but it's `gcc`. Also when setting `CC` in makefile to clang by hand, then a segmentation fault occurs in a test script.

```
==> Starting check()...
./test.lua
General tests for LPeg library
version 1.0.2
+
+
+
+
+
OK
./test.lua
make: *** [makefile:47: test] Segmentation fault
==> ERROR: A failure occurred in check().
    Aborting...
```

FractalU: The behavior that `${CC}` in PKGBUILD always defaults to gcc even if running from clang64 or clang32 happens on my machine regardless which package it is. I'm gonna update this PR and see what happens.

Biswa96: Please use `makepkg-mingw` command to create package.

FractalU: Thanks, that fixed the issue with CC. The segmentation fault error on the other hand is still present.

FractalU: The tests here have passed I don't know what causes the segmentation fault on my machine. I guess it has something to do with my installation.

I commented out CC as`#CC = gcc` in the makefile patch. It's the most consistent with other lua packages. It also works well with the other commented line `#COPT = -O2 -DNDEBUG` in the makefile patch.

FractalU: The segmentation fault turns out to be a real bug in this package. This error occurs when `mingw-w64-{mingw_arch}-x86_64-lua-lpeg` package is not installed. It mostly happens on a fresh msys2 installation. This is due to a shebang line in `test.lua`. The line `#!/usr/bin/env lua` makes this script to be executed from `lua.exe` This script loads then the lpeg package from `lpeg.dll`. If the lua runtime and the loaded dll is incompatible then a segmentation fault error appears. This is the case when the lua runtime is `lua.exe` and the loaded dll is compatible with lua5.1. If `mingw-w64-{mingw_arch}-x86_64-lua-lpeg` is installed, then the loaded dll is always compatible with lua5.4 because the installed `lpeg.dll` is being loaded first. This is why on most platforms this package builds.

I'm currently trying to figure out how to resolve this issue properly. So this PR is not ready to be merged yet.

FractalU:
The segmentation fault bug has been fixed by the commit https://github.com/msys2/MINGW-packages/pull/17650/commits/84416cf223abb8b6f8a8c3f75004403991ec0b7d, which also fixes the testing of freshly built `lpeg` libraries. Previously, the testing didn't work at all. It seemed to work because it tested the already installed `lpeg` libraries instead of the ones that have been built. I removed the `make test` in both lua 5.4 and 5.1 versions and replaced it with `lua -e "package.path=[[?.lua]] package.cpath=[[?.dll]]" test.lua` and `${lua51} -e "package.path=[[?.lua]] package.cpath=[[?.dll]]" test.lua`. The modification of the search paths ensures that only the respective library in the current working directory is being loaded, which is exactly the one that has been built. For lua 5.1 I added the `lua51` variable which is either `lua5.1` or `luajit`. This variable is determined by the condition whether luajit is installed or not. I did it that way because the installed binary cannot be determined from the package config.

All issues I know of this package has been fixed. This PR can be merged now.

FractalU:
I've added some small improvements.

FractalU:
I've deleted a superfluous `sed` line. The command in that line tries to modify the shebang line in `test.lua`. Previously, I modified the `PKGBUILD` file to run `test.lua` directly from lua/lua51 to resolve the segmentation fault bug, therefore the shebang line in `test.lua` is irrelevant. Also, the `sed` command in that superfluous line is wrong and doesn't do what intended. It assumes that the shebang line in `test.lua` is `#!/usr/bin/env lua5.1` but that line from the current lpeg version is instead `#!/usr/bin/env lua`. Therefore the `sed` command doesn't do anything at all and leaves the shebang line in both lua and lua 5.1 builds`unchanged instead of `#!/usr/bin/env lua` for the lua build and `#!/usr/bin/env lua5.1` for the lua 5.1 build.

PR #17655, opened 2023-06-27
lcms2: Enable Fast Float plugin to be built

cmyk-student:
There was a request for GIMP to add an option for LittleCMS optimization (https://gitlab.gnome.org/GNOME/gimp/-/issues/9536). This requires the lcms2 Fast Float plug-in to be built with lcms2. [A comment by lillolollo](https://gitlab.gnome.org/GNOME/gimp/-/issues/9536#note_1760997) indicated this can be done by adding the -Dfastfloat=true flag to PKGBUILD.

I have tested this locally and confirmed it still builds lcms2, now with the Fast Float plug-in as well. If anything else is needed, please let me know!

### DIFF
--- a/mingw-w64-lcms2/PKGBUILD
+++ b/mingw-w64-lcms2/PKGBUILD
@@ -46,7 +46,7 @@ build() {
       --auto-features=enabled \
       --buildtype=plain \
       -Ddefault_library=both \
-      ../${_realname}-${pkgver}
+      ../${_realname}-${pkgver} \
       -Dfastfloat=true
 
   meson compile
--- a/mingw-w64-lcms2/PKGBUILD
+++ b/mingw-w64-lcms2/PKGBUILD
@@ -47,6 +47,7 @@ build() {
       --buildtype=plain \
       -Ddefault_library=both \
       ../${_realname}-${pkgver}
+      -Dfastfloat=true
 
   meson compile
 }
--- a/mingw-w64-lcms2/PKGBUILD
+++ b/mingw-w64-lcms2/PKGBUILD
@@ -5,7 +5,7 @@ _realname=lcms2
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.15
-pkgrel=2
+pkgrel=3
 pkgdesc="Small-footprint color management engine, version 2 (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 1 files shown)


Biswa96:
Please bump the pkgrel value. Also, put the realname-pkgvar path at last in meson setup command, just for aesthetic.

cmyk-student:
@Biswa96 Ah, sorry about that. I think I fixed it - let me know if there are other issues. Thanks again for the help!

kmilos:
The fast float plugin (and other plugins) were not enabled on purpose for a reason - the license is different and incompatible to the main package.

Please revert, or package the plugins separately under the respective licenses.

cmyk-student:
> The fast float plugin (and other plugins) were not enabled on purpose (and are default to off) for a reason - their license is different and incompatible to the main library.
> 
> Please revert, or package the plugins separately under the respective licenses. @Biswa96 @cmyk-student

@kmilos Hi! According to the [readme](https://github.com/mm2/Little-CMS/blob/master/plugins/README.1ST), LCMS2 is MIT and fast float is GPL3. And if I'm reading this right, ["MIT"](https://www.gnu.org/licenses/license-list.en.html#Expat) is compatible with GPL. That said, I'm not an expert in licensing so I may be mistaken.

In that case, what should the new package be called? And would it just be a copy of what we have now, and then the main lcms2 package would remove the -Dfastfloat=true line?

kmilos:
The other way: GPL is not compatible w/ MIT.

Not necessarily a new package, sub-packages can be created from one PKGBUILD, but @Biswa96 can explain better, I haven't done it yet...

cmyk-student:
@kmilos I see. Thanks for letting me know! Happy to make whatever changes are needed.

kmilos:
Or perhaps it's sufficient to ship the separate `COPYING-fast_float` license file and update the package `license=()` metadata? I'll let @Biswa96 and @lazka decide...

The other problem I see is that it ends up in `lcms2.pc` so not that easy to decaople for projects that don't want to link to it... That might have to be solved upstream though.

Biswa96:
OK, I shall try to send patch to split pkgconfig files.

cmyk-student:
> OK, I shall try to send patch to split pkgconfig files.

@Biswa96 Thanks, and sorry for the trouble!

kmilos:
@Biswa96 Good luck, it looks like the author already decided against this... 🤷

Biswa96:
BTW, wouldn't splitting the lcms2 plugin break compatibility with other use cases? For example, one has to add lcms2_fast_float explicitly in linking libraries. But in other systems, adding lcms2 only would be enough.

kmilos:
Possibly. Hence the request upstream so it's uniform for all systems.

Biswa96:
> OK, I shall try to send patch to split pkgconfig files.

The patch can not applied in upstream. lcms2 maintainer do not want to break the compatibility. Also, he showed that if a program do not use the plugin APIs, it would not link with fastfloat or threaded library. It would not cause any license issue, right?

PR #17661, opened 2023-06-29
lcms2: update licensing info

kmilos:
Following up from https://github.com/msys2/MINGW-packages/pull/17655 @Biswa96 @cmyk-student 

Let's provide all the licensing information at the very least.

I can't verify/confirm if linking w/ `lcms2_fast_float` but not using any of its symbols is sufficient to avoid actual linkage. Perhaps it relies on `strip` being run on the object, don't know if that would always be the case...

### DIFF
--- a/mingw-w64-lcms2/PKGBUILD
+++ b/mingw-w64-lcms2/PKGBUILD
@@ -5,12 +5,12 @@ _realname=lcms2
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.15
-pkgrel=3
+pkgrel=4
 pkgdesc="Small-footprint color management engine, version 2 (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> I can't verify/confirm if linking w/ `lcms2_fast_float` but not using any of its symbols is sufficient to avoid actual linkage. Perhaps it relies on `strip` being run on the object, don't know if that would always be the case...

It's not about strip. It's about `-Wl,--as-needed`, which meson uses by default. This avoids generating linkage tags during the link stage if the DLL never gets used.

Not all buildsystems set this, although it's always possible to inject it via LDFLAGS.

kmilos:
> It's not about strip. It's about `-Wl,--as-needed`, which meson uses by default (helpful if the program linking to lcms2 uses meson). This avoids generating linkage tags during the link stage if the DLL never gets used.
> 
> Not all buildsystems set this, although it's always possible to inject it via LDFLAGS.

Thanks. Would you mind letting the author know at the upstream issue this is not always implicit?

PR #17663, opened 2023-06-29
opencv: update to 4.8.0

MehdiChinoune:


lazka:
should we drop minwg32 gmic?

MehdiChinoune:
I think so.

PR #17669, opened 2023-06-30
angleproject: update to 2.1.r21358

raedrizqie:
- enable vulkan backend
- linked with internal vulkan statically

### DIFF
--- a/mingw-w64-angleproject/001-add-mingw-toolchain.patch
+++ b/mingw-w64-angleproject/001-add-mingw-toolchain.patch
@@ -110,7 +110,7 @@ index 35c9a6275..7cf9f574d 100644
      }
    }
 diff --git a/config/BUILDCONFIG.gn b/config/BUILDCONFIG.gn
-index cc7e92ddb..680b5b6c9 100644
+index 1fa42b1b1..bab13adae 100644
 --- a/config/BUILDCONFIG.gn
 +++ b/config/BUILDCONFIG.gn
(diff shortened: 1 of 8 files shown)


raedrizqie:
@lazka @Biswa96 can we merge this?

Biswa96:
> linked with internal vulkan statically

Isn't it possible to use the mingw vulkan packages?

raedrizqie:
maybe, but needs to patch vulkan deps' build.gn

lazka:
arm64 failed with:

```
2023-07-13T06:51:06.1366400Z ../../third_party/SwiftShader/src/System/Timer.cpp:63:24: error: use of undeclared identifier 'ARM64_PMCCNTR_EL0'
2023-07-13T06:51:06.1366936Z 
2023-07-13T06:51:06.1367243Z         return _ReadStatusReg(ARM64_PMCCNTR_EL0);
```

lazka:
swiftshader is 1.7GB and makes the source package 2.5GB, which is too large for github

lazka:
I moved it to a .zip from git, maybe that helps

lazka:
now arm64 fails with:

```
ld.lld: error: undefined symbol: llvm::sys::CompareAndSwap(unsigned int volatile*, unsigned int, unsigned int)
ld.lld: error: undefined symbol: llvm::sys::MemoryFence()
```

also in SwiftShader

raedrizqie:
..or maybe you could try using the latest swiftshader?

PR #17670, opened 2023-06-30
Add new fltk-snapshot package

felixxing:
This is the weekly snapshot version of fltk.
I believe it is stable enougth, so we can accept this new package.


Biswa96:
Beta versions are not accepted as packages when a stable version is present already.

felixxing:
Oh ok. sorry.

PR #17672, opened 2023-07-01
openal: Add Cflags.private in pkgconfig file

Biswa96:
This defines AL_LIBTYPE_STATIC macro while static linking. Previously, that macro was only defined in pkgconfig file of static openal build.


Biswa96:
This removes the following symbols in libavdevice.a, one of  the static library in ffmpeg 6.

```
__imp_alcCaptureCloseDevice
__imp_alcCaptureOpenDevice
__imp_alcCaptureSamples
__imp_alcCaptureStart
__imp_alcCaptureStop
__imp_alcGetError
__imp_alcGetIntegerv
__imp_alcGetString
```

PR #17671, opened 2023-07-01
librsvg: Update to 2.56.2

lazka:

PR #17676, opened 2023-07-01
librsvg: remove clangarm64 workaround

lazka:
No longer needed with windows-targets 0.48.1
see https://github.com/msys2/MINGW-packages/pull/17671#discussion_r1248543246


jeremyd2019:
Thanks.  I considered it a "feature" of using `--versioned-dirs` that the sed would fail due to file not found when 0.48.0 was no longer pulled in, hopefully making it obvious when the workaround was no longer needed.

PR #17692, opened 2023-07-02
pkgconf: update to 1.9.5 and add patch to disable sorting

1480c1:
Based on work done by @MehdiChinoune, should fix https://github.com/msys2/MINGW-packages/issues/13390.

Tested by compiling FFmpeg and a bunch of other dependencies, and seems to work.


lazka:
I guess there is no explanation as to why this fixes things?

1480c1:
Not really. I guess because qsort is unstable is the only real explanation? Not sure why it doesn't break on Linux, unless they don't test pure static compilation?

MehdiChinoune:
Could this PR merged.

lazka:
if you both feel reasonably confident, sure.

I was planning to write some pkgconf tests, but I can always do that later.

1480c1:
This PR could wait if you feel like you can get tests written soon, but would of course prefer to be merged sooner than later. As far as I know, nothing that I know of requires 1.9.5 vs our current version, as it seems 1.9+ is mainly performance and dependency resolving.

lazka:
I'm busy over the weekend, so likely not too soon. Let's see :)

PR #17718, opened 2023-07-03
suitesparse: update to 7.1.0

mmuetzel:


mmuetzel:
I downloaded and installed the MINGW64 build artifact from the CI run of this PR. With it, the test suite of Octave still passes.

PR #17720, opened 2023-07-04
cmake: Fix building to shared protobuf

MehdiChinoune:


MehdiChinoune:
reported in: https://gitlab.kitware.com/cmake/cmake/-/issues/25011
MR: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8609

PR #17732, opened 2023-07-06
Add a new package: sigutils

antoniovazquezblanco:
This package does not use Ninja as CMake backend due to a bug in the program.

A fix has been commited to the original sw (https://github.com/BatchDrake/sigutils/pull/48) but a release is going to take some time.


antoniovazquezblanco:
Added an updated git version of the package

lazka:
Sorry for the delay:

*  we don't generally have both a stable and a git version in the repo, so I removed the git repo. We can also switch that around if you want though.
* I've added a patch so it installs the DLL in the right directory
* I've removed the arm64 build as one dependency isn't available there.

I'm fine with merging this as is, but let me know your thoughts first.

antoniovazquezblanco:
That looks perfect to me. Stable version is the priority.

Git version is interesting for GitHub CI builds. I would love to have it too. Is there any repo for git builds?

Thank you so much!

lazka:
> Git version is interesting for GitHub CI builds. I would love to have it too. Is there any repo for git builds?

as in pacman repo? no. There is https://github.com/msys2/MINGW-packages-dev from when we cleaned this repo up, but it hasn'T seen much maintenance since then.

You could build it yourself in CI maybe?

Biswa96:
> Git version is interesting for GitHub CI builds. I would love to have it too. Is there any repo for git builds?

That CI build should be added in upstream repository.

LABPBS:
> Git version is interesting for GitHub CI builds. I would love to have it too. Is there any repo for git builds?

Whenever i need that i fork this repo and use the GHA based CI there

Im gonna take this space to thank the developer(s) of that 😄

PR #17758, opened 2023-07-12
libva & libva-utils: update to 2.19.0

dvrogozh:
Custom patch we were applying before got removed due to similar fix was done on libva side:
```
1114bd2 va/win32: rely on compiler to define link names
```
Setting LIBVA_DRIVERS_PATH is still needed to fetch vaon12 driver:
```
export LIBVA_DRIVERS_PATH=/mingw64/bin
```


Biswa96:
Thank you for the contribution and upstreaming the changes.

PR #17494, opened 2023-06-10
ncurses: Correctly report Backspace key

Sonic-Amiga:
Recognize and correctly report Backspace key to applications.

PR #17759, opened 2023-07-12
ncurses: update to 6.4.20230708

mmuetzel:
That should be including a fix for the issue also addressed in #17494.

### DIFF
--- a/mingw-w64-ncurses/PKGBUILD
+++ b/mingw-w64-ncurses/PKGBUILD
@@ -4,7 +4,7 @@ _realname=ncurses
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 _base_ver=6.4
-_date_rev=20230211
+_date_rev=20230708
 pkgver=${_base_ver}.${_date_rev}
 pkgrel=1
 pkgdesc="System V Release 4.0 curses emulation library (mingw-w64)"
@@ -20,7 +20,7 @@ source=("https://invisible-mirror.net/archives/ncurses/current/${_realname}-${_b
         002-ncurses-config-win-paths.patch
         ncurses-6.3-cflags-private.patch
         ncurses-6.3-pkgconfig.patch)
-sha256sums=('657d41b7340c5ac5362ae84ba1ece70c385250dc16a467d6ec748bc1b372e280'
+sha256sums=('aa8f8e49b02643d759bf154142f5e766fb1eaa5e41780a5e8fdaaf845e69e47a'
             'SKIP'
             '5367d8f49aff92884b9daa014502df13e1812f1b7ee1b3a3cb18139f10039408'
             '3107029dfb807e338d34641d78329cd6725c58e6b873352621f4b9611a8380bf'


Sonic-Amiga:
By the way, since you're updating the library, you can also fix one more problem, if you want. If $TERM variable is not empty, it discards the API driver and tries to use terminfo and control codes. But it fails to load the terminfo because default path appears to be not translated to Windows. I. e. it says "/usr/share/terminfo" instead of "C:/MSYS2/mingw64/share/terminfo". I see there's a 002-ncurses-config-win-paths.patch exactly for this purpose, but it's missing TERMINFO_DIR, or how it's called, i forgot.

Symptom of this problem: when $TERM is not empty, applications refuse to run with "Failed to open terminal".

Note, however, that this is not the only problem with this. If you fix it, you'll see the next issue. Any attempt to use control codes ($TERM is set to a proper name) will end up in garbled screen. I don't know the root cause. I remember that couple years ago, back in Windows 7 time, i had one program at work, which also used ncurses, and i ported it to Windows. And everything worked fine in both modes. But WinPTY support in MinTTY wasn't implemented yet; so mintty could only use control codes; and only native cmd.exe could use API driver. I know now things have changed on both sides; perhaps something conflicts.

I am thinking of making another patch to at least work around this problem. It's possible to detect WinPTY without relying on $TERM.

mmuetzel:
Thank you for the feedback.

You seem to know already pretty well how to reproduce the issues that you are describing, and you have a good idea how to fix them. Feel free to open a PR with the necessary changes. 👍
It would be ideal if the fix for the second issue could (also) be sent upstream. It sounds like other projects might profit from it as well. 
The relocation issue (or path translation?) on the other hand might be something that upstream might not want to care about. But a patch for MSYS2 would be appreciated.

If you don't mind, I'd like to merge this patch in its current form because it already fixes the issue tackled in #17494. The other changes could be made in follow-up PRs if you like. Feel free to ping me on those PRs.
Please, let me know if you think the other issues should be tackled before this is merged. (If no one opposes, I'll merge probably tomorrow.)

Sonic-Amiga:
Of course, let's merge. The second problem is a bit more complex and more experimental.

Sonic-Amiga:
... i can post path fix for your reference, but IMHO hard error fills better than garbled screen. So perhaps should be held.

mmuetzel:
Ah. I think I got your point now: Currently ncurses crashes. But with the recent changes it no longer crashes, but garbled text is printed because of another bug.
Did I understand correctly? 

I agree that not merging is probably the better option in this case.

Closing this PR.

Sonic-Amiga:
No! You're wrong.

Sonic-Amiga:
This PR as it is would fix backspace and not do any harm. 
 I was suggesting more changes on top of this, which would have such an effect.

mmuetzel:
Thanks for clarifying. From your previous comments, it wasn't clear to me which change should be held because of which reason...

Merging now.

Sonic-Amiga:
Sorry for thee confusion. Perhaps too much of a brain dump.

PR #17761, opened 2023-07-12
glib2: add cflags.private to all .pc files

lazka:
by hackily extending extra_cflags (stolen from vlc)


lazka:
too hacky imo, let's patch it in PKGBUILD instead

PR #17764, opened 2023-07-12
librest: Update to 0.9.1

lazka:


lazka:
if the libsoup change is causing any problems, let us know. It's hard to tell when we should make the switch.

PR #17767, opened 2023-07-13
capstone: update to 5.0

WerWolv:


mingw-w64-capstone/PKGBUILD
```diff
@@ -15,10 +15,10 @@ license=('BSD')
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
 makedepends=("${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-cc"
-             "${MINGW_PACKAGE_PREFIX}-python-setuptools")
-source=("${_realname}-${_pkgver}.tar.gz::https://github.com/aquynh/capstone/archive/${_pkgver}.tar.gz"
+             "${MINGW_PACKAGE_PREFIX}-python-pip")
```
Biswa96, line 18: Why pip?

WerWolv: Previously to install the python bindings, setup.py was called directly which is apparently deprecated now. The recommended way now is to use pip

lazka: We use build/installer instead of pip, please see other python packages, like https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-faker/PKGBUILD

WerWolv: I honestly have trouble figuring out how these commands work. Every documentation I've found on the subject just recommended to use pip instead (including capstone's readme).

`pip` is doing everything correctly but `build` somehow tries to build the whole library again and fails while doing it
```
CMake Error at CMakeLists.txt:24 (project):
  Running

   'nmake' '-?'

  failed with:

   The system cannot find the file specified


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
The system cannot find the file specified
CMake Error: Generator: execution of make failed. Make command was: nmake -f Makefile /nologo &&
error: [Errno 2] No such file or directory: 'libcapstone.dll'
```

This is using pretty much the exact same code as all the other python packages do
```
  cd "${srcdir}"
  cp -r "${_realname}-${pkgver}" "python-build-${MSYSTEM}" && cd "python-build-${MSYSTEM}/bindings/python"
  ${MINGW_PREFIX}/bin/python -m build --wheel --skip-dependency-check --no-isolation
```

lazka: Ok thanks for trying, I'll have a look later

Biswa96:
The pip command did not work. If you look into the `[file-diff] mingw-w64-x86_64-python-capstone` section in CI log, it removed all the files in python-capstone package. I have created a separate pull request to update the capstone package. I shall create a separate pull request to use pep517 in python package.

mingw-w64-capstone/PKGBUILD
Biswa96: >  but build somehow tries to build the whole library again and fails while doing it

Because you probably did not add `LIBCAPSTONE_PATH` variable. So, it tries to rebuild the DLL again instead of using that path.

WerWolv:
Sorry, I usually don't do much with python. Thanks for fixing it ^^

PR #17768, opened 2023-07-13
angleproject: try to fix the build with clangarm64

lazka:
mingw-w64 doesn't have _ReadStatusReg, so use inline asm


lazka:
@Biswa96 something like this?

Biswa96:
I do not know anything about assembly, it was just my guess 😓

PR #17773, opened 2023-07-15
libpaper: fix relocation

lazka:
gnulib assumes the path format matches what we pass to configure, so make sure we only pass unix paths to the gnulib relocation functions.


naveen521kk:
This doesn't seem to fix the texlive issue. Now it returns a slightly different error message
```
xdvipdfmx:fatal: Unrecognized paper format: # Simply write the paper name. See papersize(5) for possible values
```

naveen521kk:
Also, running `xpdf` which also depends on `libpaper` just segfaults after this update (previously it just issued a warning for missing paper sizes).

lazka:
Meh, ok, thanks for testing

naveen521kk:
Thanks! looks like it works.

lazka:
nice :)

sorry for the breakage

PR #17777, opened 2023-07-15
Rust: disable mingw packaging via configuration option

mati865:
One less downstream patch to maintain (once the fix reaches the release....).

Upstream PR: https://github.com/rust-lang/rust/pull/113725


lazka:
thanks!

lazka:
Looks like this broke things: https://github.com/msys2/msys2-tests/actions/runs/5766194049/job/15633657127

lazka:
Reverted in 1717c9e1f31701 for now

mati865:
Sorry, looks like we cannot get rid of `0008-disable-self-contained.patch`.
I must have tested intermediate version with `rm -f dest-rust/${MINGW_PREFIX}/lib/rustlib/$OSTYPE/lib/self-contained/*` commented out.

PR #17792, opened 2023-07-18
Some cython rebuilds

lazka:


lazka:
I'd say we move the cython 3 transition to after the python 3.11 rebuilds. Otherwise we might end up with lots of non-building packages.

PR #17798, opened 2023-07-19
Switch default printf attribute to gnu_printf for ucrt64

TheShermanTanker:
This patch changes the default printf attribute to be gnu_printf on ucrt64, so that code using the generic printf attribute doesn't get routed to ms_printf instead


lazka:
Wouldn't that break things like?

```c
// gcc -Wall a.c

#include <stdio.h>
#include <stdarg.h>

int my_printf(const char* format, ...) __attribute__((format(gnu_printf, 1, 2)));

int my_printf(const char* format, ...) {
    va_list args;
    va_start(args, format);
    int result = vprintf(format, args);
    va_end(args);
    return result;
}

int main() {
    unsigned __int64 foo = 42;

    my_printf("%I64u", foo);

    return 0;
}
```

TheShermanTanker:
Not sure why the declaration in the example has the format set to gnu_printf? That would make it not compatible with the %I64u with or without this change

lazka:
I was trying to replicate a printf which defaults to gnu_printf as your PR title suggests.

TheShermanTanker:
Oh, my bad. However, when linking to ucrt we really shouldn't be using ms_printf any longer, nor encouraging code to use it and the horrendously outdated msvcrt format specifiers to avoid the warning. All of our headers already try (and in the case of strftime, fail, even if gnu_printf was explicitly specified, so this is actually also a bugfix) to use gnu_printf as the format when on ucrt64, and naively one would also expect the generic printf format to use gnu_printf when putting that in the format attribute (if all the google searches of confused people wondering why printf doesn't recognize standard format specifiers is indicative). In my opinion we should really switch it to enable use of Standard C format specifiers, when on ucrt they are the preferred ones as well. There is still the regular mingw64 system for code that needs to be compiled with ms_printf as default

TheShermanTanker:
gnu_printf supports all the Standard C and C++ format specifiers past C99 and C++98, ms_printf does not, making it impossible to write crossplatform code, or use standard format specifiers such as %T and %z, which we should especially allow on ucrt64 since it supports all those well established format specifiers

TheShermanTanker:
@lazka Bumping, hehe

TheShermanTanker:
Bumping

lazka:
If ms_printf does not match the printf provided my ucrt then it needs to be fixed. This comes to mind https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c51f1e7427e6a5ae2a6d82b5a790df77a3adc99a

TheShermanTanker:
That commit unfortunately doesn't have any effect on this issue, since the default explicit printf attribute will still default to ms_printf. The linked commit fixes gnu_printf being explicitly specified, but not the default printf

lazka:
that commit extended ms_printf to match the real implementation in more cases, from what I see.

PR #17800, opened 2023-07-19
cargo-local-registry: update windows-targets crate

jeremyd2019:
for windows-gnullvm fix.  Less hacky than the previous sed approach used
in other packages before 0.48.1 was published.


jeremyd2019:
Successful ARM64 CI run: https://github.com/msys2-arm/MINGW-packages/actions/runs/5595424409/jobs/10231284974

PR #17806, opened 2023-07-19
mesa: Split opencl clover into separate package

Biswa96:


Biswa96:
@pal1000 Would you like to review this pull request please? The primary goal driving this change is to decrease dependency and package size.

pal1000:
clang and opencl-icd must remain runtime dependencies because they are both linked dynamically by  both mesa mclc and clover.

mingw-w64-mesa/PKGBUILD
```diff
@@ -2,33 +2,34 @@
 
 _realname=mesa
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-opencl-clover-mesa")
```
pal1000, line 6: "${MINGW_PACKAGE_PREFIX}-opencl-clover-${_realname}" would work as well.

mmuetzel:
> clang and opencl-icd must remain runtime dependencies because they are both linked dynamically by both mesa mclc and clover.

Afaict, they are still runtime dependencies of that split package in the updated rules:
https://github.com/msys2/MINGW-packages/pull/17806/files#diff-ba4e3147bf8bbef34ce853a6b5e4d04d7ad898a8d8a7d4e1cb3eedafab7f11aeR190

Biswa96:
> clang and opencl-icd must remain runtime dependencies because they are both linked dynamically by both mesa mclc and clover.

Understood. Then this package spilt would not reduce the dependencies which I thought previously.

mmuetzel:
> Understood. Then this package spilt would not reduce the dependencies which I thought previously.

But it does: Both mesa mclc and clover are split into the package `"${MINGW_PACKAGE_PREFIX}-opencl-clover-mesa"` by your change. Only that package will depend on those packages. The remainder will no longer depend on them.
I understood that was your point. And afaict, your changes reached that point...

Biswa96:
> your changes reached that point...

My change was incorrect. As @pal1000 pointed that clang and opencl-icd are required for both mesa and opencl-clover-mesa. But I put clang and opencl-icd as dependencies for opencl-clover-mesa only.

mmuetzel:
`libclang-cpp.dll` is only required as a dependency for `MesaOpenCL.dll`. For the remainder, `libLLVM-16.dll` should be enough. (At least according to the runtime dependency analysis in the CI.)

Are there other binaries the CI misses somehow?

PR #17821, opened 2023-07-21
qemu: Update to 8.1.0

heljkon:
see https://wiki.qemu.org/Planning/8.1 for further information

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -6,12 +6,12 @@ pkgname=(
   "${MINGW_PACKAGE_PREFIX}-qemu-guest-agent"
   "${MINGW_PACKAGE_PREFIX}-qemu-image-util"
 )
-_base_ver="8.0.3"
+_base_ver="8.1.0"
 # QEMU Versioning of RC-SourcePackage and RC-Version differs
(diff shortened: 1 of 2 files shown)


heljkon:
I've expected failures because of major changes qemu applied to the build system, see https://wiki.qemu.org/ChangeLog/8.1#Build_Information

I'll go for it later.

heljkon:
I've created an issue for the clang build failures: https://gitlab.com/qemu-project/qemu/-/issues/1782
Currently the build for clang succeeds, if dbus is disabled: 
`./configure --disable-dbus-display`

heljkon:
@lazka @Biswa96 @jeremyd2019 
Please have a look!

Biswa96:
qemu install libftd related files which are provided by dtc package also. Should those be removed from qemu package?

lazka:
> Library fdt found: NO

maybe it's just missing a dependency on https://packages.msys2.org/base/mingw-w64-dtc ?

lazka:
In case it helps, I've updated dtc to the newest version in #18274

heljkon:
Adding the already existing package dtc as dependency solved the duplicate problems.

Biswa96:
Thank you.

PR #17825, opened 2023-07-21
Update the arm-none-eabi-toolchain to gcc 13.1

redXhammer:
binutils: 2.40
gcc: 13.1.0
gdb: 13.2
newlib: 4.3.0.20230120 (with gcc 13.1.0)


redXhammer:
I've compiled MINGW64 using the .ci/ci-build.sh script on my personal computer (8 cores) and it succeeded after approximately 5 hours. How can we extend the 6 hours limitation of github?

CLang64 is failing while compiling newlib, I'm working on it to see how to fix it. Maybe we need a patch for the "-r" issue.

mmuetzel:
For what it's worth, here the part of the log with the error messages (because GitHub makes it difficult to scroll to the end of the logs - at least for me):
```
  clang++  -fno-PIE -r -nostdlib utf8rc-mingw32.o sym-mingw32.o -o utf8-mingw32.o
  1 warning generated.
  lld: error: unknown argument: -r
  lld: error: unknown argument: -r
  make[1]: *** [../../gcc/config/i386/x-mingw32-utf8:50: utf8-mingw32.o] Error 1
  make[1]: *** Waiting for unfinished jobs....
```

Afaict, there is no way to increase the timeout of the GitHub hosted runners to longer than 6 hours.
To reduce the time required for the test build, could we update in chunks? E.g., only binutils first? Or would that already require a rebuild of the remaining packages?

MehdiChinoune:
binutils was updated to 2.41
newlib was updated to 4.3.0
gcc couldn't be updated to 13.2.0 as It fails to build on clang environments.

PR #17827, opened 2023-07-21
alacritty: new package

ognevnydemon:
Alacritty - a cross-platform, OpenGL terminal emulator.

Successfully builds, but I can't test it now.
@jeremyd2019 any thoughts on CLANGARM64 compatibility?

### DIFF
--- a/mingw-w64-alacritty/PKGBUILD
+++ b/mingw-w64-alacritty/PKGBUILD
@@ -1,0 +1,69 @@
+# Maintainer: Maksim Bondarenkov <maksapple2306@gmail.com>
+
+_realname=alacritty
+pkgbase=mingw-w64-${realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.12.2
+pkgrel=1
+pkgdesc="A cross-platform, OpenGL terminal emulator (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang32' 'clang64' 'clangarm64')
+url="https://github.com/alacritty/alacritty"
+license=('spdx:Apache-2.0 OR MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ncurses"
+             "${MINGW_PACKAGE_PREFIX}-desktop-file-utils"
+             "${MINGW_PACKAGE_PREFIX}-gdb"
+             'git')
+depends=("${MINGW_PACKAGE_PREFIX}-freetype"
+         "${MINGW_PACKAGE_PREFIX}-fontconfig")
+checkdepends=("${MINGW_PACKAGE_PREFIX}-ttf-dejavu")
+optdepends=("${MINGW_PACKAGE_PREFIX}-ncurses: for alacritty terminfo database")
+source=("git+${url}.git#tag=v${pkgver}?signed")
+validpgpkeys=('4DAA67A9EA8B91FCC15B699C85CDAE3C164BA7B4'  # Christian Dürr <contact@christianduerr.com>
+              'A56EF308A9F1256C25ACA3807EA8F8B94622A6A9') # Kirill Chibisov <contact@kchibisov.com>
+sha256sums=('SKIP')
+
+prepare() {
+  cd "${srcdir}/${_realname}"
+
+  if [[ $MINGW_PACKAGE_PREFIX == *-clang-aarch64 ]]; then
+    cargo fetch --locked --target aarch64-pc-windows-gnullvm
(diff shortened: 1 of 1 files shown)


ognevnydemon:
why check() is [skipped](https://github.com/msys2/MINGW-packages/actions/runs/5625465360/job/15244587292)?

Biswa96:
Checks are disabled for CI because those take too much time or failed most of the time. See the `--nocheck` option in `.ci/ci-build.sh` file.

ognevnydemon:
are there any plans to enable it?

Biswa96:
> are there any plans to enable it?

Nope. It is assumed that package maintainers/contributors did some rudimentary tests.

mingw-w64-alacritty/PKGBUILD
```diff
@@ -0,0 +1,69 @@
...
+              'A56EF308A9F1256C25ACA3807EA8F8B94622A6A9') # Kirill Chibisov <contact@kchibisov.com>
+sha256sums=('SKIP')
+
+prepare() {
+  cd "${srcdir}/${_realname}"
+
+  if [[ $MINGW_PACKAGE_PREFIX == *-clang-aarch64 ]]; then
+    cargo fetch --locked --target aarch64-pc-windows-gnullvm
+  else
+    cargo fetch --locked --target "${CARCH}-pc-windows-gnu"
```
jeremyd2019, line 34: I don't remember any other package needing to specify a target to cargo like this.  Why is this necessary?

jeremyd2019:
> Successfully builds, but I can't test it now. @jeremyd2019 any thoughts on CLANGARM64 compatibility?

Do you just want me to test it?  Actually running opengl things on arm64 is kind of interesting (needs a compatibility store app thing from Microsoft).

mingw-w64-alacritty/PKGBUILD
ognevnydemon: I find that it uses lots of unneeded dependencies (~230), after this this number has been greatly reduced (~110)

ognevnydemon:
> > Successfully builds, but I can't test it now. @jeremyd2019 any thoughts on CLANGARM64 compatibility?
> 
> Do you just want me to test it?  Actually running opengl things on arm64 is kind of interesting (needs a compatibility store app thing from Microsoft). 

yes, why not? I don't have machine for this, so how can I know?

mingw-w64-alacritty/PKGBUILD
```diff
@@ -0,0 +1,69 @@
...
+pkgdesc="A cross-platform, OpenGL terminal emulator (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang32' 'clang64' 'clangarm64')
+url="https://github.com/alacritty/alacritty"
+license=('spdx:Apache-2.0 OR MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ncurses"
+             "${MINGW_PACKAGE_PREFIX}-desktop-file-utils"
+             "${MINGW_PACKAGE_PREFIX}-gdb"
```
jeremyd2019, line 17: Why is gdb needed to build?  There is no gdb package for arm64, so it fails here right now.

ognevnydemon: hmm, didn't find gdb in upstream, but only in arch's package. maybe we can try to remove it

jeremyd2019:
bunch of errors building on arm64:
```
  error[E0412]: cannot find type `ULONG_PTR` in this scope
    --> C:\Users\Administrator\.cargo\registry\src\index.crates.io-6f17d22bba15001f\winapi-0.2.8\src\basetsd.rs:86:19
     |
  86 | pub type SIZE_T = ULONG_PTR;
     |                   ^^^^^^^^^ not found in this scope
  
```

and a bunch of similar in winapi-0.2.8, having to do with pointer-sized integral types.  perhaps that version is too old for windows on arm64 support?

ognevnydemon:
> perhaps that version is too old for windows on arm64 support?

maybe. it seems that upstream updated deps, but hasn't released this yet

jeremyd2019:
Yes, winapi 0.2 has only x86 and x86_64 definitions of sized types in https://github.com/retep998/winapi-rs/blob/0.2/src/basetsd.rs.  Whereas 0.3 got arm support (both arm64 and arm32), and https://github.com/retep998/winapi-rs/blob/0.3/src/shared/basetsd.rs refers instead to `target_pointer_width` cfg attr.  Unfortunately, alacritty master Cargo.lock still refers to winapi 0.2.8, apparently via mio 0.6.23?

ognevnydemon:
I opened issue in upstream and I received this answer: https://github.com/alacritty/alacritty/issues/7104#issuecomment-1646640836

PR #17828, opened 2023-07-21
sundials: update to 6.6.0

mmuetzel:


mmuetzel:
I've installed the MINGW64 build artifacts from this PR. With them, Octave's test suite still passes.
According to the package grokker results, the other depender of sundials (mfem) doesn't seem to be affected by the version suffix change of some of the built libraries.

Biswa96:
Does octave depend on ghostscript? Arch's octave depends on ghostscript but the CI log here does not show libgs-*.dll in runtime deps list.

mmuetzel:
> Does octave depend on ghostscript? Arch's octave depends on ghostscript but the CI log here does not show libgs-*.dll in runtime deps list.

As far as I can tell, Octave doesn't link to any ghostscript libraries. It's using it in some printing toolchains if it can find it though. I.e., it calls `gs.exe` on Windows if available.

In case you are asking if Octave needs to be rebuilt if GhostScript is updated: That shouldn't be necessary. If command line switches changed between versions, Octave's code would need to be adapted to that change. Rebuilding alone wouldn't help...

PR #17829, opened 2023-07-21
Test rebuilding gdk-pixbuf2

jeroen:
Does not build for me locally anymore. Triggering a CI to reproduce the problem...

### DIFF
--- a/mingw-w64-gdk-pixbuf2/PKGBUILD
+++ b/mingw-w64-gdk-pixbuf2/PKGBUILD
@@ -7,7 +7,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-docs")
 pkgver=2.42.10
-pkgrel=3
+pkgrel=4
 pkgdesc="An image loading library (mingw-w64)"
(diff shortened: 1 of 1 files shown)


Biswa96:
Hi @jeroen, Thank you for reporting the issue. I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request?

jeroen:
Thank you, this fixes my issue!

One related question:

I am trying to [build a version](https://github.com/jeroen/MINGW-packages/commit/1683f57ae5be5f37bd682f2fed9326fc293d3fc1) locally with `-Dbuiltin_loaders=all` that avoids the dynamic loaders alltogether. This fails with linking errors for libtiff, libpng, libpjeg: https://github.com/jeroen/MINGW-packages/actions/runs/5629734874/job/15255024264

How would I fix this? Should I add the same entries in meson.build as you did in https://github.com/msys2/MINGW-packages/pull/17830/files but then for  libtiff, libpng, libpjeg ?

Biswa96:
> Should I add the same entries in meson.build

Yes. All the dependencies are aggregated in `loaders_deps` variable in meson.build file. You can add `loaders_deps` in `dependencies` list also.

But, I kind of expected that `-Dbuiltin_loaders=all` option should add the dependencies automatically. 🤔

PR #17831, opened 2023-07-22
starship: update to 1.16.0

ognevnydemon:
don't know why, but it fails on i686. should we disable it?

upd.: i686 is disabled, but it will be fixed soon. see further conversation for details

### DIFF
--- a/mingw-w64-starship/PKGBUILD
+++ b/mingw-w64-starship/PKGBUILD
@@ -23,8 +23,6 @@ noextract=("${_realname}-${pkgver}.tar.gz")
 prepare() {
   tar -xzf "${srcdir}"/${_realname}-${pkgver}.tar.gz -C ${srcdir} || true
   cd "${srcdir}/${_realname}-${pkgver}"
-  # update windows-targets to fix windows-gnullvm dependency specification
-  "${MINGW_PREFIX}/bin/cargo.exe" update -p windows-targets@0.48.0 --precise 0.48.1
   "${MINGW_PREFIX}/bin/cargo.exe" fetch --locked
 }
 
--- a/mingw-w64-starship/PKGBUILD
+++ b/mingw-w64-starship/PKGBUILD
@@ -7,7 +7,9 @@ pkgver=1.15.0
 pkgrel=1
 pkgdesc="The cross-shell prompt for astronauts (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
+# i686 is disabled, see
+# https://github.com/msys2/MINGW-packages/pull/17831#issuecomment-1648392525
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://starship.rs"
 license=('spdx:ISC')
 depends=("${MINGW_PACKAGE_PREFIX}-libgit2")
--- a/mingw-w64-starship/PKGBUILD
+++ b/mingw-w64-starship/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=starship
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.11.0
-pkgrel=2
+pkgver=1.15.0
+pkgrel=1
 pkgdesc="The cross-shell prompt for astronauts (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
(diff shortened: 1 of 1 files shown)


ognevnydemon:
> don't know why, but it fails on i686. should we disable it?

@jeremyd2019 any ideas?

jeremyd2019:
Guess: sha1-asm crate isn't properly mangling asm names for x86 windows? https://github.com/RustCrypto/asm-hashes/blob/master/sha1/src/x86.S  should have leading underscore like the apple case I think.  I might find some time to hack at it tomorrow.

jeremyd2019:
@mati865 are you familiar with sha1-asm crate?  I get the impression it isn't really expected to be used on Windows, as I see no signs of MSVC support.  While it looks like it should be possible for it to work for -windows-gnu, maybe there's something else going on that is causing it to be used here on Windows instead of something more Windows-specific.

mati865:
I have never used it myself but looks like they could figure out i686 CI so they only support amd64 https://github.com/RustCrypto/asm-hashes/pull/19 ...
The only immediate solution I can think of is building with `--no-default-features` on i686 which will likely render git support as not working because of Cygwin paths.
For future releases I think we need to make gitoxide buildable on i686, it avoids `sha1` on msvc and should do the same for i686-windows-gnu: https://github.com/Byron/gitoxide/blob/146c1498c7de4c8c00b2d1df52795b8338635697/gix-features/Cargo.toml#L155

ognevnydemon:
so our task for now is build gitoxide package?

jeremyd2019:
> @mati865 are you familiar with sha1-asm crate? I get the impression it isn't really expected to be used on Windows, as I see no signs of MSVC support. While it looks like it should be possible for it to work for -windows-gnu, maybe there's something else going on that is causing it to be used here on Windows instead of something more Windows-specific.

```toml
# Assembly doesn't yet compile on MSVC on windows, but does on GNU, see https://github.com/RustCrypto/asm-hashes/issues/17
# At this time, only aarch64, x86 and x86_64 are supported.
[target.'cfg(all(any(target_arch = "aarch64", target_arch = "x86", target_arch = "x86_64"), not(target_env = "msvc")))'.dependencies]
sha1 = { version = "0.10.0", optional = true, features = ["asm"] }
```

jeremyd2019:
I will try to fix build on i686-pc-windows-gnu and open a PR against RustCrypto/asm-hashes, but for now probably the best option is to figure out how to disable the `gix/fast-sha1` feature for i686.  I'm not that familiar with cargo, but there doesn't seem to be an option to turn *off* a feature, so you may need `--no-default-features` and enable all the other features that should be enabled.  Or just drop i686 packages for starship for now.

ognevnydemon:
I decided to disable i686 for now

jeremyd2019:
I still need to test on clangarm64 too.  It looks like at least windows-targets needs to be updated to 0.48.1 (which is at least cleaner than hacking up 0.48.0 with sed)

mati865:
Minimal patch for `gitoxide` that makes it build on i686:
```diff
$ git diff -- gix-features
diff --git a/gix-features/Cargo.toml b/gix-features/Cargo.toml
index e1cd1cdea..b03d44e00 100644
--- a/gix-features/Cargo.toml
+++ b/gix-features/Cargo.toml
@@ -150,9 +150,9 @@ libc = { version = "0.2.119" }
 bstr = { version = "1.3.0", default-features = false }


-# Assembly doesn't yet compile on MSVC on windows, but does on GNU, see https://github.com/RustCrypto/asm-hashes/issues/17
+# Assembly doesn't yet compile on MSVC and x86 GNU on windows, but does on x86_64 GNU, see https://github.com/RustCrypto/asm-hashes/issues/17
 # At this time, only aarch64, x86 and x86_64 are supported.
-[target.'cfg(all(any(target_arch = "aarch64", target_arch = "x86", target_arch = "x86_64"), not(target_env = "msvc")))'.dependencies]
+[target.'cfg(all(any(target_arch = "aarch64", target_arch = "x86", target_arch = "x86_64"), not(any(target_env = "msvc", all(target_arch = "x86", target_env = "gnu")))))'.dependencies]
 sha1 = { version = "0.10.0", optional = true, features = ["asm"] }

 [package.metadata.docs.rs]
```
> I will try to fix build on i686-pc-windows-gnu and open a PR against RustCrypto/asm-hashes

I don't know if they will accept any PR: https://github.com/RustCrypto/asm-hashes/pull/58

jeremyd2019:
CLANGARM64 builds now.  Probably don't want to merge this yet, until python rebuilds are done.

ognevnydemon:
> CLANGARM64 builds now.  Probably don't want to merge this yet, until python rebuilds are done.

python rebuilds are done. can we merge it now?

ognevnydemon:
updated package to 1.16.0 and removed windows-targets workaround

jeremyd2019:
Ugh, 1.16.0 failed on clangarm64 due to not finding `-lwindows`.  It seems now it only uses windows 0.39.0 and 0.48.x.  0.39.0 is too old to have gnullvm in its Cargo.toml, and 0.48.0 now uses a version number in its "windows" lib.  Does this need messing with 0.39.0's Cargo.toml to pull in the MSVC variant for aarch64, like back when aarch64 rust was first built @mati865 ?  ☹️

PR #17838, opened 2023-07-22
darktable: update to 4.4.2

kmilos:


Biswa96:
This removed the libdarktable.dll.a file. Is that expected?

kmilos:
Yes. It was unnecessarily installed in /bin

Issue #17840, opened 2023-07-23
[ffmpeg] The procedure entry point expf could not be located


kasper93:
### Description / Steps to reproduce the issue

``` C
#include <stdio.h>
#include <libavfilter/avfilter.h>

int main() {
    printf("%u\n", avfilter_version());
}
```
```
cc main.c `pkgconf --libs --cflags libavfilter` && ./a.exe
```

Example failing job from mpv CI:
https://github.com/mpv-player/mpv/actions/runs/5636172683/job/15267962388

### Expected behavior

Prints version

### Actual behavior

```
The procedure entry point expf could not be located in the dynamic link library C:\MSYS\clang32\bin\avfilter-9.dll. 
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW32_NT-10.0-22000

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

Biswa96:
Thank you for reporting the issue. I can reproduce the issue in my system but in a weird way. If I run the compiled executable in default clang32/mintty terminal, the program does not show any error. Only the exit code is 127.

If I run the executable from Command Prompt, the program shows the following error window. I have to add the clang32/bin directory in %PATH% e.g. `set PATH=F:\msys64\clang32\bin;%PATH%`

![image](https://github.com/msys2/MINGW-packages/assets/31443074/9aa13a6c-717f-4d2c-8377-5502d572a756)

kasper93:
Yes, those dialogs doesn't show when run from msys2 process, like bash. It is `SEM_NOGPFAULTERRORBOX` which is set by defualt. See https://www.msys2.org/wiki/JIT-Debugging/

Biswa96:
tl;dr simple rebuild of ffmpeg fixes the issue.

Previously, zimg did not export any symbols from import library in clang environments. The zimg static library exports some standard C math functions e.g. expf, log10f, logf, powf etc. Here is the output of nm command with zimg static library before it was fixed.

```
$ nm libzimg.a | grep -i expf
00000000 R .refptr._zimg_x_expf
         U _zimg_x_expf
00000000 T _expf
00000000 D _zimg_x_expf
```

[...]

PR #17842, opened 2023-07-23
ffmpeg: Rebuild for zimg fix

Biswa96:
Peviously, zimg did not export any symbols from import library in clang
    environments. The zimg static library exports some standard C math
    functions e.g. expf, log10f, logf, powf etc.

    So, the avfilter library links the C math function (here expf) with zimg
    static library. After this commit 92d0f0f63f0f42760eae61a71be04d23ab6eb9e6,
    those C math functions were removed from zimg static library. And avfilter
    fails to find expf functions. This rebuild fix that issue.

Fixes https://github.com/msys2/MINGW-packages/issues/17840


kasper93:
ffmpeg4.4 might be also affected in same way. Maybe rebuild it too, just to be safe.

Biswa96:
> ffmpeg4.4 might be also affected in same way. Maybe rebuild it too, just to be safe.

Yeah, ffmpeg 4.4 is also affected. I shall bump the pkgrel in separate commit.

```
$ Dependencies.exe -imports /var/cache/pacman/pkg/avfilter-7.dll | grep -A 10 -i libzimg-2.dll
Import from module libzimg-2.dll :
         Function expf
         Function powf
         Function zimg_filter_graph_build
         Function zimg_filter_graph_free
         Function zimg_filter_graph_get_tmp_size
         Function zimg_filter_graph_process
         Function zimg_get_last_error
         Function zimg_graph_builder_params_default
         Function zimg_image_format_default
Import from module libfontconfig-1.dll :
```

PR #17844, opened 2023-07-23
Add Jolt-Physics package

ImperatorS79:
What is the policy of msys2 regarding vector extensions ? What minimum processor (and thus extensions)  do you expect ?


ImperatorS79:
There is lots of symbol missing in the .a, so I have to investigate a bit.

ImperatorS79:
Ok it is working know.

lazka:
> What is the policy of msys2 regarding vector extensions ? What minimum processor (and thus extensions) do you expect ?

In theory we target the same CPUs as Windows 10 (see https://www.msys2.org/news/#2022-10-18-new-minimum-hardware-requirements-cpus-from-20067) but in reality there are probably various special case packages like video codecs, math/scientific libraries requiring more. It would be nice to have some kind of rule for this, or to know how other distros handle this.

PR #17845, opened 2023-07-23
Assimp: fix wrong version in .pc file

ImperatorS79:
I don't know why, but the .pc file says version 5.2.4.


Biswa96:
I think this should be fixed in the code instead of using sed command.

ImperatorS79:
> I think this should be fixed in the code instead of using sed command.

Yes, this is a quick fix. But I did not have time to check where is the problem in the code.

ImperatorS79:
> The project version was fixed in this upstream commit [assimp/assimp@8ce6f5a](https://github.com/assimp/assimp/commit/8ce6f5ad272cb42118a70d19668e15c1dd1615e7)
> 
> Is it possible to import that patch in PKGBUILD ?

I am doing the patch manually, as I don't know how to import it. It will be the same results at the end.

Biswa96:
Thank you for the fix. I have verified this fixes the version number in pkgconfig file.

PR #17847, opened 2023-07-23
Update GTK3

lillolollo:
A bug that affect all the popup windows is fixed https://gitlab.gnome.org/GNOME/gtk/-/issues/5974 Since gimp is releasing a new version this avoid crash in multiple tools


Biswa96:
The version number is not correct. pkgver variable is automatically set by the pkgver function while building the package.

lillolollo:
@Biswa96 @lazka Thanks :-)

PR #17848, opened 2023-07-24
aws: update packages

Biswa96:


Biswa96:
Some executables are removed after this update because tests are disabled. Those are just part of sample and test binaries.

PR #17856, opened 2023-07-25
Remove sip4 and wxPython4.0

lazka:
It no longer builds with Python 3.11, and while there might be a way to patch it, there are no users left in MSYS2, so maybe it's time to drop it?


lazka:
@stahta01 any thoughts on this? if you think there is value in keeping it we can have a look.

stahta01:
Removing it seems reasonable to me. KiCAD was what needed it in the past. As long as KiCAD still builds and runs without it I see no reason to not remove those two packages.

PR #17857, opened 2023-07-25
arm-none-eabi-gdb: Fix rebuilding with readline 8.2 or newer.

mmuetzel:


mmuetzel:
Two of the necessary patches from upstream didn't apply as-is. One because some files have been moved to a different directory in newer versions. And the other because the parts concerning changelog files didn't apply.
I added modified versions of the upstream patches. But left comments with the links to the original patches.

I felt a bit uncomfortable to update to a new major version because I don't know how to test this. I hope cherry-picking a minimal amount of patches to allow rebuilding with current tools is ok.

mmuetzel:
Reopening because it looks like #17825 fails with a linker error on CLANG64.

This PR could work as an interim fix for the Python rebuilds.
Feel free to close when the other PR is ready in time.

Issue #6191, opened 2020-02-13
Does not support inet_pton on MinGW?


zxgm:
Does not support inet_pton on MinGW?

lazka:
it should be supported: https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inet_pton

Biswa96:
It's not in mingw-w64 repo, working...

lazka:
https://github.com/mirror/mingw-w64/blob/1b5ae4b567ee1699f69cd918ae2bc693898edbb2/mingw-w64-headers/include/ws2tcpip.h#L455

Biswa96:
But the actual function declaration isn't present.

Biswa96:
@zxgm Can you provide any compiler error message or any minimal code to reproduce this issue?

zxgm:
Thank you very much.I have found the reason,the macro _WIN32_WINNT value is 0x502 that was defined  in _mingw.h,so inet_pton is disabled in ws2tcpip.h.I redefine the macro _WIN32_WINNT value 0x0600,that is work.

zxgm:
Now,I hit the other problems.The code is follow:

> setOption(IPPROTO_TCP, TCP_KEEPIDLE, &idle, sizeof(idle))
but compile the code use mingw,that error:

> error: 'TCP_KEEPIDLE' was not declared in this scope
On windows, the macro TCP_KEEPIDLE defined in ws2ipdef.h,but the same header file in MinGW,that have not TCP_KEEPIDLE,where was the macro TCP_KEEPIDLE  defined in MinGW?

lazka:
Doesn't look like mingw-w64 has it. https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-tcp-socket-options states that it is only supported since Windows 10, which could explain it.

zxgm:
Thanks a lot.I think mingw-w64 not support the macro 'TCP_KEEPIDLE' so far.

Biswa96:
I may add that in a PR...

PR #17858, opened 2023-07-25
postgresql: update to 15.3

mmuetzel:
Fixes errors when rebuilding with Clang 16.

### DIFF
--- a/mingw-w64-postgresql/PKGBUILD
+++ b/mingw-w64-postgresql/PKGBUILD
@@ -9,7 +9,7 @@ pkgdesc="Libraries for use with PostgreSQL (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.postgresql.org/"
-license=('custom:PostgreSQL')
+license=('spdx:PostgreSQL')
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc" "${MINGW_PACKAGE_PREFIX}-autotools")
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-gettext"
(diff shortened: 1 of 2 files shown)


mingw-w64-postgresql/PKGBUILD
```diff
@@ -65,8 +65,13 @@ prepare() {
     postgresql-14.0-use-mingw-setjmp-on-ucrt.patch
   #patch -p1 -i ${srcdir}/postgresql-9.4.1-mingw-enable-readline.patch
 
-  sed -s "s|2\\.69|2\\.71|g" -i configure.ac
   autoreconf -fiv
+
+  # Clang 16 requires strict C99 compatibility by default.
+  # `inet_pton` is not declared in src/interfaces/libpq/fe-secure-common.c but can be used.
+  # Disable the corresponding error as a workaround.
+  sed -i 's/^CFLAGS = .*/& -Wno-implicit-function-declaration/' src/Makefile.global.in
```
Biswa96, line 73: Instead of disabling the error, could you check if changing the MIN_WINNT in src/include/port/win32.h from 0x0501 to 0x0600 helps?

mmuetzel: Thanks for the suggestion. That seems to working, too, for me locally.
I updated the PR with a patch and removed that workaround.

Biswa96:
Just for reference. The compiler error was as following.

```
fe-secure-common.c:219:7: error: call to undeclared function 'inet_pton'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    if (inet_pton(AF_INET6, host, &addr) == 1)
        ^

fe-secure-openssl.c:560:7: error: call to undeclared function 'inet_pton'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    || (inet_pton(AF_INET6, host, &dummy6) == 1)
        ^
```

This same issue was brought up in this mingw-w64 thread https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/777490f2-46ed-9f39-b6c2-040a001020f0@126.com/ where OP suggested to bump the _WIN32_WINNT value.

Also, another similar issue https://github.com/msys2/MINGW-packages/issues/6191

PR #17859, opened 2023-07-25
mypaint: fix compatibility with Python 3.11

mmuetzel:
"U" as the mode for universal line endings has been deprecated for a while and has been removed in Python 3.11. Universal line endings are the default now.

The CLANG* environments are failing currently (because numpy hasn't been rebuilt yet).
But this should fix the build error at least with GCC.

### DIFF
--- a/mingw-w64-mypaint/PKGBUILD
+++ b/mingw-w64-mypaint/PKGBUILD
@@ -26,16 +26,29 @@ depends=("${MINGW_PACKAGE_PREFIX}-adwaita-icon-theme"
 makedepends=("${MINGW_PACKAGE_PREFIX}-pkg-config"
              "${MINGW_PACKAGE_PREFIX}-python-setuptools"
              "${MINGW_PACKAGE_PREFIX}-swig"
-             "${MINGW_PACKAGE_PREFIX}-cc")
+             "${MINGW_PACKAGE_PREFIX}-cc"
+             $([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-gcc-compat"))
 options=('strip' '!debug' 'staticlibs')
-license=("GPL2" 'spdx:ISC')
+license=('spdx:GPL-2.0-or-later' 'spdx:ISC')
 url="http://mypaint.org"
-source=(https://github.com/mypaint/mypaint/releases/download/v${pkgver}/mypaint-${pkgver}.tar.xz)
-sha256sums=('f3e437d7cdd5fd28ef6532e8ab6b4b05d842bcdd644f16a0162dad3d8e57bb16')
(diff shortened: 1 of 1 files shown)


Biswa96:
Just for reference. This was fixed in upstream https://github.com/mypaint/mypaint/commit/032a155b72f2b021f66a994050d83f07342d04af

ognevnydemon:
> Just for reference. This was fixed in upstream https://github.com/mypaint/mypaint/commit/032a155b72f2b021f66a994050d83f07342d04af

but latest release was 3 years ago...

mmuetzel:
Thanks for the hint. I updated the PR to prefer the patch from upstream over a home-baked one.

mmuetzel:
Oops. Forgot to bump pkgrel.

Could someone delete the failed package from earlier in staging please?

mingw-w64-mypaint/PKGBUILD
```diff
@@ -26,16 +26,29 @@ depends=("${MINGW_PACKAGE_PREFIX}-adwaita-icon-theme"
 makedepends=("${MINGW_PACKAGE_PREFIX}-pkg-config"
              "${MINGW_PACKAGE_PREFIX}-python-setuptools"
              "${MINGW_PACKAGE_PREFIX}-swig"
-             "${MINGW_PACKAGE_PREFIX}-cc")
+             "${MINGW_PACKAGE_PREFIX}-cc"
+             $([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-gcc-compat"))
```
Biswa96, line 30: Forgot to ask, why gcc-compat was used here?

lazka:
> Could someone delete the failed package from earlier in staging please?

done

mingw-w64-mypaint/PKGBUILD
mmuetzel: Without it, compilation failed for me because it was trying to use `gcc` from the MSYS2 environment.

Biswa96: gcc is not installed in msys2 by-default.

mmuetzel: It was for me locally. Is having that build dependency an issue?

Biswa96: Not an issue. But gcc-compat is added when there is no other way to fix compiler detection in build system. In case of python, $CC and $CXX is detected most of the time.

mmuetzel: Aren't those set by `makepkg-mingw` by default?

Biswa96: Yup.

mmuetzel: Then, I didn't understand what you meant.

PR #17860, opened 2023-07-25
pygments: disable docs temporarily during rebuilds

jeremyd2019:
This should be reverted once jinja is built


Biswa96:
A `_bootstrap` flag might be useful in future (like we did in other PKGBUILDs).

PR #17871, opened 2023-07-26
python-bsddb3: update to 18.1.6

mmuetzel:
This is an attempt to fix building python-bsddb3.

IIUC, the current version already has a mismatch between the prototype of function pointers that are passed to the Berkeley DB and the actual prototype of those callback functions. So, I guess it would already likely crash when trying to invoke those function pointers.
Clang 16 seems to have better analysis for that issue. So, it already fails to compile instead of crashing at runtime.
That issue seems to be fixed in the latest version of python-bsddb3. 
However, that version is no longer compatible to the version of the Berkeley DB that we currently have. So, I updated to the same version that Arch Linux ships.

With those two packages updated, compilation of the package still fails on CLANG64 for me. For some reason, a `-R` flag gets inserted to the linker flags that `lld` doesn't understand.
To be honest: I have no idea where that `-R` flag comes from.

Does someone have an idea how to fix that?


Biswa96:
The project name was changed to https://pypi.org/project/berkeleydb/. It seems that berkeleydb and bsddb3 are internally different. Here is a snippet from setup3.py in bsddb3.

> Take note that upgrading to [berkeleydb](https://pypi.org/project/berkeleydb/) is easy but not transparent. Notably, keys and values are bytes in [berkeleydb](https://pypi.org/project/berkeleydb/) lib, while in ‘bsddb3’ they are strings. You would only need to change your code to add or remove type encoding/decoding, depending of how you use the library. The process should be simple, nevertheless.

jeremyd2019:
IIRC, I think there was a license change in db, which may be the reason it was still at the older version.

mmuetzel:
This is turning out to be more complicated than I expected.

Looking at the code that Clang is failing to compile now, I'd guess that the version that we are currently distributing (for any environment) is crashing if it tries to execute that code path. I don't know since when this is the case. But given that we haven't heard anything about it yet probably means that there are no users of this package.

So, instead of spending too much time to try and get this fixed, we could also just remove the (broken) package.
If someone needs it, it could still be added again in the future.

What do you think?

Biswa96:
> So, instead of spending too much time to try and get this fixed, we could also just remove the (broken) package.

I agree with that. Looking at the clang error, the number of arguments of that function point is already different. So, the package may be broken already.

mmuetzel:
I opened #17874 that removes this package.

Closing this PR.

Biswa96:
> Looking at the clang error, the number of arguments of that function pointer is already different

For the record, the compiler error was as following.

```
Modules/_bsddb.c:2729:46: error: incompatible function pointer types passing 'int (DB *, const DBT *, const DBT *)' (aka 'int (struct __db *, const struct __db_dbt *, const struct __db_dbt *)') to parameter of type 'int (*)(DB *, const DBT *, const DBT *, size_t *)' (aka 'int (*)(struct __db *, const struct __db_dbt *, const struct __db_dbt *, unsigned long long *)') [-Wincompatible-function-pointer-types]
    err = self->db->set_bt_compare(self->db, _db_compareCallback);
                                             ^~~~~~~~~~~~~~~~~~~
Modules/_bsddb.c:2851:47: error: incompatible function pointer types passing 'int (DB *, const DBT *, const DBT *)' (aka 'int (struct __db *, const struct __db_dbt *, const struct __db_dbt *)') to parameter of type 'int (*)(DB *, const DBT *, const DBT *, size_t *)' (aka 'int (*)(struct __db *, const struct __db_dbt *, const struct __db_dbt *, unsigned long long *)') [-Wincompatible-function-pointer-types]
    err = self->db->set_dup_compare(self->db, _db_dupCompareCallback);
                                              ^~~~~~~~~~~~~~~~~~~~~~
```

hgohel:
@mmuetzel @Biswa96 Dropping by to say that `bsddb3` package is being used by [Gramps software](gramps-project.org/) and as the maintainers attempt to create a new release, lack of availability of this package targeting a newer version of Python is a stumbling block.

Despite the coding issue (which was previously a warning, not error), the package worked as-is. For a proper solution I've reached out to the author. In the short-term however, is it possible to build bsddb3 6.2.9 and/or berkeley 18.1.6 by disabling the stricter setting which is treating this issue an error? See: [-Wincompatible-function-pointer-types](https://clang.llvm.org/docs/DiagnosticsReference.html#wincompatible-function-pointer-types)? Thank you!

Biswa96:
> is it possible to build bsddb3 6.2.9 and/or berkeley 18.1.6 by disabling the stricter setting which is treating this issue an error?

That above compiler error is not just a warning that can be disable. Do you think that a function pointer with different number of parameters should be passed as is?

hgohel:
@Biswa96 First, let me state that my knowledge of bsddb3/berkeleydb is about 30-minutes deep :) I'm learning on the fly by examining this PR and the bsddb3 project pages. That said, could it be that the fourth parameter is included because the conditional around it is being triggered by the first commit in this PR f23def1b7c637ac5747f29e4c16cbed2a57419e6 which upgrades db version to 6.2?

What about another approach? Instead of upgrading db and bsddb3, what if the existing bsddb3 6.2.9 package was rebuilt as a sanity check with the newer toolchain? I imagine clang will require disabling the strict check, but we would avoid the fourth parameter issue being triggered if db version isn't changed. Your thoughts?

Biswa96:
Wouldn't that require lower version of db package compared to current one?

hgohel:
AFAICT, mingw-w64-db is still on db version 6.0.19 ([repo package](https://packages.msys2.org/base/mingw-w64-db)). The upgrade to db 6.2.x in this PR hasn't been merged yet so no downgrade would be required to rebuild bsddb3 6.2.29 and current Python version.

mmuetzel:
My only motivation to have opened this PR was because the current version of bsddb3 doesn't rebuild with current Clang. The reason for that build error has been discussed at length in the comments already.

hgohel:
@mmuetzel Understood. To move things along I've created PR #18057 to restore a working build of bsddb3 6.2.29.

PR #17878, opened 2023-07-27
ogre3d: De-activate Python bindings for MINGW32

mmuetzel:
`cc1plus` OOMs when trying to compile the Python bindings on MINGW32.

I also tried to switch to `-DCMAKE_BUILD_TYPE=None` and adding `-Os` and/or `-flarge-source-files`. But that doesn't make any difference.

Without the Python bindings, the rest still builds without issues for me on MINGW32. (I don't know if they are needed by any reverse dependencies though.)


lazka:
thanks. for context, the problem is the generated `OgrePYTHON_wrap.cxx` which has 630000 lines

PR #17879, opened 2023-07-27
python-winsdk: skip broken mingw-w64 header

lazka:


lazka:
I'm dropping mingw32 for now. If anyone wants to bring it back, feel free.

Biswa96:
I have reported the winrt header issue in wine, here https://bugs.winehq.org/show_bug.cgi?id=55347

PR #17884, opened 2023-07-27
pycryptodome: fix arm64 build

jeremyd2019:
See also Legrandin/pycryptodome#757.  This may also be due to a bug in mingw-w64, that `__cpuid` prototype is present on ARM.


jeremyd2019:
Successful CLANGARM64 run: https://github.com/msys2-arm/MINGW-packages/actions/runs/5685481550/job/15410442015

PR #17888, opened 2023-07-28
gobject-introspection: generate pyc files

lazka:


MehdiChinoune:
I think VTK also requires that.

lazka:
yeah, there are lots more: https://gist.github.com/lazka/08d7f5a69f84b2005ac7b856d2570399

I'm not sure if we want to block on that

PR #17890, opened 2023-07-28
GnuCOBOL update to 3.2

GitMensch:


GitMensch:
I have a question related to the failing builds - the reason for the failure is that `-I@abs_top_srcdir@` is translated to `-I/C/_/B/src/gnucobol-3.2/build-x86_64/..` which GCC then cannot find directly afterwards:

https://github.com/msys2/MINGW-packages/actions/runs/5776172390/job/15654995192#step:11:784

I'm sure that nothing changed there since the last version - but it now doesn't build any more.
@Biswa96 also noted that he can reproduce the issue locally - so that's no CI issue.

Any insights/hints what's happening and how to fix that? It seems it does need a windows "native" path in there now, is that correct? If yes, is there an option to get the windows native version of the absolute source directory from within the PKGBUILD so we can just override that?
Note: the place where it fails is `make` in the `extras` directory where the `pre-inst-env` script is used to compile something.

@Biswa96 could you please recheck if the _current_ package builds and if it really does (I don't think so, but won't get to a windows machine this week to test myself) compare the content of both pre-inst-env scripts (or attach them here)?

GitMensch:
@Biswa96 do you have any clue why this isn't available via pacman yet?

Biswa96:
According to vercmp, this is aversion downgrade.

```
$ vercmp "3.2" "3.2_rc2"
-1
```

Either you can add epoch or fake the version number.

PR #17903, opened 2023-07-30
python-tomli cleanup

lazka:
In Python 3.11 tomli is also available in the stdlib and various packages use it if available instead of the tomli package.
Remove the tomli dependency for packages which don't require it with Python 3.11 anymore.


lazka:
I've checked them all manually, so CI doesn't add much, so merging right away.

PR #17904, opened 2023-07-30
libiff: Fix building shared library

Biswa96:
* Add -no-undefined option to build shared library.
* Add -export-symbols options with libiff.def file to export proper APIs.
* Remove LIBRARY name in libiff.def to use proper name in import library.


Biswa96:
This solves some issues here https://gitlab.gnome.org/GNOME/gimp/-/merge_requests/922

PR #17907, opened 2023-07-30
starship: add workarounds for arm64 and x86

jeremyd2019:
Patched windows-0.39.0 to link to windows_aarch64_msvc since there was no windows_aarch64_gnullvm at that point.

Since I had to patch dependency crates anyway, added patch for x86 Windows to sha1-asm (from RustCrypto/asm-hashes#61) and re-enabled i686 builds.


jeremyd2019:
Successful CLANGARM64 CI run: https://github.com/msys2-arm/MINGW-packages/actions/runs/5708922289/job/15467065919

jeremyd2019:
@mati865 I requested your review in case you know any less invasive way to get this to work on aarch64-pc-windows-gnullvm.

ognevnydemon:
thanks!

mati865 (approved):
Ideally `tauri-winrt-notification` would upgrade `windows` crate version they depend on but having it as a patch would be more annoying to maintain.

PR #17914, opened 2023-07-31
coin: update to 4.0.0.r12283.g5297f6c

Biswa96:
* Remove upstream patches.
    * Use ninja instead of make.
    * Use git repository as source.


Biswa96:
The package is now way smaller than previous - due to enabling strip and splitting docs. For example, in case of ucrt64, the installation size is
* Before: 143.16 MiB
* After: 23.85 MiB (1/6 than before)

PR #17917, opened 2023-07-31
qt6: update to 6.5.2 (2/3)

olafmandel:
Part two of three, split out from #17913. Depends on #17916

PR #17916, opened 2023-07-31
qt6: update to 6.5.2 (1/3)

olafmandel:
Part one of three, plit out from #17913

PR #17913, opened 2023-07-31
qt6: update to 6.5.2

olafmandel:
Bump from Qt6 6.5.1 to 6.5.2. The removed CVE patch was removed because it is now included in the 6.5.2 release.

PR #17918, opened 2023-07-31
qt6: update to 6.5.2 (3/3)

olafmandel:
Part three of three, split out from #17913. Depends on #17916 and #17917


Biswa96:
Previous commit was added in this PR.

olafmandel:
See https://github.com/msys2/MINGW-packages/pull/17917#issuecomment-1660543699: this PR was split into #17927 and #17928.

Issue #17921, opened 2023-08-01
Add yt-dlp


cmdrkotori:
### Package name

yt-dlp

### Brief description of package

yt-dlp is a youtube-dl fork based on the now inactive youtube-dlc. The main focus of this project is adding new features and patches while also keeping up to date with the original project.

The latest release of yt-dlp was 2023-07-06, while the latest release of youtube-dl is 20**21**-12-17. You may want to consider dropping youtube-dl from the repo in favor of yt-dlp. By packaging yt-dlp in favor of youtube-dl, I would be able to easily bundle a working version of the stream downloader from inside a CI server.

### URL for package's homepage

https://github.com/yt-dlp/yt-dlp

### Provide a basic test case to validate the package's functionality.

`yt-dlp <url>`

### MINGW environments where you need the package

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

No because I only have two or so hours free a day, but there is an Arch PKGBUILD at https://gitlab.archlinux.org/archlinux/packaging/packages/yt-dlp/-/blob/main/PKGBUILD.

Biswa96:
Hi @cmdrkotori, I have added a pull request to add the package. Would you like to test the packages from GitHub Actions artifacts of that pull request?

PR #17923, opened 2023-08-01
yt-dlp: Add package

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/17921


raedrizqie:
@Biswa96 isn't this a python extension?

Biswa96:
What is python extension?

raedrizqie:
a python module package? i mean, the package name should have a 'python' prefix, maybe?

Biswa96:
Though there are some rules about python package naming https://wiki.archlinux.org/title/Python_package_guidelines, here I just followed ArchLinux's package name. To me, everything is a python module 😄

PR #17937, opened 2023-08-03
icu: Update to 73.2

lazka:
0014-mingwize-pkgdata.mingw.patch: refreshed


lazka:
I think this fails due to build cycles being wrongly broken. Lots of distros have already update, so.. yolo

Issue #17942, opened 2023-08-03
[python-lief] ImportError: dynamic module does not define module export function (PyInit_lief)


WillyAguilera1977:
### Description / Steps to reproduce the issue

By running the following code:

```
import lief

# PE
binary = lief.parse("C:\\Windows\\explorer.exe")
print(binary)

```
It gives this error:
C:\Programming\Python\Learn\LiefTest>python3 main.py
Traceback (most recent call last):
   File "C:\Programming\Python\Learn\LiefTest\main.py", line 1, in <module>
     import lief
ImportError: dynamic module does not define module export function (PyInit_lief)

### Expected behavior

I would have to import the lief module and show information about the executable

### Actual behavior

It shows the error:
ImportError: dynamic module does not define module export function (PyInit_lief)

and finish

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

Biswa96:
Thank you for reporting the issue. I can reproduce it in my system.

```
$ python
Python 3.11.4 (main, Aug  2 2023, 21:04:34)  [GCC UCRT 13.2.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import lief
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: dynamic module does not define module export function (PyInit_lief)
>>>
```

We shall look into it.

Biswa96:
There are many issues in the project. I shall try to send those changes to upstream first.

Biswa96:
I have added a pull request to fix this issue. Would you like to test the packages from GitHub Actions artifacts of that pull request?

hartwork:
@Biswa96 thanks for #17947! Not being on Windows myself (but Linux), I lack bandwidth to do the suggested testing myself right now. Maybe from inside CI later, likely doable but not a 5 minute thing. I hope that's okay with you.

Biswa96:
No problem. I have tested it with the example OP provided.

hartwork:
@Biswa96 thanks for the quick fix! Any idea when the binary host will start serving the fixed `0.13.2-2`?

[...]

PR #17947, opened 2023-08-04
python-lief: Fix exported API from python module

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/17942


Biswa96:
@lazka Is this an expected behavior? In python 3.10, the exported API was PyInit_lief. It seems like:

* python 3.10 did `"PyInit" + "_lief"`
* python 3.11 did `"PyInit_" + "_lief"`

lazka:
lgtm

it looks like upstream broke it in https://github.com/lief-project/LIEF/commit/9da24666093c1db2fd2750158f17c2ef039d7f4b

I'd guess they use setuptools now and not cmake, so didn't notice cmake uses the old module name still.

PR #17952, opened 2023-08-04
pkgconf: Update to 2.0.0

lazka:
This removes the patch added in #17692 since the real issues is supposed to be fixed in this new release.


lazka:
@1480c1 want to have a look? Or can you tell me things I should try to build with it so I can test?

1480c1 (approved):
Tested using the opusurl test 

```bash
gcc -static $(./bui/pkgconf --static --cflags opusurl) temp.c $(./bui/pkgconf --static --libs opusurl)
```
```c
#include <stdint.h>
#include <opus/opusfile.h>

int main() { return (intptr_t)opus_server_info_init; }
```

PR #17969, opened 2023-08-06
python: update

naveen521kk:
see https://github.com/msys2-contrib/cpython-mingw/pull/144


lazka:
let's give this a try in staging

PR #17979, opened 2023-08-07
Rust: upgrade to 1.71.1 and replace patch with config options

mati865:
This time tested the final packages...


jeremyd2019:
Uh oh, failure on CLANGARM64
```
Unexpected rustc version: 1.71.0, we should use 1.70.x/1.71.1 to build source with 1.71.1
```
Was 1.71.0 so broken they decided to block it from building 1.71.1?  Do we need to do some bootstrapping hackery again?

jeremyd2019:
https://github.com/rust-lang/rust/commit/69c7d82f42eb91354dbbe83ce4ffed63c8e65ee7
vs
https://github.com/rust-lang/rust/commit/50f47d907d3470047b29529f694d5a3093604e65

mati865:
Oof, it's a bug. We gotta backport https://github.com/rust-lang/rust/commit/31a81a08786826cc6e832bd0b49fb8b934e29648
I'm going to sleep right now and will have some time in the evening, so if it blocks AArch64 runners feel free to revert it.

Issue #17550, opened 2023-06-16
Possible problems with GCC's -foptimize-sibling-calls implementation


cyrilarnould:
See #17549 / #17374. The `-foptimize-sibling-calls` optimization (which is enabled by default with `-O2`) breaks the `mingw-w64-emacs` build when `--with-native-compilation` is passed to the configure script. Could there be something wrong with its implementation in `mingw-w64-gcc`?

svraka:
Just to note, currently no other MINGW patches packages turn off `-foptimize-sibling-calls`, although some packages might do that upstream, e.g. grepping in `MINGW-packages` I found OpenBLAS, although they've been doing it for years [for seemingly different reason](https://github.com/xianyi/OpenBLAS/wiki/Faq#compiling-openblas-with-gccs--fbounds-check-actually-triggers-aborts-in-programs), so presumably it's unrelated to this issue.

Also, AFAIK MINGW packages are not tested by rebuilding after compiler and toolchain upgrades, and `-foptimize-sibling-calls` didn't cause any problems on GCC 12. So can expect more similar issues popping up as packages are built on GCC 13?

revelator:
might be an ICE with gcc-13 should probably report it upstream just in case.

in the meantime a small fix like -fno-optimize-sibling-calls in flags should do the trick for emacs i reckon ?.

cyrilarnould:
An upstream bug report has now been filed:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115132

The problem is that the `-foptimize-sibling-calls` optimization messes up the use of `__builtin_unwind_init`. It is not yet clear if it is a bug in GCC or a misuse of `__builtin_unwind_init`. Regardless, the issue has now been patched in Emacs:

https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=19c983ddedf083f82008472c13dfd08ec94b615f

svraka:
> Regardless, the issue has now been patched in Emacs:
> 
> https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=19c983ddedf083f82008472c13dfd08ec94b615f

Which is included in version 30, so merged here as well 3cc390c6c73af4e8ff3304237843db2ffd6bab2f. The GCC bug seems to be a WONTFIX. I think this issue should be closed.

mati865:
[...]

Issue #17343, opened 2023-05-27
Possible problems with GCC's _FORTIFY_SOURCE implementation


cyrilarnould:
See #17342. Using the `_FORTIFY_SOURCE` option to build Emacs breaks the binary quite badly, resulting in issues #16413 #16375 and #16190. Could there be something wrong with its implementation in `mingw-w64-gcc`?

lb90:
Is emacs broken also in clang64?

svraka:
> Is emacs 28.2-4 broken even in clang64 or in GCC builds only?

Yes, see #16190.

lb90:
Mmh could it have to do with a mingw-w64-crt update then? See also https://github.com/msys2/MINGW-packages/issues/5803#issuecomment-1276784281

oscarfv:
> Mmh could it have to do with a mingw-w64-crt update then? See also [#5803 (comment)](https://github.com/msys2/MINGW-packages/issues/5803#issuecomment-1276784281)

That's a very likely possibility. Emacs defines stpcpy in `lib/string.in.h`:

```
428:13:#if @GNULIB_STPCPY@
429:15:# if @REPLACE_STPCPY@
431:11:#   undef stpcpy
432:12:#   define stpcpy rpl_stpcpy
434:19:_GL_FUNCDECL_RPL (stpcpy, char *,
```

Interesting, uh?

cyrilarnould:
Previously I opened an upstream bug report where the discussion could be continued in case the problem does fall back on Emacs: 

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=63752

I can not judge where exactly this needs to be fixed though.

oscarfv:
IIRC C compilers can do, as far as they maintains semantics, whatever they please with functions provided by the platform libraries (or whatever the C-legalese word is, for Linux that amounts to the Standard C Runtime plus whatever POSIX Linux supports and the functions provided by Linux itself). If a user defines a function with the same name as any other provided by the platform libraries, that is undefined behavior.

Here, gcc is replacing a function that does not belong to the Standard C Runtime nor to other Windows platform libraries with what seems a faulty replacement. So gcc is at fault.

[...]

PR #17988, opened 2023-08-08
emacs: Update to 29.1

cyrilarnould:
`001-ucrt.patch` and `002-clang-fixes.patch` still work even though the line numbers don't match anymore. But I guess it's better to leave them as is to avoid introducing gratuitous changes?

As for the modification to the `build()` function, the emacs-29.1 update has added a `make sanity-check` target that invokes `emacs --batch --eval '(progn (defun f (n) (if (= 0 n) 1 (* n (f (- n 1))))) (princ (f 10)))'` after compilation. In the MinGW builds this returns `3628800`, but on the UCRT builds (including Clang) it returns nothing, which is why the check fails. 

I've modified the recipe so that it skips the sanity check for these versions: I believe the `001-ucrt.patch` is what causes this behaviour change (that or writing to `stdout` is fundamentally broken in the UCRT emacs build). Invoking the sanity check on the emacs-28.2 UCRT build doesn't return anything either, so at the very least this does not introduce a new issue.

Side note: Issues #17550 and #17343 are still present with the new Emacs/GCC versions, I've checked.

### DIFF
--- a/mingw-w64-emacs/PKGBUILD
+++ b/mingw-w64-emacs/PKGBUILD
@@ -2,11 +2,12 @@
 # Maintainer: Oscar Fuentes <ofv@wanadoo.es>
 
 _enable_jit=$([[ "${MINGW_PREFIX}" =~ /clang.* ]] || echo yes)
+_sanity_check=$([[ "${MINGW_PREFIX}" =~ /clang.* ]] || [[ "${MINGW_PREFIX}" =~ /ucrt.* ]] || echo yes)
 _realname=emacs
 pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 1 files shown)


Biswa96:
> but on the UCRT builds (including Clang) it returns nothing, which is why the check fails.

Just out of curiosity, is it possible to fix the issue instead of adding a workaround?

cyrilarnould:
>Just out of curiosity, is it possible to fix the issue instead of adding a workaround?

I think `001-ucrt.patch` is itself also just a workaround rather than a true fix if I'm reading it correctly. I might take a more in depth look into the issues once I have some more time, but at the moment that's the best I can do.

Issue #17998, opened 2023-08-11
[headers] windows.ui.composition.h uses wrong type for enums


dlech:
### Description / Steps to reproduce the issue

We have some CI tests that are failing to build: https://github.com/pywinrt/pywinrt/actions/runs/5827414350/job/15803460259#step:5:31
```
In file included from D:/a/_temp/msys64/ucrt64/include/windows.graphics.capture.interop.h:41,
                 from D:/a/pywinrt/pywinrt/projection/pywinrt/winrt/src/py.Windows.Graphics.Capture.Interop.cpp:4:
D:/a/_temp/msys64/ucrt64/include/windows.ui.composition.h:4919:30: error: use of enum 'DirectXAlphaMode' without previous declaration
 4919 |                         enum DirectXAlphaMode *value) = 0;
      |                              ^~~~~~~~~~~~~~~~
D:/a/_temp/msys64/ucrt64/include/windows.ui.composition.h:4922:30: error: use of enum 'DirectXPixelFormat' without previous declaration
 4922 |                         enum DirectXPixelFormat *value) = 0;
      |                              ^~~~~~~~~~~~~~~~~~
D:/a/_temp/msys64/ucrt64/include/windows.ui.composition.h:5707:30: error: use of enum 'DirectXPixelFormat' without previous declaration
 5707 |                         enum DirectXPixelFormat format,
      |                              ^~~~~~~~~~~~~~~~~~
D:/a/_temp/msys64/ucrt64/include/windows.ui.composition.h:5708:30: error: use of enum 'DirectXAlphaMode' without previous declaration
 5708 |                         enum DirectXAlphaMode mode,
      |                    
```

### Expected behavior

Code should compile.

### Actual behavior

Compiling fails due to bad header.

Apparently, it is using C style `enum` instead of C++. 

Instead of `enum DirectXAlphaMode` it should be `ABI::Windows::Graphics::DirectX::DirectXAlphaMode` and so on.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22621

### MINGW environments affected

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [X] CLANGARM64

### Are you willing to submit a PR?

if someone can point me to the source code

Biswa96:
[...]

PR #17999, opened 2023-08-11
headers: Fix compiler error with winrt header

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/17998


mingw-w64-headers-git/0002-heades-add-full-name-winrt.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/mingw-w64-headers/include/windows.ui.composition.h b/mingw-w64-headers/include/windows.ui.composition.h
+index faf73ea0b..76b4c2172 100644
+--- a/mingw-w64-headers/include/windows.ui.composition.h
++++ b/mingw-w64-headers/include/windows.ui.composition.h
+@@ -4916,10 +4916,10 @@ namespace ABI {
+                 ICompositionDrawingSurface : public IInspectable
+                 {
+                     virtual HRESULT STDMETHODCALLTYPE get_AlphaMode(
+-                        enum DirectXAlphaMode *value) = 0;
++                        enum Windows::Graphics::DirectX::DirectXAlphaMode *value) = 0;
```
dlech, line 10: Since this is C++ shouldn't the `enum` keyword be omitted here?

Biswa96: Yes, I didn't noticed it. Does that cause any issue?

PR #18000, opened 2023-08-11
v8: update to 11.8.7

raedrizqie:


jeroen:
I noticed abseil-cpp is both declared as a pacman dependency but also cloned in bare-clones/abseil-cpp. Do we need both?

raedrizqie:
ah.. yes, i forgot to remove it from depends

PR #18001, opened 2023-08-12
qemu: Update to 8.0.4

heljkon:
The QEMU v8.0.4 stable release contains security fixes.

See https://lists.gnu.org/archive/html/qemu-devel/2023-08/msg02106.html


heljkon:
It seems the 8.0.4 release contains some problems...

I will drop this, because 8.1.0 will be released soon

PR #18002, opened 2023-08-12
LLVM: update to 17.0.1 - Part1

MehdiChinoune:
MLIR fails to build on GNU environments.
```
 [1601/1612] Linking CXX executable bin\mlir-lsp-server.exe
 FAILED: bin/mlir-lsp-server.exe 
 cmd.exe /C "cd . && D:\M\msys64\ucrt64\bin\clang++.exe -march=nocona -msahf -mtune=generic -O2 -pipe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -Werror=mismatched-tags -O3 -DNDEBUG -pipe -Wl,--stack,16777216    -Wl,--gc-sections @CMakeFiles\mlir-lsp-server.rsp -o bin\mlir-lsp-server.exe -Wl,--out-implib,tools\mlir-lsp-server\libmlir-lsp-server.dll.a -Wl,--major-image-version,0,--minor-image-version,0  && cd ."
 lib/Dialect/LLVMIR/libMLIRLLVMDialect.a(LLVMDialect.cpp.obj):LLVMDialect.cpp:(.text$_ZN4mlir4LLVM15AtomicCmpXchgOp15getInherentAttrEPNS_11MLIRContextERKNS0_6detail33AtomicCmpXchgOpGenericAdaptorBase10PropertiesEN4llvm9StringRefE[_ZN4mlir4LLVM15AtomicCmpXchgOp15getInherentAttrEPNS_11MLIRContextERKNS0_6detail33AtomicCmpXchgOpGenericAdaptorBase10PropertiesEN4llvm9StringRefE]+0x1d): relocation truncated to fit: IMAGE_REL_AMD64_REL32 against `.rdata'
 lib/Dialect/LLVMIR/libMLIRLLVMDialect.a(LLVMDialect.cpp.obj):LLVMDialect.cpp:(.text$_ZN4mlir4LLVM15AtomicCmpXchgOp15setInherentAttrERNS0_6detail33AtomicCmpXchgOpGenericAdaptorBase10PropertiesEN4llvm9StringRefENS_9AttributeE[_ZN4mlir4LLVM15AtomicCmpXchgOp15setInherentAttrERNS0_6detail33AtomicCmpXchgOpGenericAdaptorBase10PropertiesEN4llvm9StringRefENS_9AttributeE]+0x24): relocation truncated to fit: IMAGE_REL_AMD64_REL32 against `.rdata'
...
```
Building LLVM fails on MINGW32:
```
 [5861/5950] Linking CXX executable bin\lli.exe
 FAILED: bin/lli.exe 
 cmd.exe /C "cd . && D:\M\msys64\mingw32\bin\clang++.exe -march=pentium4 -mtune=generic -O2 -pipe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -O3 -DNDEBUG -pipe -Wl,--no-seh -Wl,--large-address-aware -Wl,--large-address-aware -Wl,--stack,16777216    -Wl,--gc-sections tools/lli/CMakeFiles/lli.dir/lli.cpp.obj tools/lli/CMakeFiles/lli.dir/ExecutionUtils.cpp.obj -o bin\lli.exe -Wl,--out-implib,lib\liblli.dll.a -Wl,--major-image-version,0,--minor-image-version,0  lib/libLLVM-17-rc1.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
 D:/M/msys64/mingw32/bin/ld: cannot export _alloca: symbol not found
```


ognevnydemon:
FYI, it will be good to rebuild Rust after LLVM update, because there is some [performance boost](https://perf.rust-lang.org/compare.html?start=03a119b0b0e310d22d94399b24ed030056050f13&end=443c3161dd04f4c1b656a626f9079921bee9c326&stat=instructions:u)

mmuetzel:
Can the MINGW64 build errors be avoided by adding `-Wa,-mbig-obj` to the compiler flags?

mmuetzel:
Not sure. But maybe the MINGW32 build error is caused by this change:
https://reviews.llvm.org/D150555

@mstorsjo: Could you please have a look if that might be related?

MehdiChinoune:
Flang is also broken.

mmuetzel:
> Flang is also broken.

Does it fail to build? Or do the resulting binaries not work? What is the exact error?

mstorsjo:
> Not sure. But maybe the MINGW32 build error is caused by this change: https://reviews.llvm.org/D150555
> 
> @mstorsjo: Could you please have a look if that might be related?

Yes, it seems probable that it’s caused by that change - I’ll try to have a look a bit later.

MehdiChinoune:
> > Flang is also broken.
> 
> Does it fail to build? Or do the resulting binaries not work? What is the exact error?

- It fails to build if building shared libraries is enabled (`-DBUILD_SHARED_LIBS=ON`) because of some duplicate referenced symbols from a Flang library and LLVMSupport library
- It succeeds to build with `-DBUILD_SHARED_LIBS=OFF`, but It looks like some libc++ symbols slipped into those static libraries because of the included headers.
```
  -- Check for working Fortran compiler: C:/msys64/clang64/bin/flang.exe - broken
  CMake Error at C:/msys64/clang64/share/cmake/Modules/CMakeTestFortranCompiler.cmake:59 (message):
    The Fortran compiler
  
      "C:/msys64/clang64/bin/flang.exe"
  
    is not able to compile a simple test program.
  
    It fails with the following output:
  
      Change Dir: 'C:/_/B/src/build-CLANG64/CMakeFiles/CMakeScratch/TryCompile-b30cjz'
      
      Run Build Command(s): C:/msys64/clang64/bin/ninja.exe -v cmTC_2a314
      [1/4] cmd.exe /C "C:\msys64\clang64\bin\flang.exe -cpp    -E C:\_\B\src\build-CLANG64\CMakeFiles\CMakeScratch\TryCompile-b30cjz\testFortranCompiler.f > CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f-pp.f && C:\msys64\clang64\bin\cmake.exe -E cmake_ninja_depends --tdi=CMakeFiles\cmTC_2a314.dir\FortranDependInfo.json --lang=Fortran --src=CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f-pp.f --out=CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f-pp.f --dep=CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f-pp.f.d --obj=CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f.obj --ddi=CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f.obj.ddi"
      [2/4] C:\msys64\clang64\bin\cmake.exe -E cmake_ninja_dyndep --tdi=CMakeFiles\cmTC_2a314.dir\FortranDependInfo.json --lang=Fortran --dd=CMakeFiles\cmTC_2a314.dir\Fortran.dd @CMakeFiles\cmTC_2a314.dir\Fortran.dd.rsp
      [3/4] C:\msys64\clang64\bin\flang.exe -IC:\_\B\src\build-CLANG64\CMakeFiles\CMakeScratch\TryCompile-b30cjz -ffixed-line-length-72 -c CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f-pp.f -o CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f.obj
      [4/4] cmd.exe /C "cd . && C:\msys64\clang64\bin\flang.exe  -pipe CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f.obj -o cmTC_2a314.exe -Wl,--out-implib,libcmTC_2a314.dll.a -Wl,--major-image-version,0,--minor-image-version,0   && cd ."
      FAILED: cmTC_2a314.exe 
      cmd.exe /C "cd . && C:\msys64\clang64\bin\flang.exe  -pipe CMakeFiles\cmTC_2a314.dir\testFortranCompiler.f.obj -o cmTC_2a314.exe -Wl,--out-implib,libcmTC_2a314.dll.a -Wl,--major-image-version,0,--minor-image-version,0   && cd ."
      ld.lld: error: undefined symbol: std::__1::__libcpp_verbose_abort(char const*, ...)
      >>> referenced by libFortranRuntime.a(io-api.cpp.obj):(std::__1::__throw_bad_variant_access[abi:v170000]())
      >>> referenced by libFortranRuntime.a(io-stmt.cpp.obj)
      >>> referenced by libFortranRuntime.a(unit.cpp.obj)
```

mmuetzel:
Sorry for bothering with these questions. I'm having trouble building LLVM here locally. (Maybe, time to purge my MSYS2 installation again and start from scratch?)
You seem to have come a lot further than myself.

Regarding the build error for Flang with shared libraries: Could you please upload the build log with that error? Is that a regression in Flang 17?

When it comes to the linker error in the CMake project: That looks like it is missing to link against libc++. Does the following change make a difference?
```
diff --git "a/clang/lib/Driver/ToolChain.cpp" "b/clang/lib/Driver/ToolChain.cpp"
index 8dafc3d481c2..4fd06480b34e 100644
--- "a/clang/lib/Driver/ToolChain.cpp"
+++ "b/clang/lib/Driver/ToolChain.cpp"
@@ -1187,7 +1187,7 @@ void ToolChain::AddClangCXXStdlibIsystemArgs(
 }
 
 bool ToolChain::ShouldLinkCXXStdlib(const llvm::opt::ArgList &Args) const {
-  return getDriver().CCCIsCXX() &&
+  return (getDriver().CCCIsCXX() || getDriver().IsFlangMode()) &&
          !Args.hasArg(options::OPT_nostdlib, options::OPT_nodefaultlibs,
                       options::OPT_nostdlibxx);
 }
```

MehdiChinoune:
> Regarding the build error for Flang with shared libraries: Could you please upload the build log with that error?

You know It takes a long time to build, I will try to upload it anyway.

> Is that a regression in Flang 17?

Building shared libraries for Flang was enabled recently (after 16.x release).

> When it comes to the linker error in the CMake project: That looks like it is missing to link against libc++. Does the following change make a difference?

Flang is not supposed to depend on libc++.

> ```
> diff --git "a/clang/lib/Driver/ToolChain.cpp" "b/clang/lib/Driver/ToolChain.cpp"
> index 8dafc3d481c2..4fd06480b34e 100644
> --- "a/clang/lib/Driver/ToolChain.cpp"
> +++ "b/clang/lib/Driver/ToolChain.cpp"
> @@ -1187,7 +1187,7 @@ void ToolChain::AddClangCXXStdlibIsystemArgs(
>  }
>  
>  bool ToolChain::ShouldLinkCXXStdlib(const llvm::opt::ArgList &Args) const {
> -  return getDriver().CCCIsCXX() &&
> +  return (getDriver().CCCIsCXX() || getDriver().IsFlangMode()) &&
>           !Args.hasArg(options::OPT_nostdlib, options::OPT_nodefaultlibs,
>                        options::OPT_nostdlibxx);
>  }
> ```

You want to link to libc++ by default?

mmuetzel:
> You know It takes a long time to build, I will try to upload it anyway.

Thank you.

> You want to link to libc++ by default?

Yeah. That might be an easy workaround. (I agree that chances upstream would ever accept that are low.)

mmuetzel:
I'm not exactly sure if that's correct. But it might be that the Fortran runtime depends on the C++ runtime because `std::abort()` is called from within it. `grep`ing for it, I find it in `terminator.cpp` and in `stop.cpp`.
If that is the case, a possible way to avoid that dependency is to avoid using that function in the Fortran runtime. I don't know how it could be replaced though. So, depending on the C++ runtime is the "next best thing" that could work.

mmuetzel:
Maybe, the following also works (and might avoid the dependency on libc++):
```
diff --git "a/flang/runtime/io-api.cpp" "b/flang/runtime/io-api.cpp"
index 2c17f10e7769..4eb0248d4c98 100644
--- "a/flang/runtime/io-api.cpp"
+++ "b/flang/runtime/io-api.cpp"
@@ -1515,3 +1515,15 @@ enum Iostat IONAME(CheckUnitNumberInRange128)(common::int128_t unit,
 #endif
 
 } // namespace Fortran::runtime::io
+
+// Provide own definition for `std::__libcpp_verbose_abort` to avoid dependency
+// on the version provided by libc++.
+
+void std::__libcpp_verbose_abort(char const* format, ...) {
+  va_list list;
+  va_start(list, format);
+  std::vfprintf(stderr, format, list);
+  va_end(list);
+
+  std::abort();
+}
```

(I can't test myself unfortunately.)

MehdiChinoune:
`BUILD_SHARED_LIBS=ON`:
```
  [223/359] Linking CXX shared library bin\libFIRDialect.dll
  FAILED: bin/libFIRDialect.dll lib/libFIRDialect.dll.a 
  cmd.exe /C "cd . && C:\msys64\clang64\bin\clang++.exe -march=nocona -msahf -mtune=generic -O2 -pipe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -Wno-deprecated-copy -Wno-string-conversion -Wno-ctad-maybe-unsupported -Wno-unused-command-line-argument -Wstring-conversion           -Wcovered-switch-default -Wno-nested-anon-types -O3 -DNDEBUG  -pipe -fuse-ld=lld   -Wl,--gc-sections -shared -o bin\libFIRDialect.dll -Wl,--out-implib,lib\libFIRDialect.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles\FIRDialect.rsp  && cd ."
  ld.lld: error: duplicate symbol: llvm::errs()
  >>> defined at libLLVMSupport.a(raw_ostream.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::raw_ostream::operator<<(llvm::formatv_object_base const&)
  >>> defined at libLLVMSupport.a(raw_ostream.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::raw_ostream::flush_nonempty()
  >>> defined at libLLVMSupport.a(raw_ostream.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::raw_ostream::operator<<(long)
  >>> defined at libLLVMSupport.a(raw_ostream.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::raw_null_ostream::~raw_null_ostream()
  >>> defined at libLLVMSupport.a(raw_ostream.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::raw_ostream::indent(unsigned int)
  >>> defined at libLLVMSupport.a(raw_ostream.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::cl::parser<long long>::parse(llvm::cl::Option&, llvm::StringRef, llvm::StringRef, long long&)
  >>> defined at libLLVMSupport.a(CommandLine.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
  
  ld.lld: error: duplicate symbol: llvm::cl::basic_parser_impl::getOptionWidth(llvm::cl::Option const&) const
  >>> defined at libLLVMSupport.a(CommandLine.cpp.obj)
  >>> defined at libFIRDialectSupport.dll
...
```

mmuetzel:
To avoid that link issue, I'd guess that LLVMSupport would also need to be built as a shared library. Afaict, currently there is only a static version:
https://packages.msys2.org/package/mingw-w64-x86_64-llvm?repo=mingw64

I don't know if that changed for LLVM 17.

It looks like that static library is linked into the shared library FIRDialectSupport (with `--export-all-symbols` -- maybe implicitly?). That probably leads to this collision of symbols...

MehdiChinoune:
> Maybe, the following also works (and might avoid the dependency on libc++):
> 
> ```
> diff --git "a/flang/runtime/io-api.cpp" "b/flang/runtime/io-api.cpp"
> index 2c17f10e7769..4eb0248d4c98 100644
> --- "a/flang/runtime/io-api.cpp"
> +++ "b/flang/runtime/io-api.cpp"
> @@ -1515,3 +1515,15 @@ enum Iostat IONAME(CheckUnitNumberInRange128)(common::int128_t unit,
>  #endif
>  
>  } // namespace Fortran::runtime::io
> +
> +// Provide own definition for `std::__libcpp_verbose_abort` to avoid dependency
> +// on the version provided by libc++.
> +
> +void std::__libcpp_verbose_abort(char const* format, ...) {
> +  va_list list;
> +  va_start(list, format);
> +  std::vfprintf(stderr, format, list);
> +  va_end(list);
> +
> +  std::abort();
> +}
> ```
> 
> (I can't test myself unfortunately.)

It works

mmuetzel:
> It works

Thank you for checking.
I proposed that change to upstream: https://reviews.llvm.org/D158957

mmuetzel:
That patch might have broken compilation of Flang with GCC. Sorry for that if it did.
Guarding that definition with `#if defined(_LIBCPP_VERBOSE_ABORT)` might help in that case.

mati865:
> That patch might have broken compilation of Flang with GCC. Sorry for that if it did. Guarding that definition with `#if defined(_LIBCPP_VERBOSE_ABORT)` might help in that case.

I think building with GCC doesn't work anyway because GCC doesn't use symbol visibility directives and hits DLL limit.

mmuetzel:
> I think building with GCC doesn't work anyway because GCC doesn't use symbol visibility directives and hits DLL limit.

What makes you think that's an issue with Flang?

mati865:
> What makes you think that's an issue with Flang?

What I meant building Flang with GCC is untested and thus likely broken already. In that case you are not breaking it since it's broken already.

mmuetzel:
> What I meant building Flang with GCC is untested and thus likely broken already. In that case you are not breaking it since it's broken already.

But it is already distributed by MSYS2: https://packages.msys2.org/package/mingw-w64-x86_64-flang?repo=mingw64

mati865:
> But it is already distributed by MSYS2: [packages.msys2.org/package/mingw-w64-x86_64-flang?repo=mingw64](https://packages.msys2.org/package/mingw-w64-x86_64-flang?repo=mingw64)

Yes but it's built with Clang: https://github.com/msys2/MINGW-packages/blob/9c112027fd09b60d634c5faf6dc819f57066afe2/mingw-w64-flang/PKGBUILD#L1

mstorsjo:
> > But it is already distributed by MSYS2: [packages.msys2.org/package/mingw-w64-x86_64-flang?repo=mingw64](https://packages.msys2.org/package/mingw-w64-x86_64-flang?repo=mingw64)
> 
> Yes but it's built with Clang:
> 
> https://github.com/msys2/MINGW-packages/blob/9c112027fd09b60d634c5faf6dc819f57066afe2/mingw-w64-flang/PKGBUILD#L1

Yes, but that Clang build in the mingw64 environment still uses libstdc++, not libc++.

MehdiChinoune:
Flang fails to build on UCRT64 because another reason:
```
  [268/359] Linking CXX executable bin\tco.exe
  FAILED: bin/tco.exe 
  cmd.exe /C "cd . && C:\msys64\ucrt64\bin\clang++.exe -march=nocona -msahf -mtune=generic -O2 -pipe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -Wno-deprecated-copy -Wno-string-conversion -Wno-ctad-maybe-unsupported -Wno-unused-command-line-argument -Wstring-conversion           -Wcovered-switch-default -Wno-nested-anon-types -O3 -DNDEBUG -pipe -fuse-ld=lld -Wl,--stack,16777216    -Wl,--gc-sections @CMakeFiles\tco.rsp -o bin\tco.exe -Wl,--out-implib,lib\libtco.dll.a -Wl,--major-image-version,0,--minor-image-version,0  && cd ."
  ld.lld: error: libMLIRLLVMDialect.a(LLVMDialect.cpp.obj): .rdata should not refer to special section 0
  
  [269/359] Linking CXX executable bin\bbc.exe
  FAILED: bin/bbc.exe 
  cmd.exe /C "cd . && C:\msys64\ucrt64\bin\clang++.exe -march=nocona -msahf -mtune=generic -O2 -pipe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -Wno-deprecated-copy -Wno-string-conversion -Wno-ctad-maybe-unsupported -Wno-unused-command-line-argument -Wstring-conversion           -Wcovered-switch-default -Wno-nested-anon-types -O3 -DNDEBUG -pipe -fuse-ld=lld -Wl,--stack,16777216    -Wl,--gc-sections @CMakeFiles\bbc.rsp -o bin\bbc.exe -Wl,--out-implib,lib\libbbc.dll.a -Wl,--major-image-version,0,--minor-image-version,0  && cd ."
  ld.lld: error: libMLIRLLVMDialect.a(LLVMDialect.cpp.obj): .rdata should not refer to special section 0
  
```

mstorsjo:
> Not sure. But maybe the MINGW32 build error is caused by this change: https://reviews.llvm.org/D150555
> 
> @mstorsjo: Could you please have a look if that might be related?

I had a look at this and I managed to reproduce it. I believe this change should fix it: https://github.com/mstorsjo/llvm-project/commit/lli-export-chkstk I haven't submitted it upstream yet, and I'll run some more build tests with it before I do that.

mstorsjo:
> > Not sure. But maybe the MINGW32 build error is caused by this change: https://reviews.llvm.org/D150555
> > @mstorsjo: Could you please have a look if that might be related?
> 
> I had a look at this and I managed to reproduce it. I believe this change should fix it: [mstorsjo/llvm-project@lli-export-chkstk](https://github.com/mstorsjo/llvm-project/commit/lli-export-chkstk) I haven't submitted it upstream yet, and I'll run some more build tests with it before I do that.

That didn’t quite work (even if it worked in a small standalone testcase) - the dummy declaration of the function `_alloca` (the name of the chkstk function on i386) conflicts with the C macro `_alloca`, if the headers defining that have been included. I’ll try to rework the workaround…

MehdiChinoune:
some LLVM developers are discussing supporting building [Supporting LLVM_BUILD_LLVM_DYLIB on Windows](https://discourse.llvm.org/t/supporting-llvm-build-llvm-dylib-on-windows/58891) by using `__declspec(dllexport)`, hopefully that will help get rid of those errors about allocations/relocations/symbols.

mstorsjo:
> some LLVM developers are discussing supporting building [Supporting LLVM_BUILD_LLVM_DYLIB on Windows](https://discourse.llvm.org/t/supporting-llvm-build-llvm-dylib-on-windows/58891) by using `__declspec(dllexport)`, hopefully that will help get rid of those errors about allocations/relocations/symbols.

That would make things easier wrt `LLVM_BUILD_LLVM_DYLIB` with e.g. GCC which doesn't support the pseudo-hidden visibility on Windows. But it won't help for the chkstk issue with lli, that's just a hack to make lli tests run successfully in mingw configs...

mstorsjo:
> > > Not sure. But maybe the MINGW32 build error is caused by this change: https://reviews.llvm.org/D150555
> > > @mstorsjo: Could you please have a look if that might be related?
> > 
> > 
> > I had a look at this and I managed to reproduce it. I believe this change should fix it: [mstorsjo/llvm-project@lli-export-chkstk](https://github.com/mstorsjo/llvm-project/commit/lli-export-chkstk) I haven't submitted it upstream yet, and I'll run some more build tests with it before I do that.
> 
> That didn’t quite work (even if it worked in a small standalone testcase) - the dummy declaration of the function `_alloca` (the name of the chkstk function on i386) conflicts with the C macro `_alloca`, if the headers defining that have been included. I’ll try to rework the workaround…

I fixed it by adding an `#undef _alloca` and posted for review at https://reviews.llvm.org/D159085 now (and updated the commit in my fork on github too).

mmuetzel:
I tried to rebuild this PR locally from scratch. But I see the following file conflicts:
[test-install.log](https://github.com/msys2/MINGW-packages/files/12500586/test-install.log)

Am I doing something wrong?

mmuetzel:
Some of the build errors in GCC environments might be fixed by the change that was proposed to upstream here:
https://reviews.llvm.org/D159404

MehdiChinoune:
> Some of the build errors in GCC environments might be fixed by the change that was proposed to upstream here: https://reviews.llvm.org/D159404

It doesn't fix it, as we already using clang compiler which enable big-objects by default as mentioned by mstorsjo in the cited PR.

mmuetzel:
> It doesn't fix it, as we already using clang compiler which enable big-objects by default as mentioned by mstorsjo in the cited PR.

Are you sure that clang enables this automatically both in GCC and MSVC mode? MSYS2 is using Clang in "GCC mode".

mstorsjo:
> > It doesn't fix it, as we already using clang compiler which enable big-objects by default as mentioned by mstorsjo in the cited PR.
> 
> Are you sure that clang enables this automatically both in GCC and MSVC mode? MSYS2 is using Clang in "GCC mode".

Yes, the flag is entirely a no-op in Clang. I can try to make testcases to show it off (and to verify that MSVC really needs the flag) later.

MehdiChinoune:
> > It doesn't fix it, as we already using clang compiler which enable big-objects by default as mentioned by mstorsjo in the cited PR.
> 
> Are you sure that clang enables this automatically both in GCC and MSVC mode? MSYS2 is using Clang in "GCC mode".

I tried the patch It didn't fix the issue.

mmuetzel:
Yeah. Come to think of it, I might have mixed up two different options in my head.

Does adding `-mcmodel=large` (or `-mcmodel=medium`?) make a difference in the GCC environments?
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#index-mcmodel_003dmedium-1

MehdiChinoune:
It's Unix only, doesn't apply to Windows.

mmuetzel:
According to this (older) bug report it *does* do something on Windows:
https://bugs.llvm.org/show_bug.cgi?id=42983#c5

And if I understand correctly, it might help if relative offsets are large.

mingw-w64-clang/0002-lli-Make-sure-the-exported-__chkstk-functions-are-included-when-exporting-them.patch
mmuetzel, line 1: This change has been applied upstream in the meantime:
https://github.com/llvm/llvm-project-release-prs/commit/b51021f7b39deae7450fc2f722c1e5ec655fe105

It's part of 17.0.1.

mmuetzel:
The package grokker lists the following packages to be affected by an API change:
```
mingw-w64-include-what-you-use
mingw-w64-spirv-llvm-translator
mingw-w64-mesa
mingw-w64-qt-creator
mingw-w64-creduce
```

IIUC, the package grokker runs for UCRT64.
Is it possible that more packages are affected by that change in the CLANG* environments? Is there a way to check that?

MehdiChinoune:
> IIUC, the package grokker runs for UCRT64.
> Is it possible that more packages are affected by that change in the CLANG* environments? Is there a way to check that?

clone package-grokker repo and run your own job (on GHA).

mmuetzel:
Thanks for checking.

The CI will very likely time out if we tried to rebuild those packages as part of this PR.
Should we check beforehand (i.e., locally) if those projects are already compatible to the API changes? Or is it good enough to check in staging if they compile?

MehdiChinoune:
> Should we check beforehand (i.e., locally) if those projects are already compatible to the API changes? Or is it good enough to check in staging if they compile?

I think we should checkc before merging.

MehdiChinoune:
I have rebuilt those packages on CLANG64 and UCRT64:
- qt6-tools, qt-creator and IWYU doesn't require any changes (qt6-static building ...)
- spirv-llvm-translator: update to 17.0.0 as It is required by mesa which doesn't accept a version lower than the major version of LLVM.
- mesa: backport a MR and merged patch from main branch.
- creduce: Fails as usual #csmith-project/creduce/issues/262

I think we could merge this PR.

PR #18010, opened 2023-08-12
blosc2: update to 2.10.1

MehdiChinoune:


Biswa96:
The cmake files are installed in wrong directory.

Biswa96:
> The cmake files are installed in wrong directory.

It will be fixed by this pull request https://github.com/msys2/MINGW-packages/pull/18095

PR #18057, opened 2023-08-12
Restore bsddb3, fix clang build and bump package release

hgohel:
Restore bsddb3 6.2.29 which was recently removed with #17874 due to build failures with clang, and also work around the clang build error.

Clang generates a compiler error against code which gcc treats as a warning. This change disables that error (only for clang) so the build can succeed. The code in question needs to be fixed (has been raised with the project), but this fix allows the build to succeed so downstream consumers (e.g. Gramps) can use the package.

Bumped up the package release. Built mingw64 gcc, clang and ucrt builds locally with success.

FYI @mmuetzel @Biswa96

### DIFF
--- a/mingw-w64-python-bsddb3/PKGBUILD
+++ b/mingw-w64-python-bsddb3/PKGBUILD
@@ -1,0 +1,58 @@
+# Contributor: Josip <bpisoj@gmail.com>
+
+_realname=bsddb3
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+provides=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+conflicts=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+replaces=("${MINGW_PACKAGE_PREFIX}-python3-${_realname}")
+pkgver=6.2.9
+pkgrel=5
+pkgdesc="Python bindings for Oracle Berkeley DB (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
+license=('BSD')
+url="https://www.jcea.es/programacion/pybsddb.htm"
+depends=("${MINGW_PACKAGE_PREFIX}-python"
+         "${MINGW_PACKAGE_PREFIX}-db")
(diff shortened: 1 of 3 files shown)


Biswa96:
Did you look at the compiler error that clang shows? It still passing a function pointer of different number of parameters.

Biswa96:
The error could be solved with db version 6.1 or greater. In the _bsddb.c file, the `#if (DBVER >= 61)` condition adds the 4 th parameter in `_db_compareCallback` function.

hgohel:
Hi @Biswa96 The goal of this PR is to restore the availability of the package to help consumers blocked by its recent removal. 

For gcc builds there is no change in behavior from the previously posted packages which were used by consumers for a long time. The only thing this PR does is to set a compiler option in the clang build to match the default behavior in the gcc build.

Hopefully this is acceptable until the code is improved, and/or db package is bumped up as you suggested? (not something I'm familiar with). Thanks for the review.

Biswa96:
Should we also ignore the warning from upstream?

https://www.jcea.es/programacion/pybsddb.htm

> bsddb3 supports up to Python 3.9. If you are running a more modern Python versión, you must upgrade to berkeleydb library.

hgohel:
@Biswa96 Initially I shared your concern about the warning to move to berkeleydb for Python versions greater than 3.9 made by the maintainer of bsddb3, but I've come to interpret that with a grain of salt since:

Both bsddb3 6.2.29 and Python 3.9 were released in 2020 and at the time bsddb3 testsuite was verified with Python 3.9. Soon after that the project migrated to berkeleydb 18.x and the bsddb3 6.2.x line was no longer officially supported and that IMO, is the risk.

As of today, anecdotal data points to bsddb3 6.2.29 as working with Python 3.11 (currently in a beta build of Gramps software). It would boost confidence to verify the testsuite, but the instructions and/or test code seem to be outdated and I wasn't able to run them.

mmuetzel:
It was probably a reasonable move by the LLVM developers to make that warning an error by default. In the end, a program will most likely crash horribly should it ever try to execute that code.

Having written that, a program that never executes that part of the code might "be fine". So, there could still be a valid reason to distribute this package with the known defect.

But any users should be aware that they can't use that callback ever.

It would be nice though if someone could confirm that the remainder of the package is still working with the new Python version.

lazka:
I can confirm that the basics work (reading/writing some things from/to a DB)

PR #18096, opened 2023-08-13
db: split docs into separate package

Biswa96:


Biswa96:
Just for the record. The clang build did not succeed due to the following failure with configure script.

```
checking whether C++ supports the wstring class... checking for thread local storage (TLS) class... pthread
```

And the config.log showed

```
conftest.c:34:49: error: incompatible pointer to integer conversion passing 'pthread_key_t *' (aka 'unsigned int *') to parameter of type 'pthread_key_t' (aka 'unsigned int'); remove & [-Wint-conversion]
             return (void *)pthread_getspecific(&key);
                                                ^~~~
pthread.h:285:62: note: passing argument to parameter 'key' here
void *      WINPTHREAD_API pthread_getspecific(pthread_key_t key);
                                                             ^
conftest.c:37:35: error: incompatible pointer to integer conversion passing 'pthread_key_t *' (aka 'unsigned int *') to parameter of type 'pthread_key_t' (aka 'unsigned int'); remove & [-Wint-conversion]
              pthread_setspecific(&key, p);
                                  ^~~~
pthread.h:286:62: note: passing argument to parameter 'key' here
int         WINPTHREAD_API pthread_setspecific(pthread_key_t key, const void *value);
                                                             ^
1 warning and 2 errors generated.
```

PR #18160, opened 2023-08-13
db: update to 6.2.32

Biswa96:
/cc @hgohel @mmuetzel


mmuetzel:
@jeremyd2019 mentioned in #17871 that there was a license change for this package and that might be the reason that it wasn't updated yet.
Is that no longer an issue?

Biswa96:
There could be, the latest version is 18 according to https://repology.org/project/db/versions. But this is just a minor version update and the version matches with ArchLinux.

hgohel:
Oracle appears to have [moved to AGPL license with db version 6.0](https://docs.oracle.com/cd/E17076_05/html/installation/license_change60.html) and this PR is bumping up from 6.0.19 which was based on the same license. Here's [Oracle's current FAQ on bdb licensing](https://www.oracle.com/database/technologies/berkeleydb-db-faq.html#DoesthemovetotheAGPLopensourcelicenseaffectcustomerswhopurchasedacommerciallicense). Warning this is not my area of expertise - this is just based on documents found via searching.

Biswa96:
> There appears to have been been concerns about bumping up db version previously as well. See [6c041b0](https://github.com/msys2/MINGW-packages/commit/6c041b0d483b8127d9e4e1fdcbe067022ef78c58)

Good catch. openldap does not depend on db now. db was commented out in openldap PKGBUILD file.

Biswa96:
@GitMensch I have fix the gnucobol build issue here. Would you like to review the change please?

GitMensch:
It is good to have Berkeley DB updated, but as it has a different license it seems to be more reasonable to have both bdb (under current license) and bdb-agpl (newest one, under new license), so people can choose.

... not sure how to handle the "conflict" part then...

GitMensch:
@Biswa96: Thanks for checking the GnuCOBOL issue.
The patch tackles the right thing, but that should be done in the templates ([build_aux/pre-inst-env.in](https://github.com/OCamlPro/gnucobol/blob/gnucobol-3.2/build_aux/pre-inst-env.in) and [tests/atlocal.in](https://github.com/OCamlPro/gnucobol/blob/gnucobol-3.2/tests/atlocal.in))

The "real issue" is that those are already in there:

```sh
# options that are also used in atlocal (always add to both)
COB_CFLAGS="-I@abs_top_srcdir@ @COB_CFLAGS@ ${COB_CFLAGS}"
```
https://github.com/OCamlPro/gnucobol/blob/gnucobol-3.2/build_aux/pre-inst-env.in#L41
https://github.com/OCamlPro/gnucobol/blob/gnucobol-3.2/tests/atlocal.in#L231

I therefore think something in MSYS2 must have changed, because these lines were changed 3 years ago (before the previous release, when that worked).

One thing that comes to mind: Does it work better if there's a space after the `-I` (and in this case also the `-L`)?

mmuetzel:
> Oracle appears to have moved to AGPL license with db version 6.0. Here's Oracle's current FAQ on bdb licensing.

According to Wikipedia Oracle moved to the Affero GPL starting with version 6.0.21. The currently packaged version 6.0.19 still uses a less restrictive license.

Biswa96:
So, can this update be merged with the current dependencies? This update is just to fix the python-bsddb3 build.

Biswa96:
I gave up. I am not familiar with license and related stuff.

GitMensch:
@Biswa96 Would you mind to copy the package along with your change as a new mingw-w64-db-agpl package?
If the python-bsddb3 needs the new one (which I'd find strange because that wouldn't be "BSD" as the name implies) then this could be changed to depend on the new package, too.

Biswa96:
> Would you mind to copy the package along with your change as a new mingw-w64-db-agpl package?

I am not sure about that. python-bsddb3 and swi-prolog are both BSD licensed. Feel free to copy my changes and create a separate pull request.

PR #18162, opened 2023-08-13
libtommath: use libtool for building again

lazka:
This partially reverts #13790. Since libtool is now fixed since #18058 we can use the default makefile again for building. This alos gives us back the .pc file and leaves installation to the makefile.

Since the DLL name changed rebuild libtomcrypt.


lazka:
not sure why it's failing, maybe the CLI length limit. I'll try to move libtomcrypt to libtool too, maybe that changes things.

PR #18167, opened 2023-08-14
python-pip: Use python-build to build

Biswa96:
This fixes the following error with pip install command.

    //python.exe: can't open file '\\\\pip-script.py': [Errno 2] No such file or directory


Biswa96:
@alex-tee Thank you for reporting this issue.

Biswa96:
The actual issue was in python-setuptools (or in mingw cpython). It was fixed here https://github.com/msys2/MINGW-packages/pull/18180

PR #18168, opened 2023-08-14
cglm: split docs into separate package

Biswa96:
Also enable default packaging options


Biswa96:
For UCRT64,
* Before: Installed Size  : 14.23 MiB
* After: Installed Size  : 1524.25 KiB

PR #18169, opened 2023-08-14
libvterm: Update to 0.3.3

lazka:


Biswa96:
libvterm should be in depends list of neovim now. I shall add that in separate pull request with libtermkey and unibilium.

lazka:
oops, I missed that, thanks

PR #18170, opened 2023-08-14
python-tifffile: update to 2023.8.12

kmilos:


Biswa96:
Would it be possible to split the changes into separate commits?

kmilos:
> Would it be possible to split the changes into separate commits?

Sure. Single PR still ok I presume?

Biswa96:
> Single PR still ok I presume?

Yes.

kmilos:
Done.

PR #18083, opened 2023-08-13
onefetch: update to 2.18.1

MehdiChinoune:

PR #18182, opened 2023-08-14
onefetch: update windows-targets for gnullvm support

jeremyd2019:
Fixes https://github.com/msys2/MINGW-packages/pull/18083#issuecomment-1676330759
See https://github.com/o2sh/onefetch/issues/1143

Successful CLANGARM64 CI run: https://github.com/msys2-arm/MINGW-packages/actions/runs/5858975316/job/15884039932

### DIFF
--- a/mingw-w64-onefetch/PKGBUILD
+++ b/mingw-w64-onefetch/PKGBUILD
@@ -1,7 +1,7 @@
 _realname=onefetch
 pkgbase=mingw-w64-$_realname
 pkgname=$MINGW_PACKAGE_PREFIX-$_realname
-pkgver=2.17.0
+pkgver=2.18.1
 pkgrel=1
 pkgdesc="Git repository summary on your terminal (mingw-w64)"
 url="https://github.com/o2sh/onefetch"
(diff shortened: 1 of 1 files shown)


mingw-w64-onefetch/PKGBUILD
```diff
@@ -1,8 +1,8 @@
 _realname=onefetch
 pkgbase=mingw-w64-$_realname
 pkgname=$MINGW_PACKAGE_PREFIX-$_realname
-pkgver=2.17.0
-pkgrel=1
+pkgver=2.18.1
+pkgrel=2
```
Biswa96, line 5:
```suggestion
pkgrel=1
```

jeremyd2019: I bumped pkgrel after the revert of the revert :grin:

Biswa96: The current onefetch version is still 2.17.0-1 and this is a version bump.

jeremyd2019: But I believe 2.18.1-1 did exist in autobuild and/or staging for a while (until it was reverted after just clangarm64 failed to build) so it seemed safer to bump it to -2.

PR #18183, opened 2023-08-14
graphicsmagick: Fix writing TIFF images with transparency

mmuetzel:
The latest version of GraphicsMagick is stricter when it comes to reading TIFF images with transparency. It is now a requirement that `TIFFTAG_EXTRASAMPLES` is defined in the image file for the alpha channel:
http://hg.code.sf.net/p/graphicsmagick/code/rev/5c6fd2a5e177

Afaict, that is conform to the standard:
https://www.awaresystems.be/imaging/tiff/tifftags/extrasamples.html

However, the TIFF files with transparency that are created by GraphicsMagick itself don't write that tag (with default parameters).

Add a patch that writes the necessary tag for TIFF images with transparency with default parameters.

With this patch, Octave's test for `imwrite` passes again.

It is still a pity that GraphicsMagick now refuses to read files completely that it wrote itself in previous versions. As it stands, the alpha channel is lost.

### DIFF
--- a/mingw-w64-graphicsmagick/002-tiff-transparency.patch
+++ b/mingw-w64-graphicsmagick/002-tiff-transparency.patch
@@ -1,0 +1,19 @@
+# HG changeset patch
+# User Markus Mützel <markus.muetzel@gmx.de>
+# Date 1692044411 -7200
+#      Mon Aug 14 22:20:11 2023 +0200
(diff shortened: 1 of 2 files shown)


lazka:
Should we report this upstream?

mmuetzel:
> Should we report this upstream too?

Done: https://sourceforge.net/p/graphicsmagick/bugs/718/

Issue #18166, opened 2023-08-14
[go] missing go.env


AeliusSaionji:
### Description / Steps to reproduce the issue

On first install, the go environment variables are not all populated. In particular, GOPROXY is empty and this prevents downloading dependencies.

It appears a recent change in go expects a `go.env` in GOROOT?
**https://github.com/golang/go/issues/61928#issuecomment-1673658972**

archlinux appears to be inlcuding this file, https://gitlab.archlinux.org/archlinux/packaging/packages/go/-/commit/97e19acbc78f2b2a19f6ed25aee510669d70fc65 (thanks biswa96)

Reproduce build failures by...
1. `go clear -cache -modcache`
2. try to build something with a remote dependency.

### Expected behavior

Go environment populated with whatever defaults required to build packages.

### Actual behavior

Attempts to build packages fail as remote dependencies cannot be downloaded

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

MehdiChinoune:
Building android-tools fails:
```
  [184/488] Generating err_data.c
  FAILED: vendor/boringssl/crypto/err_data.c C:/_/B/src/build-CLANG64/vendor/boringssl/crypto/err_data.c 
  cmd.exe /C "cd /D C:\_\B\src\android-tools-33.0.3p2\vendor\boringssl\crypto\err && C:\msys64\clang64\bin\go.exe run err_data_generate.go > C:/_/B/src/build-CLANG64/vendor/boringssl/crypto/err_data.c"
  go: golang.org/x/crypto@v0.0.0-20210513164829-c07d793c2f9a: GOPROXY list is not the empty string, but contains no entries
```
I think the only thing that changed since latest successful build is "[go: Update to 1.21.0](https://github.com/msys2/MINGW-packages/commit/21426a8199a2c2fcaf8df5a77a70f1e73b257418)"

PR #18188, opened 2023-08-15
go: Include temporary fix for default environment variables

Biswa96:
Fixes https://github.com/msys2/MINGW-packages/issues/18166

### DIFF
--- a/mingw-w64-go/PKGBUILD
+++ b/mingw-w64-go/PKGBUILD
@@ -5,7 +5,7 @@ _realname=go
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.21.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Go Lang (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


Biswa96:
This is copied from ArchLinux https://gitlab.archlinux.org/archlinux/packaging/packages/go/-/commit/97e19acbc78f2b2a19f6ed25aee510669d70fc65

ognevnydemon:
it fixes a bit more issues regarding GOROOT variable

lazka:
> it fixes a bit more issues regarding GOROOT variable. or not?

I don't know much about golang, but `go env GOROOT` returns the same thing before and after this change.

ognevnydemon:
> > it fixes a bit more issues regarding GOROOT variable. or not?
> 
> I don't know much about golang, but `go env GOROOT` returns the same thing before and after this change.

anyway I can set it in this file, can't I?

Biswa96:
> it fixes a bit more issues regarding GOROOT variable. or not?

This is not related to GOROOT variable. go.env sets only GOPROXY, GOSUMDB and GOTOOLCHAIN variables. 

Though, we could add GOROOT in that go.env file. I am not sure about that. shell's rc file is another option.

PR #18196, opened 2023-08-16
starship: optdepend on nerd fonts

ognevnydemon:
needed for theme prompts

important question: can we use package group as dependency? (edit: for the answer read comments below https://github.com/msys2/MINGW-packages/pull/18196#issuecomment-1680036783)

### DIFF
--- a/mingw-w64-starship/PKGBUILD
+++ b/mingw-w64-starship/PKGBUILD
@@ -13,7 +13,7 @@ license=('spdx:ISC')
 depends=("${MINGW_PACKAGE_PREFIX}-libgit2")
 makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
              "${MINGW_PACKAGE_PREFIX}-cmake")
-optdepends=("${MINGW_PACKAGE_PREFIX}-nerd-fonts: Nerd Font Symbols preset")
+optdepends=("${MINGW_PACKAGE_PREFIX}-ttf-font-nerd: Nerd Font Symbols preset")
 options=('staticlibs' 'strip')
 source=("${_realname}-${pkgver}.tar.gz::https://github.com/starship/starship/archive/refs/tags/v${pkgver}.tar.gz"
         "windows-0.39.0-gnullvm.patch"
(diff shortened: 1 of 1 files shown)


Biswa96:
> can we use package group as dependency?

optdepends just show a message at the time of installing a package. One can write anything in optdepends. User has to decide what to install.

lazka:
> important question: can we use package group as dependency?

no

ognevnydemon:
> > important question: can we use package group as dependency?
> 
> no

I meant optdepend

lazka:
> I meant optdepend

not sure, it should also point to real packages, but I don't know if pacman will fail if it isn't.

ognevnydemon:
can you test it in staging?

lazka:
I think the "proper" fix would be to remove the group from nerd-fonts, add an additional package to nerd-fonts, which is just named "<...>-nerd-fonts" and depends on all the other packages (so like a meta package).

ognevnydemon:
> I think the "proper" fix would be to remove the group from nerd-fonts, add an additional package to nerd-fonts, which is just named "<...>-nerd-fonts" and depends on all the other packages (so like a meta package).

ok. I hope I'll try it in Sunday

ognevnydemon:
won't this conflict?

lazka:
> won't this conflict?

With what?

I see that Arch solves this via a provides: https://archlinux.org/packages/extra/any/ttf-profont-nerd/

ognevnydemon:
> With what?

sorry, at first I didn't understand you right. now I'm sure that it won't conflict

ognevnydemon:
> I see that Arch solves this via a provides (ttf-font-nerd), which is also an option: https://archlinux.org/packages/extra/any/ttf-profont-nerd/

in that case we can left package group as is, and just use this provides, right?

lazka:
> in that case we can left package group as is, and just use this provides, right?

if it's named the same then the group will be ignored by pacman, otherwise sure. Though we have made the experience that groups just confuse users at installation time, so at least I kinda avoid them.

ognevnydemon:
oops, there are some merge conflicts... I'm not pro in git, how can I resolve it without `merge branch...` commit?

ognevnydemon:
> I see that Arch solves this via a provides (ttf-font-nerd), which is also an option: https://archlinux.org/packages/extra/any/ttf-profont-nerd/

done in #18228

ognevnydemon:
should be mergable now

PR #18197, opened 2023-08-16
Update mingw-w64

lazka:
0003-headers-add-default-value-xaudio2.patch: included upstream now


lazka:
@Biswa96 ok? just double checking because of the dropped patch https://github.com/mingw-w64/mingw-w64/blob/e89de847dd3e05bb8e46344378ce3e124f4e7d1c/mingw-w64-headers/include/xaudio2.h#L1579C15-L1586

Biswa96:
Yeah, that patch was added tomorrow.

Issue #13608, opened 2022-10-16
cmake: sometimes fails to parse implicit include dirs


lazka:
### Description

This is a follow up from #10761

If cmake fails to parse the gcc output of implicit include dirs, it will add the system paths via `-isystem`, which is not supported by GCC and will result in something like:

```c
C:/msys64/mingw64/include/c++/11.2.0/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
   75 | #include_next <stdlib.h>
      |               ^~~~~~~~~~
```

I've noted the packages for which this happens in CI recently:

* mingw-w64-qt6-3d
* mingw-w64-qt6-webview
* mingw-w64-aws-crt-cpp
* qt-creator
* qt6-webchanne
* qt6-webchannel
* openshadinglanguage

so in all cases cmake+ninja.

I've had a similar issue happen due to a ninja issue, see https://github.com/msys2/MINGW-packages/pull/12262.
There you could see in "CMakeOutout.log" `you should find a warn: unable to parse implicit include dirs!`

My guess is that cmake calls gcc via ninja, and stderr/stdout gets re-ordered sometimes.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22621

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Expected behavior

It builds

### Actual behavior

It fails, sometimes

### Repro steps

unclear

mmuetzel:
That might be related to https://github.com/msys2/MSYS2-packages/pull/3259
Maybe, those additional include paths are messing with the order of `isystem` paths?

MehdiChinoune:
> That might be related to [msys2/MSYS2-packages#3259](https://github.com/msys2/MSYS2-packages/pull/3259) Maybe, those additional include paths are messing with the order of `isystem` paths?

It affects only UCRT64, so I don't think It's related.

lazka:
If someone knows a way to reproduce this locally (even if only in some cases), please let us know. I've only seen it in CI up until now.

Biswa96:
I have reproduced this issue once but forgot which package. The issue disappears after deleting pkg and src directories. I ran ninja multiple times before deleting the build directory and it fails with same compiler error.

[...]

PR #18210, opened 2023-08-17
glslang: Enable optimization support

kasper93:


mingw-w64-glslang/PKGBUILD
```diff
@@ -19,16 +19,19 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-python")
 source=(${_realname}-${pkgver}.tar.gz::https://github.com/KhronosGroup/glslang/archive/${pkgver}.tar.gz
         001-respect-destdir-for-compat-symlink.patch::https://github.com/KhronosGroup/glslang/commit/171a3220.patch
-        002-remove-deprecated-mingw-secure-api.patch)
+        002-remove-deprecated-mingw-secure-api.patch
+        https://github.com/KhronosGroup/glslang/commit/ffd07992f1b1f9e8a609338973c5beaa53d7bfa9.patch)
```
Biswa96, line 23: This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository. If the patch is not in main glslang repository, you can add the patch file in this repository. And add the PR link as a comment above `patch` command.

Biswa96:
Some questions.
1. Should glslang be linked with packaged version of spirv-tools? The glslang repository uses different version of spirv-tools than ours.
2. Should we wait for that patch to be merged in upstream?
3. Does this require to re-build dependent packages?

mingw-w64-glslang/PKGBUILD
kasper93: Ok, let's wait a bit, it should be merged soon https://github.com/KhronosGroup/glslang/pull/3152

kasper93:
> Should glslang be linked with packaged version of spirv-tools? The glslang repository uses different version of spirv-tools than ours.

So for full compatibility both packages in MSYS2 should have compatible versions. Which are tracked with https://github.com/KhronosGroup/glslang/blob/main/known_good.json

I tested the current versions works correctly. But indeed the `known_good.json` is something to keep in mind when making spit packages.

> Should glslang be linked with packaged version of spirv-tools?

Yes, especially that we provide static libraries too.

kasper93:
> Should glslang be linked with packaged version of spirv-tools? The glslang repository uses different version of spirv-tools than ours.

So for full compatibility both packages in MSYS2 should have compatible versions. Which are tracked with https://github.com/KhronosGroup/glslang/blob/main/known_good.json

I tested the current versions works correctly. But indeed the `known_good.json` is something to keep in mind when making spit packages.

> Should glslang be linked with packaged version of spirv-tools?

Yes, especially that we provide static libraries too.

> Should we wait for that patch to be merged in upstream?

Yes, sure let's do that.

> Does this require to re-build dependent packages?

Yes, if they are linking to static version. for shared one shouldn't be a difference. But I rebuild all just in case.

kasper93:
> Should we wait for that patch to be merged in upstream?

Merged now, updated with the version of a patch from main branch.

kasper93:
```
   D:/M/msys64/ucrt64/include/c++/13.2.0/cstdlib:79:15: fatal error: stdlib.h: No such file or directory
     79 | #include_next <stdlib.h>
        |               ^~~~~~~~~~
  compilation terminated.
```
huh?

lazka:
> huh?

#13608 - a retry usually fixes it

kasper93:
> #13608 - a retry usually fixes it

I see. I needed to update vulkan-validation-layers, because build failed with vulkan-headers from the repo.

lazka:
you also need to update the checksum.

lazka:
hm, it was revered https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/e6bdb8d71409a96a4174589ea195d0dc1e920625

kasper93:
> hm, it was revered

Yes, I see that, but it still doesn't compile. There are a lot of dependencies between those vulkan packages. I'm trying to unwind what can be done here.

lazka:
with the revert I get lots of other errors, it seems like we need to update SPIRV-Tools also maybe

kasper93:
> good luck, I'm out :)

For now the only way is to revert vulkan headers update and wait for approprieate spirv-tools tag.

https://github.com/msys2/MINGW-packages/pull/18212

lazka:
I guess we can't get ucrt64 to pass here :/

good to merge besides that?

kasper93:
> good to merge besides that?

good on my end.

PR #18212, opened 2023-08-17
Revert "vulkan-headers: update to 1.3.261"

kasper93:
VVL 1.3.255 cannot be build with vulkan-headers 1.3.261. VVL 1.3.261 cannot be build with SPIR-V Tools 1.3.250.0 which is currently latest release. So revert header and loader update to bring back the compatibility.

Those 4 packages have to be updated together, because they have dependencies between each other and not all of them have the same release cycle.
SPIR-V Tools
glsllang
VVL
vulkan-headers

### DIFF
--- a/mingw-w64-vulkan-headers/PKGBUILD
+++ b/mingw-w64-vulkan-headers/PKGBUILD
@@ -4,7 +4,7 @@
 _realname=Vulkan-Headers
 pkgbase=mingw-w64-vulkan-headers
 pkgname=("${MINGW_PACKAGE_PREFIX}-vulkan-headers")
-pkgver=1.3.261
+pkgver=1.3.256
 pkgrel=1
 pkgdesc='Vulkan header files (mingw-w64)'
 arch=('any')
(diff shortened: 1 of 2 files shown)


lazka:
@MehdiChinoune since you updated them last, any objections?

MehdiChinoune:
I prefer to build spirv-tools from git to fix building VVL instead of reverting those updates.

lazka:
sounds good, they are already in the RC phase anyway

MehdiChinoune:
> sounds good, they are already in the RC phase anyway

They use two tag variants, 1.x.y (1.2.250.1) and YYYY.X (2023.2), We were tracking the later until I changed last time https://github.com/msys2/MINGW-packages/commit/74a618ee93fb3c5edcd8c4e4b7c89423f2c42596
I don't know what to do now with this new tag 2023.4, should we revert to YYYY.X or wait for the next 1.x.y tag?
SPIRV-HEADERS don't have these YYYY.X tags!

lazka:
yeah, I was wondering that myself. According to https://repology.org/projects/?search=spirv-tools&maintainer=&category=&inrepo=&notinrepo=&repos=&families=&repos_newest=&families_newest= most distros use the YYYY.X variant here.

kasper93:
> > sounds good, they are already in the RC phase anyway
> 
> They use two tag variants, 1.x.y (1.2.250.1) and YYYY.X (2023.2), We were tracking the later until I changed last time [74a618e](https://github.com/msys2/MINGW-packages/commit/74a618ee93fb3c5edcd8c4e4b7c89423f2c42596) I don't know what to do now with this new tag 2023.4, should we revert to YYYY.X or wait for the next 1.x.y tag? SPIRV-HEADERS don't have these YYYY.X tags!

They are tracking two release cycles. The YYYY.X is kinda internal for the project. The 1.3.250 is tided to Vulkan version and Vulkan SDK version as seen here https://vulkan.lunarg.com/sdk/home

PR #18213, opened 2023-08-18
fdk-aac: rebuild for libtool update

raedrizqie:


lazka:
lgtm

just curious: does this change anything?

raedrizqie:
nothing is changed except now we are using mingw libtool

lazka:
From what I understand there is no difference in the context of autotools, only when libtool is called directly, since with autotools all files get re-generated anyway for the respective target. But maybe I'm missing something.

But can't hurt I guess.

raedrizqie:
just want to clean this list.. https://packages.msys2.org/package/libtool?repo=msys&variant=x86_64

msys libtool should only be used for msys packages imo :)

lazka:
oh, I missed that. good reason :)

PR #18214, opened 2023-08-18
jbigkit: rebuild for libtool update

raedrizqie:


raedrizqie:
this is actually a rebuild for clangarm64, to switch to mingw libtool

PR #18215, opened 2023-08-18
Update vulkan-validation-layers/spirv-tools to 1.3.261/2023.4rc2

MehdiChinoune:
@kasper93

### DIFF
--- a/mingw-w64-spirv-tools/PKGBUILD
+++ b/mingw-w64-spirv-tools/PKGBUILD
@@ -4,8 +4,11 @@
 _realname=spirv-tools
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-epoch=1
-pkgver=1.3.250.0
+epoch=2
+_pkgver=2023.4
+_rc=rc2
+_tag=${_pkgver}.${_rc}
+pkgver=${_pkgver}${_rc}
 pkgrel=1
 pkgdesc="API and commands for processing SPIR-V modules (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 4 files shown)


lazka (approved):
lgtm

mingw-w64-spirv-tools/PKGBUILD
```diff
@@ -19,15 +22,8 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-ninja"
              "${MINGW_PACKAGE_PREFIX}-python"
              "${MINGW_PACKAGE_PREFIX}-spirv-headers")
-source=("${_realname}-${pkgver}.tar.gz::https://github.com/KhronosGroup/SPIRV-Tools/archive/sdk-${pkgver}.tar.gz"
-        002-Do-not-statically-link-to-the-C-runtime-with-mingw.patch)
```
lazka, line 23: oh, was that dropped on purpose? the code still seems to be there upstream

MehdiChinoune: I will take a look.

kasper93:
I'm not a big fan of switching to RC (or beta) versions for main repo. But if it works and you are ok with it, LGTM too.

kasper93:
> Better than reverting to previous version.

Why? What's wrong with stable and tested version?

lazka:
@MehdiChinoune hm?

MehdiChinoune:
What?

lazka:
Would it be a good idea to follow the sdk release versions in the future (https://vulkan.lunarg.com/sdk/latest.json), to avoid similar problems? Or are there any reasons why we should update single components independently?

MehdiChinoune:
Should we revert all?

kasper93:
> Would it be a good idea to follow the sdk release versions in the future (https://vulkan.lunarg.com/sdk/latest.json), to avoid similar problems? Or are there any reasons why we should update single components independently?

Vulkan SDK releases are pretty slow. We can update those packages sooner. I would just stick to some stable tags and if something is not building correctly just wait for new release instead of chasing latest version possible.

> Should we revert them all?

IMHO, at this point keep it as is.

PR #18218, opened 2023-08-18
zlib: update to 1.3

MehdiChinoune:
06-add-64bit-functions-and-fix-typos.patch doesn't apply, I don't know what issue does it fix!


lazka:
yeah, weird patch, probably best to drop it. There have been upstream fixes for that code recently https://github.com/madler/zlib/commits/develop/contrib/minizip so if something comes up we should report it upstream.

Issue #18226, opened 2023-08-20
[libtiff] tif_hash.h is missing


joda01:
### Description / Steps to reproduce the issue

When I try to compile my project with libtif included I got the compiler error

```
C:/msys64/mingw64/include/tiffiop.h:49:10: fatal error: tif_hash_set.h: No such file or directory                                                                 if\image_loader_tif.hpp:15,
   49 | #include "tif_hash_set.h"                                                if\image_loader_tif.cpp:14:
      |          ^~~~~~~~~~~~~~~~        
```` 

### Expected behavior

`tif_hash_set.h` be part of `C:/msys64/mingw64/include`

### Actual behavior

`tif_hash_set.h` ìs not installed in `C:/msys64/mingw64/include`

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

Yes

Biswa96:
Thank you for reporting this issue. This seems to be an upstream issue. nix packages uses a patch https://github.com/NixOS/nixpkgs/tree/release-23.05/pkgs/development/libraries/libtiff

@kmilos Should this be reported and fixed in upstream?

lazka:
@joda01 Note that upstream considers tiffiop.h a private header, so you shouldn't use it ideally.

joda01:
@lazka Thank's for your hint you're right it's a private include. So it's not a bug but an expected behaviour!

lazka:
There might be a valid reason to still use it, since others are patching it too. I don't know the reason though.

PR #18227, opened 2023-08-20
libtiff: Add missing tif_hash_set.h

Biswa96:
Also move the extra headers install logic to separate patch file

Fixes https://github.com/msys2/MINGW-packages/issues/18226


Biswa96:
The patch file is copied from nixpkgs https://github.com/NixOS/nixpkgs/tree/release-23.05/pkgs/development/libraries/libtiff. Thanks to the nix developers ❤️

PR #18228, opened 2023-08-20
nerd-fonts: add missing fonts and improve recipe

ognevnydemon:
I left package group as is. if it confuses someone I'll remove it. also added missing fonts


brisingraerowing:
From looking at the other fonts in the PKGBUILD, you're using the wrong license directories for the ones you added. They need to be under the font name, not the package name.

ognevnydemon:
> From looking at the other fonts in the PKGBUILD, you're using the wrong license directories for the ones you added. They need to be under the font name, not the package name.

thanks, fixed it

Issue #9045, opened 2021-02-04
boost: context lacking asm implementation for MinGW ARM/ARM64


jeremyd2019:
This is pretty expected, but entering here to record it.  It seems that context is still lacking an arm64 implementation for Windows, even with MSVC.

Biswa96:
Do I have to I compile all the dependencies (e.g. Boost::assert, Boost::config, Boost::core etc.) for that target CPU architecture before porting boost context?

olk:
Only the libraries boost.context depends on. If you compile boost.context
with the appropriate boost.build properties (architecture, binary format,
abi etc.) then all these libs are compiled automatically.

Biswapriyo Nath ***@***.***> schrieb am Do., 8. Juli 2021,
08:02:

dennisameling:
> It seems that context is still lacking an arm64 implementation for Windows, even with MSVC.

FYI the MSVC version is available since a month or two: https://github.com/boostorg/context/pull/201

olk:
> > It seems that context is still lacking an arm64 implementation for Windows, even with MSVC.
> 
> FYI the MSVC version is available since a month or two: [boostorg/context#201](https://github.com/boostorg/context/pull/201)

#201 merged

mstorsjo:
I was pointed here... I'm not familiar with the libraries here or how to test them, but it looks like it shouldn't be too hard to make it build.

For mingw arm64 cases, it uses GAS style assembly, and there are very few ABI differences to generic AAPCS on other platforms. In many cases, ELF GAS work as is, or with very few modifications. There seems to be a few ELF specific assembly directives in those files; if they are wrapped with `#ifdef __ELF__`, I think it should work just fine. Looking at https://github.com/boostorg/context/blob/697299ea554c04ae8a5cfb1b5cc3ec9caf61724d/src/asm/jump_arm64_aapcs_elf_gas.S, the `.type`, `.size` and `.section` directives aren't needed/supported for COFF targets, so by wrapping them it should build.

That is, whatever ASM files are picked for the MSVC case, `*_pe_armasm.asm`, use the corresponding `*_elf_gas.S` files instead, and add ifdefs to them. But before doing that, it's kinda important to check that there aren't any functional differences in the assembly itself between the `_elf_gas.S` and `_pe_armasm` files; check with diff that the only differences are in the directives wrapping it, not the instructions themselves.

[...]

PR #18229, opened 2023-08-20
boost: update to 1.83.0

MehdiChinoune:
@jeremyd2019 Maybe you could extend https://github.com/boostorg/context/commit/697299ea554c04ae8a5cfb1b5cc3ec9caf61724d for clangarm64!


jeremyd2019:
That's #9045 - I have been thinking about that.  I don't know of anything being blocked by this lack so it hasn't been a priority.  Let me know if it is blocking anything and I will consider reprioritizing.

PR #18230, opened 2023-08-21
fastfetch 2.0.2

CarterLi:
Also removed dependency cjson, which is no longer used.


CarterLi:
I think I encountered the [gcc bug 85282](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282). Because I don't test gcc on Windows personally. I decide to disable gcc toolchains.

MehdiChinoune:
> I decide to disable gcc toolchains.

I don't think, You can do that.
An old version available for all environments is better than the latest version for just some environments.
UCRT64 is the primary/main environment.

CarterLi:
@MehdiChinoune Fixed

PR #18231, opened 2023-08-21
Fetch less crates in prepare step

mati865:


ognevnydemon:
hm, can we set "${CARCH}-pc-windows-gnu" as default value and just override it for CLANGARM64?

mmuetzel:
Remove the spaces around the `=` when assigning values to environment variables.
Nit: Use same indentation in `if` and `else` branches.

mati865:
> Remove the spaces around the = when assigning values to environment variables.

Oof, I haven't wrote shell scripts for so long I forgot about the spaces (I tried to daily drive Nushell for a long time) .

> Nit: Use same indentation in if and else branches.

I should've spent some time to configure some editor instead of grabbing VS Code with only few extensions installed...

ognevnydemon:
> hm, can we just set "${CARCH}-pc-windows-gnu" as default value and just override it for CLANGARM64?

I have a strange idea: can we just write `_tatget+="llvm"` instead of `_target="${CARCH}-pc-windows-gnullvm"`?

Biswa96:
> can we just write `_tatget+="llvm"` instead of `_target="${CARCH}-pc-windows-gnullvm"`?

The current logic appears to be more straightforward.

mati865:
> > hm, can we just set "${CARCH}-pc-windows-gnu" as default value and just override it for CLANGARM64?
> 
> I have a strange idea: can we just write `_tatget+="llvm"` instead of `_target="${CARCH}-pc-windows-gnullvm"`?

I had considered it but I think it would be worse from maintenance POV. It would save one repetition of short string but as it currently is obvious. just take a look and within milliseconds you know what it does.
Ofc I'll change it if that is preferred by the majority.

I have bigger issue that I haven't had time to spend on yet, when building `sccache` Cargo for weird some reason wants to download more crates than during the fetch...

mati865:
What causes CI failure: when checking if all dependencies are present `cargo build` seems to require code for all of them instead of only enabled ones. It'll filter them later to build only required ones.
Looks like this is just the way Cargo works (or there is a bug) and this cannot be changed by some options.

PR #18234, opened 2023-08-22
hdf5: update to 1.14.2

mkitti:


Biswa96:
This seems to happen after git update to 2.42.0. The ci-build.sh script fails at `git remote add upstream` command. Apology for this inconvenience. Please wait.

Biswa96:
The git update was reverted and the CI is green again.

PR #18240, opened 2023-08-22
python-cryptography: drop patch disabling abi3

lazka:
it should work now after https://github.com/msys2/MINGW-packages/pull/18238


MehdiChinoune:
We could also build pyside6 with Py_LIMITED_API https://github.com/msys2/MINGW-packages/blob/a7a713ce1562ece2006448680176bf6099109876/mingw-w64-pyside6/PKGBUILD

lazka:
> We could also build pyside6 with Py_LIMITED_API https://github.com/msys2/MINGW-packages/blob/a7a713ce1562ece2006448680176bf6099109876/mingw-w64-pyside6/PKGBUILD

thanks, yeah. This will likely need more patching though, because pyside6 hardcodes MSVC file names.

lazka:
> > We could also build pyside6 with Py_LIMITED_API https://github.com/msys2/MINGW-packages/blob/a7a713ce1562ece2006448680176bf6099109876/mingw-w64-pyside6/PKGBUILD
> 
> thanks, yeah. This will likely need more patching though, because pyside6 hardcodes MSVC file names from what I can see.

On a second thought, the main reason the limited API exists is to allow wheel builds for pypi, which we don't do, so we don't have much to gain from it unless it allows us to remove patches.

lazka:
true, that as well. Though I personally would try to rebuild it anyway to make sure things are still building fine after a major Python update.

PR #18245, opened 2023-08-23
python-maturin: Update to 1.2.3

lazka:
Remove clangarm64 workaround, should no longer be needed since windows-targets is new enough now.

Also enables the default feature set, so commands like "new" and "init" are now available, as if built with cargo.

### DIFF
--- a/mingw-w64-python-maturin/PKGBUILD
+++ b/mingw-w64-python-maturin/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=maturin
 pkgbase=mingw-w64-python-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-python-${_realname}
-pkgver=1.1.0
-pkgrel=4
+pkgver=1.2.3
+pkgrel=1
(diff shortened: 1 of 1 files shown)


jeremyd2019:
```
  thread 'main' panicked at '"C:\\Users\\Administrator\\.cargo\\registry\\src\\index.crates.io-6f17d22bba15001f\\ring-0.16.20\\pregenerated\\aesv8-armx-linux64.obj": Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }', C:\Users\Administrator\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ring-0.16.20\build.rs:769:10
```

I was afraid of that when I saw `rustls`, as I expected that to pull in ring.  briansmith/ring#1614 for windows-on-arm64 support (don't know if the support for -msvc will also have straightened out -gnullvm or not)

lazka:
thanks, I disabled that feature again

jeremyd2019:
It sounds like ring 0.17.0 was released with windows-on-arm64 support, so hopefully once that (semver) version makes its way through the ecosystem some of these things that were failing due to ring can be re-enabled.

PR #18254, opened 2023-08-24
7zip: upgrade to 23.01

dscho:
For release notes, see
https://sourceforge.net/p/sevenzip/discussion/45797/thread/3f550826d8/


lazka:
the comment suggests just dropping the patch, BUT then there also is https://github.com/msys2/MINGW-packages/pull/13928#pullrequestreview-1169465373

dscho:
> the comment suggests just dropping the patch, BUT then there also is [#13928 (review)](https://github.com/msys2/MINGW-packages/pull/13928#pullrequestreview-1169465373)

Right, I noticed that the first hunk might no longer be needed, so I removed it, but the others seem to be relevant, still.

PR #18265, opened 2023-08-25
binutils: backport patch "optimize handle_COMDAT"

### LINKED ISSUE #7660: Extremely slow linking with ld, taking several minutes
I apologize if this is the wrong place for this report. In this case I would be grateful to be pointer to a more appropriate one.

We have a pretty big C++ project built since several years with the MINGW toolchain on Windows. During the last years, our builds took always longer, and we could pinpoint it last year to the linker needing extremely long times, far above one minute, even 9 minutes in the most extreme case. It impacts in particular the executables containing unit tests, slowing down the local iterations for the project developers. The situation is degrading with time, probably due to the increasing code size.

As an experiment, we built the project with lld from the clang toolchain, and got immediately much smaller link times, typically below one second. Just a handful of executables need a few seconds at most to link. 

From my online research, we are not the only one to have performance issues with ld on Windows, but I didn't find any report at this scale. It looks like we hit a pathological case. I cannot share the code base, and since the problem is obviously related to the code size, I cannot provide a minimal example. But I would be ready to provide more information and spend time investigating the issue if it helps, even though we are probably going to switch this project to lld.

oltolm:
Makes `ld` faster. Should fix #7660.

### DIFF
--- a/mingw-w64-binutils/PKGBUILD
+++ b/mingw-w64-binutils/PKGBUILD
@@ -6,7 +6,7 @@ _realname=binutils
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.41
-pkgrel=1
+pkgrel=2
 pkgdesc="A set of programs to assemble and manipulate binary and object files (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


mmuetzel:
Thanks for your patch and the PR.
Does the upstream commit apply directly to the latest release tarball? In that case, it might be better to get it from there instead of duplicating it here:
https://sourceware.org/git/?p=binutils-gdb.git;a=patch;h=6aadf8a04d162feb2afe3c41f5b36534d661d447;hp=fb9b7fbf17f50fcfabf6e3d7d06a93e1f17c52b7

If there is a reason to keep the copy of the patch in this repository, please add a comment with a link to the respective commit from upstream.

Biswa96:
> In that case, it might be better to get it from there instead of duplicating it here

Once lazka mentioned that patch checksum changes due to the git version at the end of patch file.

lazka:
Maybe this works? https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff_plain;h=6aadf8a04d162feb2afe3c41f5b36534d661d447

oltolm:
I updated the PR.

lazka:
binutils master just gained some follow up fixes which seem related: https://sourceware.org/git/?p=binutils-gdb.git

oltolm:
I pushed another commit.

lazka:
Thanks :) I'd prefer to wait a few days to see if there'll be more fixes upstream.

mmuetzel:
It's been a week since the last update. 
I'm not following binutils development. Have there been any updates?
Could this be merged? Or should we wait a bit longer?

PR #18269, opened 2023-08-26
python-poetry: Update to 1.6.1

lazka:


lazka:
I'm gonna update all these in #18270 instead

PR #18276, opened 2023-08-27
Remove redundant dependency

jeroen:
See comment here: https://github.com/msys2/MINGW-packages/pull/18000#issuecomment-1694662796


raedrizqie:
you forgot to bump the pkgrel

jeroen:
I don't think that is needed? Nothing will change really, the dependency was not used right?

sskras:
Having two different configuration of the package under the same version would confuse users who report the bug / does QA testing.

lazka:
> Having two different configurations of a package published under the same version would confuse users who report bugs / does QA testing, IMO.

without pkgrel bump nothing will reach the user

PR #18286, opened 2023-08-30
libavif: update to 1.0.1

kmilos:
Deps need patching/update because of API change and/or `SameMajor` CMake version matching policy.

### DIFF
--- a/mingw-w64-SDL2_image/PKGBUILD
+++ b/mingw-w64-SDL2_image/PKGBUILD
@@ -4,7 +4,7 @@ _realname=SDL2_image
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
 pkgver=2.6.3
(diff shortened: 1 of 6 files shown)


lazka:
Arch completed the rebuild here: https://archlinux.org/todo/libavif-10/ see the packages for related patches they apply.

kmilos:
IMHO there are some gd and sdl2 patches missing.

darktable and kimageformats are patched upstream.

kmilos:
> Should we switch to `CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION=ON` while at it?

Good idea, done.

MehdiChinoune:
The recommended way to add SOVERSION to library name is using [CMAKE_DLL_NAME_WITH_SOVERSION](https://cmake.org/cmake/help/latest/variable/CMAKE_DLL_NAME_WITH_SOVERSION.html)

PR #18300, opened 2023-09-01
python-jupyter* updates

raedrizqie:
depends on #18288


raedrizqie:
rdeps check failed at
```
Pip Output (without reverse dependencies): mingw-w64-ucrt-x86_64-python-poetry-1.6.1-2-any.pkg.tar.zst
  Pip Check Failed for mingw-w64-ucrt-x86_64-python-poetry
Here is what it failed.
build 0.10.0 requires colorama, which is not installed.
```

colorama is actually an optdepend in python-build, but it is listed under dependencies here:
https://github.com/pypa/build/blob/1fff01ed7440d5b5626fec31422026a4920b7764/pyproject.toml#L38

Biswa96:
Would you like to force push the changes in a clean branch? Some commits like jupyter-events and jupyter-nbconvert updates are shown twice.

raedrizqie:
ok, will do later when im home

raedrizqie:
@Biswa96 rdeps check failed on spyder.. should i also update spyder?

PR #18312, opened 2023-09-04
openblas: update to 0.3.24

MehdiChinoune:
0.3.24 now supports Flang, but unfortunately Flang doesn't fully support OpenMP.
So We either build LAPACK with Flang and disable OpenMP or leave as It is (Build C_LAPACK)


mmuetzel:
OpenBLAS supports Flang 17 or newer. Until that is released and distributed by MSYS2, we should continue to build the C version on the LLVM environments.

PR #18316, opened 2023-09-04
rpds-py: update to 0.10.2, fix crates' fetching

ognevnydemon:


Biswa96:
++pkgrel

raedrizqie:
while at it, update to v0.10.2

ognevnydemon:
> while at it, update to v0.10.2

hm, should I close my PR then?

ognevnydemon:
> why did you delete? you just need to update it

oops... I'm hurried a bit, sorry

raedrizqie:
you forgot to update the sha256sums

ognevnydemon:
I messed some things up... will this have bad effect?

ognevnydemon:
@raedrizqie everything is fine now?

mingw-w64-python-rpds-py/PKGBUILD
```diff
@@ -17,14 +17,18 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-python-build"
...
   rm -rf python-build-${MSYSTEM} | true
   cp -r "${_realname/-/_}-${pkgver}" "python-build-${MSYSTEM}"
 
   cd "${srcdir}/python-build-${MSYSTEM}"
-  cargo fetch --locked --target "${CARCH}-pc-windows-gnu"
+  local _target="${CARCH}-pc-windows-gnu"
+  if [[ $MINGW_PACKAGE_PREFIX == *-clang-aarch64 ]]; then
+    _target="${CARCH}-pc-windows-gnullvm"
+  fi
+  cargo fetch --locked --target ${_target}
```
Biswa96, line 31: Is the `--target` option needed here? @mati865 tried that in this pull request https://github.com/msys2/MINGW-packages/pull/18231

mati865: It turned out in https://github.com/msys2/MINGW-packages/pull/18231 `cargo fetch --target x` will download correct dependencies but `cargo build` might want dependencies for other targets. So using `cargo fetch --target x` can cause `cargo build --release` to fail if **`--offline`** is added as well.

ognevnydemon: good catch. so there is no sense in it and I can just remove it?

raedrizqie: i just follow how Arch done it in their PKGBUILD: https://gitlab.archlinux.org/archlinux/packaging/packages/python-rpds-py/-/blame/main/PKGBUILD#L35

ognevnydemon: me too when I made [alacritty](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-alacritty#L34) recipe

mati865: @raedrizqie yeah, I tried the same in https://github.com/msys2/MINGW-packages/pull/18231 for other crates but as it turns out, this either doesn't work well on Windows or doesn't work well in general.

ognevnydemon: anyway we can left it now as is for this package

PR #18320, opened 2023-09-05
scrcpy: Help scrcpy find the server and icon file

### LINKED ISSUE #17778: scrcpy can't find /share/scrcpy-server file

1. ensure that device is connected using `adb devices`
2. execute `scrcpy`

### Expected behavior

program starts properly

### Actual behavior

scrcpy-server exists but program can't detect it
```console
$ scrcpy
scrcpy 2.1.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  2C050DLH2000WA                  device  Pixel_7
stat: No such file or directory
ERROR: '/clang64/share/scrcpy/scrcpy-server' does not exist or is not a regular file
ERROR: Server connection failed
```

### Verification

### Windows Version

MINGW64_NT-10.0-22621

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

_No response_

StableAgOH:
Fix #17778.

Scrcpy uses `PREFIX+"/share/scrcpy/scrcpy-server"` as the default server path, and `PREFIX` is `MINGW_PREFIX` (determined at build time). Before using the server file, scrcpy checks to see if the path to the server file is legal. Since the default server path is an MSYS2 path and not a Windows path, the path must not be legal, and scrcpy will not be able to find the server file.

Scrcpy implements a function (`sc_file_get_local_path`) that gets the program directory and allows you to concatenate a string after the path. I used it to splice the program directory with the relative path to the server file as the server path, which solved the problem.

Same for icon file.


Biswa96:
Thank you for the contribution. I was planning to use pathtools for scrcpy like we did with curl, openssl and other packages. But this also works, I have verified with the ucrt64 package in my system.

Would you like to bump the pkgrel value please?

StableAgOH:
> Thank you for the contribution. I was planning to use pathtools for scrcpy like we did with curl, openssl and other packages. But this also works, I have verified with the ucrt64 package in my system.
> 
> Would you like to bump the pkgrel value please?

Of cource, done.

PR #18325, opened 2023-09-06
Update gst-plugins-rs to 0.11.0

realmazharhussain:


realmazharhussain:
It seems like 32-bit builds require more RAM than what is currently available. Does anyone have any suggestions?

Biswa96:
> It seems like 32-bit builds require more RAM than what is currently available. Does anyone have any suggestions?

32 bit packages can be removed if it can not be built.

PR #18396, opened 2023-09-06
blosc2: update to 2.10.2

MehdiChinoune:


Biswa96:
The patch file was not removed.

PR #18410, opened 2023-09-07
lsd: update to 1.0.0

ognevnydemon:
notable changes:
- changed upstream URL
- nerd-fonts are now in repo, uncomment optdepends
- install license in correct directory


ognevnydemon:
arm64 patch still applies, but I would prefer to have an extra check

Biswa96:
> * disabled check() because it fails

FYI, it is not required to disable, one can build the package without the check function using `makepkg-mingw --nocheck`

ognevnydemon:
> > * disabled check() because it fails
> 
> FYI, it is not required to disable, one can build the package without the check function using `makepkg-mingw --nocheck`

thanks

PR #18417, opened 2023-09-07
python-build: Update to 1.0.3

lazka:


MehdiChinoune:
Why colorama wasn't added as a dependency?

raedrizqie:
yes, i think we should patch it out, but keep it as optdepends..

lazka:
I've created #18423

PR #18470, opened 2023-09-12
mpc,mpfr: enable static libs

jeroen:
I am building a static gcc toolchain, but these static libraries are missing. Is there any reason they are disabled? I think there is no harm in adding them, they seem to be included already for libgmp (which is part of this family of libs).

### DIFF
--- a/mingw-w64-mpc/PKGBUILD
+++ b/mingw-w64-mpc/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mpc
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.3.1
-pkgrel=1
+pkgrel=2
 pkgdesc="Multiple precision complex arithmetic library (mingw-w64)"
 url='http://www.multiprecision.org'
 license=('LGPL')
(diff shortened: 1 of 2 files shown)


ognevnydemon:
by the way, are these static libraries needed for `gmp-mpfr-sys` crate in Rust? for now it's recompiling

ognevnydemon:
> by the way, are these static libraries needed for `gmp-mpfr-sys` crate in Rust? for now it's recompiled

yes! I've just checked out and now compile time of my project reduced from 10 minutes to 30 seconds

lazka:
I've diffed the packages and don't see any differences for the shared part, so fine I guess.

mmuetzel:
Are the concerns from, e.g., https://github.com/msys2/MINGW-packages/pull/14146#issuecomment-1320422207 no longer an issue?

lazka:
> Are the concerns from, e.g., [#14146 (comment)](https://github.com/msys2/MINGW-packages/pull/14146#issuecomment-1320422207) no longer an issue?

Maybe. But this doesn't change the shared part, so I'd assume it doesn't affect us, and still somehow works for @jeroen, assuming he tested things.

jeroen:
The problem is that`gmp.h` hardcodes the value of `#define __GMP_LIBGMP_DLL` (which sets the `__declspec`) depending on whether or not a shared library was built for gmp, without the option to skip this for static linking as is done for most other libraries using a macro such as `LIBXML_STATIC` or `CURL_STATICLIB` and so on.

 So in order to actually link against the `libgmp.a` static library we have to somehow override this. I can send a small non-breaking patch introducing such a macro.

lazka:
Worth a try. Something like `GMP_STATIC` ?

PR #18493, opened 2023-09-15
riscv64-unknown-elf-binutils: remove conflict with binutils

MehdiChinoune:


lazka:
oh, I just pushed the same change and missed this PR

PR #18504, opened 2023-09-17
suitesparse: update to 7.2.0

mmuetzel:
The issues on 32-bit platforms are fixed in this version of SuiteSparse.

Rebuild all dependencies to depend on the newest version (instead of SuiteSparse 5) also on MINGW32 and CLANG32.

I didn't remove the suitesparse5 package just yet. It would probably be good to keep that package around at least a couple of months to allow for a smooth transition downstream.

(Also fixes the license identifier for ceres-solver.)


mmuetzel:
I downloaded and installed the MINGW32 build artifacts from this PR.
With them, the test suite of Octave no longer crashes (like it did for SuiteSparse 6 to 7.1).
There are no new regressions compared to the test results with SuiteSparse 5.

PR #18531, opened 2023-09-18
fastapi reverts

lazka:
it's not compatible with newer anyio and starlette

see https://github.com/msys2/MINGW-packages/commit/41fd5f246e4a4a2fccdd61b0490d4b31c6f5dfd9#commitcomment-127546512


lazka:
@podsvirov ^

PR #18546, opened 2023-09-18
jj: link to AUR's jujutsu

### LINKED ISSUE #17553: mingw-w64-jj page links to wrong AUR's jj
page of mingw-w64-jj package which is VCS links to [jj package](https://aur.archlinux.org/packages/jj) from AUR which is FIFO and filesystem (they are not the same)

ognevnydemon:
fixes #17553 

maybe will update package while at it (if needed) but later

just for interest, do we need to switch to versioned tag?


ognevnydemon:
> > just for interest, do we need to switch to versioned tag?
> 
> what do you mean?

currently we use commit tag and run pkgver() function. but we can set pkgver manually and use v${pkgver} tag

PR #18556, opened 2023-09-19
jj: update to 0.9.0

ognevnydemon:
notable changes:
- build with all features
- add split package for docs
- add completions for shells

### DIFF
--- a/mingw-w64-jj/PKGBUILD
+++ b/mingw-w64-jj/PKGBUILD
@@ -1,12 +1,12 @@
 # Maintainer: Johannes Schindelin <johannes.schindelin@gmx.de>
 
 _realname=jj
-_commit=c59dcd76a4e085de85577df7c515bcf7884c8373
 pkgbase=mingw-w64-${_realname}
-pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=0.2.0.586.c59dcd76
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-docs")
+pkgver=0.9.0
 pkgrel=1
-pkgdesc='Jujutsu (an experimental VCS) (mingw-w64)'
+pkgdesc="Jujutsu (an experimental VCS) (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
 msys2_references=(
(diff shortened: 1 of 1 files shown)


mingw-w64-jj/PKGBUILD
```diff
@@ -15,53 +15,64 @@ msys2_references=(
 url="https://github.com/martinvonz/jj"
 license=('spdx:Apache-2.0')
 makedepends=("${MINGW_PACKAGE_PREFIX}-rust" "git")
-source=("git+https://github.com/martinvonz/jj#commit=${_commit}")
+source=("${_realname}::git+${url}#tag=v${pkgver}")
```
mmuetzel, line 18: Would it make sense to build from the release tarball instead?
``` suggestion
source=("${_realname}-${pkgver}.tar.gz::https://github.com/martinvonz/jj/archive/refs/tags/v${pkgver}.tar.gz")
```
(Untested)

ognevnydemon: in this way we also install source code. the only difference is that it's already signed and .git directory is included

mmuetzel: > in this way we also install source code.

Source tarballs usually contain source code. Is that different for that project?

> .git directory is included
Do you need that directory?

ognevnydemon: > Source tarballs usually contain source code. Is that different for that project?
no
> Do you need that directory?
no

I just left source as it was. IMO there is no reason to change the way how source is downloaded

mmuetzel: If possible try to stay close to the PKGBUILD templates to keep maintenance easy.

In this case, the closest template is probably:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.cargo-tarball.sh

ognevnydemon: anyway this is note essential. if I can leave this as is, I'll leave it as is

lazka:
either build a tarball or a commit, git tags can be changed without us noticing.

ognevnydemon:
> either build a tarball or a commit, git tags can be changed without us noticing.

oh, ok. thanks for the answer. will change today

ognevnydemon:
> oh, ok. thanks for the answer. will change today

done

ognevnydemon:
oops, binary was removed...

ognevnydemon:
any other objections?

lazka:
I find the split package for 180kb docs questionable, but ymmv :)

ognevnydemon:
> I find the split package for 180kb docs questionable, but ymmv :)

these docs are actually compiled with mkdocs, but IMO it will cool to let someone to read them using good reader

PR #18565, opened 2023-09-20
Update mingw-w64

lazka:
https://github.com/mingw-w64/mingw-w64/compare/0605217f5d60257a53ec875e2e4ff36a3e14f825...833753684d3a520ebcd3cd73e614c97bbb55ffb8


raedrizqie:
while you are at it, can you please add this patch (needed for chromium)
```
diff --git a/mingw-w64-headers/include/windows.system.userprofile.h b/mingw-w64-headers/include/windows.system.userprofile.h
index 06613e00d..fda640dfc 100644
--- a/mingw-w64-headers/include/windows.system.userprofile.h
+++ b/mingw-w64-headers/include/windows.system.userprofile.h
@@ -157,7 +157,7 @@ namespace ABI {
                         HSTRING *value) = 0;
 
                     virtual HRESULT STDMETHODCALLTYPE get_WeekStartsOn(
-                        enum DayOfWeek *value) = 0;
+                        enum ABI::Windows::Globalization::DayOfWeek *value) = 0;
 
                 };
             }
```

lazka:
> while you are at it, can you please add this patch (needed for chromium)

Would it make sense to report this upstream?

Biswa96:
> Would it make sense to report this upstream?

https://bugs.winehq.org/show_bug.cgi?id=55347

lazka:
Looks like winehq.org is down currently :/

Either way, feel free to create a PR, and add an inline comment as to why this is needed, when it can be removed, and any related upstream links.

PR #18571, opened 2023-09-21
numba: update to 0.58.0

ognevnydemon:
includes numpy 1.25 compat


MehdiChinoune:
> also llvmlite 0.41.0 is released, but it requires some extra changes

numba 0.58 requires llvmlite 0.41 see: https://github.com/numba/numba/blob/0.58.0/setup.py#L27

ognevnydemon:
> > also llvmlite 0.41.0 is released, but it requires some extra changes
> 
> numba 0.58 requires llvmlite 0.41 see: https://github.com/numba/numba/blob/0.58.0/setup.py#L27

oh...

PR #18578, opened 2023-09-22
jj: switch to cargo install

ognevnydemon:
~~this provides 2 extra binaries: `fake-editor` and
`fake-diff-editor`~~ now not, see further conversation

### DIFF
--- a/mingw-w64-jj/PKGBUILD
+++ b/mingw-w64-jj/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-docs")
 pkgver=0.9.0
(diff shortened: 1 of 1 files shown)


lazka:
Aren't those internal testing cli tools? in other words, should they be installed?

ognevnydemon:
> Aren't those internal testing cli tools? in other words, should they be installed?

hm, as I understood from guidelines, upstream recommend to install this via `cargo install --git https://github.com/martinvonz/jj --bin jj jj-cli`. jj-cli includes these binaries (see manifest at cli/Cargo.toml). but I haven't tested this installation method. it's not essential now, so I will test it later then resolve, do we need this or not

ognevnydemon:
also maybe `testing` directory is for experimental things

ognevnydemon:
I tried to install via this command. it didn't install these binaries. sorry, I was wrong. I'll push the fix to remove this binaries so I can left `cargo install`

PR #18580, opened 2023-09-22
rust: rebuild for newer LLVM

ognevnydemon:
it will give some performance boost


lazka:
> it will give some performance boost

just wondering, why is that?

ognevnydemon:
> just wondering, why is that?

https://perf.rust-lang.org/compare.html?start=03a119b0b0e310d22d94399b24ed030056050f13&end=443c3161dd04f4c1b656a626f9079921bee9c326&stat=instructions:u

Biswa96:
The title says `Comparing instructions:u between #114344 and #114048`. How is that related to llvm version?

ognevnydemon:
> The title says `Comparing instructions:u between #114344 and #114048`. How is that related to llvm version?

because #114048 was about upgrade to 17.0.0

ognevnydemon:
forgot that we need to backport some things. so I can leave this until 1.73.0 update

PR #18585, opened 2023-09-22
python-tldr: new package

ognevnydemon:
also adds python-shtab and python-shpinx-argparse as dependencies.

~~currently I receive an error while building tldr. I have no idea what's wrong~~ fixed

### DIFF
--- a/mingw-w64-python-shtab/PKGBUILD
+++ b/mingw-w64-python-shtab/PKGBUILD
@@ -1,0 +1,65 @@
+# Maintainer: Maksim Bondarenkov <maksapple2306@gmail.com>
+
+_realname=shtab
+pkgbase=mingw-w64-python-${_realname}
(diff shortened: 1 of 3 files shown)


MehdiChinoune:
requirments.txt file is not included in PyPi tarball, consider using the one from GitHub.

ognevnydemon:
> requirments.txt file is not included in PyPi tarball, consider using the one from GitHub.

thanks! now it can't find man page, but `make -C docs` from arch doesn't work...

ognevnydemon:
> > requirments.txt file is not included in PyPi tarball, consider using the one from GitHub.
> 
> thanks! now it can't find man page, but `make -C docs` from arch doesn't work...

works, but now sphinxarg is now required. is there package in MSYS2 with it?

ognevnydemon:
pushed all things. should work now

Issue #18588, opened 2023-09-23
gtksourceview3 is missing the DLL library in CLANGARM64


hmartinez82:
### Description / Steps to reproduce the issue

The dynamic library file is completely missing from the package.

### Expected behavior

`/clangarm64/bin/libgtksourceview-3.0-1.dll` is present

### Actual behavior

`/clangarm64/bin/libgtksourceview-3.0-1.dll` is not present

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22621

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [X] CLANGARM64

### Are you willing to submit a PR?

Yes

hmartinez82:
gtksourceview3 is now failing to be built with a clean env. I think that the last time this package was built was before the switch to LLVM 16.
In CLANG64:
```
...
gtksourceview-enumtypes.c:273:1: warning: '/*' within block comment [-Wcomment]
/* enumerations from "gtksourceview.h" *C:/msys64/
^
gtksourceview-enumtypes.c:347:1: warning: '/*' within block comment [-Wcomment]
/* Generated data ends here */
^
11 warnings generated.
  CCLD     libgtksourceview-core.la
  CCLD     libgtksourceview-3.0.la

*** Warning: linker path does not have real file for library -luuid.
*** I have the capability to make that library automatically link in when
*** you link to this library.  But I can only do this if you have a
*** shared version of the library, which you do not appear to have
*** because I did check the linker path looking for a file starting
*** with libuuid and none of the candidates passed a file format test
*** using a file magic. Last file checked: C:/msys64/clang64/lib/libuuid.a
*** The inter-library dependencies that have been dropped here will be
*** automatically added whenever a program is linked with this library
*** or is declared to -dlopen it.

*** Since this library must not contain undefined symbols,
*** because either the platform does not support them or
*** it was explicitly requested with -no-undefined,
*** libtool will only create a static version of it.
  GISCAN   GtkSource-3.0.gir
[...]

PR #18591, opened 2023-09-23
gtksourceview3: Fix compiling with clang

Biswa96:
* workaround warning: linker path does not have real file for library -luuid.
* workaround: glib-mkenums adds wrong comments in gtksourceview-enumtypes.c.
* disable gtk-doc due to wrong paths in build command.

Fixes https://github.com/msys2/MINGW-packages/issues/18588


lazka:
fyi, we don't have any users left for gtksourceview3 with clang, so if that complicates things, consider dropping it and gtksourceviewmm3 for clang envs.

Biswa96:
Please wait before removing this package. This reveals a bug in python. The `/clang64/lib/python3.11/distutils/cygwinccompiler.py` file has `-mdll` compiler option. It was not supported in clang for long time. clang 17 raises that to compiler error. We've to fix it first.

Biswa96:
Could you fix it in python please?

lazka:
sure

hmartinez82:
@lazka So, I was trying to run quodlibet (know that project? 😉) in CLANGARM64. And these lines in textedit.py were being the issue:
```
try:
    import gi
    gi.require_version("GtkSource", "3.0")
    from gi.repository import GtkSource
except (ValueError, ImportError):
    TextView = Gtk.TextView
    TextBuffer = Gtk.TextBuffer
```
Interestingly if gtksourceview3 is not installed the exception path takes place (at least that's what I remember seeing while debugging) and it works, but if gtksourceview3 is installed then that gi import goes thru and the code crashes in the next line: `class TextBuffer(GtkSource.Buffer)` saying that GtkSource.Buffer is void.

in CLANGARM64 where gtksourceview4 does have the dll, I was able to change the require above to gi.require_version("GtkSource", "4") and it worked.

lazka:
> @lazka So, I was trying to run quodlibet (know that project? 😉) in CLANGARM64. And these lines in textedit.py were being the issue:

I see. I took the opportunity to get rid of it there: https://github.com/quodlibet/quodlibet/pull/4377

MehdiChinoune:
> Please wait before removing this package. This reveals a bug in python. The `/clang64/lib/python3.11/distutils/cygwinccompiler.py` file has `-mdll` compiler option. It was not supported in clang for long time. clang 17 raises that to compiler error. We've to fix it first.

I have encountered this issue with arrow package on CLANG* environments.

lazka:
pkg-config segfaults after the last rebuild

MehdiChinoune:
> pkg-config segfaults after the last rebuild, but works under a debugger, meh.

because of Hardening, Maybe!

mmuetzel:
> pkg-config segfaults after the last rebuild, but works under a debugger, meh.

Sounds like it could be a threading issue.

PR #18594, opened 2023-09-23
Fix libraw.pc

jeroen:
The `libraw.pc` file hardcodes `-lstdc++` which is but unnecessary and breaks building dependents on clang.

```
lld: error: unable to find library -lstdc++
```

Example: https://github.com/r-windows/libs/actions/runs/6283365115/job/17063752821


Biswa96:
Shouldn't the issue be fixed in the source code or reported in upstream?

jeroen:
I can try that too, but I'm not sure if upstream is active and cares.

Biswa96:
Thanks for trying that. I have created a pull request to fix it in proper way.

Biswa96:
Fixed in https://github.com/msys2/MINGW-packages/pull/18595

PR #18596, opened 2023-09-24
update cairo/gnuplot to 1.18.0/5.4.9

MehdiChinoune:


jeroen:
Maybe update pango then, see https://github.com/msys2/MINGW-packages/pull/17369#issuecomment-1597224460

Issue #18602, opened 2023-09-24
[fmt] fatal error: '__std_stream' file not found - llvm 17 issue


lazka:
Reported in discord:

```
[build] In file included from C:/src/dolphin/Source/Core/VideoCommon/Assets/DirectFilesystemAssetLibrary.cpp:9:
[build] In file included from C:/msys64/clang64/include/fmt/std.h:23:
[build] C:/msys64/clang64/include/fmt/ostream.h:17:12: fatal error: '__std_stream' file not found
[build]    17 | #  include <__std_stream>
[build]       |            ^~~~~~~~~~~~~~
[build] 1 error generated.
```
`__std_stream` has been renamed to `std_stream.h` in this commit https://github.com/llvm/llvm-project/commit/40a187bb622959023f751ec3a11f2af5fa8b48fa and is also no longer installed

They were using a private header... :/

lazka:
https://github.com/fmtlib/fmt/issues/3654

lazka:
Upstream went with the same fix we added, so we can close this.

PR #18603, opened 2023-09-24
fmt: workaround for llvm 17

lazka:
See #18602


lazka:
For future reference:

* The code seems to have a fallback in place at least (might break unicode output.. not sure)
* While this might break some use cases if rebuilt with this, it unbreaks the build

PR #18609, opened 2023-09-25
graphicsmagick: update to 1.3.42

kmilos:


Biswa96:
The 002-tiff-transparency.patch file was not removed.

kmilos:
Ooops, sorry. Wishful thinking - "patches are downloaded from upstream" 😉

PR #18611, opened 2023-09-25
msitools: Use Windows paths for DATADIR and LOCALEDIR

jhol:


Biswa96:
Using full path in source code will not work because not all msys2 are installed in same directory. Here CI will attach a different path. The proper way is to relocate path like in curl, openssl etc. packages.

lazka:
Since this is a vala package the fix would be similar to this: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-vala/0001-relocate-plugin-path.patch

Please report this upstream.

PR #18614, opened 2023-09-26
mcfgthread: Update to 1.6

lhmouse:
This major release contains the following changes since v1.5-ga.2:

1. A compatibility manifest has been added to the DLL, declaring compatibility with Windows 7+.
2. The size of a TLS table is now decreased correctly during rehashing if deleted keys are detected. Previously it caused the table size to grow if the process kept allocating and deallocating TLS keys.

**Full Changelog**: https://github.com/lhmouse/mcfgthread/compare/v1.5-ga.2..v1.6-ga.1


lhmouse:
Thank you!

PR #18636, opened 2023-09-27
Python: use `-shared` instead if `-mdll`

MehdiChinoune:
Confirm it by reenabling glib for arrow on clang environments where It was failing because if this bug.


lazka:
oh, this broke all the extension modules somehow

lazka:
ah, might be that this overlapped with the broken pkg-config.

we really need to replace all pkg-config with pkgconf

Biswa96:
> oh, this broke all the extension modules somehow. I somehow missed that, oops

Please explain the issue.

lazka:
this build lost tkinter and ssl extensions, because it couldn't find them via pkg-config, because that crashes. all good now.

PR #18642, opened 2023-09-28
Update freetds to 1.4.2

taozuhong:


Biswa96:
Thank you for this update.

I found an issue that some file names are now prefixed with target triplet. I have tried to compile freetds after removing `--host` and `--target` options. That keeps the file name same as before. Could you check that please? If it works you can remove those options from PKGBUILD file.

PR #18643, opened 2023-09-28
wireshark: Update package to 4.2 branch

randstr:
Update the package to track the release-4.2 branch, that will eventually become the 4.2.0 stable release.

Other changes:
- Add new opencore-amr dependency
- Remove patch that was committed upstream
- Add a comment about Lua 5.2 and make CMake option explicit
- Add new test target dependencies
- Add command to build test-programs required for the test target
- Install development headers so users can use this package to develop wireshark plugins


Biswa96:
Thank you for the update. I have installed the ucrt64 CI artifact and it worked in my system.

PR #18644, opened 2023-09-28
python: use -shared instead if -mdll (v2)

lazka:


lazka:
after I reverted #18636, somewhat prematurely.

PR #18651, opened 2023-09-29
gumbo-parser: update to 0.12.1, change upstream

Biswa96:


Biswa96:
@lazka Could you provide any hint why the build fails with glib-mkenums?

Biswa96:
The build error may be related to this upstream issue https://github.com/mesonbuild/meson/issues/12271

PR #18655, opened 2023-09-29
meson: Update to 1.2.2

lazka:


ognevnydemon:
forgot to delete patches

lazka:
thanks

MehdiChinoune:
I think there is a regression, gexiv2 was building fine before this update now it fails:
```
ninja: error: 'D:/M/msys64/mingw32/bin/glib-mkenums', needed by 'gexiv2/gexiv2-enums.h', missing and no known rule to make it
```
but glib-mkenums.exe already exists

Biswa96:
It is related to this upstream issue https://github.com/mesonbuild/meson/issues/12271

MehdiChinoune:
But It started failing only with this update.

lazka:
#18659 should fix it

PR #18659, opened 2023-09-29
glib2: add .exe extension in .pc file for wrappers we create

lazka:
While the .pc file already contains absolute paths to .exe glib tools without .exe in the pc, a recent meson change in meson made it no longer find the tools.

Since we created those .exe we might as well adjust the path in the .pc files, so that any meson error is at least no longer MSYS2 specific.


lazka:
Motivated by https://github.com/msys2/MINGW-packages/pull/18655#issuecomment-1741387747

Biswa96:
Shouldn't gobject-query also have the .exe extension?

Biswa96:
Now build fails with the following error.

```
ninja: error: 'D:/M/msys64/clang32/bin/glib-compile-resources', needed by 'talkatu/talkaturesources.h', missing and no known rule to make it
```

PR #18665, opened 2023-09-30
hwloc: Update to 2.9.3

lazka:


lazka:
mwindows needs to be moved to ldflags. And then something else failed and I gave up.

PR #18666, opened 2023-09-30
meson: revert commit that breaks glib tools lookup

lazka:
See https://github.com/mesonbuild/meson/issues/12271#issuecomment-1741686263


xclaesse:
Note that I believe the same issue always existed for dinner tools we were already looking with pkgconfig, such as glib-compile-schemas, right?

lazka:
Since this revert fixes glib-compile-resources, and glib-compile-schemas is the same in the .pc file and also an exe I'd assume it is fixed by this as well. I don't know any project using it though so I can test.

PR #18667, opened 2023-09-30
gcc: Pass `--disable-libstdcxx-time` when building a non-posix thread model

lhmouse:
When building libstdc++, by default (`--enable-libstdcxx-time=auto` [seen in source code](https://github.com/gcc-mirror/gcc/blob/fd35d72a3dcd5ba14d81a1890236acd0145497e1/libstdc%2B%2B-v3/acinclude.m4#L1748)) the configure script attempts to detect existence of `nanosleep()`. As long as winpthreads is installed, the test succeeds and `#define _GLIBCXX_USE_NANOSLEEP 1` in 'c++config.h' is emitted, and libstdc++ will have a dependency on winpthreads. This is undesired when building GCC with a non-posix thread model, namely, `win32`, `mcf` or `single`.

We only build `posix` thread model, so this has no effect on the MSYS2 gcc.

### DIFF
--- a/mingw-w64-gcc/PKGBUILD
+++ b/mingw-w64-gcc/PKGBUILD
@@ -26,7 +26,7 @@ pkgver=13.2.0
 #_majorver=${pkgver:0:1}
 #_sourcedir=${_realname}-${_majorver}-${_snapshot}
 _sourcedir=${_realname}-${pkgver}
-pkgrel=2
+pkgrel=3
 pkgdesc="GCC for the MinGW-w64"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


lhmouse:
Needs some investigation... Maybe this is not necessary.

I would like to convert this back to draft but could not find the button... close for the time being.

MehdiChinoune:
> I would like to convert this back to draft but could not find the button

Under "Reviewers"

lhmouse:
Ah many thanks. However there is no such option on my side. Maybe it requires write access to the repo.


---

This PR really seems unnecessary. GCC should be able to detect `nanosleep()` according to the current configuration, as tested on https://github.com/niXman/mingw-builds/pull/659#issuecomment-1741725147. The reason why it is defined for GCC with the `posix` thread model is because GCC has `-lpthread` as a default library, so it's kind of by design: https://github.com/niXman/mingw-builds/pull/659#issuecomment-1741796626

PR #18669, opened 2023-09-30
exiv2: update to 0.28

MehdiChinoune:


lazka:
We stopped this last time because we would lose unicode path support. Has anything changed there?

MehdiChinoune:
> We stopped this last time because we would lose unicode path support with 0.28.0. Has anything changed there?

I think exiv2 did the opposite It defaults to UNICODE.

lazka:
How can it do that?

MehdiChinoune:
> How can it do that?

I don't know, but according to their README they only supports UCRT which supports UTF-8 by default.
And they also add `-municode` flags on MINGW.

lazka:
The utf-8 locale needs to be set by the process: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170#utf-8-support.

So every application directly or indirectly linking to exiv2 will need to call setlocale when using ucrt somewhere.

Assuming upstream doesn't fix this, I think your approach of splitting things up between ucrt and msvcrt looks good though.

lazka:
@kmilos any thoughts on this?

kmilos:
Agree w/ the approach, but I kinda held back on pushing 0.28.0 on purpose - there are a few regressions and I'd rather have upstream release 0.28.1 (w/ or w/o Unicode restored), and I don't feel like porting back so many patches. @Exiv2/exiv2

kmilos:
Might as well wait for qgis 3.34 including the patch, should come in October AFAICT...

MehdiChinoune:
> Might as well wait for qgis 3.34 including the patch, should come in October AFAICT...

I have already backported the patch

What regressions are you talking about?

MehdiChinoune:
I have backported some fixes (copied from ArchLinux).

kmilos:
> I have backported some fixes (copied from ArchLinux).

Thanks. There's a few more I'd like to get in.

Is.there a particular rush to ship 0.28.0?

MehdiChinoune:
> > I have backported some fixes (copied from ArchLinux).
> 
> Thanks. There's a few more I'd like to get in.
> 
> Is.there a particular rush to ship 0.28.0?

Yes, I have spent a lot of time to backport fixes and make sure that all packages rebuild successfully.
I want to update it as soon as possible before we get any new change from other packages that disrupt the update.

kmilos:
Ok, go ahead then w/ the merge, I'll patch more later as it won't require rebuilds.

PR #18675, opened 2023-10-01
cargo-create-tauri-app: New package

podsvirov:
I'm participant of the [hacktoberfest](https://hacktoberfest.com/), please add the label `hacktoberfest-accepted` to this PR when merging.

### DIFF
--- a/mingw-w64-cargo-create-tauri-app/PKGBUILD
+++ b/mingw-w64-cargo-create-tauri-app/PKGBUILD
@@ -4,7 +4,7 @@ _basename=create-tauri-app
 _realname=cargo-${_basename}
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=3.7.3
+pkgver=3.8.0
 pkgrel=1
 pkgdesc="Rapidly scaffold out a new tauri app project (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
No other distro is packaging it, so I wonder if we should: https://repology.org/projects/?search=tauri&maintainer=&category=&inrepo=&notinrepo=&repos=&families=&repos_newest=&families_newest=

podsvirov:
> No other distro is packaging it...

This is strange for me because this is the first tool I use when I start learning `Tauri` development. We can see suggestions for using this tool on every getting started ([1](https://tauri.app), [2](https://tauri.app/v1/guides/getting-started/setup), [3](https://tauri.app/v1/guides/getting-started/setup/vite)...) page on the `Tauri` website.

podsvirov:
> actually you can install it easily with `cargo install`...

You are absolutely right! This can be done even without using a distribution such as `MSYS2`! Why peoples use distros?..

lazka:
We can't really package every npm/pypi/cpan/cargo package out there. With the package managers doing dynamic linking that also not possible since all packages have to be compatible. The line has to be drawn somewhere.

podsvirov:
I understand you, but in this case, we are talking about a command-line utility that will be useful for creating new projects. Users will not need to track its current version and periodically rebuild and install this utility on their own - our package manager will take care of this for users of the distribution.

In any case, thank you for reviewing this `PR` and its merging!

PR #18685, opened 2023-10-02
exiv2: backport some more fixes for 0.28.0

kmilos:
@MehdiChinoune


Biswa96:
The patch file needs to be added manually. The CI build fails with the following error.

```
  File test/data/pr_2612_poc.heic: git binary diffs are not supported.
```

kmilos:
> The patch file needs to be added manually.

Bummer... Ok, will do in a bit.

PR #18705, opened 2023-10-04
python-spin: new package

kmilos:
Required for updating scikit-image and other "scientific Python" libraries.

NB: Haven't figured out how to actually update scikit-image using this though...


Biswa96:
It would be better if the new packages are added with those python package updates. We could verify the dependencies in one run.

kmilos:
> It would be better if the new packages are added with those python package updates.

Yeah, let's maybe continue that over at https://github.com/msys2/MINGW-packages/pull/18706.

I'm not convinced how this "spin" works, or if it's really required... Arch seems to build scikit-image w/ just the meson backend just fine?

MehdiChinoune:
OK to merge?

kmilos:
The packaging is ready, but it's not really required by anything AFAICT, so...

PR #18706, opened 2023-10-04
python-lazy-loader: new package, update scikit-image

kmilos:
Required for updating scikit-image and other "scientific Python" libraries: https://scientific-python.org/specs/

See also https://aur.archlinux.org/packages/python-lazy-loader

### DIFF
--- a/mingw-w64-python-lazy-loader/PKGBUILD
+++ b/mingw-w64-python-lazy-loader/PKGBUILD
@@ -1,0 +1,56 @@
+# Maintainer: Miloš Komarčević <miloskomarcevic@aim.com>
+
+_pyname=lazy_loader
+_realname=lazy-loader
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=0.3
+pkgrel=1
+pkgdesc="Populate library namespace without incurring immediate import costs (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+msys2_references=(
+  'pypi: lazy_loader'
+)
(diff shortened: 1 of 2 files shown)


kmilos:
Need to remove the .dll.a implibs somehow, or patch meson commands somewhere to build modules rather than shared libs... Any suggestions are welcome.

kmilos:
@lazka Still need to remove all the .dll.a implibs from the scikit-image package!

lazka:
But they don't hurt, right?

kmilos:
> Should we report it upstream?

No side-effects, no, just wasteful/ugly.

> Should we report it upstream?

Don't know who's to blame here for shared_library() vs shared_module() - meson-python, or the scikit-image package itself, or a somewhere else?

lazka:
In pycairo I use this https://github.com/pygobject/pycairo/blob/64058391b37957f719246c2e8c51997ca10f73ac/cairo/meson.build#L75 which doesn't seem to create it.

kmilos:
This is from scikit-image:

https://github.com/scikit-image/scikit-image/blob/441fe68b95a86d4ae2a351311a0c39a4232b6521/skimage/_shared/meson.build

Can't really tell what's different...

lazka:
Oops, I was wrong, we delete it here too: https://github.com/msys2/MINGW-packages/blob/1c6698776a7c9bc984c564afe90adb4d3f24b2b2/mingw-w64-pycairo/PKGBUILD#L51

kmilos:
Will you add this than as well please before it hits staging?

kmilos:
And I guess this should now be reported as a meson bug on https://mesonbuild.com/Python-module.html#extension_module ?

lazka:
Yeah

lazka:
> Will you add this than as well please before it hits staging?

ok d76c50eb43c06663

kmilos:
> Yeah

Done.

PR #18798, opened 2023-10-05
dav1d: Update to 1.3.0

lazka:


kmilos:
What about rebuilds, changelog says ABI has changed?

Also, might be an idea to start including soname in the DLL filename while at it?

lazka:
I missed that, thanks

lazka:
reverted for now

kmilos:
> Also, might be an idea to start including soname in the DLL filename while at it?

I guess we'd need to patch
https://code.videolan.org/videolan/dav1d/-/blob/47107e384bd1dc25674acf04d000a8cdc6195234/src/meson.build#L307-L311

kmilos:
@lazka I can work on this if you weren't planning to?

lazka:
Already done in the other PR :)

PR #18807, opened 2023-10-06
gimp3: remove llvm workaround that is no longer needed with llvm 17

lazka:
See https://github.com/llvm/llvm-project/issues/57334#issuecomment-1751059398


ognevnydemon:
`updpkgsums`

PR #18811, opened 2023-10-06
rav1e: fix PKGBUILD to create static library

Deyvan:


lazka:
Looks like that doesn't work somehow? any ideas?

Deyvan:
It's working fine for me. Must be a timout.
What is frustrating is the problem in linking two static rust libraries

lazka:
Fixed in #19278

PR #18815, opened 2023-10-07
OpenBLAS: Enable Fortran with flang

MehdiChinoune:


mmuetzel:
This change will likely cause issues on systems with AVX-512 instructions.
See: https://github.com/OpenMathLib/OpenBLAS/pull/4180

Issue #18323, opened 2023-09-05
mingw-w64-v8 for clangarm


jeroen:
### Description / Steps to reproduce the issue

I am trying to build `mingw-w64-v8` for clangarm. I have the arm64 builds working for linux and macos so I am optimistic it should be possible :-)

In `PKGBUILD` we set `_arch=arm64`

https://github.com/msys2/MINGW-packages/blob/6a50f6416f3182b73f178591ca31d980f42a1fcb/mingw-w64-v8/PKGBUILD#L157

But then the build fails with:

```
clang++: error: unsupported argument '-mbig-obj' to option '-Wa,'
```

This flag is set in one of the patches, so I tried commenting that out:

https://github.com/msys2/MINGW-packages/blob/6a50f6416f3182b73f178591ca31d980f42a1fcb/mingw-w64-v8/001-add-mingw-toolchain.patch#L703-L708

But now the build fails with:

```
==> Starting build()...
Done. Made 449 targets from 128 files in 480ms
ninja: Entering directory `out/Release-CLANGARM64'
[1/1751] CXX obj/cppgc_base/compaction-worklists.o
FAILED: obj/cppgc_base/compaction-worklists.o
[...]

PR #18820, opened 2023-10-08
v8: update to latest stable and fix aarch64

jeroen:
Yay this now builds on arm64 🎉 
Fixes https://github.com/msys2/MINGW-packages/issues/18323


jeroen:
OK this is ready. Can you squash-and-merge, or should I squash it on my end? Builds take super long so I rather do the first.

hmartinez82:
@jeroen does that mean QtWebEngine next? :)

PR #18821, opened 2023-10-08
objfw 1.0.4: new package

Midar:


mingw-w64-objfw/PKGBUILD
```diff
@@ -0,0 +1,164 @@
...
+         "${MINGW_PACKAGE_PREFIX}-ofarc"
+         "${MINGW_PACKAGE_PREFIX}-ofdns"
+         "${MINGW_PACKAGE_PREFIX}-ofhash"
+         "${MINGW_PACKAGE_PREFIX}-ofhttp")
+pkgbase=mingw-w64-${_realname}
+pkgver=1.0.4
+pkgrel=1
+pkgdesc="Portable, lightweight framework for the Objective-C language"
+arch=('any')
+mingw_arch=('mingw32' 'clang64' 'clangarm64')
```
lazka, line 15: just wondering, why this unusual selection of environments?

Midar: The idea is that it uses clang anyway for the much better Objective-C support, which means the 3 targets allow to:

* Target old Windows versions in 32 bit with MSVCRT via MINGW32, as there is no MINGWCLANG32 and CLANG32 requries UCRT.
* Newer 64-bit versions of Windows support UCRT anyway, so that's targeted by CLANG64
* CLANGARM64 for anything using ARM64

The idea was that this covers all use cases. The reason for not just supporting them all was that in the past, platforms often broke, so the fewer platforms to support, the less likely things will break.

Does that make sense?

Midar:
Thanks for reviewing and merging!

Are there any steps required to create a binary package that gets added to the repo?

PR #18823, opened 2023-10-09
c-ares: update to 1.20.0

Biswa96:


lazka:
1.20.1 is out

Biswa96:
Thanks, I shall update that.

I was wondering, should libcares-static.pc be removed? The static compiler flag can be added with Cflags.private in libcares.pc.

Biswa96:
> The static compiler flag can be added with Cflags.private in libcares.pc.

For the record, here are the changes with pkgconfig file.

* Before:

```make
prefix=/ucrt64
exec_prefix=/ucrt64/bin
libdir=/ucrt64/lib
includedir=/ucrt64/include

Name: c-ares
URL: https://c-ares.org/
Description: asynchronous DNS lookup library
Version: 1.20.0
Requires: 
Requires.private: libcares-static
Cflags: -I${includedir}
Libs: -L${libdir} -lcares
Libs.private: 
```

* After:

```make
prefix=/ucrt64
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include

Name: c-ares
URL: https://c-ares.org/
Description: asynchronous DNS lookup library
Version: 1.20.0
Requires: 
Requires.private: 
Cflags: -I${includedir}
Cflags.private: -DCARES_STATICLIB
Libs: -L${libdir} -lcares
Libs.private:  -lws2_32 -ladvapi32 -liphlpapi
```

PR #18830, opened 2023-10-10
qt6: update to 6.6.0 - Part1

MehdiChinoune:


MehdiChinoune:
Findzstd.cmake from llvm is interfering!

Biswa96:
Just for the record, would you like to document what failed in this qt6 build and how it was fixed/workaround in llvm update?

MehdiChinoune:
After https://github.com/llvm/llvm-project/commit/5279e6a7d67
LLVM/Findzstd.cmake only finds libzstd_static which makes zstd/zsdtConfig.cmake throws an error because of a missing libzstd_shared target.
I have patched LLVM/Findzstd.cmake to find both targets.

PR #18834, opened 2023-10-11
purpose: New package

hmartinez82:
On the way to building kdenlive

### DIFF
--- a/mingw-w64-purpose/PKGBUILD
+++ b/mingw-w64-purpose/PKGBUILD
@@ -1,59 +1,69 @@
 
-_realname=purpose
-pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+_variant=-${KF5_VARIANT:-shared}
+source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
+_kde_f5_init_package "${_variant}" "purpose"
 pkgver=5.110.0
 pkgrel=1
-pkgdesc="Framework for providing abstractions to get the developers purposes fulfilled (mingw-w64)"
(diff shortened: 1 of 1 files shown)


Biswa96:
kde packages follow a template. See other kde PKGBUILD files.

mingw-w64-purpose/PKGBUILD
```diff
@@ -0,0 +1,69 @@
+
+_variant=-${KF5_VARIANT:-shared}
+source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
+_kde_f5_init_package "${_variant}" "purpose"
+pkgver=5.110.0
+pkgrel=1
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+msys2_references=(
+  'archlinux: purpose'
```
Biswa96, line 10:
```suggestion
  'archlinux: purpose5'
```

hmartinez82: I'll make the change, but could explain what's the rationale here so I learn for the future? I used purpose because https://archlinux.org/packages/kde-unstable/x86_64/purpose/

Biswa96: Doesn't "kde-unstable" in the link explain it? 😄 Basically, ArchLinux is preparing for KDE6 and that "kde-unsable" repository is for that.

hmartinez82: Ahh ok, that's the rationale then. I had no idea about it. I mean there's no way I would understand the intent and arrive in this conclusion myself. 
And correct, version 5.240.0 of purpose on Arch is the Qt6 based one that's why we are building 5.110.0 here

PR #18835, opened 2023-10-11
qt6-static: update to 6.6.0

MehdiChinoune:
Fails on GNU environments while packaging `No space left on device.`


MehdiChinoune:
Build only essential modules because the huge size of the build and install directories which leads to failure of CI job.
We could add another module if someone asks for it.

PR #18841, opened 2023-10-12
vlc: Update to 3.0.19

lazka:


Biswa96:
0023-opencv4-support.patch isn't used.

PR #18848, opened 2023-10-13
gimp3: Enable support for Lua plugins in aarch64 builds

hmartinez82:


hmartinez82:
I forget like 90% of the time 😅

PR #18850, opened 2023-10-13
qt6: update to 6.6.0 - Part3

MehdiChinoune:


omichel:
It seems the update did not yet reached the pacman servers as I am still getting version 6.5.3 for the qt6-websockets package.

Biswa96:
The queue is stuck with self hosted ARM runner. See the progress here https://packages.msys2.org/queue

PR #18856, opened 2023-10-13
libipt update

GitMensch:
* version bump
* changed to use tags instead of commit id
* adjustment to current version by disabling C11 threads

### DIFF
--- a/mingw-w64-libipt/PKGBUILD
+++ b/mingw-w64-libipt/PKGBUILD
@@ -1,7 +1,7 @@
 _realname=libipt
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgbase="mingw-w64-${_realname}"
-pkgver=2.0.6.r0.gba98732
+pkgver=2.1
 pkgrel=1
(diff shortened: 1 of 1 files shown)


ognevnydemon:
as it was discussed in #18556, source tarballs are prefered more, than versioned tag

GitMensch:
We should have a general issue about that, just search for `#tag` (it "obviously" was so common in here that I've changed to that - will revert).

Biswa96:
`#tag` is ok here. The source repository is small. We use it for other small packages also.

ArchLinux prefers `#tag` instead of source tarball (most of the time) because it helps to verify the source using git.

GitMensch:
... ok, then getting back to `#tag` with the next commit...

Biswa96:
It is not an requirement. You are welcome to use either `#commit` or `#tag` as per your preference for this package.

lazka:
Note that tags can change, commits can't.

GitMensch:
If you tell me that you want to part from Arch and want to replace all `#tag` by `#commit` "over time", then I'm totally adjusting this one. Otherwise I'd leave it as is and consider this PR finished.

PR #18858, opened 2023-10-13
update blender/oiio to 3.6.4/2.5.4.0

MehdiChinoune:


MehdiChinoune:
Unfortunately krita fails to build against ocio 2.3.0

PR #18863, opened 2023-10-14
luajit: Update to latest version and enable aarch64 builds

hmartinez82:
Lots of recent work gone into LuaJIT for AArch64, and for Windows  on ARM


lazka:
Could we use the version scheme Arch uses?

hmartinez82:
I just checked. Arch is using the same commit as this PR. So you mean using the same name pattern?

hmartinez82:
@lazka Done 🚀

lazka:
Thanks. It's mainly so our scripts notice new versions when comparing with Arch.

PR #18866, opened 2023-10-15
qemu: Enable hax and whpx

MehdiChinoune:


MehdiChinoune:
They are already enabled!

PR #18874, opened 2023-10-16
libheif: update to 1.17.0

kmilos:


Biswa96:
Just wondering, why does libheif need two jpeg libraries?

kmilos:
> Just wondering, why does libheif need two jpeg libraries?

Not providing the same codec: one is for regular  JPEG, the other for JPEG 2000.

PR #18875, opened 2023-10-16
[new-package] uutils-coreutils 0.0.22

MehdiChinoune:


ognevnydemon:
by the way: all binaries have the same name as coreutils' ones. which one will be preferred: from mingw envs or MSYS's ones?

MehdiChinoune:
> by the way: all binaries have the same name as coreutils' ones. which one will be preferred: from mingw envs or MSYS's ones?

Of course MinGW-w64 ones.

> upd.: you can set prefix with `PROG_PREFIX` variable while `make install`

I know about it, It's useful to avoid conflict with coreutils.
In our case there is no conflict.
I expect anyone using this experimental package to be aware if the consequences. It also allows running scripts without modifying them.

PR #18886, opened 2023-10-18
gimp3: Use Luajit in aarch64 now that it's available

hmartinez82:


Biswa96:
Just wondering, did you check if it works?

PR #18887, opened 2023-10-18
wxwidgets3.2: Fix CLANG scripts causing wxPython build error

stahta01:
wxPython building errors out from use of "-mwindows"; this change removes "-mwindows" in CLANG64/32 packages.

### DIFF
--- a/mingw-w64-wxwidgets3.2/PKGBUILD
+++ b/mingw-w64-wxwidgets3.2/PKGBUILD
@@ -14,7 +14,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-common"
          "${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-msw"
          "${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-gtk3")
 pkgver=${_wx_basever}.3
-pkgrel=1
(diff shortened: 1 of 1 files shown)


Biswa96:
AFAIK, that compiler error was fixed in latest version of llvm/clang. Also, `-mwindows` should not be removed entirely if it is required by the project. The compiler error happened because `-mwindows` was used with CFLAGS, it should be with LDFLAGS.

stahta01:
So, it looks like I need to figure out if "-mwindows" should be moved to --libs which is for linking.

stahta01:
> AFAIK, that compiler error was fixed in latest version of llvm/clang. Also, `-mwindows` should not be removed entirely if it is required by the project. The compiler error happened because `-mwindows` was used with CFLAGS, it should be with LDFLAGS.

I had to remove it from --libs (LDFLAGS) because it errors out; has the version of CLAND/LLVM with the fix had the binary built and uploaded?

Biswa96:
> has the version of CLAND/LLVM with the fix had the binary built and uploaded?

I think so, it was fixed in clang 17.0.2. Release notes https://discourse.llvm.org/t/llvm-17-0-2-released/73840 and commit https://github.com/llvm/llvm-project/commit/f6cf58eed9737cb8718465c4fb6917ca854b1bc3

PR #18896, opened 2023-10-19
libid3tag: Update to 0.16.3

lazka:
Move to new upstream which includes all the patches we had


lazka:
If someone is missing the static build, let me know and I'll have a look.

PR #18902, opened 2023-10-19
update pyqt6/pyqt6-sip to 6.5.3/13.6.0

MehdiChinoune:
PyQt6 doesn't support Qt6 6.6 yet.
Disable QtBluetooth bindings for now.


Biswa96:
> Disable QtBluetooth bindings for now.

Please explain the reason.

MehdiChinoune:
> > Disable QtBluetooth bindings for now.
> 
> Please explain the reason.

Because It fails to link. I said above it the reason Why. PyQt6 6.5.3 doesn't support Qt6 6.6.

PR #18903, opened 2023-10-19
keepassxc: new package

StableAgOH:
[KeepassXC](https://github.com/keepassxreboot/keepassxc) is a modern, secure, and open-source password manager.


StableAgOH:
@Biswa96 Sorry to bother you. When running KeepassXC for the first time, a configuration file will be generated. Where is the best place for it to be placed?

(Currently it is placed in `%LOCALAPPDATA%/keepassxc`)

Biswa96:
> (Currently it is placed in `%LOCALAPPDATA%/keepassxc`)

tl;dr no 1 rule is not to change default path. More info here https://www.msys2.org/docs/configuration/

StableAgOH:
> > (Currently it is placed in `%LOCALAPPDATA%/keepassxc`)
> 
> tl;dr no 1 rule is not to change default path. More info here https://www.msys2.org/docs/configuration/

Thanks for your reply. Then I think I'm ready for review.

Zxilly (approved):
lgtm

Tiphereth-A (approved):
LGTM

lazka:
This still uses a different configuration directory compared to the upstream installer. I have both installed and they don't use the same config.

StableAgOH:
> This still uses a different configuration directory compared to the upstream installer. I have both installed and they don't use the same config.

It's my negligence. Fixed.

lazka:
thanks!

PR #18906, opened 2023-10-20
suitesparse: update to 7.2.2

mmuetzel:


mmuetzel:
Checked with Octave and the test suite still passes.

Issue #18460, opened 2023-09-10
[python-psutil]  Packages (1) mingw-w64-x86_64-python-psutil-5.9.5-2 broken.


sendust:
### Description / Steps to reproduce the issue

Packages (1) mingw-w64-x86_64-python-psutil-5.9.5-2

run python code below...

import psutil
print(psutil.pid_exists(300))

### Expected behavior

Return should be True or False

### Actual behavior

*** stack smashing detected ***: terminated

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19042

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

lazka:
confirmed at least. stack trace:

```c
Thread 1 received signal ?, Unknown signal.
0x00007ffb3310b1ce in __fastfail (code=2) at D:/a/msys64/ucrt64/include/_mingw.h:605
605     D:/a/msys64/ucrt64/include/_mingw.h: No such file or directory.
(gdb) bt
#0  0x00007ffb3310b1ce in __fastfail (code=2) at D:/a/msys64/ucrt64/include/_mingw.h:605
#1  __stack_chk_fail () at C:/M/B/src/mingw-w64/mingw-w64-crt/ssp/stack_chk_fail.c:18
#2  0x00007ffb33101b58 in psutil_pid_exists (self=<optimized out>, args=<optimized out>)
    at psutil/_psutil_windows.c:80
```

lazka:
_Py_PARSE_PID is "L" (long long), and it uses it to write to a DWORD (unsigned long) which leads to a stack overflow: https://github.com/giampaolo/psutil/blob/77e5b7445748d30d22c0e3b2e651414da96a88b4/psutil/arch/windows/proc.c#L46

* pit_d doesn't exist in the Windows API, so CPython randomly defines it to int via https://hg.python.org/cpython/rev/0410bf251e10
* mingw-w64 still defines it as int64 for some reason which our autotools based build system picks up
* psutil randomly uses this type for unrelated things in various places.

Not quite sure what to do next here.

PR #18908, opened 2023-10-20
python-psutil: quick fix for psutil stack *** stack smashing detected ***

FMode:
Issue 18460
https://github.com/msys2/MINGW-packages/issues/18460


mmuetzel:
Wouldn't it be better to change the type of `pid` from `DWORD` to `pid_t` here?
https://github.com/giampaolo/psutil/blob/77e5b7445748d30d22c0e3b2e651414da96a88b4/psutil/arch/windows/proc.c#L43

FMode:
The patch is for MinGw only. DWORD is the datatype passed to the Win32Api. So currently it is converted from (mingw) python int to DWORD...

lazka:
Since this masks the only difference between mingw and msvc CPython headers I think this is OK as a workaround for now. It just needs to be moved into a separate patch file and in the PKGBUILD a comment linking to https://github.com/msys2/MINGW-packages/issues/18460

PR #18915, opened 2023-10-21
Upgrade kicad packages to version 7.0.8

stahta01:


stahta01:
Note: The UCRT64 Kicad does not load when ran on my hardware; but, the package without this changes also did not run.

Tim S.

PR #18921, opened 2023-10-21
yt-dlp: update to 2023.10.13

ognevnydemon:


ognevnydemon:
while at it can I add pycryptodomex as it's required?

lazka:
Yeah, I remember trying to package it (but not very hard), and giving up

PR #18932, opened 2023-10-22
cython: Update to 3.0.4; add cython0

lazka:


ognevnydemon:
sorry for tagging a lot, but there is problem with missing `"` :)

lazka:
ah, my replace regex was broken... thanks

PR #18944, opened 2023-10-23
rapidfuzz-cpp: Add package for python-rapidfuzz

Biswa96:


Biswa96:
@lazka This will touch python-levenshtein. Should I replace cython0 to cython there?

lazka:
sure

PR #18967, opened 2023-10-26
qt6-static: depend on harfbuzz

birdsTalks:
Qt6 need harfbuzz depends , or CMake an template project in QtCreator may fail. We hop that tiny change can slove this problem.
It would be appreciated if an expert could see if this initiative actually solved this failure :
[18966](https://github.com/msys2/MINGW-packages/issues/18966)


lazka:
@birdsTalks does it solve your problem (you can download the CI result and install the package with `pacman -U`) ?

PR #18983, opened 2023-10-27
gettext: update to 0.22.3

mmuetzel:
Remove a patch that was cherry-picked before.
Also remove a hunk from a patch that has been applied upstream.

Additionally, add a couple of patches to fix build issues. I tried to give a short motivation for those patches in the patch files directly.


MehdiChinoune:
Is it possible to merge this PR?

mmuetzel:
`gettext` has quite a few reverse dependencies. I don't know if people would like to run some tests before merging this.

I didn't test much apart from building it...

lazka:
xgettext sadly crashes with STATUS_STACK_BUFFER_OVERRUN here when using it in my projects (I don't have a small reproducer yet)

lazka:
@mmuetzel I think we need to sync https://src.fedoraproject.org/rpms/gettext/blob/rawhide/f/gettext-0.22-disable-libtextstyle.patch maybe?

mmuetzel:
I think I did that (and removed the hunk that has been applied upstream in the meantime).

lazka:
I see, Fedora doesn't have the newest version. Arch does though and seems to have synced it differently: https://gitlab.archlinux.org/archlinux/packaging/packages/gettext/-/blob/main/gettext-0.22-disable-libtextstyle.patch?ref_type=heads

(just a guess since the function there is in the stacktrace)

mmuetzel:
Ah. That hunk from the patch isn't what was applied upstream.
The patch:
``` diff
+static inline ostream_t
+noop_styled_ostream_create (ostream_t destination, bool pass_ownership)
+{
+  if (pass_ownership)
+    return destination;
+  else
+    return fdopen (dup (fileno (destination)), "w");
+}
```

Upstream:
```
static inline noop_styled_ostream_t
noop_styled_ostream_create (ostream_t destination, bool pass_ownership)
{
  if (!pass_ownership)
    /* Not supported without the real libtextstyle.  */
    abort ();
  return destination;
}
```
Yeah. That will abort...

lazka:
all my tests pass now

dscho:
This update causes a regression where formats like `%I64u` are no longer understood by `libintl_vsnprintf()`. Example:

```c
#include <stdio.h>
#include <stdint.h>
#include <libintl.h>

static int sn(char *buf, size_t max, const char *fmt, ...)
{
	va_list ap;
	int len;

	va_start(ap, fmt);
	len = libintl_vsnprintf(buf, max, fmt, ap);
	va_end(ap);

	return len;
}

int main(int argc, char **argv)
{
	char buffer[1024];

	printf("%d, '%s'\n", sn(buffer, 1024, "hello %I64u\n", (uint64_t)123), buffer);

	return 0;
}
```

With `mingw-w64-gettext` v0.21.1, this used to print:

```
9, 'hello 123'
```

With v0.22.3 (no rebuild necessary, just an updated `libintl-8.dll`), it now prints:

```
-1, '☻'
```

Since Git relies on the `%I` family of formats, this breaks Git for Windows thusly: https://github.com/git-for-windows/git-sdk-64/actions/runs/6727744023/job/18286014121#step:5:27

dscho:
The culprit is probably https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=a80b9021019fe311b81771d67138496f512ef35a which assumes that gnulib's `snprintf()` family of functions is behaving identically to the removed code, which is obviously false.

lazka:
uh, thanks.

I wonder if gnulib doesn't support this at all, or if some of the 100 configure things went wrong.

mmuetzel:
Looking at the documentation of the `snprintf` replacement function by gnulib, I don't see any mention of format specifiers like `I64`:
https://www.gnu.org/software/gnulib/manual/html_node/snprintf.html

They are referring to the following standard:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/snprintf.html

So, if we asked the gnulib developers, they'd probably answer that this is a "non-standard conformant" usage of the function...

Not sure what to do in that case. Just to check options: Would it be possible to use standard format specifiers in Git for Windows?

dscho:
> Just to check options: Would it be possible to use standard format specifiers in Git for Windows?

That assumes that everybody builds Git for Windows from scratch. Which is not the case.

No, we cannot cut corners here. By Hyrum's law, we have to reinstate  `%I64` handling in `libintl_*printf()`. If that be via a hack, then so be it. (We could transform `%I64u` to `%llu` on the fly, for example, something like `if (strstr(fmt, "%I")) { char *fmt2 = strdup(fmt); ... replace `%I64` -> `%ll` in-place ... use it ... free(fmt2); }`)

What we cannot do is to pretend that the status quo is somehow okay.

mmuetzel:
Playing devil's advocate: The `I64` prefix is a Microsoft extension to the interface of the printf functions in its CRT. It might never have been intended that the `libintl_*printf` functions work for this extension. Git for Windows might have fallen victim of Hyrum's law by relying on unspecified behavior.

dscho:
> The `I64` prefix is a Microsoft extension to the interface of the printf functions in its CRT. It might never have been intended that the `libintl_*printf` functions work for this extension.

Repeat after me: the `%I64u` flag used to work with `mingw-w64-gettext`'s `libintl_vsnprintf()` right up until version 0.21.1. Starting with 0.22.3, it does not work.

That's Hyrym's law right there. In plain sight.

dscho:
> Git for Windows might have fallen victim of Hyrum's law by relying on unspecified behavior.

Just to clarify what Hyrym's law says. It does not say that users do the wrong thing by relying on unspecified behavior. It says that maintainers are forced to reinstate behavior, even if it contradicts the documentation, when too many users have come to rely on the existing behavior. And there's nothing you can do about it: by virtue of offering a software that exposed a certain behavior for _years_, you are bound to keep that behavior working, whether you are happy about it or not does not matter one bit.

mmuetzel:
> Just to clarify what Hyrym's law says. It does not say that users do the wrong thing by relying on unspecified behavior. It says that maintainers are forced to reinstate behavior, even if it contradicts the documentation, when too many users have come to rely on the existing behavior. And there's nothing you can do about it: by virtue of offering a software that exposed a certain behavior for _years_, you are bound to keep that behavior working, whether you are happy about it or not does not matter one bit.

Good luck with that argument with the gnulib devs (which happens to be the same as the ones of gettext). 😞

MehdiChinoune:
I agree with @mmuetzel , GfW should fix their code instead of relying on undocumented/unspecified behavior.
Gettext maintainer said before that he will never support msys2.
Don't expect those GNU projects to do anything about it unless it breaks something on Linux.
Windows+(relying on unspecified behavior) => almost impossible to get any response.

dscho:
@MehdiChinoune please understand that Git for Windows is merely one active and collaborative partner.

The problem at hand is that the behavior of MSYS2's `mingw-w64-gettext` changed in a backwards-incompatible way that requires fixing in every consumer that relied on the previous behavior.

> Good luck with that argument with the gnulib devs (which happens to be the same as the ones of gettext). 😞

@mmuetzel But I was not talking about taking it up with the gnulib developers at all! I was talking about what we always do in MSYS2 when something like this happens: we add patches to `MINGW-packages`.

My GP demanded that I rest _a lot_ this week, therefore I had hoped that those responsible for introducing the breakage would pick up the task of adding [the previously-mentioned "hack"](https://github.com/msys2/MINGW-packages/pull/18983#issuecomment-1792569182), and I have not given up hope yet.

dscho:
> GfW should fix their code instead of relying on undocumented/unspecified behavior.

@MehdiChinoune I fear you're misunderstand the impossibility of what you just suggested.

If I "fix" Git for Windows' code, and somebody runs it with an older `mingw-w64-gettext` version, it will now break for them!

It would be fixable in Git for Windows' source code if this was truly a compile-time decision to make. But that's not the case now, is it? Switch out `libintl-8.dll` and the semantics of `printf()` change in incompatible ways. Switch it back, and it changes again. You cannot decide at compile-time what needs to be done, yet that's precisely what we would need to fix this.

_That_ is why backwards-compatibility is such a paramount principle. You don't break it. And if you break it, you fix it. If you cannot fix it properly, you add workarounds.

MehdiChinoune:
> If I "fix" Git for Windows' code, and somebody runs it with an older `mingw-w64-gettext` version, it will now break for them!

Isn't possible to use a documented behavior that works with both old and new versions of gettext.

mmuetzel:
Does Git require a C99 compatible compiler? Could you use format macro constants like `PRIu64` to maintain compatibility?
https://en.cppreference.com/w/c/types/integer

dscho:
> > If I "fix" Git for Windows' code, and somebody runs it with an older `mingw-w64-gettext` version, it will now break for them!
> 
> Isn't possible to use a documented behavior that works with both old and new versions of gettext.

@MehdiChinoune I think you still misunderstand. There are basically two standards on Windows how `printf()` formats specify 64-bit integers: The actual MSVC standard (`%I64d`, which is what MINGW should imitate, as the name states that is is a _minimal_ layer on top of Windows) and what is commonly used on Linux (`%lld`, which is not supported by all MSVC versions).

In other words, Git for Windows _does_ use what is documented. And what `libintl_printf()` used to implement, too.

> Does Git require a C99 compatible compiler? Could you use format macro constants like `PRIu64` to maintain compatibility?
> https://en.cppreference.com/w/c/types/integer

@mmuetzel you still operate under the assumption that this is a bug in Git for Windows. You are incorrect.

The bug is a regression in `mingw-w64-gettext`. And just because Git for Windows was hit by it early doesn't mean that _every single user of `mingw-w64-gettext`_ is potentially broken by the regression you introduced.

mmuetzel:
According to Microsoft's own documentation `ll` *is* a valid size specifier:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170

They document themselves that "The I (uppercase i), I32, I64, and w argument size modifier prefixes are Microsoft extensions and are not ISO C-compatible.".

Using the C compatible size specifiers should always work in Windows compatible code according to Microsoft's documentation. It would also work on other platforms. So, switching Git for Windows to be standard C-compliant would be a working fix if I understand correctly.

Could you show an example where that would not work?

MehdiChinoune:
> According to Microsoft's own documentation `ll` _is_ a valid size specifier: https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170

If I have understood him, This is only supported on UCRT which he is refusing to move to.

dscho:
> According to Microsoft's own documentation `ll` _is_ a valid size specifier:
> https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170

That's the very latest MSVC version. Previous MSVC versions mishandled `%llu`.

> Could you show an example where that would not work?

I cannot whip one example out right now, but there _is_ a reason why Git for Windows _had_ to use `%I64d`.

In any case, we're still not on the same page. Previous versions of `mingw-w64-gettext` supported `%I64d`. As of 0.22.3, the entire `libintl_*printf()` _fails_ with such a format.

That's a regression, fair and square. Regressions need fixing and not arguing about what documentation consumers should, and should not, rely on.

I'm on my way fixing it and will therefore stop responding here.

MehdiChinoune:
> In any case, we're still not on the same page. Previous versions of `mingw-w64-gettext` supported `%I64d`. As of 0.22.3, the entire `libintl_*printf()` _fails_ with such a format.
> 
> That's a regression, fair and square. Regressions need fixing and not arguing about what documentation consumers should, and should not, rely on.

Please, report it upstream.

mmuetzel:
> I'm on my way fixing it and will therefore stop responding here.

Thanks for working on a fix if you are needing it yourself so desperately.

> That's the very latest MSVC version. Previous MSVC versions mishandled `%llu`.

Fwiw, the oldest documentation I can find is for Visual Studio 2015:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-140
As far as I can see, it doesn't differ from the documentation for Visual Studio 2022 in that respect.

dscho:
A tentative fix (still in draft until I am confident that it works correctly) is here: https://github.com/msys2/MINGW-packages/pull/19047. I'll have to take a break now, as I'm still not healthy.

PR #18998, opened 2023-10-31
python-pyglm: Update to 2.7.1

podsvirov:


Biswa96:
Would it be possible to use source tarball from pypi? That way the pkgver change may not require. See the python file in PKGBUILD-template directory.

podsvirov:
> Would it be possible to use source tarball from pypi? That way the pkgver change may not require.

It is not difficult for me to track the version and I prefer to use the source code from the original repository.

PR #19018, opened 2023-11-04
mpv: switch to yt-dlp

lazka:
youtube-dl is basically dead


Biswa96:
> youtube-dl is basically dead

Not dead _yet_ - still in development 😞

lazka:
Would you prefer to keep it?

Biswa96:
>Would you prefer to keep it?

Not sure. Feel free to remove it.

lazka:
My motivation was that Arch dropped it too btw

PR #19020, opened 2023-11-05
suitesparse: update to 7.3.1

mmuetzel:
Also skip building factory kernels for GraphBLAS and rely on the JIT compiler instead as proposed by upstream.


mmuetzel:
Installed the build artifact for the MINGW64 environment and Octave's test suite is passing without any new errors.

ognevnydemon:
I find that suitesparse has dropped 160 MiB (~180 -> ~20). what happened?

mmuetzel:
> I found that suitesparse has dropped 160 MiB (~180 -> ~20). what happened?

See the original comment: The factory kernels of GraphBLAS are no longer built.

ognevnydemon:
> See the original comment: The factory kernels of GraphBLAS are no longer built.

thanks for the answer

PR #19026, opened 2023-11-06
exiv2: update to 0.28.1

kmilos:


Biswa96:
cmake files are now installed in /share instead of /lib. Is that ok? IIRC, /share is used for header only packages.

kmilos:
Good catch, I wasn't aware of that change. I think it should be in /lib/cmake as well, will work on a patch...

MehdiChinoune:
cmake search in both lib and share folders for config and module files.

kmilos:
> cmake search in both lib and share folders for config and module files

Indeed, but might still be worth keeping some consistency across one platform...

PR #19047, opened 2023-11-10
gettext: fix the regression where `%I64d` support was broken

dscho:
The `%I64` printf format family is a Windows-specific extension that has been handled for, like, eternity in `mingw-w64-gettext`. But the recent update to v0.22.3 broke that.

The reason for that is a story of two episodes:

- gnulib slipped in [a major refactoring](https://git.savannah.gnu.org/cgit/gnulib.git/commit/?id=480a59ba60fa0b43a1cebe218bbcea6a947f1e86) of the `printf()` format parsing into a commit whose purpose was to support new size specifiers.

  Unfortunately, the code handling `%I64` was converted incorrectly, where it previously skipped the `64` part before looking for the conversion specifier (correct) but now no longer does so (incorrect) ([look for `'I'`](https://git.savannah.gnu.org/cgit/gnulib.git/diff/lib/printf-parse.c?id=480a59ba60fa0b43a1cebe218bbcea6a947f1e86) and notice that the `cp += 3` was replaced with `cp++`).

- gettext saw [a major refactoring](https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=a80b9021019fe311b81771d67138496f512ef35a) with the intention to avoid duplicating gnulib's `printf()` logic, dropping its own code in favor of the latter.

The result is that as of `mingw-w64-gettext` v0.22.3, `%I64d` is _started_ to be parsed but then fails because it mistakes `6` for the conversion character (instead of recognizing `d` as such).

Work around this regression by skipping the `64` part again.

This patch needs to be upstreamed to the gnulib project, yet we need to fix the `mingw-w64-gettext` regression at a more timely pace and need to integrate the patch early.

### DIFF
--- a/mingw-w64-gettext/PKGBUILD
+++ b/mingw-w64-gettext/PKGBUILD
@@ -5,7 +5,7 @@ _realname=gettext
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.22.3
-pkgrel=1
+pkgrel=2
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.gnu.org/software/gettext/"
@@ -27,7 +27,8 @@ source=(https://ftp.gnu.org/pub/gnu/${_realname}/${_realname}-${pkgver}.tar.gz{,
         gettext-0.22-disable-libtextstyle.patch
         0021-replace-fsync.patch
         0022-libasprintf.patch
-        0023-gnulib.patch)
+        0023-gnulib.patch
+        0001-printf_parse-fix-off-by-two.patch)
 sha256sums=('839a260b2314ba66274dae7d245ec19fce190a3aa67869bf31354cb558df42c7'
             'SKIP'
             'cbc2f533012d646521afa20f8b256917fce040741ff42cf53fb6dd7123a6670a'
@@ -35,7 +36,8 @@ sha256sums=('839a260b2314ba66274dae7d245ec19fce190a3aa67869bf31354cb558df42c7'
             'a28a27192f336f0b0908bdbf840d3b19d7b587c4ac52cad635cb43e95eb3c78d'
             '380dbddee2f9e2feee4c1435e8a942b5d11d0125e60c3350ebb10c19b19011aa'
             'c354f6a7021069c99b90f1c6d6f6a4ccf40a01e9f6742b866df2b5a7286cb868'
-            '4f34906eeb535c74fa3f2936729b59c36d05d503a274e2b850fc770263c60b46')
+            '4f34906eeb535c74fa3f2936729b59c36d05d503a274e2b850fc770263c60b46'
+            'e6187cc1f1190fcb2267e4b811ade0049126e78dd8fd37d3a86a716dfdbf713e')
 validpgpkeys=('462225C3B46F34879FC8496CD605848ED7E69871'  # Daiki Ueno
               '9001B85AF9E1B83DF1BDA942F5BE8B267C6A406D') # Bruno Haible
 
@@ -62,7 +64,8 @@ prepare() {
   apply_patch_with_msg \
     0021-replace-fsync.patch \
     0022-libasprintf.patch \
-    0023-gnulib.patch
+    0023-gnulib.patch \
+    0001-printf_parse-fix-off-by-two.patch
 
   autoreconf -fiv
 }
@@ -103,6 +106,29 @@ build() {
   sed -s "s|${MINGW_PREFIX}|$(cygpath -m ${MINGW_PREFIX})|g" -i gettext-tools/config.h
 
   make
+}
+
+check () {
+  cd ${srcdir}/build-${MSYSTEM}
+
+  # ensure that the Windows-specific `%I64*` format family is recognized correctly
+  cat >test-I64d.c <<EOF
+#include <stdio.h>
+#include <stdint.h>
+#include <libintl.h>
+
+int main(int argc, char **argv)
+{
+	char buf[1024] = "(failure)";
+	int len = libintl_snprintf(buf, 1024, "hello %llu %I64u 0:%d\n", (uint64_t)123, (uint64_t)123, 0);
+
+	printf("%d, '%s'\n", len, buf);
+
+	return len < 0 ? 1 : 0;
+}
+EOF
+  gcc -Wall -g -o test-I64d.exe test-I64d.c ./gettext-runtime/intl/.libs/libintl.a || return 1
+  ./test-I64d.exe || return 1
 }
 
 package() {
--- a/mingw-w64-gettext/0001-printf_parse-fix-off-by-two.patch
+++ b/mingw-w64-gettext/0001-printf_parse-fix-off-by-two.patch
@@ -1,0 +1,36 @@
+From 45cd4b6d5f3db4ac26219ddde951fc09589e74a0 Mon Sep 17 00:00:00 2001
+From: Johannes Schindelin <johannes.schindelin@gmx.de>
+Date: Fri, 10 Nov 2023 12:16:02 +0100
+Subject: [PATCH] printf_parse: fix off-by-two
+
+In gnulib's 480a59ba60 (*printf-posix: ISO C 23: Support size specifiers
+'wN' and 'wfN'., 2023-03-24), a major refactoring hides a bug in the
+conversion of the code handling Windows' `%I64*` family of `printf()`
+formats: before the refactoring, the `64` part was skipped (as desired),
+but afterwards that part is not skipped and therefore the `6` is
+mistaken for a conversion character. Which is invalid, of course,
+causing the code to error out.
+
+Fix this by skipping the `64` part again.
+
+Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
+---
+ gettext-runtime/intl/gnulib-lib/printf-parse.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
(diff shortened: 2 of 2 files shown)


mmuetzel:
Looks good to me with your explanations.

If I understand Microsoft's documentation for the format specifiers for the `*printf*` family of functions correctly, there is also a `I` prefix that isn't followed by `32` nor `64` (for types `ptrdiff_t` or `size_t`).
I didn't have a look at the code surrounding your proposed change. So, please let me know if this question doesn't make sense: Is that case already covered? Or would it break with the proposed change?

dscho:
> > Please do, though, otherwise your review will not have much value.
> 
> You are really trying to make it as hard as possible to work with you...

I am sorry that you get that feeling. I am very familiar with how this feels in this context.

mingw-w64-gettext/PKGBUILD
```diff
@@ -105,6 +108,29 @@ build() {
...
+  # ensure that the Windows-specific `%I64*` format family is recognized correctly
+  cat >test-I64d.c <<EOF
+#include <stdio.h>
+#include <stdint.h>
+#include <libintl.h>
+
+int main(int argc, char **argv)
+{
+	char buf[1024] = "(failure)";
+	int len = libintl_snprintf(buf, 1024, "hello %llu %I64u 0:%d\n", (uint64_t)123, (uint64_t)123, 0);
```
mmuetzel, line 123: The gnulib code only tries to handle `%I64d` according to the inline comment.
``` suggestion
	int len = libintl_snprintf(buf, 1024, "hello %llu %I64d 0:%d\n", (uint64_t)123, (int64_t)123, 0);
```

dscho:
@lazka I noticed that the `check` function of `PKGBUILD` is skipped... apparently we [build with `--nocheck`](https://github.com/msys2/MINGW-packages/blob/945b8bc1d1fc37410678fe6a7323a9fdde44af33/.ci/ci-build.sh#L136)? Is that intentional?

ognevnydemon:
> @lazka I noticed that the `check` function of `PKGBUILD` is skipped... apparently we [build with `--nocheck`](https://github.com/msys2/MINGW-packages/blob/945b8bc1d1fc37410678fe6a7323a9fdde44af33/.ci/ci-build.sh#L136)? Is that intentional?

yes, checks are skipped now

mingw-w64-gettext/PKGBUILD
dscho: The [gnulib code](https://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/printf-parse.c#n536) pretty clearly intends to handle `%I64` as a synonym of `%ll`, expecting the conversion character next:

```c
[...]
#if defined _WIN32 && ! defined __CYGWIN__
            /* On native Windows, PRIdMAX is defined as "I64d".
               We cannot change it to "lld" because PRIdMAX must also
               be understood by the system's printf routines.  */
            else if (*cp == 'I' && cp[1] == '6' && cp[2] == '4')
              {
                if (64 / 8 > sizeof (long))
                  {
                    /* __int64_t = long long */
                    signed_type = TYPE_LONGLONGINT;
                    unsigned_type = TYPE_ULONGLONGINT;
                    pointer_type = TYPE_COUNT_LONGLONGINT_POINTER;
                    /* For backward compatibility only.  */
                    floatingpoint_type = TYPE_LONGDOUBLE;
                  }
                else
                  {
                    /* __int64_t = long */
                    signed_type = TYPE_LONGINT;
                    unsigned_type = TYPE_ULONGINT;
                    pointer_type = TYPE_COUNT_LONGINT_POINTER;
                  }
                cp++;
              }
#endif

            /* Read the conversion character.  */
            c = *cp++;
            switch (c)
              {
              case 'd': case 'i':
                type = signed_type;
                break;
              case 'b': case 'o': case 'u': case 'x': case 'X':
              #if SUPPORT_GNU_PRINTF_DIRECTIVES \
                  || (__GLIBC__ + (__GLIBC_MINOR__ >= 35) > 2)
              case 'B':
              #endif
                type = unsigned_type;
                break;
[...]
```

In other words, while the comment talks about `%I64d` specifically, the code itself speaks a more generic language, really trying to handle `%I64` plus any conversion character.

Since the regression was detected in Git for Windows, which uses `%I64u` extensively, it is only appropriate to test for that conversion in particular.

BTW there was a bug (maybe half a duck?) in this part of the code, I would have accepted a fix for that (but fixed it myself in the meantime): I mistakenly named the shell function `verify` which has no meaning in the `PKGBUILD` file.

dscho:
@ognevnydemon could you also answer the much more important question whether that is intentional (with the corresponding reasoning if it is)?

dscho:
> checks are skipped now, because they are usually failed

Oh wow. Is that considered an acceptable situation that checks fail before deploying the packages?

ognevnydemon:
> @ognevnydemon could you also answer the much more important question whether that is intentional (with the corresponding reasoning if it is)?

I asked several months ago, I received this answer
> Checks are disabled for CI because those take too much time or failed most of the time. See the `--nocheck` option in `.ci/ci-build.sh` file.

that's all thing I know

lazka:
All we have atm is integration tests which get run before deployment: https://github.com/msys2/msys2-tests

dscho:
> All we have atm is integration tests which get run before deployment: https://github.com/msys2/msys2-tests to make sure the basics work.

Well, okay. I would really like to get us to a point where we can drop that `--nocheck`, though. It does not exactly build confidence.

dscho:
I verified manually that this indeed fixes the regression. I plan on upstreaming the patch to the gnulib project soon-ish.

lazka:
> Well, okay. I would really like to get us to a point where we can drop that `--nocheck`, though. It does not exactly build confidence.

Yeah. Some issues (a bit off topic here.. sorry):

* We need to selectively enable it for packages which are known to pass.
* It's unclear which types of tests we should run, all of them (usually fails), just add some smoke tests which easily pass, or make it selectable. Should we run tests even if they take 30 minutes etc.
* Running tests works against the CI time limit blocks rebuilds.

dscho:
> Some issues (a bit off topic here.. sorry):
> 
>  * We need to selectively enable it for packages which are known to pass. Can be easily done though.
> 
> * It's unclear which types of tests we should run, all of them (usually fails), just add some smoke tests which easily pass, or make it selectable. Should we run tests even if they take 30 minutes etc.
> 
> * Running tests works against the CI time limit, blocking rebuilds.

Understandable.

Would it not make more sense, though, to do this by commenting out the respective package's `check` function (or insert a `return 0` at the beginning, with a comment) instead of a wholesale `--nocheck`?

That would not only be more selective, it would also document this issue better and leave breadcrumbs for volunteers to fix things "enough" to re-enable the checks.

In any case, on-topic again, I would like to see this fix merged and deployed soon.

lazka:
The patch lgtm, thanks

lazka:
> Would it not make more sense, though, to do this by commenting out the respective package's `check` function (or insert a `return 0` at the beginning, with a comment) instead of a wholesale `--nocheck`?

Yeah, I sometimes use the tests to compare the amount of failing things on major updates.

dscho:
> > Would it not make more sense, though, to do this by commenting out the respective package's `check` function (or insert a `return 0` at the beginning, with a comment) instead of a wholesale `--nocheck`?
> 
> Yeah, I sometimes use the tests to compare the amount of failing things on major updates, but commenting out works of course.

Obviously I understand that this is a rather big change for the MSYS2 project, because after `--nocheck` is dropped, every package update, however minor it may be, may need to comment out `check` manually.

Therefore I would understand if making that change would need to be discussed more widely, or even be postponed indefinitely.

dscho:
> I plan on upstreaming the patch to the gnulib project soon-ish.

[Submitted](https://lists.gnu.org/archive/html/bug-gnulib/2023-11/msg00103.html).

dscho:
> > I plan on upstreaming the patch to the gnulib project soon-ish.
> 
> [Submitted](https://lists.gnu.org/archive/html/bug-gnulib/2023-11/msg00103.html).

And lo and behold, not even 1h later, [it was accepted](https://lists.gnu.org/archive/html/bug-gnulib/2023-11/msg00106.html).

The "answer that this is a non-standard conformant usage of the function..." [someone expected](https://github.com/msys2/MINGW-packages/pull/18983#issuecomment-1792035180) never materialized. The [prediction that the maintainers would not accept that change only for MSYS2](https://github.com/msys2/MINGW-packages/pull/18983#issuecomment-1793324097) turned out to be untrue, thankfully. As was the [suggestion that I am insisting on undocumented behavior](https://github.com/msys2/MINGW-packages/pull/18983#issuecomment-1802058362). I am also delighted to report that the maintainer seemed not to find it [hard to work with me](https://github.com/msys2/MINGW-packages/pull/19047#issuecomment-1805668909).

Instead, the maintainer gratefully accepted the bug fix, applied it immediately to the main and also to the `stable` branch.

mmuetzel:
That's perfect! And demonstrates the power of open source: If an issue is discovered, different options are explored. And anyone who is bothered (the most) can find and propose a solution. 👍

lazka:
Starting with 0.22.5, upstream gettext now errors out again for the Windows specific format specifiers: https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commit;h=eef429bbccbff2bca08bafc2ff76f7ec2be93d2a

Not sure what to do. Should we patch it to use `ms_printf`, or revert, or?

dscho:
Oh no! Does the `ms_printf` annotation really work? (First time I hear about it...) If so, that'd be my preference.

lazka:
It matches the msvcrt ABI+syntax, so no, it will break for other C99 cases and long double.

I think revert is all we can do (or leave it as is and force users to ignore the warnings)

dscho:
> I think revert is all we can do (or leave it as is and force users to ignore the warnings for non-C99 specifiers)

I agree.

PR #19048, opened 2023-11-10
arrow: Update to 14.0.1

raulcd:


Biswa96:
Are you sure that this version update needs a rebuild? According to release page, this arrow update only fixes two bugs https://arrow.apache.org/release/14.0.1.html

raulcd:
Hi @Biswa96 , thanks, this release fixes a critical security vulnerability on the Python package (PyArrow) and we (the Apache Arrow team) wanted to have version consistency on the rest of our packages. I am not entirely sure how do you manage those on msys2 so I would follow your advice but if I want to publish our ruby gems for 14.0.1 I need the msys2 14.0.1 version published.
CC @kou in case you have other thoughts.

raulcd:
Some more context about the vulnerability here: https://arrow.apache.org/blog/2023/11/09/14.0.1-release/

lazka:
@raulcd in case there was a misunderstanding, I think the question is why groonga and gdal need to be rebuild as well?

raulcd:
Sorry, you are right I misunderstood. We have a shell script that helps us with the update and we update the groonga and gdal to be rebuilt when there is a new Arrow version. You are correct we probably don't need a rebuild. I'll remove those commits.

PR #19049, opened 2023-11-11
[new-package] arb 2.23.0

MehdiChinoune:


MehdiChinoune:
too bad, arb is now part of flint>3.0.0
https://flintlib.org/doc/history.html#flint-3-0-0

PR #19052, opened 2023-11-11
flint: update to 3.0.1

MehdiChinoune:


lazka:
I think the mingw-w64-arb removal was included by accident?

MehdiChinoune:
> I think the mingw-w64-arb removal was included by accident?

No, I added it just this morning, It won't hurt to remove it now.

lazka:
ah, I see, sounds good

PR #19054, opened 2023-11-11
octave: update to 8.4.0

mmuetzel:


mmuetzel:
I downloaded and installed the MINGW64 build artifact of this PR. The test suite of Octave is still passing without any unexpected errors for me.

PR #19056, opened 2023-11-11
suitesparse: Fix relocation issue with GraphBLAS JIT compiler

mmuetzel:
There is a relocation issue when it comes to the path to the compiler that is used for the JIT compilation in GraphBLAS. (Sorry, I didn't notice before. I accidentally tested with a locally compiled library without realizing.)

This should be fixed by cherry-picking the upstreamed change.


mmuetzel:
If I'm not fooling myself again, the GraphBLAS JIT compiler should be actually working now.

PR #19062, opened 2023-11-11
Pastix: update to 6.3.0

ImperatorS79:
The "0001-gitmodules.patch" is directly taken from AUR.

Unfortunately on my computer all tests fail with error code 0xc0000139, but for my c++ app pastix seems to work fine, so I do not really know what is happening.


ImperatorS79:
I don't understand why the CI fails at that step.

Biswa96:
> I don't understand why the CI fails at that step.

I have added a workaround for the issue in the latest commit here. The issue may not be related to upstream pastix project.

Biswa96:
> Unfortunately on my computer all tests fail with error code 0xc0000139

One potential reason for this issue could be that the test executables are unable to locate the DLLs in the build directory.

One thing that should be fixed in upstream is the following compiler warning.

```
pastix/spm/src/common.h: In function 'spm_setenv':
pastix/spm/src/common.h:117:53: warning: unused parameter 'overwrite' [-Wunused-parameter]
  117 | spm_setenv( const char *var, const char *value, int overwrite ) {
      |                                                 ~~~~^~~~~~~~~
```

Please let us know if this pull request is ready to merge.

ImperatorS79:
Ok for me.

PR #19065, opened 2023-11-12
sundials: update to 6.6.2

mmuetzel:
SuperLUDist needs OpenMP 4.5 or newer. The OpenMP from LLVM doesn't have support for that yet. So, disable OpenMP for SuperLUDist on the CLANG* environments.

Configuration fails with the following error otherwise:
```
-- Found OpenMP_C: -fopenmp=libomp  
  -- Found OpenMP_CXX: -fopenmp=libomp  
  -- Found OpenMP_Fortran: -fopenmp (found version "3.0") 
  -- Found OpenMP: TRUE (found version "3.0")  
  -- Checking whether OpenMP supports device offloading
  -- Checking whether OpenMP supports device offloading -- no
  CMake Error at cmake/macros/SundialsCMakeMacros.cmake:89 (message):
    ************************************************************************
  
    ERROR: The found OpenMP version does not support device offloading.
  
    ************************************************************************
    cmake/tpl/SundialsOpenMP.cmake:87 (print_error)
    cmake/SundialsSetupTPLs.cmake:36 (include)
    CMakeLists.txt:170 (include)
  
  
```

(So, it might be LLVM Flang which is limiting the max. OpenMP version?)


lazka:
> SuperLUDist needs OpenMP 4.5 or newer. The OpenMP from LLVM doesn't have support for that yet. So, disable OpenMP for SuperLUDist on the CLANG* environments.

I don't know anything about this, but clang defaults to openmp 5.1 in theory. Is that not a superset of 4.5?

mmuetzel:
> I don't know anything about this, but clang defaults to openmp 5.1 in theory. Weird..

Turns out this was (artificially) limited by LLVM Flang. But we don't actually need a Fortran compiler to build SUNDIALS. It is only used to automatically detect name mangling during configuration. (But we know the name mangling of our Fortran compilers.)

lazka:
ok, I just did a quick run and it seems as if FindOpenMP is broken. It sets OPENMP_FOUND, and also OpenMP_C_SPEC_DATE matching 5.1, but OpenMP_C_VERSION is empty.

lazka:
It looks like cmake doesn't know about 5.1? https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake#L437C7-L437C27

mmuetzel:
They have the following in their `SundialsTPLOptions.cmake`:
```cmake
# Advanced option to skip OpenMP device offloading support check.
# This is needed for a specific compiler that doesn't correctly
# report its OpenMP spec date (with CMake >= 3.9).
sundials_option(OPENMP_DEVICE_WORKS BOOL
                "Skip the OpenMP device offloading support check" OFF
                ADVANCED)
```

I guess with "specific compiler" they are referring to LLVM OpenMP?

lazka:
> It looks like cmake doesn't know about 5.1? https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake#L437C7-L437C27

I can confirm that if I add 5.1 and 5.2 the sundails configure check passes. Time for a cmake MR I guess. I wonder why I can't find anything related this this upstream.

> I guess with "specific compiler" they are referring to LLVM OpenMP?

I can't find any reason why in the git history (the option was named SKIP_OPENMP_DEVICE_CHECK previously)

mmuetzel:
> I can confirm that if I add 5.1 and 5.2 the sundials configure check passes. Time for a cmake MR I guess. I wonder why I can't find anything related this this upstream.

There probably aren't many distributions that are based on the LLVM libraries.
I agree that it should be fixed in CMake.

mmuetzel:
On another note: I'm always uncertain when it comes to adding `clangarm64` to `mingw_arch`. Should we just add it and wait if it compiles?

lazka:
> On another note: I'm always uncertain when it comes to adding `clangarm64` to `mingw_arch`. Should we just add it and wait if it compiles?

That's fine with me. Removing it again is easy.

mmuetzel:
> That's fine with me. Removing it again is easy.

Ok. Did that. 👍

lazka:
I usually commit the removal with a short log of the build error, if there was one.

mmuetzel:
@lazka: Thank you for fixing that in CMake.

I'll wait for #19067 and then remove the work-around again from this PR.

mmuetzel:
I downloaded and installed the build artifacts for MINGW64. With them, the test suite of Octave still passes without any unexpected errors.

Thanks again for tracking down and fixing the issue with CMake and LLVM OpenMP.

MehdiChinoune:
Before enabling any package on clangam64, at least make sure that its dependencies are already available.
MSMPI is not available for aarch64.

lazka:
well, now we know :) f6c0c528abcf940b53eaae53efbf205b

PR #19069, opened 2023-11-12
python-watchfiles: New package

podsvirov:


lazka:
The python package version is "0.0.0", which suggests that there is some dynamic version detection going which failed. Not sure what though.

podsvirov:
> edit: from what I see maturin should pull the version from cargo.toml.. weird.

Good catch. In Cargo.toml currently `0.0.0`. Should I add patch?

lazka:
Indeed. Yeah, I think patching is needed there.

lazka:
Ah, the tarball from https://pypi.org/project/watchfiles/ contains a proper version. Maybe use that instead?

podsvirov:
Okay!

PR #19073, opened 2023-11-13
Next attempt to build sundials for Windows on ARM

mmuetzel:
I *think* I got rid of all dependencies that require MSMPI. But I can't actually test on Windows on ARM (because I don't have access to it).
I just checked if that configuration would build for a MINGW64 environment. It does and all checks still pass.

There should be no functional difference for the already distributed binaries.

### DIFF
--- a/mingw-w64-octave/PKGBUILD
+++ b/mingw-w64-octave/PKGBUILD
@@ -23,7 +23,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-pcre2"
          "${MINGW_PACKAGE_PREFIX}-qhull"
          "${MINGW_PACKAGE_PREFIX}-suitesparse"
-         $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || echo "${MINGW_PACKAGE_PREFIX}-sundials")
+         "${MINGW_PACKAGE_PREFIX}-sundials"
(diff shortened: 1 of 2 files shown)


MehdiChinoune:
I think both petsc and superlu_dist could be built without mpi, why not trying to enable them first on clangarm64?

mmuetzel:
> I think petsc could be built without mpi, why not trying to enable it first on clangarm64?

It might be possible to build parts of PETSc without MPI. But it's pretty large. I wanted to get the rebuild of Octave out of the way first...
Would that be ok?

mmuetzel:
I tried to enable PETSc without MPI. But that doesn't work:
```
-- Checking if LAPACK works with SUNDIALS... OK
CMake Error at cmake/macros/SundialsCMakeMacros.cmake:89 (message):
  ************************************************************************

  ERROR: MPI is required for PETSc support.  Set ENABLE_MPI to ON.

  ************************************************************************
  cmake/tpl/SundialsPETSC.cmake:41 (print_error)
  cmake/SundialsSetupTPLs.cmake:149 (include)
  CMakeLists.txt:170 (include)


```

mmuetzel:
> because petsc were built with MPI enabled.
> Build petsc without MPI and retry.

That's not the reason. Check `SundialsPETSC.cmake`:
```
# Using PETSc requires building with MPI enabled
if(ENABLE_PETSC AND NOT ENABLE_MPI)
  print_error("MPI is required for PETSc support. Set ENABLE_MPI to ON.")
endif()
```

PR #19084, opened 2023-11-14
manim: Update to 0.18.0

lazka:


lazka:
pillow errors are the same as before, so no regression there

PR #18884, opened 2023-10-18
qemu: update to 8.1.2

heljkon:
see https://lists.gnu.org/archive/html/qemu-devel/2023-10/msg04870.html for details

PR #19085, opened 2023-11-15
qemu: update to 8.2.0

heljkon:


heljkon:
Ignoring clang64 failure this time, logged error was:
```
  curl: (92) HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)
  ==> ERROR: Failure while downloading https://download.qemu.org/qemu-8.2.0-rc0.tar.xz
      Aborting...
```

heljkon:
@lazka @Biswa96 @MehdiChinoune @jeremyd2019 
QEMU 8.2 was released.

QEMU now comes with plugins. Plugin builds failed for CLANG though.
I've put the plugins into $MINGW_PREFIX/lib/qemu/plugins and added some plugin tests to the test script.

heljkon:
> Maybe, just keep 32-bit disabled.

I removed 32-bit as proposed by @Biswa96, because I had build issues, no time to fix and I thought no one would use 32-bit, which resulted in with https://github.com/msys2/MINGW-packages/pull/18884.

When I realized https://github.com/msys2/MINGW-packages/issues/19200, I fixed the 32-bit builds in context of this PR. Why should I remove my fix now? I took some time to fix the MINGW32 build and there are still people who want to use 32-bit.

MehdiChinoune:
There were only one who complained about it, and he didn't give a convincing argument why he needs 32-bit qemu.
Anyway, If you insist on keeping mingw32, please remove clang32.

MehdiChinoune:
Fixes #19200

PR #19086, opened 2023-11-15
mesa: Trying to reproduce an unexpected build failure

pal1000:
in gallium megadriver that I encountered while testing 23.3.0-rc3 at pal1000/mesa-dist-win and then I kept encountering with 23.3.0-rc1 and 23.2.1


Biswa96:
Does this happen only with static build? Also, is it required to mention c_link_args and cpp_link_args options? Doesn't `--prefer-static` handle that job?

pal1000:
> Does this happen only with static build?

Yes.
> Also, is it required to mention c_link_args and cpp_link_args options? Doesn't `--prefer-static` handle that job?

In theory yes assuming all Mesaa3D dependencies static libraries are perfect and have no shared library dependencies creeping in, but that's not the case. I know for sure LLVM static libraries depend on zlib.

Biswa96:
Should `shared-llvm` be disabled for static linking?

pal1000:
Thanks for spotting that. I still have a bit of work to do to make CI reproduce the  same error I see on my Pettrosky VPS instance where I do the mesa-dist-win releases.

pal1000:
> Thanks for spotting that. I still have a bit of work to do to make CI reproduce

Done.

pal1000:
I figured out the root cause of this error
````
FAILED: src/gallium/targets/wgl/libgallium_wgl.dll
"c++" @src/gallium/targets/wgl/libgallium_wgl.dll.rsp
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/ucrt64/bin/../lib/libintl.a(relocatable.o):(.text+0xc0): multiple definition of `DllMain'; src/gallium/targets/wgl/libgallium_wgl.dll.p/wgl.c.obj:wgl.c:(.text$DllMain+0x0): first defined here
````
As this error points out there is a clash between gettext `jibintl library and [Mesa3D WGL](https://gitlab.freedesktop.org/search?group_id=1155&project_id=176&repository_ref=main&scope=blobs&search=DllMain). This regressed in #18983.

mmuetzel:
`libintl` now pulls in `relocatable.o` from gnulib. `libtool` has a hard time building shared and static libraries in one go when convenience libraries come into play. We'll probably need to build static and shared libraries separately for `gettext`.

I'm looking into it now.

mmuetzel:
@pal1000: Thank you for tracking this down to `gettext`. 👍

I opened #19155 that should be fixing this. (At least, it no longer exported that symbol when I built locally with that change.)

pal1000:
Excellent, [this got solved without me having to do any change at my end](https://github.com/pal1000/mesa-dist-win).

PR #19089, opened 2023-11-15
Update Kicad packages to version 7.0.9 and add MINGW32 and CLANG32 packages

stahta01:
Re-did kicad-library to use MSys2 cmake and make tools; they work to build 32 bit packages.

Made kicad to optionally depend on  kicad-library to match ArchLinux depends and to enable building of 32 bit packages.


MehdiChinoune:
> Re-did kicad-library to use MSys2 cmake and make tools; they work to build 32 bit packages.

Why?

stahta01:
> > Re-did kicad-library to use MSys2 cmake and make tools; they work to build 32 bit packages.
> 
> Why?

Because I could not get the 32 bit packages to build without it.
There was two problems, the one I remember was path logic related. It kept using the full path instead of the short path when installing the package.

Tim S.

mingw-w64-kicad-doc/PKGBUILD
```diff
@@ -15,11 +15,11 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-ca"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-pl"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-ru"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-zh")
-pkgver=7.0.8
+pkgver=7.0.9
 pkgrel=1
 pkgdesc="Documentation for KiCad (mingw-w64)"
 arch=(any)
-mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64') # 'mingw32' 'clang32')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64' 'mingw32' 'clang32')
```
MehdiChinoune, line 22: We agreed before to start dropping 32bit packages, not the opposite.

mingw-w64-kicad-doc/PKGBUILD
```diff
@@ -30,7 +30,11 @@ optdepends=("${MINGW_PACKAGE_PREFIX}-kicad-meta"
             "${MINGW_PACKAGE_PREFIX}-kicad")
 
 source=("https://kicad-downloads.s3.cern.ch/docs/kicad-doc-${pkgver}.tar.gz")
-sha256sums=('923752b5667f76cb4b3a9cbd00c14f36f1bf8cd3930d273c25d67331412f954b')
+sha256sums=('3e276fd4d5ca1a2cc11cb35770c3ef6a33e02ec4a20243b28d110b647abef00b')
+
+#prepare() {
+  #sha256sum kicad-doc-${pkgver}.tar.gz
+#}
```
MehdiChinoune, line 37: Please, delete these lines.

mingw-w64-kicad/PKGBUILD
```diff
@@ -44,29 +42,34 @@ makedepends=(
...
 )
+optdepends=()
 _doc=("ca" "de" "en" "es" "fr" "id" "it" "ja" "nl" "pl" "ru" "zh")
 for _doclang in ${_doc[@]}; do
   optdepends+=("${MINGW_PACKAGE_PREFIX}-${_realname}-doc-${_doclang}: for documentation (${_doclang})");
 done
+optdepends+=("${MINGW_PACKAGE_PREFIX}-footprints: for footprints libraries")
+optdepends+=("${MINGW_PACKAGE_PREFIX}-symbols: for symbols libraries")
+optdepends+=("${MINGW_PACKAGE_PREFIX}-templates: for template libraries")
+optdepends+=("${MINGW_PACKAGE_PREFIX}-packages3D: for 3D model libraries")
```
MehdiChinoune, line 55: Please, Fix packages name and put them in a single array.

mingw-w64-kicad/PKGBUILD
```diff
@@ -87,6 +90,8 @@ prepare() {
   # used for the production version of the package
   #apply_patch_with_msg \
     #007-ki-6.0-manifest-remove-win10-11-manifest-support.patch
+  # The PLUGIN_CONTENT_MANAGER allows UCRT kicad to run on my PC
+  #apply_patch_with_msg 091-Kicad-manager-build-the-PLUGIN_CONTENT_MANAGER-only-.patch
```
MehdiChinoune, line 94: I see the patch wasn't applied, So better to be removed.

stahta01: It took me more than two days to find that fix; I have decided you guys do not want this package fixed, good bye.

MehdiChinoune: You added a patch without applying it, How do you explain that?
Either apply it or remove it. That's what I was talking about.

mmuetzel: I can confirm that the current version of KiCad doesn't open in the UCRT64 environment.
If that patch fixes that for you, it would be best to apply it for everyone.

PR #19094, opened 2023-11-16
libxml2: Update to 2.11.6

lazka:


lazka:
(there is a v2.12, but that has some ABI/API changes, so needs a good look first)

PR #19100, opened 2023-11-17
kicad: update to 7.0.9

MehdiChinoune:
This is a clean version of #19089


mingw-w64-kicad/PKGBUILD
```diff
@@ -33,8 +33,6 @@ depends=(
   "${MINGW_PACKAGE_PREFIX}-zlib"
   $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-aarch64* ]] || echo "${MINGW_PACKAGE_PREFIX}-opencascade")
   "${MINGW_PACKAGE_PREFIX}-cairo"
-  "${MINGW_PACKAGE_PREFIX}-kicad-footprints"
-  "${MINGW_PACKAGE_PREFIX}-kicad-symbols"
```
mmuetzel, line 37: If I recall correctly, KiCAD didn't start correctly if these two libraries aren't installed. Is that no longer the case? Why have they been moved to `optdepends`?

MehdiChinoune: The author of #19089 claims that they are just optional dependencies
https://github.com/msys2/MINGW-packages/pull/19089/commits/817a9036c5916448fd747239f1ff005a606d5e57

mmuetzel:
I downloaded and installed the UCRT64 build artifacts from this PR. With them, the initial window now opens (compared to the current version of KiCad).
However, when I try to click anywhere, the window freezes until I kill the process.

The MINGW64 build artifacts seem to be working though.

MehdiChinoune:
I tried using kicad 7.0.8 and 7.0.9 on my PC, both of them open and run correctly!

PR #19117, opened 2023-11-19
CI: Run package-grokker on both ucrt64 and clang64

MehdiChinoune:
Or should it run on all environments?


lazka:
I guess we could run it on all, but this is a good start

PR #19128, opened 2023-11-19
yt-dlp: update to 2023.11.16

ognevnydemon:


ognevnydemon:
added 2 extra deps as required by `pip check` output

PR #19146, opened 2023-11-22
libbotan: Update to 3.2.0

lazka:
001-boost-link.patch: seems to no longer be needed (?)
002-winsock-link.patch: refresh
003-enable-shared.patch: pushed upstream

openmp support is gone, so remove the build option and dependency


lazka:
If someone still needs libbotan2, let us know, and we can bring it back as a separate package if needed.

PR #19150, opened 2023-11-22
libinih: fix version mismatch

kmilos:


mmuetzel:
Is it really necessary to cherry-pick this change? Afaict, all compilers distributed by MSYS2 already default to C++11 or newer.

Biswa96:
This fixes the version number of the library.

kmilos:
> This fixes the version number of the library.

This - the pkg-cofig file currently has `Version: 56`.

mmuetzel:
Ah, I see. I missed that part. 👍

kmilos:
Btw, let me add the ref to the Cygwin package as well in a sec...

kmilos:
> Please add the archlinux also https://archlinux.org/packages/core/x86_64/libinih/

No need? The package name is the same for Arch and MSYS2 (on purpose), so they're already linked, no?

Biswa96:
According to the _developer_, if it is absent "msys2-web will try to guess based on the naming". So, we could help poor msys2-web from that guesswork.

PR #19152, opened 2023-11-22
msitools: Make DATADIR and LIBDIR relative to runtime installation location

jhol:
This patch-set supersedes an earlier attempt: https://github.com/msys2/MINGW-packages/pull/18611


lazka:
thanks, looks good.

I see that LOCALEDIR is used in one more place here: https://github.com/GNOME/msitools/blob/4f36db7ba241fc8503fa85039b6d34280b126764/tools/msiextract.vala#L155

Should that also be adjusted?

jhol:
> I see that LOCALEDIR is used in one more place here: https://github.com/GNOME/msitools/blob/4f36db7ba241fc8503fa85039b6d34280b126764/tools/msiextract.vala#L155
> 
> Should that also be adjusted?

Good catch. I patched that one also now.

lazka:
thanks!

PR #19155, opened 2023-11-22
gettext: build static and shared libraries separately

mmuetzel:
In recent versions of `gettext`, `relocatable.o` is no longer linked into `libintl.a` directly, but indirectly via the (convenience) library `libgnu.a`.
While two versions of `relocatable.o` are still built (one that exports `DllMain` and one that doesn't), only one of those object files (the one that exports `DllMain`) is linked into the convenience library `libgnu.a`. That means that the *static* library `libintl.a` (that links all objects from that convenience library) now also exports `DllMain`.

Prevent that from happening by building static and shared libraries in separate steps.

This should be fixing the issue described in #19086.

### DIFF
--- a/mingw-w64-gettext/PKGBUILD
+++ b/mingw-w64-gettext/PKGBUILD
@@ -5,7 +5,7 @@ _realname=gettext
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.22.4
-pkgrel=2
+pkgrel=3
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.gnu.org/software/gettext/"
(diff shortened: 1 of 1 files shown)


mmuetzel:
I downloaded and installed the MINGW64 build artifact from this PR. With them, `libintl.a` no longer exports `DllMain`:
```
$ nm -C /mingw64/lib/libintl.a | grep DllMain

$
```

The static libraries `libgettextlib.a` and `libgettextsrc.a` are no longer installed with this change. But that seems to be intended by upstream:
https://github.com/autotools-mirror/gettext/blob/ba381585217f86f5e380fafa7dc5e5086287b612/gettext-tools/configure.ac#L133-L138
``` autoconf
dnl Package-private libraries do not need to be installed when we build only
dnl static libraries. Don't install them in this case. This avoids a build
dnl error on AIX with --enable-shared, when some other version of GNU gettext,
dnl built with --disable-shared, is already installed in the same $prefix and
dnl the CFLAGS or LDFLAGS happen to contain "-L$prefix/lib".
AM_CONDITIONAL([INSTALL_PRIVATE_LIBRARIES], [test "$enable_shared" = yes])
```

PR #19157, opened 2023-11-23
ruby: fix the build with recent llvm windres

lazka:
This fails now:

windres.exe --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc a.rc -o a.o

We should probably report it to upstream to see what they say. This patch just removes the --preprocessor args, since the windres defaults are equivalent anyway.

### DIFF
--- a/mingw-w64-ruby/81a1fca366d9f5b49836e42acdb19f3ff0c8173c.patch
+++ b/mingw-w64-ruby/81a1fca366d9f5b49836e42acdb19f3ff0c8173c.patch
@@ -1,0 +1,60 @@
+From 81a1fca366d9f5b49836e42acdb19f3ff0c8173c Mon Sep 17 00:00:00 2001
+From: Pierrick Bouvier <pierrick.bouvier@linaro.org>
+Date: Tue, 21 Nov 2023 16:13:21 +0400
+Subject: [PATCH] [win32] detect llvm-windres (used for windows-arm64)
+
+When adding preprocessor option for llvm-windres (using clang as
+parameter), it fails. Thus, do not add this.
+
+It's needed to be able to compile windows-arm64 version, because MSYS2
+toolchain is LLVM based (instead of GCC/binutils).
+---
+ configure.ac          |  4 ++++
+ cygwin/GNUmakefile.in | 11 ++++++++---
+ 2 files changed, 12 insertions(+), 3 deletions(-)
+
+diff --git a/configure.ac b/configure.ac
+index 805cbaae18ee2..588fdec3827a6 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -512,6 +512,9 @@ AS_CASE(["$target_os"],
+ ],
+ [hiuxmpp*], [AC_DEFINE(__HIUX_MPP__)])    # by TOYODA Eizi <toyoda@npd.kishou.go.jp>
+ 
++USE_LLVM_WINDRES=no
++windres_version=`windres --version | grep LLVM`
++test -z "$windres_version" || USE_LLVM_WINDRES=yes
+ 
+ AC_PROG_LN_S
+ AC_PROG_MAKE_SET
+@@ -4245,6 +4248,7 @@ AC_SUBST(MINIOBJS)
+ AC_SUBST(THREAD_MODEL)
+ AC_SUBST(COROUTINE_TYPE, ${coroutine_type})
+ AC_SUBST(PLATFORM_DIR)
++AC_SUBST(USE_LLVM_WINDRES)
+ 
+ firstmf=`echo $FIRSTMAKEFILE | sed 's/:.*//'`
+ firsttmpl=`echo $FIRSTMAKEFILE | sed 's/.*://'`
+diff --git a/cygwin/GNUmakefile.in b/cygwin/GNUmakefile.in
+index 09298590304b5..192a8cc7112b3 100644
+--- a/cygwin/GNUmakefile.in
++++ b/cygwin/GNUmakefile.in
+@@ -3,9 +3,14 @@ gnumake = yes
+ include Makefile
+ 
+ ENABLE_SHARED=@ENABLE_SHARED@
+ DLLWRAP = @DLLWRAP@ --target=@target_os@ --driver-name="$(CC)"
+-windres-cpp := $(CPP) -xc
+-windres-cpp := --preprocessor=$(firstword $(windres-cpp)) \
+-	$(addprefix --preprocessor-arg=,$(wordlist 2,$(words $(windres-cpp)),$(windres-cpp)))
++ifeq (@USE_LLVM_WINDRES@,yes) # USE_LLVM_WINDRES
++	# llvm-windres fails when preprocessor options are added
++	windres-cpp :=
++else
++	windres-cpp := $(CPP) -xc
++	windres-cpp := --preprocessor=$(firstword $(windres-cpp)) \
++		$(addprefix --preprocessor-arg=,$(wordlist 2,$(words $(windres-cpp)),$(windres-cpp)))
++endif
+ WINDRES = @WINDRES@ $(windres-cpp) -DRC_INVOKED
+ STRIP = @STRIP@
+ 
--- a/mingw-w64-ruby/PKGBUILD
+++ b/mingw-w64-ruby/PKGBUILD
@@ -10,7 +10,7 @@ _realname=ruby
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.1.3
-pkgrel=2
+pkgrel=3
 pkgdesc="An object-oriented language for quick and easy programming (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
@@ -32,7 +32,8 @@ source=("https://cache.ruby-lang.org/pub/ruby/${pkgver%.*}/${_realname}-${pkgver
         0003-fix-check-types.patch
         0007-nm-use-full-options.patch
         0008-ucrt-32-win11-insns.patch
-        0009-windows-arm64.patch)
+        0009-windows-arm64.patch
+        81a1fca366d9f5b49836e42acdb19f3ff0c8173c.patch)
 ## Populated by the updpkgprovs script
 provides=(
     "${MINGW_PACKAGE_PREFIX}-ruby-debug=1.6.3"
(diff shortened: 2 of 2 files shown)


lazka:
@hmartinez82 I'll leave the rest to you for now

hmartinez82:
I'll probably just wait for https://github.com/ruby/ruby/pull/8995 to be upstreamed :)

pbo-linaro:
Great to see this merged. I'm not sure it's the best approach (substitution for a variable in Makefile) but it's the only way I could make it work (I'm not experienced with autotools).

mstorsjo:
It's not that llvm-windres fails when adding `--preprocessor-arg` - that option should work fine.

The root issue is that binutils did a backwards incompatible change in how the `--preprocessor` option worked; before, it was documented that if you pass this option, it should be a string that does include the options `-E -xc-header -DRC_INVOKED` (those are the default, but if you set the option, you had to manually include them). This was changed, bugs were reported, e.g. https://sourceware.org/bugzilla/show_bug.cgi?id=27594, they never acknowledged that this was a breaking change but claimed that it wasn't supposed to work the way it did before, and updated the documentation to no longer say what it used to say: https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=5edb8e3f5ad8d74a83fc0df7f6e4514eed0aa77f and https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=749c700282097cf679ff019a9674d7c762f48619

This forced projects that were using `--preprocessor` to abandon using it, e.g. this change in ffmpeg, https://github.com/ffmpeg/ffmpeg/commit/f9626d1065c43f1d51afe66bdf988b9f33729440, as there is no way to use this argument that works across both older and newer binutils. (This isn't ancient binutils, but versions from before 2021.)

llvm-windres had implemented the option according to how it used to work, to make it work with e.g. ffmpeg which was using the option in the old form, before this change.

I haven't done the corresponding breaking change in llvm-windres yet, but I guess I should just do that at this point, as others using the old form of the option have been forced to adapt (or stick to old versions of binutils anyway), even if binutils themselves haven't acknowledged that this was a breaking change.

@pbo-linaro I'm curious in this case though; what was the actual output of llvm-windres before this change, as you just say that "llvm-windres fails when preprocessor options are added"? Can you copypaste the actual windres command that was executed by make? And can you manually re-execute that command while adding `-v` to see what exact preprocessing command it does execute? Thanks!

pbo-linaro:
```
+ make V=1
windres --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc -DRC_INVOKED --include-dir . --include-dir . --include-dir ./win32 aarch64-ucrt-ruby330.rc aarch64-ucrt-ruby330.res.o
llvm-rc: Preprocessing failed.
make: *** [GNUmakefile:54: aarch64-ucrt-ruby330.res.o] Error 1
+ windres --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc -DRC_INVOKED --include-dir . --include-dir . --include-dir ./win32 aarch64-ucrt-ruby330.rc aarch64-ucrt-ruby330.res.o -v
 clang aarch64-ucrt-ruby330.rc -o C:/work/recipes/ruby/ruby/msys64/tmp/preproc-269f33.rc -E -xc -D RC_INVOKED -I . -I . -I ./win32
llvm-rc: Preprocessing failed.
+ windres --version
llvm-windres, compatible with GNU windres
LLVM (http://llvm.org/):
  LLVM version 17.0.4
  Optimized build.
+ clang --version
clang version 17.0.4
Target: aarch64-w64-windows-gnu
Thread model: posix
InstalledDir: C:/work/recipes/ruby/ruby/msys64/clangarm64/bin
```

When running clang itself, it says the file `aarch64-ucrt-ruby330.rc` does not exist.
I found that removing all llvm-windres args solved the issue.
If you have a better fix or suggestion, you're welcome to contribute it to llvm and ruby :+1: +1

mstorsjo:
After having another, closer look at this issue...

> This forced projects that were using `--preprocessor` to abandon using it, e.g. this change in ffmpeg, [FFmpeg/FFmpeg@f9626d1](https://github.com/FFmpeg/FFmpeg/commit/f9626d1065c43f1d51afe66bdf988b9f33729440), as there is no way to use this argument that works across both older and newer binutils. (This isn't ancient binutils, but versions from before 2021.)

Actually, I somewhat misremembered; this particular way of using the `--preprocessor` option does work across both old and new binutils. When specifying `--preprocessor`, it overrides the default `-E -xc -DRC_INVOKED` options, so those has to be specified manually. Previously one could do that within the `--preprocessor` option string, but with newer binutils versions that's no longer possible.


> ```
> + make V=1
> windres --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc -DRC_INVOKED --include-dir . --include-dir . --include-dir ./win32 aarch64-ucrt-ruby330.rc aarch64-ucrt-ruby330.res.o
> llvm-rc: Preprocessing failed.
> make: *** [GNUmakefile:54: aarch64-ucrt-ruby330.res.o] Error 1
> + windres --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc -DRC_INVOKED --include-dir . --include-dir . --include-dir ./win32 aarch64-ucrt-ruby330.rc aarch64-ucrt-ruby330.res.o -v
>  clang aarch64-ucrt-ruby330.rc -o C:/work/recipes/ruby/ruby/msys64/tmp/preproc-269f33.rc -E -xc -D RC_INVOKED -I . -I . -I ./win32
> llvm-rc: Preprocessing failed.
> ```

Thanks! I tried doing the same myself, and I see that the issue is in how the options from `--preprocessor-arg` are ordered WRT the input file. See the command that gets executed; `clang aarch64-ucrt-ruby330.rc -o C:/work/recipes/ruby/ruby/msys64/tmp/preproc-269f33.rc -E -xc -D RC_INVOKED -I . -I . -I ./win32`. If I try to execute something like that, I get the following messages:
```
clang: warning: '-x c' after last input file has no effect [-Wunused-command-line-argument]
clang: warning: test.rc: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D RC_INVOKED' [-Wunused-command-line-argument]
```
So the issue is that the options passed with `--preprocessor-arg` need to be included before the input file name. Then this should work as expected. That should be fairly easy to fix.


> When running clang itself, it says the file `aarch64-ucrt-ruby330.rc` does not exist.

That's odd, I presume it's some file which is generated as part as some other rule when building?

> I found that removing all llvm-windres args solved the issue. If you have a better fix or suggestion, you're welcome to contribute it to llvm and ruby 👍 +1

No I think this fix is quite fine. I don't see why they would need to pass those options to binutils windres either, unless they explicitly want to use a non-default compiler executable for preprocessing (and I don't see why they would need to). So ideally they could fix this by just omitting all of those arguments. But in any case, the fix you made should be just fine.

pbo-linaro:
Initially, those options were added for windres 2.36 (https://github.com/ruby/ruby/commit/952b2dcc8804cbf094f8f626e04bcb3f8e151005). We can let it like this for now, no option needed seems the best for me, but I'm not sure if GNU windres accepts the same thing.

mstorsjo:
> Initially, those options were added for windres 2.36 ([ruby/ruby@952b2dc](https://github.com/ruby/ruby/commit/952b2dcc8804cbf094f8f626e04bcb3f8e151005)). We can let it like this for now, no option needed seems the best for me, but I'm not sure if GNU windres accepts the same thing.

GNU windres, in general, definitely accepts the same thing - one normally shouldn't need to manually specify the preprocessor to use.

It looks like it was added originally in https://github.com/ruby/ruby/commit/835a5edf52c2ffd321678aba45264f6be42fcd35, although I don't see an explicit reason for why it was added at the time.

lazka:
Just my two findings while testing things for this PR: *.rc gets deleted by the makefile if the builds fails, which is why it is missing afterwards. The "Preprocessing failed." error can be avoided by passing an absolute path of clang to windres instead of just "clang".

mstorsjo:
> The "Preprocessing failed." error can be avoided by passing an absolute path of clang to windres instead of just "clang".

Hmm, that's weird. Can you add a `-v` argument to the windres call here, check the output clang command and try running it manually as well (if you can manage to keep the .rc file without make deleting it)? As of my testing yesterday, it looks like the `--preprocessor-arg=-xc` gets passed after the input `.rc` file in the clang command, which means that clang doesn't pick it up, and doesn't know how to handle the `.rc` file.

lazka:
Just a minimal example outside of ruby. I can try with ruby again though if wanted.

```
user@desktop CLANG64 ~
$ cat a.rc
VS_VERSION_INFO VERSIONINFO
BEGIN
END

user@desktop CLANG64 ~
$ windres.exe --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc -v a.rc -o a.o
 clang a.rc -o C:/msys64/tmp/preproc-39c2f0.rc -E -xc
llvm-rc: Preprocessing failed.

user@desktop CLANG64 ~
$ windres.exe --preprocessor=$(which clang) --preprocessor-arg=-E --preprocessor-arg=-xc -v a.rc -o a.o
 C:/msys64/clang64/bin/clang a.rc -o C:/msys64/tmp/preproc-9b85ba.rc -E -xc
clang: warning: '-x c' after last input file has no effect [-Wunused-command-line-argument]
clang: warning: a.rc: 'linker' input unused [-Wunused-command-line-argument]

user@desktop CLANG64 ~
$ ls a.o
a.o
```

mstorsjo:
> Just a minimal example outside of ruby. I can try with ruby again though if wanted.
> 
> ```
> user@desktop CLANG64 ~
> $ cat a.rc
> VS_VERSION_INFO VERSIONINFO
> BEGIN
> END
> 
> user@desktop CLANG64 ~
> $ windres.exe --preprocessor=clang --preprocessor-arg=-E --preprocessor-arg=-xc -v a.rc -o a.o
>  clang a.rc -o C:/msys64/tmp/preproc-39c2f0.rc -E -xc
> llvm-rc: Preprocessing failed.
> 
> user@desktop CLANG64 ~
> $ windres.exe --preprocessor=$(which clang) --preprocessor-arg=-E --preprocessor-arg=-xc -v a.rc -o a.o
>  C:/msys64/clang64/bin/clang a.rc -o C:/msys64/tmp/preproc-9b85ba.rc -E -xc
> clang: warning: '-x c' after last input file has no effect [-Wunused-command-line-argument]
> clang: warning: a.rc: 'linker' input unused [-Wunused-command-line-argument]
> 
> user@desktop CLANG64 ~
> $ ls a.o
> a.o
> ```

Oh, indeed! Yes; if `--preprocessor` isn't specified, llvm-windres internally tries to look up the clang executable to run (looking for it in the same directory as llvm-windres itself, and in path, with a long list of fallback methods). But with `--preprocessor`, that command is used literally, with a subprocess execution API that doesn't look things up in the path.

So to improve things, llvm-windres should at least:
- Make it look up arguments from `--preprocess` in `$PATH`
- Place arguments from `--preprocessor-arg` before the input file
- Possibly: Adopt the binutils 2.36 breaking change to not treat `--preprocessor` as a list of arguments but as one single string pointing to the preprocessor executable.

Thanks for investigating further!

(And as repeated foot note; in normal use of windres, one shouldn't really need to use any of these options, unless one has a specific desire to use a different preprocessor executable.)

lazka:
fyi, I've backported the llvm windres changes in #19524

PR #19166, opened 2023-11-23
openssl: update to 3.2.0

MehdiChinoune:
package-grokker shows nothing.

### DIFF
--- a/mingw-w64-openssl/002-relocation.patch
+++ b/mingw-w64-openssl/002-relocation.patch
@@ -1,11 +1,11 @@
 --- a/crypto/build.info
 +++ b/crypto/build.info
 @@ -107,7 +107,7 @@
-         mem.c mem_sec.c \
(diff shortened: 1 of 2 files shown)


r4sas:
I'm proposing to revert that update, because it has broken EdDSA: https://github.com/openssl/openssl/issues/23075

Biswa96:
> I'm proposing to revert that update, because it has broken EdDSA

Would it be possible to wait for a fix in upstream?

r4sas:
Today we release new version of i2pd, so msys2 repository build will be broken.

MehdiChinoune:
If the current i2pd works, then there is no hurry to update.

r4sas:
Okay, in that case do not merge PR's with i2pd updates with using this PR as blocker till further updates.

orignal:
Use own EdDSA implementation with 3.2.0
https://github.com/PurpleI2P/i2pd/commit/0e98dd5c7049d79bdcdba341357c504e562eb61f

PR #19167, opened 2023-11-23
[new package] keepawake 0.4.5

ognevnydemon:


mingw-w64-keepawake-rs/PKGBUILD
```diff
@@ -0,0 +1,61 @@
...
+pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
+pkgver=0.4.5
+pkgrel=1
+pkgdesc="Keep your computer awake (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang32' 'clang64' 'clangarm64')
+url="https://github.com/segevfiner/keepawake-rs"
+license=('spdx:MIT')
+msys2_references=(
+    'aur'
```
Biswa96, line 14: Would you like to explain the reason for this blank reference, please?

lazka: See https://www.msys2.org/dev/pkgbuild/ "Defining a key without a value means there is no mapping and the package shouldn't be linked."

Ideas on how to improve things welcome

Biswa96: There is no keepawake-rs package in AUR.

lazka: Though I don't understand why this would be needed for this package?

ognevnydemon: because there is a [package](https://aur.archlinux.org/packages/gnome-shell-extension-keepawake-git) that also has `keepawake` on its name. I think it will otherwise link to it

Biswa96: That is a different package name. Only exact matches are show, AFAIK.

ognevnydemon: > There is no keepawake-rs package in AUR.

true, but sometimes I see that some packages has absolutely random link to AUR package just because it contains the same word as name

ognevnydemon: removed this for now

lazka: The auto linking only uses the full name

lazka:
Since the crate and the program are called just "keepawake" I wonder if we should just name it that too?

ognevnydemon:
> Since the crate and the program are called just "keepawake" I wonder if we should just name it that too?

done, returned the reference for AUR

PR #19172, opened 2023-11-24
libaec: update to 1.1.2

kmilos:
Hm, not sure why the CLANG64 grokker thinks all the DLL symbols were removed?


lazka:
The symbols are indeed gone.

lazka:
A random guess would be because clang since v16 (which was after the last rebuild of this package) respects hidden visibility.

kmilos:
Indeed, thanks. Created https://github.com/MathisRosenhauer/libaec/pull/32

kmilos:
There's also the related https://github.com/MathisRosenhauer/libaec/pull/31 so I'll just wait for the next version.

PR #19180, opened 2023-11-24
libplacebo: add Cflags.private to .pc file; use xxhash

lazka:
so that passing "--static --cflags" gives "-DPL_STATIC"


lazka:
@kasper93 would this help?

kasper93:
Probably. Upstream does that depending on the build type https://code.videolan.org/videolan/libplacebo/-/blob/795600a44b03fcd52c055981a403ad60ee5d027a/src/meson.build#L275-281

kasper93:
> While xxhash also provides a library, libplacebo inlines everything from xxhash,
so this is only a build time dependency.

@lazka: There is currently bug in upstream. Even though xxhash is inlined it is still added to .pc file as `Required.private`. This will be fixed upstream eventually, but before release, would you patch it here?

PR #19183, opened 2023-11-24
python-pivy: rebuild with newest soqt-qt5

podsvirov:


Biswa96:
Would you like to explain the reason for this change and the addition of 02-pivy-0.6.8-fake-header-cmath.patch file, please?

podsvirov:
The patch added to avoid swig error - unable to find... Upstream PR  https://github.com/coin3d/pivy/pull/111.

MehdiChinoune:
Could you build it against soqt-qt6 on 64bit environments.

podsvirov:
> Could you build it against soqt-qt6 on 64bit environments.

Looks like [freecad can use pivy](https://wiki.freecad.org/Pivy), but currently it build with qt5. This may be the cause of conflicts if we will use qt6 now.

MehdiChinoune:
> Looks like [freecad can use pivy](https://wiki.freecad.org/Pivy), but currently it build with qt5. This may be the cause of conflicts if we will use qt6 now.

It was built against qt6, before I moved it to qt5 because of python-pivy.
soqt is an optional dependency of pivy (it could be without it).

PR #19202, opened 2023-11-26
python-pywavelets: update to 1.5.0

kmilos:


lazka:
https://github.com/mesonbuild/meson/issues/12547

kmilos:
Hm... It builds fine on my (up to date) machine?!

lazka:
ah, wrong pkg-config, needs mingw one

kmilos:
Any reason why pkgconf isn't a dependency of meson-python then (or meson for that matter)? It's not listed as a dependency in the project docs...

lazka:
I guess in theory it should also work without it, assuming the meson project doesn't need it for other deps. But that's broken right now.

lazka:
But yeah, one could say it's such a common dependency that we might as well make it a default.

kmilos:
> But yeah, one could say it's such a common dependency that we might as well make it a default.

Seems to be the case when it comes to cmake...

lazka:
#19204

PR #19208, opened 2023-11-26
numba: fix exe launcher

lazka:
* don't strip
* don't call sed on the binary launcher

remove cython dep while at it, likely not needed


ognevnydemon:
> remove cython dep while at it, likely not needed

btw I've found that it's also not needed for llvmlite

PR #19217, opened 2023-11-28
xxhash: add static build

lazka:


mingw-w64-xxhash/PKGBUILD
```diff
@@ -28,20 +28,37 @@ build() {
     _extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
   fi
 
-  mkdir -p "${srcdir}/build-${MSYSTEM}" && cd "${srcdir}/build-${MSYSTEM}"
+  mkdir -p "${srcdir}/build-static-${MSYSTEM}" && cd "${srcdir}/build-static-${MSYSTEM}"
```
Biswa96, line 31: The directory name should be opposite.

lazka: thanks

PR #19221, opened 2023-11-28
ripgrep: update to 14.0.2

ognevnydemon:
~~- enabled clang32 (why wasn't it enabled?)~~ see https://github.com/msys2/MINGW-packages/pull/19221#issuecomment-1830400981
- cleanup
- fixed missing `--features 'pcre2'` in `cargo install`

### DIFF
--- a/mingw-w64-ripgrep/PKGBUILD
+++ b/mingw-w64-ripgrep/PKGBUILD
@@ -3,47 +3,64 @@
 _realname=ripgrep
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=13.0.0
-pkgrel=2
+pkgver=14.0.2
+pkgrel=1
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
> * enabled clang32 (why wasn't it enabled?)

The question why enabling it? No one uses 32bit OS today.

ognevnydemon:
> > * enabled clang32 (why wasn't it enabled?)
> 
> The question why enabling it? No one uses 32bit OS today.

agree, but we can then disable `mingw32` just for balance

jeremyd2019:
> > * enabled clang32 (why wasn't it enabled?)

My script that I used to enable things for clang32 in bulk was overly particular on the format of the mingw_arch line (regex matched most PKGBUILDs and that was good enough for this), and the lack of single quotes would have confused it.  Thereafter It was probably just overlooked from doing it manually.

> The question why enabling it? No one uses 32bit OS today.

👋

PR #19232, opened 2023-11-30
cppwinrt: generate and install header files

MehdiChinoune:


alvinhochun:
I do *not* recommend generating the headers from either the system-bundled WinMD files or the Windows SDK due to licensing issues. The only source I would use for the Windows.winmd file is from https://github.com/microsoft/windows-rs/tree/master/crates/libs/bindgen/default, which is explicitly licensed MIT or Apache-2.0.

I would suggest to consider using the [`-license` parameter](https://github.com/microsoft/cppwinrt/pull/1262) to add a license preamble to make it clear that the header is indeed MIT license. It can be the default template, or a custom template with SPDX comments like the following:

```
// SPDX-FileCopyrightText: Copyright (c) Microsoft Corporation
// SPDX-License-Identifier: MIT
```

I am also unsure about stuffing the winrt headers directly into `/prefix/include`. The officially recommended way to do it in MSVC is to have each projects generate the headers during build, but... well, I don't know what would be best for msys2.

MehdiChinoune:
What license issue are you talking about?

lazka:
Thanks for the input, I've created #19261

PR #19245, opened 2023-12-01
cargo-tauri: Update to 1.5.7

podsvirov:


jeremyd2019:
ooh, confirmed this Cargo.lock specifies ring 0.17.x.  Test case for clangarm64!  /me tries it...

PR #19247, opened 2023-12-01
cargo-tauri: enable clangarm64

jeremyd2019:
This uses ring 0.17 now, which is compatible with Windows on ARM64, so re-enable clangarm64.


jeremyd2019:
Successful CLANGARM64 CI run: https://github.com/msys2-arm/MINGW-packages/actions/runs/7063672548/job/19230146823

PR #19250, opened 2023-12-01
python-ipywidgets: Update to 8.1.1

podsvirov:


mingw-w64-python-jupyterlab-widgets/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname/_/-}")
+pkgver=3.0.9
+pkgrel=1
+pkgdesc="JupyterLab extension to use ipywidgets (mingw-w64)"
+url="https://github.com/jupyter-widgets/ipywidgets"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+msys2_references=(
+  'pypi: jupyterlab_widgets'
+)
```
MehdiChinoune, line 14: There is no such package on PyPi.
It's called jupyterlab-widgets
Also on Arch They called it the same https://archlinux.org/packages/extra/any/jupyterlab-widgets/

MehdiChinoune: I don't know!

lazka: Not easily right now... it fails for "pypi" only in https://github.com/msys2/MINGW-packages/actions/runs/7067651203/job/19241292399 since we use the references to fetch additional info from pypi.

I'll put it on my list

PR #19251, opened 2023-12-02
cmake: update to 3.27.9

MehdiChinoune:


MehdiChinoune:
gcc is now faster than clang!

ognevnydemon:
> gcc is now faster than clang!

finally! I always wondered, that UCRT64 is so slow...

PR #19253, opened 2023-12-02
libplacebo: remove xxhash dependency from .pc file

lazka:
it's a build time only dependency, see
https://github.com/msys2/MINGW-packages/pull/19180#issuecomment-1833353609

### DIFF
--- a/mingw-w64-libplacebo/PKGBUILD
+++ b/mingw-w64-libplacebo/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libplacebo
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=6.338.1
-pkgrel=3
+pkgrel=4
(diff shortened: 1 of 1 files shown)


lazka:
@kasper93 ^

This looks related to https://github.com/mesonbuild/meson/issues/11080

Maybe you can add that to the videolan gitlab, my account there is blocked and I can't log in anymore.

kasper93:
In this case it is more like https://github.com/mesonbuild/meson/issues/106 because include dirs of dependencies should be added to .pc (in cases they are referenced in public headers). Unless this is private dependency, internal one, then they should not be added. cmake can distinguish those, meson assumes everything is public.

https://github.com/mesonbuild/meson/issues/11080 is in similar area though. There are few issues reported related this dependency topic. meson has no way to customize behavior what should be exported and what not.

kasper93:
LGTM

You could wait for @haasn to approve and possibly merge it, then instead referencing PR.patch, we can directly reference commit.patch.

lazka:
We can always update it again.. :)

PR #17915, opened 2023-07-31
Update arm-none-eabi-gdb to v13.2-1 based on gdb v13.2-2

redXhammer:

PR #19259, opened 2023-12-03
gdb: Update to 14.1

lazka:
* remove patch included in the release
* work around broken "--with-foo" defaults by explicitely passing a prefix. thanks to lhmouse for the fix


MehdiChinoune:
patch wasn't deleted

MehdiChinoune:
you forgot avr-gdb.

mingw-w64-gdb/PKGBUILD
```diff
@@ -143,6 +134,9 @@ package_gdb() {
 package_gdb-multiarch() {
   pkgdesc="GNU Debugger (supports all targets)"
   depends=("${MINGW_PACKAGE_PREFIX}-${_realname}=${pkgver}")
+  provides=("${MINGW_PACKAGE_PREFIX}-arm-none-eabi-gdb=${pkgver}")
```
MehdiChinoune, line 137: is `=${pkgver}` required here?

lazka:
> you forgot avr-gdb.

oh, ok, thanks

mingw-w64-gdb/PKGBUILD
lazka: It makes version requirements work as expected, see https://wiki.archlinux.org/title/PKGBUILD#provides

We rarely use them though....

MehdiChinoune: `, unless at least one of them uses a conflicts array.`

lazka: The version makes things like "depends=(arm-none-eabi-gdb>=13)" continue to work.

I'm fine with removing it though, if you'd prefer.

selectstriker2:
> Could you please make gdb-multiarch provides&replaces arm-none-eabi-gdb and avr-gdb as suggested in #17915

This now prevents me from updating gdb-multiarch as I need to maintain a specific avr-gdb version for a testing purposes

lazka:
> I'm using gdb-multarch for ARM development so I need both versions of gdb. I don't believe they actually conflict with each other, so I'm not sure "replaces" is correct.

Yeah, in theory both can be installed, this was more meant for guiding users to the new package.

Can you expand a bit on why you need specific avr-gdb version? Is it just an older version, is it patched? why?

selectstriker2:
Yes, I need to maintain my AVR tools at a specific, older version (gcc 11.3.0-1, binutils 2.38-1, gdb-11.2-1)

MehdiChinoune:
> Yes, I need to maintain my AVR tools at a specific, older version (gcc 11.3.0-1, binutils 2.38-1, gdb-11.2-1) that were used for development of a safety certified product.

And you the community to maintain them for you?
I expect someone who is involved in developing a "safety certified" product to rely on paid/commercial support.
MSYS2 is not RHEL or SLE.

lazka:
> And you the community to maintain them for you?

I don't think that was implied, only that the "replaces" makes the old name no longer usable for their own packages.

@selectstriker2 just curious, are you building those old versions yourself, or are you just using our old packages?

selectstriker2:
> I don't think that was implied, only that the "replaces" makes the old name no longer usable for their own packages, since pacman will remove them on upgrades.

Correct, I now have a package conflict between gdb-multiarch and avr-gdb.

> @selectstriker2 just curious, are you building those old versions yourself, or are you just using our old packages?
I built the packages myself at the time because I was looking to use the latest (at the time) version of gcc (11.3.0), but the repo version was older (8.4.0?)

PR #19269, opened 2023-12-04
lcms2: update to 2.16

kmilos:


lazka:
Should we pass `-Dutils=true` to keep the executables?

kmilos:
> Should we pass `-Dutils=true` to keep the executables?

Definitely, sorry for missing the new option...

kmilos:
@lazka P.S. Thanks for the vote of confidence and the invite, but can't really commit to the project more in earnest at this time, sorry.

lazka:
No problem. There are no extra obligations, just more permissions, in case you reconsider.

PR #19314, opened 2023-12-08
rust: update to 1.74.1

filnet:


filnet:
MINGW64 build fails with:

```
Building bootstrap
     Compiling winapi-x86_64-pc-windows-gnu v0.4.0
     Compiling winapi v0.3.9
     Compiling proc-macro2 v1.0.60
     Compiling memchr v2.5.0
  error: linker `gcc` not found
    |
    = note: program not found
```

CLANG64 build fails with:

```
   ==> Starting build()...
  /C/_/B/PKGBUILD: line 104: clang: command not found
  Usage: cygpath (-d|-m|-u|-w|-t TYPE) [-f FILE] [OPTION]... NAME...
         cygpath [-c HANDLE] 
         cygpath [-ADHOPSW] 
         cygpath [-F ID] 
```

filnet:
Was something changed in CI or elsewhere that could explain that ?

mmuetzel:
Maybe, you'd need to add `"${MINGW_PACKAGE_PREFIX}-cc"` explicitly to `makedepends` now?

MehdiChinoune:
It's lldb that was depending on clang  which equal to cc on clang environments and depends on gcc on others. I split clang into clang-libs+clang and make lldb depends on clang-libs.
https://github.com/msys2/MINGW-packages/commit/e6f0258153c6ea0c01847111e370b83deb038b42

podsvirov:
It's possible also enable wasm32 target?  Like via rustup?

```
rustup target add wasm32-unknown-unknown
```

MehdiChinoune:
> It's possible also enable wasm32 target? Like via rustup?
> 
> ```
> rustup target add wasm32-unknown-unknown
> ```

By looking at Arch PKGBUILD, I think It's a compilcated process to enable wasm

PR #19317, opened 2023-12-08
zxing-cpp: update to 2.2.0

axxel:


Biswa96:
Thank you for the contribution ❤️

PR #19323, opened 2023-12-09
asio: update to 1.28.1, don't build standalone library

oscarfv:


lazka:
Note that this also drops the pkg-config files, not sure if that is relevant.

@podsvirov any thoughts on this, since you added the package?

podsvirov:
I have checked all two dependent packages (`restbed` and `krpc-cpp`) and both do not use `*.pc` files and can be built with the new version of `asio`.
I have the opinion that `asio` users should not rely on official support for any build system.
To summarize: let's just take and update these headers!

PR #19328, opened 2023-12-10
rtl-sdr: Change package source and update version to 2.0.1

antoniovazquezblanco:
As per requested in https://github.com/msys2/MINGW-packages/pull/17459 :)


Biswa96:
Thank you for the update ❤️

PR #19350, opened 2023-12-13
asio: Re-add building with autotools

Biswa96:
This reverts some hunks from 64bda161072457ec0a52c1ca3d52852b3cee0702 commit and installs pkgconfig file also.

See #19323 

@oscarfv @podsvirov


oscarfv:
Masking the fact that the build is broken makes me a bit uneasy. IMO it is a good thing to be explicit about the results we expect.

Biswa96:
My plan was to re-add the pkgconfig file without any workaround. So, I let autotools and make to do its thing and install the files manually.

PR #19357, opened 2023-12-13
krpc-cpp: Rebuild with upstream asio and protobuf

podsvirov:


Biswa96:
Would you like to explain the reason for this change?

podsvirov:
There is a namespace conflict: the global namespace pb is already declared in the updated protobuf headers.

Biswa96:
Thanks for the info. Did you report it to upstream?

MehdiChinoune:
> Thanks for the info. Did you report it to upstream?

https://github.com/krpc/krpc/issues/766

PR #19365, opened 2023-12-14
fix up some cmake config file path references

kmilos:


kmilos:
dcmtk build failure is a known issue, so dropping it until upstream has a fix: https://support.dcmtk.org/redmine/issues/1094

PR #19367, opened 2023-12-14
jasper: update to 4.1.1

kmilos:
Alternative/updated  patch for temp filename buffer overflow.
The JAS_ENABLE_AUTOMATIC_DEPENDENCIES build option has been removed.


Biswa96:
Would you like to add some commit message explaining the reasons behind the patch and option removal?

kmilos:
Hm, maybe dropping that patch is not 100% right - `L_tmpnam` is not the same as PATH_MAX for MSVCRT envs like it is for UCRT ones...

lazka:
I don't quite get why the old patch was dropped, it looks more correct. Jasper assumes the buffer is PATH_MAX large, so L_tmpnam is wrong there.

kmilos:
They made this change upstream, so it didn't apply cleanly anymore: https://github.com/jasper-software/jasper/commit/0a013da1e7f4cde51f7b4d9fa06e836a4b9c93b1

We could/should probably ask upstream to change this in a more "correct" way w/o using L_tmpnam...

lazka:
Ok, look more into it, since tmpnam in ucrt returns absolute paths (despite the docs at https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=msvc-170 still describing the old behavior) it probably fair to assume that it's something that can hold a random path, so despite being ugly MSVCRT is the odd one out here... so I guess this is fine.

kmilos:
The idea was that now we end up w/ 260 in any case...

PR #19368, opened 2023-12-14
gtk3: Import proposed patch to support high-DPI displays

Cimbali:
Patch is taken from:
https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3275

Changes allow gtk to scale its windows itself instead of having Windows do the scaling. In short  this fixes bluriness issues on high-DPI screens,which I’ve encountered when distributing gtk3 software on Windows, see Cimbali/pympress#302.


lb90:
Hi @Cimbali! 🙂 

The upstream MR is quite a partial solution: while it's true that windows won't be upscaled, they can become tiny. That can be tested with mixed DPI setups, having at least a monitor set to fractional scaling (e.g 150x)

Support for per-monitor DPI awareness in GTK3 could be implemented in two ways:

1. Render at next integer factor and let the compositor scale down (as with GTK3 on Linux)
2. Support per-window text scaling factors

Anyway, GTK4 is getting native support for rendering at fractional scales directly 😉

lb90:
Anyway, if your app can benefit from this, you don't need a patched GTK. Simply call `SetProcessDpiAwareness (PROCESS_PER_MONITOR_DPI_AWARE)` from your main function before `gtk_init()` or, better yet, add a manifest XML to opt-in for per monitor DPI awareness

Cimbali:
Thanks for the pointers @lb90 will investigate further 🙂

PR #19374, opened 2023-12-14
gitlab-cli: New package

podsvirov:


Biswa96:
Don't we already have glab https://packages.msys2.org/base/mingw-w64-glab ?

podsvirov:
Oops, thanks for note! :-)

PR #19378, opened 2023-12-15
mesa: Update to 23.3.1

pal1000:
dzn started failing to build for unknown reasons. Trying to build 23.3,0 also fails in dzn so ir's a bug in a dependency. Maybe we need to update Directx-headers to the untagged v1.611.1.


Biswa96:
> Maybe we need to update Directx-headers to the untagged v1.611.1.

Nope, just some casting fixes the error. I shall try to upstream the patch.

pal1000:
I am curious how 23.3.0 didn't fail to build 2 weeks ago, maybe directx-headers is still a factor.

PR #19384, opened 2023-12-16
qt6-pdf: Enable mingw64

podsvirov:


MehdiChinoune:
First, When you enable a new environment you don't have to bump pkgrel.
Second, We agreed also to push users towards UCRT64. Are you encoutering a problem with it.

podsvirov:
> First, When you enable a new environment you don't have to bump pkgrel.

Ok. Done.

> Second, We agreed also to push users towards UCRT64. Are you encoutering a problem with it.

A many years I use ming64, why usrt64 should be cool for me?

MehdiChinoune:
> > First, When you enable a new environment you don't have to bump pkgrel.
> 
> Ok. Done.
> 
> > Second, We agreed also to push users towards UCRT64. Are you encoutering a problem with it.
> 
> A many years I use ming64, why ucrt64 should be cool for me?

Because It's the future.

podsvirov:
> Because It's the future.

Okay. Currently we just enable mingw64 too.

MehdiChinoune:
Please, No.
I really don't like this behavior. Someone spends hours to build and fix a package and somone else comes and says "Well, I know better".
This is not like an environment like clangarm64 where someone has to make sure it builds to enable. I left for a reason. I did that so people like you move to UCRT64.
This is not the first package I have enabled only UCRT64 and CLANG64.

podsvirov:
I don't like being pushed. I like to make informed choices. Is there any public information comparing mingw64 and ucrt64 environments?

MehdiChinoune:
It's very strange to see a contributor not knowing about environments
https://www.msys2.org/docs/environments/
qt6-pdf is a new package, you could work without it like before.

podsvirov:
Thanks for the link. If the future is inevitable, then I will try to use the UCRT64 environment.

PR #19385, opened 2023-12-17
mingw-w64-directxmath - document ref to Arch's aur.

JPeterMugaas:


Biswa96:
When adding msys2_reference, it is not necessary to bump pkgrel.

PR #19386, opened 2023-12-17
mingw-w64-directxtex - Update to latest version and reference Arch's …

JPeterMugaas:
…aur.


brisingraerowing:
I just tried using the URL in the PKGBUILD, which doesn't work due to the trailing /. Removing it makes the URL work. Just a side note.

PR #19389, opened 2023-12-17
icu: Update to 74.2

lazka:


lazka:
Blocked by https://github.com/unicode-org/icu/pull/2749

PR #19438, opened 2023-12-18
ragel: rebuild

lazka:


MSP-Greg:
@lazka

JFYI, I work with a repo that checks ragel results with Ubuntu, macOS, & Windows in GHA.  For quite a while, all three platforms matched.

After this build, the line directives in the generated C code have the incorrect line numbers.  And, MSFT/vcpkg has a ragel package, and using it matches the new build from this PR.  Confused.  I just ran some quick local tests; I hope to look at it more tonite.

lazka:
For some extra context, this was rebuilt to get rid of some outdated signing keys of old packages, and the last rebuild was ~2018, so quite a while ago, so lots of things have changed since then.

PR #19471, opened 2023-12-18
update python-{spyder,spyder-kernels,qtconsole,lsp-server,lsp-jsonrpc,qdarkstyle} to {5.5.0,2.5.0,5.5.1,1.9.0,1.1.2,3.2.3}

MehdiChinoune:


mingw-w64-python-lsp-server/0001-python-lsp-server-1.5.0-allow-newlest-depends.patch
```diff
@@ -1,13 +1,11 @@
...
-@@ -13,7 +13,7 @@ readme = "README.md"
- license = {text = "MIT"}
- requires-python = ">=3.7"
+@@ -15,7 +15,7 @@ readme = "README.md"
  dependencies = [
--    "jedi>=0.17.2,<0.19.0",
+     "docstring-to-markdown",
+     "importlib_metadata>=4.8.3;python_version<\"3.10\"",
+-    "jedi>=0.17.2,<0.20.0",
 +    "jedi>=0.17.2",
```
Biswa96, line 8: This patch was not required probably. We've [jedi 0.19.1-1](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-python-jedi).

MehdiChinoune: for then future!

PR #19479, opened 2023-12-19
arrow: Update to 14.0.2

raulcd:


raulcd:
It seems that my PGP signature was failing:
```
      apache-arrow-14.0.2.tar.gz ... FAILED (unknown public key 275C532289DD0F4A)
```
I've updated with the new ones that have been added, see: https://dist.apache.org/repos/dist/release/arrow/KEYS

Biswa96:
Same question as I asked previously, does this arrow update require to rebuild dependencies (here groonga and gdal)? Dependencies are rebuilt when there is a breaking change like API/ABI changes.

raulcd:
oh, I thought as there are several critical bug fixes fixing some segmentation faults it might be required a rebuild. If it's only necessary on ABI/API change I'll remove those commits.

PR #19514, opened 2023-12-22
ruff: update to 0.1.9

ognevnydemon:


ognevnydemon:
added workaround of `--locked` issue as recommended by https://github.com/astral-sh/ruff/commit/a3e06e5a9d36bcc5708c3a720375594b7f19b2b4

ognevnydemon:
> added workaround of `--locked` issue as recommended by https://github.com/astral-sh/ruff/commit/a3e06e5a9d36bcc5708c3a720375594b7f19b2b4

this didn't help... will try to patch later

ognevnydemon:
done. should work now. also added completions

ognevnydemon:
build issue reported https://github.com/astral-sh/ruff/issues/9245

lazka:
I remember that old versions had similar issues, which prevented me to package it initially.

ognevnydemon:
> I remember that old versions had similar issues, which prevented me to package it initially. No idea though.

okay, just hope this will be fixed in future release

ognevnydemon:
there is new 0.1.11 release. I can't push it now

ognevnydemon:
> build issue reported https://github.com/astral-sh/ruff/issues/9245

update: the issue is caused by build profile. if I understand right, we just need to patch Cargo.toml with
```patch
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,7 +88,7 @@
 rest_pat_in_fully_bound_structs = "warn"

 [profile.release]
-lto = "thin"
+lto = "fat"
# Note that we set these explicitly, and these values
# were chosen based on a trade-off between compile times
# and runtime performance[1].
```

mati865:
Yup, but only for GCC based subsystems. Dunno about MINGW32, I cannot think of a reason why it wouldn't be affected but apparently it works.

ognevnydemon:
> Yup, but only for GCC based ~~subsystems~~ envs. Dunno about MINGW32, I cannot think of a reason why it wouldn't be affected but apparently it works.

I disabled 32-bit

MehdiChinoune:
I pushed 0.1.11

PR #19519, opened 2023-12-22
freerdp: Enable Azure AD authentication (WITH_AAD=ON)

MehdiChinoune:
after cjson package cmake files were fixed

### DIFF
--- a/mingw-w64-freerdp/0002-undef-interface.patch
+++ b/mingw-w64-freerdp/0002-undef-interface.patch
@@ -1,0 +1,12 @@
+--- a/libfreerdp/core/gateway/arm.c
++++ b/libfreerdp/core/gateway/arm.c
+@@ -55,6 +55,9 @@
+ //#define WITH_AAD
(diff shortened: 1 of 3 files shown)


mingw-w64-freerdp/0002-undef-interface.patch
```diff
@@ -0,0 +1,12 @@
+--- a/libfreerdp/core/gateway/arm.c
++++ b/libfreerdp/core/gateway/arm.c
+@@ -55,6 +55,9 @@
+ //#define WITH_AAD
+ #ifdef WITH_AAD
+ #include <cjson/cJSON.h>
++#ifdef interface
```
Biswa96, line 7: When encountering any compiler errors while using the `interface` macro, it is recommended to consider using the `#define _WIN32_LEAN_AND_MEAN` workaround.

MehdiChinoune: I don't know about it, thanks.

PR #19523, opened 2023-12-22
libjxl: update to 0.9.1

kmilos:
Dry run for now - there's quite a few dependencies to be checked and rebuilt as the API has changed.

darktable 4.6 was already proofed ahead for 0.9.x
gimp ?
gimp3 ?
graphicsmagick ?
imagemagick ?
kimageformats-qt5 ?
krita ?
libvips ?
python-imagecodecs ?
SDL2_image ?

### DIFF
--- a/mingw-w64-libjxl/PKGBUILD
+++ b/mingw-w64-libjxl/PKGBUILD
@@ -4,11 +4,12 @@
 _realname=libjxl
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.8.2
-pkgrel=2
+pkgver=0.9.1
+pkgrel=1
 pkgdesc='JPEG XL image format reference implementation (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+msys2_repository_url='https://github.com/libjxl/libjxl'
 url="https://jpeg.org/jpegxl/"
 license=('spdx:BSD-3-Clause')
 depends=("${MINGW_PACKAGE_PREFIX}-brotli"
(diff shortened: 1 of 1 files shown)


lazka:
Looking at the debian code search for the changed APIs, there are some already fixed, and some that likely need small patches,

kmilos:
python-imagecodecs v2024.1.1 should be ready for libjxl 0.9.x

lazka:
(We could disable it in krita for now if wanted)

kmilos:
> We could disable it in krita for now if wanted

I don't think there's any particular rush...

lazka:
all ready now

kmilos:
Let's wait another day or two - from the upstream backport PRs it looks like 0.9.1 is around the corner...

kmilos:
0.9.1 landed, so good to go now!

Unfortunately I won't manage to bump all the deps...

Cf. also https://archlinux.org/todo/libjxl-090/

Biswa96:
graphicsmagick failed to build with clang due to some weird issue.

libxml2 exports `xmlNanoFTPNewCtxt` function but it does not define `LIBXML_FTP_ENABLED` symbol in libxml/xmlversion.h. So, autoconf detects `xmlNanoFTPNewCtxt` at configure time but compiler does not find the `xmlNanoFTPNewCtxt` declaration in libxml/nanoftp.h.

kmilos:
> graphicsmagick failed to build with clang due to some weird issue

I guess this could be related to a recent libxml2 update, I've seen some other projects had to adapt (for different issues though).

Shall we ping GM upstream? I don't see this [reported over there](https://sourceforge.net/p/graphicsmagick/bugs/) yet...

kmilos:
> Or this could be a libxml2 clang build/packaging issue - both ucrt64 and clang64 libxml/xmlversion.h do not define `LIBXML_FTP_ENABLED`, but maybe default visibility attribute for clang builds exports all symbols while it shouldn't?

Hm, not it - exported symbols are the same in both DLLs, and neither exports `xmlNanoFTPNewCtxt`, so wonder how this check succeeded?!

Biswa96:
> neither exports `xmlNanoFTPNewCtxt`, so wonder how this check succeeded?!

See the libxml2 import library file libxml2.dll.a.

kmilos:
Maybe we drop `libxml2.syms` explicitly defined export table during libxm2 build and let the linker figure out the actual exported ones? Or patch to comment out disabled symbols?

lazka:
> Edit 2: We build libxml2 `--with-legacy`, maybe just turn it off?

Lots of removed symbols, but at least in MSYS2 there are no users: #19732

PR #19526, opened 2023-12-23
notepad++: new package, version 8.6

heljkon:


heljkon:
In my opinion it is ready to get merged.

It seems the first CLANG64-builds failed because g++ was not available (missing make-dependency "${MINGW_PACKAGE_PREFIX}-gcc")

Now the builds passed 3 times in a row. So there is still hope.

PR #19539, opened 2023-12-24
vulkan: update to v1.3.274

oltolm:
This is necessary for yuzu.


ognevnydemon:
`updpkgsums`

MehdiChinoune:
We have encountered some issues with Vulkan updates before.
I think we track the latest SDK tags `vulkan-sdk`, which still 1.3.268.0

lazka:
Looking at the release history, they release about every two months, so shouldn't be too far out hopefully.

lazka:
@oltolm update is in the repo now

PR #19545, opened 2023-12-25
[CI] run PKGBUILD's function CHECK if env var CI_MAKEPKG_RUN_CHECK is set

jannick0:
... plus commit from MSYS2 repo to `ci-build.sh` to keep things in sync.

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -65,7 +65,7 @@ install_packages() {
 list_dll_deps(){
     local target="${1}"
     echo "$(tput setaf 2)MSYS2 DLL dependencies:$(tput sgr0)"
-    find "$target" -regex ".*\.\(exe\|dll\)" -print0 | xargs -0 -r ldd | GREP_COLOR="1;35" grep --color=always "msys-.*\|" \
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
I still doesn't understand the goal of this PR. Running `check()` is disabled by default.
Do you want let contributors running `check()` in their forks?

jannick0:
Exactly - a very easy way to regularly run `CHECK` on branches pushed to forked repos, but toggling it is simple enough without fiddling around with commits first calling `CHECK` somehow in `PKGBUILD`, then removing the call again in a second forced push.  

Certainly, I am open to other feasible ideas for the issue when you optionally want `MAKEPKG` to run `CHECK` on a forked repo and save extra CI runs (which is a good thing for everyone per se) to generate PRs to the main repos.  **Any ideas anybody?**

And again: monitoring `CHECK` results is much more than a tedious exercise beyond updating a release number in `PKGBUILD` and looking that `BUILD` and `PACKAGE` run through somehow, it tremendously increases the reliable usability of those published packages, of course.  In my forks I definitely want `MAKEPKG` to run `CHECK` for those packages I boxed out of unmaintainable corners or I am looking after whenever I can spare time.

PR #19562, opened 2023-12-27
sl: new package

### LINKED ISSUE #18430: Add SL
### Package name

mingw-w64-sl

A joke ncurses program that displays a train every time you type `sl` instead of `ls`

I hacked together a [PKGBUILD](https://gist.github.com/That-Human-Being/3ae12404aa60566fcebf8a1564181035) from the Arch Linux package a while ago that looking back on now I see so many problems with (namely using the `ncurses` and `nurses-devel` packages, like I said this was very hacked together)

https://github.com/eyJhb/sl

Just run the command (Maybe with the disco or accident flags (`-d` and `-a` respectively)

### MINGW environments where the dependencies are available

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [X] CLANGARM64

Kreijstal:
This would fix #18430

### DIFF
--- a/mingw-w64-sl/PKGBUILD
+++ b/mingw-w64-sl/PKGBUILD
@@ -1,0 +1,33 @@
+pkgbase="mingw-w64-sl"
+pkgname="${MINGW_PACKAGE_PREFIX}-sl"
+pkgver=5.05
+pkgrel=1
+url="https://github.com/eyJhb/sl"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64' 'mingw32' 'clang32') 
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc")
+pkgdesc='Steam Locomotive runs across your terminal when you type "sl" as you meant to type "ls"'
+license=('custom')
+depends=("${MINGW_PACKAGE_PREFIX}-ncurses")
+source=(
+		"$pkgname-$pkgver.tar.gz::https://github.com/eyJhb/sl/archive/$pkgver.tar.gz"
+)
+sha512sums=('4a57737a60c90a1eec2210ce7c28db937b46a5c35006bd1b5ff43d20ef617a004c49c2b2897ebc0fb8c57b6687fe933b2c9834f5800fdc0bf9912e07542a3b30')
+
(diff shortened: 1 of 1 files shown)


mingw-w64-sl/PKGBUILD
```diff
@@ -0,0 +1,33 @@
+pkgbase="mingw-w64-sl"
+pkgname="${MINGW_PACKAGE_PREFIX}-sl"
+pkgver=5.05
+pkgrel=1
+url="https://github.com/eyJhb/sl"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64' 'mingw32' 'clang32')
```
Biswa96, line 7: Please remove 32 bit arch. https://www.msys2.org/news/#2023-12-13-starting-to-drop-some-32-bit-packages

mingw-w64-sl/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+depends=("${MINGW_PACKAGE_PREFIX}-ncurses")
+source=(
+		"$pkgname-$pkgver.tar.gz::https://github.com/eyJhb/sl/archive/$pkgver.tar.gz"
+)
+sha512sums=('4a57737a60c90a1eec2210ce7c28db937b46a5c35006bd1b5ff43d20ef617a004c49c2b2897ebc0fb8c57b6687fe933b2c9834f5800fdc0bf9912e07542a3b30')
+
+build() {
+	cd "sl-$pkgver"
+
+        cc -O3 -o sl sl.c -lncursesw -I "${MINGW_PREFIX}/include/ncurses/" -L "${MINGW_PREFIX}/lib"
```
Biswa96, line 20: Would it be possible to add CFLAGS, CPPFLAGS and LDFLAGS here?

Kreijstal: should I patch the makefile, to use these flags? or just in the PKGBUILD script?

Biswa96: The command is not using `make`. So the change should be in PKGBUILD only.

PR #19573, opened 2023-12-28
[new-package] python-versioneer 0.29

MehdiChinoune:
This package is required for the latest constantly and pandas python packages.
But It fails to build because the prefix disk `D:\` is different that source disk `C:\`.

### DIFF
--- a/mingw-w64-python-versioneer/001-fix-build.patch
+++ b/mingw-w64-python-versioneer/001-fix-build.patch
@@ -1,0 +1,11 @@
+--- a/setup.py
++++ b/setup.py
+@@ -125,7 +125,7 @@
+         s = Path("src/installer.py").read_text()
+         s = ver(s.replace("@VERSIONEER-INSTALLER@", v_b64))
+         with tempfile.TemporaryDirectory() as tempdir:
+-            installer = Path(tempdir) / "versioneer.py"
++            installer = Path(os.getcwd()) / "versioneer.py"
+             installer.write_text(s)
+ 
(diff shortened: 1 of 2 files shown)


Kreijstal:
setuptools bug?

Biswa96:
> I managed to install versioneer here 😄 

You did not add all the options in build() function in PKGBUILD file.

I managed to workaround the error with the following change.

```diff
--- a/setup.py
+++ b/setup.py
@@ -125,7 +125,7 @@
         s = Path("src/installer.py").read_text()
         s = ver(s.replace("@VERSIONEER-INSTALLER@", v_b64))
         with tempfile.TemporaryDirectory() as tempdir:
-            installer = Path(tempdir) / "versioneer.py"
+            installer = Path(os.getcwd()) / "versioneer.py"
             installer.write_text(s)
 
             self.package_dir.update({'': os.path.relpath(installer.parent)})
```

I can not explain why the issue happens. I just trace some files in ProcMon and saw path change with versioneer.py file.

Kreijstal:
which options are missing?

Biswa96:
> which options are missing?

Try to compare with the template file i.e. mingw-w64-PKGBUILD-templates/PKGBUILD.python.sh

Kreijstal:
ah it seems  without --no-isolation it builds in a virtual enviroment... however it seems to work fine. hmm but yeah that's a weird bug

MehdiChinoune:
Thanks for help

Biswa96:
May I ask a further investigation from @lazka ? :detective:

PR #19577, opened 2023-12-30
suitesparse: update to 7.4.0

mmuetzel:
The headers are installed in a subfolder now. That might need handling when updating dependencies.

(I can try to look into that when time permits.)


mingw-w64-suitesparse/PKGBUILD
```diff
@@ -18,10 +18,10 @@ depends=("${MINGW_PACKAGE_PREFIX}-openblas"
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja")
-source=(${_realname}-${pkgver}.tar.gz::"https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/v${pkgver}.tar.gz"
-        0001-relocate-jit-compiler.patch::https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/c697550beb30046c5078c318f2d7e58751b5b2fd.patch)
-sha256sums=('b512484396a80750acf3082adc1807ba0aabb103c2e09be5691f46f14d0a9718'
-            'df1e5a59d8eb6cd56dbd6f590c743c39f8afbc7aea721c9db84c6bc683b65d97')
+optdepends=("${MINGW_PACKAGE_PREFIX}-cc: GraphBLAS JIT compiler"
+            "sh: GraphBLAS JIT compiler")
```
Biswa96, line 22: There is always a shell in msys2 environment. It may not require to add sh in optdepends.

mmuetzel: Is that only needed when installing the package with pacman?  Or could that information also be useful when packaging for distributing outside of MSYS2?

mmuetzel:
I downloaded and installed the MINGW64 build artifact. With it, the test suite of Octave crashes when it reaches the first test that calls a function from libcholmod.
It looks like there is an ABI breakage from the previous version. I wonder why the package grokker didn't detect it.
How to find out if other reverse dependencies are affected as well?
Maybe, just rebuilt them all?

mmuetzel:
I rebuilt Octave (and SUNDIALS) locally. After that, Octave's test suite no longer crashes.
I've bumped the pkgrel of all reverse dependencies (apart from julia which is broken since a long time anyway).

mingw-w64-suitesparse/PKGBUILD
Biswa96: > that information also be useful when packaging for distributing outside of MSYS2

Maybe it would be useful in that scenario, I agree on that point.

mmuetzel:
Oops. I had the SuiteSparse headers installed in both the old and the new location when building SUNDIALS here.
Looking into the error...

mmuetzel:
I've downloaded the build artifacts of Octave, SUNDIALS, and SuiteSparse for MINGW64 from this PR. With them, the test suite of Octave passes without unexpected errors.
I'm assuming that rebuilding the other reverse dependencies would have fixed any potential ABI breakage, too.

mmuetzel:
Regarding the ABI breakage in this version, see also: https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/741

So, there will likely be another ABI breaking change coming soon (that will restore ABI compatibility with versions before 7.4.0).

PR #19579, opened 2023-12-30
Spectra: new package

JPeterMugaas:


raedrizqie:
this package has wrong dir structure

ognevnydemon:
> Could you please elaborate what is wrong with the dir structure of this package?

it's just called "spectra"

raedrizqie:
without 'mingw-w64-' prefix for its dirname?

mmuetzel:
> without 'mingw-w64-' prefix for its dirname?

Sorry. I still don't get the issue. Which file is installed in which directory? And where should it be installed instead?

Or does "dirname" mean something different for you? In that case, could you please explain what you mean?

mmuetzel:
Ah. Now I see what you mean. The `PKGBUILD` is in the wrong directory in the repository.

That is true. 👍

MehdiChinoune:
I think they mean cmake files, but cmake search in both lib and share directories for cmake config files.

MehdiChinoune:
I think they mean cmake files, but cmake search in both lib and share directories for cmake config files.

JPeterMugaas:
I think I see the problem.  It's directory the PKGBUILD was checked into.  It was named "spectra" instead of "mingw-w64-spectra".  Fixed in:  https://github.com/msys2/MINGW-packages/pull/19601

PR #19582, opened 2023-12-31
android-tools: Fix compiler error due to SetThreadDescription name collision

Biswa96:
SetThreadDescription was added in https://github.com/mingw-w64/mingw-w64/commit/3a137bd87e commit.

Thanks to MehdiChinoune who found this issue.


Biswa96:
This patch has been sent to upstream to review https://android-review.googlesource.com/c/platform/packages/modules/adb/+/2891118

Biswa96:
The patch has been added in upstream commit https://android.googlesource.com/platform/packages/modules/adb/+/5627a03c374fca9fa3076d1ec79df3590a89b2ab (though the diff is bit different).

PR #19587, opened 2024-01-01
mingw-w64-directxmath - Update to latest hotfix version.

JPeterMugaas:
Note that the version numbers did not increase this time.

### DIFF
--- a/mingw-w64-directxmath/PKGBUILD
+++ b/mingw-w64-directxmath/PKGBUILD
@@ -5,8 +5,8 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 # Version from CMakeLists.txt
 pkgver=3.1.8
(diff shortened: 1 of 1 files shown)


Biswa96:
Should we ask upstream to bump the version number?

JPeterMugaas:
I made a comment on a bug report to that effect.  See: https://github.com/microsoft/DirectXMath/issues/169 .

JPeterMugaas:
I hate to ask but I wonder if you can make an exception for this library and the newer dec2023 version even if the CMake project version numbers have NOT changed.  I ask because the Hotfix I mentioned earlier concerns mingw-w64 support.  It has the following:

Use two-argument cpuid when using recent MinGW -https://github.com/microsoft/DirectXMath/pull/172

add pkg-config support. - https://github.com/microsoft/DirectXMath/pull/174

This last item is important because a new version of DirectXTex also includes pkg-config support, and that support lists this library as a dependency.

PR #19589, opened 2024-01-01
feat: add picotool

thewh1teagle:
Add picotool package from
https://github.com/raspberrypi/picotool
Inspired by picotool arch Linux package.

Statically linked libraries so users can use it outside msys too.
Final binary size is ~1.32MB

### DIFF
--- a/mingw-w64-picotool/PKGBUILD
+++ b/mingw-w64-picotool/PKGBUILD
@@ -1,0 +1,49 @@
+# Contributor: thewh1teagle
+
+_realname=picotool
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver="1.1.2"
+pkgrel=1
+pkgdesc="Tool for inspecting RP2040 binaries and interacting with RP2040 devices."
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
> Statically linked libraries so users can use it outside msys too. Final binary size is ~1.32MB

MSYS2 is like any other packaging system, it relies on Dynamic linking.
If you want static libraries, you could build both. You could find some examples in other PKGBUILDs like https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aws-c-common/PKGBUILD

MehdiChinoune:
Also, please follow PKGBUILD-template https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball.sh

thewh1teagle:
@MehdiChinoune 
I havn't found example for PKGBUILD which build static + shared binary (not library) 
where should I place both?
should I build 2 and one binary name will be `picotool` and the second will be `picotool_static`?

Also, about the template format, I'm not sure where is the mistakes.

ognevnydemon:
> I havn't found example for PKGBUILD which build static + shared binary (not library) 
> where should I place both?
> should I build 2 and one binary name will be `picotool` and the second will be `picotool_static`?

a good example https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-sfml%2FPKGBUILD

MehdiChinoune:
I have fixed the PKGBUILD and merged the changes.

PR #19594, opened 2024-01-02
directxtex - update to latest version and depend upon OpenEXR.

JPeterMugaas:


Biswa96:
This one seems fine. The CMakeLists.txt[^1] file has correct 2.0.2 version.

[^1]: https://github.com/microsoft/DirectXTex/blob/dec2023/CMakeLists.txt

JPeterMugaas:
I find that this does compile if you insert the line `#undef __cpuid` before the `#include <psdk_inc/intrin-impl.h>` in /mingw64/include/intrin.h .

PR #19596, opened 2024-01-02
mingw-w64-uvatlas - new package.

JPeterMugaas:


Biswa96:
There is a newer dec2023 version.

Biswa96:
Shared library was not built in last iteration. Is that expected?

JPeterMugaas:
Yes, the library's. CMakeLists.txt only makes static versions.

PR #19598, opened 2024-01-02
Fix lxml crashing

lazka:
the build options forcing cython to run were no longer passed to setup.py which meant the cython patch preventing a crash was no longer included in the compiled C code, which re-surfaced crashes in the meson test suite for example.


eli-schwartz:
Yay, and of course the terrible setuptools PEP517 build layer just says nothing whatsoever about this when you get it wrong?

lazka:
Yeah, I think it just ignores unknown options

PR #19601, opened 2024-01-02
mingw-w64-spectra: Fix dirname for commit.

JPeterMugaas:


mmuetzel:
Oops. Looks like we've crossed.
Should already be fixed with this change: https://github.com/msys2/MINGW-packages/commit/0a7ff1899064f991f27852aa3795d086e9b2b9ab

PR #19607, opened 2024-01-04
feat: add serust

thewh1teagle:
Add `serust` 
Serial monitor CLI tool
used for debug microcontrollers over USB with serial connection,
or with arduino / raspberry pi, etc...
from https://github.com/thewh1teagle/serust

### DIFF
--- a/mingw-w64-serust/PKGBUILD
+++ b/mingw-w64-serust/PKGBUILD
@@ -16,10 +16,7 @@ sha256sums=('95e99134925e4ce4c7cccb0d0f7577327ed506b830bd1f220ff534162c78d812')
 
 prepare() {
   cd "${srcdir}/${_realname}-${pkgver}"
-
-  cargo vendor \
-    --locked \
-    --versioned-dirs
+  cargo fetch --locked
 }
 
 build() {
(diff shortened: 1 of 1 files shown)


ognevnydemon:
no need to vendor deps if you don't patch them. just use `cargo fetch --locked`

thewh1teagle:
Fixed

ognevnydemon:
> Fixed

sorry for late answer. also you need to remove `mingw32` and `clang32`

thewh1teagle:
@ognevnydemon 
removed `mingw32` and `clang32`, and added `email` on top.

ognevnydemon:
> added `email` on top.

it's not needed, also this one is incorrect because it's automatically created by github

lazka:
So this seems to be a small personal project that's not packages anywhere. I'm wondering if we should add it.

thewh1teagle:
It's already packaged into `Winget`. If you think you will add it, just reopen / merge.

lazka:
I was just curious what others were thinking about it, since I haven't worked with serial connections in a while, and can't judge how useful it would be for our users.

PR #19608, opened 2024-01-04
Fix problem with intrin.h that effected PR 19594.

JPeterMugaas:
See: https://github.com/msys2/MINGW-packages/pull/19594#issuecomment-1876069202 ,


MehdiChinoune:
Paches for mingw-w64 should be accepted upstream first. MSYS2 works very closely with mingw-w64 project.

JPeterMugaas:
I tried their mailing list but I did not see my report on their mailing list.  I can't submit a bug report because the administrator has not authorized me to submit one.  So I'm stuck making the patch I did.

MehdiChinoune:
About your patch, _cpuid is msvc specific. So instead of trying to mess with mingw-w64. Try to fix your code to use intrinsics used by mingw-w64. I have encountered this issue before.

PR #19609, opened 2024-01-04
mingw-w64-directxtex - Fixed, updated, and reuploaded.

JPeterMugaas:
Depend on OpenEXR.


MehdiChinoune:
Isn't possible to build it as a shared library?

JPeterMugaas:
The CMakeLists.txt file only makes a static version of the library.  I "guess" it could be possible to have it be a shared library if someone is willing to put in the work.  Shared libraries have their own complexities.

MehdiChinoune:
I see It's hardcoded as STATIC https://github.com/microsoft/DirectXTex/blob/8c86e0e4249e32a461f65d0ac8b43084b5d1fa49/CMakeLists.txt#L201

Kreijstal:
> Isn't possible to build it as a shared library?

This would be nice too

JPeterMugaas:
The author has made it very clear that they do not support making shared libraries and I respect that decision.

JPeterMugaas:
I have decided to make an updated package that includes more dependencies including one of the efxc2 program.  See: https://github.com/msys2/MINGW-packages/pull/19970 .

PR #19615, opened 2024-01-04
libheif: enable kvazaar

kmilos:


kmilos:
Upstream report: https://github.com/strukturag/libheif/issues/1089

Biswa96:
Interestingly, I can not reproduce the compiler error in libheif git repository and there is no change in that function compared to this libheif version.

kmilos:
CI seems to be happy now w/ the upstream patch...

PR #19616, opened 2024-01-04
mingw-w64-directxmath - Fix for cpuid.h being used instead of intrin.h.

JPeterMugaas:
This could effect things besides DirectxTex.


MehdiChinoune:
I think clang will start support cpuid.h with 18.x

MehdiChinoune:
It looks like It was supported before GCC-11 https://github.com/mingw-w64/mingw-w64/commit/d2374f898457b0f4ea8bd4084a94f2dafc87a99a

lazka:
What does this fix and should it be reported upstream?

MehdiChinoune:
> What does this fix and should it be reported upstream?

Fix #19609

JPeterMugaas:
I have reported the issue upstream.  See: https://github.com/microsoft/DirectXMath/pull/176 .

PR #19630, opened 2024-01-05
setuptools: use patched internal distutils by default

lazka:
To work around the included distutils in setuptools not supporting mingw, we changed the default of SETUPTOOLS_USE_DISTUTILS to make it use the stdlib distutils, which supports mingw obviously.

With Python 3.12 no longer providing a distutils this no longer works.

Instead use the pending distutils PR which adds mingw support to the new external distutils in https://github.com/pypa/distutils/pull/184 and patch the vendored version in setuptools.

This makes Python 3.12 work (outside of venvs and isolated build envs at least) and also gives the PR some real world testing when we build MSYS2 Python packages with it.

If this breaks something you can set SETUPTOOLS_USE_DISTUTILS=stdlib to get the old behaviour, but please report any issues either way.


lazka:
@naveen521kk fyi

naveen521kk (approved):
lgtm, it's better to have some testing before actually deploying ig.

lazka:
Various packages built today (not all use setuptools though) and no problems so far: https://github.com/msys2/MINGW-packages/pulls?page=1&q=is%3Apr+is%3Aclosed+created%3A2024-01-06

PR #19665, opened 2024-01-06
gtkada: update to 24.0.0

sad-poet:


MehdiChinoune:
There is a collection of adacore packages all of them are at 23.0.0.
I don't know if it's safe to update some to 24.0.0 while the others still at 23.0.0?
Maybe contact Adacore maintainers before.
ArchLinux aur update all of them to same version by checking tags and specified commits.

Biswa96:
@Fabien-Chouteau Would you like to provide any info if it is ok to update ada pacakges to version 24?

MehdiChinoune:
https://github.com/AdaCore/langkit/tags
https://github.com/AdaCore/libadalang/tags

MehdiChinoune:
The question is: Is it safe to update some to 24.0.0 and leave some in 23.0.0 (libadalang and langkit).

MehdiChinoune:
https://alire.ada.dev/crates/langkit_support
https://alire.ada.dev/crates/libadalang
Both are 23.0.0

Fabien-Chouteau:
Hi @sad-poet and @MehdiChinoune,

You should always take the same major version number for all libraries. For this specific update of GTKAda, there is no problem as there is no dependency on other AdaCore libraries. If you want to build libadalang `24.0.0` on the other hand, you have to use `24.0.0` releases for all the dependencies.

About the `ada-language-server` `24.0.3` "release", this is a preview and you should take the `24.0.0` tags instead.

PR #19681, opened 2024-01-08
mingw-w64-opencv: rebuild with cxx_std=17

### LINKED ISSUE #19677: Cannot load ONNX models using dnn

When attempting to load an ONNX model using dnn, the following exception is thrown:

```
"OpenCV(4.9.0) C:/M/B/src/opencv-4.9.0/modules/dnn/src/onnx/onnx_importer.cpp:4097: error: (-2:Unspecified error) DNN/ONNX: Build OpenCV with Protobuf to import ONNX models in function 'readNetFromONNX'\n"
```
It seems that opencv was built without protobuf support.
(see https://github.com/opencv/opencv/blob/9691a2dccf63f527122dc515b0a082d33e088572/modules/dnn/src/onnx/onnx_importer.cpp#L23)

Code snipet to reproduce:

```
try{
   cv::dnn::Net mNet {cv::dnn::readNet(onnxNetPath)}
 } catch(const std::exception &ex) {
   std::string what = ex.what();
   std::cout << what << std::endl;
}
```

### Expected behavior

No excpetion is thrown when loadig model.

### Actual behavior

Exception is thrown when model is loading.

### Verification

### Windows Version

MSYS_NT-10.0-19045

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

_No response_

joda01:
Set C++ standard to 17 to fix broken protobuf support.

Thank's to @Biswa96 who came with the link
https://github.com/opencv/opencv/blob/4.9.0/cmake/OpenCVFindProtobuf.cmake#L84

Closes #19677

### DIFF
--- a/mingw-w64-opencv/PKGBUILD
+++ b/mingw-w64-opencv/PKGBUILD
@@ -7,7 +7,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-python-${_realname}")
 pkgver=4.9.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Open Source Computer Vision Library (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


joda01:
Closes #19677

Biswa96:
Thank you for the contribution 🎉

The closes/fixes comment should be in the OP to close the issue automatically after merge.

PR #19707, opened 2024-01-11
imagemagick: Enable libtool option

Biswa96:
@MehdiChinoune @sboukortt


MehdiChinoune:
Could you just write a clear comment in front of options, so no one will ever consider removing it. like `# libtool files are required in imagemagick to relocate modules path.`

Biswa96:
Sure. Thanks for the review.

Biswa96:
I have verified that adding libtool file fixes the following command.

```
$ identify foobar.png
foobar.png PNG 980x350 980x350+0+0 8-bit sRGB 4533B 0.008u 0:00.007
```

PR #19710, opened 2024-01-11
coin: update to 4.0.2

ognevny:


ognevny:
> it's better to do it separately

it's already at 1.6.2 https://packages.msys2.org/base/mingw-w64-soqt

podsvirov:
@ognevny, thanks!

PR #19711, opened 2024-01-11
inih: enable static linking using pkgconf

kmilos:


Biswa96:
Please use a patch file and upstream the changes if possible.

kmilos:
> Please use a patch file and upstream the changes if possible.

IMHO there's no point in upstreaming this particular workaround, as the "ideal" solution needs to address this in the projects's Meson recipe directly ([reported upstream](https://github.com/benhoyt/inih/issues/162)).

Unfortunately, it would seem Meson doesn't support setting `Cflags.private` in their .pc generator yet...

Biswa96:
Oh, I did not notice that the build system is meson.

PR #19713, opened 2024-01-11
ruff: split into bin and python lib packages 

ognevny:
like in Arch. ruff itself doesn't depend on python module


Biswa96:
++pkgrel

Biswa96:
The python ruff package should have different pkgdesc.

ognevny:
> The python ruff package should have different pkgdesc.

oh, yes. I'll fix these things a bit later. thanks

ognevny (commented):
done

Biswa96:
Thank you.

PR #19750, opened 2024-01-14
oh-my-posh: update to 19.5.2

ognevny:


Biswa96:
It seems that oh-my-posh releases new tags more frequently.

ognevny:
> It seems that oh-my-posh releases new tags more frequently.

I found that some deps I maintain make new releases ~1-2 times/week

PR #19760, opened 2024-01-15
quassel: Add aarch64 builds

hmartinez82:
image


mingw-w64-snoregrowl/PKGBUILD
```diff
@@ -25,6 +25,7 @@ build() {
   mkdir -p "${srcdir}"/build-${MSYSTEM} && cd "${srcdir}"/build-${MSYSTEM}
 
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  CXXFLAGS="-D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR" \
```
hmartinez82, line 28: This was required for all CLANG builds, not just ARM64

Biswa96: Could you explain why that option is required?

hmartinez82: std::auto_ptr not available by default in clang anymore

PR #19765, opened 2024-01-15
suitesparse: update to 7.5.1

mmuetzel:


mmuetzel:
I downloaded and installed the MINGW64 build artifact from this PR. With it, the test suite of Octave doesn't show any new errors.
Also, rebuilding Octave seems to work without issues.

PR #19767, opened 2024-01-15
Clang: Backport fixes for Objective-C on mingw

qmfrederik:
This pull request:

Backports the following fixes for clang:
- https://github.com/llvm/llvm-project/pull/77797
- https://github.com/llvm/llvm-project/pull/77385 
- https://github.com/llvm/llvm-project/pull/77255

which support compiling Objective-C code on mingw using Clang.

Since these fixes are specific to mingw, I hope it's appropriate to backport them to this repository before they land in the next clang release.

Additionally, around the time this PR was being created, the llvm repository switched from a state where the first 14 characters of the commit hash were sufficient to uniquely identify a commit (e.g. `f55e5434dce360`)  to a state where 15 characters were required (e.g. `f55e5434dce360b`).

For example, I was able to have [llvm/llvm-project@10b78cc8cea65e7e77d227af4027963f39402724 ](https://github.com/llvm/llvm-project/commit/10b78cc8cea65e7e77d227af4027963f39402724.patch) return the following two fragments (note the difference in the `index` line).

```diff
diff --git a/llvm/test/tools/llvm-rc/preproc.test b/llvm/test/tools/llvm-rc/preproc.test
index f55e5434dce360..7a6d8b3db03662 100644
--- a/llvm/test/tools/llvm-rc/preproc.test
+++ b/llvm/test/tools/llvm-rc/preproc.test
@@ -1,3 +1,3 @@
```

vs.

```diff
diff --git a/llvm/test/tools/llvm-rc/preproc.test b/llvm/test/tools/llvm-rc/preproc.test
index f55e5434dce360b..7a6d8b3db036623 100644
--- a/llvm/test/tools/llvm-rc/preproc.test
+++ b/llvm/test/tools/llvm-rc/preproc.test
@@ -1,3 +1,3 @@
```

This caused the builds to fail intermittently, depending on which (cached) version of the patch was returned by GitHub (presumably because different front-end servers were being hit, with a different cached version).

As a result, I believe using `https://github.com/llvm/llvm-project/commit/{commit-id}.patch` together with sha256 checksums is unstable, as the representation of the patch can change, and I've added the commits themselves to the mingw-w64-clang directory.

### DIFF
--- a/mingw-w64-clang/10b78cc8cea65e7e77d227af4027963f39402724.patch
+++ b/mingw-w64-clang/10b78cc8cea65e7e77d227af4027963f39402724.patch
@@ -1,0 +1,59 @@
+From 10b78cc8cea65e7e77d227af4027963f39402724 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Martin=20Storsj=C3=B6?= <martin@martin.st>
+Date: Fri, 15 Dec 2023 20:15:07 +0200
+Subject: [PATCH] [llvm-windres] Pass user preprocessor arguments before the
+ input filename (#75389)
+
+If passing the windres option --preprocessor, the default arguments "-E
+-xc -DRC_INVOKED" aren't passed. If these are passed explicitly by the
+user via --preprocessor-arg instead, we need to make sure that "-xc" is
+passed before the input filename, as this compiler/preprocessor option
+only has an effect on input files that follow it.
+
+This fixes one of the issues with llvm-windres observed in
+https://github.com/msys2/MINGW-packages/pull/19157.
+---
+ llvm/test/tools/llvm-rc/preproc.test         | 2 +-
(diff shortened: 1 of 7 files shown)


lazka:
thanks

qmfrederik:
Thanks, @lazka .  Are you OK with this PR as is, or would you like me to make any changes?  It's a bit noisy because the .patch files are now part of the repo, but I believe that's the right thing to do (especially for a very active repo such as llvm-project).

lazka:
sounds reasonable, thanks!

MehdiChinoune:
@qmfrederik I couldn't find any link for c65ae76326040369bcd3a85678ab76e402140036.patch, Is it your own patch?

qmfrederik:
@MehdiChinoune It's the equivalent of https://github.com/llvm/llvm-project/pull/77797, which had merge conflicts.

PR #19776, opened 2024-01-16
86box: add gentoo ref

kmilos:


Biswa96:
I am not sure if it is required to add repository references individually with every packages. Those references can be added later while updating the package.

kmilos:
Yeah, just hitting a few random ones... As you say, probably best to update naturally as one goes....

PR #19777, opened 2024-01-16
libBlocksRuntime: Mark public symbols with dllexport

qmfrederik:
Functions which are marked with `BLOCK_EXPORT`  are supposed to be externally visible, but miss the `__attribute__((dllexport))` attribute.  As a result, they are not exported from a .dll when compiled with clang.

Additionally, methods marked with `__attribute__((dllexport))`  cannot have the `__attribute__((weak))` attribute applied to them, so make that conditional on being compiled for non-Windows platforms.

Together with #19767, this helps compiling Objective C code on MinGW using the clang compiler.

### DIFF
--- a/mingw-w64-libblocksruntime/0002-libblocksruntime-0.4.1-export-symbols.patch
+++ b/mingw-w64-libblocksruntime/0002-libblocksruntime-0.4.1-export-symbols.patch
@@ -1,0 +1,87 @@
+diff --git libblocksruntime-0.4.1/Block.h libblocksruntime-0.4.1/Block.h
+index 55cdd01..fb8671b 100644
+--- libblocksruntime-0.4.1/Block.h
++++ libblocksruntime-0.4.1/Block.h
+@@ -33,6 +33,12 @@
+ #   endif
+ #endif
+ 
++#if defined _WIN32 || defined __CYGWIN__
++#   define BLOCK_PUBLIC __attribute__((dllimport))
++#else
++#	define BLOCK_PUBLIC
++#endif
++
+ #if defined(__cplusplus)
+ extern "C" {
+ #endif
+@@ -41,10 +47,10 @@ extern "C" {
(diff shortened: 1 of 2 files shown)


Biswa96:
Thank you for the contribution. Just curious, should there be any condition for static vs. shared libraries?

qmfrederik:
That's a good question.  It doesn't look like the blocksruntime build setup is set up to really differentiate for this at build time -- it uses the same object files for creating the static and shared library:

```
libtool: link: clang -shared  .libs/libBlocksRuntime_la-runtime.o .libs/libBlocksRuntime_la-data.o    -g -O2 -fuse-ld=lld   -o .libs/libBlocksRuntime-0.dll -Wl,--enable-auto-image-base -Xlinker --out-implib -Xlinker .libs/libBlocksRuntime.dll.a
libtool: link: ar cr .libs/libBlocksRuntime.a  libBlocksRuntime_la-runtime.o libBlocksRuntime_la-data.o
```

Additionally, I don't really see a use case where you would want to use this as a static library.  It's part of the (modern) Objective C runtime.  Your application probably uses multiple shared libraries, all of which would reference the blocks runtime as a shared library.

It doesn't look like this package has been maintained very will since it was added 7 years ago (#3123).  It also seems [no other package depends on this](https://github.com/search?q=repo%3Amsys2%2FMINGW-packages%20libblocksruntime%20&type=code).

A option could be to just stop publishing static versions of libblocksruntime?

@Adsun701 any thoughts?

Adsun701:
@qmfrederik I think it would be a good idea to stop using static versions of libblocksruntime.

qmfrederik:
@Biswa96 @Adsun701 I've removed the static build of libblocksruntime.  Let me know if there's anything else you need!

Adsun701:
@qmfrederik @Biswa96 I think that's good!

PR #19780, opened 2024-01-16
[new package] ripgrep-all 0.10.6

ognevny:


ognevny:
I have found that `rust-toolchain.toml` causes some problems at build, including extra deps fetched at `build()`

PR #19781, opened 2024-01-16
uvicorn: update to 0.26.0

podsvirov:


podsvirov:
Hm... Why `Homepage` and `Repository` is the some (see [there](https://packages.msys2.org/base/mingw-w64-uvicorn))?

Cc: @lazka

lazka:
`url` is part of the built package metadata, and only updated once the updated package enters the repo
`msys2_*` is extra metadata which pacman doesn't know about and doesn't end up in the package metadata, which we parse separately in CI and gets updated on git push.

tl;dr: it will fix itself once the update is in the repo

podsvirov:
Thanks for the detailed answer. Let's wait a bit :smile:

PR #19797, opened 2024-01-18
python-lsprotocol: update to 2023.0.1

ognevny:


ognevny:
fixed the things from check. should work now

PR #19800, opened 2024-01-18
kvazaar: update to 2.3.0

kmilos:


Biswa96:
The kvazaar pkgconfig file is incorrect in this new version. Here is the pkgconfig file in ucrt64.

```ini
prefix=@CMAKE_INSTALL_PREFIX@
exec_prefix=${prefix}
libdir=${prefix}/lib
incdir=${prefix}/include

Name: libkvazaar
Description: @CMAKE_PROJECT_DESCRIPTION@
URL: @CMAKE_PROJECT_HOMEPAGE_URL@
Version: @PROJECT_VERSION@
Libs: -L${libdir} -lkvazaar
Libs.private: -lpthread  -lcryptopp
Cflags: -I${incdir}
```

I was planning to port the kvazaar to cmake.

kmilos:
Argh, didn't check... I'll drop this then, please feel free to port to cmake then. NB: Still need to rebuild libheif (it checks for new logging functionality at build time), although soversion wasn't bumped...

Biswa96:
I was not sure if the issue should be reported to upstream. Or, upstream now forces to use cmake instead of autotools.

kmilos:
> Or, upstream now forces to use cmake instead of autotools.

From the release notes:

> Experimental support for CMake, in the future we would like to get rid of automake and visual studio so if there are any issues with the CMakeLists.txt please report them

kmilos:
There also seems to be an issue w/ missing soversion w/ cmake: https://github.com/ultravideo/kvazaar/issues/383

kmilos:
Turns out it was already reported upstream: https://github.com/ultravideo/kvazaar/issues/381

PR #19801, opened 2024-01-18
remove firebird2-git

MehdiChinoune:


sskras:
But why?

Biswa96:
It is now provided by firebird https://packages.msys2.org/base/mingw-w64-firebird

sskras:
OK, didn't have that in my current cache. Just updated, thanks.

lazka:
lgtm

Issue #19744, opened 2024-01-13
Remove mingw-w64-bmake


iahung3:
mingw-w64-bmake is out rightly broken, always segmentation fault and no longer maintained.

iahung3:
> When did it start segfaulting please?

Sometime in the last year. But even before that, it has never worked as expected. For example, if you specify make jobs, it will fail for no reason.

PR #19802, opened 2024-01-18
remove mingw-w64-bmake

iahung3:
https://github.com/msys2/MINGW-packages/issues/19744


lazka:
I agree that in the current crashing state, and given the huge patches, and it being unmaintained, we should remove it. And if needed bring it back as a cygwin package.

@ofry do you have any opinions on this. Are you still using this package?

ofry:
No. You can safely remove it.

PR #19807, opened 2024-01-19
libgda: enable mysql, sqlcipher and web

raedrizqie:
- enable mysql, sqlcipher and web
- fix arch refs

### DIFF
--- a/mingw-w64-libgda/004-fix-libgcrypt-deps.patch
+++ b/mingw-w64-libgda/004-fix-libgcrypt-deps.patch
@@ -1,0 +1,21 @@
+--- a/meson.build
++++ b/meson.build
+@@ -246,7 +246,8 @@
+ 	'-DISO_CODES_PREFIX="'+iso_codes_prefix+'"'
+ 	]
+ 
+-libcryp_config = find_program('libcryp-config', required : false)
++gcrypt_dep = dependency('libgcrypt')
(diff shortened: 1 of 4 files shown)


mingw-w64-libgda/005-use-mysql-from-mariadb.patch
```diff
@@ -0,0 +1,11 @@
...
++++ libgda-6.0.0/meson.build	2021-12-21 13:42:39.324465066 -0600
+@@ -64,7 +64,7 @@
+ libxslt_dep = dependency('libxslt', required: false)
+ jsonglib_dep = dependency('json-glib-1.0',required : false)
+ postgres_dep = dependency ('libpq', required: false)
+-mysql_dep = dependency ('mysqlclient', required: false)
++mysql_dep = dependency ('libmariadb', required: false)
+ mysql_args = []
+ if mysql_dep.found ()
+ if mysql_dep.version ().version_compare ('>21.0.0')
```
MehdiChinoune, line 11: I don't think this patch is required after #19808

raedrizqie: nice! i will update the PR later

MehdiChinoune:
Some svg files removed!

raedrizqie:
hmm.. even just a rebuild in CI produced a package without all those svg files.. maybe something about the updated dependencies?

PR #19810, opened 2024-01-20
emacs: update to 29.2

raedrizqie:
- remove obsolete patch


lazka:
just wondering should we also enable `--with-native-compilation=aot` like Linux distros?

raedrizqie:
from a quick google-fu:
```
If AOT is enabled, some (all?) elisp files in the source distribution are compiled to .eln in advance, as opposed to being compiled as required. AOT is carried out as part of building emacs from source (and it really increases build time).
```
maybe we can try..

PR #19816, opened 2024-01-20
mpdecimal: Update to 4.0.0

lazka:
See https://github.com/msys2/MSYS2-packages/issues/4361


lazka:
A regression in msys2-runtime: https://github.com/msys2/msys2-runtime/issues/190

PR #19823, opened 2024-01-21
bat: redesign recipe

ognevny:
- make it to match template
- add changes for dynamic linking
- set `GIT_DIR=/dev/null` to resolve

<details>

```
Failed to add dependency on the git state: failed to canonicalize ~/MINGW-packages/.git\logs/HEAD: Системе не удается найти указанный путь. (os error 3). Git state changes might not trigger a rebuild.
Failed to add dependency on the git state: failed to canonicalize ~/MINGW-packages/.git\index: Системе не удается найти указанный путь. (os error 3). Git state changes might not trigger a rebuild.
Failed to add dependency on the git state: failed to canonicalize ~/MINGW-packages/.git\logs/HEAD: Системе не удается найти указанный путь. (os error 3). Git state changes might not trigger a rebuild.
Failed to add dependency on the git state: failed to canonicalize ~/MINGW-packages/.git\index: Системе не удается найти указанный путь. (os error 3). Git state changes might not trigger a rebuild.
```

</details>

- removed cargo install as it's always recompiled (I didn't find a solution for it)

### DIFF
--- a/mingw-w64-bat/PKGBUILD
+++ b/mingw-w64-bat/PKGBUILD
@@ -4,7 +4,7 @@ _realname=bat
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=0.24.0
-pkgrel=2
+pkgrel=3
 pkgdesc="Cat clone with syntax highlighting and git integration (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


ognevny:
excitement: enforce dynamic linking for system libs

lazka:
fyi, git issues can usually be worked around by `export GIT_DIR=/dev/null`

ognevny:
> fyi, git issues can usually be worked around by `export GIT_DIR=/dev/null`

I'll try it. thanks

lazka:
(unless the build really needs git, but usually it will bail out if git reports there is no repo)

ognevny (commented):
> (unless the build really needs git, but usually it will bail out if git reports there is no repo) 

I didn't find any mention about it in build scripts. with GIT_DIR=/dev/null it builds without any error

ognevny:
currently I'm slowly working on enforcing dynamic linking for some rust-based packages. let's try this one first

Issue #19718, opened 2024-01-11
[mingw-w64-fox] update to 1.7.84


iahung3:
### Package name

mingw-w64-fox

### New version number

1.7.84

### Other information that may be useful

Source tarball: http://fox-toolkit.org/ftp/fox-1.7.84.tar.gz

### Are you willing to submit a PR?

No

Biswa96:
version 1.7.84 is masked as development version here http://www.fox-toolkit.org/

iahung3:
> version 1.7.84 is marked as development version here http://www.fox-toolkit.org/

Their naming scheme sucks. The so called development version is what really being used by people now. You could think it's like Debian versioning scheme: FOX stable = Debian stable, FOX development = Debian Testing, and FOX snapshot = Debian Unstable.

FOX stable is no longer maintained. People with old code base that still using it having to patch it themselves. The FOX developer no longer care about it.

Long story short, FOX toolkit is semi-rolling release like Debian Testing.

You can see, people all target FOX development in their customized build system of FOX:

https://github.com/franko/fox

https://github.com/devinsmith/fox

Btw, FOX still uses autotools which sucks, this is the reason for these alternative build systems of FOX to be created. I'm currently use the Meson based build system as the cmake based build system is not as actively maintained.

iahung3:
@Biswa96 I created a package request for FOX 1.7: https://github.com/msys2/MINGW-packages/issues/19725. This way, we can keep FOX stable but also have the latest version of FOX. Please let me know what you think about it.

iahung3:
With franko's Meson based build system, building from source is so easy. Closed.

PR #19829, opened 2024-01-21
Add mingw-w64-fox-dev

iahung2:


iahung2:
> There is already a mingw fox package, see https://packages.msys2.org/base/mingw-w64-fox. I am not sure if both versions should exist in different package.

This is FOX 1.7. I don't know how to put it in the same mingw fox package.

lazka:
Are fox 1.6 and 1.7 not compatible? i.e. is there a need to provide both versions?

iahung2:
> Are fox 1.6 and 1.7 not compatible? i.e. is there a need to provide both versions?

Yes. They are not compatible.

Biswa96:
ArchLinux has 1.6.xx version but Gentoo has 1.7.xx. There are two ways I can think of.
1. update the fox package as usual from 1.6 to 1.7.
2. create separate package with 1.7 version, similar as pangomm, cairomm etc.

lazka:
According to http://fox-toolkit.org/download.html 1.7 is a development version and 1.6 is the stable version.

If there is a good reason why we should package a development version we can think about it, otherwise I'd close this.

iahung2:
> According to http://fox-toolkit.org/download.html 1.7 is a development version and 1.6 is the stable version.
> 
> If there is a good reason why we should package a development version we can think about it, otherwise I'd close this.

Explained in detail here: https://github.com/msys2/MINGW-packages/issues/19718#issuecomment-1888729049

iahung2:
Btw, building from source is easy and fast enough with franko's Meson based build system. I will drop this PR. I created it as my first attempt to contribute to the MSYS2 project. I will try something different.

lazka:
I see thanks. Given this situation I can understand the need for a separate package, so I'm not against adding one.

PR #19837, opened 2024-01-22
sccache: update to 0.7.6

ognevny:
upstream finally updated ring dependency! enabling clangarm64


ognevny:
oops, forgot to update checksums.... 🤦

PR #19860, opened 2024-01-25
suitesparse: update to 7.6.0

mmuetzel:
ABI compatibility with previous versions of libcholmod was (accidentally) broken in version 7.4.x of SuiteSparse (see https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/741). Version 7.6.0 restores ABI compatibility with previous versions. I.e., it breaks ABI compatibility with versions 7.5.x.

Rebuild all reverse dependencies (apart from julia) for the changed ABI (again).


MehdiChinoune:
Haven't you misse coin-or-cgl and coin-or-clp?

mmuetzel:
Oops. I looked at the reverse dependencies on MINGW64.
Is there a reason why those aren't packaged for that environment?

MehdiChinoune:
> Oops. I looked at the reverse dependencies on MINGW64. Is there a reason why those aren't packaged for that environment?

Because there is UCRT64, I don't want to maintain an obsolete environment.

MehdiChinoune:
> Is MINGW64 also obsolete now? Was there a discussion about that that I missed?

That's my classification.

lazka:
> Is MINGW64 also obsolete now? Was there a discussion about that that I missed?

These packages never had mingw64 builds to begin with. Imho its fine to decide to not package for it if there is no need in the beginning.

PR #19871, opened 2024-01-26
uutils-coreutils: update to 0.0.24

ognevny:


ognevny:
I don't understand the CLANG64 failure

ognevny:
0.0.23 builds fine by the way. seems like this is a regression from upstream's site

mati865:
That command is quite long, maybe too long for Windows?
You could try to skip zsh suggestions, by removing them from the makefile. I don't know this format but looks like it might be enough to remove them from https://github.com/uutils/coreutils/blob/33785c93a327ab15dc0f5838891ed90b14e9f32f/GNUmakefile#L348, if not you'd have to remove them with sed or awk.

mingw-w64-uutils-coreutils/dont-generate-man-completions.patch
```diff
@@ -0,0 +1,47 @@
...
+-completions: build-coreutils
+-	mkdir -p $(BUILDDIR)/completions/zsh $(BUILDDIR)/completions/bash $(BUILDDIR)/completions/fish
+-	$(foreach prog, $(INSTALLEES), \
+-		$(BUILDDIR)/coreutils completion $(prog) zsh > $(BUILDDIR)/completions/zsh/_$(PROG_PREFIX)$(prog); \
+-		$(BUILDDIR)/coreutils completion $(prog) bash > $(BUILDDIR)/completions/bash/$(PROG_PREFIX)$(prog); \
+-		$(BUILDDIR)/coreutils completion $(prog) fish > $(BUILDDIR)/completions/fish/$(PROG_PREFIX)$(prog).fish; \
+-	)
+-
+-install: build manpages completions
++install: build
```
Biswa96, line 24: Only this change would be enough :)

PR #19903, opened 2024-01-30
binutils: Update to 2.42

lazka:
Remove all backports which are included in this new release.

Remove specify-timestamp.patch, looks like this was implemented upstream in:
https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=6badd1020f5bebd3f
and it seems to work here.

Work around binutils linking against the system libiberty instead of
the included one.

Also binutils now links against libintl, so add gettext as a dependency


lazka:
Some related things:

* We should split gettext up into gettext-runtime, so the toolchain doesn't pull in the gettext tools
* I've added an integration test for the reproducible DLLs, see https://github.com/msys2/msys2-tests/pull/61 which helped to decide which Debian patches are still needed.

PR #19906, opened 2024-01-30
emacs: Enable ahead of time native compilation

svraka:
There was some talk on enabling it in https://github.com/msys2/MINGW-packages/pull/19810 with the concern that compilation might take too long. Turns out, it doesn't, the CI finished in a bit more than 30 minutes, compared to around 15 minutes without AOT. That's quite reasonable (Github [recently bumping their hosted runners to 4 CPU cores](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/) certainly helps), thus I think AOT should be enabled to save a lot of compilation for users.

### DIFF
--- a/mingw-w64-emacs/PKGBUILD
+++ b/mingw-w64-emacs/PKGBUILD
@@ -8,7 +8,7 @@ _realname=emacs
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=29.2
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


Biswa96:
Please bump the pkgrel value.

svraka:
Oops, sorry. Done.

svraka:
Strange, in this run something is stuck for MINGW64. It didn't happen in the CI before I opened this PR, nor during local testing.

(Those git errors are unrelated, building from a tarball doesn't need git but the build process still checks its presence.)

PR #19913, opened 2024-01-31
gettext: split into gettext-runtime and gettext-tools

lazka:
90% is tools which is not needed at runtime, and since it's easy to split this way just split the two sub projects into two packages.

Other packages can switch to add gettext-tools to makedepends and just gettext-runtime to depends.

Also removes expat dependency, turns out it doesn't need it since 0.19.7

### DIFF
--- a/mingw-w64-gettext/PKGBUILD
+++ b/mingw-w64-gettext/PKGBUILD
@@ -3,19 +3,17 @@
 
 _realname=gettext
 pkgbase=mingw-w64-${_realname}
-pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-runtime" "${MINGW_PACKAGE_PREFIX}-${_realname}-tools")
 pkgver=0.22.4
-pkgrel=3
+pkgrel=4
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://www.gnu.org/software/gettext/"
 pkgdesc="GNU internationalization library (mingw-w64)"
 # GPL3 for the package as a whole and LGPL for some parts, see the license files
 license=('spdx:GPL-3.0-or-later AND LGPL-2.1-or-later')
-depends=("${MINGW_PACKAGE_PREFIX}-expat"
-         "${MINGW_PACKAGE_PREFIX}-gcc-libs"
-         "${MINGW_PACKAGE_PREFIX}-libiconv"
-        )
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hello @lazka!

We're getting some failures when installing packages [here](https://github.com/rubygems/rubygems/actions/runs/7755580051/job/21151329683?pr=7422):

```
:: mingw-w64-x86_64-gettext-runtime and mingw-w64-x86_64-gettext are in conflict. Remove mingw-w64-x86_64-gettext? [Y/n] 
error: failed to prepare transaction (could not satisfy dependencies)
:: removing mingw-w64-x86_64-gettext breaks dependency 'mingw-w64-x86_64-gettext' required by mingw-w64-x86_64-gdbm
:: removing mingw-w64-x86_64-gettext breaks dependency 'mingw-w64-x86_64-gettext' required by mingw-w64-x86_64-libidn2
:: removing mingw-w64-x86_64-gettext breaks dependency 'mingw-w64-x86_64-gettext' required by mingw-w64-x86_64-libpsl
:: removing mingw-w64-x86_64-gettext breaks dependency 'mingw-w64-x86_64-gettext' required by mingw-w64-x86_64-make
:: removing mingw-w64-x86_64-gettext breaks dependency 'mingw-w64-x86_64-gettext' required by mingw-w64-x86_64-p11-kit
```

I assume they are caused by changes related to this PR. Maybe there are fixed by https://github.com/msys2/MINGW-packages/commit/d1369dab579dbcd580055eae42702516e1745592?

Just mentioning in case it's some other issue you have not yet noticed.

Thanks for your work!

Biswa96:
> We're getting some failures when installing packages [here](https://github.com/rubygems/rubygems/actions/runs/7755580051/job/21151329683?pr=7422):

Could you check if packages were updated with `pacman -Syu` before installing compiler? If possible try to use setup-msys2 action https://www.msys2.org/docs/ci/

lazka:
Thanks for the report.

It seems you are doing a partial upgrade "-Sy", which is (1) not supported and (2) will break in such renames. So either do `-Suy`, or just `-S`

lazka:
> I assume they are caused by changes related to this PR. Maybe there are fixed by [d1369da](https://github.com/msys2/MINGW-packages/commit/d1369dab579dbcd580055eae42702516e1745592)?

This should also fix it, as a side effect, but only if you don't have gettext installed before installing clang.

deivid-rodriguez:
That's great info, thanks so much!

I'm not running these commands myself, but I will let the github action that's running those, so they can follow up accordingly :+1:.

lazka:
hm, no, I see the actions unpacks some old packages into the msys2 install which depend on gettext, so I don't see a way to make hack around that from our side.

deivid-rodriguez:
No problem, I'll try to get this fixed in https://github.com/ruby/setup-ruby-pkgs!

PR #19917, opened 2024-02-01
glib2: move to new gettext deps

lazka:


MehdiChinoune:
I don't think you could rebuild any package now unless you remove gettext form DBs

lazka:
Yeah, annoying. Done now.

PR #19939, opened 2024-02-03
gcc: refactor installation and add gcc-lto-dump package

MehdiChinoune:
ArchLinux named just lto-dump, should we follow them or name gcc-lto-dump?


lazka:
nice! especially the install cleanup

Issue #11942, opened 2022-06-27
[clang] Undefined references some reason.  Possible code generation issues.


gcasa:
On the branch

```
[gregc@mother-win-vm libobjc2] % git status                                                     [0]
On branch msys2_experimental
Your branch is up to date with 'origin/msys2_experimental'.
```

I am seeing the following undefined references when trying to build with the script build-windows-libobjc2    

git@github.com:gnustep/tools-scripts.git

The repo for libobjc2 is here....

git@github.com:gnustep/libobjc2

The errors are below...

```
[ 81%] Building C object CMakeFiles/objc.dir/blocks_runtime.m.obj
[ 85%] Building C object CMakeFiles/objc.dir/properties.m.obj
[ 88%] Building C object CMakeFiles/objc.dir/gc_none.c.obj
[ 92%] Building CXX object CMakeFiles/objc.dir/arc.mm.obj
[ 96%] Building CXX object CMakeFiles/objc.dir/eh_win32_msys2.cc.obj
[100%] Linking C shared library libobjc.dll
Warning: corrupt .drectve at end of def file
C:/tools/msys64/mingw64/bin/ld: cannot export objc_enumerationMutation: symbol wrong type (2 vs 3)
C:/tools/msys64/mingw64/bin/ld: CMakeFiles/objc.dir/objects.a(objc_msgSend.S.obj):C:/tools/msys64/home/gregc/Development/gnustep/libobjc2/objc_msgSend.x86-64.S:292: undefined reference to `__guard_dispatch_icall_fptr'
C:/tools/msys64/mingw64/bin/ld: CMakeFiles/objc.dir/objects.a(objc_msgSend.S.obj):C:/tools/msys64/home/gregc/Development/gnustep/libobjc2/objc_msgSend.x86-64.S:296: undefined reference to `__guard_dispatch_icall_fptr'
C:/tools/msys64/mingw64/bin/ld: CMakeFiles/objc.dir/objects.a(NSBlocks.m.obj):NSBlocks.m:(.xdata+0x10): undefined reference to `__gnustep_objc_personality_v0'
C:/tools/msys64/mingw64/bin/ld: CMakeFiles/objc.dir/objects.a(Protocol2.m.obj):Protocol2.m:(.data+0xd0): undefined reference to `.objc_selector_conformsTo:_i240:816'
[...]

Issue #8287, opened 2021-04-05
Help wanted with exception handling models in MinGW


triplef:
We’ve been [working](https://github.com/gnustep/libobjc2/pull/190) on adding support for MinGW to the GNUstep Objective-C Runtime (aka [libobjc2](https://github.com/gnustep/libobjc2)). Most things are working now, except for interoperability with C++ exception handling in Objective-C++.

In order to work with different C++ runtimes, libobjc2 probes the C++ exception structure layout at runtime. @davidchisnall, the libobjc2 maintainer, wrote the following about his attempts to get this working with MinGW (see https://github.com/gnustep/libobjc2/pull/190#issuecomment-812854070):

> There's something odd with MinGW's bridging of the SEH and DWARF EH models that is breaking the code that tries to figure out the layout of the C++ exception structure. I've pushed changes that let me pass all of the tests except those related to C++ exception interop.
> 
> It looks as if the __gxx_personality_v0 symbol isn't exported from the C++ runtime with MinGW, so we can't wrap it for Objective-C++ anyway. I don't see any way of getting C++ exception interop working with MinGW. If you know anyone on the MinGW project, perhaps they could help? [...]

> Thinking a bit more about it, the right strategy for MinGW may be to simply punt all exception handling to the C++ runtime, as Apple does.

Is there anyone here who could help shed some light on the above, or is there a better place for me to post this?

Using Objective-C in MinGW currently only works using GCC and its Objective-C runtime, which is lacking support for many more modern language features that libobjc2 provides. This is why MinGW support in libobjc2 has been a much requested feature, and I would appreciate any help to make this happen. Thanks!

mati865:
Looking at `llvm-readobj --coff-exports /mingw32/bin/libstdc++-6.dll` output it seems properly exported:
```
Export {
  Ordinal: 6599
  Name: __gxx_personality_v0
  RVA: 0xEBAE0
}
```
I guess you are not looking for the right symbol, for 32-bit mingw-w64 it is `___gxx_personality_v0` (note there are 3 `_`) and for 64-bit mingw-w64 it is `__gxx_personality_seh0`.

davidchisnall:
[...]

PR #19948, opened 2024-02-04
libobjc2: New package

qmfrederik:
libobjc2 is a modern Objective C runtime.

The v2.2 release of libobjc2 supports building on msys2 (at least mingw64 and ucrt64, via https://github.com/gnustep/libobjc2/pull/254 and https://github.com/gnustep/libobjc2/pull/267); and the clang patches which enable this (https://github.com/llvm/llvm-project/pull/77797, https://github.com/llvm/llvm-project/pull/77385, https://github.com/llvm/llvm-project/pull/77255) have been backported to msys (https://github.com/msys2/MINGW-packages/pull/19767).

This PR adds libobjc2 to msys2.

Closes https://github.com/msys2/MINGW-packages/issues/11942
Closes https://github.com/msys2/MINGW-packages/issues/8287


lazka:
The installed files look wrong:

```
  +/ucrt64/lib/objc.dll
  +/ucrt64/lib/objc.lib
```

This should be:

```
  +/ucrt64/bin/objc.dll
  +/ucrt64/lib/objc.dll.a
```

qmfrederik:
@lazka Thanks, I created a patch to fix the `bin` vs `lib` mismatch and hope to get it upstreamed.  The `.dll.a` vs `.lib` difference seems to stem from using clang instead of gcc (libobjc2 supports clang only).  I couldn't find a CMake variable which would influence this behavior, so I renamed the file in the `package` phase.

qmfrederik:
ok - that patch landed upstream just now: https://github.com/gnustep/libobjc2/pull/275

PR #19953, opened 2024-02-04
entt: documentation

podsvirov:


Biswa96 (approved):

Biswa96:
The change itself looks good to me but I am not sure whether the documentation is required or not.

podsvirov:
As for me, I found usefull to have local documentation when develop with some library. For example, we have [docs](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-flecs-docs?repo=ucrt64) for similar package.

PR #19957, opened 2024-02-05
Update mingw-w64-arm-none-eabi-gcc to V13.2.0

TheBadPhantom:
- Change versions
- Remove !strip option according to changes in th parent repo
- Adjusted checksum


MehdiChinoune:
I tried before to update it, but it fails to build on clang64.
!strio was added in purpose because its a different target, so stripping could corrupt binaries.

TheBadPhantom:
Sadly you are right. It doesn't build.
I've just taken over the !strip option from the arch repo.

PR #19963, opened 2024-02-05
geany{,-plugins}: update to 2.0

ognevny:
geany:
- migrate to meson:
- remove autotools related things
- 0001-mingw-build.patch: add patch
- 0002-export.patch: refresh
- cleanup

geany-plugins:
- try to enable clangarm64
- 001-no-undefined.patch: geanypy no longer exists
- 002-change-for-new-libgit2.patch: fixed upstream
- cleanup

~~currently doesn't build with lots of linker errors, I don't know why~~
currently IDE can't run precompiled binaries (with system can't find path error)


ognevny:
@neheb is 0003-export.patch still needed?

neheb:
No idea

ognevny:
> No idea

but anyway geany builds and works with it (tested on CLANG64)

ognevny:
looks like geany's static libs are broken

neheb:
why not switch to meson?

ognevny:
> why not switch to meson?

I tried to leave many things as possible. will try later

ognevny:
> why not switch to meson?

thanks! locally it finally builds, but geany now can't run compiled binaries somehow...

Biswa96:
I have no idea about this one.

ognevny:
new issue: after I'd enforced GEANY_DATA_DIR as `${prefix}/share/geany` it can't startup because it finds `${prefix}/data/geany.glade` file

eht16:
Meson support in Geany is new and still work in progress (https://github.com/geany/geany/blob/master/README#L84).
Even more important, the Geany Meson build on Windows is nearly untested, I'm not surprised at all about the problems you described above.

I only wonder what was the idea to switch the package build from a working build system setup to a very unstable one.

ognevny:
> Meson support in Geany is new and still work in progress (https://github.com/geany/geany/blob/master/README#L84).
> Even more important, the Geany Meson build on Windows is nearly untested, I'm not surprised at all about the problems you described above.
> 
> I only wonder what was the idea to switch the package build from a working build system setup to a very unstable one.

got it, but that helped me to fix static lib for plugins

ognevny:
the new one https://github.com/msys2/MINGW-packages/pull/20298 has much less (or maybe even no) issues. closing

PR #19989, opened 2024-02-08
libavif: update to 1.0.4

kmilos:


Biswa96:
Is it required to remove werror cmake option? That option may require in future with different version of compiler toolchain or libavif.

kmilos:
> Is it required to remove werror cmake option?

I guess not, it's just default to OFF in this new version...

Biswa96:
ok, it was done in this commit https://github.com/AOMediaCodec/libavif/commit/9f6e55058322cb299f7e2028ac8298bbe781a35e

PR #20003, opened 2024-02-09
add mingw64 to mingw-w64-cppwinrt

ReenigneArcher:
This PR adds mingw64 architecture to mingw-w64-cppwinrt.


MehdiChinoune:
You know UCRT64 is the same as MINGW64 except that it uses ucrt runtime instead of msvcrt.
I think you may consider moving to UCRT64 to fix https://github.com/LizardByte/Sunshine/issues/832

ReenigneArcher:
@MehdiChinoune thanks for the suggestion. I had never heard of ucrt until today... I'll look into it.

MehdiChinoune:
> @MehdiChinoune thanks for the suggestion. I had never heard of ucrt until today... I'll look into it.

see https://www.msys2.org/docs/environments/

MehdiChinoune:
Could you report those issues, It's important for maintainers to investigate such problems.
I don't see any proof that the High-CPU usage is due UCRT, You are comparing two cases with different configurations. It's not like you had MINGW64+cppwinrt then you moved to UCRT64+cppwinrt and the issue start occuring.

Anyway It was just my POV to not add mingw64, other maintainers could consider merging your PR.

FrogTheFrog:
With the UCRT64 I'm encountering this bug (I believe as the behavior is very similar):
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98723

See also:
https://github.com/Exiv2/exiv2/issues/2831
https://github.com/Exiv2/exiv2/issues/2858
https://gitlab.gnome.org/GNOME/gimp/-/merge_requests/1102#note_1916367

In my specific case the regex is always run when a connection is established by the client. The regex takes a minute to compile and by that time the connection just times out. The only options I have was to just not use UCRT64 or migrate to `boost::regex`.

MehdiChinoune:
> With the UCRT64 I'm encountering this bug (I believe as the behavior is very similar): https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98723
> 
> See also: [Exiv2/exiv2#2831](https://github.com/Exiv2/exiv2/issues/2831) [Exiv2/exiv2#2858](https://github.com/Exiv2/exiv2/issues/2858) https://gitlab.gnome.org/GNOME/gimp/-/merge_requests/1102#note_1916367
> 
> In my specific case the regex is always run when a connection is established by the client. The regex takes a minute to compile and by that time the connection just times out. The only options I have was to just not use UCRT64 or migrate to `boost::regex`.

That issue occurs in both MINGW64 and UCRT64, It occured in Exiv2 0.28 because It dropped MSVCRT runtime. So It uses UCRT only. Same for GIMP when they considered to move to UCRT64 they encountered that issue with Exiv2 because It's still at 0.27 on MINGW64.
Moving to MINGW64 won't fix the issue with std::regex. If you look at the gcc issue you mentioned the bug occured on MINGW64.

MehdiChinoune:
I couldn't even reopen the PR, It looks like GH disabled repoening after a certain time.
Consider reopen a new separate PR.

MehdiChinoune:
d

FrogTheFrog:
Well I can't comment on their experience, but strangely the regex issue does not occur with the MINGW64, but occurs with UCRT64.

To be transparent I'm also working on Sunshine, by I have encountered this issues months ago when I was installing MSYS2 and UCRT64 was recommended. I then switched back to MINGW64 and the issue was gone.

Biswa96:
OP could restore the branch, not sure if that possible now.

ReenigneArcher:
I deleted my fork, I will re-create it.

ReenigneArcher:
> I couldn't even reopen the PR, It looks like GH disabled repoening after a certain time. Consider reopen a new separate PR.

Could you try now? I re-created the fork and branch. If it doesn't work I can open a separate PR.

FrogTheFrog:
@MehdiChinoune I have created this repo with the artifacts build by github:
https://github.com/FrogTheFrog/regex-issue

This is the outcome of compiling a regex on my PC:
image

MehdiChinoune:
Still not possible to re-open.

ReenigneArcher:
Re-opened as https://github.com/msys2/MINGW-packages/pull/20477

mati865:
Closed PR cannot be reopened if commits checksum differs from the value when PR was closed. In this case it should be possible to reset the branch to commit `3387a51fb8aadced304b2f35ec0112413e43122b` and then reopen but since there is a new PR, consider this message just as an explanation why it happened.

PR #20012, opened 2024-02-10
pikchr: update

podsvirov:


Biswa96:
Is this built from a source repository? The source URL looks bit different. Would it be possible to use official GitHub mirror mentioned in upstream website?

podsvirov:
> Is this built from a source repository?

Yes.

Authors use self-hosted  [fossil](https://packages.msys2.org/base/mingw-w64-fossil) server.

> The source URL looks bit different.

But home site and source repository is the same resource.

> Would it be possible to use official GitHub mirror mentioned in upstream website?

I prefer to use original source repository.

Issue #20000, opened 2024-02-09
Add `python-httpx` required for `python-jupyterlab` (since 4.1.0)


podsvirov:
### Package name

python-jupyterlab

### Brief description of package

For example see `ArchLinux` [recipe](https://gitlab.archlinux.org/archlinux/packaging/packages/jupyterlab/-/commit/22e9009758f1efe918d13191176fa8b152a3b5ae#9b9baac1eb9b72790eef5540a1685306fc43fd6c_15_15).

### URL for package's homepage

https://github.com/encode/httpx

### MINGW environments where you need the package

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [X] CLANGARM64

### Are you willing to submit a PR?

Maybe, but I'm not sure

podsvirov:
Oops. I add [it](https://github.com/msys2/MINGW-packages/commit/3583071027bfae32b613baa61b7b6aa451809c4f).

PR #20013, opened 2024-02-10
python-httpcore: new package

podsvirov:


Biswa96:
Is this for python-jupyterlab? Would it be possible add the required packages and update jupyterlab in this one pull request?

podsvirov:
> Is this for python-jupyterlab?

Yes.

> Would it be possible add the required packages and update jupyterlab in this one pull request?

Not yet. I'm currently working on #20000 than depends on package from this `PR`.

PR #20019, opened 2024-02-11
Add xmlto

### LINKED ISSUE #20018: Add mingw variant of xmlto
While it's just a bash script, it references various cygwin xml tools. By adjusting some prefixes and paths for tools it should be possible
to replace those build deps with mingw variants for all mingw packages depending on it: https://packages.msys2.org/package/xmlto

For example docbook-xsl could be replaced with mingw-w64-docbook-xsl etc.

lazka:
This is like the msys xmlto but calls mingw tools where possible.

Fixes #20018


lazka:
Need to evaluate the man page issue first.

lazka:
man page is due to docbook-xsl brokenness, so not a blocker (edit: ok, kinda still was since xmlto is used to generate man pages, so that would be a regression if mingw xmlto happens to be in path)

PR #20024, opened 2024-02-12
libobjc2: Add clang64 support

qmfrederik:
Upstream libobjc2 now has clang64 support (via https://github.com/gnustep/libobjc2/pull/276).

Cherry-pick that commit and add clang64 to the build matrix for libobjc2.  This cherry-picked commit also fixes the ".dll.a" suffix for library objects, so we no longer need to manually rename that file.

### DIFF
--- a/mingw-w64-libobjc2/PKGBUILD
+++ b/mingw-w64-libobjc2/PKGBUILD
@@ -28,16 +28,20 @@ source=("https://github.com/gnustep/${_realname}/archive/refs/tags/v${pkgver}.zi
         # Install runtime files in CMAKE_INSTALL_BINDIR when GNUstep is not installed
         "https://github.com/gnustep/libobjc2/commit/639c676bb8033422539e19a2e2888bdbd06507e5.patch"
         # Support building on msys/clang64
-        "https://github.com/gnustep/libobjc2/commit/1ff5e1298dd76aa370d5a12b690857f9a47b4b56.patch")
+        "https://github.com/gnustep/libobjc2/commit/1ff5e1298dd76aa370d5a12b690857f9a47b4b56.patch"
+        # MinGW: Update library prefix/suffix
+        "https://github.com/gnustep/libobjc2/commit/d3e81f7435dd25a3fa8caeb9dad15216ff4d1a38.patch")
 sha256sums=('3f72ea4955e3e4671d64c64c339f4947b6112b1b2dd02377c5c982057bbeda04'
             'aaa23146b3c2ec442b5dbf20606020bddd15230041ad738ef3a06e92d13605c3'
-            'fba7d33aa03ee7383b559c5b845063acd16b45396520250849f22559577adcc9')
+            'fba7d33aa03ee7383b559c5b845063acd16b45396520250849f22559577adcc9'
+            'bc6c6ab3fa79aa2242bf8b3480d03329551c73afe7e155f92fff758143bb0fe2')
 
 prepare() {
   cd ${srcdir}/${_realname}-${pkgver}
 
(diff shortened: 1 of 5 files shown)


MehdiChinoune:
The is a mistake, the import library should be named `libobjc2.dll.a` while the dynamic libraru `objc2.dll`

in https://github.com/gnustep/libobjc2/pull/276, it should be

```CMake
if (MINGW)
	set(CMAKE_IMPORT_LIBRARY_SUFFIX ".dll.a")
	set(CMAKE_IMPORT_LIBRARY_PREFIX "lib")
	set(CMAKE_LINK_LIBRARY_SUFFIX "")
	set(CMAKE_SHARED_LIBRARY_PREFIX "")
	set(CMAKE_OBJCXX_IMPLICIT_INCLUDE_DIRECTORIES ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
endif ()
```

qmfrederik:
Thanks @MehdiChinoune , I submitted a patch for upstream and included it in this PR.

With this, we now have:

/mingw64/lib/libobjc.dll.a
/mingw64/bin/objc.dll

Is this what you were expecting?

mingw-w64-libobjc2/PKGBUILD
```diff
@@ -7,23 +7,41 @@ pkgver=2.2
 pkgrel=1
 pkgdesc="Objective-C runtime library intended for use with Clang. (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw64' 'ucrt64')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
 url="https://github.com/gnustep/libobjc2"
 license=('spdx:MIT')
+# libobjc2 depends on the C++ runtime, which is libc++ on clang64 and libstdc++ on mingw64/ucrt64
+if [[ ${MSYSTEM} == CLANG* ]]; then
```
ognevny, line 14: libc++ is equivalent for gcc-libs in CLANG* environments

qmfrederik: Correct -- so the current setup is:
clang* -> ${MINGW_PACKAGE_PREFIX}-libc++
mingw*/ucrt* -> ${MINGW_PACKAGE_PREFIX}-gcc-libs
which should be correct, right?

ognevny: gcc-libs is enough. it's defined in all envs

qmfrederik: I can't seem to find gcc-libs in clang64: https://packages.msys2.org/search?t=binpkg&q=gcc-libs.
Also, libobjc2 is linking with the C++ runtime, which is libc++ and not libstdc++ on clang64, right?
Just wanting to make sure I get this right.

ognevny: https://packages.msys2.org/package/mingw-w64-clang-x86_64-gcc-libs

qmfrederik: Aha. Got it now. Thanks for bearing with me!

MehdiChinoune:
@qmfrederik could you try cmake from #20028

qmfrederik:
@MehdiChinoune Thank you for all your help.  The libobjc2 patch got merged and I updated PKGBUILD to use the merged version of that patch.  The installed files are now:

```
  -- Installing: C:/_/B/pkg/mingw-w64-clang-x86_64-libobjc2/clang64/lib/libobjc.dll.a
  -- Installing: C:/_/B/pkg/mingw-w64-clang-x86_64-libobjc2/clang64/bin/libobjc-4.6.dll
```

which I believe is what we want.

MehdiChinoune:
The name of dll was changed, so I think all reverse-dependencies should be rebuilt.

PR #20026, opened 2024-02-12
keepawake: update to 0.5.1

ognevny:


ognevny:
add expiremental C API

ognevny:
@Biswa96 actually issue with `dovi.pc` could be fixed with `MSYS2_ARG_CONV_EXCL="--prefix="` :)

PR #20028, opened 2024-02-12
cmake: support OBJC/OBJCXX on MinGW-w64

MehdiChinoune:


MehdiChinoune:
a PR opened upstream https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9244

qmfrederik:
@MehdiChinoune Thanks!

With this patch, I get the following names:

- libobjc.dll.a
- libobjc.dll

I still need to `set(CMAKE_OBJCXX_IMPLICIT_INCLUDE_DIRECTORIES ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})` for compilation on clang64 to work.

MehdiChinoune:
OBJC/OBJCXX support still not completed. One shouldn't add any LDFLAGS for standard libraries. They should be added automatically by CMake. I don't have any knoweldge/experience with OBJC/OBJCXX, I'll let to someone else to do the job.

qmfrederik:
@MehdiChinoune The LDFLAGS do two things:

1. Ensure Objective C files are linked with a C++ runtime (libobjc2 uses a c++ runtime for its exception support)
2. Ensure lld is used as the linker (the only linker which works well with Objective C)

I'm not sure to what extent you could hardcode 2. in cmake, but I assume there should be a way to determine what the default C++ runtime is and link that with any OBJC application?  I don't know the inner details of cmake well enough to submit patches for this, unfortunately.

PR #20034, opened 2024-02-13
taglib: Update to 2.0

lazka:


MehdiChinoune:
Why not dropping 32-bit from taglib?

lazka:
there is also mingw-w64-gst-plugins-good as a reverse dep and I didn't want to deal with that right now.

MehdiChinoune:
OK

MehdiChinoune:
I have pushed a commit to drop 32-bit from taglib and its reverse-dependencies.

PR #20043, opened 2024-02-14
mingw-w64-gnustep-base: add patches to fix config check failure for thread-safe +initialize

jcaselman-keysight:
Adds the following patches from the main branch of [gnustep/libs-base](https://github.com/gnustep/libs-base) to allow the config check [gnustep/libs-base/config/config.initialize.m](https://github.com/gnustep/libs-base/blob/master/config/config.initialize.m) to build and pass during the gnustep-base package build:

* https://github.com/gnustep/libs-base/commit/fe863c981d61e50e3ea9cbe7f2b23135c1f2faf1
* https://github.com/gnustep/libs-base/commit/87fc1f5e2e4094f49b330bece89ba8aa4436c684

The practical effect is the build process will detect that it is building against a runtime with thread-safe +initialize, and builds linking against gnustep-base in the msys2 environment will no longer print this warning message when they become multi-threaded:
```
WARNING your program is becoming multi-threaded, but you are using an ObjectiveC runtime library which does not have a thread-safe implementation of the +initialize method. Please see README.initialize for more information.
```

### DIFF
--- a/mingw-w64-gnustep-base/PKGBUILD
+++ b/mingw-w64-gnustep-base/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gnustep-base
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.29.0
-pkgrel=2
+pkgrel=3
 pkgdesc="GNUstep Base library (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


Biswa96:
Thank you for the contribution. @qmfrederik Would you like to share your opinion about this change?

qmfrederik:
Yup, both patches are part of upstream GNUstep and improve the build on Windows.

You can see the impact:
- Current build: configure reports: [checking for thread-safe +initialize in runtime... no](https://github.com/msys2/MINGW-packages/pull/20024/checks#step:11:1679)
- With this PR: configure reports: [ checking for thread-safe +initialize in runtime... yes
](https://github.com/msys2/MINGW-packages/pull/20043/checks#step:11:517)

Thanks @jcaselman-keysight

Issue #19981, opened 2024-02-07
[texlive-extra-utils] PANIC: unprotected error in call to Lua API (zlib library version does not match - header: 1.3, library: 1.3.1)


ArrSeaJay:
### Description / Steps to reproduce the issue

try to run `latexmk`

### Expected behavior

No error message

### Actual behavior

PANIC: unprotected error in call to Lua API (zlib library version does not match - header: 1.3, library: 1.3.1)

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

lazka:
Currently blocked on #19921 since texlive fails to build.

Biswa96:
Please ignore my previous question. It seems that the build error would happen without the patch file.

reneparis:
Since https://github.com/msys2/MINGW-packages/issues/20510 was closed due to duplication, I'd like to know what the current status of this issue is, or if it is possible for me to help you out here:

> As far as I understand it, fmtutil.exe is the source. Maybe recompiling it with zlib 1.3.1 will help, but I'm not really into this.
If there is anything I can do to narrow the issue further down, let me know.

MehdiChinoune:
texlive is unusable.
The issue is open for too long without any workaround.
There must be a decision whether fix it or drop it.

jhol:
> texlive is unusable. The issue is open for too long without any workaround. There must be a decision whether fix it or drop it.

I'm greatly in need of a fix for this bug. I would be happy to put time into finding a solution, but it's not clear to me where to look.

naveen521kk:
> texlive is unusable. The issue is open for too long without any workaround. There must be a decision whether fix it or drop it.

It's a rebuild provided we fix https://github.com/msys2/msys2-runtime/issues/208 or find a workaround for it.

MehdiChinoune:
> > texlive is unusable. The issue is open for too long without any workaround. There must be a decision whether fix it or drop it.
> 
> It's simply a rebuild provided we fix [msys2/msys2-runtime#208](https://github.com/msys2/msys2-runtime/issues/208) or find a workaround for it.

I am aware of that, but I don't see any activity or attempt to fix the issue.
[...]

PR #19921, opened 2024-02-02
texlive: use debian patch for a more sensible zlib check

lazka:
See https://github.com/msys2/MINGW-packages/issues/18251#issuecomment-1923348362

Issue #18251, opened 2023-08-24
[texlive-luatex] zlib library version does not match - header: 1.2.13, library: 1.3


That-Human-Being:
### Description / Steps to reproduce the issue

Run a command like `fmtutil`, `updmap`, `mtxrun`, etc.
(These are the commands that Pacman runs after installing any texlive packages)

### Expected behavior

The commands run without error?
Quite literally my google searching suggested that `luatex` just needs to be rebuilt with the new zlib version

### Actual behavior

The packages install fine, it's just the post transaction hooks that fail.

```
:: Running post-transaction hooks...
(1/3) Updating TeXLive filename database...
>>> Running mktexlsr
>>> Running mtxrun (this might take a while)
PANIC: unprotected error in call to Lua API (zlib library version does not match - header: 1.2.13, library: 1.3)
/
(2/3) Updating TeXLive format files...
>>> Running fmtutil (this might take a while)
PANIC: unprotected error in call to Lua API (zlib library version does not match - header: 1.2.13, library: 1.3)
error: command (/usr/share/libalpm/scripts/mingw-w64-x86_64-texlive-fmtutil.script /usr/share/libalpm/scripts/mingw-w64-x86_64-texlive-fmtutil.script ) failed to execute correctly
(3/3) Updating TeXLive font maps...
>>> Running updmap (this might take a while)
PANIC: unprotected error in call to Lua API (zlib library version does not match - header: 1.2.13, library: 1.3)
error: command (/usr/share/libalpm/scripts/mingw-w64-x86_64-texlive-updmap.script /usr/share/libalpm/scripts/mingw-w64-x86_64-texlive-updmap.script ) failed to execute correctly
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19044

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [x] UCRT64
- [x] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

lazka:
thanks.

huh, https://github.com/TeX-Live/texlive-source/blob/0d226662f0ab16f2bd11d5ab4e1cdbf2b7407b79/texk/web2c/luatexdir/luazlib/lzlib.c#L551

I wonder what's the point of that, besides everything breaking.

lazka:
fixed via #18253

thanks for the report

[...]

PR #20046, opened 2024-02-14
mingw-w64-texlive-luatex: rebuild for dependencies

bustercopley:
Error:
"PANIC: unprotected error in call to Lua API (zlib library version does not match - header: 1.3, library: 1.3.1)"

(See commit e0c02d4c9b95c912d200cf73f11d4b79bd20801a of Tue Jan 23 08:19:08 2024 +0100, "zlib: Update to 1.3.1".)

I don't know how much of texlive needs to be rebuilt.


bustercopley:
Duplicate (#18251, #19921, #19981).

PR #20049, opened 2024-02-14
gsoap: Enable use of openssl which is possible since version 2.8.132.

mkolleck:


Biswa96:
Thank you for the contribution. Please add openssl in depends list and bump the pkgrel value.

PR #20053, opened 2024-02-15
swi-prolog: update to 4.2.1

mgondan:
Patch 01-plhome has been reported upstream.

Some help is needed: For some reasons, I need to MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" in the cmake command. This causes the path to the main executable, say, /ucrt64/bin/swipl.exe to be left untouched in the build process. Everything installs fine, but unfortunately, "/ucrt64/bin/swipl.exe" ends up in /ucrt64/lib/cmake/swipl/SWIPLTargets.cmake (not translated to a native path). This, in turn, causes problems when compiling swi-prolog extensions ("packs") that use cmake. I do not know how to fix this.


mgondan:
closing this request because of wrong version number

PR #20054, opened 2024-02-15
libdispatch: new package

qmfrederik:
PR for upstream changes is here: https://github.com/apple/swift-corelibs-libdispatch/pull/814


ognevny:
thanks! this package actually conflicts with https://packages.msys2.org/base/mingw-w64-libblocksruntime

qmfrederik:
Maybe we should just remove the old libblocksruntime package (which is based off a tarball in the Debian package repositories, which I _think_ contains an old copy of the BlocksRuntime as part of LLVM's compiler-rt project) and use the libblocksruntime which ships with Swift instead?  That version has the advantage of being actively maintained.

ognevny:
> Maybe we should just remove the old libblocksruntime package (which is based off a tarball in the Debian package repositories, which I _think_ contains an old copy of the BlocksRuntime as part of LLVM's compiler-rt project) and use the libblocksruntime which ships with Swift instead?  That version has the advantage of being actively maintained.

maybe. what do you think, @lazka?

mingw-w64-libdispatch/PKGBUILD
```diff
@@ -0,0 +1,59 @@
...
+
+  declare -a extra_config
+  if check_option "debug" "n"; then
+    extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    LDFLAGS="-fuse-ld=lld" \
```
ognevny, line 40:
```suggestion
    -LDFLAGS="-fuse-ld=lld" \
    +LDFLAGS=" -fuse-ls=lld" \
```

lazka:
> maybe. what do you think, @lazka?

sure, why not

lazka:
thanks!

PR #20055, opened 2024-02-15
swi-prolog: update to V9.2.1

mgondan:
Two things:

FWIW, patch 01-plhome has been reported upstream.

Help needed: For some reasons, I need to MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" in the cmake command of PKGBUILD. I know what it does, but I don't really know why it is needed. Anyway, it causes the path to the main executable, say, /ucrt64/bin/swipl.exe to be left untouched in the build process. Everything installs fine, but unfortunately, "/ucrt64/bin/swipl.exe" ends up in /ucrt64/lib/cmake/swipl/SWIPLTargets.cmake, that is, it is not translated to a native path. This, in turn, causes problems when compiling swi-prolog extensions ("packs") that use cmake. I do not know how to fix this.

To reproduce the problem:

pacman -S mingw-w64-ucrt-x86_64-swi-prolog (the new version)
swipl
pack_install(environ)
(respond with yes)


lazka:
please try again now

mgondan:
Regarding gmp: Does it make sense to have the dependency in depends as well as in optdepends? The thing is, swi-prolog can replace gmp by libbf during the build, and gmp's license is more restrictive.

mgondan:
Actually, libbf is part of swi-prologs source tree, no need for packaging it. IIRC it is just a single file.

lazka:
How do we tell to to link to libbf instead of gmp then?

mgondan:
If gmp is not installed, the bundled LibBF is automatically used. If it is installed, one has to enable it -DUSE_GMP=ON during compilation.

Ok, two choices:
a) take gmp out completely (from optdepends, makedepends, depends). SWI-Prolog will still be able to do big integer calculations because of the bundled LibBF.
b) Use gmp (makedepends, depends, cmake switch). Do you know if I need to change the license?

I tend to the latter, b.

mgondan:
> please try again now

Works, great! Big thank you!

lazka:
> Do you know if I need to change the license?

Not sure what you mean. Upstream says BSD-2, so it seems correct. The reality is as usual more complicated though: https://metadata.ftp-master.debian.org/changelogs//main/s/swi-prolog/swi-prolog_9.0.4+dfsg-3_copyright :)

mgondan:
Nothing more to add. Thanks again for your help, much appreciated.

PR #20056, opened 2024-02-15
[new package] zed 0.136.2

ognevny:
~~currently fails to build. see https://github.com/zed-industries/zed/issues/7847~~

### DIFF
--- a/mingw-w64-zed/001-download-pc-windows-gnu-assets.patch
+++ b/mingw-w64-zed/001-download-pc-windows-gnu-assets.patch
@@ -1,0 +1,39 @@
+diff --git a/crates/languages/src/rust.rs b/crates/languages/src/rust.rs
+index 0b8e449f4..b50e6dd2b 100644
+--- a/crates/languages/src/rust.rs
++++ b/crates/languages/src/rust.rs
+@@ -77,7 +77,7 @@ impl LspAdapter for RustLspAdapter {
+         let os = match consts::OS {
+             "macos" => "apple-darwin",
+             "linux" => "unknown-linux-gnu",
+-            "windows" => "pc-windows-msvc",
++            "windows" => "pc-windows-gnu",
+             other => bail!("Running on unsupported os: {other}"),
+         };
+         let asset_name = format!("rust-analyzer-{}-{os}.gz", consts::ARCH);
+diff --git a/extensions/deno/src/deno.rs b/extensions/deno/src/deno.rs
+index 02231765d..8e1ab4959 100644
+--- a/extensions/deno/src/deno.rs
++++ b/extensions/deno/src/deno.rs
+@@ -47,7 +47,7 @@ impl DenoExtension {
+             os = match platform {
+                 zed::Os::Mac => "apple-darwin",
+                 zed::Os::Linux => "unknown-linux-gnu",
+-                zed::Os::Windows => "pc-windows-msvc",
++                zed::Os::Windows => "pc-windows-gnu",
+             },
+         );
+ 
+diff --git a/extensions/gleam/src/gleam.rs b/extensions/gleam/src/gleam.rs
+index 8deef729d..bfafaa435 100644
+--- a/extensions/gleam/src/gleam.rs
++++ b/extensions/gleam/src/gleam.rs
+@@ -47,7 +47,7 @@ impl GleamExtension {
+             os = match platform {
+                 zed::Os::Mac => "apple-darwin",
+                 zed::Os::Linux => "unknown-linux-musl",
+-                zed::Os::Windows => "pc-windows-msvc",
++                zed::Os::Windows => "pc-windows-gnu",
+             },
+         );
+ 
--- a/mingw-w64-zed/002-link-sys-libs-dynamically.patch
+++ b/mingw-w64-zed/002-link-sys-libs-dynamically.patch
@@ -1,0 +1,65 @@
+diff --git a/Cargo.lock b/Cargo.lock
+index d8681c4be..335a0a399 100644
+--- a/Cargo.lock
++++ b/Cargo.lock
+@@ -3056,8 +3056,7 @@ dependencies = [
+ [[package]]
+ name = "curl-sys"
+ version = "0.4.67+curl-8.3.0"
+-source = "registry+https://github.com/rust-lang/crates.io-index"
+-checksum = "3cc35d066510b197a0f72de863736641539957628c8a42e70e27c66849e77c34"
++source = "git+https://github.com/ognevny/curl-rust.git?branch=sys-0.4.67-dynamic-mingw#077c35c6b5331042e3ba22e879c8b0bebe7bbf93"
+ dependencies = [
+  "cc",
+  "libc",
+@@ -5824,8 +5823,7 @@ dependencies = [
+ [[package]]
+ name = "libz-sys"
+ version = "1.1.12"
+-source = "registry+https://github.com/rust-lang/crates.io-index"
+-checksum = "d97137b25e321a73eef1418d1d5d2eda4d77e12813f8e6dead84bc52c5870a7b"
++source = "git+https://github.com/ognevny/libz-sys.git?branch=sys-1.1.12-dynamic#c7fac58eed1671bf350fc28e5fb193c7f37cd6b8"
+ dependencies = [
+  "cc",
+  "libc",
+@@ -13221,12 +13219,10 @@ dependencies = [
+ 
+ [[package]]
+ name = "zstd-sys"
+-version = "2.0.8+zstd.1.5.5"
+-source = "registry+https://github.com/rust-lang/crates.io-index"
+-checksum = "5556e6ee25d32df2586c098bbfa278803692a20d0ab9565e049480d52707ec8c"
++version = "2.0.10+zstd.1.5.6"
++source = "git+https://github.com/ognevny/zstd-rs.git?branch=sys-2.0.10-dynamic#ae8ea299d53178d5b947f7c04b2d738f06d1330f"
+ dependencies = [
+  "cc",
+- "libc",
+  "pkg-config",
+ ]
+ 
+diff --git a/Cargo.toml b/Cargo.toml
+index 5785e5b7a..5b32a12c3 100644
+--- a/Cargo.toml
++++ b/Cargo.toml
+@@ -286,10 +286,7 @@ hex = "0.4.3"
+ ignore = "0.4.22"
+ indoc = "1"
+ # We explicitly disable http2 support in isahc.
+-isahc = { version = "1.7.2", default-features = false, features = [
+-    "static-curl",
+-    "text-decoding",
+-] }
++isahc = { version = "1.7.2", default-features = false, features = ["text-decoding"] }
+ itertools = "0.11.0"
+ lazy_static = "1.4.0"
+ libc = "0.2"
+@@ -429,6 +426,9 @@ features = [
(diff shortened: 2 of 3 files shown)


Biswa96:
The Windows support issue in upstream is not fixed. We should wait for proper upstream support.

ognevny:
> Not sure if it's actionable at this point: there's no support for Windows (yet) in Zed hence mingw failures are not something we're ready to consider too.

ognevny:
I'll make it draft. maybe they will reconsider it in future...

ognevny:
I found a workaround for a new build fail in AUR. will try later

ognevny:
I don't understand this failure
```
error: failed to get `tree-sitter-yaml` as a dependency of package `languages v0.1.0 (C:\_\B\src\zed-0.126.2\crates\languages)`
  Caused by:
    failed to load source for dependency `tree-sitter-yaml`
  Caused by:
    Unable to update https://github.com/zed-industries/tree-sitter-yaml?rev=f545a41f57502e1b5ddf2a6668896c1b0620f930#f545a41f
  Caused by:
    failed to update submodule `yaml-test-suite`
  Caused by:
    failed to fetch submodule `yaml-test-suite` from https://github.com/yaml/yaml-test-suite
  Caused by:
    failed to authenticate when downloading repository
    * attempted to find username/password via git's `credential.helper` support, but failed
    if the git CLI succeeds then `net.git-fetch-with-cli` may help here
    https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
  Caused by:
    failed to acquire username/password from local configuration
 ```

**Comment by MehdiChinoune** [comment]
Try adding git in makedepends.

**Comment by ognevny** [comment]
successful local build 🎉

**Comment by ognevny** [comment]
> > Try adding git in makedepends.
> 
> CI still fails :(

https://github.com/yaml/yaml-test-suite returns 404...

**Comment by mati865** [comment]
I guess CI is stuck on Git authentication dialog.
Overall `git` should be unnecessary as Cargo does use libgit2 (to be replaced with gitoxide in the future). Maybe there are awkward entries in git files or somebody force rebased the repo and that commit is no longer there.

**Comment by ognevny** [comment]
> I guess CI is stuck on Git authentication dialog. Overall `git` should be unnecessary as Cargo does use libgit2 (to be replaced with gitoxide in the future). Maybe there are awkward entries in git files or somebody force rebased the repo and that commit is no longer there.

I've already found that one dependncy fetchies the git repository that doesn't exist now. when using `git` executable it gets stuck at it. I guess that I have this repo fetched locally in the past

**Comment by ognevny** [comment]
the same issue with `thin` lto for non-clang environments

**Comment by ognevny** [comment]
non-clang envs are failing even with disabled lto. is there a way to enforce clang+lld for them?

**Comment by ognevny** [comment]
```
❯ zed
[2024-03-16T12:48:13+03:00 ERROR fs] watch path does not exist: C:\Users\maksa\AppData\Roaming\Zed\keymap.json
[2024-03-16T12:48:13+03:00 ERROR fs] watch path does not exist: C:\Users\maksa\AppData\Roaming\Zed\settings.json
[2024-03-16T12:48:13+03:00 ERROR util] crates\settings\src\settings_file.rs:76: EOF while parsing a value at line 1 column 0
[2024-03-16T12:48:13+03:00 ERROR fs] watch path does not exist: C:\Users\maksa\AppData\Roaming\Zed\extensions\installed
[2024-03-16T12:48:13+03:00 ERROR util] crates\extension\src/extension_store.rs:266: Системе не удается найти указанный путь. (os error 3)
[2024-03-16T12:48:13+03:00 ERROR util] crates\zed\src/zed.rs:619: EOF while parsing a value at line 1 column 0
[2024-03-16T12:48:13+03:00 ERROR util] crates\zed\src/main.rs:679: Системе не удается найти указанный путь. (os error 3)
[2024-03-16T12:48:15+03:00 ERROR fs] watch path does not exist: C:\Users\maksa\AppData\Roaming\Zed\tasks.json
fish: Job 1, 'zed' terminated by signal SIGSEGV (Address boundary error)
```

**Comment by ognevny** [comment]
a new issue with v0.127.3: https://github.com/zed-industries/zed/issues/9757

**Comment by ognevny** [comment]
patch sent upstream: https://github.com/zed-industries/zed/pull/9815, but after applying it binary segfaults

**Comment by mikayla-maki** [comment]
FYI all non-macOS builds of Zed should be considered WIP alpha builds, rather than anything stable or functional :)

We'll let folks know when we're ready to properly support Windows development.

**Comment by ognevny** [comment]
> FYI all non-macOS builds of Zed should be considered WIP alpha builds, rather than anything stable or even functional :)
> 
> We'll let folks know when we're ready to properly support Windows development.

ok, thanks!

**Comment by Kreijstal** [comment]
> patch sent upstream: [zed-industries/zed#9815](https://github.com/zed-industries/zed/pull/9815), but after applying it binary segfaults

I guess it's not the rust part?

**Comment by ognevny** [comment]
> I guess it's not the rust part?

I don't know

**Comment by ognevny** [comment]
switched to git HEAD so the fixes will be tested earlier

**Comment by Kreijstal** [comment]
Interesting, it did work for me
grafik
I got some error messages tho

```
[2024-03-27T17:24:41+01:00 ERROR util] crates\settings\src\settings_file.rs:76: EOF while parsing a value at line 1 column 0
[2024-03-27T17:24:41+01:00 ERROR util] crates\settings\src\keymap_file.rs:89: invalid binding value for keystroke escape, context Some("ChatPanel > MessageEditor")

Caused by:
    no action type registered for chat_panel::CloseReplyPreview
[2024-03-27T17:24:41+01:00 ERROR util] crates\zed\src\zed.rs:629: EOF while parsing a value at line 1 column 0
[2024-03-27T17:24:41+01:00 ERROR util] crates\zed\src/main.rs:720: Das System kann den angegebenen Pfad nicht finden. (os error 3)
[2024-03-27T17:24:42+01:00 ERROR util] crates\workspace\src/workspace.rs:912: Error in last_window, select_row_bound expected single row result but found none for: SELECT
  display,
  window_state,
  window_x,
  window_y,
  window_width,
  window_height,
  fullscreen
FROM
  workspaces
WHERE
  workspace_location IS NOT NULL
ORDER BY
  timestamp DESC
LIMIT
  1
[2024-03-27T17:24:42+01:00 ERROR util] crates\zed\src/main.rs:1032: Zugriff verweigert (os error 5)
[2024-03-27T17:24:50+01:00 ERROR util] crates\welcome\src/welcome.rs:136: not yet implemented
[2024-03-27T17:25:20+01:00 ERROR util] crates\copilot_ui\src\copilot_completion_provider.rs:207: copilot was not started because of an error: failed to iterate over archive
```
Probably because of not logging in? But cool project

**Comment by ognevny** [comment]
> Interesting, it did work for me
> grafik
> I got some error messages tho
> 
> ```
> [2024-03-27T17:24:41+01:00 ERROR util] crates\settings\src\settings_file.rs:76: EOF while parsing a value at line 1 column 0
> [2024-03-27T17:24:41+01:00 ERROR util] crates\settings\src\keymap_file.rs:89: invalid binding value for keystroke escape, context Some("ChatPanel > MessageEditor")
> 
> Caused by:
>     no action type registered for chat_panel::CloseReplyPreview
> [2024-03-27T17:24:41+01:00 ERROR util] crates\zed\src\zed.rs:629: EOF while parsing a value at line 1 column 0
> [2024-03-27T17:24:41+01:00 ERROR util] crates\zed\src/main.rs:720: Das System kann den angegebenen Pfad nicht finden. (os error 3)
> [2024-03-27T17:24:42+01:00 ERROR util] crates\workspace\src/workspace.rs:912: Error in last_window, select_row_bound expected single row result but found none for: SELECT
>   display,
>   window_state,
>   window_x,
>   window_y,
>   window_width,
>   window_height,
>   fullscreen
> FROM
>   workspaces
> WHERE
>   workspace_location IS NOT NULL
> ORDER BY
>   timestamp DESC
> LIMIT
>   1
> [2024-03-27T17:24:42+01:00 ERROR util] crates\zed\src/main.rs:1032: Zugriff verweigert (os error 5)
> [2024-03-27T17:24:50+01:00 ERROR util] crates\welcome\src/welcome.rs:136: not yet implemented
> [2024-03-27T17:25:20+01:00 ERROR util] crates\copilot_ui\src\copilot_completion_provider.rs:207: copilot was not started because of an error: failed to iterate over archive
> ```
> Probably because of not logging in? But cool project
> 

wow! I also had this error, but the binary segfault, maybe because I have locally set `RUSTFLAGS="-C target-cpu=native"`. I'll test it tomorrow

**Comment by ognevny** [comment]
It segfaults on CLANG64, but on UCRT64 not

**Comment by ognevny** [comment]
currently comment `cargo vendor` due to https://github.com/zed-industries/zed/issues/10460

**Comment by ognevny** [comment]
most bugs are fixed, the things look pretty (tested on UCRT64)
image

**Comment by ognevny** [comment]
artifacts have very huge size

**Comment by ognevny** [comment]
> artifacts have very huge size

looks like it's not zed issue, because the previous update I had ([63a5f46df4a694cf5ccfea43628a2ac8a4f053da](https://github.com/zed-industries/zed/commit/63a5f46df4a694cf5ccfea43628a2ac8a4f053da)) had normal size, but now even its size is huge

**Comment by ognevny** [comment]
looks like that this is a issue from msys2 side. what can cause a such regression?

**Comment by ognevny** [comment]
turns out that rust compiler creates lots of debug information. I'll enable stripping

**Comment by ognevny** [comment]
compilation fail for UCRT64 because of GCC 14 moment. is it ok to add `-Wno-incompatible-pointer-types` to the C{,XX}FLAGS?

**Comment by ognevny** [comment]
almost all issues fixed. I'll wait until stable 0.136.x release and then push it for `ucrt64`

**Comment by ognevny** [comment]
I've switched to zed as a main editor and I didn't face any bug or glitch. I've removed CLANG64 as segfault issue didn't dissappear. later I'll try to add an entry in zed docs with small guide for installing it in msys2

**Comment by Kreijstal** [comment]
for mingw64, I guess we need to compile this with mingw64 from the pkgbuild?

**Comment by ognevny** [comment]
> for mingw64, I guess we need to compile this with mingw64 from the pkgbuild?

don't understand. you mean provide zed for mingw64? the editor (also rustc) has Windows 10 as minimum supported version. for Windows 10 and higher it's recommended to use ucrt64

**Comment by mati865** [comment]
Clang64 crash is hard to debug because it happens in this C code:
image
and obtaining full debug build is impossible because of Win32 symbols limit.

**Comment by ognevny** [comment]
> Clang64 crash is hard to debug...

I had the crash at startup

**Comment by ognevny** [comment]
> I had the crash at startup

```
(lldb) Process 17072 stopped
* thread #497, stop reason = Exception 0xc0000005 encountered at address 0x7ff6f7b30000: User-mode data execution prevention (DEP) violation at location 0x7ff6f7b30000
    frame #0: 0x00007ff6f7b30000 Zed.exe
->  0x7ff6f7b30000: popq   %r10
    0x7ff6f7b30002: js     0x7ff6f7b30004
    0x7ff6f7b30004: addl   %eax, (%rax)
    0x7ff6f7b30006: addb   %al, (%rax)
```

mati865:
> > I had the crash at startup
> 
> backtrace

This is what I had already posted, I couldn't debug the highlighted C code though for some reason (using gdb and lldb since RustRover doesn't support C debugging).

ognevny:
I see. the whole wasm thing is problematic here. it doesn't even compile with ld, lld only works here

PR #20100, opened 2024-02-18
fontconfig: port to meson

lazka:
* python-lxml doesn't seem to be used
* bzip2 also
* iconv support is disabled by upstream by default in meson, we could patch it back in if needed
* gettext support isn't included in the upstream meson config, we could patch it back if needed

Removed patches: all patching tests or autotools files which are no longer used by default in the meson port

Added patches:
0010-fix-relocation.patch: set DLL_EXPORT, so the config relocation logic is active, also set dllexport
0011-conf-copy-instead-of-symlink.patch: copy default config files since we can't package symlinks


lazka:
I'm gonna report some issues upstream first.

lazka:
I've filed:

* https://gitlab.freedesktop.org/fontconfig/fontconfig/-/merge_requests/311
* https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/407
* https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/408

lazka:
tested some gtk things using fontconfig, all looks fine.

PR #20104, opened 2024-02-18
[libobjc2]: Fix exception handling

qmfrederik:
libobjc2 used Vectored Exception Handlers to catch uncaught exceptions on mingw. This implementation is too greedy, and invokes `_objc_unexpected_exception` for (certain) exceptions which would be handled by the application itself.  This surfaced in the gnustep-based unit tests: https://github.com/gnustep/libs-base/pull/374#issuecomment-1939782669

This patch uses `_Unwind_RaiseException` to throw exceptions instead of `__cxa_throw`, allowing us to process unhandled exceptions correctly.

Upstream PR: https://github.com/gnustep/libobjc2/pull/278


Biswa96:
++pkgrel

qmfrederik:
> ++pkgrel

🤦

Biswa96:
This changes the DLL and import library names. Should dependent packages be rebuilt also?

qmfrederik:
Hmm, weird, I thought that change came in via https://github.com/msys2/MINGW-packages/pull/20024, but seems we missed the pkgrel update for libobjc2 in tha tone.  Bumped the pkgrel of the dependencies, hopefully everything is in sync after that.

lazka:
Is this expected?

```
mingw-w64-ucrt-x86_64-libobjc2: /ucrt64/lib/libobjc.dll.a exists in filesystem (owned by mingw-w64-ucrt-x86_64-gcc-objc)
````

PR #20122, opened 2024-02-20
[new package] nodejs-napi-rs{,-lzma,-tar}

raedrizqie:
preparing for `nodejs-swc`

### DIFF
--- a/mingw-w64-nodejs-napi-rs-tar/PKGBUILD
+++ b/mingw-w64-nodejs-napi-rs-tar/PKGBUILD
@@ -1,0 +1,36 @@
+# Contributor: Raed Rizqie <raed.rizqie@gmail.com>
+
+_realname=napi-rs-tar
+pkgbase=mingw-w64-nodejs-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-nodejs-${_realname}"
+pkgver=0.1.0
+pkgrel=1
+pkgdesc='lzma-rs binding to Node.js via napi.rs (mingw-w64)'
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
+url='https://napi.rs/'
+msys2_repository_url='https://github.com/napi-rs/tar'
+license=('spdx:MIT')
+depends=("${MINGW_PACKAGE_PREFIX}-nodejs"
+         "${MINGW_PACKAGE_PREFIX}-nodejs-napi-rs")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-rust")
+source=(https://github.com/napi-rs/tar/archive/refs/tags/v0.1.0.tar.gz)
+sha256sums=('a6573d816aec246d9dbda89ea0747612346ed3f0444f51b8b40da386b4a5b416')
+
+package() {
+  cd ${srcdir}
+  npm install -g @napi-rs/tar@0.1.0 \
+    --prefix "${pkgdir}${MINGW_PREFIX}"
+
+  local _arch=x64
+  if [[ ${CARCH} == aarch64 ]]; then
+    _arch=arm64
+  fi
+
+  cd ${srcdir}/tar-${pkgver}
+  npm install --ignore-scripts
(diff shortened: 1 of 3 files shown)


Kreijstal:
I mean if it is a binary, then wouldn't it make sense for msys2 to package it?

raedrizqie:
the first one:
```
npm install -g @napi-rs/lzma@1.2.1 \
    --prefix "${pkgdir}${MINGW_PREFIX}"
```
is to automate the installation of scripts, cause i dont know how else to do it (not manually)

the second one is to compile the rust based dll and replace the one installed above. if not, the package will use the prebuilt (downloaded) msvc compiled dll (which is not compatible in mingw envs)

Kreijstal:
in that case shouldnt the dev dependencies also be packaged by msys2 as well explicitly?

Kreijstal:
> they usually were used while building the package itself, not a runtime requirement, i think..

yeah that's what makedepends is for

raedrizqie:
i dont think we need to package all dev deps.. `nodejs-swc-core` for example, have 1000+ dev deps

Kreijstal:
> i dont think we need to package all dev deps.. `nodejs-swc-core` for example, have 1000+ dev deps

Yeah that's probably why this won't get packaged in msys unless someone wants to become a mantainer of this

Kreijstal:
Maybe there is another way to compile this without using those 1000 dependencies?

raedrizqie:
yes, just by running:
```
npm install -g @napi-rs/lzma@1.2.1 \
    --prefix "${pkgdir}${MINGW_PREFIX}"
```
but, then the installed `lzma.win32-x64-msvc.node` and `tar.win32-x64-msvc.node` are from prebuilt binaries from npmjs.org
they were built using msvc compiler and have vcruntime14.dll as dependency

Kreijstal:
> yes, just by running:
> 
> ```
> npm install -g @napi-rs/lzma@1.2.1 \
>     --prefix "${pkgdir}${MINGW_PREFIX}"
> ```
> 
> but, then the installed `lzma.win32-x64-msvc.node` and `tar.win32-x64-msvc.node` are from prebuilt binaries from npmjs.org.. they were built using msvc compiler and have `vcruntime140.dll` as dependency
> 
> you can check it here: https://github.com/Brooooooklyn/lzma/releases/download/v1.2.1/lzma.win32-x64-msvc.node

There must be a misunderstanding, downloading builds from the Internet is not compiling

raedrizqie:
either build it from source and provide the package here in the repo, or build it from source and upload it to www.npmjs.org

then what do you suggest?

Kreijstal:
> either build it from source and provide the package here in the repo, or build it from source and upload it to [www.npmjs.org](http://www.npmjs.org)
> 
> then what do you suggest?

you can certainly do second one more easily, but under your own namespace, that means you can't do `npm install package` but you'd have to do something like `npm install @mingw/package` or something like that, or you'd have to convince upstream to build it using msys2/mingw

raedrizqie:
easily? yeah.. you know that we support many environments right? how will you automate the installation between msvct vs ucrt vs clang vs aarch64?

mingw-w64-nodejs-napi-rs-tar/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+  local _arch=x64
+  if [[ ${CARCH} == aarch64 ]]; then
+    _arch=arm64
+  fi
+
+  cd ${srcdir}/tar-${pkgver}
+  npm install --ignore-scripts
+  npm run build --verbose
+
+  install -Dm755 tar.win32-${_arch}-gnu.node ${pkgdir}${MINGW_PREFIX}/lib/node_modules/@napi-rs/tar/node_modules/@napi-rs/tar-win32-${_arch}-msvc/tar.win32-${_arch}-msvc.node
```
Kreijstal, line 35: What confuses me is that you are installing msvc builds, can't we get the gnu builds?

Kreijstal:
hopefully this wolves the issue
https://github.com/napi-rs/napi-rs/pull/2026
and you can work normally

PR #19954, opened 2024-02-05
qt5-static: patches cleanups and updates

raedrizqie:
- patches cleanups and updates
- simplified PKGBUILD recipe
- revert back to use mingw32-make
- build in silent mode
- build out-of-tree
- qtimageformats now linked with static libwebp
- qtnetwork now linked with static openssl
- make sure we link with static zlib and zstd
- fix qmltyperegistrar 'Cannot open foreign types file' error

Fixes #19891 (hopefully)

@MehdiChinoune

PR #20127, opened 2024-02-20
qt-installer-framework: update to 4.7.0

MehdiChinoune:
Build it against shared Qt6 and external libarchive

@lazka Could you test it with msys2-installer

### DIFF
--- a/mingw-w64-qt-installer-framework/0005-fix-shared-build.patch
+++ b/mingw-w64-qt-installer-framework/0005-fix-shared-build.patch
@@ -1,0 +1,194 @@
+--- a/installerfw.pri
++++ b/installerfw.pri
+@@ -106,8 +106,6 @@
+     $$IFW_SOURCE_TREE/src/libs/ifwtools \
+     $$IFW_SOURCE_TREE/src/libs/installer
+ 
+-CONFIG(libarchive): INCLUDEPATH += $$IFW_SOURCE_TREE/src/libs/3rdparty/libarchive
+-
+ CONFIG(lzmasdk) {
+     INCLUDEPATH += $$IFW_SOURCE_TREE/src/libs/3rdparty/7zip
+     win32:INCLUDEPATH += $$IFW_SOURCE_TREE/src/libs/3rdparty/7zip/win/CPP \
+@@ -177,36 +175,7 @@
+ }
+ 
+ CONFIG(libarchive):equals(TEMPLATE, app) {
+-    LIBS += -llibarchive
+-    !isEmpty(IFW_ZLIB_LIBRARY) {
+-        LIBS += $$IFW_ZLIB_LIBRARY
+-    } else:!contains(QT_MODULES, zlib) {
+-        unix:LIBS += -lz
+-        win32:LIBS += -lzlib
+-    }
+-    !isEmpty(IFW_BZIP2_LIBRARY) {
+-        LIBS += $$IFW_BZIP2_LIBRARY
+-    } else {
+-        unix:LIBS += -lbz2
+-        win32:LIBS += -llibbz2
+-    }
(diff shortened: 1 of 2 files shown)


lazka:
Will do

lazka:
I don't quite get how that's supposed to work :) It's all linked dynamically, so we can't use it for the installer.

MehdiChinoune:
> I don't quite get how that's supposed to work :) It's all linked dynamically, so we can't use it for the installer.

Isn't a tool to generate installers like NSIS and WiX? (I am seriously asking)

podsvirov:
> Isn't a tool to generate installers like NSIS and WiX? (I am seriously asking)

Yes.

The `installerbase.exe` executable used as base for created installer. It should be linked statically (have no extra dependencies to run on clean system).

lazka:
> Isn't a tool to generate installers like NSIS and WiX? (I am seriously asking)

It is, but from what I understand it does that by just using the executable from the the package and adding some binary stuff at the end.

MehdiChinoune:
> Yes.
> 
> The `installerbase.exe` executable used as base for created installer. It should be linked statically (have no extra dependencies to run on clean system).

We could use windepoyqt to add required dlls

podsvirov:
> We could use windepoyqt to add required dlls

It's bad idea... We should use static linked qt and link with static runtime.

MehdiChinoune:
> It's bad idea

Why?
I see Many apps deploying Qt DLLs with their apps.

lazka:
Note that we need to get one exe in the end, which is the installer. There can't be any DLLs.

podsvirov:
> I see Many apps deploying Qt DLLs with their apps.

This framework can be used to install any application, no qt-based only.
Application himself can use windeployqt to prepare components, that can be packed to installer for late installation.

MehdiChinoune:
> This framework can be used to install any application, no qt-based only.

I know that, It's already used for msys2-installer which is not a Qt App.

> Application himself can use windeployqt to prepare components, that can be packed to installer for late installation.

What does that to do with the installer?

podsvirov:
Just for note: I'm author of the [CPack IFW Generator](https://cmake.org/cmake/help/latest/cpack_gen/ifw.html) and @lazka wrote right thinks.

podsvirov:
> We could use windepoyqt to add required dlls

> What does that to do with the installer?

MehdiChinoune:
qt-installer-framework is build on Linux against shared Qt. Linux users use it.

Could I ask, Have you tried it or just a speculation?

I have tried building msys2-installer
```
Caught exception: Component "com.msys2.root.base" does not contain a package description (meta/package.xml is missing).
```

MehdiChinoune:
I have successfully generate msys2-installer, but It doesn't launch because of missing dlls and there is no way to fix it.

MehdiChinoune:
I tried to build it against qt6-static but It fails because qt6-static is not fully static.

lazka:
> Could I ask, Have you tried it or just a speculation?

I tried it using the script in https://github.com/msys2/msys2-installer (I had to disable the pacman package downgrade in there though)

MehdiChinoune:
Bad news, It fails to build against qt5-static

raedrizqie:
could you please try building it with qt5-static from #19954 ?

MehdiChinoune:
> could you please try building it with qt5-static from #19954 ?

It's not about static linking.
IFW 4.7.0 was developed against Qt6, while they claim support for Qt5>=5.15.2, It looks like It's broken.

raedrizqie:
i see.. ok then

MehdiChinoune:
I have fixed it to build against qt6-static, just waiting for it from https://github.com/msys2/MINGW-packages/commit/354da6ce to finish building

PR #20130, opened 2024-02-20
qt-installer-framework: update to 4.7.0

MehdiChinoune:


MehdiChinoune:
It builds with LLVM toolchain but fails with GNU one. because LLVM doesn't have an issue with circular static dependency.

MehdiChinoune:
Finally, It builds against Qt6 with some hacks

@lazka Could you try it with msys2-installer just to make sure It doesn't need any DLL, altough `ntldd -R` shows nothing

MehdiChinoune:
I tried it myself, It generates a working installer.

PR #20133, opened 2024-02-21
libobjc2: conflicts with the GCC Objective C runtime

qmfrederik:
Both libobjc2 and gcc-objc provide an Objective C runtime, so these packages conflict with each other.


lazka:
thanks!

Biswa96:
Should this be added only for non-clang msystem?

qmfrederik:
@Biswa96 Feels like a bit of a corner case (and I _think_ you could package GCC's Objective C runtime for the clang environments if you really wanted), but pushed an update so that the conflicts statement applies to non-clang only.

PR #20134, opened 2024-02-21
cmake: no import lib for executable

raedrizqie:


lazka:
is there an upstream issue for this?

raedrizqie:
not that i am aware of

raedrizqie:
> Is there any package affected by this issue? Sometime import library is created from executable to create plugins.

on windows? can you please give an example?

Biswa96:
> on windows? can you please give an example?

I can not recall any actual project but meson supports creating import library from executable https://mesonbuild.com/Reference-manual_functions.html#executable_implib

MehdiChinoune:
postgresql link to posgresql.exe itself, you could check that by `ntldd -R`.

raedrizqie:
both examples are not from cmake build system itself.. but yeah, they exist.. closing

PR #20137, opened 2024-02-21
gitui: update to 0.25.0

ognevny:
- local build showed that libgit2 dynamic lib is used (zlib not)
- remove vendor-openssl feature which broke clangarm64 build


ognevny:
> Why default features are disabled?

it has `vendor-openssl` feature enabled by default. this a bit ugly workaround disables this feature
gitui Cargo.toml: https://github.com/extrawurst/gitui/blob/3f69f5fe62118952cba82a9eff84845d0051723e/Cargo.toml#L70-L78
build script of openssl-sys: https://github.com/sfackler/rust-openssl/blob/ada168bd5baf5f2f3e4f37c08ba76472ffedb83d/openssl-sys/build/main.rs#L47-L57

PR #20144, opened 2024-02-23
gettext: revert upstream adding a printf format attribute

lazka:
It doesn't match what the gettext functions allow, so revert to avoid build warnings for things like %I64u


lazka:
This should fix these failing tests: https://github.com/msys2/msys2-tests/actions/runs/8014128179/job/21892244034

PR #20149, opened 2024-02-23
ffmpeg: enable libjxl

kmilos:


Biswa96:
Should this also be enabled for ffmpeg4.4?

kmilos:
> Should this also be enabled for ffmpeg4.4?

I didn't check if it was actually supported there, sorry... Maybe leave it out if 4.4 is being phased out anyway?

lazka:
We only have 4.4 because some packages aren't compatible with anything newer, so I think it's fine to just leave it as is, until we can delete it.

lazka:
I've created #20151 for tracking the 4.4 removal

PR #20169, opened 2024-02-24
qt6-static: build against internal libraries

MehdiChinoune:


lazka:
I'm wondering, what would be the advantage?

MehdiChinoune:
> I'm wondering, what would be the advantage?

Not needing to fix any linking issue against qt6-static.

Alexpux:
I dont think it is good idea. I prefer to leave as is for now. It allow to us to use latest versions of libraries.

MehdiChinoune:
> I dont think it is good idea. I prefer to leave as is for now. It allow to us to use latest versions of libraries.

First, we don't rebuild against when any library update. We rebuild only with LLVM update because It's dynamically linked to libclang-cpp.
Second Qt always update their internal libraries before any update of Qt. They make sure to use the latest versions.
There is one issue is thet Qt doesn't have internal mng and jasper. Either external or none.

PR #20170, opened 2024-02-24
[new package] stoken 0.92

mpaouris:
Add the mingw-w64-stoken package.

### DIFF
--- a/mingw-w64-stoken/PKGBUILD
+++ b/mingw-w64-stoken/PKGBUILD
@@ -1,0 +1,47 @@
+# Contributor: Marios Paouris <mspaourh@gmail.com>
+
+_realname=stoken
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.92
+pkgrel=1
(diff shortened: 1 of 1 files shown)


Biswa96:
Thank you for the contribution.

Would it be possible to use the https://github.com/stoken-dev/stoken source code repository? According to the project page, the development was moved to that repository. Also, the latest release is 0.93.

mpaouris:
Thank you for your prompt comment!

According to the project's GitHub page, the project's home page is still at http://stoken.sf.net/, where it mentions that "Release tarballs are available on the download page" which is at https://sourceforge.net/projects/stoken/files/

At GitHub there aren't any release tarballs, only tags.

It seems that version v0.93 hasn't been "officially' released.

All major Linux distributions include the v0.92 release:

https://src.fedoraproject.org/repo/pkgs/rpms/stoken/

https://packages.ubuntu.com/search?keywords=stoken

https://packages.debian.org/sid/stoken

https://archlinux.org/packages/?sort=&q=stoken&maintainer=&flagged=

FWIW, openconnect (my next contribution), which depends on stoken builds with 0.92 on all supported platforms, including MinGW cross-compiled from Fedora ( see https://gitlab.com/openconnect/mingw-packages/-/merge_requests/1#note_1687644766 )

PR #20175, opened 2024-02-25
python-milc: update to 1.8.0

fauxpark:


fauxpark:
Needs `types-colorama` now. Will sort that out in a bit.

lazka:
I think you could also patch that out: https://github.com/clueboard/milc/issues/71

PR #20187, opened 2024-02-26
python-pivy: update to 0.6.9.a0

podsvirov:


MehdiChinoune:
python-pivy is at 0.6.9 now.

PR #20189, opened 2024-02-27
new package: mingw-w64-ardour

Kreijstal:
First things first, so this pkgbuild is based from arch and biswa96 patches, I managed to build it. And it runs, but I have trouble with the license path.  but hey it works?
image

Cannot build on clang because there is no jack2 dependency in clang afaik.  There is a bug in waf when it detects CC or CXX as env variables it attempts to get it as a file, it does not check the path for these, it treats it as absolute values. I am not a musician, so no idea if it is functional.

There might be some path conversion bugs.

### DIFF
--- a/mingw-w64-ardour/0001-mingw-link-fixes.patch
+++ b/mingw-w64-ardour/0001-mingw-link-fixes.patch
@@ -1,0 +1,44 @@
+Link fixes for mingw build.
+
+1. fftw3f_threads is available in the MSYS2 fftw package, so allow it
+   on mingw — fftwf_make_planner_thread_safe is otherwise unresolved.
+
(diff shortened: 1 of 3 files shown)


Biswa96:
> There might be some path conversion bugs.

Did you follow the PKGBUILD I provided https://github.com/msys2/MINGW-packages/discussions/20183 ? Use the `/usr/bin/python` before any waf command. The mingw waf package may have some issue. Use the waf which is automatically downloaded. Also, you forgot the whole point about path relocation from that discussion.

Kreijstal:
I've manged to build latest ardour, there is still a problem of $SRC directory appearing on the dll.

PR #20207, opened 2024-02-29
Enable aarch64 builds in packages that depend on scotch

hmartinez82:


mingw-w64-pastix/PKGBUILD
```diff
@@ -20,6 +20,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-openblas"
 makedepends=("${MINGW_PACKAGE_PREFIX}-ninja"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-fc"
```
hmartinez82, line 23: For some reason Flang is being installed transitively in CLANG64, but not in CLANGARM64. And then pastix failed to build without CMake finding a FORTRAN compiling. I'm adding this since pastix has a direct makedep on a FORTRAN compiler. 

I'll investigate which package is not transitively bringing Flang on CLANGARM64, but it is in CLANG64 soon.

PR #20208, opened 2024-02-29
winfile: Fix building with Clang after Clang 17 and enable aarch64 builds

hmartinez82:
Windows 3.11 ARM64

image


Biswa96:
Would it be possible to fix the actual compiler error instead of suppressing it?

hmartinez82:
@Biswa96 I'll give it a try :)

lazka:
(I think that's only worth it if you plan to upstream them)

hmartinez82:
@lazka Correct 😅, but the usage of `register` alone has 100 hits in 19 files 😑

PR #20239, opened 2024-03-02
New package: mediainfo

hmartinez82:
Depends on https://github.com/msys2/MINGW-packages/pull/20245
image


hmartinez82:
@Biswa96 Can't merge this yet because https://github.com/msys2/MINGW-packages/pull/20245 needs to go first. That PR, among other things, fixes the name of the DLL that this package will use.

PR #20245, opened 2024-03-03
libmediainfo: Use Cmake instead

hmartinez82:


hmartinez82:
@Biswa96 Is this failing because the libzen package already exists, but the newer one (that install .cmake files) is not installed yet?

Biswa96:
Yeah, that seems to be the issue.

hmartinez82:
@Biswa96 ok. Can you restart the check?

PR #20246, opened 2024-03-03
New package: frameworkintegration

hmartinez82:
https://archlinux.org/packages/extra/x86_64/frameworkintegration5/

### DIFF
--- a/mingw-w64-frameworkintegration-qt5/PKGBUILD
+++ b/mingw-w64-frameworkintegration-qt5/PKGBUILD
@@ -1,0 +1,57 @@
+
+_variant=-${KF5_VARIANT:-shared}
+source "$(dirname ${BASH_SOURCE[0]})"/../mingw-w64-PKGBUILD-common/kde-frameworks5
+_kde_f5_init_package "${_variant}" "frameworkintegration"
+pkgver=5.115.0
+pkgrel=1
+arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
@Biswa96 any thoughts on this re the move to kf6?

hmartinez82:
I can create the kf6 version too when all the upstream deps are out
Same for kdenlive

Sent with Shortwave <https://www.shortwave.com?utm_medium=email&utm_content=signature&utm_source=aGVybmFuLmMubWFydGluZXpAZ21haWwuY29t>

Biswa96:
kf6 does not have all the packages yet. I am working on it. I think having an older kf5 version of this/kdenlive would help to find any regression in kf6.

Biswa96:
BTW, which program depends on this library?

hmartinez82:
@Biswa96 I guess you are right. I originally went thru this script https://github.com/KDE/kdenlive/blob/master/packaging/kdenlive-msys2.sh looking at what depds Kdenlive required. As you can see, one of the things it does is to manually download frameworkintegration and build it since MSYS2 doesn't have it. 

But then, when I started writing Kdenlive's PKGBUILD looking at Arch's I didn't add this dependency, and now Kdenlive is working without it. Should we close this without merging then?

Biswa96:
If this is not required, I would choose to close this.

hmartinez82:
Closing since there are not downstream deps.

Issue #20253, opened 2024-03-04
octave 8.4.0 fails to build on clang64


MehdiChinoune:
### Description / Steps to reproduce the issue

build octave from its PKGBUILD

### Expected behavior

succeeds to build

### Actual behavior

fails to
```
    CXXLD    libgui/src/libgui-src.la
  clang++: warning: argument unused during compilation: '-pthread' [-Wunused-command-line-argument]
  ld.lld: error: undefined symbol: pthread_mutex_lock
  >>> referenced by libgnu.a(libgnu_la-localename.o):(get_lcid)
  
  ld.lld: error: undefined symbol: pthread_mutex_unlock
  >>> referenced by libgnu.a(libgnu_la-localename.o):(get_lcid)
  >>> referenced by libgnu.a(libgnu_la-localename.o):(get_lcid)
  
  ld.lld: error: undefined symbol: pthread_mutexattr_init
  >>> referenced by libgnu.a(libgnu_la-lock.o):(glthread_recursive_lock_init_multithreaded)
  
  ld.lld: error: undefined symbol: pthread_mutexattr_settype
  >>> referenced by libgnu.a(libgnu_la-lock.o):(glthread_recursive_lock_init_multithreaded)
  
  ld.lld: error: undefined symbol: pthread_mutexattr_destroy
  >>> referenced by libgnu.a(libgnu_la-lock.o):(glthread_recursive_lock_init_multithreaded)
  >>> referenced by libgnu.a(libgnu_la-lock.o):(glthread_recursive_lock_init_multithreaded)
  
  ld.lld: error: undefined symbol: pthread_mutex_init
  >>> referenced by libgnu.a(libgnu_la-lock.o):(glthread_recursive_lock_init_multithreaded)
  
  ld.lld: error: undefined symbol: pthread_once
  >>> referenced by libgnu.a(libgnu_la-lock.o):(glthread_once_multithreaded)
  
  ld.lld: error: undefined symbol: pthread_getspecific
  >>> referenced by libgnu.a(libgnu_la-strsignal.o):(strsignal)
  
  ld.lld: error: undefined symbol: pthread_setspecific
  >>> referenced by libgnu.a(libgnu_la-strsignal.o):(strsignal)
  >>> referenced by libgnu.a(libgnu_la-strsignal.o):(free_key_mem)
  
  ld.lld: error: undefined symbol: pthread_key_create
  >>> referenced by libgnu.a(libgnu_la-strsignal.o):(init)
[...]

PR #20254, opened 2024-03-04
sundials: update to 7.0.0

mmuetzel:
Remove some patches for issues that have been fixed upstream (differently).

Please, don't merge yet. The API changed in a backward-incompatible way and dependent packages need time to adapt for those changes.


mmuetzel:
As discussed in #20253, disable building with MPI for all targets.
The reason is that there is no actual MPI package in MSYS2. `mingw-w64-msmpi` only contains import libraries. A user would need to install MSMPI from Microsoft manually to be able to execute a program that links to SUNDIALS (even if the program that links to SUNDIALS doesn't actually use MPI).

Still left the logic for building with MPI intact. So, reverting this part should be easy.

mmuetzel:
Octave has been updated to version 9 in MSYS2. It should be compatible to SUNDIALS 7.
The other dependent in MSYS2, mfem, is not compatible yet. See: https://github.com/mfem/mfem/issues/4169

mmuetzel:
It looks like the dependency of mfem on SUNDIALS has been removed in the meantime.

Bumping the pkgrel of Octave to rebuild against the new version of SUNDIALS. (That requires cherry-picking an upstream patch to fix a build issue with GCC 14.)

mmuetzel:
I installed and tested the MINGW32 and MINGW64 build artifacts from this PR. There are no new errors when running the test suite of Octave.

PR #20276, opened 2024-03-07
arrow: Update to 15.0.1

raulcd:


MehdiChinoune:
This is just a point release update, so why rebuilding dependent packages. Did arrow break ABI?

PR #18872, opened 2023-10-16
mlt: Update to 7.22.0

hmartinez82:
image

PR #20247, opened 2024-03-03
New package: kdenlive 24.02.0 🎬

hmartinez82:
image

PR #20291, opened 2024-03-09
shotcut: Update to 24.02

hmartinez82:
Requires https://github.com/msys2/MINGW-packages/pull/18872


hmartinez82:
@Biswa96 Can you kick the rebuild of this one? It should work now that the new mlt was merged. The same goes for https://github.com/msys2/MINGW-packages/pull/20247

PR #20295, opened 2024-03-10
gcab, msitools: Update gcab to 1.6 and enable msitools aarch64 builds

hmartinez82:


hmartinez82:
UCRT64 having a fit, again 😑

Biswa96:
I was a network issue from Gnome side. I also get download failure with Gnome mirror links most of the time.

PR #20296, opened 2024-03-10
swi-prolog: new version 9.2.2

mgondan:


mgondan:
help needed: I am wondering if the unit tests from PKGBUILD/check() are invoked. I cannot see them above under "All checks have passed".

Biswa96:
> I am wondering if the unit tests from PKGBUILD/check() are invoked.

Checks/tests are not performed in CI. It is assumed that package maintainer/contributor already ran that. Also, tests may fail most of the time but the actual program/library may work fine.

PR #20307, opened 2024-03-11
objfw: Update to 1.0.12

Midar:


MehdiChinoune:
@Midar I want just to let know that I have made some changes to this package in https://github.com/msys2/MINGW-packages/commit/3b4bfcff which are:
- drop mingw32 (as part of phasing out 32-bit)
- enable ucrt64
- build with gcc on ucrt64

I think building with GCC is supported/allowed, Isn't it?

Midar:
@MehdiChinoune Please revert this!

1. mingw32 with clang and 32 bit is actively supported and one of the two *only* supported configurations (the other being clang64). This is one of the few configurations where unwinding isn't broken.
2. ucrt64 is actively not supported! This environment isn't tested at all and in the past was *known* to be broken.
3. Please do not build with gcc! clang is a very deliberate choice! gcc has incomplete, buggy and almost useless support for ObjC.

All 3 of those changes will result in unhappy users and bogus bug reports to me. This was limited to mingw32 and clang64 because everything else was an unmaintainable nightmare that regularly broke due to not well tested changes to msys2. If you look at the PR importing this package, this has been detailed there and these are all *very* deliberate decisions.

Thank  you for understanding, I hope this will convince you to revert.

MehdiChinoune:
OK, I will make it build with clang on ucrt64.

Midar:
ucrt64 with clang is identical to clang64, so what's the point?
mingw64 would use GCC as well, I suppose? Please don't add any GCC based builds. Those are extremely buggy on Windows. Or did you mean you'll readd mingw32 with Clang? That would be appreciated.

MehdiChinoune:
I won't add mingw32 with clang (We are dropping 32-bit), I am going to add mingw64+clang.
We have environements for everyone to choose, ucrt64 and clang64 are not identical.
please see: https://www.msys2.org/docs/environments/

Midar:
OK, thanks for letting me know. With mingw32 being dropped, I will have to tell my users to use something other than MSYS2 then.

MehdiChinoune:
Who are your users?

MehdiChinoune:
@ I have restored mingw32.

Midar:
Thank you for restoring!

My users use anything from Windows 98 to Windows 11. That's one of the selling features, that you can have a single binary that runs everywhere (portability is one of the main goals of the project, not only on Windows) and detects at runtime which features you can use (e.g. whether to use the A or W Win32 functions). This is all abstracted away so that neither a developer or user needs to care about it.

Midar:
Ah, one more thing: Did you test ucrt64 + mingw64 with Clang actually work? Last time I checked, `make check` failed. Should `make check` be added to the build phase to ensure this?

ognevny:
> Ah, one more thing: Did you test ucrt64 + mingw64 with Clang actually work? Last time I checked, `make check` failed. Should `make check` be added to the build phase to ensure this?

you can use `check()` function for tests. this is skipped in build process (`--nocheck` flag)

Midar:
@MehdiChinoune This entire discussion prompted me to set up CI for MSYS2. It turns out mingw32 in MSYS2 is broken due to a compiler bug: The landing pad generated to catch an exception during init when ARC is enabled just crashes. This does not happen with other mingw-w64 distributions. E.g. installing mingw32 on Fedora and then using `./configure --host=i686-w64-mingw32 OBJC='clang -target i686-w64-mingw32' results in binaries that work, while using MSYS2 does not.

Is there any chance of this ever being fixed in MSYS2, as it's clearly a bug introduced by MSYS2 as upstream mingw-w64 works? If not, I think mingw32 should indeed be replaced by clang32 in this package (clang32 works).

MehdiChinoune:
I don't think anyone is going to waste his time on 32-bit.

PR #20311, opened 2024-03-12
New version: GTK 4.14.1

taozuhong:


mingw-w64-gtk4/PKGBUILD
```diff
@@ -29,10 +29,11 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-json-glib"
          "${MINGW_PACKAGE_PREFIX}-libepoxy"
          "${MINGW_PACKAGE_PREFIX}-pango"
+         "${MINGW_PACKAGE_PREFIX}-shaderc"
```
lazka, line 32: also needs a dep on vulkan-loader

lazka:
The vulkan backend isn't compatible with 32bit pointers (see https://gitlab.gnome.org/GNOME/gtk/-/issues/6033), so it needs to be disabled on 32bit arches.

Let me know if you need any help.

lazka:
And unrelated to the build, it crashes right away here.

mingw-w64-gtk4/PKGBUILD
taozuhong: The CI result shows that vulkan-loader is not required.

lazka: It links against vulkan-1.dll

and the clang CI has a broken dependency: `vulkan-1.dll => not found`

lazka:
With 4.14.1 and newer pango it no longer crashes, but there are glitches and font rendering is broken:

image

lb90:
Bisected to https://gitlab.gnome.org/GNOME/gtk/-/commit/b569470b87d143d30c2388046c9a6450696f19b8

pbs3141:
What happens if you revert https://gitlab.gnome.org/GNOME/gtk/-/commit/b569470b87d143d30c2388046c9a6450696f19b8 as suggested above, and implement the icon fix from https://gitlab.gnome.org/GNOME/gtk/-/issues/6564#note_2059394? Then it might be possible to get this update out the door.

lazka:
yeah, worth a shot

lazka:
I'm getting lots of pango crashes with this, I've filed https://gitlab.gnome.org/GNOME/pango/-/issues/789

Maybe we'll have to revert.

lazka:
Workaround in #20482

PR #20314, opened 2024-03-12
faudio: switch from SDL2 dep to PLATFORM_WIN32

rkitover:
Use the -DPLATFORM_WIN32=TRUE cmake parameter to activate the win32-specific code instead of using SDL2, remove the SDL2 dependency and increase the pkgrel.

### DIFF
--- a/mingw-w64-faudio/PKGBUILD
+++ b/mingw-w64-faudio/PKGBUILD
@@ -4,13 +4,12 @@ _realname=FAudio
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=24.01
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


mingw-w64-faudio/PKGBUILD
```diff
@@ -19,6 +18,7 @@ sha256sums=('28839f2f962180e0f8e0d3cb10061758f3cab7ea7eb92a8ba62ae6d0077e9fb6')
 
 build() {
   declare -a extra_config
+  extra_config+=("-DPLATFORM_WIN32=TRUE")
```
mmuetzel, line 21: `extra_config` is usually used only for conditional flags. Flags that are used unconditionally are usually added verbatim to the `cmake` command.

Biswa96: It can be used for common flags too. e.g. ffmpeg, curl, pango etc. The output is same.

PR #20315, opened 2024-03-12
mesa: disable llvm

MehdiChinoune:
for llvm-18
@pal1000

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=24.0.2
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


mmuetzel:
The main reason for me to use mesa on Windows is llvmpipe. Without it, it's pretty useless to me.

MehdiChinoune:
You are welcome to fix it to build against llvm-18.

mmuetzel:
Did you report it upstream?

MehdiChinoune:
> Did you report it upstream?

yes, https://gitlab.freedesktop.org/mesa/mesa/-/issues/10783

lazka:
Could this be a general meson problem? Meson usually needs adjustmets for every llvm release

MehdiChinoune:
> Could this be a general meson problem? Meson usually needs adjustmets for every llvm release

I think so, I should try cmake before using llvm-config. It looks like It couldn't use the new LLVM cmake files.
I tried it with LLVM-17, using llvm-config fails.

thesamesam:
This ended up being reported as https://github.com/mesonbuild/meson/issues/12961.

PR #20337, opened 2024-03-14
rust: rebuild for LLVM 18

ognevny:
LLVM 18 support will be added only at rust 1.78.0, so applying upstream changes now


MehdiChinoune:
What's the benefit?

ognevny:
> What's the benefit?

it's not about a benefit actually, this is needed for update to 1.77.0, which won't support llvm 18

MehdiChinoune:
So, let's wait for 1.77.0

Issue #20300, opened 2024-03-11
KiCad fails to startup


stahta01:
### Description / Steps to reproduce the issue

Install and try to run Kicad

### Expected behavior

The Kicad program opens a GUI window

### Actual behavior

Under MINGW64 I get this result (Likely the same under UCRT64 and CLANG64):
Edit: CLANG64 had different error; but, the same package installation fixed it.
```
kicad.exe: error while loading shared libraries: wxmsw32u_stc_gcc_custom.dll: cannot open shared object file: No such file or directory
```
Installing "${MINGW_PACKAGE_PREFIX}-libgit2" fixes the problem under MINGW64.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [X] UCRT64
- [X] CLANG64

### Are you willing to submit a PR?

Yes

hmartinez82:
Looking

hmartinez82:
Yes. It's required at runtime, not just at build time
![image](https://github.com/msys2/MINGW-packages/assets/1100440/32b73ef3-13aa-4d42-82db-23b0f9b5adab)

PR #20345, opened 2024-03-15
kicad: Require libgit2 at runtime

hmartinez82:
Fixes https://github.com/msys2/MINGW-packages/issues/20300


MehdiChinoune:
bump pkgrel

PR #20347, opened 2024-03-15
mbedtls: Add pkgconfig file

Biswa96:
For this thread https://github.com/m-ab-s/media-autobuild_suite/issues/2593


Biswa96:
@LigH-de this should fix the mbedtls static linking with librist.

PR #20354, opened 2024-03-15
breeze: Bring parity with KF5

hmartinez82:
Reintroduces breeze-settings like the KF5 package has.


Biswa96:
Could you also repeat the tar extraction command twice? In second time, most of the symlinks are resolved but not all. See breeze-icons for example.

hmartinez82:
Done. unfortunately they don't provide .zip tarballs

PR #20359, opened 2024-03-16
libevent: Enable aarch64 build

hmartinez82:


Biswa96:
Could you please explain why the patch file is required?

hmartinez82:
Clang was failing to build because `evutil_socket` is now a `long long` (declared in util.h) instead of an `int`. Then the function pointer declaration was with `long long` instead, thus this `do_term` method needs sig changing too.

I guess they forgot to built test the sample file? This change to `long long` only happens on Windows.

Biswa96:
Would it be possible to report the issue or upstream the change?

hmartinez82:
I've created a PR for upstream. Also modified my patch because we still build 32 bits libevent and I saw this in corecrt.h:
```
#ifndef _INTPTR_T_DEFINED
#define _INTPTR_T_DEFINED
#ifndef __intptr_t_defined
#define __intptr_t_defined
#undef intptr_t
#ifdef _WIN64
__MINGW_EXTENSION typedef __int64 intptr_t;
#else
typedef int intptr_t;
#endif /* _WIN64 */
#endif /* __intptr_t_defined */
#endif /* _INTPTR_T_DEFINED */
```

PR #20362, opened 2024-03-16
octave: update to 9.1.0

mmuetzel:
Build with some packages that are available on CLANGARM64 now.
Build with Qt5 packages for MINGW32, build with Qt6 packages otherwise.
Suppress more warnings from Clang that seem to break the terminal somehow.
Drop some configure flags that are no longer needed.
Refresh a patch that no longer applied cleanly.


mmuetzel:
A bug was reported that likely affects building Octave packages.
Converting to draft until there is a fix for that.

PR #20370, opened 2024-03-17
cargo-tarball template: fetch target triple in another way at preparation step

ognevny:
arch now uses `"$(rustc -vV | sed -n 's/host: //p')"` to fetch the target triple, which is more minimalist way than we use now


ognevny:
any objections?

PR #20378, opened 2024-03-18
typstudio: update to 0.1.0dev33

podsvirov:


MehdiChinoune:
You know those PKGBUILDs should be buildable for anyone with clean installation.
Your changes means that the user should install node 20.11.1 in /C/hostedtoolcache/windows/node/20.11.1/x64
Next time when GHA node get updated It should be updated to reflect that!

podsvirov:
Unfortunately, now we cannot build a client using the our MSYS2's node.

Can you suggest any solution?

Biswa96:
> Can you suggest any solution?

Leave it as it is for now.

MehdiChinoune:
seriously!
@Biswa96 means to leave typstudio as it is for now, don't update it.

podsvirov:
@Biswa96 what did you mean? Should I revert `typstudio` to `0.1.0dev32`?

Biswa96:
Yes, I meant not to update it if the package can not be built.

podsvirov:
Oops.

But what we will do with N-API? It seems that this is not an easy question.

MehdiChinoune:
What is "N-API"?
Maybe Fix mingw-w64-nodejs!

Kreijstal:
> Oops.
> 
> But what we will do with N-API? It seems that this is not an easy question.
I am trying to tackle it #20407 but lots of things are needed, like rust wasm target (because n-api requires rollup, which requires napi and wasm-pack which requires rust wasm target)

PR #20385, opened 2024-03-19
New version: Pango 1.52.1

taozuhong:


lazka:
Is https://gitlab.gnome.org/GNOME/pango/-/issues/767 fixed with this?

lb90:
Hi! I suggest disabling creation of Cairo DWrite fonts in Pango, that way rendering will be the same as with the current version. That can be done in two ways:

1. (easier) disable lookup of cairo-dwrite.pc in meson.build, e.g by replacing the names in https://gitlab.gnome.org/GNOME/pango/-/blob/1.52.1/meson.build?ref_type=tags#L380 with an empty string. Works thanks to https://mesonbuild.com/Reference-manual_functions.html#dependency:
   ```
   If dependency_name is '', the dependency is always not found. So with required: false, this always returns a dependency object for which the found() method returns false
   ```
2. (a bit awkward) remove calls to [cairo_dwrite_font_face_create_for_dwrite_fontface](https://www.cairographics.org/manual/cairo-DWrite-Fonts.html#cairo-dwrite-font-face-create-for-dwrite-fontface) in https://gitlab.gnome.org/GNOME/pango/-/blob/1.52.1/pango/pangocairo-win32font.c?ref_type=tags

Beside that, we should also include https://gitlab.gnome.org/GNOME/pango/-/merge_requests/720

THis should also fix the crash in GTK 4.14 reported in https://github.com/msys2/MINGW-packages/pull/20311

PR #20389, opened 2024-03-20
transmission: Enable aarch64 builds

hmartinez82:
image


hmartinez82:
Could it be a new dependency that I happen to have in my ARM64 machine? 🤔

Biswa96:
Try to add either mingw itstool or gettext-tools or both.

hmartinez82:
@Biswa96 that worked \o/

PR #20408, opened 2024-03-22
glib2: Update to 2.80.0

lazka:
Remove patch included upstream.
Add new python-packaging dep for gdbus-codegen
Remove gtk-doc dep, no longer used and we don't build docs anyway atm

Disable introspection for bootstrapping


Biswa96:
Would it be possible to add a `_enable_gir=no` flag (or something similar) for the bootstrapping? The similar approach was done with mingw gcc.

lazka:
sure, I'll try, but in a future PR.

Biswa96:
waiting...

mingw-w64-glib2/PKGBUILD
```diff
@@ -17,31 +17,30 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-gettext-runtime"
          "${MINGW_PACKAGE_PREFIX}-pcre2"
          "${MINGW_PACKAGE_PREFIX}-libffi"
+         "${MINGW_PACKAGE_PREFIX}-python-packaging"
```
Biswa96, line 20: Should python-packaging be in optdepends like ArchLinux does?

lazka: I saw that, but it wasn't clear to me why it should be.

PR #20409, opened 2024-03-22
cargo-c: update to 0.9.31

ognevny:


ognevny:
looks like PKG_CONFIG_PATH is broken
```
cargo:warning=failed to probe system libgit2: 
    pkg-config exited with status code 1
    > PKG_CONFIG_PATH=D:\M\msys64\ucrt64\lib\pkgconfig;D:\M\msys64\ucrt64\share\pkgconfig PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags libgit2 libgit2 >= 1.7.2 libgit2 < 1.8.0
  
    The system library `libgit2` required by crate `libgit2-sys` was not found.
    The file `libgit2.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
    PKG_CONFIG_PATH contains the following:
        - D
        - \M\msys64\ucrt64\lib\pkgconfig;D
        - \M\msys64\ucrt64\share\pkgconfig
 ```

**Comment by Biswa96** [comment]
> looks like PKG_CONFIG_PATH is broken

Yeah, the libgit2 pkgconfig file is incorrect.

**Comment by ognevny** [comment]
> Yeah, the libgit2 pkgconfig file is incorrect.

but in my machine it compiles. there even with explicitly set `PKG_CONFIG_PATH` it's parsed as
```
- D
- /M/msys64/clang64/lib/pkgconfig
```
maybe disabling pathconv will help?

PR #20410, opened 2024-03-22
lsd: update to 1.1.0

ognevny:
the patch is no longer needed


ognevny:
the same build issue as #20409

PR #20413, opened 2024-03-23
libgit2: Fix pkgconfig file path relocation

Biswa96:
This fixes the pkgconfig issue as discussed in #20409


ognevny:
> This fixes the pkgconfig issue as discussed in #20409

the issue was the thing that I forgot to put `pkgconf` as makedepend. btw I've found that pkg-config crate doesn't use `Libs.private` field: https://github.com/rust-lang/pkg-config-rs/issues/152

Biswa96:
The pkgconfig in libgit2 package is also incorrect. The libdir and includedir are not properly relocated and Libs.private has CI build directory.

PR #20414, opened 2024-03-23
gitui: update to 0.25.2

ognevny:


mingw-w64-gitui/PKGBUILD
```diff
@@ -24,18 +24,21 @@ sha256sums=('78d31ba66de1521477aef1642c798a385106ff4858f59e79775ba08a694d0ae4')
...
 
-  local _target="${CARCH}-pc-windows-gnu"
-  if [[ $MINGW_PACKAGE_PREFIX == *-clang-aarch64 ]]; then
-    _target="${CARCH}-pc-windows-gnullvm"
-  fi
-  cargo fetch --locked --target "${_target}"
+  cargo fetch --locked --target "$(rustc -vV | sed -n 's/host: //p')"
+}
+
+_env() {
```
Biswa96, line 30: It is not required to export variables in all functions. Just exporting those once in build or prepare function would be sufficient.

PR #20420, opened 2024-03-23
onefetch: redesign the recipe

ognevny:


ognevny:
all the libraries are linked staticly there

PR #20448, opened 2024-03-26
cairomm: update to 1.18.0

lb90:


Biswa96:
We've cairomm-1.16 https://packages.msys2.org/base/mingw-w64-cairomm-1.16

lb90:
Ah, now I see! Right, closing this PR :)

PR #20455, opened 2024-03-26
qt5-static: updates

raedrizqie:
- simplified PKGBUILD recipe
- build in silent mode
- make sure we link with static zlib and zstd
- fix qmltyperegistrar 'Cannot open foreign types file' error
- fix race condition when building qtactiveqt tools
- fix *.pc & *.prl files further by removing full path to libs


MehdiChinoune:
@raedrizqie Please If you want your PR to be reviewed then drop `patches cleanups and updates`.
and spread your changes to separate commits. I don't think anyone could review your changes.

raedrizqie:
ok, all patches were restored..

PR #20457, opened 2024-03-27
dust: update to 1.0.0

ognevny:


mingw-w64-dust/PKGBUILD
```diff
@@ -41,23 +40,27 @@ END
 }
 
 build() {
-  cp -r ${_realname}-${pkgver} build-${MSYSTEM}
-  cd build-${MSYSTEM}
+  cd "${srcdir}/${_realname}-${pkgver}"
+
```
MehdiChinoune, line 44: We do that to make possible for maintainers to build for different environments in the same machine.

ognevny: I don't think it's necessary, but we can set `CARGO_TARGET_DIR="target-${MSYSTEM}"` to change it

MehdiChinoune: Please, Revert the unnecessary change.

PR #20397, opened 2024-03-20
rust: update to 1.77.1

filnet:
https://blog.rust-lang.org/inside-rust/2024/03/17/1.77.0-prerelease.html

PR #20471, opened 2024-03-29
librsvg: Update to 2.58.0

lazka:


lazka:
issue likely fixed by #20397

PR #20473, opened 2024-03-29
libadwaita: Update to 1.5.0

lazka:


lazka:
lots of crashes with the newer gtk4, so waiting a bit

lazka:
Worked around via #20482

PR #20477, opened 2024-03-29
add mingw64 to mingw-w64-cppwinrt

ReenigneArcher:
replacement for #20003


psyke83:
I think I can help shed some light on the situation. In Sunshine and the [regex-issue](https://github.com/FrogTheFrog/regex-issue) testcase we are setting the locale to `.utf-8`, but some debugging reveals that MINGW64 fails to set it correctly, leaving it as the default (`C`).

Results using my [fork](https://github.com/psyke83/regex-issue/tree/print_info) of: https://github.com/FrogTheFrog/regex-issue

MINGW64:
```
setlocale info: C
Before regex
After regex. Took 0ms
```

UCRT64:
```
setlocale info: English_Ireland.utf8
Before regex
After regex. Took 194383ms
```

So, @MehdiChinoune was probably correct in stating that the issue is likely to be present in MINGW64, but we're not seeing the issue simply because `setlocale` was not having an effect. I tried to hardcode the locale as `English_Ireland.utf8`, it still returns as `C`  when built against the MINGW64 runtime, so I'm not sure what's happening.

It seems that `std::regex` works OK with the UCRT64 runtime if the locale is unset/set to the default `C`. Maybe this PR is not necessary, unless there are other major issues aside from locale?

FrogTheFrog:
> I think I can help shed some light on the situation. In Sunshine and the [regex-issue](https://github.com/FrogTheFrog/regex-issue) testcase we are setting the locale to `.utf-8`, but some debugging reveals that MINGW64 fails to set it correctly, leaving it as the default (`C`).
> 
> Results using my [fork](https://github.com/psyke83/regex-issue/tree/print_info) of: https://github.com/FrogTheFrog/regex-issue
> 
> MINGW64:
> 
> ```
> setlocale info: C
> Before regex
> After regex. Took 0ms
> ```
> 
> UCRT64:
> 
> ```
> setlocale info: English_Ireland.utf8
> Before regex
> After regex. Took 194383ms
> ```
> 
> So, @MehdiChinoune was probably correct in stating that the issue is likely to be present in MINGW64, but we're not seeing the issue simply because `setlocale` was not having an effect. I tried to hardcode the locale as `English_Ireland.utf8`, it still returns as `C` when built against the MINGW64 runtime, so I'm not sure what's happening.
> 
> It seems that `std::regex` works OK with the UCRT64 runtime if the locale is unset/set to the default `C`. Maybe this PR is not necessary, unless there are other major issues aside from locale?

Can confirm this. You either have to explicitly set it to `C` or leave it unset (as C++ sets it to `C` anyway). Setting it to default locale `""` (empty string) would still produce the regex bug if the Windows locale is set to UTF-8 system-wide.

PR #20479, opened 2024-03-29
xz: Switch from upstream tarball to git tarball

lazka:
That doesn't contain a configure, so run autogen.
Also needs extra build depends for things already included
in the tarball.

And remove the signing key from the bad actor.


Biswa96:
Would it be possible to keep the old link + the reason as a comment?

lazka:
@Biswa96 something like this?

Biswa96:
Yeah, just as a reminder. Thanks.

PR #20482, opened 2024-03-30
gtk4: default to the old "gl" renderer to work around crashes

lazka:
Upstream report: https://gitlab.gnome.org/GNOME/pango/-/issues/789

This switches the default from ngl to gl. The crashes can still be reproduced via:

GSK_RENDERER=ngl adwaita-1-demo.exe


pbs3141:
The same PR, only upstream and for all platforms - https://gitlab.gnome.org/GNOME/gtk/-/issues/6411

I agree disabling NGL is the right call. Upstream clearly didn't even bother to run the code once on Windows before releasing it.

lazka:
It's not a PR, but thanks for the link.

lazka:
This will be reverted in #20525

PR #20485, opened 2024-03-30
mold: update to 2.30.0

3rav:


ognevny:
does it still segfault?

PR #20487, opened 2024-03-30
[new package] procs 0.14.5

ognevny:


mingw-w64-procs/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=0.14.5
+pkgrel=1
+pkgdesc="A modern replacement for ps written in Rust (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/dalance/procs'
+license=('spdx:MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust")
+options=('!lto')
```
Biswa96, line 14: `lto` is already disabled by-default https://github.com/msys2/MSYS2-packages/blob/e711dd5410eb8b0b8d9ac294ed6f5e9d1a8472c9/pacman/makepkg_mingw.conf#L165 , is it required to mention here?

Biswa96:
man page is missing.

mingw-w64-procs/PKGBUILD
ognevny: in case lto will be enabled by default that won't be applied there. it's already enabled in Cargo.toml manifest

ognevny:
> man page is missing.

should be fixed

mingw-w64-procs/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+  cd "${srcdir}/${_realname}-${pkgver}"
+
+  cargo install \
+    --offline \
+    --no-track \
+    --frozen \
+    --path . \
+    --root "${pkgdir}${MINGW_PREFIX}"
+
+  local _complete="./target/release/procs --gen-completion"
```
Biswa96, line 46:
```suggestion
  local _complete="./target/release/procs --gen-completion-out"
```

The generated shell completions are wrong. You can also simplify those shell completions install command with the above command `$_complete > path_goes_here` (all those pipe and stdin can be replaced).

ognevny: oops, I received a error from fish while `procs --help`

ognevny: fixed, thanks again

PR #20511, opened 2024-03-31
slim-simulator: update to version 4.2

rdinnager:
Just bumping up to the latest version.


mingw-w64-slim-simulator/PKGBUILD
```diff
@@ -1,8 +1,8 @@
 _realname=slim-simulator
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=4.1
-pkgrel=2
+pkgver=4.2
+pkgrel=3
```
Biswa96, line 5: `pkgrel` should be reset to 1 when there is a version bump.

rdinnager: Right you are. My bad. Just pushed fix.

PR #20514, opened 2024-04-01
gtkmm-4.0: update to 4.14.0

mgmax:


mgmax:
I checked the following:
- `makepkg` on MINGW64
- Read changelogs
- Compile Inkscape with the new package (with some bugfixes to the Inkscape build tooling)

lazka:
Thanks!

PR #20523, opened 2024-04-02
rebuild lldb-mi for new version of LLVM

oltolm:
It does not work and needs a rebuild.


Biswa96:
Could you share any error output?

lazka:
It also seems to work here (on a first glance)

oltolm:
Sorry, I built it locally and it was linked against llvm-18.1.dll. There is no problem with the package.

PR #20527, opened 2024-04-03
python-pillow: update to 10.3.0

kmilos:


kmilos:
The existing 10.2.0 is already "breaking" these as well, don't know what we did for it, if anything?

lazka:
yeah, not a regression, those other packages need to be patched...

PR #20530, opened 2024-04-03
tk: Update to 8.6.13

lazka:
010-win-non-x86.patch removed in favor of
https://core.tcl-lang.org/tk/info/18d10428af88089b (there is some difference in TkFinalize() no tbeing called, assume upstream knows best)

011-win-arm64.patch removed in favor of
https://core.tcl-lang.org/tk/info/aab87cd547e2f590

Updating to 8.6.13 (and not .14) to match the version of tcl again. Them matching is required by plplot for example.


lazka:
fingers crossed for arm64

PR #20535, opened 2024-04-03
podman: Update to 5.0.1

lazka:
This is not compatible with podman 4, so all machines need to be reset:

With 4.9 (downgrade if needed):

podman machine stop
podman machine rm

With 5.0:

podman machine init
podman machine start

Also see https://blog.podman.io/2024/03/migration-of-podman-4-to-podman-5-machines/


lazka:
also see https://github.com/containers/podman/issues/22144

lazka:
wrongly installs things into `/etc`, needs to be fixed.

PR #20544, opened 2024-04-04
czmq: add package

gdelazzari:
Add a package for the [CZMQ](https://github.com/zeromq/czmq) higher level library (that depends on `libzmq` which is already provided by the `mingw-w64-zeromq` package). I obtained the `PKGBUILD` is obtained starting from the `mingw-w64-zeromq` one.


peeter123:
Thanks, I was looking for this. Tried to compile from source without success, this builds and works without issues for ucrt at least.

PR #20548, opened 2024-04-04
mingw-w64-directxmesh - update to latest version

JPeterMugaas:


ognevny:
the new executable requires `gcc-libs` to be installed

JPeterMugaas:
I have moved "gcc-libs" to make it a run-time dependency.

PR #20549, opened 2024-04-04
mingw-w64-duvatlas - update to latest version.

JPeterMugaas:


ognevny:
the same as https://github.com/msys2/MINGW-packages/pull/20548#issuecomment-2037798701

ognevny:
this is not optdepend, it's needed to run the binary

JPeterMugaas:
meshconvert is the binary executable.  That's what I thought you meant.

JPeterMugaas:
I have moved "gcc-libs" to make it a run-time dependency.

Biswa96:
> I have moved "gcc-libs" to make it a run-time dependency.

Please move it to depends list.

PR #20553, opened 2024-04-04
drogon: new package

StableAgOH:
add [drogon](https://github.com/drogonframework/drogon) and its dependency [trantor](https://github.com/an-tao/trantor).


StableAgOH:
Looks like I can't add two packages that have dependencies on each other at the same time...

Biswa96:
You did not add trantor in drogon dependency list.

StableAgOH:
> You did not add trantor in drogon dependency list.

My negligence, thx.

PR #20556, opened 2024-04-05
yarn: add clangarm64 support

Suyun114:


Suyun114:
oh maybe i need to bump `pkgrel`?

Biswa96:
> oh maybe i need to bump `pkgrel`?

No, it is not required for simple addition in mingw_arch.

Issue #20493, opened 2024-03-31
Add [LF]


LinArcX:
### Package name

LF

### Brief description of package

Terminal file manager

### URL for package's homepage

https://github.com/gokcehan/lf

### MINGW environments where you need the package

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [X] CLANGARM64

### Are you willing to submit a PR?

maybe. if i know how to make a package for msys2 :)

striezel:
The package [`lf` / mingw-w64-lf](https://packages.msys2.org/base/mingw-w64-lf) is now available for CLANG64, CLANGARM64,  MINGW32, MINGW64, and  UCRT64. Therefore, this issue can be closed.

Note however that the MINGW32 variant of the package may (or may not) get dropped in the nearer future.

PR #20559, opened 2024-04-05
lf: new, 31

Suyun114:
#20493


striezel:
I think `lf` is a good package to have in MSYS2 / MINGW64, so I made a few updates and improvements and rebased it against the current master branch. It should be fine to merge as soon as the CI passes.

striezel:
Still had to [remove clang32 from the arch list](https://github.com/msys2/MINGW-packages/commit/97aa596fbfcea58c1f9197a82a6b5b9b8f67b609), because the [CLANG32 environment was removed](https://www.msys2.org/news/#2024-12-18-removal-of-the-clang32-environment) between the creation of this pull request and its merge.

PR #20585, opened 2024-04-07
mcfgthread: Update to v1.7.1

lhmouse:
1. Migrate from autotools to meson.
2. Fix x87 `long double` test on ARM64EC (thanks to @Blackhex)
3. Make use of `GetSystemTimePreciseAsFileTime()` (win8) and `QueryInterruptTime()` (win10) when they are available.
4. Exclude system suspension from relative timeout intervals, like `CLOCK_MONOTONIC` on Linux.
5. Make headers usable with VC++. For C additional experimental options are required. See [devblog](https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/).

### DIFF
--- a/mingw-w64-mcfgthread/PKGBUILD
+++ b/mingw-w64-mcfgthread/PKGBUILD
@@ -4,17 +4,17 @@ _realname=mcfgthread
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}"-libs)
-pkgver=1.6.1
+pkgver=1.7.1
 pkgrel=1
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
 url="https://github.com/lhmouse/mcfgthread"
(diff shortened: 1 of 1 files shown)


mingw-w64-mcfgthread/PKGBUILD
```diff
@@ -4,69 +4,44 @@ _realname=mcfgthread
...
-
-  make
+  # Build the DLL first, as Meson creates the DLL in the working directory,
+  # which might get picked up by cc1.exe when it is only half-baked.
+  ninja libmcfgthread-1.dll
+  meson compile
 
   rm -rf "${srcdir}${MINGW_PREFIX}"
-  make -j1 DESTDIR=${srcdir} install
+  DESTDIR="${srcdir}" meson install --strip
```
Biswa96, line 39: Is strip option required here? Binary stripping is done automatically before packaging. See options array above.

lhmouse: Yeah you are right, it's not necessary. Updated now.

There is still a difference though: `meson install --strip` passes `--strip-all` which is more aggressive than `makepkg` which passes `--strip-unneeded`. This way, `meson install --strip` produces slightly smaller executables.

See also: https://github.com/msys2/msys2-pacman/pull/20

lhmouse:
Thank you!

PR #20588, opened 2024-04-07
octave: rebuild with Qt5

mmuetzel:
Since Qt6 has been updated to version 6.7.0, Octave segfaults when trying to plot. Rebuild with Qt5 which is still working.

Also rebuild HDF5 because the MINGW32 versions depends on a library that is no longer distributed.
That might have happened while building HDF5 was disabled for a short time last December:
https://github.com/msys2/MINGW-packages/commit/69ab0328c4553515cf5293618b2df273b585414f

I probably rebuilt that library locally and then forgot about it. So, Octave was working for me locally until I installed MSYS2 from scratch. (Thank you @MehdiChinoune for notifying.)

I'll try to ask during next maintainer's meeting of Octave (in the last week of April), how the community would like to proceed with regards to MSYS2. An original plan was to maybe rely on MSYS2 as a distributor for Windows. But I'm doubting if that is a good idea. It looks like MSYS2 is very fast in adapting new versions (sometimes without proper quality control). That might be good for some users. But it is too fast for Octave's small number of maintainers to keep up with.
That probably means that it cannot be a "general distributor" for third party software. (Imho, that's a shame. Windows users would really profit from that.)

Are there other projects that see themselves more like a "distributor" for stable binaries on Windows?


mmuetzel:
HDF5 failed to build in the CLANG64 environment with the following error:
```
      FAILED: cmTC_e2b71.exe 
      C:\Windows\system32\cmd.exe /C "cd . && D:\M\msys64\clang64\bin\flang.exe  -pipe CMakeFiles\cmTC_e2b71.dir\testFortranCompiler.f.obj -o cmTC_e2b71.exe -Wl,--out-implib,libcmTC_e2b71.dll.a -Wl,--major-image-version,0,--minor-image-version,0   && cd ."
      flang: error: unknown argument: '-pipe'
  
```

Something similar also happen for Octave. I thought this was something that autotools did wrong and worked around it by overriding `LDFLAGS=`.
But now that it is failing also with CMake it might be something more general that would better be fixed somewhere else. (Maybe, in `makepkg-mingw`?)

MehdiChinoune:
flang started refusing that linking flag.

mmuetzel:
Afaict, the `-pipe` in `LDFLAGS` is coming from here:
https://github.com/msys2/MSYS2-packages/blob/1f3663d93163421aa6d7f868ed9d5b344a460464/pacman/makepkg_mingw.conf#L71

MehdiChinoune:
Octave was broken on mingw32 for 3 months without anyone noticing/reporting. That prove only one thing "no one uses octave on mingw32".
mingw32 should be dropped from octave because there is no user.

mmuetzel:
> Octave was broken on mingw32 for 3 months without anyone noticing/reporting. That proves only one thing "no one uses octave on mingw32".
> mingw32 should be dropped from octave because there is no user.

Like I already wrote, I'll try to bring that up during next developer meeting of Octave. Please, wait until then.
(Tbh, I'm starting to get tired to get into the same argument again and again...)

lazka:
@MehdiChinoune  mmuetzel is obviously a user, and the maintainer on top. Please stop those useless arguments.

mmuetzel:
Converting to draft for now.
It may need further changes depending on how #20591 will turn out.

mmuetzel:
Shouldn't the CLANG64 runner have downloaded the latest pacman from staging?

lazka:
Tha staging repo gets rebuilt every 15 minutes only

PR #20591, opened 2024-04-07
flang: ignore command line argument `-pipe`

mmuetzel:
Flang 18 started to emit an error when called with the command line argument `-pipe`. Afaict, Flang 17 just ignored it.

Add a patch to ignore that command line argument again.

I'm not sure if we actually need to make that flag visible for `cc1` and `fc1`. But I figured ignoring it for those won't do much harm in case that's not actually needed.

Untested as of now.


lazka:
I can confirm that flang 17 "supported" -pipe and --pipe

lazka:
Arch has -pipe only in CFLAGS: https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/blob/main/makepkg.conf?ref_type=heads maybe something to research

mmuetzel:
> Arch has -pipe only in CFLAGS

It probably doesn't make much difference (unless you are building on a network file system - or have very limited memory).

lazka:
I'd vote for removing it from LDFLAGS.

mmuetzel:
> I'd vote for removing it from LDFLAGS. I'm not even sure it does anything for linking.

That sounds reasonable.

Should I just close this PR? Or would it still make sense to allow Flang to ignore `-pipe` again?

lazka:
> Should I just close this PR? Or would it still make sense to allow Flang to ignore `-pipe` again?

I wouldn't mind it being closed.

mmuetzel:
I opened https://github.com/msys2/MSYS2-packages/pull/4515 for the change of the default `LDFLAGS`.

PR #20592, opened 2024-04-08
imagemagick: update to 7.1.1.30

kmilos:


Biswa96:
wmf module is not built. autotools can not detect wmf library.

kmilos:
What changed since -29?

kmilos:
Looks like same happened on Arch: https://github.com/ImageMagick/ImageMagick/issues/7230

Issue #20490, opened 2024-03-30
notepad++ does not start from msys shell if you have a local language


Kreijstal:
### Description / Steps to reproduce the issue

go to say mingw64 enviroment, install notepad++
then do 
`LANG="en_GB.UTF-8" notepad++`
simulating you have a non en_US locale.

### Expected behavior

opens

### Actual behavior

![it does not open](https://github.com/msys2/MINGW-packages/assets/2415206/cdedf5e6-9c8b-45af-a363-a27b6457b5e7)

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22631

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

lazka:
Looks like this is a known problem: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12417

Biswa96:
Does this happen in mingw64 only?

Kreijstal:
Ucrt as well don't know about clang

heljkon:
@Kreijstal
With https://github.com/msys2/MINGW-packages/pull/20599 a solution was provided. 
Please update your packages and verify.

Biswa96:
Is this issue fixed with newer notepad++ package?

Kreijstal:
bug is solved!

PR #20599, opened 2024-04-09
notepad++: fix exception on posix locale, use localization option

heljkon:
Addresses https://github.com/msys2/MINGW-packages/issues/20490


lazka:
I guess makes sense. Could use a link to the upstream issue and some inline comments on what it's trying to do.

I wonder if that language setting is actually needed, or if it would use the preferred Windows display language anyway if those env vars are unset.

heljkon:
@lazka 
On commenting I read some more documentation.
Because of this I changed the priority for the setting localization option 
See
https://gcc.gnu.org/onlinedocs/libstdc++/manual/localization.html
https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html

PR #20607, opened 2024-04-10
[new package] openconnect 9.12

mpaouris:
Add the mingw-w64-openconnect package.

OpenConnect needs a TUN/TAP driver to operate and supports both the [TAP-Windows](https://community.openvpn.net/openvpn/wiki/GettingTapWindows) driver shipped with OpenVPN and the [Wintun](https://www.wintun.net) layer 3 driver.

The TAP-Windows driver must be pre-installed on the system and a network adapter must have been created prior to running OpenConnect.

The Wintun driver, on the other hand, can be loaded dynamically as a DLL, which installs the driver and creates a network adapter that exists during OpenConnect execution, making it more convenient for users of OpenConnect.
 
The driver needs to be digitally signed in order to be installed on the system, and Wintun is officially distributed as a signed DLL.
OpenConnect downloads the zip distribution, verifies it using it's checksum and extracts the DLL as part of its build process.

Since this DLL *may* be provided by another package (none exists presently), it is packaged independently as mingw-w64-openconnect-wintun and this package is listed an a optional dependency of mingw-w64-openconnect.


lazka:
Puh, we normally don't do pre-built. I personally would prefer a download script which explains it to the user and downloads it it the user agrees, but I've asked what others think in chat.

Regarding the license, it states that it's not allowed to separately redistribute the DLL, so I'm not quite sure if putting it in a separate package is a license violation, but could be that both being in the same repo and having a dependency relation could count as "alongside".

lazka:
The consensus seems to be that we are not going to package prebuilt DLLs.

This means the way forward could be a package which would hint at a command for downloading during install, and in the script educates the user that it's going to download it from a third party website, point to the website, and the license for the binrary https://git.zx2c4.com/wintun/tree/prebuilt-binaries-license.txt, requires the users to agree to download, and checks the checksums.

Feedback welcome.

mpaouris:
Thanks for your input.

Your proposition is to convert the openconnect-wintun package, or the plain openconnect package for simplicity, to inform the user of the required external dependency, and install a downloader script? And the user should run the downloader script to perform the download, checksum checks and installation to the system?

Or is it something else?

Are you aware of a package that does something similar, so it can be studied as a reference?

Kreijstal:
just one small (maybe unrelated) question, I know this vpn software but in case for example I just want to send packages specifically to one vpn without assuming there is an interface installed on the system because windows has no drivers, could it be done a la poor man's socks proxy? Can I create tcp connections through the VPN without it being 'installed?'

mpaouris:
> just one small (maybe unrelated) question, I know this vpn software but in case for example I just want to send packages specifically to one vpn without assuming there is an interface installed on the system because windows has no drivers, could it be done a la poor man's socks proxy? Can I create tcp connections through the VPN without it being 'installed?'

Well, there is an option that can be used in such scenarios, `--script-tun` but it is not supported on windows, see eg. https://gitlab.com/openconnect/openconnect/-/blob/17b45cdd406d3d4218ba08a0835da839a6d79465/main.c#L241 and https://gitlab.com/openconnect/openconnect/-/blob/17b45cdd406d3d4218ba08a0835da839a6d79465/main.c#L1016

Kreijstal:
> > just one small (maybe unrelated) question, I know this vpn software but in case for example I just want to send packages specifically to one vpn without assuming there is an interface installed on the system because windows has no drivers, could it be done a la poor man's socks proxy? Can I create tcp connections through the VPN without it being 'installed?'
> 
> Well, there is an option that can be used in such scenarios, `--script-tun` but it is not supported on windows, see eg. https://gitlab.com/openconnect/openconnect/-/blob/17b45cdd406d3d4218ba08a0835da839a6d79465/main.c#L241 and https://gitlab.com/openconnect/openconnect/-/blob/17b45cdd406d3d4218ba08a0835da839a6d79465/main.c#L1016

is it because it needs the drivers, I suppose? I thought the drivers were only to create a network interface to redirect all traffic..

lazka:
> Are you aware of a package that does something similar, so it can be studied as a reference?

We don't have anything like that. But if you remove that part from this PR I can look into creating such a package separately.

Kreijstal:
Also this package is extremely useful but idk how hard is it to port to windows
https://aur.archlinux.org/packages/openconnect-sso

mpaouris:
@lazka (and everyone!) I did an attempt for an installer. Can you please provide feedback?

mpaouris:
Can you please provide your feedback on the installer? I understand that time is a scarce resource, but I want to check that I am on the right track on this. Thanks

lazka:
Sorry for the delay, I'll try to find some time.

mpaouris:
> Sorry for the delay, I'll try to find some time.

Thank you, I really appreciate it!

lazka:
Are the scripts installed into sbin on purpose? since they are not in PATH that way.

mpaouris:
I've eventually found the time to address the issues about the download location and the license :)
I also copied the license file to the directory containing openconnect's license, to comply with the rest of the system.

Regarding the installation into sbin, this comes from the original package. I guess that the authors chose this since the executables must be ran as root on Linux system and elevated on Windows.

That could be easily fixed by specifying `--sbindir=${MINGW_PREFIX}/bin` during `./configure` .
Do you want me to fix it ?

I've used makepkg's 'lib' as a proven of doing downloads, verifications, argument parsing etc, but this turned out more complex than I originally thought. I'll give it another try to use only the bits of interest.

mpaouris:
Removed the dependency from makepkg's library by keeping only the parts of interest.
Also, the installation is now performed into bin instead of sbin.

@lazka can you kindly review the latest changes?

lazka:
thanks, at least it's self contained now.

Note that your upgrade logic wont work since that waits on user input from what I see, which isn't available in post_upgrade.
You could just print a warning that it needs updating? Or just leave it be for now.

mpaouris:
You are right. I've introduce a new argument to the install, --check-upgrade, to use in post_upgrade.
Please let me know if you want me to squash the commits into a single one.

Thanks!

mpaouris:
@lazka can you kindly review the latest changes?

mpaouris:
Hello everyone! Is there anything that needs to be done for getting this MRerged, besides rebasing it against latest head?

Kreijstal:
> Hello everyone! Is there anything that needs to be done for getting this MRerged, besides rebasing it against latest head?

hmm would you like me to package it in a custom repo

mpaouris:
> hmm would you like me to package it in a custom repo

How does that work?

PR #20609, opened 2024-04-10
[new-package] netcdf-fortran 4.6.1

xueke114:
There is a `netcdf.inc` file in the hdf4 installation directory. I am about to submit a new package, netcdf-fortran, which also has a `netcdf.inc` in the installation directory, and the two cause a conflict. According to Debian's solution (refer to [libhdf-dev](https://packages.debian.org/bookworm/amd64/libhdf4-dev/filelist) and [libnetcdff-dev](https://packages.debian.org/bookworm/amd64/libnetcdff-dev/filelist)), it is proposed to move the installation directory of hdf4 from include to include/hdf, please discuss.

### DIFF
--- a/mingw-w64-netcdf-fortran/PKGBUILD
+++ b/mingw-w64-netcdf-fortran/PKGBUILD
@@ -1,0 +1,68 @@
+# Maintainer: Zheng Xueke <xueke0114@foxmail.com>
+
+_realname=netcdf-fortran
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=4.6.1
+pkgrel=1
+pkgdesc="Unidata NetCDF Fortran Library (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url="https://docs.unidata.ucar.edu/netcdf-fortran/current/"
+msys2_repository_url="https://github.com/Unidata/netcdf-fortran"
+license=('spdx:Apache-2.0')
+depends=("${MINGW_PACKAGE_PREFIX}-netcdf"
+         "${MINGW_PACKAGE_PREFIX}-libxml2"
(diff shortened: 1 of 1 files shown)


ognevny:
++pkgrel

Biswa96:
Please explain why this is needed. Would this cause any issue with other projects which depends on hdf4? Also, please add the package in this same PR so that we can verify the change.

xueke114:
> Please explain why this is needed. Would this cause any issue with other projects which depends on hdf4? Also, please add the package in this same PR so that we can verify the change.

After hdf4 has enabled FORTRAN support, `netcdf.inc` will be generated under include. netcdf-fortran will also generate `netcdf.inc` under include after compilation (Figure 1), so the two packages, netcdf-fortran and hdf4, constitute a conflict on `netcdf.inc`. So the two packages netcdf-fortran and hdf4 constitute a conflict on `netcdf.inc`. If you don't modify the include directory of one of the packages, you can't install both packages at the same time.

Obviously, there are two options about solving this file conflict problem, one is to modify the installation directory of hdf4, and the other is to set the installation directory of netcdf-fortran. Currently there is no netcdf-fortran package in Arch Linux, so there is no direct reference object. So I refer to Debian's solution and modify the installation directory of hdf4.

Regarding the dependencies, at present, the only one that depends on hdf4 is gdal. gdal and hdf4 are both built with cmake, theoretically there will not be a problem of not finding the hdf4 include directory, but for the sake of rigor, I also conducted a test. After installing the modified hdf4 package, I used gdal's PKGBUILD to compile and install gdal, and there were no errors during the compilation process, and after installation, running `gdalinfo --format HDF4` also proved that hdf4 can still be supported (Figure 2). So this PR won't have any impact on gdal.

The PKGBUILD for netcdf-fortran is also submitted in this PR, please check.

图片

图片

MehdiChinoune:
@xueke114 I have rebased your PR and made some little changes.

xueke114:
@MehdiChinoune Better.

PR #20622, opened 2024-04-12
templates: Pass build directory as command line option of build system

Biswa96:
This helps to rule out all the cd commands and jumping between various directories.

### DIFF
--- a/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-git.sh
+++ b/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-git.sh
@@ -36,8 +36,6 @@ prepare() {
 }
 
 build() {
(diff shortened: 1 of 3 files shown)


ognevny:
so it's better to apply this change for all packages with cmake/meson?

Biswa96:
> so it's better to apply this change for all packages with cmake/meson? (after merge)

IMO, it is not required to change all the PKGBUILDs but new packages may follow these templates.

ognevny:
> > so it's better to apply this change for all packages with cmake/meson? (after merge)
> 
> IMO, it is not required to change all the PKGBUILDs but new packages may follow these templates.

what about package updates? there is not merged yet update for shaderc

Biswa96:
> what about package updates? there is not merged yet update for shaderc

It can be kept as is. As I said, only new PKGBUILDs which will be added later.

PR #20625, opened 2024-04-13
hugo: update to 0.124.1

podsvirov:


Biswa96:
Could you add gcc-libs in depends list like other PKGBUILDs? The missing DLLs are reported here https://github.com/msys2/MINGW-packages/actions/runs/8673429269/job/23784780915?pr=20625#step:11:573

podsvirov:
@Biswa96, thanks for suggestion.

PR #20637, opened 2024-04-15
swig: Add boost as a check dependency

pfusik:


lazka:
Could you expand on why and what this fixes?

pfusik:
I've tried building this package locally and its tests (`make check`) fail if `boost` is not installed.

PR #20641, opened 2024-04-15
gitui: update to 0.26.1

ognevny:


ognevny:
git is required here for build: https://github.com/extrawurst/gitui/blob/master/build.rs

PR #20642, opened 2024-04-15
mosquitto: Add openssl library in pkgconfig files for static linking

### LINKED ISSUE #20604: [mosquitto] Static linking broken

I try to use the libmosquitto package in my project with static linking. I use `-static` linker flag and I get the linker flags for the package like this:  `pkg-config --libs --static libmosquitto`. The output of the pkg-config is `-lmosquitto`.

### Expected behavior

The flags provided by pkg-config should work as is.

### Actual behavior

The linker fails like this: `cannot find -lmosquitto: No such file or directory`
It happens because the static library is called `libmosquitto.dll.a`, see the file list [here](https://packages.msys2.org/package/mingw-w64-x86_64-mosquitto).
If I link with the `-lmosquitto.dll` flag, the link succeeds.
I think the `libmosquitto.dll.a` should be renamed to `libmosquitto.a`, but I don't know where to look at.
I think `libmosquittopp` is affected too, but I haven't tried it.

### Verification

### Windows Version

MINGW64_NT-10.0-19045

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

_No response_

Biswa96:
Fixes #20604

@silabs-beta

### DIFF
--- a/mingw-w64-mosquitto/0002-mosquitto-2.0.18-pkgconf-static-cflags.patch
+++ b/mingw-w64-mosquitto/0002-mosquitto-2.0.18-pkgconf-static-cflags.patch
@@ -1,18 +1,20 @@
 diff -Naur mosquitto-2.0.18.orig/libmosquitto.pc.in mosquitto-2.0.18/libmosquitto.pc.in
 --- mosquitto-2.0.18.orig/libmosquitto.pc.in	2024-04-10 00:32:55.137531100 +0300
 +++ mosquitto-2.0.18/libmosquitto.pc.in	2024-04-10 21:47:48.600521000 +0300
-@@ -7,4 +7,5 @@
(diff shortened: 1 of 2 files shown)


Biswa96:
@jeremyd2019 xsltproc fails to build man pages for this package in clangarm64 only. Could you share any hint to troubleshoot the issue?

jeremyd2019:
This seems familiar.  I attempted to build this package in the MINGW-packages ci on clangarm64, and it succeeded.

jeremyd2019:
I also stopped the runner and built the package manually on the same instance as the autobuild runner runs (from the same git checkout even), and it also succeeded.  I manually uploaded that package.

PR #20663, opened 2024-04-18
python-spyder: update to 5.5.4

Biswa96:


Biswa96:
The new version of python-spyder is working fine in my setup.

image

PR #20664, opened 2024-04-18
mingw-w64-python-xmltodict - new package.

JPeterMugaas:


Biswa96:
The upstream repository is not maintained anymore and last commit was a year ago. Do we need to package this for other projects?

JPeterMugaas:
We probably don't need that package.

PR #20665, opened 2024-04-18
texinfo: Add package for info files

### LINKED ISSUE #631: Emacs info files are not included on top tree
Start Emacs and show the info tree top (`C-h i`) The Emacs manuals are not listed.

Likewise if you use the command-line `info` viewer.

Apparently the Emacs manuals are not installed correctly on the info index (the files are there, though.)

Biswa96:
Fixes #631

The hook files are copied from msys2/texinfo directory.
The sed command in package function is copied from mingw glib2 directory.

### DIFF
--- a/mingw-w64-texinfo/PKGBUILD
+++ b/mingw-w64-texinfo/PKGBUILD
@@ -1,0 +1,62 @@
+# Maintainer: Biswapriyo Nath <nathbappai@gmail.com>
+
+_realname=texinfo
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=7.1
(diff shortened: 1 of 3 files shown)


mmuetzel:
Nice! 🎉

That might mean that Octave would no longer need the work-around to be able to call the (Perl-based) `makeinfo` from the MSYS2 environment. (I haven't tested that though...)

Biswa96:
One PITA though, I have to reinstall mingw emacs to update the info dir file. It feels like expected behavior but I am not sure enough.

mmuetzel:
Quick test with the build artifact for MINGW64: It seems to do the job for displaying help at the command prompt. 👍

I'll try to update Octave once this is merged and I had the opportunity to run a couple more tests.

PR #20672, opened 2024-04-19
slim-simulator: update to 4.2.1 (sorry so soon, major bug fix)

rdinnager:
The was a major bug in slim-simulator version 4.2 so we have to update to 4.2.1, which has fix. I know it is sooner that is encouraged for updating but hopefully this is considered a reasonable exception.


ognevny:
it's ok when there is a need to fix a major bug, sometimes important changes are backported before the official release. so, no problem :)

rdinnager:
Ok, thanks for the clarification. Cheers :)

PR #20674, opened 2024-04-19
octave: depend on texinfo (mingw-w64)

mmuetzel:
The help system is a central part of Octave that should never have been an *optional* dependency. Calling almost any function results in odd output without it.
Fix that. And switch to the MinGW version of texinfo.


mmuetzel:
The build artifacts seem to be working fine.

PR #20700, opened 2024-04-22
imagemagick: update to 7.1.1.31

kmilos:


mingw-w64-imagemagick/PKGBUILD
```diff
@@ -108,8 +106,7 @@ prepare() {
 
   # https://github.com/ImageMagick/ImageMagick/issues/7230
```
Biswa96, line 107: The comment can be removed now.

kmilos: Good catch, thanks.

PR #20702, opened 2024-04-22
meson: add optdepends on sccache and ccache

ognevny:
see https://mesonbuild.com/Feature-autodetection.html#ccache and https://mesonbuild.com/Release-notes-for-0-61-0.html#added-support-for-sccache

### DIFF
--- a/mingw-w64-meson/PKGBUILD
+++ b/mingw-w64-meson/PKGBUILD
@@ -4,7 +4,7 @@ _realname=meson
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.4.0
-pkgrel=2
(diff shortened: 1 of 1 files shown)


ognevny:
any objections?

Biswa96:
This not required in my opinion. One can install sccache/ccache and choose to use it without meson. Or, meson does not need sccache/ccache for general use cases.

ognevny:
> This not required in my opinion. One can install sccache/ccache and choose to use it without meson. Or, meson does not need sccache/ccache for general use cases.

that's why I added them to `optdepends`. I accidentally found, that meson supports `sccache` without any extra options to be passed

lazka:
fair enough I guess. It's just one of many things meson can use, but a very useful and easy to miss one. I've never used sccache though, only ccache.

PR #20711, opened 2024-04-23
manim: update to 0.18.0.post0

ognevny:
~~`sed`ing is not required now~~


ognevny:
I was wrong, the `0.18.0.post0` tag still has ugly requirements

PR #20714, opened 2024-04-23
libminpack: fix broken link on spdx license

luau-project:
## Description

The current PKGBUILD for **mingw-w64-libminpack** has an invalid identifier for spdx license. As such, the webpage for the library points to a link that does not exist.

## How

If I understood correctly how spdx licenses work on PKGBUILD, the term that comes after 'spdx:' seems to be the **Short Identifier** as defined by spdx website.

Thus, in this PR, I replaced the old license value (```license=('spdx:BSD-4-Clause-Modified')```) with ```license=('spdx:Minpack')``` on PKGBUILD. In the new behavior,
it should point to [Minpack License on SPDX](https://spdx.org/licenses/Minpack.html) and fix the problem.

> [!NOTE]
> 
> Since this fix does not change binaries, I didn't increase ```pkgrel``` field on PKGBUILD. I don't know the MSYS2 position on this subject, so the reviewer can take that action in the proper manner.

## Actual behavior

actual-behavior

## Expected behavior after the fix

expected-behavior


Biswa96:
According to SPDX page, the minpack license in not a free software license. Should the package be removed?

ognevny:
the license will be changed only after rebuild, so, please, bump `pkgrel`

luau-project:
@Biswa96 

> According to SPDX page, the minpack license in not a free software license. Should the package be removed?

I'm not sure, but I think this package is on debian repositories for so long that I don't see a real problem. It seems, like the old value had, a BSD-4-Clause with minor modifications. Anyway, it is up to MSYS2 folks to discuss and take the best road.

@ognevny 

> the license will be changed only after rebuild. so, please, bump `pkgrel`

Done.

PR #20719, opened 2024-04-24
starpu: update to 1.4.5

3rav:


Biswa96:
Would you like to bump pkgrel in pastix for this update? See https://github.com/msys2/MINGW-packages/actions/runs/8818547312/job/24208569547?pr=20719#step:3:40

Biswa96:
@jeremyd2019 starpu fails to build in clangarm64 because the configure script finds python executable somewhere. Could you reproduce the error? Here is the difference in build log.

* clangarm64

```
checking for python3... python3
checking for python3 version... 3.11
checking for Python.h... yes
checking for PyErr_Print in -lpython3.11... yes
...
```

* ucrt64

```
checking for python3... no
configure: WARNING: python3 missing, cannot build StarPU python interface
```

It could be possible to disable python module with `--disable-starpupy` option but I am not sure enough.

jeremyd2019:
The arm self-hosted runners have mingw-w64-clang-aarch64-toolchain pre-installed.  python gets pulled in as a dependency of lldb I think.  I would either look for a configure option to tell it not to look for python, or else try to enable python for the other arches too.

Issue #20687, opened 2024-04-20
kiconthemes: KIconTheme needs to add BreezeIcons as build-time dependency and needs to be patched for init icon resources correctly


BLumia:
### Description / Steps to reproduce the issue

There are two issues:

1. It's suggested to build KIconTheme with BreezeIcons since icon theme works in a different way under Windows. when adding BreezeIcons as a build-time dependency, KIconTheme will help init icon resources and ensure app also find icons from Qt's resource system, so programs can use BreezeIcons as they do under Linux.
2. KIconTheme's 6.1.0 release has a typo: https://github.com/KDE/kiconthemes/commit/d8ee0caf5b940a310020dd68ecb6905e67413ed4 which will cause the icon resource not be able to init correctly. KDE applied a patch to fix this issue from their build system (Craft), see https://github.com/KDE/craft-blueprints-kde/blob/9335df8a24b10043295086e8079909736f551627/kde/frameworks/tier3/kiconthemes/kiconthemes-6.1.0-20240415.diff

And another issue I haven't figure it out right now:

- KDE also [apply another patch in their Craft build system](https://github.com/KDE/craft-blueprints-kde/blob/9335df8a24b10043295086e8079909736f551627/kde/frameworks/tier3/kiconthemes/svgiconengine.diff) to [enforce iconengine plugin is there on (mac and) Windows](https://github.com/KDE/craft-blueprints-kde/blob/9335df8a24b10043295086e8079909736f551627/kde/frameworks/tier3/kiconthemes/kiconthemes.py#L11). I'm not quite sure but we likely might also need to apply this patch too.

### Expected behavior

kiconthemes is build with BreezeIcons and patched, so KDE apps like Kate can find breeze icons, as lone as they are depend on kiconthemes.

### Actual behavior

kiconthemes (currently `ucrt64/mingw-w64-ucrt-x86_64-kiconthemes 6.1.0-1`) is not build with BreezeIcons and not patched.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22631

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

[...]

PR #20720, opened 2024-04-24
kate: update to 24.02.2

BLumia:
The patch is also upstreamed to kate's repo: https://invent.kde.org/utilities/kate/-/merge_requests/1473


Biswa96:
I was planning to update all KDE apps at once after fixing breeze and breeze icons because many icons are missing for those packages.

BLumia:
> after fixing breeze and breeze icons because many icons are missing for those packages.

I guess the "missing icon" issue is likely caused by https://github.com/msys2/MINGW-packages/issues/20687 (no icon), then caused by the symlink issue you previously mentioned. I'm currently trying to figure out how to address https://github.com/msys2/MINGW-packages/issues/20687 and then planned to also look into the symlink issue if it's still not addressed at that time. But are these two issues considered as blocking issues? If not, I personally suggest we update some of these KDE apps if it's okay. For detail, I plan to update Kate (this PR) and Konsole currently so it can also be consumed by `KParts`' API.

Biswa96:
ok, I shall add some modifications in this pkgbuild. please wait.

PR #20721, opened 2024-04-24
apply svgiconengine patch to kiconthemes package

BLumia:
Related discussion:

- https://invent.kde.org/utilities/kcharselect/-/merge_requests/20
- https://github.com/msys2/MINGW-packages/issues/20687

This is marked as a draft for now since it doesn't really work as expected currently, i.e. still have the same issue mentioned in the links above. Any help and/or discussions are appreciated.

### DIFF
--- a/mingw-w64-kiconthemes/PKGBUILD
+++ b/mingw-w64-kiconthemes/PKGBUILD
@@ -24,6 +24,7 @@ depends=(
   "${MINGW_PACKAGE_PREFIX}-kwidgetsaddons"
   "${MINGW_PACKAGE_PREFIX}-qt6-base"
   "${MINGW_PACKAGE_PREFIX}-qt6-svg"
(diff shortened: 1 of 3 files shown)


Biswa96:
Are these patch required for kiconthemes 6.2.0 ? See https://invent.kde.org/packaging/craft-blueprints-kde/-/blob/b06953082144acfac6388a518b573a3ccc85286c/kde/frameworks/tier3/kiconthemes/kiconthemes.py

Those patches are applied only for kiconthemes < 6.2.0.

BLumia:
> Are these patch required for kiconthemes 6.2.0 ?

No, only 6.1.x needs such a patch.

BLumia:
Closing since we now have kiconthemes 6.2.0 and this patch is no longer needed.

PR #20725, opened 2024-04-24
kdiff3: update to 1.11.0

podsvirov:


lazka:
lgtm, what about the kparts change?

podsvirov:
> lgtm, what about the kparts change?

Looks like the kparts [no more used](https://invent.kde.org/sdk/kdiff3/-/commit/14070b6161be7c9425c5c2be03191feca7ce276b).

PR #20732, opened 2024-04-25
python-cx-freeze: restore mingw32 build

cedk:


Biswa96:
Please explain why you need the 32 bit package which have being dropped since last year https://www.msys2.org/news/#2023-12-13-starting-to-drop-some-32-bit-packages

cedk:
The cx-freeze 32 bits have been dropped recently db875ac3d46e7c6ad103a09ae93efc59adb65a2c
We need it to build the Windows client for https://www.tryton.org/ on 32bits for which we still have users.

PR #20748, opened 2024-04-26
qt6-base: fix inclusion of OpenGL header after Qt6 header

mmuetzel:
The Qt6 header `qopengl.h` includes `<GL/gl.h>` while trying to avoid the inclusion of `windows.h`. It does that by defining two preprocessor macros that would normally be defined by including `windows.h` in `<GL/gl.h>`. If these macros are defined, `<GL/gl.h>` does not include `windows.h`. At the end of `qopengl.h`, these preprocessor macros are undefined.

If a project tries to include `<GL/glu.h>` after (indirectly) including `qopengl.h`, the following happens:
* `<GL/glu.h>` includes `<GL/gl.h>` and expects that the necessary preprocessor macros are getting defined by including that header.
*  `<GL/gl.h>` was already included before (in `qopengl.h`) and the inclusion guards lead to `#include <GL/gl.h>` being a no-op at that point.
*  The necessary preprocessor macros are undefined which leads to compilation errors.

This PR adds a patch that
* adds definitions for two more preprocessor macros in `qopengl.h`  that are needed when including `<GL/glu.h>` in user code and
* no longer undefines these (four) preprocessor macros at the end of `qopengl.h`.

### DIFF
--- a/mingw-w64-qt6-base/005-qt-6.7.0-opengl-header.patch
+++ b/mingw-w64-qt6-base/005-qt-6.7.0-opengl-header.patch
@@ -1,0 +1,45 @@
+Avoid issue for projects that include <GL/glu.h> after <QOpenGLContext>
+on Windows.
+
+diff -urN qtbase-everywhere-src-6.7.0/src/gui/opengl/qopengl.h.orig qtbase-everywhere-src-6.7.0/src/gui/opengl/qopengl.h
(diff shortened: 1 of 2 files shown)


mmuetzel:
I downloaded and installed the build artifact for MINGW64 from this PR. With it, it is possible to build a current head of the default branch of Octave with Qt6 without the previous issues.
Maybe, we could switch back to Qt6 once Octave 9.2.0 is released. (The crash on `figure()` with Qt6 version 6.7.0 should also be fixed in Octave 9.2.0.)

PR #20750, opened 2024-04-26
python-pytest-datafiles: add new package

marcelotduarte:
I will use it to implement check() in python-cx-freeze.


Biswa96:
Are 32 bit packages required?

marcelotduarte:
I copied mingw_arch from other pytest plugins. But I think only mingw32 will be needed because I saw that cedk will need it for now in cx_Freeze. So if you want, I can remove clang32.

marcelotduarte:
By the way, I was planning to update pytest and its plugins, so I can disable clang32 for everyone too?

Biswa96:
> so I can disable clang32 for everyone too?

If there is not reverse dependencies, yes absolutely.

marcelotduarte:
Updated. Can you merge?

PR #20752, opened 2024-04-27
swi-prolog: update to 9.2.4

mgondan:
now includes the test suite, run with ?- test_installation.

### DIFF
--- a/mingw-w64-swi-prolog/PKGBUILD
+++ b/mingw-w64-swi-prolog/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=swi-prolog
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=9.2.3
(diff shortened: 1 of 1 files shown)


Biswa96:
Are those tests require for normal usage?

lazka:
Looks like they are just installed tests: https://github.com/SWI-Prolog/swipl-devel/blob/c60103a2df235d6c52221576a6b531df67bbcbd6/doc/Testing.md?plain=1#L57

mgondan:
Sorry for not responding earlier, but you found it out anyway. The tests are not strictly required---on the other hand, professional users will probably appreciate them.

Biswa96:
> The tests are not strictly required---on the other hand, professional users will probably appreciate them.

I was wondering if tests should be in a separate package e.g. swi-prolog-tests or swi-prolog-data. As tests are not strictly required, not all users need them.

mgondan:
Good point. I'll put them into a separate package (in the next release, if that is ok).

PR #20780, opened 2024-04-30
avrdude: A better patch

ikspress:
https://github.com/avrdudes/avrdude/commit/c4b5097b1fdfb19c1ebd31193692e1aac2ebbfd7

https://github.com/msys2/MINGW-packages/pull/20652 I found a better patch.


Biswa96:
The actual result is same though.

PR #20785, opened 2024-05-02
emacs: depend on texinfo for proper info directory update upon install

oscarfv:
This completes the fix for issue #631

### DIFF
--- a/mingw-w64-emacs/PKGBUILD
+++ b/mingw-w64-emacs/PKGBUILD
@@ -8,7 +8,7 @@ _realname=emacs
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=29.3
-pkgrel=2
+pkgrel=3
 pkgdesc="The extensible, customizable, self-documenting, real-time display editor (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


oscarfv:
Looks like the ucrt64 build timed out. Forcing a new check run...

oscarfv:
@cyrilarnould : do you plan to submit a PR for the workaround mentioned on https://debbugs.gnu.org/cgi/bugreport.cgi?bug=70889#25 ?

Otherwise I will.

Thanks to you and @lazka for analyzing the faulty gcc 14.1 build.

cyrilarnould:
I'm currently checking whether the `PKGBUILD` recipe works on my machine with the workaround :)

But feel free to integrate it into your PR, my suggestion would have been

```
diff --git a/mingw-w64-emacs/PKGBUILD b/mingw-w64-emacs/PKGBUILD
index 469d137ca..7c088c6bc 100644
--- a/mingw-w64-emacs/PKGBUILD
+++ b/mingw-w64-emacs/PKGBUILD
@@ -83,6 +83,9 @@ build() {
   CFLAGS=${CFLAGS//"-Wp,-D_FORTIFY_SOURCE=2"}
   # -foptimize-sibling-calls breaks native compilation (GCC 13.1)
   CFLAGS+=" -fno-optimize-sibling-calls"
+  # configure script can not deal with the warnings that were turned
+  # into errors in GCC 14
+  CFLAGS+=" -Wno-error=implicit-function-declaration"

   ../${_realname}-${pkgver}/configure \
     --prefix="${MINGW_PREFIX}" \

```

oscarfv:
Okay, PR updated, thanks @cyrilarnould .

The occasional CI build failure seems to be a timing issue either on emacs' build system or in the nativecomp machinery.

cyrilarnould:
Upstream has fixed the issue `-Wno-error=implicit-function-declaration` with a different patch: 

https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=5216903ae6c3f91ebefb1152af40753f723cbc39

Maybe using this might be more appropriate?

In the meantime, they've also integrated the `-fno-optimize-sibling-calls`

lazka:
thanks, I've referenced the commits in 0d77e7c68b6698f40c45f88399eed1caaf59baf3, so we can drop the workarounds maybe on the next upstream release.

PR #20800, opened 2024-05-03
rust: update to 1.78.0

filnet:


filnet:
There is something wonky going on with embed-manifest.
First, I had to change its sha twice and then the build fails with this:

```
error: the listed checksum of `C:\_\B\src\rustc-1.78.0-src\vendor\embed-manifest\sample.exe.manifest` has changed:
  expected: 01e80ef76de2b628d452c7e80022654b9e0c8aee72ec64ee522c7083d835f4df
  actual:   a0e53ad5d65af69fe608999bcdab2e7c705a27c4a722c444d2d12cb89de68b6b
```

We are not patching the embed-manifest crate so the provided checksums should be fine.

filnet:
We are actually patching the checksums...

jeremyd2019:
Well, that's *odd* - it seems the line endings in sample.exe.manifest changed (from CRLF in the .tar.gz with sha256sum `f1c65919a5f182376ecbfed69f72935abbebad5dc62bd32b2038905258c49453` to LF in `24ef6d949c0b5b1940c1d6a7aad78d86012152fb8845a1644bc939350d7b75e2`).  running `dos2unix` on the version in the old .tar.gz results in it matching the file in the new .tar.gz.  I didn't see any new commits, maybe gitlab suddenly started respecting `.gitattributes` when generating the .tar.gz exports?  1.4.0 changed .gitattributes to say `*.manifest` should have crlf and that one has crlf in the sample.exe.manifest (which is now in testdata subdirectory)

PR #20810, opened 2024-05-05
update sundials to 7.0.0

MehdiChinoune:
based on #20254


mmuetzel:
Please, don't remove SUNDIALS support from Octave (for any environment).

PR #20820, opened 2024-05-06
uutils-coreutils: update to 0.0.26

ognevny:
the issue didn't disappear, but I found that generating doesn't fail for some progs. draft until I investigate it


ognevny:
generating works fine, BUT both `> file` and `| install /dev/stdin` segfault

PR #20829, opened 2024-05-07
GCC: rebuild

lazka:
There were some reports that there might be problems with newer cygwin

let's try...


lazka:
Luckily all looks fine

PR #20835, opened 2024-05-08
onefetch: update to 2.21.0

ognevny:


Biswa96:
Could you provide the source of the patch file? Or, is it mingw specific?

ognevny:
> Could you provide the source of the patch file? Or, is it mingw specific?

I wrote this patch myself. I'm planning to create forks with this change and use [patch.crates-io]

Biswa96:
Understood. Although, I would allow onefetch (or any other rust projects) to link statically if it does not support shared linking.

ognevny:
> Understood. Although, I would allow onefetch (or any other rust projects) to link statically if it does not support shared linking.

this is not about upstream projects, but developers who use this bindings. the main idea is fully static linking in rust (which is even documented in homepage), but cargo and rustc have a stable support of shared libraries (e.g. windows system libraries). so why not?

Biswa96:
> so why not?

This kind of patches are harder to maintain.

ognevny:
> > so why not?
> 
> This kind of patches are harder to maintain. KISS

I can :). as I said I have a plans how to make things easier. but it requires some time. will try tomorrow

PR #20841, opened 2024-05-08
[sentencepiece] Add new package. 

### LINKED ISSUE #20832: Add [python-sentencepiece]
### Package name

python-senetencepiece and sentencepiece

Python sentencepiece is a google project which is basically a tokenizer for LLM if I understand correctly?
they can be either distributed as two different packages or just one, the reason to request here is that it can't be build with a mere pip install

_No response_

`import sentencepiece` in python,

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

_No response_

Kreijstal:
Sentencepiece is a llibrary that is essentially a tokenizer for llms

fixes #20832

### DIFF
--- a/mingw-w64-sentencepiece/0001-mingw-build.patch
+++ b/mingw-w64-sentencepiece/0001-mingw-build.patch
@@ -1,0 +1,39 @@
+From: Kreijstal <kreijstal@hotmail.com>
+Subject: [PATCH] Build cleanly under MinGW-w64
+
+The CMake `if(WIN32)` block force-disables shared libraries; constrain it
+to MSVC so MinGW-w64 (and its protobuf-shared deps) can build the DLLs.
+
+`setup.py`'s `os.name == 'nt'` branch hard-codes MSVC import-library
+paths. Skip it when Python was built with GCC (mingw-w64) so the unix
+build_ext path runs and links against the staged `../build/root` install
+through pkg-config.
+---
+ CMakeLists.txt  | 2 +-
+ python/setup.py | 2 +-
+ 2 files changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -46,7 +46,7 @@ if (SPM_CROSS_SYSTEM_PROCESSOR)
+ endif()
+
+ # Disable shared build on windows
+-if(WIN32)
++if(MSVC)
+   set(SPM_ENABLE_SHARED OFF)
+ endif()
+
+diff --git a/python/setup.py b/python/setup.py
+--- a/python/setup.py
++++ b/python/setup.py
+@@ -161,7 +161,7 @@ def get_win_arch():
+   return arch
+
+
+-if os.name == 'nt':
++if os.name == 'nt' and 'GCC' not in sys.version:
+   # Must pre-install sentencepice into build directory.
+   arch = get_win_arch()
+   if os.path.exists('..\\build\\root_{}\\lib'.format(arch)):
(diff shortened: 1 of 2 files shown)


mingw-w64-python-sentencepiece/0001-make-it-more-friendly-for-mingw-enviroments.patch
```diff
@@ -0,0 +1,40 @@
...
+index 2d3ecfc..fbd9140 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -43,7 +43,7 @@ if (SPM_CROSS_SYSTEM_PROCESSOR)
+ endif()
+ 
+ # Disable shared build on windows
+-if(WIN32)
++if(MSVC)
+   set(SPM_ENABLE_SHARED OFF)
```
Biswa96, line 21: Can this option be set from cmake command instead of patching it here?

mingw-w64-python-sentencepiece/0001-make-it-more-friendly-for-mingw-enviroments.patch
```diff
@@ -0,0 +1,40 @@
...
+index d600321..d0a1552 100755
+--- a/python/setup.py
++++ b/python/setup.py
+@@ -102,8 +102,7 @@ class build_ext(_build_ext):
+     ext.extra_link_args = libs
+     _build_ext.build_extension(self, ext)
+ 
+-
+-if os.name == 'nt':
++if os.name == 'nt' and not os.environ.get('SENTENCEPIECE_MINGW_SKIP'):
```
Biswa96, line 34: Environment variable is not required here. Just check "gcc" string in sys.version.

Biswa96:
Why is there two different PKGBUILDs? Can those be done in one PKGBUILD, the source tarball is same?

mingw-w64-python-sentencepiece/0001-make-it-more-friendly-for-mingw-enviroments.patch
Kreijstal: How? If I enable it win32 will disable it... No? I have no idea, can you assist?

Kreijstal:
> Why is there two different PKGBUILDs? Can those be done in one PKGBUILD, the source tarball is same?

One is a c++ library and the other one is just the python interface, I only manage to solve building in Python when the c++ library is already installed, and you can only have one build function instead of two even on multipackage pkgbuilds, if you want you can close this pull request and just show me how you'd do it so I can learn as a reference? I'd like some help here

mingw-w64-python-sentencepiece/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+pkgrel=1
+pkgdesc="Unsupervised text tokenizer for Neural Network-based text generation (MinGW-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/google/sentencepiece'
+license=('Apache-2.0')
+depends=("${MINGW_PACKAGE_PREFIX}-sentencepiece")
+makedepends=("${MINGW_PACKAGE_PREFIX}-python-build"
+"${MINGW_PACKAGE_PREFIX}-python-installer" 
+"${MINGW_PACKAGE_PREFIX}-python-setuptools
```
ognevny, line 16:
```suggestion
"${MINGW_PACKAGE_PREFIX}-python-setuptools"
```

Kreijstal: Funny if seems this broke the package?? How?

Kreijstal:
> Why is there two different PKGBUILDs? Can those be done in one PKGBUILD, the source tarball is same?

Because I had skill issues,..  but I still do! for some reason it is installing on /msys64/ucrt64 and not on /ucrt64, I don't know why

Kreijstal:
I finally did it, it took forever

ognevny:
CI still fails

Kreijstal:
> CI still fails

I'm too eager

PR #20855, opened 2024-05-10
libslirp: Update to 4.8.0

lazka:
* Drop patch included in the new release
* Now contains symlinks, so do the double extract trick


lazka:
the clang build lost a lot of symbols, but looks like those were never intended to be exported, so looks fine.

PR #20874, opened 2024-05-11
Fortran-stdlib: v0.6.0

zoziha:
### Description

`Fortran-stdlib` comes with modern Reference-BLAS built into v0.6.0, and also provides macro definitions to use external BLAS and LAPACK.
This PR upgrades `Fortran-stdlib` to v0.6.0 and links to OpenBLAS for optimized performance by adding a CMake patch.

### DIFF
--- a/mingw-w64-fortran-stdlib/001-using-openblas.patch
+++ b/mingw-w64-fortran-stdlib/001-using-openblas.patch
@@ -1,0 +1,36 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index b10e1f7..0029d06 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -42,6 +42,17 @@ if(NOT DEFINED CMAKE_MAXIMUM_RANK)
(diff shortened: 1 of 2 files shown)


Biswa96:
Did you try to upstream the patch?

zoziha:
> Fortran-stdlib does not address enabling OpenBLAS or other implementations in CMake yet. Current focus is to implement linear algebra subroutines, that will come next.

Fortran-stdlib requires reasonable references to external BLAS libraries from complex sources (e.g. intel MKL, OpenBLAS, etc.), which requires a long implementation plan. Since the OpenBLAS package is included in MSYS2, my simple idea is to take advantage of it directly to get optimized performance, and I will continue to follow the progress of Fortran-stdlib.

@Biswa96, if you think it's **not appropriate** to add a CMake patch for Fortran-stdlib to MSYS2 **at this stage**, I'll remove the CMake linking OpenBLAS patch for Fortran-stdlib and **revert to** **updating only the Fortran-stdlib version**.

PR #20883, opened 2024-05-12
objfw: Update to 1.1.3

Midar:


ognevny:
why draft?

Midar:
I always create as draft first until CI passed :)

ognevny:
> I always create as draft first until CI passed :)

ok

PR #20939, opened 2024-05-18
qt5-base: rebuild

lazka:


MehdiChinoune:
No Vulkan!

MehdiChinoune:
I see now, vulkan-headers is missing in makedepends()
I remove dependency of vulkan-loader on vulkan-headers in https://github.com/msys2/MINGW-packages/commit/27d35e9e

lazka:
ah, yeah, so related to https://github.com/KhronosGroup/Vulkan-Loader/issues/1463

lazka:
fixed via c0e699eefafe97e337881a613999a54cd7e9d63c then.

PR #20954, opened 2024-05-20
kdiff3: update to 1.11.1, port to kf6

podsvirov:


Biswa96:
Thank you for the update. I have added icoutils for adding icons in executable and mentioned gettext-tools explicitly.

PR #20955, opened 2024-05-20
vsgpoints: new package

podsvirov:


lazka:
I wonder why it fails in autobuild

lazka:
doxygen update maybe?

Biswa96:
Do we really need docs for this or any vulkanscenegraph packages? Those are just API docs and not built by default.

lazka:
likely fixed via #20967

PR #20968, opened 2024-05-21
gnustep-base: patch in some xml fixes from main branch of gnustep-base

jcaselman-keysight:
Adds patch for the merge commit from https://github.com/gnustep/libs-base/pull/405 to msys2 `gnustep-base` package

cc @qmfrederik


jcaselman-keysight:
I see the CI failures, will re-open once I have a solution

PR #20972, opened 2024-05-22
kicad: update to 8.0.2

mmuetzel:
The newest doc tarball that I could find is for version 8.0.0-rc3.

Does someone know where newer doc tarballs could be found?
Should we remove the doc packages? Users could still find the online documentation.

### DIFF
--- a/mingw-w64-kicad-doc/PKGBUILD
+++ b/mingw-w64-kicad-doc/PKGBUILD
@@ -15,8 +15,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}-ca"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-pl"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-ru"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-zh")
-pkgver=8.0.0
-_rcver=rc3
+pkgver=8.0.2
 pkgrel=1
 pkgdesc="Documentation for KiCad (mingw-w64)"
 arch=(any)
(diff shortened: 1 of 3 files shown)


Biswa96:
> The newest doc tarball that I could find is for version 8.0.0-rc3.

Would it be same from git tag archive? https://gitlab.com/kicad/services/kicad-doc/-/tags

mmuetzel:
> > The newest doc tarball that I could find is for version 8.0.0-rc3.
> 
> Would it be same from git tag archive? https://gitlab.com/kicad/services/kicad-doc/-/tags

I tried to build the documentation from the Git repository. However, we don't have most of the necessary build tools in MSYS2. So, I stopped looking into that option.

mmuetzel:
Is it ok to merge this as it is (i.e., with the doc package from the release candidate)? Or should the `*-kicad-doc` packages be removed?

mmuetzel:
Alternatively, we could try to use the build artifact from here:
https://gitlab.com/kicad/services/kicad-doc/-/jobs/6731879825

But I don't know if that will expire at some point in the future...

mmuetzel:
The build artifacts for the `8.0.2` release tag have the following note on their webpage:
https://gitlab.com/kicad/services/kicad-doc/-/jobs/6731879825
> These artifacts are the latest. They will not be deleted (even if expired) until newer artifacts are available.

The same for the build artifacts for the `8.0.1` release tag:
https://gitlab.com/kicad/services/kicad-doc/-/jobs/6388910557

So, I'm assuming these build artifacts will be available for the foreseeable future.
I've added a commit that switches to the build artifacts as the source for the documentation.

I've downloaded and installed the build artifacts for the MINGW64 environment. Kicad seems to be working fine with them.

PR #20975, opened 2024-05-22
sqlite3/cairo/gc: rebuild

### LINKED ISSUE #20974: cairo: pkgconfig files reference `-ldl` in Libs.private on clangarm64
I believe this is the cause of the build error of inkscape currently being seen.  I don't know why that is though.  This is not present in the version in clang64.

jeremyd2019:
somehow they ended up with -ldl in Libs.private in their pkgconfig files on clangarm64.

Add workaround for gcc errors from https://github.com/msys2/MINGW-packages/issues/20974#issuecomment-2125518371

Fixes #20974


jeremyd2019:
I did a clangarm64 CI run with also rebuilding inkscape, at https://github.com/msys2-arm/MINGW-packages/actions/runs/9196521728/job/25294798745, but that was before adding the gcc workaround.

PR #20982, opened 2024-05-23
swi-prolog: new version 9.2.5

mgondan:
swi-prolog: new version 9.2.5


lazka:
I guess it's fine that way too, but I think @Biswa96 meant a split package (??)

mgondan:
Can you guide me how to create a split package? Is there an example on MSYS2 that I can copy from?

lazka:
I pushed something.

I noticed that the `BUILD_TESTS` option doesn't exist and there is a `BUILD_TESTING` option instead. I didn't touch it though, in case you want to have a look.

mgondan:
I think it correct as it is.

https://github.com/SWI-Prolog/swipl-devel/blob/master/CMAKE.md

BUILD_TESTING and INSTALL_TESTS

lazka:
You added BUILD_TESTS in the first commit

mgondan:
You're right. Fixed, thanks for noticing.

mgondan:
I see what you mean with the conflicts, thank you.

mgondan:
One of the swipl "packs" shows a strange dependency. If I `makepkg` swi-prolog on my local msys2, "janus.dll" is linked correctly against python's dll:

ldd /ucrt64/lib/swipl/lib/x64-win64/janus.dll
        ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x7ffda6910000)
     ...
        libpython3.dll => /ucrt64/bin/libpython3.dll (0x7ffd99320000)
     ...
        libpython3.11.dll => /ucrt64/bin/libpython3.11.dll (0x7ffd36b50000)

If I use the binary from msys2.org (pacman -S mingw-w64-ucrt-x86_64-swi-prolog), I get

$ ldd /ucrt64/lib/swipl/lib/x64-win64/janus.dll
    ...
        python311.dll => not found

This is strange, since the python package from msys2 does not seem to include a dll with that name.

MehdiChinoune:
> One of the swipl "packs" shows a strange dependency. If I `makepkg` swi-prolog on my local msys2, "janus.dll" is linked correctly against python's dll:
> 
> ldd /ucrt64/lib/swipl/lib/x64-win64/janus.dll ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x7ffda6910000) ... libpython3.dll => /ucrt64/bin/libpython3.dll (0x7ffd99320000) ... libpython3.11.dll => /ucrt64/bin/libpython3.11.dll (0x7ffd36b50000)
> 
> If I use the binary from msys2.org (pacman -S mingw-w64-ucrt-x86_64-swi-prolog), I get
> 
> $ ldd /ucrt64/lib/swipl/lib/x64-win64/janus.dll ... python311.dll => not found
> 
> This is strange, since the python package from msys2 does not seem to include a dll with that name.

A fix is coming, CMake picked system Python (MSC) instead on mingw-w64-python

MehdiChinoune:
What does `tests` package serve?

mgondan:
As suggested previously, swi-prolog has been split into the normal package and an extra package containing the unit tests.

MehdiChinoune:
Are those unit tests required for someone who use swi-prolog (I mean Is it like gtest, pytest ...etc)?
Or just tests packaged separately?

MehdiChinoune:
I couldn't find any distribution packaging those unit tests!

mgondan:
It's the unit tests for Swi-Prolog itself. It's many, 7 MB, therefore the suggestion to put them into a separate package. By the way, I'll suggest to do the same in future distributions of Swi-Prolog on debian etc.

mgondan:
See discussion here,

https://github.com/msys2/MINGW-packages/pull/20752#issuecomment-2081111890

MehdiChinoune:
I am gonna remove it, We don't package tests.
You could build and run them in check() step.

lazka:
There is the concept of installed tests, that are not run at build time. I don't know what these here are though.

mgondan:
Please reconsider your decision. IIRC, check() is not even invoked on the msys2 package, and users of Swi-Prolog might appreciate that option.

MehdiChinoune:
> Please reconsider your decision. IIRC, check() is not even invoked on the msys2 package, and users of Swi-Prolog might appreciate that option.

You could run it locally, that's How we test packages.
check() doesn't run on CI.

passing -DINSTALL_TESTS=OFF remove those tests from the package.

mgondan:
I know. Package maintainers can run the unit tests on their system, msys2 could, in principle, do the same (but doesn't). For the users (I.e., those who type pacman -S ...), both is not really relevant. They might wish to run the tests after they have installed swi-prolog, invoking test_installation from the prolog prompt.

The tests were included by default on 9.2.4, but used quite a bit of space. Therefore, I've been asked to package them separately, and that's what I did now (well, with quite some help from lazka).

PR #20987, opened 2024-05-24
xmake: update to 2.9.2

tqfx:
What is https://github.com/msys2/MINGW-packages/actions/runs/9217271624/job/25358959159?pr=20987#step:11:1234? Although MINGW32/xmake works on my device.


Biswa96:
> What is https://github.com/msys2/MINGW-packages/actions/runs/9217271624/job/25358959159?pr=20987#step:11:1234?

Those are DLLs from CI server and can be ignored in this case.

PR #20991, opened 2024-05-25
sccache: update to 0.8.1

ognevny:


Biswa96:
Adding own forked repository in dependency seems a bit strange to me. Same goes to zed package also.

Biswa96:
> this fork has only 1 extra commit which changes only 1 line of code. the same thing was before with vendoring

Can't it be added in upstream with an environment variable like zlib, oniguruma, libgit etc. ?

ognevny:
> > this fork has only 1 extra commit which changes only 1 line of code. the same thing was before with vendoring
> 
> Can't it be added in upstream with an environment variable like zlib, oniguruma, libgit etc. ?

I didn't try. guess the upstream maintainers would like to have static-only

lazka:
A more transparent strategy seems to be used by librsvg: https://github.com/msys2/MINGW-packages/blob/2030516c40175691c481e0b8640626f72a2eee3e/mingw-w64-librsvg/PKGBUILD#L41-L54

I'd prefer that too.

ognevny:
> A more transparent strategy seems to be used by librsvg: https://github.com/msys2/MINGW-packages/blob/2030516c40175691c481e0b8640626f72a2eee3e/mingw-w64-librsvg/PKGBUILD#L41-L54
> 
> I'd prefer that too.

sounds good. thanks

ognevny:
done

PR #20992, opened 2024-05-25
mesa: Update to 24.0.8

pal1000:
24.1.0 has many problems:
- [zink](https://gitlab.freedesktop.org/mesa/mesa/-/issues/11220), [llvmpipe and softpipe](https://gitlab.freedesktop.org/mesa/mesa/-/issues/11254) broke on Windows, leaving only dd3d12 OpenGL driver working;
- we can't build with `-Dshared-llvm=enabled` anymore.

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=24.0.6
+pkgver=24.0.8
 pkgrel=1
 pkgdesc="Open-source implementation of the OpenGL, Vulkan and OpenCL specifications (mingw-w64)"
 url="https://www.mesa3d.org/"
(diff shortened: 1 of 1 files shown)


Biswa96:
> For some reason this fails to build clc in CI, but I couldn't reproduce locally.

I can reproduce the compiler error in my system. Please make sure all packages are updated in your system.

MehdiChinoune:
compilers complain about missing definition of Dl_info which is part of libdl (dlfcn).
Maybe dlfcn is installed in your system.

I think mesa should fixed by using Win32 APIs to load dynamic libraries.

MehdiChinoune:
dlfcn.h is required for UNIX 
They use libdl for UNIX+LLVM-shared https://gitlab.freedesktop.org/mesa/mesa/-/blob/24.1/src/compiler/clc/clc_helpers.cpp?ref_type=heads#L64
with LLVM static, libdl is not needed
https://gitlab.freedesktop.org/mesa/mesa/-/blob/24.1/src/compiler/clc/clc_helpers.cpp?ref_type=heads#L851

What about Windows+LLVM-shared?

Biswa96:
I have asked for suggestion in the MR https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25568

pal1000:
> > For some reason this fails to build clc in CI, but I couldn't reproduce locally.
> 
> I can reproduce the compiler error in my system. Please make sure all packages are updated in your system.

Before trying this MR I built release for mesa-dist-win. That links statically even with MinGW by default as that project goal is making binaries portable. No wonder I couldn't reproduce.

PR #20997, opened 2024-05-27
curl: enable http3

MehdiChinoune:
Enabling nghttp3 requires enabling openssl-quic or ngtcp2. I have enabled openssl-quic following Arch.
openssl-quci requires to use a single ssl backend which requires to disable `schannel` with openssl backend.

### DIFF
--- a/mingw-w64-curl/PKGBUILD
+++ b/mingw-w64-curl/PKGBUILD
@@ -3,10 +3,11 @@
 _realname=curl
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
-         "${MINGW_PACKAGE_PREFIX}-${_realname}-gnutls"
-         "${MINGW_PACKAGE_PREFIX}-${_realname}-winssl")
+         $([[ ${CARCH} == i686 ]] || echo \
+           "${MINGW_PACKAGE_PREFIX}-${_realname}-gnutls" \
+           "${MINGW_PACKAGE_PREFIX}-${_realname}-winssl"))
 pkgver=8.8.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Command line tool and library for transferring data with URLs (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
I don't know what to do?. Enable openss-quic backend, ngtcp2 backend or leave it as it is.
curl says that only ngtcp2 is stable and all other backends are considered experimental (including openssl-quic). https://curl.se/docs/http3.html
Official Windows Installer was built with ngtcp2 backend https://curl.se/windows/
Arch enable openssl-quic. https://gitlab.archlinux.org/archlinux/packaging/packages/curl/-/blob/8.8.0-1/PKGBUILD?ref_type=tags#L88
Fedora doesn't support it https://src.fedoraproject.org/rpms/curl/blob/rawhide/f/curl.spec
same for OpenSUSE https://build.opensuse.org/projects/openSUSE:Factory/packages/curl/files/curl.spec?expand=1

lazka:
We don't ship any of the TLS libraries that would work with ngtcp2, so I don't think that's an option right now.

I would just go with openssl, ideally that will improve in the future. From what I see http3 isn't enabled by default, so no harm.

re `schannel`, got any info why that is disabled then?

MehdiChinoune:
> re `schannel`, got any info why that is disabled or does not work then?

I don't understand the question. I have disabled schannel because http3 doesn't accept multi-ssl (openssl+schannel).

lazka:
> I don't understand the question. I have disabled schannel because http3 doesn't accept multi-ssl (openssl+schannel).

I was just wondering where you read that, or where in the build code it fails. I don't understand why it couldn't support different levels of http depending on the ssl backend selected at runtime.

MehdiChinoune:
> I was just wondering where you read that, or where in the build code it fails. I don't understand why it couldn't support different levels of http depending on the ssl backend selected at runtime.

https://github.com/curl/curl/blob/curl-8_8_0/configure.ac#L4770

lazka:
perfect, thanks

PR #20999, opened 2024-05-27
Add rv32 target!

Kreijstal:
Add gcc cross compiler for RV32

### DIFF
--- a/mingw-w64-riscv32-unknown-elf-binutils/PKGBUILD
+++ b/mingw-w64-riscv32-unknown-elf-binutils/PKGBUILD
@@ -1,0 +1,57 @@
+# Maintainer: Fabien Chouteau <chouteau@adacore.com>
+# Based on mingw-w64-arm-none-eabi-binutils
+
+_realname=binutils
+_target=riscv32-unknown-elf
+
+pkgbase=mingw-w64-${_target}-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_target}-${_realname}")
+pkgver=2.41
+pkgrel=2
+pkgdesc='GNU Tools for RISC-V Embedded Processors - Binutils (mingw-w64)'
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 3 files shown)


Kreijstal:
It does build, for non-believers: https://github.com/kreijstal-contributions/MINGW-packages/actions/runs/9253416924

lazka:
Just curious, what do you need it for?

Kreijstal:
You use this to target RISCV processors that can run either in simulation, in fpga devices or custom microcontrollers. Very useful

lazka:
Yeah, I'm a bit hesitant as gcc targets tend to not get updated/maintained after they are added.

Could we maybe reduce the amount of environments it's available in? In which env would you use it?

lazka:
ah, that's good to know. So this being a copy doesn't make it worse at least.

let me re-run CI to see if it still builds.

Kreijstal:
it even works with clangarm64 :D

jhol:
I would also like to see this included.

Kreijstal:
> I would also like to see this included.

Optionally there are gcc flags to use multi arch on rv64... Some distros do it like that..

Kreijstal:
is there a mirror of ftp.gnu that we could use in ci lol

Peter0x44:
You can use (and are actually supposed to) use ftpmirror.gnu.org
though it can return a nonworking mirror sometimes too. Kernel.org typically works as a backup, that isn't as slow.

Peter0x44:
We recommend using this generic ftpmirror.gnu.org address wherever possible in links, documentation, etc., to reduce load on the main GNU server. https://www.gnu.org/server/mirror.en.html

https://github.com/mstorsjo/llvm-mingw/commit/a96bbc0c7fc69e0a7d86678fb3d8c0234d7eb7d2

https://github.com/skeeto/w64devkit/commit/e97a594ca059b1c7b1355304a43c4a190a51622f

Kreijstal:
> You can use (and are actually supposed to) use ftpmirror.gnu.org though it can return a nonworking mirror sometimes too. Kernel.org typically works as a backup, that isn't as slow.

this is a copy of rv64 so you should update it there first and I just copy + sed it here

Issue #21000, opened 2024-05-27
Build nsis for CLANGARM64


kmilos:
### Package name

nsis

### For which environment do you need the package?

CLANGARM64

### Provide a brief description why you need this package in the selected environment

Used by many projects to package Windows installers, but currently not able to package ARM64 installers natively.

Note that also the CLANG64 build is missing, so this could be equally down to the compiler support and not just the platform...

PR #21003, opened 2024-05-27
nsis: enable clang

raedrizqie:
Fixes: #21000 

I cannot test for clangarm64 though..


lazka:
great!

lazka:
Just curious, did mingw32 got dropped because it caused issues?

lazka:
```
scons: *** [build/urelease/System/unicode/Source/Call-arm64CPP.o] Source `Contrib/System/Source/Call-arm64CPP.S' not found, needed by target `build/urelease/System/unicode/Source/Call-arm64CPP.o'.
```

should we disable arm64 again?

raedrizqie:
> Just curious, did mingw32 got dropped because it caused issues?

I drop mingw32 because now we are not using msys2 scons, and there is no mingw32 scons.

raedrizqie:
> ```
> scons: *** [build/urelease/System/unicode/Source/Call-arm64CPP.o] Source `Contrib/System/Source/Call-arm64CPP.S' not found, needed by target `build/urelease/System/unicode/Source/Call-arm64CPP.o'.
> ```
> 
> should we disable arm64 again?

yee, i guess. hoping somebody else could help with arm64.

lazka:
> yee, i guess. hoping somebody else could help with arm64.

OK, thanks, I'll disable it later

hmartinez82:
Oh wow, sorry 😅. That ASM file is way out of my league. It's not like I know ARM64 assembly anyway.

PR #21007, opened 2024-05-28
uv: update to 0.2.4

ognevny:
dropped libz-sys change, as it doesn't matter when libgit2 is linked


ognevny:
I don't know exactly, but maybe rust use something like `--as-needed` link option, so if the linked dll has needed symbols, there is no need to link this dependency directly (openssl and zlib are dependencies of libgit2). but then I don't know whether I should keep zlib-sys change. the size change for that update is very small

PR #21008, opened 2024-05-28
curl: Fix static library

### LINKED ISSUE #21006: -DNGHTTP3_STATICLIB CFLAGS option has to be specified when building static curl to link against libnghttp3.a

I tried compiling my library using `msys2/setup-msys2@v2` in GitHub Actions with `CURL_STATICLIB` enabled. This issue is similar to  #5440

### Expected behavior

It was supposed to compile successfully, like the previous `mingw-w64-x86_64-curl` build.

### Actual behavior

undefined reference to `__imp_nghttp3_*'

### Verification

### Windows Version

MINGW64_NT-10.0-20348

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

_No response_

MehdiChinoune:
Fix #21006


MehdiChinoune:
@sammycage Could you please download the artifacts and test if they fix your issue.

raedrizqie:
i think those flags should also be added to `Cflags.private` in libcurl.pc

MehdiChinoune:
> i think those flags should also be added to `Cflags.private` in libcurl.pc

They should be added for libnghttp2.pc and libnghttp3.pc

sammycage:
> @sammycage Could you please download the artifacts and test if they fix your issue.

@MehdiChinoune Thank you for providing the artifacts. Unfortunately, I am unable to use them directly. I can only incorporate them through the `msys2/setup-msys2` in our GitHub Actions workflow. 

For your reference, the GitHub Action is still encountering the same errors, even after clearing the cache.

Biswa96:
It is possible to verify shared library imports in static library. The output of `nm /path/to/libfoo.a | grep __imp | sort | uniq` command should have Windows APIs **only**.

MehdiChinoune:
> @MehdiChinoune Thank you for providing the artifacts. Unfortunately, I am unable to use them directly.

If you want the issue to be fixed, then you should test the fix. The package is ready just download it and install it.

> I can only incorporate them through the `msys2/setup-msys2` in our GitHub Actions workflow.

This is not possible until the PR got merged, the package got built and deplyed.

> For your reference, the GitHub Action is still encountering the same errors, even after clearing the cache.

Of course.

sammycage:
> If you want the issue to be fixed, then you should test the fix. The package is ready just download it and install it.

I have tested it, and it works as expected. Thanks for the quick fix!

PR #21013, opened 2024-05-29
curl: build with cmake

MehdiChinoune:


MehdiChinoune:
> CMAKE_UNITY_BUILD should also be enabled, which is supported upstream.

We don't use that mode in packaging.

lazka:
lgtm

lazka:
I guess CMAKE_UNITY_BUILD would give us faster builds? I don't have much experience with it.

MehdiChinoune:
> I guess CMAKE_UNITY_BUILD would give us faster builds? I don't have much experience with it.

Neither me.

lazka:
I just tried locally, and there is not much difference all in all, both take around 3m20s.

But sure, why not I guess.

lazka:
It's faster in CI at least, likely because of the fewer cores.

lazka:
Turns out relocation is broken somehow:

```
$ curl https://www.msys2.org
curl: (77) error setting certificate file: C:/msys64/:/a/msys64/clang64/etc/ssl/certs/ca-bundle.crt
```

vszakats:
Proposing this patch: https://github.com/curl/curl/pull/13898, that would allow dropping one of the custom patches. Let me know what you think under the PR.

PR #21025, opened 2024-05-30
[new package] cminpack

luau-project:
## Description

According to the description given by the author Frédéric Devernay on [https://github.com/devernay/cminpack](https://github.com/devernay/cminpack), **cminpack** is a C/C++ rewrite of the MINPACK software (originally in FORTRAN) for solving nonlinear equations and nonlinear least squares problems.

## Positive points

* cminpack is a mature library, which has been around for almost 20 years;
* reentrant code;
* API is more user friendly than that one found in the original Fortran minpack;
* Eigen (a scientific library) mentions to have used cminpack in their implementation of nonlinear optimization [https://eigen.tuxfamily.org/dox/unsupported/index.html](https://eigen.tuxfamily.org/dox/unsupported/index.html), [https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/NonLinearOptimization?ref_type=heads#L45](https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/NonLinearOptimization?ref_type=heads#L45);
* Available on a large number of repositories (Debian, Ubuntu, Fedora, Arch) for almost a decade [https://repology.org/project/cminpack/versions](https://repology.org/project/cminpack/versions).

## Website

[http://devernay.free.fr/hacks/cminpack/](http://devernay.free.fr/hacks/cminpack/)

### DIFF
--- a/mingw-w64-cminpack/PKGBUILD
+++ b/mingw-w64-cminpack/PKGBUILD
@@ -1,0 +1,82 @@
+# Maintainer: luau-project <luau.project@gmail.com>
+
+_realname=cminpack
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=1.3.9
+pkgrel=1
+pkgdesc="A C/C++ rewrite of the MINPACK software for solving nonlinear equations and nonlinear least squares problems (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='http://devernay.free.fr/hacks/cminpack'
+license=('spdx:Minpack')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
+source=("https://github.com/devernay/cminpack/archive/v1.3.9.tar.gz")
+sha256sums=("aa37bac5b5caaa4f5805ea5c4240e3834c993672f6dab0b17190ee645e251c9f")
+
+prepare() {
+  # on Windows, EOL must be CRLF (\r\n)
+  # for reference test files,
+  # but the tarball is packed with LF (\n).
+  # So, we replace LF with CRLF.
+  for ref in ./${_realname}-${pkgver}/examples/ref/*.ref;
+  do
+    mv $ref $ref.bak;
+    unix2dos -n $ref.bak $ref;
+  done
+}
+
+build() {
+  declare -a extra_config
+  if check_option "debug" "n"; then
+    extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  # Configure the static version
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  "${MINGW_PREFIX}"/bin/cmake.exe \
+    -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+    "${extra_config[@]}" \
+    -DUSE_BLAS=OFF \
+    -DBUILD_SHARED_LIBS=OFF \
+    -S "${_realname}-${pkgver}" \
+    -B build-${MSYSTEM}-static
+
+  "${MINGW_PREFIX}"/bin/cmake.exe --build build-${MSYSTEM}-static
+
+  # Configure the shared version
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  "${MINGW_PREFIX}"/bin/cmake.exe \
+    -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+    "${extra_config[@]}" \
+    -DUSE_BLAS=OFF \
+    -DBUILD_SHARED_LIBS=ON \
+    -S "${_realname}-${pkgver}" \
+    -B build-${MSYSTEM}-shared
+
+  "${MINGW_PREFIX}"/bin/cmake.exe --build build-${MSYSTEM}-shared
+}
+
+check() {
+  # test the static version
+  "${MINGW_PREFIX}"/bin/cmake.exe --build build-${MSYSTEM}-static --target test
+
+  # test the shared version
+  "${MINGW_PREFIX}"/bin/cmake.exe --build build-${MSYSTEM}-shared --target test
+}
+
+package() {
+  # install the static version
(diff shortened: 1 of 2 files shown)


mingw-w64-cminpack/PKGBUILD
```diff
@@ -0,0 +1,82 @@
...
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
+source=("https://github.com/devernay/cminpack/archive/v1.3.9.tar.gz")
+sha256sums=("aa37bac5b5caaa4f5805ea5c4240e3834c993672f6dab0b17190ee645e251c9f")
+
+prepare() {
+  # on Windows, EOL must be CRLF (\r\n)
+  # for reference test files,
+  # but the tarball is packed with LF (\n).
+  # So, we replace LF with CRLF.
```
lazka, line 23: Is this really needed? Most things can deal with LF nicely nowadays.

mingw-w64-cminpack/PKGBUILD
```diff
@@ -0,0 +1,82 @@
...
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  # Configure the static version
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+  "${MINGW_PREFIX}"/bin/cmake.exe \
+    -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+    "${extra_config[@]}" \
+    -DUSE_BLAS=OFF \
```
lazka, line 45: Any particular reason BLAS isn't used?

lazka:
The static libs have  "_s" suffix, which likely means they won't work when used with pkg-config at least. Not sure what's this about and how much of an issue that is.

mingw-w64-cminpack/PKGBUILD
luau-project: Unfortunately, in the current CMake build present on the released tarball, it is.

However, due your valuable input, I already solved this puzzle on a fork and I am submitting a PR on cminpack's repository.

mingw-w64-cminpack/PKGBUILD
luau-project: If you look at Ubuntu's page for cminpack [https://packages.ubuntu.com/source/noble/cminpack](https://packages.ubuntu.com/source/noble/cminpack), there is no dependency on BLAS. So, I choose to disable it as well.

luau-project:
@lazka 

> The static libs have "_s" suffix, which likely means they won't work when used with pkg-config at least. Not sure what's this about and how much of an issue that is.

You have given really helpful tips. I confirmed that the current files generated for pkg-config on CMake build are not as they would be. So, looks like you've found an issue on cminpack build with CMake. Locally, I addressed this issue as well and I am submitting another PR on cminpack's repository.

Once I get a response about all these PRs, I'll get back and drop updates here. Thank you.

lazka:
You can also "backport" patches, as long as you link to the upstream PR. But waiting is also an option of course.

luau-project:
## Updates

Hello, @lazka 

I'll comment the changes I have made to the cminpack insertion:
1. Added a patch file (```001-fix-cmake-eol-for-tests.patch```) to avoid EOL problems, which was submitted and approved (but not yet merged) as upstream PR [https://github.com/devernay/cminpack/pull/63](https://github.com/devernay/cminpack/pull/63);
2. Added another patch file (```002-fix-pkg-config.patch```) to fix some pkg-config related issues that you noted on your previous post. This patch was also submitted as PR in the upstream repo, but it was not analyzed yet [https://github.com/devernay/cminpack/pull/64](https://github.com/devernay/cminpack/pull/64);
3. Removed the static library version. **Reason**: At the current state with these 2 patches, pkg-config files are fixed, but the resulting CMake files for both shared and static libraries would not be ok (static would be broken). In order to fix CMake files for the static library version, it is required a major rework of their CMake build. In the future, I plan to help them in this area, but it is a long journey. In short, for now, static library is disabled;
4. Followed the same procedure on Debian distros, which builds and installs only the double-precision version of the library (```-DCMINPACK_PRECISION=d``` on CMake configuring step).

lazka:
> Followed the same procedure on Debian distros, which builds and installs only the double-precision version of the library (`-DCMINPACK_PRECISION=d` on CMake configuring step).

I don't see anything like that here: https://salsa.debian.org/science-team/cminpack/-/blob/master/debian/rules

luau-project:
## Reply

> > Followed the same procedure on Debian distros, which builds and installs only the double-precision version of the library (`-DCMINPACK_PRECISION=d` on CMake configuring step).
> 
> I don't see anything like that here: https://salsa.debian.org/science-team/cminpack/-/blob/master/debian/rules

This ```CMINPACK_PRECISION``` argument was introduced on version 1.3.8, which you can compare analyzing the commits of the released tarballs [1.3.7](https://github.com/devernay/cminpack/blob/5af798d3c4fcd58af3cc580393f4dc949c5e539b/CMakeLists.txt) and [1.3.8](https://github.com/devernay/cminpack/blob/cb7c3f6433ccea7eef58ff57b3a9a4c2563eb375/CMakeLists.txt#L32). On the other hand, Debian repositories uses 1.3.6, which back in the time, it was building only the double precision version (can be confirmed by the [installed files on Debian](https://packages.debian.org/bookworm/amd64/libcminpack1/filelist) judging by the absence of suffix in the library name, discussed below).

## A bit of context

In the current version [1.3.9](https://github.com/devernay/cminpack/blob/master/CMakeLists.txt#L32), the argument ```CMINPACK_PRECISION``` has 4 possible values (```s```, ```d```, ```ld```, ```all```)

* ```s```: builds the single precision version (adds ```s``` suffix to the library name, e.g.: ```libcminpacks.dll```);
* ```d```: builds the double precision version (CMake tests were implemented targeting this precision. This option doesn't add suffixes to the library name, e.g.: ```libcminpack.dll```);
* ```ld```: builds the long-double precision version (adds ```ld``` suffix to the library name, e.g.: ```libcminpackld.dll```);
* ```all```: builds all the previous precisions (default nowadays when ```CMINPACK_PRECISION``` is not provided).

### Motivation to use only the double precision version

The main motivation to use only the double precision version of cminpack is due the fact that CMake tests were designed to run for that precision, and that precision is the version which is around for a long time. In order to avoid shipping a version "untested" by PKGBUILD check method, I sticked to that precision.

On the other hand, if you want, I can remove that ```CMINPACK_PRECISION``` argument and we take a route like [Arch Linux](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cminpack) does. Since they don't apply any ```CMINPACK_PRECISION``` argument in the configure step, they are using ```CMINPACK_PRECISION=all```, building and delivering all the precisions.

## Conclusion

Currently on version 1.3.9:
* CMake test do test the binaries built by double-precision version;
* CMake tests do **NOT** test the binaries built by single or long-double precision versions.

PR #21047, opened 2024-06-01
Revert "imagemagick: Use native Windows thread APIs"

### LINKED ISSUE #20983: Re-enable multi-threading for mingw-w64-ucrt-x86_64-imagemagick and variants

Hello,

This commit - [mingw-w64-ucrt-x86_64-imagemagick](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-imagemagick?repo=ucrt64) added the "--without-threads" line to the ImageMagick configure section of PKGBUILD. This means that the packages (I have tested the normal one, ucrt64 and clang64) are only single threaded. I removed the line and rebuilt all three packages locally and multi-threaded support is restored. I do not need all three packages as the ucrt one is the most suitable. The commit mentions that adding this line also fixed a clang compiler error but I tested building using clang64 with and without the line and I could not see any difference but I must admit I am no expert.

If this line does need to be there, then any guidance as to how to enable native Windows threads for mingw-w64-ucrt-x86_64-imagemagick would be most welcome. I have spent a few days trying to make it work but without any success.

Many thanks for any pointers or advice!

### Expected behavior

That the package would use all CPUs to process images.

### Actual behavior

The package only uses a single CPU to process images.

### Verification

### Windows Version

MSYS_NT-10.0-22631

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

No

Biswa96:
This reverts commit 0dff9a633d6dd0cb9f117fce7c9bde311e9a39c7
    
    This was disabled due to compiler error with clang which was fixed in upstream.
    https://github.com/ImageMagick/ImageMagick/commit/2fbf9383c574d08327f7e41db50d613003857604
    https://github.com/ImageMagick/ImageMagick/commit/d32677a65e630fa4a36bd1994498234ca280cae2

Fixes #20983

### DIFF
--- a/mingw-w64-imagemagick/PKGBUILD
+++ b/mingw-w64-imagemagick/PKGBUILD
@@ -7,7 +7,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
 _basever=7.1.1
 _rc=-33
 pkgver=${_basever}${_rc//-/.} # pkgver doesn't have "," "/", "-" and space.
-pkgrel=1
+pkgrel=2
 pkgdesc="An image viewing/manipulation program (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


ognevny (commented):
remove a comment before `configure`

PR #21064, opened 2024-06-04
uv: update to 0.2.6

ognevny:


ognevny:
uv no longer links libssh2 and libgit2

PR #21076, opened 2024-06-05
Groonga: Update to 14.0.4

otegami:
Groonga version14.0.4 has just released [here](https://github.com/groonga/groonga/releases/tag/v14.0.4).


otegami:
I raised issue https://github.com/msys2/MINGW-packages/issues/21077 about the CI fails.

PR #21079, opened 2024-06-05
mingw-w64-directxtk - update to latest release.

JPeterMugaas:


Biswa96:
Would it be possible to add mingw gcc-libs in depends array? CI error https://github.com/msys2/MINGW-packages/actions/runs/9380362017/job/25834058316?pr=21079#step:11:488

lazka:
Failed on arm64: `clang++: error: unsupported option '-msse2' for target 'aarch64-w64-windows-gnu'`

PR #21081, opened 2024-06-05
tiledb: cleanup and migrate from legacy dependency management mode

teo-tsirpanis:
This PR updates the `tiledb` package to obviate the need for `001-do-not-use-vcpkg.patch`, and updates the build command to build only the necessary targets.

See each commit message for more details, and also macports/macports-ports#24239 and https://github.com/TileDB-Inc/TileDB/issues/4960#issuecomment-2118864650 for more background.

Validated locally by running `makepkg-mingw --noconfirm --syncdeps`.


mingw-w64-tiledb/003-disable-lz4-cmake.patch
teo-tsirpanis, line 1: This patch is also applied in the MacPorts and Conda distributions of TileDB. It will not be required with TileDB-Inc/TileDB#5021.

mingw-w64-tiledb/PKGBUILD
```diff
@@ -62,24 +62,22 @@ build() {
     CXXFLAGS+=" -Wno-c++11-narrowing"
   fi
 
-  TILEDB_DISABLE_AUTO_VCPKG=0 \
+  TILEDB_DISABLE_AUTO_VCPKG=1 \
```
teo-tsirpanis, line 65: Either 0 or 1 would work (TileDB checks the existence of the variable), but I changed it to 1 for clarity.

MehdiChinoune:
The real solution is to keep a CMake option to disable VCPKG, This is the first time I see someone using an environment variable to enable/disable something.

teo-tsirpanis:
The vcpkg auto-download mechanism [was based on the Azure SDK for C++](https://github.com/Azure/azure-sdk-for-cpp/blob/459c148d06c6c04645f5f8093245c1c285c3f7a4/cmake-modules/AzureVcpkg.cmake#L17), which uses an environment variable for reasons I don't know.

I opened TileDB-Inc/TileDB#5048 to add a CMake option in addition to the environment variable. I would suggest you to merge this PR either way.

MehdiChinoune:
Isn't it weird to see the upstream patching the downstream for their own issue.
Why not applying the patch on tiledb?

teo-tsirpanis:
I don't understand. The MinGW package builds fine right now without any patches. I opened the upstream PR because I found the feature request reasonable and useful by itself, but that upstream PR should not block this PR.

MehdiChinoune:
> I don't understand. The MinGW package builds fine right now without any patches. I opened the upstream PR because I found the feature request reasonable and useful by itself, but that upstream PR should not block this PR.

You are working on TileDB and you want to patch Findlz4 for MinGW-packages. Why don't you fix it in TileDB first.

mingw-w64-tiledb/PKGBUILD
MehdiChinoune: DTILEDB_VCPKG=OFF is better.

MehdiChinoune:
This PR doesn't fix anything.

teo-tsirpanis:
Indeed it does not fix anything now, but will make updating easier after a couple releases. If you prefer it, you can defer these changes and make them when 2.26 gets released.

MehdiChinoune:
Let's wait for the next release and see if they come with a friendly way to package tiledb (like any other library Instead of forcing vcpkg on all platforms)

PR #21082, opened 2024-06-05
octave: update to 9.2.0

mmuetzel:
Build with Qt6 again on environments where it is available.

Drop cherry-picked patches that are part of the new release.


mmuetzel:
The build artifacts from this PR seem to be working.

PR #21084, opened 2024-06-05
zed: update to 0.138.4

ognevny:


ognevny:
I've updated to 0.138.4 on my Arch machine and now I get random crashes. not sure about Windows

PR #21098, opened 2024-06-07
curl: Enable nghttp3 for 32 bit

Biswa96:
This fixes static linking with curl in 32 bit. It is not feasible to maintain different pkgconfig patch for 32 bit.


Biswa96:
appstream-glib fails to build for 32 bit with the following error.

```
  ../appstream-glib-appstream_glib_0_8_3/meson.build:77:10: ERROR: Dependency lookup for libcurl with method 'pkgconfig' failed: Could not generate cflags for libcurl:
  Package libnghttp3 was not found in the pkg-config search path.
  Perhaps you should add the directory containing `libnghttp3.pc'
  to the PKG_CONFIG_PATH environment variable
  Package 'libnghttp3', required by 'libcurl', not found
```

Biswa96:
There is another issue. nghttp2 and nghttp3 should not be in gnutls and winssl versions of curl. So, the pkgconfig patch should be added in cmake file.

PR #21099, opened 2024-06-07
curl: Fix Requires.private field pkgconfig file

Biswa96:
This adds the reference in Requires.private when the dependencies are found. Hence, this fixes pkgconfig file in gnutls and winssl versions.


lazka:
lgtm

Issue #21028, opened 2024-05-30
Linking curl broken since roughly 2 days ago


ReenigneArcher:
### Description / Steps to reproduce the issue

Curl is failing to statically link since ~2 days ago.

1. Install `mingw-w64-ucrt-x86_64-curl`
2. Find curl like: https://github.com/LizardByte/Sunshine/blob/f0a00ae35683dfd5742952fb4cf99edf268b57fd/cmake/dependencies/common.cmake#L16
3. Statically link curl like: https://github.com/LizardByte/Sunshine/blob/f0a00ae35683dfd5742952fb4cf99edf268b57fd/cmake/compile_definitions/windows.cmake#L15-L18
4. Include curl like: https://github.com/LizardByte/Sunshine/blob/f0a00ae35683dfd5742952fb4cf99edf268b57fd/cmake/compile_definitions/common.cmake#L140
5. Final link: https://github.com/LizardByte/Sunshine/blob/f0a00ae35683dfd5742952fb4cf99edf268b57fd/cmake/targets/common.cmake#L25

### Expected behavior

Curl properly links as it did previously.

### Actual behavior

The initial error (~2 days ago) was many `undefined references` (https://github.com/LizardByte/Sunshine/actions/runs/9278791852/job/25530652407?pr=2604#step:7:501)

```txt
[ 81%] Linking CXX executable sunshine.exe
D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.1.0/../../../../lib\libcurl.a(libcurl_la-curl_osslq.o):(.text+0x430): undefined reference to `__imp_nghttp3_conn_read_stream'
D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.1.0/../../../../lib\libcurl.a(libcurl_la-curl_osslq.o):(.text+0x5a2): undefined reference to `__imp_nghttp3_strerror'
[...]

PR #21100, opened 2024-06-07
curl: exclude pathtools.c from unity mode

vszakats:
`pathtools.c` contains code that is also present in other static
MSYS2/mingw-w64-built libraries. To allow the linker to de-duplicate
this common object and avoid "multiple definition" linker errors, the
object must be kept separate from the rest of the library, that is
compiled into a single, "unity" object after
ee2184d939d4a99ab017b0f61222b0555d6b1467 #21021.

Ref: #21028


lazka:
Thanks!

PR #21101, opened 2024-06-08
rust: add split package for source code

ognevny:
it's not required for a compiler, but for rust-analyzer. the same thing is already implemented in arch

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -17,9 +17,10 @@ embed_manifest_url=https://gitlab.com/careyevans/embed-manifest/-/archive/v${emb
 _realname=rust
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
-         $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-rust-docs"))
+         $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-rust-docs")
+         "${MINGW_PACKAGE_PREFIX}-rust-src")
 pkgver=1.78.0
-pkgrel=1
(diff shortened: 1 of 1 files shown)


lazka:
@filnet thoughts?

mingw-w64-rust/PKGBUILD
```diff
@@ -256,6 +258,8 @@ package_rust() {
   mv "${pkgdir}${MINGW_PREFIX}/etc/bash_completion.d/cargo" \
      "${pkgdir}${MINGW_PREFIX}/share/bash-completion/completions/cargo"
 
+  install -d "${srcdir}/${MSYSTEM}/dest-src${MINGW_PREFIX}/lib/rustlib/src"
```
filnet, line 261: Why is it not done in `package_rust-src` like `package_rust-docs` does ?

PS: I am far from being an expert in PKGBUILD.

filnet:
> @filnet thoughts?

I am fine with the proposed change but did ask a question.

mingw-w64-rust/PKGBUILD
ognevny: the files should be moved before the function ends. otherwise, the whole src will be packaged as it's already installed in $pkgdir

filnet:
LGTM

mingw-w64-rust/PKGBUILD
jeremyd2019: rust-docs moves the files away in build (circa line 224 I think).  That seems to be the way it's usually done.  It may not be a good idea to rely on this, but the package functions are called in the order the packages are listed in the pkgname array, so theoretically you could list rust-src and rust-docs before rust in there, and move the files out in their package functions.  I don't think I've seen any other package do it this way though.

ognevny: you can refactor it while updating package

PR #21106, opened 2024-06-08
[new package] ffmpegthumbnailer 2.2.2

ognevny:


Biswa96:
Does this require in Windows?

ognevny:
> Does this require in Windows?

I want to add one package that use this as optional dependency. I can close if it's not actually needed

ognevny:
> I want to add one package that uses this as optional dependency. I can close if it's not actually needed

I didn't find any OS-specific conditionals for that package: https://github.com/search?q=repo%3Asxyazi%2Fyazi+ffmpegthumbnailer&type=code

ognevny:
I don't understand CI failure

Biswa96:
> I want to add one package that uses this as optional dependency. I can close if it's not actually needed

That's probably in Linux, right? I have used this program in Linux with thunar file manager in XFCE to show media file thumbnail (mkv, wmv, avi etc.). I do not know if this is required in Windows.

ognevny:
ok, will close for now

Biswa96:
> I don't understand CI failure

The patch command failed with "different line endings" error because the line endings do not match with actual file. e.g. the file has LF but the patch has CRLF or vice-versa. dos2unix or unix2dos programs can help to convert line endings.

ognevny:
> > I don't understand CI failure
> 
> The patch command failed with "different line endings" error because the line endings do not match with actual file. e.g. the file has LF but the patch has CRLF or vice-versa. dos2unix or unix2dos programs can help to convert line endings.

forgot to say that the patch applied in my Arch machine. maybe there is different behaviour on Windows?

Biswa96:
This also affect the outcome.

https://github.com/msys2/MINGW-packages/blob/ba76417a09f9aad220047d7d434cb25c69ebf076/.gitattributes#L2

PR #21112, opened 2024-06-09
SQLite3MC: Update to 1.8.5

taozuhong:


mingw-w64-sqlite3mc/PKGBUILD
```diff
@@ -34,7 +34,8 @@ build() {
     --host=${MINGW_CHOST} \
     --target="${MINGW_CHOST}" \
     --enable-static \
-    --enable-shared
+    --enable-shared \
+    SQLITE_USER_AUTHENTICATION=1
```
ognevny, line 38: what this option do?

taozuhong: SQLite3MC 1.8.4 disable user authentication extension by default, for compatibility, add this switch to enable it.

ognevny:
diffs don't show that something disappeare

PR #21129, opened 2024-06-11
libc++: linked with compiler-rt only

raedrizqie:
- merge libunwind into libc++ (as per current clang package)
- remove dependency of libgcc_s
- all libraries are standalone now


MehdiChinoune:
Why?

raedrizqie:
why not?

PR #20651, opened 2024-04-16
Rust test

filnet:

PR #21141, opened 2024-06-12
rust: update to 1.79.0

filnet:


ognevny:
I've rerun jobs now, server back online

jeremyd2019 (changes requested):
Should we consider #20651 at this point (minus the ci change) rather than this straight update?  I'd like to figure out which PR to make suggestions on 😉

filnet:
I want to keep the rust test branch open for testing (i.e. with the ci changes).

So the question is what do we want to move here ? Everything ?

jeremyd2019:
I think we would want these patches at least:
>        "https://github.com/rust-lang/rust/commit/eac0b3a1d1b8d1db8fa95c50e118de5b9471e5af.patch"
>        "https://github.com/rust-lang/backtrace-rs/commit/adc9f5ca6205a7bde95b5f1a3f029e590ce6c694.patch"
>        "https://github.com/rust-lang/backtrace-rs/commit/e15130618237eb3e2d4b622549f9647b4c1d9ca3.patch"

Also it would be a good opportunity to move the moving of the src files into `build` instead of `package` function like the doc files.

Otherwise, the other PR only changes the check function and checkdepends.  Those should be safe enough.

ognevny:
> I think we would want these patches at least:
> >        "https://github.com/rust-lang/rust/commit/eac0b3a1d1b8d1db8fa95c50e118de5b9471e5af.patch"
> >        "https://github.com/rust-lang/backtrace-rs/commit/adc9f5ca6205a7bde95b5f1a3f029e590ce6c694.patch"
> >        "https://github.com/rust-lang/backtrace-rs/commit/e15130618237eb3e2d4b622549f9647b4c1d9ca3.patch"
> 
> Also it would be a good opportunity to move the moving of the src files into `build` instead of `package` function like the doc files.
> 
> Otherwise, the other PR only changes the check function and checkdepends.  Those should be safe enough.

ok. feel free to do it :)

filnet:
I'll do the suggested changes.

mingw-w64-rust/PKGBUILD
```diff
@@ -8,8 +8,8 @@ if [[ $MINGW_PACKAGE_PREFIX == *-clang-aarch64 ]]; then
   _bootstrapping=no
 fi
 
-rust_dist_server=https://static.rust-lang.org/dist
-#rust_dist_server=https://dev-static.rust-lang.org/dist/2024-03-17
+#rust_dist_server=https://static.rust-lang.org/dist
+rust_dist_server=https://dev-static.rust-lang.org/dist/2024-06-10
```
jeremyd2019, line 12: Should we wait for the files to be on the 'normal' dist location rather than the 'dev' one?

filnet: Yes ! That's why the PR is a draft.

filnet:
@mati865 the release of Rust includes a contribution if yours : https://github.com/rust-lang/rust/pull/121712

mingw-w64-rust/PKGBUILD
jeremyd2019: I converted it back to a draft now

ognevny:
> I am pretty sure I created the PR as a draft... or wanted to do so...

I thought it's ok to merge, sorry 😁

jeremyd2019:
> I think we would want these patches at least:
> 
> > ```
> >    "https://github.com/rust-lang/rust/commit/eac0b3a1d1b8d1db8fa95c50e118de5b9471e5af.patch"
> >    "https://github.com/rust-lang/backtrace-rs/commit/adc9f5ca6205a7bde95b5f1a3f029e590ce6c694.patch"
> >    "https://github.com/rust-lang/backtrace-rs/commit/e15130618237eb3e2d4b622549f9647b4c1d9ca3.patch"
> > ```

The last two are already present in 1.79.0.  Sorry for the mistake.

jeremyd2019 (approved):
I think that's everything, with the possible exception of including the changes to the check function/checkdepends from #20651, which shouldn't hurt anything since they are never used by any automated process (such as msys2-autobuild)

filnet:
LGTM

jeremyd2019:
> @mati865 this release of Rust includes a contribution of yours : [rust-lang/rust#121712](https://github.com/rust-lang/rust/pull/121712)

Would that make it possible to bootstrap -gnullvm, or does the bootstrap require a native rustc for the target?

mati865:
Maybe it would be possible to hack it to cross compiler but proper bootstrap requires native build.

PR #21142, opened 2024-06-12
qemu: update to 9.0.1

heljkon:


Biswa96:
qemu tarball contains many symlinks. Are those required for normal usage? There is a script to resolve those symlinks.

Also, is there any configure or meson option to disable building tests?

heljkon:
> qemu tarball contains many symlinks. Are those required for normal usage? There is a script to resolve those symlinks.

I have never experienced any runtime issue because of missing symlinks.
On the other hand https://wiki.qemu.org/Hosts/W32 suggests

> On newer versions of Windows 10, unprivileged accounts can create symlinks if Developer Mode is enabled. When Developer Mode is not available/enabled, the SeCreateSymbolicLinkPrivilege privilege is required, or the process must be run as an administrator. 

The only WA needed is integrated in PKGBUILD
```
noextract=(${_tarname}.tar.xz)

prepare() {
  [[ -d "${srcdir}"/${_tarname} ]] && rm -rf "${srcdir}"/${_tarname}
  # tar cannot create links here, therefore hide the failure
  tar -xf "${srcdir}"/${_tarname}.tar.xz -C "${srcdir}" || true
}
```
Could you elaborate "There is a script", please?

> Also, is there any configure or meson option to disable building tests?

I've searched multiple times but always failed in finding the option which would save me hours when performing git bisect.

Biswa96:
> Could you elaborate "There is a script", please?

See the mingw breeze package as an example. The python script replaces symlinks with the target file.

PR #21153, opened 2024-06-13
[new package] makelink 1.0

moskensoap:
# Motivation
In the MSYS2 environment, users often copy and rename `mingw32-make.exe` to `make.exe` to maintain a Linux-like `make` command behavior on Windows. However, this approach has several drawbacks:

1. The renamed `make.exe` cannot be managed by `pacman`, the package manager for MSYS2.
2. Setting aliases or creating alias scripts like `make.cmd` works in shells but is not recognized by tools such as VS Code's Make extension, which requires an actual `make.exe`.

# Solution
To address these issues and maintain the familiar Linux workflow, a good idea is to compile a new `make.exe` that simply forwards all command-line arguments to `mingw32-make.exe` and executes it.

## Comparison with (msys)make.exe

- `(msys)make.exe` outputs Unix-like directory structures, while `mingw32-make.exe` (and the new `make.exe`) outputs Windows local directory structures.
- `(msys)make.exe` can execute Unix-style shell scripts directly, whereas `mingw32-make.exe` might require Windows batch scripts or MinGW-specific tools.

## Potential conflicts

MSYS2 isolates `make.exe` instances in different sub-environments using PATH's priority, preventing conflicts.

# Website
https://github.com/moskensoap/makelink

### DIFF
--- a/mingw-w64-makelink/PKGBUILD
+++ b/mingw-w64-makelink/PKGBUILD
@@ -1,0 +1,33 @@
+# Maintainer: moskensoap <141073988+moskensoap@users.noreply.github.com>
+
+_realname=makelink
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=1.0
+pkgrel=1
+pkgdesc="Creat make.exe linking to mingw32-make.exe (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64')
+url="https://github.com/moskensoap/makelink"
+
+license=('MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             "${MINGW_PACKAGE_PREFIX}-make")
+depends=("${MINGW_PACKAGE_PREFIX}-make")
+source=("https://github.com/moskensoap/makelink/releases/download/${pkgver}/makelink-${pkgver}.tar.gz")
+sha256sums=('d012b80e1c31d7c5e5ee0fd03f2ce294aabf6abe700424e86df49ec7bcdf2b7e')
+
+
+build() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+  mkdir -p "${srcdir}/build-${MSYSTEM}" && cd "${srcdir}/build-${MSYSTEM}"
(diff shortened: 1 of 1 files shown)


moskensoap:
It's a pleasure to find that cc.exe is a common component of both gcc and clang when solving the makefile error.

Biswa96:
> users often copy and rename `mingw32-make.exe` to `make.exe` to maintain a Linux-like `make`

Wouldn't it be possible to shell alias?

moskensoap:
> > users often copy and rename `mingw32-make.exe` to `make.exe` to maintain a Linux-like `make`
> 
> Wouldn't it be possible to shell alias?

Alias is a good way, but there are two limitations:

1.If you add /ucrt64/bin to your PATH and use MSYS2 commands in CMD, CMD does not support alias settings.
2.If you have two packages installed, `msys/make` and `ucrt64/mingw-w64-ucrt-x86_64-make`, there will be two executables, `/usr/bin/make.exe` and `/ucrt64/bin/mingw32-make.exe`. Once you set the alias of `make` to `mingw32-make` in .bashrc, you can type `make` to run `mingw32-make` in the UCRT64 sub-environment. However, in the MSYS sub-environment, you cannot run `/usr/bin/make.exe` itself anymore.

Compiling a new make.exe that points to mingw32-make.exe is an optional global solution to these problems.

Biswa96:
> you can type `make` to run `mingw32-make` in the UCRT64 sub-environment.

And what is the wrong to type `mingw32-make` to run `mingw32-make` in UCRT64 environment?

moskensoap:
> > you can type `make` to run `mingw32-make` in the UCRT64 sub-environment.
> 
> And what is the wrong to type `mingw32-make` to run `mingw32-make` in UCRT64 environment?

There is nothing wrong with typing `mingw32-make`. To honor the pioneers of the mingw32 project, I should type `mingw32-make` whenever possible. However, many software packages are configured to use `make.exe` by default. I think changing it to `make` might be more convenient. This configuration has been my personal setup for a long time, and I just want to share it so that users with similar needs can easily find a solution. If it doesn't align with the aesthetics of the MSYS2 project, I am more than willing to withdraw this pull request.

Biswa96:
Yes, this is not required. Users can run msys2 make from UCRT64 environment and this will overwrite that use case.

PR #21176, opened 2024-06-15
mingw-w64-python-cssselect2 - new package

JPeterMugaas:
mingw-python-svglib - new package


Biswa96:
Please use the source link as shown in PKGBUILD.python.sh template. In both cases, the prepare function are not required and can be removed.

Biswa96:
Added in d4f493df08e248d25d03fcf8e14c99e075758167 and f15203578df0dd22b7850db4e08011f87f8c0ac9

PR #21206, opened 2024-06-20
gstreamer: Update to 1.24.5

lazka:


lazka:
something in staging is broken, will look later

PR #21211, opened 2024-06-21
sundials: update to 7.1.0

mmuetzel:
Remove patches for changes that have been applied upstream.


mmuetzel:
I downloaded and installed the MINGW64 build artifact from this PR. Octave's test suite still passes with it.

PR #21216, opened 2024-06-22
mingw-w64-mpv-git

shoober420:
Git version of the mingw-w64-mpv package.

Can be added to mpv.io download page below stable PKGBUILD.
https://mpv.io/installation/


Biswa96:
Most packages are provided only with released tag. There is mingw-w64-mpv package available.

shoober420:
mpv does recommend to use git master builds, according to the mpv.io website. Please consider adding the git version.

I made an updated PKGBUILD here.
https://github.com/shoober420/windows-configs/blob/main/C%3A/msys2/home/shoober420/mingw-w64-mpv-git/PKGBUILD

ognevny:
> mpv does recommend to use git master builds, according to the mpv.io website

I can't find a such recommendation

shoober420:
At the installation page https://mpv.io/installation/, it states under "Installation" at the top of the page: Using git master is recommended.

PR #21217, opened 2024-06-22
mold: update to 2.32.0

ognevny:
and backport fix for lto support: https://github.com/rui314/mold/pull/1287


ognevny:
I guess that should be fine

PR #21225, opened 2024-06-22
[TEST] build mold with itself and lto on

ognevny:
like arch does: https://gitlab.archlinux.org/archlinux/packaging/packages/mold/-/blob/main/PKGBUILD?ref_type=heads#L27

### DIFF
--- a/mingw-w64-mold/PKGBUILD
+++ b/mingw-w64-mold/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mold
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.32.0
-pkgrel=1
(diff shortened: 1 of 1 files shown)


mmuetzel:
Afaict, mold doesn't support PE files. So, it won't work to link Windows binaries with it.
Iiuc, it can only be used in MSYS2 when cross-building for Linux (or other targets that use ELF binaries).

ognevny:
> Afaict, mold doesn't support PE files. So, it won't work to link Windows binaries with it. Iiuc, it can only be used in MSYS2 when cross-building for Linux (or other targets that use ELF binaries).

I see, I opened an issue for it: https://github.com/rui314/mold/issues/1288

mmuetzel:
Since upstream has no intention to implement support for PE/COFF in mold, this will probably never work.

PR #21230, opened 2024-06-23
opencv: ffmpeg 7 test build

lazka:
I can't get it to work locally, so try in CI


lazka:
indeed, in CI it finds ffmpeg, weird.

PR #21249, opened 2024-06-26
mingw-w64-efxc2 - Update to latest release.

JPeterMugaas:


Biswa96:
Please use better commit messages, for example `efxc2: update to 0.0.13.252`

PR #21254, opened 2024-06-26
zed: update to 0.141.3

ognevny:
for a list of breaking changes see https://github.com/zed-industries/zed/releases/tag/v0.141.2


ognevny:
libgit2 1.8.1 is required for it. I guess it can be now updated with package rebuilds

ognevny:
I've decided to push it with vendored libgit2 at the moment

PR #21260, opened 2024-06-27
nodejs: update to 22.3.0

raedrizqie:


lazka:
That OS=msvc and clang-cl stuff is a bit confusing, but I assume it's not easy otherwise.

Thanks!

raedrizqie:
yeah.. its not like we are gonna upstream these changes anyway :)

PR #21264, opened 2024-06-27
kokkos: add new package

3rav:


Biswa96:
Is this package required for that trilinos issue?

MehdiChinoune:
I don't think this package is needed on windows.
kokkos was made as an abstract to work with different parallel APIs (CUDA, OpenMP, OpenACC...) for GPUs/Accelerators.
Giving that those none of those APIs is supported on Windows (No CUDA, No OpenMP Offload, No OpenACC on GPU). It doesn't make sense to add this package.

3rav:
> Is this package required for that trilinos issue?

I think yes (also kokkos-kernels), but: https://github.com/trilinos/Trilinos/issues/13177

Kreijstal:
I will test this package

Kreijstal:
this package conflicts with the existing trilinos package, you can't install both at the same time, not sure it is a good idea.

Issue #20892, opened 2024-05-13
[lua] cannot seek to the end of large file


sewbacca:
### Description / Steps to reproduce the issue

This bug prevents calculating the size of large files. It works as expected on the same system using WSL.
Reproduce with any file:
```lua
f = assert(io.open("somefile.txt"))
f:seek("set", 0x200000000)
```

Reproduce with a large file:
```lua
f = assert(io.open("largefile.txt"))
print(f:seek("end"))
```

### Expected behavior

`seek("set", 0x200000000)` should succeed (even if the actual file is smaller).
and `seek("end")` should return the correct file size.

### Actual behavior

`seek("set", 0x200000000)` fails with `bad argument #2 to 'seek' (not an integer in proper range)`
and `seek("end")` returns `-1` as file size.

It probably has something to with how [l_seeknum](https://github.com/lua/lua/blob/master/liolib.c#L116-L143) is defined at compile time.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [X] CLANGARM64

### Are you willing to submit a PR?

no

luau-project:
Hello @sewbacca . Based on your input, I tried to fix your issue on [https://github.com/msys2/MINGW-packages/pull/21266](https://github.com/msys2/MINGW-packages/pull/21266). Please, download the artifacts generated by the build and tell me whether it works or not.

PR #21266, opened 2024-06-27
fix: file seek for MinGW compilers on Lua 5.4 and Lua 5.3

luau-project:
## Description

Fixes [https://github.com/msys2/MINGW-packages/issues/20892](https://github.com/msys2/MINGW-packages/issues/20892)

> [!NOTE]
> 
> Even though I could, I decided to not patch for Lua 5.1 as well, because Lua 5.1 had a file seek limitation on all platforms, not only the MSYS2 binaries. So, I would be changing Lua 5.1 as a whole, and that is not my intent.

## How to test

1. You need to have any file larger than 4,294,967,295 bytes (32-bits max value). By the way, I used Ubuntu 24.04 iso, which is 6.1 GB.
2. Write the following Lua script to a file (```fseek-test.lua```):
    ```lua
    f = assert(io.open("ubuntu-24.04-desktop-amd64.iso"))
    print(f:seek("end"))
    ```
3. Run the script on current Lua ```lua fseek-test.lua``` or Lua  5.3 ```lua5.3 fseek-test.lua``` 
4. Compare the results
   * Old behavior: would print ```-1```
   * Actual behavior: prints ```6114656256``` for Ubuntu's iso (the file size).


sewbacca:
Successfully tested this PR manually on UCRT64, for both lua 5.3 and 5.4.

mingw-w64-lua/0004-fseek-posix-on-mingw.patch
```diff
@@ -0,0 +1,12 @@
+diff -Naur a/src/liolib.c b/src/liolib.c
+--- a/src/liolib.c	2024-06-13 19:15:09.000000000 -0300
++++ b/src/liolib.c	2024-06-27 14:17:25.906754600 -0300
+@@ -115,7 +115,7 @@
+ 
+ #if !defined(l_fseek)		/* { */
+ 
+-#if defined(LUA_USE_POSIX)	/* { */
++#if defined(__MINGW32__) || defined(LUA_USE_POSIX)	/* { */
```
lazka, line 9: Shouldn't we try to enable this for mingw instead https://github.com/lua/lua/blob/c1dc08e8e8e22af9902a6341b4a9a9a7811954cc/liolib.c#L125-L126 so it uses proper Windows APIs.?

luau-project: Makes sense.

Submitted a new commit with your suggested changes and tested it locally. Worked for me.

PR #21267, opened 2024-06-27
harfbuzz: update to 9.0.0

MehdiChinoune:
- Split package the same as Arch  (utils, icu and cairo)
- Enable chafa support (on non i686)

### DIFF
--- a/mingw-w64-harfbuzz/PKGBUILD
+++ b/mingw-w64-harfbuzz/PKGBUILD
@@ -4,14 +4,18 @@
 _realname=harfbuzz
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-cairo"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-icu"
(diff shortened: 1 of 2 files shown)


mingw-w64-harfbuzz/PKGBUILD
```diff
@@ -60,10 +82,11 @@ build() {
     -Dcpp_std=c++17 \
     -Dgdi=enabled \
     -Dgraphite=enabled \
-    -Dchafa=disabled \
     -Ddirectwrite=enabled \
+    -Dexperimental_api=true \
```
lazka, line 86: Why do we need this?

otherwise lgtm

MehdiChinoune: Just to let people test experimental features.

lazka: From what I understand (looking at some PRs using it) it also enables experimental behavior changes and optimizations, so I'm a bit worried.

khaledhosny: Also, experimental APIs are not stable might change without notice, so updating HarfBuzz might require re-compiling all packages that depend on it.

MehdiChinoune: I thooucht It just adds extra experimental features.

PR #21268, opened 2024-06-28
fastfetch: update to 2.17.0

CarterLi:


Biswa96:
fastfetch got a library, interesting!

CarterLi:
The library is built for calling WinRT apis, which are not available on Windows 7. Therefore on Windows 7 fastfetch still work despite the dll fails to load.

ognevny:
> fastfetch got a library, interesting!

import libraries should be located in ${prefix}/lib, shouldn't it?

Biswa96:
> import libraries should be located in ${prefix}/lib, shouldn't it?

Yeah, missed it 😝

CarterLi:
[Will be fixed in next release](https://github.com/fastfetch-cli/fastfetch/commit/d50ebc86c708060b61fe8c2cc229cfc5d0dc4e22).

PR #21272, opened 2024-06-28
llvm: update to 18.1.8

oscarfv:


oscarfv:
CLANG64 timed out :(

ognevny:
> CLANG64 timed out :(

all-in-one updates will take long to compile anyway

luau-project:
I already built some llvm-related packages on my own with different compilers. The most annoying is LLVM flang, which is memory hungry and takes a lot of time to compile. Checking the logs, it was exactly on flang compilation that timed out your build.

MehdiChinoune:
> I already built some llvm-related packages on my own with different compilers. The most annoying is LLVM flang, which is memory hungry and takes a lot of time to compile. Checking the logs, it was exactly on flang compilation that timed out your build.

It doesn't time-out because of flang. It times out because clang is built with PGO on CLANG* environments.

PR #21280, opened 2024-06-30
libgridxc: No MPI on aarch64

hmartinez82:
Fixes:
image


MehdiChinoune:
sorry, already pushed 68e2fad4f

PR #21290, opened 2024-07-01
texlive-bin: Rebuild with current runtime and compiler

mmuetzel:
This is an attempt to fix some build errors of the texlive-bin package. It is based on the changes from #19921.

It works around the issue with the conversion of path lists (https://github.com/msys2/msys2-runtime/issues/208) by manually converting the respective paths from relative to absolute using `cygpath`.
It also cherry-picks some fixes from upstream and uses the `-Wno-error=incompatible-pointer-types` work-around to avoid some build errors with GCC 14.

It still fails with errors (in an ICU header iiuc). Maybe, someone has hints how to fix that or would like to continue from here.

It fails with errors like the following for me locally:
```
In file included from C:/msys64/mingw64/include/unicode/ubidi.h:26,
                 from ../../../texk/web2c/xetexdir/XeTeXLayoutInterface.cpp:37:
C:/msys64/mingw64/include/unicode/localpointer.h:561:26: error: 'auto' parameter not permitted in this context
  561 | template <typename Type, auto closeFunction>
      |                          ^~~~
C:/msys64/mingw64/include/unicode/localpointer.h:573:76: error: template argument 2 is invalid
  573 |     explicit LocalOpenPointer(std::unique_ptr<Type, decltype(closeFunction)> &&p)
      |                                                                            ^
C:/msys64/mingw64/include/unicode/localpointer.h:583:78: error: template argument 2 is invalid
  583 |     LocalOpenPointer &operator=(std::unique_ptr<Type, decltype(closeFunction)> &&p) {
      |                                                                              ^
C:/msys64/mingw64/include/unicode/localpointer.h:599:59: error: template argument 2 is invalid
  599 |     operator std::unique_ptr<Type, decltype(closeFunction)> () && {
      |                                                           ^
C:/msys64/mingw64/include/unicode/ubidi.h:579:1: note: invalid template non-type parameter
  579 | U_DEFINE_LOCAL_OPEN_POINTER(LocalUBiDiPointer, UBiDi, ubidi_close);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/mingw-w64-texlive-bin/0005-untighten_version_check_zlib.patch
+++ b/mingw-w64-texlive-bin/0005-untighten_version_check_zlib.patch
@@ -1,0 +1,16 @@
+Description: Untighten version check of runtime linked zlib. Fix for #1056183.
+Forwarded: not-done, could be done
+Author: Hilmar Preuße <hille42@web.de>
+Last-Update: 20231119
+
+--- texlive-bin.orig/texk/web2c/luatexdir/luazlib/lzlib.c
++++ texlive-bin/texk/web2c/luatexdir/luazlib/lzlib.c
+@@ -546,7 +546,7 @@
+ 
+     /* make sure header and library version are consistent */
+     const char* version = zlibVersion();
+-    if (strncmp(version, ZLIB_VERSION, 4))
++    if (strncmp(version, ZLIB_VERSION, 2))
+     {
+         lua_pushfstring(L, "zlib library version does not match - header: %s, library: %s", ZLIB_VERSION, version);
+         lua_error(L);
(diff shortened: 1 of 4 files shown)


jhol:
Those errors look like what you would get from g++ if the C++ language standard was not set to C++11 or newer.

mmuetzel:
Aaah. Doesn't ICU require C++17 now?
It's set to C++14 currently. Let me try that...

mmuetzel:
That seems to have done the trick (at least for me locally in the MINGW64 environment).

ReinhardBiegelIntech:
That looks like a lot of effort, considering similar fixes might also be necessary for other packages. Maybe we could give https://github.com/msys2/msys2-runtime/pull/213 another go? @lazka?

PR #21315, opened 2024-07-04
zed: update to 0.142.5

ognevny:


atahrijouti:
@ognevny not sure where to ask outside of here, but does the msys2 version of zed use the same path for the settings.json file? or is it placed somewhere else?

ognevny:
> @ognevny not sure where to ask outside of here, but does the msys2 version of zed use the same path for the settings.json file? or is it placed somewhere else?

it isn't patched. it's located in default path defined for Windows

atahrijouti:
@ognevny Very fast reply! So fast, you didn't give me time to realize I was missing a comma in my settings xD

Thank you

PR #21319, opened 2024-07-04
rtaudio: Update to 6.0.1

EvergreenTheTree:
This technically does skip a minor version (6.0.0), let me know if it should be updated to 6.0.0 first and then 6.0.1


ognevny:
> let me know if it should be updated to 6.0.0 first and then 6.0.1

it doesn't matter, update process deletes old files and installs new after that

ognevny:
please, add a commit for `mlt` rebuild (just bump pkgrel for it)

ognevny:
please, also add `${MINGW_PACKAGE_PREFIX}-gcc-libs` to `depends`

ognevny:
deleted previous message, ntdll somehow didn't found gcc-libs but they are already in depends...

ognevny:
> deleted previous message, ntdll somehow didn't found gcc-libs dlls for librtaudio but they are already in depends...

update: there is a typo, please fix it to `depends` :D

ognevny:
thanks!

PR #21335, opened 2024-07-07
boost: backport fix for fix aliasing violation in boost::container

### LINKED ISSUE #17977: [boost] boost::container::flat_map iterator has different behavior depending on compiler optimization flags with gcc

g++ v13.1.0, boost v1.82
The following example has different behavior depending on compiler optimization flags:
```C++
#include <boost/container/flat_map.hpp>

#include <cstdint>
#include <iostream>
#include <string>

int main()
{
    boost::container::flat_map<std::string, size_t> map{};
    const std::string key = "test";
    const size_t value = 13;
	
    auto [it, succeeded] = map.emplace(key, value);
	
    if (succeeded)
    {
        std::cout << "emplace result: " << it->first << " -> " << it->second << "\n";
        std::cout << "emplace iterator result: " << "it == map.end() is " << std::boolalpha << (it == map.end()) << "\n";
		
        it = map.find(key);
        std::cout << "find result: " << it->first << " -> " << it->second << "\n";
        std::cout << "find iterator result: " << "it == map.end() is " << std::boolalpha << (it == map.end()) << "\n"; 
    }
	
    return 0;
}
```

### Expected behavior

Without optimization flags (-O0, -O1) everything fine.
g++ -O0 -o boost_v1.82_flat_map_it_bug boost_v1.82_flat_map_it_bug.cpp

> $ ./boost_v1.82_flat_map_it_bug
emplace result: test -> 13
emplace iterator result: it == map.end() is false
find result: test -> 13
find iterator result: it == map.end() is false

### Actual behavior

With optimization flags (-O2, -O3) iterator has bad value.
g++ -O2 -o boost_v1.82_flat_map_it_bug boost_v1.82_flat_map_it_bug.cpp

>  ./boost_v1.82_flat_map_it_bug
emplace result: test -> 13
emplace iterator result: it == map.end() is true
find result: test -> 13
find iterator result: it == map.end() is true

### Verification

### Windows Version

MINGW64_NT-10.0-19044

_No response_

lazka:
backport patch taken from gentoo (thanks!)

Fixes #17977

### DIFF
--- a/mingw-w64-boost/PKGBUILD
+++ b/mingw-w64-boost/PKGBUILD
@@ -8,7 +8,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.85.0
 _boostver=${pkgver//./_}
-pkgrel=2
+pkgrel=3
 pkgdesc="Free peer-reviewed portable C++ source libraries (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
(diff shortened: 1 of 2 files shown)


lazka:
Confirmed locally that this fixes the reported issue.

PR #21363, opened 2024-07-10
llvm-15: Add ARM and RISCV targets

thomasfire:
We need it at https://github.com/m-labs/artiq . 

Previously, llvmlite was depending on LLVM 14, which had these targets. Now it retargeted to newer LLVM, which broke both our platforms.

### DIFF
--- a/mingw-w64-llvm-15/PKGBUILD
+++ b/mingw-w64-llvm-15/PKGBUILD
@@ -19,7 +19,7 @@ _version=15.0.7
 _rc=""
 _tag=llvmorg-${_version}${_rc}
 pkgver=${_version}${_rc/-/}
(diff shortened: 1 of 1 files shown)


Biswa96:
> Previously, llvmlite was depending on LLVM 14, which had these targets. Now it retargeted to newer LLVM, which broke both our platforms.

I agree. llvm-14 was removed in 849293932ce45098f6e0a96b634247d6897b4c5a and it had those targets.

lazka:
I think the reason it was kept minimal is that its main use is so we can build some packages that are slow to update to newer llvm, so mainly for MSYS2-internal use.

But this doesn't make things more complicated, so fine I guess.

PR #21364, opened 2024-07-10
python-python-multipart: new package

podsvirov:


ognevny:
why python is doubled in pkgname?

podsvirov:
> why `python` is doubled in pkgname?

Because [python-starlette](https://archlinux.org/packages/extra/any/python-starlette/) optionaly depends on [python-python-multipart](https://archlinux.org/packages/extra/any/python-python-multipart/) and [python-multipart](https://archlinux.org/packages/extra/any/python-multipart/) is another package.

PR #21384, opened 2024-07-11
libavif: update to 1.1.0

kmilos:


Biswa96:
This update caused a compiler error with kimageformats. I have imported a patch from upstream here 1a6cbd0854e2c7f84ef6533360b609d16fb57317

PR #21397, opened 2024-07-13
add package python-igraph

### LINKED ISSUE #21316: Add [mingw-w64-python-igraph]
### Package name

python-igraph

This is an interface for igraph in python, not so easy to install sometimes.

https://github.com/igraph/python-igraph

import igraph

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

_No response_

Kreijstal:
This fixes #21316, also thank you for the feedback on texttable
also related
https://github.com/igraph/python-igraph/issues/769
https://github.com/igraph/python-igraph/issues/574

### DIFF
--- a/mingw-w64-python-igraph/PKGBUILD
+++ b/mingw-w64-python-igraph/PKGBUILD
@@ -8,9 +8,9 @@ arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url='https://pypi.org/project/igraph/'
 license=('spdx:GPL-2.0')
-depends=("${MINGW_PACKAGE_PREFIX}-python" "${MINGW_PACKAGE_PREFIX}-python-texttable")
+depends=("${MINGW_PACKAGE_PREFIX}-python" "${MINGW_PACKAGE_PREFIX}-python-texttable" "$MINGW_PACKAGE_PREFIX-igraph")
 makedepends=("${MINGW_PACKAGE_PREFIX}-python-build" "${MINGW_PACKAGE_PREFIX}-python-installer" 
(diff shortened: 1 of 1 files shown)


szhorvat:
Just a heads up that if you decide to make this depend on `igraph`, it is your responsibility to update `python-igraph` and `igraph` concurrently. `python-igraph` is only guaranteed to be compatible with a single patch release of `igraph`. Even though in practice compatibility is not broken frequently, there is no guarantee that a given `python-igraph` will be compatible with an earlier or later `igraph`.

Kreijstal:
> Just a heads up that if you decide to make this depend on `igraph`, it is your responsibility to update `python-igraph` and `igraph` concurrently. `python-igraph` is only guaranteed to be compatible with a single patch release of `igraph`. Even though in practice compatibility is not broken frequently, there is no guarantee that a given `python-igraph` will be compatible with an earlier or later `igraph`.
> 
> You may choose to use the version of `igraph` included in `python-igraph` (instead of setting `IGRAPH_USE_PKG_CONFIG=1`)and then you will not need to be concerned about this.

But you would then have 2 possibly conflicting igraph versions?

szhorvat:
No, you would not. `python-igraph` includes the full source code of the C/igraph library, and this is used by default when you compile `python-igraph` (i.e. when you don't set `IGRAPH_USE_PKG_CONFIG`). This does not mean that the C igraph library will be installed. It won't. It will be statically linked into `python-igraph`.

Kreijstal:
> No, you would not. `python-igraph` includes the full source code of the C/igraph library, and this is used by default when you compile `python-igraph` (i.e. when you don't set `IGRAPH_USE_PKG_CONFIG`). This does not mean that the C igraph library will be installed. It won't. It will be statically linked into `python-igraph`.

It is all good as long as there are always python-igraph releases that match igraph itself

szhorvat:
There aren't (usually for lack of time). If you're in doubt about compatibility, feel free to ping me ... Sorry that I can't offer a better solution at the moment.

PR #21399, opened 2024-07-14
blake3: update to 1.5.3

ognevny:
added a package for a C library. it can be used for ccache (but #21401 is required first)


MehdiChinoune:
Could you enable SOVERSION for blake3.

PR #21404, opened 2024-07-14
msmpi: Avoid using gfortran extensions that are disabled by default.

mmuetzel:
Avoid non-standard conform usage of BOZ literals.

That fixes an issue if downstream projects use `mpifort` with gfortran 10 or newer (introduced in 6a55d50fac595fd7c961505888efb0ecf53d64b2).

See also: https://github.com/ElmerCSC/elmerfem/issues/482


MehdiChinoune:
I haven't introduced any new options, they were added by default and I made them only for gfortran

mmuetzel:
If I understand that changeset correctly, it moved a couple of compiler flags from the `mpifort` wrapper to the `PKGBUILD` rules.
But these flags were needed by downstream projects, too, when building with `gfortran`. (Not only when building this package itself.)

This PR makes the code in the headers more standard conform. Thus, avoiding the need for those compiler flags entrirely.

MehdiChinoune:
But, I didn't introduce the non-conforming declarations in mpi.f
Needs just clarification.

mmuetzel:
> But, I didn't introduce the non-conforming declarations in mpi.f
> Needs just clarification.

Sure. These non-conforming usages of BOZ literals in `mpif.h` was there since that header was originally added in MSYS2. `gfortran` (in version 10 and newer) needs additional flags to not fail for that non-conforming usage.
The commit I pointed to removed those (necessary) additional flags. Thus, causing compilation errors in downstream projects.

I hope that clarifies the situation and the proposed fix.

PR #21422, opened 2024-07-16
qt5-base: Change OpenGL to "dynamic" mode to support fallbacks

evilJazz:
Changing to -opengl dynamic will enable fallback code for loading ANGLE (libEGL.dll and libGLESv2.dll from package mingw-*-angleproject) as well as software rasterizers (opengl32sw.dll) at runtime. Currently -opengl desktop will only work with full-fledged OpenGL 2.x+ implementation, that is: Qt Quick 2 will not work in VMs or systems without 3D acceleration.

### DIFF
--- a/mingw-w64-qt5-base/PKGBUILD
+++ b/mingw-w64-qt5-base/PKGBUILD
@@ -13,7 +13,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-${_realname}-debug"))
 _qtver=5.15.14
 pkgver=${_qtver}+kde+r141
-pkgrel=1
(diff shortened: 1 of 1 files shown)


evilJazz:
Referring to https://www.msys2.org/dev/update-package/ , do you also want me to increment the pkgver from r141 to r142 or is this handled by the maintainer? Thank you!

lazka:
Without knowing too much about this, the change seems sensible.

* Upstream moved to this mode for their binaries long ago: https://www.qt.io/blog/2014/11/27/qt-weekly-21-dynamic-opengl-implementation-loading-in-qt-5-4 (this also a an explanation regarding what and why)
* With qt6 the ANGLE support was removed again, but the mesa support still exists: https://doc.qt.io/qt-6/opengl-changes-qt6.html#removal-of-angle

lazka:
There seems to be some fallout #21477

PR #21423, opened 2024-07-16
python-cx-freeze: update to 7.2.0

marcelotduarte:


Biswa96:
It is not required to move the source and checksum above.

lazka:
Not worth waiting on it though, so merging anyway

PR #21430, opened 2024-07-17
Some improvements with arrow package

Biswa96:


ognevny:
does this actually require a rebuild?

Biswa96:
> does this actually require a rebuild?

New commits are coming soon ™️

PR #21436, opened 2024-07-17
python-installer: Zero lpReserved2 and cbReserved2 fields in simple_launcher STARTUPINFOW

jhol:
simple_launcher currently copies the values of lpReserved2 and cbReserved2 from the parent process to the child. In some cases, these fields are used to pass information. For example, the MSVC C runtime uses the fields as part of its _spawn implementation to pass file descriptor information to the child. However, this information is invalid in the grand-child proces.

This issue is a cause of a bug when CMake attempts to launch a python-installer child process:

  * https://gitlab.kitware.com/cmake/cmake/-/issues/25996

This patch corrects the issue by applying a patch to zero the fields.

### DIFF
--- a/mingw-w64-python-installer/004-zero-reservered-fields-in-startupinfo.patch
+++ b/mingw-w64-python-installer/004-zero-reservered-fields-in-startupinfo.patch
@@ -1,0 +1,37 @@
+From 63deda9f0a881c7577a70744fbd19e73f7a6dcdb Mon Sep 17 00:00:00 2001
+From: Joel Holdsworth <jholdsworth@nvidia.com>
+Date: Wed, 17 Jul 2024 11:35:16 -0700
+Subject: [PATCH] Zero lpReserved2 and cbReserved2 fields in STARTUPINFOW
+
+simple_launcher currently copies the values of lpReserved2 and cbReserved2 from
+the parent process to the child. In some cases, these fields are used to pass
+information. For example, the MSVC C runtime uses the fields as part of its
+_spawn implementation to pass file descriptor information to the child. However,
+this information is invalid in the grand-child proces.
+
+This issue is a cause of a bug when CMake attempts to launch a python-installer
+child process:
+
+  * https://gitlab.kitware.com/cmake/cmake/-/issues/25996
+
+This patch corrects the issue by zeroing the fields.
+---
+ launcher.c | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/launcher.c b/launcher.c
+index c965d19..d15ab91 100644
+--- a/launcher.c
++++ b/launcher.c
+@@ -519,6 +519,8 @@ run_child(wchar_t * cmdline)
+     assert(ok, "Job information setting failed");
+     memset(&si, 0, sizeof(si));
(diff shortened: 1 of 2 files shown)


lazka:
Should we look into upstreaming this?

jhol:
> Should we look into upstreaming this?

Yes, but I'm not quite sure what to do.

The upstream project is here: https://bitbucket.org/vinay.sajip/simple_launcher.git . However, the BitBucket project is not configured to allow allow Merge Requests. So I reached out to the author with a patch in May. He replied to me, thanking me for the patch, said he would apply it, but never did. He doesn't seem to be very actively maintaining the project.

There is also the msys2-contrib fork repository: https://github.com/msys2-contrib/simple_launcher . This is a fork from some time in 2021. It lacks a few of the newer changes in the original repository.

Then there are the 2 (now 3), patches to simple_launcher in MINGW-packages which are not in either repository.

It's all rather a mess!

jhol:
> Should not those reserved fields be ignored? Are those used in any program or in CreateProcess internally?

I don't think `CreateProcess` uses them internally. The data is just handed to the sub-process. C libraries use it to pass info about fake UNIX file descriptors for Windows, and in the case of CMake (and possibly others) causes the msys2-mingw build of python to launch with `sys.stderr` set to `None`.

jhol:
Do we need to do anything to trigger the 566 downstream packages to rebuild?

lazka:
> Yes, but I'm not quite sure what to do.

I'll ask pypa folks on Discord

> The upstream project is here: https://bitbucket.org/vinay.sajip/simple_launcher.git . However, the BitBucket project is not configured to allow allow Merge Requests. So I reached out to the author with a patch in May. He replied to me, thanking me for the patch, said he would apply it, but never did. He doesn't seem to be very actively maintaining the project.

ok, thanks for trying

> There is also the msys2-contrib fork repository: https://github.com/msys2-contrib/simple_launcher . This is a fork from some time in 2021. It lacks a few of the newer changes in the original repository.

This was from when I tried to convince upstream to move to github and build the binaries in CI using this branch: https://github.com/msys2-contrib/simple_launcher/tree/meson-ci

> Then there are the 2 (now 3), patches to simple_launcher in MINGW-packages which are not in either repository.
> 
> It's all rather a mess!

normal open source business :)

lazka:
> Do we need to do anything to trigger the 566 downstream packages to rebuild?

If there are any specific packages you need fixes feel free to bump them.

For the others I should probably write a script to filter the real ones, not all reverse dependencies end up with launchers.

jhol:
> If there are any specific packages you need fixes feel free to bump them.

https://github.com/msys2/MINGW-packages/pull/21440 

(By the way, thanks for all your work on this. It's amazing to be able to get issues resolved so quickly).

lazka:
> For the others I should probably write a script to filter the real ones, not all reverse dependencies end up with launchers.

that left 163 packages: f8b7077f8a36f31f8e822c0953f

re simple_launcher: I've asked upstream again for accepting some help: https://bitbucket.org/vinay.sajip/simple_launcher/issues/9/ci-and-meson-support#comment-67053236

PR #21478, opened 2024-07-23
qxorm: new package

podsvirov:


MehdiChinoune:
Why not using `PREFIX=${MINGW_PREFIX}` and `INSTALL_ROOT=${pkgdir}`.

podsvirov:
> Why not using `PREFIX=${MINGW_PREFIX}` and `INSTALL_ROOT=${pkgdir}`.

Because that not working (qmake magic).

PR #21485, opened 2024-07-24
zig: enable mingw64 & ucrt64 (linked with static llvm)

raedrizqie:
- builds a pure llvm toolchain on top of MinGW
- builds libc++ and libunwind with builtin compiler-rt on all systems
- emulated TLS is disabled when building with Clang
- remove dependency of libgcc_s and libpthread on non Clang systems
- fix building of libc++abi shared library

Notes:
We previously build with `-DLIBCXXABI_ENABLE_STATIC_UNWINDER=ON` which actually leads to libc++ not depends on libunwind, with both libc++ and libunwind exporting the same unwind functions.

If we set `-DLIBCXXABI_ENABLE_STATIC_UNWINDER=OFF`, then libc++ will correctly depends on libunwind.


ognevny:
without target triple zig compiler segfaulted for me. why did you remove it?

raedrizqie:
im so confused right now.. on clang, it will always fails at `build()` step.. mingw works fine though

ArchGuyWu:
> im so confused right now.. on clang, it will always fails at `build()` step.. mingw works fine though
> 
> ```
> 2024-07-25T04:42:45.6816477Z ==> Starting build()...
> 2024-07-25T04:42:45.7760069Z ==> ERROR: A failure occurred in build().
> ```

I downloaded your PKGBUILD file for mingw-w64-clang-x86_64-zig, and built it locally.

your PKGBUILD file:
[PKGBUILD.txt](https://github.com/user-attachments/files/16470782/PKGBUILD.txt)

the built package:
[mingw-w64-clang-x86_64-zig-0.13.0-2-any.pkg.tar.zst.zip](https://github.com/user-attachments/files/16470797/mingw-w64-clang-x86_64-zig-0.13.0-2-any.pkg.tar.zst.zip)

the build script (powershell):
[makepkg.txt](https://github.com/user-attachments/files/16470804/makepkg.txt)

mingw-w64-libc++/PKGBUILD
```diff
@@ -49,61 +53,79 @@ sha256sums=('bdecf90be0072bc720fd5c9c8ab061cdb197edd0c8ad3e170dc3e6bfaa49f388'
...
   else
-    _extra_config+=(-DLIBCXXABI_USE_LLVM_UNWINDER=OFF)
+    _extra_config+=(-DLIBUNWIND_USE_COMPILER_RT=ON)
   fi
 
   # Targeting Win 7 will just lead to libc++ looking
   # up new APIs at runtime, so there is no downside really
   local _win32_winnt
-  if [[ ${MSYSTEM} == CLANGARM64 ]]; then
+  if [[ ${MSYSTEM} == CLANG* ]]; then
```
ognevny, line 115: why?

raedrizqie: i hit linker errors from undefined stdlib implementations (_free, _memcp, etc) when compiling ruby v3.3.4.. just want to experiment something here..

ognevny: it also makes sense as UCRT is available only in Windows 10 or later

lazka:
> @lazka please review

I'm afraid I don't know enough about llvm to know what this is trying to change any why, so I can't.

raedrizqie:
this change is needed to make zig build on mingw64 and clang64, because zig needs llvm with builtin compiler-rt

mati865:
> * emulated TLS is disabled when building with Clang

Do I correctly understand this enables native TLS for LLVM libs?
If so, this will break linking to LLVM libs with GCC (or Clang with emulated TLS enabled).

raedrizqie:
> > * emulated TLS is disabled when building with Clang
> 
> Do I correctly understand this enables native TLS for LLVM libs?

Yes, becuse it will fail when linking zig on stage 3, otherwise..

>If so, this will break linking to LLVM libs with GCC (or Clang with emulated TLS enabled).

The resulting llvm lib from this PR does not linked to gcc libs. In the case of using emutls on clang, I never tried it before though.

mati865:
> The resulting llvm lib from this PR does not linked to gcc libs. In the case of using emutls on clang, I never tried it before though.

We have tried to link LLVM with libc++ in the past and had to revert it because packages using LLVM libs (like mesa) broke on GNU envs. So this part is basically a no-go. I guess the same will be true with TLS change.

If Zig needs LLVM that is not compatible with all other packages in the env, then it should probably build its own variant.

raedrizqie:
> > The resulting llvm lib from this PR does not linked to gcc libs. In the case of using emutls on clang, I never tried it before though.
> 
> We have tried to link LLVM with libc++ in the past and had to revert it because packages using LLVM libs (like mesa) broke on GNU envs. So this part is basically a no-go. I guess the same will be true with TLS change.

I see..

> If Zig needs LLVM that is not compatible with all other packages in the env, then it should probably build its own variant.

I will try to make a static llvm bootstrap package for zig. It will only be used when building the zig package.

PR #21492, opened 2024-07-25
firebird: update to 5.0.1.1469 and port to cmake

raedrizqie:
- port to cmake
- do not remove icu from depends, because it is dynamically loaded at runtime


raedrizqie:
randomly fails to build on mingw64 and ucrt64.. closing

PR #21506, opened 2024-07-26
Update GW version PKGBUILD

kcleal:


MehdiChinoune:
Please use a better commit message next time.

PR #21521, opened 2024-07-29
libtool: update to 2.5.1

ikspress:
**Changes in this PR**
`0007-fix-cr-for-awk-in-configure.all.patch` is no longer needed and has been removed.

**Issues**
1. I have no idea to refresh `0003-Pass-various-runtime-library-flags-to-GCC.mingw.patch`, help is appreciated. :)
2. `0009-libtool-2.4.2.418-msysize.patch` is not a good idea. Since `configure` no longer exists, it is possible to modify `autotools` to achieve the desired effect?


lazka:
why don't you use the release tarball which contains the configure?

ikspress:
> why don't you use the release tarball which contains the configure?

There is no `configure` in [libtool-2.5.1.tar.gz](https://git.savannah.gnu.org/cgit/libtool.git/snapshot/libtool-2.5.1.tar.gz). It is a snapshot version that not exist on [ftp.gnu.org](https://ftp.gnu.org/pub/gnu/libtool/). Is there any way to download the tarball with `configure`?

ikspress:
Oh, I noticed `0009-libtool-2.4.2.418-msysize.patch` were fixed, issue 1 solved.

ikspress:
Some changes in `0003-Pass-various-runtime-library-flags-to-GCC.mingw.patch` were fixed in 2.5.1 and `libtool` handles `-static-*` properly now (but -shared-* won't). issue 1 may fixed?

lazka:
> There is no `configure` in [libtool-2.5.1.tar.gz](https://git.savannah.gnu.org/cgit/libtool.git/snapshot/libtool-2.5.1.tar.gz). It is a snapshot version that not exist on [ftp.gnu.org](https://ftp.gnu.org/pub/gnu/libtool/). Is there any way to download the tarball contains `configure`?

https://alpha.gnu.org/gnu/libtool/

ikspress:
> > There is no `configure` in [libtool-2.5.1.tar.gz](https://git.savannah.gnu.org/cgit/libtool.git/snapshot/libtool-2.5.1.tar.gz). It is a snapshot version that not exist on [ftp.gnu.org](https://ftp.gnu.org/pub/gnu/libtool/). Is there any way to download the tarball contains `configure`?
> 
> https://alpha.gnu.org/gnu/libtool/

Oh, okay. I've updated `0004-libtool-msysize.patch`.

One noticable thing is I modified the handling of `msys-*.dll` in `libtool.m4`, the original code is confusing.
Modified code:
```shell
    cygwin*)
      # Cygwin DLLs use 'cyg' prefix rather than 'lib'
      soname_spec='`echo $libname | $SED -e 's/^lib/cyg/'``echo $release | $SED -e 's/[[.]]/-/g'`$versuffix$shared_ext'
m4_if([$1], [],[
      sys_lib_search_path_spec="$sys_lib_search_path_spec /usr/lib/w32api"])
      ;;
    msys*)
      # MSYS DLLs use 'msys' prefix rather than 'lib'
      soname_spec='`echo $libname | $SED -e 's/^lib/msys-/'``echo $release | $SED -e 's/[[.]]/-/g'`$versuffix$shared_ext'
m4_if([$1], [],[
      sys_lib_search_path_spec="$sys_lib_search_path_spec /usr/lib/w32api"])
      ;;
```
Original code:
```shell
    cygwin* | msys*)
      # Cygwin DLLs use 'cyg' prefix rather than 'lib'
      soname_spec='`echo $libname | $SED -e 's/^lib/msys-/'``echo $release | $SED -e 's/[[.]]/-/g'`$versuffix$shared_ext'
m4_if([$1], [],[
      sys_lib_search_path_spec="$sys_lib_search_path_spec /usr/lib/w32api"])
      ;;
```

lazka:
That is on purpose: https://github.com/msys2/MSYS2-packages/pull/2779

ikspress:
> That is on purpose: [msys2/MSYS2-packages#2779](https://github.com/msys2/MSYS2-packages/pull/2779)

Okay. Thank you for your patience. I revert my changes back. But I think the following comment in original code is incorrect.
`# Cygwin DLLs use 'cyg' prefix rather than 'lib'`
And then I changed it to
`# MSYS DLLs use 'msys-' prefix rather than 'lib'`
If you are not satified with this change, please feel free to tell me...

I've noticed some changes in `0014-Support-llvm-objdump-f-output.patch` and  have not applied to the current `configure` file in mingw-w64-libtool. Should I proceed with applying it?

Here is how I generated `0004-libtool-msysize.patch`. Hopefully it may help contributors to review my changes.

1. Build mingw-w64-libtool 2.5.1 at first.
2. Download `libtool-2.5.1.tar.gz` and extract it to libtool-2.5.1 and  libtool-2.5.1-orig.
3. Apply `0004-libtool-msysize.patch` to fix `ltmain.in`, `libtool.m4`, `ltdl.m4` and etc.
4. Run `autoreconf -fi` to regenerate `configure`, `build-aux/config.guess` and etc with MSYS target support.
5. Run `rm -rf  configure~ config-h.in~ autom4te.cache`  to remove the obslete files which can reduce the patch size (`autoreconf` may have option to remove obslete files, but I don't know...).
6. Run `git diff --minimal --no-prefix --output=1.patch libtool-2.5.1-orig libtool-2.5.1` to generate patch.
7. Remove unapplied changes I mentioned above and the version string changes manually.
8. Repatch `1.patch` in a fresh libtool-2.5.1 directory to get the correct line number.
9. Generate patch and output it to `mingw-w64-libtool/0004-libtool-msysize.patch`.
10. OPTIONAL! You can follow these steps again...It will get the same patch.

I compared `0004-libtool-msysize.patch` with the original `0009-libtool-2.4.2.418-msysize.patch`. Then I found that `ltargz.m4` treated cygwin as lack of argz.h (but it exists). Should I add it for msys target?

ikspress:
Just run `autoreconf -fi` will failed with some file missing. It's too complex to regenerate `configure` in tarball without .git directory. So I think switch to clone git tag is better...

ikspress:
Simply running `autoreconf -fi` fails due to missing files. Additionally, manually patching `configure` like `0004-libtool-msysize.patch` does is unnecessarily complex.
Instead, regenerating `configure` by `bootstrap` would be a good approach to add MSYS target support and include changes in `0014-Support-llvm-objdump-f-output.patch` I mentioned above. It can also make it easier to review `0004-libtool-msysize.patch` and updates in future. However, this process requires needs a .git directory present. This situation is quite frustrating...
I found that Arch Linux always using clone git repository instead of tarball file. Could we follow them?
Therefore, reverting back to cloning the git tag might be a more appropriate solution...

ikspress:
I will find a clear way to configure libtool, but close this PR now.
Thank you for your patience, lazka.

Issue #21313, opened 2024-07-04
neovim update to 0.10.0


hjm7:
### Package name

mingw-w64-clang-x86_64-neovim

### New version number

 0.10.0

### Other information that may be useful

Nvim 0.10 introduces a new 'termsync' option (enabled by default) which enables [synchronized output](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036) feature if it is supported by the terminal emulator.

prgreadonly:
Would be nice to see this for ucrt64 as well

sewbacca:
I gave updating neovim to v0.10.0 a go, since lazydev requires this version, but didn't finish it yet. I got halfway through though, so if anyone wants to give it a try, here is what I've found out so far:
- The patch `0001-cmake-disable-bundling.patch` needs to be updated to:
```diff
diff --git a/src/nvim/CMakeLists.txt b/src/nvim/CMakeLists.txt
index d9cc695c..42465eca 100644
--- a/src/nvim/CMakeLists.txt
+++ b/src/nvim/CMakeLists.txt
@@ -726,7 +726,7 @@ if(ENABLE_LTO)
 endif()

 add_custom_target(nvim_runtime_deps)
-if(WIN32)
+if(FALSE)
   # Copy DLLs and third-party tools to bin/ and install them along with nvim
   add_custom_command(TARGET nvim_runtime_deps
     COMMAND ${CMAKE_COMMAND} -E ${COPY_DIRECTORY} ${PROJECT_BINARY_DIR}/windows_runtime_deps/
```
- Add `0002-cmake-fix-FindLpeg.patch` to fix finding `lpeg` both during configuration and compilation
```diff
diff --git a/cmake/FindLpeg.cmake b/cmake/FindLpeg.cmake
index 3d0ff5929..7e6bca45a 100644
--- a/cmake/FindLpeg.cmake
+++ b/cmake/FindLpeg.cmake
@@ -1,3 +1,4 @@
+set(CMAKE_FIND_LIBRARY_SUFFIXES .dll ${CMAKE_FIND_LIBRARY_SUFFIXES})
 find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} PATH_SUFFIXES lua/5.1)

 find_package_handle_standard_args(Lpeg DEFAULT_MSG LPEG_LIBRARY)
[...]

PR #21522, opened 2024-07-29
neovim: update to 0.10.1

sewbacca:
This PR updates neovim from 0.9.5 to 0.10.1 and fixes an issue during the compilation process while linking against `lpeg.dll` which is not easily possible. To circumvent the problem, neovim is patched to load the dll through Lua's `dlopen` mechanism instead. See #21313 for more details.

### DIFF
--- a/mingw-w64-neovim/0001-cmake-disable-bundling.patch
+++ b/mingw-w64-neovim/0001-cmake-disable-bundling.patch
@@ -1,11 +1,13 @@
+diff --git a/src/nvim/CMakeLists.txt b/src/nvim/CMakeLists.txt
+index d9cc695c..42465eca 100644
 --- a/src/nvim/CMakeLists.txt
 +++ b/src/nvim/CMakeLists.txt
-@@ -668,7 +668,7 @@
-   endif()
+@@ -726,7 +726,7 @@ if(ENABLE_LTO)
  endif()
  
+ add_custom_target(nvim_runtime_deps)
 -if(WIN32)
 +if(FALSE)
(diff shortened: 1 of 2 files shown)


lazka:
Is there an upstream issue for this patch?

sewbacca:
No I couldn't find one. Shall I open a issue explaining the problem to them?

lazka:
Yes please. I don't think it would be wise to maintain this downstream unless it is a MSYS2 specific issue (or is it?)

sewbacca:
I've submitted an issue at neovim/neovim#30022
This is a Windows issue, so I am not sure whether the fix should be applied here or upstream.

sewbacca:
Over on the neovim issue, clason pointed out to me the `BUILD.MD`, which helped me to figure out some stuff. We could simply solve this issue, by statically compiling against `lpeg`, which does compile just fine.

Ideally we would link dynamically against `lpeg.dll`, but then we would need to fix this issue / wait for the fix to happen.

I gonna wait and see if they have a quickfix for me to link against `lpeg.dll`. If not, then shall I link statically against `lpeg` instead? From a maintainance point of view, this seems the easiest fix.

sewbacca:
Okay so I've changed this PR to compile `lpeg` statically, which gets rid of the patch for linking against the installed `lpeg`. This is of course not ideal, as we have lpeg as a package available. But due to the outlined issues, linking against it comes with a bunch of issues. The original patch can be viewed at dace21b1b7fc9af22eca252e5a38bcd2f6854eea.

lazka:
Sounds like a good compromise given the circumstances, thanks.

PR #21527, opened 2024-07-30
uv: update to 0.2.32

ognevny:


ognevny:
rustc 1.80.0 is required

PR #21540, opened 2024-08-01
qemu: update to 9.1.0

heljkon:
see https://wiki.qemu.org/ChangeLog/9.1

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -6,19 +6,19 @@ pkgname=(
   "${MINGW_PACKAGE_PREFIX}-qemu-guest-agent"
   "${MINGW_PACKAGE_PREFIX}-qemu-image-util"
 )
-_base_ver="9.0.2"
+_base_ver="9.1.0"
 # QEMU Versioning of RC-SourcePackage and RC-Version differs
 # e.g. qemu-6.1.0-rc0.tar.xz contains 6.0.90, qemu-6.1.0-rc1.tar.xz contains 6.0.91
 # Unset _rc_no to create Release-Build
 #_rc_no="4"
-_rc_base_ver="8.2.9${_rc_no}"
+_rc_base_ver="9.0.9${_rc_no}"
 _rc_file_ver="rc${_rc_no}"
 _tarname=$( [ -z "${_rc_no}" ] && echo ${_realname}-${_base_ver} || echo ${_realname}-${_base_ver}-${_rc_file_ver} )
 pkgver=$( [ -z "${_rc_no}" ] && echo ${_base_ver} || echo ${_rc_base_ver} )
 pkgrel=1
 pkgdesc="QEMU - a generic and open source machine emulator and virtualizer (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+mingw_arch=('mingw64' 'ucrt64')
 license=('spdx:GPL-2.0-or-later AND LGPL-2.1-or-later AND BSD-2-Clause AND BSD-3-Clause AND MIT')
 url="https://qemu.org/"
 msys2_repository_url="https://gitlab.com/qemu-project/qemu"
@@ -113,11 +113,11 @@ source=(
   msys2.examples.tests.sh
 )
 sha256sums=(
-  'a8c3f596aece96da3b00cafb74baafa0d14515eafb8ed1ee3f7f5c2d0ebf02b6'
+  '816b7022a8ba7c2ac30e2e0cf973e826f6bcc8505339603212c5ede8e94d7834'
   'SKIP'
   '51625fd83c0a63729942d3dfc6d48be3491a700fe0f7cb8e88935b46081c9016'
   'f2f9eeb31023d002f54637a9941ea0d8fae3c6f0a66c05c033857b305bd1470d'
-  'a740a5cfce50083dabbb6af2f24c320ccc9d6b762b5298911750bc6e3cfbaba1'
+  '1478e41f27a4646e14c2abca764d9fd2fd158a4a4cd6421e077fba101d4a74ea'
 )
 validpgpkeys=('CEACC9E15534EBABB82D3FA03353C9CEF108B584') # Michael Roth <flukshun@gmail.com>
 # tar cannot create links, to keep build running, manual extraction is required
@@ -252,7 +252,7 @@ package_qemu() {
   if enablePlugins
   then
     mkdir -pv lib/qemu/plugins/test lib/qemu/plugins/contrib
-    install -t lib/qemu/plugins/test "${srcdir}"/build-${MSYSTEM}/tests/plugin/*.dll
+    install -t lib/qemu/plugins/test "${srcdir}"/build-${MSYSTEM}/tests/tcg/plugins/*.dll
     install -t lib/qemu/plugins/contrib "${srcdir}"/build-${MSYSTEM}/contrib/plugins/*.dll
   fi
 }
--- a/mingw-w64-qemu/msys2.examples.tests.sh
+++ b/mingw-w64-qemu/msys2.examples.tests.sh
@@ -876,6 +876,7 @@ function qemuPlugins {
 	local -A PLUGIN_OPTS
 	PLUGIN_OPTS["libmem.dll"]=",inline=true"
 	PLUGIN_OPTS["libbb.dll"]=",inline=true,idle=true"
+	PLUGIN_OPTS["libstoptrigger.dll"]=",addr=0xd4"
 	PLUGIN_OPTS["libinsn.dll"]=",inline=false,sizes=true"
 	PLUGIN_OPTS["libhotpages.dll"]=",sortby=reads,io=on"
 	PLUGIN_OPTS["libexeclog.dll"]=",ifilter=msr"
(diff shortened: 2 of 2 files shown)


heljkon:
For the clang64 build failure: https://gitlab.com/qemu-project/qemu/-/issues/2476

heljkon:
@lazka @Biswa96 @jeremyd2019
Based on ticket [Regression 9.1.0-rc0: Msys2/Clang64 build fails](https://gitlab.com/qemu-project/qemu/-/issues/2476) in QEMU repository a discussion came up about the fitness of CLANG64/CLANGARM64 for QEMU.

[thuth](https://gitlab.com/thuth), a core QEMU developer stated:

> Ok, so Clang does not support gcc_struct yet. You should not use it to build QEMU on MSYS2. While 95% of such a build might work fine, there are likely some emulated devices where this gcc_struct attribute makes a difference and those won't work in such a build.

As I am only contributing to msys2, I ask you the members of msys2 for discussion and decision.

lazka:
If upstream doesn't want to fix their struct packing to not depend on gcc_struct there is not much we can do I guess.

Not sure, your call.

I look at why they added it in the first place: https://lists.gnu.org/archive/html/qemu-devel/2011-08/msg01877.html

jeremyd2019:
> [thuth](https://gitlab.com/thuth), a core QEMU developer stated:
> 
> > Ok, so Clang does not support gcc_struct yet. You should not use it to build QEMU on MSYS2. While 95% of such a build might work fine, there are likely some emulated devices where this gcc_struct attribute makes a difference and those won't work in such a build.

@mstorsjo are you aware of this limitation?  I figured clang has to support both struct packing styles, since it emulates both gcc and msvc frontends, but I guess maybe not both on the same target?

mstorsjo:
> > [thuth](https://gitlab.com/thuth), a core QEMU developer stated:
> > > Ok, so Clang does not support gcc_struct yet. You should not use it to build QEMU on MSYS2. While 95% of such a build might work fine, there are likely some emulated devices where this gcc_struct attribute makes a difference and those won't work in such a build.
> 
> @mstorsjo are you aware of this limitation? I figured clang has to support both struct packing styles, since it emulates both gcc and msvc frontends, but I guess maybe not both on the same target?

Yes, I'm aware of this. It's indeed a case where Clang does support both, it just doesn't support the attribute. It does IIRC support `ms_struct` for opting in to the MS struct layout when building for a target that otherwise uses the default layout, but it doesn't yet support `gcc_struct` for requesting the reverse, when building in a configuration where the MS layout is the default. See https://github.com/llvm/llvm-project/issues/24757 for the general tracking issue, and https://github.com/llvm/llvm-project/pull/71148 for a recent PR that attempts to implement this.

heljkon:
@lazka @Biswa96 @jeremyd2019 @mstorsjo 
Upstream has added a compiler test to make sure, CLANG64 is not build. I don't want to patch this, so I've disabled clang builds.
```
==> Starting build()...
...
../qemu-9.1.0-rc3/meson.build:321:4: ERROR: Problem encountered: Your compiler does not support __attribute__((gcc_struct)) - please use GCC instead of Clang

A full log can be found at .../mingw-w64-qemu/src/build-CLANG64/meson-logs/meson-log.txt

ERROR: meson setup failed

    Aborting...
```
 .../mingw-w64-qemu/src/build-CLANG64/meson-logs/meson-log.txt:
```
-----------
Command line: `clang++ -m64 ../tmpud8icv9j/testfile.cpp -o ../tmpud8icv9j/output.obj -c -march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Werror=implicit-function-declaration -Werror=unknown-warning-option -Werror=unused-command-line-argument -Werror=ignored-optimization-argument -std=gnu++11` -> 0
Running compile:
Working directory:  ./
Code:
struct x { int y; } __attribute__((gcc_struct));
-----------
Command line: `clang++ -m64 ../tmpoo9iwq6a/testfile.cpp -o ../tmpoo9iwq6a/output.obj -c -march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Werror=implicit-function-declaration -Werror=unknown-warning-option -Werror=unused-command-line-argument -Werror=ignored-optimization-argument -std=gnu++11 -Werror` -> 1
stderr:
../tmpoo9iwq6a/testfile.cpp:1:36: error: unknown attribute 'gcc_struct' ignored [-Werror,-Wunknown-attributes]
    1 | struct x { int y; } __attribute__((gcc_struct));
      |                                    ^~~~~~~~~~
1 error generated.
-----------

../qemu-9.1.0-rc3/meson.build:321:4: ERROR: Problem encountered: Your compiler does not support __attribute__((gcc_struct)) - please use GCC instead of Clang
```

jeremyd2019:
That's too bad - my "default" has been using CLANG64 qemu, generally with whpx but sometimes with tcg for arm64 target.  I suppose I'll have to switch to UCRT64 until clang gains support for gcc_struct.

heljkon:
@lazka @jeremyd2019 @Biswa96 
Please review!

As already discussed, QEMU packages for clang weren't build, because upstream has introduced a test which currently results in failing clang builds:

> ERROR: Problem encountered: Your compiler does not support __attribute__((gcc_struct)) - please use GCC instead of Clang

heljkon:
Just a reminder: still waiting for review.

jeremyd2019:
I miss having qemu-img on arm64 ☹️

pbo-linaro:
I tried recently to push a series fixing clang build on windows. However, maintainers prefer to wait for official clang support for gcc_struct attribute. That said, you can enable it on msys2 side with those two patches: [1](https://patchew.org/QEMU/20241031040426.772604-1-pierrick.bouvier@linaro.org/20241031040426.772604-8-pierrick.bouvier@linaro.org/) [2](https://patchew.org/QEMU/20241031040426.772604-1-pierrick.bouvier@linaro.org/20241031040426.772604-9-pierrick.bouvier@linaro.org/).

IMHO supporting windows-arm64 and clang on windows is worth the risk to introduce any problem with the packing stuff.

pbo-linaro:
@jeremyd2019 You can use mingw64 or ucrt64 msys2 env for that. It will work thanks to x64 emulation (if you're on win11). The overhead should not be a problem for running a simple tool like qemu-img.

pbo-linaro:
I think too it's more a FUD argument than a real problem, QEMU is one of the only big open source projects to use gcc_struct attribute.
That said, MSYS2 is free to apply patches and enable clang for upcoming 9.2 release.

bonzini:
> I think too it's more a FUD argument than a real problem, QEMU is one of the only big open source project to use gcc_struct attribute.

One solution is to add `QEMU_BUILD_BUG_ON(sizeof(...) == ...)` to all structs in QEMU that use bitfields. That will prove very quickly if there are issues or not.

jeremyd2019:
In addition to testing if any struct changes size, each such struct would need to be evaluated to see if it actually matters if it is a different size - not every struct would be exposed to a guest, which is pretty much the only thing I can think of where the layout would matter that much (since this is a Windows host, any struct exposed to the host system would be expected to be laid out with ms bitfields).

pbo-linaro:
Thinking about it today, how about we simply dump dwarf info from qemu-binaries (using llvm-dwarfdump) and simply compare information with following builds configs on windows:
- gcc: using gcc_struct
- gcc: not using gcc_struct
- gcc: using globally option -mno-ms-bitfields
- clang: using globally option -mno-ms-bitfields

It will only cover the types we used, but if we don't, I don't think we should be worried about those.

pbo-linaro:
After a deep investigation, I will propose once again to switch to using `-mno-ms-bitfields` instead of `gcc_struct` attribute. It is, I think, a saner and better default on Windows.
As a consequence, it will allow us to enable clang builds, which supports this option.
 
Details of the investigation are on the [QEMU mailing list](https://lore.kernel.org/qemu-devel/66c346de-7e20-4831-b3eb-1cda83240af9@linaro.org/T/#u).

jeremyd2019:
I would still be concerned about it changing layouts for structs passed to external APIs.  If you can prove that isn't happening, great.

Potential badness: `name:_GIOChannel size:0x70 DW_AT_decl_file:C:/msys64/mingw64/include/glib-2.0/glib/giochannel.h DW_AT_decl_line:100 DW_AT_decl_column:8  ` this would have been built with the default layout in the glib binaries in msys2.

pbo-linaro:
All struct used through QEMU source code are present in debug info, so the listing should be exhaustive. Is there any other potential case that could have been missed?

For `_GIOChannel`, we don't manipulate directly, nor allocate this structure, only glib does. So the size difference has no impact.

Yes, I did this investigation with gcc only for now. There is no point to push further if other devs don't agree in this direction. If we switch to using this option, any difference with clang would be a compiler bug.

mstorsjo:
FWIW, I don't remember if this had been discussed, but I think it probably could be safe to grab the Clang patch for `gcc_struct` support and apply it to the msys2 Clang. The patch iirc requires a little bit more work for upstream (at least adding some tests, maybe something more), but if you'd give it a quick practical test to see that it seems to do what is expected (give structs the same layout as with `-mno-ms-bitfields`) for some cases, I think it's probably quite safe to apply and use.

bonzini:
The conclusion of the discussion is that it *is* safe to drop the usage of `gcc_struct` (and we now have the data to prove it).

pbo-linaro:
Thanks for your help on this Paolo!

pbo-linaro:
Note: This change will not be included officially in 9.2 release, but MSYS2 is free to apply the needed patches (I'll post them here) to enable clang builds.

PR #21547, opened 2024-08-01
[new package] wasm-component-ld

ArchGuyWu:
wasm-component-ld is a wrapper around two pieces of functionality used to produce a [WebAssembly Component](https://component-model.bytecodealliance.org/), which will first invoke wasm-ld and then run the componentization process to produce a final component.


ArchGuyWu:
I'll add another package wasi-libc later, then we're ready to add package rust-wasm.

ArchGuyWu:
I'll try to add package rust-wasm by myself, so close thi for now.

PR #21551, opened 2024-08-01
libuhd: fix python library install path

ognevny:
@chrisjohgorman I tried this and it worked _locally_

I think that `-DUHD_PYTHON_DIR` can be removed at all


ognevny:
> I tried this and it worked _locally_

it also succeeded in CI
> I think that `-DUHD_PYTHON_DIR` can be removed at all

waiting for your opinion :)

chrisjohgorman:
Hi Maksim,

Thank you.  This appears to fix a bug that I have a poor hack prepared
to fix.  I will run a test to see if -DUHD_PYTHON_DIR can be removed.
I approve of your PR, but I'm not sure who has to commit it.

PR #21553, opened 2024-08-02
gnuradio: Add deps needed by gnuradio-companion.

MarkMoss:
This PR adds python-gobject and gtk3 as dependencies for gnuradio.  These dependences are needed to run gnuradio-companion

Without python-gobject, attempting to run gnuradio-companion fails with the following error:

```
$ gnuradio-companion.exe
ModuleNotFoundError

Failed to initialize GTK. If you are running over ssh, did you enable X forwarding and start ssh with -X?

(No module named 'gi')
```

Without gtk3, attempting to run gnuradio-companion fails with the following error:

```
$ gnuradio-companion.exe
While trying to display an error message, another error occurred
Namespace Gtk not available
The original error message follows.
ValueError

Failed to initialize GTK. If you are running over ssh, did you enable X forwarding and start ssh with -X?

(Namespace Gtk not available)
```

### DIFF
--- a/mingw-w64-gnuradio/PKGBUILD
+++ b/mingw-w64-gnuradio/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gnuradio
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=3.10.11.0
-pkgrel=3
+pkgrel=4
 pkgdesc="General purpose DSP and SDR toolkit (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64')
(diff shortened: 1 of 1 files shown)


ognevny:
@chrisjohgorman

chrisjohgorman:
@MarkMoss thank you for the bugfix.  I agree with your changes, nice catch.  @ognevny can you commit Mark's changes?

mmuetzel:
Would it make sense to remove these packages from `makedepends` now that they are in the `depends` list? Or alternatively, move them to `optdepends` if they are only needed for some use-cases?

chrisjohgorman:
@mmutzel thank you for your oversight.  Indeed you are right.  I should have caught this.  We should remove these packages from the `makedepends` list.  I am ready to make a PR for this, but wanted to know if you thought this change should warrant a `pkgrel` bump?

mmuetzel:
No worries. No big harm done at all.

Bumping the `pkgrel` would probably not hurt for that change.

chrisjohgorman:
Thanks @mmuetzel .  Unfortunately, there are some more dependency problems other than the ones outlined here.  Specifically, `makedepends` packages that should be in the `depends` list.  I'm making a PR to fix these errors as well as removal of `gtk3` and `python-gobject` from the `makedepends` list.

PR #21555, opened 2024-08-02
libuhd: bugfixes

chrisjohgorman:
- There was a small section of code commented out because I couldn't get it to build. A simple cast fixes this. Now we can use pthread_setname_np to set thread names.

- Use cygpath.exe to set a proper path for the _DEFAULT_INSTALL_PATH for uhd_images_downloader

- Move uhd_images_downlader.py and usrp2_card_burner.py to ${MINGW_PREFIX}/bin and change their names to uhd_images_downloader and usrp_card_burner respectively. The reflects the contents of /usr/bin in the arch package.


chrisjohgorman:
Checked build artifacts.  This looks good to merge.

PR #21556, opened 2024-08-03
poppler:update to 24.08.0

sunderme:
built on Clang64, the other crashed for unrelated reasons


lazka:
thanks.

the gdal rebuild was missing; I pushed that to master.

PR #21566, opened 2024-08-04
msgpack-c: update to 6.0.2

ikspress:


lazka:
Just curious, why was 32bit dropped?

ikspress:
> Just curious, why was 32bit dropped?

32-bit cjson is no longer exists, so it will failed to build examples. Additionally, I was too tried last night, so I just disable it. Do you think disable examples for 32-bit might be a better approach? Or disable examples and tests like `msgpack-cxx` does? (but it also not provide 32-bit package) Oh, it might be a dependency of binutils, so 32-bit package is meaningful.

lazka:
> 32-bit cjson is no longer exists, so it will failed to build examples

Ah, thanks, I missed that. I think it's find. If someone complains we can have another look.

MehdiChinoune:
> This PR is ready for review and merge now. A small question, is there any reason why disable mingw64 build for `msgpack-cxx`?

The contributor who added it (me) chose to.
There is no reason for someone using Windows 10+ to use MINGW64.

PR #21573, opened 2024-08-05
gnuradio: fixup dependencies

chrisjohgorman:
- bump pkgrel
- remove makedepends on `gtk3` and `python-gobject` that were added to depends by https://github.com/msys2/MINGW-packages/pull/21553
- enable check to run test suite.  one case, `qa_throttle`, fails.
- add numpy-2.0.patch from arch to fix test case `qa_correlate_access_code`
- used `namcap` and `ntldd.exe` fix dependencies
  - moved packages from makedepends to depends as they were determined to be dependencies
  - removed dependencies that were installed as a dependency of another package eg. `libiio` is a dependency of `libad9361-iio`, `python-cairo` is a dependency of `python-gobject` and `fmt` is a dependency of `spdlog`.
  - added packages that were python dependencies determined by `namcap`, `python-lxml`, `pyqtgraph` and `python-pyqt5`
  - added dependency to `python-pyopengl` as it is a dependency `gnuradio-companion` in arch and an optional dependency for `pyqtgraph`.
  - added `optdepends` for `soapyrtlsdr`

### DIFF
--- a/mingw-w64-gnuradio/PKGBUILD
+++ b/mingw-w64-gnuradio/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gnuradio
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=3.10.11.0
-pkgrel=4
+pkgrel=5
 pkgdesc="General purpose DSP and SDR toolkit (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64')
@@ -15,12 +15,14 @@ msys2_references=(
 )
 license=(spdx:GPL-3.0-or-later)
 depends=("${MINGW_PACKAGE_PREFIX}-codec2"
+  "${MINGW_PACKAGE_PREFIX}-boost"
   "${MINGW_PACKAGE_PREFIX}-gcc-libs"
   "${MINGW_PACKAGE_PREFIX}-gmp"
   "${MINGW_PACKAGE_PREFIX}-gsl"
   "${MINGW_PACKAGE_PREFIX}-gsm"
   "${MINGW_PACKAGE_PREFIX}-gtk3"
   "${MINGW_PACKAGE_PREFIX}-libad9361-iio"
+  "${MINGW_PACKAGE_PREFIX}-libiio"
   "${MINGW_PACKAGE_PREFIX}-libuhd" 
   "${MINGW_PACKAGE_PREFIX}-libunwind"
   "${MINGW_PACKAGE_PREFIX}-fftw"
@@ -29,6 +31,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-codec2"
   "${MINGW_PACKAGE_PREFIX}-python-click"
   "${MINGW_PACKAGE_PREFIX}-python-click-plugins"
   "${MINGW_PACKAGE_PREFIX}-python-gobject"
+  "${MINGW_PACKAGE_PREFIX}-python-cairo"
   "${MINGW_PACKAGE_PREFIX}-python-jsonschema"
   "${MINGW_PACKAGE_PREFIX}-python-mako"
   "${MINGW_PACKAGE_PREFIX}-python-matplotlib"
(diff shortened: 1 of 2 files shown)


mingw-w64-gnuradio/PKGBUILD
```diff
@@ -38,37 +39,35 @@ depends=("${MINGW_PACKAGE_PREFIX}-codec2"
...
-             "${MINGW_PACKAGE_PREFIX}-python-gobject"
              "${MINGW_PACKAGE_PREFIX}-python-packaging"
              "${MINGW_PACKAGE_PREFIX}-python-pip"
-             "${MINGW_PACKAGE_PREFIX}-python-pyqt5"
-             "${MINGW_PACKAGE_PREFIX}-python-pyqtgraph"
-             "${MINGW_PACKAGE_PREFIX}-python-pytest" # only required for tests
-             "${MINGW_PACKAGE_PREFIX}-zeromq")
-checkdepends=("${MINGW_PACKAGE_PREFIX}-python-jsonschema"
-              "${MINGW_PACKAGE_PREFIX}-python-setuptools")
+             "${MINGW_PACKAGE_PREFIX}-python-pytest") # only required for tests
```
ognevny, line 63: add it in `checkdepends` then

chrisjohgorman: I'll try moving it, although I'm not 100% sure that it isn't needed in `makedepends`.  That's where it is in the arch package, although we diverge from arch in a few ways.  I'll try it out on my home machine and let you know what I get.

mingw-w64-gnuradio/PKGBUILD
```diff
@@ -38,37 +39,35 @@ depends=("${MINGW_PACKAGE_PREFIX}-codec2"
...
+  "${MINGW_PACKAGE_PREFIX}-spdlog"
+  "${MINGW_PACKAGE_PREFIX}-zeromq")
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja"
-             "${MINGW_PACKAGE_PREFIX}-boost"
              "${MINGW_PACKAGE_PREFIX}-cppzmq"
              "${MINGW_PACKAGE_PREFIX}-doxygen"
-             "${MINGW_PACKAGE_PREFIX}-gtk3"
-             "${MINGW_PACKAGE_PREFIX}-libiio"
```
mmuetzel, line 56: According to the package grokker, gnuradio depends directly on libiio:
```
  /mingw64/bin/libgnuradio-iio.dll:
  	libgcc_s_seh-1.dll => D:\M\msys64\mingw64\bin\libgcc_s_seh-1.dll (0x00000125569d0000)
  	libwinpthread-1.dll => D:\M\msys64\mingw64\bin\libwinpthread-1.dll (0x0000012556a00000)
  	libstdc++-6.dll => D:\M\msys64\mingw64\bin\libstdc++-6.dll (0x0000012556b70000)
  	libad9361.dll => D:\M\msys64\mingw64\bin\libad9361.dll (0x00000125569d0000)
  	libboost_thread-mt.dll => D:\M\msys64\mingw64\bin\libboost_thread-mt.dll (0x00000125569d0000)
  	libfmt.dll => D:\M\msys64\mingw64\bin\libfmt.dll (0x00000125569d0000)
  	libiio.dll => D:\M\msys64\mingw64\bin\libiio.dll (0x0000012556a30000)
```

So, even if the package "happens" to be pulled in by another dependency currently, it might make sense to keep it here as a direct dependency.
Only if the dependency is only indirect, it should be removed.

It's probably not a big deal either way. But if you decide to put it in again, it should be moved to the `depends` list.

(I didn't check any of the other dependencies here.)

chrisjohgorman: I hummed and hawed about whether or not to include indirect dependencies, and thought it best to remove them.  Perhaps I did this in haste.  I have no problems adding them back on the `depends` list.  It looks like I'll need to move `python-pytest` per @ognevny 's observation, so one more change won't hurt.

mmuetzel: You shouldn't add indirect dependencies. But adding direct dependencies would probably be a good idea even if they would be pulled in via other dependent packages.
I'm not sure how the package grokker picks the packages it checks on an update. But it would probably help if all direct dependencies appear in the `depends` list.

chrisjohgorman: My bad, I'm not communicating properly here.  What I meant by indirect dependencies were dependencies pulled in by other dependent packages.  In this case, specifically, `fmt`, `libiio` and `python-cairo`.  These are all pulled in by other packages at the moment and I have a pending change to re add them to the `depends` list.  Thanks again both Markus and Maksim for your time on this.

chrisjohgorman:
I have tested both `ucrt64` and `mingw64` packages and they install and run on a fresh install of msys2.  These are ready for review.

PR #21579, opened 2024-08-06
[new package] rust-analyzer 20240805

ognevny:
~and add `rust` as `rust-src` dependency~


ognevny:
let me know if someone needs i686 rust-analyzer

PR #21580, opened 2024-08-06
suitesparse: update to 7.8.0

mmuetzel:


mmuetzel:
I downloaded and installed the MINGW64 build artifact from this PR. With it, all relevant tests in Octave still pass.

PR #21581, opened 2024-08-06
rust: don't generate debuginfo on i686

ognevny:
it takes too much memory to do that


ognevny:
I also backported change from #21579

jeremyd2019:
I think that makepkg is stripping rust, so I don't know what good building with extra debug info does when it's just stripped back out when packaging.  Maybe you want to enable `options=('strip' 'debug')` and get split debug packages?  Note: if you try that, please check that CFLAGS are still what you expect, I am kind of concerned doing that might change optimization.

Issue #16833, opened 2023-04-17
Integrate Control Flow Guard into MSYS2


lazka:
This is just a meta issue to keep track of things:

* [x] mingw-w64: enable CFG https://github.com/msys2/MINGW-packages/pull/16827
* [x] add integration tests https://github.com/msys2/msys2-tests/pull/34
* [ ] teach peflags to show/set cfg https://github.com/msys2/MSYS2-packages/pull/3733
* [ ] consider adding a build option to PKGBUILD, so packages can opt-in (and later opt-out if we want to)
* [ ] document CFG in the docs
   * how to enable it
   * how to see if it's enabled (peflags, genpeimg, llvm-readobj --coff-load-config)
   * how mixing cfg/non-cfg works
* [ ] announce the support

ognevny:
we can enable CFG in rustc since it's supported in Windows 10+ for a long time

mati865:
@Andarwinux without any context, it's hard to understand what you mean. If there are issues with mingw-w64 that prevent CFG from working correctly, please report them upstream.

PR #21598, opened 2024-08-09
rust: update to 1.80.1

ognevny:
- add debuginfo only if `debug` option is set (see https://github.com/msys2/MINGW-packages/pull/21581#issuecomment-2276354980)
- add missing wasm32-unknown-unknown target
- remove self-contained directory for all targets
- `rust`: remove license installation as `rustdoc` is now a part of main package (it installs license files in doc directory)
- `rust-wasm`: don't strip binaries, add missing `wasi-libc` dependency
- enable frame-pointers ~and CFG (addresses #16833)~ in config.toml

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -19,8 +19,8 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-rust-wasm")
          "${MINGW_PACKAGE_PREFIX}-rust-src")
-pkgver=1.80.0
-pkgrel=4
+pkgver=1.80.1
+pkgrel=1
 pkgdesc="Systems programming language focused on safety, speed and concurrency (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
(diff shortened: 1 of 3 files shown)


ArchGuyWu:
We may still need to set wasi-sysroot for all wasm32-wasip* targets.

ArchGuyWu:
But is it possible to only disable self-contained for target $OSTYPE?

ognevny:
> But is it possible to only disable self-contained for target $OSTYPE?

I pushed the change, but it'll probably not work because the environment variable I used usually generated for `build.rs`...

ArchGuyWu:
try TARGET env: https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/environment-variables.md#environment-variables-cargo-sets-for-build-scripts

ognevny:
> try TARGET env: https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/environment-variables.md#environment-variables-cargo-sets-for-build-scripts

it's also set for build.rs

mingw-w64-rust/config.toml
```diff
@@ -42,9 +42,11 @@ codegen-units-std = 1
 channel = "stable"
 description = "Rev%PKGREL%, Built by MSYS2 project"
 rpath = false
+frame-pointers = true
```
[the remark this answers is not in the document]
ognevny: ok, CFG will work fine, right?

mingw-w64-rust/PKGBUILD
```diff
@@ -266,6 +265,8 @@ package_rust-wasm() {
   depends=("${MINGW_PACKAGE_PREFIX}-lld"
            "${MINGW_PACKAGE_PREFIX}-rust"
            "${MINGW_PACKAGE_PREFIX}-wasm-component-ld")
+  # stripping files fail
```
mati865, line 268: What does fail? Maybe it happens only in GNU environments as we had issues with the Binutils strip somewhere else.

ognevny: objdump and strip throw errors 'unsupported object file'

mingw-w64-rust/config.toml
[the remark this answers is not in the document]
mati865: Sorry, I have confused function sections with frame pointers. Upstream has already enabled it for the standard library so it's fine: https://github.com/rust-lang/rust/blob/899eb03926be23f2e5d2ffcaa1d6f9ac40af7f13/src/bootstrap/src/core/build_steps/compile.rs#L593
Although I don't see how enabling it for the compiler itself would be helpful.

mingw-w64-rust/PKGBUILD
mati865: Perhaps they encounter wasm objects and bail out. I think LLVM tools should be fine out of the box.

ognevny: I saw the same errors for CLANG64

mingw-w64-rust/config.toml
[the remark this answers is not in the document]
ognevny: > Sorry, I have confused function sections with frame pointers. Upstream has already enabled it for the standard library so it's fine

no problem

alvinhochun:
> CFG

I afraid it may not be that simple: https://github.com/rust-lang/rust/pull/74103

mingw-w64-rust/PKGBUILD
ArchGuyWu: There is a package called wabt in archlinux repo, which has wasm-strip, I'll add it to msys2.

ognevny:
> I afraid it may not be that simple: https://github.com/rust-lang/rust/pull/74103

that means that it can't be enable at the moment. I'll comment it in config.toml so we can mlve back to it in future

ognevny:
rust-wasm now compiles with self-contained objects properly and `_start` symbol isn't stripped anymore!

PR #21601, opened 2024-08-09
libheif: Update to 1.18.2

lazka:


lazka:
https://github.com/strukturag/libheif/issues/1262

PR #21608, opened 2024-08-10
[new package] wasmer 4.3.5

ognevny:


ognevny:
tun tap patch is taken from https://github.com/vorner/tuntap/pull/14

ArchGuyWu:
- https://github.com/wasmerio/wasmer/blob/c0122bc62ba162087660f69bb0f187a3c3633030/lib/vm/src/probestack.rs#L42
- https://github.com/rust-lang/compiler-builtins/blob/master/src/x86_64.rs#L16

ognevny:
> - https://github.com/wasmerio/wasmer/blob/main/lib/vm/src/probestack.rs#L45
> - https://github.com/rust-lang/compiler-builtins/blob/master/src/x86_64.rs#L16

I see that they have different names

ArchGuyWu:
> > * https://github.com/wasmerio/wasmer/blob/main/lib/vm/src/probestack.rs#L45
> > * https://github.com/rust-lang/compiler-builtins/blob/master/src/x86_64.rs#L16
> 
> I see that they have different names

Yes, I Googled it, and ___chkstk_ms is the one for mingw.

ognevny:
> > > * https://github.com/wasmerio/wasmer/blob/main/lib/vm/src/probestack.rs#L45
> Yes, I Googled it, and ___chkstk_ms is the one for mingw.

so patching this line will help?

ArchGuyWu:
> > > > * https://github.com/wasmerio/wasmer/blob/main/lib/vm/src/probestack.rs#L45
> > > >   Yes, I Googled it, and ___chkstk_ms is the one for mingw.
> 
> so patching this line will help?

Don't know, but it's worth to try.

mingw-w64-wasmer/mingw-build.patch
```diff
@@ -0,0 +1,72 @@
...
++++ b/lib/vm/src/probestack.rs
+@@ -42,7 +42,7 @@ cfg_if::cfg_if! {
+             // ___chkstk (note the triple underscore) is implemented in compiler-builtins/src/x86_64.rs
+             // by the Rust compiler for the MinGW target
+             #[cfg(all(target_os = "windows", target_env = "gnu"))]
+-            pub fn ___chkstk();
++            pub fn ___chkstk_ms();
+         }
+         /// The probestack for Windows when compiled with GNU
+         pub const PROBESTACK: unsafe extern "C" fn() = ___chkstk;
```
ArchGuyWu, line 72: This may need to replace, too.

ognevny: yes, it now throws an error

ArchGuyWu:
It's usable.

ArchGuyWu:
We can use LLVM as [native backend](https://docs.wasmer.io/runtime).
Build doc: https://github.com/wasmerio/wasmer/blob/56894ff7b73dde5555675a8b11ce75363281b2a7/docs/BUILD.md?plain=1#L65
The least LLVM version requirement: https://github.com/wasmerio/wasmer/tree/main/lib/compiler-llvm#requirements

ArchGuyWu:
This might also help in getting rid of some patches.

ognevny:
> This might also help in getting rid of some patches.

which? tun-tap should be with up-to-date tokio so it doesn't depend on very old windows bindings. but hardcoding `IS_WINDOWS` probably can be dropped

ArchGuyWu:
> > This might also help in getting rid of some patches.
> 
> which? tun-tap should be with up-to-date tokio so it doesn't depend on very old windows bindings. but hardcoding `IS_WINDOWS` probably can be dropped

Ok, and use llvm from msys2 to compile it maybe better.

ognevny:
x86_64 Windows should with everything, the problems can be only for aarch64

ArchGuyWu:
> x86_64 Windows should work with everything, the problems can be only for aarch64
> 
> comment from Makefile
> ```
> # Here is what works and what doesn't:
> #
> # * Cranelift works everywhere,
> #
> # * LLVM works on Linux+Darwin/`amd64`,
> #   and linux+`aarch64`, linux+`riscv`
> #   but it doesn't work on Darwin/`aarch64` or Windows/`aarch64`.
> #
> # * Singlepass works on Linux+Darwin+Windows/`amd64`,
> #   and Linux+Darwin/`aarch64`
> #   it doesn't work on */`riscv`.
> #
> # * Windows isn't tested on `aarch64`, that's why we consider it's not
> #   working, but it might possibly be.
> # * The Only target for `riscv` familly of processor is the RV64, with the `GC` extensions
> ```

OK.

ArchGuyWu:
We can get rid of [mingw-build.patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-wasmer/mingw-build.patch) since the next version:. https://github.com/wasmerio/wasmer/pull/5017#event-13927079535

PR #21618, opened 2024-08-11
libdovi: Update to 3.3.1

lazka:


ognevny:
please, add `MSYS2_ARG_CONV_EXCL` for `cargo cbuild` so workaround wouldn't be required

lazka:
thanks

PR #21621, opened 2024-08-11
libtool: update to 2.5.1

ikspress:
Thanks for lazka's patience.


ikspress:
A accidental thing, I have encountered a failure in compiling mingw-w64-gcc with libtool v2.4.7 and v2.5.1. ld just returned an error code 127 (or others, I've forgot). Then I can't reproduce this error anymore. Could anyone please attempt to compile gcc in their local machine?

PR #21633, opened 2024-08-12
binutils: update to 2.43

ikspress:


mingw-w64-binutils/PKGBUILD
```diff
@@ -140,12 +140,7 @@ package() {
   make DESTDIR=${pkgdir} install
 
   # https://github.com/msys2/MINGW-packages/issues/7890
-  rm "${pkgdir}${MINGW_PREFIX}/lib/bfd-plugins/libdep.a"
-
-  find ${pkgdir}${MINGW_PREFIX}/share -type f -iname "opcodes.mo" -o -iname "bfd.mo" | xargs -rtl1 rm
+  rm -f "${pkgdir}${MINGW_PREFIX}/lib/bfd-plugins/libdep.a"
```
lazka, line 143: Why -f?

mingw-w64-binutils/PKGBUILD
```diff
@@ -140,12 +140,7 @@ package() {
   make DESTDIR=${pkgdir} install
 
   # https://github.com/msys2/MINGW-packages/issues/7890
-  rm "${pkgdir}${MINGW_PREFIX}/lib/bfd-plugins/libdep.a"
-
-  find ${pkgdir}${MINGW_PREFIX}/share -type f -iname "opcodes.mo" -o -iname "bfd.mo" | xargs -rtl1 rm
```
lazka, line 145: These files are shipped by gdb currently. We can't have them in both packages.

mingw-w64-binutils/2001-ld-option-to-move-default-bases-under-4GB.patch
```diff
@@ -138,5 +128,16 @@ index c6fd8b8f8d9..312f8d6d566 100644
...
+diff --git a/ld/ldlex.h b/ld/ldlex.h
+index defe3fcbbb9..6c5e631171b 100644
+--- a/ld/ldlex.h
++++ b/ld/ldlex.h
+@@ -389,6 +389,8 @@ enum option_values
+   OPTION_DISABLE_NO_BIND,
+   OPTION_DISABLE_WDM_DRIVER,
+   OPTION_DISABLE_TERMINAL_SERVER_AWARE,
++  OPTION_DEFAULT_IMAGE_BASE_LOW,
++  OPTION_DEFAULT_IMAGE_BASE_HIGH,
```
lazka, line 140: nitpick: they should be added below the `/* Used by emultempl/pep.em.  */` comment

lazka:
otherwise lgtm

mingw-w64-binutils/PKGBUILD
ikspress: It failed when configured with `--enable-shared`.

lazka: But we are not using that. but OK.

downside is we don't notice when the file path changes

PR #21643, opened 2024-08-13
rufus: enable ucrt64

ognevny:
test target is not defined, removing


Biswa96:
Would it be possible to link with system libcdio https://packages.msys2.org/base/mingw-w64-libcdio ? Then that patch may not be required.

ognevny:
> Would it be possible to link with system libcdio https://packages.msys2.org/base/mingw-w64-libcdio ? Then that patch may not be required.

it's not supported upstream AFAIK

ognevny:
the title was totally wrong (it should be ucrt64), sorry

PR #21651, opened 2024-08-15
codeblocks: Enable wxSmith plugin

### LINKED ISSUE #21650: CodeBlocks package does not have wxSmith Plugin
The CodeBlocks package does not have wxSmith Plugin; reason it fails to build.
I am going to try to fix that by a round about method because the Makefile.am build system is too complex for me to fix it that way.
So, I plan to create or edit an MSys2 MinGW package to provide the cbp2make command.
Then, I will use cbp2make in an second package codeblocks-wxsmith that uses cbp2make to create a makefile to build wxsmithlib.dll and the wxsmith plugin. Likely be weeks to months before I am done.

Edit: The options for cbp2make are very complex looking

Tim S.

- [x] cbp2make package
- [x] wxsmith cbp file created
- [ ] makefile created using wxsmith cbp file and cbp2make
- [ ] wxsmith package created

Biswa96:
Fixes #21650

### DIFF
--- a/mingw-w64-codeblocks/001-makefile-wxsmith-add-no-undefined.patch
+++ b/mingw-w64-codeblocks/001-makefile-wxsmith-add-no-undefined.patch
@@ -1,0 +1,11 @@
+--- a/src/plugins/contrib/wxSmith/Makefile.am
++++ b/src/plugins/contrib/wxSmith/Makefile.am
+@@ -7,7 +7,7 @@
+ 
+ lib_LTLIBRARIES = libwxsmithlib.la
+ 
+-libwxsmithlib_la_LDFLAGS = -version-info 0:1:0 -shared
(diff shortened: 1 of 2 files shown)


stahta01:
wxSmith plugin fails to load under CLANG64; going to try UCRT64 next.

Biswa96:
> I am going to remove the grep/sed line and see if I can fix the wxSmith build error.

It is not required. I did not notice that one DLL is installed in lib/ instead of bin/

Biswa96:
All the wxSmith plugins are loaded correctly.

image

stahta01:
Both UCRT64 and CLANG64 plugins loaded for me. I used the UCRT64 plugin to open some wxs files in an CB project. And, then built the changes; the changes looked okay to me. So, I think this PR is working okay.

PR #21659, opened 2024-08-15
eza: add mingw64

srouquette:
until git-for-windows is built on something else that mingw64, I guess we're stuck...


Biswa96:
Just a little trick, you could also do something like `alias eza='/ucrt64/bin/eza'` in bash to make eza available in any mingw environment.

srouquette:
@Biswa96 thanks for the heads up. would you recommand using UCRT64 by default and install installing git-for-windows only as MINGW64 and do an alias the same way?

I check the [environments](https://www.msys2.org/docs/environments/) and I was wondering if there's a comparison page somewhere ?

PR #21682, opened 2024-08-17
Rebuild some packages without python-wheel

lazka:
it should not longer be needed with newer setuptools


Biswa96:
Does removing python-wheel from makedepends change anything in the package? In other words, is pkgver required in this case?

lazka:
It should not change anything, this was just a test to see if there are any file changes etc.

Are you suggesting that we just drop it for everything else without rebuilding?

Biswa96:
> Are you suggesting that we just drop it for everything else without rebuilding?

Nope, just asking.

PR #21683, opened 2024-08-18
wasm-pack: update to 0.13.0

ognevny:
curl isn't required there


ognevny:
with rust-wasm as dependency it's usable

PR #21684, opened 2024-08-18
librsvg: Update to 2.59.1

lazka:
so we can move to meson, and don't have to deal with libtool at least


mingw-w64-librsvg/0002-avoid-link-whole.patch
```diff
@@ -0,0 +1,20 @@
...
++    link_with: librsvg_c_lib,
+     link_args: link_args,
+     link_depends: librsvg_ver,
+     dependencies: library_dependencies,
+@@ -342,7 +342,7 @@
+       librsvg_gir = shared_library(
+         'rsvg-gir',
+         rsvg_dummy,
+-        link_whole: librsvg_c_lib,
++        link_whith: librsvg_c_lib,
```
ognevny, line 17: typo

lazka: oops, thanks

ognevny:
also PKGBUILD L17: change autotools to meson :)

lazka:
I'll try to upstream the gdk-pixbuf patch once this builds.

lazka:
Two issues:

* cargo-c uses the wrong linker for the static build, worked around by not linking anything
* binutils strip fails to strip the static archive, which is now huge

lazka:
https://gitlab.gnome.org/GNOME/librsvg/-/merge_requests/1015

ognevny:
patch for clang is not pushed

lazka:
> binutils strip fails to strip the static archive, which is now huge

This is still not solved, but having a newer version is more important, so I'm merging this.

I'll try to upstream the clang patch later.

lazka:
> I'll try to upstream the clang patch later.

https://gitlab.gnome.org/GNOME/librsvg/-/merge_requests/1052

PR #21687, opened 2024-08-19
SWI-Prolog: new version 9.2.6

mgondan:
I have split the package into the components defined by cmake_install.cmake (core, packages, archive etc.)


lazka:
Your call, but doesn't this just complicate things for not much gain (not splitting up any large dependency chains, nor reducing install size much)?

mgondan:
I was basically trying to align the package organization a bit with the one on debian. But your argument still holds.

The dependencies can be split up, e.g. the dependency on libarchive be moved to swi-prolog-archive etc. That was the plan for the next release.

mgondan:
Moved the dependencies to the subpackages.

mgondan:
Thank you for the fixes and amendments, very instructive

Issue #20407, opened 2024-03-22
node rollup does not work


Kreijstal:
### Description / Steps to reproduce the issue

require node rollup
use `rollup -c` will give you an esoteric error, but it is clear that it was made with msvc in mind, we can see this, because when we `npm install` `rollup` we get: https://www.npmjs.com/package/@rollup/rollup-win32-x64-msvc ,
Not sure if either npm would have to be patched, or rollup would be just patched and installed as a msys2 package.. (If that can be figured out).
Anyway, also reporting this for future viewers. And people that happen to come across this https://github.com/tauri-apps/tauri/issues/6685

### Expected behavior

gives a meaningful error or it works

### Actual behavior

```
$ npm run build

> js-interpreter-npm@1.0.8 build
> rollup -c

thread '<unnamed>' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\napi-sys-2.3.0\src\functions.rs:7:3:
Must load N-API bindings
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-22631 ayylmai 3.4.10.x86_64 2024-02-10 08:39 UTC x86_64 Msys

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [X] CLANGARM64

### Are you willing to submit a PR?

if I figure out a solution

Kreijstal:
Alternatively, it seems webpack has no trouble with it.

raedrizqie:
it is not about npm

Kreijstal:
going through the dependency trail now attempting to make it work with napi-rs. https://github.com/napi-rs/napi-rs/issues/2001

``` 
Load Node-API [napi_get_last_error_info] from host runtime failed: GetProcAddress failed
Load Node-API [napi_get_uv_event_loop] from host runtime failed: GetProcAddress failed
Load Node-API [napi_fatal_exception] from host runtime failed: GetProcAddress failed
Load Node-API [napi_create_threadsafe_function] from host runtime failed: GetProcAddress failed
Load Node-API [napi_create_date] from host runtime failed: GetProcAddress failed
[...]

PR #21691, opened 2024-08-19
typstudio: update to 0.1.0dev.33

ognevny:


raedrizqie:
rollup needs to be fixed first.. see #20407

PR #21698, opened 2024-08-20
Add Libkqueue Package

johnathan-becker:
This MR adds a libkqueue package to MINGW-packages


Biswa96:
Do you want 2.6.2 tag or any specific commit?

johnathan-becker:
> Do you want 2.6.2 tag or any specific commit?

I am fine with using the commit from v2.6.2 or just using the v2.6.2 tag. Which would you prefer?

Biswa96:
There are many issues in this package. I shall try to fix those in separate pull request. I am merging this as is for now.

PR #21702, opened 2024-08-21
wasm-bindgen: update lock file

lazka:
link upstream issue re missing lock file

because github complains that it contains outdated dependencies, but the relevant packages can't be updated, so this wont help.

But update anyway while I'm already at it.


lazka:
@ognevny ok?

ognevny:
> because github complains that it contains outdated dependencies, but the relevant packages can't be updated, so this wont help.

where I can see such logs?

ognevny:
> @ognevny ok?

ok anyway

lazka:
image

PR #21729, opened 2024-08-23
[new package] wezterm 20240203.110809.5046fc22

ognevny:
@Kreijstal

### DIFF
--- a/mingw-w64-wezterm/PKGBUILD
+++ b/mingw-w64-wezterm/PKGBUILD
@@ -1,0 +1,63 @@
+# Maintainer: Maksim Bondarenkov <maksapple2306@gmail.com>
+
+_realname=wezterm
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+_basever=20240203.110809.5046fc22
+pkgver=20240203.110809.5046fc22+r325+g30345b36d
+pkgrel=1
+pkgdesc="A GPU-accelerated cross-platform terminal emulator and multiplexer (mingw-w64)"
+arch=('any')
+mingw_arch=('ucrt64' 'clang64')
+url="https://wezfurlong.org/wezterm/"
+msys2_repository_url='https://github.com/wez/wezterm'
+license=('spdx:MIT')
+depends=("${MINGW_PACKAGE_PREFIX}-openssl"
+         "${MINGW_PACKAGE_PREFIX}-libssh"
+         "${MINGW_PACKAGE_PREFIX}-libgit2")
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
+             "${MINGW_PACKAGE_PREFIX}-pkgconf"
+             "${MINGW_PACKAGE_PREFIX}-zstd"
+             "${MINGW_PACKAGE_PREFIX}-openssl"
+             'git')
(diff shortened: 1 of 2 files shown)


mingw-w64-wezterm/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+        "mingw-build.patch")
+sha256sums=('4b893f15c5330ee40aee2c341028234fb0a78825e0d7207ff43dfd9aeed71799'
+            '4daf44978645a72b3de7a9f972141db809aa4a607265ed3b19b3210d867fa636')
+
+prepare() {
+  cd "${_realname}"
+
+  patch -Np1 -i "${srcdir}/mingw-build.patch"
+
+  git submodule update --init --recursive
```
Kreijstal, line 33: isnt it better to manually get the dependencies yourself in sources?

ognevny: what's wrong with this method?

Kreijstal: you can't verify the sources, it is not reproducible if the sources change, no?
isnt that why archlinux repos do it like that?, so that it is verifiable and reproducible?

ognevny: in Arch commit is not specified, and checksums are `'SKIP'` for all sources

ognevny:
изображение

Kreijstal:
> изображение

Is this a censored version of your shell or is it a bug?
this is how mine looks
grafik
but this is from my own modified pkgbuild

ognevny:
I don't know, just what is shown in my setup

ognevny:
> but this is from my own modified pkgbuild

can you share me an output from `tar -tvf *.pkg.*`?

Kreijstal:
> > but this is from my own modified pkgbuild
> 
> can you share me an output from `tar -tvf *.pkg.*`?

```
# tar -tvf *.pkg.*
-rw-r--r-- root/root     37351 2024-08-23 09:30 .BUILDINFO
-rw-r--r-- root/root       365 2024-08-23 09:30 .MTREE
-rw-r--r-- root/root       493 2024-08-23 09:30 .PKGINFO
drwxr-xr-x root/root         0 2024-08-23 09:30 ucrt64/
drwxr-xr-x root/root         0 2024-08-23 09:30 ucrt64/bin/
-rwxr-xr-x root/root  73220181 2024-08-23 09:30 ucrt64/bin/wezterm-gui.exe
-rwxr-xr-x root/root  34267182 2024-08-23 09:30 ucrt64/bin/wezterm.exe
```
you need libgit2 and libssh thats it

ognevny:
> you need libgit2 and libssh thats it

they're already in `depends`

Kreijstal:
I tried your package in my computer and got the same bug, maybe it is because it is not in master branch?

There is a bug with wezterm ssh on windows tho

image

left `wezterm ssh`, right ssh in wezterm, not sure if this is because of msys2 or a wezterm bug. oh well.

Kreijstal:
Observations: sixel only works when doing ./wezterm ssh, doesn't work when doing wezterm and then ssh.
image
for the original build.
And for my own build.
image
but it seems it still has the newline error problem.

Kreijstal:
Let me check today

Kreijstal:
Hi, it looks really stable. I think it is ready!

ognevny:
ok, I'll turn back to versioned tag after a new release

PR #21731, opened 2024-08-23
Add mingw-w64-i686-python-reportlab

manfromarce:
I need this package for x86 as a dependency for another, and it successfully builds when adding mingw32 to the PKGBUILD, all the dependencies for it are already available for x86.


Biswa96:
Why do you need the x86 version of that another project?

lazka:
And just curious, which project needs this?

manfromarce:
Yes, I don't necessarily need this anymore

PR #21739, opened 2024-08-25
ncurses: Update to 6.5

lazka:


lazka:
Got handled in #21868

PR #21791, opened 2024-08-30
libspelling: update to 0.3.0

taozuhong:


lazka:
We don't package unstable releases without a good reason.

taozuhong:
Got it, I marked it as draft.

Issue #21768, opened 2024-08-28
long double in clang msys2


wmurw:
### Description / Steps to reproduce the issue

package mingw-w64-clang-x86_64-toolchain
clang version 18.1.8 
code:
```c++
cout << "sizeof(double) = " << sizeof(double) << " --- ";
cout << std::numeric_limits<double>::max() << "\n";
cout << "sizeof(long double) = " << sizeof(long double) << " --- ";
cout << std::numeric_limits<long double>::max() << "\n";
```
it's coming out:
```
sizeof(double) = 8 --- 1.79769e+308
sizeof(long double) = 16 --- 2.9972e-312
```
instead of `std::numeric_limits<long double>::max()` outputs something in between `::min()` and `::denorm_min() `
`long double` takes 16 bytes, but uses 64 bits. Does not use 80 bits, even with a key `--target=x86_64-w64-mingw32`.
clang from [https://github.com/llvm/llvm-project]( https://github.com/llvm/llvm-project ) under windows by default uses headers from msvc and uses 64 bit long double,
but with `--target=x86_64-w64-mingw32` it uses 80 bits. (`::max()` = 1.18973e+4932)

### Expected behavior

coming out:
```
sizeof(double) = 8 --- 1.79769e+308
sizeof(long double) = 16 --- 1.18973e+4932
```
mingw-w64-x86_64-toolchain gives this exact result.

### Actual behavior

```
sizeof(double) = 8 --- 1.79769e+308
sizeof(long double) = 16 --- 2.9972e-312
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-19045

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

lazka:
This should be fixed now via #21796. Can you confirm?

lazka:
your example works here now, so closing.

Issue #17795, opened 2023-07-18
cmake on clang32 reports wrong configuration time


madebr:
### Description / Steps to reproduce the issue

Recent cmake releases reports the time it takes to configure a project in seconds at the end.
It looks like on clang32, this is somehow broken.

For example, see this output of a GitHub workflow:
https://github.com/libsdl-org/SDL/actions/runs/5589750584/jobs/10218501241#step:9:271

### Expected behavior

```
 -- Configuring done (32.9s)
-- Generating done (0.1s)
-- Build files have been written to: D:/a/SDL/SDL/build
```

### Actual behavior

```
 -- Configuring done (-9123292659426253850436719799017442136586774209552161337536467353746771162918819034520641703491042846408815082120540550597342102467005169222967783506967346641314978059609663676974096259599798996998937607570840069391076737937496429167352331447416802841585171636136443462884803862862218511601527855135072452608.0s)
-- Generating done (-0.0s)
-- Build files have been written to: D:/a/SDL/SDL/build
```

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

windows-latest GitHub runner

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

### Are you willing to submit a PR?

I have no experience with clang :(

Biswa96:
Thank you for reporting this issue. I have reproduced the issue with the following example. The code is based on the commit which calculates the time difference after cmake configure https://github.com/Kitware/CMake/commit/5f0c5ec49b553cbe4482d9a8cf7e8889dc90e759

```cpp
#include <chrono>
#include <iomanip>
#include <iostream>
#include <thread>

int main()
{
    auto startTime = std::chrono::steady_clock::now();
    std::this_thread::sleep_for(std::chrono::milliseconds(3000));
    auto endTime = std::chrono::steady_clock::now();
    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(endTime - startTime);
    std::cout << "Configuring done (" << std::fixed << std::setprecision(1) << ms.count() / 1000.0L << "s)\n";
}
```

* UCRT64:
```
$ g++ test.cpp && ./a.exe && echo $?
Configuring done (3.0s)
0
```

* CLANG64

```
[...]

PR #21796, opened 2024-08-30
libc++: Pass global CPPFLAGS to CMake via CXXFLAGS

mstorsjo:
CMake doesn't react to the global env variable CPPFLAGS; feed the contents of it to CMake via CXXFLAGS.

This makes sure that libc++ is built with -D__USE_MINGW_ANSI_STDIO=1, which fixes the formatting output of long doubles on x86.

This should fix issues #21768 and #17795.

### DIFF
--- a/mingw-w64-libc++/PKGBUILD
+++ b/mingw-w64-libc++/PKGBUILD
@@ -12,7 +12,7 @@ _version=18.1.8
 _rc=""
 _tag=llvmorg-${_version}${_rc}
 pkgver=${_version}${_rc/-/}
-pkgrel=1
+pkgrel=2
 arch=(any)
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
 url="https://libcxx.llvm.org/"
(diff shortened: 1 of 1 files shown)


mstorsjo:
CC @mati865 @jeremyd2019 @Biswa96 @lazka 

This obviously doesn't fix the root cause, where the makepkg global CPPFLAGS isn't picked up by any of the PKGBUILD files that build things with CMake, but I think this might be an important enough special case to warrant fixing directly.

Biswa96:
> where the makepkg global CPPFLAGS isn't picked up by any of the PKGBUILD

In that case, should the CPPFLAGS be added with CXXFLAGS in mingw_makepkg.conf?

mstorsjo:
> > where the makepkg global CPPFLAGS isn't picked up by any of the PKGBUILD
> 
> In that case, should the CPPFLAGS be added with CXXFLAGS in mingw_makepkg.conf?

In a generic case, `CPPFLAGS` should be used for both C and C++, so it would have to be added to both `CFLAGS` and `CXXFLAGS`. Whether that's the right thing to do overall in msys2, I'd leave up to you guys though.

lazka:
Arch also moved away from using CPPFLAGS (see https://gitlab.archlinux.org/archlinux/rfcs/-/blob/master/rfcs/0003-buildflags.rst) which we copied for `_FORTIFY_SOURCE` https://github.com/msys2/MSYS2-packages/blob/09f5a74e9745bfac184a9566ef1b167486c44d0f/pacman/makepkg_mingw.conf#L58 so probably worth a try. I've created https://github.com/msys2/MSYS2-packages/issues/4860

lazka:
Thanks!

lazka:
Added a regression test in https://github.com/msys2/msys2-tests/pull/72

PR #21814, opened 2024-09-01
neocmakelsp: init at 0.8.1

### LINKED ISSUE #21813: Add neocmakelsp
### Package name

https://github.com/neocmakelsp/neocmakelsp

Another cmake lsp 

https://github.com/neocmakelsp/neocmakelsp

neocmakelsp --version

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [X] CLANGARM64

_No response_

Freed-Wu:
Fix #21813

### DIFF
--- a/mingw-w64-neocmakelsp/PKGBUILD
+++ b/mingw-w64-neocmakelsp/PKGBUILD
@@ -7,40 +7,36 @@ pkgver=0.8.1
 pkgrel=1
 pkgdesc='CMake LSP implementation based on Tower and Tree-sitter'
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt32' 'ucrt64' 'clang32' 'clang64' 'clangarm64')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url='https://github.com/Decodetalkers/neocmakelsp'
 license=('MIT')
-makedepends=("${MINGW_PACKAGE_PREFIX}-rust")
-source=("${_realname}-v${pkgver}.tar.gz::https://github.com/Decodetalkers/neocmakelsp/archive/refs/tags/v${pkgver}.tar.gz")
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
+             "${MINGW_PACKAGE_PREFIX}-meson"
+             "${MINGW_PACKAGE_PREFIX}-ninja")
+source=("https://github.com/Decodetalkers/neocmakelsp/archive/v${pkgver}/${_realname}-${pkgver}.tar.gz")
 sha256sums=('23d2fd6f6bd0152dad9b6bdf9b5d6932d97ccd106bfb47d6d6fee563ea5a7eec')
 
 prepare() {
-  cd "${srcdir}/${_realname}-${pkgver}"
+  cd "${_realname}-${pkgver}"
 
   cargo fetch --locked --target "$(rustc -vV | sed -n 's/host: //p')"
 }
 
 build() {
-  cd "${srcdir}/${_realname}-${pkgver}"
+  MSYS2_ARG_CONV_EXCL="--prefix=" \
+    meson setup --prefix="${MINGW_PREFIX}" "build-${MSYSTEM}" "${_realname}-${pkgver}"
 
-  cargo build --release --frozen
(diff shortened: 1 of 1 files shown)


ognevny:
please, follow a [template](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.cargo-tarball.sh)

mingw-w64-neocmakelsp/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+url='https://github.com/Decodetalkers/neocmakelsp'
+license=('MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust")
+source=("${_realname}-v${pkgver}.tar.gz::https://github.com/Decodetalkers/neocmakelsp/archive/refs/tags/v${pkgver}.tar.gz")
+sha256sums=('23d2fd6f6bd0152dad9b6bdf9b5d6932d97ccd106bfb47d6d6fee563ea5a7eec')
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+
+  # if cargo wants to make an http request at build stage, use `cargo fetch --locked` instead
```
ognevny, line 20: the comment isn't needed

ognevny:
I've pushed a change so meson is used for build and install. it adds completions to the package

ognevny:
how `&&` is passed as argument?

ognevny:
> So I use cargo not meson to avoid this problem :smiling_face_with_tear:

that's bug, not a problem

Biswa96:
How about using meson and cargo both? I'll just add a patch to disable the cargo build in meson and build the program with cargo only. Then install other stuff using meson.

I am merging this as is and shall create a new PR with the changes.

Decodetalkers:
Emm.. sorry, I copy the meson build from fractal, whose meson build use '&&', and it always works on my archlinux. I think I have fixed it in master.

Biswa96:
> Emm.. sorry, I copy the meson build from fractal, whose meson build use '&&', and it always works on my archlinux. I think I have fixed it in master.

The cp command does not work in Command Prompt. Powershell has an alias `cp` pointing to `copy-item`.

Biswa96:
Also, for some situations like cross compiling or using certatin toolchain, some options need to be passed to cargo. For example, here https://github.com/termux/termux-packages/commit/c8da2e75f61bc535dc6a01058a6edea3ba2e5173 the meson.build file was deleted to use cargo build. There the build system passes various option with cargo build internally for cross compilation.

Decodetalkers:
> Also, for some situations like cross compiling or using certatin toolchain, some options need to be passed to cargo. For example, here [termux/termux-packages@c8da2e7](https://github.com/termux/termux-packages/commit/c8da2e75f61bc535dc6a01058a6edea3ba2e5173) the meson.build file was deleted to use cargo build. There the build system passes various option with cargo build internally for cross compilation.

then now https://github.com/neocmakelsp/neocmakelsp/blob/master/meson.build#L40-L58 can it works? Then how can I do cp in windows?

Biswa96:
For build systems issue, may I summon @eli-schwartz please? 🧞‍♂️

eli-schwartz:
Certainly you may. :)

> Emm.. sorry, I copy the meson build from fractal, whose meson build use '&&', and it always works on my archlinux. I think I have fixed it in master.

Using `&&` in a meson `command: [....]` does not work, because meson only supports and permits a *command* with argv parameters. The `&&` is, rather, shell syntax. Meson doesn't guarantee that it will run commands via a shell such as /bin/sh and this is notably not the case on Windows. In most cases, meson will actually end up quoting shell metacharacters for safety because it assumes the `&&` is a string argv parameter to the command.

You should instead use something like `cargo_wrapper.py` to handle both programs you want to run (`cargo build` as well as `shutil.move()`.)

eli-schwartz:
It is also possible to use meson's builtin support for the rust language:
```meson
executable('foobar', 'foobar.rs')
```

You can depend on crates.io crates as subprojects, synthesized via Cargo.toml (similar to how meson can synthesize a cmake subproject):

https://mesonbuild.com/Wrap-dependency-system-manual.html#cargo-wraps

PR #21815, opened 2024-09-01
Adjust PKGBUILDs for CPPFLAGS removal from makepkg

lazka:
See https://github.com/msys2/MSYS2-packages/issues/4860 It was moved from CPPFLAGS to CFLAGS/CXXFLAGS.

### DIFF
--- a/mingw-w64-grpc/PKGBUILD
+++ b/mingw-w64-grpc/PKGBUILD
@@ -6,7 +6,7 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-python-grpcio")
 pkgver=1.64.2
-pkgrel=2
+pkgrel=3
 _opencensus_proto_ver=0.4.1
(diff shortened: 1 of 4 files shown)


ognevny:
for rufus remove `unset CPPFLAGS`. instead remove `-D__USE_MINGW_ANSI_STDIO=1` from it, as upstream redefines it to 0

lazka:
> for rufus remove `unset CPPFLAGS`. instead remove `-D__USE_MINGW_ANSI_STDIO=1` from CFLAGS, as upstream redefines it to 0

thanks, I hadn't looked for "unset". rufus ignores CFLAGS (should we fix that?), and even if it would it would override our default, so I think it is fine. The unset could be removed though.

ognevny:
> > for rufus remove `unset CPPFLAGS`. instead remove `-D__USE_MINGW_ANSI_STDIO=1` from CFLAGS, as upstream redefines it to 0
> 
> thanks, I hadn't looked for "unset". rufus ignores CFLAGS (should we fix that?), and even if it would it would override our default, so I think it is fine. The unset could be removed though.

oh I didn't see it. understood

lazka:
the check thing is a false positive, so merging

PR #21817, opened 2024-09-01
typstfmt: init at 0.2.10

### LINKED ISSUE #16663: Add typst-fmt
### Package name

typst-fmt

A formatter for the Typst language

https://github.com/astrale-sharp/typst-fmt/

_No response_

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANG32
- [X] CLANGARM64

No

Freed-Wu:
Fix #16663


lazka:
Sorry for the late response.

Upstream was archived before this was opened and it looked like no fork emerged since then.

The alternative seems to be https://github.com/Enter-tainer/typstyle - should we package that instead?

Freed-Wu:
https://github.com/Myriad-Dreamin/tinymist support both typstfmt and typstyle. So it should be OK.

PS: comment from the author: https://github.com/Myriad-Dreamin/tinymist/issues/794#issuecomment-2469694128

Myriad-Dreamin:
I'm not against typstfmt, but developing on typstfmt may prevent our future deprecation and remove of typstfmt. If you love some features in typstfmt which is not in typstyle, you can open an issue on typstyle and see if they will handle it. As far I know the well-known prose wrap feature, provided by typstfmt, was implemented in typstyle v0.13 through development for a long time.

Freed-Wu:
> developing on typstfmt may prevent our future deprecation and remove of typstfmt.

Both Accept and decline are OK :smile:

ognevny:
the project was archived by the author, meaning that this is abandoned...

PR #21831, opened 2024-09-01
codeblocks: Update and apply patch to fix run-time issue

stahta01:
The patch removes an invalid message when installing an cbplugin file.

Tim S.


Biswa96:
Did you report that issue in upstream? Could you also report the wxSmith plugin issue which I added as workarounds in previous two commits in codeblocks package?

stahta01:
> Did you report that issue in upstream? Could you also report the wxSmith plugin issue which I added as workarounds in previous two commits in codeblocks package?

I reported the patch I added upstream; but, since the issue only happens when using configure/make build under windows I do not expect a fast response.  Your way of fixing the wxSmith issue is not likely to be accepted; so, I did not pass it on. But, another github project that is being discussed used that method; so, people on the CB Forum knows about the solution.

Tim S.

PR #21839, opened 2024-09-03
angleproject: fix building on arm64

raedrizqie:
Can someone test building this on arm64?


lazka:
seems to have done the trick. thank you!

PR #21863, opened 2024-09-06
aom: CPU detection fixed upstream

kmilos:


Biswa96:
Could you provide the commit which fixed the issue in upstream?

Biswa96:
Thank you. I have merged the commit with modified commit message.

PR #21876, opened 2024-09-08
abseil-cpp: Workaround compiler error due to absence of ICalendar definition

Biswa96:
This fixes the following compiler error.
time_zone_lookup.cc:102:14: error: use of undeclared identifier 'RuntimeClass_Windows_Globalization_Calendar'


Biswa96:
I have sent a merge request to add the ICalendar definition in wine headers https://gitlab.winehq.org/wine/wine/-/merge_requests/6460

It'll later be imported to mingw-w64.

PR #21892, opened 2024-09-11
zed: update to 0.152.3

ognevny:


ognevny:
I can't reproduce issue locally

PR #21895, opened 2024-09-11
arpack: package PARPACK

mmuetzel:
Package files for PARPACK in a split package to keep the dependency on MSMPI separate.

### DIFF
--- a/mingw-w64-arpack/PKGBUILD
+++ b/mingw-w64-arpack/PKGBUILD
@@ -7,9 +7,10 @@
 _realname=arpack
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
-         $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-${_realname}64"))
+         $([[ "${CARCH}" == "i686" ]] || echo "${MINGW_PACKAGE_PREFIX}-${_realname}64")
+         $([[ "${CARCH}" != "x86_64" ]] || echo "${MINGW_PACKAGE_PREFIX}-p${_realname}"))
 pkgver=3.9.1
-pkgrel=1
+pkgrel=2
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
 pkgdesc="Fortran77 subroutines designed to solve large scale eigenvalue problems (mingw-w64)"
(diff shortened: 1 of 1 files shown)


mingw-w64-arpack/PKGBUILD
```diff
@@ -125,6 +131,26 @@ package_arpack() {
     sed -e "s|${MINGW_PREFIX}|\$\{_IMPORT_PREFIX\}|g" -i ${_f}
   done
 
+  if [ "${CARCH}" == "x86_64" ]; then
```
ognevny, line 134:
```suggestion
  if [ "${CARCH}" != "i686" ]; then
```

mingw-w64-arpack/PKGBUILD
```diff
@@ -125,6 +132,26 @@ package_arpack() {
...
+    mv ${pkgdir}${MINGW_PREFIX}/lib/pkgconfig/parpack* ${srcdir}/${MSYSTEM}-PARPACK/lib/pkgconfig/
+  fi
+}
+
+package_parpack() {
+  pkgdesc="Fortran77 subroutines designed to solve large scale eigenvalue problems with MPI (mingw-w64)"
+  depends+=("${MINGW_PACKAGE_PREFIX}-arpack"
+            "${MINGW_PACKAGE_PREFIX}-openblas"
+            "${MINGW_PACKAGE_PREFIX}-msmpi")
+
```
Biswa96, line 152: Here should be a `provides=()` if you do not want to add provides for this sub-package.

mmuetzel: Good point! 👍

I copied that and a couple more assignments from the `package_arpack64` from the same file.

mmuetzel:
I downloaded and installed the build artifacts from this PR. With them, Octave's test suite still passes.
I also built ElmerFEM locally against libarpack and libparpack. It built without issues and no additional self-tests are failing afaict.

mingw-w64-arpack/PKGBUILD
```diff
@@ -125,6 +132,29 @@ package_arpack() {
...
+    mv ${pkgdir}${MINGW_PREFIX}/lib/libparpack* ${srcdir}/${MSYSTEM}-PARPACK/lib/
+    mv ${pkgdir}${MINGW_PREFIX}/lib/pkgconfig/parpack* ${srcdir}/${MSYSTEM}-PARPACK/lib/pkgconfig/
+  fi
+}
+
+package_parpack() {
+  pkgdesc="Fortran77 subroutines designed to solve large scale eigenvalue problems with MPI (mingw-w64)"
+  depends+=("${MINGW_PACKAGE_PREFIX}-arpack"
+            "${MINGW_PACKAGE_PREFIX}-openblas"
+            "${MINGW_PACKAGE_PREFIX}-msmpi")
```
MehdiChinoune, line 151: ${MINGW_PACKAGE_PREFIX}-msmpi is not a runtime dependency. The user should instead install the MS-MPI from Microsoft which should install msmpi.dll in `C:\Windows\system32` folder.

mmuetzel: True. I'm not sure how to reflect that in the dependencies though.

Feel free to change that if you have an idea how to show that requirement correctly.

ognevny: warning via .install file?

MehdiChinoune: This affect all MPI packages, one should come up with a solution to tell users to install MS-MPI.
${MINGW_PACKAGE_PREFIX}-msmpi is a package that depends on the compilers C/C++/Fortran which is huge size to install.

PR #21901, opened 2024-09-11
rust: add gnullvm target for CLANG{32,64} envs

mati865:
Prerequisite to disable bootstrap for CLANG envs: #219002

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -20,7 +20,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-rust-wasm")
          "${MINGW_PACKAGE_PREFIX}-rust-src")
 pkgver=1.81.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Systems programming language focused on safety, speed and concurrency (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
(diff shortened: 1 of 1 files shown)


ognevny:
maybe some config should be changed via config.toml

mingw-w64-rust/PKGBUILD
```diff
@@ -213,6 +213,11 @@ build() {
     sed -i '/target = \[/a\  "wasm32-unknown-unknown", "wasm32-wasip1", "wasm32-wasip1-threads", "wasm32-wasip2",' "${srcdir}/${_realname}c-${pkgver}-src/config.toml"
   fi
 
+  # Add target *-gnullvm
+  if [[ $MINGW_PACKAGE_PREFIX == *-clang-i686 || $MINGW_PACKAGE_PREFIX == *-clang-x86_64 ]]; then
+    sed -i '/target = \[/a\  "${CARCH}-pc-windows-gnullvm",' "${srcdir}/${_realname}c-${pkgver}-src/config.toml"
```
ognevny, line 218:
```suggestion
    sed -i "/target = \[/a\  \"${CARCH}-pc-windows-gnullvm\"," "${srcdir}/${_realname}c-${pkgver}-src/config.toml"
```

ognevny:
it now builds, can be squahed & merged?

mati865:
> it now builds, can be squashed & merged?

Thanks, I had tested only x86_64 build and tried to quickly address CI errors.
I'll try the 32-bit build first and then we can move on.

> maybe some config for gnullvm should be changed via config.toml

This is PR is only for adding additional core and std libs that will be removed in the next PR. If we want additional changes, they should happen in subsequent PRs.

ognevny:
we can move forward with this IMO. anyway clang32 rust ill be dropped at all

mati865:
Agreed

lazka:
Could this error be related? https://github.com/msys2/MINGW-packages/actions/runs/11334394015/job/31520493787?pr=21684

```
  librsvg-2.59.1/meson.build:345:4: ERROR: Problem encountered: error[E0463]: can't find crate for `std`
    |
    = note: the `x86_64-pc-windows-gnu` target may not be installed
    = help: consider downloading the target with `rustup target add x86_64-pc-windows-gnu`
  
  error: aborting due to 1 previous error
  
  For more information about this error, try `rustc --explain E0463`.
  ```

mati865:
Yes, something is passing target that is now wrong: `Running command: D:/M/msys64/clang64/bin/python.exe C:/_/B/src/librsvg-2.59.1/meson/query-rustc.py D:\M\msys64\clang64\bin/rustc.EXE --target x86_64-pc-windows-gnu --query=native-static-libs`

lazka:
Ah, I see, thanks!

PR #21902, opened 2024-09-11
rust: migrate CLANG{32,64} envs to gnullvm targets

mati865:
Requires to be built with packages from https://github.com/msys2/MINGW-packages/pull/21901


mati865:
With CLANG32 migrated to gnullvm some of our packages fail with `lld: error: unable to find library -lwindows.0.48.5` due to outdated dependencies. Although we don't package them any longer, so that shouldn't be a problem, I suppose.

ognevny:
done. I'll squash and merge it after CI is green

PR #21903, opened 2024-09-12
virglrenderer: Update to 1.1.0

heljkon:
No patching or CFLAGS required for MINGW64/UCRT64.

Only minimal patches for CLANG64.

For CLANG64 the same issues as documented in QEMU PR https://github.com/msys2/MINGW-packages/pull/21540 occur as they did in previous builds.
```
[32/47] Compiling C object src/libvirgl.a.p/vrend_blitter.c.obj
../virglrenderer-1.1.0/src/vrend_blitter.c:94:8: warning: unknown attribute 'gcc_struct' ignored [-Wunknown-attributes]
   94 | struct PACKED blit_prog_key {
      |        ^~~~~~
../virglrenderer-1.1.0/src/mesa/util/macros.h:207:35: note: expanded from macro 'PACKED'
  207 | #    define PACKED __attribute__((gcc_struct,__packed__))
      |                                   ^~~~~~~~~~
1 warning generated.
```

If QEMU is the only package requiring virglrenderer, maybe we should drop CLANG-builds for virglrenderer as we did with QEMU until https://github.com/llvm/llvm-project/pull/71148 is merged?

Is there a way to check the reverse dependencies of virglrenderer?

### DIFF
--- a/mingw-w64-virglrenderer/PKGBUILD
+++ b/mingw-w64-virglrenderer/PKGBUILD
@@ -3,11 +3,12 @@
 _realname=virglrenderer
 pkgbase=mingw-w64-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-${_realname}
-pkgver=1.0.1
-pkgrel=2
+pkgver=1.1.0
+pkgrel=1
 pkgdesc='A virtual 3D GPU library, that allows the guest operating system to use the host GPU to accelerate 3D rendering (mingw-w64)'
 arch=('any')
-mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+# clang disabled until attribute gcc_struct is supported, see https://github.com/msys2/MINGW-packages/pull/21903#issue-2521206281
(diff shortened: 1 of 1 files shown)


lazka:
> If QEMU is the only package requiring virglrenderer, maybe we should drop CLANG-builds for virglrenderer as we did with QEMU until [llvm/llvm-project#71148](https://github.com/llvm/llvm-project/pull/71148) is merged?

Sounds reasonable, whatever you prefer.

> Is there a way to check the reverse dependencies of virglrenderer?

You can see them here under "Required": https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-virglrenderer

or via

```console
$ pactree -rs mingw-w64-ucrt-x86_64-virglrenderer
mingw-w64-ucrt-x86_64-virglrenderer
└─mingw-w64-ucrt-x86_64-qemu
```

heljkon:
After disabling clang builds the remaining builds succeed without requiring any patches or additional CLFAGS.

PR #21910, opened 2024-09-13
templates: remove some `${srcdir}` references

ognevny:
every functions start at `${srcdir}`. it's not required to pass full paths to `cd`


ognevny:
maybe remove `msg` from python template too?

PR #21916, opened 2024-09-13
fastfetch: update optional dependencies

CarterLi:


ognevny:
maybe bump pkgrel?

Biswa96:
It will be added in next version update. fastfetch is really _fast_ with new version every week/month.

PR #21927, opened 2024-09-14
sonixflasher: update to 2.0.6

dexter93:


Biswa96:
Could you please remove the static flags in Makefile? We prefer dynamic linking and it will reduce the binary file size.

dexter93:
> Could you please remove the static flags in Makefile? We prefer dynamic linking and it will reduce the binary file size.

done, thank you!

Biswa96:
My apology for the confusion, I did not mean to remove the static option in upstream. You could patch is here only. Some may want static linking for a standalone program but for packaging purposes shared linking is preferred.

dexter93:
> My apology for the confusion, I did not mean to remove the static option in upstream. You could patch is here only. Some may want static linking for a standalone program but for packaging purposes shared linking is preferred.

No worries, it was actually a leftover from initial bringup. However, worth noting that I didn't see a binary size difference
Anyway, hope it's ok now :)

PR #21931, opened 2024-09-14
jsoncpp: Update to 1.9.6

lazka:


lazka:
https://github.com/curl/curl/issues/14923

lazka:
We have to rethink cmake bootstrapping, a so-bump in a dependency always requires manual intervention otherwise.

PR #21936, opened 2024-09-16
helix: update to 24.07

avnyu:
icon.patch: use windres to add icon when target_env is gnu
disable-lto.patch: PKGBUILD failed at cargo test --locked --release, alternatively test without --release can work too
This is my first time to create a pull request, sorry in advanced if I cause any problems.

### DIFF
--- a/mingw-w64-helix/PKGBUILD
+++ b/mingw-w64-helix/PKGBUILD
@@ -12,7 +12,6 @@ url="https://github.com/helix-editor/helix"
 license=('spdx:MPL-2.0')
 makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
              "${MINGW_PACKAGE_PREFIX}-mdbook"
-             "patch"
              "git")
 source=("$url/archive/$pkgver/${_realname}-${pkgver}.tar.gz"
         'icon.patch'
--- a/mingw-w64-helix/PKGBUILD
+++ b/mingw-w64-helix/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=helix
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=23.10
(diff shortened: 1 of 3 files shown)


mingw-w64-helix/PKGBUILD
```diff
@@ -12,11 +12,16 @@ url="https://github.com/helix-editor/helix"
 license=('spdx:MPL-2.0')
 makedepends=("${MINGW_PACKAGE_PREFIX}-rust"
              "${MINGW_PACKAGE_PREFIX}-mdbook"
+             "patch"
```
ognevny, line 15: there is no need to add this package. it's always installed in build machines

ognevny:
can be icon.patched added upstream?

mingw-w64-helix/icon.patch
```diff
@@ -0,0 +1,64 @@
...
+         println!("cargo:rustc-link-lib=dylib=resource");
+     }
+ 
++    #[cfg(target_env = "gnu")]
++    fn compile_with_windres_gnu(output: PathBuf, input: PathBuf) {
++        let mut command = process::Command::new("windres.exe");
++        let command = command.arg(format!(
++            "--include-dir={}",
++            env::var("CARGO_MANIFEST_DIR")
++                .expect("CARGO_MANIFEST_DIR should have been set by Cargo")
```
ognevny, line 34:
```suggestion
+                .unwrap()
```
nit: `Result::Err` won't be returned in this case, no need to handle it woth `expect`

ognevny:
don't forget to run `updpkgsums` after changing files :)

avnyu:
sorry about that, about the icon.patch, it should work upstream, but i didn't tested

ognevny:
add `${MINGW_PACKAGE_PREFIX}-gcc-libs` to depends

ognevny:
I tested it on CLANG64. sorry for delay, thanks for contribution

PR #21956, opened 2024-09-17
enter-tex: add package, renamed from gnome-latex

Biswa96:


Biswa96:
I can not figure out the meson error.

Biswa96:
@eli-schwartz could you provide any hint to troubleshoot this meson issue?

eli-schwartz:
It's not precisely a meson issue (maybe not a meson issue at all). Meson successfully configured.

When compiling `src/gtex/Gtex-1.gir` using gnome's g-ir-scanner, it appears to be instructed to parse libgtex-static.a instead of libgtex-shared.dll, and the gnome introspection tooling emits an error that it cannot find a shared DLL at that location (the .a file).

eli-schwartz:
https://gitlab.gnome.org/swilmet/enter-tex/-/blob/main/src/gtex/meson.build#L129

```meson.build
GNOME.generate_gir(
  GTEX_STATIC_LIB,
  header: 'gtex.h',
```

Interesting.

Biswa96:
> the gnome introspection tooling emits an error that it cannot find a shared DLL at that location (the .a file).

Do you mean that should be changed to shared library target gtex_shared_lib?

eli-schwartz:
It sounds like it, yes. At the very least, ask upstream whether they think it might make sense to do so. :p

Biswa96:
I have reported the issue in upstream https://gitlab.gnome.org/swilmet/enter-tex/-/issues/21

PR #21961, opened 2024-09-18
[TEST] rust: update config.toml

ognevny:
change the way how values are substituted, use `envsubst`
explicitly disable parallel compiler (nightly only feature)
add commented config values which are uncommented in some conditions (`_bootstrap=no` or `debug = true`)

DO NOT MERGE, I don't want to break gnullvm-related PRs


ognevny:
> Question: Why did we switch from generating the `config.toml` file to providing it (and substituting some variables) ?

it makes wasm integrating easier

PR #21988, opened 2024-09-20
Add cmake-bootstrap

lazka:
A cmake variant that doesn't depend on any external dependencies and can be used for bootstrapping an env, and building a full cmake.

### DIFF
--- a/mingw-w64-cmake-bootstrap/0003-fix-find-python-on-mingw-aarch64.patch
+++ b/mingw-w64-cmake-bootstrap/0003-fix-find-python-on-mingw-aarch64.patch
@@ -1,0 +1,13 @@
+--- a/Modules/FindPython/Support.cmake
++++ b/Modules/FindPython/Support.cmake
+@@ -970,8 +970,8 @@
+         ERROR_QUIET
+         OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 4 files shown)


Biswa96:
Identity crisis - if standalone cmake-bootstrap is present why do we need the system dependent cmake?

lazka:
> Identity crisis - if standalone cmake-bootstrap is present why do we need the system dependent cmake?

It statically links everything, which isn't ideal I guess. But yeah, I'm not sure what the best approach is.. this is more of an experiment.

Biswa96:
Alpine Linux links cmake selectively https://git.alpinelinux.org/aports/tree/main/cmake/APKBUILD

```
# jsoncpp/cppdap/curl/nghttp2 needs cmake to build so to avoid recursive build
# dependency, we use the bundled version.
# Do NOT remove --no-system-*
```

lazka:
oh, and the qt-gui stuff would be missing, since that can't be vendored in.

lazka:
One difference to Linux distributions is that they are usually fine with using autotools in the cmake dependencies to avoid cycles. But in our case I'd prefer if we would switch to cmake for packages that support it, like libarchive and libuv.

lazka:
I expanded the commit message a bit and changed cmake to build with cmake-bootstrap.

PR #21991, opened 2024-09-20
soqt/quarter: update to 1.6.3/1.2.2

podsvirov:


Biswa96:
Why do we need a rebuild for python package?

podsvirov:
> Why do we need a rebuild for python package?

The `pivy` use `swig` to generate native python module (dll) (for example, `/ucrt64/lib/python3.11/site-packages/pivy/_coin.cp311-mingw_x86_64_ucrt.pyd).

Biswa96:
> The `pivy` use `swig` to generate native python module (dll)

That is true for any other projects which uses swig to generate native python module. Did you notice any API/ABI break with the soqt/quarter update?

podsvirov:
> Did you notice any API/ABI break with the soqt/quarter update?

You were right. I checked and it seems that rebuilding is not required now.

Biswa96:
Just as a suggestion, could you rebase the PRs while merging? It is not a requirement. It would just help to make the git history linear.

podsvirov:
> Just as a suggestion, could you rebase the PRs while merging? It is not a requirement. It would just help to make the git history linear.

It's already been done.

PR #22013, opened 2024-09-22
boost: update to 1.86

MehdiChinoune:
stacktrace.from_exception was disabled because It's not implemented yet for mingw-w64. It was enabled for Windows with 1.86 release because they added an MSVC implementation.

package-grokker shows nothing need to be rebuilt.


Biswa96:
> stacktrace.from_exception was disabled because It's not implemented yet for mingw-w64.

Could you provide the source file in boost project or any compiler error related to that feature?

MehdiChinoune:
> Could you provide the source file in boost project or any compiler error related to that feature?
```
  gcc.compile.c++ bin.v2\libs\stacktrace\build\gcc-14\release\x86_64\pch-off\threadapi-win32\threading-multi\visibility-hidden\from_exception.o
  libs/stacktrace/build/../src/from_exception.cpp:160:10: fatal error: dlfcn.h: No such file or directory
    160 | #include <dlfcn.h>
        |          ^~~~~~~~~
  compilation terminated.
```
It was enabled for Windows (unconditionnaly) in https://github.com/boostorg/stacktrace/commit/d1b7a613

PR #22018, opened 2024-09-22
python-imagecodecs: update to 2024.9.22

kmilos:


mingw-w64-python-imagecodecs/PKGBUILD
```diff
@@ -43,13 +44,18 @@ makedepends=("${MINGW_PACKAGE_PREFIX}-python-build"
 checkdepends=("${MINGW_PACKAGE_PREFIX}-python-pytest")
 options=('!strip')
 source=("https://pypi.org/packages/source/${_realname::1}/${_realname}/${_realname}-${pkgver}.tar.gz"
+        "https://patch-diff.githubusercontent.com/raw/cgohlke/imagecodecs/pull/110.patch"
```
Biswa96, line 47: It seems that the patch file adds new feature. Generally, the patch files are added to fix current issues. Is this patch required here?

Biswa96: Or you could update this without the patch file.

lazka: I don't mind mingw specific backports at least.

PR #22021, opened 2024-09-23
python-tifffile: update to 2024.9.20

kmilos:


Biswa96:
@lazka clang32 fails in CI. Should it be disabled?

ognevny:
ntldd was removed, so ALL CI builds will fail

lazka:
> ntldd was removed, so ALL CI builds will fail

oh, please someone add it back

PR #22023, opened 2024-09-23
CI: Install ntldd only when there is a list of packages to be built.

MehdiChinoune:


ognevny:
maybe just enable ntldd for clang32? anyway it won't be so long, but CI will succeed

PR #22036, opened 2024-09-25
cfitsio: update to 4.5.0

kmilos:
Will need all deps rebuilt because of soname change:

mingw-w64-ucrt-x86_64-gdal
mingw-w64-ucrt-x86_64-gimp3
mingw-w64-ucrt-x86_64-libvips
mingw-w64-ucrt-x86_64-python-imagecodecs


lazka:
python-imagecodecs seems to have vendored parts of cfitsio and is no longer using it directly

PR #22039, opened 2024-09-26
zed: update to 0.154.2

ognevny:


ognevny:
tree-sitter has been updated upstream and finally linking issue with ld **disappeared**. I will try to build on clang64 soon

PR #22040, opened 2024-09-26
boost: build with cmake

MehdiChinoune:


lazka:
nice build performance boost :)

vejbomar:
@MehdiChinoune can I ask what was the motivation for this change?

I'm working on `mingw-w64-gcc` for `aarch64-w64-mingw32` target and we want to build Boost with our toolchain to evaluate our toolchain. I started with `b2` but I'm having some errors with that, so I wonder if building with `cmake` isn't actually better supported on Windows?

MehdiChinoune:
> @MehdiChinoune can I ask what was the motivation for this change?
> 
> I'm working on `mingw-w64-gcc` for `aarch64-w64-mingw32` target and we want to build Boost with our toolchain to evaluate our toolchain. I started with `b2` but I'm having some errors with that, so I wonder if building with `cmake` isn't actually better supported on Windows?

CMake is a better than b2, you could observe yourself that patches were reduced and the build was faster.

PR #22042, opened 2024-09-26
dtc: update to 1.7.1

Biswa96:
Remove 0001-meson-disable-tests.patch file because tests can be disabled with different option now.


Biswa96:
I need some help with the python stuff. Those are built but not installed for some reason.

raedrizqie:
maybe this could fix it?
https://gitlab.archlinux.org/archlinux/packaging/packages/dtc/-/blob/main/pyscm.patch?ref_type=heads

Biswa96:
> maybe this could fix it? https://gitlab.archlinux.org/archlinux/packaging/packages/dtc/-/blob/main/pyscm.patch?ref_type=heads

That patch is not used in ArchLinux dtc PKGBUILD. I have added commands to install python bindings separately by building a python wheel.

PR #22047, opened 2024-09-27
rust: rebuild

ognevny:
- rebuild to remove patch for cross-build (used for 'bootstrapping' gnullvm target)
- remove hack that used for `*-pc-windows-gnu` targets under CLANG{32,64} environments

merge only after rust 1.81.0-3 is available in repo


ognevny:
> * remove hack that used for `*-pc-windows-gnu` targets under CLANG{32,64} environments

@mati865 will you mind if I do so?

mati865:
No problem, it's getting more and more broken (as expected) as the new releases come out anyway.

ognevny:
CLANG32 fails as it require to have newer version of windows-rs. I guess it can be updated easily. will try tomorrow

ognevny:
cherry-picked #21961 while at it

ognevny:
> CLANG32 fails as it require to have a newer version of windows-rs. I guess it can be updated easily. will try tomorrow

that workaround will be removed in 1.82.0 update

PR #22049, opened 2024-09-27
mingw-w64: default msvcrt to msvcr80 on MINGW64

MehdiChinoune:
I have encountered an issue today while updating posgresql. It fails to build on MINGW64 because PG uses some functions only available when `__MSVCRT_VERSION>=0x800` (which correspond to `MSVC-2005`). but the default value of `__MSVCRT_VERSION__` when choosing msvcrt is `0x700`.
Many functions are not available for `__MSVCRT_VERSION<0x800` https://github.com/search?q=repo%3Amingw-w64%2Fmingw-w64+__MSVCRT_VERSION__+0x800&type=code
This PR makes MINGW64 default to `msvcr80` (`__MSVCRT_VERSION=0x800`).

I don't know what implications/consequences does this change have!. Does the user have to install MSVCR80.dll?

@mstorsjo @Biswa96 @lazka

### DIFF
--- a/mingw-w64-crt-git/PKGBUILD
+++ b/mingw-w64-crt-git/PKGBUILD
@@ -4,7 +4,7 @@ _realname=crt
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgver=12.0.0.r297.g0be496546
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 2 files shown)


lazka:
> I don't know what implications/consequences does this change have!. Does the user have to install MSVCR80.dll?

Yes, that's not really an option.

I'd suggest to either patch postgres, or drop mingw64

PR #22052, opened 2024-09-27
[new package] rust-bindgen 0.70.1

podsvirov:


ognevny:
does it actually depend on clang? om arch `libclang` libs are linked dynamically, while it doesn't work there due to issue with `clang-sys` crate: https://github.com/KyleMayes/clang-sys/issues/173

ognevny:
> libs are linked dynamically

oh, my apologies, I should have checked it. it doesn't actually, it's not listed in soname lists

PR #22068, opened 2024-09-29
[new-package] mdcat 2.5.0

srouquette:


podsvirov:
@srouquette, plesase add license information and install LICENSE file (for example like [there](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-nss/PKGBUILD))

podsvirov:
@srouquette, thanks for your efforts!

PR #22092, opened 2024-10-02
[new package] c3c 0.6.3

ognevny:


ognevny:
lld is installed, but it's not found. looks like it requires manual setting...

Biswa96:
Do we need this new toolchain for any program?

ognevny:
> Do we need this new toolchain for any program?

I saw they have CI pipeline for build it with MSYS2, so I decided to provide as it should be supported. anyway, why not?

Biswa96:
I was just curious. No objection.

PR #22101, opened 2024-10-04
[new package] wasmtime 25.0.1

ognevny:
`libwasmtime` is used in tree-sitter for wasm support


ognevny:
@mati865 .dll.a import library is not generated for gnullvm build. why?

mati865:
Because of the lack of https://github.com/rust-lang/cargo/pull/14451

ognevny:
thanks! will backport later

PR #22120, opened 2024-10-05
wasmer: update to 4.4.0

ognevny:


ognevny:
it now requires #22119

ognevny:
turns out this also requires huge backport for one crate...

PR #22124, opened 2024-10-06
tree-sitter: update to 0.24.2

ognevny:
- switch to cmake, patch is not required
- enable wasm feature


ognevny:
there is missing `wasmtime/conf.h` header: https://github.com/bytecodealliance/wasmtime/issues/9382

ognevny:
.pc file is now installed in `$prefix/shared/` is it bad?

PR #22128, opened 2024-10-06
libwasmtime: fix LICENSE file destination

podsvirov:
Fix error when install both wasmtime and libwasmtime:
```
/ucrt64/share/licenses/wasmtime/LICENSE exists in both 'mingw-w64-ucrt-x86_64-wasmtime' and 'mingw-64-ucrt-x86_64-libwasmtime'
```


ognevny (approved):
good catch, thanks!

PR #22130, opened 2024-10-06
qt5: restore some MINGW32 packages

SlySven:
Following a discussion on Discord, see:
https://www.discord.com/channels/792780131906617355/794220101741838401/1283064557660803083 brought about by https://github.com/msys2/MINGW-packages/commit/a4f0579e14509d27927a94498c35e1ccddf29bf5 I suggested to restore for a while a couple of Qt5 packages that the [Mudlet](https://github.com/Mudlet/Mudlet) project I code for uses for its 32-Bit Windows builds until we have migrated as many of our Windows users up to 64-Bit builds as possible.

I was hoping that one of the MSYS2+Mingw-w64 team would have been able to find the time to make this change but after a couple of weeks of nothing happening I have decided to try and do it myself.

### DIFF
--- a/mingw-w64-qt5-speech/PKGBUILD
+++ b/mingw-w64-qt5-speech/PKGBUILD
@@ -3,12 +3,12 @@
 _realname=qt5-speech
 pkgbase="mingw-w64-${_realname}"
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
-         "${MINGW_PACKAGE_PREFIX}-${_realname}-debug")
+         $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-${_realname}-debug"))
 pkgver=5.15.15
-pkgrel=1
+pkgrel=2
 pkgdesc="Qt module to make text to speech and speech recognition easy (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url='https://www.qt.io/'
 msys2_references=(
   "cpe: cpe:/a:qt:qt"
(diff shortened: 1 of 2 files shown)


mingw-w64-qt5-speech/PKGBUILD
```diff
@@ -3,12 +3,12 @@
 _realname=qt5-speech
 pkgbase="mingw-w64-${_realname}"
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
-         "${MINGW_PACKAGE_PREFIX}-${_realname}-debug")
+         $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-${_realname}-debug"))
 pkgver=5.15.15
-pkgrel=1
+pkgrel=2
```
SlySven, line 8: I suppose this change in this file and the other one is technically redundant - the packages that are produced for the other build types are not altered in any way so that the end-user will not see any changes when their `mingw-w64-*-qt5-winextras` and `mingw-w64-*-qt5-speech` packages are removed and replaced with exactly the same things despite this number being incremented - and since the previous packages for MINGW32 were for Qt 5.15.**14** there is no problem with the new packages having `1` here.

lazka:
Is there an upstream issue where one can track the move to 64bit?

Biswa96:
Would it be possible to migrate to 64 bit now instead of adding 32 bit qt5?

vadi2:
Unfortunately no, because we need to do at least one 32bit update that will guide users towards installing 64bit. If we dont do this, the vast majority of our userbase will never receive an update for the next version.

mingw-w64-qt5-speech/PKGBUILD
lazka: correct

lazka:
Thanks. We'll restore them for now and re-evaluate in 3 months.

vadi2:
Appreciated! It would also be nice to know in the future if such packages are planned to be dropped by the project in advance so that we can plan for this.

MehdiChinoune:
@SlySven @vadi2 Did you complete the latest 32-bit release.

vadi2:
We have now added a 64bit version of our application: https://www.mudlet.org/download

But folks have only had 1 week to switch from 32bit so far, which is not enough. Could we have one more month to switch? Deprecating 32bit in February would work well for us :+1:

MehdiChinoune:
> But folks have only had 1 week to switch from 32bit so far, which is not enough. Could we have one more month to switch? Deprecating 32bit in February would work well for us 👍

I don't understand, both 32-bit and 64-bit versions are available they could move any time they want!
It's just about new versions being 64-bit only.

vadi2:
We have thousands of people - 90% of windows users - using the 32bit version that receives regular updates. If we drop support for 32bit, they will no longer receive updates, and they'll be stuck on an older version until they manually fix the problem.

We've added code to migrate those users to 64bit when they update, and we'd like to have a chance to issue several updates to ensure that everyone can move over.

MehdiChinoune:
We are in February, now. I guess It's safe to drop them again now.

PR #22137, opened 2024-10-07
wasmtime: configure build type

podsvirov:


ognevny:
does it make sense? it's just rust library compiled into .dll or .a file. `-DWASMTIME_FASTEST_RUNTIME` actually does the thing. also `"${extra_config[@]}` is correct

podsvirov:
> does it make sense?

Just to match the CMake package template.

It's can be useful if some CMake export will be added later.

PR #22138, opened 2024-10-07
pywin32: update to 308

raedrizqie:


ognevny:
why not python-build and python-inataller?

raedrizqie:
> why not using python-build and python-inataller?

done

raedrizqie:
@lazka can we merge this?

PR #22159, opened 2024-10-10
Add Specific commits to fix issues with MSYS2

johnathan-becker:
We are currently undergoing a lot of changes with the GNUstep libs-gui repo to make it work correctly in MSYS2 environments. Until that work is done I don't see an upcoming release happening. We do need some of these changes though to continue with the effort. This MR adds two specific commit patches to fix outstanding issues.


johnathan-becker:
I should mention there will probably be more of these patches coming over the next few weeks. If the maintainers have a more preferred way to handle this situation please feel free to comment on this MR.

Biswa96:
Please add the patches in `sources` list and use proper commit message, for example, "gnustep-gui: import patch to fix the-issue-name".

Biswa96:
Would you like to add a proper commit message as I requested before? The reason is not clear behind those upstream commits.

johnathan-becker:
> Would you like to add a proper commit message as I requested before? The reason is not clear behind those upstream commits.

I apologize. I have amended the commit message but it looks like the last commit was a merge commit. Do you want me to add a new commit with the correct message so that is the last one?

PR #22172, opened 2024-10-11
SDL2_mixer: Do not link to LGPL library

sharkwouter:
I'd like to statically link the binary for my game, but I cannot do so if SDL2_mixer relies on libmpg123, since it is LGPL licensed and my software is not GPL licensed. Luckily SDL2_mixer supports using the single header library called minimp3 now, which is included in the SDL2_mixer repo. Building with it works perfectly for PSP and Vita, so I'm assuming it should work just fine for Windows as well.

I've marked this one as draft, because I'm not 100% sure how I'll test this yet. I just wanted to have it up in case someone spots a clear error in my work here. I'll update once I've tested this.

### DIFF
--- a/mingw-w64-SDL2_mixer/PKGBUILD
+++ b/mingw-w64-SDL2_mixer/PKGBUILD
@@ -4,7 +4,7 @@ _realname=SDL2_mixer
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.8.0
-pkgrel=1
+pkgrel=2
 pkgdesc="A simple multi-channel audio mixer (Version 2) (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
Sorry, but we don't do that here.
If you GPL/LGPL free libraries just them yourself.

sharkwouter:
But why? This does not change the functionality of SDL2_Mixer at all.

Biswa96:
The reason for this pull request does not sound like msys2's issue. Many Linux distributions links sdl2-mixer package with mpg123 library.

* Chimera Linux https://github.com/chimera-linux/cports/blob/master/main/sdl_mixer/template.py
* Fedora https://src.fedoraproject.org/rpms/SDL2_mixer/blob/rawhide/f/SDL2_mixer.spec
* Arch Linux https://gitlab.archlinux.org/archlinux/packaging/packages/sdl2_mixer/-/blob/main/PKGBUILD?ref_type=heads
* Alpine Linux https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/sdl2_mixer/APKBUILD

sharkwouter:
I would say Linux is a special case, though. Generally, you'll want to ship binaries that are not statically linked to Linux users, because they are one of the few user groups you can ask to install libraries. For Window you will either have to statically link, or manually add all the library DLLs to your release.

In the end this PR shows one of the ways I can make it so my non-GPL project can be statically linked, so let me know if I'm pushing too much. I really do think this could also benefit other users, though.

PR #22178, opened 2024-10-12
minidump-rustwalk: update to 0.22.2

ognevny:
what about packaging more utilities from rust-minidump?


ognevny:
oopsie, stackwalk* xD

PR #22180, opened 2024-10-12
cmake: remove _bootstrap

ognevny:
cmake-bootstrap is now used for such purpose, so no need to keep "dead code" in general cmake package


lazka:
lgtm

PR #22182, opened 2024-10-12
libjpeg-turbo: Include LICENSE.md in package

sharkwouter:
I was compiling the list of licenses I'm using for a project I'm working on and found that the license file was missing from the libjpeg-turbo package. This PR adds it in.


ognevny:
thanks

PR #22192, opened 2024-10-14
Add OpenH264 option for ffmpeg

johnathan-becker:
This MR adds the ability to use openh264 from Cisco instead of libx264. OpenH264 does not require a commercial license so it can be a good option if you do not have a H264 license.


lazka:
What is the H264 license?

johnathan-becker:
> What is the H264 license?
> 
> Looking around I see that the openh264 encode is worse than the x264 one, so there has to be a good reason to switch.

This Merge Request proposes to retain libx264 as the default encoder while offering users the option to build and package FFmpeg with openh264 as an alternative.

libx264 is licensed under the GNU General Public License (GPL) version 2.0 or later, which imposes certain restrictions, especially for commercial applications. The GPL is a copyleft license, meaning any application that distributes libx264 must also be licensed under the GPL. This requires making the source code available and allowing users to modify and redistribute the entire application under the same terms. For companies developing proprietary or paid software, complying with the GPL's copyleft requirements is often not feasible.

As a result, openh264 is offered as an alternative, since it is licensed under a more permissive license (BSD-style), which does not impose the same copyleft requirements. However, users must manually build and package FFmpeg with openh264 to ensure compliance with licensing terms.

Although openh264 may not provide the same level of encoding quality as libx264, its license makes it a more viable option for commercial software developers who need to avoid the obligations imposed by the GPL, while still including H.264 encoding functionality in their applications.

lazka:
I see, thanks, I assumed as much, but was confused because the GPL wasn't mentioned.

Regarding the GPL this doesn't change much for ffmpeg, as there are more libraries in the dependency chain of ffmpeg that are GPL, like vid.stab, xvidcore, etc. Then there are also patents of certain things on top of that, in certain parts of the world.

As for GPL variants of things in general, this request has come up before, and I'm unsure if that's something we should support. On the one hand it's not something we care about as an open source project, on the other hand if there really is need and would result in more contributions overall then it might be still worth it.

johnathan-becker:
> I see, thanks, I assumed as much, but was confused because the GPL wasn't mentioned.
> 
> Regarding the GPL this doesn't change much for ffmpeg, as there are more libraries in the dependency chain of ffmpeg that are GPL, like vid.stab, xvidcore, etc. Then there are also patents of certain things on top of that, in certain parts of the world.
> 
> As for non-GPL variants of things in general, this request has come up before, and I'm unsure if that's something we should support. On the one hand it's not something we care about as an open source project, on the other hand if there really is a need and would result in more contributions overall then it might be still worth it.
> 
> Maybe a non-gpl build only flag as you suggested is a compromise.

I think that is a fair compromise. It may take me some time to sort all of that out but I will look into it today.

lazka:
This was just meant as a discussion starter for other project members.

oech3:
Is this PR still needed? It seems this is something user can change PKGBUILD easily.

PR #22210, opened 2024-10-16
pyinstaller: update to 6.11.0

podsvirov:
Windows Security treat `PyInstaller/bootloader/Windows-32bit-intel/*.exe` files as `Trojan:Win32/Wacatac.B!ml` for me.
But for bootloaders from `6.10.0` version all good.
Can anybody check or confirm this?


podsvirov:
I'm just deleting an unused loader. There is no executable file - no problem!

PR #22232, opened 2024-10-18
librsvg: rebuild and fix build with newer rust

lazka:
rust 1.82.0 doesn't like the target temp file being open while writing to it


ognevny:
can such change be upstreamed?

lazka:
yes, see the linked issue

ognevny:
it isn't reproducible on my arch machine though

lazka:
It's about replacing a file that is opened by another process, so a Windows specific issue.

PR #22236, opened 2024-10-19
Remove pytest-test-runner (and some wheel)

lazka:
pytest-test-runner should no longer be used (archived upstream, uses deprecated setuptools feature):

* replace the dep with pytest if it's not there, or remove it
* replace setup.py calls with calls to pytest directly

While at it, also remove some wheel dependencies, since wheel is now vendored in setuptools.


lazka:
some US mirror is having issues

PR #22246, opened 2024-10-20
nanobind: new package

marcelotduarte:
It will be used to build LIEF 0.15.1


marcelotduarte:
@lazka @Biswa96 
Please check if it is correct to name this package `python-nanobind` or just `nanobind` since I saw that pybind11 is just pybind11.

Biswa96:
I follow ArchLinux package name, so 'nanobind' only https://archlinux.org/packages/extra/any/nanobind/

Was mingw-w64-python-scikit-build-core rebuilt for any reason?

> It will be used to build LIEF 0.15.1

Would it be possible to add the LIFE update in this pull request? It might help us to check any issues with python LIFE package.

marcelotduarte:
> Was mingw-w64-python-scikit-build-core rebuilt for any reason?

Just to remove the pip lines that I used for testing and forgot to delete before the last PR.

> Would it be possible to add the LIFE update in this pull request? It might help us to check any issues with python LIFE package.

No, there are still some (maybe many) problems to face. Unless you want me to send you the version with errors. But I think it's better to just rename the package now, and during the week I should do something else and send it.

PR #22254, opened 2024-10-22
add python-aioquic

totaam:
Add [`python-aioquic`](https://github.com/aiortc/aioquic) so we can add support for the [QUIC transport](https://github.com/Xpra-org/xpra/blob/master/docs/Network/QUIC.md) in [`xpra`](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-xpra/PKGBUILD). (this optional dependency will be submitted later as a separate PR)

This PR depends on `python-pylsqpack`: https://github.com/msys2/MINGW-packages/pull/22255


Biswa96:
I have added necessary stuff to follow pkgbuild template and python-service-identity dependency.

PR #22255, opened 2024-10-22
add pylsqpack

totaam:
This library is required by [`aioquic`](https://github.com/aiortc/aioquic), PR: https://github.com/msys2/MINGW-packages/pull/22254

---

Sorry about the duplicate (closed) PR https://github.com/msys2/MINGW-packages/pull/22253, the `cc` dependency was missing and when I forced pushed the wrong commit head, it closed that PR.

Incidentally, what is the correct (and preferably efficient) process for verifying that the list of dependencies of a new `PKGBUILD` is correct when working with an existing installation?
Isn't waiting for the CI to run is a little bit of waste of your github runner resources?

### DIFF
--- a/mingw-w64-python-pylsqpack/PKGBUILD
+++ b/mingw-w64-python-pylsqpack/PKGBUILD
@@ -1,0 +1,40 @@
+# Maintainer: Antoine Martin <totaam@xpra.org>
+
+_realname=pylsqpack
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=0.3.18
+pkgrel=1
+pkgdesc="a wrapper around the ls-qpack library. It provides Python Decoder and Encoder objects to read or write HTTP/3 headers compressed with QPACK (mingw-w64)"
+arch=('any')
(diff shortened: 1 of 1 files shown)


Biswa96:
I have added necessary stuff to follow pkgbuild template.

Biswa96:
> what is the correct (and preferably efficient) process for verifying that the list of dependencies of a new `PKGBUILD` is correct when working with an existing installation?

I copy-paste the pkgbuild template in package directory. Then, I look for the dependencies in pyproject.toml, requirements.txt, setup.cfg, setup.py etc. files.

The `cc` dependency is required when the python project uses C/C++ file. Similarly, rust dependency is required if a python project uses rust written components. I hope this clear things a bit. Please let us know if there is any further issues.

totaam:
Thanks @Biswa96.
What I really would like to be able to do is to trigger the kinds of build errors that happen on the CI, but before I submit the PR. So that my PRs are in better shape, and don't cost unnecessary CI runs.
Preferably on my local dev env.

Biswa96:
> What I really would like to be able to do is to trigger the kinds of build errors that happen on the CI

You can reproduce most* of the error in your system. Just build the package in clean msys2 installation.

PR #22263, opened 2024-10-22
mingw-w64: disable wildcard by default

lazka:


lazka:
Some docs: https://github.com/msys2/msys2.github.io/pull/347

lazka:
news entry: https://github.com/msys2/msys2.github.io/pull/348

mati865:
IIRC this will break Rust build or Rust testsuite on envs using Binutils.

lazka:
Thanks. Did someone try this last year?

If all fails we can re-enable it on a per-package basis.

mati865:
I was able to find the old error: `C:/a/rust/rust/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find C:\a\rust\rust\build\i686-pc-windows-gnu\stage2\lib\rustlib\i686-pc-windows-gnu\lib/liballoc-*.rlib: Invalid argument`.

It was previously called from Makefile: https://github.com/rust-lang/rust/pull/128407/files#diff-53c7f7d0d39446b7fbe61af298fbdbdcdfcd1a7f083095343edc545a8926c6f6L14
Since it's rewritten in Rust, this won't fail anymore, but there might be other places (also in different packages) where the toolchain is expected to handle globs, just like on Linux.

lazka:
Thanks, from what I understand enabling wildcards in gcc would also fix this case? https://github.com/gcc-mirror/gcc/blob/7f67acf60c5429895d7c9e5df81796753e2913e0/gcc/configure.ac#L2263

mati865:
Sounds likely but won't work for Binutils tools that don't go through GCC like when calling `ar`, but that might be an uncommon thing to do.

lazka:
let's see :)

PR #22267, opened 2024-10-22
qt6: update to qt 6.8.0

sunderme:
Update to qt 6.8.0

updated patches for base and pdf
added patch for base to fix bug (QTBUG-130319)
packet for languageserver changed as qt wants to build static library only


Biswa96:
May I split this PR into two parts? CI can not build all packages at once.

Question. Why did you remove the debug package in mingw-w64-qt6-languageserver ?

sunderme:
Yes, you can split the PR.

Languageserver now does not build dlls, only static libraries (.a)
There are no .debug files and the packaging part fails with an error.
I tried to patch it to build dlls again, but then qdeclarative failed.

Biswa96:
> added patch for base to fix bug (QTBUG-130319)

That upstream bug was fixed in different way than you patch https://codereview.qt-project.org/c/qt/qtbase/+/596940. Would it be similar if I added the patch from upstream commit?

sunderme:
upstream leads to the same solution, so that works as well

PR #22268, opened 2024-10-22
typst: update to 0.12.0

podsvirov:


podsvirov:
For `CLANGARM64`:
```
error: failed to run custom build command for `psm v0.1.23`
```
Looks like [psm](https://crates.io/crates/psm) not supperted for `aarch64` on `Windows` currently.

lazka:
Should we drop for clangarm64?

PR #22272, opened 2024-10-23
binutils: Add security patches

lhmouse:
Before this commit, the `.CRT` section (which contains pointers to image TLS callbacks) and the `.idata` section (which contains `__imp_` pointers to dllimport functions) that had been created by the GNU linker were writeable.

As these function pointers are essential for programs to be functional, they could be exploited to inject malicious code, like the well-known IFUNC backdoor in XZ Utils. The Microsoft linker does not create these sections as writeable; instead, they seem to be merged into `.rdata` and are not modifiable, unless unprotected.

LLD also does the same, suggesting these sections be merged into `.rdata`: https://github.com/llvm/llvm-project/blob/ebeb56af5f8f1ff9da8f5a7e98348f460d223de1/lld/COFF/Driver.cpp#L2034-L2048

This commit includes the following countermeasures:

1. Patch 3005 ensures the final `.CRT` and `.idata` sections (if any) will not be writeable in the final image.
2. Patch 3006 merges known sub-sections of `.CRT`, as well as `.ctors` and `.dtors` (which were merged into `.text` instead), into `.rdata`. Merging `.idata` into `.rdata` seems to prevent dllimport from working, so there's still an `.idata` section in the final image. See also: https://stackoverflow.com/questions/22651433/pe-idata-section

Reference: https://sourceware.org/bugzilla/show_bug.cgi?id=32264

### DIFF
--- a/mingw-w64-binutils/3005-Make-.CRT-and-.idata-read-only.patch
+++ b/mingw-w64-binutils/3005-Make-.CRT-and-.idata-read-only.patch
@@ -1,0 +1,31 @@
+From 4af2c5ca7d2df81c4ae4a1e63aab92c323a75a43 Mon Sep 17 00:00:00 2001
+From: LIU Hao <lh_mouse@126.com>
+Date: Tue, 22 Oct 2024 15:06:17 +0800
+Subject: [PATCH] Make `.CRT` and `.idata` read-only
+
+Signed-off-by: LIU Hao <lh_mouse@126.com>
+---
+ bfd/peXXigen.c | 3 ++-
+ 1 file changed, 2 insertions(+), 1 deletion(-)
+
+diff --git a/bfd/peXXigen.c b/bfd/peXXigen.c
+index 57cb3f5a184..2f503d264a6 100644
+--- a/bfd/peXXigen.c
++++ b/bfd/peXXigen.c
+@@ -995,11 +995,12 @@ _bfd_XXi_swap_scnhdr_out (bfd * abfd, void * in, void * out)
+ 
+     pe_required_section_flags known_sections [] =
+       {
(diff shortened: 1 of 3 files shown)


lhmouse:
I have rebuilt binutils, CRT, GCC and binutils again in MINGW32, MINGW64 and UCRT64, and no error has been observed.

Please examine this carefully.

lhmouse:
Previous patch broke pseudo relocation, as `__RUNTIME_PSEUDO_RELOC_LIST_END__` was defined to be the immediate end of `.rdata` (and why was it not just `__rt_psrelocs_end`):

https://github.com/msys2/MINGW-packages/commit/51247e19ff0845c576505c3499088605ecafb823#diff-a641beeafacd689f0c4ac5f1141e724abbce202d57be6339df410b00401971f8R121-R131

lhmouse:
Also note that the pseudo relocator code which unprotects read-only memory was added in 2009, much later than the relocator itself: https://github.com/mingw-w64/mingw-w64/commit/e2097782b75046dba93d9458dd085f3e8fa94907

This might explain why those sections were required to be writeable. They no longer are.

lhmouse:
I am closing this as the patches have been committed to binutils trunk: https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=db00f6c3aceabbf03acdb69e74b59b2d2b043cd7

lazka:
We could still backport them if wanted

lhmouse:
> We could still backport them if wanted

I don't have a very strong opinion on that. It has been like that for two decades.

lazka:
binutils 2.44, which includes this patch, is now in the repo

PR #22252, opened 2024-10-22
add libspng

totaam:
This library can be used by [`xpra`](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-xpra/PKGBUILD) to speedup `png` (de)compression.

PR #22284, opened 2024-10-24
update xpra dependencies

totaam:
* `libspng` added as it is now available: #22252 (merged yesterday)
* `python-lz4` is no longer used by xpra: https://github.com/Xpra-org/xpra/issues/3601
* `python-rencode` was also removed a while back: https://github.com/Xpra-org/xpra/commit/9e040307d968b2a054bff984d0a3f303e1339730
* update info strings for gstreamer modules: these are used for both audio and video since v5: https://github.com/Xpra-org/xpra/issues/3706
* add optional dependency on `aioquic` (merged yesterday: #22254) to provide the QUIC transport: https://github.com/Xpra-org/xpra/blob/8926b62f435548f05cb78df061c23bf54f5b67b3/docs/Network/QUIC.md


totaam:
> Does that match your use case?

Yes, I think so. Most people use TCP, websockets, SSL or SSH.
QUIC is much newer and somewhat more difficult to setup.

PR #22297, opened 2024-10-25
[new package] ada 2.9.2

ognevny:
required for telegram-desktop


mmuetzel:
Would it make sense to chose a name for the package that makes it clear that this is not a compiler for the programming language of that name:
https://en.wikipedia.org/wiki/Ada_(programming_language)

Maybe `ada-url`?

PR #22316, opened 2024-10-26
opencascade: update to 7.8.1

mmuetzel:
And rebuild dependencies.

The dependencies are pretty large. The CI will probably timeout.
Additionally, there is a cyclic dependency to VTK. (The updated OpenCASCADE should be built first.)


mmuetzel:
CI started with the wrong package...

I built the packages locally. (Albeit a couple of days ago before the Qt update. And not for all environments.) So, I hope they will also build for staging.

PR #22063, opened 2024-09-28
scotch: switch to cmake

3rav:

PR #22321, opened 2024-10-27
scotch: update & switch to cmake

3rav:
Draft:
https://github.com/msys2/MINGW-packages/pull/22063


mingw-w64-scotch/PKGBUILD
```diff
@@ -36,111 +41,82 @@ prepare() {
...
 }
 
-package_scotch-int64() {
+package_scotch64() {
   pkgdesc="Graph partitioning and sparse matrix ordering package with 64-bit indexing (mingw-w64)"
-  provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-  conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-  _package_scotch "64"
+  provides=()
+  conflicts=()
```
mmuetzel, line 110: The package with the new name should probably conflict, replace, and provide(?) the old name of this package.

mmuetzel:
The package grokker indicates that the following package needs a rebuild:
`mingw-w64-mumps`

(Potentially, because one library seems to be split into three libraries now?)

mmuetzel:
IIUC, the previous build rules manually created package config files for scotch and ptscotch. They are not created with the CMake build rules.
By the looks of it, mumps requires these .pc files.

Maybe, you could keep the part of the old build rules that created these files manually?

mingw-w64-scotch/0005-name-suffix.patch
```diff
@@ -0,0 +1,220 @@
...
+   EXPORT ptesmumpsTargets
+diff -urN scotch-v7.0.5_orig/src/libscotch/CMakeLists.txt scotch-v7.0.5/src/libscotch/CMakeLists.txt
+--- scotch-v7.0.5_orig/src/libscotch/CMakeLists.txt	2024-08-13 11:00:50.000000000 +0200
++++ scotch-v7.0.5/src/libscotch/CMakeLists.txt	2024-10-27 11:15:02.323859300 +0100
+@@ -567,7 +567,7 @@
+   target_compile_definitions(scotch PRIVATE SCOTCH_DEBUG_LIBRARY1)
+ endif()
+ 
+-add_library(scotcherr library_error.c)
++add_library(scotcherr STATIC library_error.c)
```
mmuetzel, line 71: IIUC, you'd like that the shared library also exports the symbols from this source file without having it be its own (shared) library.
If that is your goal here, it would probably be better to have `scorcherr` be an `OBJECT` library target instead.
Otherwise, users would need to link to the shared library *and* this static library.

3rav:
Info from scotch about .pc
https://gitlab.inria.fr/scotch/scotch/-/issues/37#note_1103036

> IIUC, the previous build rules manually created package config files for scotch and ptscotch. They are not created with the CMake build rules. By the looks of it, mumps requires these .pc files.
> 
> Maybe, you could keep the part of the old build rules that created these files manually?

mmuetzel:
> Info from scotch about .pc https://gitlab.inria.fr/scotch/scotch/-/issues/37#note_1103036
> 
> > IIUC, the previous build rules manually created package config files for scotch and ptscotch. They are not created with the CMake build rules. By the looks of it, mumps requires these .pc files.
> > Maybe, you could keep the part of the old build rules that created these files manually?

Like I wrote, the previous build rules created the package config files *manually*. See this part of the `package` function before your changes:
```
  echo "
	prefix=${MINGW_PREFIX}
	libdir=\${prefix}/lib
	includedir=\${prefix}/include
	Name: scotch
	URL: ${url}
	Version: ${pkgver}
	Description: Serial graph partitioning and sparse matrix ordering package
	Cflags: -I\${includedir}
	Libs.private: -lscotcherr
	Libs: -L\${libdir} -lscotch
  " | sed '/^\s*$/d;s/^\s*//' > "${pkgdir}${MINGW_PREFIX}/lib/pkgconfig/scotch.pc"

  if [[ ${CARCH} != aarch64 ]]; then
    echo "
    prefix=${MINGW_PREFIX}
    libdir=\${prefix}/lib
    includedir=\${prefix}/include
    Name: ptscotch
    URL: ${url}
    Version: ${pkgver}
    Description: Parallel graph partitioning and sparse matrix ordering package
    Requires: scotch
    Requires.private: msmpi
    Cflags: -I\${includedir}
    Libs.private: -lptscotcherr
    Libs: -L\${libdir} -lptscotch
    " | sed '/^\s*$/d;s/^\s*//' > "${pkgdir}${MINGW_PREFIX}/lib/pkgconfig/ptscotch.pc"
fi
```

So, no surprise that upstream doesn't know about files that are only created by the MSYS2 packaging rules...

3rav:
Ok, now I see that the files were created manually,  is it also possible to do it this way for the cmake version?

mmuetzel:
> Ok, now I see that the files were created manually, is it also possible to do it this way for the cmake version?

I don't know why it wouldn't be possible.
You might need to check if all libraries are called the same and are located at the same directory as before though, and potentially update the corresponding lines in the .pc file generation.

raedrizqie:
> Ok, now I see that the files were created manually, is it also possible to do it this way for the cmake version?

You need to provide *.pc.in file with every required data in it. Then just use cmake's `configure_file` function to substitue the data in *.pc.in and generate *.pc file from it.

PR #22325, opened 2024-10-27
rust: enable Control Flow Guard on gnullvm

mati865:
It works but not sure whether we should do it.

### DIFF
--- a/mingw-w64-rust/config.toml
+++ b/mingw-w64-rust/config.toml
@@ -50,8 +50,8 @@ lld = false
 codegen-tests = false
 deny-warnings = false
 backtrace-on-ice = true
-# FIXME: CFG can be enabled only for MSVC targets as for https://github.com/rust-lang/rust/pull/74103
-# control-flow-guard = true
+# FIXME: CFG is enabled for gnullvm targets only
+#control-flow-guard = true
 
 [dist]
 compression-formats = ["gz"]
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -20,7 +20,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-rust-wasm")
          "${MINGW_PACKAGE_PREFIX}-rust-src")
 pkgver=1.83.0
-pkgrel=2
+pkgrel=3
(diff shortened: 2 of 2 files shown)


ognevny:
for a future reference, why can't we enable it for other environments? it depends on support from local mingw-w64 runtime, right?

mingw-w64-rust/config.toml
```diff
@@ -48,8 +48,7 @@ lld = false
 codegen-tests = false
 deny-warnings = false
 backtrace-on-ice = true
-# FIXME: CFG can be enabled only for MSVC targets as for https://github.com/rust-lang/rust/pull/74103
```
ognevny, line 51: please, don't remove FIXME comment. it can be changed to `CFG is enabled for gnullvm targets only`

mingw-w64-rust/0002-allow-CFGuard-on-windows-gnullvm.patch
```diff
@@ -0,0 +1,31 @@
...
+index 3fc153c6cd4..7634d3c88ad 100644
+--- a/compiler/rustc_codegen_llvm/src/context.rs
++++ b/compiler/rustc_codegen_llvm/src/context.rs
+@@ -280,8 +280,12 @@ pub(crate) unsafe fn create_module<'ll>(
+         }
+     }
+ 
+-    // Control Flow Guard is currently only supported by the MSVC linker on Windows.
+-    if sess.target.is_like_msvc {
++    // Control Flow Guard is currently only supported by the MSVC and LLVM on Windows.
```
filnet, line 20:
```suggestion
+    // Control Flow Guard is currently only supported by MSVC and LLVM on Windows.
```

mati865:
> for building rust itself?

It's enabled for libstd (equivalent of libstdc++ from GCC), that is used by Rust and things built with Rust. The comment mentions only libstd, but I think it might be enabled for the compiler as well.

> for things built with rustc?

Not explicitly, it's similar to mingw-w64. Some checks will be pulled from the library, but to fully benefit from the feature (at a small performance penalty), it has to be enabled manually.

> for users to enable it with rustc?

It allows enabling that option.

> and are there any differences to the official upstream builds?

AFAICT upstream doesn't enable it in any provided builds.

ognevny:
the patch is merged upatream: https://github.com/rust-lang/rust/pull/132965. CFG can be enabled for gnullvm in 1.84.0 version. PR can be closed then

mati865:
It will be released early next year, so it depends whether we want to test it sooner.

Regardless, there is still no consensus whether it should be enabled.

PR #22336, opened 2024-10-29
box2d: update to 3.0.0

felixaszx:
This suppose to be an update of box2d, but I realize that there are so many breaking changes from 2.4 to 3.0. Therefore I think it is a good idea to add a new package.


felixaszx:
Please do not merge this yet, there are conflicts that are not yet found in my commits.

lazka:
> This suppose to be an update of box2d, but I realize that there are so many breaking changes from 2.4 to 3.0. Therefore I think it is a good idea to add a new package.

From a distro POV there would be no problem as we don't have any users of that package. Do you think there will be external users that will require the old version to stay around for longer?

felixaszx:
> > This suppose to be an update of box2d, but I realize that there are so many breaking changes from 2.4 to 3.0. Therefore I think it is a good idea to add a new package.
> 
> From a distro POV there would be no problem as we don't have any users of that package. Do you think there will be external users that will require the old version to stay around for longer?

From my POV, there should not be much people using it. Yeah, just being a bit too conservative here.  Maybe it is better to just update box2d to the lastest version anyway.

lazka:
OK, I'd vote for just replacing it. If someone complains we can add a 2.4 variant back.

PR #22340, opened 2024-10-29
poppler:update to 24.10.0

### LINKED ISSUE #22328: [poppler] update to 24.10.
### Package name

poppler

### New version number

24.10.0

### Other information that may be useful

I have updated and run PKGBUILD in my fork.
The problem is that inkscape does not build anymore as poppler uses now smartpointers for some functions.
What is the policy here ?

yes

sunderme:
* Fixes #22328


Biswa96:
Inkscape has merged the fix for newer poppler. Also, there is a new version of poppler. I am not sure if it is compatible with Inkscape yet.

Biswa96:
Just curious, is it required to update poppler so frequently? poppler breaks things with every release and all the reverse dependencies required to be rebuilt.

sunderme:
poppler updates every month and more often than not, they talk about "fix crash/memory handling".
I have not had any user reports on poppler issues recently but after all, it is used to handle data from the internet ...

mingw-w64-inkscape/PKGBUILD
```diff
@@ -6,7 +6,7 @@ _realname=inkscape
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.4
-pkgrel=1
+pkgrel=3
```
Biswa96, line 9:
```suggestion
pkgrel=2
```

sunderme: it is 
`
pkgver=1.4
pkgrel=2
` in master ?

Biswa96: Oh, I did not notice it. Then, it's ok.

sunderme:
now gdcm fails

sunderme:
and another bug report: https://gitlab.freedesktop.org/poppler/poppler/-/issues/1537

sunderme:
and now inkscape/clang64 fails to build, i have included  the official patch

sunderme:
looks like c++20 & pdf2djvu issue ?

lazka:
> Since pdf2djvu is out of maintenance for 2 years, this is out of my scope.

I've dropped the clang builds in 014a8c2ae0157625f3867874bfce6f224c8ad152

lazka:
dia rebuild is missing, I'll push that separately

PR #22342, opened 2024-10-29
hdf5: update to 1.14.5

mmuetzel:


rouault:
For your information, this update seems to cause test failures in the GDAL continuous integration: https://github.com/OSGeo/gdal/issues/11181

mmuetzel:
I don't know what exactly is happening in the tests you pointed out. What exactly is done that results in that error?

IIUC, HDF5 introduced a new data type in 1.14.5:
https://github.com/HDFGroup/hdf5/blob/hdf5_1_14_5/release_docs/HISTORY-1_14.txt#L68-L93

Maybe, that shuffled around some enums? Potentially, more packages need to be rebuilt.

rouault:
@mmuetzel  I've now done a local GDAL build against mingw-w64-x86_64-hdf5-1.14.5-1 and I now get the following detailed errors (truncated to the first ones emitted by libhdf5). The following doesn't involve Python at all and uses the libhdf5 C API:
```
$ gdalmdiminfo ../autotest/gdrivers/data/hdf5/u8be.h5
{
  "type": "group",
  "driver": "HDF5",
  "name": "/"HDF5-DIAG: Error detected in HDF5 (1.14.5):
  #000: C:/M/B/src/hdf5-1.14.5/src/H5Gdeprec.c line 971 in H5Gget_objinfo(): can't get info for object: 'TestArray'
    major: Symbol table
    minor: Can't get value
  #001: C:/M/B/src/hdf5-1.14.5/src/H5VLcallback.c line 4648 in H5VL_group_optional(): unable to execute group optional callback
    major: Virtual Object Layer
    minor: Can't operate on object
  #002: C:/M/B/src/hdf5-1.14.5/src/H5VLcallback.c line 4616 in H5VL__group_optional(): unable to execute group optional callback
    major: Virtual Object Layer
    minor: Can't operate on object
```

Interestingly we also have a fedora:rawhide CI target (https://github.com/OSGeo/gdal/actions/runs/11599287039/job/32297104747) that is at hdf5 == 1.14.5-1.fc42 and those tests pass just fine there.

mmuetzel:
If I'm reading their packaging rules correctly, they are building with autotools:
https://src.fedoraproject.org/rpms/hdf5/blob/rawhide/f/hdf5.spec

MSYS2 is using their CMake build system. I wonder if that makes a difference?

Fwiw, that file cannot be read using HDF5's tools directly either:
```
$ h5stat u8be.h5
Filename: u8be.h5
h5stat error: unable to traverse objects/links in file "u8be.h5"
```

rouault:
> I wonder if that makes a difference?

I build official 1.14.5 on my Linux system (Ubuntu 20.04) using CMake and h5stat works fine on the file.

mmuetzel:
If I downgrade to HDF5 1.14.3, it still works:
```
$ pacman -U /var/cache/pacman/pkg/mingw-w64-x86_64-hdf5-1.14.3-1-any.pkg.tar.zst
[...]
$ h5stat -f u8be.h5
Filename: u8be.h5
File information
        # of unique groups: 1
        # of unique datasets: 1
        # of unique named datatypes: 0
        # of unique links: 0
        # of unique other: 0
        Max. # of links to object: 1
        Max. # of objects in group: 1
```
So definitely a regression from the update...
Not sure where though or how to avoid it.

mmuetzel:
Rebuilt 1.14.3 with the current toolchain (in a MINGW64 environment): The file can still be read successfully by `h5stat`. 
So, it's not any dependencies. More likely some change in upstream HDF5.

rouault:
manually checkout hdf5-1_14_4 and this is a bad one as well

rouault:
987a734e759c82c65a661ae6090b2252d63a7aec is a good one

rouault:
Victory... or almost ... so I managed to get a conclusive git bisect session that points to:
```
$ git bisect bad
b69c6fcbf624222d9c00583cf6f1f386e80fa86b is the first bad commit
commit b69c6fcbf624222d9c00583cf6f1f386e80fa86b (HEAD)
Author: Larry Knox <lrknox@hdfgroup.org>
Date:   Thu Mar 21 09:17:22 2024 -0500

    1 14 dev sync lrk (#4200)

[ ... snip ... hundreds of commit message lines !!!! .... ]
 479 files changed, 14745 insertions(+), 6114 deletions(-)
```

mmuetzel:
That's... something.

They might have squashed these 100+ commits for some reason:
https://github.com/HDFGroup/hdf5/pull/4200/commits

That makes it hard now...
Do you have an idea how to handle that?

rouault:
> Do you have an idea how to handle that?

Digging further requires too deep expertise that I don't have. I've reported the issue upstream at https://github.com/HDFGroup/hdf5/issues/5052

The failure is not on all HDF5 files luckily. This seems to be specific to this file using this odd data type H5T_STD_U8BE (unsigned integer 8 bit big endian)

Was the update to 1.14.5 motivated by a particular reason, or just to keep up with things?
Because I really question upstream choice of changing thousands of files in *patch* releases and adding a new data type (Float16) that requires changes in downstream packages. Seems a total violation of semantic versioning expectations.

In my own point of view, reverting to working 1.14.3 could be a resolution.

mingw-w64-hdf5/PKGBUILD
```diff
@@ -69,19 +74,29 @@ build() {
...
+  fi
+
   CFLAGS+=" -Wno-implicit-function-declaration" \
   MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
     ${MINGW_PREFIX}/bin/cmake \
       -Wno-dev \
       -GNinja \
       -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
       -DCMAKE_DLL_NAME_WITH_SOVERSION=ON \
+      -DCMAKE_C_FLAGS="-Wno-error=incompatible-pointer-types" \
```
rouault, line 94: That's scary... I believe the issue is only present in sample programs. I believe I avoided that by adding ``-DHDF5_BUILD_EXAMPLES=OFF -DHDF5_TEST_EXAMPLES=OFF``

mmuetzel: IIRC, the point were I saw that error could have been resolved by just casting to the "correct" pointer type. IIUC, that is what is happening when that warning isn't treated as an error (like it was before GCC 14).

I only had a superficial look at that part of the sources though.
Might be worth reporting upstream (in case they didn't fix it already).

mmuetzel:
> Was the update to 1.14.5 motivated by a particular reason, or just to keep up with things?

To be honest, there was no particular reason for the update other than that the existing rules didn't build any more.

> In my own point of view, reverting to working 1.14.3 could be a resolution.

If the issue is limited to just one particular (obscure) data type, we could also wait a couple of days if upstream finds a fix and cherry-pick that.
Is the failing test blocking further development for you?

rouault:
> If the issue is limited to just one particular (obscure) data type, we could also wait a couple of days if upstream finds a fix and cherry-pick that.

sounds good

> Is the failing test blocking further development for you?

no, I've disabled HDF5 support for now from GDAL's msys2 CI configuration (we have tons of other CI configurations that test HDF5 support)

PR #22349, opened 2024-10-30
CI: change order of jobs

ognevny:
so it will be sorted by priority: UCRT64 -> CLANG64 -> MINGW64 -> MINGW32 -> CLANG32. this changes the default job shown in "checks" tab, as it's a bit annoying to switch to UCRT64 job because MINGW64 is disabled for a particular package

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,10 +15,10 @@ jobs:
       fail-fast: false
       matrix:
         include: [
-          { msystem: MINGW64, runner: windows-2022 },
-          { msystem: MINGW32, runner: windows-2022 },
           { msystem: UCRT64, runner: windows-2022 },
           { msystem: CLANG64, runner: windows-2022 },
+          { msystem: MINGW64, runner: windows-2022 },
+          { msystem: MINGW32, runner: windows-2022 },
           { msystem: CLANG32, runner: windows-2022 },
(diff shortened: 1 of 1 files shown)


ognevny:
how it looks with the change

screen

felixaszx:
I believe tests are running in parallel anyway.
It is not really necessary to order them.

ognevny:
> this changes the default job shown in "checks" tab, as it's a bit annoying to switch to UCRT64 job because MINGW64 is disabled for a particular package

I maintain some packages, that don't have MINGW64 version, so this is actual reason

felixaszx:
> > this changes the default job shown in "checks" tab, as it's a bit annoying to switch to UCRT64 job because MINGW64 is disabled for a particular package
> 
> I maintain some packages, that don't have MINGW64 version, so this is actual reason
Ahh, I got your point. Just my personal oppinion, is it a good idea to swap the priority of MINGW64 and CLANG64?

ognevny:
no, because MINGW64 uses MSVCRT, which is older, than UCRT, which is used in all CLANG environments (including CLANG32, but we're about to drop it, so it's in the lowest priority)

felixaszx:
> no, because MINGW64 uses MSVCRT, which is older, than UCRT, which is used in all CLANG environments (including CLANG32, but we're about to drop it, so it's in the lowest priority)

I see, thank you for replying.

lazka:
sure, why not, thanks

PR #22355, opened 2024-10-30
godot: update to 4.3

felixaszx:
Update to the lastest godot game engine, PKGBUILD is modofied from Arch Linux repo.


lazka:
the build issue seems to be this one: https://github.com/godotengine/godot/issues/97882

felixaszx:
> the build issue seems to be this one: [godotengine/godot#97882](https://github.com/godotengine/godot/issues/97882)

Oh thank you for finding that out, I was desperately digging other sources and not realize this. I will now work on a fix as soon as possible.

lazka:
What's the reasoning behind removing all the dependencies?

felixaszx:
> What's the reasoning behind removing all the dependencies?

Godot also provide source of those libraries in their repository. Since MSYS2 only provides the binary package of the engine; I think it is better to use their specific version of those libraries to avoid potential compability issues.

lazka:
Hm, ok. Normally we try to use system libs if that is supported upstream, so it also includes our patches and gets security updates more easily. But if they support this setup and it makes maintenance easier (given no one bothered to update it in a while) we could make an exception.

felixaszx:
> Hm, ok. Normally we try to use system libs if that is supported upstream, so it also includes our patches and gets security updates more easily. But if they support this setup and it makes maintenance easier (given no one bothered to update it in a while) we could make an exception.

I see, then I will try to make using external libs possible in the next following days.

Kreijstal:
> > Hm, ok. Normally we try to use system libs if that is supported upstream, so it also includes our patches and gets security updates more easily. But if they support this setup and it makes maintenance easier (given no one bothered to update it in a while) we could make an exception.
> 
> I see, then I will try to make using external libs possible in the next following days.

It is desirable to use system libs (libraries provided by msys2 rather than vendored libraries for packaging purposes), in case of bugs we can raise bug reports or patch the code itself.

felixaszx:
It now depends on MSYS2 libs, but a few of them are still compiled using godot's sources. Because those  libraries are not presented in MSYS2 repository or they require too many patches to work.

Comparing to the official build, this build is linked dynamicly againts required libraries.

mingw-w64-godot/PKGBUILD
```diff
@@ -1,104 +1,102 @@
...
+url='https://godotengine.org/'
 msys2_references=(
   "cpe: cpe:/a:godotengine:godot"
   "cpe: cpe:/a:godotengine:godot_engine"
 )
 license=('spdx:MIT')
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64')
+makedepends=("make"
+             "${MINGW_PACKAGE_PREFIX}-gcc"
```
Kreijstal, line 20: cc instead of gcc, think clang
```suggestion
             "${MINGW_PACKAGE_PREFIX}-cc"
```

lazka:
Perfect, thank you

Kreijstal:
image
Not sure if this is related

felixaszx:
> image Not sure if this is related

This seems like an issue with your graphics driver. If it is running on your VM, makey sure you are running it on hyper-v with a working GPU-Partitioning setup.

Kreijstal:
> This seems like an issue with your graphics driver. If it is running on your VM, makey sure you are running it on hyper-v with a working GPU-Partitioning setup.

it works if I try opengl3 instead it says opengl3 fails (it doens't crash and switches to ANGLE), but it works, yes it is a VM, under KVM. With Remote-Desktop. 

Not sure if this is a godot bug or a packaging bug

felixaszx:
> > This seems like an issue with your graphics driver. If it is running on your VM, makey sure you are running it on hyper-v with a working GPU-Partitioning setup.
> 
> it works if I try opengl3 instead it says opengl3 fails (it doens't crash and switches to ANGLE), but it works, yes it is a VM, under KVM. With Remote-Desktop.
> 
> Not sure if this is a godot bug or a packaging bug

Can you ensure that your VM has a proper Vulkan driver working?

PR #22362, opened 2024-10-31
svt-av1: update to 2.3.0

kmilos:


Biswa96:
The total file size went double with the version update https://github.com/msys2/MINGW-packages/actions/runs/11614724193/job/32343561041?pr=22362#step:11:1281. Could you know the reason for that change?

lazka:
Is the LTO with the static build actually usable? Maybe we could disable it there?

lazka:
ok, we this becomes more common we might have to rethink static packaging.

lazka:
There is fallout reported in #22439

PR #22368, opened 2024-11-01
python3.12: add `EXTERNALLY-MANAGED` for PEP 668

naveen521kk:
mintty_uiVaiD9ym6

### DIFF
--- a/mingw-w64-python3.12/EXTERNALLY-MANAGED
+++ b/mingw-w64-python3.12/EXTERNALLY-MANAGED
@@ -1,0 +1,15 @@
+[externally-managed]
+Error=To install Python packages system-wide, try 'pacman -S
+ $MINGW_PACKAGE_PREFIX-python-xyz', where xyz is the package you
+ are trying to install.
+
+ If you wish to install a non-MSYS2-packaged Python package,
+ create a virtual environment using 'python -m venv path/to/venv'.
+ Then use path/to/venv/bin/python and path/to/venv/bin/pip.
+
+ If you wish to install a non-MSYS2 packaged Python application,
+ it may be easiest to use 'pipx install xyz', which will manage a
+ virtual environment for you. Make sure you have $MINGW_PACKAGE_PREFIX-python-pipx
+ installed via pacman.
+
+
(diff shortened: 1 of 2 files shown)


mingw-w64-python3.12/EXTERNALLY-MANAGED
```diff
@@ -0,0 +1,15 @@
...
+Error=To install Python packages system-wide, try 'pacman -S
+ $MINGW_PACKAGE_PREFIX-python-xyz', where xyz is the package you
+ are trying to install.
+
+ If you wish to install a non-MSYS2-packaged Python package,
+ create a virtual environment using 'python -m venv path/to/venv'.
+ Then use path/to/venv/bin/python and path/to/venv/bin/pip.
+
+ If you wish to install a non-MSYS2 packaged Python application,
+ it may be easiest to use 'pipx install xyz', which will manage a
```
ognevny, line 11: do we actually have pipx?

naveen521kk: Oh, we don't have? I assumed we had it 😅.

I guess we should first package `pipx`.

naveen521kk: I've created https://github.com/msys2/MINGW-packages/pull/22371

mingw-w64-python3.12/EXTERNALLY-MANAGED
```diff
@@ -0,0 +1,15 @@
...
+ virtual environment for you. Make sure you have python-pipx
```
ognevny, line 12:
```suggestion
 virtual environment for you. Make sure you have $MINGW_PACKAGE_PREFIX-python-pipx
```

lazka:
As I said in chat, I'm slightly worried that this will break many CI setups, since venvs in MSYS2 aren't always that nice to work with due to missing binary wheels. Maybe I'm wrong though.

One thing we could do is patch pip (https://github.com/pypa/pip/blob/420435903ff2fc694d6950a47b896427ecaed78f/src/pip/_internal/utils/misc.py#L40) to warn instead of raise for starters (and maybe sleep a minute to be annoying)

naveen521kk:
I guess we can merge this, as we've patched up pip. https://github.com/msys2/MINGW-packages/pull/22382

lazka:
A PR to make the warning an error outside of CI: #27642

PR #22370, opened 2024-11-01
mcfgthread: Upgrade to 1.9.1

lhmouse:
https://github.com/lhmouse/mcfgthread/releases/tag/v1.9-ga.1

### DIFF
--- a/mingw-w64-mcfgthread/PKGBUILD
+++ b/mingw-w64-mcfgthread/PKGBUILD
@@ -4,14 +4,14 @@ _realname=mcfgthread
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}"-libs)
-pkgver=1.8.4
+pkgver=1.9.1
 pkgrel=1
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64' 'clang32')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://github.com/lhmouse/mcfgthread"
 license=('spdx:LGPL-3.0-or-later' 'custom')
 options=('staticlibs' 'strip' '!buildflags')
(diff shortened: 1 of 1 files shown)


lhmouse:
Is CLANG32 still supported? I see it's being dropped now: https://www.msys2.org/news/

Biswa96:
If you want clang32 binaries it could be kept.

lhmouse:
> If you want clang32 binaries it could be kept.

No I am not requesting you keep it. If it has been marked obsolete, should it be removed from CI?

Biswa96:
Please feel free to remove it as you see fit if it is not necessary. The build failure in clang32 does not seem related to msys2 https://github.com/msys2/MINGW-packages/actions/runs/11627215910/job/32380171035?pr=22370#step:11:432

lhmouse:
Yes it's a Clang bug:  https://github.com/llvm/llvm-project/pull/114533

And it's not the only one. Clang generates a 16-bit push for `push OFFSET _do_call_once_seh_uhandler`, so I wonder whether it's worth fixing (by writing machine code, for example).

lhmouse:
I have removed `clang32` from `mingw_arch` now.

mingw-w64-mcfgthread/PKGBUILD
```diff
@@ -26,8 +26,8 @@ build() {
   [[ -d "${srcdir}/build-${MSYSTEM}" ]] && rm -rf "${srcdir}/build-${MSYSTEM}"
   mkdir -p "${srcdir}/build-${MSYSTEM}" && cd "${srcdir}/build-${MSYSTEM}"
 
-  export CFLAGS+=' -Os -g'
   MSYS2_ARG_CONV_EXCL="--prefix=" meson setup -Dbuildtype=plain  \
+    -Dc_flags='-Os -g' -Dc_link_flags='-Os'  \
```
ognevny, line 30:
```suggestion
    -Dc_args='-Os -g' -Dc_link_args='-Os'  \
```
but I don't know whether it rewrites CFLAGS environment variable either appends to it

ognevny:
nit: can you use versioned tag in `source` instead of commit sha?

lhmouse:
> nit: can you use versioned tag in `source` instead of commit sha?

Of course.

PR #22371, opened 2024-11-01
python-pipx: new package

naveen521kk:
and it's dependencies.

### DIFF
--- a/mingw-w64-python-pipx/PKGBUILD
+++ b/mingw-w64-python-pipx/PKGBUILD
@@ -1,0 +1,56 @@
+# Maintainer: Naveen M K <naveen521kk@gmail.com>
+
+_realname=pipx
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=1.7.1
+pkgrel=1
+pkgdesc="Install and Run Python Applications in Isolated Environments (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/pypa/pipx'
+msys2_repository_url='https://github.com/pypa/pipx'
+msys2_references=(
+  'pypi: pipx'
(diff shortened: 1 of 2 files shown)


mingw-w64-python-pipx/PKGBUILD
```diff
@@ -0,0 +1,57 @@
...
+  "${MINGW_PACKAGE_PREFIX}-python-argcomplete"
+  "${MINGW_PACKAGE_PREFIX}-python-packaging"
+  "${MINGW_PACKAGE_PREFIX}-python-platformdirs"
+  "${MINGW_PACKAGE_PREFIX}-python-colorama"
+)
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-python-build"
+  "${MINGW_PACKAGE_PREFIX}-python-installer"
+  "${MINGW_PACKAGE_PREFIX}-python-hatchling"
+  "${MINGW_PACKAGE_PREFIX}-python-wheel"
```
ognevny, line 30: wheel is integrated into setuptools, so python-wheel can be removed

mingw-w64-python-userpath/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+  "${MINGW_PACKAGE_PREFIX}-python-wheel"
+)
+options=('!strip')
+source=("https://pypi.org/packages/source/${_realname::1}/${_realname}/${_realname}-${pkgver}.tar.gz")
+sha256sums=('6c52288dab069257cc831846d15d48133522455d4677ee69a9781f11dbefd815')
+
+prepare() {
+  cd "${_realname}-${pkgver}"
+
+  export SETUPTOOLS_SCM_PRETEND_VERSION=${pkgver}
```
ognevny, line 35: is this actually needed?

naveen521kk: I think so since `hatch-vcs` is used, see https://github.com/pypa/pipx/blob/99c998dc3b798559ebf602fdbe8bb5a24f1c98dd/pyproject.toml#L4

PR #22373, opened 2024-11-01
mcfgthread: apply small fixes

ognevny:
@lhmouse that's what I meant :) 
I suppose that tag has ${major}.${minor}-ga.${patch} format, so this pkgver expansion should be always correct. pushed repo checksums while at it

### DIFF
--- a/mingw-w64-mcfgthread/PKGBUILD
+++ b/mingw-w64-mcfgthread/PKGBUILD
@@ -5,13 +5,13 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}"-libs)
 pkgver=1.9.1
-pkgrel=1
+pkgrel=2
+pkgdesc="An efficient gthread implementation, providing C++11 threading support (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://github.com/lhmouse/mcfgthread"
-license=('spdx:LGPL-3.0-or-later' 'custom')
-options=('staticlibs' 'strip' '!buildflags')
-_tag='v1.9-ga.1'
+license=('spdx:LGPL-3.0-or-later WITH GCC-exception-3.1')
+options=('!buildflags')
 makedepends=(
   "${MINGW_PACKAGE_PREFIX}-cc"
   "${MINGW_PACKAGE_PREFIX}-meson"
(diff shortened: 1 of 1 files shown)


lhmouse:
Actually I hade been thinking about this for quite a while. For example, in a recent GCC version string, the third field is almost always zero. The branch state is indicated by the second field: 0 for pre-release, and 1+ for minor releases.

In my schedule, tags are named like 'v${abi}.${minor}-${suffix}'. The suffix is a string and it indicates the status of a branch (alpha, beta, rc, ga). There is not necessarily a revision or build number field. But as you have suggested, I can try to keep a revision suffix for ga tags.

ognevny:
so `ga` is for 'stable' releases, right? you can keep your system with beta, rc, etc. but the only version is built here is with ga suffix

lhmouse:
> so `ga` is for 'stable' releases, right? you can keep your system with beta, rc, etc. but the only version is built here is with ga suffix

yes, exactly.

Biswa96:
`_tag='v1.9-ga.1'` seems easier to read and comprehend compared to the pkgver magic.

ognevny:
for me it's better when you need to change pkgver only

lhmouse:
> for me it's better when you need to change pkgver only

I agree with this, like with other packages.

mingw-w64-mcfgthread/PKGBUILD
lhmouse, line 1:
```sh
package_mcfgthread() {
  pkgdesc="An efficient gthread implementation, providing C++11 threading support (devel)"
  depends=("${MINGW_PACKAGE_PREFIX}-${_realname}"-libs)
```

I just saw this today; I think `depends` should be 

```sh
  depends=("${MINGW_PACKAGE_PREFIX}-${_realname}-libs=${pkgver}")
```

lhmouse:
also the license in SPDX form is `LGPL-3.0-or-later WITH GCC-exception-3.1`.

ognevny:
everything is done, changed PR name since it now contains several changes. also bumped pkgrel

lhmouse:
What do `staticlibs` and `strip` mean in `options`?

ognevny:
> What do `staticlibs` and `strip` mean in `options`?

staticlibs: allow staticlibs in package (disabled by default in Arch, but enabled by default in MSYS2)
strip: run strip for each executable/library file (enabled by default)

lhmouse:
> staticlibs: allow staticlibs in package (disabled by default in Arch, but enabled by default in MSYS2) strip: run strip for each executable/library file (enabled by default)

Thanks for the explanation. So they are on by default. The change looks good to me.

ognevny:
> The change looks good to me.

all changes or the exact one?

lhmouse:
I meant this _commit_ looks good to me. 😅

ognevny:
> I meant this _commit_ looks good to me. 😅 

that's what "all changes" mean :)

lhmouse:
thanks; apologies for the confusion.

ognevny:
no problems

Biswa96:
> What do `staticlibs` and `strip` mean in `options`?

`man PKGBUILD` and `man makepkg.conf`

lhmouse:
> > What do `staticlibs` and `strip` mean in `options`?
> 
> 
> 
> `man PKGBUILD` and `man makepkg.conf`

I was actually a bit confused how `staticlib` should work. To me it looks like 'whether staticlibs should be _built_' rather than to _exclude_ them; however for this package there is no way to not build a static lib.

`strip` is for `strip --strip-unneeded` and is actually less confusing. (still a little because it's not `--strip-all`.)

PR #22382, opened 2024-11-02
pip: in case of EXTERNALLY-MANAGED only warn instead of erroring out

lazka:
We want users to move away from breaking their systems with pip, and upstream provides PEP 668 for this where pip will error out if the system packages are marked as externally managed (by pacman in this case).

In MSYS2 though it's not that easy to use venvs (mixing with system site packages is sometimes required), though things have improved with setuptools/meson-python/rust tooling in the last year. Also in many cases MSYS2 is mainly used in CI, so it's fine to break things as it's just a temporary VM.

To get the message still out there we downgrade the error to warning.

In the future we can escalate this further by making it an error in non-CI envs, and adding a sleep in CI, so it gets annoying but doesn't break things right away.

Screenshot 2024-11-02 112516


naveen521kk (approved):
Lgtm

PR #22387, opened 2024-11-02
objfw: Update to 1.2

Midar:


Midar:
I currently don't have access to a Windows machine, so it would be good if someone could check that all files indeed end up in the correct subpackages. Thanks!

Midar:
I currently don't have access to a Windows machine, so it would be good if someone could check that all files indeed end up in the correct subpackages. Thanks!

lazka:
CI logs show a diff of the included files compared to the previous version. Looks fine from a quick look.

Issue #19949, opened 2024-02-04
[nim] - new major version 2.0.2


MirKml:
### Description / Steps to reproduce the issue

Hello, it's possible to upgrade nim package? Current version is https://packages.msys2.org/base/mingw-w64-nim 1.6.12-1
upstream is 2.0.2.

Thanks

### Expected behavior

latest nim version will be available.

### Actual behavior

Old version is presented

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-22631 PC 3.4.10.x86_64 2023-12-22 10:06 UTC x86_64 Msys

### MINGW environments affected

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANG32
- [ ] CLANGARM64

ognevny:
I tried to update it a month ago. it failed at installation step without a meaningful error. I'll try again later

striezel:
Nim has been updated to version 2.2.6 in <https://github.com/msys2/MINGW-packages/commit/bf181d3f8b7f8cafeece538025940501427bcd04>, so this update request is finally fulfilled.

PR #22413, opened 2024-11-05
Update Nim to 2.2.0

VeronikaGrishina:
Fixes [#19949](https://github.com/msys2/MINGW-packages/issues/19949)
1. Updated "0003-Fix-library-path.patch" to match "options.nim";
2. Removed "0006-Fix-sqlite3-dll-name.patch" since there's no "sqlite3.dll" now;
3. Since 2.0 Nim [is using csources_v2](https://github.com/nim-lang/Nim/commit/3b965f463b41eace340393af636e4abcd9c223d6)


VeronikaGrishina:
Removing MINGW32 that does make things easier, but `package` fails for some reason on CLANG64 without a proper error message, while building goes well.

ognevny:
I see, that's just random segfault...

ognevny:
trying to run koch.exe manually
```console
$ src/Nim-2.2.4/koch.exe
oserrors.nim(92)         raiseOSError
Error: unhandled exception: Системе не удается найти указанный путь.
Additional info: lib [OSError]
```

MehdiChinoune:
I am going to update it and drop clang64.

MehdiChinoune:
I have pushed bf181d3f8b7f8cafeece538025940501427bcd04

PR #22415, opened 2024-11-05
pyside2: update to 5.15.15

mmuetzel:


MehdiChinoune:
It requires more that that to be compatible with Python 3.12
see: https://salsa.debian.org/qt-kde-team/qt/pyside2/-/commit/34279c27272f57c3be78ed71bde92c477bb3a894
and https://github.com/NixOS/nixpkgs/pull/327976

lazka:
I hope the Debian stuff is enough (??) let's see

mmuetzel:
@MehdiChinoune: Thank you for the pointers.
@lazka: Thank you for picking the patches.

PR #22421, opened 2024-11-06
python-py2exe: disable building for any environment

mmuetzel:
It is currently incompatible with Python 3.12 or newer:
https://github.com/py2exe/py2exe/issues/191


ewancg:
Seems like it could be re-enabled:
https://github.com/py2exe/py2exe/issues/191#issuecomment-3077581435

PR #22422, opened 2024-11-06
ruby: update to 3.3.6

raedrizqie:
- update ruby to latest version
- remove all patches that were already included in this version
- rebuild all externally packaged gems


lazka:
the commits contain much more than you mention without any explanation why it was changed, or why it is OK to remove certain things. I can't really review that.

raedrizqie:
@lazka i have updated the commit change

ognevny:
seems fine now, doesn't it?

mingw-w64-asciidoctor/PKGBUILD
```diff
@@ -4,10 +4,10 @@ _realname=asciidoctor
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=2.0.23
-pkgrel=1
+pkgrel=2
 pkgdesc="An implementation of AsciiDoc in Ruby (mingw-w64)"
 arch=('any')
-mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clang32' 'clangarm64')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
```
mmuetzel, line 10: `asciidoctor` is a build dependency of `ccache` which is a toolchain package. Those should be removed last.

JPeterMugaas:
I think the reasoning behind the provides is still valid.

ognevny:
@raedrizqie can you do `chmod +x` for `updpkgprovs`? it had 0755 mode before

ognevny:
IMO it's better to just drop clang32 here

mmuetzel:
> IMO it's better to just drop clang32 here

If we prefer to do that, we'd need to drop all reverse dependencies including the build dependencies.

If the consensus is that the CLANG32 environment as currently distributed by MSYS2 is broken beyond repair, it would be better imho to remove it in one fowl swoop instead of continue to distribute something that is broken. 

This is a larger decision than updating Ruby.

@lazka: What do you think?

raedrizqie:
One of the reason I want to update ruby is because of this warning when configuring `qtwebkit`:
```
...
2024-11-17T16:21:20.8217076Z -- Found Perl: D:/M/msys64/usr/bin/perl.exe (found suitable version "5.38.2", minimum required is "5.10.0")
2024-11-17T16:21:20.8599084Z -- Found PythonInterp: D:/M/msys64/clang64/bin/python3.exe (found suitable version "3.12.7", minimum required is "2.7.0")
2024-11-17T16:21:22.8290085Z -- Could NOT find Ruby (missing: Ruby_LIBRARY) (found suitable version "3.1.5", minimum required is "1.9")
...
```

mmuetzel:
It looks like the linker doesn't pull in any CRT libraries on CLANG32 for some reason. (I wonder why that is the only affected environment.)

The following change fixes the build for me:
[0001-ruby-link-UCRT.patch](https://github.com/user-attachments/files/17842815/0001-ruby-link-UCRT.patch)

mingw-w64-ruby/0010-configure-fixes.patch
```diff
@@ -0,0 +1,39 @@
+--- a/configure.ac
++++ b/configure.ac
+@@ -1361,7 +1361,7 @@
+ AC_CHECK_HEADERS(sys/event.h)
+ AC_CHECK_HEADERS(stdatomic.h)
+ 
+-AS_CASE("$target_cpu", [x64|x86_64|i[3-6]86*], [
++AS_CASE("$target_cpu", [x64|x86_64|[i[3-6]86*]], [
```
mmuetzel, line 8: Why is this change needed? Doesn't the pattern match a string with any of these characters now (instead of the characters in that specific order)?

raedrizqie: without the change, it will be translated to:
```
case "$target_cpu" in #(
  x64|x86_64|i3-686*) :
```
you can check it in `configure` file, which will make the check failed for i386

mmuetzel: Ah. Thanks for clarifying.

mingw-w64-ruby/updpkgprovs
mmuetzel, line 1: Could you please revert the change to the file mode and keep this script executable?

mmuetzel:
> Let me try something first.. if it failed, I will try your fix..

Nice.
So, this was caused by a new hunk of a patch?

raedrizqie:
> > Let me try something first.. if it failed, I will try your fix..
> 
> Nice. So, this was caused by a new hunk of a patch?

Yeah.. but the dll naming for i386 ucrt should be `ucrt-ruby330.dll`, without the extra prefx `i386-`
We can force it by adding;
```
    $( [[ ${MSYSTEM} == CLANG32 ]] && \
      echo '--with-soname=ucrt-$(RUBY_BASE_NAME)$(MAJOR)$(MINOR)0' || true )
```
to configure (like before)

mingw-w64-ruby/updpkgprovs
mmuetzel: I could force-push a change that resets that flag if that's ok with you...

raedrizqie: Ok.. thank you

mmuetzel:
In that case, should the same be done for CLANG64 which also has the extra prefixes now (from the file-diff section of the CI)?
```
  -/clang64/bin/ruby310.dll
   /clang64/bin/rubyw.exe
  +/clang64/bin/syntax_suggest
  +/clang64/bin/syntax_suggest.bat
   /clang64/bin/typeprof
   /clang64/bin/typeprof.bat
  +/clang64/bin/x64-ucrt-ruby330.dll
```

mingw-w64-ruby/updpkgprovs
mmuetzel: Done

raedrizqie:
> In that case, should the same be done for CLANG64 which also has the extra prefixes now (from the file-diff section of the CI)?
> 
> ```
>   -/clang64/bin/ruby310.dll
>    /clang64/bin/rubyw.exe
>   +/clang64/bin/syntax_suggest
>   +/clang64/bin/syntax_suggest.bat
>    /clang64/bin/typeprof
>    /clang64/bin/typeprof.bat
>   +/clang64/bin/x64-ucrt-ruby330.dll
> ```

For x64 ucrt based dll, the name is correct with `x64-ucrt-ruby330.dll`. Only i386 should strip the `i386-` prefix

mmuetzel:
> For x64 ucrt based dll, the name is correct with `x64-ucrt-ruby330.dll`. Only i386 should strip the `i386-` prefix, same as MINGW32 (`msvcrt-ruby330.dll`)

I don't know why the naming pattern should be different depending on the architecture. But I trust you know the reason.

raedrizqie:
> > For x64 ucrt based dll, the name is correct with `x64-ucrt-ruby330.dll`. Only i386 should strip the `i386-` prefix, same as MINGW32 (`msvcrt-ruby330.dll`)
> 
> I don't know why the naming pattern should be different depending on the architecture. But I trust you know the reason.

Yeah.. it is because of historical reasons.. see here: https://github.com/ruby/ruby/pull/11364#issuecomment-2283960114

ognevny:
I wonder, can Rust be used for YJIT here?

mmuetzel:
Would you like to take over again and restore that configure flag? (I could also amend your last commit if you prefer that.)

raedrizqie:
> Would you like to take over again and restore that configure flag? (I could also amend your last commit if you prefer that.)

Ok.. its easier if you could do it..

raedrizqie:
> I wonder, can Rust be used for YJIT here?

Sorry, I know nothing about rust :(

ognevny:
> Sorry, I know nothing about rust :(

from what I see the only required think is to add rust into makedepends. I can try it in separate PR

mmuetzel:
> > Would you like to take over again and restore that configure flag? (I could also amend your last commit if you prefer that.)
> 
> Ok.. its easier if you could do it..

Trying to do so leads to the old issue where no CRT seems to have been linked.

Tbh, I didn't understand from the link you showed earlier why the naming convention needs to differ for i386. What would be the issue if we just used the naming convention from upstream?

raedrizqie:
> > > Would you like to take over again and restore that configure flag? (I could also amend your last commit if you prefer that.)
> > 
> > 
> > Ok.. its easier if you could do it..
> 
> Trying to do so leads to the old issue where no CRT seems to have been linked.
> 

Ah.. I see.. It may be the way they hardcoded soname when compiling..

> Tbh, I didn't understand from the link you showed earlier why the naming convention needs to differ for i386.

Sorry, it should be this one.. https://github.com/ruby/ruby/commit/576b2e64cdc5ea42ad345dd3c1c215e006c06fca

> What would be the issue if we just used the naming convention from upstream?
> 

No issue actually.. upstream don't have CI job for testing x86 ucrt anyway..

> Edit: If I drop the `ucrt-` prefix, it seems to work...

Ok

mmuetzel:
> > What would be the issue if we just used the naming convention from upstream?
> 
> No issue actually.. upstream don't have CI job for testing x86 ucrt anyway..

So, could this be merged like it currently is?

raedrizqie:
> > > What would be the issue if we just used the naming convention from upstream?
> > 
> > 
> > No issue actually.. upstream don't have CI job for testing x86 ucrt anyway..
> 
> So, could this be merged like it currently is?

I think so, yes

raedrizqie:
Thank you!

ognevny:
> from what I see the only required thing is to add rust into makedepends. I can try it in separate PR

it's not possible currently: https://bugs.ruby-lang.org/issues/19325. I currently don't have a time to test it :(

PR #22424, opened 2024-11-06
python-jsonrpc-server: unbundle vendored versioneer

mmuetzel:
The bundled version is outdated and incompatible with Python 3.12.

### DIFF
--- a/mingw-w64-python-jupyter-nbviewer/PKGBUILD
+++ b/mingw-w64-python-jupyter-nbviewer/PKGBUILD
@@ -4,7 +4,7 @@ _realname=nbviewer
 pkgbase=mingw-w64-python-jupyter-${_realname}
 pkgname=${MINGW_PACKAGE_PREFIX}-python-jupyter-${_realname}
 pkgver=1.0.1
-pkgrel=5
+pkgrel=6
 pkgdesc='Render Jupyter Notebooks as static web pages (mingw-w64)'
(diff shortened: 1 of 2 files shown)


mmuetzel:
Same for `python-jupyter-nbviewer`.

mingw-w64-python-jupyter-nbviewer/PKGBUILD
```diff
@@ -27,11 +27,14 @@ depends=(
 makedepends=(
     ${MINGW_PACKAGE_PREFIX}-python-setuptools
     ${MINGW_PACKAGE_PREFIX}-python-invoke
+    ${MINGW_PACKAGE_PREFIX}-python-versioneer
 )
 source=("${_realname}-${pkgver}.tar.gz::https://github.com/jupyter/${_realname}/archive/${pkgver}.tar.gz")
 sha256sums=('2fe0b0ee8ec5c8e2e173b8f15a22e42ba583fc6d4fa60ddb7aa40e9ee972f268')
 
 prepare() {
+  rm -f "${srcdir}"/${_realname}-${pkgver}/versioneer.py | true
```
ognevny, line 36: nit: -f option ignore non existing files, so it will never fail. it can be `rm "${_realname}-${pkgver}"/versioneer.py` so in case script is removed upstream this line can also be removed

PR #22426, opened 2024-11-06
python-language-server: unbundle vendored versioneer

mmuetzel:
The bundled version is outdated and incompatible with Python 3.12.


mmuetzel:
It looks like the "check" step downloaded an old version of `python-jsonrpc-server` instead of the one from staging. Other than that this is looking ok afaict.

PR #22435, opened 2024-11-07
gimp3: Update to RC1

hmartinez82:
Needs https://github.com/msys2/MINGW-packages/pull/22434
image


ognevny:
would it be better if we wait for a final release?

mmuetzel:
> would it be better if we wait for a final release?

IIUC, we were packaging beta versions earlier. I don't know why a release candidate would be any worse.
But it could also be that I'm misunderstanding upstream's versioning scheme...

ognevny:
seems it won't be so soon, so it's fine to have it now

PR #22436, opened 2024-11-07
jsoncpp: disable object library

raedrizqie:
- disable object library
- install license file


lazka:
just wondering, does that rename work with cmake users?

raedrizqie:
ah.. i forgot about cmake compatibility (i was just considering pkgconfig pc file)

i think we need to fix this in cmake itself then

raedrizqie:
i have reverted the static lib name change

PR #22440, opened 2024-11-07
libheif: update to 1.19.2

kmilos:
Enable OpenH264 codec
Disable tests, fail to build


ognevny:
what happened to package size? (5 MiB -> 2.4 MiB)

PR #4302, opened 2018-08-24
gpgme: make mkdefsinc use CPPFLAGS

lovetox:
This fixes the build on the CI and some Windows systems

see #4165 

thanks to @vyache-slav for providing the patch

Issue #4165, opened 2018-08-08
GPGME package doesnt find gpg.exe


lovetox:
Hi,

This code shows that gpgme doesnt find gpg.exe

```
import gpg
def print_engine_infos():
    print("gpgme version:", gpg.core.check_version(None))
    print("engines:")

    for engine in gpg.core.get_engine_info():
        print(engine.file_name, engine.version)

    for proto in [gpg.constants.protocol.OpenPGP, gpg.constants.protocol.CMS]:
        print("Have {}? {}".format(gpg.core.get_protocol_name(proto),
                                   gpg.core.engine_check_version(proto)))
print_engine_infos()
```

This yields

```
gpgme version: 1.11.1
engines:
C:\msys64\mingw32\bin\gpgconf.exe 1.0.0
/nonexistent 1.0.0
Have OpenPGP? False
Have CMS? False
```

same in MSYS env yields

```
gpgme version: 1.11.1
engines:
/usr/bin/gpg 2.2.9-unknown
/usr/bin/gpgsm 2.2.9-unknown
/usr/bin/gpgconf 2.2.9-unknown
/nonexistent 1.0.0
Have OpenPGP? True
Have CMS? True
```

lovetox:
Weird thing is that it shows gpgconf.exe 1.0.0 and i checked in

`C:\msys64\mingw32\bin\gpgconf.exe` 2.2.9-unknown

sooo im out of ideas very weird

lovetox:
Ok investigated further, it seems we fail to query gpgconf

This is the problem i think

```
GPGME 2018-08-08 23:24:55 <0x1e30>  _gpgme_io_spawn: check: path=0050cab8, _gpgme_mkstemp failed: File exists
GPGME 2018-08-08 23:24:55 <0x1e30>  _gpgme_io_spawn: error: File exists
```

Full log

```
GPGME 2018-08-08 23:24:55 <0x1e30>  gpgme_debug: level=9
GPGME 2018-08-08 23:24:55 <0x1e30>  gpgme_debug: gpgme='C:\msys64\mingw32\bin'
GPGME 2018-08-08 23:24:55 <0x1e30>  gpgme_check_version: call: 0=00000000, req_version=(null), VERSION=1.11.1
GPGME 2018-08-08 23:24:55 <0x1e30>  gpgme_check_version_internal: call: 0=00000000, req_version=(null), offset_sig_validity=32
[...]

PR #22461, opened 2024-11-08
gpgme: add __USE_MINGW_ANSI_STDIO=1 to CPPFLAGS

jeremyd2019:
build tool mkdefsinc uses (sn)printf '%n', which is blocked for security
 reasons in UCRT (perhaps only recent versions).  Because it is a build
 tool (built for the build rather than target system in the case of a
 cross-compile), it does not respect CFLAGS, but it does respect
 CPPFLAGS.  Define __USE_MINGW_ANSI_STDIO=1 in CPPFLAGS so it uses that
 implementation that allows '%n'.

See https://github.com/msys2/MINGW-packages/pull/22442#issuecomment-2463055922, #4302, https://github.com/msys2/MINGW-packages/issues/4165#issuecomment-413152233


jeremyd2019:
https://github.com/msys2-arm/MINGW-packages/actions/runs/11747637878/job/32730007024

PR #22475, opened 2024-11-10
python-pygit2: Update to 1.16.0

lazka:


lazka:
I guess needs the newer libgit

PR #22479, opened 2024-11-10
tree-sitter: update to 0.24.4

ognevny:


ognevny:
it's annoying when github pulls such silly name of PRs :(

ognevny:
tree sitter is now split into library and CLI package as tree-sitter.exe became huge with wasm feature enabled

PR #22481, opened 2024-11-11
gimp3: Build GIMP non relocatable

brunvonlope:
This drops a superfluous .patch and build GIMP as non-relocatable, which is the expected way to run it in MSYS2 shells.

ping @hmartinez82


lazka:
This disables lots of relocation code in gimp (see `ENABLE_RELOCATABLE_RESOURCES`). Are you sure this is what we want?

brunvonlope:
> Are you sure this is what we want?

@lazka No, not sure. I'm new to MSYS2 GUI packages so that depends on what the MSYS2 project usually expects from its packages. They should be relocatable or non-relocatable?

If we suppose that the clang64 folder can be copied or moved freely, so `relocatable-bundle=yes`. If no so the project assumes that the clang64 works like a system prefix in the UNIX way, so `relocatable-bundle=no`.

mmuetzel:
All packages distributed by MSYS2 must be relocatable in the sense that a user can choose any prefix for their MSYS2 installation. For one user, that might result in `/c/msys2/ucrt64` as the prefix for their UCRT64 environment. For another user, that might be `/d/my_msys2/ucrt64`. For a third user, it might be something completely different.

Do you know what "relocatable" means for GIMP? Does it still support different prefixes even if it is built non-relocatable?

brunvonlope:
@mmuetzel Thanks. For GIMP, "relocatable" means exactly what you described. We will probably drop this bundling code being run by default or rework to be less pervasive, as @Jehan already pointed it to me on IRC.

brunvonlope:
Not needed anymore due to: https://gitlab.gnome.org/GNOME/gimp/-/commit/31218e69da938d0f93ef1c9543ab0f60c97ff85b

Closing.

Issue #22459, opened 2024-11-08
Add Nix


Ericson2314:
### Package name

Nix

### Brief description of package

Nix, the purely functional package manager.

Nix has a Windows port in progress, which I am involved with. It is not yet feature-complete, but it does build. We understand if packaging software that isn't yet useful is putting the cart before the horse, but we would love to better do native builds of Nix while we finish the port, and also learn a bit more about MSYS / toolchain setups in the process.

Once we're far enough to be attempting to build Nixpkgs on Windows natively with Cygwin/MSYS, we'll also need to learn a lot from you all in general. (See https://github.com/NixOS/nixpkgs/pull/354137 for the first steps on that.) So consider this a hybrid packaging requestion / general knowledge sharing and collaboration request.

(And of course, if someday you all want to ditch `pacman` for `nix`, we won't complain :))

### URL for package's homepage

https://github.com/nixos/nix

### Provide a basic test case to validate the package's functionality.

We have unit tests for the libraries that run, and some in-progress work to run the shell-script-based integration tests.

### MINGW environments where you need the package

- [X] MINGW64
- [X] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANGARM64

### Are you willing to submit a PR?

I think one of us porting Nix to Windows could do it but we would need hand-holding. Ideally one of you could drive the PR, and we would be highly available to answer questions.

lazka:
Would nix target the Windows APIs via C++, or is this a cygwin port?

Ericson2314:
@lazka No Cygwin! :tada: We're cross building from Linux with MinGW now, and a previous version of the port (https://github.com/nix-windows/nix/commit/windows) in fact got things building with MSVC.

The newish C++ `filesystem` library helps a lot, here :)

puffnfresh:
I'll take this on, since I've got an [automated build of Nix using mingw64](https://github.com/nix-windows/nix/commit/f493fe78d4fbd54f2f12eb6c87e6b0fd63877e66#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR61-R107). There are two dependencies missing:

* [ ] toml11 #22409
* [ ] libgit2, with a patch which is only in HEAD, not in a release yet #6875 (I think we should just wait)

Kreijstal:
[...]

PR #22492, opened 2024-11-11
libgit2: backport a feature for nix

lazka:
See https://github.com/msys2/MINGW-packages/issues/22459#issuecomment-2466576725


Ericson2314 (approved):
Thank you so much!!

PR #22513, opened 2024-11-12
emacs: fix libtree-sitter linking

### LINKED ISSUE #22508: libtree-sitter-0.24.4-1 and tree-sitter-0.24.4-1 breaks emacs

reproduce:
1. update msys2 to latest, {lib}tree-sitter 0.24.4-1 and emacs 29.4-2 included
2. open emacs
3. emacs says 
```
The following tree-sitter grammars are/were missing:
c
cpp
...

Install missing grammars? (y or n)
```
I can confirm the grammars are installed under `%appdata%\.emacs.d\tree-sitter`. Re-installation doesn't help. 
7. Uninstall libtree-sitter, downgrade tree-sitter to 0.24.3-2 and emacs to 29.4-1, and emacs works again.

Appendix:

I believe emacs was rebuilt, but something still went wrong.
https://github.com/msys2/MINGW-packages/commit/130eacba51530ea18fa514b6f3a8b79d3ae40be5

Appendix:

I run strace, and no tree-sitter relevant dll is loaded.

### Expected behavior

Emacs Works.

### Actual behavior

Emacs prompt

```
The following tree-sitter grammars are/were missing:
...
```

and syntax highlighting doesn't work.

### Verification

### Windows Version

MINGW64_NT-10.0-26100

- [ ] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [ ] CLANGARM64

I'll try if you find out the problem.

ognevny:
introducing soversion broke emacs build due to hardcoded library names. I'm actually not familiar with this syntax so I added hardcoded name with soversion that must be changed at every libtree-sitter update... contributions welcome 
fixes #22508

### DIFF
--- a/mingw-w64-emacs/004-libtree-sitter.patch
+++ b/mingw-w64-emacs/004-libtree-sitter.patch
@@ -1,0 +1,11 @@
+--- a/lisp/term/w32-win.el
++++ b/lisp/term/w32-win.el
+@@ -314,7 +314,7 @@
+ '(gccjit "libgccjit-0.dll")
+        ;; MSYS2 distributes libtree-sitter.dll, without API version
+        ;; number...
+-       '(tree-sitter "libtree-sitter.dll" "libtree-sitter-0.dll")))
++       '(tree-sitter "libtree-sitter.dll" "libtree-sitter-0.dll" "libtree-sitter-0.24.dll")))
+
+ ;;; multi-tty support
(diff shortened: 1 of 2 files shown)


raedrizqie:
why dont we just disable `-DCMAKE_DLL_NAME_WITH_SOVERSION=ON` on every package.. this is a mess

PR #22515, opened 2024-11-12
Lexbor 2.4.0

vtorri:


Biswa96:
I have tried to workaround the build issue but now it fails with clang. Are those extra compiler flags required?

vtorri:
the copy is necessaray to find the DLL, maybe updating the PATH would be sufficient ?

Biswa96:
The DLL issue has been workaround. It probably has some race condition - copying file before it is built. There is a new compiler error in clang.

vtorri:
in source/lexbor/ports/windows_nt/config.cmake, add

```
        set(LEXBOR_CXX_FLAGS "-Wall -pedantic -pipe")
```

just after `LEXBOR_C_FLAGS` line 14, and without your patch about copying the DLL. It works for me in the clang64 environment

PR #22533, opened 2024-11-14
zed: update to 0.161.2

ognevny:
and try to enable CLANGARM64


ognevny:
bruh, I thought ring is updated in this version. in 0.162.x it should be 0.17+ only...

ognevny:
thanks anyway, I put references in commit

PR #22545, opened 2024-11-15
python-milc: update to 1.9.0

zvecr:
Bumping to latest release as requirement for updating `mingw-w64-x86_64-python-qmk` to `1.1.6`.


Biswa96:
FYI, Multiple dependent packages can be updated/changed in same pull request. CI will build them according to the dependency.

PR #22548, opened 2024-11-15
scotch: update to 7.0.5

### LINKED ISSUE #22540: Does libscotch.dll from scotch-int64 and mingw-scotch-int64 packages really provide int64 interface?

I downloaded [mingw-w64-x86_64-mumps](https://packages.msys2.org/packages/mingw-w64-x86_64-mumps) package, and used it correctly with libscotch.dll from [mingw-w64-x86_64-scotch](https://packages.msys2.org/packages/mingw-w64-x86_64-scotch) package. Then I downloaded [mingw-w64-x86_64-scotch-int64](https://packages.msys2.org/packages/mingw-w64-x86_64-scotch-int64) and [mingw-w64-ucrt-x86_64-scotch-int64](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-scotch-int64), and these packages also provide libscotch.dll, but when I replace old libscotch.dll with the one provided by these packages, mumps still works correctly. From the other hand, when I try to link "int64" libscotch.dll to 64-bit integer mumps, it return error -52 that means it thinks 32-bit scotch was provided. How should I check which functions does libscotch.dll export and are they 32-bit-int or 64-bit-int? 

### Expected behavior

Expected behaviour: I guess mumps-int32 should not work fine with scotch-int64.

### Actual behavior

Actual behaviour: mumps-int32 works fine with libscotch.dll obtained from scotch-in64 and ucrt-scotch-int64 packages.

### Verification

### Windows Version

Windows 10 Pro 22H2

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [ ] CLANG64
- [ ] CLANGARM64

_No response_

mmuetzel:
Also, take better care to not mix build artifacts for 32-bit and 64-bit indexing. 
Fixes #22540.

~~Additionally, add missing dependency to libwinpthread.~~ (Edit: That might not have actually been missing. But it might be a new feature in version 7.0.5. I haven't checked that though.)

### DIFF
--- a/mingw-w64-scotch/PKGBUILD
+++ b/mingw-w64-scotch/PKGBUILD
@@ -5,15 +5,16 @@ pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-int64")
 pkgdesc='Graph partitioning and sparse matrix ordering package (mingw-w64)'
-pkgver=7.0.4
-pkgrel=3
+pkgver=7.0.5
+pkgrel=1
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://www.labri.fr/perso/pelegrin/scotch/"
 msys2_repository_url="https://gitlab.inria.fr/scotch/scotch"
 license=('spdx:CECILL-C')
 depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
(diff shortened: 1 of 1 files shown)


mmuetzel:
Pushed another commit to no longer claim that the 64-bit indexing package provides libraries that are compatible with the libraries from the 32-bit indexing package. They are fundamentally incompatible.

mmuetzel:
There are some obvious packaging errors in the `-int64` package that is currently distributed.
Even if there were still issues in that part after this PR (which is still unclear), this is a step in the right direction. Remaining issues could also be fixed in follow-ups.

Merging this as it currently is might also facilitate tests for #22540.

PR #22562, opened 2024-11-17
python-ics: update to 917.18

ognevny:
- add dunamai package for a build


ognevny:
seems that this package is unbuildable because it requires to build in git repo, but it still fails... maybe just drop this package? no one distro packages it, no users here

lazka:
No objection to dropping it.

PR #22575, opened 2024-11-18
fastfetch: update to 2.30.1

CarterLi:
Also remove mingw64 which is known to be buggy


Biswa96:
> Also remove mingw64 which is known to be buggy

What bug did it cause? Though, it is recommended to use ucrt64/clang64 in Windows 10/11.

CarterLi:
`strftime("%F %T")` doesn't work

PR #22576, opened 2024-11-18
mesa: Restore optimization lost in 21ac180

pal1000:
21ac180 switched to default makepkg-mingw CFLAGS and CXXFLAGS, losing CPU march optimization.
The optimization could have been restored in next commit, keeping the other compiler flags, but it wasn't..

### DIFF
--- a/mingw-w64-mesa/PKGBUILD
+++ b/mingw-w64-mesa/PKGBUILD
@@ -4,7 +4,7 @@ _realname=mesa
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=24.2.7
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
The optimization is already part of CFLAGS/CXXFLAGS. take a look at /etc/makepkg-mingw.conf file.
About aarch64, I think there were a discussion on which march should b the default, but not a final decision.

pal1000:
I looked and there is a performance drop, mostly affecting llvmpipe and lavapipe. That change downgraded march from core2 to nocona losing SSSE3.

pal1000:
I'll keep `-march=core2` on mesa-dist-win downstream and merge it with MSYS2 MinnGW-W64 defaults.

PR #22585, opened 2024-11-19
jsoncpp: correct name of static library

### LINKED ISSUE #22584: jsoncpp pkgconfig file reports wrong libs for static builds since 1.9.6 update

When building an application which uses jsoncpp statically and linking using the output of `pkgconfig --static --libs jsoncpp`, linking will fail. This started happening since jsoncpp was updated to version 1.9.6. I am running into this issue with my game OceanPop, available [here](https://github.com/sharkwouter/oceanpop). Below is a smaller example that should reproduce it.

main.cpp:
```
#include <string>
#include <json/json.h>
#include <iostream>

int main(void)
{
  Json::Value data;
  Json::Reader reader;
  std::string text = "{\"message\":\"hello json\"}";
  
  reader.parse(text, data);
  std::cout << data.get("message", "Could not load json message") << "\n";
  
  return 0;
}
```
CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.10)

project(hello_json)

set(CMAKE_CXX_FLAGS "-static ${CMAKE_CXX_FLAGS}")

include(FindPkgConfig)
pkg_search_module(JSONCPP Required jsoncpp)

include_directories(${JSONCPP_INCLUDE_DIRS})
add_executable(${PROJECT_NAME} main.cpp)
target_link_libraries(${PROJECT_NAME} ${JSONCPP_STATIC_LIBRARIES})
```

Put these in a directory, then run the following commands there:
```
mkdir build && cd build
cmake -G "MinGW Makefiles" ..
cmake --build .
```

### Expected behavior

As with jsoncpp version 1.9.5, I was expecting the build to link correctly and result in a working binary.

### Actual behavior

As seen in the failure of [this GitHub workflow](https://github.com/sharkwouter/oceanpop/actions/runs/11912612163/job/33196581768), building with jsoncpp version 1.9.6 results in the following error:
```
 [100%] Linking CXX executable oceanpop.exe
D:/a/_temp/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -ljsoncpp: No such file or directory
D:/a/_temp/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -ljsoncpp: No such file or directory
mingw32-make[2]: *** [CMakeFiles\oceanpop.dir\build.make:517: oceanpop.exe] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:86: CMakeFiles/oceanpop.dir/all] Error 2
mingw32-make: *** [Makefile:135: all] Error 2
```
This same code build just fine before the update.

### Verification

### Windows Version

MINGW64_NT-10.0-20348

I have contributed before, but I don't know if I'll have time to fix this myself.

mmuetzel:
Fixes #22584.

### DIFF
--- a/mingw-w64-jsoncpp/0001-static-lib-name.patch
+++ b/mingw-w64-jsoncpp/0001-static-lib-name.patch
@@ -1,0 +1,14 @@
+Append suffix to static library name only for MSVC.
+
+diff -urN jsoncpp-1.9.6/src/lib_json/CMakeLists.txt.orig jsoncpp-1.9.6/src/lib_json/CMakeLists.txt
+--- jsoncpp-1.9.6/src/lib_json/CMakeLists.txt.orig	2024-09-10 05:18:29.000000000 +0200
++++ jsoncpp-1.9.6/src/lib_json/CMakeLists.txt	2024-11-19 14:07:18.639721600 +0100
+@@ -143,7 +143,7 @@
+ 
+     # avoid name clashes on windows as the shared import lib is also named jsoncpp.lib
+     if(NOT DEFINED STATIC_SUFFIX AND BUILD_SHARED_LIBS)
+-        if (WIN32)
++        if (MSVC)
+             set(STATIC_SUFFIX "_static")
+         else()
+             set(STATIC_SUFFIX "")
--- a/mingw-w64-jsoncpp/PKGBUILD
+++ b/mingw-w64-jsoncpp/PKGBUILD
@@ -4,7 +4,7 @@ _realname=jsoncpp
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.9.6
-pkgrel=2
+pkgrel=3
 pkgdesc="A C++ library for interacting with JSON (mingw-w64)"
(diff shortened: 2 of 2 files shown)


sharkwouter (approved):
This is the fix I needed, thanks!

sharkwouter:
It seems this change was made in this PR to enable Clang to work: https://github.com/open-source-parsers/jsoncpp/pull/1480

Maybe upstream has an idea on how to make this condition only activate for MSVC and Clang instead of every WIN32 build.

mmuetzel:
A better condition for upstream would probably be `MSVC OR ("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")`. But MSYS2 always targets MinGW (not MSVC). So, the second part of that condition would never be true when packaging for MSYS2.

mmuetzel:
Thank you for testing and for digging up that PR.

I created a PR for upstream with a change that should work with any compiler when targeting MSVC: https://github.com/open-source-parsers/jsoncpp/pull/1579

PR #22588, opened 2024-11-19
libheif: update to 1.19.5

kmilos:


lazka:
For the curious:

* `WITH_UNCOMPRESSED_CODEC`: enable support for uncompressed images according to ISO/IEC 23001-17:2024. This is *experimental*
   and not available as a dynamic plugin. When enabled, it adds a dependency to `zlib`, and optionally will use `brotli`.

lazka:
Just wondering, is this a feature you want for something specific? Since there have been CVEs against it... on the other hand upstream is active and we are too, so it's probably not adding much.

lazka:
> Not really, I just saw [upstream recommends it in the "release" preset](https://github.com/strukturag/libheif/blob/11da9e48ca5ec1329106c82d95429730dc3e56ad/CMakePresets.json#L68), while confusingly it is indeed OFF by default in CMakeLists.txt...

Oh, indeed, weird.

PR #22589, opened 2024-11-19
Add libp11

mlt:


lazka:
OK with my changes?

(we can always improve things later if needed)

mlt:
Good catches! I should probably clean up the readme as well (at least the last paragraph). I was hesitating if it belongs there, but I think it might be helpful on how to configure things.

I took quite a detour trying to figuring out how to make things work on Windows with PATHs & escaping in configs, MSVC vs gcc, token names with vendor module vs emulated, etc. Then I realized that everything normally is going through p11-kit on all platforms and should be on Windows as well to unify configuration and minimize that for end apps. I even have *opensc* package for msys2 but I believe it is deadborn for 2 reasons: 1) the official installer and pkcs11 module work fine as is (no need to bring extra DLLs on PATH like with ykcs11), 2) we can't build minidriver anyway as it requires *Cryptographic Provider Development Kit*. So unless you suggest to submit a PR, I'd bury it.

lazka:
I don't really know anything about these projects, so I can't really say.

lazka:
I'm just gonna merge this get things moving...

PR #22597, opened 2024-11-19
gedit: update list of dependencies

swilmet:
gedit: update list of dependencies.


Biswa96:
Could you please bump the pkgrel value in both PKGBUILD files and add that in the respective commits?

PR #22620, opened 2024-11-21
fltk: update to 1.4

felixaszx:
Update FLTK to the latest 1.4 version.


mmuetzel:
The package grokker detected API changes that require a rebuild of the following packages:
```
mingw-w64-mathgl
mingw-w64-tigervnc
mingw-w64-gmsh
```

I wonder why Octave isn't in that list.
Does the package grokker include optdepends?

mmuetzel:
> It's probably a good idea to rebuild everything

Sounds reasonable.
See: https://www.fltk.org/articles.php?L1955
> FLTK 1.4 is intended to be mostly API compatible with FLTK 1.3.x so you don't need to change source code when you upgrade to FLTK 1.4. However, all programs must be recompiled with FLTK 1.4 because the ABI (Application Binary Interface) has changed.

felixaszx:
Looks like we need tp patch mathgl.
`C:/_/B/src/mathgl-8.0.1/mgllab/dialogs.cpp:422:36: error: call of overloaded 'value(NULL)' is ambiguous`
`  422 |                 {       p[i]->value(NULL);      s[i]->value(5); c[i]->value(0); }`

felixaszx:
I am surprise that those 64-bit tests do take forever.

lazka:
> I am surprised that those 64-bit builds do take forever.
> They are twice as long as the mingw32 one.

some of the rebuilds are 64bit only packages

mingw-w64-fltk/0001-cmake-fix-install.patch
```diff
@@ -1,38 +0,0 @@
-$ diff -Naur ./CMake/install.cmake.orig ./CMake/install.cmake
```
lazka, line 1: Was this patch removed on purpose?

Seems you replaced with with a mv; and the man pages are gone too, thought he related tools are also missing, so not a big issue.

felixaszx: Yes, I replace the patch with `mv`.
I actually did not tounch the man pages, it seems to be that FLTK dev team removed `sudoku`, 'checkers' and `block` pages themself.
I can only guess that it was a mistake they made before since apparently those apps are never shipped with FLTK sources itself.

mmuetzel:
It looks like, since this update, Octave no longer closes (after plotting) after the FLTK libraries have been loaded. See also:
https://octave.discourse.group/t/ci-stuck-after-make-check-on-windows/6097/5

Does anyone know if something changed in 1.4.0 which could make a difference when it comes to unloading the FLTK libraries?

mmuetzel:
If I add `-DFLTK_GRAPHICS_GDIPLUS=OFF` to the cmake flags, Octave is no longer stuck on exit after the FLTK libraries have been loaded.

Would it be ok to build FLTK with this flag?

felixaszx:
> If I add `-DFLTK_GRAPHICS_GDIPLUS=OFF` to the cmake flags, Octave is no longer stuck on exit after the FLTK libraries have been loaded.
> 
> Would it be ok to build FLTK with this flag?

I have not seen any significant negative effects of building FLTK without GDI+. I think it will be totally fine. 
Except it might cause performance issues when drwaing menus. (https://github.com/fltk/fltk/discussions/1089#discussioncomment-10927428)

mmuetzel:
> I have not seen any significant negative effects of building FLTK without GDI+. I think it will be totally fine.

I opened #23051 to disable GDI+ in the FLTK package.

PR #22622, opened 2024-11-22
godot-cpp: Update to 4.3

lazka:


felixaszx:
I think this is still in beta according to Godot themselves?

lazka:
Oh, I thought it needs to be kept in sync. Never mind then

PR #22626, opened 2024-11-23
Add vk-bootstrap

felixaszx:
Add this new and popular package that makes vulkan development way easier than it used to be.

This package should be synced with vulkan-headers in version number.

### DIFF
--- a/mingw-w64-vk-bootstrap/PKGBUILD
+++ b/mingw-w64-vk-bootstrap/PKGBUILD
@@ -1,0 +1,46 @@
+# Maintainer: Sergey Popov <popov.sa@gmail.com>
+# Contributor: Mario Emmenlauer <memmenlauer@biodataanalysis.de>
+
+_realname=vk-bootstrap
+pkgbase="mingw-w64-${_realname}"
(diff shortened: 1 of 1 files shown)


mingw-w64-vk-bootstrap/PKGBUILD
```diff
@@ -0,0 +1,45 @@
...
+          -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+          -DBUILD_SHARED_LIBS=ON \
+          -DVK_BOOTSTRAP_TEST=OFF \
+          -S "${_realname}-${pkgver}" \
+          -B "build-${MSYSTEM}"
+
+  cmake --build "build-${MSYSTEM}"
+
+  # remove the src archive
+  rm "../v${pkgver}"
```
lazka, line 37: What's this about?

lazka:
otherwise lgtm

mingw-w64-vk-bootstrap/PKGBUILD
felixaszx: It was a workaround to prevent git to upload the source archive. I have now changed it to a proper solution.

PR #22635, opened 2024-11-23
meson: backport a fix for `autolib` sections in Cargo subprojects

ognevny:
at one moment cargo subprojects start failing to extract because of new `autolib` section in Cargo.toml. add an upstream fix for this section support


ognevny:
I tested local build and now my subprojects builds again

PR #22641, opened 2024-11-24
gimp: rebuild without python2 for ucrt64 and clang64

ognevny:
*as discussed below*:

mingw32 and mingw64 is used by gimp devs to make release. for remaining envs there is no sense to keep EOLed version of python, users can switch to gimp3, which is about to release soon

### DIFF
--- a/mingw-w64-gimp/PKGBUILD
+++ b/mingw-w64-gimp/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gimp
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=2.10.38
-pkgrel=3
+pkgrel=4
 pkgdesc="GNU Image Manipulation Program (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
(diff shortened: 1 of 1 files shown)


lazka:
This is still needed by gimp upstream until 3.0 is released as far as I know

ognevny:
python plugins are optional for gimp. CI shows that it's fine to build gimp without python. as we package gimp3 here, it won't be a problem to ask users to switch to gimp3 if they need python plugins

lazka:
Yes, but gimp is using MSYS2 to create their installer.

ognevny:
I thought that gimp2 won't be updated anymore...

lazka:
That would be weird, given it is the only supported stable branch atm.

lazka:
We could drop everything besides mingw32/mingw64 though

ognevny:
dropping 32 bit here will be good, it doesn't seem that someone really needs such version of gimp

lazka:
Upstream gimp2 uses mingw32 for their installer.

ognevny:
ok, leaving it draft until gimp3 is released

ognevny:
> So, we could remove it from clang64 and ucrt64.

done

ognevny:
branch is now changed to build gimp without python2 for non-MINGW envs. also pygobject2 was forgotten to drop

PR #22647, opened 2024-11-25
Add fix for gnuradio issue #7468, gr_modtool fails for python oot

chrisjohgorman:
- this fix allows for building python oot modules with gnuradio 3.11.11.0
- a new patch 0002-issue-7468-python-oot.patch is added.
- without this patch the python `__init__.py` file for the oot block does not get created properly and the oot module will fail to load.
- update pkgrel


lazka:
I guess this is the same as https://github.com/gnuradio/gnuradio/commit/16ef847807553d65b80666b238643794b0b93d71 ? Can't we just backport that?

chrisjohgorman:
It is indeed the patch I applied to my pull request.  Is there an easier
way to apply it rather than the one I've proposed in PR #22647?  If you
know of a better method, I'm all for it.  Let me know if you want me to do
this another way.

PR #22649, opened 2024-11-25
Add qucs-s

Kreijstal:
This an eletrical schematic and simulator using ngspice and quics own simulator.

Xyce can be used but Xyce hasn't been packaged yet.

What is also great is https://github.com/ra3xdh/qucs_s has msys2 on their CI

### DIFF
--- a/mingw-w64-qucs-s/PKGBUILD
+++ b/mingw-w64-qucs-s/PKGBUILD
@@ -1,0 +1,59 @@
+# Maintainer: Kreijstal <kreijstal@hotmail.com>
+
+_realname=qucs-s
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=24.4.1
+pkgrel=1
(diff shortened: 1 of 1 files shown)


mingw-w64-qucs-s/PKGBUILD
```diff
@@ -0,0 +1,58 @@
...
+  "${MINGW_PACKAGE_PREFIX}-gcc-libs"
+  "${MINGW_PACKAGE_PREFIX}-ngspice"
+)
+makedepends=(
+  "${MINGW_PACKAGE_PREFIX}-cc"
+  "${MINGW_PACKAGE_PREFIX}-cmake"
+  "${MINGW_PACKAGE_PREFIX}-ninja"
+  "${MINGW_PACKAGE_PREFIX}-qt6-tools"
+  "${MINGW_PACKAGE_PREFIX}-qt6-svg"
+  "${MINGW_PACKAGE_PREFIX}-qt6-charts"
```
lazka, line 25: Should these two be runtime deps instead?

PR #22651, opened 2024-11-26
arrow python-pyarrow: Update to 18.1.0

amoeba:


amoeba:
Thanks for the quick merge @Biswa96.

Biswa96:
Please reset the pkgrel value to 1 next time (which I forgot to mention).

amoeba:
Noted, thanks. We have it all scripted but it looks like that was missed.

PR #22658, opened 2024-11-26
sqlite3: Update to 3.47.1

lazka:
0002-sqlite-sqldiff-makefile-add-console-io.patch: simple refresh


lazka:
needs more work I guess

PR #22684, opened 2024-11-28
putty: Update to 0.82

lazka:


jeremyd2019:
Failed on arm64 due to not finding `<consoleapi.h>`.  This seems to be in the Windows SDK, but not in mingw-w64.  I'm guessing it only failed on arm64 because default `_WIN32_WINNT` is `0x0a00` there so it enabled conpty support.  @Biswa96 ?

lazka:
See 3e35b7b8cdd106a and the linked issue

Biswa96:
I have sent the patch upstream to add those header files https://sourceforge.net/p/mingw-w64/mailman/message/58848109/ . Please test it.

PR #22685, opened 2024-11-28
rust: update psm dependency

ognevny:
I don't know whether this work, patch doesn't create checksum file... I'll go sleep now, feel free to fix if needed
@jeremyd2019


jeremyd2019:
we shall see: https://github.com/msys2-arm/MINGW-packages/actions/runs/12074687359/job/33673128814 though I bet if the patch is bad it'd fail in one of the other CI runs too.

jeremyd2019:
I wasn't sure if it'd be easier to update psm like this, or just apply the patch from https://github.com/rust-lang/stacker/commit/c94a424e1dc1c1d7f0e7cdb5150d1c0fe4e0489b, either would require messing with the checksum files, which suck.

jeremyd2019:
https://github.com/msys2-arm/MINGW-packages/actions/runs/12075324149/job/33674915084 🤞

lazka:
merge?

ognevny:
good morning, thank you :)

PR #22695, opened 2024-11-29
assimp: update to 5.4.3

ognevny:


lazka:
lgtm

ognevny:
fails on clangarm64:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
  Please set them or make sure they are set and tested correctly in the CMake files:
  DirectX_D3DX9_LIBRARY (ADVANCED)
      linked by target "assimp_viewer" in directory C:/M/B/src/assimp-5.4.3/tools/assimp_view
```
I won't be home soon, anyone please fix 🙏

Biswa96:
`-DDirectX_D3DX9_LIBRARY=${MINGW_PREFIX}/lib/libd3d9.a` may work but I have not tried that.

mingw-w64-assimp/PKGBUILD
```diff
@@ -12,42 +12,40 @@ url="https://www.assimp.org/"
...
-  ${MINGW_PREFIX}/bin/cmake.exe \
+  cmake \
     -G"Ninja" \
     -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DASSIMP_WARNINGS_AS_ERRORS=OFF \
     -DASSIMP_BUILD_ASSIMP_TOOLS=ON \
-    -DASSIMP_BUILD_ASSIMP_VIEW=$([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] && \
-                                  echo "OFF" || echo "ON") \
+    -DASSIMP_BUILD_ASSIMP_VIEW=ON \
```
MehdiChinoune, line 35: This the reason why clangarm64 fails, assimp_view requires d3dx9 library which is not available for ARM64 on Windows.

ognevny: so the condition should be for clangarm64 only

MehdiChinoune: Yes, I already pushed a commit.

PR #22706, opened 2024-12-01
mosquitto: Update to 2.0.20

lazka:
It now requires threading and fails to build if disabled (likely not tested upstream) so try using winpthreads. Hopefully good enough.


lazka:
@podsvirov fyi

PR #22707, opened 2024-12-01
cvise: colordiff is an MSYS2 package

oltolm:


lazka:
What makes you think that?

oltolm:
```
$ pacboy find colordiff
msys/colordiff 1.0.21-1 [installed]
    Diff wrapper with pretty syntax highlighting
```

BTW I just discovered that `cvise` does not work because it contains wrong paths like `D:/a/msys64/clang64/bin/clang-format.exe`. If I replace `D:/a` with `C:` it works.

lazka:
Ah, right, I mixed it up, thanks. Needs a pkgrel bump

oltolm:
I bumped the `pkgrel`. The paths are now correct for some reason.

PR #22758, opened 2024-12-07
solvespace: New Package

Kreijstal:
the 1487.patch refers to the PR, and because this patch is on release it will not exactly work for the tagged version, so the 002*.patch is a hotfix which should be removed whenever they tag a new version.

### DIFF
--- a/mingw-w64-solvespace/PKGBUILD
+++ b/mingw-w64-solvespace/PKGBUILD
@@ -1,0 +1,74 @@
+# Contributor: Kreijstal <kreijstal@hotmail.com>
+
+_realname=solvespace
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=3.2
+pkgrel=1
+pkgdesc='Parametric 2d/3d CAD (mingw-w64)'
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://solvespace.com'
+msys2_repository_url='https://github.com/solvespace/solvespace'
+msys2_changelog_url='https://github.com/solvespace/solvespace/blob/master/CHANGELOG.md'
+msys2_references=(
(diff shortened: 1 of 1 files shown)


mingw-w64-solvespace/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+  "1487.patch"
+  "002-not-in-msys2-temp.patch")
+sha256sums=('34a273ce642d0c77b8f101463730ed4eade7d90cfabfe486f3e7cbf494ff132a'
+            '76b7711320430440e646d3a93c4128ae86db797c4089cc08bbfb813aac70527e'
+            '519cda45a5e5c76d1c5a3904b9ff35fcb058a8f12a72ec06d703576290092a54')
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}"
+  patch -N -p1 -i "${srcdir}/1487.patch" || true
+  patch -N -p1 -i "${srcdir}/002-not-in-msys2-temp.patch" || true
```
lazka, line 35: why the "|| true" ?

lazka:
the resulting package is empty, so something is broken in the package() phase

mingw-w64-solvespace/PKGBUILD
Kreijstal: the first patch is from PR but it's from upstream which is not the version we are packaging, the rejects are then fixed/backported in 002-not-in-msys2-temp.patch, however the second one a || true is not necessary, you're right.

lazka: I'd argue it doesn't make sense in both cases. It shouldn't fail, and if it does we want to know.

Kreijstal:
> the resulting package is empty, so something is broken in the package() phase

It seems it's an upstream problem..

mingw-w64-solvespace/PKGBUILD
Kreijstal: I guess I can manually remove the reject chunks

PR #22768, opened 2024-12-08
qt6: update to 6.8.1

sunderme:
adapt patches
quick3d: new patches to fix 3rd party compile fail (https://bugreports.qt.io/browse/QTBUG-131962)


sunderme:
clang64 runs through, the rest run into time out
Is there another way to test an qt update ? And unfortunately it is necessary to test. This time qt6-pdf failed at first.

lazka:
The only way right now is to split it into two PRs

But since this is a minor release and clang runs through I'd suggest to maybe just merge it and see?

sunderme:
I am confident that this will build as clang runs through.
The question is rather how to do my initial check more efficiently.

lazka:
I've added a link to the upstream issue in 7eec0796c7c6afdbf

mingw-w64-qt6-base/PKGBUILD
```diff
@@ -4,7 +4,7 @@ _realname=qt6-base
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-${_realname}-debug")
-_qtver=6.8.0
+_qtver=6.8.1
 pkgver=${_qtver/-/}
 pkgrel=2
```
lazka, line 9: it's just a convention, but pkgrel should be reset on version updates. It doesn't make much of a difference though.

PR #22772, opened 2024-12-08
libssh: remove outdated workaround

mati865:


lazka:
syntax error, and the patch is still used and there

mati865:
Made the edit in a hurry and trusted that green mark on the CI too much.

PR #22790, opened 2024-12-12
vtk: update to 9.4.

MehdiChinoune:
Fails to build with libc++-19, reported upstream but as usual VTK developers don't fix issues about compatibility with versions of packages other than the ones they build with.
If someone has enough experience with c++, please consider fixing it.
The issue is blocking updating protobuf to 29.1 and gdal to 3.10.0

If no fix for the issue, I will drop some packages (freecad, avogadolibs/avogadro, paraview and itk) and rebuild the others without vtk (OpenCV and OpenCASCADE).


lazka:
Is there an upstream issue?

PR #22804, opened 2024-12-13
rust: don't build out of tree

ognevny:
this is not supported upstream, but a workaround was used. set build-dir in config.toml, make dest dirs names contain $MSYSTEM to match previous behavior

asking @filnet as a workaround's author is it ok to do so

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -17,7 +17,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          $([[ ${CARCH} == i686 ]] || echo "${MINGW_PACKAGE_PREFIX}-rust-wasm")
          "${MINGW_PACKAGE_PREFIX}-rust-src")
 pkgver=1.83.0
-pkgrel=3
+pkgrel=4
 pkgdesc="Systems programming language focused on safety, speed and concurrency (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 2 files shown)


filnet:
> asking @filnet as a workaround's author is it ok to do so

I am not sure that I understand your question.
But note that I have currently lost interested in Rust (might come back some day though).

ognevny:
I meant will be fine to drop workaround for building outside of rustc src and build inside of it?

filnet:
> I meant will be fine to drop workaround for building outside of rustc src and build inside of it?

I think that workaround was already in place when I got involved in Rust maintenance.

ognevny:
oh, I didn't look carefully at blame...

jeremyd2019:
I was going to mention, we were running right up against the `MAX_PATH` limits for this package, and moving the build directory deeper would probably make it too deep.  Sure enough:
```
BUG: compat/win32/fscache.c:128: Trying to allocate fsentry for long path 'B/src/rustc-1.83.0-src/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rustlib/rustc-src/rust/build-CLANGARM64/aarch64-pc-windows-gnullvm/stage0-sysroot/lib/rus
```

this seems even weirder than that though, some sort of circular link that git is unhappy with.

jeremyd2019:
I opened git-for-windows/git#5320 for the git BUG output, git clean seems to be following directory junctions rather than simply removing them

PR #22807, opened 2024-12-14
octave: update to 9.3.0

mmuetzel:
`libtool` and `autoconf` have been patched to better support LLVM Flang in the meantime. So, remove the workarounds that were needed earlier.


mmuetzel:
I'm pretty sure it was possible to remove a work-around that was used to correctly display help texts at some point. But it looks like that work-around is needed again. (I'm not sure why though.)

mmuetzel:
I downloaded and installed the MINGW64 build artifacts from this PR. With them, Octave's test suite is passing.

PR #22828, opened 2024-12-19
xpra 6.2.2

totaam:
also use env vars to pass arguments to setup.py because the '-C' switches don't seem to work properly


Biswa96:
This change removed the entire nvidia directory. Is that expected?

totaam:
> This change removed the entire nvidia directory `/lib/python3.12/site-packages/xpra/codecs/nvidia/`. Is that expected?

Yes, I have just double checked it and none of these codecs (`nvdec`, `nvenc`, `nvfbc` and `nvjpeg`) can ever work without having the CUDA SDK installed at build time.
The `nvidia/util` and `nvidia/cuda` modules are also useless on its own.

PR #22838, opened 2024-12-19
tfel (mfront): update to 5.0.0

3rav:


MehdiChinoune:
please update sha256sums, `SKIP` is not allowed for patches.

PR #22851, opened 2024-12-20
python-llvmlite: update to 0.44.0rc2; python-numba: update to 0.61.0rc2

ognevny:
update numba due to ImportError raised because of numpy version. this requires llvmlite update


ognevny:
https://github.com/numba/numba/issues/9838

jeanpaul98728:
[https://www.mylittlepolynesia.com/](url) thanks

ognevny:
the only way now is revert commit that pins numpy version to 2.1 (like Arch did)

jeanpaul98728:
thanks

PR #22859, opened 2024-12-21
starpu: disable opencl

3rav:
This option currently does nothing and prevents the compilation of dependent packages

### DIFF
--- a/mingw-w64-starpu/PKGBUILD
+++ b/mingw-w64-starpu/PKGBUILD
@@ -4,7 +4,7 @@ _realname=starpu
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=1.4.7
-pkgrel=1
+pkgrel=2
 pkgdesc='StarPU is a task programming library for hybrid architectures (mingw-w64)'
 arch=('any')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
Could you show what errors/issues do you face with dependent packages?

3rav:
I can't locally compile CalculiX in a fully static version via opencl, and from what I see starpu is only used in PaStiX

MehdiChinoune:
> I can't locally compile CalculiX in a fully static version via opencl, and from what I see starpu is only used in PaStiX

You could build your own starpu, but since It doesn't cause any issue with MSYS2 packagign I don't see it as a reason to drop OpenCL.

3rav:
Why when I add the compiler to use the OpenCL library (-lOpenCL) I get an error that there is no such library (and I have it installed), and at the same time pkg-config?:
image

MehdiChinoune:
Because OpenCL doesn't have a static library.

3rav:
But shouldn't the file "libOpenCL.dll.a" simulate its/her existence?
image

Kreijstal:
> I can't locally compile CalculiX in a fully static version via opencl, and from what I see starpu is only used in PaStiX

you have to use dynamic, otherwise we need static versions of ALL packages

PR #22879, opened 2024-12-23
`pyopengl[-accelerate]`: update to 3.1.8

totaam:
This update requires some explanations as [`pyopengl`](https://github.com/mcfletch/pyopengl) does not look actively maintained anymore:
* switched to github instead of pypi because the 3.1.8 release of `pyopengl-accelerate` release is missing from pypi: https://pypi.org/project/PyOpenGL-accelerate/
* added a "version" patch because the release on github forgot to bump the version: https://github.com/mcfletch/pyopengl/issues/123
* the "array" patch fixes compilation with newer toolchains, as per https://github.com/mcfletch/pyopengl/pull/112
* the checks are skipped because they fail on my test VM (even after installing `pygame`) despite `pyopengl` working just fine (perhaps they expect a real GPU to run?) - I have left them in, just commented out (I am not sure what the best practice is - I can just remove it all)

FWIW: I have been using these patches in xpra binary releases for some time.


lazka:
thanks for all the info. lgtm

PR #22892, opened 2024-12-24
filezilla: update to 3.68.1

### LINKED ISSUE #21889: [filezilla] source link does not download tarball instead redirects to html file
### Package name

filezilla

### New version number

_No response_

### Other information that may be useful

ERROR: One or more files did not pass the validity check!

I tried to build the package and got this error and can not get it to build after ignoring the error.

Tim S. 

No

Biswa96:
* Fixes #21889


Biswa96:
I have tested this latest Filezilla from UCRT64 CI artifact. See following screenshot.

image

PR #22900, opened 2024-12-25
[new package] libexcelize 2.9.0

taozuhong:


lazka:
Why is the library source part of this PR?

taozuhong:
The source extracted from project: https://github.com/xuri/excelize-py

taozuhong:
Replace with #22923

PR #22907, opened 2024-12-26
mingw-w64: Update to 03d8a40f57

lazka:
https://github.com/mingw-w64/mingw-w64/compare/a2d19218d...03d8a40f5

drop the revert since that should be fixed via https://github.com/mingw-w64/mingw-w64/commit/c9eca28decb3aa5213fe22ab1f9883df91d79806

### DIFF
--- a/mingw-w64-crt-git/PKGBUILD
+++ b/mingw-w64-crt-git/PKGBUILD
@@ -3,9 +3,9 @@
 _realname=crt
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
-pkgver=12.0.0.r446.ga2d19218d
+pkgver=12.0.0.r458.g03d8a40f5
 pkgrel=1
-_commit='a2d19218ddeff0ac61e8a38aada393c06c263b99'
+_commit='03d8a40f57649fbb773f1cdbe3a760f5e0943e76'
 pkgdesc='MinGW-w64 CRT for Windows (mingw-w64)'
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
@@ -22,7 +22,7 @@ provides=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 conflicts=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 options=('!strip' '!buildflags' '!emptydirs')
 source=("mingw-w64"::"git+https://git.code.sf.net/p/mingw-w64/mingw-w64#commit=$_commit")
-sha256sums=('71f6d0e2522447456e77d7a631d8862ccf52024ffcc4c251818da4a0057e82fa')
+sha256sums=('cda9858cf758f752306330f66840631de953f11d0e74836ea3b7892a2743db6f')
 
 pkgver() {
   cd "${srcdir}/mingw-w64"
--- a/mingw-w64-headers-git/PKGBUILD
+++ b/mingw-w64-headers-git/PKGBUILD
@@ -4,9 +4,9 @@ _realname=headers
 pkgbase=mingw-w64-${_realname}-git
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
 pkgdesc="MinGW-w64 headers for Windows (mingw-w64)"
-pkgver=12.0.0.r446.ga2d19218d
-pkgrel=2
-_commit='a2d19218ddeff0ac61e8a38aada393c06c263b99'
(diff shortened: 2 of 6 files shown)


rouault:
I strongly believe this update has caused a regression in the GDAL regression test suite.
Last good build in https://github.com/OSGeo/gdal/actions/runs/12505586623/job/34889173672 used git-12.0.0.r446.ga2d19218d-1 and first failed in https://github.com/OSGeo/gdal/actions/runs/12515627580/job/34913475983 uses -git-12.0.0.r458.g03d8a40f5-1
One failure is about long filenames starting with ``\\?\`` (not sure which calls to the Windows API  has regressed there)
The other failure is more straightforward. It is about a failure renaming "tmp/yy\u4E2D\u6587.\u4E2D\u6587" on top of itself using _wrename()

Biswa96:
Could you provide any minimal sample code to reproduce the issue?

lazka:
just to confirm that looking at the build logs the only package versions differences are due to this mingw-w64 change.

rouault:
I figured the root cause. With git-12.0.0.r446.ga2d19218d-1, the below code prints 0x700, whereas with >= git-12.0.0.r458.g03d8a40f5-1, it prints 0x600.   The GDAL code base has code paths that require ``__MSVCRT_VERSION__ >= 0x601`` to use some functions of the Win32 API

test.c:
```
#include <windows.h>
#include <stdio.h>

int main()
{
    printf("0x%X\n", __MSVCRT_VERSION__);
    return 0;
}
```

old working mingw64 install:
```
evenr@DESKTOP-J297O1T MINGW64 /c/dev
$ pacman -Q | grep git
mingw-w64-x86_64-crt-git 12.0.0.r369.g0d4221712-2
mingw-w64-x86_64-headers-git 12.0.0.r369.g0d4221712-2
mingw-w64-x86_64-libmangle-git 12.0.0.r369.g0d4221712-1
mingw-w64-x86_64-libwinpthread-git 12.0.0.r369.g0d4221712-1
mingw-w64-x86_64-tools-git 12.0.0.r369.g0d4221712-1
mingw-w64-x86_64-winpthreads-git 12.0.0.r369.g0d4221712-1
mingw-w64-x86_64-winstorecompat-git 12.0.0.r369.g0d4221712-1

evenr@DESKTOP-J297O1T MINGW64 /c/dev
$ gcc test.c -o test.exe

evenr@DESKTOP-J297O1T MINGW64 /c/dev
$ ./test.exe
0x700
```

latest mingw64 install:
```
evenr@DESKTOP-J297O1T MINGW64 /c/dev
$ pacman -Q | grep git
mingw-w64-x86_64-crt-git 12.0.0.r459.g63f3f2846-1
mingw-w64-x86_64-headers-git 12.0.0.r459.g63f3f2846-1
mingw-w64-x86_64-libwinpthread-git 12.0.0.r459.g63f3f2846-1
mingw-w64-x86_64-winpthreads-git 12.0.0.r459.g63f3f2846-1

evenr@DESKTOP-J297O1T MINGW64 /c/dev
$ gcc test.c -o test.exe

evenr@DESKTOP-J297O1T MINGW64 /c/dev
$ ./test.exe
0x600
```

rouault:
ok, so this seems to be related to commit https://github.com/mingw-w64/mingw-w64/commit/90d3f33f1135b7a9f349267476fe39c27cd96cdc
Probably that the build recipe here should use --with-default-msvcrt=msvcr70 to restore the previous behavior ?

lazka:
Here is the discussion behind it: https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/20241018181538.bahgwxwolvx2n6cj%40pali/#msg58829839

> Probably that the build recipe here should use --with-default-msvcrt=msvcr70 to restore the previous behavior ?

That would break the mingw-w64 internal checks then, since we are not using msvcr70. We could revert that linked commit temporarily, but removing the checks for `__MSVCRT_VERSION__` on your side, as you did in the PR, is probably the best way forward.

rouault:
> but removing the checks for `__MSVCRT_VERSION__` on your side, as you did in the PR, is probably the best way forward.

ok, I'll stick with that then since it doesn't seem to have obvious inconvenience (if one excepts the fact that current released versions of GDAL would be slightly broken when built against current mingw-w64, until our next release, but I guess that's life)

thanks for your assistance

lazka:
There is some talk upstream to maybe move the default to `0x06FF` instead, to avoid issues like this.

I've also created https://github.com/boostorg/align/pull/21 for boost to remove some old checks as well there.

ognevny:
the repo seems abandoned, latest commit was created 2 years ago :(

PR #22926, opened 2024-12-28
gimp3: update to 3.0.0RC2

### LINKED ISSUE #22629: gimp3: cmd.exe window is shown while executing

1. `pacman -S $MINGW_PACKAGE_PREFIX-gimp3`
2. launch gimp.exe by double clicking on executable

### Expected behavior

no window is shown but gimp one

### Actual behavior

image

### Verification

### Windows Version

MINGW64_NT-10.0-26100

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [X] CLANG64
- [ ] CLANGARM64

_No response_

ognevny:
- `0002-skip-test-env.patch`: refresh
- remove options that are set to default value
- add `omp` as dependency explicitly
- `enable-default-bin` is fixed upstream
- add libbacktrace for non-i686 environments
- explicitly disable `libunwind` as it can't be found by meson (because gimp uses libunwind not from LLVM)

### DIFF
--- a/mingw-w64-gimp3/PKGBUILD
+++ b/mingw-w64-gimp3/PKGBUILD
@@ -1,12 +1,12 @@
 # Maintainer: Alexey Pavlov <Alexpux@gmail.com>
 
 _realname=gimp3
-_pkgversuffix=RC1
+_pkgversuffix=RC2
 _pkgver=3.0.0
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=${_pkgver}${_pkgversuffix}
-pkgrel=3
+pkgrel=1
 pkgdesc="GNU Image Manipulation Program (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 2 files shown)


ognevny:
@Biswa96 I made some more changes, are they ok to land?

Biswa96:
No idea.

Biswa96:
I meant that I am not sure if changing the meson options would break anything.

ognevny:
the only breaking change is "new" set of executables, but it's not bad as there is still `gimp.exe` available :) removed options are mostly set by default

hmartinez82:
`win32-debug-console` should stay enabled until GA, no?

mingw-w64-gimp3/PKGBUILD
```diff
@@ -89,58 +91,49 @@ apply_patch_with_msg() {
...
   MSYS2_ARG_CONV_EXCL="--prefix=" \
-  ${MINGW_PREFIX}/bin/meson.exe setup \
+  meson setup \
     --buildtype=plain \
     --auto-features=enabled \
     --prefix=${MINGW_PREFIX} \
     -Daa=disabled \
     -Dalsa=disabled \
     -Dappdata-test=disabled \
+    -Dbug-report-url="https://github.com/msys2/MINGW-packages/issues" \
```
mmuetzel, line 111: Do we really want for users to report issues with GIMP here instead of with the project upstream?

ognevny: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/meson.build#L1324

ognevny:
> `win32-debug-console` should stay enabled until GA, no?

ok, I'll left it enabled, it's set to `auto` by default

brunvonlope:
> add libbacktrace for non-i686 environments

Also not needed. Linux-specific code

ognevny:
thank you for clarification. will fix when will get home

PR #22927, opened 2024-12-28
llama.cpp: update to b4394

peter277:
* Introduce epoch field to indicate that new package version numbering scheme beginning with letter b is newer than old scheme beginning with letter r
* Amend output file renaming logic since llama.cpp outputs have changed significantly since last package update: only target files not already prefixed with word 'llama' (skipping DLLs), change prefix to 'llama' instead of 'llama.cpp' (consistent with new default convention), remove logic for renaming file 'main.exe' since this no longer exists (we instead have llama-cli and llama-server)
* Optimise build configuration: include more backends for GGML (Vulkan and OpenCL), add preferred openblas64 library as dependency, add openmp as build dependency, other option tweaks similar to ArchLinux PKGBUILD script
* More informative package description

### DIFF
--- a/mingw-w64-llama.cpp/PKGBUILD
+++ b/mingw-w64-llama.cpp/PKGBUILD
@@ -1,26 +1,35 @@
 _realname=llama.cpp
-_realver=b3091
+_realver=b4394
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=b3091.r0.g2b33896
+epoch=1
+pkgver=b4394.r0.g16cdce7
 pkgrel=1
-pkgdesc="Port of Facebook's LLaMA model in C/C++ (mingw-w64)"
+pkgdesc="Library and tools for running inference with Meta's LLaMA model (and derivatives) in C/C++ (mingw-w64)"
 arch=('any')
 mingw_arch=('ucrt64' 'clang64' 'clangarm64' 'mingw64')
 url="https://github.com/ggerganov/llama.cpp"
 license=('spdx:MIT')
 depends=(
-  "${MINGW_PACKAGE_PREFIX}-openblas"
   "${MINGW_PACKAGE_PREFIX}-gcc-libs"
+  "${MINGW_PACKAGE_PREFIX}-openblas64"
+  "${MINGW_PACKAGE_PREFIX}-vulkan-loader"
+  "${MINGW_PACKAGE_PREFIX}-shaderc"
+  "${MINGW_PACKAGE_PREFIX}-opencl-icd"
+  "${MINGW_PACKAGE_PREFIX}-python"
+  "${MINGW_PACKAGE_PREFIX}-python-numpy"
 )
 makedepends=(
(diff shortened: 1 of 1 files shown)


mingw-w64-llama.cpp/PKGBUILD
```diff
@@ -1,26 +1,35 @@
...
+  "${MINGW_PACKAGE_PREFIX}-opencl-icd"
+  "${MINGW_PACKAGE_PREFIX}-python"
+  "${MINGW_PACKAGE_PREFIX}-python-numpy"
 )
 makedepends=(
   "${MINGW_PACKAGE_PREFIX}-cmake"
   "${MINGW_PACKAGE_PREFIX}-ninja"
   "${MINGW_PACKAGE_PREFIX}-cc"
-  'git'
+  "${MINGW_PACKAGE_PREFIX}-openmp"
```
ognevny, line 26:
```suggestion
  "${MINGW_PACKAGE_PREFIX}-omp"
```

mingw-w64-llama.cpp/PKGBUILD
```diff
@@ -1,26 +1,35 @@
...
-pkgdesc="Port of Facebook's LLaMA model in C/C++ (mingw-w64)"
+pkgdesc="Library and tools for running inference with Meta's LLaMA model (and derivatives) in C/C++ (mingw-w64)"
 arch=('any')
 mingw_arch=('ucrt64' 'clang64' 'clangarm64' 'mingw64')
 url="https://github.com/ggerganov/llama.cpp"
 license=('spdx:MIT')
 depends=(
-  "${MINGW_PACKAGE_PREFIX}-openblas"
   "${MINGW_PACKAGE_PREFIX}-gcc-libs"
+  "${MINGW_PACKAGE_PREFIX}-openblas64"
```
MehdiChinoune, line 15: Any reason to choose openblas64 instead of openblas (32), you know the difference, right?

peter277: Reason I replaced it with openblas64 is because based on some of the CMake output and configuration it appeared that it might be preferred by the build process, e.g. rf. https://github.com/ggerganov/llama.cpp/blob/a813badbbdf0d38705f249df7a0c99af5cdee678/ggml/src/ggml-blas/CMakeLists.txt#L28-L33  . ArchLinux script also declared both openblas and openblas64 as a dependency.

Though good idea to check the results - inspecting the DLL files (e.g. ggml.dll) using ldd does show that the build process has nonetheless linked against libopenblas.dll not libopenblas_64.dll . Guess there is further configuration in play and the references to openblas64 might be mostly redundant (or second in precedence).

peter277:
@MehdiChinoune - thanks for the fixes. I did notice that you switched off the building of the 'examples' - many of these are useful utilities in their own right, e.g. llama-cli and llama-server, and I would request/suggest these be switched back on by `setting -DLLAMA_BUILD_EXAMPLES=ON` instead. This would maximise usefulness for the user and was also included in previous builds. Otherwise the package is purely a library release without associated tools (albeit classified as 'examples').

Interested to know your thinking on this?

MehdiChinoune:
I will restore them once I get home.

PR #22935, opened 2024-12-30
Unvendor glfw from raylib

Kreijstal:
To use system version instead


ognevny:
you forgot to add an option for shared build

Kreijstal:
@lazka anything wrong?

PR #22940, opened 2024-12-31
zed: update to 0.167.1

ognevny:
updating to pre-release version as it wasn't released on 25th of December (for unknown reason for me). also backporting 2 patches, which contain fixes and improvements for Windows (they depend on each other)


ognevny:
oh, it will be released tomorrow, in 2025 :D

PR #22951, opened 2025-01-01
gettext: update to 0.23.1

MehdiChinoune:
package-grokker shows 190 packages need to rebuilt, I don't remember encountering such thing with previous versions of gettext!
The upstream refuse to support anything related to MSYS2, so I don't think we could report the issue.


MehdiChinoune:
https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=b182c17096847cb581c3e2c3f42c145c47c7c7a6 ??

Larhzu:
The `libintl_hash_string` wasn't in the public API headers and thus not part of the public API. It's unlikely that any other package relied on it, so removing it from the ABI *shouldn't* break anything. There are similar commits for [mbszero](https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=6027f38bd7c45742bbd69723d4fa5f4d6d466fcd) and [rpl_isnanf](e37cb87ec99d6ff42afc801f3e09f2355483c950). It's understandable that an automated tool may flag all dependent packages as needing rebuilding because such a tool sees an ABI change.

I was looking at MSYS2's Gettext version because Gettext 0.23.1 fixes UTF-8 code page usage on native Windows. The fix is in the Gnulib files included in libintl so it's not visible in Gettext's Git log.

XZ Utils >= 5.6.3 uses UTF-8 code page. With old Gettext, non-ASCII chars are garbled when translation support is enabled. Next XZ Utils release will allow enabling translations on native Windows only if Gettext >= 0.23.1 and using UCRT.

MehdiChinoune:
> The `libintl_hash_string` wasn't in the public API headers and thus not part of the public API. It's unlikely that any other package relied on it, so removing it from the ABI _shouldn't_ break anything. There are similar commits for [mbszero](https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=6027f38bd7c45742bbd69723d4fa5f4d6d466fcd) and [rpl_isnanf](e37cb87ec99d6ff42afc801f3e09f2355483c950). It's understandable that an automated tool may flag all dependent packages as needing rebuilding because such a tool sees an ABI change.
> 
> I was looking at MSYS2's Gettext version because Gettext 0.23.1 fixes UTF-8 code page usage on native Windows. The fix is in the Gnulib files included in libintl so it's not visible in Gettext's Git log.
> 
> XZ Utils >= 5.6.3 uses UTF-8 code page. With old Gettext, non-ASCII chars are garbled when translation support is enabled. Next XZ Utils release will allow enabling translations on native Windows only if Gettext >= 0.23.1 and using UCRT.

The package grokker look for symbols removed and actually used by dependent packages. It doesn't flag everything. sometimes some packages have ABI but not all dependent packages need to be rebuilt.

MehdiChinoune:
Even after restoring symbol `libintl_hash_string` package-grokker still showing packages need to be rebuilt, so the missing symbol that break ABI is `DllMain` (from gettext-runtime\gnulib-lib\relocatable.c) which no longer used since https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=b182c17096847cb581c3e2c3f42c145c47c7c7a6

MehdiChinoune:
I am done.

Larhzu:
Sorry, I commented this PR but I'm not able to help with this.

PR #22957, opened 2025-01-02
Remove mingw-w64-python-jose

lazka:
CVE, no release in 3-4 years, no users in MSYS2

just a proposal, @podsvirov are you still using it?


podsvirov:
Ok to remove. I will migrate to pyjwt directly.

lazka:
ok, thanks. I've updated pyjwt while at it #22970

PR #22958, opened 2025-01-02
seacas: add libexodus (provides)

3rav:


MehdiChinoune:
exodus is already a separate package that is built from the same repo but from different tag.

3rav:
But this does not change the fact that when installing the seacas package it will contain the exodus lib

MehdiChinoune:
> But this does not change the fact that when installing the seacas package it will contain the exodus lib

If your claim is true, then libexodus package should be removed and provided by seacas.
according to https://github.com/sandialabs/seacas?tab=readme-ov-file#exodus exodus is just a component of seacas that could be built separately. Giving that no package depends on libexodus I think It's safe to drop it and make seacas provide it.

PR #22976, opened 2025-01-03
gdal: Update to 3.10.0

lazka:
and backport some build fixes


MehdiChinoune:
gdal 3.10 requires both vtk and paraview to be rebuilt #22790

lazka:
thanks. closing for #22978 then

PR #22989, opened 2025-01-04
tomlplusplus: Fix wrong library location according to CMake

### LINKED ISSUE #22988: mingw-w64-tomlplusplus: Wrong library location according to CMake

Include this package in a CMake-based project:
```cmake
find_package(tomlplusplus)
```

### Expected behavior

CMake should accurately locate the library

### Actual behavior

CMake/Ninja fails to find the library because it is specified wrong, perhaps an artifact of the original porting from a Linux system not accounting for differences with Windows:
```console
$ grep IMPORTED_LOCATION $MINGW_PREFIX/lib/cmake/tomlplusplus/tomlplusplusConfig.cmake
      IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/lib/libtomlplusplus-3.dll"
$ test -f $MINGW_PREFIX/lib/libtomlplusplus-3.dll && echo yes || echo no
no
$ test -f $MINGW_PREFIX/bin/libtomlplusplus-3.dll && echo yes || echo no
yes
```
Libraries on Windows live next to executables (e.g., in `/bin` instead of `/lib`), and therefore should be adjusted:
```diff
diff --git a/1 b/2
index bc3421a..882927c 100644
--- a/1
+++ b/2
@@ -34,7 +34,7 @@ if(NOT TARGET tomlplusplus::tomlplusplus)
   # Set the path to the installed library so that users can link to it
   if (1)
     set_target_properties(tomlplusplus::tomlplusplus PROPERTIES
-      IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/lib/libtomlplusplus-3.dll"
+      IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/bin/libtomlplusplus-3.dll"
     )
     # compile_options not quoted on purpose
     target_compile_options(tomlplusplus::tomlplusplus INTERFACE -DTOML_HEADER_ONLY=0 -DTOML_SHARED_LIB=1)
```

### Verification

### Windows Version

MINGW64_NT-10.0-22631

- [X] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANGARM64

Yes

500-internal-server-error:
Fixes #22988

### DIFF
--- a/mingw-w64-tomlplusplus/PKGBUILD
+++ b/mingw-w64-tomlplusplus/PKGBUILD
@@ -28,6 +28,7 @@ build() {
       "${_realname}-${pkgver}"
 
   meson compile -C "build-${MSYSTEM}"
+  sed -i 's/lib/bin/' "build-${MSYSTEM}/src/tomlplusplusConfig.cmake"
 }
 
 package() {


mingw-w64-tomlplusplus/PKGBUILD
```diff
@@ -28,6 +28,7 @@ build() {
       "${_realname}-${pkgver}"
 
   meson compile -C "build-${MSYSTEM}"
+  sed -i 's/lib/bin/' "build-${MSYSTEM}/src/tomlplusplusConfig.cmake"
```
ognevny, line 31: is it possible to patch source `.cmake` file instead of using `sed`?

500-internal-server-error: Probably, but I figured since it's a very simple change an entire `.patch` file would be overkill/overcomplicated. The source uses a `.cmake.meson.in` file, not a `.cmake` file directly, so patching would have to be done there if not using `sed`.

Biswa96: I agree with ognevny and shall try to create a separate pull request with a proper patch. tomlplusplusConfig.cmake may contain other `lib` string and replacing every occurrence of that would be incorrect.

500-internal-server-error:
Superseded by #23009

PR #22995, opened 2025-01-04
python-cx-freeze: update to 7.2.8

marcelotduarte:


lazka:
Does that mean `*.h` is missing here: https://github.com/marcelotduarte/cx_Freeze/blob/1a7228a620a972e4af2fb7ea000f9ec55d1ee982/MANIFEST.in#L12 ?

marcelotduarte:
Yes, it will be resolved for the next versions.

PR #23009, opened 2025-01-05
tomlplusplus: Fix dll path in cmake imported location

### LINKED ISSUE #22988: mingw-w64-tomlplusplus: Wrong library location according to CMake

Biswa96:
* Fixes #22988

### DIFF
--- a/mingw-w64-tomlplusplus/0001-tomlplusplus-fix-cmake-imported-location.patch
+++ b/mingw-w64-tomlplusplus/0001-tomlplusplus-fix-cmake-imported-location.patch
@@ -1,0 +1,21 @@
+--- a/cmake/tomlplusplusConfig.cmake.meson.in
++++ b/cmake/tomlplusplusConfig.cmake.meson.in
+@@ -20,7 +20,7 @@
+   # Set the path to the installed library so that users can link to it
+   if (@compile_library@)
+     set_target_properties(tomlplusplus::tomlplusplus PROPERTIES
+-      IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/@libdir@/@lib_name@"
++      IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/@bindir@/@lib_name@"
+     )
+     # compile_options not quoted on purpose
+     target_compile_options(tomlplusplus::tomlplusplus INTERFACE @compile_options@)
+--- a/src/meson.build
++++ b/src/meson.build
+@@ -81,6 +81,7 @@
+ 			'compile_library': true,
+ 			'compile_options': cmake_compile_options,
+ 			'includedir': get_option('includedir'),
++			'bindir': get_option('bindir'),
+ 			'libdir': get_option('libdir'),
+ 			'lib_name': lib_name
+ 		})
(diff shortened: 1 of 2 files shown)


Biswa96:
@500-internal-server-error would you like to review the change please?

ognevny (approved):
thanks

500-internal-server-error:
@Biswa96 Testing locally, `bindir` is empty on Windows it seems, perhaps it needs to be hardcoded?

Biswa96:
> `bindir` is empty on Windows it seems, perhaps it needs to be hardcoded?

I have added the variable in meson file too. The cmake section now looks like following.

```cmake
  if (1)
    set_target_properties(tomlplusplus::tomlplusplus PROPERTIES
      IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/bin/libtomlplusplus-3.dll"
    )
    # compile_options not quoted on purpose
    target_compile_options(tomlplusplus::tomlplusplus INTERFACE -DTOML_HEADER_ONLY=0 -DTOML_SHARED_LIB=1)
  endif()
```

ognevny:
> Testing locally, `bindir` is empty on Windows it seems, perhaps it needs to be hardcoded?

according to https://mesonbuild.com/Builtin-options.html#directories it's always set

500-internal-server-error:
Hmm, I guess I was doing something wrong then. Downloading the build artifact confirms it is set correctly.

Biswa96:
> according to https://mesonbuild.com/Builtin-options.html#directories it's always set

`bindir` is set in meson but it is not passed to generate .cmake file. The similar thing is done with `libdir` in src/meson.build.

PR #23011, opened 2025-01-05
gettext: update to 0.23.1

MehdiChinoune:
I have restored `DllMain` symbol to silent package-grokker about rebuilding packages.


MehdiChinoune:
I observed in recent days (same as here) that sometimes buidling autotools packages hang. Maybe something related to msys2-runtime

lazka:
> I observed in recent days (same as here) that sometimes buidling autotools packages stucks. Maybe something related to msys2-runtime
> for example:

yeah, see https://github.com/msys2/msys2-runtime/pull/251

PR #23019, opened 2025-01-06
Add patch to build gr-network with boost 1.87

chrisjohgorman:
- also, update to build for clang64 and calangarm64 as they both have python-scipy again.
- remove references to unused Boost_REGEX_LIBRARY_RELEASE and Boost_SYSTEM_LIBRARY_RELEASE.
- update pkgrel.


chrisjohgorman:
This now builds thanks to an updated spdlog, which has the necessary patch in it to work with fmt 11.1.1.  I have downloaded the artifacts and run them on my machine.  The artifacts pull in the necessary dependencies and run properly, so I'm changing this to ready for review.

PR #23028, opened 2025-01-07
rust: update to 1.84.0

ognevny:
config changes:
- enable CFG for gnullvm targets (closes #22325)
- remove parallel_compiler, it's now always true (but `-Zthreads` is still unstable)


ognevny:
that clang 19 thing again
```
CFLAGS_wasm32v1_none = Some(-ffunction-sections -fdata-sections --target=wasm32v1-none -mtune=generic -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wp,-D__USE_MINGW_ANSI_STDIO=1)
cargo:rerun-if-env-changed=CC_SHELL_ESCAPED_FLAGS
CC_SHELL_ESCAPED_FLAGS = None
cargo:warning=clang: warning: argument unused during compilation: '-mtune=generic' [-Wunused-command-line-argument]
cargo:warning=error: unknown target triple 'wasm32v1-none'
```

ognevny:
@mati865 there is a new binary shipped: rust-objcopy.exe. is it better to remove it? there shouldn't be a problem if it's not shipped here

mati865:
> @mati865 there is a new binary shipped: `rust-objcopy.exe`. is it better to remove it? there shouldn't be a problem if it's not shipped here

I doubt lack of it would be a problem for us.

> * enable CFG for gnullvm targets (closes [rust: enable Control Flow Guard on gnullvm #22325](https://github.com/msys2/MINGW-packages/pull/22325), thanks mati865)

Wasn't the only blocker there the decision whether MSYS2 should enable it, since upstream does not?

ognevny:
> Wasn't the only blocker there the decision whether MSYS2 should enable it, since upstream does not?

I've been waiting for a patch that adds CFG support for gnullvm being available in stable

mati865:
Oh, I must have messed up the rebase. I have backported that patch.

PR #23047, opened 2025-01-09
eza: update to 0.20.16

ognevny:


lazka:
I guess we need to update libgit2?

ognevny:
> I guess we need to update libgit2? always such a pain...

maybe. for this package I reverted git2 crate update. I'll try to update libgit2 after merging this

lazka:
I created #23050 for starters, but feel free to take that over.

PR #23051, opened 2025-01-09
fltk: disable GDI+

mmuetzel:
If FLTK is built with GDI+, unloading the library leads to a deadlock (at least in Octave).
See also: https://octave.discourse.group/t/6097/11

Following up on the discussion in #22620.


mmuetzel:
I downloaded and installed the MINGW64 build artifacts from this PR. With them, I can no longer reproduce the issue after running the following commands in `octave --no-gui-libs`:
```
graphics_toolkit gnuplot
graphics_toolkit fltk
exit
```
(This previously lead to a deadlock on `exit`.)

PR #23064, opened 2025-01-10
[new package] libjudy

omikrun:
It's a dependency of mariadb, netdata and so on (although they are both not included in MSYS2). I am trying to build the whole mariadb rather than a client only for my local website. However, I am not sure I could...

This package is not original libjudy, it's a fork maintained by netdata. The original library does work on Linux only and it may causes undefined behavior on Windows, therefore I made possibly efforts to make it portable. Yup, I'm not tested.

Oh, unfortunately, I'm not sure whether it can run under CLANGARM64, your feedback would be greatly appreciated.


Kreijstal:
could  mariadb get compiled with this?

omikrun:
> could mariadb get compiled with this?

Actually, after a few attempts, no. :(
MariaDB doesn't compile properly on mingw-w64. Some libraries like `ws2_32.lib` were hardcoded in their `CMakeLists.txt`. Therefore, it only supports MSVC on Windows. However I think it's possible and capable.

Kreijstal:
> > could mariadb get compiled with this?
> 
> Actually, after a few attempts, no. :( MariaDB doesn't compile properly on mingw-w64. Some libraries like `ws2_32.lib` were hardcoded in their `CMakeLists.txt`. Therefore, it only supports MSVC on Windows. However I think it's possible and capable.

just replace it with libws2_32.a no?

lazka:
That fork looks outdated and only has two small patches on the original? I'd prefer the non-fork variant with patches instead of that is possible.

omikrun:
> That fork looks outdated and only has two small patches on the original? I'd prefer the non-fork variant with patches instead of that is possible.

Oh, yes, you're right. The original release is better. One of the two small patches was fixed in official 1.0.5 release. Another is applied in PKGBUILD now. :)

PR #23077, opened 2025-01-10
kicad: fix building with libc++ from LLVM 19

mmuetzel:
Avoid instantiating the std::char_traits<T> template with arbitrary types.
That is no longer possible with the implementation of the C++ STL from LLVM 19.


ognevny:
thanks

PR #23092, opened 2025-01-12
SDL3: add package

### LINKED ISSUE #23083: Add SDL3
### Package name

SDL3

SDL3 is the next major release of SDL. While there are currently only pre releases available for SDL3 before 3.2.0 releases, they are properly versioned with 3.1.x numbers, they have been heavily battle tested by Valve and have already frozen the ABI. Certain Linux distros like Fedora also already package SDL3 3.1.x releases in preparation for the full release.

I hope that would be enough to get it included in the MSYS2 repos already since it would be useful for early adopters who want to update from SDL2 early.

https://libsdl.org

SDL3 comes with some examples and demos in the `examples/` folder that can be used to test that everything works right.

- [ ] MINGW64
- [ ] MINGW32
- [X] UCRT64
- [X] CLANG64
- [X] CLANGARM64

Not me personally, but sewbacca seems to be willing to

Biswa96:
* Fixes #23083

### DIFF
--- a/mingw-w64-sdl3/PKGBUILD
+++ b/mingw-w64-sdl3/PKGBUILD
@@ -1,0 +1,54 @@
+# Maintainer: Biswapriyo Nath <nathbappai@gmail.com>
+
+_realname=sdl3
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=3.1.8
+pkgrel=1
+pkgdesc="A library for portable low-level access to a video framebuffer, audio output, mouse, and keyboard (Version 3) (mingw-w64)"
+arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
Could we go with lowercase naming like other distros i.e. sdl3?

Biswa96:
Yeah, sure. I use uppercase to make the name consistent with other mingw SDL packages. Some Linux distributions use uppercase too https://repology.org/project/sdl3/versions e.g. Fedora, OpenMandriva, OpenSUSE etc.

lazka:
Yeah, ideally we'd convert all packages to lowercase for consistency. Feel free to leave it if you want, we can change it later for all packages.

Biswa96:
Does it look ok now?

lazka:
yes

PR #23100, opened 2025-01-14
vtk: update to 9.4.1

mmuetzel:
Remove patches that have been applied upstream.
Fix compilation issues with `clang++` version 19.
Fix new build issues with fmt 11.
Cherry-pick patches from upstream VTK-m.
Update patch to make it apply cleanly.
Update DIY to version 3.6.0 to avoid compilation issue with C++ STL from LLVM 19 (`std::char_traits<T>` can no longer be instantiated with arbitrary types).


MehdiChinoune:
I can't review the changes as I don't have enough knoweldge about VTK internals, but I think they don't use diy2 as It is but make some changes to make it suitable for VTK.

mmuetzel:
Good point.
Afaict, the only changes are that they added CMakeLists.txt files (which I adapted for this change) and that they changed one include directive to include a wrapper header for MPI. See the last four commits on the branch they imported originally:
https://github.com/diatomic/diy/commits/for/vtk-20191127-gd6d04d0d

The latter (i.e., the MPI header) isn't really relevant for the package that is distributed by MSYS2 because it is built without MPI. But I agree that it should be changed nevertheless for the sake of completeness.
I updated the patch to include that change.

The only part that I didn't touch (but maybe should be adapted) is the script that they used to import that branch originally. Instead of using it, I copied the files manually (and made the changes mentioned before).
That is probably only an issue if we tried to upstream that patch as it is.

mmuetzel:
I added a couple of new patches and (optional) dependencies to fix a build issue with the InfovisBoostGraphAlgorithms and the RenderingAnari modules and to build the new IOAlembic and the (aforementioned) RenderingAnari modules.

mmuetzel:
How does one add an `optdepends` for only some `MSYSTEM`s with description?
It is working for me locally. But it fails with `==> ERROR: optdepends contains invalid characters: ':'` in CI.

mmuetzel:
Should be ready for review (again) now.

MehdiChinoune:
No, AFAIR.

mmuetzel:
Ok. Merging in its current state in that case.

PR #23123, opened 2025-01-16
ruby: update to 3.4.1

jeremyd2019:
All patches are upstream now.


jeremyd2019:
I doctored CI to run tests also, and they passed on everything except MINGW32, which seems to have gotten confused about the existence of a symbol `DllMain@12`.  https://github.com/msys2-arm/MINGW-packages/actions/runs/12816890918/job/35738860664#step:12:7633

ognevny:
drop mingw32? would it be hard to do?

jeremyd2019:
It's fine, it's just a false positive test failure.

jeremyd2019:
I wasn't sure about mruby - it doesn't link to ruby, and it doesn't install any files to the 3.3.0 gem directory, but I don't know if it needs a rebuild or update to pick up anything...  I think it's its own thing, so it should be ok 🤞

PR #23124, opened 2025-01-16
paraview: update to 5.13.2

mmuetzel:
Depend on VTK package.


MehdiChinoune:
Paraview upstream doesn't support building against external VTK other than the one build against (the same commit). It may work or not, It could break at any point.
For example VCPKG package them both fellowimg Paraview release.

mmuetzel:
Thanks for the feedback.

If I understand correctly, building with an external VTK (with shared linking) is supported by upstream since this change:
https://gitlab.kitware.com/paraview/paraview/-/merge_requests/6375

I couldn't find a more recent information that says it is not supported. Did I miss something?

My motivation for depending on the packaged version of VTK is to avoid having to maintain patches for VTK in various versions in multiple packages. Depending on the existing package might mean less maintenance in the future.

I downloaded and installed the CLANG64 build artifacts from this PR. Running `paraview.exe` seems to be working for me.

MehdiChinoune:
https://gitlab.kitware.com/paraview/paraview/-/blob/v5.13.2/Documentation/dev/build.md#L556

mmuetzel:
Thanks for the links.

So, the version of the external VTK has to match the version of the bundled VTK (e.g., 9.3). 
Since both projects seem to be collaborating quite closely, they'll probably keep close compatibility with the most recent version of VTK. Worst case will probably be that we'd need to cherry-pick a couple of upstream patches if the latest version of ParaView was released before the latest version of VTK (like right now the patches for the changed implementation of OpenGL in VTK).
I still believe that this will mean less maintenance burden for MSYS2 in the long run.

Do you expect something different?

MehdiChinoune:
I was feloowing that effort to build Paraview against external VTK since I added paraview, but I was hesitating because of their strict requirements.
If you feel that's safe, I am not going to object.

mmuetzel:
Imho, we could try to use the packaged version of VTK. If it turns out that this results in more effort than patching the bundled version, we could still revert to that bundled version of VTK in the future.

mmuetzel:
Let's see how it fares. Like I mentioned, we can still revert to the bundled VTK if this change is causing trouble...

PR #23131, opened 2025-01-17
procs: update to 0.14.9

ognevny:


mingw-w64-procs/PKGBUILD
chiefofchildren, line 1: ****

chiefofchildren (commented):
gh pr checkout 23131

PR #23135, opened 2025-01-17
qemu: reenable clang builds

pbo-linaro:
Upstream QEMU was fixed, by removing usage of gcc_struct attribute, and enabling plugins linkage using lld.
- https://gitlab.com/qemu-project/qemu/-/commit/8f5a4cfc7ed9e06e07fdd8e8fdf50ef3ea783f63
- https://gitlab.com/qemu-project/qemu/-/commit/923710b6d5b21d9b3fcecc7e6719cfa5a53de268

Those changes will be available for 10.0, but meanwhile, we can safely backport them on 9.2, as the gcc_struct attribute investigation was made on 9.2 release (see patch commit messages for more details and links).

Note: dependency virglrenderer is skipped because
mingw-w64-clang-X-virglrenderer is not available.

### DIFF
--- a/mingw-w64-qemu/PKGBUILD
+++ b/mingw-w64-qemu/PKGBUILD
@@ -15,11 +15,10 @@ _rc_base_ver="9.1.9${_rc_no}"
 _rc_file_ver="rc${_rc_no}"
 _tarname=$( [ -z "${_rc_no}" ] && echo ${_realname}-${_base_ver} || echo ${_realname}-${_base_ver}-${_rc_file_ver} )
 pkgver=$( [ -z "${_rc_no}" ] && echo ${_base_ver} || echo ${_rc_base_ver} )
-pkgrel=1
+pkgrel=2
 pkgdesc="QEMU - a generic and open source machine emulator and virtualizer (mingw-w64)"
 arch=('any')
-# clang disabled until attribute gcc_struct is supported, see https://github.com/msys2/MINGW-packages/pull/21540#issuecomment-2287923652
-mingw_arch=('mingw64' 'ucrt64')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
 license=('spdx:GPL-2.0-or-later AND LGPL-2.1-or-later AND BSD-2-Clause AND BSD-3-Clause AND MIT')
 url="https://qemu.org/"
 msys2_repository_url="https://gitlab.com/qemu-project/qemu"
@@ -101,14 +100,17 @@ depends=(
   "${MINGW_PACKAGE_PREFIX}-snappy"
   "${MINGW_PACKAGE_PREFIX}-spice"
   "${MINGW_PACKAGE_PREFIX}-usbredir"
-  "${MINGW_PACKAGE_PREFIX}-virglrenderer"
+  $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-virglrenderer")
   "${MINGW_PACKAGE_PREFIX}-xz"
   "${MINGW_PACKAGE_PREFIX}-zlib"
   "${MINGW_PACKAGE_PREFIX}-zstd"
 )
+
 options=('!emptydirs')
 source=(
   "https://download.qemu.org/${_tarname}.tar.xz"{,.sig}
+  01-win32-remove-usage-of-attribute-gcc_struct.patch::https://gitlab.com/qemu-project/qemu/-/commit/8f5a4cfc7ed9e06e07fdd8e8fdf50ef3ea783f63.patch
+  02-plugins-enable-linking-with-clang-lld.patch::https://gitlab.com/qemu-project/qemu/-/commit/923710b6d5b21d9b3fcecc7e6719cfa5a53de268.patch
   msys2.readme.txt
   msys2.qemu-guest-agent.txt
   msys2.examples.tests.sh
@@ -116,6 +118,8 @@ source=(
 sha256sums=(
   'f859f0bc65e1f533d040bbe8c92bcfecee5af2c921a6687c652fb44d089bd894'
   'SKIP'
+  '16056f047d838245bc8cc327dc2a73ab9b9ee851ffee7a62f7a16530ba21c0f3'
+  '385f3fae53afa29b5d344bb6565f24833d22cf0a6d407a1efbd2cfb0e94c407e'
   '51625fd83c0a63729942d3dfc6d48be3491a700fe0f7cb8e88935b46081c9016'
   'f2f9eeb31023d002f54637a9941ea0d8fae3c6f0a66c05c033857b305bd1470d'
   '1478e41f27a4646e14c2abca764d9fd2fd158a4a4cd6421e077fba101d4a74ea'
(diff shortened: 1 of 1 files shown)


pbo-linaro:
It's the first time I edit a MSYS2 package, so don't hesitate to point anything wrong.

I tested this on a Windows on Arm machine, by building QEMU using makepkg, and by using generated binaries (including with a plugin).

pbo-linaro:
Pinging potential reviewers based on PR for bumping version to 9.2.0.

pbo-linaro:
Fixed a small issue for conditional depends list.

mingw-w64-qemu/PKGBUILD
```diff
@@ -101,21 +100,33 @@ depends=(
...
 )
+case ${MSYSTEM} in
+  CLANG*)
+    # virglrenderer is not enabled for clang platforms.
+    ;;
+  *)
+    depends+=("${MINGW_PACKAGE_PREFIX}-virglrenderer")
+    ;;
+esac
+
```
MehdiChinoune, line 115: First, we 'depends' is an array not a string, so it should be 'depends+=("...")'
Second we don't use that form for packages available for specific envs. We use $([[ ${MINGW_PREFIX_PACKAGE} == *-clang-* ]] || echo "${MINGW_PREFIX_PACKAGE}-package").

MehdiChinoune (changes requested):

mingw-w64-qemu/PKGBUILD
pbo-linaro: I just fixed usage of `(` after seeing the CI failure :).
I'll update it to the conditional you mentioned. Thanks.

pbo-linaro:
Fixed the conditional to not depend on `virglrenderer` package for clang platforms.

mingw-w64-qemu/PKGBUILD
jeremyd2019: Something similar is already being done on `makedepends` above.

jeremyd2019 (approved):
lgtm

mingw-w64-qemu/PKGBUILD
pbo-linaro: Yes this is why I followed this pattern. However, I think makedepends change makes sense because it is a commented addition, compared to this one who is just straightforward.

mingw-w64-qemu/PKGBUILD
```diff
@@ -101,21 +100,26 @@ depends=(
...
+  $([[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-virglrenderer")
   "${MINGW_PACKAGE_PREFIX}-xz"
   "${MINGW_PACKAGE_PREFIX}-zlib"
   "${MINGW_PACKAGE_PREFIX}-zstd"
 )
+
 options=('!emptydirs')
 source=(
   "https://download.qemu.org/${_tarname}.tar.xz"{,.sig}
+  01-win32-remove-usage-of-attribute-gcc_struct.patch
```
heljkon, line 112: I would prefer retrieving patches like this:

01-win32-remove-usage-of-attribute-gcc_struct.patch::https://gitlab.com/qemu-project/qemu/-/commit/8f5a4cfc7ed9e06e07fdd8e8fdf50ef3ea783f63.patch

mingw-w64-qemu/PKGBUILD
```diff
@@ -18,8 +18,7 @@ pkgver=$( [ -z "${_rc_no}" ] && echo ${_base_ver} || echo ${_rc_base_ver} )
 pkgrel=1
```
heljkon, line 18: Please set pkgrel=2

mingw-w64-qemu/PKGBUILD
pbo-linaro: Neat! I didn't know it was possible to do it.

mingw-w64-qemu/PKGBUILD
pbo-linaro: Updated, and added a comment.

pbo-linaro:
Increased `pkgrel` and download patches directly from upstream QEMU repo.

mingw-w64-qemu/PKGBUILD
jeremyd2019: my preference would be not to have a comment there

pbo-linaro: Ok, I change this.

pbo-linaro:
Removed comment for `pkgrel`.

driver1998:
I think virglrenderer can be brought back as well. IIRC it got removed simply because QEMU is the only thing that uses it, and QEMU got removed.

heljkon:
> I think virglrenderer can be brought back as well. IIRC it was removed simply because QEMU is the only thing that uses it, and QEMU was removed. (see #21903).

It was removed, because as QEMU virglrenderer also has the issue
`warning: unknown attribute 'gcc_struct' ignored [-Wunknown-attributes]`
and QEMU was the only user

see https://github.com/msys2/MINGW-packages/pull/21903

pbo-linaro:
Thanks for the merge.

Concerning `virglrenderer`, even though QEMU has been able to remove this attribute does not mean the same thing is possible for `virglrenderer`.
Basically, library needs to be compiled with and without the attribute, with debug info, and structures needs to be compared to identify any potential ABI difference. The first patch we backport in QEMU contains a link to a conversation explaining the whole issue, and the investigation made to eradicate this from QEMU.

jeremyd2019:
As another test point for qemu (or just general messing-around) you might be interested in a workflow I wrote: https://github.com/jeremyd2019/winpebuilder/blob/main/.github/workflows/build.yml  This builds a ready-to-use-with-qemu package of a Windows-on-ARM64 WinPE iso with virtio drivers and x64 emulation support, a qcow2 image with msys2 x86_64 extracted on it, the latest edk2 image from Debian sid, and a shell script to run it under qemu.  I think the license for WinPE does not allow for redistribution, so I delete the artifacts after running it, and instead offer the script to build it for yourself.  The workflow takes an optional comma-delimited list of URLs of Latest Cumulative Updates to install on the WinPE image, which for January would be `https://catalog.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/f263aa64-f367-42f0-9cad-328f342b93f7/public/windows11.0-kb5043080-arm64_df540a05f9b118e339c5520f4090bb5d450f090b.msu,https://catalog.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/fb50861d-95ca-47ee-80f0-7bee850d0674/public/windows11.0-kb5050009-arm64_d3440f1c6b2b9471cd2cdfae7b360006d847d505.msu`

Now all I need is whpx support on Windows-on-ARM64 😜

pbo-linaro:
Nice, thanks for sharing.

I have a similar CI ([workflow](https://github.com/pbo-linaro/qemu-boot-windows-ci/blob/master/.github/workflows/boot_windows.yml) + [qemu test](https://github.com/pbo-linaro/qemu-boot-windows-ci/blob/master/tests/functional/test_aarch64_virt_windows.py)) booting Windows Validation OS (which is a stripped down version of Win11). It's focused on booting the Windows kernel (i.e. to be fast).

Having WinPE is probably better because I guess it has less limitations than Validation OS.

Did you consider having a way to automatically run a specific command in the VM?
It could be useful to have ephemeral runners for GitHub (triggered from the pipeline itself).
However, windows-arm64 GitHub runners should hopefully be available soon for free.

jeremyd2019:
Hmm, I haven't thought about even trying to get actions/runner to work on WinPE.  I'd probably have to install at least .NET and powershell, maybe more of the optional components... Might be interesting to try though.

pbo-linaro:
Or even better, booting a WinPE VM with an accessible ssh server, so it's easy to run any command/script on it.

jeremyd2019:
> Or even better, booting a WinPE VM with an accessible ssh server, so it's easy to run any command/script on it.

Done, see jeremyd2019/winpebuilder#2

Issue #21017, opened 2024-05-29
librsvg problem in clang


revelator:
### Description / Steps to reproduce the issue

building librsvg with clang produces this error ->

```
ld.lld: error: librsvg_c_api.a(bcryptprimitives.dll): .idata$4 should not refer to special section 0
clang: error: linker command failed with exit code 1 (use -v to see invocation)

```

### Expected behavior

should produce the librsvg build like with gcc

### Actual behavior

bugs out with the above error

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-26100

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [X] CLANG64
- [X] CLANG32
- [ ] CLANGARM64

Biswa96:
I can reproduce the issue with clang64.

revelator:
yeah not really sure whats going on there ???

revelator:
looks like an invalid section in the bcryptprimitives export though it is not even part of the mingw-w64-abi only bcrypt.
seems it links directly to the dll but fails and it is not clear if t is a problem with the compiler or something else.

revelator:
ah i smell a mingw specific linker script sec investigating.

mati865:
There are not linker/version scripts involved since LLD doesn't support them at all and would just error out.
The library is linked by Rust since https://github.com/rust-lang/rust/pull/121337 by creating own import library through LLVM backend.

revelator:
exactly what i was after but yeah librsvg does not use a linker script so it seems to be something else going on.

revelator:
ok it is rust it seems, tried reverting to a previous version and now it builds...

mati865:
How to reproduce this problem?
For me clean checkout of librsvg from git (commit `b4d6f2071342ef5be36f25253a9bf1bf9277f8ec`) and running with Rust from CLANG64 works:
```
$ cargo b
...
Finished `dev` profile [unoptimized + debuginfo] target(s) in 44.55s

$ cargo clean

$ cargo b -r
...
Finished `release` profile [optimized] target(s) in 51.56s

$ rustc -vV
rustc 1.78.0 (9b00956e5 2024-04-29) (Rev1, Built by MSYS2 project)
binary: rustc
commit-hash: 9b00956e56009bab2aa15d7bff10916599e3d6d6
[...]

PR #23136, opened 2025-01-18
llvm: update to 19.1.7 and add import libs fix for LLD

mati865:
Fixes https://github.com/msys2/MINGW-packages/issues/21017


lazka:
lgtm. ok to merge despite it timing out?

mati865:
I had built them each one in CLANG64 env before committing, so it should be fine, but if you want, I can just split it into two PRs.

lazka:
I think it's fine to just merge

mingw-w64-llvm/PKGBUILD
```diff
@@ -112,6 +115,14 @@ apply_patch_with_msg() {
...
 }
 
+apply_git_patch_with_msg() {
+  for _patch in "$@"
+  do
+    msg2 "Applying ${_patch}"
+    patch -Nbp1 -i "${srcdir}/${_patch}"
+  done
+}
+
```
jeremyd2019, line 125: This function looks exactly like the `apply_patch_with_msg` function immediately above it.  Is there some reason for another copy?

mati865: Good point, this is a leftover.
At first I tried to apply the patch to a specific directory (`lld/`) but git patches require `-p2` so I created this helper. Then I found out it's not going to work because the patch modifies `llvm/` and `lld/` at once. And here comes my mistake for modifying the new helper instead of using the previous one.

mati865: Created https://github.com/msys2/MINGW-packages/pull/23152

PR #23137, opened 2025-01-18
librsvg: drop LLD workaround

mati865:
Stacked on top of https://github.com/msys2/MINGW-packages/pull/23136


mati865:
Retried build after LLVM update: https://github.com/mati865/MINGW-packages/actions/runs/12853599587

lazka:
thanks!

Biswa96:
The patch file was not removed.

lazka:
thanks, removed

PR #23150, opened 2025-01-19
sox: Update to latest master

orgads:


lazka:
the version is auto generated on build, so it was fine as is.

orgads:
I see, thanks. Squashed your commit and removed the message.

PR #23152, opened 2025-01-19
llvm: remove redundant helper

mati865:


jeremyd2019:
As long as you're doing that, you could fix the comment on line 80 as well, the patch is 09xx not 99xx as the comment suggests

PR #23164, opened 2025-01-22
Update skia to m132

kcleal:
This updates skia to the latest stable release


kcleal:
Patching BUILD.gn looks like a nightmare. Might make sense to create a new build file, rather than applying a huge patch.

PR #23167, opened 2025-01-22
Test runtime hangs

lazka:
rebuild some things which are likely to produce hangs


jeremyd2019:
No hang that time, with 3.5.5-3

lazka:
Nice

lazka:
second time also.

thanks @tyan0 !

lazka:
passed two times with 3.5.6, so closing

PR #23200, opened 2025-01-26
[new package] crystal 1.15.1

HertzDevil:
Adds the [Crystal programming language](https://crystal-lang.org/).

Crystal must be compiled from Crystal itself. An initial 1.15.0 compiler is cross-compiled from Ubuntu using a [GitHub workflow](https://github.com/crystal-lang/crystal/blob/9e475f6442465b4dda485ff12d2dfd70cda25efe/.github/workflows/mingw-w64.yml); the language's forward compatibility guarantee ensures that this initial compiler should be able to build any subsequent 1.x compiler.

Does not include the [Crystal-specific package manager](https://github.com/crystal-lang/shards) yet.


mingw-w64-crystal/PKGBUILD
```diff
@@ -0,0 +1,70 @@
...
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+_major=1
+_minor=15
+_patch=0
+_version="${_major}.${_minor}.${_patch}"
+_tag=${_version}
+pkgver=${_version}
+pkgrel=1
+pkgdesc="Fast and statically typed, compiled language with Ruby-like syntax"
+arch=('x86_64')
```
Biswa96, line 14:
```suggestion
arch=('any')
```

mingw-w64-crystal/PKGBUILD
```diff
@@ -0,0 +1,70 @@
...
+pkgbase="mingw-w64-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+_major=1
+_minor=15
+_patch=0
+_version="${_major}.${_minor}.${_patch}"
+_tag=${_version}
+pkgver=${_version}
+pkgrel=1
+pkgdesc="Fast and statically typed, compiled language with Ruby-like syntax"
```
Biswa96, line 13:
```suggestion
pkgdesc="Fast and statically typed, compiled language with Ruby-like syntax (mingw-w64)"
```

mingw-w64-crystal/PKGBUILD
```diff
@@ -0,0 +1,70 @@
...
+source=(
+  https://github.com/crystal-lang/crystal/releases/download/1.15.0/crystal-1.15.0-windows-x86_64-gnu-unsupported.zip # stage 0 compiler
+)
+sha256sums=(
+  '7d7df1b4a99cb3f938106ba45e0b55ce1d9913491999f1acaa57da168a18becb'
+)
+
+prepare() {
+  # workaround for lack of native symlink support in Git
+  git clone https://github.com/crystal-lang/crystal.git "$pkgname-$pkgver" -c core.symlinks=true || \
```
Biswa96, line 52: This should be in source array with proper `#tag` in the URL. see https://wiki.archlinux.org/title/VCS_package_guidelines

mingw-w64-crystal/PKGBUILD
```diff
@@ -0,0 +1,70 @@
...
+  "${MINGW_PACKAGE_PREFIX}-zlib"
+)
+makedepends=(
+  "git"
+)
+checkdepends=(
+  "${MINGW_PACKAGE_PREFIX}-lld" # needed for linking std_spec
+)
+source=(
+  https://github.com/crystal-lang/crystal/releases/download/1.15.0/crystal-1.15.0-windows-x86_64-gnu-unsupported.zip # stage 0 compiler
```
Biswa96, line 44:
```suggestion
  https://github.com/crystal-lang/crystal/releases/download/${pkgver}/crystal-${pkgver}-windows-x86_64-gnu-unsupported.zip # stage 0 compiler
```

HertzDevil: The stage 0 compiler is always 1.15.0, it doesn't change with the MSYS2 package version

mingw-w64-crystal/PKGBUILD
HertzDevil: I tried, but our repository contains internal, recursive symlinks that seem to break the prepare step

mingw-w64-crystal/PKGBUILD
Biswa96: But older releases in crystal repository have those compiled tarball also, not the gnu one. Wouldn't future releases have that *-gnu-unsupported.zip ?

HertzDevil: Maybe, but if we always use a recent version of the compiler then I believe Crystal itself can be added to `makedepends` instead once it is available in MSYS2, so there is never a need to use any version other than 1.15.0 here

mingw-w64-crystal/PKGBUILD
ognevny: try something like https://github.com/msys2/MSYS2-packages/blob/c78ec19593de93e984e378bea4814771e4465a3b/msys2-runtime/PKGBUILD#L153

PR #23205, opened 2025-01-27
[new package] kstatusnotifieritem

BLumia:
This is a tier 2 KDE Framework package/library, which is required to build some KDE applications (NeoChat for example).

Note: GPG key is not tested since I have no idea about how to test it properly.


Biswa96:
> Note: GPG key is not tested since I have no idea about how to test it properly.

```sh
cd mingw-w64-the-package-name
../.ci/fetch-validpgpkeys.sh
makepkg-mingw
```

PR #23208, opened 2025-01-27
[new-package] premake3 3.7

llm96:
Needed this for another package I'm hoping to contribute after this one

Used the [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/premake3/-/blob/main/PKGBUILD) package & the [existing](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-premake4/PKGBUILD) `premake4` package as a reference

### DIFF
--- a/mingw-w64-premake3/PKGBUILD
+++ b/mingw-w64-premake3/PKGBUILD
@@ -1,0 +1,27 @@
+# Maintainer: llm96 <llm96@fervidex.net>
+
+_realname=premake
+pkgbase=mingw-w64-${_realname}3
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}3")
+pkgver=3.7
+pkgrel=1
+pkgdesc="A build configuration tool. Describe your build using Lua and generate \
+the project files for your specific toolset (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
Could you metion the project that need premake3 to build.
I searched for it in Linux distros, It only exists on Arch with no user.
I think the project you are going to package is most probably unmaintained/abandoned/outdated, why isn't using premake4/5?

llm96:
> Could you metion the project that need premake3 to build. I searched for it in Linux distros, It only exists on Arch with no user. I think the project you are going to package is most probably unmaintained/abandoned/outdated, why isn't using premake4/5?

It's needed to build [aacskeys](https://aur.archlinux.org/packages/aacskeys). It's definitely outdated and unmaintained but I think it's still useful.

lazka:
Does aacskeys need java? or is it optional?

This all is very ancient and without a real upstream. So while I don't mind in general, I don't think this is worth the effort/maintenance.

llm96:
It's possible to build it without Java. I've been using it in a personal repository with [this PKGBUILD](https://gist.github.com/llm96/dab61e183c125a9594a8684a20af663f), which is still in need of some cleaning up, but builds the standalone binary just fine.

I don't disagree though, it's all very messy. No worries if you'd rather close this.

Edit: aacskeys itself needs a bunch of patches which is a whole other mess, so I'll just close this for now.

PR #23203, opened 2025-01-26
hackrf: Initial package

antoniovazquezblanco:
A new package for interacting with an RF device called HackRF :)

PR #23213, opened 2025-01-28
SoapyHackRF: Initial package

antoniovazquezblanco:
:)


antoniovazquezblanco:
CI fails because it seems https://github.com/msys2/MINGW-packages/pull/23203 is still not available as a dependency.

Biswa96:
Is this similar as other soapy* packages? https://packages.msys2.org/search?t=pkg&q=soapy

antoniovazquezblanco:
The package named `soapysdr` is the base package wich is different.

This [soapyrtlsdr](https://packages.msys2.org/base/mingw-w64-soapyrtlsdr) package is an extension of the base package should be almost identical to the one in this PR.

Biswa96:
I've restarted the CI checks and it failed for another reason.

antoniovazquezblanco:
Yay! :)

PR #23249, opened 2025-01-31
rio: update to 0.2.5

podsvirov:


Biswa96:
Need `depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")` 
see https://github.com/msys2/MINGW-packages/pull/23249/checks#step:11:1013

podsvirov:
> Need `depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")`

Done.

PR #23257, opened 2025-02-01
fix: add `ca-certificates` for HTTPS(SSL) dependency

zent1n0:


zent1n0:
I'm wondering if it is possible to use shared package `ca-certificates` for SSL certs.
It runs successful when passing argument `--ca-certificate=/usr/ssl/cert.pem`, but configure with `--with-ca-bundle="/usr/ssl/cert.pem"` will result in `[ERROR] Failed to load trusted CA certificates from /usr/ssl/cert.pem. Cause: error:80000003:system library::No such process`. I believe it is related to file path convertion, as the configuration could compile and run on my ArchLinux machine.

> I'm wondering if it is possible to use shared package `ca-certificates` for SSL certs.

So I would appreciate it if anyone tells me about why `ca-certificates` locates in `/usr/` instead of `/etc/`, or it's possible to move these architecture-independent files into shared normal places `/etc/`.

lazka:
We need autoreconf for clang builds to work, the DLL is missing now with this.

zent1n0:
> We need autoreconf for clang builds to work, the DLL is missing now without it.

I removed autoreconf commands because it can't build on my Windows machine (nor could Github CI), which raise an error when checking gettext macro version.



> I'm not sure about the gettext error, needs investigating.

This error should be easily reproduced through an rebuild of the package, or more exactly, run `autoreconf -i` on any system not bundled with gettext 18.0, as version macro was defined with aria2 release 1.73.0.
On Arch there's no autoreconf preparation, so I just removed it to give a quick fix for CI. Maybe you can provide a better solution to fix the build issue.



> yes, but both cases are broken since you can't hardcode a Windows path in the build, users have MSYS2 installed in different locations.

Sorry but I don't agree with this. What I hardcoded at build time is a Cygwin path, which is relative to MSYS2 installation. The issue is that their build configure doesn't support Cygwin path, while runtime argument supports. This really confuses me when trying to share certificates from /usr.
BTW it's not the critical problem. The true issue is this package is **unable to be built** now.

lazka:
See #23332

> Sorry but I don't agree with this. What I hardcoded at build time is a Cygwin path, which is relative to MSYS2 installation. The issue is that their build configure doesn't support Cygwin path, while runtime argument supports.

runtime only works because cygwin auto-converts arguments to Windows paths (see https://www.msys2.org/docs/filesystem-paths/#automatic-unix-windows-path-conversion). aria2 is native so doesn't know anything about cygwin and cygwin paths.

zent1n0:
@lazka Thanks for the fix. One more question, is it possible to move certs into /etc so aria2 could find shared certs without further configuration?

> So I would appreciate it if anyone tells me about why `ca-certificates` locates in `/usr/` instead of `/etc/`, or it's possible to move these architecture-independent files into shared normal places `/etc/`.

lazka:
certs are under `$MINGW_PREFIX/etc` already, see https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-ca-certificates for example

zent1n0:
> certs are under `$MINGW_PREFIX/etc` already, see https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-ca-certificates for example

Yes. But I wonder why we are not sharing the package `ca-certificates` (instead of split package for every env), as this package should be platform independent.

lazka:
> Yes. But I wonder why we are not sharing the package `ca-certificates` (instead of split package for every env), as this package should be platform independent.

In theory yes, but we try to keep things separate, so that you could copy `$MINGW_PREFIX` to another machine, or a different path, and the contained things should still work. Which is useful for bundling up apps. Also making things relocatable is tricky as it is, and making the relocation code depend on a different prefix than where the software was installed to complicates things even more (currently it's just "<dll>/../etc", then it would be "<dll>/../../usr/etc", which would make things depend on the MSYS2 filesystem layout)

zent1n0:
I see your consideration 😄 btw let's move back to the commit itself. I guess this should fix the missing deps.

PR #23280, opened 2025-02-04
libressl: update to 4.0.0

omikrun:
I want to update libressl package in MSYS2 with the following changes.

1. Port to CMake to avoid install failure in man pages
2. Rename libraries prefixed with libre rather than liblibre (both shared and static)
3. Solve relocation issues (following `mingw-w64-openssl` package)
4. Rename man pages with `libressl-` following Arch Linux
5. Change the config dir following Arch Linux

In LibreSSL 3.1.1-4:

```
$ libressl req -new -nodes -batch -keyout test.key -out test.csr
8712:error:02FFF00D:system library:func(4095):Permission denied:../../libressl-3.1.1/crypto/bio/bss_file.c:122:fopen('D:/a/msys64/clang64/share/libressl/openssl.cnf', 'rb')
8712:error:20FFF002:BIO routines:CRYPTO_internal:system lib:../../libressl-3.1.1/crypto/bio/bss_file.c:127:
8712:error:0EFFF002:configuration file routines:CRYPTO_internal:system lib:../../libressl-3.1.1/crypto/conf/conf_def.c:202:
```

It's obvious that original libressl finds the wrong location.

In LibreSSL 4.0.0-1:

```
$ libressl-openssl req -new -nodes -batch -keyout test.key -out test.csr
Generating a 1234 bit RSA private key
.........................................
..........
writing new private key to 'test.key'
-----
```

Now, after editing `/clang64/etc/libressl/openssl.cnf` with some information (like `countryName`, `stateOrProvinceName`) In particular, I changed `default_bits` to 1234 to indicate my patch is correct.

However, CI failed because `rename` command in `util-linux` that I used is buggy now (it fixed in upstream). Could anyone please update the `util-linux` package?


ognevny:
may I ask you again: is there any issue with updated package? if not, I can merge it

omikrun:
> may I ask you again: is there any issue with updated package? if not, I can merge it

No, it works fine. :) Feel free to merge.

Oh, I've forgot to mention that it brokes the backwards compatibility. But there is no package depend on it, so I think it's ok.

Issue #10312, opened 2021-12-10
Latest ncurses update broke linking using pkg-config


shermp:
The latest ncurses update (#10285) seems to have broken linking when using pkg-config. 

Minimal test program (from https://tldp.org/HOWTO/NCURSES-Programming-HOWTO/helloworld.html):

```c
#include <ncurses.h>

int main()
{	
	initscr();			/* Start curses mode 		  */
	printw("Hello World !!!");	/* Print Hello World		  */
	refresh();			/* Print it on to the real screen */
	getch();			/* Wait for user input */
	endwin();			/* End curses mode		  */

	return 0;
}
```

Compile with:
```sh
gcc $(pkg-config ncursesw --cflags) ncurses.c -o ncurses-test $(pkg-config ncursesw --libs)
```

Results in the following output
```
$ gcc $(pkg-config ncursesw --cflags) ncurses.c -o ncurses-test $(pkg-config ncursesw --libs)
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\msys64\tmp\ccCrSAYo.o:ncurses.c:(.text+0x10): undefined reference to `__imp_initscr'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\msys64\tmp\ccCrSAYo.o:ncurses.c:(.text+0x23): undefined reference to `__imp_printw'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\msys64\tmp\ccCrSAYo.o:ncurses.c:(.text+0x2c): undefined reference to `__imp_stdscr'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\msys64\tmp\ccCrSAYo.o:ncurses.c:(.text+0x39): undefined reference to `__imp_wrefresh'
[...]

PR #23283, opened 2025-02-04
update to ngspice 45.2

Kreijstal:
I would like assistance updating ngspice to 44.2 as you can see the .dll is missing, there are some libtool warnings, not sure what I did wrong :)


Kreijstal:
hmm it seems on CI it compiles fine

mingw-w64-ngspice/PKGBUILD
```diff
@@ -70,7 +76,7 @@ build() {
     --host=${MINGW_CHOST} \
     "${_common_flags[@]}" \
     --with-ngshared
-  make
+  make -j
```
lazka, line 79: why? it's passed automatically

mingw-w64-ngspice/curses.patch
```diff
@@ -0,0 +1,40 @@
+diff --git a/configure.ac b/configure.ac
+index f97a93ea6..232fa7ac1 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -1280,6 +1280,20 @@ else
+                 AC_CHECK_LIB([readline], [readline],
+                     [LIBS="$LIBS -lreadline"],
+                     [AC_MSG_ERROR([Couldn't find readline libraries.])])
++
++PKG_CHECK_MODULES_STATIC([NCURSES], [ncursesw], [
```
lazka, line 10: why is curses needed? It's not linking against ist.

mingw-w64-ngspice/curses.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/src/xspice/verilog/Makefile.am b/src/xspice/verilog/Makefile.am
+index 262d63384..1a298cff0 100644
+--- a/src/xspice/verilog/Makefile.am
++++ b/src/xspice/verilog/Makefile.am
+@@ -45,7 +45,9 @@
+ install-exec-hook:
+ 	cd $(DESTDIR)$(pkglibdir); \
+ 	rm -f ivlng*a libvvp* ; \
+-	mv ivlngvpi.* ivlng.vpi
++	if [ -f ivlngvpi.* ]; then \
```
lazka, line 35: Do you know why this is needed?

mingw-w64-ngspice/curses.patch
Kreijstal: only statically, it does get used. 
https://github.com/imr/ngspice/blob/e011d14f6fd9595723b2acd6e21324826f25aef5/src/frontend/terminal.c#L26
because if you link dynamically we get another error, namely: 
https://github.com/msys2/MINGW-packages/issues/10312

mingw-w64-ngspice/PKGBUILD
Kreijstal: when testing it, it didn't seem it was getting passed, or maybe it was a placebo effect, but I thought -j made things faster.

mingw-w64-ngspice/curses.patch
Kreijstal: We compile 3 times, correct, when doing the nonshared version (ngspice_con), ivlngvpi doesn't get generated, probably libtool shenanigans, 
https://sourceforge.net/p/ngspice/bugs/735/

mingw-w64-ngspice/curses.patch
Kreijstal: ahhh you mean curses should not be runtime dependency, yeah,I think you're right, sorry about that.

mingw-w64-ngspice/PKGBUILD
Kreijstal: I've tested it on my laptop, and I'm pretty sure the makefiles that ngspice seems configuring "-j" is not passed automatically, i mean unless makefile also listen to an env var im Not aware of? This is outside the PKGBUILD

Kreijstal:
I notice that after simply updating the branch suddenly it cannot be built with clang, while obviously, it in the past could, is this a clang bug? or a ngspice regression or more likely clang doesnt accept this anymore?

Kreijstal:
@lazka managed to make it buildable with 45.2 and clang again..

PR #23334, opened 2025-02-10
Groonga: Update to 15.0.0

otegami:
Groonga version15.0.0 has just released [here](https://github.com/groonga/groonga/releases/tag/v15.0.0).


otegami:
Thanks you so much every time!

Issue #22809, opened 2024-12-15
gtk4: Emoji rendering broken


cal0pteryx:
### Description / Steps to reproduce the issue

* install [mingw-w64-gtk4](https://packages.msys2.org/base/mingw-w64-gtk4) (current version: 4.16.7-1)
* start gtk4-demo.exe
* run any demo where it's possible to add emojis, e.g. with a GtkEntry, right click, insert emoji

https://gitlab.gnome.org/GNOME/gtk/-/issues/7144 has more info on this and it seems that for MSYS, some patches are applied which break emoji rendering on MSYS.

### Expected behavior

Emojis should be rendered correctly

### Actual behavior

Emojis are not rendered correctly:

![image](https://github.com/user-attachments/assets/0f879e82-7c8c-4309-abc8-b00ea3a1575c)

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-26100

Biswa96:
gtk4 has been updated to 4.16.12 version. Does the issue still remain?

lovetox:
i dont know why it should, the linked issue is open, and the mingw package still aplies a font fix `[001-fix-font-rendering.patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gtk4/001-fix-font-rendering.patch)` which upstream said to remove

lovetox:
@lazka you were involved when the dwrite patch was added.

It seems the dwrite disable patch was added because of https://gitlab.gnome.org/GNOME/pango/-/issues/767

In this issue there is now a user that says that the [001-fix-font-rendering.patch](https://github.com/msys2/MINGW-packages/blob/7de2c7c49aa8feb1f6e4a2bf8c99bf98bd8e50f0/mingw-w64-gtk4/001-fix-font-rendering.patch) in Gtk should be enough to solve the issue, so maybe we can drop this patch in pango now?

lazka:
> In this issue there is now a user that says that the [001-fix-font-rendering.patch](https://github.com/msys2/MINGW-packages/blob/7de2c7c49aa8feb1f6e4a2bf8c99bf98bd8e50f0/mingw-w64-gtk4/001-fix-font-rendering.patch) in Gtk should be enough to solve the issue, so maybe we can drop this patch in pango now?

[...]

PR #23337, opened 2025-02-10
Draft: Cairo: Add patch for DWrite fonts

lb90:
Still a draft because I want to add another enhancement...

Fixes https://github.com/msys2/MINGW-packages/issues/22809


cal0pteryx:
I see https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/602 has just been merged :tada:

lb90:
Ok, should be ready to go now!

lazka:
Thanks, this looks kinda acceptably now.

Is it expected that there is less/worse font hinting with dwrite?

lazka:
hm, it's worse with gtk4. I let you decide :)

old:
gtk4_old

new:
gtk4_new

cal0pteryx:
Even if font rendering is slightly degraded (which may always be improved in later Cairo versions), this new release _fixes_ rendering Emojis, which wouldn't render at all before.

g-willems:
The visual regressions are due to gtk enabling an "automatic" font rendering mode, with some tweaks that are better suitable for the fontconfig renderer.

It's easy to check: open the inspector > Global > Settings , and set Rendering to "manual" to use the original properties that are set from `gdk_win32_display_get_setting()`.

Maybe we could patch gdk-win32 to enforce `gtk-font-rendering = manual`?

lb90:
Hi everyone!

https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/602 was meant to align the DWrite font backend more closely to the FT2 font backend. For example, if you run GTK applications with `PANGOCAIRO_BACKEND=fc` (FontConfig + FreeType2), you shall see an output that's very similar to DWrite (you have to use Pango from this PR, of course!).

However, that's not quite integrated with the way text is rendered on Windows. Most of that is due to some limitations in Cairo / Pixman, see the links.

Anyway I can update this PR to make DWrite rendering resemble GDI rendering more closely...

Stay tuned! 😎 

References:
1. https://lists.nongnu.org/archive/html/freetype-devel/2013-10/msg00056.html
2. https://bugs.freedesktop.org/show_bug.cgi?id=28549
3. https://gitlab.freedesktop.org/cairo/cairo/-/issues/2

lovetox:
@lb90 Any time yet to update this MR? thanks

cal0pteryx:
@lazka @lb90 can this get merged one way or the other? #22809 is now open >4 months and we'd really like to release Windows versions of our software again.

lb90:
Rebased on top master, now I'm going to add the actual Cairo patch

kleisauke:
> Maybe we could patch gdk-win32 to enforce `gtk-font-rendering = manual`?

AFAIK (and correct me if I wrong), this is already handled by [this patch](https://github.com/msys2/MINGW-packages/blob/7516af05679bc9be51170ff81b061f02426fec27/mingw-w64-gtk4/001-fix-font-rendering.patch) and the revert of commit https://gitlab.gnome.org/GNOME/gtk/-/commit/b569470b87d143d30c2388046c9a6450696f19b8.

@lb90 Do you think we can drop these GTK patches now?:
https://github.com/msys2/MINGW-packages/blob/7516af05679bc9be51170ff81b061f02426fec27/mingw-w64-gtk4/PKGBUILD#L51-L56

lb90:
Ok, this is now ready for review!

cal0pteryx:
Who would be able to do a review of this PR? Sorry for bumping, but I'd really like to have emojis working again on Windows :smiling_face_with_tear:

lazka:
I've updated the gist: https://gist.github.com/lazka/c426a63282d3c575fda6ee28c694ea03

it's not much different compared to the last version

cal0pteryx:
Would it be an option to merge this in its current state, and further improve the situation afterwards?

lovetox:
I see a very minimal regression from old to current for normal text, but at the same time if emoji support now works, this would be a huge plus. I dont think the regression makes it unusable, but no emoji support is a big problem which infact does not allow us to make new releases at all.

lazka:
ok, let's merge it then.

lazka:
If someone ends up here, `export PANGOCAIRO_BACKEND=fc` avoids the too light font rendering with gtk3 for now. It also gets rid of the rainbow colors with dark mode.

lazka:
I wonder if it could be somehow disabled for gtk3 only.

PR #23340, opened 2025-02-10
gimp: Update to 3.0.0-RC3

hmartinez82:


mingw-w64-gimp/0003-dont-install-external-files.patch
hmartinez82, line 1: This is already in the source code where both flags default to `false`. https://gitlab.gnome.org/GNOME/gimp/-/blob/GIMP_3_0_0_RC3/meson.build?ref_type=tags#L2133

PR #23360, opened 2025-02-11
crystal: Support non-bootstrapped builds

HertzDevil:
The bootstrap will be back again when we have a stage 0 compiler for ARM64 Windows.

### DIFF
--- a/mingw-w64-crystal/PKGBUILD
+++ b/mingw-w64-crystal/PKGBUILD
@@ -1,10 +1,11 @@
 # Maintainer: Quinton Miller <nicetas.c@gmail.com>
 
+_bootstrap=0
 _realname=crystal
 pkgbase="mingw-w64-${_realname}"
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.15.1
-pkgrel=1
+pkgrel=2
 pkgdesc="Fast and statically typed, compiled language with Ruby-like syntax (mingw-w64)"
 arch=('any')
 mingw_arch=('ucrt64' 'mingw64' 'clang64')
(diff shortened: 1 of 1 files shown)


mingw-w64-crystal/PKGBUILD
```diff
@@ -31,16 +32,19 @@ depends=(
 )
 makedepends=(
   "git"
+  $( (( !_bootstrap )) && echo "${MINGW_PACKAGE_PREFIX}-crystal")
```
ognevny, line 35: nit: we usually do something like
```suggestion
  $( (( _bootstrap )) || echo "${MINGW_PACKAGE_PREFIX}-crystal")
```

mingw-w64-crystal/PKGBUILD
```diff
@@ -31,16 +32,19 @@ depends=(
...
   "${MINGW_PACKAGE_PREFIX}-lld" # needed for linking std_spec
 )
 source=(
   "$pkgname-$pkgver::git+${msys2_repository_url}.git#tag=${pkgver}"
-  https://github.com/crystal-lang/crystal/releases/download/1.15.0/crystal-1.15.0-windows-x86_64-gnu-unsupported.zip # stage 0 compiler
+  $( (( _bootstrap )) && echo "https://github.com/crystal-lang/crystal/releases/download/1.15.0/crystal-1.15.0-windows-x86_64-gnu-unsupported.zip" ) # stage 0 compiler
+)
+sha256sums=(
+  '626fed60399d7f1dc83e2e5a2ecde20d235e19a3dcf6d70d6f52ec4de063b32b'
+  $( (( _bootstrap )) && echo '7d7df1b4a99cb3f938106ba45e0b55ce1d9913491999f1acaa57da168a18becb' )
```
ognevny, line 46: please, rewrite this to use
```bash
if (( _bootstrap )); then
  sha256sums+=('7d7df1b4a99cb3f938106ba45e0b55ce1d9913491999f1acaa57da168a18becb')
fi
```

ognevny: the same for source field. [example](https://github.com/msys2/MINGW-packages/blob/77bbfafbd4a7ddde2903f589f968dd1f88ad9cdd/mingw-w64-rust/PKGBUILD#L73)

ognevny:
there is a [llvm-libs](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-llvm-libs) which provide DLL for LLVM. is it possible to add `llvm` to `makedepends` and `llvm-libs` to `depends`?

Issue #23316, opened 2025-02-07
[libimagequant] Segmentation fault


radarhere:
### Description / Steps to reproduce the issue

Over at https://github.com/python-pillow/Pillow, our MinGW GitHub Actions job recently showed a segfault when libimagequant is used. The job passed with mingw-w64-x86_64-libimagequant-4.3.3-1, but started failing when mingw-w64-x86_64-libimagequant-4.3.3-2 was released. That change came from https://github.com/msys2/MINGW-packages/commit/2d14aef55bc34f2855cb417e34bffb5aea5bcac4, so perhaps this is related to cargo-c?

For good measure, I updated to libimagequant 4.3.4 in https://github.com/msys2/MINGW-packages/pull/23245, but the problem continued to occur.

I've created a minimal reproduction at https://github.com/radarhere/mingw_libimagequant. The segfault occurs in https://github.com/radarhere/mingw_libimagequant/actions/runs/13198684016/job/36845578278, but [without the call](https://github.com/radarhere/mingw_libimagequant/commit/c5ba9e4a16122304645b1d20493c0fccb6787288) to libimagequant's `liq_version()`, it passes in https://github.com/radarhere/mingw_libimagequant/actions/runs/13198259889

### Expected behavior

No segmentation fault.

### Actual behavior

Segmentation fault.

### Verification

- [x] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-20348

### MINGW environments affected

- [x] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANGARM64

lazka:
related #23358

lazka:
Upstream issue: https://github.com/lu-zero/cargo-c/issues/443

A workaround is now in the repo: #23364 

I'll keep this open for now, since the underlying issue is not fixed.

kleisauke:
For triaging purposes, does this only occur with libimagequant? rav1e is also built with cargo-c, which seems to work fine, at least with cargo-c 0.10.9 (see #23165).

lazka:
@kleisauke I've extended the reproducer with another project that also crashes. Running avifdec in MSYS2 also segfaults in rav1e_version_full(), so rav1e is also affected.

PR #23365, opened 2025-02-12
libimagequant: use new cargo-c meson-paths option

lazka:
So we can get rid of patching this in cargo-c (https://github.com/msys2/MINGW-packages/blob/3f0d0ff907b9b967fa612b9106b8e63b73829e2d/mingw-w64-cargo-c/001-add-lib-prefix-on-windows-gnu.patch)
Results in imagequant.dll -> libimagequant.dll, so rebuild everything


ognevny:
I can drop patch for cargo-c, rebuild packages. maybe create template for cargo-c packages too?

lazka:
Sure, but be aware that anything re-linking to cargo-c build libraries might be broken due to #23316

ognevny:
> I can drop patch for cargo-c and rebuild packages. maybe create template for cargo-c packages too?

...or patch this line instead https://github.com/lu-zero/cargo-c/blob/master/src%2Fcli.rs#L56

ognevny:
> Sure, sounds good, but be aware that anything re-linking to cargo-c built libraries might be broken at runtime due to #23316.

we've already rebuilt all rdeps once, so it shouldn't change anything I guess

lazka:
> ...or patch this line instead https://github.com/lu-zero/cargo-c/blob/master/src%2Fcli.rs#L56

That means all DLLs will get a "lib" prefix on the next rebuild.

> we've already rebuilt all rdeps once, so it shouldn't change anything I guess

It's about rdeps of libraries using cargo-c, not the libraries themselves. So if you rebuild libplacebo now it will likely be broken.

lazka:
Confirmed that all packages are broken now. I'll extend the workaround applied in #23364 to all cargo-c packages, then we can do the rebuild.

PR #23381, opened 2025-02-12
cargo-c: remove patch to force lib prefix for import libs

lazka:
Instead users should pass --meson-paths to cargo-c commands to get the meson compatible naming.

Slightly inconvenient, but less magic, and we behave like upstream again.


ognevny:
thanks

PR #23391, opened 2025-02-14
zed: update to 0.173.10

ognevny:


ognevny:
now pthread functions are undefined somehow

PR #23395, opened 2025-02-14
plasma: update to 6.3.0

Biswa96:


Biswa96:
I can not figure out the qt6 cmake issue. Please feel free to have a look into it.

hmartinez82:
This is happening (See inline comments): 
I'm not very versed with CMake. I don't know why `Qt6Core_FOUND` would go "out of scope".
```
    )

    unset(QUERY_EXECUTABLE CACHE)
endfunction()   < ----  #1 Once we get here, Qt6Core_FOUND is set to 1, but this line clears it
if(BUILD_QT6)
    build_Qt6()
endif()

function(build_colors_and_cursors)
    # While it is technically not 100% correct to use versionless
    # KDEInstallDirs while (maybe) building for both Qt versions,
    # it doesn't really matter at the end, because colors and cursors
    # only use ICONDIR and DATADIR which both do not depend on the version
    include(KDEInstallDirs)
    add_subdirectory(colors)
    add_subdirectory(cursors) < --- #3 This looks for Qt6:Core again, with an unset `Qt6Core_FOUND` which in turn tries get to that point where Qt6EntryPointPrivateConfig.cmake is called again.
endfunction()

build_colors_and_cursors() <----- #2 this is called 
```

hmartinez82:
@Biswa96 Give it a try https://github.com/Biswa96/MINGW-packages/pull/1

Biswa96:
Thanks for the help. Please let me ask upstream about this issue before adding any change.

hmartinez82:
I filed this upstream https://bugs.kde.org/show_bug.cgi?id=500106

hmartinez82:
Also, I just realized, by the 4 line comment, that those icons and cursors seem to be intended to be shared across Qt5 and Qt6 builds of Breeze. So maybe we need to move them to their own package? 

The double resolution of Qt6::Core seems to be affect only Qt6 builds because the `cursors` subdir require compiling a tool at build time as seen at https://invent.kde.org/plasma/breeze/-/blob/master/cursors/src/CMakeLists.txt

PR #23410, opened 2025-02-16
python-crashtest: Remove unmaintained package

Biswa96:


lazka:
lgtm

needs to be removed from poetry as well

PR #23419, opened 2025-02-17
python-geoalchemy2: support spatialite databases

podsvirov:
Patch for auto resolve path to spatialite module (currently [path should be set manually](https://geoalchemy-2.readthedocs.io/en/latest/spatialite_dialect.html)).

I suggest checking the patch here, then suggest it to the upstream repository.


Biswa96:
MINGW_PREFIX environment variable will not be set if msys2 is used by appending PATH e.g. `set PATH=C:\msys64\ucrt64\bin;%PATH%`.

Biswa96:
This reminds of a python thing which allows/disallows loading DLLs from /bin. I can not remember the exact environment variable. lazka may provide some info.

PR #23434, opened 2025-02-18
arrow python-pyarrow: Update to 19.0.1

amoeba:


MehdiChinoune:
Rebuilds are only required with ABI break

amoeba:
Thanks for the quick review @MehdiChinoune. I removed the rebuilds. Can you take a look again?

amoeba:
Thanks!

PR #23437, opened 2025-02-19
new package: python-hererocks

tobil4sk:
Based on: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hererocks and existing mingw python packages

### DIFF
--- a/mingw-w64-python-hererocks/PKGBUILD
+++ b/mingw-w64-python-hererocks/PKGBUILD
@@ -1,0 +1,38 @@
+# Contributor: Tobi Laskowski <tobil4sk@outlook.com>
+
+_realname=hererocks
+pkgbase=mingw-w64-python-${_realname}
+pkgname=${MINGW_PACKAGE_PREFIX}-python-${_realname}
+pkgver=0.25.1
+pkgrel=0
+pkgdesc="Python script for installing Lua/LuaJIT and LuaRocks into a local directory (mingw-w64)"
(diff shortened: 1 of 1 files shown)


mingw-w64-python-hererocks/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/luarocks/hererocks'
+msys2_references=(
+  'purl: pkg:pypi/hererocks'
+)
+license=('spdx:MIT')
+depends=("git"
+         "${MINGW_PACKAGE_PREFIX}-python"
+         "unzip")
```
ognevny, line 18: if seems that unzip is not needed for the package

ognevny: see [requirements](https://github.com/luarocks/hererocks/tree/latest?tab=readme-ov-file#requirements) page from upstream READMW

tobil4sk: It doesn't seem needed for hererocks itself, however, hererocks is used to install luarocks/lua binaries in a local directory and unzip is a requirement for luarocks: https://github.com/luarocks/luarocks/wiki/Installation-instructions-for-Unix#1ubuntudebian-user

PR #23447, opened 2025-02-20
Groonga: Update to 15.0.1

otegami:
Groonga version15.0.1 has just released [here](https://github.com/groonga/groonga/releases/tag/v15.0.1).


otegami:
Thank you so much!

PR #23452, opened 2025-02-20
[new-package] glaze 4.4.3

llm96:


mingw-w64-glaze/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    cmake \
+      -GNinja \
+      -DCMAKE_INSTALL_PREFIX="${MINGW_PREFIX}" \
+      "${extra_config[@]}" \
+      -DBUILD_{SHARED,STATIC}_LIBS=ON \
```
Biswa96, line 35:
```suggestion
```

Those options are not required here because this is just header only library.

PR #23463, opened 2025-02-21
[new-package] python-kbinxml 2.1

llm96:


lazka:
please no new packages using setup.py, use build/installer instead: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.python.sh

PR #23468, opened 2025-02-22
cargo-c: update to 0.10.11

ognevny:
- remove backport
- remove workarounds for rdeps for future updates/rebuilds


ognevny:
there is a more recent version available, but lockfile is not published for it yet

ognevny:
stripping fails for some packages in CLANG64 environment
```
-> Stripping unneeded symbols from binaries and libraries...
D:/M/msys64/clang64/bin/strip.exe: error: unsupported object file format
```

PR #23472, opened 2025-02-22
cargo-leptos: update to 0.2.29

podsvirov:


ognevny:
have no idea actually. I won't be home soon, so can't reproduce now...

ognevny:
try to use clang/lld. gcc doesn't have good wasm support IIRC

podsvirov:
> try to use clang/lld. gcc doesn't have good wasm support IIRC

You right! All good with `CLANG64` builds. But how to tell `cargo` use custom linker? :thinking:

ognevny:
> > try to use clang/lld. gcc doesn't have good wasm support IIRC
> 
> You right! All good with `CLANG64` builds. But how to tell `cargo` use custom linker? :thinking:

`export RUSTFLAGS="-C linker=clang -C link-arg=-fuse-ld=lld"`. but try to use lld only first

podsvirov:
> export RUSTFLAGS="$RUSTFLAGS -C linker=clang -C link-arg=-fuse-ld=lld"

Thanks! Will try...

podsvirov:
Only `lld` fix work as expected:
```
export RUSTFLAGS="$RUSTFLAGS -C link-arg=-fuse-ld=lld"
```

PR #23490, opened 2025-02-23
silc-toolkit: Enable aarch64 builds

hmartinez82:


ognevny:
why bumping pkgrel? adding new env doesn't require rebuild for all packages

hmartinez82:
@ognevny because I had to change the `CFLAGS`

ognevny:
> @ognevny because I had to change the `CFLAGS` 

this doesn't require a rebuild. CFLAGS could be used only for clangarm64 for now

PR #23497, opened 2025-02-24
rustscan: update to 2.4.1

ognevny:


ognevny:
https://github.com/RustScan/RustScan/commit/b1f008a9c987f7e879ccf74bdab5ecfea9e74fb9 should be included in this tag, but it isn't

Issue #23261, opened 2025-02-01
neovim creates wrong directory for parsers


dominicattal:
### Description / Steps to reproduce the issue

After installing neovim using `pacman -S mingw-w64-ucrt-x86_64-neovim`, whenever I tried to open a help page with <code>:h</code>, I would get this error message:
```
Error detected while processing BufNewFile Autocommands for "*":
Error executing lua callback: ...pData\Local\msys2\ucrt64\share\nvim\runtime\filetype.lua:35: Error executing lua: ...pData\Local\msys2\ucrt64\share\nvim\runt
ime\filetype.lua:36: BufNewFile Autocommands for "*"..FileType Autocommands for "*"..function <SNR>1_LoadFTPlugin[20]..script C:\Users\attal\AppData\Local\msy
s2\ucrt64\share\nvim\runtime\ftplugin\lua.lua: Vim(runtime):E5113: Error while calling lua chunk: ...crt64\share\nvim\runtime/lua/vim/treesitter/language.lua:
107: no parser for 'lua' language, see :help treesitter-parsers
stack traceback:
        [C]: in function 'error'
        ...crt64\share\nvim\runtime/lua/vim/treesitter/language.lua:107: in function 'add'
        ...4\share\nvim\runtime/lua/vim/treesitter/languagetree.lua:111: in function 'new'
        ...l\msys2\ucrt64\share\nvim\runtime/lua/vim/treesitter.lua:41: in function '_create_parser'
        ...l\msys2\ucrt64\share\nvim\runtime/lua/vim/treesitter.lua:108: in function 'get_parser'
        ...l\msys2\ucrt64\share\nvim\runtime/lua/vim/treesitter.lua:416: in function 'start'
        ...a\Local\msys2\ucrt64\share\nvim\runtime\ftplugin\lua.lua:2: in main chunk
        [C]: in function 'nvim_cmd'
        ...pData\Local\msys2\ucrt64\share\nvim\runtime\filetype.lua:36: in function <...pData\Local\msys2\ucrt64\share\nvim\runtime\filetype.lua:35>
        [C]: in function 'nvim_buf_call'
        ...pData\Local\msys2\ucrt64\share\nvim\runtime\filetype.lua:35: in function <...pData\Local\msys2\ucrt64\share\nvim\runtime\filetype.lua:10>
stack traceback:
        [C]: in function 'nvim_cmd'
[...]

PR #23499, opened 2025-02-24
[new-packages] tree-sitter-c, tree-sitter-lua, tree-sitter-markdown, tree-sitter-query, tree-sitter-vim, tree-sitter-vimdoc

calebabutler:
These packages are dependencies for newer versions of neovim. Currently neovim gives an error when opening `*.lua`, `*.c`, `*.md`, `*.vim`, and when opening vim documentation since the program lacks these dependencies.

Proof they are now dependencies [here](https://github.com/neovim/neovim/blob/master/BUILD.md#how-to-build-without-bundled-dependencies) and [here](https://archlinux.org/packages/extra/x86_64/neovim/).


calebabutler:
https://github.com/msys2/MINGW-packages/issues/23261

calebabutler:
@Biswa96 neovim maintainer

mingw-w64-neovim/PKGBUILD
```diff
@@ -93,4 +99,7 @@ package() {
   DESTDIR="${pkgdir}" "${MINGW_PREFIX}"/bin/cmake.exe --install .
 
   install -Dm644 "${srcdir}/${_realname}-${pkgver}/LICENSE.txt" "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/LICENSE.txt"
+
+  # Symlink tree-sitter grammars directory to nvim runtime path
+  ln -s "${MINGW_PREFIX}/lib/tree_sitter" "${pkgdir}${MINGW_PREFIX}/share/nvim/runtime/parser"
```
ognevny, line 104: symlink won't work as you expect: https://www.msys2.org/docs/symlinks/

calebabutler: You are correct; this cannot be the best solution as it inadvertently bundles the `tree-sitter-*` so files with neovim. I changed it so that the `tree-sitter-*` packages add to the neovim runtime path directly in their respective packages.

ognevny:
I believe you've done a good work on testing it

I rebased some commits for cleaner history. removed `make` entries in makedepends as it's always pulled with `base-devel`.

thanks

PR #23507, opened 2025-02-25
kicad: update to 9.0.0

Kreijstal:
it has some breaking changes, first it depends on nng which only builds on ucrt.
clang64 breaks as well. I'd like assistance with it 

<details><summary>Details</summary>
<p>

```
  In file included from C:/_/B/src/kicad-9.0.0/common/settings/bom_settings.cpp:23:
  C:/_/B/src/kicad-9.0.0/libs/core/include/core/json_serializers.h:55:8: error: explicit specialization of 'nlohmann::adl_serializer<wxString>' after instantiation
     55 | struct adl_serializer<wxString>
        |        ^~~~~~~~~~~~~~~~~~~~~~~~
  C:/_/B/src/kicad-9.0.0/thirdparty/nlohmann_json/nlohmann/json.hpp:3520:37: note: implicit instantiation first required here
   3520 | using from_json_function = decltype(T::from_json(std::declval<Args>()...));
  
        |                                     ^
  
  1 error generated.
```

This reminds me, maybe we can unvendor nhlohman json, not sure if it will solve the problem.

</p>
</details>


Kreijstal:
while this might build, it doesn't run. I am out of ideas.

Kreijstal:
weirdly enough clang64 version does run ucrt64 doesn't

Kreijstal:
in another computer it seems kicad ucrt64 just works?

Kreijstal:
grafik

I think there might me a problem when kicad encounters old configuration, user might have to delete previous user configuration for kicad to start properly. it seems everything works fine...

Kreijstal:
I wonder if we should build with caching since it takes 1 hour to build, so if we have to build twice..

Kreijstal:
clangarm64 added

MehdiChinoune:
I pushed it with some changes.

PR #23510, opened 2025-02-25
kicad-[doc,library]: Update to 9.0.0

hmartinez82:
Requires https://github.com/msys2/MINGW-packages/pull/23507 to be built first


MehdiChinoune:
pushed.

PR #23515, opened 2025-02-25
blis: Enable aarch64 builds

hmartinez82:


MehdiChinoune:
Isn't possible to fix building armsve on Windows?

hmartinez82:
@MehdiChinoune Yes. It's going to be a lot of code, but I'll file a PR with upstream first. 
SVE is somewhat new, and https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent doesn't support it yet. I'll have to read the register and hunt the flag myself, that's why I want to push this to upstream first.

hmartinez82:
I stand corrected. I guess MSDN is not up to date 🤣
https://gitlab.winehq.org/toadking/wine/-/blob/24694dabfc53fc850cf035e187a4c15b77d6bc77/include/winnt.h#L1101

Issue #22559, opened 2024-11-16
Wildcard removal and gcc


lazka:
There was someone in chat reporting that the wildcard removal broke something which is also referenced in some MS docs: https://code.visualstudio.com/docs/cpp/config-mingw#_modifying-tasksjson and I don't know if there is a good workaround currently.

It seems that cl.exe for example supports them, and since gcc has an official option to enable this at gcc build time I wonder if we should enable it. It's unlikely there is untrusted things being passed to build tools, and it's a common thing to deal with multiple files there.

for the gcc option: https://github.com/gcc-mirror/gcc/blob/b085fc9965b7d1e99725892587682822068f2baa/gcc/configure.ac#L2262-L2270

Kreijstal:
shouldn't it be a commandline on gcc at runtime not at compile time?

lazka:
#23517 

* There were more wishes for this in chat.
* llvm/clang does this too, though they implement their own wildcard handling

lazka:
> shouldn't it be a commandline on gcc at runtime not at compile time?

Not sure what you mean? Like have a "--wildcard" at runtime that enables this? If that, then in theory, yes, but I doubt that anyone would find that feature then.

PR #23517, opened 2025-02-26
mingw-w64-gcc: Allow the gcc driver to expand wildcards

Peter0x44:
https://github.com/msys2/MINGW-packages/issues/22559

https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/driver-mingw32.cc

This options sets _dowildcard to 1 which makes mingw-w64 expand the wildcard somehow.


lazka:
thanks

PR #23526, opened 2025-02-26
[gnuradio] - upgrade to new version and fix check() environment

chrisjohgorman:
- upgrade to version 3.10.12.0
  - remove references to git patches that have become obsolete.
  - remove file and reference for the numpy-2.0.patch which is no longer needed.

- in check()
  - check was non functional before without an installed instance of gnuradio.
  - made changes so the check will run even when a gnuradio instance is not installed.
  - copy all dlls to a directory and add the directory to the path
  - add python call to os.add_dll_directory to all (needed) test scripts.


chrisjohgorman:
Tested mingw64, ucrt64 and clang64 artifacts and they all run fine on my machine.

PR #23543, opened 2025-02-28
innoextract: update to the most recent version ("1.10-dev")

dscho:
There have been 82 revisions since version 1.9 was tagged, over 1,600 days passed, and https://github.com/dscharrer/innoextract/issues/176 does not exactly suggest that any new version will be tagged any time soon.

Yet version 1.9 is incompatible with many an InnoSetup version, while the code actually has been updated to handle the most recent one in https://github.com/dscharrer/innoextract/commit/e58f295d80c3.

So let's switch to the most recent development version (according to https://github.com/dscharrer/innoextract/blob/HEAD/VERSION it is "1.10-dev").

I verified manually that this version of `innoextract` can handle the recent Git for Windows installers, for example, while the most recent one in MSYS2 cannot.

### DIFF
--- a/mingw-w64-innoextract/PKGBUILD
+++ b/mingw-w64-innoextract/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=innoextract
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.9
-pkgrel=10
+pkgver=1.10dev.82.6e9e34e
+pkgrel=1
 pkgdesc="A tool to extract installers created by Inno Setup (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
```
$ vercmp.exe 1.10.dev 1.10
1
```
1.10.dev is considered newer than 1.10, please change it to 1.10dev
```
$ vercmp.exe 1.10dev 1.10
-1
```

MehdiChinoune:
We use `pkgver` function when package is built from a specific commit.

dscho:
> 1.10.dev is considered newer than 1.10, please change it to 1.10dev

Makes sense to me.

> We use `pkgver()` function when package is built from a specific commit.

I imitated [`mingw-w64-ada-libfswatch`'s `pkgver()` function](https://github.com/msys2/MINGW-packages/blob/165ac9302a67db484e346ab2c7a7c81f05d5924a/mingw-w64-ada-libfswatch/PKGBUILD#L42-L45) (it was one of the first hits of `git grep  '^pkgver *[^=]' \*/PKGBUILD`).

At the same time, I realized that the `prepare()` function was still left as a semi-no-op, and dropped it.

dscho:
Thank you @MehdiChinoune!

PR #23548, opened 2025-02-28
libuv: backport a fix for Windows 1607.

### LINKED ISSUE #23531: mingw-w64-x86_64-cmake fails to start

1. all this is done in an `MINGW2 MinGW 64-bit` shell
2. Update `mingw64` to the latest and greatest
    ```sh
    pacman -Suy
    ```
   (on my system this currently updates to `filesystem 2025.02-23.1`)
3. install `mingw-w64-x86_64-cmake`
    ```sh
    pacman -S "${MINGW_PACKAGE_PREFIX}-cmake"
    ```
4. run cmake
    ```sh
    $ cmake || echo $?
    127
    ```

### Expected behavior

i expect to be greeted with a the Cmake banner:

```
Usage

  cmake [options] <path-to-source>
  cmake [options] <path-to-existing-build>
  cmake [options] -S <path-to-source> -B <path-to-build>

Specify a source directory to (re-)generate a build system for it in the
current working directory.  Specify an existing build directory to
re-generate its build system.

Run 'cmake --help' for more information.


```

### Actual behavior

nothing is printed, and instead i get an error code `127` (afair this is "file not found").

it took me a while to narrow this down, as I already had the `cmake` package (without the `${MINGW_PACKAGE_PREFIX}-`) installed, and ran `cmake` before updating.
this would then cache the path to the `cmake` binary as `/usr/bin/cmake`, and subsequent runs would use the binary from the `cmake`-package, which *is* working correctly.
starting up a new shell, would clear the cache (so `cmake` is now resolved to `${MINGW_PREFIX}/bin/cmake` - e.g. `/mingw64/bin/cmake`), and things start to fail.

### Verification

### Windows Version

MSYS_NT-10.0-14393

- [x] MINGW64
- [x] MINGW32
- [x] UCRT64
- [x] CLANG64
- [ ] CLANGARM64

if I new how to fix the problem, I probably would submit a PR (but I don't)

MehdiChinoune:
Fixes #23531

### DIFF
--- a/mingw-w64-libuv/.gitignore
+++ b/mingw-w64-libuv/.gitignore
@@ -1,0 +1,2 @@
+*.sign
+001-windows-1607.patch
--- a/mingw-w64-libuv/PKGBUILD
+++ b/mingw-w64-libuv/PKGBUILD
@@ -5,31 +5,37 @@ _realname=libuv
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=1.50.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Cross-platform asynchronous I/O (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://libuv.org/'
+msys2_repository_url='https://github.com/libuv/libuv/'
+msys2_documentation_url='https://docs.libuv.org/'
 msys2_references=(
   'archlinux: libuv'
   "cpe: cpe:/a:libuv:libuv"
   "cpe: cpe:/a:libuv_project:libuv"
 )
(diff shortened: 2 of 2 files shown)


MehdiChinoune:
@umlaeute could you try the artifacts from build jobs and see if they fix the issue.

umlaeute:
after installing the mingw64 package with `pacman -U ./mingw-w64-x86_64-libuv-1.50.0-2-any.pkg.tar.zst`, I now get:

```
# /mingw64/bin/cmake.exe 
Usage

  cmake [options] <path-to-source>
  cmake [options] <path-to-existing-build>
  cmake [options] -S <path-to-source> -B <path-to-build>

Specify a source directory to (re-)generate a build system for it in the
current working directory.  Specify an existing build directory to
re-generate its build system.

Run 'cmake --help' for more information.

#
```

so yes, this seems to fix the problem indeed! :tada:

PR #23561, opened 2025-03-03
drop 32-bit for some librsvg rdeps

ognevny:
they're not used in gimp2


ognevny:
who starts CI, huh? no need to run CI for such changes :)

ognevny:
the process of `i686-pc-windows-gnu` target to tier 2 [has begun](https://github.com/rust-lang/rfcs/pull/3771#issuecomment-2718863911), so let's start with it

PR #23563, opened 2025-03-03
flang: inherit target specific code for BIND(C) on Windows

mmuetzel:
Prompted by @MehdiChinoune's comment in https://github.com/llvm/llvm-project/issues/114035#issuecomment-2693354786.

I tried building the example in the original comment in that issue. It is executing fine for me with the changes from here.

@MehdiChinoune: Did you come across other projects that failed to compile with that error?

### DIFF
--- a/mingw-w64-flang/0003-flang-Inherit-target-specific-code-for-BIND-C-types.patch
+++ b/mingw-w64-flang/0003-flang-Inherit-target-specific-code-for-BIND-C-types.patch
@@ -1,0 +1,50 @@
+From afd990ed84c308d5d4d06f461d6bf80b670ddf16 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Markus=20M=C3=BCtzel?= <markus.muetzel@gmx.de>
+Date: Mon, 3 Mar 2025 10:32:22 +0100
+Subject: [PATCH] [flang] Inherit target specific code for BIND(C) types on
+ Windows (#114035)
+
+Inherit target specific code for Windows i386 and x86_64 from the classes that
+define that code for the respective processors on non-Windows operating
+systems.
+Only overload parts that differ.
+
+That allows re-using the existing implementation for BIND(C) types on
+non-Windows x86_64 also for Windows x86_64 targets.
+---
+ lib/Optimizer/CodeGen/Target.cpp | 12 ++++--------
+ 1 file changed, 4 insertions(+), 8 deletions(-)
+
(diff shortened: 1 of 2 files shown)


MehdiChinoune:
There is fgsl which I tried to build recently with flang-20.
I tried yo build some, but I don't remember which ones failed because if this issue.

mmuetzel:
I'm unable to build fgsl. It fails linking with the following error:
```
/bin/sh ./libtool  --tag=FC   --mode=link /clang64/bin/flang  -g -O2 -version-info 1:6:0 -no-undefined  -o libfgsl.la -rpath /clang64/lib fgsl.lo libfgsl_la-fgsl_utils.lo  -lgsl -lgslcblas -lm
libtool: link: /clang64/bin/flang -shared  .libs/fgsl.o .libs/libfgsl_la-fgsl_utils.o   -lgsl -lgslcblas  -g -O2   -o .libs/libfgsl-1.dll -Wl,--enable-auto-image-base -Xlinker --out-implib -Xlinker .libs/libfgsl.dll.a
flang: error: unknown argument: '-Xlinker'
flang: error: unknown argument: '--out-implib'
flang: error: unknown argument: '-Xlinker'
flang: error: no such file or directory: '.libs/libfgsl.dll.a'
```
(But that's probably an issue that should be fixed in libtool.)

Up to that point, there was no error about `BIND(C)` though.

MehdiChinoune:
fgsl succeeds to build with flang-20 after applying your patch without any change on libtool.

mmuetzel:
> fgsl succeeds to build with flang-20 after applying your patch without any change on libtool.

Thanks for testing.
That probably means that `flang` now no longer fails for the `-Xlinker` flag. That's good news. And probably means that we don't need #23564 and can just wait for LLVM 20 instead.
What do you think?

I haven't checked yet whether this change breaks the original use-case that motivated the older change in Flang. I.e., is the ABI for complex floating point values still correct?
Do you recall where we first ran into that? Was that OpenBLAS?

MehdiChinoune:
I think we should wait for llvm 20, and yes tests were failing for OpenBLAS.

mmuetzel:
I opened a PR with that patch upstream: https://github.com/llvm/llvm-project/pull/129579

MehdiChinoune:
There is also HDF5 with Fortran support (mentioned by the llvm issue).

MehdiChinoune:
HDF5 now builds successfully when enabling Fortran support.

MehdiChinoune:
Giving that there are some Fortran projects using CMake build-system, I think we could merge this PR and wait for autotools build-system packages until LLVM-20.

mmuetzel:
See #23590.

PR #23564, opened 2025-03-03
libtool: fix using LLVM Flang 19 as the linker driver for shared libraries

mmuetzel:
If LLVM Flang 19 is used as the linker driver, it fails on the `-Xlinker` flag.
Pass options to the linker with the `-Wl,` flag instead.

This change is required to build fgsl with LLVM Flang (see #23563).

### DIFF
--- a/mingw-w64-libtool/0008-flang-linker-driver-options.patch
+++ b/mingw-w64-libtool/0008-flang-linker-driver-options.patch
@@ -1,0 +1,14 @@
+If LLVM Flang is used as the linker driver, it fails on the `-Xlinker` flag.
+Pass options to the linker with the `-Wl,` flag instead.
+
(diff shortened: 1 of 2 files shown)


mmuetzel:
Hmm...
The error on CLANG64 looks like some file wasn't completely written when it was attempted to use it.

Is it ok to build with `make -j1`?

(Let's just try if that avoids the issue.)

mmuetzel:
Aahh. We never built libltdl with its Fortran parts. And flang isn't installed. 😖

Removing that again helped.

mmuetzel:
According to @MehdiChinoune's testing in #23563, this change is no longer needed for the Flang in LLVM 20.

I'll leave this as a draft for now in case someone needs this before LLVM 20 will be released and packaged in MSYS2.

PR #23578, opened 2025-03-05
godot: update to 4.4

felixaszx:
Update Godot game engine to 4.4-stable.

There are few changes:
1. Jolt-physic is introduced in this new release, but for some reasons their build scripts do not allow using an external instance of it. Therefore we can not dynamically link to msys2's Jolt-physic. Maybe it contains modifications just for Godot.
2. libsquish is removed.
3. Patch used for LLVM workaround is no longer needed but ketp just in case the issue comes back again. I would like to keep it for at least 1 stable release and see how things going on 4.5-stable.


felixaszx:
BTW, I was trying to get their C# scripting working again, but there is way too much pain to get .Net working on MSYS2. Seems like Microsoft just do not want to support their .Net with Mingw.
I can try a external .Net + MSYS2 mix building, but tbh, it is very likely to be a disaster and sounds terrible anyway.

Issue #23577, opened 2025-03-04
Regression with delaylib when using gcc + mingw


p-b-o:
### Description / Steps to reproduce the issue

Around a month ago (between feb 1 and feb 3), we started to notice segmentation fault when using qemu with plugins, **without any source code change**. Plugins are basically shared objects that use a defined API exported by qemu binary itself. This is done thanks to delay load mechanism.

On Linux side, symbols are resolved by the first occurrence (starting with the current binary, and then shared objects loaded). On Windows, symbols are imported from a specific binary/dll instead. Delay load mechanism allows to deal with that situation, by creating a "special" library that defines this semantic.

I'm not sure where the regression is located, but it's present when using gcc only, and not with clang.

Given those sources files:

`main.c`
```
#include <assert.h>
#include <stdio.h>
#include <windows.h>

__declspec(dllexport) int main_f(void) { return 42; }

int main() {
  HMODULE lib = LoadLibraryA("lib.dll");
  assert(lib);
  printf("lib.dll loaded\n");
  FARPROC func = GetProcAddress(lib, "lib_f");
  assert(func);
  printf("lib_f retrieved\n");
  int lib_return = func();
  printf("lib_f called\n");
  assert(lib_return == main_f());
  printf("OK\n");
}
```

`main.def`
```
EXPORTS
    main_f
```

`lib.c`
```
#include <assert.h>
#include <stdio.h>
#include <windows.h>
#include <delayimp.h>

FARPROC WINAPI dll_failure_hook(unsigned dliNotify, PDelayLoadInfo pdli);
PfnDliHook __pfnDliFailureHook2 = dll_failure_hook;
FARPROC WINAPI dll_failure_hook(unsigned dliNotify, PDelayLoadInfo pdli) {
  if (dliNotify == dliFailLoadLib) {
    printf("Loading DLL %s\n", pdli->szDll);
    if (strcmp(pdli->szDll, "does_not_exist.exe") == 0) {
      /* we return main executable */
      return (FARPROC)GetModuleHandle(NULL);
    }
    assert(0);
  }
  /* ignore other requests */
  return 0;
}

__declspec(dllimport) int main_f(void);

int lib_f() {
  printf("lib: call main_f\n");
  int ret = main_f();
  printf("lib: return value f\n");
  return ret;
}
```

We can reproduce with gcc, in a `mingw64` or `ucrt64` environment, execution crashes when calling `main_f` from the library:

```
[...]

PR #23580, opened 2025-03-05
mingw-w64-crt-git: fix delay load libraries

p-b-o:
Details: https://github.com/msys2/MINGW-packages/issues/23577


pbo-linaro:
Built and tested with use case reported on https://github.com/msys2/MINGW-packages/issues/23577.

MehdiChinoune:
changes related to mingw-w64 (crt) are preffered to be upstreamed first.
Please consider sending changes to mingw-w64 mailing list.

pbo-linaro:
The original author, @lhmouse will upstream it, but meanwhile, it would be nice to have to a fix on msys side, as it introduces a segfault for delay load libraries, which is pretty inconvenient when a program is concerned.

Is there an msys2 policy to deal with issues like this?

lazka:
> Please consider sending changes to mingw-w64 mailing list.

It was https://sourceforge.net/p/mingw-w64/mailman/message/59156290/

PR #23582, opened 2025-03-05
safetyhook: update to 0.6.5

llm96:


aixxe:
> The commit has two authors. Does another author aware of this commit?

All good. 👍

PR #23590, opened 2025-03-06
Revert "flang: inherit target specific code for BIND(C) on Windows"

mmuetzel:
This reverts commit https://github.com/msys2/MINGW-packages/commit/32fb1dcb2837a02ab2728a952e295521270a088b.

Also rebuild packages that had been miscompiled with that patched version of LLVM Flang.

@MehdiChinoune: Are there other packages that have been compiled with the broken `BIND(C)` implementation?


mmuetzel:
Why doesn't HDF5 rebuild?
It fails with errors like this:
```
  D:/_/B/src/hdf5-1.14.6/src/H5Tconv_float.c: In function 'H5T__conv_float__Float16':
  D:/_/B/src/hdf5-1.14.6/src/H5Tconv_float.c:1735:65: error: 'FLT16_MAX' undeclared (first use in this function); did you mean 'INT16_MAX'?
   1735 |     H5T_CONV_Ff(FLOAT, FLOAT16, float, H5__Float16, -FLT16_MAX, FLT16_MAX);
        |                                                                 ^~~~~~~~~
```

What has changed since the last rebuild two days ago?

mmuetzel:
It still builds locally without issues in the MINGW64 environment. So, it's likely some package that is currently in the staging repositories.
Could some change in [mingw-w64-headers-git](https://packages.msys2.org/base/mingw-w64-headers-git) have caused this?

mmuetzel:
HDF5 has the following in `H5private.h`:
```

#include <assert.h>
#include <ctype.h>
#include <errno.h>
#include <fcntl.h>
#include <fenv.h>
#include <setjmp.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>

/* Define __STDC_WANT_IEC_60559_TYPES_EXT__ for _FloatN support, if available */
#define __STDC_WANT_IEC_60559_TYPES_EXT__
#include <float.h>
```

Maybe, one of the earlier included headers now includes `float.h` before `#define __STDC_WANT_IEC_60559_TYPES_EXT__`. And an inclusion guard prevents these macros from being defined...

Let's try defining `__STDC_WANT_IEC_60559_TYPES_EXT__` earlier in that header...

mmuetzel:
Might have been this change:
https://github.com/mingw-w64/mingw-w64/commit/b40b6a021dcb6c2396e248e22f682b7bd2599d7a

mmuetzel:
Proposed that patch to upstream HDF5: https://github.com/HDFGroup/hdf5/pull/5356

MehdiChinoune:
eccodes.

PR #23599, opened 2025-03-07
python-poetry/-core: update to 2.1.1

podsvirov:


Biswa96:
Please reset the pkgrel to 1 in python-poetry.

Issue #22600, opened 2024-11-20
freecad: Segfault when opening Prefereces


hmartinez82:
### Description / Steps to reproduce the issue

- Install FreeCAD 1.0rc2-5
- Click on the Edit menu then Preferences

This is also happening with 1.0.0

### Expected behavior

Preferences dialog shows up

### Actual behavior

```
[Thread 23096.0x5410 exited with code 0]

Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffc058974bc in Gui::PythonWrapper::getWrapperName (this=this@entry=0x5f9537, obj=obj@entry=0x5223ca70)
    at X:/Github/MINGW-packages/mingw-w64-freecad/src/FreeCAD-1.0.0/src/Gui/PythonWrapper.cpp:848
848         const QMetaObject* meta = obj->metaObject();
(gdb) bt
#0  0x00007ffc058974bc in Gui::PythonWrapper::getWrapperName (this=this@entry=0x5f9537, obj=obj@entry=0x5223ca70)
    at X:/Github/MINGW-packages/mingw-w64-freecad/src/FreeCAD-1.0.0/src/Gui/PythonWrapper.cpp:848
#1  0x00007ffc0589f9c6 in Gui::UiLoaderPy::load (this=0x5220d4f0, args=...)
    at X:/Github/MINGW-packages/mingw-w64-freecad/src/FreeCAD-1.0.0/src/Gui/UiLoader.cpp:607
#2  0x00007ffc05961aa0 in Py::PythonExtension<Gui::UiLoaderPy>::method_varargs_call_handler (
    _self_and_name_tuple=<optimized out>, _args=0x1d69ece0)
    at X:/Github/MINGW-packages/mingw-w64-freecad/src/FreeCAD-1.0.0/src/CXX/Python3/ExtensionOldType.hxx:309
#3  0x00007ffcc3553460 in ?? () from C:\msys64\ucrt64\bin\libpython3.12.dll
#4  0x00007ffcc3518e3f in ?? () from C:\msys64\ucrt64\bin\libpython3.12.dll
#5  0x00007ffcc35191db in ?? () from C:\msys64\ucrt64\bin\libpython3.12.dll
#6  0x00007ffcc35c87a2 in ?? () from C:\msys64\ucrt64\bin\libpython3.12.dll
#7  0x00007ffcc3476982 in ?? () from C:\msys64\ucrt64\bin\libpython3.12.dll
[...]

PR #23624, opened 2025-03-11
freecad: Disable Addon Manager

hmartinez82:
Fixes https://github.com/msys2/MINGW-packages/issues/22600
image


mingw-w64-freecad/PKGBUILD
```diff
@@ -38,6 +38,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-cmake"
              "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-nlohmann-json"
```
hmartinez82, line 41: This was required due to VTK 9.4, but only at build time

PR #23627, opened 2025-03-11
fpm: update to v0.11.0

zoziha:
Compile mingw-w64-fpm using a more stable single-source fpm bootstrap file fpm-0.8.2.F90 and update the mingw-w64-fpm version to 0.11.0.

### DIFF
--- a/mingw-w64-fpm/001-fix-windows-CRLF.patch
+++ b/mingw-w64-fpm/001-fix-windows-CRLF.patch
@@ -1,0 +1,105 @@
+diff --git a/src/fpm_filesystem.F90 b/src/fpm_filesystem.F90
+index a0066708d..0e7e221b7 100644
+--- a/src/fpm_filesystem.F90
++++ b/src/fpm_filesystem.F90
+@@ -7,7 +7,7 @@ module fpm_filesystem
+                                OS_UNKNOWN, OS_LINUX, OS_MACOS, OS_WINDOWS, &
+                                OS_CYGWIN, OS_SOLARIS, OS_FREEBSD, OS_OPENBSD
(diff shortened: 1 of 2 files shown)


MehdiChinoune:
Why using 0.8.2?, was there an issue with the latest 0.11.0?

zoziha:
Yes, there is a file reading bug under Windows in fpm-0.11.0.F90. It has now been fixed (see https://github.com/fortran-lang/fpm/pull/1107), but this fix will be released in the next fpm version, and I had better add a git patch for mingw-w64-fpm to address this.

fpm-0.8.2.F90 is considered a robust version.

MehdiChinoune:
> fpm-0.8.2.F90 is considered a more robust version.

Who said that, I mean why not 0.9.x or 0.10.x? Is there an upstream statement labeling it as a "more robust" release? or Is it just a personal estimation/opinion?
I assume that the most recent release has less issues, unless the previous one is an LTS.

zoziha:
Sorry, I didn't explain this point earlier.

In the fpm repository, [fpm-0.8.x.F90 is used for CI testing](
https://github.com/fortran-lang/fpm/blob/1ae7351a840e572043e954207ec7c30b03cb3521/.github/workflows/CI.yml#L58). The version 0.8.x is [what the maintainer of fpm suggested to me](https://github.com/fortran-lang/fpm/issues/1105#issuecomment-2712968555).

At present, fpm does not have an LTS version. Given that there is a bug with fpm-0.11.0.F90, we can choose to use 0.8.x or the relatively latest 0.10.x for fpm bootstrapping.

MehdiChinoune:
pushed.

PR #23651, opened 2025-03-13
[new package] typescript 5.8.2

podsvirov:


ognevny:
it will become a Go based package this year. maybe wait for it?

podsvirov:
> it will become a Go based package this year. maybe wait for it?

Or we can have both `typescript` and `typescript-go` as alternativaties if they will be compatible?

PR #23658, opened 2025-03-14
gst-plugins-rs: update to 1.26.0

ognevny:
currently fails with absolutely strange cmake error (that wasn't a Zed case)


ognevny:
this error occurs also in case of aws-lc-sys@v0.25.0 which didn't failed in Zed

ognevny:
wait a second, it failed locally with _another_ error. dough!

ognevny:
@lazka it should be ready now (I also enabled dav1d plugin)

lazka:
Awesome, thanks!

PR #23684, opened 2025-03-15
codeblocks: Update to SVN revision 13634 and

stahta01:
add cbp2make utility


stahta01:
> Please, don't build outside of build-${MSYSTEM}, copy directory if necessary.

okay

stahta01:
See #23687

PR #23686, opened 2025-03-15
handbrake: Enable aarch64 builds

hmartinez82:


Biswa96:
`$MSYSTEM_CARCH` is only defined if /etc/msystem is `source`-ed in shell. It will not be defined if someone uses mingw environment with `$PATH`. Is there any alternative?

hmartinez82:
@Biswa96 What is the proper way to get `x86_64` and `aarch64` given the target arch of the build?

Biswa96:
I can not think of any workaround right now.

PR #23687, opened 2025-03-16
codeblocks: Update to SVN revision 13634 and

stahta01:
add cbp2make utility

### DIFF
--- a/mingw-w64-codeblocks/PKGBUILD
+++ b/mingw-w64-codeblocks/PKGBUILD
@@ -6,7 +6,7 @@ _realname=codeblocks
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 _basever=20.03
-_revision=13620
+_revision=13634
 pkgver=${_basever}.r${_revision}
 pkgrel=1
 pkgdesc="A free C, C++ and Fortran IDE (mingw-w64)"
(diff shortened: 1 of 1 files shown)


mingw-w64-codeblocks/PKGBUILD
```diff
@@ -24,6 +24,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
          "${MINGW_PACKAGE_PREFIX}-tinyxml"
          "${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-msw-libs")
 makedepends=("${MINGW_PACKAGE_PREFIX}-cc"
+             $([[ ${MINGW_PACKAGE_PREFIX} != *-clang-* ]] || echo "${MINGW_PACKAGE_PREFIX}-gcc-compat")
```
MehdiChinoune, line 27: This is not required, just override CC, CXX and LD after make command.

MehdiChinoune:
You don't have to close and open new PR, just amend and force-push.

mingw-w64-codeblocks/PKGBUILD
stahta01: That would require adding a patch to the makefile and is not worth the effort to maintain in my opinion; do you disagree?

stahta01: So, you are the cc=gcc in the makefile will not override that, because I think other wise.

PR #23690, opened 2025-03-16
libsystre: port to meson

lazka:
And get rid of the archive.
Sadly makepkg can't deal with directories, so this means the src pkg is broken now. But better than a binary blob.


lazka:
The motivation here is to simplify things and avoid the tarball in git. Feedback welcome.

lazka:
removed the dir and added all files to source now

PR #23699, opened 2025-03-17
gimp: Update to 3.0.0

lazka:


MehdiChinoune:
Could gimp2 be dropped now? (with python2 packages)

PR #23703, opened 2025-03-17
gimp: disable debug console

ognevny:
it's not needed for release binaries

fixes https://github.com/msys2/MINGW-packages/issues/22629


MehdiChinoune:
gimp 3.0.0 was released.

ognevny:
> gimp 3.0.0 was released.

yes, it's already in repo

I meant that there is no need to have debug console enabled for a release version of gimp

MehdiChinoune:
I mean, just update gimp to 3.0.0 and make this change part of it.

PR #23705, opened 2025-03-17
libavif: update to 1.2.1

kmilos:
I'm puzzled how 1.2.0 was built, nothing substantial was changed w.r.t. libargparse dependency since?!

https://github.com/AOMediaCodec/libavif/compare/v1.2.0...v1.2.1

It now looks like git needs to be added as makedepends for FetchContent to work (and also libxml2 for some new functionality...).


ognevny:
I see it's trying to run git binary

you can switch to git source

ognevny:
ignore previous comment...

actually it's required to apply patches for some vendored sources. upstream uses `git apply` command. so probably the _only_ way to fix this is to use git source (anyway you must add `git` to makedepends)

PR #23709, opened 2025-03-17
drop some 32-bit librvg rdeps (3/N)

ognevny:
that should be a final portion before full librsvg drop

follow-up to #23663


TeaRex73:
I'd like to keep 32-bit builds available at for this year as long as 32-bit Windows 10 is still supported. At least a few people downloaded my alpha release in 32 bit (though the 64 bit build had more downloads of course), so I think the interest is not zero. The affected project is [https://github.com/TeaRex73/xu4-deutsch](https://github.com/TeaRex73/xu4-deutsch) if that matters. I wanted to release a beta today and my 32-bit build failed because of this one (and I noticed all other SDL1 packages are still available in 32-bit). Is there any supported place to download the last msys2 binary package of SDL1's SDL_mixer? Then I could use that in the build process and hope it keeps working at least during 2025.

ognevny:
there is a msys2-archive repo which contains old states of 32-bit repos (2022). you can try to download [SDL2_mixer here](https://github.com/msys2/msys2-archive/releases/download/2022-12-16-mingw32/mingw-w64-i686-SDL2_mixer-2.6.2-1-any.pkg.tar.zst) and install with `pacman -U`

TeaRex73:
SDL2_mixer is not the problem, it's SDL_mixer from SDL1. But thanks for the hint, I will try whether it still works.

ognevny:
> SDL2_mixer is not the problem, it's SDL_mixer from SDL1. But thanks for the hint, I will try whether it still works.

oh, sorry :) there are more packages in that tag. [SDL_mixer](https://github.com/msys2/msys2-archive/releases/download/2022-12-16-mingw32/mingw-w64-i686-SDL_mixer-1.2.12-9-any.pkg.tar.zst)

TeaRex73:
Thanks a lot!!
Downloading the package from that link via `wget` and installing it via `pacman --noconfirm -U` in my GitHub action solved my problem for now. It's a bit ugly but it works.

PR #23713, opened 2025-03-18
Drop gimp2

MehdiChinoune:


ognevny:
makes sense. arch doesn't provide an alternative for gimp (it's 3.0.0 already) and gimp2 will appear probably in AUR

lazka:
I've asked in IRC

lazka:
There will be no 2.12 release in the future. So this is good to go I think.

PR #23730, opened 2025-03-20
Bring back 'mingw32' architecture for gtkmm3

### LINKED ISSUE #23729: Build gtkmm3 for mingw32
### Package name

mingw-w64-i686-gtkmm3

### For which environment do you need the package?

MINGW32

### Provide a brief description why you need this package in the selected environment

Please kindly bring back the gtkmm3 build for Win32.

I'm doing some development for FontForge, which still maintains the Win32 build, and I'm going to need it. This environment was dropped recently in #23709.

Yes.

iorsh:
Closes #23729 - see comment there.


MehdiChinoune:
Rust dropped support for mingw-i686, as a result all rust packages were dropped except librsvg.
All reverse-dependencies of librsvg were dropped (including gtkmm3), except those required for gimp because it has a printing plugin that could only built in 32-bit.

ognevny:
> Rust dropped support for mingw-i686

let me clarify things.

this wasn't actual support drop, but _disabling tests_. in next weeks upstream Rust won't run any tests for i686-pc-windows-gnu, but they'll still be ensured that the target is _buildable_.

I'm sure that will lead to unexpected bugs happening with compiler and other tools with i686-pc-windows-gnu host, so there will be extra demotion with a library-only support, without host tools

eventually i686-pc-windows-gnu will become a cross-compiled target.

for us cross-compiling is not good, as we want the same host and target for each environment. so we can support i686-pc-windows-gnu host until it becomes uncompilable

Issue #23715, opened 2025-03-18
mold Segmentation fault


SquallATF:
### Description / Steps to reproduce the issue

A segmentation error will occur when mold exits
`mold --version`
![Image](https://github.com/user-attachments/assets/acacefd1-ef45-49af-b4d7-b2af0c0be4c5)
`mold --help`
![Image](https://github.com/user-attachments/assets/1e01326f-764a-4162-a695-b48e680f5597)

the reason is mimalloc loading incorrectly
`MIMALLOC_VERBOSE=1 mold --version `
```
mimalloc: option 'show_errors': 0
mimalloc: option 'show_stats': 0
mimalloc: option 'verbose': 1
mimalloc: option 'eager_commit': 1
mimalloc: option 'arena_eager_commit': 2
mimalloc: option 'purge_decommits': 1
mimalloc: option 'allow_large_os_pages': 0
mimalloc: option 'reserve_huge_os_pages': 0
mimalloc: option 'reserve_huge_os_pages_at': -1
mimalloc: option 'reserve_os_memory': 0 KiB
mimalloc: option 'deprecated_segment_cache': 0
mimalloc: option 'deprecated_page_reset': 0
mimalloc: option 'abandoned_page_purge': 0
mimalloc: option 'deprecated_segment_reset': 0
mimalloc: option 'eager_commit_delay': 1
mimalloc: option 'purge_delay': 10
mimalloc: option 'use_numa_nodes': 0
mimalloc: option 'disallow_os_alloc': 0
mimalloc: option 'os_tag': 100
mimalloc: option 'max_errors': 32
mimalloc: option 'max_warnings': 32
mimalloc: option 'max_segment_reclaim': 10
mimalloc: option 'destroy_on_exit': 0
mimalloc: option 'arena_reserve': 1048576 KiB
mimalloc: option 'arena_purge_mult': 10
mimalloc: option 'purge_extend_delay': 1
mimalloc: option 'abandoned_reclaim_on_free': 0
mimalloc: option 'disallow_arena_alloc': 0
mimalloc: option 'retry_on_oom': 400
mimalloc: option 'visit_abandoned': 0
mimalloc: option 'guarded_min': 0
mimalloc: option 'guarded_max': 1073741824
mimalloc: option 'guarded_precise': 0
mimalloc: option 'guarded_sample_rate': 0
mimalloc: option 'guarded_sample_seed': 0
mimalloc: option 'target_segments_per_thread': 0
[...]

PR #23741, opened 2025-03-21
mimalloc: update to 2.2.2

ognevny:


yujincheng08:
Maybe we can also add patch from https://github.com/msys2/MINGW-packages/issues/23715#issuecomment-2732309843?

ognevny:
> Maybe we can also add patch from https://github.com/msys2/MINGW-packages/issues/23715#issuecomment-2732309843?

this is a patch for mold, not mimalloc

ognevny:
I'm not an actual maintainer of this package. feel free to contribute if something should be fixed. I didn't apply a prefix hunk because it doesn't seem to have any sense

PR #23747, opened 2025-03-23
whisper.cpp : update to b2266 and refactor build script to align with latest codebase

peter277:
Updated to b2266, refactored build script to align with latest codebase changes, and introduced llama.cpp as a dependency for GGML library.

Note: Introduced llama.cpp as a temporary dependency for the common GGML library. This will remain until GGML is packaged separately upstream, which will require proper versioning to be implemented.


MehdiChinoune:
Please, keep the previous prefix 'whisper.cpp' instead 'whisper'.

peter277:
> Please, keep the previous prefix 'whisper.cpp' instead 'whisper'.

The MSYS2 package for whisper.cpp has not been updated in over a year, and upstream has since changed its naming conventions for executables. Previously, loosely named binaries like main.exe and server.exe have been replaced with more structured names such as whisper-cli.exe and whisper-server.exe. To align with upstream, I have changed the prefix from whisper.cpp- to whisper-, which follows the current upstream naming conventions.

Example evidence from upstream changes:

[whisper-cli.exe introduced](https://github.com/ggerganov/whisper.cpp/blob/7fe4979f25ea1b2e0696a11007c9d46e168d18b3/examples/cli/CMakeLists.txt#L1) (previously main.exe)

[whisper-stream.exe introduced](https://github.com/ggerganov/whisper.cpp/blob/7fe4979f25ea1b2e0696a11007c9d46e168d18b3/examples/stream/CMakeLists.txt#L2)

[whisper-server.exe introduced](https://github.com/ggerganov/whisper.cpp/blob/7fe4979f25ea1b2e0696a11007c9d46e168d18b3/examples/server/CMakeLists.txt#L1)

Since upstream has standardized on whisper- as the prefix for its binaries, I think it makes sense to follow suit for consistency. Keeping the older whisper.cpp- prefix would diverge from upstream and could lead to unnecessary confusion.

mingw-w64-whisper.cpp/PKGBUILD
```diff
@@ -1,44 +1,29 @@
-# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
```
MehdiChinoune, line 1: Why?

mingw-w64-whisper.cpp/PKGBUILD
```diff
@@ -78,22 +61,18 @@ package() {
...
-  # Install the examples by prefixing them with whisper.cpp
-  # except main.exe, which gets installed as plain whisper.cpp.exe
-  for i in bin/*.exe; do
-    install -Dm755 "${i}" \
-      "${pkgdir}${MINGW_PREFIX}/bin/${_realname}-${i//bin\//}"
+  # Rename output files by prefixing them with whisper (if not already prefixed and if not DLL)
+  for i in "${pkgdir}"${MINGW_PREFIX}/bin/*.{exe,py,sh}; do
+    filename=$(basename "$i")
+    
+    if [[ -e $i && ! ("$filename" =~ ^whisper || "$filename" =~ \.dll$) ]]; then
```
MehdiChinoune, line 68: `"$filename" =~ \.dll$` is not required, you are renaming only files with `*.{exe,py,sh}` extension.

mingw-w64-whisper.cpp/PKGBUILD
```diff
@@ -1,44 +1,29 @@
...
+pkgrel=1
 pkgdesc="Port of OpenAI's Whisper model in C/C++ (mingw-w64)"
 arch=('any')
 mingw_arch=('ucrt64' 'clang64' 'clangarm64' 'mingw64')
 url='https://github.com/ggerganov/whisper.cpp'
 license=('spdx:MIT')
 depends=(
-  "${MINGW_PACKAGE_PREFIX}-openblas"
-  "${MINGW_PACKAGE_PREFIX}-SDL2"
+  "${MINGW_PACKAGE_PREFIX}-curl"
```
MehdiChinoune, line 15: looking at runtime dependencies and whisper code, curl library is not used at all.

mingw-w64-whisper.cpp/PKGBUILD
peter277: Seemed semi-abandoned since package was not updated for a relatively long time. I've added maintainer name back in.

mingw-w64-whisper.cpp/PKGBUILD
peter277: Amended.

mingw-w64-whisper.cpp/PKGBUILD
peter277: Had included due to WHISPER_CURL compile option being switched on (rf. https://github.com/ggerganov/whisper.cpp/blob/7fe4979f25ea1b2e0696a11007c9d46e168d18b3/CMakeLists.txt#L84 ) - was not aware it was not actually used. Have removed dependency for now, though nonetheless it will be included transitively through llama.cpp package.

PR #23754, opened 2025-03-24
mimalloc: remove the `lib` prefix when enabling mimalloc-redirect

SquallATF:
The mimalloc-redirect feature supports only the following library names: `mimalloc.dll`, `mimalloc-override.dll`, `mimalloc-secure.dll`, `mimalloc-secure-debug.dll`, `mimalloc-debug.dll`, and `mimalloc-release.dll`, none of which include the `lib` prefix.
https://github.com/microsoft/mimalloc/pull/1045


ognevny:
seems reasonable, thanks

lazka:
I just noticed that mimalloc-redirect is not actually open source. Is there an alternative?

ognevny:
it can be opt out with few cmake options IIUC

lazka:
yeah, done in #23756

PR #23756, opened 2025-03-24
mimalloc: drop mimalloc-redirect, not open source

lazka:
It just copies pre-built binaries that mimalloc has in git.

See https://github.com/microsoft/mimalloc/issues/836


ognevny:
oh, you was faster :)

PR #23765, opened 2025-03-25
crystal: rebuild with llvm 20.1.1

HertzDevil:
Needed for an LLVM regression ([llvm/llvm-project#130632](https://github.com/llvm/llvm-project/issues/130632))


MehdiChinoune:
Isn't crystal dynamically linked to llvm, so It doesn't require a rebuild.

PR #23767, opened 2025-03-25
svt-av1: decoder is no longer shipped

kmilos:
This is just for repo metadata, no real need for a new build...


Biswa96:
Scheduled CI has not started yet in queue https://packages.msys2.org/queue. The change will be added as usual without pkgrel bump.

PR #23779, opened 2025-03-26
neovim: update to 0.11.0

podsvirov:


podsvirov:
The `MINGW64` error:
```
  D:/_/B/src/neovim-0.11.0/src/nvim/main.c:114:4: error: #error UCRT is the only supported C runtime on windows
    114 | #  error UCRT is the only supported C runtime on windows
        |    ^~~~~
```

Biswa96:
mingw64 can be removed from the list.

podsvirov:
> mingw64 can be removed from the list.

And for all dependent packagens.

ognevny:
there is a bit more dependencies. you can drop everything with `tree-siitter-` in its name (obviously excluding `tree-sitter` itself)

podsvirov:
> there is a bit more dependencies. you can drop everything with `tree-siitter-` in its name (obviously excluding `tree-sitter` itself)

Done.

PR #23785, opened 2025-03-27
kcachegrind: New package

lhmouse:


lhmouse:
FWIW we can also provide QCacheGrind with qt6-static. dunno whether anyone wants it.

lhmouse:
should be ready now.

Biswa96:
Would you mind if I modify the PKGBUILD to follow other KDE packages in this repository?

lhmouse:
Of course not.

Biswa96:
Does the change look ok?

lhmouse:
not sure about `depends` though. I checked the executable with Dependencies and not all of these DLLs were directly referenced. Do we have to list indirect dependencies, too?

Biswa96:
The following screenshot of Dependencies shows the required DLLs with kcachegrind

image

Also, ntldd shows same output.

lhmouse:
Fair enough. The change looks good to me.

Issue #23346, opened 2025-02-11
Dropping 32bit Rust


lazka:
@workingjubilee approached us yesterday in chat asking if it would be a problem if upstream rust would drop "i686-pc-windows-gnu"

> I'm preparing an RFC with a peer, but basically we have a lot of problems because
> - no one tests the target enough
> - no one contributes enough maintenance to it
> - the people who do maintain it are already doing a lot and this is like their quaternary thing after also helping out with more general windows  support
> - we don't even have many downloads for it

@ognevny and @1480c1 came out in support of dropping 32bit rust.

Looking at https://packages.msys2.org/packages/mingw-w64-i686-rust we have:

* libimagequant
* libdovi
* rav1e
* librsvg

The first three we can look into dropping, for librsvg I'll ask GIMP devs again what their status on gimp2 releases is.

workingjubilee:
you'll have at least a few months to figure out what the story will be for librsvg, as the very fastest possible turnaround for us deciding to Officially drop it going forward will be in Rust 1.87

lazka:
good to know, thanks

Jehan:
> I'll ask GIMP devs again what their status on gimp2 releases will be in the future. @Jehan

GIMP 2 should not be a problem since we should have GIMP 3 released by then. But this is about 32-bit support, so even GIMP 3 is concerned. Right now our Windows installer is a universal x86/x86_64/AArch64 installer (though our MSIX build which is brand new is x86_64/AArch64 only).

I *think* (i.e. if I recall correctly) that we wanted to have at least a few releases of GIMP 3 with x86 build so that people can at least get a somewhat early version of GIMP 3 at their disposal even if we were to drop this architecture later on.
I.e. that even if people would not have updates (and worse: security updates) anyway — so their situation is definitely not good —, they would at least have some reasonably modern version of GIMP (GTK3, Python3, Wayland support, etc.).

@brunvonlope Was that the last state of our discussions? Does Jernej also have an account on Github by the way?

Personally I wished we would not basically deprecate a whole range of perfectly working hardware (which will end up in trash at some point when it will have no update whatsoever from any software), but I know I can't fight against this current. Even less all by myself. Ecologically this is a shame. But I am powerless here. 🤷

brunvonlope:
> Was that the last state of our discussions?

[...]

PR #23789, opened 2025-03-27
rust: build only rustc and cargo for i686

ognevny:
cc #23346

there is no sense of using rustfmt or clippy in i686 if you have x86_64 working

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -18,7 +18,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
          "${MINGW_PACKAGE_PREFIX}-rust-wasm" \
          "${MINGW_PACKAGE_PREFIX}-rust-src"))
 pkgver=1.85.1
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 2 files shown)


ginnyTheCat:
I'm not sure if this was intended but this removes rustdoc as well, preventing doctests from running in i686 CI. Am I understanding correctly that with this step mingw i686 rust is deprecated and it should not be used in CI anymore?

ognevny:
for what do you need exactly i686 rustdoc? what's wrong with x86_64 one?

and starting with Rust 1.88.0 it will slowly become a tier-2 target *without* host tools, which will eventually mean, that rustdoc and others won't be guaranteed to work or even to compile

ginnyTheCat:
> for what do you need exactly i686 rustdoc? what's wrong with x86_64 one?

We run tests in i686 mingw, and idk if x86_64 rustdoc can run doctests for i686. We are testing x86_64 as well, but as long as 32bit windows rust is not deprecated, I'd prefer testing it in CI. But i'm more than fine with removing i686 from the CI if you think that's more the path to the future.

Edit: I do not know a lot about mingw so it is very likely that I'm misunderstanding sth.

PR #23790, opened 2025-03-27
SDL2_image: drop mingw32

ognevny:
actually it should be `drop some 32-bit for librsvg rdeps (5/N)`, but there's only one package...

follow-up to #23723


ognevny:
@mmuetzel will you mind if I drop `octave`? I can keep it if you don't want

mmuetzel:
Octave 10 will probably be released in the next couple of days. We planned to drop mingw32 with that.

Could you still wait until then?

PR #23791, opened 2025-03-27
cmake: update to 4.0.0

MehdiChinoune:
Breaking change: drop compatibility with CMake<3.5, If cmake throws an error about it just pass `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`, but keep in mind that It doesn't always work. see https://github.com/microsoft/vcpkg/pull/44273 as an example.


Biswa96:
Would it be possible to patch the error as a warning?

MehdiChinoune:
CMake 3.5 was released in 2016, I side with CMake maintainers. There was a warning about compatibility with old CMake, but many developers ignored it. mingw-w64-cmake is not just used for packaging. Those developers should update their projects.

ognevny:
it would be problematic for some rust packages that use cmake-rs to compile which didn't care about cmake warnings...

PR #23802, opened 2025-03-29
libgme: update to 0.6.4 and update URLs.

sezero:


lazka:
Why did you remove the static build?

sezero:
> Why did you remove the static build?

I did not. Look at the build options passed to cmake.

lazka:
I see, thanks, I only looked at the file diff:

```
  -/ucrt64/lib/libgme.a
   /ucrt64/lib/libgme.dll.a
  +/ucrt64/lib/libgme-static.a
```

So gme renames the static build if both are build. We probably still need to build twice to get the right naming back.

sezero:
Added a patch which I just submitted to mainstream at https://github.com/libgme/game-music-emu/pull/123

Wohlstand:
Whoops... It was supposed to add "-static" suffix to MSVC-built only thing, thanks to @sezero for the quick patch, I accepted it.

sezero:
OK, I finalized the PR. Anything else needed?

lazka:
thanks!

sezero:
You're welcome

PR #23805, opened 2025-03-30
aichat: new package

wszqkzqk:
AIChat is an all-in-one LLM CLI tool featuring Shell Assistant, Chat-REPL, RAG, AI Tools & Agents, with access to OpenAI, Claude, Gemini, Ollama, Groq, and more.

### DIFF
--- a/mingw-w64-aichat/PKGBUILD
+++ b/mingw-w64-aichat/PKGBUILD
@@ -1,0 +1,48 @@
+# Maintainer: wszqkzqk <wszqkzqk@qq.com>
+
+_realname=aichat
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=0.29.0
+pkgrel=1
+pkgdesc="All-in-one AI CLI tool featuring Chat-REPL, Shell Assistant, RAG, AI tools & agents (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/sigoden/aichat'
+msys2_references=(
+  'archlinux: aichat'
+)
+license=('MIT OR Apache-2.0')
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust" "${MINGW_PACKAGE_PREFIX}-pkgconf")
+source=("${_realname}-${pkgver}.tar.gz::${url}/archive/v${pkgver}.tar.gz")
+sha256sums=('0b586419ce4e29e02eb165e0ab668e0661fac305840348467ab5f45e42551a5a')
+
+prepare() {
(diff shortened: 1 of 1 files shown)


Biswa96:
Please remove `/usr` from target directory.

wszqkzqk:
> Please remove `/usr` from target directory.

Understood regarding removing `/usr` as the target directory. In that case, what is the standard/expected location for installing the shell completion files?

Biswa96:
> what is the standard/expected location for installing the shell completion files?

I just follow whatever Arch uses. Just replace `/usr` in Arch PKGBUILD with `${pkgdir}${MINGW_PREFIX}`.

wszqkzqk:
> > what is the standard/expected location for installing the shell completion files?
> 
> I just follow whatever Arch uses. Just replace `/usr` in Arch PKGBUILD with `${pkgdir}${MINGW_PREFIX}`.

I see.

mingw-w64-aichat/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+)
+license=('MIT OR Apache-2.0')
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+makedepends=("${MINGW_PACKAGE_PREFIX}-rust" "${MINGW_PACKAGE_PREFIX}-pkgconf")
+source=("${_realname}-${pkgver}.tar.gz::${url}/archive/v${pkgver}.tar.gz")
+sha256sums=('0b586419ce4e29e02eb165e0ab668e0661fac305840348467ab5f45e42551a5a')
+
+prepare() {
+  cd "${_realname}-${pkgver}"
+  export RUSTUP_TOOLCHAIN=stable
```
ognevny, line 23: we don't use rustup, please remove this

mingw-w64-aichat/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+  cargo fetch --locked --target "$(rustc -vV | sed -n 's/host: //p')"
+}
+
+build() {
+  cd "${_realname}-${pkgver}"
+  export RUSTUP_TOOLCHAIN=stable
+  export CARGO_TARGET_DIR=target
```
ognevny, line 30: the same for above 2 lines

ognevny (changes requested):

mingw-w64-aichat/PKGBUILD
ognevny: I meant CARGO_TARGET_DIR should be also removed 


these environment variables are used in AUR packages for easier and reproducible local builds, because users can set nightly toolchain as their default. also they can change target directory name. this is not our case, as Rust is provided by a separate package. it's guaranteed to be stable and the packages are built with automatic tools, which prevents from setting any unneded environment variables

ognevny (commented):
I added a comment to the previous review which is marked unresolved

wszqkzqk:
> I added a comment to the previous review which is marked unresolved

Done.

PR #23810, opened 2025-03-30
libmikmod: disable openal output.

sezero:
windows native output drivers are more than enough.

( CC @Biswa96 as the last updater of the pkg )

### DIFF
--- a/mingw-w64-libmikmod/PKGBUILD
+++ b/mingw-w64-libmikmod/PKGBUILD
@@ -4,7 +4,7 @@ _realname=libmikmod
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
 pkgver=3.3.12
-pkgrel=1
+pkgrel=2
 pkgdesc="A portable sound library (mingw-w64)"
 arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
What's the motivation for this? What about the xaudio2/sdl2 changes?

sezero:
> What's the motivation for this?

Windows builds of libmikmod already have two native output drivers, dsound and winmm,
enabled by defult. Openal here is a completely pointless extra dependency.

> What about the xaudio2/sdl2 changes?

sdl2 option depends on sdl: its a choice of sdl version. so removed sdl2 part.
xaudio2 is a 'while I was there' thing. Both sdl[2] and xaudio2 drivers (as well
as openal) are disabled by default already, but since this packaging explicitly
disables sdl[2], I just added xaudio2 among them.

I can just remove thos `--diable-xxx` for those drivers which are already disabled
by default, too, if that is prefrerred.

lazka:
> I can just remove those `--disable-xxx` for those drivers which are already disabled
> by default, too, if that is prefrerred.

No, makes sense, thanks for the explanation.

sezero:
Rebased to cureent git.

Is there any changes needed for this?

lazka:
If someone ends up here because they miss openal, let us know.

PR #23812, opened 2025-03-31
Groonga: Update to 15.0.4

otegami:
Groonga version15.0.4 has just released [here](https://github.com/groonga/groonga/releases/tag/v15.0.4).

### DIFF
--- a/mingw-w64-groonga/PKGBUILD
+++ b/mingw-w64-groonga/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=groonga
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=15.0.3
+pkgver=15.0.4
 pkgrel=1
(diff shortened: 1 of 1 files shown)


otegami:
Thank you so much!

Biswa96:
Is it allowed to use system provided llama.cpp with groonga? The version will differ from vendored llama.cpp in groonga.

Biswa96:
cmake shows error with system provided llama.cpp library in groonga.

```
CMake Error at lib/CMakeLists.txt:191 (target_link_libraries):
  The link interface of target "libgroonga" contains:

    Groonga::llama

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

otegami:
Thanks for the report! I’ve confirmed the build error when using the system-provided llama.cpp.
We’ll fix it in the following PR, so it’ll be resolved in the upcoming release.
- https://github.com/groonga/groonga/pull/2289

PR #23814, opened 2025-03-31
Qt6: update to 6.8.3

sunderme:
update to qt6.8.3
fix conflict in multimedia-ffmeg package


MehdiChinoune:
Qt 6.9.0 is going to be released on April 2nd. I think you should wait a little and update to it.

PR #23819, opened 2025-03-31
octave: update to 10.1.0

mmuetzel:
Also drop mingw32.


MehdiChinoune:
It fails on CLANG64 because of some C++ issues and on UCRT64 because of missing TeX installation. I tried it yesterday.

mmuetzel:
> It fails on CLANG64 because of some C++ template issues and on UCRT64 because of missing TeX installation.

I thought that this would be addressed already with the changes here.
Let's see how the CI fares...

MehdiChinoune:
OK, I see "--disable-lib-visibility-flags" and "--disable-docs"

ognevny:
thanks! some its dependencies will be dropped later :)

PR #23831, opened 2025-03-31
libuhd: update to 4.8.0.0

chrisjohgorman:
- Removed use of cast of thrd->native_handle() to (pthread_t) in uhd::set_thread_name opting for no thread naming on windows.  This follows gnuradio which doesn't name threads on windows.
- Add new dependencies and makedepends for package.
- Renamed and rebuilt patch against new uhd sources.
- Rebuild gnuradio against new libuhd.


mingw-w64-libuhd/PKGBUILD
```diff
@@ -67,12 +71,18 @@ build() {
...
+  #run tests with dll directory added to the path
+  PATH=$PWD/lib:$PATH \
+    "${MINGW_PREFIX}"/bin/ctest.exe --timeout 90 --output-on-failure || warning "Tests failed"
 }
 
 package() {
   DESTDIR="${pkgdir}" "${MINGW_PREFIX}"/bin/cmake.exe --install "build-${MSYSTEM}"
 
+  cd "${srcdir}"/"build-${MSYSTEM}"/"python"
+  pip install --ignore-installed --prefix="${pkgdir}"/"${MINGW_PREFIX}" dist/uhd*.whl
```
ognevny, line 84: is it possible to implement this with `python-install`? https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates%2FPKGBUILD.python.sh#L53

mingw-w64-libuhd/PKGBUILD
```diff
@@ -18,23 +18,27 @@ depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs"
...
+        0001-libraries_headers_and_no_pthread_setname_np.patch)
+sha256sums=('a2159491949477dca67f5a9b05f5a80d8c2b32e91b95dd7fac8ddd3893e36d09'
+            'aa8a56652d90c9ad869e03c2aebf70a741b63266f681fc0fc0fd4fc856781846')
 
 prepare() {
   cd "${srcdir}"/uhd-${pkgver}
 
-  patch -Np1 -i "${srcdir}"/0001-libraries_headers_and_pthread_setname_np_cast.patch
+  patch -Np1 -i "${srcdir}"/0001-libraries_headers_and_no_pthread_setname_np.patch
+
```
ognevny, line 41: unneded newline :)

chrisjohgorman: Hi and than you @ognevny.  I have added the requested changes.  I am keeping this as a work in progress as the build machine artifacts did not run properly on my machine.  I hope with the suggested changes they will run properly after the build.  (My local builds ran fine, but the build machine artifacts couldn't load a library properly.)

ognevny: they probably won't resolve issue. that was nits to better match common PKGBUILD style

chrisjohgorman:
Well @ognevny you were correct, it didn't fix the problem.  I think I have
narrowed it down.  The build machine links to python312.dll and my machine
links to libpython3.12.dll.  Off to see why.

chrisjohgorman:
A rewrite of one of the points in my git commit with typos corrected.
- add some guard code to UHDPython.cmake to ensure that cmake's find_package()
  command doesn't use the libraries in the buildmachine's c:/hostedtoolcache
  directory.

This looks good to go all packages install and the library link bug has been fixed.

chrisjohgorman:
I've made all of @ognevny 's recommended changes, and fixed the linking error made by cmake's find_package(Python3) call.  I've squashed my 3 commits into 1 and run.  I don't own any Ettus Research USRPs, so I can't test functionality of the updated software on real hardware, so I tested the ability of python to import the uhd module and ran the test cases.  For the test cases I get 94% passing on ucrt64, 96% passing on mingw64 and 98% passing on clang64.  I get two test failures and between 0 and 4 test timeouts depending on which MINGW_ARCH the tests run under. 

I believe this is ready for inclusion, but if there are more changes to be done, please let me know.

PR #23843, opened 2025-04-02
dust: update to 1.2.0

ognevny:


ognevny:
it seems that our `crt-static = false` config is broken
```
/clang64/bin/dust.exe:
  	libunwind.dll => not found
```

PR #23847, opened 2025-04-02
rust: update to 1.86.0

ognevny:
0009-support-LLVM20.patch: refresh

config change:

- try thinlto


ognevny:
>     * try thinlto

small performance boost?
```
Finished `release` profile [optimized] target(s) in 14m 50s - no lto
Finished `release` profile [optimized] target(s) in 14m 25s - lto
```

PR #23855, opened 2025-04-03
lib3mf: Update to 2.4.1

kintel:


kintel:
@Biswa96 FYI as I saw you bumped the previous version.

mingw-w64-lib3mf/0002-lib3mf-cmake-fix-cmake-4-build.patch
```diff
@@ -0,0 +1,8 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,4 +1,4 @@
+-cmake_minimum_required (VERSION 3.0)
++cmake_minimum_required (VERSION 3.5)
```
lazka, line 5: fyi, I think you an also just pass ``-DCMAKE_POLICY_VERSION_MINIMUM=3.5`

lazka:
I guess patching in a `#include <algorithm>` might help?

mingw-w64-lib3mf/0002-lib3mf-cmake-fix-cmake-4-build.patch
Biswa96: I have added that option and removed the patch file.

PR #23884, opened 2025-04-05
icu: update to 77.1

MehdiChinoune:
I managed to rebuild all dependent packages except qt5-base and qtwebkit. The latter needs the first to be built first.
There is a regression in msys2-runtime, but I don't know when did it happen.

### DIFF
--- a/mingw-w64-icu/.gitignore
+++ b/mingw-w64-icu/.gitignore
@@ -1,0 +1,1 @@
+reverse_of_ICU-22610.patch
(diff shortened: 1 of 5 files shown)


lazka:
Thanks for tracking it down. So it has perl scripts checking for cygwin in there?

lazka:
ah, I guess this needs to be changed to "cygwin": https://invent.kde.org/qt/qt/qtbase/-/blob/75279b03791c65953b0f5a921cf4b539298ad445/bin/syncqt.pl#L57

lazka:
Or something like:

```perl
my $normalizePath_fixDrive = ($^O eq "msys" || (defined $ENV{MSYSTEM} && $^O eq "cygwin") ? 1 : 0);
```

lazka:
I can give it a try if wanted.

MehdiChinoune:
Now, qtwebkit fails and It seems that failure is related to perl.

PR #23898, opened 2025-04-08
xmake: Update to 2.9.9 and enable aarch64 builds

hmartinez82:


mingw-w64-xmake/PKGBUILD
```diff
@@ -27,6 +27,11 @@ prepare() {
 
 build() {
     cd "${_realname}-${pkgver}"
+
+    if [[ ${CARCH} == aarch64 ]]; then
+        export CFLAGS="-D_M_ARM64"
```
Biswa96, line 32:
```suggestion
        export CFLAGS+=" -D_M_ARM64"
```

Or, something similar like that.

PR #23900, opened 2025-04-08
wabt: Update to 1.0.37 and enable aarch64 builds

hmartinez82:


mingw-w64-wabt/0002-win32-tests.patch
```diff
@@ -0,0 +1,12 @@
+diff -bur wabt-orig/test/run-spec-wasm2c.py wabt/test/run-spec-wasm2c.py
+--- wabt-orig/test/run-spec-wasm2c.py	2025-04-08 01:30:33.953857800 -0600
++++ wabt/test/run-spec-wasm2c.py	2025-04-08 01:31:42.016723500 -0600
+@@ -33,7 +33,7 @@
+ SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+ WASM2C_DIR = os.path.join(find_exe.REPO_ROOT_DIR, 'wasm2c')
+ SIMDE_DIR = os.path.join(find_exe.REPO_ROOT_DIR, 'third_party/simde')
+-IS_WINDOWS = sys.platform == 'win32'
++IS_WINDOWS = False
```
Biswa96, line 9: could you explain this change please? It seems counterintuitive.

hmartinez82: Yes. I had to research all over the code. This is all about using cl.exe instead of cc :(

I can swap this with fixing where this is actually used.

Also, this is only affecting the `check()` stage.

https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L455
https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L462
https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L495
https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L510
https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L618
https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L657

PR #23936, opened 2025-04-11
rust: build msys2 target

Berrysoft:
Build MSYS2 target based on `x86_64-pc-cygwin`. Users need to pass `--target x86_64-pc-cygwin` to target MSYS2.

The target name might be a little confusing, though. I don't add a new target because some other crates can only recognize `x86_64-pc-cygwin` (e.g., `cc`). I didn't find a way to add a target alias, so let's keep it simple.

This target is patched to use `libmsys-2.0.a` and `x86_64-pc-msys-gcc` as default. I guess no cygwin user would use a MINGW/MSYS2 rust so they won't find out this change:)


mingw-w64-rust/0010-fix-msys2-link.patch
ognevny, line 1: why is a binary file? GitHub don't show a text

mingw-w64-rust/config.toml
```diff
@@ -115,3 +115,10 @@ default-linker = "wasm-ld.exe"
 sanitizers = false
 profiler = false
 wasi-root = "$MINGW_PREFIX/share/wasi-sysroot"
+
+[target.x86_64-pc-cygwin]
+cc = "$MSYS2_PREFIX/bin/gcc.exe"
+cxx = "$MSYS2_PREFIX/bin/g++.exe"
+ar = "$MSYS2_PREFIX/bin/ar.exe"
+ranlib = "$MSYS2_PREFIX/bin/ranlib.exe"
+linker = "x86_64-pc-msys-gcc.exe"
```
ognevny, line 124:
```suggestion
linker = "x86_64-pc-msys-gcc.exe"
default-linker = "x86_64-pc-msys-gcc.exe"
```
default-linker is for rustc runtime, linker is for rustc build time

ognevny (changes requested):
good start, but few suggestions needs to be addressed

mingw-w64-rust/0010-fix-msys2-link.patch
Berrysoft: I don't know why...

ognevny: can you past a text of patch here, maybe I can try to push it

Berrysoft:
```patch
diff --git a/compiler/rustc_target/src/spec/base/cygwin.rs b/compiler/rustc_target/src/spec/base/cygwin.rs
index 819d1d68a71..c4f0e91e442 100644
--- a/compiler/rustc_target/src/spec/base/cygwin.rs
+++ b/compiler/rustc_target/src/spec/base/cygwin.rs
@@ -14,7 +14,7 @@ pub(crate) fn opts() -> TargetOptions {
         LinkerFlavor::Gnu(Cc::Yes, Lld::No),
         &["-Wl,--disable-dynamicbase", "-Wl,--enable-auto-image-base"],
     );
-    let cygwin_libs = &["-lcygwin", "-lgcc", "-lcygwin", "-luser32", "-lkernel32", "-lgcc_s"];
+    let cygwin_libs = &["-lmsys-2.0", "-lgcc", "-lmsys-2.0", "-luser32", "-lkernel32", "-lgcc_s"];
     let mut late_link_args =
         TargetOptions::link_args(LinkerFlavor::Gnu(Cc::No, Lld::No), cygwin_libs);
     crate::spec::add_link_args(
diff --git a/compiler/rustc_target/src/spec/targets/x86_64_pc_cygwin.rs b/compiler/rustc_target/src/spec/targets/x86_64_pc_cygwin.rs
index eac4caf41c8..7f8d684bc86 100644
--- a/compiler/rustc_target/src/spec/targets/x86_64_pc_cygwin.rs
+++ b/compiler/rustc_target/src/spec/targets/x86_64_pc_cygwin.rs
@@ -6,7 +6,7 @@ pub(crate) fn target() -> Target {
     base.add_pre_link_args(LinkerFlavor::Gnu(Cc::No, Lld::No), &["-m", "i386pep"]);
     base.add_pre_link_args(LinkerFlavor::Gnu(Cc::Yes, Lld::No), &["-m64"]);
     base.max_atomic_width = Some(64);
-    base.linker = Some("x86_64-pc-cygwin-gcc".into());
+    base.linker = Some("x86_64-pc-msys-gcc".into());
     Target {
         llvm_target: "x86_64-pc-cygwin".into(),
         pointer_width: 64,
```

mingw-w64-rust/config.toml
Berrysoft: Fixed.

mingw-w64-rust/0010-fix-msys2-link.patch
Berrysoft: Oh, I see. powershell generates a UTF 16 file.

mingw-w64-rust/PKGBUILD
```diff
@@ -237,6 +245,18 @@ package_rust-wasm() {
   install -Dm644 ../LICENSE-{APACHE,MIT} -t "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}-wasm"
 }
 
+package_rust-cygwin() {
+  pkgdesc="MSYS2 target for Rust (mingw-w64) (--target=x86_64-pc-cygwin)"
```
ognevny, line 249:
```suggestion
  pkgdesc="MSYS2 target for Rust (mingw-w64)"
```
users should be aware of how to cross-compile with rust :)

Berrysoft:
Well... I just realized that rust 1.86 only introduces cygwin target but not std support. Maybe I should wait for 1.87 :(

ognevny (approved):
thanks for working on this!

AFAIK std support will be shipped in Rust 1.87.0. what about host tools? do you have plans on supporting them for cygwin (at least rustc and cargo)?

@mati865 one question: will it compile on aarch64? probably it worth to gate for x86_64 only

ognevny:
oh, I forgot to check: why llvm from cygwin (msys) is not in makedepends?

Berrysoft:
> do you have plans on supporting them for cygwin

Not now. It seems a little hard for me to build host tools on cygwin.

> will it compile on aarch64?

I have tried and succeeded.

> why llvm from cygwin (msys) is not in makedepends?

It's a cross-compile target so only needs the host llvm, I think.

ognevny:
unfortunately bootstrap compiler doesn't support cygwin yet. to resolve this, you have to set _bootstrapping=no or wait until 1.87.0 release

Berrysoft:
Just wait, then. Anyway we need to wait for 1.87:)

ognevny:
turns out you need to use msys llvm... but I don't know whether `llvm-config = "$MSYS2_PREFIX/bin/llvm-config.exe` will work, but it worth to try

mingw-w64-rust/config.toml
```diff
@@ -115,3 +115,11 @@ default-linker = "wasm-ld.exe"
...
 profiler = false
 wasi-root = "$MINGW_PREFIX/share/wasi-sysroot"
+
+[target.x86_64-pc-cygwin]
+cc = "$MSYS2_PREFIX/bin/gcc.exe"
+cxx = "$MSYS2_PREFIX/bin/g++.exe"
+ar = "$MSYS2_PREFIX/bin/ar.exe"
+ranlib = "$MSYS2_PREFIX/bin/ranlib.exe"
+linker = "$MSYS2_PREFIX/bin/gcc.exe"
+default-linker = "x86_64-pc-msys-gcc.exe"
```
ognevny, line 125: try
```suggestion
default-linker = "x86_64-pc-msys-gcc.exe"
llvm-config = "$MINGW_PREFIX/bin/llvm-config.exe"
```

MehdiChinoune:
I suggest to open this PR in MSYS2-packages repo.

Berrysoft:
> I suggest to open this PR in MSYS2-packages repo.

It's not possible to build rustc & cargo targeting MSYS2. This is a cross-compile target.

Berrysoft:
> the same error again. I suggest to put msys llvm in makedepends and set `llvm-config = "$MSYS2_PREFIX/bin/llvm-config.exe"` for cygwin target

That's *not* right. The MinGW rustc should link to MinGW LLVM. Rust std *doesn't* depend on LLVM, it's the compiler does.

See the comments here: https://github.com/rust-lang/rust/blob/master/bootstrap.example.toml#L857-L860

Actually I don't think it will be used. I should remove all entries expect `linker` in `[target.x86_64-pc-cygwin]` because we are not building LLVM here.

ognevny:
> Actually I don't think it will be used. I should remove all entries expect `linker` in `[target.x86_64-pc-cygwin]` because we are not building LLVM here.

I don't believe this will fix an issue...

Berrysoft:
Let me try this one.

ognevny:
lots of libc crate compile errors. worth a wait for 1.87.0 (15th of may) to try again :)

Berrysoft:
It's right now, I think. The std could not be built because... it is just incomplete in 1.86.

mingw-w64-rust/config.toml
```diff
@@ -115,3 +115,14 @@ default-linker = "wasm-ld.exe"
...
 wasi-root = "$MINGW_PREFIX/share/wasi-sysroot"
+
+[target.x86_64-pc-cygwin]
+cc = "$MINGW_PREFIX/bin/clang.exe"
+cxx = "$MINGW_PREFIX/bin/clang++.exe"
+ar = "$MINGW_PREFIX/bin/llvm-ar.exe"
+ranlib = "$MINGW_PREFIX/bin/llvm-ranlib.exe"
+linker = "x86_64-pc-msys-gcc.exe"
+default-linker = "x86_64-pc-msys-gcc.exe"
+llvm-config = "$MINGW_PREFIX/bin/llvm-config.exe"
```
mati865, line 126: I'm pretty sure this is unnecessary since this is just a target, not the host.

mati865:
I kinda miss the point here, why should MSYS2 add cross-compile target from native to Cygwin/MSYS2?

Berrysoft:
Because I want to package the latest fish-shell, and the only way is to cross-compile from native to msys2.

mati865:
I don't think this is the right way. MSYS2 packages have never depended on other environments.  It should be possible to build a host toolchain for MSYS2.

Berrysoft:
>  It should be possible to build a host toolchain for MSYS2.

Maybe, but I don't have time to explore that:)

ognevny:
@MehdiChinoune ????

MehdiChinoune:
> @MehdiChinoune why did you close PR?

Because It doesn't make sense, want rust for MSYS then build it there, especially after llvm was updated to 20.1.x

ognevny:
> > @MehdiChinoune why did you close PR?
> 
> Because It doesn't make sense, want rust for MSYS then build it there, especially after llvm was updated to 20.1.x

but why do you prevent from building cross-target here? probably rustc and cargo will require patches to be build as host tools

MehdiChinoune:
> but why do you prevent from building cross-target here? probably rustc and cargo will require patches to be build as host tools

MINGW-packages are for native packages not for msys2. Even for msys2 only those required for MINGW-packages are generally allowed. Adding new target will just complicate things and waste time and resources.

jeremyd2019:
I'm trying to get a working cross-to-Cygwin rust in https://github.com/jeremyd2019/MINGW-packages/commits/refs/heads/rust-beta-cygwin/ but I don't know that it's really relevant to include in MINGW-packages, except maybe as a step in bootstrapping an MSYS2-packages/rust once that's possible.

jeremyd2019:
So far, the "trick" seems to be that to use /usr/bin/gcc it needs a wrapper to translate to POSIX paths.  (using clang might work, but it would a) need patches from MSYS2-packages/llvm, and b) need to specify sysroot somehow (it failed building compiler-builtins due to not finding stdlib.h)).

```c
#define _GNU_SOURCE 1
#include <ctype.h>
#include <stdlib.h>
#include <string.h>
#include <sys/cygwin.h>
#include <unistd.h>

#define iswinabspath(p) (isalpha((unsigned)*(p)) && (p)[1] == ':' && \
                         ((p)[2] == '/' || (p)[2] == '\\'))

int main(int argc, char **argv)
{
        char newcmd[260] = "/usr/bin/";
        char *newargv[argc + 1];
        newargv[0] = strcat (newcmd, basename (argv[0]));
        for (int i = 1; i < argc; ++i)
        {
                if (iswinabspath (argv[i]))
                        newargv[i] = cygwin_create_path (CCP_WIN_A_TO_POSIX, argv[i]);
                else
                        newargv[i] = argv[i];
        }
        newargv[argc] = NULL;
        execv(newargv[0], newargv);
        return 42;
}
```

```bash
  if [[ ${CARCH} != i686 ]]; then
    /usr/bin/gcc -o gcc.exe wrapper.c
    for dst in g++ ar ranlib x86_64-pc-msys-gcc; do
      ln -s gcc.exe ${dst}.exe
    done
  fi
```

Berrysoft:
@MehdiChinoune it's a pity that you close this PR... it's far more complicated to build a host toolchain for MSYS2. I would call for people that are interested in this work to try and contribute to the dependencies that needed by rustc.

jeremyd2019:
FYI, https://gist.github.com/jeremyd2019/27f079c5134390353123543f58655f7c if you want to be able to try beta easily.

jeremyd2019:
https://github.com/jeremyd2019/MINGW-packages/commit/30349b4498c371a5cb1b5e5ea10508aee59d9a6c built successfully

Berrysoft:
I would also like to point out that there is also a way to cross-compile to MSYS2 with official **nightly** rustc: `-Zbuild-std`. However, due to different naming of target triple and lib name `libcygwin.a`, you need some adjustments. The linker could be changed by env vars or cargo configs, and `libcygwin.a` could be a linker script redirect to `libmsys-2.0.a`.

Berrysoft:
@jeremyd2019 I don't quite understand your `wrapper.c`. I didn't face the problems about compiler-builtins. Previously I use `-Zbuild-std` or the rustc I built myself. They could find the compilers correctly.

I also package fish-shell in my own repo: https://github.com/Berrysoft/fish-msys2/ and there are no such workarounds needed.

Kreijstal:
> > I suggest to open this PR in MSYS2-packages repo.
> 
> It's not possible to build rustc & cargo targeting MSYS2. This is a cross-compile target.

maybe wait till mrustc supports it?

Berrysoft:
rust 1.87 has been released.

ognevny:
yes, I just forgot to merge https://github.com/msys2/MINGW-packages/pull/24268

mingw-w64-rust/bootstrap.toml
```diff
@@ -110,3 +110,14 @@ linker = "$MINGW_PREFIX/bin/wasm-ld.exe"
...
+
+[target.x86_64-pc-cygwin]
+cc = "$SRCDIR/gcc.exe"
+cxx = "$SRCDIR/g++.exe"
+ar = "$SRCDIR/ar.exe"
+ranlib = "$SRCDIR/ranlib.exe"
+linker = "$SRCDIR/x86_64-pc-msys-gcc.exe"
+default-linker = "x86_64-pc-msys-gcc.exe"
+llvm-config = "$MINGW_PREFIX/bin/llvm-config.exe"
+sanitizers = false
```
ognevny, line 122:
```suggestion
```
this lines can be removed

mati865:
You can build Rust version N with versions N-1 and N. So you can use official beta host toolchain (bootstrap process will download it if you don't configure it to use local one) to cross-compile Cygwin target (or host toolchain once all the bits are merged - Cargo support is still missing).

Berrysoft:
You might want to use this one as stage 0: https://github.com/Berrysoft/rust/releases/tag/1.89.0-dev

It adds the "fork_no_reload" patch to complete building. Although the patch is not safe, it should be OK just to compile rustc. You can reproduce that fork bug when trying to build a stage 2 rustc.

jeremyd2019:
hitting some stupidity in my attempted cross build, might be helpful to somebody in future:
1. complaint about not finding libLLVMTargetX86MC, due to bootstrap being Windows rather than Cygwin, but getting a Cygwin `-L/usr/lib`, hack in compiler/rustc_llvm/build.rs along the lines of
   ```
            println!("cargo:rustc-link-search=native={stripped}");
            println!("cargo:rustc-link-search=native=c:\\test\\msys64{stripped}");
   ```
   I'm thinking another wrapper exe to translate `-L<PATH>` coming out of llvm-config might be better than hacking at this.
2. multiple definitions linker errors on llvm-related stuff.  I switched to linking to llvm shared instead of static
3. panic that it couldn't find x86_64-pc-windows-gnullvm\\stage1-rustc\\x86_64-pc-cygwin\\release\\rustc-main, fix to add cygwin to bootstrap/src/utils/shared_helpers.rs:
   ```
   /// Given an executable called `name`, return the filename for the
   /// executable for a particular target.
   pub fn exe(name: &str, target: &str) -> String {
       if target.contains("windows") || target.contains("cygwin") {
           format!("{name}.exe")
   ```

I'm currently hacking at cargo's Cargo.lock to try to get the updated crates that build for Cygwin

mati865:
> I'm thinking another wrapper exe to translate -L\u003CPATH> coming out of llvm-config might be better than hacking at this.

Maybe a patch to llvm-config to output the desired format? I recall LLD has something in this spirit for arguments quoting.

Berrysoft:
@jeremyd2019 Congratulations! I have also met the first two problems. The third problem is the contrary. I had to revert the patch. I think it relates to the python you use. I used the MSYS2 one.

Berrysoft:
> Maybe a patch to llvm-config to output the desired format? I recall LLD has something in this spirit for arguments quoting.

After a successful cross-compiled bootstrap completes, the output compiler could be stage-0 host tool. Then no cross-compiling is needed. So I think it's OK to keep it as-is. The dirty work is only in the beginning.

Berrysoft:
How did you run `x`? I think `./x install` or `./x dist` or `./x build --stage 2` should reproduce the fork bug. The stage 0 compiler uses dlfork so `stage 0 -> stage 1` doesn't trigger the bug. If you use master branch, `stage 1 -> stage 2` should trigger it.

Berrysoft:
Well... If you build your stage 2 compiler successfully, I think I also need to try again then...

jeremyd2019:
My `MSYS=winsymlinks:sys` setting is because bootstrap wanted to make a symlink that deepcopy couldn't do (target is a parent of the link location).  Are you by chance using native or nativestrict links?  I've run into a case before where those cause Windows to do funny things with binaries when there is some symlink slight-of-hand going on.

Berrysoft:
I use nativestrict and it works fine.

Berrysoft:
I packaged a new one for rust nightly. This one is on master and I didn't applied any `dlfork` workarounds.

https://github.com/Berrysoft/rust/releases/tag/1.89.0-dev-2

PR #23940, opened 2025-04-11
tulip: Update to 6.0.1 and enable aarch64 builds

hmartinez82:
image


mingw-w64-tulip/PKGBUILD
```diff
@@ -79,10 +84,11 @@ build() {
     _extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
   fi
 
-  if [[ ${CC} == gcc ]]; then
-    CXXFLAGS+=" -Wno-template-id-cdtor"
+  if [[ ${MINGW_PACKAGE_PREFIX} == *-clang-* ]]; then
+    CXXFLAGS+=" -Wno-dtor-name"
+    _extra_config+=("-DCLANG=ON")
```
hmartinez82, line 89: Question? Wasn't CMake supposed to already define this? If don't do this here then this `IF` https://github.com/msys2/MINGW-packages/blob/b9b0d3a4165632dea9a6b8a0384debf5489c22de/mingw-w64-tulip/003-fix-clang-omp-link.patch#L20 doesn't return true.

MehdiChinoune: CLANG is not a variabe in CMake language, when checking for Clang compiler one should use `if(CMAKE_C_COMPILER_ID STREQUAL Clang)` (or CMAKE_CXX_COMPIELR_ID in case of C++).

MehdiChinoune: Unless tulip is defining CLANG somewhere.

hmartinez82: Ahh ok. I guess so many projects uses that idiom and defining `CLANG` that I thought it was a built-in feature 😅

PR #23951, opened 2025-04-12
python-sphinxcontrib-programoutput: update to 0.18

podsvirov:


podsvirov:
@Biswa96, thanks for improve package info!

PR #23966, opened 2025-04-13
meld: Update to 3.23 and give the executable an icon

hmartinez82:
Requires https://github.com/msys2/MINGW-packages/pull/23965


hmartinez82:
@Biswa96 omg, I thought it was the total opposite 😅, hence the draft here.

OK. Now I know for the future :) I'll wait until rcedit it deployed then mark this as ready for reviews.

PR #23974, opened 2025-04-14
soil: Enable aarch64 builds

hmartinez82:


hmartinez82:
🤔 I've never seen these errors before. An I/O error with all the runners?

hmartinez82:
@MehdiChinoune wait, what??
I build this locally in both UCRT64 and CLANGARM64, what on earth is going on? Is this something very recent that I'll have to pay attention to in other packages?

MehdiChinoune:
> @MehdiChinoune wait, what?? I build this locally in both UCRT64 and CLANGARM64, what on earth is going on? Is this something very recent that I'll have to pay attention to in other packages?

Well, previously gcc was installed with base-devel before It was removed and made required to be included in `makedepends()`. The job failed because the PKGBUILD wasn't updated.

hmartinez82:
Ahhh ok. Phew. Thank you :)

PR #23983, opened 2025-04-15
_blindly_ enable some clangarm64

ognevny:
probably some packages obviously don't support ARM64, but I don't have much time to recheck it carefully

MinHook: unknown (at least I couldn't find it) compile error. also requires gcc to be installed with gcc-compat.
alure: source checksums changed
openocd: `__asm__` is x86 specific
bscan-spi-bitstreams: depends on openocd
creduce: badly maintained, llvm related problems
editorconfig-qtcreator: CMake specfific issues
efl: mono is required
ffcall: looks like need a small code change
```
  avcall-arm64.s:6:8: error: expected absolute expression
   .type avcall_call,%function
         ^
  avcall-arm64.s:324:2: error: unknown directive
   .size avcall_call,.-avcall_call
   ^
```


MehdiChinoune:
I don't think one PR is the right way to do, It will take much time because of fix after another.
Maybe enable some packages in separate branches for each package where other environments are disabled, and then push the ones succeed to build.

ognevny:
> I don't think one PR is the right way to do, It will take much time because of fix after another. Maybe enable some packages in separate branches for each package where other environments are disabled, and then push the ones succeed to build.

ok, I have some already built successfully. I'll start pushing them later

MehdiChinoune:
This force-push after another will just waste resources/time. I think we should appreciate that GitHub/Microsoft is offering Actions for free.
Please take another approach that doesn't require force-pushing after another and without running jobs for other environments.

PR #24022, opened 2025-04-17
[new package] eslint 9.24.0

podsvirov:


ognevny:
yeah, new optdepend for Zed. thanks

PR #24044, opened 2025-04-20
ffts: Update to r818 and enable aarch64 builds

hmartinez82:


mingw-w64-ffts/PKGBUILD
```diff
@@ -39,6 +54,7 @@ build() {
   cd build
   MSYS2_ARG_CONV_EXCL=-DCMAKE_INSTALL_PREFIX= cmake \
     -G "MSYS Makefiles" \
```
Biswa96, line 56: Would it be possible to use ninja now?

hmartinez82: Let me try

hmartinez82: It worked, and the pkg structure stayed the same.

PR #24046, opened 2025-04-20
fzf: Update to 0.61.2 and enable aarch64 builds

hmartinez82:


mingw-w64-fzf/PKGBUILD
```diff
@@ -34,7 +47,7 @@ build() {
   export CGO_CXXFLAGS="${CXXFLAGS}"
   export GOFLAGS="-trimpath -mod=readonly -modcacherw -buildvcs=false"
   case "${CARCH}" in
-    i686|x86_64)
+    i686|x86_64|aarch64)
       GOFLAGS+=" -buildmode=pie"
```
MehdiChinoune, line 51: Is there other arch? I think the condition should be removed.

hmartinez82: I'll remove it

PR #24062, opened 2025-04-20
safetyhook: update to 0.6.6

llm96:


MehdiChinoune:
`gcc-libs` is missing from `depends()`

PR #24064, opened 2025-04-21
irrlicht: Enable aarch64 builds

hmartinez82:


ognevny:
please try to add `https://github.com/pnggroup/libpng.git` into source() field. it's common to not do vendor steps from prepare() or build()

```bash
if [[ ${MSYSTEM} == CLANGARM64 ]]; then
  source+=("git+https://github.com/pnggroup/libpng.git#tag=${_pngver}") # _pngver probably should be set manually
  sha256sums+=('aaaaaaaaaaaaaaaaaaaaaaa') # replace with actual repo checksum
fi
```

ognevny:
> - append to the source in an imperative way

[example](https://github.com/msys2/MINGW-packages/blob/af0156c1676292b7ec6abfec3182883848a189e6/mingw-w64-rust/PKGBUILD#L69)

> - have "if" outside the functions? wow, this is really new to me

PKGBUILD is just a bash script, so there is no special restrictions on having a script outside of functions

ognevny:
> Ah. I just saw your comment `pngver probably should be set manually`

yeah, that's the actual problem with my suggestion...

hmartinez82:
@ognevny it worked ^_^

PR #24065, opened 2025-04-21
isa-l: Update to 2.31.1 and enable aarch64 builds

hmartinez82:


mingw-w64-isa-l/PKGBUILD
```diff
@@ -36,13 +50,10 @@ build() {
...
   ../${_realname}-${pkgver}/configure \
     --prefix=${MINGW_PREFIX} \
-    --build=${MINGW_CHOST} \
-    --host=${MINGW_CHOST} \
-    --target=${MINGW_CHOST} \
     --enable-static \
     --enable-shared
 
-  make
+  make -j1
```
MehdiChinoune, line 56: I prefer `make || make -j1`

hmartinez82: Done :)

PR #24078, opened 2025-04-22
python-cx-freeze: update to 8.2.0 (and add dependencies)

marcelotduarte:


marcelotduarte:
Should I submit a PR with the new packages and then one with cx_Freeze?

PR #24079, opened 2025-04-23
New package mingw-w64-mylibrary

ProfessorNavigator:
New package MyLibrary (home librarian)


Biswa96:
I am not sure if personal projects should be added as a package. It is only present in AUR.

PR #24087, opened 2025-04-24
powersession-rs: Update to r15 since 0.1.11 and enable aarch64 builds

hmartinez82:


ognevny:
of it won't be so simple here. you must update windows-rs first

hmartinez82:
@ognevny I tried, but I can't, I can't even update to 0.39.0 :( because its release notes literally starts with https://github.com/microsoft/windows-rs/releases/tag/0.39.0 and this project uses `alloc`.

That would be something that only Powersession-rs' developer could try to update given I'm not a Rust developer.

ognevny:
I'll try to do something with it. but not soon, I have much things to do in the university...

ognevny:
great, upstream developers decides to do updates themselves

ognevny:
one more thing to do: you need to add `depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")`, because *-gnullvm targets links libunwind dynamically for this binary. that is a result of an external library being linked into an executable (openssl for powersession)

Biswa96:
There is a new version of this project https://github.com/Watfaq/PowerSession-rs/releases/tag/v0.1.12. The pkgver function could be removed.

PR #24119, opened 2025-04-27
arrow python-pyarrow: Update to 20.0.0

assignUser:


assignUser:
Thanks for the quick merge!

PR #24138, opened 2025-04-29
notepad++: Update to 8.8

heljkon:
see ReleaseNotes https://notepad-plus-plus.org/news/v88-we-are-with-ukraine/


Biswa96:
Please remove the actual patch file if is not required.

Biswa96:
I meant `rm mingw-w64-notepad++/002-windres-codepage.patch`

PR #24151, opened 2025-05-01
python-lief: update to 0.16.5

marcelotduarte:


lazka:
thanks! I tried and failed multiple times.

lazka:
Just a note that this drops the C++ library part. But I assume that is not needed.

marcelotduarte:
LIEF generates static and dynamic libraries separately upstream, and conda also generates the dynamic one, as a separate package. In other words, only the python-lief package has always been available here. But if necessary, we can generate a liblief package together with python-lief.

PR #24154, opened 2025-05-01
new package: python-pyodbc

RoDuth:
Python ODBC bridge

### DIFF
--- a/mingw-w64-python-pyodbc/PKGBUILD
+++ b/mingw-w64-python-pyodbc/PKGBUILD
@@ -1,0 +1,49 @@
+# Maintainer: Ross Demuth <rossdemuth123@gmail.com>
+
+_realname=pyodbc
+pkgbase=mingw-w64-python-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=5.2.0
+pkgrel=1
+pkgdesc="ODBC module for Python"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/mkleehammer/pyodbc'
+license=('spdx:MIT-0')
+depends=("${MINGW_PACKAGE_PREFIX}-python")
+makedepends=("${MINGW_PACKAGE_PREFIX}-python-build"
+             "${MINGW_PACKAGE_PREFIX}-python-installer"
+             "${MINGW_PACKAGE_PREFIX}-python-setuptools"
+             "${MINGW_PACKAGE_PREFIX}-cc")
+options=('!strip')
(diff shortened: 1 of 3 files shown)


mingw-w64-python-pyodbc/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+pkgname=("${MINGW_PACKAGE_PREFIX}-python-${_realname}")
+pkgver=5.2.0
+pkgrel=1
+pkgdesc="ODBC module for Python"
+arch=('any')
+mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
+url='https://github.com/mkleehammer/pyodbc'
+license=('MIT')
+depends=("${MINGW_PACKAGE_PREFIX}-python"
+         "${MINGW_PACKAGE_PREFIX}-unixodbc")
```
lazka, line 14: Is unixodbc actually needed? It's not linking against it and the headers are provided by mingw-w64 from what I see.

RoDuth: No your right, shouldn't. depend.  I believe its still needed to supply `SQL_OV_ODBC3_80` for build though.

RoDuth: ...actually I'm now thinking this approach RoDuth/MINGW-packages@b1f4a96 may be preferable?  Happy to close this and open a fresh PR if you like.

lazka: If that works, it does look nicer at least. Feel free to force push here.

What's the `#if (ODBCVER >= 0x0380)` about, i.e. why is it needed?

RoDuth: (with the caveat that c/cpp is not my strong point!) it comes down to mingw-w64-headers defines `ODBCVER 0x0351` and doesn't define `SQL_OV_ODBC3_80` at all which causes `defaultVersion = (SQLPOINTER)SQL_OV_ODBC3_80;` (pyodbc/src/pyodbcmodule.cpp line 310) to fall over if it is reached. unixodbc on the other hand has both `#define ODBCVER 0x0380` and `#define SQL_OV_ODBC3_80  380UL`.

So the options seem to be use the unicODBC headers or avoid line 310 of pyodbcmodule.cpp (and `#include <windows.h>`).

To be honest I'm not sure which is preferable and would appreciate your opinion.  (I guess the other option would be to update mingw-w64 to support newer ODBC versions but I'm not the one for that!)

lazka: I see, thanks. I've filed https://github.com/mingw-w64/mingw-w64/issues/93

Maybe just defining it would be enough, not sure.

lazka:
cleaned/split it up a bit. does it look OK?

RoDuth:
Looks perfect, thanks for the effort.

PR #24155, opened 2025-05-01
podofo: Fix build with Clang 20 and enable aarch64 builds

hmartinez82:


ognevny:
I remeber trying upstreaming my patches for this package, it was not so easy :D. anyway thanks

PR #24159, opened 2025-05-01
osgearth: Update to 3.7.2 and enable aarch64 builds

hmartinez82:
image


mingw-w64-osgearth/0003-osg-plugin-prefix.patch
```diff
@@ -0,0 +1,12 @@
+diff -bur osgearth-osgearth-3.7.2-o/cmake/osgearth-macros.cmake osgearth-osgearth-3.7.2/cmake/osgearth-macros.cmake
+--- osgearth-osgearth-3.7.2-o/cmake/osgearth-macros.cmake	2025-05-01 06:16:07.023054400 -0600
++++ osgearth-osgearth-3.7.2/cmake/osgearth-macros.cmake	2025-05-01 06:16:26.286111700 -0600
+@@ -45,7 +45,7 @@
+     
+     # no prefixes on plugin library files, please.
+     # without this, plugins will have names like "libosgdb_osgearth.so" and OSG doesn't want the "lib"
+-    set_target_properties(${MY_TARGET} PROPERTIES PREFIX "")
++    set_target_properties(${MY_TARGET} PROPERTIES PREFIX "mingw_")
```
ognevny, line 9: just my curiosity: why setting this?

hmartinez82: Because otherwise, it doesn't follow the pattern that OpenSceneGraph is using and it fails to find any osgEarth plugins

PR #24165, opened 2025-05-02
cvise: fix paths in cvise.py

oltolm:
cvise only worked when compiled locally because `cvise.py` contains absolute paths. I fixed this problem.


mingw-w64-cvise/fix-paths.patch
```diff
@@ -0,0 +1,80 @@
...
++                path = shutil.which(prog, path=search)
+ 
+         if not path:
+             path = shutil.which(prog)
+@@ -93,6 +100,10 @@
+ 
+     # Special case for clang-format
+     programs['clang-format'] = '@CLANG_FORMAT_PATH@'
++    if not os.path.exists(programs['clang-format']):
++        programs['clang-format'] = shutil.which('clang-format', path=script_path)
```
lazka, line 75: why the `path=script_path`?

lazka:
While this works it will never be testable locally as the cmake things will win. Do you mind if I adjust it a bit?

oltolm:
> While this works it will never be testable locally as the cmake things will win (or breakage gets only noticed in the repo). Do you mind if I adjust it a bit?

No I don't mind.

lazka:
I added a small hack to make bash files work too now

PR #24187, opened 2025-05-06
ossia score: update to 3.5.2

jcelerier:


hmartinez82:
Maybe it will work now with the latest GCC patches @lazka ?

jcelerier:
>  cc1plus.exe: out of memory allocating 65536 bytes


:|

lazka:
#24747

PR #24189, opened 2025-05-06
gcc: Backport PCH fix from master

lhmouse:
Reference: https://github.com/gcc-mirror/gcc/commit/16e301adf7b653c72d5c3cae1c7287877f3f20ff


lhmouse:
This restores a patch that was removed in https://github.com/msys2/MINGW-packages/commit/f59921184b35858d4ceb91679578de0d62475cbf.

@MehdiChinoune

Issue #24148, opened 2025-04-30
gcc 15.1.0 internal compiler error building Babl


Wormnest:
### Description / Steps to reproduce the issue

As part as our CI for GIMP we build Babl for MINGW32 if the Windows-installer label is set.

Build babl master for MINGW32 after updating MSYS2.

### Expected behavior

Build succeeds.

### Actual behavior

Build fails, reporting several internal compiler errors. 

### Verification

- [x] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Whatever win32-creiter-4 vJWzEqDv uses

### MINGW environments affected

- [ ] MINGW64
- [x] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANGARM64

Wormnest:
Ah, forgot to link to the job with the build details: [https://gitlab.gnome.org/GNOME/gimp/-/jobs/5029270](https://gitlab.gnome.org/GNOME/gimp/-/jobs/5029270)

Biswa96:
I think the internal compiler error should be reported to gcc bugzilla. Also, a minimal example to reproduce the issue would be very helpful. Thank you.

lhmouse:
Please provide a **pre-processed** source file to reproduce the issue.

Wormnest:
I totally understand the need for a minimal source file to make it easier to reproduce. However, I personally don't have MINGW32 installed anymore to test it locally and we have a huge surge in issues after the release of GIMP 3.0 that take up all my limited time. So, I won't be able to look at this anytime soon.

Biswa96:
I can reproduce the issue in my setup.
1. Install dependencies using babl PKGBUILD and `makepkg-mingw -seoc` command.
2. git clone https://gitlab.gnome.org/GNOME/gimp
3. meson setup builddir
4. ninja -C builddir

preprocess file: [test.c.txt](https://github.com/user-attachments/files/19982828/test.c.txt)
command for ICE: `gcc -O2 -gcodeview -c test.c`

lazka:
thanks, c-reduced to:

```
// gcc -O1 -gcodeview -c test.c
void a();
static void b();
void c() { b(); }
void b() { a(); }
```

lazka:
I've filed https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120051

maharmstone:
[...]

PR #24205, opened 2025-05-07
gcc: apply codeview fixes

lazka:
See #24148


Jehan:
I see you updated the patch earlier today. Did that work in compiling `babl`? 🤞

lazka:
no, still crashes

PR #24166, opened 2025-05-02
gst-plugins-rs: Enable aarch64 builds

hmartinez82:

PR #24217, opened 2025-05-09
remove aws-lc-rs workaround; gst-plugins-rs: enable aarch64 builds

ognevny:
as described per each commit

aws-lc PR: https://github.com/aws/aws-lc/pull/2381

closes #24166


ognevny:
pushed zed update because why not

PR #24223, opened 2025-05-09
rustup: Enable aarch64 builds

hmartinez82:


mingw-w64-rustup/PKGBUILD
```diff
@@ -5,7 +5,7 @@ pkgver=1.28.1
 pkgrel=1
 pkgdesc="The Rust toolchain installer (mingw-w64)"
 arch=('x86_64')
```
ognevny, line 7: FYI this should be set to ('any')

hmartinez82: oh I was wondering why it worked in my machine 😅

PR #24255, opened 2025-05-12
New Package: dukpy

RoDuth:
Simple JavaScript interpreter for Python


RoDuth:
needs `python-webassets` for full functionality (i.e. `dukpy-install` script).

RoDuth:
retract that `webassets` is an optional extra and not needed (at least in my use case).

PR #24266, opened 2025-05-13
cmakerc: rebuild with cmake 4

podsvirov:


MehdiChinoune:
Does it link to cmake or what?
I mean, why rebuild? Did it stop working, was there an issue?

MehdiChinoune:
This rebuild does not make any sense.
The package contains only two files 
```
/ucrt64/lib/cmake/cmakerc/cmakerc-config.cmake
/ucrt64/share/licenses/cmakerc/LICENSE.txt
```
there is no difference between before and after rebuild
```
diff -Nau cmakerc-config_3.cmake cmakerc-config.cmake
```
So, what's the point?

podsvirov:
> there is no difference between before and after rebuild

Oh, you right!

Please see #24275.

PR #22778, opened 2024-12-09
rust: update to 1.89.0

ognevny:
- perform PGO for rustc via opt-dist tool (addresses #19273):
  - various backports (available in 1.90.0) to match our environment better and fix some bugs
  - do not profile LLVM, because it makes backend performance even worse: https://github.com/rust-lang/rust/issues/144318
  - on UCRT64 do a separate LLVM build with clang+lld so it will be linkable with PGOed rustc
  - bootstrap.toml changes to match requirements for opt-dist (and get some more optimizations)
- CXXFLAGS workaround to actually perform LTO for llvm-wrapper (from rustc_llvm)
- ensure that clang+lld is used for whole build on non-i686

PR #24268, opened 2025-05-14
rust: update to 1.87.0

ognevny:
- 0008-disable-self-contained-for-gnu-targets.patch: refresh
- LLVM 20 is now supported
- extract some changes from https://github.com/msys2/MINGW-packages/pull/22778 (e.g. use more system libs)

config changes:
- config.toml is now bootstrap.toml
- description field was moved
- remove sanitizers field as they're not supported anyway


mingw-w64-rust/PKGBUILD
```diff
@@ -127,14 +124,9 @@ build() {
...
-    # Use $CFLAGS for $OSTYPE only
-    export CFLAGS_${OSTYPE//-/_}="$CFLAGS"
-
-    # Remove unsupported options for target wasm32-*
-    local _unsupported_options="-march=nocona -msahf"
-    export CFLAGS="${CFLAGS#$_unsupported_options}"
-  fi
+  # use supported flags for all targets
+  export CFLAGS="-O2 -pipe -Wp,-D__USE_MINGW_ANSI_STDIO=1"
+  export CXXFLAGS="$CFLAGS"
```
MehdiChinoune, line 129: I don't understand why replacing global flags. I understand the part about wasm32-*, but why remove it altogether. You know those flags were added to get more optimizations and performance.

ognevny: I took that change from #22778 as is because in that PR I tried to make LLVM build as fast as possible. there is no problem to revert the change in this PR

PR #24285, opened 2025-05-16
crystal: update to 1.16.3

HertzDevil:


MehdiChinoune:
Is it possible to provide a prebuilt crystal for aarch64, so we could build it for clangarm64.

HertzDevil:
Not yet

HertzDevil:
However, we are about to arrange our first Windows ARM64 build here, since GitHub CI runners are now available: crystal-lang/crystal#15794

PR #24287, opened 2025-05-16
llvm: split a new package llvm-tools

MehdiChinoune:
This is to minimize toolchain installation size on clang* environments.

### DIFF
--- a/mingw-w64-llvm/PKGBUILD
+++ b/mingw-w64-llvm/PKGBUILD
@@ -10,9 +10,7 @@
 if [[ $MINGW_PACKAGE_PREFIX == *-clang-* ]]; then
   _clangprefix=1
   _compiler=clang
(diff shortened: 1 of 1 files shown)


lazka:
Is it on purpose that the tools contain the import libs and llvm-config?

lazka:
An it's a dependency anyway. Still seems weird.

lazka:
and clang also depends on it, so I don't understand the "This is to minimize toolchain installation size" part. It will always be installed.

lazka:
ah, the llvm dep is gone. ok, never mind.

MehdiChinoune:
> Is it on purpose that the tools contain the import libs and llvm-config?

I just let cmake install all what's in tools directory, which includes llvm shared libraries.

PR #24292, opened 2025-05-17
mingw-w64-wxwidgets3.2: Edit/add provides and conflicts

stahta01:
In preparation for wxWidgets 3.3.x release package

### DIFF
--- a/mingw-w64-wxwidgets3.2/PKGBUILD
+++ b/mingw-w64-wxwidgets3.2/PKGBUILD
@@ -16,7 +16,7 @@ pkgname=("${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-common"
          "${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-qt-libs"
          "${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-qt")
 pkgver=${_wx_basever}.8.1
-pkgrel=3
+pkgrel=4
 pkgdesc="A C++ library that lets developers create applications for Windows, Linux and UNIX (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw64' 'ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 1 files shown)


mingw-w64-wxwidgets3.2/PKGBUILD
```diff
@@ -454,7 +456,7 @@ package_wxwidgets3.2-common() {
 
 package_wxwidgets3.2-msw-cb_headers() {
   pkgdesc="private wxWidgets ${_wx_basever} MSW headers needed by Code::Blocks (mingw-w64)"
-  depends=("${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-common")
+  optdepends=("${MINGW_PACKAGE_PREFIX}-wxwidgets${_wx_basever}-common")
```
lazka, line 459: These headers depend on headers in common, so I don't understand this change.

mingw-w64-wxwidgets3.2/PKGBUILD
```diff
@@ -374,6 +374,8 @@ package_wxwidgets3.2-msw-libs() {
 
 package_wxwidgets3.2-gtk3() {
   pkgdesc="GTK+3 implementation of wxWidgets ${_wx_basever} C++ GUI API (mingw-w64)"
+  provides=("${MINGW_PACKAGE_PREFIX}-wxconfig-gtk3")
+  conflicts=("${MINGW_PACKAGE_PREFIX}-wxconfig-gtk3")
```
lazka, line 378: What is wxconfig-gtk3?

stahta01: These changes allow upgrading the wxWidgets package version to be done with less user manual package operation.

stahta01:
To all: wxWidgets 3.3.0 is a few days late it was due to be released on May 15 2025; when it is released someone not me will think it is a good idea to create a package for it. I plan to wait till 3.3.1 because the 3.3 series is a development series and it is normally more stable on the .1 version than the .0 version. But, 3.3.0 has midnight support and someone will want it. Therefore I think adding these changes that make the upgrade easier is a good idea.

mingw-w64-wxwidgets3.2/PKGBUILD
lazka: Now that 3.3 is in the repo too, I think see what you mean. So you want to be able to depend on either 3.2 and 3.3 with some external package?

mingw-w64-wxwidgets3.2/PKGBUILD
stahta01: The wxWidgets 3.3 and 3.2 do not conflict with each other therefore I can either falsely say they conflict or falsely say they opt depnds on common in order to make switching from wx 3.2 to wx 3.3. Do you wish me to change to conflicts instead? They both will work for me.

stahta01: I have decided to go with just conflicts; but, I keep missing an conflict mainly because of typos. But, I am getting close on my local PC I hope to push the changes later today.

PR #24294, opened 2025-05-17
codeblocks: patches gcc 15.x bug issue

stahta01:
The provides helps to avoid future run-time issues with CB plugins like the FORTRAN plugin using a different base version of wxWidgets. The CB plugin needs to depend on the wxWidgets versioned of the Code::Blocks IDE. 

The maintainer rejected fixing the problem before it happens.

### DIFF
--- a/mingw-w64-codeblocks/PKGBUILD
+++ b/mingw-w64-codeblocks/PKGBUILD
@@ -6,7 +6,7 @@ _realname=codeblocks
 pkgbase=mingw-w64-${_realname}
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
 pkgver=25.03
-pkgrel=2
+pkgrel=3
 pkgdesc="A free C, C++ and Fortran IDE (mingw-w64)"
 arch=('any')
 mingw_arch=('ucrt64' 'clang64' 'clangarm64')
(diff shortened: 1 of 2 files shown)


mingw-w64-codeblocks/PKGBUILD
```diff
@@ -75,6 +79,7 @@ check() {
 }
 
 package() {
+  provides=("${MINGW_PACKAGE_PREFIX}-${_realname}-wx${_wx_basever}")
```
MehdiChinoune, line 82: 1. There is only one package, so this declaration should be in top of PKGBUILD.
2. I don't see any benefit from this `provides`. Why?

stahta01: I will move it to the top; I am testing code::blocks with wxWidgets 3.3 git master and if the CB plugins are not depending on the correct wxwidgets version of codeblocks you get a weird run-time error.

MehdiChinoune: What plugins?, there is only plugin which is Fortran project.
I don't think a developer would build codeblocks and its plugins against different wxwidgets versions. It passes through a PR and a review after all.
I am still not convinced by this change.

stahta01: So, you will rather have users complaining about weird errors when they start using wxWidgets 3.3.0 as many will because it adds midnight mode.

MehdiChinoune: Which users, you are imagining a situation that will never happen.
If we build codeblocks against wxwidgets 3.3.0, of course we will build fortran against the same wxwidgets version.

stahta01: I disagree; but, you are allowed to make the discussion.

PR #24299, opened 2025-05-18
Rename gcc-libs to cc-libs

lazka:
They are not provided by gcc in the clang envs, so make this more clear.


lazka:
@MehdiChinoune do you know why it failed?

lazka:
ah, you pushed something

PR #24315, opened 2025-05-20
libarchive: Update to 3.8.0

lazka:


Alexpux:
@lazka error is the same probably that I fix for libevent. 32-bit mingw still use by default 32-bit time_t. I wonder why mingw-w64 developers not switch it by default to 64-bit. So to fix 32-bit build you need add: 
```
  if [ "${CARCH}" = "i686" ]; then
    # Need for 32-bit toolchain to have 64bit time
    CFLAGS+=" -D__MINGW_USE_VC2005_COMPAT"
  fi
```

Alexpux:
@lazka od add this compile flag by default to 32-bit flags in makepkg configs for 32-bit

lazka:
thanks! I'll try later

lazka:
> @lazka od add this compile flag by default to 32-bit flags in makepkg configs for 32-bit

I'm a bot afraid of ABI breaks at this point, unless we rebuild everything.

PR #24326, opened 2025-05-22
cbindgen: update to 0.29.0

ognevny:


ognevny:
I shouldn't have merge it...

PR #24346, opened 2025-05-24
ruff: update to 0.11.11

ognevny:


ognevny:
what's wrong with github? the repo and its commit is here, but something wrong with SSL certificate (??)

PR #24348, opened 2025-05-24
Revert "ng-log: Replacing glog library with its fork"

lazka:
This reverts commit be8c109f2f153e42866c9104b800b5fc56f8d9fb.

Turns out it's not API compatible, so we can only switch once all downstreams are fixed.


lazka:
We can look into switching again, but only if the dependencies are also fixed.

PR #24361, opened 2025-05-25
llama.cpp: Update to b5481 + ggml/whisper

lazka:


lazka:
I had to update whisper to a snapshot to keep it in sync with ggml

PR #24376, opened 2025-05-26
fastfetch: Update to 2.44.0

Suyun114:


Biswa96:
This removed libffwinrt.dll in ucrt due the gcc 15 issue. So, the user experience would be different with ucrt vs clang.

PR #24383, opened 2025-05-27
rav1e: update to 0.8.0

ognevny:
lockfile is provided in upstream tarball

### DIFF
--- a/mingw-w64-rav1e/PKGBUILD
+++ b/mingw-w64-rav1e/PKGBUILD
@@ -3,8 +3,8 @@
 _realname=rav1e
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=0.7.1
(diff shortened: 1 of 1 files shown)


ognevny:
ucrt64 static lib is not stripped: `D:\M\msys64\ucrt64\bin\strip.exe: error: the input file 'ucrt64/lib/librav1e.a(std-8f4200ddc3508613.std.7e39b4733639c7ef-cgu.0.rcgu.dwo)' has no sections`

lazka:
librsvg also is not stripped afair, but there it was another error

ognevny:
> librsvg also is not stripped afair, but there it was another error

that was clang issue IIRC

Morilli:
For what it's worth using `llvm-strip` allows stripping the `librav1e.a` file. I assume some of the files contained in the .a are not supported by gnu strip.

ognevny:
hm, thanks for a hint!

PR #24393, opened 2025-05-28
swi-prolog: new version 9.3.24

mgondan:


mgondan:
Basically, because the current version did not work anymore on my machine (reason not clear), and there wasn’t any newer stable release.

lazka:
ok. what didn't work for example, just for future reference.

mgondan:
If I start swipl from the MSYS2 console, it just returns back to the prompt. If I double-click on swipl.exe from the Windows Explorer, I get an error message, telling me that "sem_timedwait" was not found in libswipl-9.dll. I was unsure how to fix this problem, and since it does not seem to appear in 9.3.24, I thought I just submit an update.

PR #24399, opened 2025-05-28
grpc: Rebuild for protobuf 31.0

kou:
See also: #24370


kou:
I haven't confirmed whether this fixes the `protoc --plugin=protoc-gen-grpc=.../grpc_cpp_plugin.exe ...` crash or not. Sorry. But we can confirm it on apache/arrow CI if this is accepted.
See also: https://github.com/apache/arrow/issues/46634

MehdiChinoune:
You could confirm by testing the artifact built by CI in your local repo.

PR #24406, opened 2025-05-30
bzip2: Update to 1.1.0 (git)

Alexpux:
Update bzip2 from git to switch to cmake. Implement cmake scripts for switch between shared/static link


lazka:
Do we know if this is a problem:

```diff
  -/ucrt64/bin/bunzip2.exe
  +/ucrt64/bin/bunzip.exe
   /ucrt64/bin/bzcat.exe
  -/ucrt64/bin/bzcmp
  -/ucrt64/bin/bzdiff
  -/ucrt64/bin/bzegrep
  -/ucrt64/bin/bzfgrep
  -/ucrt64/bin/bzgrep
   /ucrt64/bin/bzip2.exe
   /ucrt64/bin/bzip2recover.exe
  -/ucrt64/bin/bzless
  -/ucrt64/bin/bzmore
```

mingw-w64-bzip2/PKGBUILD
```diff
@@ -4,8 +4,9 @@
 _realname=bzip2
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.0.8
```
lazka, line 7: I think we should use a pkgver() function to get something like 1.0.8.rX.gY. Otherwise we won't notice when there is a new version.

Alexpux:
Other mingw packages dont have in bin folder scripts and probably we dont need it as it bash scripts. MSYS bin folder have it and its enougth if it needed for builds

PR #24408, opened 2025-05-30
double-conversion: Implement cmake config to switch between shared/static link

Alexpux:
Cmake shared/static linking


Biswa96:
Would it be possible to create a pull request in upstream to add this change?

Alexpux:
@Biswa96 I dont know. Mostly ALL cmake packages dont have proper support to link static/shared by default

PR #24413, opened 2025-05-30
winpthreads-stub: Add package to be link without pthreads library

Alexpux:


Alexpux:
@lazka if you just uninstall winpthreads pqckage then tou cannot compile anything because by default toolchain link with libpthread and mingw headers include winpthread includes. So to be able compile you need stubs. So im add empty library and empty headers

lazka:
I see, thanks

lazka:
We no longer use groups.

I've created #24423 for this

lazka:
Now that everything is renamed I've re-applied this.

Alexpux:
Also i think you dont need add to provides packages with -git suffix. It not necessarily

lazka:
We don't know who is using the old name still though.

PR #24418, opened 2025-05-30
ngtcp2: build with openssl support; curl: build with ngtcp2 backend

lazka:
this is possible with openssl 3.5+ and newer ngtcp2 versions


lazka:
ah, thanks, that's a better fix :)

PR #24423, opened 2025-05-30
winpthreads: remove -git suffix from package

lazka:
See #24413


lazka:
> what about renaming all mingw-w64 packages to remove `-git` suffix?

yeah, let's see if this works, and then do it for all the others.

PR #24425, opened 2025-05-31
gstreamer: update to 1.26.2

Biswa96:


ognevny:
I'll try to use our skia to fix its plugin a bit later, it seems possible to implement

PR #24427, opened 2025-05-31
new package: systemc

Kreijstal:
Introduce a new package for SystemC, providing C++ classes and macros for event-driven simulation in hardware modeling.


Kreijstal:
ah, I guess it works only with the git branch

Kreijstal:
@striezel fixed

PR #24456, opened 2025-06-02
luajit: update to 2.1.1748459687

luau-project:
notable changes: restored https on LuaJIT links that were disabled in the previous update.


luau-project:
The update seems to be fine (I ran it locally on my computer on CLANG64). I guess that rerunning the failed job would build nicely.

PR #24464, opened 2025-06-03
octave: update to 10.2.0

mmuetzel:
Remove patch that was applied upstream:
https://github.com/gnu-octave/octave/commit/16ef607e71f21f68c30b8491b1f69a060d4cedf4

No longer disable documentation (for which no rebuild is triggered without that local patch).


mmuetzel:
I downloaded and installed the MINGW64 build artifact from this PR. It is working as expected here.

Issue #24460, opened 2025-06-02
TkImg 2.0.1: stack smash


oscarfv:
### Description / Steps to reproduce the issue

[stacksmash.zip](https://github.com/user-attachments/files/20559478/stacksmash.zip)

On the clang64 environment, unzip the attached file and execute

```
$ pacman -S mingw-w64-clang-x86_64-tkimg
$ wish stack.tcl
starting...
*** stack smashing detected ***: terminated
```

After rebuilding the package with debug symbols:

```
$ gdb --args wish stack.tcl
GNU gdb (GDB) 16.3
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-w64-mingw32".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from wish...
(No debugging symbols found in wish)
(gdb) r
Starting program: C:\apps\msys64\clang64\bin\wish.exe stack.tcl
[New Thread 5224.0x27f8]
[New Thread 5224.0x2ed4]
[New Thread 5224.0x2ef8]
[New Thread 5224.0x4b0]
gdb: unknown target exception 0xc0000409 at 0x7fff2cc54327

Thread 1 received signal ?, Unknown signal.
0x00007fff2cc54327 in __fastfail (code=2) at C:/M/msys64/clang64/include/_mingw.h:632
warning: 632    C:/M/msys64/clang64/include/_mingw.h: No such file or directory
(gdb) bt
#0  0x00007fff2cc54327 in __fastfail (code=2) at C:/M/msys64/clang64/include/_mingw.h:632
#1  __stack_chk_fail () at C:/W/B/src/mingw-w64/mingw-w64-crt/ssp/stack_chk_fail.c:18
#2  0x00007fff2cc51ff1 in CommonMatch (png_ptr=0x0, handle=<optimized out>,
[...]

PR #24478, opened 2025-06-04
Revert "tkimg: Update to 2.0.1", bump epoch

oscarfv:
This reverts commit 05feb3964815b5cddd12724fac086a87f6436ea6.

This version crashes in some common scenarios. See

https://github.com/msys2/MINGW-packages/issues/24460


MehdiChinoune:
It crashes only on clang64 environment, so I don't think that's sufficient to degrade it.

oscarfv:
> It crashes only on clang64 environment, so I don't think that's sufficient to degrade it.

That's an strange thing to say. Is clang64 some type of third-tier environment that we don't care about?

BTW, I didn't check  the test case on the other environments, except for mingw64. Plus, the fact that it does not crash in mingw64 **with the provided test case** does not mean that the bug does not exist there. It could crash (or silently corrupt data) in some other scenario.

Until upstream analyzes the bug and demonstrates that the problem is our toolset and not their code, we must regard the problematic version as utterly broken.

oscarfv:
It is a memory corruption problem, how much more severe can it be?

Upstream being slow to respond to reports is another good reason for reverting to the stable version.

oscarfv:
Yes, it is possible that the fault is in the hardening mechanism. But disabling hardening and seeing the (trivial) test case "work" wouldn't demonstrate anything. OTOH, we build all packages with hardening enabled with no problem. So the line of action that consists on covering the "check engine" light with duct tape does not look sensible to me.

MehdiChinoune:
see https://github.com/msys2/MINGW-packages/blob/f80ad7b66caf29066a9b13e50fe9e50692682495/mingw-w64-emacs/PKGBUILD#L84

oscarfv:
> see
> 
> https://github.com/msys2/MINGW-packages/blob/f80ad7b66caf29066a9b13e50fe9e50692682495/mingw-w64-emacs/PKGBUILD#L84

On that case the hardening system was fine. It was Emacs' fault because it was redefining a C runtime function the wrong way. See https://github.com/msys2/MINGW-packages/issues/17343#issuecomment-2368903501

BTW, that hack on Emacs' s PKGBUILD is no longer needed. I'll submit a PR later.

oscarfv:
The source of the problem was found, it is very likely that a patch may be made. The discussion continues in #24460.

PR #24495, opened 2025-06-06
icu: export C API on MinGW

### LINKED ISSUE #24493: QT6 cannot build icu by cmake.

There is a issue on build program using qt6.I installed msys2 on windows10, and setup mingw64 tools-chain and mingw_x86 icu. But when i tried to build a qt6 program with including <unicode/ucsdet.h>. There's some error on building program. What's wrong with it?

### Expected behavior

~~~cpp
// my code for testing icu working properly
#include "icu_lib.h"
#include <unicode/ucsdet.h>
#include <iostream>
Icu_lib::Icu_lib(QWidget *parent)
    : QWidget{parent}
{}

void Icu_lib::runIcu()
{




    const char *text = "Hello, 世界!";
    char *detected = nullptr;
    if (detectTextEncoding(text, strlen(text), &detected)) {
        std::cout << "Detected encoding: " << detected << std::endl;
        free(detected);
    } else {
        std::cout << "Failed to detect encoding." << std::endl;
    }
    return ;

}

bool Icu_lib::detectTextEncoding(const char *data, int32_t len, char **detected)
{
    UErrorCode status = U_ZERO_ERROR;
    UCharsetDetector *csd = ucsdet_open(&status);
    if (U_FAILURE(status)) return false;

    ucsdet_setText(csd, data, len, &status);
    if (U_FAILURE(status)) return false;

    int32_t matchCount = 0;
    const UCharsetMatch **csm = ucsdet_detectAll(csd, &matchCount, &status);
    if (U_FAILURE(status)) return false;

    if (matchCount > 0) {
        *detected = strdup(ucsdet_getName(csm[0], &status));
        if (U_FAILURE(status)) return false;
    }

    ucsdet_close(csd);
    return true;
}


~~~

### Actual behavior

C:\Users\WIN10\Documents\icu_test\icu_lib.cpp:-1: error: undefined reference to `ucsdet_open_77'
:-1: error: C:/Users/WIN10/Documents/icu_test/icu_lib.cpp:32:(.text+0x1a1): undefined reference to `ucsdet_setText_77'
:-1: error: C:/Users/WIN10/Documents/icu_test/icu_lib.cpp:36:(.text+0x1dc): undefined reference to `ucsdet_detectAll_77'
:-1: error: C:/Users/WIN10/Documents/icu_test/icu_lib.cpp:40:(.text+0x214): undefined reference to `ucsdet_getName_77'
:-1: error: C:/Users/WIN10/Documents/icu_test/icu_lib.cpp:44:(.text+0x24c): undefined reference to `ucsdet_close_77'
:-1: error: collect2.exe: error: ld returned 1 exit status
:-1: error: ninja: build stopped: subcommand failed.

### Verification

### Windows Version

win10 

- [x] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANGARM64

_No response_

mmuetzel:
Probably fixes #24493.

### DIFF
--- a/mingw-w64-icu/0001-Add-visibility-attributes-for-MinGW.patch
+++ b/mingw-w64-icu/0001-Add-visibility-attributes-for-MinGW.patch
@@ -1,0 +1,45 @@
+From 0a92e43809209c210cf77a4d63a255e4093ea73f Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Markus=20M=C3=BCtzel?= <markus.muetzel@gmx.de>
+Date: Fri, 6 Jun 2025 09:42:02 +0200
+Subject: [PATCH] Add visibility attributes for MinGW.
+
+---
+ source/common/unicode/platform.h | 4 ++++
+ source/config/icu.pc.in          | 1 +
+ 2 files changed, 5 insertions(+)
+
+diff --git a/source/common/unicode/platform.h b/source/common/unicode/platform.h
+index 45741a83deb..60be4987eb2 100644
+--- a/source/common/unicode/platform.h
++++ b/source/common/unicode/platform.h
+@@ -780,6 +780,8 @@
+ #elif defined(_MSC_VER) || (UPRV_HAS_DECLSPEC_ATTRIBUTE(__dllexport__) && \
+                             UPRV_HAS_DECLSPEC_ATTRIBUTE(__dllimport__))
+ #   define U_EXPORT __declspec(dllexport)
++#elif defined (_WIN32) || defined (__CYGWIN__)
++#   define U_EXPORT __attribute__ ((dllexport))
+ #elif defined(__GNUC__) || defined(__open_xl__)
+ #   define U_EXPORT __attribute__((visibility("default")))
+ #elif (defined(__SUNPRO_CC) && __SUNPRO_CC >= 0x550) \
+@@ -806,6 +808,8 @@
+                             UPRV_HAS_DECLSPEC_ATTRIBUTE(__dllimport__))
+     /* Windows needs to export/import data. */
+ #   define U_IMPORT __declspec(dllimport)
++#elif defined (_WIN32) || defined (__CYGWIN__)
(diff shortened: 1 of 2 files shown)


mmuetzel:
Apparently, this change leads to a lot of symbols that are no longer being exported from the shared libraries. But the missing symbols from #24493 are exported from `libicuin77.dll` now.

I checked a few of the symbol names that are no longer exported. And afaict, they corresponded to internal functions (that aren't defined in the installed headers and don't have API attributes in the sources).
Also, the package grokker didn't detect any packages that are using those symbols.

lazka:
lgtm

In theory icu-uc.pc could be extended with `Cflags.private: -DU_STATIC_IMPLEMENTATION` so the dllimport is removed for "pkgconf --static" again. Not sure how likely that is that users depended on that.

mmuetzel:
Good point!

I amended the patch to also add that line to the pkg-config files.

hessenfarmer:
Hi,   
since this patch the static libs are not working anymore. therefore we in widelands pinned the icu version to the last version before this however due to the changes in the runtime and gcc16 thi does not work anymore.      
after some digging I beleive we would need the following line addedbefore line 26 in the patch of this PR 

"#elif defined(U_STATIC_IMPLEMENTATION)"
"#   define U_IMPORT"

this should make the nelwy introduced lines 28 and 29 being ignored if a static library is build

could anyone please fix this? @lazka @mmuetzel 
or shall I try to provide the PR

lazka:
@hessenfarmer like this https://github.com/msys2/MINGW-packages/pull/30111 ?

hessenfarmer:
> @hessenfarmer like this #30111 ?

@lazka 
in principle yes. however I finally fond out that our original issue was due to that that we in widelands did not define U_STATIC_IMPLEMENTATION correctly. After setting it properly the current builds 78.3-2 are working fine. However this is only true as long as icu does not need additional static libs to be linked. So to safeguard this for the future I would still deem the proposed change of #30111 necessary although it cannot be tested properly at the moment.

Issue #24494, opened 2025-06-06
Meld not blocking anymore


HazardyKnusperkeks:
### Description / Steps to reproduce the issue

Running meld from command line.

This breaks using meld as git difftool, because the temporary files are deleted before meld actually starts.

### Expected behavior

That the shell is blocked until meld is closed.

### Actual behavior

It doesn't block, e.g. the help text appears after the next command prompt:
```
$ meld --help

bjoern@hades MSYS /e
$ Usage:
  meld-script.pyw                              Beim Start kein Fenster öffnen
  meld-script.pyw <Datei|Ordner>               Versionskontrollvergleich starten
  meld-script.pyw <Datei> <Datei> [<Datei>]    Im Zwei- oder Dreiwegevergleich starten
  meld-script.pyw <Ordner> <Ordner> [<Ordner>] Einen Zwei- oder Dreiwegevergleich starten

```

### Verification

- [x] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-19045

### MINGW environments affected

- [ ] MINGW64
- [ ] MINGW32
- [x] UCRT64
- [ ] CLANG64
- [ ] CLANGARM64

HazardyKnusperkeks:
CC @hmartinez82

hmartinez82:
Did you try using `meld-script.pyw` instead of `meld.exe` ?

`meld.exe` is indeed not blocking, but the script is:

![Image](https://github.com/user-attachments/assets/5687cfcd-e951-4f87-9033-53c610bc40f2)

HazardyKnusperkeks:
Yeah, for now I use the script, but now my gitconfig has to differ between Linux and Windows.

lazka:
> Did you try using `meld-script.pyw` instead of `meld.exe` ?

the main motivation/reason we created the .exe though was that git gui needed an exe and couldn't work with a py script

could we make a meldw.exe maybe? Or is there any other convention for this?

hmartinez82:
@lazka In the Windows world there's a hack for this exact scenario. I just don't know if MSYS2 behaves the same way:

Applications that need two flavors of the same executable with a Windows subsystem and a Console subsystem employ a very old trick. They have `a.com` and `a.exe` and Windows defaults to `a.com` first when looking the PATH for `a`.
That's how, `devenv` (Visual Studio) for instance, works:

[...]

PR #24502, opened 2025-06-07
meld: Create separate console and gui executables

hmartinez82:
There's now a blocking, console based, `meld.exe`  and a non-blocking, gui based `meldw.exe`

Fixes https://github.com/msys2/MINGW-packages/issues/24494

### DIFF
--- a/mingw-w64-meld3/pyscript2exe.py
+++ b/mingw-w64-meld3/pyscript2exe.py
@@ -6,14 +6,20 @@ foobar.py -> foobar.exe + foobar-script.py
 import sys
 import re
 import os
+import shutil
 from setuptools.command.easy_install import get_win_launcher
 
 path = sys.argv[1]
+root, ext = os.path.splitext(path)
 with open(path, "rb") as f:
     data = f.read()
 with open(path, "wb") as f:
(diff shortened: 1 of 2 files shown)


mingw-w64-meld3/pyscript2exe.py
```diff
@@ -6,14 +6,20 @@
...
     data = f.read()
 with open(path, "wb") as f:
-    f.write(re.sub(b"^#![^\n\r]*", b'#!pythonw.exe', data))
-root, ext = os.path.splitext(path)
+    f.write(re.sub(b"^#![^\n\r]*", b'#!bin/python.exe', data))
 with open(root + ".exe", "wb") as f:
+    f.write(get_win_launcher("cli"))
+shutil.copy(path, root + "-script.py")
+with open(path, "wb") as f:
+    f.write(re.sub(b"^#![^\n\r]*", b'#!bin/pythonw.exe', data))
```
lazka: why "bin/" now?

hmartinez82: @lazka I just removed it. Also a side effect of my broken setuptools :( that I forgot to revert

HazardyKnusperkeks:
Thanks!

PR #24505, opened 2025-06-07
ninja: build with cmake+make; fix missing manifest

### LINKED ISSUE #24479: Ninja 1.12.1 from the repos is not long paths aware

1. Download the release from https://github.com/ninja-build/ninja/releases/download/v1.12.1/ninja-win.zip
2. Search for "longPathAware" inside the executable file. It contains the string.
3. Install ninja inside mingw-w64 environment: `pacman -S mingw-w64-x86_64-ninja`.
4. Open the executable file and search for "longPathAware" inside the executable file. It doesn't contain the string.

### Expected behavior

The executable file should contain a manifest resource which enables long paths operation in Windows.

### Actual behavior

Ninja complains that paths are longer than 260 chars even if long paths support is enabled in Windows registry.

### Verification

### Windows Version

MINGW64_NT-10.0-19045

- [x] MINGW64
- [ ] MINGW32
- [ ] UCRT64
- [ ] CLANG64
- [ ] CLANGARM64

_No response_

lazka:
Fixes #24479

the static lib is gone now, but from what I understand that's private anyway

### DIFF
--- a/mingw-w64-ninja/0001-fix-manifest-build.patch
+++ b/mingw-w64-ninja/0001-fix-manifest-build.patch
@@ -1,0 +1,17 @@
+--- ninja-1.12.1/CMakeLists.txt.orig	2025-06-07 20:26:58.068523900 +0200
++++ ninja-1.12.1/CMakeLists.txt	2025-06-07 20:27:07.480625600 +0200
+@@ -193,7 +193,7 @@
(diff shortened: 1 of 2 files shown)


lazka:
Since cmake for convenience depends on ninja this adds a dep cycle. Can be solved by removing the ninja dep from cmake-bootstrap and using that when bootstrapping.

PR #24506, opened 2025-06-07
python-numpy: Update to 2.3.0

lazka:
enable the new openmp support


ognevny:
numba should be patched again...

PR #24519, opened 2025-06-09
[new package] python-pylablib 1.4.3

fsagbuya:
### Description

Add `python-pylablib` package.


fsagbuya:
It builds when using `makepkg-mingw`. Can it be triggered from the CI?

fsagbuya:
@MehdiChinoune Thanks!

striezel:
No idea why this took so long, but I think we can merge this pull request. I've just updated the package from 1.4.3 to 1.4.4, and it looks like the CI also passed with the new version. So I'm going to merge this.

PR #24525, opened 2025-06-10
new package: gssdp

benthetechguy:


mingw-w64-gssdp/PKGBUILD
```diff
@@ -0,0 +1,83 @@
...
+
+_realname=gssdp
+pkgbase=mingw-w64-${_realname}
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}"
+         "${MINGW_PACKAGE_PREFIX}-${_realname}-docs")
+pkgver=1.6.3
+pkgrel=1
+pkgdesc="GObject-based API for handling resource discovery and announcement over SSDP (mingw-w64)"
+arch=(any)
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
```
lazka: do you need 32bit builds? We avoid adding new packages for it unless needed by something else.

lazka:
lgtm otherwise

mingw-w64-gssdp/PKGBUILD
benthetechguy: Ok, that makes things a little simpler with the optional gtk4 dependency!

PR #24537, opened 2025-06-11
blender: Update to 4.4.3

Alexpux:


mmuetzel:
Regarding the build errors with Clang, the following snippet in `source/blender/blenlib/BLI_math_base.h` should probably be extended to define `_USE_MATH_DEFINES` also with Clang:
```
#if defined(_MSC_VER) && !defined(_USE_MATH_DEFINES)
#  define _USE_MATH_DEFINES
#endif
```
Or, since it shouldn't hurt with GCC, define it on Windows in general. E.g., something like this:
```
#if defined(_WIN32) && !defined(_USE_MATH_DEFINES)
#  define _USE_MATH_DEFINES
#endif
```

Alexpux:
This changes already in patches

Alexpux:
@MehdiChinoune this is strange because this define is in headers. I will inspect it

MehdiChinoune:
I have this issue before with OpenVDB or OpenEXR, `_USE_MATH_DEFINES` should be defined before include math headers, but some headers include them implicitly so you couldn't track where to put that macro.
Apparently in GNU toolchan math header is not included implicitly, while It is on LLVM toolchain one.

PR #24547, opened 2025-06-12
poppler: Update to 25.06.0

lazka:


lazka:
huh, I wonder if that's the old CRLF issue

edit: I can't reproduce

lazka:
#24677 also fails this way, but smaller build

PR #24549, opened 2025-06-12
libaec: update to 1.1.4

mmuetzel:
Remove patches for issues that have been addressed upstream.


mmuetzel:
Some symbols are no longer exported from the libraries with GCC. That is expected because `dllexport`/`dllimport` attributes are now set correctly by upstream and internal symbols are no longer exported.

PR #24550, opened 2025-06-13
[do not merge] mupdf bump version to 1.26.2

3ncor:
I made this PR originaly just for myself, but since msys2's mupdf haven't got a update for a long time, I think it might be helpful to upstream it, hopfully someone with better knowledge can take over from this point.

Potential Issues:
1. Not sure if the dependencies are correct.
2. `shared=yes` build does not work, but most linux distros use it.
3. Upstream removed MINGW support from Makefile since 1.26.0

### DIFF
--- a/mingw-w64-mupdf/001-makerules-search-system-libraries.patch
+++ b/mingw-w64-mupdf/001-makerules-search-system-libraries.patch
@@ -1,11 +1,24 @@
 --- a/Makerules
 +++ b/Makerules
-@@ -171,7 +171,7 @@
-   CFLAGS += -DCLUSTER
+@@ -90,7 +93,12 @@
+   SO := dylib
+ else
+   LDREMOVEUNREACH := -Wl,--gc-sections
+-  SO := so
++  ifeq ($(OS),$(filter $(OS),Linux MINGW Windows_NT))
++    SO := dll
++    EXE := .exe
++  else
++    SO := so
++  endif
  endif
  
--ifeq ($(OS),Linux)
-+ifeq ($(OS),$(filter $(OS),Linux MINGW Windows_NT))
-     LINUX_OR_OPENBSD := yes
- endif
- ifeq ($(OS),OpenBSD)
+ SANITIZE_FLAGS += -fsanitize=address
+@@ -216,7 +224,7 @@
+ 
+ else
+ 
+-  ifeq ($(OS),Linux)
++  ifeq ($(OS),$(filter $(OS),Linux MINGW Windows_NT))
+     HAVE_OBJCOPY := yes
+   endif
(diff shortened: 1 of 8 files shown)


3ncor:
OK, all CI builds succesful ,I think its ready for merge.

3ncor:
~~Building `libmupdfcpp` currently requires python bindings for clang and Visual Studio, clang python for generating source code for libmupdfcpp and VS because wrap scripts demand it on windows. I have no idea how to fix it, if we don't have libmupdfcpp the pymupdf package can't build.~~

Well it only takes a small patch to make libmupdfcpp for gcc builds, but the error message for clang builds make no sense to me... 😩

ousia:
Many thanks for this newer version, @3ncor.

I updated `mingw-w64-x86_64-mupdf-gl-1.26.3-1` (no `mingw-w64-x86_64-mupdf` installed in any version) and I had two binaries: current `mupdf.exe` (which was actually `mupdf-gl.exe`) and the older version of `mupdf-gl.exe`.

I guess this comes from naming the binary in `mingw-w64-x86_64-mupdf-gl-1.26.3-1` to `mupdf.exe` instead of `mupdf-gl.exe`.

ousia:
BTW, the new `mupdf-gl.exe` displays the following message:

> ```
> This version of MuPDF has been built WITHOUT clipboard or unicode input support!
> Please file a complaint with your friendly local distribution manager.
> ```

I suggest to add the following patch (or a similar one), that is [included with _Fedora_](https://github.com/mjg/mupdf-spec/blob/spec/0001-Do-not-complain-to-your-friendly-local-distribution-.patch).

3ncor:
> Many thanks for this newer version, @3ncor.
> 
> I updated `mingw-w64-x86_64-mupdf-gl-1.26.3-1` (no `mingw-w64-x86_64-mupdf` installed in any version) and I had two binaries: current `mupdf.exe` (which was actually `mupdf-gl.exe`) and the older version of `mupdf-gl.exe`.
> 
> I guess this comes from naming the binary in `mingw-w64-x86_64-mupdf-gl-1.26.3-1` to `mupdf.exe` instead of `mupdf-gl.exe`.

I made a mistake by not renaming the executable correctly in some of my earlier commits, this should be fixed now, you can safely remove the mupdf-gl.exe.



> BTW, the new `mupdf-gl.exe` displays the following message:
> 
> > ```
> > This version of MuPDF has been built WITHOUT clipboard or unicode input support!
> > Please file a complaint with your friendly local distribution manager.
> > ```
> 
> I suggest to add the following patch (or a similar one), that is [included with _Fedora_](https://github.com/mjg/mupdf-spec/blob/spec/0001-Do-not-complain-to-your-friendly-local-distribution-.patch).

Thanks, just added this patch.

ousia:
I was totally sure I replied to this yesterday, @3ncor (with a text reply besides the emoji reaction).

Not sure how to get a review and merging, but mentioning @Biswa96 (being the PyMuPDF maintainer).

Many thanks for your help.

ousia:
Many thanks for the new PR, @3ncor.

Sorry for contacting you, @lazka, but has been mupdf somehow left behind or deprecated?

PR #24554, opened 2025-06-13
binutils: backport upstream commits

### LINKED ISSUE #23577: Regression with delaylib when using gcc + mingw

jeremyd2019:
This fixes #23577,
https://sourceware.org/bugzilla/show_bug.cgi?id=14339, and https://sourceware.org/bugzilla/show_bug.cgi?id=32675 at least.  Also backport support for filling in the load config directory with _load_config_used.  This is not directly a bugfix, but avoided a conflict with the other patches.

### DIFF
--- a/mingw-w64-binutils/5000-bfd-properly-use-bfd_get_symbol_leading_char-in-peXX.patch
+++ b/mingw-w64-binutils/5000-bfd-properly-use-bfd_get_symbol_leading_char-in-peXX.patch
@@ -1,0 +1,54 @@
+From 9d5eb91175c5ddb601ec24d3249ad23e81f92619 Mon Sep 17 00:00:00 2001
+From: Jeremy Drake <sourceware-bugzilla@jdrake.com>
+Date: Mon, 7 Apr 2025 13:18:46 +0200
+Subject: [PATCH 5000/5007] bfd: properly use bfd_get_symbol_leading_char() in
+ peXXigen
+
+This function returns the leading char to use, so we cannot just assume
+it will always be '_' or '\0'.
+
+Signed-off-by: Jeremy Drake <sourceware-bugzilla@jdrake.com>
+---
+ bfd/peXXigen.c | 12 ++++++------
+ 1 file changed, 6 insertions(+), 6 deletions(-)
+
+diff --git a/bfd/peXXigen.c b/bfd/peXXigen.c
+index 29012688adf..561f8585766 100644
+--- a/bfd/peXXigen.c
++++ b/bfd/peXXigen.c
+@@ -4403,6 +4403,7 @@ _bfd_XXi_final_link_postscript (bfd * abfd, struct coff_final_link_info *pfinfo)
+   struct coff_link_hash_entry *h1;
+   struct bfd_link_info *info = pfinfo->info;
+   bool result = true;
++  char name[20];
+ 
+   /* There are a few fields that need to be filled in now while we
+      have symbol table access.
+@@ -4540,10 +4541,9 @@ _bfd_XXi_final_link_postscript (bfd * abfd, struct coff_final_link_info *pfinfo)
+ 	}
+     }
+ 
+-  h1 = coff_link_hash_lookup (coff_hash_table (info),
+-			      (bfd_get_symbol_leading_char (abfd) != 0
+-			       ? "__tls_used" : "_tls_used"),
+-			      false, false, true);
++  name[0] = bfd_get_symbol_leading_char (abfd);
++  strcpy (name + !!name[0], "_tls_used");
++  h1 = coff_link_hash_lookup (coff_hash_table (info), name, false, false, true);
+   if (h1 != NULL)
+     {
+       if ((h1->root.type == bfd_link_hash_defined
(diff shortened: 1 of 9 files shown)


jeremyd2019:
I tested this using https://github.com/msys2/msys2-tests/pull/81, which fails with existing packages and succeeds with the packages from this PR for MINGW32 MINGW64 and UCRT64

PR #24564, opened 2025-06-15
Static libraries fix and pkg-config .pc file for mingw-w64-angleproject package

tdcosta100:
The `libGLESv2.a` static library bundled with `mingw-w64-angleproject` has lots of undefined symbols, because its `libANGLE` dependency is not built as a static library, but as a group of objects that are linked directly in the shared DLL library. While this works if you are building with ANGLE sources, it have problems as a package. So these modifications bring the `libangle.a` static library, so we can link against `libGLESv2.a` without any missing dependency.

Additionally, this PR also brings the `angleproject.pc`, which allows using `pkg-config` to get all the required information about ANGLE's libraries.

### DIFF
--- a/mingw-w64-angleproject/002-buildflags-fixes.patch
+++ b/mingw-w64-angleproject/002-buildflags-fixes.patch
@@ -1,5 +1,5 @@
 diff --git a/BUILD.gn b/BUILD.gn
-index 964d51123..14bc8fb6c 100644
+index 964d511238..1a99c6095d 100644
 --- a/BUILD.gn
 +++ b/BUILD.gn
 @@ -14,7 +14,10 @@ if (angle_has_build) {
@@ -117,15 +117,16 @@ index 964d51123..14bc8fb6c 100644
    }
  
    if (angle_enable_metal) {
-@@ -1469,6 +1476,7 @@ if (angle_enable_gl_desktop_frontend) {
+@@ -1469,6 +1476,8 @@ if (angle_enable_gl_desktop_frontend) {
  
  angle_static_library("libGLESv2_static") {
    sources = libglesv2_sources
++  complete_static_lib = true
 +  output_name = "libGLESv2"
    if (angle_enable_gl_desktop_frontend) {
      sources += libglesv2_gl_sources
      defines = [ "ANGLE_ENABLE_GL_DESKTOP_FRONTEND" ]
-@@ -1614,6 +1622,7 @@ if (angle_enable_vulkan) {
+@@ -1614,6 +1623,7 @@ if (angle_enable_vulkan) {
  }
  
  libEGL_template("libEGL_static") {
--- a/mingw-w64-angleproject/004-swiftshader-updates.patch
+++ b/mingw-w64-angleproject/004-swiftshader-updates.patch
@@ -1,5 +1,5 @@
 diff --git a/src/Reactor/BUILD.gn b/src/Reactor/BUILD.gn
-index 67dfeb0ec..ceb96a5d7 100644
+index 67dfeb0ec..5df50ba8a 100644
 --- a/src/Reactor/BUILD.gn
 +++ b/src/Reactor/BUILD.gn
 @@ -16,7 +16,7 @@ import("reactor.gni")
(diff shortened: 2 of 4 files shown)


raedrizqie:
Have you tried linking those libs with `-langle_common`?

tdcosta100:
Yes, I did. But lots of symbols are missing even with `libangle_common.a`, `libpreprocessor.a` and `libtranslator.a`.

tdcosta100:
Build on `CLANG64` corrected.

tdcosta100:
> Have you tried linking those libs with `-langle_common`?

Explaining better: in the ANGLE's build system (Google GN), some targets (specially `libANGLE`) are meant to not generate a static library, but just a collection of objects that will be linked as dependencies of a library consuming the "virtual" target who they belong to. This works fine if you are building everything together from the sources, that is, using the same build system to build your own library and adding the ANGLE's targets as dependencies.

When you are distributing ANGLE as a bundle, this doesn't work, because the static libraries generated are just a small part of the objects necessary to build the complete ANGLE implementation.

This is not a problem if you link against the shared libraries (`libEGL` and `libGLESv2`), because at build time, they have all objects to put together (the objects generated by the "virtual" target `libANGLE` are linked inside them), so they work fine. But if you want to use the bundled static libraries instead, they don't work because the needed objects of the "virtual" target are not present.

The solution to this is to make `libANGLE` a static library, so it will be part of the bundle, and then will bring all the needed symbols currently missing.

raedrizqie:
This will introduce a new library (libangle.a), but it should only be used for static linking. Your new pc file should only add `-langle` to `Libs.private`.

Can we just add libangle objects directly in `libEGL.a` or `libGLESv2.a`, so that we dont have to create an extra libangle.a library?

tdcosta100:
Yes, the `libANGLE` objects are in `libGLESv2` in the shared library, so I can try to reproduce the same in the static library. Doing this can lead to the same experience as linking against `libEGL.dll.a` and `libGLESv2.dll.a`. By the way, it's my first experience writing a `.pc` file, and I'm thinking if there should be a way to choose wether to link against the shared libraries or the static ones. Does it makes sense?

raedrizqie:
>By the way, it's my first experience writing a .pc file, and I'm thinking if there should be a way to choose wether to link against the shared libraries or the static ones. Does it makes sense?

Maybe something like this?
```
Libs: -L${libdir} -lEGL -lGLESv2 -lcfgmgr32 -lgdi32 -ld3d9 -ldxgi -ldxguid
Libs.private: -langle_common -langle_gpu_info_util -langle_image_util -lpreprocessor -ltranslator
Cflags: -I${includedir}
Cflags.private: -DKHRONOS_STATIC
```

tdcosta100:
Nice. I made `libGLESv2` a complete library, so all needed objects are incorporated to it. I checked the dependencies, and `libGLESv2` does not depend on `libangle_common`, `libangle_gpu_info_util`, `libangle_image_util`, `libpreprocessor` and `libtranslator`.

Here are the lists of the dependencies, generated with `gn refs <build dir> <target>`:

**angle_common**
//:angle_capture_common
//:angle_gl_enum_utils
//:angle_gpu_info_util
//:angle_image_util_headers
//:angle_json_serializer
//:libANGLE_headers
//:libANGLE_no_vulkan
//:libEGL
//:libEGL_vulkan_secondaries
//:preprocessor
//:translator
//:translator_gl_d3d_only
//src/common/spirv:angle_spirv_base
//src/common/spirv:angle_spirv_builder
//src/common/spirv:angle_spirv_headers
//src/common/spirv:angle_spirv_parser
//src/common/vulkan:angle_libvulkan_loader
//src/common/vulkan:angle_vulkan_icd


**angle_gpu_info_util**
//:libANGLE_no_vulkan
//:libfeature_support
//src/libANGLE/renderer/d3d:angle_d3d_shared
//src/libANGLE/renderer/gl:angle_gl_backend
//src/libANGLE/renderer/vulkan:angle_vulkan_backend
//src/libANGLE/renderer/vulkan:angle_vulkan_secondaries_backend


**angle_image_util**
//:libANGLE_no_vulkan
//src/libANGLE/renderer/d3d:angle_d3d_shared
//src/libANGLE/renderer/gl:angle_gl_backend
//src/libANGLE/renderer/vulkan:angle_vulkan_backend
//src/libANGLE/renderer/vulkan:angle_vulkan_secondaries_backend


**preprocessor**
//:angle_static
//:translator
//:translator_gl_d3d_only


**translator**
//:angle_shader_translator
//:angle_static
//:libANGLE_no_vulkan
//:translator_fuzzer

Note `libGLESv2` does not depend directly on any of the above targets.

lazka:
@raedrizqie any more thoughts on this?

raedrizqie:
@lazka I think this is good to be merged

tdcosta100:
Great! Thank you guys! I was waiting for this change to bring MSYS2 build support of Maplibre Native.

PR #24585, opened 2025-06-17
Update tree sitter grammars

ognevny:


ognevny:
it probably worths to switch to cmake...

ognevny:
does it worth to use .so extension and for compiled lib? it actually can be .dll, but copied as .so for nvim

podsvirov:
What about to update [wasmtime](https://packages.msys2.org/base/mingw-w64-wasmtime) too?

ognevny:
https://github.com/msys2/MINGW-packages/pull/24610, I'm waiting for a correct fix of tree-sitter's compile errors

podsvirov:
> #24610, I'm waiting for a correct fix of tree-sitter's compile errors

Thank you, we will be waiting for the opportunity to update `wasmtime`.

ognevny:
it doesn't seem that tree-sitter devs will fix it soon, so I have to suppress this warnings

PR #24592, opened 2025-06-18
vulkan: Restore fix for proper import library and pkg-config file

Alexpux:


Alexpux:
@lazka  CI cannot build multiple packages from pull request in order of it dependencies between each other?

lazka:
hm, in theory it should, not sure what's wrong there.

I'd guess providing the same package name twice with different versions doesn't work.

PR #24593, opened 2025-06-18
Fix msmpi for urtc64 running on aarch64

robtaylor:
I hit a problem where the standard msmpi package had an undefined reference to WinMain when running mpicc. This seems to fix the issue.

### DIFF
--- a/mingw-w64-msmpi/PKGBUILD
+++ b/mingw-w64-msmpi/PKGBUILD
@@ -67,12 +67,12 @@ build() {
     sed -e "s|gfortran|flang|g" -i ${mpic}
   fi
   cflags="-s -O2 -DNDEBUG"
-  $CC ${cflags} -o bin/mpicc.exe -DCC "${mpic}"
-  $CC ${cflags} -o bin/mpicxx.exe -DCXX "${mpic}"
+  $CC ${cflags} -mwindows -o bin/mpicc.exe -DCC "${mpic}"
+  $CC ${cflags} -mwindows -o bin/mpicxx.exe -DCXX "${mpic}"
   cp bin/mpicxx.exe bin/mpic++.exe
 
   # Fortran Compiler wrappers
-  $CC ${cflags} -o bin/mpifort.exe -DFC "${mpic}"
+  $CC ${cflags} -mwindows -o bin/mpifort.exe -DFC "${mpic}"
   cp bin/mpifort.exe bin/mpif77.exe
   cp bin/mpifort.exe bin/mpif90.exe
 


ognevny:
please rebase your branch on current master branch. I fixed an issue that made package build fail

robtaylor:
Done, that was very helpful, thank you @ognevny !

ognevny:
I can't review such change here, I'll leave this for others...

mmuetzel:
I might be misunderstanding what `-mwindows` actually does. But the compiler wrappers are console applications and should not be built with that command line argument afaict.

Which error are you seeing without it? How do you reproduce the issue that you'd like to resolve with the proposed changes?

ognevny:
maybe it should be `-municode` instead?

mmuetzel:
> maybe it should be `-municode` instead?

I don't think so.
The entry function of the compiler wrapper is `main` (not `wmain`):
https://github.com/msys2/MINGW-packages/blob/0570a94246e7c5f68960714c4f766a05f5831453/mingw-w64-msmpi/mpi.c#L18

ognevny:
could it mean that issue appears with a stuff processed by mpicc (just a wrong code), but when not trying to run mpicc itself?

robtaylor:
No, the issue is specifically running mpicc or mpic++, fails to run on
urtc64 running on an aarch64 system. Note this problem doesn't appear to
happen on an x86_64 system.

ognevny:
please, can you provide steps to reproduce the issue? what output is shown?

mmuetzel:
> No, the issue is specifically running mpicc or mpic++, fails to run on
> urtc64 running on an aarch64 system. Note this problem doesn't appear to
> happen on an x86_64 system.

What exactly is "the issue"? Which error message or warning or message box do you get? Which exact command do you run when "the issue" happens?

robtaylor:
I think i descibed it fine at the start of this thread, but as you don't seem to believe me...


```
Rob Taylor@mac-vm UCRT64 /build/Xyce-build
# pacman -S mingw-w64-ucrt-x86_64-msmpi
warning: mingw-w64-ucrt-x86_64-msmpi-10.1.1-13 is up to date -- reinstalling
resolving dependencies...

Packages (1) mingw-w64-ucrt-x86_64-msmpi-10.1.1-13

Total Installed Size:  3.25 MiB
Net Upgrade Size:      0.00 MiB

:: Proceed with installation? [Y/n] y
(1/1) checking keys in keyring                                             [#########################################] 100%
(1/1) checking package integrity                                           [#########################################] 100%
(1/1) loading package files                                                [#########################################] 100%
(1/1) checking for file conflicts                                          [#########################################] 100%
(1/1) checking available disk space                                        [#########################################] 100%
(1/1) reinstalling mingw-w64-ucrt-x86_64-msmpi                             [#########################################] 100%

Rob Taylor@mac-vm UCRT64 /build/Xyce-build
# mpicc
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/ucrt64/lib/libmingw32.a(lib64_libmingw32_a-crtexewin.o): in function `main':
C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtexewin.c:67:(.text.startup+0xc5): undefined reference to `WinMain'
```

mmuetzel:
It's not like we don't believe you. We'd like to understand what is happening.

I can reproduce that on Windows 11 (on AMD64):
```
$ mpicc
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/ucrt64/lib/libmingw32.a(lib64_libmingw32_a-crtexewin.o): in function `main':
D:/W/B/src/mingw-w64/mingw-w64-crt/crt/crtexewin.c:66:(.text.startup+0xb5): undefined reference to `WinMain'
```

However, I don't think this is an actual issue though.
The compiler wrapper isn't meant to be called without any source or object file as the input. If you call the compiler wrapper with actual input arguments, it works just fine. E.g.:
```
Markus@Markus-PC UCRT64 ~
$ cat hello_world.c
#include <stdio.h>

int main (void)
{
  printf("Hello World!");
  return 0;
}

Markus@Markus-PC UCRT64 ~
$ mpicc hello_world.c

Markus@Markus-PC UCRT64 ~
$ ./a.exe
Hello World!
```


Your "fix" just "eats" the output with the error message without displaying it on the console.

robtaylor:
hmm, no, it's not meant to crash without input. That causes cmake configuration checks to fail.

The correct behaviour is:
```
% mpicc
clang: error: no input files
```

mmuetzel:
> hmm, no, it's not meant to crash without input.

Afaict, it doesn't crash. It fails with an error.

> The correct behaviour is:

So, an error is the expected behavior?
How is that error handled differently from the error that you currently see? Which CMake project are you attempting to build?

robtaylor:
hmm, though looking again, I do suspect there is something weirder going on. I'll investigate further with the cmake check failures with the mainline package.

mmuetzel:
Note that unlike most other packages from MSYS2, this package installs only the *import* library (and compiler wrappers).
To actually use MSMPI, you'll also need to install the actual libraries from Microsoft:
https://learn.microsoft.com/en-us/message-passing-interface/microsoft-mpi
(And potentially add the path to those libraries to the environment variable `PATH`.)

robtaylor:
Yes, I'm aware of that =)

robtaylor:
> > hmm, no, it's not meant to crash without input.
> 
> Afaict, it doesn't crash. It fails with an error.
> 
> > The correct behaviour is:
> 
> So, an error is the expected behavior? 

Fair. It would be preferable if it was the right (meaningful) error message, but I don't think now that was causing the issue (see below)

> How is that error handled differently from the error that you currently see? Which CMake project are you attempting to build?

Trilinos - for full context https://github.com/ChipFlow/Xyce-build/

Having said that, I now can't reproduce the problem >< (I even tested uninstalling microsoft-mpi  while rolling back msmpi to sanity check). 

I guess we can close this for now and I'll revisit if/when it hits again...

PR #24601, opened 2025-06-19
starpu: update to 1.4.8

3rav:


ognevny:
why patch isn't applied?

3rav:
Because it is no longer needed, the fix is ​​already in the source code.

ognevny:
can you remove it entirely from recipe, please? such commenting may be confusing for someone

PR #24608, opened 2025-06-20
pdcurses: update to PDCursesMod 4.5.2

GitMensch:
additional: compile with standard flags again, PDCursesMod is portable again (update contains C90/C23 fixes)

(See [`docs/HISTORY.md`](https://github.com/Bill-Gray/PDCursesMod/blob/v4.5.2/docs/HISTORY.md) for more details, and/or the [Git log](https://github.com/Bill-Gray/PDCursesMod/commits/v4.5.2) for extreme detail.)

PDCursesMod 4.5.2 - 2025 June 19 (msys2 filtered)
================================

Minor new features
------------------

- Fixes for C90 compliance : removed some C++-style // comments,  moved variables to the top of code blocks,  used TRUE/FALSE instead of true/false.

Bug fixes
---------

- getch() triggered an assert() if fed a character that expanded to a multi-byte string in PDC_FORCE_UTF8 builds (reported by
  Robin Haberkorn).  

- WinCon/pdckbd.c contained a poorly-written cast to which the C23 standard objects,  resulting in warnings/errors when built to that standard. Fix supplied by Gisle Vanem.  

- The `PDC_KEY_MODIFIER_REPEAT` modifier was broken in SDL2.


GitMensch:
Note: As seen in CI WinGUI currently fails with C23, I'll add it back or adjust the SHA if 4.5.2 is repackaged

GitMensch:
Updated to current version which should really solve C23...

PR #24620, opened 2025-06-20
Add mingw-w64-python-pykwalify

jhol:
pyKwalify is a library for YAML/JSON validation.


jhol:
Added python docopt which is a required dependency.

PR #24621, opened 2025-06-21
add uboot-tools

jhol:


MehdiChinoune:
u-boot-tools is already available as an MSYS package https://packages.msys2.org/base/u-boot-tools

jhol:
> u-boot-tools is already available as an MSYS package https://packages.msys2.org/base/u-boot-tools

oops! well that's good because packaging was not going well.

PR #24626, opened 2025-06-22
zed: remove some workarounds

ognevny:
rebuild with our protoc and with ld on GCC environments

locally tested UCRT64 artifact, anyone feel free to do the same (previous time it also succeeded for me but not for other users)


ognevny:
ok, let's see if issues appear

PR #24630, opened 2025-06-23
[new package] libmem

luadebug:
Looks like it is ready to go.


lazka:
From what I understand the patch was just there so silence warnings (?). Are you OK with just ignoring the warnings instead, so we don't have to maintain that patch?

otherwise lgtm

luadebug:
> From what I understand the patch was just there so silence warnings. Are you OK with just ignoring the warnings instead, so we don't have to maintain that patch?

Yes, that is the case, sure lets use flags instead.

PR #24645, opened 2025-06-25
sundials: update to 7.4.0

mmuetzel:


mmuetzel:
I've downloaded and installed the MINGW64 build artifact from this PR. With it, the test suite of Octave is still passing without unexpected errors.

PR #24647, opened 2025-06-25
rust: update to 1.88.0

ognevny:
this is then first Rust relase that downgrades i686-mingw target to tier 2. that means that std and host tools are guaranteed to build, but full test suite won't be run for this target: https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-2-target-policy. currently this doesn't mean that 32-bit Rust will be dropped soon, but now you should be aware of possible issues while using it as a host

config change: 
- remove crt-static option, it didn't do what expected (this will link libunwind dynamically for gnullvm tools)

@mati865 I guess removing stage0 hack for aarch64 gnullvm won't hurt, right? and I saw you put an effort to make gnullvm targets tier 2 with **host tools**, so we can truly bootstrap all the targets soon

### DIFF
--- a/mingw-w64-rust/PKGBUILD
+++ b/mingw-w64-rust/PKGBUILD
@@ -9,7 +9,7 @@ if [[ $MINGW_PACKAGE_PREFIX == *-clang-* ]]; then
 fi
 
 rust_dist_server=https://static.rust-lang.org/dist
-#rust_dist_server=https://dev-static.rust-lang.org/dist/2025-05-10
+#rust_dist_server=https://dev-static.rust-lang.org/dist/2025-06-23
 
 _realname=rust
 pkgbase=mingw-w64-${_realname}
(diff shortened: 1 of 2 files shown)


mati865:
> @mati865 I guess removing stage0 hack for aarch64 gnullvm won't hurt, right? and I saw you put an effort to make gnullvm targets tier 2 with **host tools**, so we can truly bootstrap all the targets soon

I think so. Starting with 1.89 `-gnullvm` hosts should be possible to bootstrap using upstream stage0, just like regular `-gnu` hosts.

mati865:
> Starting with 1.89 `-gnullvm` hosts should be possible to bootstrap using upstream stage0, just like regular `-gnu` hosts.

Oops, I forgot to update the manifest, so this has slipped to 1.90 release. At least it works now:
```
$ rustc +nightly -vV
rustc 1.90.0-nightly (f26e58023 2025-06-30)
binary: rustc
commit-hash: f26e58023071e71636a3c72ac3a2bf89b1f76706
commit-date: 2025-06-30
host: x86_64-pc-windows-gnullvm
release: 1.90.0-nightly
LLVM version: 20.1.7

$ cargo +nightly -vV
cargo 1.90.0-nightly (930b4f62c 2025-06-28)
release: 1.90.0-nightly
commit-hash: 930b4f62cfcd1f0eabdb30a56d91bf6844b739bf
commit-date: 2025-06-28
host: x86_64-pc-windows-gnullvm
libgit2: 1.9.0 (sys:0.20.2 vendored)
libcurl: 8.12.1-DEV (sys:0.4.80+curl-8.12.1 vendored ssl:Schannel)
os: Windows 10.0.26100 (Windows 11 Professional) [64-bit]
```

ognevny:
yeah, I saw an upstream PR. there is just no any way to backport to beta? :(

mati865:
No, it's not a bug fix.

PR #24688, opened 2025-06-30
qt6-base: better fallback path for freetype

mmuetzel:
Use a more reasonable fallback directory for fonts on Windows. The existing fallback points to a directory that no longer exists by default which causes warnings like the following on run-time:

```
QFontDatabase: Cannot find font directory C:/msys64/mingw64/qt6/libs/fonts.
Note that Qt no longer ships fonts. Deploy some (from
https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
```

The change is motivated by reports from Octave users. The proposed change would affect *all* downstream users of `QFontDatabase` though (in a positive way - I think).
@MehdiChinoune, @lazka, others: Is this change acceptable?

Note: Qt6 doesn't support fontconfig for Windows.

### DIFF
--- a/mingw-w64-qt6-base/008-freetype-fonts-fallback-dir.patch
+++ b/mingw-w64-qt6-base/008-freetype-fonts-fallback-dir.patch
@@ -1,0 +1,33 @@
+Use a more reasonable fallback directory for fonts on Windows.
+
+The existing fallback points to a directory that no longer exists by default
+which causes warnings like the following on run-time:
+
+```
+QFontDatabase: Cannot find font directory C:/msys64/mingw64/qt6/libs/fonts.
+Note that Qt no longer ships fonts. Deploy some (from
+https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
+```
+
+diff -urN qtbase-everywhere-src-6.9.1/src/gui/text/qplatformfontdatabase.cpp.orig qtbase-everywhere-src-6.9.1/src/gui/text/qplatformfontdatabase.cpp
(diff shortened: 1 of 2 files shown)


lazka:
Just curious, what would be the downside of just have an empty dir there? At least some basic testing shows qt6 happily finding all my fonts without doing anything special. But I might be missing something.

mmuetzel:
It's difficult to track down the exact code path that is taken when it comes to this. But if I understand correctly, Qt6 is using the Windows font API for on-screen rendering. So, the FreeType library isn't involved at all in this case.
But apparently, the Windows font API doesn't work(?) for off-screen rendering. In that case, Qt6 falls back to FreeType.
Octave uses off-screen rendering to print to PDF files.

To reproduce the issue, you can run the following commands in Octave:
```
close all
plot(1:10)
print ("foo.pdf")
```

You'll see the warning for those lines. Note that the axis labels are still rendered for this example because Octave ships a couple of font files that it uses as a fallback (e.g., in `<PREFIX>\share\octave\10.2.0\fonts`) that it registers explicitly. These fonts are found by Qt independent on the font backend that it is using.

With the following example, you can see more clearly that the text appears in the PDF but with the wrong font before the change from here:
```
close all
annotation('textbox', [0.5, 0.5, 100, 100], 'String', 'test', 'FontName', 'Comic Sans MS', 'LineStyle', 'none')
print ("foo.pdf")
```

With the change from here, the PDF prints with the correct font.

lazka:
Thanks for the explanation. This patch looks reasonable to me then.

mmuetzel:
Thanks for the feedback.

PR #24689, opened 2025-06-30
Update GW to v1.2.4

kcleal:
Added angleproject to dependencies

### DIFF
--- a/mingw-w64-gw/PKGBUILD
+++ b/mingw-w64-gw/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=gw
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=1.2.3
+pkgver=1.2.4
 pkgrel=1
 pkgdesc="A genome browser for bam/cram and vcf files (mingw-w64)"
 arch=(any)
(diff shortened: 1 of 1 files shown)


mingw-w64-gw/PKGBUILD
```diff
@@ -21,12 +21,14 @@ depends=("${MINGW_PACKAGE_PREFIX}-htslib"
          "${MINGW_PACKAGE_PREFIX}-libpng"
          "${MINGW_PACKAGE_PREFIX}-zlib"
          "${MINGW_PACKAGE_PREFIX}-icu"
+         "${MINGW_PACKAGE_PREFIX}-angleproject"
          )
 makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-angleproject"
```
lazka: From what I understand it is a runtime-only dep(?), indirectly via glfw, so this is not needed.

mingw-w64-gw/PKGBUILD
```diff
@@ -21,12 +21,14 @@ depends=("${MINGW_PACKAGE_PREFIX}-htslib"
          "${MINGW_PACKAGE_PREFIX}-libpng"
          "${MINGW_PACKAGE_PREFIX}-zlib"
          "${MINGW_PACKAGE_PREFIX}-icu"
+         "${MINGW_PACKAGE_PREFIX}-angleproject"
```
lazka, line 24: so this is always using GLES? are the freeglut/glew deps still needed then?

mingw-w64-gw/PKGBUILD
kcleal: I agree, this should be a runtime dep. However, without it in the runtime dependencies then I get a missing libEGL error, so glfw might not be pulling it in properly

mingw-w64-gw/PKGBUILD
kcleal: It uses GLES yes, but can be configured to use opengl at runtime

lazka: ok, thanks

PR #24698, opened 2025-07-01
tzdata: Include tzdata.zi

HazardyKnusperkeks:
This is the file, which is used by libstdc++ and libc++ for the time zone database.


mingw-w64-tzdata/PKGBUILD
```diff
@@ -4,23 +4,44 @@ _realname=tzdata
...
 pkgver=2025b
-pkgrel=1
+pkgrel=2
 pkgdesc="Sources for time zone and daylight saving time data (mingw-w64)"
 arch=('any')
 mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64' 'clangarm64')
 url="https://www.iana.org/time-zones"
 license=('custom: public domain')
-makedepends=('tzcode')
+makedepends=('tzcode' 'make' 'gawk')
```
ognevny: not needed change. both packages are included with `base-devel`

lazka: I personally don't mind the explicitness, ymmv

lazka: although gawk is not used..

HazardyKnusperkeks: `awk` is used in the makefile to create `tzdata.zi`.

I'm fine with removing them.

HazardyKnusperkeks:
I'd keep the patch, because at least when I build it locally it uses the MINGW-packages git, and when you try something it is automatically `-dirty`.

lazka:
It doesn't here, as long as I don't change any dependent file.

What do you think about my last change? (I'd just like to avoid patches to make updates easier..)

HazardyKnusperkeks:
That will do the trick, nice.

lazka:
Thanks!

PR #24707, opened 2025-07-02
libheif: Update to 1.20.0

lazka:


lazka:
https://github.com/strukturag/libheif/issues/1542

PR #24708, opened 2025-07-02
python-cmake-format: new package for formatting CMakeLists.txt files

chrisjohgorman:
- tried to get modern python -m build to work with this package, but it wouldn't build properly with it.  Reverting to python setup.py bdist_wheel works.

- This package provides tools for parsing and providing a consistent format for CMakeLists.txt files.


chrisjohgorman:
- Made and fixed syntax error, missing dependency for git.

- Ready for review.

lazka:
you probably know, but note that there is a good chance this will break in the future since it is unmaintained for years.

chrisjohgorman:
Thanks @lazka.  Yeah, I noticed this has been unmaintained for years.  I needed to install it on my machine to build a project (gnuradio4) on msys2, so I thought it would be a good addition for however long it works and would make it easier for others to develop for gnuradio4 on msys2.  (All they need to do is pacman -S.)  Thanks for fixing this up with the correct name.  All the best.  Chris.

PR #24713, opened 2025-07-02
gettext: Update to 0.25.1

lazka:


lazka:
ah, gnulib... same as b83f1ef105685eb778ec2b4dc20 I assume

PR #24717, opened 2025-07-03
libxml2: Update to 2.13.8

lazka:
re-add relocation support that was dropped upstream


lazka:
re grokker: we might need to re-enable legacy APIs, and check if the packages using them right now really need them.

zlib can xz can also be removed I think.

PR #24737, opened 2025-07-05
python-pyarrow: rebuild

lazka:


lazka:
wrong package.. python-pyarrow...

PR #24743, opened 2025-07-07
opencv: update to 4.12.0

Biswa96:


Biswa96:
@MehdiChinoune could you provide any hint about the itk build issue?

MehdiChinoune:
It's written in the log
```
D:/_/B/src/ITK-5.4.2/Modules/Core/Common/include/itkFloatingPointExceptions.h:23:1: note: 'uint8_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
```
see: https://gcc.gnu.org/gcc-15/porting_to.html

Biswa96:
Oh, I scroll from the bottom of the log.

PR #24748, opened 2025-07-07
gexiv2: update to 0.14.6

brunvonlope:
This fixes a bug in gimp: https://gitlab.gnome.org/GNOME/gimp/-/issues/14384 which we plan to release a revision


ognevny:
gtkdoc fails again on aarch64...

PR #24750, opened 2025-07-07
GCC: use msys tzdata at runtime

HazardyKnusperkeks:
Instead of the hard compiled version shipped with gcc.

### DIFF
--- a/mingw-w64-gcc/0001-libstdc-Search-for-tzdata-on-Windows-msys.patch
+++ b/mingw-w64-gcc/0001-libstdc-Search-for-tzdata-on-Windows-msys.patch
@@ -1,0 +1,78 @@
+From 6b50be573c35011012140e7e191c235719495cab Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Bj=C3=B6rn=20Sch=C3=A4pers?= <bjoern@hazardy.de>
+Date: Mon, 7 Jul 2025 13:10:49 +0200
+Subject: [PATCH] libstdc++: Search for tzdata on Windows (msys)
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Windows does not provide a tzdata.zi, but msys does. Use this, if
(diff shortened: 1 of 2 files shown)


Biswa96:
Could you please explain the reason for this change?

HazardyKnusperkeks:
As stated in the patch file and in the description, this enables to use the tzdata.zi (which I added in #24698) provided by msys instead of the version shipped with the gcc source code, which is hard compiled into `libstdc++-6`. As such an update of `tzdata` will be reflected immediately, when used with `std::chrono::reload_tzdb()` even without program restart.

lazka:
Thanks. What happens if the file is missing compared to the status quo, for example if someone copies the DLL into their app without tzdata.

HazardyKnusperkeks:
In that case it still tries to open `../share/zoneinfo/tzdata.zi`. If that doesn't exist, it falls back to the still embedded version, so we are back at where we came from. But if by any chance the file does exist, but is not in the right format it will throw an exception and possibly ending the program.

In theory the author can provide `const char* __gnu_cxx::zoneinfo_dir_override()` to state a different path, which takes precedence. But I did not get that working, but didn't try really hard. Do weak symbols work on Windows?

All in all, I think it's a good thing, and the patch is basically already accepted (but not yet pushed) upstream: https://gcc.gnu.org/pipermail/libstdc++/2025-July/062432.html

On a side note, I have also another patch proposed to make `std::chrono::current_zone()` work on Windows, should I include it in this PR, or open a new one, once this is merged? I have no feedback yet, so I'd wait at least for the first comments from Jonathan.

lazka:
ok, thanks, sounds reasonable. that shouldn't cause any regressions then.

re weak, it's not really supported with gcc/binutils afaik, but clang does.

new PR is fine.

dscho:
Does this really have to be a hard dependency on `tzdata` rather than a suggested one?

lazka:
> Does this really have to be a hard dependency on `tzdata` rather than a suggested one?

Not strictly, but in reality no one would install it then. I was wondering that as well though, so I wouldn't mind changing it.

PR #24764, opened 2025-07-09
python-rencode: update to 1.0.8

striezel:
Update python-rencode to 1.0.8.

Depends on python-pytest for checks, even in 1.0.6.

### DIFF
--- a/mingw-w64-python-rencode/0001-Remove-extra-compile-args.patch
+++ b/mingw-w64-python-rencode/0001-Remove-extra-compile-args.patch
@@ -1,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: "Jan Alexander Steffens (heftig)" <heftig@archlinux.org>
+Date: Fri, 20 Jun 2025 18:48:26 +0200
+Subject: [PATCH] Remove extra compile args
+
+These are hazardous and cause unintended incompatibilities,
+`-march=native` especially so, as it's dependent on the build machine.
+
+Fixes: https://github.com/aresch/rencode/issues/38
+---
+ build.py | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/build.py b/build.py
+index 658ada290a7a..0d170a3f7c15 100644
+--- a/build.py
++++ b/build.py
+@@ -11,7 +11,7 @@ from setuptools import Extension
+ from setuptools.command.build_ext import build_ext
+ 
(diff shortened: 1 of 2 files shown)


striezel:
That did not work. Converting to draft until I figured out what's wrong here.

ognevny:
try to refer to arch recipe: https://gitlab.archlinux.org/archlinux/packaging/packages/python-rencode/-/commit/0354c675c540314a1d5cad7b1e4d8834e5d359d5

striezel:
@ognevny Thanks, that helped to get most of the builds pass.
However, clangarm64 still fails:

```
  ==> Starting build()...
  Compiling rencode/_rencode.pyx because it changed.
  [1/1] Cythonizing rencode/_rencode.pyx
  clang: error: unsupported option '-msse' for target 'aarch64-w64-windows-gnu'
  clang: error: unsupported option '-msse2' for target 'aarch64-w64-windows-gnu'
  clang: error: unsupported option '-mfma' for target 'aarch64-w64-windows-gnu'
```

Seems reasonable, because ARM does not have the SSE, SSE2 and FMA instruction set extensions, unlike x86 processors.

ognevny:
something in the source applies these flags unconditionally

striezel:
> something in the source applies these flags unconditionally

Yes, and it looks like there already is a fix for that in the Git repository: https://github.com/aresch/rencode/commit/591b9f4d85d7e2d4f4e99441475ef15366389be2 A later commit even removed those conditional compilation flags altogether: https://github.com/aresch/rencode/commit/e7ec8ea718e73a8fee7dbc007c262e1584f7f94b They are not part of a release (yet). So putting that changes in a patch file and applying it before the build should be enough to make sure the ARM build succeeds - if there are no other obstacles in the way. :D

ognevny:
backporting first commit should be enough. please remove `"-march=native"` from it, because it will apply optimizations specific to CI machine CPU

ognevny:
OR better use this patch from Arch: https://gitlab.archlinux.org/archlinux/packaging/packages/python-rencode/-/blob/main/0001-Remove-extra-compile-args.patch?ref_type=heads

striezel:
> OR better use this patch from Arch: https://gitlab.archlinux.org/archlinux/packaging/packages/python-rencode/-/blob/main/0001-Remove-extra-compile-args.patch?ref_type=heads

I've chosen this approach. And finally the builds pass for all jobs, even clangarm64.
Thank you for the guidance, @ognevny. It would have taken way longer without your input.

PR #24781, opened 2025-07-10
live-media: update to 2022.12.01

striezel:
Update live-media to 2022.12.01.

There are newer versions of live-media, but this one builds without having to got through patches or adjusting the build logic. I figured having a small update, although not to the newest version, is still better than no update at all. However, I am not sure whether this is acceptable for you. If you feel differently, then you are free to reject this pull request.


lazka:
sure, any update helps

PR #24785, opened 2025-07-10
sccache: build dist-server binary

ognevny:
I don't know whether this fixes clang issue (very likely not), but I've looked into Arch recipe and found this. so why not


ognevny:
it's actually a unixy tool

PR #24786, opened 2025-07-11
dxvk: new

felixaszx:
A DirectX3D translation layer base on Vulkan. It requires a very specific/modify version of `libdisplay-info`; the version they requires does not provide installation scripts and behaves differently than the "normal" one. Therefore I fetch it individually...

### DIFF
--- a/mingw-w64-dxvk/PKGBUILD
+++ b/mingw-w64-dxvk/PKGBUILD
@@ -15,8 +15,11 @@ makedepends=(git
              "${MINGW_PACKAGE_PREFIX}-cc"
              "${MINGW_PACKAGE_PREFIX}-meson"
              "${MINGW_PACKAGE_PREFIX}-ninja"
-             "${MINGW_PACKAGE_PREFIX}-vulkan-devel"
+             "${MINGW_PACKAGE_PREFIX}-spirv-headers"
+             "${MINGW_PACKAGE_PREFIX}-vulkan-headers"
 )
+optdepends=("${MINGW_PACKAGE_PREFIX}-vulkan-loader: use MSYS2 Vulkan loader when possible")
+
 _libdisplay_info_name=libdisplay-info
 _libdisplay_info_commit=275e6459c7ab1ddd4b125f28d0440716e4888078
 source=("https://github.com/doitsujin/dxvk/archive/refs/tags/v$pkgver.tar.gz"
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
This package doesn't make any sense on Windows, on Linux this package helps wine and alike to run DirectX games and programs through this dxvk, but here on Windows DirectX could be used directly.

ognevny:
some games have better performance when using dxvk libs injected into a core game files

MehdiChinoune:
> some games have better performance when using dxvk libs injected into a core game files

Which? any benchmarks?. Do they support the same features. I read once about some games having more fps using dxvk on Linux than Windows native DirectX but bacause dxvk doesn't support some Graphics features.
Anyway, So shipping Windows native games/programs with DXVK doesn't make sense.

ognevny:
I used to play Train Simulator with dxvk dll being installed in the root directory of the game. I didn't do any benchmarks but performance became better. this is probably because TS is old game which doesn't support modern PCs well (they still have an _experimental_ dx12 support and it's bad IIRC)

felixaszx:
> > some games have better performance when using dxvk libs injected into a core game files
> 
> Which? any benchmarks?. Do they support the same features. I read once about some games having more fps using dxvk on Linux than Windows native DirectX but bacause dxvk doesn't support some Graphics features. Anyway, So shipping Windows native games/programs with DXVK doesn't make sense.

https://www.gamingonlinux.com/2022/12/intel-using-dxvk-part-of-steam-proton-for-their-windows-arc-gpu-dx-9-drivers/
In fact Intel even replaced their dx9 driver with dxvk...
This is not a must-have, but certainly is a good-to-have for lots of people.

MehdiChinoune:
I think Intel did that to not waste effort/time/money developing a driver for their latest GPUs for an old Graphics API. As they focus more on DirectX 12+.

felixaszx:
So does the same logic apply on other drivers. 

dxvk is nice-to-have because from now on, hardware providers will surely prioritize later generation APIs development. Although I have no clear evidents that dxvk provides significant improvements over native D3D9/10 APIs, it is widely reported from both Linux/Windows communities that dxvk improves older softwares.

It is probably unnecessary to have dxvk as a development target but as a general-purpose software distribution platform, there is nothing wrong to distribute it.

________________________________

PR #24787, opened 2025-07-11
 python-aiohttp: update to 3.12.14

striezel:
Update python-aiohttp to 3.12.14.


striezel:
Looks like this update also needs [mingw-w64-python-aiosignal](https://packages.msys2.org/base/mingw-w64-python-aiosignal) to be updated. I'll prepare a pull request for that, too.

PR #24842, opened 2025-07-15
ossia-score: Enable aarch64 builds

hmartinez82:
image


lazka:
@jcelerier fyi

jcelerier:
yay!

jcelerier:
@hmartinez82 do you have the build error that led to having to include windows.h ?

jcelerier:
(I'd like to avoid including windows.h here because this file may end up being included in a ton of other files and windows.h bloats compile time like crazy)

hmartinez82:
@jcelerier Yes, `YieldProcessor()` undefined.
I could've found which files include `audio_spin_mutex.hpp` and only add `Windows.h` there, but then I would take all the fun from you ;-)

jcelerier:
thanks! indeed, that one comes straight from winnt.h.. strange that it worked before that

jcelerier:
also @hmartinez82 it looks like there's a lot of graphical glitches on the screenshot? those definitely don't happen with non-arm64 builds (or on arm64 builds on other platforms, either linux or macOS)

jcelerier:
@hmartinez82 current head commit of score has the patch applied :)

jcelerier:
(but i'm still worried about these artifacts on your screenshot)

hmartinez82:
@jcelerier it's just the snipping tool compression, (or capturing), sometimes it does that, sorry for the alarm. Here's another one from the same machine.

image

jcelerier:
ah thanks for the reassurance!

PR #24854, opened 2025-07-16
dbus: update 1.16.2

felixaszx:
For some reasons I can not build ghostscript and efl on my machine, probably some weird environment issues. Try it on MSYS2 server instead


felixaszx:
Some quick explaination of the changes:

`dbus`: `autotools` are no longer avaibale from the upstream. In theory `meson` should be prefered but it can not properly detect `SF_UNIX` for Winsocks2. Probably a bug.

`efl`: The upstream has not yet been ready for GCC15.

felixaszx:
Looks like C23 breaks lots of legacy stuffs...

lazka:
I see you forced c17 and then reverted it, why didn't that work?

felixaszx:
> I see you forced c17 and then reverted it, why didn't that work?

Some c++ codes will not be compiled...
I can try it later again

felixaszx:
> I see you forced c17 and then reverted it, why didn't that work?

No, that does not work, I have tried c99,c11,c17 + c++11,c++14,c++17,c++20. Their codes are really non-standard.
I have raised an issue to the upstream, not sure how they will act on it.

MehdiChinoune:
> No, that does not work, I have tried c99,c11,c17 + c++11,c++14,c++17,c++20. Their codes are really non-standard. I have raised an issue to the upstream, not sure how they will act on it.

Did you try `gnu99`, `gnu11` + `gnu++11`,`gnu++11`. Their code could be non-standard but It should be built with gcc.

felixaszx:
Make it gnu17/gnu++17 works, the upstream has also changed the code to make it work with GCC 15. But their changes will be applied in next release I suppose.

PR #24855, opened 2025-07-16
rust: change some config

ognevny:
- use fat lto as a way to get a bit more optimizations
- integrate some config from #22778


ognevny:
tested UCRT64. performance is almost the same, but binaries' size are reduced a lot!

striezel:
Just out of curiosity: Why is `#lto = "fat"` in bootstrap.toml commented out and uncommented later via `sed` command? Wouldn't it be easier / simpler to just remove the '#'  in bootstrap.toml in the first place? :thinking:

ognevny:
look closer into a diff. it's actually commented under certain condition (non-i686 environment). this is done because LLVM codegen in 32-bit compiler can't handle fat-LTO for building Rust due to memory restrictions of this architecture...

striezel:
Ah. Thanks for the explanation.

PR #24860, opened 2025-07-16
zopfli: add msys2_references + fix CMake build error

striezel:


striezel:
Looks like zopfli hasn't been updated to work with current CMake versions:

```
   ==> Starting build()...
  CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
    Compatibility with CMake < 3.5 has been removed from CMake.
  
    Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
    to tell CMake that the project requires at least <min> but has been updated
    to work with policies introduced by <max> or earlier.
  
    Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
  
  
      Aborting...
```

striezel:
Builds again, although with a deprecation notice:
```
  ==> Starting build()...
  CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
    Compatibility with CMake < 3.10 will be removed from a future version of
    CMake.
```
So this probably will not work anymore when CMake 5.0 is released (whenever that will be), but until then this should be fine.

PR #24861, opened 2025-07-16
zlib-ng: add msys2_references

striezel:
Not sure whether this works, because there are two ArchLinux packages, zlib-ng and zlib-ng-compat, which are part of the same package in MSYS2.


ognevny:
pkgbase is zlib-ng, so you should leave zlib-ng only. and IIUC multiple reference of the same repo are not handled as you intended

striezel:
> and IIUC multiple reference of the same repo are not handled as you intended

Would have been to easy that way, would it? :D
OK, I've removed the reference to zlib-ng-compat.

PR #24862, opened 2025-07-16
z3: add msys2_references + disable strip option for post-build check

striezel:


striezel:
Package builds succeed, but the final check fails:
```

Run python .ci/ci-check.py "D:\a\_temp/artifacts"
  python .ci/ci-check.py "D:\a\_temp/artifacts"
  shell: D:\a\_temp\setup-msys2\msys2.CMD {0}
  env:
    MSYSTEM: UCRT64

Create temporary Repository:
  ==> Adding package '/d/a/_temp/artifacts/mingw-w64-ucrt-x86_64-z3-4.14.1-1-any.pkg.tar.zst'
    -> Computing checksums...
    -> Creating 'desc' db entry...
    -> Creating 'files' db entry...
  ==> Creating updated database file '/d/a/_temp/artifacts/ci.db.tar.gz'
Error: Package `mingw-w64-ucrt-x86_64-z3` is using `python-installer` and has executables but strip option is not disabled in the PKGBUILD.
```
So why did it not fail in earlier builds? Was the check step changed recently?

lazka:
> So why did it not fail in earlier builds? Was the check step changed recently?

it was directly pushed/updated since the check was added I guess.

PR #24863, opened 2025-07-16
shards: enable clangarm64 builds

HertzDevil:


ognevny:
no need to bump pkgrel. autobuild will manage a separate build for aarch64 without rebuilding the package for other environments

PR #24877, opened 2025-07-17
vigra: add msys2_references

striezel:


striezel:
clang64 build failed due to network problems:
```
  :: Retrieving packages...
   curl-8.15.0-1-x86_64 downloading...
   libcurl-8.15.0-1-x86_64 downloading...
  error: failed retrieving file 'curl-8.15.0-1-x86_64.pkg.tar.zst.sig' from distrohub.kyiv.ua : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds
  warning: failed to retrieve some files
  error: failed to commit transaction (download library error)
  Error: The process 'C:\Windows\system32\cmd.exe' failed with exit code 1
```
Somebody should probably restart that job, and hopefully it will pick another mirror then.

PR #24884, opened 2025-07-18
llvm: enable pgo for non-i686 GCC environments

### LINKED ISSUE #24881: [suggestion] optimize LLVM with PGO for GCC environments
currently PGO is enabled for clang environments only. is there a chance of having it for other environments too? when I tried #22778 I found that LLVM pipeline is not required for CLANG* at all, because LLVM built with opt-dist has even worse performance (this is also shown in https://github.com/rust-lang/rust/pull/143898#issuecomment-3069916409). I want to reduce maintanance costs of having separate 3 pipelines with different timings (32-bit - no PGO; GCC - PGO both rustc and LLVM because current library can't be used for a proper profiling; Clang - PGO rustc only) in case of (finally) I got successful builds in #22778

update: previous builds of #22778 have shown that PGO pipeline is possible for GCC envs, but lld and compiler-rt are required to be used

ognevny:
locally built successfully. I'll try with #22778 later and update this comment. enabling PGO for GCC envs requires _clangprefix'ed dependencies to be used + lld for a build.

fixes #24881

update: this doesn't work in my PR due to lld crash...
update2: hello world compiles with gcc+lld and clang+lld

### DIFF
--- a/mingw-w64-llvm/PKGBUILD
+++ b/mingw-w64-llvm/PKGBUILD
@@ -13,6 +13,9 @@ if [[ $MINGW_PACKAGE_PREFIX == *-clang-* ]]; then
   _pgo=1
 else
   _compiler=clang
+  if [[ ${CARCH} != i686 ]]; then
+    _pgo=1
+  fi
 fi
 
 _realname=llvm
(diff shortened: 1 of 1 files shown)


ognevny:
anyone feel free to change this PR or create a new one

mati865:
On CI it fails with:
```
/d/_/B/PKGBUILD: line 244: /ucrt64/bin/llvm-profdata: No such file or directory
```

ognevny:
> On CI it fails with:
> 
> ```
> /d/_/B/PKGBUILD: line 244: /ucrt64/bin/llvm-profdata: No such file or directory
> ```
ahh, typo. llvm-tools should be installed

ognevny:
timeout...

PR #24889, opened 2025-07-19
openimageio: update to 3.0.8.1

striezel:


striezel:
The patch for commit https://github.com/AcademySoftwareFoundation/OpenImageIO/commit/09250af27d11f6ea761872490403d074424b6e62 is no longer required in 3.0.8.**1**, because it is part of that release, see commit https://github.com/AcademySoftwareFoundation/OpenImageIO/commit/83a90bd05023acfaa2498999e98bb5fe467b8f20.

